PR #116, opened 2023-03-20
Actions

qc00:
Addresses #5


build_tooling/parallel_test.sh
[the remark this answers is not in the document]
qc00: It runs two pytests concurrently in one runner to make use of the two CPU.

_Surprisingly_, this cuts the test runtime by rougly half and the `sed` command inside adds a worker id in front of every log line, so it's not confusing.

It's written to be environment-agnostic and as a drop-in replacemnt for the `pytest` command, so can run locally as well. (Trivial change required to make it run more than two workers)

cpp/CMake/CheckSupportsPreset.cmake
```diff
@@ -0,0 +1 @@
+cmake_minimum_required(VERSION 3.21 FATAL_ERROR)
```
poodlewars, line 1: Why don't we just update the root `CMakeLists.txt` with this and remove this file?

cpp/CMakeLists.txt
```diff
@@ -40,27 +32,42 @@ if(NOT CMAKE_BUILD_TYPE)
...
-    set(ENV_BASE_DIR "/default-medusa-venv")
-
-    set(BUILD_PYTHON_VERSION 3.6)
-
-    set(PYTHON_LIBRARIES "${ENV_BASE_DIR}/lib")
-    set(PYTHON_EXECUTABLE "${ENV_BASE_DIR}/bin/python")
-    set(PYTHON_INCLUDE_DIRS "${ENV_BASE_DIR}/include/python${BUILD_PYTHON_VERSION}m")
-    set(PYTHON_LIBRARY_SO "${ENV_BASE_DIR}/lib/libpython${BUILD_PYTHON_VERSION}m.so")
-endif()
+include(PythonUtils) # Must be called before Pybind (third_party) to override its finding mechanism
```
poodlewars, line 62: I'm a bit lost here - how does this override the finding mechanism?

cpp/CMakePresets.json
```diff
@@ -0,0 +1,78 @@
...
+      },
+      "toolset": {
+        "value": "host=x64",
+        "strategy": "external"
+      },
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_C_COMPILER": "cl",
+        "CMAKE_CXX_COMPILER": "cl",
+        "VCPKG_TARGET_TRIPLET": "x64-windows-static-md"
```
poodlewars, line 30: what does `-md` mean? Can we improve the naming of the triplet?

cpp/CMakePresets.json
```diff
@@ -0,0 +1,78 @@
...
+    {
+      "name": "linux-release",
+      "inherits": "linux-debug",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release"
+      }
+    }
+  ],
+  "buildPresets": [
+    {"name": "windows-cl-debug",   "configurePreset": "windows-cl-debug",   "targets": "arcticdb_ext" },
```
poodlewars, line 73: did you consider putting some default for `jobs` in here?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -3,52 +3,73 @@ option(SSL_LINK "Link against SSL libraries." ON)
...
 include(GoogleTest)
 include(GenerateExportHeader)
 
-find_package(Threads REQUIRED)
-find_package(Zstd REQUIRED)
-find_package(LZ4 REQUIRED)
+# System and third-party libraries with built-in find_package Config
+cmake_policy(PUSH)
+    if (EXISTS "/usr/local/lib64/aws-c-cal/cmake/modules/FindLibCrypto.cmake") # Workaround old AWS SDK bug
+        cmake_policy(SET CMP0045 OLD)
```
poodlewars, line 16: The PUSH and POP is quite fragile considering we may not necessarily set a policy at all?
Why do we need this, we'll get the AWS lib from VCPKG each time rather than relying on a system install? Does the VCPKG version still have this bug?

Oh - this is for compat with the internal `man.arcticdb` build?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -3,52 +3,73 @@ option(SSL_LINK "Link against SSL libraries." ON)
...
 find_package(prometheus-cpp CONFIG REQUIRED)
+find_package(Threads REQUIRED)
 
-if (WIN32)
-    find_package(Fizz REQUIRED)
-    find_package(protobuf REQUIRED)
-    find_package(wangle REQUIRED)
-    find_package(Zstd REQUIRED)
+# Libraries compiled by vcpkg and also baked into our internal image:
+if (ARCTICDB_USING_VCPKG)
```
poodlewars, line 30: We can just assume we are always using VCPKG, no?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -3,52 +3,73 @@ option(SSL_LINK "Link against SSL libraries." ON)
...
     find_package(mongocxx REQUIRED)
     find_package(bsoncxx REQUIRED)
-    find_package(aws-cpp-sdk-s3 CONFIG REQUIRED)
-    find_package(RdKafka REQUIRED)
-    find_package(unofficial-sodium REQUIRED)
-    find_package(unofficial-pcre REQUIRED)
-else ()
+    set(ARCTICDB_MONGO_LIBS mongo::mongocxx_static mongo::bsoncxx_static)
+
+    find_package(zstd CONFIG REQUIRED)  # "CONFIG" bypass our cpp/CMake/FindZstd.cmake module
```
poodlewars, line 42: s/bypass/bypasses

cpp/arcticdb/CMakeLists.txt
```diff
@@ -3,52 +3,73 @@ option(SSL_LINK "Link against SSL libraries." ON)
...
+    set(ARCTICDB_MONGO_LIBS ${LIBMONGOCXX_STATIC_LIBRARIES_NO_SASL} ${LIBBSONCXX_STATIC_LIBRARIES})
+
+    get_target_property(NEW_FOLLY_DEPS Folly::folly_deps INTERFACE_LINK_LIBRARIES)
+    list(TRANSFORM NEW_FOLLY_DEPS REPLACE "/usr/local/lib/libsodium\\.so" "libsodium.a")
+    set_target_properties(Folly::folly_deps PROPERTIES INTERFACE_LINK_LIBRARIES "${NEW_FOLLY_DEPS}")
+endif()
+
+# Libraries baked into the internal and external Linux images are found by searching various paths
+# On Windows, vcpkg will provide a "Config" which takes precedence
+find_library(Pcre_LIBRARY NAMES pcre PATHS /usr/local/lib/libpcre.a REQUIRED)
```
poodlewars, line 63: why not just use vcpkg?

cpp/arcticdb/CMakeLists.txt
poodlewars, line 1: Brilliant clean up of this

setup.cfg
```diff
@@ -35,22 +36,16 @@ install_requires =
...
     boto3
-    arctic
-    kafka-python
-    kazoo
-    hypothesis
     moto
-
-[tool:pytest]
-cpp_files=cpp/tests/*
+    flask  # Used by moto
```
poodlewars, line 49: (nit) I think `moto[server]` should pick up the flask stuff and leave it a bit more obvious why we need it at all.

setup.py
```diff
@@ -97,63 +70,79 @@ def run(self):
...
-            os.symlink(extension_file, so_file)
-
-    def compile(self, build_dir):
+        dest = os.path.abspath(self.get_ext_fullpath(ext.name))
+        print(f"Destination: {dest}")
+        install_args = [f"-DCMAKE_INSTALL_PREFIX={os.path.dirname(dest)}"]
+
+        cmake = shutil.which("cmake")  # Windows safe
+        if not cmake and platform.system() != "Windows":
+            print("Did not find cmake on the path. Will try to resolve from shell", file=sys.stderr)
```
poodlewars, line 79: Probably safer to blow up at this point - clever defaulting never ends well

cpp/CMake/CheckSupportsPreset.cmake
poodlewars: For compat with our internal Docker image?

setup.py
```diff
@@ -97,63 +70,79 @@ def run(self):
...
+        if preset == "*":
+            suffix = "-debug" if self.debug else "-release"
+            preset = ("windows-cl" if platform.system() == "Windows" else platform.system().lower()) + suffix
+        print(
+            f"Did not find build directory with '{search}'. Will configure and build using cmake preset {preset}",
+            file=sys.stderr,
+        )
+        subprocess.check_call([cmake, *common_args, "--preset", preset], cwd="cpp")
+
+    def _configure_for_legacy_image(self, common_args, build_dir):
```
poodlewars, line 119: Probably cleaner to just update the cmake in the internal Docker image than support legacy CMAKE?

.github/workflows/build_steps.yml
```diff
@@ -0,0 +1,176 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.3.0
+        with:
+          submodules: recursive # Just in case a dep has its own third-party deps
+
+      - name: Deep checkout vcpkg
+        run: cd cpp/vcpkg ; git fetch --unshallow origin master
+
+      - name: Configure sccache
```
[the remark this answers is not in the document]
qc00: Can use it locally by pointing the `CMAKE_*_COMPILER_LAUNCHER` to it.

It's also a remote/distributed compiler luncher at the same time and I am hoping of setting up at a later date:
* A shared cache interally (cannot easily reuse the GitHub one)
* A set of distributed worker nodes (instead of everyone trying to get larger headnodes)

.github/workflows/build_steps.yml
```diff
@@ -0,0 +1,176 @@
...
+        if: inputs.job_type == 'cpp-tests'
+        run: cd cpp; cmake --build --preset $ARCTIC_CMAKE_PRESET --target install
+
+      - name: C++ Rapidcheck
+        if: inputs.job_type == 'cpp-tests'
+        run: cpp/out/install/arcticdb_rapidcheck_tests
+
+      - name: C++ unit tests
+        if: inputs.job_type == 'cpp-tests'
+        run: "cpp/out/install/test_unit_arcticdb --gtest_filter=-TestNfsBackedStorage.*:TestS3Storage.*"
```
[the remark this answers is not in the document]
qc00: There's no moto in C++ so S3 tests can't run. Raise #127

.github/workflows/build_steps.yml
```diff
@@ -0,0 +1,176 @@
...
+      fail-fast: false
+      matrix:
+        os: [linux]
+        type: ${{fromJSON(needs.compile.outputs.test_dirs)}}
+        include:
+          - os: linux
+            distro: ubuntu-latest
+            cibw_format: "cp{0}-manylinux_x86_64"
+          - type: hypothesis
+            cmd: '{hypothesis,nonreg}' # There are so few nonreg tests, run them in the hypothesis runner
```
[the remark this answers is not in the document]
qc00: Just to save some overhead. It takes ~3 mins to get a runner to the point to be able to run a test and the nonreg test takes 10s.

build_tooling/avoid_oom.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# Avoids compiling $file with other parallel compiler executions
+
+file=expression_node.cpp
```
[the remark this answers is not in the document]
qc00: Raise #128

build_tooling/parallel_test.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+set -e
+
+tooling_dir="$(dirname $BASH_SOURCE)"
+echo Saving results to ${test_output:="$(realpath "$tooling_dir/../cpp/out")"}
```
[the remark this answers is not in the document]
qc00: That's only the default directory.

In a previous incarnation of the container, the `out` folder is the only one that's archived. This variable is now set by the caller.

build_tooling/parallel_test.sh
```diff
@@ -0,0 +1,40 @@
...
+    # Each test also get a separate directory since there's a mystery lock somewhere preventing concurrent runs (even)
+    # from different Python processes
+    mkdir $new_root # Will fail the function if it exists for some reason
+    ln -s "$(realpath "$tooling_dir/../python/tests")" $new_root
+    cd $new_root
+
+    python -m pytest -v --show-capture=no --log-file="$test_output/pytest-logger.$group.log" \
+        --junitxml="$test_output/pytest.$group.xml" \
+        --splits 2 --group $group --durations-path="$duration_file" --store-durations \
+        "$@" 2>&1 | sed -r "s#^(tests/.*/([^/]+\.py))?#$group: \2#"
```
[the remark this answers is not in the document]
qc00: It does two things:

* Adds a `1: ` or `2: ` indicator at the beginning of every log line so we know which concurrent runner logged it
* Strips `random/long/path/in/the/build/container/to/python/tests/unit/arcticdb/some_test.py` to just `some_test.py`. Note each runner only (mostly) run one kind of test, so the file name is not ambiguous.

cpp/CMake/PythonUtils.cmake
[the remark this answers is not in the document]
qc00: I really wish I wrote these checks at the beginning, would have saved me so much time....

Took me two weeks to realise `Python_EXECUTABLE` and `PYTHON_EXECUTABLE` are different variables used by different packages in CMake for example. Our scripts was basically overwriting half of the detection of pybind11 but not correctly configuring FindPython at all, causing all kinds of subtle breakage.

cpp/CMake/PythonUtils.cmake
```diff
@@ -0,0 +1,145 @@
...
+# Enhanced FindPython
+if(DEFINED ARCTICDB_FIND_PYTHON_DEV_MODE)
+    message("Using supplied ARCTICDB_FIND_PYTHON_DEV_MODE=${ARCTICDB_FIND_PYTHON_DEV_MODE}.")
+    if("${ARCTICDB_FIND_PYTHON_DEV_MODE}" STREQUAL "pybind11")
+        set(ARCTICDB_FIND_PYTHON_DEV_MODE "")
+    endif()
+elseif(WIN32)
+    # Static linking is not necessary on Windows and Development.Module needs the runtime lib anyway:
+    set(ARCTICDB_FIND_PYTHON_DEV_MODE Development)
+elseif(${CMAKE_VERSION} VERSION_LESS 3.15)
```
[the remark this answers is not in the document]
qc00: This is to support the internal image (and any unfortuante external people).

The `CMake/CheckSupportsPreset.cmake` is only called from `setup.py` to check the version.

.github/workflows/build_steps.yml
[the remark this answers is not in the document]
poodlewars: Ah fair enough. Thought it was more to reuse this bit of the build script.

cpp/arcticdb/CMakeLists.txt
qc00: Yes. Newer AWS SDK seems to be fine.

poodlewars: Why don't we leave this out then? We won't need `man.arcticdb` once this is in.

cpp/arcticdb/storage/mongo/mongo_client.cpp
```diff
@@ -19,17 +19,27 @@
...
+#include <mongocxx/config/version.hpp>
 #include <arcticdb/util/composite.hpp>
 
 namespace arcticdb::storage::mongo {
 
 namespace detail {
 
+template<typename ElementType>
+std::string get_string_element(const ElementType& element) {
+#if (MONGOCXX_VERSION_MAJOR * 1000 + MONGOCXX_VERSION_MINOR) >= 3007
```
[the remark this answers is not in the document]
poodlewars: Why? Once this is merged everything can use vcpkg.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -548,17 +530,25 @@ target_include_directories(arcticdb_ext
...
+install(TARGETS arcticdb_ext
+        LIBRARY DESTINATION . COMPONENT Python_Lib
+        RUNTIME DESTINATION . COMPONENT Python_Lib)
+
 ## Unit Tests ##
-if(TEST)
+if(${TEST})
+    message("Python for test compilation:")
+    unset(Python_USE_STATIC_LIBS)
+    find_package(Python 3 COMPONENTS Interpreter Development REQUIRED)
```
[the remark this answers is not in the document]
qc00: This will require a dedicated academic paper to explain, but the criticla points are:

* the `Development` component can't be found for main code compilation if the only way to satisfy it is a statically linked `libpython.a`
* the `Development` component must be present to link the C++ tests and it can be either static or dynamic at this point
* CMake 3.18+ tries to solve this problem by splitting `Development` into `Module` and `Embed`, but can't assume we have that version

cpp/CMakeLists.txt
```diff
@@ -40,27 +32,42 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release)
 endif()
 
+if(${CMAKE_TOOLCHAIN_FILE} MATCHES "vcpkg.cmake$")
+    message(STATUS "CMAKE_TOOLCHAIN_FILE is set to vcpkg, so will use find_package to find the libraries")
+    set(ARCTICDB_USING_VCPKG ON)
+endif()
+
+if(NOT DEFINED ENV{CMAKE_BUILD_PARALLEL_LEVEL})
```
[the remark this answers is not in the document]
qc00: The problem is I don't have an easy access to set the number in the pegasus build. We'll either have to do cgroup stuff in Python or use CMake and I prefer CMake.

Also gives people who runs CMake directly from IDEs a nice warning to set the concurrency.

cpp/CMakeLists.txt
qc00: By setting variables before pybind11 is included.

cpp/CMakePresets.json
qc00: There was, but `setup.py` ended up not using the build presets so the only thing `jobs` would affect are the IDEs, which should use a machine-based default.

setup.py
qc00: Not wanting to spend another 3 weeks fixing compilation dependency issues :P

cpp/arcticdb/CMakeLists.txt
qc00: It still requires some effort to get vcpkg to download reliably and similarly for a new dev image.

If it only takes a few lines to keep the old builds working, I would rather do that.

PR #129, opened 2023-03-20
Add specific error message when range searching on unsorted data

willdealtry:
At the moment if you try to do a range search on unsorted data it throws an InternalException, which is not ideal as those are intended for assertion failures and really should not be used for anything that users might see (even under less-than-ideal circumstances). Throw a specific exception with the correct error code.

### DIFF
--- a/cpp/arcticdb/util/error_code.hpp
+++ b/cpp/arcticdb/util/error_code.hpp
@@ -28,7 +28,8 @@ enum class ErrorCategory : BaseType {
     NORMALIZATION = 2,
     MISSING_DATA = 3,
     SCHEMA = 4,
-    STORAGE = 5
+    STORAGE = 5,
+    SORTING = 6
 };
 
 // FUTURE(GCC9): use magic_enum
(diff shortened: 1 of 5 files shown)


mehertz:
Can we update all the third party modules in a seperate PR please.

cpp/arcticdb/util/error_code.hpp
```diff
@@ -57,7 +59,8 @@ inline std::unordered_map<ErrorCategory, const char*> get_error_category_names()
     ERROR_CODE(3000, E_NO_SUCH_VERSION)  \
     ERROR_CODE(4000, E_DESCRIPTOR_MISMATCH)  \
     ERROR_CODE(5000, E_KEY_NOT_FOUND) \
-    ERROR_CODE(5001, E_DUPLICATE_KEY)
+    ERROR_CODE(5001, E_DUPLICATE_KEY) \
```
mehertz, line 62: Update the docs please!

cpp/arcticdb/util/error_code.hpp
```diff
@@ -57,7 +59,8 @@ inline std::unordered_map<ErrorCategory, const char*> get_error_category_names()
...
+    ERROR_CODE(6000, E_UNSORTED_DATA)
```
qc00, line 63: Please leave a `\` at the end so we don't end up touching the previous line each time we add one

PR #140, opened 2023-03-22
Small tweaks to get Python tests running on Windows

poodlewars:
There are still a few failures that I'll look at next, but this gets us as far as having most of the tests pass. It also lets us build the extension with a Python of our choice easily, and debug the extension from Python tests.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -14,8 +14,8 @@ if(WIN32)
 
     # Guide to MSVC compilation warnings https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c4000-through-c4199?view=msvc-170
     set(CMAKE_CXX_FLAGS "/DWIN32 /D_WINDOWS /w /GR /bigobj /EHsc /w /wd4244 /wd4267")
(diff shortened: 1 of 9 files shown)


cpp/CMakeLists.txt
```diff
@@ -59,14 +59,18 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-include(PythonUtils) # Must be called before Pybind (third_party) to override its finding mechanism
+if(NOT WIN32)
```
qc00, line 62: Is this still necessary given what we've learned the real cause was?

python/tests/conftest.py
```diff
@@ -42,20 +48,16 @@
 BUCKET_ID = 0
 
 
+def run_server(port):
```
qc00, line 51: Can we make it clear what kind of server this is?

cpp/CMakeLists.txt
poodlewars: Yes, it had issues discovering the Python static lib with this left in.

python/tests/conftest.py
poodlewars: That is fair - I'll make it private too.

PR #148, opened 2023-03-24
Do not rely on hashing of different types for IsIn operations

poodlewars:
Which has some bugs.

Instead cast the valueset and the data column to the same type before doing the inclusion test.


cpp/arcticdb/entity/types.hpp
```diff
@@ -476,17 +476,7 @@ struct FieldDescriptor {
         data_.mutable_type_desc()->CopyFrom(static_cast<arcticdb::proto::descriptors::TypeDescriptor>(td));
         data_.set_name(n);
     }
-/*
```
qc00, line 479: Gem will reinstate these soon and no other lines in this file need changing.

python/arcticdb/util/hypothesis.py
```diff
@@ -65,6 +65,12 @@ def integral_type_strategies(draw):
     return draw(from_dtype(draw(st.one_of([unsigned_integer_dtypes(), integer_dtypes()]))).filter(non_infinite))
 
 
+@st.composite
+def dataframes_with_names_and_dtypes(draw, names, dtype_strategy):
```
poodlewars, line 69: This is to help with the flaky numeric isin tests. With latest pandas, hypothesis sometimes provides a df containing floats even when using `elements=integral_type_strategies()`. It is safer to explicitly declare the types and leave the elements to be inferred, rather than the other way around.

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -120,7 +120,7 @@ struct unary_arithmetic_promoted_type {
 };
 
 template <class LHS, class RHS, class Func>
-struct binary_arithmetic_promoted_type {
+struct type_arithmetic_promoted_type {
```
poodlewars, line 123: Renaming since it is now used for more than just _binary_ arithmetic.

alexowens90: IsIn/IsNotIn are still binary operations. The "arithmetic" part of the name just referred to the fact that it was covering addition/subtraction/multiplication/division. binary_promoted_type is probably more accurate

alexowens90 (approved):

cpp/arcticdb/processing/operation_types.hpp
poodlewars: IMO - they never were binary operations. The others (roughly) are mappings

f: N x N -> N

whereas IsIn maps,

f: N x [N...] -> N

which is not what binary operation typically means.

alexowens90: I view it as one of the operands being a set, not a hill I'm willing to die on though

PR #180, opened 2023-03-31
Manual segment defragmentation

phoebusm:
Add support of manual segment defragmentation
Tests, APIs and docstrings are included along with the implemenation

### DIFF
--- a/cpp/arcticdb/pipeline/frame_utils.cpp
+++ b/cpp/arcticdb/pipeline/frame_utils.cpp
@@ -170,40 +170,22 @@ void adjust_slice_rowcounts(const std::shared_ptr<pipelines::PipelineContext>& p
 
 size_t adjust_slice_rowcounts(std::vector<pipelines::SliceAndKey> & slice_and_keys) {
     using namespace arcticdb::pipelines;
-    auto current_col = slice_and_keys[0].slice_.col_range;
-    std::queue<std::pair<size_t, size_t>> columns_first_offset;
-    columns_first_offset.emplace(slice_and_keys[0].slice_.col_range.second, 0);
-    size_t offset = 0u;
-    for (auto& slice_and_key : slice_and_keys) {
-        if (slice_and_key.slice_.col_range != current_col) {
-            //for handling column-sliced repeated compaction
-            //start_col end_col start_row   end_row
-            //1         4       0           2
-            //1         3       2           3
-            //3         4       2           3
-            if (slice_and_key.slice_.col_range.first != current_col.first) {
-                while (true){
-                    if (columns_first_offset.empty()){ //Non-continuous column in segments, after filtering (test_filter_column_slicing_different_segments)
-                        offset = 0u;
-                        break;
-                    }
-                    else if (slice_and_key.slice_.col_range.first == columns_first_offset.front().first) {
-                        offset = columns_first_offset.front().second;
-                        columns_first_offset.pop();
-                        break;
-                    }
-                    columns_first_offset.pop();
-                }
-            }
-            columns_first_offset.emplace(slice_and_key.slice_.col_range.second, offset);
-            current_col = slice_and_key.slice_.col_range;
-        }
-        size_t rows = slice_and_key.slice_.row_range.diff();
-        RowRange updated{offset, offset + rows};
-        offset += rows;
-        slice_and_key.slice_.row_range = updated;
-    }
-    return offset;
+    if(slice_and_keys.empty())
+		return 0u;
+	
+	auto offset = slice_and_keys[0].slice_.row_range.first;
+	auto diff = slice_and_keys[0].slice_.row_range.diff();
+    auto col_begin = slice_and_keys[0].slice_.col_range.first;
+	
+	for(auto it = slice_and_keys.begin(); it != slice_and_keys.end(); ++it) {
+		if(it != slice_and_keys.begin() && it->slice_.col_range.first == col_begin) {
+			offset += diff;
+			diff = it->slice_.row_range.diff();
+		}
+		it->slice_.row_range = RowRange{offset, offset + diff};
+	}
+	
+	return offset + diff;
 }
 
 size_t get_slice_rowcounts(std::vector<pipelines::SliceAndKey> & slice_and_keys) {
--- a/cpp/arcticdb/stream/schema.hpp
+++ b/cpp/arcticdb/stream/schema.hpp
@@ -27,8 +27,11 @@ class FixedSchema {
         desc_(std::move(desc)),
         index_(index){}
 
-    static FixedSchema default_schema() {
-        return FixedSchema(StreamDescriptor(), stream::TimeseriesIndex::default_index());
+    static FixedSchema default_schema(const Index& index) {
+        return util::variant_match(index, [] (auto idx) {
+            using IndexType = std::remove_reference_t<decltype(idx)>;
+            return FixedSchema(StreamDescriptor(), IndexType::default_index());
+        });
     }
 
     void check(std::size_t pos, TypeDescriptor td) const {
@@ -79,8 +82,11 @@ public:
         desc_(default_dynamic_descriptor(desc, index)),
         index_(index) { }
 
-    static DynamicSchema default_schema() {
-        return DynamicSchema(StreamDescriptor(), stream::TimeseriesIndex::default_index());
+    static DynamicSchema default_schema(const Index& index) {
+        return util::variant_match(index, [] (auto idx) {
+            using IndexType = std::remove_reference_t<decltype(idx)>;
+            return DynamicSchema(StreamDescriptor(), IndexType::default_index());
+        });
     }
 
     void check(std::size_t pos ARCTICDB_UNUSED, TypeDescriptor td ARCTICDB_UNUSED) const {
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -716,7 +716,9 @@ void copy_segments_to_frame(const std::shared_ptr<Store>& store, const std::shar
 SegmentInMemory prepare_output_frame(std::vector<SliceAndKey>&& items, const std::shared_ptr<PipelineContext>& pipeline_context, const std::shared_ptr<Store>& store, const ReadOptions& read_options) {
     pipeline_context->clear_vectors();
     pipeline_context->slice_and_keys_ = std::move(items);
-    std::sort(std::begin(pipeline_context->slice_and_keys_), std::end(pipeline_context->slice_and_keys_));
+	std::sort(std::begin(pipeline_context->slice_and_keys_), std::end(pipeline_context->slice_and_keys_), [] (const auto& left, const auto& right) {
+		return std::tie(left.slice_.row_range, left.slice_.col_range) < std::tie(right.slice_.row_range, right.slice_.col_range);
+	});
     adjust_slice_rowcounts(pipeline_context);
     const auto dynamic_schema = opt_false(read_options.dynamic_schema_);
     mark_index_slices(pipeline_context, dynamic_schema, pipeline_context->bucketize_dynamic_);
@@ -1038,7 +1040,7 @@ VersionedItem sort_merge_impl(
             aggregator.commit();
         },
         [&](const auto &) {
-            util::raise_rte("Not supported index type for sort merge implementation");
+            util::raise_rte("Sort merge only supports datetime indexed data. You data does not have a datetime index.");
         }
         );
 
@@ -1093,8 +1095,9 @@ VersionedItem compact_incomplete_impl(
     std::vector<folly::Future<VariantKey>> fut_vec;
     std::vector<FrameSlice> slices;
     bool dynamic_schema = write_options.dynamic_schema;
-    auto policies = std::make_tuple(index_type_from_descriptor(first_seg.descriptor()), 
-                                    dynamic_schema ? VariantSchema{DynamicSchema::default_schema()} : VariantSchema{FixedSchema::default_schema()}, 
+    auto index = index_type_from_descriptor(first_seg.descriptor());
+    auto policies = std::make_tuple(index, 
+                                    dynamic_schema ? VariantSchema{DynamicSchema::default_schema(index)} : VariantSchema{FixedSchema::default_schema(index)}, 
                                     sparsify ? VariantColumnPolicy{SparseColumnPolicy{}} : VariantColumnPolicy{DenseColumnPolicy{}}
                                     );
     util::variant_match(std::move(policies), [
@@ -1153,7 +1156,7 @@ PredefragmentationInfo get_pre_defragmentation_info(
         if (slice.row_range.diff() < segment_size && !compaction_start_info)
             compaction_start_info = {slice.row_range.start(), segment_idx};
             
-        if (slice.col_range.start() == 1){
+        if (slice.col_range.start() == pipeline_context->descriptor().index().field_count()){//where data column starts
             first_col_segment_idx.emplace_back(slice.row_range.start(), segment_idx);
             if (new_segment_row_size == 0)
                 ++num_to_segments_after_compact;
@@ -1195,8 +1198,9 @@ VersionedItem defragment_symbol_data_impl(
     // in the new index segment, we will start appending after this value
     std::vector<folly::Future<VariantKey>> fut_vec;
     std::vector<FrameSlice> slices;
-    auto policies = std::make_tuple(index_type_from_descriptor(pre_defragmentation_info.pipeline_context->descriptor()),
-                                    options.dynamic_schema ? VariantSchema{DynamicSchema::default_schema()} : VariantSchema{FixedSchema::default_schema()}
+    auto index = index_type_from_descriptor(pre_defragmentation_info.pipeline_context->descriptor());
+    auto policies = std::make_tuple(index,
+                                    options.dynamic_schema ? VariantSchema{DynamicSchema::default_schema(index)} : VariantSchema{FixedSchema::default_schema(index)}
                                     );
 
     util::variant_match(std::move(policies), [
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -2508,7 +2508,7 @@ class NativeVersionStore:
                 )
             )
 
-    def is_symbol_fragmented(self, symbol: str, segment_size: int = None) -> bool:
+    def is_symbol_fragmented(self, symbol: str, segment_size: Optional[int] = None) -> bool:
         """
         Check whether the number of segments that would be reduced by compaction is more than or equal to the
         value specified by the configuration option "SymbolDataCompact.SegmentCount" (defaults to 100).
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -1419,7 +1419,7 @@ class Library:
         """
         self._nvs.version_store.reload_symbol_list()
 
-    def is_symbol_fragmented(self, symbol: str, segment_size: int = None) -> bool:
+    def is_symbol_fragmented(self, symbol: str, segment_size: Optional[int] = None) -> bool:
         """
         Check whether the number of segments that would be reduced by compaction is more than or equal to the
         value specified by the configuration option "SymbolDataCompact.SegmentCount" (defaults to 100).
--- a/python/tests/unit/arcticdb/version_store/test_append.py
+++ b/python/tests/unit/arcticdb/version_store/test_append.py
@@ -450,7 +450,7 @@ def test_append_mix_sorted_non_sorted(lmdb_version_store):
 @given(
     col_per_append_df=st.integers(2, 100),
     col_name_set=st.integers(1, 10000),
-    num_of_rowss=st.lists(st.lists(st.integers(1, 20), min_size=1, max_size=10), min_size=1, max_size=10),
+    num_rows_per_test_cycle=st.lists(st.lists(st.integers(1, 20), min_size=1, max_size=10), min_size=1, max_size=10),
     column_group_size=st.integers(2, 100),
     segment_row_size=st.integers(2, 100),
     dynamic_schema=st.booleans(),
@@ -461,7 +461,7 @@ def test_append_with_defragmentation(
     sym,
     col_per_append_df,
     col_name_set,
-    num_of_rowss,
+    num_rows_per_test_cycle,
     get_wide_df,
     column_group_size,
     segment_row_size,
@@ -546,7 +546,7 @@ def test_append_with_defragmentation(
 
     assume(col_per_append_df <= col_name_set)
     assume(
-        num_of_row % 2 != 0 for num_of_rows in num_of_rowss for num_of_row in num_of_rows
+        num_of_row % 2 != 0 for num_of_rows in num_rows_per_test_cycle for num_of_row in num_of_rows
     )  # Make sure at least one successful compaction run per cycle
 
     set_config_int("SymbolDataCompact.SegmentCount", 1)
@@ -559,7 +559,7 @@ def test_append_with_defragmentation(
         dynamic_strings=dynamic_strings,
         reuse_name=True,
     )
-    for num_of_rows in num_of_rowss:
+    for num_of_rows in num_rows_per_test_cycle:
         before_compact, index_offset = run_test(
             lib,
             col_per_append_df,
--- a/cpp/arcticdb/column_store/memory_segment.hpp
+++ b/cpp/arcticdb/column_store/memory_segment.hpp
@@ -427,6 +427,10 @@ public:
         return output;
     }
 
+    StreamId get_index_col_name() const{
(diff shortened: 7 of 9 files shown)


python/arcticdb/version_store/library.py
```diff
@@ -1261,6 +1261,77 @@ def reload_symbol_list(self):
...
+        Notes
+        ----------
+        Config map setting - SymbolDataCompact.SegmentCount will be replaced by a library setting
+        in the future. This API will allow overriding the setting as well.
+        
+        Returns
+        -------
+        bool
+        """
+        return self._nvs.is_symbol_fragmented(symbol, segment_size)
```
mehertz, line 1286: This should raise an exception with an error code!

python/tests/conftest.py
```diff
@@ -247,7 +247,7 @@ def wrapped(*args, **kwargs):
         for storage in cfg.storage_by_id.values():
             lmdb = LmdbConfig()
             lmdb.ParseFromString(storage.config.value)
-            shutil.rmtree(lmdb.path)
+            shutil.rmtree(lmdb.path, True)
```
qc00, line 250: Minor: consider using `ignore_errors=True` to improve readability

python/tests/conftest.py
```diff
@@ -166,7 +164,11 @@ def _version_store_factory_impl(
     name = name or default_name
     if name == "_unique_":
         name = name + str(len(used))
-    assert (name not in used) or reuse_name, f"{name} is already in use"
+    if name in used:
+        if reuse_name:
+            return used[name]  # Windows does not allow LMDB open mmap with the same path twice
```
qc00, line 169: The point of using `reuse_name` is to have a different library config while reusing the underlying data.

This change is not doing the former.

python/tests/conftest.py
```diff
@@ -255,16 +258,27 @@ def lmdb_version_store_prune_previous(version_store_factory):
 
 @pytest.fixture
 def lmdb_version_store_big_map(version_store_factory):
-    return version_store_factory(lmdb_config={"map_size": 2**30})
+    return version_store_factory(lmdb_config={"map_size": 2 ** 30})
+
 
-@pytest.fixture(scope='function', params=[False])
```
qc00: Wait, where did these get changed to single quotes in the first place?

qc00 (changes requested):

cpp/arcticdb/version/version_core.cpp
```diff
@@ -963,27 +928,25 @@ VersionedItem compact_incomplete_impl(
...
-        using IndexType = decltype(idx);
-
-        if(sparsify) {
-            do_compact<IndexType, DynamicSchema, RowCountSegmentPolicy, SparseColumnPolicy>(
+    bool dynamic_schema = write_options.dynamic_schema;
+    auto policies = std::make_tuple(index_type_from_descriptor(first_seg.descriptor()), 
+                                    dynamic_schema ? VariantSchema{DynamicSchema::default_schema()} : VariantSchema{FixedSchema::default_schema()}, 
+                                    sparsify ? VariantColumnPolicy{SparseColumnPolicy{}} : VariantColumnPolicy{DenseColumnPolicy{}}
+                                    );
+    util::variant_match(std::move(policies), [
```
qc00, line 936: This is a lot of indirection just to call `do_compact()` with different values known only at run time.

I wonder, then, if having these as template parameters is worth it.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1000,4 +963,109 @@ VersionedItem compact_incomplete_impl(
...
+        const WriteOptions& options,
+        size_t segment_size) {
+    util::check(update_info.previous_index_key_.has_value(), "No latest undeleted version found for data compaction");
+
+    auto pipeline_context = std::make_shared<PipelineContext>();
+    pipeline_context->stream_id_ = stream_id;
+    pipeline_context->version_id_ = update_info.next_version_id_;
+
+    ReadQuery read_query;
+    read_indexed_keys_to_pipeline(store, pipeline_context, *(update_info.previous_index_key_), read_query, defragmentation_read_options_generator(options));
```
qc00, line 979: Calling this low-level stuff scares me. Please write a test to double-check that check recursive normaliser works with this method?

cpp/arcticdb/version/version_core.hpp
```diff
@@ -96,8 +96,25 @@ VersionedItem compact_incomplete_impl(
     bool append,
     bool convert_int_to_float,
     bool via_iteration,
-    bool sparsify);
-
+    bool sparsify,
+    const WriteOptions& write_options);
+
+std::tuple<std::shared_ptr<PipelineContext>, ReadQuery, size_t, std::optional<size_t>> get_pre_defragmentation_info(
```
qc00, line 102: Please write some docs. Returning a `ReadQuery` is most unusual.

The last two items effectively have the same type. Should really return a `struct` with each field properly named.

python/arcticdb/version_store/_store.py
```diff
@@ -2278,6 +2278,80 @@ def scan_object_sizes(self):
...
+        bool
+        """
+        return self.version_store.is_symbol_fragmented(symbol, segment_size)
+
+    def defragment_symbol_data(self, symbol: str, segment_size: int = None) -> VersionedItem:
+        """
+        Compacts fragmented segments by merging row-sliced segments (https://docs.arcticdb.io/technical/on_disk_storage/#data-layer).
+        This method calls `is_symbol_fragmented` to determine whether to proceed with the defragmentation operation. 
+
+        CAUTION - One caveat of this method is that the resulting data will not be column sliced which may impact the performance
```
qc00, line 2311: Wait, if we then `update/append` data to a compacted symbol, aren't we creating that column groups size mismatch situation we say we aren't going to support?

python/arcticdb/version_store/_store.py
```diff
@@ -2278,6 +2278,80 @@ def scan_object_sizes(self):
...
+            Symbol name.
+        segment_size: `int`
+            Target for maximum no. of rows per segment, after compaction.
+            If parameter is not provided, library option - "segment_row_size" will be used
+            Note that no. of rows per segment, after compaction, may exceed the target.
+            It is for achieving smallest no. of segment after compaction. Please refer to below example for further explantion.
+
+        Returns
+        -------
+        VersionedItem object that contains a .data and .metadata element.
```
qc00, line 2326: A bit surprised we return the data.

python/tests/conftest.py
```diff
@@ -256,12 +258,27 @@ def lmdb_version_store_prune_previous(version_store_factory):
...
+    return version_store_factory(lmdb_config={"map_size": 2 ** 30})
 
 
-@pytest.fixture
-def lmdb_version_store_column_buckets(version_store_factory):
-    return version_store_factory(dynamic_schema=True, column_group_size=3, segment_row_size=2, bucketize_dynamic=True)
+@pytest.fixture(scope="function", params=[False])
+def lmdb_version_store_column_buckets(version_store_factory, request):
+    return version_store_factory(
+        dynamic_schema=True, column_group_size=3, segment_row_size=2, bucketize_dynamic=True, reuse_name=request.param
```
qc00, line 267: I am not sure what you're trying to achieve with this?

python/tests/conftest.py
```diff
@@ -256,12 +258,27 @@ def lmdb_version_store_prune_previous(version_store_factory):
...
+    )
+
+
+@pytest.fixture(scope="function", params=[False])
+def lmdb_version_store_column_buckets_dynamic_string(version_store_factory, request):
```
qc00, line 272: Please don't create a global fixture that's only used in one test. We need to limit the proliferation of these.

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -280,3 +282,173 @@ def test_(initial: InputFactories, append: InputFactories, match, swap, lmdb_ver
...
+    [(4, 20, 100, 1, 5, 2, True, True, True, True, True, True)],
+    indirect=[
+        "lmdb_version_store_column_buckets",
+        "lmdb_version_store_tiny_segment",
+        "lmdb_version_store_tiny_segment_dynamic_string",
+        "lmdb_version_store_tiny_segment_dynamic",
+        "lmdb_version_store_tiny_segment_dynamic_dynamic_string",
+        "lmdb_version_store_column_buckets_dynamic_string",
+    ],
+)
```
qc00, line 311: Using test fixtures in this style is unnecessarily opaque, and borderline unreadable.

Please just inject the `version_store_factory` pass the necessary config combinations.

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -280,3 +282,173 @@ def test_(initial: InputFactories, append: InputFactories, match, swap, lmdb_ver
...
+                lmdb_version_store_tiny_segment_dynamic_string,
+            ],
+        ),
+    ]:
+        for lib in libs:
+            for segment_row_size in [1, 3]:
+                for df_in_str in [True, False]:
+                    for _ in range(num_of_repeat):
+                        before_compact = pd.DataFrame()
+                        for cycle in range(num_of_compact_per_cycle):
```
qc00, line 425: Don't reinvent `parameterize` or `hypothesis`. Please use the `version_store_factory` method I mentioned above.

qc00 (changes requested):

python/tests/conftest.py
phoebusm: Oops. I have run black formatting with this file. Will remove the last commit.

python/tests/conftest.py
phoebusm: Sorry it was trying to get away from the lmdb open mmap file issue in Windows. It is a newly introduced limitation, which makes a test fixture unable to be passed more than one lmdb storage fixture.
I guess I have to bite the bullet and re-design the defragmentation test case.

cpp/arcticdb/util/variant.hpp
```diff
@@ -14,14 +14,42 @@ namespace arcticdb::util {
...
+template<size_t... I, class... TupleTs, class...Ts>
+auto variant_match(std::index_sequence<I...>, std::tuple<TupleTs...> &&v, Ts... ts){
+    return std::visit(overload{ts...}, std::get<I>(v)...);
+}
+
 template<class Variant, class... Ts>
 auto variant_match(Variant && v, Ts... ts){
-    return std::visit(overload{ts...}, v);
+    if constexpr(is_tuple<std::remove_cv_t<std::remove_reference_t<Variant>>>::value)
+    {
```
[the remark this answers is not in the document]
phoebusm: Oops bad habit

cpp/arcticdb/version/version_core-inl.hpp
```diff
@@ -97,4 +98,56 @@ void merge_frames_for_keys(
...
+    std::vector<folly::Future<VariantKey>>& fut_vec,
+    std::vector<pipelines::FrameSlice>& slices,
+    const std::shared_ptr<Store>& store,
+    bool convert_int_to_float,
+    std::optional<size_t> segment_size){
+        auto index = stream::index_type_from_descriptor(pipeline_context->descriptor());
+        stream::SegmentAggregator<IndexType, SchemaType, SegmentationPolicy, DensityPolicy>
+        aggregator{
+            [&slices](pipelines::FrameSlice slice) {
+                slices.emplace_back(std::move(slice));
```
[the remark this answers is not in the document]
phoebusm: I didn't put any thought about it as this is the code passed by Pat.
And yes this part can be optimized per your suggestion.

cpp/arcticdb/version/version_core.cpp
phoebusm: I don't have another choice or I need to revamp quite a lot of other classes, e.g. `SegmentAggregator` and `Aggregator`, which `do_compact()` calls.

cpp/arcticdb/version/version_core.cpp
phoebusm: I am not sure what recursive normaliser is. Care to point me to a test method to let me refer to it?

cpp/arcticdb/version/version_core.hpp
phoebusm: 1. `ReadQuery` is required by `Clause`, (`clause.hpp`)
2. It is arguable. If in doubt,  meaning of fields can be found in the caller of `get_pre_defragmentation_info`, as the variables are named properly over there.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -715,7 +715,34 @@ VersionedItem LocalVersionedEngine::compact_incomplete_dynamic(
...
+    version_map_->write_version(store_, versioned_item.key_);
+
+    if(cfg_.symbol_list())
+        symbol_list().add_symbol(store_, stream_id);
+
+    return versioned_item;
+}
+
+bool LocalVersionedEngine::is_symbol_fragmented(const StreamId& stream_id, std::optional<size_t> segment_size) {
+    auto update_info = get_latest_undeleted_version_and_next_version_id(
```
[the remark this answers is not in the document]
phoebusm: Good spot. Although it won't have a hit on the performance, unnecessary code should be added/run.

python/arcticdb/version_store/_store.py
```diff
@@ -2278,6 +2278,80 @@ def scan_object_sizes(self):
...
+
+        Raises
+        ------
+        1002 ErrorCategory.INTERNAL:E_ASSERTION_FAILURE
```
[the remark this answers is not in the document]
phoebusm: I have added them in the error message in another branch. I think it will be better if the user can understand the exception on the spot, instead of looking up the errorcode list.

python/tests/conftest.py
phoebusm: These newly added lmdb fixtures also serve as the reminders that "Certain combinations of db setting is needed to be considered".
These combinations are not widely test in other test cases, which I did have found bugs with them, during the development of this project.

python/tests/unit/arcticdb/version_store/test_append.py
phoebusm: I will revamp this test method per your suggestions.

cpp/arcticdb/version/version_core.hpp
mehertz: Is the `ReadQuery` returned empty?

python/arcticdb/version_store/_store.py
mehertz: What does merged mean in this scenario?

cpp/arcticdb/version/version_core.cpp
mehertz: Do you mean check that recursively-normalized data can be compacted?

python/arcticdb/version_store/_store.py
mehertz: mhmm. If the data is very large (which it could be - hence why we are de-fragmenting!) this will likely OOM when handing over to Pandas. I think we probably shouldn't return the data to be honest. Can imagine that being a problem.

cpp/arcticdb/version/version_core.cpp
phoebusm: If so, it is covered in the current test case.
I am also writing a new test case to address Qi's comment, which will utilize hypothesis. It should have covered this scenario in a better way.

python/arcticdb/version_store/_store.py
phoebusm: Remove column slicing.
Steps for defragmentation:
1. Remove column slicing
2. Remove row slicing as much as possible (per library setting `segment_row_size`)

phoebusm: Remove column slicing.
Steps for defragmentation:
1. Remove column slicing
2. Remove row slicing as much as possible (per library setting `segment_row_size`)

mehertz: Ok - I think this will hit the case that Gem hit. The end result is that QueryBuilder queries breaks.

So just to clarify - we defragment the data, so we'll write out a new index key that no column slicing. If we append data, those rows will be column sliced. As a result the column slicing will differ from one row to another in the same index key, which I believe is the issue that @GemDot saw.

mehertz: At the very least we should add a test for this. Defragement, append, query using QueryBuilder.

willdealtry (changes requested):

python/arcticdb/version_store/library.py
```diff
@@ -1294,6 +1294,82 @@ def reload_symbol_list(self):
...
+
+    def defragment_symbol_data(self, symbol: str, segment_size: Optional[int] = None) -> None:
+        """
+        Compacts fragmented segments by merging row-sliced segments (https://docs.arcticdb.io/technical/on_disk_storage/#data-layer).
+        This method calls `is_symbol_fragmented` to determine whether to proceed with the defragmentation operation. 
+
+        CAUTION - One caveat of this method is that the resulting data will not be column sliced which may impact the performance
+        of read operations.
```
mehertz, line 1327: I'm keen to make this as clear as possible. What do you think of this:

```
CAUTION - Please note that a major restriction of this method at present is that any column slicing present on the data will be removed in the new version created as a result of this method. As a result, if the impacted symbol has more than 127 columns (default value), the performance of selecting individual columns of the symbol (by using the `columns` parameter) may be negatively impacted in the defragmented version. If your symbol has less than 127 columns this caveat does not apply. For more information, please see `columns_per_segment` here:

https://docs.arcticdb.io/api/arcticdb/arcticdb.LibraryOptions
```

cpp/arcticdb/version/version_core.cpp
```diff
@@ -998,4 +961,109 @@ VersionedItem compact_incomplete_impl(
...
+
+    using CompactionStartInfo = std::pair<size_t, size_t>;//row, segment_append_after
+    std::vector<CompactionStartInfo> first_col_segment_idx;
+    first_col_segment_idx.reserve(pipeline_context->slice_and_keys_.size());
+    std::optional<CompactionStartInfo> compaction_start_info;
+    size_t segment_idx = 0, num_to_segments_after_compact = 0, new_segment_row_size = 0;
+    for(auto rev_it = pipeline_context->slice_and_keys_.begin(); rev_it != pipeline_context->slice_and_keys_.end(); rev_it++) {
```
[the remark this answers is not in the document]
phoebusm: Oops should be `it`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -998,4 +961,109 @@ VersionedItem compact_incomplete_impl(
...
+        auto &slice = rev_it->slice();
+
+        if (slice.row_range.diff() < segment_size && !compaction_start_info)
+            compaction_start_info = {slice.row_range.start(), segment_idx};
+            
+        if (slice.col_range.start() == 1){
```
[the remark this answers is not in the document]
phoebusm: Just test again and `col_range` is 1-indexed.

cpp/arcticdb/stream/schema.hpp
```diff
@@ -25,7 +25,11 @@ class FixedSchema {
   public:
     FixedSchema(StreamDescriptor desc, const Index& index) :
         desc_(std::move(desc)),
-        index_(std::move(index)){}
+        index_(index){}
+
+    static FixedSchema default_schema() {
+        return FixedSchema(StreamDescriptor(), stream::TimeseriesIndex::default_index());
+    }
```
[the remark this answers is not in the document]
phoebusm: They are served merely with `variant_match` to runtime dispatch template parameters. The choice of index carries no meaning. It references `inline Index default_index_type_from_descriptor(const IndexDescriptor::Proto &desc);`

cpp/arcticdb/pipeline/pipeline_context.hpp
```diff
@@ -121,24 +121,36 @@ struct PipelineContext : public std::enable_shared_from_this<PipelineContext> {
     std::optional<std::unordered_set<std::string_view>> filter_columns_set_;
     std::optional<SegmentInMemory> multi_key_;
     std::optional<util::BitSet> overall_column_bitset_;
-    util::BitSet compacted_;
+    std::vector<uint32_t> compacted_;
```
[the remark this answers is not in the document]
phoebusm: Let me try. Theoretically anything but `bool` should work.

cpp/arcticdb/version/version_core.cpp
willdealtry: This is the only game in town, we need to somehow move from multiple run-time values to the associated types. It might be easier on the eye if it created the objects as l-values with names and then moved them into the tuple, rather than quite a lot of ternary operator

cpp/arcticdb/pipeline/frame_utils.cpp
```diff
@@ -170,12 +170,33 @@ void adjust_slice_rowcounts(const std::shared_ptr<pipelines::PipelineContext>& p
 
 size_t adjust_slice_rowcounts(std::vector<pipelines::SliceAndKey> & slice_and_keys) {
     using namespace arcticdb::pipelines;
-    auto current_col = slice_and_keys[0].slice_.col_range.first;
+    auto current_col = slice_and_keys[0].slice_.col_range;
```
willdealtry, line 173: Doing this by columns is quite complicated. Would it not be easier just to sort by row range first: 
```suggestion
   size_t adjust_slice_rowcounts(std::vector<pipelines::SliceAndKey> & slice_and_keys) {
	if(slice_and_keys.empty())
		return;

	std::sort(std::begin(slice_and_keys), std::end(slice_and_keys), [] (const auto& left, const auto& right) {
		return std::tie(left.slice.row_range, left.slice.col_range) < std::tie(right.slice.row_range, right.slice.col_range);
	}
	
	auto offset = slice_and_keys[0].slice_.row_range.first;
	auto diff = slice_and_keys[0].slice_.row_range.diff();
	
	for(auto& slice_and_key : slice_and_keys) {
		if(slice_and_key.row_range.first != offset) {
			offset += diff;
			diff = slice_and_key.slice_.row_range.diff();
		}
		slice_and_key.row_range = RowRange{offset, diff};
	}
	
	return offset + diff;
}
```

willdealtry: Once we standardize the index sorting on rows-first, we can even drop the sort

cpp/arcticdb/stream/schema.hpp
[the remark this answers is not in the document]
willdealtry: I still think it's a bit of a code smell. If the index type matters then we are setting it wrongly, as it could be a row count index or a string index. If it doesn't matter, why are we setting it at all?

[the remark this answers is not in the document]
willdealtry: Could we pass the index type in as an argument?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -439,4 +441,150 @@ def test_append_mix_sorted_non_sorted(lmdb_version_store):
...
\ No newline at end of file
+    assert info["sorted"] == "UNSORTED"
+
+
+@use_of_function_scoped_fixtures_in_hypothesis_checked
+@settings(deadline=None, max_examples=10)
+@given(
+    col_per_append_df=st.integers(2, 100),
+    col_name_set=st.integers(1, 10000),
+    num_of_rowss=st.lists(st.lists(st.integers(1, 20), min_size=1, max_size=10), min_size=1, max_size=10),
```
mehertz, line 452: Typo: `num_of_rowss`

mehertz (changes requested):

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -439,4 +441,150 @@ def test_append_mix_sorted_non_sorted(lmdb_version_store):
...
+    column_group_size=st.integers(2, 100),
+    segment_row_size=st.integers(2, 100),
+    dynamic_schema=st.booleans(),
+    dynamic_strings=st.booleans(),
+    df_in_str=st.booleans(),
```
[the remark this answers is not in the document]
phoebusm: Only time-series index is supported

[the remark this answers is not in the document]
phoebusm: Or exception `Not supported index type for sort merge implementation` will be given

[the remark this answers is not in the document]
mehertz: Ah, ok. That's not a great error message though IMO - can you please change to:

```
Sort merge only supports datetime indexed data. You data does not have a datetime index.
```

python/tests/unit/arcticdb/version_store/test_append.py
mehertz: oh wow! can't we just call it `num_rows_per_test_cycle`? I fear everybody else who sees this will just think it's a typo as well unless this is a convention I've never seen before.

mehertz (approved):

Issue #153, opened 2023-03-27
Fix small remaining issues with Windows Python tests


poodlewars:
There are a few trivial issues left over with some of the Windows Python tests. For example:

- Flipping from int32 on write to int64 on read in some circumstances
- LMDB running out of memory on some stress tests. Either use a bigger LMDB for these tests only, or just skip them on Windows.

This ticket is to cover getting a "clean" Python test run on Windows.

PR #182, opened 2023-03-31
Small changes to get Python tests to pass and clean up on Windows

poodlewars:
#153


python/tests/conftest.py
```diff
@@ -202,6 +218,29 @@ def _version_store_factory_impl(
...
 
+def lmdb_version_store_cleanup(version_store_fixture):
+
+    @functools.wraps(version_store_fixture)
+    def wrapped(*args, **kwargs):
+        result: NativeVersionStore = version_store_fixture(*args, **kwargs)
+
+        yield result
+
+        #  pytest holds a member variable `cached_result` equal to `result` above which keeps the storage alive and
```
poodlewars, line 229: This was a nightmare to figure out

python/tests/stress/arcticdb/version_store/test_strings_flat_memory.py
poodlewars, line 1: This test works fine on Windows, but it's quite slow and doesn't assert anything, so I've just removed it.

python/tests/conftest.py
willdealtry: We should be able to award medals for valour above and beyond the call of duty.

python/tests/stress/arcticdb/version_store/test_strings_flat_memory.py
willdealtry: Yeah it was more of an investigation than anything else, it's not worth slowing down the builds for it

python/tests/unit/arcticdb/version_store/test_empty_writes.py
```diff
@@ -20,20 +20,19 @@ def test_write_no_rows(lmdb_version_store, sym):
     assert not lmdb_version_store.is_symbol_pickled(sym)
     df.index = df.index.astype("datetime64[ns]")
     df["a"] = df["a"].astype("float64")
-    # df.iloc[:, 2] = df.iloc[:, 2].astype("string")
     assert_frame_equal(lmdb_version_store.read(sym).data, df)
 
     df2 = pd.DataFrame([[1.3, 6, "test"]], columns=column_names, index=[pd.Timestamp(0)])
     df2 = df.append(df2)
-    # coercing not needed no more
```
willdealtry, line 28: Had a kind of R&B vibe to it previously. Very weird.

willdealtry (approved):

mehertz (approved):

python/tests/conftest.py
```diff
@@ -202,6 +218,29 @@ def _version_store_factory_impl(
     return out
 
 
+def lmdb_version_store_cleanup(version_store_fixture):
```
qc00, line 221: Some docs on how this should be used? I guess it's sensitive to the ordering with `@pytest.fixture`?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -1101,9 +1102,10 @@ def test_coercion_to_float(lmdb_version_store_string_coercion):
 
     assert df["col"].dtype == np.object_
 
-    with pytest.raises(ArcticNativeNotYetImplemented):
-        # Needs pickling due to the obj column
-        lib.write("test", df)
+    if sys.platform != "win32":  # SKIP_WIN Windows always uses dynamic strings
```
qc00, line 1105: GitHub issue reference unless we don't ever want to support static strings on Windows?

python/tests/unit/arcticdb/version_store/test_deletion.py
```diff
@@ -30,15 +27,15 @@ def test_delete_version_with_update(version_store_factory, pos, sym):
     symbol = sym
 
     idx = pd.date_range("1970-01-01", periods=100, freq="D")
-    df = pd.DataFrame({"a": range(len(idx))}, index=idx)
+    df = pd.DataFrame({"a": range(len(idx))}, index=idx, dtype=np.int64)
```
qc00, line 30: A bit worried about why this change is necessary.

Is this a behavioural change in Pandas that we should actually support?

Issue #150, opened 2023-03-27
Filter numeric isin overflow issues


poodlewars:
```
def test_filter_numeric_isin_single(lmdb_version_store):
    max_signed = 2 ** 63
    df = pd.DataFrame({"a": [0, 1, 2 ** 62]})
    lmdb_version_store.write("test_filter_numeric_isin_single", df)

    vals = [0, 1]
    q = QueryBuilder()
    q = q[q["a"].isin(vals)]
    result = lmdb_version_store.read("test_filter_numeric_isin_single", query_builder=q).data

    expected = pd.DataFrame({"a": [0, 1]})
    assert_frame_equal(expected, result)  # result actually includes 2 ** 62
```

The type arithmetic for numeric isins is incorrect.

PR #148 looks at this. It needs fixing up from its current state to:

- Always "widen" the valueset type out to be at least as wide as the column data
- We should be able to leave the column data as narrow as possible (unlike the PR which always widens it)

Issue #149, opened 2023-03-27
Filter numeric isin tests can fail with new Pandas and Hypothesis


poodlewars:
For example,

```
2: =========================== short test summary info ============================
2: FAILED tests/unit/arcticdb/version_store/test_filtering.py::test_filter_numeric_isnotin - assert False
2: Falsifying example: test_filter_numeric_isnotin(
2:     df=
2:                       a
2:         0  9.007199e+15
2:     ,
2:     vals=frozenset({int64(9007199254740993)}),
2:     lmdb_version_store=NativeVersionStore: Library: local.test_2023-03-23T19:47:25.829980, Primary Storage: lmdb_storage.,
2: )
2: 
2: You can reproduce this example by temporarily adding @reproduce_failure('6.70.0', b'AXicY2RgYGZgYFBgAANGRjAPymVkAAAEtQBL') as a decorator on your test case
```

We originally thought this was a Windows issue, but it is happening on both platforms.

The issue is with the test setup. Despite specifying `integral_type_strategies` hypothesis sometimes uses examples with float dtype in the test dataframe.

poodlewars:
PR #148

Issue #128, opened 2023-03-20
Reduce expression_node.cpp compilation RAM usage


qc00:
I have discussed with @willdealtry and @alexowens90 about this. The general idea is to split the file up or cut down the template expansions somehow.

There's a work around script linked below.

PR #183, opened 2023-03-31
Fix rebase error from Numeric IsIn work

poodlewars:
Refactor for PR #128 accidentally removed the fix from PR #148 which addressed issues #149 and #150.

This just puts the main fix back, and adds a little more test coverage. Strangely it also fixes compilation on MSVC.

It also fixes a couple of clang warnings for good measure.


alexowens90 (approved):

qc00 (approved):
So an entire change went missing? How did that happen?

PR #185, opened 2023-04-02
Add quickstart instructions to build documentation

mehertz:


docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
-Setting up your OS dependencies
-===============================
+Docker Quickstart
+=================
 
-ManyLinux Docker image
-----------------------
-ManyLinux(https://github.com/pypa/manylinux) captures the common denominator of many Linux distributions,
-giving the confidence that code worked in this image runs everywhere.
+#### 1) Start the ArcticDB build docker image
```
qc00, line 4: Does not apply to Windows/MacOS

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+git submodule init && git submodule update
+```
+
+#### 3) Kick off the build
+
+```bash
+/opt/python/cp39-cp39/bin/python3 -m pip install -U pip setuptools wheel grpcio-tools
+ARCTIC_CMAKE_PRESET=skip /opt/python/cp39-cp39/bin/python3 setup.py develop
+# Change the below Python_EXECUTABLE value to build against a different Python version
+cmake -DPython_EXECUTABLE=/opt/python/cp39-cp39/bin/python -DTEST=off --preset linux-debug cpp
```
qc00, line 31: This python cannot be used to compile the C++ tests, which will be unnecessarily confusing

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+git clone https://github.com/man-group/ArcticDB.git
+cd ArcticDB
+git submodule init && git submodule update
+```
+
+#### 3) Kick off the build
+
+```bash
+/opt/python/cp39-cp39/bin/python3 -m pip install -U pip setuptools wheel grpcio-tools
+ARCTIC_CMAKE_PRESET=skip /opt/python/cp39-cp39/bin/python3 setup.py develop
```
qc00, line 29: For a quick start, it will be sufficient to run `pip install -ve .` (without `ARCTIC_CMAKE_PRESET=skip`).

The original version was trying to explain what above does internally which most users can then replace with IDE built-in features.

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+# Change the below Python_EXECUTABLE value to build against a different Python version
+cmake -DPython_EXECUTABLE=/opt/python/cp39-cp39/bin/python -DTEST=off --preset linux-debug cpp
+# NOTE: Set parallel level below to number of cores!
+cmake --build --parallel 8 --preset linux-debug cpp
```
qc00, line 33: Having a path while also using a preset is not officially supported by CMake ([doc](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project) indicates the two call patterns are distinct).

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+```
+
+#### 4) Run ArcticDB
+
+Ensure the below is run in the Git project root:
+
+```bash
+# PYTHONPATH first part = Python module, second part compiled C++ binary
+PYTHONPATH=`pwd`/python:`pwd`/cpp/out/linux-debug-build/arcticdb/ /opt/python/cp39-cp39/bin/python3
```
qc00, line 42: As long as `ARCTIC_CMAKE_PRESET=skip` is not set, `pip install/setup.py` will build and copy the module to the correct location. It's not necessary to fiddle with the `PYTHONPATH`.

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -26,80 +95,50 @@ See [pybind11 configuration](#pybind11-configuration).
...
+```bash
+cmake -DPython_EXECUTABLE=<path to python> -DTEST=ON --preset linux-debug cpp
 ```
 
-This will build and install ArcticDB in the current working directory.
-
-Compiling step by step
-----------------------
-During day-to-day development, it's quicker to rebuild only the part that changed.
+Note that `<path to python>` must point to a Python that is compatible with [Development.Embed](https://cmake.org/cmake/help/latest/module/FindPython.html). This will probably be the result of installing `python3-devel` from your distributions dependency manager.
```
poodlewars, line 125: Missing apostrophe distributions

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -26,80 +95,50 @@ See [pybind11 configuration](#pybind11-configuration).
...
+-----------------
 
-```bash
-# Recursively checkout ArcticDB
-# Make sure the Python version you want is the first on the PATH
+Configure ArcticDB with TEST=on:
 
-python -m pip install -ve .
+```bash
+cmake -DPython_EXECUTABLE=<path to python> -DTEST=ON --preset linux-debug cpp
```
poodlewars, line 122: TEST=ON is the default, no?

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+
+See [running tests below](#running-c-tests).
+
+Detailed Build Information
+==========================
+
+Docker Image Construction
+-------------------------
+
+The above docker image is built form [ManyLinux](https://github.com/pypa/manylinux).
```
poodlewars, line 77: s/form/from

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+See [running tests below](#running-c-tests).
+
+Detailed Build Information
+==========================
+
+Docker Image Construction
+-------------------------
+
+The above docker image is built form [ManyLinux](https://github.com/pypa/manylinux). 
+Build script located [here](https://github.com/man-group/ArcticDB/blob/master/build_tooling/build_many_linux_image.sh).
```
poodlewars, line 78: The build script is

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
-Linux
------
+#### How do I run the C++ tests?
+
+See [running tests below](#running-c-tests).
+
+Detailed Build Information
+==========================
+
+Docker Image Construction
```
poodlewars, line 74: Do we need to write about this at all?

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+The above docker image is built form [ManyLinux](https://github.com/pypa/manylinux). 
+Build script located [here](https://github.com/man-group/ArcticDB/blob/master/build_tooling/build_many_linux_image.sh).
+
+GitHub output [here](https://github.com/man-group/ArcticDB/pkgs/container/cibuildwheel_manylinux).
+
+**We recommend you use this image for compilation and testing!**
+
+Setting up Linux
+----------------
 The codebase and build system can work with any reasonably recent Linux distribution with at least
```
poodlewars, line 86: Not your bit of the docs but maybe we should be more proscriptive:
"We support GCC 10, CMake 3.21+ and MSVC v142+"

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -1,19 +1,88 @@
...
+```
+
+#### 3) Kick off the build
+
+```bash
+/opt/python/cp39-cp39/bin/python3 -m pip install -U pip setuptools wheel grpcio-tools
+ARCTIC_CMAKE_PRESET=skip /opt/python/cp39-cp39/bin/python3 setup.py develop
+# Change the below Python_EXECUTABLE value to build against a different Python version
+cmake -DPython_EXECUTABLE=/opt/python/cp39-cp39/bin/python -DTEST=off --preset linux-debug cpp
+# NOTE: Set parallel level below to number of cores!
```
poodlewars, line 32: Qi's cmake stuff already tries to guess a suitable number of cores I think

poodlewars (commented):
Happy with this besides Qi's suggestions, which I agree with.

docs/mkdocs/docs/technical/contributing.md
mehertz: I know. Frankly don't think it's a build deal - this should be self evident. I can specify that this applies only to Linux.

docs/mkdocs/docs/technical/contributing.md
mehertz: Yeh I tried that. Couldn't get it to work for love nor money to be honest.

This does work and honestly I think is more descriptive.

docs/mkdocs/docs/technical/contributing.md
mehertz: Otherwise you're left building against Python3.6 which is extremely old and had many other issues when I was trying this. 

I strongly believe this to be the simplest solution, next to ensuring an up to date version of Python that is suitable for test development is available in the image.

docs/mkdocs/docs/technical/contributing.md
mehertz: Didn't realise. Will change.

docs/mkdocs/docs/technical/contributing.md
mehertz: ...my grammer isn't good enough for this. I'm removing the word `distribution` 😆

docs/mkdocs/docs/technical/contributing.md
mehertz: Discussed - appreciate `pip install -ve .` is one line but I think this is more descriptive and still can be copied pasted

docs/mkdocs/docs/technical/contributing.md
poodlewars: Worrying that 3.6 has `many other issues`, in which case there is not a Python installation in the image that works for tests too. I strongly think we should either fix the tests-link-Python problem or include development Python builds for each version as soon as possible.

docs/mkdocs/docs/technical/contributing.md
mehertz: Discussed. Accept this might change with later versions for for now want to keep as is.

docs/mkdocs/docs/technical/contributing.md
mehertz: Leaving for now.

qc00 (approved):

docs/mkdocs/docs/technical/contributing.md
mehertz: No... but given that it's there I'm happy to leave it.

PR #186, opened 2023-04-02
Support multiple Python protobuf versions in one wheel

qc00:
Supports both Protobuf Python binding versions 3 and 4 in on wheel via namespace packages and dynamic loading.

This allows us to support Python 3.6 and 3.11 using the same codebase.

Contributes towards #6


mehertz:
@qc00 please provide a description of where this would be useful.

setup.py
```diff
@@ -150,7 +147,8 @@ def _configure_for_legacy_image(self, common_args, build_dir):
     setup(
         ext_modules=[CMakeExtension("arcticdb_ext")],
         package_dir={"": "python"},
-        packages=find_packages(where="python", exclude=["tests", "tests.*"]),
+        packages=find_packages(where="python", exclude=["tests", "tests.*"])
+        + find_namespace_packages(where="python", include=["arcticdb.proto.*"]),
```
mehertz, line 151: There's a literal `+` here!

qc00: That's the list concatenation operator? `black` forced it into that location.

python/arcticc/__init__.py
mehertz, line 1: That's a shame :/

python/arcticdb/__init__.py
```diff
@@ -1,10 +1,14 @@
...
-try:
-    import arcticdb_ext
-except Exception as e:
-    traceback.print_exc()
+_proto_ver = _protobuf.__version__.split(".")[0]
+if _proto_ver in "34":
+    # Use the namespace package (https://peps.python.org/pep-0420) feature to select the pb2 files matching the protobuf
+    _sys.path.append(os.path.join(os.path.dirname(__file__), "proto", _proto_ver))
+else:
+    raise NotImplementedError(f"We do not support protobuf {_protobuf.__version__}")
```
mehertz, line 11: `ArcticDB only supports protobuf version 3 or 4. You have protobuf version {_protobuf.__version__}.`.

python/arcticc/__init__.py
mehertz: Why is this required? Could deal with `arcticc` in `proto`, having it in `python` too is super sad IMO.

mehertz: Cool, in which case this is the exact opposite of a shame!

mehertz (approved):

python/arcticdb/__init__.py
```diff
@@ -13,5 +17,6 @@
 from arcticdb.options import LibraryOptions
 from arcticdb.tools import set_config_from_env_vars
 
-
 set_config_from_env_vars(os.environ)
+
+del os
```
poodlewars, line 22: ?

poodlewars (approved):

mehertz (approved):

PR #202, opened 2023-04-04
#184 Apply library options to LMDB too in order to turn on dynamic strings

poodlewars:
It turns out that library options were not being applied correctly to LMDB at all, so this applies all of them.

This also takes the opportunity to run many of the `test_arctic.py` tests against the LMDB backend too, whose test coverage was pretty non-existent. The LMDB backed tests execute extremely quickly so build time is not a concern.

### DIFF
--- a/python/arcticdb/adapters/arctic_library_adapter.py
+++ b/python/arcticdb/adapters/arctic_library_adapter.py
@@ -11,6 +11,26 @@ from arcticdb_ext.storage import Library
 from abc import ABC, abstractmethod
 
 
+def set_library_options(lib_desc: "LibraryConfig", options: LibraryOptions):
+    write_options = lib_desc.version.write_options
+
+    write_options.dynamic_strings = True
+    write_options.recursive_normalizers = True
+    write_options.use_tombstones = True
+    write_options.fast_tombstone_all = True
+    lib_desc.version.symbol_list = True
+
+    write_options.prune_previous_version = False
+    write_options.pickle_on_failure = False
+    write_options.snapshot_dedup = False
+    write_options.delayed_deletes = False
+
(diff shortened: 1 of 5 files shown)


python/tests/conftest.py
```diff
@@ -121,12 +121,23 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
     ] + ":" + bucket + "?access=" + aws_access_key + "&secret=" + aws_secret_key + "&port=" + port
 
 
-@pytest.fixture(scope="function")
-def arctic_library(moto_s3_uri_incl_bucket):
-    ac = Arctic(moto_s3_uri_incl_bucket)
+@pytest.fixture(scope="function", params=("S3", "LMDB"))
```
poodlewars, line 124: I only discovered this nice pytest feature this morning. It will effectively create one test run for each value of params for each test that uses the fixture. Therefore all tests that use the `arctic_library` fixture will now run against LMDB as well as S3.

python/arcticdb/adapters/arctic_library_adapter.py
```diff
@@ -11,6 +11,26 @@
 from abc import ABC, abstractmethod
 
 
+def set_library_options(lib_desc: "LibraryConfig", options: LibraryOptions):
```
poodlewars, line 14: This is just pulled up unchanged from the S3 adapter.

mehertz:
Closes #47

python/tests/conftest.py
mehertz: Neat!

mehertz (approved):

qc00 (approved):

poodlewars:
closes #184

PR #203, opened 2023-04-04
Prevent out-of-order data being written unless explicitly requested

GemDot:


cpp/arcticdb/version/version_core.cpp
```diff
@@ -525,6 +536,10 @@ void check_column_and_date_range_filterable(const pipelines::index::IndexSegment
     "The data for this symbol is pickled and does not support date_range, row_range, or column queries");
     util::check(index_segment_reader.has_timestamp_index() || !std::holds_alternative<IndexRange>(read_query.row_filter),
             "Cannot apply date range filter to symbol with non-timestamp index");
+    util::check(index_segment_reader.tsd().stream_descriptor().sorted() == arcticc::pb2::descriptors_pb2::SortedValue::UNKNOWN ||
+        index_segment_reader.tsd().stream_descriptor().sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING ||
+        !std::holds_alternative<IndexRange>(read_query.row_filter),
+            "When filtering data using _date_range_, the symbol must be sorted in ascending order. ArcticDB believes (SYMBOL) to not be sorted in ascending order and cannot therefore filter the data using _date_range_.");
```
mehertz, line 542: The underscores in `_date_range_` won't be rendered.

mehertz, line 542: `(SYMBOL)` should be replaced by the actual symbol!

python/arcticdb/version_store/library.py
```diff
@@ -256,6 +256,7 @@ def write(
         metadata: Any = None,
         prune_previous_versions: bool = True,
         staged=False,
+        ensure_sorted=False,
```
mehertz, line 259: Needs to be documented in the docstring below

python/arcticdb/version_store/library.py
```diff
@@ -426,7 +428,7 @@ def _raise_if_unsupported_type_in_write_batch(payloads):
         raise ArcticUnsupportedDataTypeException(error_message)
 
     def write_batch(
-        self, payloads: List[WritePayload], prune_previous_versions: bool = True, staged=False
+        self, payloads: List[WritePayload], prune_previous_versions: bool = True, staged=False, ensure_sorted=True
```
mehertz, line 431: Needs documenting below. This will apply to every payload member?

python/arcticdb/version_store/library.py
```diff
@@ -535,7 +538,12 @@ def write_batch_pickle(
         )
 
     def append(
-        self, symbol: str, data: NormalizableType, metadata: Any = None, prune_previous_versions: bool = False
+        self,
+        symbol: str,
+        data: NormalizableType,
+        metadata: Any = None,
+        prune_previous_versions: bool = False,
+        ensure_sorted: bool = True,
```
mehertz, line 546: Documentation

python/tests/unit/arcticdb/version_store/test_date_range.py
```diff
@@ -40,3 +42,80 @@ def test_arctic_date_range_max():
...
+
+    assert df.index.is_monotonic_increasing == False
+    assert df.index.is_monotonic_decreasing == False
+    lmdb_version_store.write(symbol, df)
+    info = lmdb_version_store.get_info(symbol)
+    assert info["sorted"] == "UNSORTED"
+    with pytest.raises(InternalException) as e_info:
+        data = lmdb_version_store.read(
+            symbol, date_range=(DateRange(pd.Timestamp("2019-01-03"), pd.Timestamp("2019-01-06")))
+        ).data
```
mehertz, line 79: Can we also test what happens when you call date range and it's not even a DateTime index please

cpp/arcticdb/version/version_core.cpp
```diff
@@ -132,6 +137,8 @@ folly::Future<AtomKey> async_append_impl(
     bool bucketize_dynamic = index_segment_reader.bucketize_dynamic();
     auto row_offset = index_segment_reader.tsd().total_rows();
     util::check_rte(!index_segment_reader.is_pickled(), "Cannot append to pickled data");
+    util::check(!ensure_sorted || frame.desc.get_sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING || !std::holds_alternative<stream::TimeseriesIndex>(frame.index),
+        "When calling _append_ on sorted data and ensure_sorted is enabled, the new data must be sorted. ArcticDB relies on Pandas to detect if data is sorted - you can call DataFrame.index.is_monotonic_increasing on your input DataFrame to see if Pandas believes the data to be sorted");
```
mehertz, line 141: Remove underscores

mehertz, line 141: Error code

cpp/arcticdb/version/version_core.cpp
```diff
@@ -95,14 +96,17 @@ folly::Future<entity::AtomKey> async_write_dataframe_impl(
...
+    bool ensure_sorted
     ) {
     ARCTICDB_SAMPLE(DoWrite, 0)
 
     // Slice the frame according to the write options
     frame.set_bucketize_dynamic(options.bucketize_dynamic);
     auto slicing_arg = get_slicing_policy(options, frame);
     auto partial_key = IndexPartialKey{frame.desc.id(), version_id};
+    util::check(!ensure_sorted || frame.desc.get_sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING || !std::holds_alternative<stream::TimeseriesIndex>(frame.index),
+                "When calling _write_ and ensure_sorted is enabled, the new data must be sorted. ArcticDB relies on Pandas to detect if data is sorted - you can call DataFrame.index.is_monotonic_increasing on your input DataFrame to see if Pandas believes the data to be sorted");
```
mehertz, line 109: Error code

python/arcticdb/version_store/_normalization.py
```diff
@@ -769,9 +772,22 @@ def normalize(self, item, string_max_len=None, dynamic_strings=False, coerce_col
             norm_meta.df.common.columns.name = item.columns.name
         else:
             norm_meta.df.common.columns.fake_name = True
+
+        sort_status = _SortedValue.UNKNOWN
+        if hasattr(item.index, "is_monotonic_increasing"):
```
qc00, line 777: Should really `MultiIndex`

python/arcticdb/version_store/_store.py
```diff
@@ -484,6 +484,8 @@ def write(
             Pickle `data` if it can't be normalized.
         kwargs :
             passed through to the write handler
+                ensure_sorted: bool, default=False
+                    If True, will verify new data is sorted, but only if it contains a datetime index.
```
qc00, line 488: Indentation

qc00, line 488: Indentation

python/arcticdb/version_store/_store.py
```diff
@@ -22,7 +22,7 @@
 from typing import Any, Optional, Union, List, Mapping, Iterable, Sequence, Tuple, Dict, TYPE_CHECKING
 from contextlib import contextmanager
 
-from arcticc.pb2.descriptors_pb2 import TypeDescriptor
+from arcticc.pb2.descriptors_pb2 import TypeDescriptor, SortedValue
```
qc00, line 25: Unused?

python/arcticdb/version_store/_store.py
```diff
@@ -556,7 +559,7 @@ def write(
                 return None
             else:
                 vit = self.version_store.write_versioned_dataframe(
-                    symbol, item, norm_meta, udm, prune_previous_version, sparsify_floats
+                    symbol, item, norm_meta, udm, prune_previous_version, sparsify_floats, ensure_sorted
```
qc00, line 562: Consider creating an options object that passes from Python all the way to the `*_impl` in C++.

python/arcticdb/version_store/library.py
qc00, line 431: Why would `write_batch` has a different default from `write`??

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -288,3 +288,34 @@ def test_(initial: InputFactories, append: InputFactories, match, swap, lmdb_ver
     to_append, _ = append.make(abs(next_start), 1)
     with pytest.raises(NormalizationException):
         lib.append("s", to_append, match=match)
+
+
```
qc00, line 292: This is not a small change, so really need a lot more tests for each. I would expect:

* `append(ensure_sorted=False)`
* `RangeIndex`
* `MultiIndex`
* Various combinations of ascending, descending and unsorted in both existing data and the argument (you might want to look into Hypothesis)

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -288,3 +288,34 @@ def test_(initial: InputFactories, append: InputFactories, match, swap, lmdb_ver
...
+    info = lmdb_version_store.get_info(symbol)
+    assert info["sorted"] == "ASCENDING"
+
+    num_rows = 20
+    initial_timestamp = pd.Timestamp("2020-01-01")
+    dtidx = np.roll(pd.date_range(initial_timestamp, periods=num_initial_rows), 3)
+    df2 = pd.DataFrame(
+        {"uint8": _random_integers(num_rows, np.uint8), "uint32": _random_integers(num_rows, np.uint32)}, index=dtidx
+    )
+    assert df2.index.is_monotonic_increasing == False
```
qc00, line 318: The fixture setup is a bit too long, which makes it hard to see what the test is actually checking. (IMO, the code doesn't care about the types of the columns nor the number of rows nor randomness.)

Please consider using existing fixture creation functions.

python/tests/unit/arcticdb/version_store/test_date_range.py
```diff
@@ -40,3 +42,80 @@ def test_arctic_date_range_max():
...
+        },
+        index=np.roll(dtidx, 3),
+    )
+
+    assert df.index.is_monotonic_increasing == False
+    assert df.index.is_monotonic_decreasing == False
+    lmdb_version_store.write(symbol, df)
+    info = lmdb_version_store.get_info(symbol)
+    assert info["sorted"] == "UNSORTED"
+    with pytest.raises(InternalException) as e_info:
```
qc00, line 76: `InternalException` is a bit generic.

For all new errors, a unique error code should be assigned and asserted in tests.

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -468,3 +468,43 @@ def test_update_pickled_data(lmdb_version_store):
     df2 = pd.DataFrame({"a": [1000]}, index=idx[1:2])
     with pytest.raises(InternalException) as e_info:
         lmdb_version_store.update(symbol, df2)
+
+
+def _random_integers(size, dtype):
```
qc00, line 473: Do not copy and paste these.

python/arcticdb/version_store/_normalization.py
mehertz: This is a great point actually - if multiindex is set is `index` not set?

python/arcticdb/version_store/library.py
mehertz: I think `write` is wrong - it looks to me like it defaults to `False` but it should default to `True` in this API.

GemDot: yes, we set it it common to all the cases, isn't it?

python/arcticdb/version_store/_normalization.py
qc00: `MultiIndex` is just a sub-class of `Index`, but it doesn't directly have `is_monotonic_increasing` as it wouldn't make sense. Will have to drill down to the first level of a `MultiIndex`.

python/arcticdb/version_store/library.py
GemDot: yes, that should be True as well, thanks

python/arcticdb/version_store/_store.py
GemDot: it is used in 2250

python/arcticdb/version_store/_normalization.py
GemDot: better to discuss with Matt

python/tests/unit/arcticdb/version_store/test_date_range.py
GemDot: I added the test, but the check was already on place as internal error. I didn't change the error code in this case, for comp issues

python/arcticdb/version_store/_normalization.py
GemDot: The logic we have right now already works for multi index. The reason is because _index_to_records function is doing a reset index for the index levels beyond than zero. which basically, converts the index to its single version (relying only in the first level). I added a test for it

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -368,8 +368,10 @@ VersionedItem LocalVersionedEngine::update_internal(
             auto versioned_item =  write_dataframe_impl(store_,
                                                         update_info.next_version_id_,
                                                         std::move(frame),
-                                                        write_options
-                                                        );
+                                                        write_options,
+                                                        std::make_shared<DeDupMap>(),
+                                                        false,
+                                                        true);
```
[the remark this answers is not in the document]
mehertz: Can we put this in `write_options`? Does that make sense?

mehertz (approved):

python/arcticdb/version_store/_normalization.py
```diff
@@ -842,7 +860,13 @@ def normalize(self, obj, **kwargs):
         column_val = np.array(memoryview(self._buffer[: size * 8]), np.uint8).view(np.uint64)
 
         return NormalizedInput(
-            item=NPDDataFrame(index_names=[], index_values=[], column_names=["bytes"], columns_values=[column_val]),
+            item=NPDDataFrame(
+                index_names=[],
+                index_values=[],
+                column_names=["bytes"],
+                columns_values=[column_val],
+                sorted=_SortedValue.UNKNOWN,
```
alexowens90, line 868: I think this should be `UNSORTED` for pickled data?

python/tests/conftest.py
```diff
@@ -249,7 +248,9 @@ def version_store_factory(arcticdb_test_lmdb_config, lib_name):
     `name` can be a magical value "_unique_" which will create libs with unique names."""
     used = {}
 
-    def version_store(col_per_group: Optional[int] = None, row_per_segment: Optional[int] = None, **kwargs) -> NativeVersionStore:
+    def version_store(
+        col_per_group: Optional[int] = None, row_per_segment: Optional[int] = None, **kwargs
+    ) -> NativeVersionStore:
```
alexowens90, line 253: Please avoid checking in random formatting/linting changes like this in files that are otherwise untouched

python/arcticdb/version_store/_store.py
```diff
@@ -482,6 +483,8 @@ def write(
             Removes previous (non-snapshotted) versions from the database.
         pickle_on_failure: `bool`, default=False
             Pickle `data` if it can't be normalized.
+        ensure_sorted: bool, default=False
+            If True, will verify new data is sorted, but only if it contains a datetime index.
```
alexowens90, line 487: Am I right in saying:

- We are just using `is_monotonic_increasing`/`is_monotonic_decreasing` from Pandas for this verification, we are not doing any of our own checks if Pandas does not know?
- We will not reorder unsorted data, it is only a check?

If so, I think it is worth being explicit in the docstring about both of these, as well as specifying what exception will be thrown if this flag is true but unsorted data is provided.

cpp/arcticdb/version/version_store_api.hpp
```diff
@@ -92,7 +93,8 @@ class PythonVersionStore : public LocalVersionedEngine {
         const py::object &norm,
         const py::object & user_meta,
         bool upsert,
-        bool prune_previous_versions);
+        bool prune_previous_versions,
+        bool ensure_sorted);
 
     VersionedItem update(
```
alexowens90, line 99: Am I right in assuming there's no `ensure_sorted` flag for `update` because it is effectively always true?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -288,3 +289,100 @@ def test_(initial: InputFactories, append: InputFactories, match, swap, lmdb_ver
...
         lib.append("s", to_append, match=match)
+
+
+def test_append_non_sorted_exception(lmdb_version_store):
+    symbol = "bad_append"
+
+    num_initial_rows = 20
+    initial_timestamp = pd.Timestamp("2019-01-01")
+    dtidx = pd.date_range(initial_timestamp, periods=num_initial_rows)
+    df = pd.DataFrame({"c": [0 for i in range(0, num_initial_rows)]}, index=dtidx)
```
alexowens90, line 300: nit: `np.arange` produces data like this more concisely

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -468,3 +469,76 @@ def test_update_pickled_data(lmdb_version_store):
...
+    assert df.index.is_monotonic_increasing == True
+    lmdb_version_store.write(symbol, df)
+
+    symbol = "bad_write"
+    num_rows = 20
+    dtidx = pd.RangeIndex(0, num_rows, 1)
+    df = pd.DataFrame({"c": [0 for i in range(0, num_rows)]}, index=dtidx)
+    assert df.index.is_monotonic_increasing == True
+    with pytest.raises(SortingException):
+        lmdb_version_store.update(symbol, df)
```
alexowens90, line 544: Why is this a new test? `update` has always failed with non-timestamp indexed data hasn't it?

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -468,3 +469,76 @@ def test_update_pickled_data(lmdb_version_store):
     df2 = pd.DataFrame({"a": [1000]}, index=idx[1:2])
     with pytest.raises(InternalException) as e_info:
         lmdb_version_store.update(symbol, df2)
+
+
+def test_update_non_sorted_exception(lmdb_version_store):
```
alexowens90, line 474: I think we also need a test that writes some `UNSORTED` data, and then calls `update` with some sorted data, as this should fail.

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -288,3 +289,100 @@ def test_(initial: InputFactories, append: InputFactories, match, swap, lmdb_ver
     to_append, _ = append.make(abs(next_start), 1)
     with pytest.raises(NormalizationException):
         lib.append("s", to_append, match=match)
+
+
+def test_append_non_sorted_exception(lmdb_version_store):
```
alexowens90, line 294: I think we also need a test that writes some `UNSORTED` data and then calls `append` with `ensure_sorted=True`, as this should fail.

cpp/arcticdb/entity/types.hpp
```diff
@@ -653,6 +661,20 @@ struct StreamDescriptor {
         data_->mutable_index()->CopyFrom(idx.data_);
     }
 
+    void set_sorted(SortedValue sorted) {
+        if(sorted == SortedValue::UNSORTED){
```
alexowens90, line 665: Use a switch statement

python/arcticdb/version_store/_store.py
```diff
@@ -447,6 +447,7 @@ def write(
         metadata: Optional[Any] = None,
         prune_previous_version: Optional[bool] = None,
         pickle_on_failure: Optional[bool] = None,
+        ensure_sorted: bool = False,
```
alexowens90, line 450: Is `ensure_sorted` the best name for this? If `is_monotonic_decreasing` is `True` on the input data, it's definitely sorted, but we will reject anything that isn't monotonically _increasing_

alexowens90 (changes requested):

cpp/arcticdb/version/version_core.cpp
```diff
@@ -281,6 +290,8 @@ VersionedItem update_impl(
 
     auto index_segment_reader = index::get_index_reader(*(update_info.previous_index_key_), store);
     util::check_rte(!index_segment_reader.is_pickled(), "Cannot update pickled data");
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(frame.desc.get_sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING && std::holds_alternative<stream::TimeseriesIndex>(frame.index),
```
qc00, line 293: The `TimeseriesIndex` is already checked on line 298 below. You might want to move that check up.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -525,6 +536,10 @@ void check_column_and_date_range_filterable(const pipelines::index::IndexSegment
     "The data for this symbol is pickled and does not support date_range, row_range, or column queries");
     util::check(index_segment_reader.has_timestamp_index() || !std::holds_alternative<IndexRange>(read_query.row_filter),
             "Cannot apply date range filter to symbol with non-timestamp index");
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(index_segment_reader.tsd().stream_descriptor().sorted() == arcticc::pb2::descriptors_pb2::SortedValue::UNKNOWN ||
+        index_segment_reader.tsd().stream_descriptor().sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING ||
+        !std::holds_alternative<IndexRange>(read_query.row_filter),
```
qc00, line 541: The index type check is already preformed above.

You might also want to extract `index_segment_reader.tsd().stream_descriptor().sorted()` out to a variable.

python/arcticdb/version_store/_normalization.py
qc00: `UNKNOWN` is probably better considering the deserialized data could have a sorted index.

GemDot: I will leave as is now, since assuming unsorted could cause unnecessary exceptions, in case we deal with the scenario that Qi commented

python/arcticdb/version_store/_store.py
GemDot: yes. you are right with both. 

For your suggestion: regarding your second point I thing we already are explicit on that in the docstring.  Your first point is fair to add in the doctoring in a similar way we are already doing in the exception message.

cpp/arcticdb/version/version_store_api.hpp
GemDot: correct. As the error message says:  When calling update on sorted data, the new data must be sorted. ArcticDB relies on Pandas to detect if data is sorted - you can call DataFrame.index.is_monotonic_increasing on your input DataFrame to see if Pandas believes the data to be sorted.

There is no point to update otherwise considering our logic

python/tests/unit/arcticdb/version_store/test_update.py
GemDot: yes it is correct. Matt asked for this test If I don't miss remember

python/tests/unit/arcticdb/version_store/test_update.py
GemDot: Initially we were doing it, but after some discussions with Matt we moved towards explicitly check only the new data, not the existing one. I will leave this comment open, so we remember to discuss it next week

python/tests/unit/arcticdb/version_store/test_append.py
GemDot: Idem to previous comment

cpp/arcticdb/version/version_core.cpp
```diff
@@ -95,14 +96,17 @@ folly::Future<entity::AtomKey> async_write_dataframe_impl(
...
+    bool ensure_sorted
     ) {
     ARCTICDB_SAMPLE(DoWrite, 0)
 
     // Slice the frame according to the write options
     frame.set_bucketize_dynamic(options.bucketize_dynamic);
     auto slicing_arg = get_slicing_policy(options, frame);
     auto partial_key = IndexPartialKey{frame.desc.id(), version_id};
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(!ensure_sorted || frame.desc.get_sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING || !std::holds_alternative<stream::TimeseriesIndex>(frame.index),
+                "When calling write and ensure_sorted is enabled, the new data must be sorted. ArcticDB relies on Pandas to detect if data is sorted - you can call DataFrame.index.is_monotonic_increasing on your input DataFrame to see if Pandas believes the data to be sorted");
```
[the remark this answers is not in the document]
GemDot: I will explain it there instead. It was (again) after deeply discussing with Matt

cpp/arcticdb/version/local_versioned_engine.cpp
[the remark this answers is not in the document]
GemDot: As we commented 2 weeks ago, by now we will leave as is

python/arcticdb/version_store/_store.py
GemDot: Correct, but Arcticc point of view does not support currently the decreasing approach. So, as its eyes, it is not sorted. But I will leave it open to discussion for next week

cpp/arcticdb/version/version_core.cpp
```diff
@@ -95,14 +96,17 @@ folly::Future<entity::AtomKey> async_write_dataframe_impl(
...
+    bool sparsify_floats,
+    bool ensure_sorted
     ) {
     ARCTICDB_SAMPLE(DoWrite, 0)
 
     // Slice the frame according to the write options
     frame.set_bucketize_dynamic(options.bucketize_dynamic);
     auto slicing_arg = get_slicing_policy(options, frame);
     auto partial_key = IndexPartialKey{frame.desc.id(), version_id};
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(!ensure_sorted || frame.desc.get_sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING || !std::holds_alternative<stream::TimeseriesIndex>(frame.index),
```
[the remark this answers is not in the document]
GemDot: As commented above. We have to agree all together. We already discussed the names with Matt. Otherwise it is a back and forth in the PR reviews, which is not efficient.

cpp/arcticdb/version/version_core.cpp
GemDot: I disagree here. We do that check much later and only under a specific type of the row filter. On the other hand, by doing this ordering check earlier, we save unnecessary computation, in case the update is not permitted because unsorted data

cpp/arcticdb/version/version_core.cpp
GemDot: Certainly, it could be factored out and save some extra check. However, the impact of this saving would be marginal compared, at expenses of losing clarity in the code to understand what exactly we are checking here, which I think it is  important. On the other hand, we might want to change the definition of some of the checks in the future, and it could become a problem.

python/arcticdb/version_store/_normalization.py
alexowens90: We can't append or update pickled data, so if that's all this enum is for I still think it should be `UNSORTED`, but not a hill I'm willing to die on.

python/tests/unit/arcticdb/version_store/test_update.py
mehertz: `update` should fail regardless of whether the data is sorted or not. Agree this would be a good descriptive test, but given the contents of the data doesn't matter this needn't block the PR IMO.

python/tests/unit/arcticdb/version_store/test_append.py
mehertz: Is this identical to the comment about update? Update is illogical. Append isn't.

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
mehertz: Change it.

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
mehertz: I'm not concerned about the naming honestly. I think this parameter is code for `sorted as far as ArcticDB is concerned`, and given this may change in the future (we should support descending data) I don't think we should be too fine-grained with the naming. If we name it `is_monotonic_ascending` we'd have to complicate matters once it supports descending data.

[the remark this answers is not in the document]
qc00: But we cannot change what `ensure_sorted` means once this is merged as it's in the public Python API.

If we ever want to support querying descending data, we have to introduce yet another conflicting argument to the Python...

python/tests/unit/arcticdb/version_store/test_update.py
mehertz: Spoke with @GemDot. Code doesn't currently reject updates if the existing data is unsorted. Modifying it to do that and adding a test.

cpp/arcticdb/version/version_core.cpp
qc00: Under this change, if the user passes a non-Timeseries argument to `update`, s/he will get an irrelevant error message about sorting instead of the correct error message that the argument type must be Timeseries indexed.

python/arcticdb/version_store/_store.py
GemDot: validate_index_ordering

GemDot: validate_index

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
GemDot: ok

cpp/arcticdb/version/version_core.cpp
GemDot: move 299 above the new check

cpp/arcticdb/version/version_core.cpp
```diff
@@ -281,10 +291,12 @@ VersionedItem update_impl(
 
     auto index_segment_reader = index::get_index_reader(*(update_info.previous_index_key_), store);
     util::check_rte(!index_segment_reader.is_pickled(), "Cannot update pickled data");
+    util::check(index_desc.kind() == IndexDescriptor::TIMESTAMP, "Update will not work as expected with a non-timeseries index");
```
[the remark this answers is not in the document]
mehertz: This should have an error code/category.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -281,10 +291,12 @@ VersionedItem update_impl(
...
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(frame.desc.get_sorted() == arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING && std::holds_alternative<stream::TimeseriesIndex>(frame.index) && index_segment_reader.tsd().stream_descriptor().sorted() == pb2::descriptors_pb2::SortedValue::ASCENDING,
+        "When calling update on sorted data, the new data must be sorted.");
```
mehertz, line 296: `When calling update on sorted data, the input data must be sorted.`

python/arcticdb/version_store/_store.py
```diff
@@ -614,6 +624,11 @@ def append(
             not combined in any way with the metadata stored in the previous version.
         prune_previous_version
             Removes previous (non-snapshotted) versions from the database.
+        validate_index: bool, default=False
+            If True, will verify new data is sorted, but only if it contains a datetime index. ArcticDB relies on Pandas to detect if data is sorted -
```
mehertz, line 628:
```
If True, will verify that resulting symbol will support date range searches and update operations. This in effect tests that the previous version of the data and `data` are both sorted in ascending order. 
```

python/arcticdb/version_store/_store.py
```diff
@@ -991,6 +1012,11 @@ def batch_write(
             Remove previous versions from version list. Uses library default if left as None.
         pickle_on_failure : `Optional[bool]`, default=None
             Pickle results if normalization fails. Uses library default if left as None.
+        validate_index: bool, default=False
+            If True, will verify new data is sorted, but only if it contains a datetime index. ArcticDB relies on Pandas to detect if data is sorted -
+            you can call DataFrame.index.is_monotonic_increasing on your input DataFrame to see if Pandas believes the data to be sorted
```
mehertz, line 1017: Same comment as write. Mention that it applies to each entry in the batch.

python/arcticdb/version_store/_store.py
```diff
@@ -1109,12 +1143,22 @@ def batch_append(
             i-th entry corresponds to i-th element of `symbols`.
         prune_previous_version : `Optional[bool]`, default=None
             Remove previous versions from version list. Uses library default if left as None.
+        validate_index: bool, default=False
+            If True, will verify new data is sorted, but only if it contains a datetime index. ArcticDB relies on Pandas to detect if data is sorted -
```
mehertz, line 1147: Same comment as batch_write

python/arcticdb/version_store/library.py
```diff
@@ -296,6 +297,9 @@ def write(
             Removes previous (non-snapshotted) versions from the database.
         staged : bool, default=False
             Whether to write to a staging area rather than immediately to the library.
+        validate_index: bool, default=False
+            If True, will verify new data is sorted, but only if it contains a datetime index. ArcticDB relies on Pandas to detect if data is sorted -
```
mehertz, line 301: Same comment as in `_store.py`.

mehertz (changes requested):

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
GemDot: it does. category: sorting. Error code: ErrorCode. It is the same as other checks like normalization::check<ErrorCode::E_INCOMPATIBLE_INDEX>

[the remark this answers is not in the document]
mehertz: This is just a plain `util::check`. I think I must be missing how that would raise a sorting error?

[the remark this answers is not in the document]
GemDot: It is how I discussed with William and seems consistent to the rest of checks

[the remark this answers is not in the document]
GemDot: **util::check(index_desc.kind() == IndexDescriptor::TI**ME...  Ah sorry... this is not part of my change, it is another existing check that I just moved (as was requested by Qi). It is not related to the sort per see

GemDot (commented):
Addressed all the comments

GemDot (commented):
Addressed all the review comments

mehertz (approved):

PR #207, opened 2023-04-06
Fix for static schema when the updated frame has a different column g…

GemDot:


mehertz (approved):

cpp/arcticdb/pipeline/read_frame.hpp
```diff
@@ -37,19 +37,10 @@ std::optional<util::BitSet> check_and_mark_slices(
...
-        if(has_column_groups) {
-          is_first = row_ranges.insert(slice.row_range).second;
-        } else if (slice.col_range != col_range) {
-            if(!dynamic_schema)
-                is_first = slice.col_range == index_slice.col_range;
-            
-            col_range = slice.col_range;
-        }
-        row_ranges.insert(slice.row_range);
+        bool is_first = row_ranges.insert(slice.row_range).second;
```
qc00, line 43: Minor: `is_first` is only needed in the `if`

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -468,3 +468,34 @@ def test_update_pickled_data(lmdb_version_store):
     df2 = pd.DataFrame({"a": [1000]}, index=idx[1:2])
     with pytest.raises(InternalException) as e_info:
         lmdb_version_store.update(symbol, df2)
+
+
+def test_update_with_daterange_different_column_group_size(arcticdb_test_lmdb_config, lib_name):
+    local_lib_cfg = arcticdb_test_lmdb_config(lib_name)
```
qc00, line 474: If you create a library by hand from config, then you must clear it up. Otherwise, it will waste disk space in Windows builds.

To avoid this complexity, the `version_store_factory` has an argument `reuse_name` which you can set to `True` in order to recreate another library instance sharing the storage of an existing (i.e. also pass `name=...` to the `factory`) instance but with different configuration.

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -468,3 +468,34 @@ def test_update_pickled_data(lmdb_version_store):
...
+    def get_frame_for_date_range(start, end):
+        df = pd.DataFrame(index=pd.date_range(start, end, freq="h"))
+        num_columns = 45
+        for i in range(num_columns):
+            df["value_" + str(i)] = df.index.day
+        return df
+
+    df1 = get_frame_for_date_range("2010-01-01", "2023-01-01")
+    lmdb_version_store.write("test", df1)
+    info = lmdb_version_store.get_info("test")
```
qc00, line 488: Unused

qc00 (changes requested):

PR #210, opened 2023-04-10
docs: Update pieces of the documentation

jjerphan:
Hi,

Thank you for developing ArcticDB!

Here are some suggestions to improve the documentation based on my experience compiling, installing, and using ArcticDB.

I do not know if the `<figure markdown>` is needed or not (at least it seems that their path is wrong).

### DIFF
--- a/docs/mkdocs/docs/technical/architecture.md
+++ b/docs/mkdocs/docs/technical/architecture.md
@@ -8,18 +8,12 @@ Users interact with the C++ storage engine, and therefore the storage itself, vi
 
 Please note that there is **no required server** component.
 
-<figure markdown>
-  ![Image title](../../images/ArcticArchitecture.png)
-  <figcaption>ArcticDB Architecture</figcaption>
-</figure>
+![ArcticDB Architecture](../images/ArcticArchitecture.png)
 
 ## ArcticDB DataFlow
 
 The below diagram visualises the flow of data through ArcticDB from source to storage and back again:
 
-<figure markdown>
-  ![Image title](../../images/ArcticDetailedArchitecture.png)
-  <figcaption>ArcticDB Architecture</figcaption>
-</figure>
+![ArcticDB Detailed Architecture](../images/ArcticDetailedArchitecture.png)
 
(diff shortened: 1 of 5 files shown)


docs/mkdocs/docs/technical/architecture.md
```diff
@@ -8,18 +8,12 @@ Users interact with the C++ storage engine, and therefore the storage itself, vi
 
 Please note that there is **no required server** component.
 
-<figure markdown>
-  ![Image title](../../images/ArcticArchitecture.png)
-  <figcaption>ArcticDB Architecture</figcaption>
-</figure>
+![ArcticDB Architecture](../images/ArcticArchitecture.png)
```
qc00, line 11: I agree there isn't much value add from `<figure>` at the moment, especially given the captions are never changed.

@mehertz What was the original intention?

docs/mkdocs/docs/technical/contributing.md
```diff
@@ -34,7 +34,10 @@ $MY_PYTHON -m pip install -U pip setuptools wheel grpcio-tools
 ARCTIC_CMAKE_PRESET=skip $MY_PYTHON setup.py develop
 # Change the below Python_EXECUTABLE value to build against a different Python version
 cmake -DPython_EXECUTABLE=$MY_PYTHON -DTEST=off --preset linux-debug cpp
-cmake --build --preset linux-debug cpp
```
qc00, line 37: This command is indeed using argument combinations not officially supported by `cmake`. Did it not work for you?

The more portable command is

```bash
cd cpp
cmake --build --preset linux-debug
```

jjerphan: Indeed `cmake --build --preset linux-debug cpp` was not working on my machine.

```
cmake version 3.26.0
```

Do the proposed command (i.e. `cmake -S cpp --preset linux-release`) work for you or would you like to use the ones you have proposed?

qc00: `cmake` without the `--build` parameter does the configure step. We want to do the build step at this point.

qc00 (approved):

PR #213, opened 2023-04-11
Fix incorrect end index while handling non-cstyle numpy array

phoebusm:
- Changes for supporting direct access reverse c-style array's elements
- Relevant test

### DIFF
--- a/cpp/arcticdb/entity/native_tensor.hpp
+++ b/cpp/arcticdb/entity/native_tensor.hpp
@@ -107,10 +107,11 @@ struct NativeTensor {
         util::warn(strides_condition,
                    "Cannot safely ptr_cast to type of size {} when strides ({}) is not a multiple of elsize ({}) in NativeTensor with dtype {}",
                    sizeof(T), strides_[0], elsize_, data_type());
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -107,10 +107,11 @@ struct NativeTensor {
         util::warn(strides_condition,
                    "Cannot safely ptr_cast to type of size {} when strides ({}) is not a multiple of elsize ({}) in NativeTensor with dtype {}",
                    sizeof(T), strides_[0], elsize_, data_type());
+        ssize_t signed_pos = pos;
```
alexowens90, line 110: `ssize_t` is not just a signed version of `size_t`:
https://stackoverflow.com/questions/15739490/should-i-use-size-t-or-ssize-t
You probably want `int64_t` in this case

phoebusm: Nice catch! Changed to `int64_t`.

alexowens90 (approved):

PR #219, opened 2023-04-11
Add batch get descriptor in library

GemDot:
last_update_time field in SymbolDescription changed from Python datetime to Pandas Timestamp

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -249,6 +249,32 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
     return std::make_pair(version.value_or(VersionedItem{}), std::move(frame_and_descriptor));
 }
 
+std::pair<VersionedItem, std::optional<google::protobuf::Any>> LocalVersionedEngine::read_descriptor_version_internal(
+        const StreamId& stream_id,
+        const VersionQuery& version_query
+    ) {
+    ARCTICDB_SAMPLE(ReadDescriptor, 0)
+
+    auto version = get_version_to_read(stream_id, version_query);
+    version::check<ErrorCode::E_NO_SUCH_VERSION>(static_cast<bool>(version),
+                                                 "Unable to retrieve descriptor data. {}@{}: version not found", stream_id, version_query);
+
+    auto metadata_proto = store()->read_metadata(version->key_).get().second;
+    return std::pair{version.value(), metadata_proto};
+}
+
+std::vector<std::pair<VersionedItem, std::optional<google::protobuf::Any>>> LocalVersionedEngine::batch_read_descriptor_internal(
+        const std::vector<StreamId>& stream_ids,
+        const std::vector<VersionQuery>& version_queries) {
+    std::vector<folly::Future<std::pair<VersionedItem, std::optional<google::protobuf::Any>>>> results_fut;
+    for (size_t idx=0; idx < stream_ids.size(); idx++) {
+        auto version_query = version_queries.size() > idx ? version_queries[idx] : VersionQuery{};
+        results_fut.push_back(read_descriptor_version_internal(stream_ids[idx],
+                                                              version_query));
+    }
+    return folly::collect(results_fut).get();
+}
+
 void LocalVersionedEngine::flush_version_map() {
     version_map()->flush();
 }
--- a/cpp/arcticdb/version/python_bindings.cpp
+++ b/cpp/arcticdb/version/python_bindings.cpp
@@ -514,6 +514,9 @@ void register_bindings(py::module &m) {
         .def("read_descriptor",
              &PythonVersionStore::read_descriptor,
              "Get back the descriptor for a symbol.")
+        .def("batch_read_descriptor",
+             &PythonVersionStore::batch_read_descriptor,
+             "Get back the descriptor of a list of symbols.")
         .def("restore_version",
              [&](PythonVersionStore& v,  StreamId sid, const VersionQuery& version_query) {
                 auto [vit, tsd] = v.restore_version(sid, version_query);
--- a/cpp/arcticdb/version/version_store_api.cpp
+++ b/cpp/arcticdb/version/version_store_api.cpp
@@ -1039,22 +1039,36 @@ std::pair<VersionedItem, py::object> PythonVersionStore::read_descriptor(
     const StreamId& stream_id,
     const VersionQuery& version_query
     ) {
-    ARCTICDB_SAMPLE(ReadDescriptor, 0)
-
+    auto [version, metadata_proto] = read_descriptor_version_internal(stream_id, version_query);
     py::object pyobj;
-    auto version = get_version_to_read(stream_id, version_query);
-    if(!version)
-        throw NoDataFoundException(fmt::format("read_descriptor: version not found for stream", stream_id));
-
-    if (auto metadata_proto = store()->read_metadata(version->key_).get().second; metadata_proto) {
+    if (metadata_proto.has_value()) {
         arcticdb::proto::descriptors::TimeSeriesDescriptor tsd;
         metadata_proto->UnpackTo(&tsd);
         pyobj = python_util::pb_to_python(tsd);
     } else {
         pyobj = pybind11::none();
     }
+    return std::make_pair(version, pyobj);
+}
 
-    return std::pair{version.value(), pyobj};
+std::vector<std::pair<VersionedItem, py::object>> PythonVersionStore::batch_read_descriptor(
+        const std::vector<StreamId>& stream_ids,
+        const std::vector<VersionQuery>& version_queries){
+    
+    auto vector_descriptors = batch_read_descriptor_internal(stream_ids, version_queries);
+    std::vector<std::pair<VersionedItem, py::object>> output_vector;
+    for (auto [version, metadata_proto] : vector_descriptors) {
+        py::object pyobj;
+        if (metadata_proto.has_value()) {
+            arcticdb::proto::descriptors::TimeSeriesDescriptor tsd;
+            metadata_proto->UnpackTo(&tsd);
+            pyobj = python_util::pb_to_python(tsd);
+        } else {
+            pyobj = pybind11::none();
+        }
+        output_vector.push_back(std::make_pair(version, pyobj));
+    }
+    return output_vector;
 }
 
 ReadResult PythonVersionStore::read_index(
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -2192,34 +2192,7 @@ class NativeVersionStore:
     def open_mode(self):
         return self._open_mode
 
-    def get_info(self, symbol: str, as_of: Optional[VersionQueryInput] = None) -> Dict[str, Any]:
-        """
-        Returns descriptive data for `symbol`.
-
-        Parameters
-        ----------
-        symbol : `str`
-            symbol name
-        as_of : `Optional[VersionQueryInput]`, default=None
-            See documentation of `read` method for more details.
-
-        Returns
-        -------
-        `Dict[str, Any]`
-            Dictionary containing the following fields:
-
-            - col_names, `Dict`
-            - dtype, `List`
-            - rows, `int`
-            - last_update, `datetime`
-            - input_type, `str`
-            - index_type, `index_type`
-            - normalization_metadata,
-            - type, `str`
-            - date_range, `tuple`
-        """
-        version_query = self._get_version_query(as_of)
-        vit, desc = self.version_store.read_descriptor(symbol, version_query)
+    def _process_info(self, symbol: str, vit, desc, as_of: Optional[VersionQueryInput] = None) -> Dict[str, Any]:
         columns = [f.name for f in desc.stream_descriptor.fields]
         dtypes = [f.type_desc for f in desc.stream_descriptor.fields]
         index = []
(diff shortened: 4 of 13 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -249,6 +249,42 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
...
+std::pair<VersionedItem, py::object> LocalVersionedEngine::read_descriptor_version_internal(
+        const StreamId& stream_id,
+        const VersionQuery& version_query
+    ) {
+    ARCTICDB_SAMPLE(ReadDescriptor, 0)
+
+    py::object pyobj;
+    auto version = get_version_to_read(stream_id, version_query);
+    if(!version)
+        throw NoDataFoundException(fmt::format("read_descriptor: version not found for stream", stream_id));
```
mehertz, line 261: This raises with an error code, right?

Change the error message to not include the word `stream`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -249,6 +249,42 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
     return std::make_pair(version.value_or(VersionedItem{}), std::move(frame_and_descriptor));
 }
 
+std::pair<VersionedItem, py::object> LocalVersionedEngine::read_descriptor_version_internal(
```
mehertz, line 252: This method is new. What was the previous method that was used for this?

mehertz: Okay, plain `read_descriptor`

python/arcticdb/version_store/_store.py
```diff
@@ -2140,34 +2140,7 @@ def lib_cfg(self):
...
-            - last_update, `datetime`
-            - input_type, `str`
-            - index_type, `index_type`
-            - normalization_metadata,
-            - type, `str`
-            - date_range, `tuple`
-        """
-        version_query = self._get_version_query(as_of)
-        vit, desc = self.version_store.read_descriptor(symbol, version_query)
+    def process_info(self, symbol: str, vit, desc, as_of: Optional[VersionQueryInput] = None) -> Dict[str, Any]:
```
mehertz, line 2143: This is a private method and so should start with `_`

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: I tried to mimic similar structure than read_batch/read, which also share the internal_... function

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: That's exactly the same one we had previously for read_descriptor. Which is the non-parallel version, sharing now this function with the batch version. Actually, there is a bug in the error message (missing {} for the symbol)

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -249,6 +249,42 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
...
+    for (size_t idx=0; idx < stream_ids.size(); idx++) {
+        auto version_query = version_queries.size() > idx ? version_queries[idx] : VersionQuery{};
+        results_fut.push_back(read_descriptor_version_internal(stream_ids[idx],
+                                                              version_query));
+    }
+    return folly::collect(results_fut).get();
+}
+
+
+
```
mehertz, line 287: Too much whitespace

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1034,22 +1034,14 @@ std::pair<VersionedItem, py::object> PythonVersionStore::read_descriptor(
...
-        pyobj = pybind11::none();
-    }
+    return read_descriptor_version_internal(stream_id, version_query);
+}
 
-    return std::pair{version.value(), pyobj};
+std::vector<std::pair<VersionedItem, py::object>> PythonVersionStore::batch_read_descriptor(
+        const std::vector<StreamId>& stream_ids,
+        const std::vector<VersionQuery>& version_queries){
+    auto versions_and_py_objs = batch_read_descriptor_internal(stream_ids, version_queries);
```
mehertz, line 1043: `return batch_read_descriptor_internal(stream_ids, version_queries);`

python/arcticdb/version_store/_store.py
```diff
@@ -2226,6 +2199,84 @@ def get_info(self, symbol: str, as_of: Optional[VersionQueryInput] = None) -> Di
...
+            - index_type, `index_type`
+            - normalization_metadata,
+            - type, `str`
+            - date_range, `tuple`
+        """
+        version_query = self._get_version_query(as_of)
+        vit, desc = self.version_store.read_descriptor(symbol, version_query)
+        return self._process_info(symbol, vit, desc, as_of)
+
+    def get_batch_info(
```
mehertz, line 2232: `batch_get_info`

python/arcticdb/version_store/_store.py
```diff
@@ -2226,6 +2199,84 @@ def get_info(self, symbol: str, as_of: Optional[VersionQueryInput] = None) -> Di
...
+        self, symbols: List[str], as_ofs: Optional[List[VersionQueryInput]] = None
+    ) -> List[Dict[str, Any]]:
+        """
+        Returns descriptive data for a list of `symbols`.
+
+        Parameters
+        ----------
+        symbols : `str`
+            symbols: `List[str]`
+                List of symbols to read the metadata for. Elements should not be duplicated.
```
mehertz, line 2242: What does this mean? 

> Elements should not be duplicated.

Can you not return metadata for different versions of the same symbol?

python/arcticdb/version_store/library.py
```diff
@@ -1255,6 +1255,46 @@ def get_description(self, symbol: str, as_of: Optional[AsOf] = None) -> SymbolDe
             date_range=info["date_range"],
         )
 
+    def get_description_batch(
+        self, symbols: List[Union[str, ReadRequest]], as_ofs: Optional[List[AsOf]] = None
```
mehertz, line 1259: Change to use `ReadRequest` for the `as_ofs`

mehertz (changes requested):

python/arcticdb/version_store/_store.py
GemDot: Sorry, We do, yes. I will remove the comment and add the corresponding test

python/arcticdb/version_store/library.py

python/arcticdb/version_store/library.py
```diff
@@ -1288,6 +1288,75 @@ def get_description(self, symbol: str, as_of: Optional[AsOf] = None) -> SymbolDe
             date_range=info["date_range"],
         )
 
+    def get_description_batch(
+        self, symbols: List[Union[str, ReadRequest]], query_builder: Optional[QueryBuilder] = None
```
mehertz, line 1292: Remove `query_builder`. And the parameter!

python/arcticdb/version_store/library.py
```diff
@@ -1288,6 +1288,75 @@ def get_description(self, symbol: str, as_of: Optional[AsOf] = None) -> SymbolDe
...
+        """
+        Returns descriptive data for a list of ``symbols``.
+
+        Parameters
+        ----------
+        symbols : List[Union[str, ReadRequest]
+            List of symbols to read.
+            Params columns, date_range and query_builder from ReadRequest are not used
+
+        query_builder: This parameter is not used
```
mehertz, line 1303: And here!

mehertz (approved):

python/arcticdb/version_store/_store.py
```diff
@@ -2192,34 +2192,7 @@ def lib_cfg(self):
...
-            - last_update, `datetime`
-            - input_type, `str`
-            - index_type, `index_type`
-            - normalization_metadata,
-            - type, `str`
-            - date_range, `tuple`
-        """
-        version_query = self._get_version_query(as_of)
-        vit, desc = self.version_store.read_descriptor(symbol, version_query)
+    def _process_info(self, symbol: str, vit, desc, as_of: Optional[VersionQueryInput] = None) -> Dict[str, Any]:
```
qc00, line 2195: The `vit` was a poor variable name. Converting it to a parameter without type annotation makes it even more difficult to know what it represents.

When changing old code, we should improve on bad practices where possible.

python/arcticdb/version_store/library.py
```diff
@@ -1288,6 +1288,73 @@ def get_description(self, symbol: str, as_of: Optional[AsOf] = None) -> SymbolDe
...
+            as_ofs.append(None)
+
+        for s in symbols:
+            if isinstance(s, str):
+                handle_symbol(s)
+            elif isinstance(s, ReadRequest):
+                handle_read_request(s)
+            else:
+                raise ArcticInvalidApiUsageException(
+                    f"Invalid symbol type s=[{s}] type(s)=[{type(s)}]. Only [str] and [ReadRequest] are supported."
```
qc00, line 1336: Minor: I would say "Unsupported item in the symbols argument..."

`symbol type` may have other connotations.

python/arcticdb/version_store/library.py
```diff
@@ -1288,6 +1288,73 @@ def get_description(self, symbol: str, as_of: Optional[AsOf] = None) -> SymbolDe
...
+            A list of the descriptive data, whose i-th element corresponds to the i-th element of the ``symbols`` parameter.
+
+        See Also
+        --------
+        SymbolDescription
+            For documentation on each field.
+        """
+        symbol_strings = []
+        as_ofs = []
+        def handle_read_request(s_):
```
qc00, line 1315: Minor: it's unusual to have underscores at the end of variable names in Python. C++ code conventions do not apply in Python.

python/arcticdb/version_store/library.py
```diff
@@ -1288,6 +1288,73 @@ def get_description(self, symbol: str, as_of: Optional[AsOf] = None) -> SymbolDe
...
+                handle_read_request(s)
+            else:
+                raise ArcticInvalidApiUsageException(
+                    f"Invalid symbol type s=[{s}] type(s)=[{type(s)}]. Only [str] and [ReadRequest] are supported."
+                )
+
+        infos = self._nvs.batch_get_info(symbol_strings, as_ofs)
+        list_descriptions = []
+        for info in infos:
+            last_update_time = pd.to_datetime(info["last_update"]).to_pydatetime()
```
qc00, line 1342: Avoid using `to_pydatetime()` in new APIs. It will unnecessarily drop the nanoseconds

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -249,6 +249,42 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
...
+}
+
+std::vector<std::pair<VersionedItem, py::object>> LocalVersionedEngine::batch_read_descriptor_internal(
+        const std::vector<StreamId>& stream_ids,
+        const std::vector<VersionQuery>& version_queries) {
+    std::vector<folly::Future<std::pair<VersionedItem, py::object>>> results_fut;
+    for (size_t idx=0; idx < stream_ids.size(); idx++) {
+        auto version_query = version_queries.size() > idx ? version_queries[idx] : VersionQuery{};
+        results_fut.push_back(read_descriptor_version_internal(stream_ids[idx],
+                                                              version_query));
```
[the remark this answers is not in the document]
GemDot: correct. As all the cases in local engineer. It will be sort our with the folly changes we are separately doing

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -249,6 +249,42 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
...
+
+    return std::pair{version.value(), pyobj};
+}
+
+std::vector<std::pair<VersionedItem, py::object>> LocalVersionedEngine::batch_read_descriptor_internal(
+        const std::vector<StreamId>& stream_ids,
+        const std::vector<VersionQuery>& version_queries) {
+    std::vector<folly::Future<std::pair<VersionedItem, py::object>>> results_fut;
+    for (size_t idx=0; idx < stream_ids.size(); idx++) {
+        auto version_query = version_queries.size() > idx ? version_queries[idx] : VersionQuery{};
```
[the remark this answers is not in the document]
GemDot: This PR is being consistent with other methods like batch_read_internal. I don't think it is a discussion it should be here, despite I think that it is a valid discussion

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -980,6 +980,102 @@ def test_get_description(arctic_library):
...
+    lib.append("symbol2", to_append_df)
+
+    df = pd.DataFrame({"column": [1, 2, 3, 4]}, index=pd.date_range(start="1/1/2020", end="1/4/2020"))
+    df.index.rename("named_index", inplace=True)
+    lib.write("symbol3", df)
+    to_append_df = pd.DataFrame({"column": [5, 6]}, index=pd.date_range(start="1/5/2020", end="1/6/2020"))
+    to_append_df.index.rename("named_index", inplace=True)
+    lib.append("symbol3", to_append_df)
+
+    infos_multiple_version = lib.get_description_batch([ReadRequest("symbol1", as_of=0), ReadRequest("symbol2", as_of=0), ReadRequest("symbol3", as_of=0), ReadRequest("symbol1", as_of=1), ReadRequest("symbol2", as_of=1), ReadRequest("symbol3", as_of=1)])
```
qc00, line 1055: While we don't always run `black` on tests, please don't create lines longer than 120 characters.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -494,6 +494,9 @@ void register_bindings(py::module &m) {
         .def("read_descriptor",
              &PythonVersionStore::read_descriptor,
              "Get back the descriptor for a symbol.")
+        .def("batch_read_descriptor",
+             &PythonVersionStore::batch_read_descriptor,
+             "Get back the descriptor for a symbol.")
```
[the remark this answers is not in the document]
GemDot: Fixed

cpp/arcticdb/version/local_versioned_engine.cpp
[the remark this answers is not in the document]
GemDot: Addressed

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1034,22 +1034,15 @@ std::pair<VersionedItem, py::object> PythonVersionStore::read_descriptor(
...
-        throw NoDataFoundException(fmt::format("read_descriptor: version not found for stream", stream_id));
-
-    if (auto metadata_proto = store()->read_metadata(version->key_).get().second; metadata_proto) {
-        arcticdb::proto::descriptors::TimeSeriesDescriptor tsd;
-        metadata_proto->UnpackTo(&tsd);
-        pyobj = python_util::pb_to_python(tsd);
-    } else {
-        pyobj = pybind11::none();
-    }
+    auto [versioned_item, py_obj] =  read_descriptor_version_internal(stream_id, version_query);
```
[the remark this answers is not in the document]
GemDot: I mimmic the exact structure that read/read_batch operation, to be consistent. Your comment is fair, but then I believe we should change both (read/read_batch and read_descriptor/read_batch_descriptor) and maybe some other, in a separate PR

python/arcticdb/version_store/_store.py
```diff
@@ -2226,6 +2199,84 @@ def get_info(self, symbol: str, as_of: Optional[VersionQueryInput] = None) -> Di
...
+
+            - col_names, `Dict`
+            - dtype, `List`
+            - rows, `int`
+            - last_update, `datetime`
+            - input_type, `str`
+            - index_type, `index_type`
+            - normalization_metadata,
+            - type, `str`
+            - date_range, `tuple`
```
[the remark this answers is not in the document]
GemDot: I think we should leave as is, I think it is quite consistent with other APIs functions (no one refer to another)

python/arcticdb/version_store/library.py
GemDot: done

python/arcticdb/version_store/library.py
GemDot: I am going to change the normal behaviour then, since get_description also makes use of it.

python/arcticdb/version_store/_store.py
GemDot: there are literally 400 references to fit in the code...

python/arcticdb/version_store/library.py
mehertz: @GemDot you are right - but in this case I think we should change both `get_description` and the batch equivalent as part of this PR.

I don't think we should change the code in the old API however - changing types is a breaking change. Given that this API is much newer though, and given how rubbish it is to lose nanosecond resolution, I think it is reasonable to fix it in this API.

mehertz: To clarify, I think we should return a Pandas Timestamp to retain nanosecond precision.

This is a changing behaviour - please make this change clear in the PR description.

python/arcticdb/version_store/_store.py
qc00: Past bad naming is technical debt; however, continuing to use it is just sad....

PR #220, opened 2023-04-11
Add batch read metadata in library

GemDot:


python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,41 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+    lib.write_pickle("sym1", 3, {"meta1": 3})
+    lib.write_pickle("sym2", 1, {"meta2": 1})
+    lib.write_pickle("sym2", 1, {"meta2": 2})
+    lib.write_pickle("sym2", 3, {"meta2": 3})
+    # v0 and v1 should be pruned by now.
+    results_dict = lib.read_metadata_batch(["sym1", "sym2", "sym3"])
+    assert results_dict["sym1"].metadata == {"meta1": 3}
+    assert results_dict["sym2"].metadata == {"meta2": 3}
+    assert "sym3" not in results_dict
+    assert lib.read_metadata("sym1").metadata == results_dict["sym1"].metadata
```
alexowens90, line 201: This isn't really testing anything different than the previous test, as you don't try and read the metadata of the pruned versions.
It would be good to have a test for the `as_of` `kwarg` as well.

python/arcticdb/version_store/library.py
```diff
@@ -907,6 +907,35 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
         """
         return self._nvs.read_metadata(symbol, as_of)
 
+    def read_metadata_batch(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedItem:
```
qc00, line 910: Since we're introducing a new public API, we should consider addressing that problem where multiple versions of the same sample cannot be queried in batch (which will be the most efficient use of the underlying API).

One possible solution is to return a `List[VersionedItem]` in the order of the arguments.

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,41 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
         _lib = ac["pytest_test_lib"]
 
 
+@pytest.mark.parametrize("arctic_library", [["use_tombstones", "sync_passive.enabled"]], indirect=True)
+def test_read_meta_batch_with_tombstones(arctic_library):
```
qc00, line 170: The "implementation" has no logic, so probably superfluous to test.

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
GemDot: read metadata only cares about the as_of field (apart from the symbol name). So, I don't see the point to include a ReadRequest here. I would rather go for a string directly:

```
def read_metadata_batch(
        self, symbols: List[str], as_ofs: Optional[List[VersionQueryInput]] = None, **kwargs
    ) -> Dict[str, VersionedItem]:
```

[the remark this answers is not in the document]
GemDot: For your first part of the comment, totally fair

python/tests/integration/arcticdb/test_arctic.py
GemDot: All the tests are copied and adapted  from the same test of the store api: test_batch_read_meta_with_tombstones test the already existing batch_read_metadata function

python/tests/integration/arcticdb/test_arctic.py
GemDot: Removed that and improved the test

python/tests/integration/arcticdb/test_arctic.py
GemDot: Improved the tests now

python/arcticdb/version_store/library.py
GemDot: I think the outcome of this was to use another API

python/arcticdb/version_store/library.py
```diff
@@ -907,6 +907,35 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
         """
         return self._nvs.read_metadata(symbol, as_of)
 
+    def read_metadata_batch(self, symbols: List[str], as_ofs: Optional[List[AsOf]] = None) -> Dict[str, VersionedItem]:
```
poodlewars, line 910: This should return a list not a dict to allow multiple AsOf queries on a single symbol. That would be more consistent with `read_batch`. We should also have some parameters objects (something like `MetadataReadRequest`) like in `read_batch` rather than this fiddly API where the i-th element of `as_ofs` corresponds to the i-th element of `symbols`.

poodlewars (changes requested):

python/arcticdb/version_store/library.py
GemDot: That's a very fair comment. I investigated on Friday and we currently are not in a position to return it in the C++ layer with minimum changes. Therefore, this PR will have to contain the necessary C++ changes to support it. We are working now on deciding the best solution in that layer.

GemDot: We want to:

1)  Use same signature that read batch. In the doctoring we should note that we will ignore all the params than read requests, other than asofs and symbol. Add warnings!
2) Return list honouring original order

python/arcticdb/version_store/library.py
```diff
@@ -907,6 +907,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
         """
         return self._nvs.read_metadata(symbol, as_of)
 
+    def read_metadata_batch(
+        self, symbols: List[Union[str, ReadRequest]], query_builder: Optional[QueryBuilder] = None
+    ) -> List[VersionedItem]:
+        """
+        Reads multiple symbols.
```
mehertz, line 914: `Reads the metadata of multiple symbols.`

python/arcticdb/version_store/library.py
```diff
@@ -907,6 +907,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+
+        Parameters
+        ----------
+        symbols : List[Union[str, ReadRequest]
+            List of symbols to read.
+            Params columns, date_range and query_builder from ReadRequest are not used
+        
+        query_builder: This parameter is not used
```
mehertz, line 922: Why include it then?

python/arcticdb/version_store/library.py
```diff
@@ -907,6 +907,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+
+        Returns
+        -------
+        List[VersionedItem]
+            A list of the read results, whose i-th element corresponds to the i-th element of the ``symbols`` parameter.
```
mehertz, line 927: 👍

python/arcticdb/version_store/library.py
```diff
@@ -907,6 +907,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+        See Also
+        --------
+        read
+        """
+        symbol_strings = []
+        as_ofs = []
+        def handle_read_request(s_):
+            symbol_strings.append(s_.symbol)
+            if s_.date_range is not None:
+                warnings.warn("Read batch metadata makes not use of the data_range field", SyntaxWarning, stacklevel=2)
```
mehertz, line 938: `does not make use of`

python/arcticdb/version_store/library.py
```diff
@@ -907,6 +907,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+            if s_.columns is not None:
+                warnings.warn("Read batch metadata makes not use of the columns field", SyntaxWarning, stacklevel=2)
+            if s_.query_builder is not None:
+                warnings.warn("Read batch metadata makes not use of the query field", SyntaxWarning, stacklevel=2)
```
mehertz, line 942: `does not make use of`

mehertz (changes requested):

python/arcticdb/version_store/library.py
poodlewars: And similar for the `ReadRequest` - it is not the right type to use here (the giveaway is that most of its attributes are not suitable for this method).

mehertz: @poodlewars we discussed. We decided that whilst not perfect, this is preferable to creating a `ReadMetadataRequest` or having a seperate `as_of`s parameter.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -993,21 +994,43 @@ std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> Loc
...
+                }
+            }
+        }
+        auto last_version_symbol_it = latest_versions_ids->find(*stream_id);
+        if(last_version_symbol_it != latest_versions_ids->end()) {
+            auto last_version_map_key = std::make_pair(*stream_id, last_version_symbol_it->second);
+            auto last_version_it = stream_index_map->find(last_version_map_key);
+            util::check(last_version_it != stream_index_map->end(), "If we didn't specify or have not found a specific version, we must be able to index this map with the last one");
+            fut_vec.push_back(store()->read_metadata(last_version_it->second));
+        }else{
```
mehertz, line 1030: When do we hit this?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -993,21 +994,43 @@ std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> Loc
...
+                auto specific_version_it = stream_index_map->find(specific_version_map_key);
+                if (specific_version_it != stream_index_map->end()){
+                    fut_vec.push_back(store()->read_metadata(specific_version_it->second));
+                    continue;
+                }
+            }
+        }
+        auto last_version_symbol_it = latest_versions_ids->find(*stream_id);
+        if(last_version_symbol_it != latest_versions_ids->end()) {
+            auto last_version_map_key = std::make_pair(*stream_id, last_version_symbol_it->second);
```
mehertz, line 1026: Does this mean we always pull back the latest version regardless of whether a previous version has been requested?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1023,9 +1023,14 @@ std::vector<std::pair<VersionedItem, py::object>> PythonVersionStore::batch_read
 
     std::vector<std::pair<VersionedItem, py::object>> results;
     for (auto [key, meta_proto]: metadata) {
-        VersionedItem version{std::move(to_atom(key))};
-        auto res = std::make_pair(version, metadata_protobuf_to_pyobject(meta_proto));
-        results.push_back(res);
+            VersionedItem version{std::move(to_atom(key))};
+            if(meta_proto.has_value()) {
+                auto res = std::make_pair(version, metadata_protobuf_to_pyobject(meta_proto));
+                results.push_back(res);
```
mehertz, line 1029: Move this line outside of the if

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: Basically we need to track the cases where we do not specify the version (either because the vector of queries is empty or because that specific entry does not contain a specific version id). As commented above is more clear now

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: No, that was wrong (it was a confusion). It is all sorted now.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1000,26 +1001,42 @@ std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> Loc
...
-
     } else {
         specific_versions = std::make_shared<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>>();
         latest_versions = batch_get_latest_version(store(), version_map(), stream_ids, false);
     }
-    for(const auto& latest_version : *latest_versions)
-        specific_versions->try_emplace(std::make_pair(latest_version.first, latest_version.second.version_id()), latest_version.second);
 
-    return specific_versions;
+    return std::make_pair(specific_versions, latest_versions);
```
qc00, line 1009: Please remember to use `std::move()`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1000,26 +1001,42 @@ std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> Loc
...
-    auto stream_index_map = get_stream_index_map(stream_ids, version_queries);
+    auto [specific_versions_index_map, latest_versions_index_map] = get_stream_index_map(stream_ids, version_queries);
+    auto version_queries_is_empty = version_queries.empty();
     std::vector<folly::Future<std::pair<VariantKey, std::optional<google::protobuf::Any>>>> fut_vec;
-    for (const auto &[_, key] : *stream_index_map)
-        fut_vec.push_back(store()->read_metadata(key));
-
+    for(const auto& stream_id : folly::enumerate(stream_ids)) {
+        auto has_specific_version_query = version_queries_is_empty ? 
+                                          false : std::holds_alternative<SpecificVersionQuery>(version_queries[stream_id.index].content_);
```
qc00, line 1021: Using a conditional to do what can be done by a boolean expression is sad.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1000,26 +1001,42 @@ std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> Loc
...
+        if(has_specific_version_query){
+            const auto& query = version_queries[stream_id.index].content_;
+            auto specific_version_map_key = std::make_pair(*stream_id, std::get<SpecificVersionQuery>(query).version_id_);
+            auto specific_version_it = specific_versions_index_map->find(specific_version_map_key);
+            if (specific_version_it != specific_versions_index_map->end()){
+                fut_vec.push_back(store()->read_metadata(specific_version_it->second));
+            }else{
+                fut_vec.push_back(std::make_pair(VariantKey(), std::nullopt));
```
qc00, line 1029: Don't create an empty `VariantKey`. It may cause issues in other C++ code calling this function.

Wrap the entire `pair` in an `optional` or not return the `VariantKey` at all.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -986,7 +986,8 @@ std::vector<timestamp> LocalVersionedEngine::batch_get_update_times(
     return results;
 }
 
-std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> LocalVersionedEngine::get_stream_index_map(
+std::pair<std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>>, 
+std::shared_ptr<std::unordered_map<StreamId, AtomKey>>> LocalVersionedEngine::get_stream_index_map(
```
qc00, line 990: We don't have an official style guide yet, but please make such long return type more readable. For example, consider indenting like:

```cpp
std::pair<std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>>, 
          std::shared_ptr<std::unordered_map<StreamId, AtomKey>>>
LocalVersionedEngine::get_stream_index_map(
```

GemDot: done

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: In this case they are a shared pointer

python/arcticdb/version_store/_store.py
```diff
@@ -881,21 +881,33 @@ def batch_read_metadata(
             Dictionary of symbol mapping with the versioned items. The data attribute will be None.
         """
         _check_batch_kwargs(NativeVersionStore.batch_read_metadata, NativeVersionStore.read_metadata, kwargs)
-        results_dict = {}
+        meta_items = self._batch_read_meta_to_versioned_items(symbols, as_ofs, kwargs)
+        return {v.symbol: v for v in meta_items}
+
+    def _batch_read_meta_to_versioned_items(self, symbols, as_ofs, kwargs=None):
```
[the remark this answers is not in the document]
GemDot: yes, from library. I was consistent with what read batch method does (see _batch_read_to_versioned_items_)

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -300,7 +300,7 @@ class LocalVersionedEngine : public VersionedEngine {
      * @param version_queries Only explicit versions are supported at the moment. The implementation currently
      * accepts deleted versions (e.g. to support reading snapshots) and it's the caller's responsibility to verify.
      */
-    std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> get_stream_index_map(
+    std::pair<std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>>, std::shared_ptr<std::unordered_map<StreamId, VersionId>>> get_stream_index_map(
```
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: Amended now

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: improved this part

mehertz (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -986,7 +986,9 @@ std::vector<timestamp> LocalVersionedEngine::batch_get_update_times(
     return results;
 }
 
-std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>> LocalVersionedEngine::get_stream_index_map(
+std::pair<std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKey>>,
```
poodlewars, line 989: This type is very unwieldy, at the least it would benefit from some type aliases to make it more readable

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -303,8 +303,12 @@ class LocalVersionedEngine : public VersionedEngine {
      * Get the queried, if specified, otherwise the latest, versions of index keys for each specified stream.
      * @param version_queries Only explicit versions are supported at the moment. The implementation currently
      * accepts deleted versions (e.g. to support reading snapshots) and it's the caller's responsibility to verify.
+     * A pair of std unordered maps are returned. The first one contains all the Atom keys for those queries that we
```
poodlewars, line 306: Wouldn't a union type make this clearer? Ie map keys that are `pair<StreamId, Version>` and `Version = VersionId | LatestVersion`

python/arcticdb/version_store/library.py
```diff
@@ -940,6 +940,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+        self, symbols: List[Union[str, ReadRequest]], query_builder: Optional[QueryBuilder] = None
+    ) -> List[VersionedItem]:
+        """
+        Reads the metadata of multiple symbols.
+
+        Parameters
+        ----------
+        symbols : List[Union[str, ReadRequest]
+            List of symbols to read.
+            Params columns, date_range and query_builder from ReadRequest are not used
```
poodlewars, line 953: As mentioned before, this indicates that `ReadRequest` is not the right query type for this, we need a new one

python/arcticdb/version_store/library.py
```diff
@@ -940,6 +940,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+        See Also
+        --------
+        read
+        """
+        symbol_strings = []
+        as_ofs = []
+        def handle_read_request(s_):
+            symbol_strings.append(s_.symbol)
+            if s_.date_range is not None:
+                warnings.warn("Read batch metadata does not make use of of the data_range field", SyntaxWarning, stacklevel=2)
```
poodlewars, line 971: Why this `warnings` approach rather than our normal logging? How does `warnings` interact with logger configuration?

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,90 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+    lib.write_pickle("sym3", 3, {"meta3": 3}, prune_previous_versions=False)
+
+    lib.write_pickle("sym1", 1, {"meta1": 4}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 5}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 6}, prune_previous_versions=False)
+
+    lib.write_pickle("sym1", 1, {"meta1": 7}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 8}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 9}, prune_previous_versions=False)
+    # v0 and v1 should be pruned by now.
```
poodlewars, line 218: ?

python/arcticdb/version_store/library.py
```diff
@@ -940,6 +940,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+        ----------
+        symbols : List[Union[str, ReadRequest]
+            List of symbols to read.
+            Params columns, date_range and query_builder from ReadRequest are not used
+        
+        query_builder: This parameter is not used
+
+        Returns
+        -------
+        List[VersionedItem]
```
poodlewars, line 959: Should document what happens if a version is specified that does not exist

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,90 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+    assert results_dict[0].metadata == {"meta1": 1}
+    assert results_dict[1].metadata == {"meta2": 2}
+    assert results_dict[2].metadata is None
+    assert results_dict[3].metadata is None
+    assert results_dict[4].metadata == {"meta3": 3}
+
+    results_dict = lib.read_metadata_batch([ReadRequest("sym1", as_of=1), ReadRequest("sym2", as_of=1), ReadRequest("sym2", as_of=5), "sym_no_exist", ReadRequest("sym3", as_of=1)])
+    assert results_dict[0].metadata == {"meta1": 4}
+    assert results_dict[1].metadata == {"meta2": 5}
+    assert results_dict[2].metadata is None
```
poodlewars, line 241: The API isn't helping to disambiguate between "the version exists and there is no metadata" and "the version does not exist at all" (which I think is the case for "sym2" as_of=5

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,90 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+    lib.write_pickle("sym1", 1, {"meta1": 4}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 5}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 6}, prune_previous_versions=False)
+
+    lib.write_pickle("sym1", 1, {"meta1": 7}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 8}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 9}, prune_previous_versions=False)
+    # v0 and v1 should be pruned by now.
+
+    results_dict = lib.read_metadata_batch(["sym1", "sym2", "sym_no_exist" ,"sym3"])
```
poodlewars, line 220: these should not be called `results_dict`?

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,90 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
         _lib = ac["pytest_test_lib"]
 
 
+def test_read_meta_batch_with_tombstones(arctic_library):
```
poodlewars, line 169: I can't see any tests for how this behaves when there is _no_ metadata for a particular symbol?

poodlewars (dismissed):

mehertz (changes requested):

python/tests/integration/arcticdb/test_arctic.py
poodlewars: Should consider how the existing `batch_read` behaves for non-existent versions and make sure we are consistent with it.

python/arcticdb/version_store/library.py
mehertz: This has been discussed at length previously - we decided that this is a better solution then creating a seperate object for metadata methods.

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: done

python/tests/integration/arcticdb/test_arctic.py
willdealtry: I think that's the idea. We don't want to error the whole batch because a single version doesn't exist, so there are two cases where you get a None, when the Symbol/Version pair doesn't exist or when it does exist but doesn't have metadata. We could change the API but we'd need to make the case for the difference in (null) return value being useful to the end-user.

cpp/arcticdb/version/local_versioned_engine.hpp
GemDot: Not applicable in this case. Latest version is still a VersionId, but it is a special case because it represents the latest and also because we don't know in advance which VersionId is (so we wouldn't be able to index the map, unless we create a new constant type representing the meaning of "latest_id", so we can use that as a key). It also would require to change the logic of the function producing it (get_stream_index_map). I widely discussed with William about it, and we decided to leave it as these changes are now, since we want to refactor all this part eventually.

cpp/arcticdb/version/local_versioned_engine.cpp
willdealtry: I want to refactor this whole method, doing two calls to batch_read_latest then batch_read_specific is a bit nuts. What we need is a method in version_store_batch_methods.hpp that takes a list of StreamId and VersionQuery, and returns a vector of optional<AtomKey>. That way we can preserve the order (which is lost at the moment because of the two calls to get the keys depending on the latest vs specific, which removes the need for an unordered_map.

alexowens90 (changes requested):

cpp/arcticdb/async/async_store.hpp
```diff
@@ -175,7 +175,7 @@ class AsyncStore : public Store {
         return async::submit_io_task(ReadCompressedTask{key, library_, opts});
     }
 
-    folly::Future<std::pair<VariantKey, std::optional<google::protobuf::Any>>>
+    folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>>
```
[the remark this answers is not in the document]
GemDot: Qi suggested it previously

python/tests/integration/arcticdb/test_arctic.py
GemDot: bad copy paste, removed the comment now

python/arcticdb/version_store/library.py
GemDot: done

python/tests/integration/arcticdb/test_arctic.py
GemDot: oh yeah! bad copy paste again

cpp/arcticdb/async/async_store.hpp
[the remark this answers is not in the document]
willdealtry: I don't think it makes sense to have a method that returns an optional, yet the body of the method can never return std::nullopt, just to make it easier to short-circuit the method at a higher level. It's very confusing for someone who comes along and reads that method as they have no idea why it is optional in the first place. If we need to transform the pair of <AtomKey, std::optional<Any>> into a pair of optionals, we should enchain it to the future we create in the method in local_versioned_engine.cpp, like `fut_vec.push_back(store()->read_metadata(specific_version_it->second).thenValue([] (auto&& pair) { return std::make_pair(std::make_optional(std::move(pair.first), std::move(pair.second));`

python/tests/integration/arcticdb/test_arctic.py
GemDot: done the test

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,74 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+    lib.write_pickle("sym3", 3, {"meta3": 3}, prune_previous_versions=False)
+
+    lib.write_pickle("sym1", 1, {"meta1": 4}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 5}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 6}, prune_previous_versions=False)
+
+    lib.write_pickle("sym1", 1, {"meta1": 6}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 7}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 8}, prune_previous_versions=False)
+    # v0 and v1 should be pruned by now.
```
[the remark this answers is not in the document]
GemDot: I removed the comment, was a bad copy pasted

python/arcticdb/version_store/library.py
mehertz, line 971: As discussed and as highlighted by @poodlewars, please change to `log.warning`.

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,74 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+
+    results_dict = lib.read_metadata_batch(["sym1", "sym2", "sym3"])
+    assert results_dict["sym1"].metadata == {"meta1": 6}
+    assert results_dict["sym2"].metadata == {"meta2": 7}
+    assert results_dict["sym3"].metadata == {"meta3": 8}
+
+    assert lib.read_metadata("sym1").metadata == results_dict["sym1"].metadata
+    assert lib.read_metadata("sym2").metadata == results_dict["sym2"].metadata
+    assert lib.read_metadata("sym3").metadata == results_dict["sym3"].metadata
```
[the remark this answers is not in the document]

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,90 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+
+    lib.write_pickle("sym1", 1, {"meta1": 4}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 5}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 6}, prune_previous_versions=False)
+
+    lib.write_pickle("sym1", 1, {"meta1": 6}, prune_previous_versions=False)
+    lib.write_pickle("sym2", 2, {"meta2": 7}, prune_previous_versions=False)
+    lib.write_pickle("sym3", 3, {"meta3": 8}, prune_previous_versions=False)
+
+    results_dict = lib.read_metadata_batch(["sym1", "sym2", "sym_no_exist", "sym3"])
```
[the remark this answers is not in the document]
GemDot: done

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
GemDot: done

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -166,6 +166,90 @@ def test_library_management_path_prefix(moto_s3_uri_incl_bucket, boto_client):
...
+
+    results_dict = lib.read_metadata_batch(["sym1", "sym2", "sym_no_exist" ,"sym3"])
+    assert results_dict[0].metadata == {"meta1": 7}
+    assert results_dict[1].metadata == {"meta2": 8}
+    assert results_dict[2].metadata is None
+    assert results_dict[3].metadata == {"meta3": 9}
+
+    assert lib.read_metadata("sym1").metadata == results_dict[0].metadata
+    assert lib.read_metadata("sym2").metadata == results_dict[1].metadata
+    assert lib.read_metadata("sym3").metadata == results_dict[3].metadata
```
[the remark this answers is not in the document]
GemDot: done

python/arcticdb/version_store/library.py
```diff
@@ -940,6 +940,56 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
 
+    def read_metadata_batch(
+        self, symbols: List[Union[str, ReadRequest]], query_builder: Optional[QueryBuilder] = None
+    ) -> List[VersionedItem]:
+        """
+        Reads the metadata of multiple symbols.
+
+        Parameters
+        ----------
+        symbols : List[Union[str, ReadRequest]
```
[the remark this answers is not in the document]
GemDot: done

alexowens90 (approved):

python/arcticdb/version_store/library.py

python/arcticdb/version_store/library.py
poodlewars: Why? I strongly disagree, these unused fields are one of the most typical flags of incorrect API design.

python/arcticdb/version_store/library.py
```diff
@@ -14,12 +14,12 @@
...
 
 from arcticdb.version_store.processing import QueryBuilder
-from arcticdb.version_store._store import NativeVersionStore, VersionedItem
+from arcticdb.version_store._store import NativeVersionStore, VersionedItem, VersionQueryInput
 from arcticdb_ext.exceptions import ArcticException
 import pandas as pd
 import numpy as np
 import logging
-
+from arcticdb.log import version as log
```
poodlewars, line 22: err, this looks wrong to me??

python/arcticdb/version_store/library.py
```diff
@@ -201,6 +201,24 @@ class ReadRequest(NamedTuple):
     query_builder: Optional[QueryBuilder] = None
 
 
+class ReadInfoRequest(NamedTuple):
+    """ReadInfoRequest is designed to enable batching of read get descriptor / get metadata operations
```
poodlewars, line 205: `read get descriptor / get metadata operations` doesn't make sense. Say specifically which methods this is useful for.

python/arcticdb/version_store/library.py
```diff
@@ -201,6 +201,24 @@ class ReadRequest(NamedTuple):
...
 
 
+class ReadInfoRequest(NamedTuple):
+    """ReadInfoRequest is designed to enable batching of read get descriptor / get metadata operations 
+    with an API that mirrors the singular ``read`` API. Therefore, construction of this object is 
+    only required for the corresponding batch operations.
+
+    Attributes
+    ----------
+    as_of: Optional[AsOf], default=none
```
poodlewars, line 211: symbol is missing

python/arcticdb/version_store/library.py
```diff
@@ -201,6 +201,24 @@ class ReadRequest(NamedTuple):
     query_builder: Optional[QueryBuilder] = None
 
 
+class ReadInfoRequest(NamedTuple):
```
poodlewars, line 204: Could you check whether this will show up in the docs like the other one? https://docs.arcticdb.io/api/library/arcticdb.version_store.library.ReadRequest

python/arcticdb/version_store/library.py
```diff
@@ -941,6 +959,48 @@ def read_metadata(self, symbol: str, as_of: Optional[AsOf] = None) -> VersionedI
...
+        Parameters
+        ----------
+        symbols : List[Union[str, ReadInfoRequest]]
+            List of symbols to read.
+        
+        Returns
+        -------
+        List[VersionedItem]
+            A list of the read results, whose i-th element corresponds to the i-th element of the ``symbols`` parameter.
+            A None object will be returned if either that version doesn't exist, or it exists but there is no metadata
```
poodlewars, line 977: That isn't the API we put in the wiki page is it? That has `None` iff the version doesn't exist, and a `VersionedItem` with metadata=None if the version does exist

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -1052,8 +1133,8 @@ def test_get_description_batch_multiple_versions(arctic_library):
     to_append_df.index.rename("named_index", inplace=True)
     lib.append("symbol3", to_append_df)
 
-    infos_multiple_version = lib.get_description_batch([ReadRequest("symbol1", as_of=0), ReadRequest("symbol2", as_of=0), ReadRequest("symbol3", as_of=0), 
-                                                        ReadRequest("symbol1", as_of=1), ReadRequest("symbol2", as_of=1), ReadRequest("symbol3", as_of=1)])
+    infos_multiple_version = lib.get_description_batch([ReadInfoRequest("symbol1", as_of=0), ReadInfoRequest("symbol2", as_of=0), ReadInfoRequest("symbol3", as_of=0), 
+                                                        ReadInfoRequest("symbol1", as_of=1), ReadInfoRequest("symbol2", as_of=1), ReadInfoRequest("symbol3", as_of=1)])
 
     infos = infos_multiple_version[3:6]
     original_infos = infos_multiple_version[0:3]
```
poodlewars, line 1140: It would be good to add some tests to verify that `get_description_batch` follows the same semantics as `read_batch_metadata`. Ie, what happens if a symbol or a version doesn't exist?

python/arcticdb/version_store/library.py
GemDot: oh yeah, it is just the description, had not been updated

python/arcticdb/version_store/library.py
GemDot: oh yes, no longer necessary, since it was used by the warnings caused when we uses non-used fields from ReadRequest in batch read metadata

python/arcticdb/version_store/library.py
GemDot: done

python/arcticdb/version_store/library.py
GemDot: added

python/arcticdb/version_store/library.py
GemDot: Updated as:
        List[VersionedItem]
            A list of the read results, whose i-th element corresponds to the i-th element of the ``symbols`` parameter.
            A VersionedItem object with the metadata field set as None will be returned if the requested version of the 
                symbol exists but there is no metadata
            A None object will be returned if the requested version of the symbol does not exist

python/arcticdb/version_store/library.py
GemDot: As we discussed, another class field is added, common to both read_metadata_batch and read_descriptor_batch

python/tests/integration/arcticdb/test_arctic.py
GemDot: No. We are raising an exception, as I reuses the existent get_info C++ logic, which raises an exception too. But it is something we also need to address, along with the rest of methods we agreed. Good point!

GemDot: done

python/arcticdb/version_store/library.py
GemDot: ok, added the corresponding index in library.rst as discussed

mehertz (approved):

PR #226, opened 2023-04-12
Protobuf dynamic loading without relying on prior arcticdb import + compat with Man protobuf

poodlewars:
- Reduce grpcio-tools version for Proto3 generated files for compat with Python protobuf 3.8.0, to enable use in Man Group.
- Do the dynamic loading of Proto3 / Proto4 on import of arcticc.pb2, rather than relying on arcticdb having been imported first.


qc00 (approved):

python/arcticc/pb2/__init__.py
```diff
@@ -0,0 +1,13 @@
+import os as _os
+import sys as _sys
+import google.protobuf as _protobuf
+
+_proto_ver = _protobuf.__version__.split(".")[0]
+if _proto_ver in "34":
+    # Use the namespace package (https://peps.python.org/pep-0420) feature to select the pb2 files matching the protobuf
+    _protos_path = _os.path.abspath(
+        _os.path.join(_os.path.dirname(__file__), "..", "..", "arcticdb", "proto", _proto_ver, "arcticc", "pb2")
```
qc00, line 9: I am a bit worried about this. What if `arcticdb` and `arcticc` are not installed in the same directory?

For example, when someone is working on an [editable install](https://peps.python.org/pep-0662/) of `man.arcticdb` (or whatever we call it) and executed its `arcticc/pb2/__init__.py` first (which adds a non-existant path to `os.path`), will the `__init__.py` from `arcticdb`'s wheel be called again?

poodlewars: I think this will be the _only_ `arcticc/pb2` package in the codebase, `man.arcticdb` won't have one.

setup.py
```diff
@@ -16,7 +16,7 @@
 
 class CompileProto(Command):
     # When adding new protobuf versions, also update: setup.cfg, python/arcticdb/__init__.py
-    _PROTOBUF_TO_GRPC_VERSION = {"3": "<=1.48.*", "4": ">=1.49"}
+    _PROTOBUF_TO_GRPC_VERSION = {"3": "<=1.30.*", "4": ">=1.49"}
```
qc00, line 19: Can we add a Man Group detection just in case in someone else's environment there's only a version between 1.30 and 1.48?

poodlewars: I think we can cross that bridge if we come to it, no? Anyway, it would only affect _contributors_ rather than users. Users downloading the wheel would still be fine.

PR #239, opened 2023-04-18
Support UTC time-zoned indexes on modern Pandas versions

poodlewars:
Before this, the normalization layer would fail on indexes with UTC timezones.

Internal ref: AN-912


python/arcticdb/version_store/_normalization.py
```diff
@@ -39,14 +37,8 @@
 
 PICKLE_PROTOCOL = 4
 
-try:  # 0.21+ Compatibility
-    from pandas._libs.tslib import Timestamp
-    from pandas._libs.tslibs.timezones import get_timezone
-except ImportError:
```
poodlewars, line 45: I figure we can remove support for such old versions. Man doesn't use such old Pandas.

python/arcticdb/version_store/_normalization.py
```diff
@@ -39,14 +37,8 @@
...
-    try:  # 0.20.x Compatibility
-        from pandas._libs.tslib import Timestamp, get_timezone
-    except ImportError:  # <= 0.19 Compatibility
-        from pandas.tslib import Timestamp, get_timezone
+from pandas._libs.tslib import Timestamp
```
qc00, line 40: I suspect this is the same as `pd.Timestamp`?

qc00 (approved):

python/arcticdb/version_store/_normalization.py
```diff
@@ -301,11 +293,21 @@ def _normalize_single_index(index, index_names, index_norm, dynamic_strings=None
...
         if index_tz is not None:
-            index_norm.tz = index_tz
+            index_norm.tz = _ensure_str_timezone(index_tz)
 
         return index_names, ix_vals
 
 
+def _ensure_str_timezone(index_tz):
+    if is_utc(index_tz):
+        # Pandas started to treat UTC as a special case and give back a pytz object for it. We coerce it back to
```
lwfitzgerald, line 303: `pandas` also does the same for UTC tzinfo objects from other timezone libs such as `dateutil` (and [in later versions](https://github.com/jbrockmendel/pandas/blob/main/pandas/_libs/tslibs/timezones.pyx#L75), `datetime.timezones.UTC`)

```suggestion
        # Pandas started to treat UTC as a special case and give back the tzinfo object for it. We coerce it back to
```

python/arcticdb/version_store/_normalization.py
poodlewars: I should think so

python/arcticdb/version_store/_normalization.py
```diff
@@ -301,11 +293,21 @@ def _normalize_single_index(index, index_names, index_norm, dynamic_strings=None
...
 
         if index_tz is not None:
-            index_norm.tz = index_tz
+            index_norm.tz = _ensure_str_timezone(index_tz)
 
         return index_names, ix_vals
 
 
+def _ensure_str_timezone(index_tz):
+    if is_utc(index_tz):
```
lwfitzgerald, line 302: Unit tests failing here as `index_tz` is a `str` here for versions of `pandas` where `get_timezone` **does** still return a `str` for `UTC`.

```suggestion
    if not isinstance(index_tz, str) and is_utc(index_tz):
```

poodlewars: Updated with a check that `index_tz` is a `tzinfo` object to match Panda's `is_utc` strict typing.

PR #249, opened 2023-04-20
Support old Pandas (==0.22) in fix from 4638cb9

poodlewars:
`is_utc` is not importable in that version, but the workaround that we use `is_utc` in is only necessary for newer Pandas anyway. So we simply replace it with a no-op implementation for archaic Pandas.


alexowens90 (approved):

qc00:
Raised #250 to improve test coverage of this

poodlewars:
Merging - the failing test is a known flaky one and there is a ticket for it.

PR #271, opened 2023-04-24
inline error codec data

DerThorsten:
Making ErrorCodecData inline, otherwise I get a lot of duplicated symbols errors
(compiling with clang on macos)


qc00 (approved):

DerThorsten:
closing with #304

PR #278, opened 2023-04-26
Add compat exception type to bridge to the original, internal, version of this codebase and also add new universal base exception type

poodlewars:
The HEAD~ commit introduces a compatibility base exception type, similar to our old internal base type.

The HEAD commit then introduces a new, clean base exception type for all our custom C++/Python exceptions, and slightly cleans up exception translation introduced in the HEAD~ commit and some other minor feedback.

I recommend viewing commit-wise since the HEAD commit has a noisy rename from ArcticBaseException to ArcticCategorizedException.

### DIFF
--- a/cpp/arcticdb/python/python_module.cpp
+++ b/cpp/arcticdb/python/python_module.cpp
@@ -159,43 +159,7 @@ int rec_call(int i){
         return rec_call(i-1);
 }
 
-void register_exception(py::module && m) {
-
-    std::set_terminate([]{
-        auto eptr = std::current_exception();
-        try {
-            std::rethrow_exception(eptr);
-        } catch (const std::exception &e) {
-            std::string msg = fmt::format("{}({}).\n{}", arcticdb::get_type_name(typeid(e)), e.what(),
-                    arcticdb::get_stack());
-            arcticdb::log::root().error("Terminate called in thread {}: {}\n Aborting",
-                    folly::getCurrentThreadName().value_or("Unknown"), msg);
-            std::abort();
-        }
-    });
-
-    static py::exception<arcticdb::ArcticNativeCxxException> ex(m, "ArcticNativeCxxException");
-
-    py::register_exception_translator([](std::exception_ptr p) {
-        try {
-            if (p) std::rethrow_exception(p);
-        } catch (const arcticdb::ArcticNativeCxxException & e){
-            ex(e.what());
-        } catch (const py::stop_iteration &e){
-            // let stop iteration bubble up, since this is how python implements iteration termination
-            std::rethrow_exception(p);
-        } catch (const std::exception &e) {
-            std::string msg = fmt::format("{}({})", arcticdb::get_type_name(typeid(e)), e.what());
-            ex(msg.c_str());
-        }
-    });
-
-    m.def("throw_native", [](int i){
-        return rec_call(i);
-    });
-}
-
-void register_exceptions(py::module&) {
+void register_termination_handler() {
     std::set_terminate([]{
         auto eptr = std::current_exception();
         try {
@@ -208,7 +172,7 @@ void register_exceptions(py::module&) {
     });
 }
 
-void register_error_code_ecosystem(py::module& m) {
+void register_error_code_ecosystem(py::module& m, py::exception<arcticdb::ArcticException>& base_exception) {
     using namespace arcticdb;
 
     auto cat_enum = py::enum_<ErrorCategory>(m, "ErrorCategory");
@@ -227,15 +191,35 @@ void register_error_code_ecosystem(py::module& m) {
     setattr(m, "enum_value_to_prefix", enum_value_to_prefix);
     m.def("get_error_category", &get_error_category);
 
-    py::register_exception<InternalException>(m, "InternalException");
-    py::register_exception<NoSuchVersionException>(m, "NoSuchVersionException");
-    py::register_exception<NormalizationException>(m, "NormalizationException");
-    py::register_exception<MissingDataException>(m, "MissingDataException");
-    py::register_exception<SchemaException>(m, "SchemaException");
-    py::register_exception<StorageException>(m, "StorageException");
-    py::register_exception<SortingException>(m, "SortingException");
-}
+    // legacy exception base type kept for backwards compat with Man Python client
+    struct ArcticCompatibilityException : public ArcticException {};
+    auto compat_exception = py::register_exception<ArcticCompatibilityException>(
+            m, "_ArcticLegacyCompatibilityException", base_exception);
 
+    static py::exception<InternalException> internal_exception(m, "InternalException", compat_exception.ptr());
+
+    py::register_exception_translator([](std::exception_ptr p) {
+        try {
+            if (p) std::rethrow_exception(p);
+        } catch (const arcticdb::InternalException & e){
+            internal_exception(e.what());
+        } catch (const py::stop_iteration &e){
+            // let stop iteration bubble up, since this is how python implements iteration termination
+            std::rethrow_exception(p);
+        } catch (const std::exception &e) {
+            std::string msg = fmt::format("{}({})", arcticdb::get_type_name(typeid(e)), e.what());
+            internal_exception(msg.c_str());
+        }
+    });
+
+    py::register_exception<NoSuchVersionException>(m, "NoSuchVersionException");
+    py::register_exception<NormalizationException>(m, "NormalizationException", compat_exception.ptr());
+    py::register_exception<MissingDataException>(m, "MissingDataException", compat_exception.ptr());
+    py::register_exception<SchemaException>(m, "SchemaException", compat_exception.ptr());
+    py::register_exception<StorageException>(m, "StorageException", compat_exception.ptr());
+    py::register_exception<SortingException>(m, "SortingException", compat_exception.ptr());
+    py::register_exception<UnsortedDataException>(m, "UnsortedDataException", compat_exception.ptr());
+}
 
 void reinit_scheduler() {
     ARCTICDB_DEBUG(arcticdb::log::version(), "Post-fork, reinitializing the task scheduler");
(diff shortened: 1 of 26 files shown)


cpp/arcticdb/util/error_code.hpp
```diff
@@ -97,10 +98,16 @@ constexpr ErrorCategory get_error_category(ErrorCode code) {
     return static_cast<ErrorCategory>(static_cast<BaseType>(code) / error_category_scale);
 }
 
+struct ArcticMarkerException : public std::runtime_error {
```
willdealtry, line 101: What does marker mean in this context?

cpp/arcticdb/util/trace.hpp
```diff
@@ -19,6 +19,10 @@ namespace arcticdb {
 std::string get_stack();
 std::string get_type_name(const std::type_info & );
 
+/*
+ *  Backwards compatibility layer for old uses of the Python API, for which this was the base exception
+ *  type raised from the extension module.
+ */
 class ArcticNativeCxxException : public std::exception {
```
willdealtry, line 26: I'm a teeny bit sad that we have to have ArcticNativeCxx anything in ArcticDB, but I'm guessing there was no other way of doing it. 

Can we definitely not catch and rethrow in the internal wrapper?

poodlewars: Yeah. We might be able to do it in the wrapper but I actually I have no idea how, since it's really just a thing that re-writes imports. I really like the idea of keeping the wrapper as just a straight pass through because there will be less chance of breaking internal side with an ArcticDB change.

cpp/arcticdb/util/error_code.hpp
poodlewars: Hmm this is a bit of a hangover from an earlier iteration that is a bit obsolete now. Let me try and clean up a bit, thanks

cpp/arcticdb/python/python_module.cpp
```diff
@@ -227,12 +191,30 @@ void register_error_code_ecosystem(py::module& m) {
...
     m.def("get_error_category", &get_error_category);
 
-    py::register_exception<InternalException>(m, "InternalException");
-    py::register_exception<NormalizationException>(m, "NormalizationException");
-    py::register_exception<MissingDataException>(m, "MissingDataException");
-    py::register_exception<SchemaException>(m, "SchemaException");
-    py::register_exception<StorageException>(m, "StorageException");
-    py::register_exception<SortingException>(m, "SortingException");
+    // legacy exception base type kept for backwards compat
+    auto cxx_exception = py::register_exception<arcticdb::ArcticNativeCxxException>(m, "ArcticNativeCxxException");
```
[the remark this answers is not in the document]
poodlewars: Yep, see https://github.com/man-group/ArcticDB/pull/278#discussion_r1178147307

cpp/arcticdb/python/python_module.cpp
```diff
@@ -277,7 +259,7 @@ PYBIND11_MODULE(arcticdb_ext, m) {
 #endif
     // Set up the global exception handlers first, so module-specific exception handler can override it:
     auto exceptions = m.def_submodule("exceptions");
```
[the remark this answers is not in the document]
poodlewars: Nope,

```
    // Set up the global exception handlers first, so module-specific exception handler can override it:
    auto exceptions = m.def_submodule("exceptions");
    register_termination_handler();
    register_error_code_ecosystem(exceptions);
```

poodlewars:
I haven't quite thought this through properly. I've reset to draft for now.

cpp/arcticdb/python/python_module.cpp
```diff
@@ -227,12 +191,33 @@ void register_error_code_ecosystem(py::module& m) {
...
     m.def("get_error_category", &get_error_category);
 
-    py::register_exception<InternalException>(m, "InternalException");
-    py::register_exception<NormalizationException>(m, "NormalizationException");
-    py::register_exception<MissingDataException>(m, "MissingDataException");
-    py::register_exception<SchemaException>(m, "SchemaException");
-    py::register_exception<StorageException>(m, "StorageException");
-    py::register_exception<SortingException>(m, "SortingException");
+    // legacy exception base type kept for backwards compat with Man Python client
+    struct ArcticCompatibilityException : public std::runtime_error {};
```
mehertz, line 195: This is _just_ a base exception type at the end of the day. Does calling it a `CompatibilityException` clarify much more than just calling it `ArcticBaseException` or something?

qc00, line 195: Not a big fan of creating a class inside a function.

The C++ exceptions can benefit from a common, non-templated base class as well.

Also, don't use `register_exception` unless you want to see the exception type thrown in C++  as well.

poodlewars: See https://github.com/man-group/ArcticDB/pull/278#discussion_r1178147307

cpp/arcticdb/python/python_module.cpp
```diff
@@ -227,12 +191,33 @@ void register_error_code_ecosystem(py::module& m) {
...
+    auto compat_exception = py::register_exception<ArcticCompatibilityException>(
+            m, "_ArcticLegacyCompatibilityException");
+
+    py::register_exception<InternalException>(m, "InternalException", compat_exception.ptr());
```
qc00, line 199: If you read the source code of `register_exception`, you'll find it's similar to calling `py::register_exception_translator` for `InternalException`, so the first `catch` below is redundant.

cpp/arcticdb/util/trace.cpp
```diff
@@ -9,46 +9,12 @@
...
-            free(buffer);
-
-        return stack;
-#else
-        //TODO: Implement get_stack for windows
-        return std::string("");
-#endif
-    }
-
     std::string get_type_name(const std::type_info & ti){
```
qc00, line 18: Given this lacks implementations for most platforms we (are going to) support, we should switch to an [implementation ](https://github.com/facebook/folly/blob/main/folly/Demangle.cpp) from a library and rename the file?

docs/mkdocs/docs/error_messages.md
```diff
@@ -63,7 +63,7 @@ These errors relate to data being pickled, which limits the operations available
 
 Furthermore, it is not possible to partially read/update/append the data using the ArcticDB API or use the QueryBuilder with pickled symbols. 
 
-All of these errors are of type `ArcticNativeCxxException`.
+All of these errors are of type `arcticdb_ext.exceptions.InternalException`.
```
qc00, line 66: It's very sad that we're making such a big deal out of `InternalException` in the public documentation.

I would not mention the exception name at all. Users should not try to catch using the `InternalException` type because when we assign proper error codes to these, the exception type will change.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -1750,7 +1749,7 @@ def test_diff_long_stream_descriptor_mismatch(lmdb_version_store, method, num):
             lib.update("x", pd.DataFrame(bad_row, index=dr), date_range=dr)
         assert False, "should throw"
     except StreamDescriptorMismatch as e:
-        print(e)
+        assert not isinstance(e, _ArcticLegacyCompatibilityException)
```
qc00, line 1752: Wait, so `StreamDescriptorMismatch` is actually outside the class hierarchy?

I did it by accident. All ArcticDB exceptions should be catch-able via a common type.

cpp/arcticdb/util/trace.cpp
poodlewars: Not in scope

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
poodlewars: Cool. For `StreamDescriptorMismatch` we can reasonably safely put it in the hierarchy since it's newer. For `NoDataFoundException` though, I think it's too late, we could break existing programs by putting it in the hierarchy.

cpp/arcticdb/python/python_module.cpp
poodlewars: No it isn't, it's necessary to stop it being caught by the `std::exception` catch.

cpp/arcticdb/python/python_module.cpp
poodlewars: I think the use of `register_exception` is correct here. https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

cpp/arcticdb/python/python_module.cpp
qc00: But the `register_exception_translator` on line 201 is called first?

That call will throw an `InternalException` either on line 205 or 211 and be caught again by the exception handler registered on line 199, which rethrows that exception to Python.

You can, however, remove the second rethrowing overhead by adapting this example from the pybind docs:

```cpp
static py::exception<MyCustomException> exc(m, "MyCustomError");

py::register_exception_translator([](std::exception_ptr p) {
    try {
        if (p) std::rethrow_exception(p);
    } catch (const MyCustomException &e) {
        exc(e.what());
...
```

Here, the `exc` is a variable pointing at the exception class in Python. Calling that will register an exception to be thrown when switching back to Python.

cpp/arcticdb/python/python_module.cpp
poodlewars: That should never happen. It's by design to keep this here in the function and hide it as much as possible.

poodlewars: I've now added a true base exception type too.

docs/mkdocs/docs/error_messages.md
```diff
@@ -63,7 +63,7 @@ These errors relate to data being pickled, which limits the operations available
 
 Furthermore, it is not possible to partially read/update/append the data using the ArcticDB API or use the QueryBuilder with pickled symbols. 
 
-All of these errors are of type `ArcticNativeCxxException`.
+All of these errors are of type `arcticdb_ext.exceptions.ArcticException`.
```
qc00, line 66: A new thought just occurred: shall we create an alias for this in the `arcticdb.exception` module?

Just in case only one of `arcticdb-enterprise` and `arcticdb_ext` can be loaded.

qc00 (approved):

PR #281, opened 2023-04-26
Reduce usage of protobuf in stream descriptor and field encoding (WIP)

willdealtry:
Protobuf string deserialization is a source of slowness in wide dataframes, particularly in the TimeSeriesDescriptor which contains the superset of individual segment descriptors. Maintain the protobuf header as an overarching language-independent data descriptor, but move string-heavy descriptor into column objects that can be compressed as-is.


mehertz:
Recreate as draft please.

PR #282, opened 2023-04-26
Pin hypothesis to reduce disruptive test failures

poodlewars:
Master is failing due to hypothesis 6.74.0 release. We should fix the issues and relax the pin as soon as time allows.

https://github.com/man-group/ArcticDB/actions/runs/4811024777/jobs/8564968291

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,4 +48,4 @@ Testing =
     moto
     flask  # Used by moto
     flask-cors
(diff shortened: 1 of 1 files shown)


qc00:
Please add an issue to fix the broken tests and remove the restriction.

mehertz (approved):

PR #288, opened 2023-04-28
Removed unused internal function `is_indexish`

DerThorsten:
without this, I get the following error: 
```
version_functions.hpp:121:6: error: function 'is_indexish' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
```

### DIFF
--- a/cpp/arcticdb/version/version_functions.hpp
+++ b/cpp/arcticdb/version/version_functions.hpp
@@ -11,13 +11,6 @@
 #include <arcticdb/version/version_store_objects.hpp>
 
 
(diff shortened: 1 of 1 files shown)


qc00:
Clang is actually right: `is_indexish` is used as the default value, but that argument is always set to something else.

I guess better to remove the default value, which also makes the code easier to read.

DerThorsten:
> I guess better to remove the default value, which also makes the code easier to read.

agreed!

qc00 (approved):

DerThorsten:
closed by #308

PR #293, opened 2023-04-28
Exception messages and errorcode improvment

phoebusm:
- Re-organize errorcodes and error message for easier debug
- Remove `streamid` printout in the error messages
- Now python and C++ codebases share the same set of exceptions


cpp/arcticdb/util/error_code.hpp
```diff
@@ -30,19 +30,22 @@ enum class ErrorCategory : BaseType {
...
+inline decltype(auto) get_error_category_names() {
+    static const std::unordered_map<ErrorCategory, const char*> error_category_map{
         {ErrorCategory::INTERNAL, "INTERNAL"},
         {ErrorCategory::NORMALIZATION, "NORMALIZATION"},
         {ErrorCategory::MISSING_DATA, "MISSING_DATA"},
         {ErrorCategory::SCHEMA, "SCHEMA"},
         {ErrorCategory::STORAGE, "STORAGE"},
-        {ErrorCategory::SORTING, "SORTING"}
+        {ErrorCategory::SORTING, "SORTING"},
+        {ErrorCategory::USER_INPUT, "USER_INPUT"}
```
qc00, line 46: Please leave a `,` at the end where possible. Same for the `ERROR_CODES` macro below

cpp/arcticdb/util/preconditions.hpp
```diff
@@ -21,7 +21,7 @@ struct Raise {
 
     template<typename...Args>
     [[noreturn]] void operator()(fmt::format_string<Args...> format, Args&&...args) const {
-        std::string combo_format = fmt::format("{} {}", error_code_data<code>.name_, format);
+        std::string combo_format = fmt::format("{} {}:{} {}", static_cast<std::underlying_type_t<ErrorCode>>(code), get_error_category_names().at(get_error_category(code)), error_code_data<code>.name_, format);
```
qc00, line 24: The agreed format that I know of is `Ennnn: message`.

The idea is:
* The category name is encoded in the Exception name, which Python prints by default
* The `E_XXXX` mnemonic is for programmatic use only
so neither should appear in the error string

Is there a design change that I am not aware of?

qc00 (changes requested):

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -153,8 +153,12 @@ void register_bindings(py::module &m) {
         ;
 
     py::register_exception<DuplicateKeyException>(storage, "DuplicateKeyException");
-    py::register_exception<NoDataFoundException>(storage, "NoDataFoundException");
```
poodlewars, line 156: I'll have to handle this in my internal -> external bridge. Why is this being removed, and what is its replacement?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
poodlewars, line 1: These tests should pass unchanged if we are keeping backwards-compat (which we have to). Should just be adding extra assertions to check exceptions are the correct _sub_-types.

The really annoying one is `ArcticNativeNotYetImplemented` - I totally agree we should be replacing it. But it is actually used a lot in internal code and we need to be careful.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -239,7 +239,7 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
             identifier = stream_id;
         }
         else
-            throw storage::NoDataFoundException(fmt::format("read_dataframe_version: version not found for stream '{}'", stream_id));
+            throw storage::VersionNotFoundException(fmt::format("read_dataframe_version: version not found for stream"));
```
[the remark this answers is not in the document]
phoebusm: I would be very grateful if something like `renaming stream_id to symbol` is used in the ticket.
` Ensure streamid isn't revealed in the error messages`
As I am not involved in the discussion of design, I can only read the note as it is, literally...

PR #257, opened 2023-04-21
getNamedPrefix

DerThorsten:
new folly versions need the `getNamedPrefix` in  any class deriving from `folly::ThreadFactory`

PR #301, opened 2023-05-02
getNamedPrefix / fixes for new folly version 

DerThorsten:
recreation of #257 as branch instead of a fork


DerThorsten:
probably need to update folly version  for  vcpkg

DerThorsten:
closing this for now as we can just pin the used version in conda

PR #274, opened 2023-04-25
removed unused variable

DerThorsten:
removed unused variable `RowRange row_range;`

PR #306, opened 2023-05-02
Fix/remove unused var

DerThorsten:
#274 but from a branch instead of a fork

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,82 @@
+CI system guide
+===============
+
+<!--
+<tr><th></th><td></td>
+-->
+
+## [tag.yml](tag.yml)
+
+* Creates a git tag from the selected branch
+* Optionally bump the selected branch to the next version
+
+This can only be invoked manually by contributors and can require further approval using the `TestPypi`
+environment's protection rules.
+
+### Settings
+<table>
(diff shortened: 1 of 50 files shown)


qc00:
> the ci fails in the unit test are unrelated, aren't they?

Yeah, the aggregation functions behave differently from Pandas when it comes to extreme numbers. You can rerun the test job until it passes.

DerThorsten:
> 

@qc00 what about 
* https://pypi.org/project/pytest-rerunfailures/ 
* https://pypi.org/project/flaky/
* https://github.com/marketplace/actions/retry-step

All solutions seems better to me than by hand retriggering the CI until its green again

qc00:
> > 
> 
> @qc00 what about
> 
>     * https://pypi.org/project/pytest-rerunfailures/
> 
>     * https://pypi.org/project/flaky/
> 
>     * https://github.com/marketplace/actions/retry-step
> 
> 
> All solutions seem better to me, compared to retriggering the CI by hand until its green again. In particular when there are lots tiny PRs.

Those tests are failing because `hypothesis` is getting better and better at detecting numerical handling issues (e.g. #319).

I personally prefer each case to be manually inspected by the team just in case a previously unknown case is discovered.

PR #311, opened 2023-05-02
Clarify error message

phoebusm:
Replace "stream" and "stream id" with "symbol" in error message printout to reduce ambiguity

### DIFF
--- a/cpp/arcticdb/entity/types.hpp
+++ b/cpp/arcticdb/entity/types.hpp
@@ -637,14 +637,14 @@ struct StreamDescriptor {
         util::variant_match(id,
             [this] (const StringId& str) { data_->set_str_id(str); },
             [this] (const NumericId& n) {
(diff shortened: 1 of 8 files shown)


cpp/arcticdb/entity/types.hpp
```diff
@@ -637,14 +637,14 @@ struct StreamDescriptor {
         util::variant_match(id,
             [this] (const StringId& str) { data_->set_str_id(str); },
             [this] (const NumericId& n) {
-                util::check(n >= 0, "Negative NumericId is not supported");
+                util::check(n >= 0, "Negative Numeric symbol is not supported");
```
qc00, line 640: Minor: I would use a small "N"  to avoid any confusion with Numeric Investors.

The original capital "N" was part of a type name.

qc00 (approved):

PR #317, opened 2023-05-02
Fix local wheel build.

poodlewars:
It was broken before if the user had any CMake output files already present that matched the prefix, since the CMAKE_INSTALL_PREFIX would not be set to the wheel build directory properly.

This allows us to build the wheel by running,

```
pip install build wheel
python3 -m build --wheel
```


qc00:
Also created AN-967

PR #321, opened 2023-05-03
Statically link PCRE

poodlewars:
- Makes it easier to build useful wheels without using auditwheel
- No reason to think PCRE in vcpkg is less maintained than the one in manylinux CentOS

Also includes some cleanup where we assume that we _are_ using VCPKG rather than the legacy build system.

I didn't realise when I did this how easy it is to create manylinux wheels from the dev container that we use internally (just `auditwheel repair blah.whl`). Therefore I'm happy to either merge or decline this PR depending on what others think.


cpp/arcticdb/CMakeLists.txt
```diff
@@ -25,38 +25,20 @@ find_package(Gflags REQUIRED)
 find_package(LZ4 REQUIRED)
 find_package(prometheus-cpp CONFIG REQUIRED)
 find_package(Threads REQUIRED)
+find_package(unofficial-pcre REQUIRED)
+find_package(double-conversion REQUIRED)
```
qc00, line 29: Minor: the list was sorted

qc00 (approved):
All good improvements, especially since the old dev container is dead. Let's just merge regardless.

PR #333, opened 2023-05-05
build: recycle support via conda

jjerphan:
To be merged after https://github.com/man-group/ArcticDB/pull/348.


cpp/arcticdb/util/buffer_pool.hpp
```diff
@@ -9,7 +9,11 @@
 
 #include <arcticdb/util/buffer.hpp>
 
-#include <third_party/recycle/src/recycle/shared_pool.hpp>
+#ifdef ARCTICDB_USING_CONDA
+    #include <recycle/shared_pool.hpp>
+#else
+    #include <third_party/recycle/src/recycle/shared_pool.hpp>
```
qc00, line 15: Please use the include path

jjerphan: Which include paths are you thinking about? Could you do a suggestion for what you are thinking about?

jjerphan: [`ARCTICDB_USING_CONDA` has been introduced in #348 to distinguish between include paths](https://github.com/man-group/ArcticDB/pull/348/files#diff-9357e7e6906b72682dbae796b1a310682b3f7b334e3bba0cfee71c08f2e8cf57R444). 

As discussed, we should be able to modify the setup to rather have source agnostic include path, like this one:
```suggestion
    #include <recycle/shared_pool.hpp>
```

mehertz (approved):

PR #335, opened 2023-05-05
Document the hierarchy of Python exceptions.

poodlewars:


phoebusm:
It is a bit different from my design... I will amend it on my branch

phoebusm (approved):

poodlewars:
> It is a bit different from my design... I will amend it on my branch

Yep I figure we can have it as a bit of a living document. As long as release notes mention any small changes, we should be fine.

PR #341, opened 2023-05-05
Implementation for Decimal type with memory layout compatible with pyarrow's memory layout.

vasil-pashov:
* Load decimal from string
* Convert decimal to string
* Check for equality
* Add unit tests

Implementation is for 128 bit fixed point decimal. The number is stored as two little endian uint64_t variables. The format does not store the position of the decimal point.

TODO:
* Add arithmetic
* Add comparison operators

### DIFF
--- a/cpp/arcticdb/util/decimal.cpp
+++ b/cpp/arcticdb/util/decimal.cpp
@@ -1,0 +1,328 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#include <arcticdb/util/decimal.hpp>
+#include <string>
+#include <cassert>
+#include <cmath>
+#include <boost/multiprecision/cpp_int.hpp>
+#include <arcticdb/util/preprocess.hpp>
+#include <arcticdb/util/preconditions.hpp>
+
+namespace arcticdb {
+
+    namespace util {
+        // By word we refer to string of digits, not a CPU word
+        constexpr static int max_digits_per_word = std::numeric_limits<int64_t>::digits10;
+        constexpr static std::array<uint64_t, std::numeric_limits<uint64_t>::digits10 + 1> powers_of_10{
+                1,
+                10,
+                100,
+                1000,
+                10000,
+                100000,
+                1000000,
+                10000000,
+                100000000,
+                1000000000,
+                10000000000,
+                100000000000,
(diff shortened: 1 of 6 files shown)


cpp/arcticdb/util/decimal.cpp
```diff
@@ -0,0 +1,328 @@
...
+                const Decimal negated = this->operator-();
+                std::string result = negated.to_string(scale);
+                result.insert(0, "-");
+                return result;
+            }
+            if(data_[0] == 0 && data_[1] == 0) {
+                return "0";
+            }
+
+            boost::multiprecision::uint128_t number = to_uint128(
```
vasil-pashov, line 277: This also uses boost's implementation as intermediate type. Again we could change this if needed at the cost of having a bit more complicated code.

cpp/arcticdb/util/preprocess.hpp
```diff
@@ -7,11 +7,6 @@
 
 #pragma once
 
-#define stringify(x) #x
-#define static_to_string(x) stringify(x)
-#define source_loc __FILE__ ":" static_to_string(__LINE__)
-#define to_id(x) ([]() constexpr { return x; })
```
vasil-pashov, line 13: This was not used anywhere but for some reason makes the build to fail.

cpp/arcticdb/util/decimal.cpp
```diff
@@ -0,0 +1,328 @@
...
+            return digits_.data();
+        }
+
+        int NumberComponents::get_size() const {
+            return size_;
+        }
+
+        Decimal::Decimal() : data_{ 0 } {}
+
+        Decimal::Decimal(std::string_view number) : data_{0} {
```
vasil-pashov, line 235: I went with custom linear parsing, but regex might be a better idea. Any opinions?

willdealtry (changes requested):

willdealtry (approved):

cpp/arcticdb/util/decimal.cpp
```diff
@@ -0,0 +1,328 @@
...
+            }
+            if(data_[0] == 0 && data_[1] == 0) {
+                return "0";
+            }
+
+            boost::multiprecision::uint128_t number = to_uint128(
+                    data_[MOST_SIGNIFICANT_WORD_INDEX],
+                    data_[LEAST_SIGNIFICANT_WORD_INDEX]);
+
+            std::string result;
```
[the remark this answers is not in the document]
vasil-pashov: Yes, it's possible. I was thinking that for most of the cases we have the SSO will kick in. `sizeof(std::string)` is 32bytes, so about 30 chars, should fit it without allocation. If we do `result.reserve` we will be allocating always (I think).

PR #343, opened 2023-05-08
 Changes for pandas 2.0

jamesmunro:
- change version checks
- df.append has gone so use pd.concat
- df.update type normalisation has changed so test with floats

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_normalization.py
+++ b/python/tests/unit/arcticdb/version_store/test_normalization.py
@@ -127,10 +127,10 @@ def test_multiindex_with_tz(tz):
     df, norm_meta = norm.normalize(d)
     fd = FrameData.from_npd_df(df)
     denorm = norm.denormalize(fd, norm_meta)
-    if pd.__version__.startswith("1"):
-        assert_equal(d.index.to_numpy(), denorm.index.to_numpy())
-    else:
+    if pd.__version__.startswith("0"):
         assert_equal(d.index.get_values(), denorm.index.get_values())
+    else:
+        assert_equal(d.index.to_numpy(), denorm.index.to_numpy())
     denorm.index = d.index
     assert_frame_equal(d, denorm)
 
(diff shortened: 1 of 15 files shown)


jamesmunro:
There is more to do, but here is the easy stuff

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -127,10 +127,10 @@ def test_multiindex_with_tz(tz):
     df, norm_meta = norm.normalize(d)
     fd = FrameData.from_npd_df(df)
     denorm = norm.denormalize(fd, norm_meta)
-    if pd.__version__.startswith("1"):
-        assert_equal(d.index.to_numpy(), denorm.index.to_numpy())
-    else:
+    if pd.__version__.startswith("0"):
         assert_equal(d.index.get_values(), denorm.index.get_values())
+    else:
+        assert_equal(d.index.to_numpy(), denorm.index.to_numpy())
```
qc00, line 133: I wonder if `pandas.testing.assert_index_equal` existed for the `0.x` releases?

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -127,10 +127,13 @@ def test_update_repeatedly_dynamic_schema_hashed(
...
-    df = pd.DataFrame(
-        {"a": range(l), "b": range(1, l + 1), "c": range(2, l + 2), "d": range(3, l + 3), "e": range(4, l + 4)},
-        index=idx,
-    )
+    df = pd.DataFrame({
+        "a": np.arange(l, dtype='float'), 
+        "b": np.arange(1, l + 1, dtype='float'),
+        "c": np.arange(2, l + 2, dtype='float'),
+        "d": np.arange(3, l + 3, dtype='float'),
+        "e": np.arange(4, l + 4, dtype='float')
```
qc00, line 135: Also curious why `range()` stopped working.

Also, we don't have `black` set up yet, but please use double quotes.

mehertz (approved):

python/tests/unit/arcticdb/version_store/test_update.py
jamesmunro: `df.update` used to convert the the `int` columns to `float` more often than it should, that got fixed.  Switching to float here so that it works across pandas versions.

python/tests/unit/arcticdb/version_store/test_normalization.py
jamesmunro: It came in for v.0.23, I think we're still supporting v0.22 😢 
https://pandas.pydata.org/pandas-docs/version/1.0/whatsnew/v0.23.1.html?highlight=assert_index_equal

python/tests/unit/arcticdb/version_store/test_update.py
jamesmunro: I'll do the black changes

PR #351, opened 2023-05-09
Encoded field

willdealtry:
Change field encoding to reduce dependence on protobuf. Instead write timeseries and segment descriptors as columns and compress/decompress accordingly


mehertz:
Closing - think this is superceded by #401

PR #355, opened 2023-05-09
Add utility required for Mongo-backed-storage tests.

poodlewars:
We use Mongo for some of our oldest libraries internally and need to test it still works.

I split the pytest build steps by Windows/Linux since the `mongodb` service container can only run on Linux and the Github Actions schema does not allow a "conditional" service container. The steps had actually already diverged quite a lot and the duplication is actually quite minimal. I pulled out the only chunky bit of logic to a composite action.

An alternative would have been to install `mongod` in the runners each time, but having tried that `pytest-server-fixtures` still could not launch the `mongod` instance correctly. Debugging the cause would have been extremely time consuming, and we prefer to test within an unmodified manylinux image anyway, for greater fidelity with user's environments.

To note, we set `CI_MONGO_HOST` to `mongodb` as that is the name of the service container, and GHA is putting the service container on _the same_ Docker network as the test container.


qc00 (approved):

build_tooling/build_many_linux_image.sh
```diff
@@ -45,7 +45,8 @@ RUN rpmkeys --import 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x3F
 ADD sccache /usr/local/bin/
 RUN yum update -y && \
     yum install -y zip jq less devtoolset-11-gdb \
-      openssl-devel cyrus-sasl-devel devtoolset-10-libatomic-devel libcurl-devel python3-devel && \
+      openssl-devel cyrus-sasl-devel devtoolset-10-libatomic-devel libcurl-devel python3-devel \
+      mongodb-org && \
```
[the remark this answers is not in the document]
poodlewars: Hmm it's for running the Python tests

build_tooling/mongodb.repo
qc00, line 1: Let's hope mongo's mirrors don't block us as there will be ~30 installs per build

build_tooling/mongod.sh
```diff
@@ -0,0 +1,5 @@
+tooling_dir="$(dirname $BASH_SOURCE)"
+
+cp $tooling_dir/mongodb.repo /etc/yum/repos.d/
```
qc00, line 3: Build reported

    cp: cannot create regular file ‘/etc/yum/repos.d/’: Not a directory

build_tooling/mongodb.repo
poodlewars: Yeah I'm going to fiddle around with this this afternoon, maybe adding a github package.

.github/workflows/build_steps.yml
```diff
@@ -170,8 +170,8 @@ jobs:
       manylinux_image: ${{env.manylinux_image}}
       test_dirs: ${{env.test_dirs}}
 
-  python_tests:
-    if: inputs.job_type == 'follower'
+  python_tests_windows:
+    if: inputs.job_type == 'follower' && matrix.os == 'windows'
```
qc00: I think it's more trouble than it's worth to split this job by OS because we will soon have a Mac build, and the copies will start to diverge in dangerous ways.

poodlewars: There's no choice, the containerized service can only run on Linux

.github/actions/install-wheel/action.yml
```diff
@@ -0,0 +1,24 @@
+name: 'Install wheel'
+description: 'Install the wheel and dependencies'
+inputs:
+  python_impl_name:
+    required: true
+  envs:
+    required: false
+
+run: |
```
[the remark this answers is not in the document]
poodlewars: This is all work in progress, there's no need to comment on it as it goes along

.github/workflows/build_steps.yml
qc00: Ah, that's trivial to work around. Because the `services` key interprets expressions, you can do something like

```
services: ${{fromJSON(matrix.test_services)}}
```

Where `test_services` can be `null` for Windows and `{"mongo": {"image": "...."}}` for Linux.

You have to double JSON-encoded the values and put them in those giant JSON blocks in `build.yml`

qc00: I think it's rather confusing that each version's job group now has a weird skipped job in it:
image

Also, having minor differences in duplicated copies of code in an esoteric language is highly likely going to lead to maintenance issues and bugs. E.g. future maintainers will have to guess whether an inconsistency is intentional or due to carelessness. This is especially risky with tests as we can fool ourselves into believing something is being tested.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -40,6 +40,14 @@
 from tests.util.date import DateRange
 
 
+if sys.platform == "win32":
+    # leave out Mongo as spinning up a Mongo instance in Windows CI is fiddly, and Mongo support is only
+    # currently required for Linux for internal use.
+    SMOKE_TEST_VERSION_STORES = ["lmdb_version_store", "s3_version_store"]  # SKIP_WIN
```
mehertz, line 46: Very minor: We don't really use the `smoke test` terminology anywhere in the code AFAIK. I'm probably just call this `TEST_VERSION_STORES`.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -40,6 +40,14 @@
 from tests.util.date import DateRange
 
 
+if sys.platform == "win32":
```
mehertz, line 43: Can we not test if the `CI_MONGO_HOST` env var is set? If we don't, what happens if you run this locally on Linux?

.github/workflows/build_steps.yml
mehertz: Can I be a second opinion?

If they were otherwise identical I think the JSON hack is a great idea, but it looks like there's already a bunch of if statements? Like:

```
- name: Select Python (Linux)
```

```
- name: Windows Pagefile
```

```
- name: Collect crash dumps (Windows)
```

At this point it's just a question of 1 big if or many smaller ifs and I think it's probably much of a muchness as to which is "better".

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
poodlewars: No we shouldn't, since we don't insist on that being set for local runs (which work fine using pytest-server-fixtures as long as `mongod` is installed).

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
poodlewars: I prefer it as is since _every_ version store in this file is really a "test version store".

willdealtry (approved):

.github/workflows/build.yml
```diff
@@ -52,13 +52,17 @@ jobs:
       # Please declare any key you added below in build_steps.yaml:jobs.compile.strategy.matrix to aid the linting tools
       # Please update publish.yml if changing the symbols value!!
       linux_matrix: >-
-        {"os":"linux", "distro":"\"ubuntu-latest\"", "pytest_distro": "${{vars.PYTEST_DISTRO_LINUX}}",
```
qc00, line 55: Ah, this must have been confusing. I use it to divert my fork to run tests on my home cluster, so it's possible to debug the test job.

Let's remove it for now. I'll properly introduce it if we host our own runners.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -40,6 +40,14 @@
...
+    # leave out Mongo as spinning up a Mongo instance in Windows CI is fiddly, and Mongo support is only
```
qc00, line 44: Curious. I thought the Windows runner already had Mongo service installed. You just need to add a command to start it

qc00 (approved):

PR #356, opened 2023-05-09
Experiment with clang-format.

vasil-pashov:
The style is based on LLVM. With the following differences:
* AccessModifierOffset: -2 -> -4
* AlignAfterOpenBracket: Align -> DontAlign
* AllowAllArgumentsOnNextLine: true -> false
* AllowAllParametersOfDeclarationOnNextLine: true -> false
* AllowShortEnumsOnASingleLine: true -> false
* AllowShortFunctionsOnASingleLine: All -> None
* AlwaysBreakTemplateDeclarations: MultiLine -> Yes
* BinPackArguments: true -> false
* BinPackParameters: true -> false
* AllowAllConstructorInitializersOnNextLine: true -> false
* IndentWidth: 2 -> 4
* ObjCBlockIndentWidth: 2 -> 4
* ReflowComments true -> false
* SpaceAfterTemplateKeyword: true -> false
* SortIncludes: true -> false // Sorting the includes makes the build to
  fail
* ColumnLimit: 80 -> 120
* PackConstructorInitializers: BinPack -> Never
* PointerAlignment: Right -> Left
* AfterControlStatement: false -> MultiLine
*  AfterFunction: false -> true
* BreakBeforeBraces: Attach -> Custom

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -22,227 +22,263 @@ namespace arcticdb::async {
 template<class ClockType = util::SysClock>
 class AsyncStore : public Store {
 public:
-    AsyncStore(
-        std::shared_ptr<storage::Library> library,
-        const arcticdb::proto::encoding::VariantCodec &codec) :
-        library_(std::move(library)),
-        codec_(new arcticdb::proto::encoding::VariantCodec{codec}) {
+    AsyncStore(std::shared_ptr<storage::Library> library, const arcticdb::proto::encoding::VariantCodec& codec)
+        : library_(std::move(library)),
+          codec_(new arcticdb::proto::encoding::VariantCodec{codec})
+    {
     }
 
-    folly::Future<entity::VariantKey> write(
-        stream::KeyType key_type,
+    folly::Future<entity::VariantKey> write(stream::KeyType key_type,
         VersionId version_id,
         const StreamId& stream_id,
         IndexValue start_index,
         IndexValue end_index,
-        SegmentInMemory &&segment) override {
+        SegmentInMemory&& segment) override
+    {
 
         util::check(segment.descriptor().id() == stream_id,
-                    "Descriptor id mismatch in atom key {} != {}",
-                    stream_id,
-                    segment.descriptor().id());
+            "Descriptor id mismatch in atom key {} != {}",
+            stream_id,
+            segment.descriptor().id());
 
-        return async::submit_cpu_task(EncodeAtomTask{
-            key_type, version_id, stream_id, start_index, end_index, current_timestamp(),
-            std::move(segment), library_, codec_, size_t(0)
-        })
+        return async::submit_cpu_task(EncodeAtomTask{key_type,
+                                          version_id,
+                                          stream_id,
+                                          start_index,
+                                          end_index,
+                                          current_timestamp(),
+                                          std::move(segment),
+                                          library_,
+                                          codec_,
+                                          size_t(0)})
             .via(&async::io_executor())
             .thenValue(WriteSegmentTask{library_});
     }
 
-    folly::Future<entity::VariantKey> write(
-            stream::KeyType key_type,
-            VersionId version_id,
-            const StreamId& stream_id,
-            timestamp creation_ts,
-            IndexValue start_index,
-            IndexValue end_index,
-            SegmentInMemory &&segment) override {
+    folly::Future<entity::VariantKey> write(stream::KeyType key_type,
+        VersionId version_id,
+        const StreamId& stream_id,
+        timestamp creation_ts,
+        IndexValue start_index,
+        IndexValue end_index,
+        SegmentInMemory&& segment) override
+    {
 
         util::check(segment.descriptor().id() == stream_id,
-                    "Descriptor id mismatch in atom key {} != {}",
-                    stream_id,
-                    segment.descriptor().id());
+            "Descriptor id mismatch in atom key {} != {}",
+            stream_id,
+            segment.descriptor().id());
 
-        return async::submit_cpu_task(EncodeAtomTask{
-            key_type, version_id, stream_id, start_index, end_index, creation_ts,
-            std::move(segment), library_, codec_, size_t(0)
-        })
-        .via(&async::io_executor())
-        .thenValue(WriteSegmentTask{library_});
+        return async::submit_cpu_task(EncodeAtomTask{key_type,
+                                          version_id,
+                                          stream_id,
+                                          start_index,
+                                          end_index,
+                                          creation_ts,
+                                          std::move(segment),
+                                          library_,
+                                          codec_,
+                                          size_t(0)})
+            .via(&async::io_executor())
+            .thenValue(WriteSegmentTask{library_});
     }
 
-    folly::Future<VariantKey> write(PartialKey pk, SegmentInMemory &&segment) override {
+    folly::Future<VariantKey> write(PartialKey pk, SegmentInMemory&& segment) override
+    {
         return write(pk.key_type, pk.version_id, pk.stream_id, pk.start_index, pk.end_index, std::move(segment));
     }
 
-    folly::Future<entity::VariantKey> write(
-        KeyType key_type,
-        const StreamId& stream_id,
-        SegmentInMemory &&segment) override {
+    folly::Future<entity::VariantKey>
+    write(KeyType key_type, const StreamId& stream_id, SegmentInMemory&& segment) override
+    {
         util::check(is_ref_key_class(key_type), "Expected ref key type got  {}", key_type);
-        return async::submit_cpu_task(EncodeRefTask{
-            key_type, stream_id, std::move(segment), codec_
-        })
+        return async::submit_cpu_task(EncodeRefTask{key_type, stream_id, std::move(segment), codec_})
             .via(&async::io_executor())
             .thenValue(WriteSegmentTask{library_});
     }
 
-    entity::VariantKey write_sync(
-            stream::KeyType key_type,
-            VersionId version_id,
-            const StreamId& stream_id,
-            IndexValue start_index,
-            IndexValue end_index,
-            SegmentInMemory &&segment) override {
+    entity::VariantKey write_sync(stream::KeyType key_type,
+        VersionId version_id,
+        const StreamId& stream_id,
+        IndexValue start_index,
+        IndexValue end_index,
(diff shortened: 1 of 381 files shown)


jjerphan:
Thanks for making the formating uniform, @vasil-pashov.

Formatting is useful, yet it can create obfuscation.

We might want [to ignore commits in the blame view for GitHub](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) and have people [adapt their local setup with `blameignoreRevsFile` ](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt-blameignoreRevsFile).

What do you think?

cpp/arcticdb/async/python_bindings.cpp
```diff
@@ -16,15 +16,16 @@ namespace arcticdb::async {
 void register_bindings(py::module &m) {
     auto async = m.def_submodule("cpp_async", "Asynchronous processing");
 
-    py::class_<TaskScheduler, std::shared_ptr<TaskScheduler>>(async, "TaskScheduler")
```
qc00, line 19: The original line is a mere 85 cols long. Can you set the line width higher, e.g. to 120?

cpp/arcticdb/async/async_store.hpp
```diff
@@ -8,187 +8,196 @@
 #pragma once
```
qc00, line 9: Is there a way to mark specific headers as must come before others? We have a few that might be sensitive to this.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -367,67 +397,104 @@ class AsyncStore : public Store {
...
-                        (*res)[key_seg.id()] = (key_seg.atom_key());
-                        return std::make_pair(std::forward<storage::KeySegmentPair>(key_seg), true);
-                    }
-                }).thenValue([key_segs=key_segs, key_seg_mutex=key_seg_mutex](auto && pair_key_write) {
-                        if (pair_key_write.second) {
-                            std::lock_guard kl{*key_seg_mutex};
-                            ARCTICDB_DEBUG(log::version(), "Enqueuing {} for write", pair_key_write.first.atom_key());
-                            key_segs->emplace_back(std::move(pair_key_write.first));
-                        }
-                    }));
```
qc00, line 413: Below is definitely over-wrapped. We need longer lines and maybe don't enable strings to be auto-split

jjerphan:
> @jjerphan adding a `.git-blame-ignore-revs` is a great idea. I'll add one.
> 
> I didn't get this part, however:
> 
> > and have people [adapt their local setup with blameignoreRevsFile ](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt-blameignoreRevsFile).

Oh, I wanted to say that the setup is automatic on GitHub but that it isn't locally.

Locally, one needs to adapt the configuration with:

```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

cpp/arcticdb/async/async_store.hpp
```diff
@@ -8,187 +8,196 @@
...
-        stream::KeyType key_type,
-        VersionId version_id,
-        const StreamId& stream_id,
-        IndexValue start_index,
-        IndexValue end_index,
-        SegmentInMemory &&segment) override {
+    folly::Future<entity::VariantKey>
+    write(stream::KeyType key_type, VersionId version_id,
+          const StreamId &stream_id, IndexValue start_index,
+          IndexValue end_index, SegmentInMemory &&segment) override {
```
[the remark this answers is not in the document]
willdealtry: I agree. I would rather have either no wrapping or complete wrapping, the above style is just hard to read and easy to make mistakes, particularly if the wrapping ends up in different places between a source and header file.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -8,187 +8,196 @@
...
-public:
-    AsyncStore(
-        std::shared_ptr<storage::Library> library,
-        const arcticdb::proto::encoding::VariantCodec &codec) :
-        library_(std::move(library)),
-        codec_(new arcticdb::proto::encoding::VariantCodec{codec}) {
+template <class ClockType = util::SysClock> class AsyncStore : public Store {
+  public:
+    AsyncStore(std::shared_ptr<storage::Library> library,
+               const arcticdb::proto::encoding::VariantCodec &codec)
```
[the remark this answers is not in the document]
vasil-pashov: Is it though `int* a, b;`

cpp/arcticdb/async/async_store.hpp
vasil-pashov: Yep, sorting the includes breaks the build. I'll disable it.

cpp/arcticdb/async/async_store.hpp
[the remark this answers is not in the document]
qc00: I think we never use that style because of said confusion. Delcaring a reference and a non-referene variable on the same line is looking for trouble anyway.

I would consider that an unwelcome left-over from C.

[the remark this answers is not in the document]
willdealtry: Agreed, most static analyzers would flag this because of the known ambiguity

cpp/arcticdb/async/async_store.hpp
```diff
@@ -22,227 +22,263 @@ namespace arcticdb::async {
...
 public:
-    AsyncStore(
-        std::shared_ptr<storage::Library> library,
-        const arcticdb::proto::encoding::VariantCodec &codec) :
-        library_(std::move(library)),
-        codec_(new arcticdb::proto::encoding::VariantCodec{codec}) {
+    AsyncStore(std::shared_ptr<storage::Library> library, const arcticdb::proto::encoding::VariantCodec& codec)
+        : library_(std::move(library)),
+          codec_(new arcticdb::proto::encoding::VariantCodec{codec})
+    {
```
qc00, line 28: Egyptian brackets please.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -22,227 +22,263 @@ namespace arcticdb::async {
...
     }
 
-    folly::Future<entity::VariantKey> write(
-        stream::KeyType key_type,
+    folly::Future<entity::VariantKey> write(stream::KeyType key_type,
```
qc00, line 31: I prefer: once wrapping starts, the first argument should be wrapped as well.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -22,227 +22,263 @@ namespace arcticdb::async {
...
+        return async::submit_cpu_task(EncodeAtomTask{key_type,
+                                          version_id,
+                                          stream_id,
+                                          start_index,
+                                          end_index,
+                                          current_timestamp(),
+                                          std::move(segment),
+                                          library_,
+                                          codec_,
+                                          size_t(0)})
```
qc00, line 53: I really like the original two-line wrapping for this call. When I read that, my brain is able to parse the structure quickly and I can then focus on the important bits.

The long wrapping, on the other hand, robs the focus and I have to triple read to make the connection between `submit_cpu_task` and `via` below.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -22,227 +22,263 @@ namespace arcticdb::async {
...
-            IndexValue start_index,
-            IndexValue end_index,
-            SegmentInMemory &&segment) override {
+    folly::Future<entity::VariantKey> write(stream::KeyType key_type,
+        VersionId version_id,
+        const StreamId& stream_id,
+        timestamp creation_ts,
+        IndexValue start_index,
+        IndexValue end_index,
+        SegmentInMemory&& segment) override
```
qc00, line 64: When Egpyption brackets are used, the argument list needs extra indentation to offset it from the function body.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -22,227 +22,263 @@ namespace arcticdb::async {
...
-        .thenValue(UpdateSegmentTask{library_, opts});
+        return async::submit_cpu_task(EncodeSegmentTask{key, std::move(segment), library_, codec_, size_t(0)})
+            .via(&async::io_executor())
+            .thenValue(UpdateSegmentTask{library_, opts});
     }
 
-    folly::Future<VariantKey> copy(KeyType key_type, const StreamId& stream_id, VersionId version_id, const VariantKey& source_key) override {
-        return async::submit_io_task(CopyCompressedTask<ClockType>{source_key, key_type, stream_id, version_id, library_});
+    folly::Future<VariantKey>
+    copy(KeyType key_type, const StreamId& stream_id, VersionId version_id, const VariantKey& source_key) override
```
qc00, line 163: Is it possible to prefer wrapping the arguments first instead of the type?

qc00 (changes requested):

cpp/arcticdb/async/async_store.hpp
vasil-pashov: Not possible with clang format as of today.

cpp/arcticdb/async/async_store.hpp
vasil-pashov: I don't think there's a direct option to this. But you can assign some penalty value, so it prefers putting the type on the same line.

cpp/arcticdb/async/async_store.hpp
vasil-pashov: There are basically 3 things we can do:
```
void function(int param1,
    int param2,
    int param3,
    int param4) {
    //function body
}
```
Pros: most of the function will stay the same
Cons: When wrapping happens the last param and the function body are indented the same and it's kind of hard to tell where params end and where the body starts
```
void function(int param1,
    int param2,
    int param3,
    int param4)
{
    //function body
}
```
Pros: Clear separation between the end of the function params and the start of the body
Cons: Changes each function
```
void function(int param1,
                      int param2,
                      int param3,
                      int param4) {
    //function body
}
```
Pros: Provides some separation (not always) between the body and the end of the params.
Cons: (Personal opinion) looks weird.  Will change lots of functions

cpp/arcticdb/async/async_store.hpp
vasil-pashov: There are rare occasions where we don't want to enforce wrapping. If we don't enforce it it will dissolve into anarchy. On the other hand I feel like having multiple things on the same line is confusing and easy to misread.

PR #364, opened 2023-05-11
Make ArcticDB config env vars case insensitive

### LINKED ISSUE #211: Windows environment variables can only be in caps
The `ConfigsMap::get/set` methods and the type suffixes in `set_config_from_env_vars()` must be made case-insensitive to support Windows.

mehertz:
Closes #211

### DIFF
--- a/python/arcticdb/tools.py
+++ b/python/arcticdb/tools.py
@@ -14,10 +14,10 @@ from arcticdb_ext import set_config_int, set_config_string, set_config_double
 
 # Setting config from environment variables. This code is used in the package __init__.py
 _ARCTICDB_ENV_VAR_PREFIX = "ARCTICDB_"
-_TYPE_INT = "int"
-_TYPE_FLOAT = "float"
-_TYPE_STR = "str"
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/util/test/test_runtime_config.cpp
```diff
@@ -22,4 +22,19 @@ TEST(RuntimeConfig, PersistToProtobuf) {
...
     ASSERT_EQ(read_map.get_double("double_value"), 1.3);
+}
+
+TEST(RuntimeConfig, PersistToProtobufCaseInsensitive) {
+    using namespace arcticdb;
+
+    ConfigsMap test_map;
+    test_map.set_int("INT_value", 4);
+    test_map.set_string("STRING_value", "value");
+    test_map.set_double("DOUBLE_value", 1.3);
```
qc00, line 33: Minor: the actual config names we use in C++ have the format `Some.Name`. The `int_something` format is specific to that Python function (which translates it to the dot notation).

qc00 (approved):

Issue #174, opened 2023-03-29
Assume fast_tombstone_all is set when deleting data


GemDot:
When deleting data, either a `tombstone` or a `TOMBSTONE_ALL` key will be written. 

Whilst we need to respect either key type when reading data (for backwards compatibility reasons), there is no longer any reason to _not_ write `TOMBSTONE_ALL` keys. 

This change will require simplifying the flow for deleting data to remove any code branches that would execute should `fast_tombstone_all` be off.

vasil-pashov:
PR: #366

PR #366, opened 2023-05-11
Assume fast tombstone is true and fix bug in VersionMapImpl::follow_version_chain

vasil-pashov:
**Resolves**: [#174](https://github.com/man-group/ArcticDB/issues/174)

**Changes**

* Remove all branches involving fast tombstone property assuming it's always true.
* Fix a bug in follow version chain functionality.

**Bug description**
Up to now the code assumes that there are no active versions after a tombstone_all key and stops following the version chain after encountering a `tombstone_all` key type. This is not true when we use `prune_previous_versions`. Imagine the following calls:
1. Add version 0 for a symbol `sym`
2. Add version 1 for the symbol `sym`
3. Prune previous versions

This would result in the following version chain:

0. `(key_type: tombstone_all, version_id: 0)`
1. `(key_type: version, version_id: 0)`
2. `(key_type: index, version_id: 1)`
3. `(key_type: version, version_id: 1)`
5. `(key_type: index, version_id 0)`

Where we have a live version (version 1) after the `tombstone_all` key.

The bug does not appear when using `write + prune` because in that case we first add the `tombstone_all` key and after that we write the new version.

**Solution**
In case there is `tombstone_all` key, in `VersionMapImpl::follow_version_chain` we need to keep track of the oldest index which was read and stop reading only if `tombstone_all_->version_id()` is greater or equal to the `version_id` of the oldest index.


cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -322,12 +322,10 @@ TEST(VersionMap, TombstoneAllRemoveFastTombstone) {
     THREE_SIMPLE_KEYS
 
     auto version_map = std::make_shared<VersionMap>();
-    version_map->set_fast_tombstone_all(true);
     version_map->set_validate(true);
     version_map->write_and_prune_previous(store, key1, std::nullopt);
     auto maybe_prev = get_latest_version(store, version_map,  id, true, false);
     version_map->write_and_prune_previous(store, key2, maybe_prev.value());
-    version_map->set_fast_tombstone_all(false);
```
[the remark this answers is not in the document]
willdealtry: I think this one is good, however I wonder if there is another test which was previously doing the non-fast-tombstone-all version, which is now doing the exact same thing

willdealtry (approved):

cpp/arcticdb/version/test/test_version_map.cpp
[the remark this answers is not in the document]
vasil-pashov: I don't see anything similar. However, I'm not quite sure what's being tested here...

cpp/arcticdb/version/version_map.hpp
```diff
@@ -879,9 +875,13 @@ class VersionMapImpl {
                                                             std::optional<AtomKey> first_key_to_tombstone = std::nullopt,
                                                             std::shared_ptr<VersionMapEntry> entry = nullptr) {
```
qc00, line 876: I am starting to wonder if the `first_key_to_tombstone` argument is safe before this change even @willdealtry 

Calls using it basically assumes the version map has not changed without checking the timeout.

qc00 (approved):

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -217,12 +217,22 @@ inline bool load_latest_ongoing(const LoadParameter& load_params, const std::sha
...
+    if(load_params.load_type_ != LoadType::LOAD_UNDELETED) {
         return true;
+    }
 
-    ARCTICDB_DEBUG(log::version(), "Exiting because we have found an undeleted version");
-    return false;
+    if(entry->tombstone_all_) {
+        const bool is_deleted_by_tombstone_all = entry->tombstone_all_->version_id() >= oldest_loaded_index;
+        if(is_deleted_by_tombstone_all) {
+            ARCTICDB_DEBUG(log::version(), "Exiting because we have found an undeleted version");
```
willdealtry, line 228: I might add in the version that we found, as it will make it easier to debug

cpp/arcticdb/version/test/test_version_map.cpp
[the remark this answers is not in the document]
willdealtry: It's a bit less useful now, I don't mind if it stays or goes to be honest

cpp/arcticdb/version/version_map.hpp
```diff
@@ -152,17 +147,21 @@ class VersionMapImpl {
...
-        if ((load_params.load_type_ == LoadType::LOAD_LATEST || load_params.load_type_ == LoadType::LOAD_LATEST_UNDELETED ) && is_index_key_type(ref_entry.keys_[0].type())) {
+        if ((load_params.load_type_ == LoadType::LOAD_LATEST || load_params.load_type_ == LoadType::LOAD_LATEST_UNDELETED)
+            && is_index_key_type(ref_entry.keys_[0].type()))
+        {
             entry->keys_.push_back(ref_entry.keys_[0]);
         } else {
             do {
                 auto [key, seg] = store->read_sync(next_key.value());
                 std::tie(next_key, loaded_until) = read_segment_with_keys(seg, entry);
+                oldest_loaded_index_version = std::min(oldest_loaded_index_version, loaded_until);
```
willdealtry, line 160: It should be safe to assume that the new loaded_until is always less than the previously-loaded one, but I don't object to the cautious approach

willdealtry (changes requested):

cpp/arcticdb/version/version_map.hpp
vasil-pashov: That was my initial idea, but `read_segment_with_keys` returns valid version number only if an index was read. If for example a tombstone_all or tombstone is read it would return `std::numeric_limits<VersionId>::max`

cpp/arcticdb/version/version_utils.hpp
qc00, line 228: It's very annoying that `LoadType` is undocumented. My understanding of `LOAD_UNDELETED` is that it should load *all* undeleted versions.

That means the logging message is misleading: we did not exit because "an undeleted version" was found (that's the job of `LOAD_LATEST_UNDELETED`), but because we found a tombstone all key, so there couldn't be any further undeleted versions.

vasil-pashov: Done

vasil-pashov: I was about to ask about this as well. It doesn't make much sense the way it is now. Maybe it should be `Exiting because tombstone_all key deletes all versions beyond version ...`. What do you think?

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -217,12 +217,22 @@ inline bool load_latest_ongoing(const LoadParameter& load_params, const std::sha
     return false;
 }
 
-inline bool looking_for_undeleted(const LoadParameter& load_params, const std::shared_ptr<VersionMapEntry>& entry) {
-    if(!(load_params.load_type_ == LoadType::LOAD_UNDELETED && entry->tombstone_all_))
+inline bool looking_for_undeleted(const LoadParameter& load_params, const std::shared_ptr<VersionMapEntry>& entry, const VersionId& oldest_loaded_index) {
```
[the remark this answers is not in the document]
vasil-pashov: I don't get your comment.

cpp/arcticdb/version/version_map.hpp
willdealtry: Aha. Cool, good to know

cpp/arcticdb/version/version_utils.hpp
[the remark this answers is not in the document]
vasil-pashov: I see, `read_segment_with_keys` already returns the value. So I thought that doing another iteration for something that we already have is a bit wasteful.

qc00 (approved):

PR #371, opened 2023-05-12
Fix flaky test in newer pandas version

phoebusm:
In newer pandas version, values that exceed limit will be output as `np.nan` rather than `np.inf`, as in old pandas version and ArcticDB. This may lead to assertion failure in the hypothesis test.
This pull request will treat `np.nan` and `np.inf` as the same in  `test_hypothesis_mean_agg` to prevent the test from being flaky.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_aggregation.py
+++ b/python/tests/unit/arcticdb/version_store/test_aggregation.py
@@ -70,6 +70,9 @@ def test_hypothesis_sum_agg(lmdb_version_store, df):
     q = QueryBuilder()
     q = q.groupby("grouping_column").agg({"a": "sum"})
     expected = df.groupby("grouping_column").agg({"a": "sum"})
+    expected.replace(
+        np.nan, np.inf, inplace=True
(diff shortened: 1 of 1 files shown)


python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -41,6 +41,9 @@ def test_hypothesis_mean_agg(lmdb_version_store, df):
     q = QueryBuilder()
     q = q.groupby("grouping_column").agg({"a": "mean"})
     expected = df.groupby("grouping_column").agg({"a": "mean"})
+    expected.replace(
+        np.nan, np.inf, inplace=True
+    )  # New version of pandas treats values which exceeds limits as np.nan rather than np.inf, as in old version and arcticdb
```
qc00, line 46: I am happy with this change as long as `black` is happy with this very long comment

qc00 (approved):

python/tests/unit/arcticdb/version_store/test_aggregation.py
phoebusm: `black` is happy with it:)

PR #380, opened 2023-05-15
build: Remove redundant call to `std::move`

jjerphan:
This warning is emitted many times when compiling with gcc 13.1.1:
```c++
  ./ArcticDB/cpp/arcticdb/column_store/memory_segment_impl.hpp:963:40:   required from here
  ./ArcticDB/cpp/arcticdb/column_store/memory_segment_impl.hpp:968:30: warning: redundant move in initialization [-Wredundant-move]
  ./ArcticDB/cpp/arcticdb/column_store/memory_segment_impl.hpp:968:30: note: remove ‘std::move’ call
  ./ArcticDB/cpp/arcticdb/column_store/memory_segment_impl.hpp: In instantiation of ‘arcticdb::SegmentInMemoryImpl::filter(const arcticdb::util::BitSet&, bool, bool) const::<lambda(auto:116)> [with auto:116 = arcticdb::entity::TypeDescriptorTag<arcticdb::entity::DataTypeTag<arcticdb::entity::DataType::ASCII_DYNAMIC64>, arcticdb::entity::DimensionTag<arcticdb::entity::Dimension::Dim2> >]’:
```

This naively implements the changes which have been suggested by GCC. 

Since `filter_bitset` is an lvalue reference, it can't be cast to an rvalue reference and thus can't be moved. It seems that a copy is necessary to initialize `final_bitset` without a change on the reference type of `filter_bitset`.


cpp/arcticdb/column_store/memory_segment_impl.hpp
```diff
@@ -965,7 +965,7 @@ class SegmentInMemoryImpl {
                 using ColumnTagType = typename TypeDescriptorTag::DataTypeTag;
                 using RawType = typename ColumnTagType::raw_type;
 
-                util::BitSet final_bitset = std::move(filter_bitset);
```
vasil-pashov, line 968: This move makes me wonder if the param was supposed to be lvalue instead of const ref.

qc00: Yes, I have fixed this once somewhere. Maybe it's stuck on a branch. I'll dig it out.

Please don't merge this as is since it will cover up the genuine issue that the bitset is being copied by mistake.

qc00: Would you mind porting the internal change to ArcticDB? Otherwise, we will lose it when we switch.

PR #382, opened 2023-05-15
Export compile commands fom CMake

vasil-pashov:
Set CMAKE_EXPORT_COMPILE_COMMANDS to ON so that CMake exports compiler commands. This makes it easier to work on the codebase using Vim and the YouCompleteMe plugin.


cpp/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ project(arcticdb VERSION 0.0.1)
 
 enable_testing()
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
qc00, line 9: Could just inspect the ninja/make files directly?

vasil-pashov: For ninja you have compdb which can extract it, for make you have to install [python package](https://pypi.org/project/compiledb/). On the other hand this will work the same way regardless of the build system.

qc00:
```vim
Ctrl-C
ESC
Ctrl-Alt-Delete
```

qc00 (approved):

cpp/CMakeLists.txt
vasil-pashov: > Ctrl-C
ESC
Ctrl-Alt-Delete

I don't get this.

qc00: Just a joke on how difficult it is to quit Vim :)

PR #388, opened 2023-05-16
Update OoO checks should consider if previous old data was not storin…

GemDot:
…g sortness


mehertz:
So previously if the data had an unknown sorting value we'd still compare?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -293,7 +293,7 @@ VersionedItem update_impl(
     util::check_rte(!index_segment_reader.is_pickled(), "Cannot update pickled data");
     auto index_desc = check_index_match(frame.index, index_segment_reader.tsd().stream_descriptor().index());
     util::check(index_desc.kind() == IndexDescriptor::TIMESTAMP, "Update not supported for non-timeseries indexes");
-    sorting::check<ErrorCode::E_UNSORTED_DATA>(frame.desc.get_sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING && std::holds_alternative<stream::TimeseriesIndex>(frame.index) && index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING,
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(frame.desc.get_sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING && std::holds_alternative<stream::TimeseriesIndex>(frame.index) && (index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING || index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::UNKNOWN),
```
qc00, line 296: This line is way too long to be read-able.

poodlewars (changes requested):

poodlewars:
> 

Yeah, which prevents update of data for which the sorting value is not known. We should only blow up with the error when we're _sure_ that something nonsensical is being attempted. We found this as it causes issues with some internal custom normalizers.

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
GemDot: As commented, this is a different case. I believe it is better to discuss it offline.

mehertz (changes requested):
Leave a detailed description including an example of the previously acceptable and undesireable behaviour please.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -292,7 +292,11 @@ VersionedItem update_impl(
     util::check_rte(!index_segment_reader.is_pickled(), "Cannot update pickled data");
     auto index_desc = check_index_match(frame.index, index_segment_reader.tsd().stream_descriptor().index());
     util::check(index_desc.kind() == IndexDescriptor::TIMESTAMP, "Update not supported for non-timeseries indexes");
-    sorting::check<ErrorCode::E_UNSORTED_DATA>(frame.desc.get_sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING && std::holds_alternative<stream::TimeseriesIndex>(frame.index) && index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING,
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(
+        (frame.desc.get_sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING || frame.desc.get_sorted() == arcticdb::proto::descriptors::SortedValue::UNKNOWN) && 
+        std::holds_alternative<stream::TimeseriesIndex>(frame.index) && 
+        (index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING || 
+            index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::UNKNOWN),
```
qc00, line 299: I thought we agreed to switch to a state-machine-based approach?

If we're not there yet, the signal-to-noise ratio of these lines is still too low and the business logic is obscured.

Consider:
1. Extract the core decision logic to a free function with simple argument names like `existing_frame_sorted`, `input_frame_sorted`...
2. You can then use `if` statements to better specify the layers of decisions involved instead of having to cramp them all in one boolean expression.
3. `using SortedValue = arcticdb::proto::descriptors::SortedValue` in that function so to avoid the noisy prefixes
4. Do the complex object access (`index_segment_reader.mutable_tsd().stream_descriptor().sorted()`) on the calling side of that function

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -450,7 +450,7 @@ def _create_product_candles_df(arr):
     assert_frame_equal(after_arctic, before_arctic)
 
 
-def test_update_non_sorted_exception(lmdb_version_store):
+def test_update_new_non_sorted_exception(lmdb_version_store):
```
qc00, line 453: What does "new" mean in this context?

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -494,14 +494,59 @@ def test_update_existing_non_sorted_exception(lmdb_version_store):
...
+    num_rows = 20
+    initial_timestamp = pd.Timestamp("2020-01-01")
+    dtidx = reversed(pd.date_range(initial_timestamp, periods=num_rows))
+    df2 = pd.DataFrame({"c": np.arange(0, num_rows, dtype=np.int64)}, index=dtidx)
+    assert df2.index.is_monotonic_increasing == False
+    assert df2.index.is_monotonic_decreasing == True
+
+    with pytest.raises(SortingException):
+        lmdb_version_store.update(symbol, df2)
+
```
qc00, line 519: Two new lines after a free function per `black`

cpp/arcticdb/version/version_core.cpp
GemDot: even though, such function already exists, despite now it is only used by the append. Since initially, we wanted to force the update only work with sorted values, it is why in that case the state machine of the update is reduced to "resulting data always sorted", making that function unnecessary to be used there. Now, we will be able to use that. However, it lives in the write logic, so it has to be moved first to a free function. As Alex said, we decided to postpone it, in order to merge this PR ASAP.

python/tests/unit/arcticdb/version_store/test_update.py
GemDot: fair

python/tests/unit/arcticdb/version_store/test_update.py
qc00: Try `test_update_with_unsorted_input_throws`?

cpp/arcticdb/version/version_core.cpp
qc00: I think sacrificing code quality to meet arbitrary deadlines is bad. Doing so is creating technical debt which will ultimately add work for ourselves and may reduce our ability to react quickly in a real crisis.

poodlewars: Yeah we didn't agree to do the absolute bare minimum here, we simply agreed to postpone the full refactor to the state machine. Cleaning this up with a function is a totally reasonable thing to do.

GemDot: yes, sorry. I was only referring to the state machine for the update (regarding only your question). The extraction of the code into a function is something that already was done yesterday. I agree it leaves the change cleaner.

python/tests/unit/arcticdb/version_store/test_update.py
GemDot: well otherwise the test would fail isn't it

poodlewars:
@GemDot I've just merged https://github.com/man-group/ArcticDB/pull/395 and it includes a failing test case (skipped at the moment) to test unknown sortedness. Please remove the pytest skip in there as part of this change.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -277,6 +277,17 @@ VersionedItem delete_range_impl(
...
 
+void sorted_data_check(InputTensorFrame& frame, index::IndexSegmentReader& index_segment_reader){
+    bool input_data_is_sorted = frame.desc.get_sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING || 
+                                frame.desc.get_sorted() == arcticdb::proto::descriptors::SortedValue::UNKNOWN;
+    bool existing_data_is_sorted = index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::ASCENDING || 
+                                    index_segment_reader.mutable_tsd().stream_descriptor().sorted() == arcticdb::proto::descriptors::SortedValue::UNKNOWN;
+    bool is_time_series = std::holds_alternative<stream::TimeseriesIndex>(frame.index);
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(
+        input_data_is_sorted && existing_data_is_sorted && is_time_series,
+        "When calling update on sorted data, the input data must be sorted.");
```
mehertz, line 288: As @jamesmunro has suggested, let's split this into three seperate checks.

mehertz (changes requested):

cpp/arcticdb/version/version_core.cpp
mehertz, line 288: This is such an improvement we should do the same for append as well.

mehertz (changes requested):

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -450,7 +450,7 @@ def _create_product_candles_df(arr):
     assert_frame_equal(after_arctic, before_arctic)
 
 
-def test_update_non_sorted_exception(lmdb_version_store):
+def test_update_input_non_sorted_exception(lmdb_version_store):
```
mehertz, line 453: `not_sorted` not `non_sorted`

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -494,14 +495,62 @@ def test_update_existing_non_sorted_exception(lmdb_version_store):
...
+    initial_timestamp = pd.Timestamp("2020-01-01")
+    dtidx = pd.date_range(initial_timestamp, periods=num_rows)
+    df2 = pd.DataFrame({"c": np.arange(0, num_rows, dtype=np.int64)}, index=dtidx)
+    assert df2.index.is_monotonic_increasing == True
+
+    with pytest.raises(SortingException):
+        lmdb_version_store.update(symbol, df2)
+
+
+def test_update_non_sorted_input_multi_index_exception(lmdb_version_store):
```
mehertz, line 547: `not_sorted`

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -514,7 +563,24 @@ def test_update_non_sorted_multi_index_exception(lmdb_version_store):
...
+        index=pd.MultiIndex.from_arrays([dtidx1, dtidx2], names=["datetime", "level"]),
+    )
+    assert isinstance(df.index, MultiIndex) == True
+    assert df.index.is_monotonic_increasing == False
+
+    with pytest.raises(SortingException):
+        lmdb_version_store.update(symbol, df)
+
+
+def test_update_non_sorted_existing_multi_index_exception(lmdb_version_store):
```
mehertz, line 577: `not_sorted`

cpp/arcticdb/version/version_core.cpp
GemDot: done!

GemDot: done!

python/tests/unit/arcticdb/version_store/test_update.py
GemDot: there are plenty of examples in the tests where we use "non"...test_partial_write_non_contiguous, non_sorted_arr_1, for example

python/tests/unit/arcticdb/version_store/test_update.py
GemDot: I changed the name

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -230,3 +230,40 @@ void fix_stream_ids_of_index_keys(
...
+                final_state = SortingState::ASCENDING;
+            }
+            break;
+        case SortingState::DESCENDING:
+            if(input_frame  == SortingState::UNKNOWN){
+                final_state = SortingState::UNKNOWN;
+            }else if (input_frame  != SortingState::DESCENDING){
+                final_state = SortingState::UNSORTED;
+            }else{
+                final_state = SortingState::DESCENDING;
```
qc00, line 261: I guess this only matters to `append()`, but does it actually check the input is less than the existing index?

I.e. would it allow appending `[9, 8, 7]` to `[3, 2, 1]`?

GemDot: that's a good point

cpp/arcticdb/pipeline/input_tensor_frame.hpp
```diff
@@ -45,7 +45,12 @@ struct InputTensorFrame {
     }
 
     void set_sorted(SortedValue sorted) {
-        desc.set_sorted(sorted);
+        switch (sorted) {
+            case SortedValue::UNSORTED:desc.set_sorted(arcticc::pb2::descriptors_pb2::SortedValue::UNSORTED);break;
+            case SortedValue::DESCENDING:desc.set_sorted(arcticc::pb2::descriptors_pb2::SortedValue::DESCENDING);break;
+            case SortedValue::ASCENDING:desc.set_sorted(arcticc::pb2::descriptors_pb2::SortedValue::ASCENDING);break;
+            default:desc.set_sorted(arcticc::pb2::descriptors_pb2::SortedValue::UNKNOWN);
```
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/version_core.cpp
```diff
@@ -277,6 +288,23 @@ VersionedItem delete_range_impl(
     return versioned_item;
 }
 
+void sorted_data_check_update(InputTensorFrame& frame, index::IndexSegmentReader& index_segment_reader){
+    bool is_time_series = std::holds_alternative<stream::TimeseriesIndex>(frame.index);
+    sorting::check<ErrorCode::E_UNSORTED_DATA>(
+        is_time_series,
+        "When calling update on sorted data, the input data must be a time series.");
```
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/version_utils.hpp
GemDot: The answer is no. We do not allow it, but not as part of all the sorting exception new work. It is something was already controlled in arcticDB. I consider this as resolved

arcticdb_ext.exceptions.InternalException: E_ASSERTION_FAILURE Can't append dataframe with start index 2019-01-01 00:00:00.0 to existing sequence ending at 2020-01-20 00:00:00.0

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
qc00: It's still not changed?

qc00 (approved):

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -230,3 +230,40 @@ void fix_stream_ids_of_index_keys(
     const StreamId &stream_id,
     const std::shared_ptr<VersionMapEntry> &entry);
 }
+
+inline arcticdb::proto::descriptors::SortedValue deduce_sorted(arcticdb::proto::descriptors::SortedValue existing_frame, arcticdb::proto::descriptors::SortedValue input_frame){
```
poodlewars, line 234: Isn't this switch equivalent to,

```
if (input_frame == existing_frame) {
  return input_frame;
}

if ((input_frame == UNKNOWN || existing_frame == UNKNOWN) 
  && (input_frame != UNSORTED || existing_frame != UNSORTED)) {
  return UNKNOWN;
}

return UNSORTED;
```

poodlewars (changes requested):

cpp/arcticdb/version/version_utils.hpp
GemDot: yes agreed. But I didn't want to do any logic simplification here. Because for 2 reasons: 

1. We want the state machine to be as  more clear as possible.
2. Improbable, but if in the future we want to do changes to this state machine, that could need more changes in the logic, rather that just changing some values in the current state machine logic.

poodlewars (approved):

mehertz (approved):

cpp/arcticdb/version/version_utils.hpp
GemDot: Agree with point 2. But still not convinced for 1.

cpp/arcticdb/version/test/test_sorting_info_state_machine.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <arcticdb/util/bitset.hpp>
+
+#include <msgpack.hpp>
+#include <arcticdb/util/test/generators.hpp>
+#include <arcticdb/version/version_store_api.hpp>
+
+
+TEST(Sorting, ExistingUnknown) {
+    using namespace arcticdb;
+    using namespace arcticdb::proto;
```
qc00, line 22: You can put these two outside the test body (at the top of the file) so they don't need to be repeated.

Similarly, consider `using SortedValue = proto::descriptors::SortedValue` at the top of the file, instead of `descriptors::` everywhere.

I would go further by declaring `constexpr UNKNOWN = proto::descriptors::SortedValue::UNKNOWN;` at the top for all four enum items, so they don't need to be qualified in this file.

cpp/arcticdb/version/test/test_sorting_info_state_machine.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    descriptors::SortedValue existing_frame = descriptors::SortedValue::UNKNOWN;
+
+    descriptors::SortedValue input_frame = descriptors::SortedValue::UNKNOWN;
+    ASSERT_EQ(deduce_sorted(existing_frame, input_frame), descriptors::SortedValue::UNKNOWN);
```
qc00, line 27: It will be easier to follow if both parameters are in-line instead of variables, so the assertion line more closely resembles a table.

Actually, the traditional way of implementing a state machine is a look-up table which wouldn't require any tests.

cpp/arcticdb/version/test/test_sorting_info_state_machine.cpp
GemDot: I agree with all. Pushed all the changes

cpp/arcticdb/version/test/test_sorting_info_state_machine.cpp
qc00: Consider this:

```cpp
    //               existing_frame, input_frame
    ASSERT_EQ(deduce_sorted(UNKNOWN, UNKNOWN   ), UNKNOWN);
    ASSERT_EQ(deduce_sorted(UNKNOWN, ASCENDING ), UNKNOWN);
    ASSERT_EQ(deduce_sorted(UNKNOWN, DESCENDING), UNKNOWN);
    ASSERT_EQ(deduce_sorted(UNKNOWN, UNSORTED  ), UNSORTED);
```

I would say that it's equally clear what the common argument is, but also much less visual noise.

GemDot: yes, I was thinking in exactly the same option, the only caviat is that it needs the comment bo be added...but fine

GemDot: done

PR #389, opened 2023-05-16
build: Add a public version identifier

### LINKED ISSUE #110: Python package arcticdb should have a __version__ special attribute
```
In [1]: import arcticdb

In [2]: arcticdb.__version__
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 arcticdb.__version__

AttributeError: module 'arcticdb' has no attribute '__version__'
```

We are missing this conventional feature. Compare with some popular packages,

```
[ins] In [1]: import pandas

[ins] In [2]: pandas.__version__
Out[2]: '1.1.5'

[ins] In [3]: import numpy

[ins] In [4]: numpy.__version__
Out[4]: '1.19.5'
```

It might be good to wait till the Github actions work is merged, as that will affect our version tagging.

jjerphan:
Resolves #110.

The proposed identifier is based on the next version following semantic versioning and includes a development release segment as specified by PEP 440.

See: https://peps.python.org/pep-0440/#public-version-identifiers

---

If this PR gets accepted, the release process will need to adapt to follow at least those steps to:
 - commit the public version identifier not to include a development release segment
 - tag the commit of this change with the value of the public version identifier
 - create a release on GitHub
 - publish on PyPi and conda-forge
 - create a new commit bumping the version identifier with a development release segment

### DIFF
--- a/python/arcticdb/__init__.py
+++ b/python/arcticdb/__init__.py
@@ -11,3 +11,4 @@ from arcticdb.tools import set_config_from_env_vars
 
 set_config_from_env_vars(_os.environ)
 
+__version__ = "1.1.1.dev0"


mehertz:
Our PyPi release process already has a mechanism to update the version in `setup.cfg`:

https://github.com/man-group/ArcticDB/blob/master/.github/workflows/tag.yml#L27

I think we'd want to modify that to also update this version for the branch so we can't release a version with an incorrect `__version__`.

mehertz:
Is there a way to get this at runtime from the package metadata with Conda?

jjerphan:
> How is this normally handled for Conda packages/releases? CI-powered sed like above?

The version for the release on conda-forge has to be provided via the feedstock. There's no check against the one in `pyproject.toml`, in `setup.cfg`, or `__version__`, but there's one against the release of the sources on GitHub to have consistency (assuming the versions set in the aforementioned files and in `__version__` are correct).

For instance, see [how the feedstock of scikit-learn manages its version](https://github.com/jjerphan/scikit-learn-feedstock/blob/main/recipe/meta.yaml#L1-L9).

willdealtry (approved):

jjerphan:
@qc00: should we wait for your PR first?

Issue #365, opened 2023-05-11
Improve handling of exceptions during VersionMap load 


qc00:
## Background

Before the introduction of tombstones, it was possible to have a break in the version linked list if an older version was deleted.

To work around this, `iterate_on_failure` was introduced to load all version keys of a symbol using the `iterate_type` call.

Any symbol created since the introduction of tombstones (which includes all symbols written by the public release) will not benefit from the `iterate_on_failure` behaviour. A lot of new code paths are setting it to `false`. However:

## The weakness

`storage_reload` swallows all storage exceptions if `iterate_on_failure=false` so it can return partial or wrong results.

The exception is only logged at debug and the bad result will be cached in the `VersionMap` and maybe even in the `SymbolList`!!

## Fixes to consider

1. Treat read exception differently from missing key exception
2. Log exceptions as warnings
3. Do not allow incomplete reads to be cached
4. For the existing symbol lists cache keys, use the key's version to distinguish if it could be affected. Use a new config option to control whether the cache should be reloaded or simply produce a warning.

mehertz:
>wrong results

I'm not sure what wrong means here - would it not always be at worst partial (i.e. down to the first missing version)?

qc00:
"Wrong" could be:
* Live symbol being reported as deleted
* Read-able version being reported as non-exist

Such wrong result can also influence the decision of higher-level calls, e.g. the symbol list cache, making a transient storage issue have permanent effect elsewhere.

mehertz:
Okay, intermittent failures on read will be confused for missing data and cached. Understood.

willdealtry:
This is not a very accurate characterization of the history of this setting. `iterate_on_failure` was introduced very early on, when it was discovered that S3 ListObjects calls were slow, and version ref keys were introduced to deal provide a traversal- rather than an iteration-based mechanism for accessing symbol/version pairs. The setting was essentially a feature flag to allow fallback to the old, iteration-based approach.

Deleting a version never created a gap in the version chain before tombstones, which would have been too self-evidently broken to be usable. Rather, a deletion triggered an immediate version map compaction, which meant that there was subsequently only one version key. This is not ideal in terms of parallel consistency between clients, and the tombstone approach was eventually preferred.

[...]

PR #393, opened 2023-05-17
Symbol list

qc00:
**You might want to review commit by commit**

---

Refactoring for #30

* Make s3_store_factory reuse_name work
* Remove inefficient SegmentInMemory Row visit_* methods
* Hooked StorageFailureSimulator into InMemoryStore
* Implemented prefix matching in InMemoryStore
* Error code tidy up
* Clarified reload_symbol_list()
* Additional C++ interop facilities

#30 Simpler and more reliable SymbolList logic

+ Minimum fix for version map swallowing exceptions (#365)
+ Fix off-by-one error in submit_tasks_for_range() exception case


mehertz:
This is a large PR - prior to publishing as a full PR please update the description with details on the included changes.

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -272,8 +272,8 @@ inline void submit_tasks_for_range(const Inputs& inputs, TaskSubmitter submitter
 
     auto fut_itr = futs.begin();
     try {
-        for(auto input_itr = inputs.cbegin(); input_itr != inputs.cend(); ++input_itr, ++fut_itr) {
-            auto&& resolved = std::move(*fut_itr).get();
+        for(auto input_itr = inputs.cbegin(); input_itr != inputs.cend(); ++input_itr) {
```
vasil-pashov, line 275: Looks like we can simplify by using range based for.

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -7,56 +7,148 @@
...
+    static_assert(std::is_copy_assignable_v<ActionSequence>);
+
+private:
+    friend class StorageFailureSimulator;
+
+    const ActionSequence sequence_;
+    std::atomic<size_t> cursor_ {0}; // Index into sequence
+
+public:
+    FailureTypeState(ActionSequence sequence) :
```
vasil-pashov, line 106: Please mark single arg constructors explicit.

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -7,56 +7,148 @@
...
+            sequence_(sequence.empty() ? ActionSequence{action_factories::no_op} : std::move(sequence)) {}
+
+    const ActionSequence::value_type& pick_action() {
+        if (cursor_ < sequence_.size()) {
+            if (auto local = cursor_.fetch_add(1); local < sequence_.size()) {
+                return sequence_[local];
+            }
+        }
+        return sequence_[sequence_.size() - 1];
```
vasil-pashov, line 115:
```suggestion
        return sequence_.back()
```

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -7,56 +7,148 @@
...
+     * The string is a sequence of "action indicators" selecting the action to perform for each call.
+     * After this sequence is exhausted, the last action is used for all subsequent calls.
+     */
+    using Params = std::unordered_map<FailureType, ParamActionSequence>;
+
     static std::shared_ptr<StorageFailureSimulator> instance();
-    static std::shared_ptr<StorageFailureSimulator> instance_;
-    static std::once_flag init_flag_;
-    static void init(){
+    /* Clears/re-initializes the simulator. All action sequences, custom actions and states will be cleared. */
```
vasil-pashov, line 130: The comment seems redundant.

cpp/arcticdb/storage/failure_simulation.hpp
qc00: As this class cannot be moved, the map (entry) containing this can only be constructed using the `piecewise_construct` trick below, so this can't be explicit.

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -29,6 +29,7 @@
 #include <set>
 #include <mutex>
 
+// FUTURE(#297)
```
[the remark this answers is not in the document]
qc00: GitHub issues and PRs are allocated from the same sequence, so there's no confusion of what the number points to.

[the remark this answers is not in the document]
vasil-pashov: My point was that it's just a number and it's not much informative. I've worked on codebase where the there was a switch from one bug tracking system to another. Comments referring issue by ID became kind of hard to follow. Maybe a permalink would be better.

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -10,53 +10,143 @@
...
+
+    inline void operator()(FailureType type) const {
+        proxy_(type);
+    }
+};
+
+inline std::ostream& operator<<(std::ostream& out, const FailureAction& action) {
+    util::variant_match(action.description_,
+                        [&out](const char* desc) { out << desc ;},
+                        [&out](const std::string& desc) { out << desc ;});
```
vasil-pashov, line 60: If this works it could be a bit simpler:
```suggestion
    std::visit([&out](const auto& desc){out << desc;}, action.description_);
```

cpp/arcticdb/storage/failure_simulation.hpp
vasil-pashov: I don't get why can't this be explicit. Is it a compilation error, I haven't heard of any rules forbidding the explicit specifier.

cpp/arcticdb/storage/failure_simulation.hpp
qc00: I like documenting behaviour that is not verbatim in the function body to indicate intent.

cpp/arcticdb/column_store/memory_segment_impl.hpp
```diff
@@ -240,7 +207,7 @@ class SegmentInMemoryImpl {
         }
 
         bool operator<(const Row &other) const {
-            return visit_field(0, [that=this, &other](auto type_desc_tag) {
+            return entity::visit_field(parent_->field(0), [that=this, &other](auto type_desc_tag) {
                 using RawType =  typename decltype(type_desc_tag)::DataTypeTag::raw_type;
                 return that->parent_->scalar_at<RawType>(that->row_id_, 0) <
                 other.parent_->scalar_at<RawType>(other.row_id_, 0);
```
vasil-pashov, line 213: that=this is no longer necessary 
```suggestion
            return entity::visit_field(parent_->field(0), [this, &other](auto type_desc_tag) {
                using RawType =  typename decltype(type_desc_tag)::DataTypeTag::raw_type;
                return parent_->scalar_at<RawType>(that->row_id_, 0) <
                other.parent_->scalar_at<RawType>(other.row_id_, 0);
```

cpp/arcticdb/processing/test/test_clause.cpp
```diff
@@ -18,14 +19,11 @@ void segment_scalar_assert_all_values_equal(const arcticdb::ProcessingSegment& s
     slice_and_key.ensure_segment(empty);
     const arcticdb::SegmentInMemory& segment_memory = slice_and_key.segment(empty);
     segment_memory.init_column_map();
+    auto column_index = segment_memory.column_index(name.value).value();
     size_t row_counter = 0;
     for (auto row : segment_memory) {
```
vasil-pashov, line 24: It's kind of hard to know what type row is (I personally despise `auto`), but it looks like it's changed so maybe:
```suggestion
    for (const auto& row : segment_memory) {
```

qc00: It should be `SegmentInMemoryImpl::Row` which contains a pointer and a number.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -5,83 +5,104 @@
...
+    } else {
+        load_result.timestamp = store->current_timestamp();
+        load_result.symbols = load_from_version_keys(store);
     }
+    return load_result;
+}
 
-    SymbolList::CollectionType SymbolList::compact(std::shared_ptr<Store> store, const std::vector<AtomKey>& symbol_keys) {
+SymbolList::CollectionType SymbolList::load(const std::shared_ptr<Store>& store, bool no_compaction) {
+    // Steps 0-1
```
willdealtry, line 48: Please get rid of this

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -5,83 +5,104 @@
...
+    const LoadResult load_result = ExponentialBackoff<std::runtime_error>(100, 2000)
```
willdealtry, line 49: Do we actually want to catch runtime_error? What error conditions is it legitimate to ignore?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -216,27 +241,31 @@ static const StreamId compaction_id {CompactionId};
...
-            std::shared_ptr<Store> store, const KeyVector& lists) {
-        std::vector<VariantKey> vars;
-        vars.reserve(lists.size());
-        for(const auto& atom_key: lists)
-            vars.emplace_back(VariantKey{atom_key});
-        store->remove_keys(vars).get();
+    folly::Future<std::vector<Store::RemoveKeyResultType>>
+    SymbolList::delete_keys(const std::shared_ptr<Store>& store, KeyVector&& to_remove, const AtomKey& exclude) {
+        std::vector<VariantKey> variant_keys;
+        variant_keys.reserve(to_remove.size());
```
willdealtry, line 247: Are we making a complete copy of all the keys except one to remove one of them?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -216,27 +241,31 @@ static const StreamId compaction_id {CompactionId};
...
+            if (atom_key != exclude) {
+                variant_keys.emplace_back(std::move(atom_key));
+            }
+        }
+
+        return store->remove_keys(variant_keys);
     }
 
-    std::optional<SymbolList::KeyVector::difference_type> SymbolList::last_compaction(const KeyVector& keys) {
+    auto SymbolList::last_compaction(const KeyVector& keys) -> std::optional<KeyVectorItr> {
```
willdealtry, line 259: This pattern is not used anywhere else. Why is it used here?

willdealtry (approved):

cpp/arcticdb/column_store/memory_segment_impl.hpp
```diff
@@ -198,39 +198,6 @@ class SegmentInMemoryImpl {
         parent_(parent),
         row_id_(row_id_) {}
 
-        template<typename Callable>
```
willdealtry, line 201: I think that these are used in the tick data dedup

willdealtry (changes requested):

cpp/arcticdb/version/symbol_list.cpp
qc00: This is the Atom->Variant Key conversion and filtering folded into one. I've done my best with `std::move` below.

cpp/arcticdb/version/symbol_list.cpp
qc00: Left over from copy-and-paste of internal code

python/arcticdb/version_store/library.py
```diff
@@ -1415,7 +1415,10 @@ def handle_symbol(s):
 
     def reload_symbol_list(self):
         """
-        Forces the symbol list cache to be reloaded
+        Forces the symbol list cache to be reloaded.
+
+        This can take a long time on large libraries or certain S3 implementations, and once started, it cannot be
```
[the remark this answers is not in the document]
qc00: Once any of the `__add__` or `__del__` keys are deleted, the symbol list cache will be incomplete and must be re-built from the version keys.

cpp/arcticdb/version/symbol_list.cpp
qc00: There are two classes of issues to consider:

* Some `Storage` don't throw the correct exception type in all cases
* We will need to check if symbol list code itself need this

Given this implementation has been validated internally, I am more inclined to release this first and then improve.

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
mehertz: This is scary. 

If it is interrupted, is it safe to run this again?

mehertz (changes requested):

python/arcticdb/version_store/helper.py
```diff
@@ -279,7 +277,8 @@ def create_test_lmdb_cfg(lib_name: str, db_dir: str, lmdb_config: Dict[str, Any]
     return cfg
 
 
-def create_test_s3_cfg(lib_name, credential_name, credential_key, bucket_name, endpoint):
+def create_test_s3_cfg(lib_name, credential_name, credential_key, bucket_name, endpoint, *, with_prefix=True):
```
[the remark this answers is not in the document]
qc00: Inline typing will cause the line to wrap, which is less readable, IMO. In reality, all tools support both formats.

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
qc00: > If it is interrupted, is it safe to run this again?

Yes, and it must be run again.

cpp/arcticdb/processing/test/test_clause.cpp
```diff
@@ -6,6 +6,7 @@
  */
 
 #include <gtest/gtest.h>
+#include <gmock/gmock.h>
```
willdealtry, line 9: I think the inclusion of a completely different testing framework (even on that is contained in an existing one) is somethign we should maybe talk about before code review. Does the syntactic sugar for one row justify the additional project-wide dependency?

cpp/arcticdb/storage/failure_simulation.hpp
willdealtry: I do agree that clear does pretty much exactly what you expect it to, shouldn't comments be for behaviour that is surprising? The issue is really consistency with the rest of the project, is it to be policy to do this everywhere from now on, or are we just doing it for this particular file?

willdealtry (changes requested):

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
qc00: Raised #446 for a solution.

cpp/arcticdb/processing/test/test_clause.cpp
qc00: Not sure why Google put the matcher functions in a gmock, but those two projects have been merged into one repo at least. I can narrow the include to `gmock-matchers.h` if it helps?

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
qc00: @willdealtry 

> But that happens at the end, doesn't it?

Could you clarify what "that" refers to?

If you meant the re-calculation of the symbol list at the end, the weakness starts in the deletion step before that. If the `reload_symbol_list` call is interrupted in the deletion step, then any use of the symbol list cache from anywhere might return incomplete results.

cpp/arcticdb/version/symbol_list.cpp
qc00: I tried to fix the storage when we did this change internally, but it involves many layers of storage abstraction, so requires significant effort to check and test all the downstream code.

This symbol list change has been verified internally and is known to improve things, so I want it decoupled from any potential change that can break things. I have logged #447.

cpp/arcticdb/storage/failure_simulation.hpp
qc00: Will remove

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
willdealtry: Oh right, it's this thing. Pretty sure I wrote it originally as two separate methods for exactly this reason.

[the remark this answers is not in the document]
willdealtry: I don't love the name, it's too bland for what the function actually does, which is nuke the whole thing and rebuild from scratch

python/arcticdb/version_store/helper.py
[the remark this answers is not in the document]
willdealtry: We should settle on one and use it. If there are two many parameters to a function then that's a code smell that should be addressed, e.g. by a Parameter Object.

cpp/arcticdb/version/symbol_list.cpp
willdealtry: I think it would be better to make the types consistent. VariantKey is a later addition, but generally speaking we are converging on using it everywhere to avoid this sort of copying conversion. But I guess we can leave it for a future change

willdealtry (approved):

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
qc00: The deletion part is exposed via PyBind, but not on any higher-level objects.

I feel once the symbol list has been made stabled, we should remove _this_ method and expose the deletion-only method. I have added a comment to #446

PR #395, opened 2023-05-17
Support update with daterange for generic non-Pandas timeseries structures

poodlewars:
Support update with daterange for generic non-Pandas timeseries structures, such as one occasionally used in Man.

I've moved `test_update_date_range_dataframe` unchanged to a new file as `test_basic_version_store.py` is huge.

The checks will fail on this until something like https://github.com/man-group/ArcticDB/pull/388 is merged.

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -1257,7 +1257,7 @@ def restrict_data_to_date_range_only(data: T, *, start: Timestamp, end: Timestam
             start, end = _strip_tz(start, end)
         data = data.loc[pd.to_datetime(start) : pd.to_datetime(end)]
     else:  # non-Pandas, try to slice it anyway
-        if not data.timezone:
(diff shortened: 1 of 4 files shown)


qc00 (approved):

python/arcticdb/version_store/_normalization.py
```diff
@@ -1255,6 +1256,10 @@ def _strip_tz(s, e):
         if not data.index.get_level_values(0).tz:
             start, end = _strip_tz(start, end)
         data = data.loc[pd.to_datetime(start) : pd.to_datetime(end)]
+    else:  # non-Pandas, try to slice it anyway
+        if not data.timezone:
```
mehertz, line 1260: `getattr` here to reduce the requirements to only slicing?

qc00:
Seems to have broken the Windows integration tests with a change that shouldn't affect Windows?

qc00 (approved):

PR #401, opened 2023-05-18
Move encoded field and descriptors from protobuf

willdealtry:
Move encoded fields and stream descriptors out of protobuf

This change introduces a Field and EncodedField type, which replace the protobuf field and encoded_field message types respectively. These are stored as columns, the stream descriptor fields at the head of the segment, and the encoded fields at the end (because we don't know the details until we've compressed all the rest of the data).

Writing the old or new format is a library setting. The encoding type is stored on the segments and the code can now decode new or old formats. Internally we use the new encoding consistently.

Introduces C++ classes representing stream descriptor and timeseries descriptor instead of using the protobuf versions, and tasks to retrieve them individually.

Note that this introduces a new library-level option - `encoding_version`. This setting controls the encoding version used to store data segments (and only data segments!) within ArcticDB. When set to `1`, (`0` is the default), the above changes will take effect. `encoding_version` is not necessarily backwards-compatible. If an old client attempts to read data written with an encoding it cannot recognize, it will get the error 'Error decoding <codec>', with a numeric error code from the codec itself.


mehertz:
image

Can you please add the **full set of changes to the PR descriptionI**. A real quick glance shows that you are changing our dependency from `asan` to `tsan` for example so I'm guessing there are further small unrelated changes throughout.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -281,17 +293,19 @@ class AsyncStore : public Store {
...
+        auto keys = std::move(ks);
         util::check(keys.size() == continuations.size(),
                     "keys and continuations must has then same number of elements");
         std::vector<folly::Future<VariantKey>> batch;
         batch.reserve(args.batch_size_);
         std::vector<VariantKey> res;
         res.reserve(keys.size());
         for (std::size_t i = 0; i < keys.size(); ++i) {
-            auto &key = keys[i];
+            const auto &key = keys[i];
```
qc00, line 308: The variable is only used once, so might as well do `std::move(keys[i])` for efficiency?

cpp/arcticdb/async/base_task.hpp
qc00, line 1: It might make this PR more reviewable if you can exclude these file mode changes.

qc00, line 1: It might make this PR more reviewable if you can exclude these file mode changes.

mehertz (changes requested):

cpp/arcticdb/CMakeLists.txt
```diff
@@ -710,6 +714,7 @@ if(${TEST})
     if (WIN32)
         target_link_libraries(test_unit_arcticdb
                 PUBLIC
+                #asan
```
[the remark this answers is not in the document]
willdealtry: These have been there since forever so you can turn on sanitizers with a quick find and replace. I guess I''ll replace them with a cmake flag.

mehertz (changes requested):
Document the new library setting

python/arcticdb/exceptions.py
mehertz, line 1: These have still been ch-modded.

mehertz (changes requested):

willdealtry (commented):
Commented on things

mehertz (changes requested):
What error will be thrown if an old client reads new data? Please paste the error!

alexowens90 (approved):

mehertz (approved):

PR #403, opened 2023-05-19
github: Update bug-report template

jjerphan:
This proposes updating the bug report template so that users can provide minimal reproducible examples, eventually allowing us to better troubleshoot and fix issues.

The rendering of the template is previewable [here](https://github.com/man-group/ArcticDB/blob/github/update-bug-report-template/.github/ISSUE_TEMPLATE/bug_report.yaml).


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,135 @@
...
+      A clear and concise description of what the bug is.
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: Steps/Code to Reproduce
+    description: |
+      Please add a [minimal code example](https://matthewrocklin.com/minimal-bug-reports) that can reproduce the error when running it.
+    Be as succinct as possible, **do not depend on external data files**: instead you can generate synthetic data using `numpy.random`.
```
mehertz, line 25: How about adding:

```If your bug report appears to be dependant on the external data files and cannot be reproduced without, please describe your data files in as much detail as you can.```

File: .github/ISSUE_TEMPLATE/bug_report.yaml [hunk 2]
```diff
@@ -0,0 +1,135 @@
...
+    placeholder: >
+      1.1.0
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: Backend storage used
+    placeholder: >
+      S3
```
poodlewars, line 122: Maybe nudge people to be more specific eg AWS S3 / GCP S3

.github/ISSUE_TEMPLATE/bug_report.yaml
poodlewars: s/dependant/dependent
s/on the external/on external
s/without/without them

poodlewars (approved):

PR #404, opened 2023-05-19
Fix a backwards incompatible change to exceptions

poodlewars:
The NoSuchVersionException needs to be a NoDataFoundException, since read_descriptor used to throw a NoDataFoundException but now it throws a NoSuchVersionException, and we need to be compatible with earlier consumers who may be catching a NoSuchVersionException.

I implement this by lifting the affected exceptions up to the top level `python_module.cpp`. I think this makes the configuration the most declarative and clear for this slightly fiddly interaction across submodules.

Internal ref: AN-976

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ dist/
 cmake-build-*/
 dist_release/
 CMakeUserPresets.json
+
+*_pb2.py
+*.mdb
--- a/cpp/arcticdb/python/python_module.cpp
+++ b/cpp/arcticdb/python/python_module.cpp
@@ -9,6 +9,7 @@
 #include <arcticdb/codec/python_bindings.hpp>
 #include <arcticdb/column_store/python_bindings.hpp>
 #include <arcticdb/storage/python_bindings.hpp>
+#include <arcticdb/storage/storage.hpp>
 #include <arcticdb/stream/python_bindings.hpp>
 #include <arcticdb/toolbox/python_bindings.hpp>
 #include <arcticdb/version/python_bindings.hpp>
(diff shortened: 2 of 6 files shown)


cpp/arcticdb/python/python_module.cpp
```diff
@@ -271,10 +272,27 @@ PYBIND11_MODULE(arcticdb_ext, m) {
...
+    auto storage_submodule = m.def_submodule("storage", "Segment storage implementation apis");
+    auto no_data_found_exception = py::register_exception<arcticdb::storage::NoDataFoundException>(
+            storage_submodule, "NoDataFoundException", base_exception.ptr());
+    arcticdb::storage::apy::register_bindings(storage_submodule, base_exception);
+
     arcticdb::stream::register_bindings(m);
     arcticdb::toolbox::apy::register_bindings(m);
-    arcticdb::version_store::register_bindings(m, base_exception);
+
+    m.def("get_version_string", &arcticdb::get_arcticdb_version_string);
```
mehertz, line 284: I know @qc00 has been working on this but I'm surprised to see this in this PR?

mehertz: Okay, you've removed from version python_bindings.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -576,14 +574,6 @@ void register_bindings(py::module &m, py::exception<arcticdb::ArcticException>&
              "Returns latest timestamp of a symbol")
         ;
 
-    m.def("get_version_string", &get_arcticdb_version_string);
```
mehertz, line 579: I get moving the exception definition up, but why do move this up as well?

mehertz (approved):

.gitignore
```diff
@@ -33,3 +33,6 @@ dist/
 cmake-build-*/
 dist_release/
 CMakeUserPresets.json
+
+*_pb2.py
```
qc00, line 37: Wait, why would these be picked up?

cpp/arcticdb/python/python_module.cpp
qc00: Should just remove this function then. It does nothing.

Issue #402, opened 2023-05-19
Editing the top level CMakeLists.txt triggers a full rebuild


vasil-pashov:
**Is your feature request related to a problem? Please describe.**
When CMake is reconfigured, which happens if the top level CMakeLists.txt gets edited or when reconfiguring is done manually (e.g. via clicking the reconfigure button in the IDE) a full rebuild of the whole project is triggered. The rebuild is triggered even if no cpp/hpp/h files were changed.

**Describe the solution you'd like**
Fix the build process so that editing the CMakeLists.txt does not require full rebuild. If new files are added to the project they must, obviously, be compiled. But if the changes are only CMake related we must not rebuild the whole project (e.g. adding a custom target should not rebuild the project).

**Describe alternatives you've considered**
No alternatives.

**Additional context**

poodlewars:
Is this not always the case with CMake?

Even so, if it is, we should set up SCCACHE for our dev environments which would still fix most of the pain. It's already used in the CI build and should be really easy.

vasil-pashov:
> Is this not always the case with CMake?
> 
> Even so, if it is, we should set up SCCACHE for our dev environments which would still fix most of the pain. It's already used in the CI build and should be really easy.

No, reconfiguring CMake should (almost) never trigger any kind of (even) partial builds.

PR #411, opened 2023-05-19
Fix files being unnecessarily recompiled after CMake

qc00:
#402


vasil-pashov:
Do I understand correctly the root cause of the issue and the fix applied

CMake was editing files in-place at configure time. Those files were `#include`-d in lot's of places in the source, thus the edit triggered a rebuild. Now the custom command:
* We don't edit proto files in place
* We create a copy of the proto file and edit it instead
* We use the copy of the proto file
* We do this only if the proto file is changed

Maybe we can add some info of the fix in the commit message.

vasil-pashov (approved):

PR #415, opened 2023-05-22
Parallelise version map requests across version and symbol

mehertz:


cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,132 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    output.reserve(symbols.size());
+    for(const auto& symbol : folly::enumerate(symbols)) {
+        const auto it = version_data.find(*symbol);
+        util::check(it != version_data.end(), "Missing version data for symbol {}", *symbol);
+        if(it->second.count_ == 1) {
+            output.push_back(async::submit_io_task(CheckReloadTask{store, version_map, *symbol,
+                                                                   it->second.load_param_}).thenValue([version_query =
+            version_queries[symbol.index]](auto version_map_entry) {
+                return get_key_for_version_query(version_map_entry, version_query);
+            }));
```
qc00, line 229: Wrapping on line 226 is difficult to read. Might want to manually wrap `.thenValue`.

Line 234 below is even worse...

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,132 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+            if(fut == shared_futures.end()) {
+                auto [splitter, inserted] = shared_futures.emplace(*symbol, folly::FutureSplitter(async::submit_io_task(CheckReloadTask{store, version_map, *symbol,
+                                                                                                                                                                           it->second.load_param_})));
+                output.push_back(splitter->second.getFuture().thenValue([version_query = version_queries[symbol.index]](auto version_map_entry) {
+                    return get_key_for_version_query(version_map_entry, version_query);
+                }));
+            } else {
+                output.push_back(fut->second.getFuture().thenValue([version_query = version_queries[symbol.index]](auto version_map_entry) {
+                    return get_key_for_version_query(version_map_entry, version_query);
+                }));
```
qc00, line 241: These three lines are almost identical in all 3 branches. Consider extracting the part before `.thenValue` to a variable and do the `.thenValue` outside the `if`?

cpp/arcticdb/version/test/test_version_map_batch.cpp
```diff
@@ -0,0 +1,84 @@
...
+    using namespace arcticdb::pipelines;
+
+    auto store = test_store_->_test_get_store();
+    auto version_map = std::make_shared<VersionMap>();
+
+    for(auto i = 0u; i < 5; ++i) {
+        auto stream = fmt::format("stream_{}", i);
+        add_versions_for_stream(version_map, store, stream, 5);
+    }
+
```
GemDot, line 43: was the original intention to differentiate between actions across these two loops?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -605,7 +605,7 @@ if(${TEST})
             version/test/test_version_map.cpp
             version/test/test_version_store.cpp
             version/test/version_map_model.hpp
-            )
+            version/test/test_stream_version_data.cpp version/test/test_version_map_batch.cpp)
```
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/version_functions.hpp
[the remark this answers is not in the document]
GemDot: :)

cpp/arcticdb/version/test/test_version_map_batch.cpp
GemDot: fixed in another commit

GemDot: simplified in another commit

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -43,6 +50,7 @@ struct LoadParameter {
 
     LoadType load_type_ = LoadType::NOT_LOADED;
     std::optional<VersionId> load_until_ = std::nullopt;
+    std::optional<timestamp> load_from_time_ = std::nullopt;
```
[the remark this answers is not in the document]
GemDot: yes, in this case I think it is still fine and would be too late

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: done

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: Agree. Done

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,132 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+        react(version_query);
+    }
+
+    StreamVersionData() = default;
+
+    size_t count_ = 0;
+    LoadParameter load_param_ = LoadParameter{LoadType::LOAD_LATEST_UNDELETED};
+
+    void react(const pipelines::VersionQuery& version_query) {
+        util::variant_match(version_query.content_, [that = this](const auto &query) {
```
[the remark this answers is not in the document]
GemDot: I would leave @willdealtry to comment on here, since he explicitly wanted to use variants I think

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
     return output;
 }
 
+struct StreamVersionData {
```
GemDot, line 123: @vasil-pashov moving the comment here https://github.com/man-group/ArcticDB/pull/423/files#r1205054793

GemDot: However, it is a struct. I think it is a good practice to define it in the app file

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    explicit StreamVersionData(const pipelines::VersionQuery& version_query) {
+        react(version_query);
+    }
+
+    StreamVersionData() = default;
+
+    size_t count_ = 0;
+    LoadParameter load_param_ = LoadParameter{LoadType::LOAD_LATEST_UNDELETED};
```
GemDot, line 131: @vasil-pashov moving the comment here https://github.com/man-group/ArcticDB/pull/423/files#r1205055757

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+        ++count_;
+        switch(load_param_.load_type_) {
+        case LoadType::LOAD_LATEST_UNDELETED:
+            load_param_ = LoadParameter{LoadType::LOAD_DOWNTO, specific_version.version_id_};
+            break;
+        case LoadType::LOAD_DOWNTO:
+            util::check(load_param_.load_until_.has_value(), "Expect LOAD_DOWNTO to have version specificed");
+            load_param_.load_until_ = std::min(load_param_.load_until_.value(), specific_version.version_id_);
+            break;
+        case LoadType::LOAD_FROM_TIME:
```
GemDot, line 153: @vasil-pashov moving the comment here https://github.com/man-group/ArcticDB/pull/423/files#r1205057639

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    const std::vector<pipelines::VersionQuery>& version_queries) {
+    ARCTICDB_SAMPLE(BatchGetVersion, 0)
+    util::check(symbols.size() == version_queries.size(), "Symbol and version query list mismatch: {} != {}", symbols.size(), version_queries.size());
+
+    robin_hood::unordered_flat_map<StreamId, StreamVersionData> version_data;
+    for(const auto& symbol : folly::enumerate(symbols)) {
+        auto it = version_data.find(*symbol);
+        if(it == version_data.end())
+            version_data.insert(robin_hood::pair<StreamId, StreamVersionData>(*symbol, StreamVersionData{version_queries[symbol.index]}));
+        else
```
GemDot, line 215: @vasil-pashov moved the comment here https://github.com/man-group/ArcticDB/pull/423/files#r1205061424

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    for(const auto& symbol : folly::enumerate(symbols)) {
+        const auto it = version_data.find(*symbol);
+        util::check(it != version_data.end(), "Missing version data for symbol {}", *symbol);
+        auto version_entry_fut = folly::Future<std::shared_ptr<VersionMapEntry>>::makeEmpty();
+        if(it->second.count_ == 1) {
+            version_entry_fut = async::submit_io_task(CheckReloadTask{store, version_map, *symbol, it->second.load_param_});
+        } else {
+            auto fut = shared_futures.find(*symbol);
+            if(fut == shared_futures.end()) {
+                auto [splitter, inserted] = shared_futures.emplace(*symbol, folly::FutureSplitter(async::submit_io_task(CheckReloadTask{store, version_map, *symbol, it->second.load_param_})));
```
GemDot, line 232: @vasil-pashov moving your comment here https://github.com/man-group/ArcticDB/pull/423/files#r1205063759

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: If we add it, I think that if we add it, we should do the same for all the switch/cases, to be consistent. So far, I don't see any place in the code base where it is used and on my opinion that's just used to avoid potential warnings/user clarification. I don't see a big benefit.

cpp/arcticdb/version/test/test_stream_version_data.cpp
```diff
@@ -0,0 +1,101 @@
+#include <gtest/gtest.h>
+
+#include <arcticdb/version/version_map_batch_methods.hpp>
+
+TEST(StreamVersionData, SpecificVersion) {
+    using namespace arcticdb;
+    using namespace arcticdb::pipelines;
```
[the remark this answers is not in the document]
willdealtry: Why does that make it better? Honestly I preferred it as it was

cpp/arcticdb/version/version_map_batch_methods.hpp
willdealtry: It's an implementation detail, it belongs with the implementation. I don't think we should declare structs in the header file as a matter of course, when they are not part of the API that we want to expose

cpp/arcticdb/version/version_map_batch_methods.hpp
willdealtry: It's kind of nice, and we do use [[nodiscard]] etc elsewhere. The standard versions of these weren't available reliably when the project was started, so it's not everywhere. Likewise there are some preprocessor-defined GCC attributes that are now part of the standard, that we should replace at some point

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: From the time you commented it, I refactored all this code and I think now this function looks much clearer. Regarding this line in particular. As for your comment, I also splatted the line in two, to improve the clarity.

cpp/arcticdb/version/test/test_version_map_batch.cpp
```diff
@@ -0,0 +1,433 @@
...
+    StreamId stream_id,
+    size_t num_versions,
+    size_t start = 0u) {
+    for(auto i = start; i < start + num_versions; ++i) {
+        version_map->write_version(store, test_index_key(stream_id, i));
+    }
+};
+}
+
+TEST_F(VersionMapBatchStore, SimpleVersionIdQueries) {
```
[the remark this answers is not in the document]
GemDot: Comment from William, from the other PR: `I mind repetition much less in tests than I do in production code. Personally I feel like some of the rules like DRY and no explicit strings that are mandatory for real code are actually reversed in test code, where you want everything to be obvious and local; I don't really want to have to go and look elsewhere to see what a test is doing when it fails.`

cpp/arcticdb/version/version_map_batch_methods.hpp
qc00: This line looks very similar to before and is still too long!

I urge you to extract the `async::submit_io_task(...)` to a temporary variable and `std::move` its result here.

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -346,11 +358,14 @@ inline std::optional<VersionId> get_next_version_in_entry(const std::shared_ptr<
     return std::nullopt;
 }
 
-inline std::optional<AtomKey> find_index_key_for_version_id(VersionId version_id, const std::shared_ptr<VersionMapEntry>& entry) {
+inline std::optional<AtomKey> find_index_key_for_version_id(VersionId version_id, const std::shared_ptr<VersionMapEntry>& entry, bool included_deleted=true) {
```
qc00, line 361: * This line is now too long. Please wrap
* `find_index_key_for_version_timestamp` below differs by just a few bytes. Please move it up so it immediately follows this function.

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: sorry @qc00, it is exactly what I already did. But my push command went wrong and I didn't realise of that

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: To summarise:
We don’t want to define/declare this new struct in separate hpp/cpp files because it would give the wrong idea to the users (i.e. it can be used elsewhere). On the other side, even if the functions were defined in the cpp, we woudln’t want to define this struct in the cpp, otherwise the tests would not have access to its definition. It is also possible to declare a struct in the hpp and define in the cpp, but it doesn’t seem to be a common practice

cpp/arcticdb/version/test/test_version_map_batch.cpp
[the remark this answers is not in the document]
GemDot: I think most of your suggestions are fair. Only for the first one, be aware that this change is neither exposing any extra API nor using this functionality in any existing one. Let me have a second thought about trying to have the test with less duplication, which is what I agree most.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    },
+    [&version_map_entry] (const pipelines::TimestampVersionQuery& timestamp_version) {
+       return find_index_key_for_version_timestamp(timestamp_version.timestamp_, version_map_entry);
+    },
+    [&version_map_entry] (const std::monostate&) {
+       return version_map_entry->get_first_index(false);
+       },
+    [] (const auto&) -> std::optional<AtomKey> {
+        util::raise_rte("Unsupported version query type");
+    });
```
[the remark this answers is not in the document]
GemDot: -Done the first 2 bullets points.

-For the third one: which is good if more calls to get_key_for_version_query are tried to be added in the future, with a non-supported query type

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    }
+
+    using SplitterType = folly::FutureSplitter<std::shared_ptr<VersionMapEntry>>;
+    robin_hood::unordered_flat_map<StreamId, SplitterType> shared_futures;
+    std::vector<folly::Future<std::optional<AtomKey>>> output;
+    output.reserve(symbols.size());
+    for(const auto& symbol : folly::enumerate(symbols)) {
+        const auto it = version_data.find(*symbol);
+        util::check(it != version_data.end(), "Missing version data for symbol {}", *symbol);
+        auto version_entry_fut = folly::Future<std::shared_ptr<VersionMapEntry>>::makeEmpty();
```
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/version_map_entry.hpp
GemDot: done

cpp/arcticdb/version/version_map_batch_methods.hpp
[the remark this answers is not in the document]
GemDot: Actually, it your proposal was not compiling. Future needs to be initialised, so. considering it is more like a nicknack, I will leave it as it was, since I am already explicating the type in the initialisation.

cpp/arcticdb/version/test/test_version_map_batch.cpp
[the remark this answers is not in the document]
GemDot: The reason why it was decided to stress the tests with multiple requests (not just 3), is because we are running it in parallel and for the (sometimes fragile) way folly works, we wanted to be sure that there is no any hidden bug there when we do a massive parallelisation.

cpp/arcticdb/version/version_map_batch_methods.hpp
mehertz: Let's move the definition of `StreamVersionData` to it's own header file (and possibly it's own cpp file).

GemDot: We will do it in another PR, as has been commonly decided https://github.com/man-group/ArcticDB/issues/452

willdealtry (approved):

PR #416, opened 2023-05-22
Store the product's version number in C++ (#110)

qc00:
So to embed metadata in symbol list keys for diagnostics use.

In the future, we can add a background job to scan for potentially corrupt symbol list written by an old version and re-populate them.

Has to be merged after #393 because they touch the same files


cpp/arcticdb/util/writer_info.hpp
willdealtry, line 1: Really I think this belongs in the telemetry. I had originally asked for the telemetry to be an abstract interface into which we could plug various different back-ends, but alas it ended up specific to Prometheus. Ultimately this object should be a wrapper around that interface that supplies the relevant data to that, so that we can plug in OpenTelemetry which is a near-term requirement.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -159,25 +163,34 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
...
+        if (existing_descriptor) {
+            metadata.CopyFrom(*existing_descriptor);
+            if (ARCTICDB_VERSION_INT < metadata.min_writer_version()) {
+                metadata.set_min_writer_version(ARCTICDB_VERSION_INT);
+            }
+        } else {
             metadata.set_enabled(true);
-            any.PackFrom(metadata);
-            list_segment.set_metadata(std::move(any));
+            util::set_static_writer_info_fields(*metadata.mutable_initial_writer());
```
willdealtry, line 175: I don't really think that writing telemetry data into the compacted symbol list is going to help us debug anything, because compaction, although it is somewhat complex compared to the other operations, represents only 0.1% of the symbol lists actions, so knowing the version that did the compaction is likely to be a red herring in terms of diagnosing symbol list corruption.

willdealtry (changes requested):

PR #417, opened 2023-05-22
Test against older Python deps that we want to keep compat with

poodlewars:
Run the CI against older versions of some key Python dependencies against which we wish to remain compatible.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,39 +105,21 @@ jobs:
       matrix:
         format: ${{fromJson(needs.common_config.outputs.linux_cibw_formats)}}
         python3: ${{fromJson(needs.common_config.outputs.linux_python_versions)}}
-    name: 3.${{matrix.python3}} Linux
+        python_deps: [""]
+        include:
+          - python3: 6
+            python_deps: "requirements-compatibility-py36.txt"
+            format: ${{fromJson(needs.common_config.outputs.linux_cibw_formats)}}
+          - python3: 8
+            python_deps: "requirements-compatibility-py38.txt"
+            format: ${{fromJson(needs.common_config.outputs.linux_cibw_formats)}}
+    name: 3.${{matrix.python3}} Linux ${{matrix.python_deps}}
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
     permissions: {packages: write}
     with:
       job_type: follower
       python3: ${{matrix.python3}}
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -118,6 +122,29 @@ jobs:
       cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
       matrix: ${{needs.common_config.outputs.linux_matrix}}
 
+  linux-compat-tests:
+    needs: [follower-linux]
```
qc00, line 126: Have to do it in parallel with the other tests in `build_steps.yml` itself. Otherwise, it will make each build even longer.

build_tooling/requirements-compatibility-1.txt
qc00, line 1: Please include the Python version or a more descriptive name

build_tooling/requirements-compatibility-1.txt
```diff
@@ -0,0 +1,7 @@
+# The oldest supported deps, for Python 3.6
```
qc00, line 1: I would be careful about the word "support" here.

.github/workflows/build_steps.yml
```diff
@@ -221,6 +221,9 @@ jobs:
           python -V
           cd "$RUNNER_TEMP" # Works for Windows-style paths as well
           python -m pip install --force-reinstall *${{env.python_impl_name}}*.whl
+          if [[ -n ${{inputs.python_deps || ''}} ]]
```
qc00, line 224: Can actually do `if ${{inputs.python_deps}}`

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -36,7 +37,8 @@
     NPDDataFrame,
 )
 from arcticdb.version_store._common import TimeFrame
-from arcticdb.util.test import param_dict, CustomThing, TestCustomNormalizer, assert_frame_equal, assert_series_equal
+from arcticdb.util.test import \
+    param_dict, CustomThing, TestCustomNormalizer, assert_frame_equal, assert_series_equal, IS_PANDAS_ZERO
```
qc00, line 41: Not a thing in `black`

poodlewars: Not going to attempt manual Black formatting or running the auto-formatter till it's in the CI

qc00 (approved):

PR #418, opened 2023-05-23
rename isin test to avoid being skipped

GemDot:

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_filtering.py
+++ b/python/tests/unit/arcticdb/version_store/test_filtering.py
@@ -892,7 +892,7 @@ def test_filter_numeric_isin_hashing_overflows(lmdb_version_store, df_col, isin_
     assert_frame_equal(expected, result)
 
 
-def test_filter_numeric_isin_unsigned_not_hypothesis(lmdb_version_store):
+def test_filter_numeric_isin_unsigned_max_uint64(lmdb_version_store):
     df = pd.DataFrame({"a": [0, 1, 2 ** 64 - 1]})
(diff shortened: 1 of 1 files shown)


python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -892,7 +892,7 @@ def test_filter_numeric_isin_hashing_overflows(lmdb_version_store, df_col, isin_
     assert_frame_equal(expected, result)
 
 
-def test_filter_numeric_isin_unsigned(lmdb_version_store):
+def test_filter_numeric_isin_unsigned_not_hypothesis(lmdb_version_store):
```
mehertz, line 895: Calling a test `not_hypothesis` doesn't describe what the test actually _does_. Every test in the codebase could have that suffix.

This test must be testing a specific case seperately to the hypothesis test. What is that case?

mehertz (changes requested):

python/tests/unit/arcticdb/version_store/test_filtering.py
GemDot: because it is forcing the column to be uint64 (2**64 - 1 is one of the values), I would name it as *_uint64 or *_max_uint64?

mehertz: Both are better than `not_hypothesis`.

GemDot: changed it

mehertz (approved):

PR #423, opened 2023-05-24
Generic parallel get version for batch meta

GemDot:
In this PR, we provide per-symbol parallelism for the read_batch_metadata method.

Considering all this and the way folly Futures work, it is required to propagate a chain of futures all the way long for each independent read metadata, which ultimately, they converge in a single future containing all the individual meta reads. On top of these changes we need to re-use the batch_get_versions function as the start of all the future chain, where all the versions are obtained, per each symbol-version query pair.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1160,31 +1160,25 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
     return std::make_pair(specific_versions, latest_versions);
 }
 
+folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::get_metadata_async(folly::Future<std::optional<AtomKey>>&& version_fut){
+    return  std::move(version_fut)
+    .thenValue([this](std::optional<AtomKey> key){
+        if (key.has_value()){
+            return store()->read_metadata(key.value());
+        }else{
+            return folly::makeFuture(std::make_pair(std::nullopt, std::nullopt));
+        }
+    });
+}
+
 std::vector<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::batch_read_metadata_internal(
     const std::vector<StreamId>& stream_ids,
     const std::vector<VersionQuery>& version_queries
     ) {
-    auto [specific_versions_index_map, latest_versions_index_map] = get_stream_index_map(stream_ids, version_queries);
-    auto version_queries_is_empty = version_queries.empty();
+    auto versions_fut = batch_get_versions(store(), version_map(), stream_ids, version_queries);
     std::vector<folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>>> fut_vec;
-    for(const auto& stream_id : folly::enumerate(stream_ids)) {
-        if(!version_queries_is_empty && std::holds_alternative<SpecificVersionQuery>(version_queries[stream_id.index].content_)){
-            const auto& query = version_queries[stream_id.index].content_;
-            auto specific_version_map_key = std::make_pair(*stream_id, std::get<SpecificVersionQuery>(query).version_id_);
-            auto specific_version_it = specific_versions_index_map->find(specific_version_map_key);
-            if (specific_version_it != specific_versions_index_map->end()){
-                fut_vec.push_back(store()->read_metadata(specific_version_it->second));
-            }else{
-                fut_vec.push_back(std::make_pair(std::nullopt, std::nullopt));
-            }
-        }else{
-            auto last_version_it = latest_versions_index_map->find(*stream_id);
-            if(last_version_it != latest_versions_index_map->end()) {
-                fut_vec.push_back(store()->read_metadata(last_version_it->second));
-            }else{
-                fut_vec.push_back(std::make_pair(std::nullopt, std::nullopt));   
-            }
-        }
+    for (auto& version: versions_fut){
+        fut_vec.push_back(get_metadata_async(std::move(version)));
     }
     return folly::collect(fut_vec).get();
 }
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -731,7 +731,7 @@ void LocalVersionedEngine::delete_trees_responsibly(
                     [&check, &not_to_delete](auto& prev) { if (check.prev_version) not_to_delete.insert(prev);},
                     [&check, &not_to_delete](auto& next) { if (check.next_version) not_to_delete.insert(next);},
                     [v=key.version_id()](const AtomKeyImpl& key, const std::shared_ptr<VersionMapEntry>& entry) {
-                        // Can't use is_indexish_and_not_tombstoned() because the target version's index key might have
+                        // Can't use is_live_index_type_key() because the target version's index key might have
                         // already been tombstoned, so will miss it and thus not able to find the prev/next key.
                         return is_index_key_type(key.type()) && (key.version_id() == v || !entry->is_tombstoned(key));
                     });
--- a/cpp/arcticdb/version/test/test_version_map_batch.cpp
+++ b/cpp/arcticdb/version/test/test_version_map_batch.cpp
@@ -41,6 +41,11 @@ TEST_F(VersionMapBatchStore, Simple) {
         add_versions_for_stream(version_map, store, stream, 5);
     }
 
+    for(auto i = 5u; i < 10u; ++i) {
+        auto stream = fmt::format("stream_{}", i);
+        add_versions_for_stream(version_map, store, stream, 5);
+    }
+
     for(auto i = 0u; i < 5; ++i) {
         auto stream = fmt::format("stream_{}", i);
         add_versions_for_stream(version_map, store, stream, 5, 5);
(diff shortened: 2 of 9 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1160,31 +1160,29 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
...
+        return store()->read_metadata(key.value());
+    }else{
+        return folly::makeFuture(std::make_pair(std::nullopt, std::nullopt));
+    }
+}
+
+folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::get_metadata_future(folly::Future<std::optional<AtomKey>>&& version_fut){
+    return  std::move(version_fut)
+    .thenValue([that = this](std::optional<AtomKey> key){
+        return that->get_metadata_future_impl(key);
```
vasil-pashov, line 1174: that = this is a pattern that is no longer needed, we can just capture this
```suggestion
    .thenValue([this](std::optional<AtomKey> key){
        return get_metadata_future_impl(key);
```

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1160,31 +1160,29 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
     return std::make_pair(specific_versions, latest_versions);
 }
 
+folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::get_metadata_future_impl(std::optional<AtomKey> key){
```
vasil-pashov, line 1163: This type is quite long maybe we can use a typedef for this. Maybe `KeyValueFuture` or something similar.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
     return output;
 }
 
+struct StreamVersionData {
```
vasil-pashov, line 123: I'd split the whole thing into a cpp and header and put the implementation in the cpp.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    explicit StreamVersionData(const pipelines::VersionQuery& version_query) {
+        react(version_query);
+    }
+
+    StreamVersionData() = default;
+
+    size_t count_ = 0;
+    LoadParameter load_param_ = LoadParameter{LoadType::LOAD_LATEST_UNDELETED};
```
vasil-pashov, line 131: Are those supposed to be used outside of the struct? Should we make them private?

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+        ++count_;
+        switch(load_param_.load_type_) {
+        case LoadType::LOAD_LATEST_UNDELETED:
+            load_param_ = LoadParameter{LoadType::LOAD_DOWNTO, specific_version.version_id_};
+            break;
+        case LoadType::LOAD_DOWNTO:
+            util::check(load_param_.load_until_.has_value(), "Expect LOAD_DOWNTO to have version specificed");
+            load_param_.load_until_ = std::min(load_param_.load_until_.value(), specific_version.version_id_);
+            break;
+        case LoadType::LOAD_FROM_TIME:
```
vasil-pashov, line 153: I like annotating these cases with the [fallthrough](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) attribute

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    ARCTICDB_SAMPLE(BatchGetVersion, 0)
+    util::check(symbols.size() == version_queries.size(), "Symbol and version query list mismatch: {} != {}", symbols.size(), version_queries.size());
+
+    robin_hood::unordered_flat_map<StreamId, StreamVersionData> version_data;
+    for(const auto& symbol : folly::enumerate(symbols)) {
+        auto it = version_data.find(*symbol);
+        if(it == version_data.end())
+            version_data.insert(robin_hood::pair<StreamId, StreamVersionData>(*symbol, StreamVersionData{version_queries[symbol.index]}));
+        else
+            it->second.react(version_queries[symbol.index]);
```
vasil-pashov, line 216: I'd wrap those in curly braces even in the statements are one liners.
```suggestion
        if(it == version_data.end()) {
            version_data.insert(robin_hood::pair<StreamId, StreamVersionData>(*symbol, StreamVersionData{version_queries[symbol.index]}));
        } else {
            it->second.react(version_queries[symbol.index]);
        }
```

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -117,6 +120,130 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
+    for(const auto& symbol : folly::enumerate(symbols)) {
+        const auto it = version_data.find(*symbol);
+        util::check(it != version_data.end(), "Missing version data for symbol {}", *symbol);
+        auto version_entry_fut = folly::Future<std::shared_ptr<VersionMapEntry>>::makeEmpty();
+        if(it->second.count_ == 1) {
+            version_entry_fut = async::submit_io_task(CheckReloadTask{store, version_map, *symbol, it->second.load_param_});
+        } else {
+            auto fut = shared_futures.find(*symbol);
+            if(fut == shared_futures.end()) {
+                auto [splitter, inserted] = shared_futures.emplace(*symbol, folly::FutureSplitter(async::submit_io_task(CheckReloadTask{store, version_map, *symbol, it->second.load_param_})));
```
vasil-pashov, line 232: Any chance of breaking this line into a few statements?

cpp/arcticdb/version/test/test_version_map_batch.cpp
```diff
@@ -0,0 +1,433 @@
...
+    ASSERT_EQ(versions[3]->id(), StreamId{"stream_3"});
+    ASSERT_EQ(versions[3]->version_id(), 7);
+
+    ASSERT_EQ(versions[4]->id(), StreamId{"stream_8"});
+    ASSERT_EQ(versions[4]->version_id(), 4);
+}
+
+TEST_F(VersionMapBatchStore, SimpleTimestampQueries) {
+    auto store = test_store_->_test_get_store();
+    auto version_map = std::make_shared<VersionMap>();
```
vasil-pashov, line 83: These two lines are repeated in every test. We should add them in the fixture as protected members to reduce code duplication.

cpp/arcticdb/version/test/test_version_map_batch.cpp
```diff
@@ -0,0 +1,433 @@
...
+    size_t start = 0u) {
+    for(auto i = start; i < start + num_versions; ++i) {
+        version_map->write_version(store, test_index_key(stream_id, i));
+    }
+};
+}
+
+TEST_F(VersionMapBatchStore, SimpleVersionIdQueries) {
+    auto store = test_store_->_test_get_store();
+    auto version_map = std::make_shared<VersionMap>();
```
vasil-pashov, line 37: These two lines are repeated in every test. We should add them in the fixture as protected members to reduce code duplication.

cpp/arcticdb/version/test/test_version_map_batch.cpp
```diff
@@ -0,0 +1,433 @@
...
+    stream_ids.push_back(StreamId{"stream_1"});
+    version_queries.push_back(VersionQuery{SpecificVersionQuery{4}, false, false});
+    stream_ids.push_back(StreamId{"stream_4"});
+    version_queries.push_back(VersionQuery{SpecificVersionQuery{4}, false, false});
+    stream_ids.push_back(StreamId{"stream_1"});
+    version_queries.push_back(VersionQuery{SpecificVersionQuery{1}, false, false});
+    stream_ids.push_back(StreamId{"stream_3"});
+    version_queries.push_back(VersionQuery{SpecificVersionQuery{7}, false, false});
+    stream_ids.push_back(StreamId{"stream_8"});
+    version_queries.push_back(VersionQuery{SpecificVersionQuery{4}, false, false});
```
vasil-pashov, line 61: This whole thing is copy pasted into the next test case. We should really make it part of the setup function of a fixture. Same goes for the other pair of tests we have (`MultipleVersionsSameSymbolVersionIdQueries`, `MultipleVersionsSameSymbolTimestampQueries`). My suggestion is to add 
```c++
    auto store = test_store_->_test_get_store();
```
into a fixture and then make two other fixture which inherit from it: one for the `Simple*` test cases and one for the `MultipleVersionsSameSymbol*` test cases.

cpp/arcticdb/version/local_versioned_engine.cpp
willdealtry: I'm conflicted about this, as I think sometimes an alias in a signature makes it harder to read, since you have to go and look up what the type is somewhere else.

cpp/arcticdb/version/test/test_version_map_batch.cpp
willdealtry: I mind repetition much less in tests than I do in production code. Personally I feel like some of the rules like DRY and no explicit strings that are mandatory for real code are actually reversed in test code, where you want everything to be obvious and local; I don't really want to have to go and look elsewhere to see what a test is doing when it fails.

mehertz:
Detail all included changes in the PR description. Don't leave it empty please.

mehertz (changes requested):
Detail all included changes in the PR description. Don't leave it empty please.

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: @willdealtry what is exactly the reason why version_map/version_map_batch methods don't have any .cpp file currently?

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: In this particular case, we are adding a break; thought...but could be a good practice

cpp/arcticdb/version/version_map_batch_methods.hpp
mehertz: tests shouldn't change the structure of the code in my opinion. Is there another way we can structure the tests?

cpp/arcticdb/version/version_map_batch_methods.hpp
mehertz: Okay, having spoken to Gem these changes are also included in #415. This is an additional commit on top of that.

cpp/arcticdb/version/version_map_batch_methods.hpp
willdealtry: I think if we do that then we probably need to move StreamVersionData to its own files, otherwise it either looks like it's part of the API of the stream version methods (which it isn't), or becomes impossible to test.

mehertz:
To be re-opened - depends on #415

mehertz:
To be re-opened - depends on #415

GemDot:
reopening as dependent of PRs that have been merged

GemDot:
reopening as dependent of PRs that have been merged

PR #427, opened 2023-05-25
Add support for Azure blob storage (Linux and Windows)

phoebusm:
All pytest test cases test against s3 storage, have been added Azure storage fixture as well. All tests against enumerator (Azurite) pass. However, until Azurite container is added in Github Action, those test cases won't pass in Github action. Separate PR will be raised for it. 
TODO:
- [x] Add Azurite in Github Action
- [x] python layer's config and pytest fixtures still require some work
- [ ] document

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -231,14 +231,15 @@ jobs:
           disk-root: "D:"  # This is also the checkout directory. Total size 12GB.
         continue-on-error: true
 
-      - name: Install azurite
+      - name: Install npm # Linux github runner image does not come with npm
         if: matrix.os == 'linux'
-        run: |
-          yum install -y npm
-          npm install -g azurite
+        uses: actions/setup-node@v3.3.0
+        with:
+          node-version: '16'
 
       - name: Install the wheel and dependencies
         run: |
+          npm install -g azurite
           cmake -P cpp/CMake/CpuCount.cmake | sed 's/^-- //' | tee -a $GITHUB_ENV
           python -V
           cd "$RUNNER_TEMP" # Works for Windows-style paths as well
--- a/cpp/CMake/AzureVcpkg.cmake
+++ b/cpp/CMake/AzureVcpkg.cmake
@@ -1,0 +1,169 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# We need to know an absolute path to our repo root to do things like referencing ./LICENSE.txt file.
+set(AZ_ROOT_DIR "${CMAKE_CURRENT_LIST_DIR}/..")
+
+macro(az_vcpkg_integrate)
+  message("Vcpkg integrate step.")
+  # AUTO CMAKE_TOOLCHAIN_FILE:
+  #   User can call `cmake -DCMAKE_TOOLCHAIN_FILE="path_to_the_toolchain"` as the most specific scenario.
+  #   As the last alternative (default case), Azure SDK will automatically clone VCPKG folder and set toolchain from there.
+  if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
+    message("CMAKE_TOOLCHAIN_FILE is not defined. Define it for the user.")
+    # Set AZURE_SDK_DISABLE_AUTO_VCPKG env var to avoid Azure SDK from cloning and setting VCPKG automatically
+    # This option delegate package's dependencies installation to user.
+    if(NOT DEFINED ENV{AZURE_SDK_DISABLE_AUTO_VCPKG})
+      message("AZURE_SDK_DISABLE_AUTO_VCPKG is not defined. Fetch a local copy of vcpkg.")
+      # GET VCPKG FROM SOURCE
+      #  User can set env var AZURE_SDK_VCPKG_COMMIT to pick the VCPKG commit to fetch
+      set(VCPKG_COMMIT_STRING 94ce0dab56f4d8ba6bd631ba59ed682b02d45c46) # default SDK tested commit
+      if(DEFINED ENV{AZURE_SDK_VCPKG_COMMIT})
+        message("AZURE_SDK_VCPKG_COMMIT is defined. Using that instead of the default.")
+        set(VCPKG_COMMIT_STRING "$ENV{AZURE_SDK_VCPKG_COMMIT}") # default SDK tested commit
+      endif()
+      message("Vcpkg commit string used: ${VCPKG_COMMIT_STRING}")
+      include(FetchContent)
+      FetchContent_Declare(
+          vcpkg
+          GIT_REPOSITORY      https://github.com/microsoft/vcpkg.git
+          GIT_TAG             ${VCPKG_COMMIT_STRING}
+          )
+      FetchContent_GetProperties(vcpkg)
+      # make sure to pull vcpkg only once.
+      if(NOT vcpkg_POPULATED) 
+          FetchContent_Populate(vcpkg)
+      endif()
+      # use the vcpkg source path 
+      set(CMAKE_TOOLCHAIN_FILE "${vcpkg_SOURCE_DIR}/scripts/buildsystems/vcpkg.cmake" CACHE STRING "")
+    endif()
+  endif()
+
+  # enable triplet customization
+  if(DEFINED ENV{VCPKG_DEFAULT_TRIPLET} AND NOT DEFINED VCPKG_TARGET_TRIPLET)
+    set(VCPKG_TARGET_TRIPLET "$ENV{VCPKG_DEFAULT_TRIPLET}" CACHE STRING "")
+  endif()
+endmacro()
+
+macro(az_vcpkg_portfile_prep targetName fileName contentToRemove)
+  # with sdk/<lib>/vcpkg/<fileName>
+  file(READ "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/${fileName}" fileContents)
+
+  # Windows -> Unix line endings
+  string(FIND fileContents "\r\n" crLfPos)
+
+  if (crLfPos GREATER -1)
+    string(REPLACE "\r\n" "\n" fileContents ${fileContents})
+  endif()
+
+  # remove comment header
+  string(REPLACE "${contentToRemove}" "" fileContents ${fileContents})
+
+  # undo Windows -> Unix line endings (if applicable)
+  if (crLfPos GREATER -1)
+    string(REPLACE "\n" "\r\n" fileContents ${fileContents})
+  endif()
+  unset(crLfPos)
+
+  # output to an intermediate location
+  file (WRITE "${CMAKE_BINARY_DIR}/vcpkg_prep/${targetName}/${fileName}" ${fileContents})
+  unset(fileContents)
+
+  # Produce the files to help with the vcpkg release.
+  # Go to the /out/build/<cfg>/vcpkg directory, and copy (merge) "ports" folder to the vcpkg repo.
+  # Then, update the portfile.cmake file SHA512 from "1" to the actual hash (a good way to do it is to uninstall a package,
+  # clean vcpkg/downloads, vcpkg/buildtrees, run "vcpkg install <pkg>", and get the SHA from the error message).
+  configure_file(
+    "${CMAKE_BINARY_DIR}/vcpkg_prep/${targetName}/${fileName}"
+    "${CMAKE_BINARY_DIR}/vcpkg/ports/${targetName}-cpp/${fileName}"
+    @ONLY
+  )
+endmacro()
+
+macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
+  foreach(vcpkgFile "vcpkg.json" "portfile.cmake")
+    az_vcpkg_portfile_prep(
+      "${targetName}"
+      "${vcpkgFile}"
+      "# Copyright (c) Microsoft Corporation. All rights reserved.\n# SPDX-License-Identifier: MIT\n\n"
+    )
+  endforeach()
+
+  # Standard names for folders such as "bin", "lib", "include". We could hardcode, but some other libs use it too (curl).
+  include(GNUInstallDirs)
+
+  # When installing, copy our "inc" directory (headers) to "include" directory at the install location.
+  install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/inc/azure/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/azure")
+
+  # Copy license as "copyright" (vcpkg dictates naming and location).
+  install(FILES "${AZ_ROOT_DIR}/LICENSE.txt" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/${targetName}-cpp" RENAME "copyright")
+
+  # Indicate where to install targets. Mirrors what other ports do.
+  install(
+    TARGETS "${targetName}"
+      EXPORT "${targetName}-cppTargets"
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} # DLLs (if produced by build) go to "/bin"
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} # static .lib files
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} # .lib files for DLL build
+        INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR} # headers
+    )
+
+  # If building a Windows DLL, patch the dll_import_export.hpp
+  if(WIN32 AND BUILD_SHARED_LIBS)
+    add_compile_definitions(AZ_${macroNamePart}_BEING_BUILT)
+    target_compile_definitions(${targetName} PUBLIC AZ_${macroNamePart}_DLL)
+
+    set(AZ_${macroNamePart}_DLL_INSTALLED_AS_PACKAGE "*/ + 1 /*")
+    configure_file(
+      "${CMAKE_CURRENT_SOURCE_DIR}/inc/${dllImportExportHeaderPath}"
+      "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${dllImportExportHeaderPath}"
+      @ONLY
+    )
+    unset(AZ_${macroNamePart}_DLL_INSTALLED_AS_PACKAGE)
+
+    get_filename_component(dllImportExportHeaderDir ${dllImportExportHeaderPath} DIRECTORY)
+    install(
+        FILES "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_INCLUDEDIR}/${dllImportExportHeaderPath}"
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${dllImportExportHeaderDir}"
+    )
+    unset(dllImportExportHeaderDir)
+  endif()
+
+  # Export the targets file itself.
+  install(
+    EXPORT "${targetName}-cppTargets"
+      DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/${targetName}-cpp"
+      NAMESPACE Azure:: # Not the C++ namespace, but a namespace in terms of cmake.
+      FILE "${targetName}-cppTargets.cmake"
+    )
+
+  # configure_package_config_file(), write_basic_package_version_file()
+  include(CMakePackageConfigHelpers)
+
+  # Produce package config file.
+  configure_package_config_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/Config.cmake.in"
+    "${targetName}-cppConfig.cmake"
+    INSTALL_DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/${targetName}-cpp"
+    PATH_VARS
+      CMAKE_INSTALL_LIBDIR)
+
+  # Produce version file.
+  write_basic_package_version_file(
+    "${targetName}-cppConfigVersion.cmake"
+      VERSION ${AZ_LIBRARY_VERSION} # the version that we extracted from package_version.hpp
+      COMPATIBILITY SameMajorVersion
+    )
+
+  # Install package config and version files.
+  install(
+      FILES
+        "${CMAKE_CURRENT_BINARY_DIR}/${targetName}-cppConfig.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/${targetName}-cppConfigVersion.cmake"
+      DESTINATION
+        "${CMAKE_INSTALL_DATAROOTDIR}/${targetName}-cpp" # to shares/<our_pkg>
+    )
+
+  # Export all the installs above as package.
+  export(PACKAGE "${targetName}-cpp")
+endmacro()
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -58,6 +58,9 @@ include(PythonUtils) # Must be called before Pybind (third_party) to override it
 
 add_subdirectory(third_party)
 add_subdirectory(proto)
+if(NOT ${ARCTICDB_USING_CONDA})
+    include(AzureVcpkg) #AzureVcpkg.cmake is from https://github.com/Azure/azure-sdk-for-cpp/blob/main/cmake-modules/AzureVcpkg.cmake, commit ada77b3
+endif()
 
 python_utils_dump_vars_if_enabled("After Pybind")
 python_utils_check_include_dirs("accepted by pybind")
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -101,6 +101,7 @@ else ()
     find_library(Sasl2_LIBRARY NAMES sasl2 libsasl2 PATHS /usr/local/lib/libsasl2.a REQUIRED)
 endif ()
 
+
 find_path(BITMAGIC_INCLUDE_DIRS "bitmagic/bm.h")
 
 set(FIND_LIBRARY_USE_LIB64_PATHS ON)
@@ -247,13 +248,13 @@ set(arcticdb_srcs
         storage/mongo/mongo_client.hpp
         storage/mongo/mongo_instance.hpp
         storage/mongo/mongo_storage.hpp
+        storage/object_store_utils.hpp
         storage/s3/nfs_backed_storage.hpp
         storage/s3/nfs_backed_storage-inl.hpp
         storage/s3/s3_client_accessor.hpp
         storage/s3/s3_storage-inl.hpp
         storage/s3/s3_storage_tool.hpp
         storage/s3/s3_storage_tool-inl.hpp
-        storage/s3/s3_utils.hpp
         storage/storage_factory.hpp
         storage/storage_options.hpp
         storage/storage.hpp
@@ -423,9 +424,12 @@ set(arcticdb_srcs
         version/symbol_list.cpp)
 
 # Exclude vendored sources when using conda.
-if(NOT ${ARCTICDB_USING_CONDA})
+if(NOT ${ARCTICDB_USING_CONDA}) #Awaiting Azure sdk support in conda https://github.com/man-group/ArcticDB/issues/519
     list (APPEND arcticdb_srcs
         util/third_party/robin_hood.hpp
+        storage/azure/azure_storage-inl.hpp
+        storage/azure/azure_storage.hpp
+        storage/azure/azure_storage.cpp
     )
 endif()
 
@@ -484,6 +488,15 @@ set (arcticdb_core_libraries
         spdlog::spdlog
         Folly::folly # Transitively includes: double-conversion, gflags, glog, libevent, libssl, libcrypto, libiberty, libsodium
         )
+
+if(NOT ${ARCTICDB_USING_CONDA})  #Awaiting Azure sdk support in conda https://github.com/man-group/ArcticDB/issues/519
+    find_package(azure-identity-cpp CONFIG REQUIRED)
+    find_package(azure-storage-blobs-cpp CONFIG REQUIRED)
+    list (APPEND arcticdb_core_libraries
+        Azure::azure-identity
+        Azure::azure-storage-blobs
+    )
+endif()
 
 # Dependencies expose different names of their libraries depending
 # on their origin (i.e. vendored source or packages on conda-forge).
@@ -550,8 +563,8 @@ if (SSL_LINK)
         list(APPEND arcticdb_core_includes  ${KERBEROS_INCLUDE_DIR})
     endif()
 endif ()
-
 target_link_libraries(arcticdb_core_object PUBLIC ${arcticdb_core_libraries})
+
 
 target_include_directories(arcticdb_core_object
         PRIVATE
--- a/cpp/arcticdb/storage/azure/azure_storage-inl.hpp
+++ b/cpp/arcticdb/storage/azure/azure_storage-inl.hpp
@@ -1,0 +1,329 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#ifndef ARCTICDB_AZURE_STORAGE_H_
+#error "This should only be included by file_storage.hpp"
+#endif
+
+#include <arcticdb/util/preconditions.hpp>
+#include <arcticdb/util/pb_util.hpp>
+#include <arcticdb/log/log.hpp>
+#include <arcticdb/util/buffer_pool.hpp>
+
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
+#include <folly/gen/Base.h>
+#include <arcticdb/storage/storage_utils.hpp>
+#include <arcticdb/storage/object_store_utils.hpp>
+#include <arcticdb/storage/storage_options.hpp>
+#include <arcticdb/entity/serialized_key.hpp>
+#include <arcticdb/util/exponential_backoff.hpp>
+#include <arcticdb/util/configs_map.hpp>
+#include <arcticdb/util/composite.hpp>
+
+#include <azure/core.hpp>
+#include <azure/storage/blobs.hpp>
+
+#include <boost/interprocess/streams/bufferstream.hpp>
+#include <folly/ThreadLocal.h>
+
+#undef GetMessage
+
+namespace arcticdb::storage {
+
+using namespace object_store_utils;
+
+namespace azure {
+
+namespace fg = folly::gen;
+
+namespace detail {
+
+static const size_t BATCH_SUBREQUEST_LIMIT = 1000;
+
+inline Azure::Core::Context get_context(unsigned int request_timeout){
+    Azure::Core::Context requestContext; //TODO: Maybe can be static but need to be careful with its shared_ptr and ContextSharedState
+    return requestContext.WithDeadline(std::chrono::system_clock::now() + std::chrono::milliseconds(request_timeout));
+}
+
+template<class KeyBucketizer>
+void do_write_impl(
+    Composite<KeySegmentPair>&& kvs,
+    const std::string& root_folder,
+    Azure::Storage::Blobs::BlobContainerClient& container_client,
+    KeyBucketizer&& bucketizer,
+    const Azure::Storage::Blobs::UploadBlockBlobFromOptions& upload_option,
+    unsigned int request_timeout) {
+        ARCTICDB_SAMPLE(AzureStorageWrite, 0)
+        auto fmt_db = [](auto&& kv) { return kv.key_type(); };
+
+        (fg::from(kvs.as_range()) | fg::move | fg::groupBy(fmt_db)).foreach(
+            [&container_client, &root_folder, b=std::move(bucketizer), &upload_option, &request_timeout] (auto&& group) {
+                auto key_type_dir = key_type_folder(root_folder, group.key());
+                ARCTICDB_TRACE(log::storage(), "Azure key_type_folder is {}", key_type_dir);
+
+                ARCTICDB_SUBSAMPLE(AzureStorageWriteValues, 0)
+                for (auto& kv : group.values()) {
+                    auto& k = kv.variant_key();
+                    auto blob_name = object_path(b.bucketize(key_type_dir, k), k);
+                    auto& seg = kv.segment();
+
+                    std::shared_ptr<Buffer> tmp;
+                    auto hdr_size = seg.segment_header_bytes_size();
+                    auto [dst, write_size] = seg.try_internal_write(tmp, hdr_size);
+                    util::check(arcticdb::Segment::FIXED_HEADER_SIZE + hdr_size + seg.buffer().bytes() <= write_size,
+                                    "Size disparity, fixed header size {} + variable header size {} + buffer size {}  >= total size {}",
+                                    arcticdb::Segment::FIXED_HEADER_SIZE,
+                                    hdr_size,
+                                    seg.buffer().bytes(),
+                                    write_size);
+
+                    ARCTICDB_SUBSAMPLE(AzureStorageUploadObject, 0)
+                    try{
+                        auto blob_client = container_client.GetBlockBlobClient(blob_name);
+                        ARCTICDB_RUNTIME_DEBUG(log::storage(), "Writing key {}: {}, with {} bytes of data",
+                                               variant_key_type(k),
+                                               variant_key_view(k),
+                                               seg.total_segment_size(hdr_size));
+                        blob_client.UploadFrom(dst, write_size, upload_option, get_context(request_timeout));
+                    }
+                    catch (const Azure::Core::RequestFailedException& e){
+                        util::raise_rte("Failed to write azure with key '{}' {} {}: {}",
+                                            k,
+                                            blob_name,
+                                            static_cast<int>(e.StatusCode),
+                                            e.ReasonPhrase);
+                    }
+
+                    ARCTICDB_RUNTIME_DEBUG(log::storage(), "Wrote key {}: {}, with {} bytes of data",
+                                        variant_key_type(k),
+                                        variant_key_view(k),
+                                        seg.total_segment_size(hdr_size));
+                }
+            }
+        );
+}
+
+template<class KeyBucketizer>
+void do_update_impl(
+    Composite<KeySegmentPair>&& kvs,
+    const std::string& root_folder,
+    Azure::Storage::Blobs::BlobContainerClient& container_client,
+    KeyBucketizer&& bucketizer,
+    const Azure::Storage::Blobs::UploadBlockBlobFromOptions& upload_option,
+    unsigned int request_timeout) {
+        // azure updates the key if it already exists
+        do_write_impl(std::move(kvs), root_folder, container_client, std::move(bucketizer), upload_option, request_timeout);
+}
+
+struct UnexpectedAzureErrorException : public std::exception {};
+
+template<class Visitor, class KeyBucketizer>
+void do_read_impl(Composite<VariantKey> && ks,
+    Visitor&& visitor,
+    const std::string& root_folder,
+    Azure::Storage::Blobs::BlobContainerClient& container_client,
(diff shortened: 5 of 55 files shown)


cpp/arcticdb/cmake-modules/AzureVcpkg.cmake
qc00, line 1: What is this for? Typically a `vcpkg`-packaged dependency can be directly consumed without copying helper cmake scripts.

phoebusm: https://github.com/Azure/azure-sdk-for-cpp
It is the recommended way.

qc00: Gosh... M$ people can't even use their own tools properly...

Please do these instead:
1. Move this file to the existing `cpp/CMake` folder
2. Revert the 3 lines they asked you to add on our `arcticdb/CMakeLists.txt`
3. Add `include(AzureVcpkg)` to `cpp/CMakeLists.txt` after `add_subdirectory(proto)`

cpp/arcticdb/CMakeLists.txt
```diff
@@ -208,6 +208,7 @@ set(arcticdb_srcs
         storage/common.hpp
         storage/config_resolvers.hpp
         storage/failure_simulation.hpp
+        storage/object_store_utils.hpp
```
qc00, line 211: These lists are sorted alphabetically

cpp/arcticdb/CMakeLists.txt
```diff
@@ -509,6 +512,19 @@ endif ()
...
+        Azure::azure-identity
+        Azure::azure-storage-blobs
+        )
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake-modules")
+include(AzureVcpkg)
+az_vcpkg_integrate()
+
+find_package(azure-identity-cpp CONFIG REQUIRED)
+find_package(azure-storage-blobs-cpp CONFIG REQUIRED)
```
qc00, line 525: Finding dependencies should happen at the very top, along with the other dependencies.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -509,6 +512,19 @@ endif ()
 
 target_link_libraries(arcticdb_core_object PUBLIC ${arcticdb_core_libraries})
 
+set (azure_package
+        Azure::azure-identity
+        Azure::azure-storage-blobs
```
qc00, line 517: Just add these to the `arcticdb_core_libraries` list like *all* the other dependencies and get rid of the extra call to `target_link_libraries` below.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -791,8 +812,9 @@ if(${TEST})
             rapidcheck_gtest
             PRIVATE
             ${AWSSDK_LINK_LIBRARIES}
+            ${azure_package}
             )
     endif()
 
     gtest_discover_tests(arcticdb_rapidcheck_tests PROPERTIES DISCOVERY_TIMEOUT 30)
-endif()
+endif()
```
qc00, line 820: Please configure your IDE not to remove the new line at the end of files.

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
```diff
@@ -0,0 +1,340 @@
...
+}
+
+template<class KeyBucketizer>
+void do_write_impl(
+    Composite<KeySegmentPair>&& kvs,
+    const std::string& root_folder,
+    Azure::Storage::Blobs::BlobContainerClient& container_client,
+    KeyBucketizer&& bucketizer,
+    const Azure::Storage::Blobs::UploadBlockBlobFromOptions& upload_option,
+    unsigned int request_timeout) {
```
qc00, line 55: Please indent the argument list differently from the method body.

Our old code doesn't do it for historical reasons.

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
qc00, line 1: There is a lot of duplication from the S3 code. Please extract the common stuff.

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+    (fg::from(ks.as_range()) | fg::move | fg::groupBy(fmt_db)).foreach(
+        [&container_client, &root_folder, b=std::move(bucketizer), &visitor, &failed_reads,
+         opts=opts, &download_option, &request_timeout] (auto&& group) {
+        for (auto& k : group.values()) {
+            auto key_type_folder = object_store_utils::key_type_folder(root_folder, variant_key_type(k));
+            auto blob_name = object_store_utils::object_path(b.bucketize(key_type_folder, k), k);
+            try{
+                auto blob_client = container_client.GetBlockBlobClient(blob_name);
+                auto properties = blob_client.GetProperties(Azure::Storage::Blobs::GetBlobPropertiesOptions{}, get_context(request_timeout)).Value;
```
qc00, line 136: We can't really make a separate request just to allocate a buffer...

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
```diff
@@ -0,0 +1,340 @@
...
+void do_remove_impl(Composite<VariantKey>&& ks,
+                    const std::string& root_folder,
+                    Azure::Storage::Blobs::BlobContainerClient& container_client,
+                    KeyBucketizer&& bucketizer,
+                    bool connect_to_azurite,
+                    unsigned int request_timeout) {
+    ARCTICDB_SUBSAMPLE(AzureStorageDeleteBatch, 0)
+    auto fmt_db = [](auto&& k) { return variant_key_type(k); };
+    std::vector<VariantKey> failed_deletes;
+    if (connect_to_azurite){ //https://github.com/Azure/Azurite/issues/1822 due to an open issue of Azurite not filling subrequeset response after batch delete
```
qc00, line 170: We can't really branch off an entire function for a test fixture library. It defeats the purpose of testing...

cpp/arcticdb/storage/s3/s3_storage-inl.hpp
```diff
@@ -71,13 +62,13 @@ void do_write_impl(
 
     (fg::from(kvs.as_range()) | fg::move | fg::groupBy(fmt_db)).foreach(
         [&s3_client, &bucket_name, &root_folder, b=std::move(bucketizer)] (auto&& group) {
-        auto key_type_folder = s3_key_type_folder(root_folder, group.key());
+        auto key_type_folder = object_store_utils::key_type_folder(root_folder, group.key());
```
qc00, line 65: Might want to do `using namespace` at the top instead.

cpp/arcticdb/storage/storage_factory.cpp
```diff
@@ -54,6 +55,12 @@ std::shared_ptr<VariantStorageFactory> create_storage_factory(
         res = std::make_shared<VariantStorageFactory>(
             nfs_backed::NfsBackedStorageFactory(nfs_backed_config)
         );
+    } else if (type_name == azure::AzureStorage::Config::descriptor()->full_name()) {
+        azure::AzureStorage::Config azure_config;
+        storage.config().UnpackTo(&azure_config);
+        res = std::make_shared<VariantStorageFactory>(
+            azure::AzureStorageFactory(azure_config)
+        );
```
qc00, line 63: It was probably fine when we had two storage types, but in a language with macros and templates, this level of manual copy and paste makes me sad.

cpp/proto/arcticc/pb2/azure_storage.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package arcticc.pb2.azure_storage_pb2;
+
+message Config {
+    string container_name = 1;
+    string credential_name = 2;
+    string credential_key = 3;
+    string endpoint = 4;
+    uint32 max_connections = 5;
```
qc00, line 10: I know this is copied from S3, but I wonder if it's correct to have it in the permanent storage config.

cpp/proto/arcticc/pb2/azure_storage.proto
```diff
@@ -0,0 +1,15 @@
...
+    uint32 request_timeout = 6;
+    string prefix = 7;
+    bool https = 8;
+    bool connect_to_azurite = 9;
```
qc00, line 14: Can't the `endpoint` take care of it? If not, it's a pretty poorly designed mock library....

cpp/vcpkg.json
```diff
@@ -157,7 +159,9 @@
     { "name": "s2n", "version": "1.3.5" },
     { "name": "spdlog", "version": "1.11.0" },
     { "name": "wangle", "version-string": "2022.03.21.00" },
-    { "name": "zlib", "version": "1.2.13" }
+    { "name": "zlib", "version": "1.2.13" },
+    { "name": "azure-identity-cpp", "version": "1.3.0" },
+    { "name": "azure-storage-blobs-cpp", "version": "12.6.2" }
```
qc00, line 164: It's not necessary to pin the versions of new dependencies. The existing ones are pinned because we haven't had time to test their latest versions.

python/tests/stress/arcticdb/version_store/test_stress_multicolumn.py
```diff
@@ -38,7 +38,16 @@ def make_periods(start_date, end_date, freq, range_type="b"):
...
 
-@pytest.mark.parametrize("lib_type", ["lmdb_version_store_big_map", "s3_version_store", "s3_version_store"])
+@pytest.mark.parametrize(
+    "lib_type",
+    [
+        "lmdb_version_store_big_map",
+        "s3_version_store",
+        "s3_version_store",
+        "azure_version_store",
+        "azure_version_store",
```
qc00, line 48: Have you figured out why `s3_version_store` was duplicated to begin with?

python/tests/integration/toolbox/test_library_tool.py
```diff
@@ -26,7 +26,7 @@ def get_log_types():
     return [KeyType.LOG, KeyType.LOG_COMPACTED]
 
 
-@pytest.mark.parametrize("lib_type", ["lmdb_version_store", "s3_version_store"])
+@pytest.mark.parametrize("lib_type", ["lmdb_version_store", "s3_version_store", "azure_version_store"])
```
qc00, line 29: Can probably just use `object_version_store` instead of this manual `request.getfixturevalue` indirection.

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -112,17 +112,17 @@ def test_hypothesis_max_min_agg(lmdb_version_store, df):
     assert_frame_equal(expected, vit.data)
 
 
-def test_sum_aggregation(s3_version_store):
+def test_sum_aggregation(object_version_store):
```
qc00, line 115: Most of the unit tests are not storage-dependent, so we should not just blindly replace all of them with `object_version_store`, making them take twice as long with limited additional bug detection power.

python/tests/conftest.py
```diff
@@ -278,6 +306,11 @@ def s3_version_store_prune_previous(s3_store_factory):
     return s3_store_factory(prune_previous_version=True)
 
 
+@pytest.fixture(scope="function")
```
qc00, line 309: `scope="function"` is the default

python/tests/conftest.py
```diff
@@ -235,6 +248,21 @@ def s3_store_factory(lib_name, arcticdb_test_s3_config):
             lib.version_store.clear()
 
 
+@pytest.fixture
+def azure_store_factory(lib_name, arcticdb_test_azure_config):
+    """Factory to create any number of S3 libs with the given WriteOptions or VersionStoreConfig.
```
qc00, line 253: S3?

python/tests/conftest.py
```diff
@@ -156,6 +161,14 @@ def create(lib_name):
     return create
 
 
+@pytest.fixture
+def arcticdb_test_azure_config():
+    def create(lib_name):
+        return create_test_azure_cfg(lib_name)
+
+    return create
```
qc00, line 169: This is equivalent to returning `create_test_azure_cfg`

python/tests/stress/arcticdb/version_store/test_stress_multicolumn.py
```diff
@@ -38,7 +39,18 @@ def make_periods(start_date, end_date, freq, range_type="b"):
     return [r for r in ranges if len(r) > 0]
 
 
-@pytest.mark.parametrize("lib_type", ["lmdb_version_store_big_map", "s3_version_store", "s3_version_store"])
```
qc00, line 41: Please note the original duplicated `s3_version_store` is just a mistake when the `mongo_version_store` was mechanically replaced for the external release.

Since you're touching this line, you should fix that by removing the duplication.

Note: I am reviewing from the last file up and there are many similar examples above.

python/tests/integration/toolbox/test_library_tool.py
```diff
@@ -82,18 +76,18 @@ def test_write_keys(lib_type, request):
     assert len(new_keys) == len(all_keys)
 
 
-@pytest.mark.parametrize("lib_type", ["lmdb_version_store", "s3_version_store"])
-def test_count_keys(lib_type, request):
+def test_count_keys(object_and_lmdb_version_store):
     df = sample_dataframe()
```
qc00, line 80: It will be easier to review if you do `lib = object_and_lmdb_version_store` here?

This is a common pattern used in a lot of tests as the fixture name can get rather long.

Note: I am reviewing from the last file up and there are many similar examples above.

cpp/arcticdb/storage/azure/azure_storage.hpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+template<typename ConfigType>
+std::shared_ptr<Azure::Storage::StorageSharedKeyCredential> get_azure_credentials(const ConfigType& conf) {
+    return std::make_shared<Azure::Storage::StorageSharedKeyCredential>(conf.credential_name(), conf.credential_key());
+}
+
+} //namespace arcticdb::azure
+
+#define ARCTICDB_AZURE_STORAGE_H_
```
qc00, line 126: Is this ever used?

python/tests/conftest.py
```diff
@@ -117,6 +125,38 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
     ] + ":" + bucket + "?access=" + aws_access_key + "&secret=" + aws_secret_key + "&port=" + port
 
 
+@pytest.fixture
+def moto_azure_endpoint_and_credentials(azurite_port, spawn_azurite):
```
qc00, line 129: Note `moto` is the library that emulates AWS S3 (like what Azurite does for Azure), so this name makes no sense.

python/tests/conftest.py
```diff
@@ -117,6 +125,38 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
...
+    global BUCKET_ID
+
+    container = f"testbucket{BUCKET_ID}"
+    BUCKET_ID = BUCKET_ID + 1
```
qc00, line 133: All caps are normally used for constants

python/tests/conftest.py
```diff
@@ -117,6 +125,38 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
...
+
+    credential_name = "devstoreaccount1"
+    credential_key = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
+    endpoint = "0.0.0.0:" + str(azurite_port)
+    is_https = False
+    connect_to_azurite = True
+    yield endpoint, container, credential_name, credential_key, is_https, connect_to_azurite
```
qc00, line 140: I think you can use `return` here.

python/tests/conftest.py
```diff
@@ -117,6 +125,38 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
...
+        credential_name,
+        credential_key,
+        is_https,
+        connect_to_azurite,
+    ) = moto_azure_endpoint_and_credentials
+    yield "azure://" + endpoint + ":" + container + "?access=" + credential_name + "&secret=" + credential_key + "&https=" + str(
+        is_https
+    ) + "&connect_to_azurite=" + str(
+        connect_to_azurite
+    )
```
qc00, line 157: Use f-string.

cpp/proto/arcticc/pb2/azure_storage.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
```
mehertz, line 1: header

mehertz (changes requested):

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
phoebusm: Currently we can only test against Azurite (the enumerator)... I have raised this concern in the meeting with MS. I have test this function against a testing Azure storage account and it works fine.
In the near future I have to stick to Azurite for testing.
I understand that only code fully test should be released but it is a bit sad the code is not released due to the limit of enumerator. What is your thought?

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
qc00: To be more specific, all the HTTP-based blob storage should reuse the majority of the code, and the only difference should be: a) the encoding of the path and b) the actual SDK methods.

Of the 346 lines, I count about 16 lines (excluding the disputed Azurite-specific branch) that are dedicated to calling the Azure SDK. 

The lines in this file are concerned with:

* folly pipelines
* validations
* logging
* segment marshalling
* batching
* key -> path formatting

which should be the same in the S3 implementation.

A similar situation applies to the `azure_storage.?pp` files.

This calls for a different strategy of proper code reuse (than duplicating an entire storage implementation and replacing a few lines).

cpp/arcticdb/storage/storage_factory.cpp
phoebusm: I have the same thought but do not have a good idea, as I think using macro here will sacrifice too much readability but not gaining too much. After all this part of the code is very straight forward. It should not need much maintenance  anyway. Pretty much write and forget.

cpp/proto/arcticc/pb2/azure_storage.proto
phoebusm: I would say let's stick to S3 as Azure Storage support should be like a drop-in replacement for S3 here.

python/tests/stress/arcticdb/version_store/test_stress_multicolumn.py
phoebusm: I have figured this is a bad idea to just blindly copying everything from S3. So I have removed it in later commit.

python/tests/integration/toolbox/test_library_tool.py
phoebusm: Agree. So I have add a new fixture, similar to `object_version_store` but with `lmdb` additionally for it in later commit.

cpp/proto/arcticc/pb2/azure_storage.proto
qc00: I made this comment with the assumption that there should not be any Azurite-specific behaviour for it to be a meaningful mock...

cpp/vcpkg.json
qc00: No. Versions are already managed by the baseline.

The reason for these version overrides is explained on line 76/78. Your change is not related to that reason.

.gitignore
```diff
@@ -36,3 +36,4 @@ CMakeUserPresets.json
 
 *_pb2.py
 *.mdb
+azurite/
```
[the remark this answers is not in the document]
phoebusm: Azurite will create some temp files in that particular folder. They are for persistence but sadly we can't turn it off although we don't need it.
Normally my fixtures should clear them automatically but if the test crashes, they won't be cleared.
Therefore .gitignore has been updated

.gitignore
```diff
@@ -36,3 +36,4 @@ CMakeUserPresets.json
 
 *_pb2.py
 *.mdb
+azurite*/
```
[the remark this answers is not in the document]
phoebusm: I have thought about that. The temp files are created if the user runs the test locally as well. I suppose we shouldn't storing any temp files outside the repo folder, as the directory may not be available in linux distribution/Windows/Linux? Also I believe what happen in the repo, should stay in its folder as well. 
The setting has been changes in my IDE so the newline problem won't happen again.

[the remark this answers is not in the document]
phoebusm: Still, maybe I should give it a better naming for the temp folder, e.g. `temp/azurite`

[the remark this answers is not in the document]
qc00: I presume you start Azurite from Python? You can use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
phoebusm: Sadly because of the limitation by the SDK, we have to bite either of below two bullets:
1. 2 http requests (one for the size of the blob to allocate the buffer, one for the download)
2. Extra copy of object data

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
phoebusm: It has been amended. You may take a look again. More code is shared now.

cpp/CMake/AzureVcpkg.cmake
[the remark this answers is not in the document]
phoebusm: Added

cpp/arcticdb/storage/azure/azure_storage.hpp
phoebusm: Used in `azure_storage.hpp`, like other storage adapters

.gitignore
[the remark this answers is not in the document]
phoebusm: Resolved by making Azurite using a temporary directory

.github/workflows/build_steps.yml
```diff
@@ -231,10 +231,9 @@ jobs:
           disk-root: "D:"  # This is also the checkout directory. Total size 12GB.
         continue-on-error: true
 
+
       - name: Install azurite
-        if: matrix.os == 'linux'
-        run: |
-          yum install -y npm
+        uses: actions/setup-node@v3
           npm install -g azurite
```
qc00, line 237: The syntax is a bit broken: `uses:` calls an external action, so you will need a separate

```
- name: ...
  run: npm install -g azurite
```

However, AFAIK, all GitHub runner images have npm pre-installed. You don't need to do the `setup-node` (or `yum install`)

qc00 (commented):
I have reviewed almost all of the files, and they mostly look fine now.

The main thing left over is a design discussion with William.

However, I think it's also okay to merge the current design as a beta and refactor it later.

cpp/CMakeLists.txt
```diff
@@ -58,6 +58,9 @@ include(PythonUtils) # Must be called before Pybind (third_party) to override it
 
 add_subdirectory(third_party)
 add_subdirectory(proto)
+if(NOT ${ARCTICDB_USING_CONDA})
+    include(AzureVcpkg) #AzureVcpkg.cmake is from https://github.com/Azure/azure-sdk-for-cpp/blob/main/cmake-modules/AzureVcpkg.cmake, commit ada77b3
```
qc00, line 62: It would be better to put the comment in `AzureVcpkg.cmake` itself.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -403,6 +404,9 @@ set(arcticdb_srcs
 if(NOT ${ARCTICDB_USING_CONDA})
     list (APPEND arcticdb_srcs
         util/third_party/robin_hood.hpp
+        storage/azure/azure_storage-inl.hpp
+        storage/azure/azure_storage.hpp
+        storage/azure/azure_storage.cpp
```
qc00, line 409: Do we have an issue raised to support Azure in Conda? If so, could link it here.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -800,6 +813,7 @@ if(${TEST})
                 rapidcheck_gtest
                 PRIVATE
                 ${RAPIDCHECK_PRIVATE_LIBRARIES}
+                ${AWSSDK_LINK_LIBRARIES}
```
qc00, line 816: Just noticed this is saying AWS instead of Azure?

cpp/arcticdb/storage/storage_factory.cpp
```diff
@@ -54,6 +57,14 @@ std::shared_ptr<VariantStorageFactory> create_storage_factory(
...
             nfs_backed::NfsBackedStorageFactory(nfs_backed_config)
         );
+#ifndef ARCTICDB_USING_CONDA
+    } else if (type_name == azure::AzureStorage::Config::descriptor()->full_name()) {
+        azure::AzureStorage::Config azure_config;
+        storage.config().UnpackTo(&azure_config);
+        res = std::make_shared<VariantStorageFactory>(
+            azure::AzureStorageFactory(azure_config)
+        );
+#endif
```
qc00, line 67: It will be very helpful to add an `#else` here to explain Azure is not supported and a link to the GitHub issue for the user to follow.

cpp/proto/arcticc/pb2/azure_storage.proto
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright 2023 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
```
qc00, line 4: Minor: BTW, the license file is not actually at this location.

@mehertz We should really fix this. Maybe in Vasil's clang-format PR?

python/arcticdb/util/test.py
```diff
@@ -32,6 +32,8 @@
 PANDAS_VERSION = version.parse(pd.__version__)
 CHECK_FREQ_VERSION = version.Version("1.1")
 IS_PANDAS_ZERO = PANDAS_VERSION < version.Version("1.0")
+ARCTICDB_USING_CONDA = os.environ.get("ARCTICDB_USING_CONDA", "0") != "0"
+print(f"ARCTICDB_USING_CONDA {ARCTICDB_USING_CONDA}")
```
qc00, line 36: Please use feature detection instead of resuing this environment variable because we don't want to manually set the latter all the time when running tests locally.

For example, add a boolean variable like `azure_enabled` to Python in `python_modules.cpp`.

python/tests/conftest.py
```diff
@@ -118,10 +149,57 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
...
+    container = f"testbucket{bucket_id}"
+    bucket_id = bucket_id + 1
+
+    credential_name = "devstoreaccount1"
+    credential_key = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
+    endpoint = f"127.0.0.1:{azurite_port}"
+    is_https = False
+    connect_to_azurite = True
+
+    return endpoint, container, credential_name, credential_key, is_https, connect_to_azurite
```
qc00, line 165: Minor: It will be much easier to read and maintain if you return an object here with these fields instead of a random tuple of values.

You can either return a `dict` or use the `arcticc.pb2.azure_storage_pb2.Config` class which seems to have these fields.

python/tests/conftest.py
qc00: I don't think this is resolved!

You only need to use `yield` in a fixture if you need to do things with the fixture post-test. If you just want to return a fixture value, use `return`.

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -137,12 +140,30 @@ def test_separation_between_libraries_with_prefixes(moto_s3_uri_incl_bucket):
...
-    ]["Name"]
+@pytest.mark.parametrize(
+    "connection_string, client",
+    [
+        ("moto_s3_uri_incl_bucket", "boto_client"),
+        pytest.param(
+            "azurite_azure_uri_incl_bucket",
+            "azure_client",
+            marks=pytest.mark.skipif(sys.platform != "linux", reason="Pending Azure Storge Windows support"),
+        ),
```
qc00, line 151: This `pytest.param(...)` is quite long and used twice, so might want to extract that to a variable.

qc00 (dismissed):

cpp/proto/arcticc/pb2/encoding.proto
```diff
@@ -1,3 +1,10 @@
+/*
```
mehertz, line 1: Thanks for these.

python/tests/stress/arcticdb/version_store/test_stress_generated.py
```diff
@@ -82,3 +91,22 @@ def test_stress_small_row(pct_null, repeats, symbol, lmdb_version_store_tiny_seg
     elapsed = time.time() - start
     print("arctic read time: " + str(elapsed))
     assert_frame_equal(test_df, df)
+
+
+@pytest.mark.parametrize("ncols, nrows", [(5, 10000), (50, 100000), (50, 1000000), (50, 10000000), (500, 1000000)])
```
qc00, line 96: Minor: if you intend to merge this, it's not worth running the smaller test cases unless there are any stats being collected automatically.

phoebusm: I will not merge this part of the code. It will be removed after the benchmark.

cpp/CMakeLists.txt
phoebusm: I think it is better to leave it is as the file should be kept identical as the one in Azure SDK. As later this file may need to be updated, along with the SDK.

cpp/arcticdb/CMakeLists.txt
phoebusm: Hmm it is not added by me. I guess this line is removed in a particular commit, which it should have removed after rebasing.
Removed

cpp/arcticdb/CMakeLists.txt
phoebusm: Added

cpp/arcticdb/storage/storage_factory.cpp
phoebusm: Added

python/tests/conftest.py
phoebusm: I understand. I thought I have resolved it already. It is on me sorry.

cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <azure/core/http/curl_transport.hpp>
+
+namespace arcticdb::storage::azure{
+
+using namespace Azure::Storage;
+using namespace Azure::Storage::Blobs;
+
+AzureStorage::AzureStorage(const LibraryPath &library_path, OpenMode mode, const Config &conf) :
+    Parent(library_path, mode),
+    blob_container_url_(conf.connect_to_azurite() ?
```
willdealtry, line 19: This could probably be a free function

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
willdealtry: I'm not sure, they look fairly different to me. I mean, the patterns are the same, but most of the actual lines are different because they're setting up Azure objects and catching Azure exceptions. 

The intention in the short-to-medium term is to change the storage mechanism anyway, and to move away from the VariantStorage to something more pluggable. This is quite a wide-ranging change anyway, I would be tempted to keep further refactors out of it.

willdealtry (approved):

python/tests/conftest.py
phoebusm: Closed per discussion with Qi. It will be addressed in bulk in a separate PR.

python/tests/integration/arcticdb/test_arctic.py
phoebusm: Closed per discussion with Qi. It will be addressed in bulk in a separate PR.

python/tests/conftest.py
```diff
@@ -118,10 +148,69 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
...
+        container_client.exists()
+    ):  # name of container will be reused; Not clearing the files will fill up the storage while testing against Azure Blob Storage
+        blobs = list(container_client.list_blobs())  # delete container is slow, even with little amount of blobs
+        index = 0
+        max_batch_size = 256  # SDK limit
+        while len(blobs) > 0 and index < len(
+            blobs
+        ):  # delete container is slow and async; No api available to check whether the container delete operation is complete
+            batch_of_blobs = blobs[index : index + max_batch_size]
+            container_client.delete_blobs(*batch_of_blobs)
```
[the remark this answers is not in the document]
phoebusm: This part of the code has been removed.

python/tests/conftest.py
```diff
@@ -118,10 +148,69 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
...
+        container,
+        credential_name,
+        credential_key,
+        is_https,
+        connect_to_azurite,
+    ) = azurite_azure_test_connection_setting
+    container_client = azure_client.get_container_client(container=container)
+    if (
+        container_client.exists()
+    ):  # name of container will be reused; Not clearing the files will fill up the storage while testing against Azure Blob Storage
```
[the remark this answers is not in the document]
phoebusm: Code refactored.

python/tests/conftest.py
```diff
@@ -411,3 +560,80 @@ def get_df(ts, width, max_col_width):
...
+            port += 1
+    raise IOError("no free ports")
+
+
+@pytest.fixture(scope="module")
+def spawn_azurite(azurite_port):
+    if sys.platform == "linux":
+        print("Spawning Azurite")
+        port = str(azurite_port)
+        temp_folder = tempfile.TemporaryDirectory().name
```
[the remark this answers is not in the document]
phoebusm: I have tried it but sadly it is either session-scoped or function-scoped. Instead of having some workaround for tmpdir, I will keep it as it is.

python/tests/integration/arcticdb/version_store/test_metadata_support.py
```diff
@@ -20,36 +21,31 @@
 # test_rt_df stands for roundtrip dataframe (implicitly pandas given file name)
 
 
-@pytest.mark.parametrize("lib_type", ["lmdb_version_store", "s3_version_store"])
-def test_rt_df_with_small_meta(lib_type, request):
-    lib = request.getfixturevalue(lib_type)
```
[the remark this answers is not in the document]
phoebusm: Added

python/tests/conftest.py
```diff
@@ -74,6 +84,26 @@ def _moto_s3_uri_module():
...
+        endpoint,
+        container,
+        credential_name,
+        credential_key,
+        is_https,
+        connect_to_azurite,
+    ) = azurite_azure_test_connection_setting
+    credential = {"account_name": credential_name, "account_key": credential_key}
+    account_url = "https://" if is_https else "http://"
+    account_url += endpoint + "/" + credential_name if connect_to_azurite else credential_name + "." + endpoint
```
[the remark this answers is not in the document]
phoebusm: Fixed

.github/workflows/build_steps.yml
phoebusm: Seems the linux image we used does not come with npm

python/arcticdb/util/test.py
phoebusm: Fixed

cpp/arcticdb/storage/azure/azure_storage.cpp
phoebusm: It's done

python/tests/conftest.py
```diff
@@ -271,9 +408,18 @@ def mongo_version_store(mongo_store_factory):
     return mongo_store_factory()
 
 
-@pytest.fixture(scope="function")
-def s3_version_store_prune_previous(s3_store_factory):
-    return s3_store_factory(prune_previous_version=True)
+@pytest.fixture(
+    scope="function",
+    params=["s3_store_factory", "azure_store_factory"] if AZURE_SUPPORT else ["s3_store_factory"],
```
qc00, line 413: Instead of doing it here, create a new fixture `object_store_factory` with this parameterization. You can then reuse that in most places you're checking `AZURE_SUPPORT`.

mehertz (approved):

.github/workflows/build_steps.yml
```diff
@@ -231,14 +231,15 @@ jobs:
           disk-root: "D:"  # This is also the checkout directory. Total size 12GB.
         continue-on-error: true
 
-      - name: Install azurite
+      - name: Install npm # Linux github runner image does not come with npm
```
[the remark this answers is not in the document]
phoebusm: I will just stick to my original solution (for picking npm by specifying `uses: actions/setup-node@v3.3.0`

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
```diff
@@ -0,0 +1,341 @@
...
+
+template<class KeyBucketizer>
+void do_update_impl(
+    Composite<KeySegmentPair>&& kvs,
+    const std::string& root_folder,
+    Azure::Storage::Blobs::BlobContainerClient& container_client,
+    KeyBucketizer&& bucketizer,
+    const Azure::Storage::Blobs::UploadBlockBlobFromOptions& upload_option,
+    unsigned int request_timeout) {
+        // azure updates the key if it already exists
```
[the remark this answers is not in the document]
phoebusm: Ticket created.
Minor: I believe you are referring to the old, discontinued SDK. But I do remember I saw similar option available in the current SDK as well.

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
phoebusm: Hmmm the Azurite ticket is still opened and there is other person reporting the same issue, even after the fix.

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
mehertz: I think we're agreed that we're not going to refactor this in this PR.

I suspect there is scope to refactor out the common HTTP/Object-storage stuff, but will have to be weighed against how much additional investment we want to make the current storage scheme.

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
```diff
@@ -0,0 +1,341 @@
...
+                        }
+                    }
+                }
+            }
+        );
+        if (batch_counter) {
+            submit_batch(batch);
+        }
+
+        util::check(failed_deletes.empty(), "Have {} segment that have not been removed", failed_deletes.size());
```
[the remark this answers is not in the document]
phoebusm: Hmm thanks for the reminder. Actually the `failed_deletes` vector is no longer needed as deleting non-exist objects is not an error, same as S3.
P.S. Line 223 is submitting batch if the batch size is smaller than the maximum batch size.

python/tests/conftest.py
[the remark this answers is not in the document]
phoebusm: Thanks for the reminder.
Module is selected at the first place as each module could have its own Azurite so that a very basic "load-sharing" can be setup.
If it is a session-scope, the loading can easily make Azurite crash

.github/workflows/build_steps.yml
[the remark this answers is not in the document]
phoebusm: Done

cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -0,0 +1,49 @@
...
+using namespace Azure::Storage::Blobs;
+
+
+AzureStorage::AzureStorage(const LibraryPath &library_path, OpenMode mode, const Config &conf) :
+    Parent(library_path, mode),
+    container_client_(BlobContainerClient::CreateFromConnectionString(conf.endpoint(), conf.container_name(), get_client_options(conf))),
+    root_folder_(object_store_utils::get_root_folder(library_path)),
+    request_timeout_(conf.request_timeout() == 0 ? 60000 : conf.request_timeout()){
+        if (conf.ca_cert_path().empty())
+            log::version().info("Using default CA cert path");
```
mehertz, line 24: `log::storage`, not version.

cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -0,0 +1,49 @@
...
+        else
+            log::version().info("CA cert path: {}", conf.ca_cert_path());
```
mehertz, line 26: `log::storage`, not version.

cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -0,0 +1,49 @@
...
+        log::version().info("Connecting to Azure Blob Storage: {} Container: {}", conf.endpoint(), conf.container_name());
+
```
mehertz, line 28: Too much whitespace.

cpp/arcticdb/storage/azure/azure_storage-inl.hpp
```diff
@@ -0,0 +1,319 @@
...
+        auto blob_name = object_path(bucketizer.bucketize(key_type_dir, key), key);
+        try{
+            auto blob_client = container_client.GetBlockBlobClient(blob_name);
+            auto properties = blob_client.GetProperties().Value;
+
+            if (properties.ETag.HasValue())
+                return true;
+        }
+        catch (const Azure::Core::RequestFailedException& e){
+            ARCTICDB_DEBUG(log::storage(), "Failed to check azure key '{}' {} {}: {}",
```
mehertz: At the bare minimum we should raise a warning here, not log to debug.

python/arcticdb/adapters/azure_library_adapter.py
```diff
@@ -0,0 +1,134 @@
...
+        if query and query.startswith("?"):
+            query = query.strip("?")
+        elif not query:
+            raise ValueError(f"Invalid Azure URI. Missing query parameter")
+
+        parsed_query = re.split("[;]", query)
+        parsed_query = {t.split("=", 1)[0]: t.split("=", 1)[1] for t in parsed_query}
+
+        for key in parsed_query.keys():
+            field_dict = {field.name: field for field in fields(ParsedQuery)}
```
mehertz, line 96: This just resets the value of `field_dict` three times.

python/arcticdb/adapters/azure_library_adapter.py
```diff
@@ -0,0 +1,134 @@
...
+        option_regex = r"(?P<option>[^=]+)=(?P<value>.*)"
+        self._endpoint = ";".join(  # azure c++ sdk doesn't accept any non-standard string
+            [
+                f"{match[0]}={match[1]}"
+                for option_str in uri[len("azure://") :].split(";")
+                for match in re.findall(option_regex, option_str)
+                if match[0] not in additional_options
+            ]
+        )
+        self._bucket = self._query_params.Container
```
mehertz, line 57: ...it's _not_ a bucket though.

mehertz:
Hi @SylvainCorlay,

Your concerns are all valid, understood, and are planned to be addressed later on. As agreed with @phoebusm who is the author of this PR, we will later have to undo the packaging elements of this PR, publish the Azure SDK on conda-forge and use that. There is a placeholder issue for this [here](https://github.com/man-group/ArcticDB/issues/519). It's lacking in detail, but the intent is for that to track this work. 

In other words, this PR is understood to be a short-term solution to unblock users of ArcticDB who would like to use Azure.

For what it's worth, apart from anything else, currently we have split storage capabilities depending on where you've sourced the ArcticDB package from which is obviously confusing for users and something we'd like to resolve.

> Finally, this really brings up the question of the way storage types are handled in ArcticDB. ...
> Moving away from variants and visitors for the storage backend.

We are _very_ keen to do this. In fact we'd like Azure to be the last storage type implemented like this. Given the short timeframe for this work we just thought best to complete this backend first and finalise a more dynamic storage implementation afterwards.

PR #430, opened 2023-05-25
Further simplify the bug template

qc00:
* Reduce the number of questions
* Reduce the amount of text to read


jjerphan (approved):
Thank you for those additions, @qc00.

Issue #146, opened 2023-03-23
Add mypy to build


qc00:
* Would be nice to type-check Python

poodlewars:
And C++ linter while we're at it?

jjerphan:
Hi,

I guess [pre-commit](https://pre-commit.com/) provides all the hooks that ArcticDB might need. Would you be interested in setting up such a setup with this tool?

jjerphan:
To you, is https://github.com/man-group/ArcticDB/pull/445 a complete solution to this issue?

qc00:
No, it's doesn't have mypy?

mehertz:
I'm renaming to just mention mypy

PR #431, opened 2023-05-25
#146 Add Black formatting to build

poodlewars:
#### Reference Issues/PRs

Partially addresses #146


.github/workflows/build_steps.yml
```diff
@@ -55,6 +55,13 @@ jobs:
         with:
           submodules: recursive # Just in case a dep has its own third-party deps
 
+      - name: Black Python formatting
```
qc00, line 58: I resisted adding a step like this to avoid a trivial formatting error failing all concurrent build jobs, which wastes electricity and prevents fast feedback on other issues.

This was one of my biggest gripes with Man Build...

If you want to do this properly, please add a dedicated job to `build.py` for style checks.

poodlewars: Yeah good point, or something closer to this plus https://pre-commit.com/ like Julien suggested

mehertz (approved):

.github/workflows/build_steps.yml
jjerphan: IIUC, will the runs just fail early in case of misformatting in the last commits?

If this is the case, I think we can come up with a pre-commit setup that could call several hooks. This setup could include one hook for black, and potentially other ones such as another hook for clang-format. This setup could be reused on the CI allowing an identical behavior locally and on the CI and preventing maintaining several pieces of tools configuration.

What do you think?

poodlewars: Yeah the `pre-commit` setup sounds good to me. Might be a bit fiddly to get working on our corporate machines, but I'm sure it will be fine.

qc00 (approved):

jjerphan (approved):

PR #432, opened 2023-05-25
Generic parallel get version for batch description

GemDot:
In this PR, we provide per-symbol parallelism for the read_batch_description method.

Considering all this and the way folly Futures work, it is required to propagate a chain of futures all the way long for each independent read description, which ultimately, they converge in a single future containing all the individual description reads. On top of these changes we need to re-use the batch_get_versions function as the start of all the future chain, where all the versions are obtained, per each symbol-version query pair.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -353,13 +353,18 @@ std::pair<VersionedItem, std::optional<google::protobuf::Any>> LocalVersionedEng
 std::vector<std::pair<VersionedItem, std::optional<google::protobuf::Any>>> LocalVersionedEngine::batch_read_descriptor_internal(
         const std::vector<StreamId>& stream_ids,
         const std::vector<VersionQuery>& version_queries) {
-    std::vector<folly::Future<std::pair<VersionedItem, std::optional<google::protobuf::Any>>>> results_fut;
-    for (size_t idx=0; idx < stream_ids.size(); idx++) {
-        auto version_query = version_queries.size() > idx ? version_queries[idx] : VersionQuery{};
-        results_fut.push_back(read_descriptor_version_internal(stream_ids[idx],
-                                                              version_query));
+    auto versions_fut = batch_get_versions(store(), version_map(), stream_ids, version_queries);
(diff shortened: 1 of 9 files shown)


mehertz:
Detail all included changes in teh PR description. Please don't leave it empty!

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -353,13 +353,18 @@ std::pair<VersionedItem, std::optional<google::protobuf::Any>> LocalVersionedEng
...
-        auto version_query = version_queries.size() > idx ? version_queries[idx] : VersionQuery{};
-        results_fut.push_back(read_descriptor_version_internal(stream_ids[idx],
-                                                              version_query));
+    auto versions_fut = batch_get_versions(store(), version_map(), stream_ids, version_queries);
+    std::vector<folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>>> fut_vec;
+    for (auto& version: versions_fut){
+        fut_vec.push_back(get_metadata_future(std::move(version)));
     }
-    return folly::collect(results_fut).get();
+    auto metadata_vector = folly::collect(fut_vec).get();
```
willdealtry, line 361: Why not just enqueue this transformation as another thenValue?

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -315,9 +323,13 @@ struct VersionMapEntry {
...
 };
 
+inline bool is_live_index_type_key(const AtomKeyImpl& key, const std::shared_ptr<VersionMapEntry>& entry) {
+    return is_index_key_type(key.type()) && !entry->is_tombstoned(key);
+}
+
 inline std::optional<VersionId> get_prev_version_in_entry(const std::shared_ptr<VersionMapEntry>& entry, VersionId version_id) {
     //sorted in decreasing order
-    //entry->keys_ is not sorted in version_id any more (due to tombstones), we only need to fetch live index keys
+    //entry->keys_ is not sorted in version_id anymore (due to tombstones), we only need to fetch live index keys
```
mehertz, line 332: `anymore` _technically_ isn't a word. `any more` is correct.

PR #435, opened 2023-05-27
Add branch of `read_batch` that improves parallelism if there is no `QueryBuilder` query

willdealtry:
### What does this PR resolve?

Prior to this PR, `batch_read` parallelises over the slices of symbols, not over individual symbols. Large symbols are by default split (or sliced) per 100k rows or 127 columns and as a result, reading large symbols acheives a degree of parallelism as each slices will be retrieved and decompressed in parallel.

However if you are reading many small symbols, **there will be no effective parallelism**. Each symbol will be wholy retrieved and decompressed before processing the next symbol.

This PR provides a temporary resolution to this problem. As described in the next section, as long as there is no QueryBuilder query, this PR will parallelise across symbols as well as slices.

### How does this PR achieve this?

* Adds `temp_batch_read_internal_direct`. This is a special method that is only called if there is no processing required (read: no `QueryBuilder` query). This paralellises across both slices and symbols.
* Modifies `batch_read_compressed` to return `folly::Future`s enabling the scheduling of many futures to be managed higher up in the stack.
* Adds `gil_lock.hpp`. This adds a scoped means of taking and releasing the GIL. This is used to globally take the GIL when preparing the output frame. 
* Moves `read_options` one level lower to ensure that the lifetime management of objects is correct as we can no longer rely on them being maintained by the enclosing scope

### What is required after this?

#273 completes the work and enables symbol-leel parallelism even if a QueryBuilder query has been added.

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -280,36 +280,27 @@ public:
         return res;
     }
 
-    std::vector<VariantKey> batch_read_compressed(
-        std::vector<entity::VariantKey> &&keys,
-        std::vector<ReadContinuation> &&continuations,
+    folly::Future<std::vector<VariantKey>> batch_read_compressed(
+        std::vector<entity::VariantKey>&& keys,
+        std::vector<ReadContinuation>&& continuations,
         const BatchReadArgs & args) override {
-        util::check(keys.size() == continuations.size(),
-                    "keys and continuations must has then same number of elements");
-        std::vector<folly::Future<VariantKey>> batch;
-        batch.reserve(args.batch_size_);
-        std::vector<VariantKey> res;
-        res.reserve(keys.size());
-        for (std::size_t i = 0; i < keys.size(); ++i) {
-            auto &key = keys[i];
-            auto &cont = continuations[i];
-            batch.push_back(
-                async::submit_io_task(ReadCompressedTask(key, library_, storage::ReadKeyOpts{}))
-                    .via(&async::cpu_executor())
-                    .thenValue(SegmentFunctionTask{std::move(cont)}));
 
-            if(batch.size() == args.batch_size_) {
-                auto vec = folly::collect(batch).get();
-                res.insert(end(res), std::make_move_iterator(std::begin(vec)), std::make_move_iterator(std::end(vec)));
-                batch.clear();
-            }
+        util::check(!keys.empty(), "Unexpected empty keys in batch_read_compressed");
+
+        auto key_seg_futs = folly::window(keys, [*this] (auto&& key) {
+                return async::submit_io_task(ReadCompressedTask(std::move(key), library_, storage::ReadKeyOpts{}));
+            }, args.batch_size_);
+
+        util::check(key_seg_futs.size() == keys.size(), "Size mismatch in batch_read_compressed: {} != {}", key_seg_futs.size(), keys.size());
+        std::vector<folly::Future<VariantKey>> result;
+        result.reserve(key_seg_futs.size());
+        for(auto&& key_seg_fut : folly::enumerate(key_seg_futs)) {
+            result.emplace_back(std::move(*key_seg_fut).thenValue([continuation=std::move(continuations[key_seg_fut.index])] (auto&& key_seg) mutable {
+                return continuation(std::move(key_seg));
+            }));
         }
 
-        if(!batch.empty()) {
-            auto vec = folly::collect(batch).get();
-            res.insert(end(res), std::make_move_iterator(std::begin(vec)), std::make_move_iterator(std::end(vec)));
-        }
-        return res;
+        return folly::collect(result).via(&async::io_executor());
     }
 
     std::vector<Composite<ProcessingSegment>> batch_read_uncompressed(
--- a/cpp/arcticdb/pipeline/read_frame.cpp
+++ b/cpp/arcticdb/pipeline/read_frame.cpp
@@ -214,7 +214,7 @@ void decode_into_frame_static(
     SegmentInMemory &frame,
     PipelineContextRow &context,
     Segment &&s,
-    const std::shared_ptr<BufferHolder>& buffers) {
+    const std::shared_ptr<BufferHolder> buffers) {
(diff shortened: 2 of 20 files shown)


alexowens90 (changes requested):

cpp/arcticdb/async/async_store.hpp
```diff
@@ -280,36 +280,27 @@ class AsyncStore : public Store {
...
-        if(!batch.empty()) {
-            auto vec = folly::collect(batch).get();
-            res.insert(end(res), std::make_move_iterator(std::begin(vec)), std::make_move_iterator(std::end(vec)));
+        auto key_seg_futs = folly::window(keys, [*this] (auto&& key) {
+                return async::submit_io_task(ReadCompressedTask(std::move(key), library_, storage::ReadKeyOpts{}));
+            }, args.batch_size_);
+
+        util::check(key_seg_futs.size() == keys.size(), "Size mismatch in batch_read_compressed: {} != {}", key_seg_futs.size(), keys.size());
+        std::vector<folly::Future<VariantKey>> result;
+        result.reserve(key_seg_futs.size());
```
GemDot, line 296: As previously discussed, I am not requesting any changes. However, it's important to note that there is a subtle difference in performance due to the fact that the folly window method is now only applied to the ReadCompressedTask, rather than both the ReadCompressedTask and SegmentFunctionTask. While we are still throttling, there is no control over the number of threads executing the SegmentFunctionTask.

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -1019,12 +1019,12 @@ folly::Future<std::vector<VariantKey>> fetch_data(
     {
```
GemDot, line 1019: I still think that we should allocate continuations in the heap, instead of the stack, to avoid problems like this one https://github.com/man-group/ArcticDB/actions/runs/5136691407/jobs/9245918336?pr=435.

If an object is created on the stack within a function, then it will be automatically destroyed when the function execution completes, regardless if it has been moved to another thread that is still executing or not. 

Since we are using folly futures, it's possible that some threads running continuations will still be executing when we terminate this function. These continuations may access captured members, including shared pointers that have been destroyed because no one is pointing to the object anymore.

I would personally use a make_shared here for the continuations.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -231,8 +231,8 @@ inline std::vector<folly::Future<std::optional<AtomKey>>> batch_get_versions(
         } else {
             auto fut = shared_futures.find(*symbol);
             if(fut == shared_futures.end()) {
-                auto check_reload_task_fut = async::submit_io_task(CheckReloadTask{store, version_map, *symbol, it->second.load_param_});
-                auto [splitter, inserted] = shared_futures.emplace(*symbol, folly::FutureSplitter(std::move(check_reload_task_fut)));
+                auto [splitter, inserted] = shared_futures.emplace(*symbol, folly::FutureSplitter(async::submit_io_task(CheckReloadTask{store, version_map, *symbol, it->second.load_param_})));
```
GemDot, line 234: when the previous PR was submitted, there was a comment requesting splitting it in 2 statements (as is in the original code). https://github.com/man-group/ArcticDB/pull/415#discussion_r1213448013

GemDot (changes requested):
It is also needed the addition of the locks for the column string reads around reduce_and_fix_columns

alexowens90 (approved):

cpp/arcticdb/async/async_store.hpp
willdealtry: That's fine in my opinion, because we are gating the number of things that can enter the pipeline

cpp/arcticdb/pipeline/read_frame.cpp
willdealtry: The continuations are being moved into bach_read_compressed, there is no need for a shared_ptr that I can see

willdealtry: The only thing that was being captured by reference (and which could therefore go out of scope) is the buffers, which is already a shared_ptr, which is why I'm now passing it by value

cpp/arcticdb/version/version_map_batch_methods.hpp
willdealtry: I didn't like that change

cpp/arcticdb/pipeline/read_frame.cpp
GemDot: yes I see your point, and if the members from continuations didn't contain any shared pointer we would still be fine...However, considering my last paragraph, there is a rare case where the shared pointers inside the captured per copy members may reach a reference count of zero, resulting in the destruction of the object they point to.

cpp/arcticdb/version/version_map_batch_methods.hpp
mehertz: We should set up clang-format and leave that to decide line-length debates IMO.

mehertz: (not in this PR, obviously)

cpp/arcticdb/pipeline/read_frame.cpp
GemDot: yes, correct. As I commented, in this case we are fine

mehertz (changes requested):
Description please

cpp/arcticdb/version/version_map_batch_methods.hpp
mehertz: I didn't realise we were reverting this change in fact.

Is this intentional @willdealtry? Seems a strange change to intentionally revert in my opinion - certainly looks cleaner across two lines to me.

willdealtry: I guess it's a matter of taste. To me it just seems to be introducing a variable with no purpose. We can change it back again in future if people feel strongly about it

mehertz (approved):

PR #438, opened 2023-05-29
Higher utility GitHub Actions changes

qc00:
* More read-able matrix reuse syntax
* Re-organised the Python pinned dependency tests
* Display sccache stats in CIBuildWheel Linux build
* Doc-only changes can be merged now
* Cancel superseded builds
* ~~Exclude LMDB compilation warnings~~ (Didn't work)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,10 +22,36 @@ on:
         type: choice
         options: ["-", debug, release]
 run-name: Building ${{github.ref_name}} on ${{github.event_name}} by ${{github.actor}}
+concurrency:
+  group: ${{github.ref}}
+  cancel-in-progress: true
 jobs:
   common_config:
-    # Hacky way to supply common variables (as the env context is not available in jobs.*.with)
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          # Use a matrix to build the common hierarchical structure used by multiple matrix jobs in build_steps.yaml
+          # Please declare any key you added below in build_steps.yaml's dummy matrix as well to aid the linting tools
+          - linux_matrix:
+              - os: linux
+                distro: ubuntu-latest
+                cmake_preset_prefix: linux
+                cibw_build_suffix: manylinux_x86_64
+                build_dir: /tmp/cpp_build
+                symbols: "*.gz" # Please update publish.yml when changing this!!
+                do_not_archive: ["*.so", "*.[ao]", "vcpkg_installed"]
+                test_services:
+                  mongodb: { image: "mongo:4.4" }
+            windows_matrix:
+              - os: windows
+                distro: windows-latest
+                cmake_preset_prefix: windows-cl
+                cibw_build_suffix: win_amd64
+                build_dir: C:\cpp_build
+                symbols: "*.pdb" # Please update publish.yml when changing this!!
+                do_not_archive: ["*.lib", "*.ilk", "*.pyd", "*.dll", "CMakeFiles/*.dir", "vcpkg_installed"]
+                test_services: {}
     steps:
       - run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -99,26 +111,34 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(needs.common_config.outputs.linux_python_versions)}}
-        python_deps: [""]
+        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
```
poodlewars, line 114: Can we get rid of `LINUX_PYTHON_VERSIONS`, don't see why we need that additional degree of freedom?

.github/workflows/build.yml
```diff
@@ -146,15 +166,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(needs.common_config.outputs.windows_python_versions)}}
+        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
```
poodlewars, line 169: Just remove this degree of freedom?

.github/workflows/build.yml
```diff
@@ -99,26 +111,34 @@ jobs:
...
-            python_deps: "requirements-compatibility-py36.txt"
+            python_deps_ids: ["", -compat36]
+            matrix_override:
+              - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
+              - python_deps_id: -compat36
+                python_deps: requirements-compatibility-py36.txt
           - python3: 8
-            python_deps: "requirements-compatibility-py38.txt"
-    name: 3.${{matrix.python3}} Linux ${{matrix.python_deps}}
+            python_deps_ids: ["", -compat38]
```
poodlewars, line 125: Why the "" first element? I would have thought the first element in the `include` would take care of this, and no need for python_deps_ids to be a string-JSON-list?

poodlewars (approved):

.github/workflows/build.yml
qc00: I use it on my fork to reduce the number of jobs.

.github/workflows/build.yml
qc00: The way `matrix.include` is expanded is like this: 

For each map object in the `include` list:
1. if any map entry matches the existing case(s), and no entry conflict with such case(s), merge the entire map into all of those cases
2. if there were conflict(s), the map becomes its own separate case (no inheritance from any existing case)
3. if there is no overlap at all, merge the map into all existing cases
4. except if there's no existing case, then the map becomes the first case.

What I need to do in `build_steps.yml::python_tests` matrix is to: declare two `python_deps_ids` case and apply the `inputs.matrix` to both of them.

I have tried the following:
* In the `linux_matrix` of `common_config` above, add two `- python_deps_ids: ...` items before the current `- os: linux` entry. This doesn't work because there's no easy way in GitHub Actions (GHA) to address only the last entry for the `build_steps.yml::compile` matrix
* Split `common_config` into named parts like "common_matrix", "text_matrix_extra". This is difficult because GHA doesn't have a native way to join an item ("common_matrix") with a list ("text_matrix_extra"). The hack to do this will be very difficult to understand

This led me to this solution:
1. Declare the `python_deps_ids` so they create some existing cases in the matrix
2. Make sure the `- os: linux` is the first item so it can be easily selected for the `compile` job. For the `python_tests` job, as it has no keys that overlap with existing matrix cases, it will be merged into _all_ cases
3. Add extra `- python_deps_ids: -compat...` item in the `matrix_override` which will be merged into one of the existing cases

PR #463, opened 2023-06-06
Add per symbol parallelism to batch read metadata

GemDot:
In this PR, we provide per-symbol parallelism for the read_batch_metadata method.

Considering all this and the way folly Futures work, it is required to propagate a chain of futures all the way long for each independent read metadata, which ultimately, they converge in a single future containing all the individual meta reads. On top of these changes we need to re-use the batch_get_versions function as the start of all the future chain, where all the versions are obtained, per each symbol-version query pair.


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1160,31 +1160,31 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
...
-            }
-        }else{
-            auto last_version_it = latest_versions_index_map->find(*stream_id);
-            if(last_version_it != latest_versions_index_map->end()) {
-                fut_vec.push_back(store()->read_metadata(last_version_it->second));
-            }else{
-                fut_vec.push_back(std::make_pair(std::nullopt, std::nullopt));   
-            }
-        }
+    for (auto& version: versions_fut){
```
alexowens90, line 1186: This should be `auto&&`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1160,31 +1160,31 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
     return std::make_pair(specific_versions, latest_versions);
 }
 
+folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::get_metadata_future_impl(
```
alexowens90, line 1163: I believe we agreed to remove `future` from the method names. I think a reasonable convention to set would be to have `get_metadata_async` call `get_metadata`

alexowens90 (changes requested):

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: `read_metadata` currently has it's own (almost identical) implementation in `version_store_api.cpp`. Please replace that with a call to this function.

alexowens90 (changes requested):

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: agree 100%

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: correct

alexowens90 (approved):

PR #467, opened 2023-06-07
Implement per symbol parallelisation for write batch method

GemDot:
- This pull request introduces per-symbol parallelism for the write_batch method, which requires significant implementation changes using the Folly Futures API. 
- The way Folly Futures work requires creating a chain of futures for each independent read, which ultimately converges into a single future containing all the individual writes. This can be visualized as a tree with each individual read representing a branch that has its own independent per-slice read operations, which are also futures. 
- To ensure proper functioning for all write flows, it's important to avoid intermediate .get() calls from the futures of all these flows. Instead, we must call a SINGLE .get() at the end of the entire futures chain, when they converge; otherwise, the system will deadlock. 
- This changes also requires the implementation of the async version of the method `async_batch_get_latest_undeleted_version_and_next_version_id`.
- In our local environment, we observed an approximate 12-fold improvement by writing 2000 symbols occupying a single segment using the changes in this PR compared to the current master. The results with these changes showed 12.37 +- 1.27S, while the results in master were approximately 145s. These experiments were conducted on an 8-core CPU with the number of IO threads incremented to 50 (VersionStore.NumIOThreads = 50).

### DIFF
--- a/cpp/arcticdb/pipeline/slicing.hpp
+++ b/cpp/arcticdb/pipeline/slicing.hpp
@@ -67,15 +67,6 @@ SlicingPolicy get_slicing_policy(
 
 std::vector<FrameSlice> slice(InputTensorFrame &frame, const SlicingPolicy& slicer);
 
-std::vector<folly::Future<SliceAndKey>> write_slices(
-    const InputTensorFrame &frame,
-    const std::vector<FrameSlice> &slices,
-    const SlicingPolicy& slicing,
-    folly::Function<stream::StreamSink::PartialKey(const FrameSlice &)>&& partial_key_gen,
-    const std::shared_ptr<stream::StreamSink>& sink,
-    const std::shared_ptr<DeDupMap>& de_dup_map);
-
-
 inline auto slice_begin_pos(const FrameSlice& slice, const InputTensorFrame& frame) {
     return slice.row_range.first - frame.offset;
 }
--- a/cpp/arcticdb/pipeline/write_frame.cpp
+++ b/cpp/arcticdb/pipeline/write_frame.cpp
@@ -24,6 +24,7 @@
 #include <arcticdb/stream/append_map.hpp>
 #include <arcticdb/version/version_utils.hpp>
 #include <arcticdb/entity/merge_descriptors.hpp>
+#include <arcticdb/async/task_scheduler.hpp>
 
 #include <pybind11/pybind11.h>
 
@@ -32,9 +33,9 @@ namespace arcticdb::pipelines {
 using namespace arcticdb::entity;
 using namespace arcticdb::stream;
 
-std::vector<folly::Future<SliceAndKey>> write_slices(
+folly::Future<std::vector<SliceAndKey>> write_slices(
         const InputTensorFrame &frame,
-        const std::vector<FrameSlice> &slices,
+        std::vector<FrameSlice>&& slices,
         const SlicingPolicy &slicing,
         folly::Function<stream::StreamSink::PartialKey(const FrameSlice &)>&& partial_key_gen,
         const std::shared_ptr<stream::StreamSink>& sink,
@@ -107,14 +108,14 @@ std::vector<folly::Future<SliceAndKey>> write_slices(
     auto fut_writes = sink->batch_write(std::move(key_segs), de_dup_map);
 
     ARCTICDB_SUBSAMPLE_DEFAULT(WriteSlicesWait)
-    return std::move(fut_writes).thenValue([&](auto &&keys) {
-        std::vector<folly::Future<SliceAndKey>> res;
+    return std::move(fut_writes).thenValue([slices = std::move(slices)](auto &&keys) mutable {
+        std::vector<SliceAndKey> res;
         res.reserve(keys.size());
         for (std::size_t i = 0; i < res.capacity(); ++i) {
             res.emplace_back(SliceAndKey{slices[i], std::move(to_atom(keys[i]))});
         }
         return res;
-    }).get();
+    });
 }
 
 folly::Future<entity::VariantKey> write_multi_index(
@@ -131,7 +132,7 @@ folly::Future<entity::VariantKey> write_multi_index(
     return writer.commit();
 }
 
-std::vector<folly::Future<SliceAndKey>> slice_and_write(
+folly::Future<std::vector<SliceAndKey>> slice_and_write(
         InputTensorFrame &frame,
         const SlicingPolicy &slicing,
         folly::Function<stream::StreamSink::PartialKey(const FrameSlice &)>&& partial_key_gen,
@@ -140,15 +141,13 @@ std::vector<folly::Future<SliceAndKey>> slice_and_write(
         bool sparsify_floats) {
     ARCTICDB_SUBSAMPLE_DEFAULT(SliceFrame)
     auto slices = slice(frame, slicing);
-
     ARCTICDB_SUBSAMPLE_DEFAULT(SliceAndWrite)
-    auto fut_slice_keys = write_slices(frame, slices, slicing, std::move(partial_key_gen), sink, de_dup_map, sparsify_floats);
-    return fut_slice_keys;
+    return write_slices(frame, std::move(slices), slicing, std::move(partial_key_gen), sink, de_dup_map, sparsify_floats);
 }
 
 folly::Future<entity::AtomKey>
 write_frame(
-        const IndexPartialKey& key,
+        IndexPartialKey&& key,
         InputTensorFrame&& frame,
         const SlicingPolicy &slicing,
         const std::shared_ptr<Store>& store,
@@ -158,7 +157,9 @@ write_frame(
     auto fut_slice_keys = slice_and_write(frame, slicing, get_partial_key_gen(frame, key), store, de_dup_map, sparsify_floats);
     // Write the keys of the slices into an index segment
     ARCTICDB_SUBSAMPLE_DEFAULT(WriteIndex)
-    return index::write_index(std::move(frame), std::move(fut_slice_keys), key, store);
+    return std::move(fut_slice_keys).thenValue([frame = std::move(frame), key = std::move(key), &store](auto&& slice_keys) mutable {
+        return index::write_index(std::move(frame), std::move(slice_keys), key, store);
+    });
 }
 
 folly::Future<entity::AtomKey> append_frame(
@@ -193,8 +194,7 @@ folly::Future<entity::AtomKey> append_frame(
     );
 
     auto existing_slices = unfiltered_index(index_segment_reader);
-    auto fut_slice_keys = slice_and_write(frame, slicing, get_partial_key_gen(frame, key), store);
-    auto keys_fut = folly::collect(fut_slice_keys);
+    auto keys_fut = slice_and_write(frame, slicing, get_partial_key_gen(frame, key), store);
 
     auto slice_and_keys_to_append = keys_fut.wait().value();
 
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -418,7 +418,7 @@ DescriptorItem LocalVersionedEngine::read_descriptor_internal(
 std::vector<DescriptorItem> LocalVersionedEngine::batch_read_descriptor_internal(
     const std::vector<StreamId>& stream_ids,
     const std::vector<VersionQuery>& version_queries) {
-    auto versions_fut = batch_get_versions(store(), version_map(), stream_ids, version_queries);
+    auto versions_fut = batch_get_versions_async(store(), version_map(), stream_ids, version_queries);
     std::vector<folly::Future<DescriptorItem>> fut_vec;
     for(const auto& stream_id : folly::enumerate(stream_ids)) {
         fut_vec.push_back(
@@ -1014,7 +1014,7 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
     std::vector<ReadQuery> &read_queries,
     const ReadOptions &read_options) {
     py::gil_scoped_release release_gil;
-    auto versions = batch_get_versions(store(), version_map(), stream_ids, version_queries);
+    auto versions = batch_get_versions_async(store(), version_map(), stream_ids, version_queries);
     std::vector<folly::Future<std::pair<VersionedItem, FrameAndDescriptor>>> output;
     for (auto &&version : folly::enumerate(versions)) {
         auto read_query = read_queries.empty() ? ReadQuery{} : read_queries[version.index];
@@ -1058,10 +1058,33 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
     return folly::collect(results_fut).get();
 }
 
-std::vector<AtomKey> LocalVersionedEngine::batch_write_internal(
+folly::Future<VersionedItem> LocalVersionedEngine::write_index_key_to_version_map_async(
+    const std::shared_ptr<VersionMap> &version_map,
+    const AtomKey&& index_key,
+    const UpdateInfo& stream_update_info,
+    bool prune_previous_versions) {
+
+    folly::Future<folly::Unit> write_version_fut;
+
+    if(prune_previous_versions) {
+        write_version_fut = async::submit_io_task(WriteAndPrunePreviousTask{store(), version_map, index_key, stream_update_info.previous_index_key_});
+    } else {
+        write_version_fut = async::submit_io_task(WriteVersionTask{store(), version_map, index_key});
+    }
+
+    return std::move(write_version_fut)
+    .then([this, index_key = std::move(index_key)](auto &&) mutable {
+        return async::submit_io_task(WriteSymbolTask(store(), symbol_list_ptr(), index_key.id()))
+        .then([index_key = std::move(index_key)](auto &&) mutable {
+            return VersionedItem(std::move(index_key));
+        });
+    });
+}
+
+std::vector<folly::Future<AtomKey>> LocalVersionedEngine::batch_write_internal(
     std::vector<VersionId> version_ids,
     const std::vector<StreamId>& stream_ids,
-    std::vector<InputTensorFrame> frames,
+    std::vector<InputTensorFrame>&& frames,
     std::vector<std::shared_ptr<DeDupMap>> de_dup_maps,
     bool validate_index
 ) {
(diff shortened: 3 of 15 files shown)


python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -626,6 +642,34 @@ def test_write_batch_pickle_duplicate_symbols(arctic_library):
     assert not lib.list_symbols()
 
 
+@pytest.mark.parametrize("num_columns", [8, 16])
+@pytest.mark.parametrize("num_days", [10, 200])
+@pytest.mark.parametrize("num_symbols", [100, 200])
+def test_write_batch_stress(arctic_library, num_columns, num_days, num_symbols):
```
mehertz, line 648: Shouldn't this go in the stress test module?

GemDot: we will change the name of this tests

GemDot: I would like it to be in this file, since it is the external API

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -653,6 +669,34 @@ def test_write_batch_pickle_duplicate_symbols(arctic_library):
...
+        df = generate_dataframe(random.sample(columns, 6), dt, num_days, num_rows_per_day)
+        list_requests.append(WritePayload("symbol_" + str(sym), df))
+        list_dataframes[sym] = df
+
+    batch = lib.write_batch(list_requests)
+    assert all(type(w) == PythonVersionedItem for w in batch)
+
+    for sym in range(num_symbols):
+        original_dataframe = list_dataframes[sym]
+        for col in original_dataframe.columns:
```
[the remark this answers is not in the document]
GemDot: oh yeah, no idea why I did like this, probs copy paste

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -116,6 +116,25 @@ std::vector<VersionedItem> PythonVersionStore::batch_write_index_keys_to_version
     return output;
 }
 
+std::vector<VersionId> PythonVersionStore::create_version_id_and_dedup_map(
+    std::vector<version_store::UpdateInfo>& vector_update_info,
```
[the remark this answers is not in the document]
GemDot: it can ;)

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -116,6 +116,25 @@ std::vector<VersionedItem> PythonVersionStore::batch_write_index_keys_to_version
...
 }
 
+std::vector<VersionId> PythonVersionStore::create_version_id_and_dedup_map(
+    std::vector<version_store::UpdateInfo>& vector_update_info, 
+    const std::vector<StreamId>& stream_ids, 
+    std::vector<std::shared_ptr<DeDupMap>>& de_dup_maps,
+    WriteOptions& write_options){
+    std::vector<VersionId> version_ids;
+    for (size_t idx = 0; idx < stream_ids.size(); idx++) {
+        const auto& update_info = vector_update_info[idx];
```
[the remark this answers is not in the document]
GemDot: not very useful on my opinion, since vector_update_info is a logic fully controlled by us. But happy to add it.

cpp/arcticdb/version/version_store_api.cpp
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -40,28 +40,25 @@ inline std::shared_ptr<std::unordered_map<StreamId, AtomKey>> batch_get_latest_v
...
-    });
-
-    return output;
+                                                     LoadParameter{LoadType::LOAD_LATEST_UNDELETED}})
+        .thenValue([](auto&& entry){
+            auto latest_version = entry->get_first_index(true);
+            auto latest_undeleted_version = entry->get_first_index(false);
+            VersionId next_version_id = latest_version.has_value() ? latest_version->version_id() + 1 : 0;
+            return version_store::UpdateInfo{latest_undeleted_version, next_version_id};
+        }).via(&async::cpu_executor()));
```
alexowens90, line 59: I don't think we need to switch executor for such a cheap task.

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -334,6 +334,25 @@ class LocalVersionedEngine : public VersionedEngine {
         const WriteOptions& write_options
     );
 
+    folly::Future<VersionedItem> async_write_index_key_to_version_map(
```
alexowens90, line 337: `async` as suffix again

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1075,9 +1098,61 @@ std::vector<AtomKey> LocalVersionedEngine::batch_write_internal(
...
+    const std::vector<StreamId>& stream_ids,
+    std::vector<InputTensorFrame>&& frames,
+    bool prune_previous_versions,
+    bool validate_index
+) {
+    auto write_options = get_write_options();
+    auto update_info_fut_vector = async_batch_get_latest_undeleted_version_and_next_version_id(store(),
+                                                                                               version_map(),
+                                                                                               stream_ids);
+    util::check(stream_ids.size() == update_info_fut_vector.size(), "stream_ids and update_info_fut_vectormust be of the same size");
```
alexowens90, line 1125: Please use the new error code framework rather than `util::check`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1075,9 +1098,61 @@ std::vector<AtomKey> LocalVersionedEngine::batch_write_internal(
...
 }
 
+std::vector<VersionedItem> LocalVersionedEngine::batch_write_versioned_dataframe_internal(
+    const std::vector<StreamId>& stream_ids,
+    std::vector<InputTensorFrame>&& frames,
+    bool prune_previous_versions,
+    bool validate_index
+) {
+    auto write_options = get_write_options();
+    auto update_info_fut_vector = async_batch_get_latest_undeleted_version_and_next_version_id(store(),
```
alexowens90, line 1122: Prefer not specifying the container type in variable names, and making it plural instead. `_fut` is valuable to distinguish from non-future variables, but the fact it is a vector isn't important. I would use `update_info_futs`.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1075,9 +1098,61 @@ std::vector<AtomKey> LocalVersionedEngine::batch_write_internal(
...
+                                                                                               version_map(),
+                                                                                               stream_ids);
+    util::check(stream_ids.size() == update_info_fut_vector.size(), "stream_ids and update_info_fut_vectormust be of the same size");                                                                                    
+    std::vector<version_store::UpdateInfo> vector_update_info(stream_ids.size());
```
alexowens90, line 1126: This variable isn't needed, just pass the relevant information from one task to another by their return values.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1075,9 +1098,61 @@ std::vector<AtomKey> LocalVersionedEngine::batch_write_internal(
             validate_index
         ));
     }
-    return folly::collect(results_fut).get();
+    return results_fut;
+}
+
+std::pair<VersionId, std::shared_ptr<DeDupMap>> LocalVersionedEngine::create_version_id_and_dedup_map(
```
alexowens90, line 1104: Prefer a struct to `std::pair`, it makes intent much clearer. It will also help with passing generated values between tasks, eliminating the need for `vector_update_info`.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1054,10 +1054,33 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
...
+    folly::Future<folly::Unit> write_version_fut;
+
+    if(prune_previous_versions) {
+        write_version_fut = async::submit_io_task(WriteAndPrunePreviousTask{store(), version_map, index_key, stream_update_info.previous_index_key_});
+    } else {
+        write_version_fut = async::submit_io_task(WriteVersionTask{store(), version_map, index_key});
+    }
+
+    return std::move(write_version_fut)
+    .then([this, index_key, store = store()](auto &&){
```
alexowens90, line 1072: `index_key` can be captured by reference, and there's no need to capture `store` explicitly when you have `this`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1054,10 +1054,33 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
...
+        return async::submit_io_task(WriteSymbolTask(store, symbol_list_ptr(), index_key.id()))
+        .then([index_key](auto &&){
```
alexowens90, line 1074: Can capture by rvalue reference and move into the `VersionedItem` ctor

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -130,19 +131,18 @@ folly::Future<entity::VariantKey> write_multi_index(
...
+folly::Future<std::vector<SliceAndKey>> slice_and_write(
         InputTensorFrame &frame,
         const SlicingPolicy &slicing,
         folly::Function<stream::StreamSink::PartialKey(const FrameSlice &)>&& partial_key_gen,
         const std::shared_ptr<stream::StreamSink>& sink,
         const std::shared_ptr<DeDupMap>& de_dup_map,
         bool sparsify_floats) {
     ARCTICDB_SUBSAMPLE_DEFAULT(SliceFrame)
-    auto slices = slice(frame, slicing);
+    auto slices = std::make_shared<std::vector<FrameSlice>>(slice(frame, slicing));
```
alexowens90, line 142: Why is this now a `shared_ptr`

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -31,17 +32,17 @@ namespace arcticdb::pipelines {
 using namespace arcticdb::entity;
 using namespace arcticdb::stream;
 
-std::vector<folly::Future<SliceAndKey>> write_slices(
+folly::Future<std::vector<SliceAndKey>> write_slices(
         const InputTensorFrame &frame,
-        const std::vector<FrameSlice> &slices,
+        const std::shared_ptr<std::vector<FrameSlice>> slices,
```
alexowens90, line 37: `const std::shared_ptr` doesn't really make sense, generally pass by copy, or const ref if you really don't want to touch the reference counter

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -157,7 +157,9 @@ write_frame(
     auto fut_slice_keys = slice_and_write(frame, slicing, get_partial_key_gen(frame, key), store, de_dup_map, sparsify_floats);
     // Write the keys of the slices into an index segment
     ARCTICDB_SUBSAMPLE_DEFAULT(WriteIndex)
-    return index::write_index(std::move(frame), std::move(fut_slice_keys), key, store);
+    return std::move(fut_slice_keys).thenValue([&frame, key, store](auto&& slice_keys){
```
alexowens90, line 160: Capture `frame` explicitly by rvalue since it's getting moved, and others by reference

alexowens90 (changes requested):

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: I wanted to indicate that from that point I want to execute everything in the CPU executor, but fair to remove it here

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: no, I would rather remove this assertion. it is auto testing logic, no user should ever encounter it

cpp/arcticdb/pipeline/write_frame.cpp
GemDot: because it is created in this function and used later by the actual threads, after the batch_write operation is done, as part of the last ThenValue clause inside write_slices. If we don't convert it as a shared pointer, at the time it will be used, the object will be already destroyed.

cpp/arcticdb/pipeline/write_frame.cpp
GemDot: agree

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: I wanted to avoid to pass it across 2 levels and have this not-nice tuple return...

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: yes, it will help, but only for the first passing level of tasks (regarding your comment about vector_update_info)

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -717,6 +734,32 @@ def test_write_batch_pickle_duplicate_symbols(arctic_library):
     assert not lib.list_symbols()
 
 
+@pytest.mark.parametrize("num_columns", [8, 16])
+@pytest.mark.parametrize("num_days", [10, 200])
+@pytest.mark.parametrize("num_symbols", [100, 200])
+def test_write_batch_different_params(arctic_library, num_columns, num_days, num_symbols):
```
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -40,28 +40,25 @@ inline std::shared_ptr<std::unordered_map<StreamId, AtomKey>> batch_get_latest_v
 }
 
 // The logic here is the same as get_latest_undeleted_version_and_next_version_id
-inline std::unordered_map<StreamId, version_store::UpdateInfo> batch_get_latest_undeleted_version_and_next_version_id(
+inline std::vector<folly::Future<version_store::UpdateInfo>> async_batch_get_latest_undeleted_version_and_next_version_id(
```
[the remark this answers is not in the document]
GemDot: done

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: agree, changed

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -717,6 +717,33 @@ def test_write_batch_pickle_duplicate_symbols(arctic_library):
     assert not lib.list_symbols()
 
 
+def test_write_batch(arctic_library):
+    """Should be able to write different size of batch of data."""
+    lib = arctic_library
+    lib._nvs._lib_cfg.lib_desc.version.write_options.segment_row_size = 10
```
alexowens90, line 723: This is an option you can control at library creation time. Please add a fixture for creating libraries that accepts a `LibraryOptions` parameter.

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -717,6 +717,33 @@ def test_write_batch_pickle_duplicate_symbols(arctic_library):
...
+    num_symbols = 200
+    dt = datetime(2019, 4, 8, 0, 0, 0)
+    column_length = 6
+    num_rows_per_day = 1
+    list_requests = []
+    list_dataframes = {}
+    for sym in range(num_symbols):
+        columns = random_strings_of_length(num_columns, column_length, True)
+        df = generate_dataframe(random.sample(columns, 6), dt, num_days, num_rows_per_day)
+        list_requests.append(WritePayload("symbol_" + str(sym), df))
```
alexowens90, line 735: As I said before, some of these should have metadata so that the test covers at least as much as the old one did (with a corresponding assert after the read)

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Checking supposed invariants that users should never encounter is something we frequently do in the codebase. Otherwise, when there is a bug that would have failed the check, you end up with much harder to debug errors like segfaults.

alexowens90 (changes requested):

cpp/arcticdb/pipeline/write_frame.cpp
GemDot: I did the same for key, despite was not in one of your comments

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: done

cpp/arcticdb/pipeline/write_frame.cpp
GemDot: done

cpp/arcticdb/pipeline/write_frame.cpp
GemDot: done

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1054,10 +1054,33 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
...
+    if(prune_previous_versions) {
+        write_version_fut = async::submit_io_task(WriteAndPrunePreviousTask{store(), version_map, index_key, stream_update_info.previous_index_key_});
+    } else {
+        write_version_fut = async::submit_io_task(WriteVersionTask{store(), version_map, index_key});
+    }
+
+    return std::move(write_version_fut)
+    .then([this, &index_key](auto &&){
+        return async::submit_io_task(WriteSymbolTask(store(), symbol_list_ptr(), index_key.id()))
+        .then([&index_key](auto &&){
```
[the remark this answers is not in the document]
GemDot: done

python/tests/integration/arcticdb/test_arctic.py

python/tests/integration/arcticdb/test_arctic.py
GemDot: done as discussed

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -67,9 +67,9 @@ SlicingPolicy get_slicing_policy(
 
 std::vector<FrameSlice> slice(InputTensorFrame &frame, const SlicingPolicy& slicer);
 
-std::vector<folly::Future<SliceAndKey>> write_slices(
+folly::Future<std::vector<SliceAndKey>> write_slices(
     const InputTensorFrame &frame,
-    const std::vector<FrameSlice> &slices,
+    std::vector<FrameSlice> slices,
```
alexowens90, line 72: I'm confused by this. Is this meant to be the function prototype of `write_slices` from `write_frame.cpp`? Because the signatures don't match so it can't be.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -140,15 +141,13 @@ std::vector<folly::Future<SliceAndKey>> slice_and_write(
...
     ARCTICDB_SUBSAMPLE_DEFAULT(SliceAndWrite)
-    auto fut_slice_keys = write_slices(frame, slices, slicing, std::move(partial_key_gen), sink, de_dup_map, sparsify_floats);
-    return fut_slice_keys;
+    return write_slices(frame, std::move(slices), slicing, std::move(partial_key_gen), sink, de_dup_map, sparsify_floats);
 }
 
 folly::Future<entity::AtomKey>
 write_frame(
-        const IndexPartialKey& key,
+        IndexPartialKey&& key,
```
alexowens90, line 150: This should remain as `const&`

python/tests/conftest.py
```diff
@@ -231,6 +232,35 @@ def create_version_store(
...
+        library_options: Optional[LibraryOptions] = None,
+    ) -> Library:
+        return _arctic_library_factory_impl(used, lib_name, arctic_client, library_options)
+
+    return create_library
+
+
+@pytest.fixture(scope="function")
+def library_factory_small_segment_size(library_factory):
+    return library_factory(LibraryOptions(rows_per_segment=10))
```
alexowens90, line 261: The idea is to not have fixtures like this. So `test_write_batch` should use the `library_factory` fixture directly, and pass in appropriate `LibraryOptions` to create a library with the desired configuration.

alexowens90 (changes requested):

cpp/arcticdb/pipeline/write_frame.cpp
GemDot: as discussed, it is fine as is

python/tests/conftest.py
GemDot: agree

cpp/arcticdb/pipeline/slicing.hpp
GemDot: Modifying this file was just a silly mistake I did, which was not detected by the compiler because no function is implemented/used for this header. So, I removed it completely. The one I am using relates to the implementation in write_slices. Also, changed the header of write_slices to be consistent

alexowens90 (approved):

PR #476, opened 2023-06-09
Parallel version write batch  metadata method

GemDot:
- This pull request introduces per-symbol parallelism for the ```write_batch_metadata``` method, which requires significant implementation changes using the Folly Futures API.
- The way Folly Futures work requires creating a chain of futures for each independent read, which ultimately converges into a single future containing all the individual writes. This can be visualized as a tree with each individual read representing a branch that has its own independent per-slice read operations, which are also futures.
- To ensure proper functioning for all write metadata flows, it's important to avoid intermediate .get() calls from the futures of all these flows. Instead, we must call a SINGLE .get() at the end of the entire futures chain, when they converge; otherwise, the system will deadlock.
- This changes also reuses the async method async_batch_get_latest_undeleted_version_and_next_version_id, which is also used by write_batch.
- In our local environment, we observed an approximate 37-fold improvement by writing metadata for 2000 symbols using the changes in this PR compared to writing the same metadata by the non-batch version of this method (write_metadata). The results with these changes showed 5.59S by using the write batch metadata method, while the results by using the non-batch version of the method were approximately 145s. These experiments were conducted on an 8-core CPU with the number of IO threads incremented to 50 (VersionStore.NumIOThreads = 50).

### DIFF
--- a/cpp/arcticdb/entity/data_error.hpp
+++ b/cpp/arcticdb/entity/data_error.hpp
@@ -25,30 +25,32 @@ enum class VersionRequestType: uint32_t {
 class DataError {
 public:
     DataError(const StreamId& symbol,
-              const pipelines::VersionQueryType& version_query_type,
               std::string&& exception_string,
+              const std::optional<pipelines::VersionQueryType>& version_query_type=std::nullopt,
               std::optional<ErrorCode> error_code=std::nullopt) :
         symbol_(symbol),
         exception_string_(exception_string),
         error_code_(error_code){
-        util::variant_match(
-                version_query_type,
-                [this] (const pipelines::SnapshotVersionQuery& query) {
-                    version_request_type_ = VersionRequestType::SNAPSHOT;
-                    version_request_data_ = query.name_;
-                },
-                [this] (const pipelines::TimestampVersionQuery& query) {
-                    version_request_type_ = VersionRequestType::TIMESTAMP;
-                    version_request_data_ = query.timestamp_;
-                },
-                [this] (const pipelines::SpecificVersionQuery& query) {
-                    version_request_type_ = VersionRequestType::SPECIFIC;
-                    version_request_data_ = query.version_id_;
-                },
-                [this] (const std::monostate&) {
-                    version_request_type_ = VersionRequestType::LATEST;
-                }
-        );
+        if (version_query_type.has_value()){
+            util::variant_match(
+                    *version_query_type,
+                    [this] (const pipelines::SnapshotVersionQuery& query) {
+                        version_request_type_ = VersionRequestType::SNAPSHOT;
+                        version_request_data_ = query.name_;
+                    },
+                    [this] (const pipelines::TimestampVersionQuery& query) {
+                        version_request_type_ = VersionRequestType::TIMESTAMP;
+                        version_request_data_ = query.timestamp_;
+                    },
+                    [this] (const pipelines::SpecificVersionQuery& query) {
+                        version_request_type_ = VersionRequestType::SPECIFIC;
+                        version_request_data_ = query.version_id_;
+                    },
+                    [this] (const std::monostate&) {
+                        version_request_type_ = VersionRequestType::LATEST;
+                    }
+            );
+        }
     }
 
     DataError() = delete;
@@ -63,11 +65,11 @@ public:
         return fmt::format("{}", symbol_);
     }
 
-    VersionRequestType version_request_type() const {
+    std::optional<VersionRequestType>  version_request_type() const {
         return version_request_type_;
     }
 
-    std::variant<std::monostate, int64_t, std::string> version_request_data() const {
+    std::optional<std::variant<std::monostate, int64_t, std::string>> version_request_data() const {
         return version_request_data_;
     }
 
@@ -84,24 +86,26 @@ public:
     }
 
     std::string to_string() const {
-        std::string version_request_explanation;
-        switch (version_request_type_) {
-            case VersionRequestType::SNAPSHOT:
-                version_request_explanation = fmt::format("in snapshot '{}'", std::get<std::string>(version_request_data_));
-                break;
-            case VersionRequestType::TIMESTAMP:
-                version_request_explanation = fmt::format("at time '{}'", std::get<int64_t>(version_request_data_));
-                break;
-            case VersionRequestType::SPECIFIC:
-                version_request_explanation = fmt::format("with specified version '{}'", std::get<int64_t>(version_request_data_));
-                break;
-            case VersionRequestType::LATEST:
-                version_request_explanation = fmt::format("with specified version 'latest'");
-                break;
-            default:
-                internal::raise<ErrorCode::E_ASSERTION_FAILURE>(
-                        "Unexpected enum value in DataError::to_string: {}",
-                        static_cast<uint32_t>(version_request_type_));
+        std::string version_request_explanation = "UNKNOWN";
+        if (version_request_type_.has_value()) {
+            switch (*version_request_type_) {
+                case VersionRequestType::SNAPSHOT:
+                    version_request_explanation = fmt::format("in snapshot '{}'", std::get<std::string>(*version_request_data_));
+                    break;
+                case VersionRequestType::TIMESTAMP:
+                    version_request_explanation = fmt::format("at time '{}'", std::get<int64_t>(*version_request_data_));
+                    break;
+                case VersionRequestType::SPECIFIC:
+                    version_request_explanation = fmt::format("with specified version '{}'", std::get<int64_t>(*version_request_data_));
+                    break;
+                case VersionRequestType::LATEST:
+                    version_request_explanation = fmt::format("with specified version 'latest'");
+                    break;
+                default:
+                    internal::raise<ErrorCode::E_ASSERTION_FAILURE>(
+                            "Unexpected enum value in DataError::to_string: {}",
+                            static_cast<uint32_t>(*version_request_type_));
+            }
         }
         return fmt::format(
                 "Version {} of symbol '{}' unable to be retrieved: Error Category: '{}' Exception String: '{}'",
@@ -112,9 +116,9 @@ public:
     }
 private:
     StreamId symbol_;
-    VersionRequestType version_request_type_{VersionRequestType::LATEST};
+    std::optional<VersionRequestType> version_request_type_;
     // int64_t for timestamp and SignedVersionId
-    std::variant<std::monostate, int64_t, std::string> version_request_data_;
+    std::optional<std::variant<std::monostate, int64_t, std::string>> version_request_data_;
     std::string exception_string_;
     std::optional<ErrorCode> error_code_;
 };
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -592,22 +592,85 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
                                                                         stream_id,
                                                                         VersionQuery{},
                                                                         ReadOptions{});
-    util::check(update_info.previous_index_key_.has_value(), "No previous version exists for write metadata");
-    ARCTICDB_DEBUG(log::version(), "write_versioned_dataframe for stream_id: {}", stream_id);
-    auto index_key = UpdateMetadataTask{store(),
-                                        update_info,
-                                        std::move(user_meta)}();
+    if(update_info.previous_index_key_.has_value()) {
+        ARCTICDB_DEBUG(log::version(), "write_versioned_dataframe for stream_id: {}", stream_id);
+        auto index_key = UpdateMetadataTask{store(),
+                                            update_info,
+                                            std::move(user_meta)}();
 
-    if(prune_previous_versions) {
-        auto versioned_item = VersionedItem{index_key};
-        auto pruned_indexes = version_map()->write_and_prune_previous(store(), index_key, update_info.previous_index_key_);
-        delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_).get();
+        if(prune_previous_versions) {
+            auto versioned_item = VersionedItem{index_key};
+            auto pruned_indexes = version_map()->write_and_prune_previous(store(), index_key, update_info.previous_index_key_);
+            delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_).get();
+            return versioned_item;
+        }
+        else {
+            version_map()->write_version(store(), index_key);
+            return VersionedItem{index_key};
+        }
+    } else {
+        auto frame = convert::py_none_to_frame();
+        frame.desc.set_id(stream_id);
+        frame.user_meta = std::move(user_meta);
+        auto versioned_item = write_versioned_dataframe_internal(stream_id, std::move(frame), prune_previous_versions, false, false);
+        if(cfg_.symbol_list())
+            symbol_list().add_symbol(store_, stream_id);
         return versioned_item;
     }
-    else {
-        version_map()->write_version(store(), index_key);
-        return VersionedItem{index_key};
+}
+
+std::vector<std::variant<VersionedItem, DataError>> LocalVersionedEngine::batch_write_versioned_metadata_internal(
+    const std::vector<StreamId>& stream_ids,
+    bool prune_previous_versions,
+    std::vector<arcticdb::proto::descriptors::UserDefinedMetadata>&& user_meta_protos) {
+    auto stream_update_info_vector_fut = batch_get_latest_undeleted_version_and_next_version_id_async(store(),
+                                                                                                      version_map(),
+                                                                                                      stream_ids);
+    internal::check<ErrorCode::E_ASSERTION_FAILURE>(stream_ids.size() == stream_update_info_vector_fut.size(), "stream_ids and stream_update_info_vector_fut must be of the same size");
+    std::vector<folly::Future<VersionedItem>> write_metadata_versions_futs;
+    for (const auto&& [idx, stream_update_info_fut] : folly::enumerate(stream_update_info_vector_fut)) {
+        write_metadata_versions_futs.push_back(
+            std::move(stream_update_info_fut)
+            .thenValue([this, prune_previous_versions, user_meta_proto = std::move(user_meta_protos[idx]), &stream_id = stream_ids[idx]](auto&& update_info) mutable -> folly::Future<IndexKeyAndUpdateInfo> {
+                auto index_key_fut = folly::Future<AtomKey>::makeEmpty();
+                if (update_info.previous_index_key_.has_value()) {
+                    index_key_fut = async::submit_io_task(UpdateMetadataTask{store(), update_info, std::move(user_meta_proto)});
+                } else {
+                    auto frame = convert::py_none_to_frame();
+                    frame.desc.set_id(stream_id);
+                    frame.user_meta = std::move(user_meta_proto);
+                    auto version_id = 0;
+                    auto write_options = get_write_options();
+                    auto de_dup_map = std::make_shared<DeDupMap>();
+                    index_key_fut = async_write_dataframe_impl(store(), version_id, std::move(frame), write_options, de_dup_map, false, false);
+                }
+                return std::move(index_key_fut)
+                .thenValue([update_info = std::move(update_info)](auto&& index_key) mutable -> IndexKeyAndUpdateInfo {
+                    return IndexKeyAndUpdateInfo{std::move(index_key), std::move(update_info)};
+                });
+            })
+            .thenValue([this, prune_previous_versions](auto&& index_key_and_update_info){
+                auto&& [index_key, update_info] = index_key_and_update_info;
+                return write_index_key_to_version_map_async(version_map(), std::move(index_key), std::move(update_info), prune_previous_versions, !update_info.previous_index_key_.has_value());
+            }));
     }
+
+    auto write_metadata_versions = folly::collectAll(write_metadata_versions_futs).get();
+    std::vector<std::variant<VersionedItem, DataError>> write_metadata_versions_or_errors;
+    write_metadata_versions_or_errors.reserve(write_metadata_versions.size());
+    for (auto&& [idx, write_metadata_version]: folly::enumerate(write_metadata_versions)) {
+        if (write_metadata_version.hasValue()) {
+            write_metadata_versions_or_errors.emplace_back(std::move(write_metadata_version.value()));
+        } else {
+            auto exception = write_metadata_version.exception();
+            DataError data_error(stream_ids[idx], exception.what().toStdString());
+            if (exception.is_compatible_with<storage::KeyNotFoundException>()) {
+                data_error.set_error_code(ErrorCode::E_KEY_NOT_FOUND);
+            }
+            write_metadata_versions_or_errors.emplace_back(std::move(data_error));
+        }
+    }
+    return write_metadata_versions_or_errors;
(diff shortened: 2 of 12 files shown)


python/arcticdb/version_store/library.py
```diff
@@ -1028,6 +1061,54 @@ def write_metadata(self, symbol: str, metadata: Any) -> VersionedItem:
...
+
+        Returns
+        -------
+        List
+            List of versioned items. The data attribute will be None for each versioned item.
+            i-th entry corresponds to i-th element of `symbols`.
+
+        Raises
+        ------
+        ArcticDuplicateSymbolsInBatchException
```
alexowens90, line 1088: What is the behaviour if you try and write metadata for a symbol that doesn't exist?

cpp/arcticdb/version/version_store_api.hpp
```diff
@@ -265,7 +265,7 @@ class PythonVersionStore : public LocalVersionedEngine {
         bool validate_index);
 
     std::vector<VersionedItem> batch_write_metadata(
-        std::vector<StreamId> stream_ids,
+        std::vector<StreamId>& stream_ids,
```
alexowens90, line 268: const?

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -265,6 +265,11 @@ class LocalVersionedEngine : public VersionedEngine {
         bool validate_index
     );
 
+    std::vector<VersionedItem> batch_write_versioned_metadata_internal(
+        std::vector<StreamId> stream_ids,
```
alexowens90, line 269: const&?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -592,6 +592,39 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
...
 }
 
+std::vector<VersionedItem> LocalVersionedEngine::batch_write_versioned_metadata_internal(
+    std::vector<StreamId> stream_ids,
+    bool prune_previous_versions,
+    std::vector<arcticdb::proto::descriptors::UserDefinedMetadata>&& user_meta_proto_vec) {
+    auto stream_update_info_vector_fut = batch_get_latest_undeleted_version_and_next_version_id_async(store(),
+                                                                                                      version_map(),
+                                                                                                      stream_ids);
+    std::vector<folly::Future<VersionedItem>> output_vect_fut;
```
alexowens90, line 602: Prefer `output_futs`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -592,6 +592,39 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
     }
 }
 
+std::vector<VersionedItem> LocalVersionedEngine::batch_write_versioned_metadata_internal(
+    std::vector<StreamId> stream_ids,
+    bool prune_previous_versions,
+    std::vector<arcticdb::proto::descriptors::UserDefinedMetadata>&& user_meta_proto_vec) {
+    auto stream_update_info_vector_fut = batch_get_latest_undeleted_version_and_next_version_id_async(store(),
```
alexowens90, line 599: Worth an `internal::check(stream_update_info_vector_fut.size() == stream_ids.size(), ...)` as it'll crash later if not

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -592,6 +592,39 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
...
+                                                                                                      version_map(),
+                                                                                                      stream_ids);
+    std::vector<folly::Future<VersionedItem>> output_vect_fut;
+    for(const auto& stream_id : folly::enumerate(stream_ids)) {
```
alexowens90, line 603: I discovered yesterday that this syntax is also valid, and I think cleaner in this case:
`for (const auto&& [idx, stream_id] : folly::enumerate(stream_ids))`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -592,6 +592,39 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
...
+        output_vect_fut.push_back(
+            std::move(stream_update_info_vector_fut[stream_id.index])
+            .thenValue([this, &stream_id, user_meta_proto = std::move(user_meta_proto_vec[stream_id.index])](auto&& update_info) mutable {
+                util::check(update_info.previous_index_key_.has_value(),
```
alexowens90, line 607: OK, this answers my previous question about what the behaviour is if the symbol doesn't exist. `folly::collect` short-circuits on the first exception, so the user will not know which symbols (if any) had metadata written. Would be better to log and return `None` rather than a versioned item in this case.
Unfortunately we will probably need to maintain backwards compatibility with the `NativeVersionStore API`. See my PR for an example of how to do this - https://github.com/man-group/ArcticDB/pull/544

alexowens90: Also, always use `internal::check` in preference to `util::check` now

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -592,6 +592,39 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
...
+                        "Failed to find latest version to write metadata to for symbol {}",
+                        *stream_id);
+                return async::submit_io_task(UpdateMetadataTask{store(), update_info, std::move(user_meta_proto)}).thenValue([update_info = std::move(update_info)](auto&& index_key) mutable {
```
alexowens90, line 610: This `thenValue` is pointless, just capture `update_info` in the next lambda

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -592,6 +592,39 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
...
+                });
+            })
+            .thenValue([this, prune_previous_versions](auto&& index_key_update_info_pair){
+                auto index_key = std::move(index_key_update_info_pair.first);
+                if(prune_previous_versions) {
+                    return async::submit_io_task(WriteAndPrunePreviousTask{store(), version_map(), index_key, index_key_update_info_pair.second.previous_index_key_})
+                    .then([index_key = std::move(index_key)](auto&& ){return index_key;});
+                } else {
+                    return async::submit_io_task(WriteVersionTask{store(), version_map(), index_key})
+                    .then([index_key = std::move(index_key)](auto&& ){return index_key;});
```
alexowens90, line 621: This duplicated code can be in a single `thenValue` chained to this one

alexowens90 (changes requested):

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -369,6 +371,38 @@ def test_read_meta_batch_with_as_ofs_stress(arctic_library):
...
+        for sym in range(num_symbols)
+        for version in range(num_versions)
+    ]
+    results_list = lib.read_metadata_batch(read_requests)
+    for sym in range(num_symbols):
+        for version in range(num_versions):
+            idx = sym * num_versions + version
+            assert results_list[idx].metadata == {"meta_" + str(sym): version}
+
+    read_requests = ["sym_" + str(sym) for sym in range(num_symbols)]
```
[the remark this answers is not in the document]
GemDot: it is quite redundant, I agree to remove it

python/arcticdb/version_store/library.py
```diff
@@ -173,6 +173,39 @@ def __iter__(self):
...
+    def __init__(self, symbol: str, metadata: Any):
+        """
+        Constructor.
+
+        Parameters
+        ----------
+        symbol : str
+            Symbol name. Limited to 255 characters. The following characters are not supported in symbols:
+            ``"*", "&", "<", ">"``
+        metadata : Any, default=None
```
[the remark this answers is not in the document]
GemDot: The comment is incorrect, there is no default value in this case. For the rest of your questions, I am maintaining the existing behaviour

python/arcticdb/version_store/library.py
```diff
@@ -1028,6 +1061,54 @@ def write_metadata(self, symbol: str, metadata: Any) -> VersionedItem:
         """
         return self._nvs.write_metadata(symbol, metadata, prune_previous_version=False)
 
+    def write_metadata_batch(
+        self, payloads: List[WriteMetadataPayload], prune_previous_versions=None
+    ) -> List[VersionedItem]:
+        """
+        Write metadata to multiple symbols in a batch fashion. This is more efficient than making multiple `write` calls
```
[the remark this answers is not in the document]
GemDot: ok

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: I don’t understand your comment here. I cannot capture the value in the next lambda because update_info is the value wrappered inside the future, which also needs to be propagated to the next future. On the other side, I need the ThenValue because one of the elements still is a future (what submit_io_task returns) and the other is an actual value, so I need to reach the point in time where both of them are actual values.

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: I don’t understand your comment here. I cannot capture the value in the next lambda because update_info is the value wrappered inside the future, which also needs to be propagated to the next future. On the other side, I need the ThenValue because one of the elements still is a future (what submit_io_task returns) and the other is an actual value, so I need to reach the point in time where both of them are actual values.

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: agree

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: agree

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: sounds plausible

python/tests/integration/arcticdb/test_arctic.py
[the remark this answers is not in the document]

python/arcticdb/version_store/library.py
```diff
@@ -173,6 +173,39 @@ def __iter__(self):
...
+            Symbol name. Limited to 255 characters. The following characters are not supported in symbols:
+            ``"*", "&", "<", ">"``
+        metadata : Any, default=None
+            metadata to persist along with the symbol.
+        """
+        self.symbol = symbol
+        self.metadata = metadata
+
+    def __repr__(self):
+        return f"WriteArgs(symbol={self.symbol}, metadata={self.metadata})"
```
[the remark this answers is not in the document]

python/arcticdb/version_store/library.py
```diff
@@ -1028,6 +1061,54 @@ def write_metadata(self, symbol: str, metadata: Any) -> VersionedItem:
...
+        in succession as some constant-time operations can be executed only once rather than once for each element of
+        `symbols`.
+        Note that this isn't an atomic operation - it's possible for the metadata for one symbol to be fully written and
+        readable before another symbol.
+        Parameters
+        ----------
+        payloads : `List[WriteMetadataPayload]`
+            Symbols and their corresponding metadata. There must not be any duplicate symbols in `payload`.
+        prune_previous_version : `Optional[bool]`, default=None
+            Remove previous versions from version list. Uses library default if left as None.
```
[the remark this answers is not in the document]
GemDot: as discussed, that's a not completely supported yet feature for batch methods. So, I will dismiss this one

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: amended

python/arcticdb/version_store/library.py
GemDot: As we discussed --> non-batch version: It creates a symbol if it does not exist. We will leave the behaviour as is, but adding all the logic into the C++ layer. Batch version, we want to do the same that the non-batch version, so we will not raise any exception, breaking the existing API

cpp/arcticdb/version/version_store_api.hpp
GemDot: yes agree

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: agree

cpp/arcticdb/version/local_versioned_engine.hpp
GemDot: agree

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1001,39 +1001,16 @@ std::pair<VersionedItem, py::object> PythonVersionStore::read_metadata(
...
-        fut_vec.push_back(async::submit_io_task(UpdateMetadataTask{store(),
-                                                                   update_info,
-                                                                   std::move(user_meta_proto)}));
-    }
-    auto index_keys = folly::collect(fut_vec).get();
-    if(prune_previous_versions) {
-        batch_write_and_prune_previous(store(), version_map(), index_keys, stream_update_info_vector);
-    } else {
-        batch_write_version(store(), version_map(), index_keys);
+        user_meta_proto_vec.push_back(std::move(user_meta_proto));
```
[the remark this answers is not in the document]
GemDot: I don't understand exactly what you are suggesting. Could you please be a bit more explicit?

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -225,6 +225,22 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
         return lst;
     };
 
+    auto adapt_write_dfs = [](std::vector<std::variant<VersionedItem, DataError>> && ret) -> py::list {
```
alexowens90, line 228: Is this needed? `pybind11` will convert a vector to a list anyway, and I think it understands `std::variant` so this behaviour should come for free?
If it is necessary, `adapt_write_dfs` does not describe what this function does at all.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -225,6 +225,22 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
...
+        for (auto &res: ret) {
+            util::variant_match(
+                    res,
+                    [&lst] (VersionedItem& res) {
+                        lst.append(res);
+                    },
+                    [&lst] (DataError& data_error) {
+                        lst.append(data_error);
+                    }
+            );
```
alexowens90, line 239: When all of the visitors are the same, you can write it more concisely:
```
            util::variant_match(
                    elem,
                    [&lst] (auto& elem) {
                        lst.append(elem);
                    }
            );
```

cpp/arcticdb/version/version_store_api.cpp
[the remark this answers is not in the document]
alexowens90: `*stream_id`, you're just using `stream_id.index` to index into `user_meta`. So the `for` loop should just be over `user_meta`.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1009,40 +1009,17 @@ std::pair<VersionedItem, py::object> PythonVersionStore::read_metadata(
...
     const std::vector<py::object>& user_meta,
     bool prune_previous_versions) {
-    std::vector<std::pair<VersionedItem, py::object>> results;
-    std::vector<AtomKey> keys_to_read;
-    auto stream_update_info_vector_fut = batch_get_latest_undeleted_version_and_next_version_id_async(store(),
-                                                                                                      version_map(),
-                                                                                                      stream_ids);
-    auto stream_update_info_vector = folly::collect(stream_update_info_vector_fut).via(&async::io_executor()).get();
-    std::vector<folly::Future<AtomKey>> fut_vec;
+    std::vector<arcticdb::proto::descriptors::UserDefinedMetadata> user_meta_proto_vec;
```
alexowens90, line 1016: `_vec` suffix not needed

alexowens90: Fox maximum efficiency, you can `reserve` `user_meta.size()`, and use `emplace_back` rather than `push_back` below

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -51,7 +51,7 @@ inline std::vector<folly::Future<version_store::UpdateInfo>> batch_get_latest_un
                                                      version_map,
                                                      stream_id,
                                                      LoadParameter{LoadType::LOAD_LATEST_UNDELETED}})
-        .thenValue([](auto&& entry){
+        .thenValue([&stream_id](auto&& entry){
```
alexowens90, line 54: This seems to be unused

cpp/arcticdb/entity/data_error.hpp
```diff
@@ -25,30 +25,32 @@ enum class VersionRequestType: uint32_t {
 class DataError {
 public:
     DataError(const StreamId& symbol,
-              const pipelines::VersionQueryType& version_query_type,
+              const std::optional<pipelines::VersionQueryType>& version_query_type,
```
alexowens90, line 28: It is much cleaner to reorder the arguments so that `version_query_type` is defaulted to `std::nullopt`, rather than callers having to explicitly provide `std::nullopt`

alexowens90 (changes requested):

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -185,6 +186,78 @@ def test_read_meta_batch_with_as_ofs_high_amount(arctic_library):
             assert results_list[idx].metadata == {"meta_" + str(sym): version}
 
 
+def test_write_meta_batch_with_as_ofs(arctic_library):
+    lib = arctic_library
+    num_symbols = 10
+    num_versions = 20
```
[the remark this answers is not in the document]
GemDot: we want sometimes to stress a bit more the tests, to catch thread concurrency issues. But fine to cut up to 5

python/arcticdb/version_store/library.py
```diff
@@ -173,6 +173,39 @@ def __iter__(self):
...
+            Symbol name. Limited to 255 characters. The following characters are not supported in symbols:
+            ``"*", "&", "<", ">"``
+        metadata : Any
+            metadata to persist along with the symbol.
+        """
+        self.symbol = symbol
+        self.metadata = metadata
+
+    def __repr__(self):
+        return f"WriteMetadataArgs(symbol={self.symbol}, metadata={self.metadata})"
```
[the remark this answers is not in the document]
GemDot: I disagree. It is consistent with WritePayload

cpp/arcticdb/version/version_map_batch_methods.hpp
GemDot: yes, right. It was probably necessary in another commit and I forgot to remove it when it was not longer used.

cpp/arcticdb/entity/data_error.hpp
GemDot: yes it is cleaner, agree

cpp/arcticdb/version/python_bindings.cpp
GemDot: Removed the function entirely.

cpp/arcticdb/version/version_store_api.cpp
GemDot: agree x2

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -185,6 +186,78 @@ def test_read_meta_batch_with_as_ofs_high_amount(arctic_library):
...
+    for sym in range(num_symbols):
+        for version in range(num_versions):
+            idx = sym * num_versions + version
+            assert results_list[idx].metadata == {"meta_" + str(sym): version}
+
+
+def test_write_meta_batch_with_none(arctic_library):
+    lib = arctic_library
+    symbol = "symbol_"
+    symbol_mirror = "symbol_mirror_"
```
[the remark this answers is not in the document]
GemDot: agree

python/arcticdb/version_store/library.py
```diff
@@ -1037,11 +1070,63 @@ def write_metadata(self, symbol: str, metadata: Any) -> VersionedItem:
...
+        Parameters
+        ----------
+        payloads : `List[WriteMetadataPayload]`
+            Symbols and their corresponding metadata. There must not be any duplicate symbols in `payload`.
+        prune_previous_version : `Optional[bool]`, default=None
+            Remove previous versions from version list. Uses library default if left as None.
+
+        Returns
+        -------
+        List
```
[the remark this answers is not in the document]
GemDot: I wrongly added all this explanation in the non-batch version

python/arcticdb/version_store/_store.py
```diff
@@ -1205,6 +1205,28 @@ def batch_write(
...
+
+        proto_cfg = self._lib_cfg.lib_desc.version.write_options
+        prune_previous_version = self.resolve_defaults(
+            "prune_previous_version", proto_cfg, global_default=False, existing_value=prune_previous_version
+        )
+        normalized_meta = [normalize_metadata(metadata_vector[idx]) for idx in range(len(symbols))]
+        write_metadata_results = self.version_store.batch_write_metadata(
+            symbols, normalized_meta, prune_previous_version
+        )
+        cxx_versioned_items = []
```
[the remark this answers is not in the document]
GemDot: agree

cpp/arcticdb/version/version_store_api.cpp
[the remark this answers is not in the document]
GemDot: ok understood

alexowens90 (changes requested):

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
GemDot: I will update WritePayload as well then @alexowens90

python/tests/integration/arcticdb/test_arctic_batch.py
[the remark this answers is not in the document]
GemDot: I will switch to 2 symbols

alexowens90 (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -592,22 +592,85 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
...
+    if(update_info.previous_index_key_.has_value()) {
+        ARCTICDB_DEBUG(log::version(), "write_versioned_dataframe for stream_id: {}", stream_id);
+        auto index_key = UpdateMetadataTask{store(),
+                                            update_info,
+                                            std::move(user_meta)}();
+
+        if(prune_previous_versions) {
+            auto versioned_item = VersionedItem{index_key};
+            auto pruned_indexes = version_map()->write_and_prune_previous(store(), index_key, update_info.previous_index_key_);
+            delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_).get();
```
willdealtry, line 604: Is this get() okay?

PR #485, opened 2023-06-13
Set prune_previous_versions as False by default

GemDot:
prune_previous_versions=True is a usability concern. If you intend to keep history but forget to call this argument once, you lose all your history. Best to set to None (for potential future library level option), or False if we want to require users to manually clean up versions from now onwards. This PR, set this value as False by default.

### DIFF
--- a/python/tests/integration/arcticdb/test_arctic.py
+++ b/python/tests/integration/arcticdb/test_arctic.py
@@ -421,6 +421,17 @@ def test_prune_previous_versions(arctic_library):
     assert lib["symbol"].metadata == {"tres": "interessant"}
 
 
+def test_non_prune_previous_versions_by_default(arctic_library):
+    lib = arctic_library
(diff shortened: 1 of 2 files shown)


mehertz (approved):

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -687,7 +698,7 @@ def test_prune_previous_versions_with_write(arctic_library):
     assert not v1.empty
 
     # We prune by default
```
poodlewars, line 700: The comment is wrong

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -421,6 +421,17 @@ def test_prune_previous_versions(arctic_library):
     assert lib["symbol"].metadata == {"tres": "interessant"}
 
 
+def test_non_prune_previous_versions_by_default(arctic_library):
```
poodlewars, line 424: Name like `test_do_not_prune_previous_versions_by_default`

PR #487, opened 2023-06-14
Address post merging comments from prune_previous_versions set defaul…

GemDot:
Address post-merging comments from PR 485: https://github.com/man-group/ArcticDB/pull/485

### DIFF
--- a/docs/mkdocs/docs/index.md
+++ b/docs/mkdocs/docs/index.md
@@ -323,6 +323,4 @@ Note the timestamps - we've read the data prior to the _append_ operation. Pleas
 
 !!! note "Versioning & Prune Previous"
 
-    By default, `write`, `append`, and `update` operations will remove the previous versions to save on space. 
-
-    Use the `prune_previous` argument to control this behaviour. 
+    By default, `write`, `append`, and `update` operations will remove the previous versions. Please be aware that having this option enabled results in consuming more space.
--- a/docs/mkdocs/docs/index.md
+++ b/docs/mkdocs/docs/index.md
@@ -323,4 +323,6 @@ Note the timestamps - we've read the data prior to the _append_ operation. Pleas
 
 !!! note "Versioning & Prune Previous"
 
-    By default, `write`, `append`, and `update` operations will remove the previous versions. Please be aware that having this option enabled results in consuming more space.
+    By default, `write`, `append`, and `update` operations will **not** remove the previous versions. Please be aware that this will consume more space.
+    
+    This behaviour can be can be controlled via the `prune_previous_versions` keyword argument. 
--- a/python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
+++ b/python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
@@ -114,7 +114,7 @@ class VersionStoreComparison(RuleBasedStateMachine):
 
     @rule(
         sym=symbols,
-        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False, but prune is on by default
+        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False, but prune is off by default
         write_mode=st.sampled_from(WriteMode),
     )
     def write_new_version_to_symbol(self, sym: str, prune: bool, write_mode: WriteMode):
--- a/python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
+++ b/python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
@@ -114,7 +114,7 @@ class VersionStoreComparison(RuleBasedStateMachine):
 
     @rule(
         sym=symbols,
-        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False, but prune is off by default
+        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False
         write_mode=st.sampled_from(WriteMode),
     )
     def write_new_version_to_symbol(self, sym: str, prune: bool, write_mode: WriteMode):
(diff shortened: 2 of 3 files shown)


python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
```diff
@@ -114,7 +114,7 @@ def _get_latest_undeleted_version(self, sym) -> Tuple[Optional[int], Optional[Ve
 
     @rule(
         sym=symbols,
-        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False, but prune is on by default
+        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False, but prune is off by default
```
poodlewars, line 117: I think that just 

```
# booleans() shrinks towards False
```

is clearer, the default prune_previous isn't used by the logic of the test so the comment was misleading before

poodlewars (approved):

docs/mkdocs/docs/index.md
```diff
@@ -323,6 +323,4 @@ Note the timestamps - we've read the data prior to the _append_ operation. Pleas
 
 !!! note "Versioning & Prune Previous"
 
-    By default, `write`, `append`, and `update` operations will remove the previous versions to save on space. 
-
-    Use the `prune_previous` argument to control this behaviour. 
+    By default, `write`, `append`, and `update` operations will remove the previous versions. Please be aware that having this option enabled results in consuming more space.
```
mehertz, line 326:
```suggestion
    By default, `write`, `append`, and `update` operations will **not** remove the previous versions. Please be aware that this will consume more space.
    
    This behaviour can be can be controlled via the `prune_previous_versions` keyword argument. 
```

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
GemDot: removed

docs/mkdocs/docs/index.md
```diff
@@ -323,6 +323,6 @@ Note the timestamps - we've read the data prior to the _append_ operation. Pleas
 
 !!! note "Versioning & Prune Previous"
 
-    By default, `write`, `append`, and `update` operations will remove the previous versions to save on space. 
-
-    Use the `prune_previous` argument to control this behaviour. 
+    By default, `write`, `append`, and `update` operations will **not** remove the previous versions. Please be aware that this will consume more space.
```
qc00, line 326: Please place each sentence on its own line.

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
```diff
@@ -114,7 +114,7 @@ def _get_latest_undeleted_version(self, sym) -> Tuple[Optional[int], Optional[Ve
 
     @rule(
         sym=symbols,
-        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False, but prune is on by default
+        prune=st.sampled_from([True, False]),  # booleans() shrinks towards False
```
qc00, line 117: This change is incorrect: most tests don't use `library.py` so will respect the library option

GemDot: do you mean that they might use a library where prune_previous_versions option is explicitly set? @qc00

docs/mkdocs/docs/index.md
GemDot: I think they are. "Please be aware that this will consume more space" is related to previous sentence

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
qc00: Each library option has a default value which is used if not explicitly overridden when called in Python.

The `library.py` methods have your `prune_previous_version=False` changes, but they're not used in this test or most other tests. They use the `_store.py` methods, which has `prune_previous_version=None`.

GemDot: Understood. However, this commit only involves deleting the 'but prune is off by default' comment. With the default value of the library's prune_previous_version attribute and the default value of the function call for the same attribute, the comment can still be confusing. So, I think is good to remove it, as @poodlewars suggested

GemDot: Docs should be the way to disambiguate any question/doubt. The comment should relate only to the test itself

qc00: But if you remove that bit, the whole comment makes no sense.

The comment indicated the unusual behaviour of using `st.sampled_from([True, False])` instead of `st.booleans()` is because the latter shrinks towards the default library option, which has less testing power than the non-default (`True`).

docs/mkdocs/docs/index.md
mehertz: `Please be aware that this will consume more space.` is on the same line. Line 326.

It does make it marginally harder to comment.

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
mehertz: As we're manually passing in the keyword arg, and `but prune is on by default` refers to the library option which is being overridden, I think removing this part of the comment makes things clearer.

poodlewars: Yeah the comment makes sense without the last clause. It's saying that we have that line to force a shrink towards True. The default never affected the actual logic of the test.

docs/mkdocs/docs/index.md
GemDot: yes, I will leave it as is then

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
GemDot: agree

PR #489, opened 2023-06-15
Print a warning if the user should expect `list_symbols` to be slow

mehertz:
### What does this implement/fix? Explain your changes.

This PR:

* Prints a warning if the user should expect `list_symbols` to be slow, linking to the documentation.
* Fixes the symbol list logging - previously it wouldn't render.
* Fixes a typo in the documentation

### Any other comments?

There are no tests in this PR as all changes are either logging changes or documentation changes and neither are easy to test. 

The two warnings are:

```
[arcticdb] [warning] Warning - symbol list cache is significantly out of date. `list_symbols` may take longer than expected. 

See here for more information: https://docs.arcticdb.io/technical/on_disk_storage/#symbol-list-caching

To resolve, run `list_symbols` through to completion frequently.
```

and:

```
[2023-06-15 15:17:40.863] [arcticdb] [warning] Warning - no compacted symbol list cache found. `list_symbols` may take longer than expected. 

See here for more information: https://docs.arcticdb.io/technical/on_disk_storage/#symbol-list-caching

To resolve, run `list_symbols` through to completion to compact the symbol list cache.
```

This also adds the following to our documentation:

image

### DIFF
--- a/cpp/arcticdb/version/symbol_list.cpp
+++ b/cpp/arcticdb/version/symbol_list.cpp
@@ -91,8 +91,8 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
             }, std::get<std::string>(compaction_id));
     }
 
-    SYMBOL_LIST_RUNTIME_LOG("found_last=", found_last);
-    SYMBOL_LIST_RUNTIME_LOG("has_newer=", has_newer);
+    SYMBOL_LIST_RUNTIME_LOG_VAL("found_last={}", found_last);
+    SYMBOL_LIST_RUNTIME_LOG_VAL("has_newer={}", has_newer);
     return (!maybe_last_compaction || found_last) && !has_newer;
 }
 
@@ -119,7 +119,7 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
             stream_ids.push_back(id);
 
             if (stream_ids.size() == this->max_delta_ * 2 && !warned_expected_slowdown_) {
-                log::storage().warn(
+                log::version().warn(
                         "Warning - no compacted symbol list cache found. "
                         "`list_symbols` may take longer than expected. \n\n"
                         "See here for more information: https://docs.arcticdb.io/technical/on_disk_storage/#symbol-list-caching\n\n"
@@ -145,7 +145,7 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
         const StreamId& stream_id,
         timestamp creation_ts)  {
 
-        SYMBOL_LIST_RUNTIME_LOG("Writing following number of symbols to symbol list cache: ", symbols.size());
+        SYMBOL_LIST_RUNTIME_LOG_VAL("Writing following number of symbols to symbol list cache: {}", symbols.size());
         SegmentInMemory list_segment{symbol_stream_descriptor(stream_id)};
 
         for(const auto& symbol : symbols) {
@@ -177,7 +177,7 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
             const folly::Range<KeyVectorItr>& keys) {
         SYMBOL_LIST_RUNTIME_LOG("Loading symbols from symbol list keys");
         bool read_compaction = false;
-        int uncompacted_keys_found = 0;
+        uint64_t uncompacted_keys_found = 0;
         CollectionType symbols{};
         for(const auto& key : keys) {
             if(key.id() == compaction_id) {
@@ -187,7 +187,7 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
             else {
                 uncompacted_keys_found++;
                 if (uncompacted_keys_found == this->max_delta_ * 2 && !warned_expected_slowdown_) {
-                    log::storage().warn(
+                    log::version().warn(
                             "Warning - symbol list cache is significantly out of date. "
                             "`list_symbols` may take longer than expected. \n\n"
                             "See here for more information: https://docs.arcticdb.io/technical/on_disk_storage/#symbol-list-caching\n\n"
@@ -207,9 +207,9 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
                 }
             }
         }
-        SYMBOL_LIST_RUNTIME_LOG("Post load, got following number of symbols: ", symbols.size());
+        SYMBOL_LIST_RUNTIME_LOG_VAL("Post load, got following number of symbols: {}", symbols.size());
         if(!read_compaction) {
-            SYMBOL_LIST_RUNTIME_LOG("Read no compacted segment from symbol list of size:", keys.size());
+            SYMBOL_LIST_RUNTIME_LOG_VAL("Read no compacted segment from symbol list of size: {}", keys.size());
         }
         return symbols;
     }
--- a/cpp/arcticdb/version/symbol_list.cpp
+++ b/cpp/arcticdb/version/symbol_list.cpp
@@ -91,7 +91,8 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
             }, std::get<std::string>(compaction_id));
     }
 
-    SYMBOL_LIST_RUNTIME_LOG("found_last={}, has_newer={}", found_last, has_newer);
+    SYMBOL_LIST_RUNTIME_LOG("found_last=", found_last);
+    SYMBOL_LIST_RUNTIME_LOG("has_newer=", has_newer);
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -116,6 +117,16 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
         store->iterate_type(KeyType::VERSION_REF, [&] (const auto& key) {
             auto id = variant_key_id(key);
             stream_ids.push_back(id);
+
+            if (stream_ids.size() == this->max_delta_ * 2 && !warned_expected_slowdown_) {
```
poodlewars, line 121: What's the intent of the `warned_expected_slowdown_` for this one?
Why do we say "to resolve" as if some user action is required, this is already in the context of a `list_symbols` call isn't it so it's about to compact the cache anyway?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -166,13 +177,24 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
...
+        int uncompacted_keys_found = 0;
         CollectionType symbols{};
         for(const auto& key : keys) {
             if(key.id() == compaction_id) {
                 read_list_from_storage(store, key, symbols);
                 read_compaction = true;
             }
             else {
+                uncompacted_keys_found++;
+                if (uncompacted_keys_found == this->max_delta_ * 2 && !warned_expected_slowdown_) {
```
poodlewars, line 189: How did you choose `max_delta_ * 2` as the heuristic? Seems rather small, is this really so slow with 1000 symbols? 

Again, what's the point of the `warned_expected_slowdown_` state?

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -30,7 +30,10 @@
 #include <mutex>
 
 // FUTURE(GitHub #297)
-#define SYMBOL_LIST_RUNTIME_LOG(...) ARCTICDB_RUNTIME_DEBUG(log::version(), "Symbol List: {}: ", __func__, __VA_ARGS__)
```
poodlewars, line 33: don't see what was wrong with having the prefix?

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -50,6 +53,7 @@ class SymbolList {
     StreamId type_holder_;
     uint64_t max_delta_ = 0;
     std::shared_ptr<VersionMap> version_map_;
+    bool warned_expected_slowdown_ = false;
```
poodlewars, line 56: any thread-safety concerns? Think this is the only mutable state

cpp/arcticdb/version/symbol_list.cpp
mehertz: > isn't it so it's about to compact the cache anyway?

You're right, it is. However this first run of `list_symbols` might take a _really_ long time depending on how many symbols there are. Currently it appears as if it is hanging if there are a lot - so even though we're about to resolve the issue it's still worth highlighting this to the user. 

It's also possible that this won't proceed to compaction anyway though because:

* The user may not have write privileges  
* The process may be killed prior to the compaction key being written

I'm definitely open to claifying the key points in the wording. It's hard to be clear without overloading!

cpp/arcticdb/version/symbol_list.cpp
mehertz: At default settings `max_delta_ * 2` is the minimal value that will result in at least two `list_object` calls. Otherwise though it's a heuristic that I've mostly plucked out of thin air...

> is this really so slow with 1000 symbols?

Depends. One user is saying that it's taking minutes with ~1,400 so I think the answer is yes :/ 

> Again, what's the point of the warned_expected_slowdown_ state?

If the user doesn't have write privileges this will repeat for every invocation. I thought that'd be annoying. Happy to remove if you think this checking is unnecessary?

cpp/arcticdb/version/symbol_list.hpp
mehertz: Ah good question. Probably? I think this means it's best to remove!

cpp/arcticdb/version/symbol_list.cpp
poodlewars: I didn't appreciate the point about write privs, so I think what you've done makes sense.

I think we might need to tune the heuristic at some point but I guess we can see how it goes.

cpp/arcticdb/version/symbol_list.hpp
poodlewars: Reckon we probably do want it after all given what you said about,

```
If the user doesn't have write privileges this will repeat for every invocation. 
```

mehertz: Using std::atomic<bool>

cpp/arcticdb/version/symbol_list.hpp
mehertz: Nothing is wrong with the prefix. `__VA_ARGS__` includes strings with `{}` which aren't effectively replaced so most messages were just blank.

Now `message` is concatenated rather than formatted this will work.

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -30,7 +30,8 @@
 #include <mutex>
 
 // FUTURE(GitHub #297)
-#define SYMBOL_LIST_RUNTIME_LOG(...) ARCTICDB_RUNTIME_DEBUG(log::version(), "Symbol List: {}: ", __func__, __VA_ARGS__)
+#define SYMBOL_LIST_RUNTIME_LOG(message) ARCTICDB_RUNTIME_DEBUG(log::version(), "Symbol List: {}: " message, __func__)
```
willdealtry, line 33: We are jumping through a lot of hoops here to avoid creating a new log channel, which would not only indicate that the logging relates to the symbol list but would allow us to toggle it on and off independently

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -166,13 +177,24 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
...
+        uint64_t uncompacted_keys_found = 0;
         CollectionType symbols{};
         for(const auto& key : keys) {
             if(key.id() == compaction_id) {
                 read_list_from_storage(store, key, symbols);
                 read_compaction = true;
             }
             else {
+                uncompacted_keys_found++;
+                if (uncompacted_keys_found == this->max_delta_ * 2 && !warned_expected_slowdown_) {
```
qc00, line 189: This is wrong. This branch is processed entirely in RAM with no possibility of any noticeable slowdown.

docs/mkdocs/docs/technical/on_disk_storage.md
```diff
@@ -130,4 +140,4 @@ Without any housekeeping, this process could lead to unbounded growth in the num
...
 
-Astute observers will correctly take issue with basing logical decisions on timestamps in a purely client-side database with no synchronisation between clocks of different clients enforced. As such, this cache can diverge from reality, if two different clients create and delete the same symbol at about the same time, and this is the most likely cause of odd behaviour such as `lib.read(symbol)` working, but `symbol in lib.list_symols()` returning `False`. If this happens, `lib.reload_symbol_list()` should resolve the issue.
+Astute observers will correctly take issue with basing logical decisions on timestamps in a purely client-side database with no synchronisation between clocks of different clients enforced. As such, this cache can diverge from reality, if two different clients create and delete the same symbol at about the same time, and this is the most likely cause of odd behaviour such as `lib.read(symbol)` working, but `symbol in lib.list_symbols()` returning `False`. If this happens, `lib.reload_symbol_list()` should resolve the issue.
```
qc00, line 143: Matt

# PLEASE USE PUT EACH SENTENCE ON ITS OWN LINE IN MARKDOWN!!!

cpp/arcticdb/version/symbol_list.cpp
mehertz: This has to go one level higher, in other words? Line 34?

cpp/arcticdb/version/symbol_list.hpp
mehertz: That's fair - I just didn't want to change it in this set of changes really.

cpp/arcticdb/version/symbol_list.cpp
qc00: This is the "fast"/"cached" code path, so I don't think any warning is necessary in this code path. This is as fast as the user can possibly get for `list_symbols`.

mehertz: https://github.com/man-group/ArcticDB/pull/500

PR #491, opened 2023-06-15
Try removing enum-compat

poodlewars:
It shouldn't be necessary on Python > 3.4, which is all that we support anyway.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,6 @@ install_requires =
     pandas <2
     attrs
     dataclasses ; python_version < '3.7'
-    enum-compat
     grpcio-tools
     protobuf >=3.5.0.post1 # Per https://github.com/grpc/grpc/blob/v1.45.3/requirements.txt
     xxhash


jjerphan (approved):
Indeed this can be removed. The specification of the conda environment (i.e. [`environment_unix.yml`](https://github.com/man-group/ArcticDB/blob/master/environment_unix.yml)) already does not specifies `enum-compat` (hence nothing is do adapt for building with conda).

poodlewars:
> Indeed this can be removed. The specification of the conda environment (i.e. [`environment_unix.yml`](https://github.com/man-group/ArcticDB/blob/master/environment_unix.yml)) already does not specifies `enum-compat` (hence nothing is do adapt for building with conda).

But the feedstock does? https://github.com/conda-forge/arcticdb-feedstock/blob/main/recipe/meta.yaml#L69

jjerphan:
Indeed, fixed by https://github.com/conda-forge/arcticdb-feedstock/pull/11.

I do not know if we can get ride of this duplication for the dependencies' specification since different tools use their own source of truth.

PR #492, opened 2023-06-15
Add a means of caching data for each symbol such as start time for qu…

willdealtry:
In the medium term, we plan to add data to the symbol list to add some constantly-collected statistics about the start and end time-ranges for a symbol, the number of rows etc. However, whilst this is not a particularly difficult task, it is non-trivial to get correct as at the moment the symbol list assumes that data is time-independent, a symbol either exists or doesn't. Ensuring the data is correct would mean correctly selecting the latest symbol list entry on compaction.

In the meantime, this PR provides a method to generate a symbol containing the statistics for all other symbols in the library. This should be relatively quick to generate as it works in parallel, and unlike batch_read_descriptor will involve at most a handful of IO operations, so should be extremely fast to retrieve.

Written in one day, feel free to take a look but I wll tidy a bit on Monday, specifically I intend to split the statistics generation and write into two non-class functions in the cpp file, and pass in the class members such as store and version map from the now-trivial existing function

### DIFF
--- a/cpp/arcticdb/stream/protobuf_mappings.hpp
+++ b/cpp/arcticdb/stream/protobuf_mappings.hpp
@@ -29,6 +29,16 @@ inline arcticdb::proto::descriptors::NormalizationMetadata make_timeseries_norm_
     return norm_meta;
 }
 
+inline arcticdb::proto::descriptors::NormalizationMetadata make_rowcount_norm_meta(const entity::StreamId& stream_id) {
+    using namespace arcticdb::proto::descriptors;
+    NormalizationMetadata norm_meta;
+    NormalizationMetadata_PandasDataFrame pandas;
+    auto id = std::get<entity::StringId>(stream_id);
+    pandas.mutable_common()->set_name(std::move(id));
+    norm_meta.mutable_df()->CopyFrom(pandas);
+    return norm_meta;
+}
+
 /**
  * Set the minimum defaults into norm_meta. Originally created to synthesize norm_meta for incomplete compaction.
  */
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1066,6 +1066,106 @@ VersionedItem LocalVersionedEngine::append_internal(
     }
 }
 
+VersionedItem LocalVersionedEngine::refresh_symbol_cache() {
+    struct SymbolData {
+        StreamId stream_id_;
+        timestamp first_index_;
+        timestamp last_index_;
+        size_t num_rows_;
+    };
+
+    auto stream_ids = symbol_list().get_symbols(store());
+    std::vector<VersionQuery> version_queries;
+    version_queries.resize(stream_ids.size());
+    auto versions = batch_get_versions(store(), version_map(), stream_ids, version_queries);
+    std::vector<folly::Future<SymbolData>> output;
+    for (auto &&version : folly::enumerate(versions)) {
+        output.emplace_back(std::move(*version).thenValue([store = store()](auto maybe_index_key) -> ReadKeyOutput {
+            util::check(static_cast<bool>(maybe_index_key), "Version not found for stream");
+            return store->read_sync(maybe_index_key.value());
+        }).thenValue([store = store()](auto &&key_segment_pair) {
+            auto [index_key, index_segment] = std::move(key_segment_pair);
+            using FieldType = pipelines::index::Fields;
+            const auto first_index = pipelines::index::index_start_from_segment<SegmentInMemory, FieldType>(index_segment, 0);
+            const auto last_index = pipelines::index::index_end_from_segment<SegmentInMemory, FieldType>(index_segment, index_segment.get_row_id());
+            const auto last_row = index_segment.template scalar_at<std::size_t>(index_segment.get_row_id(), static_cast<position_t>(FieldType::end_row));
+            const auto& stream_id = variant_key_id(index_key);
+            util::check(std::holds_alternative<timestamp>(first_index), "First index key is not of timestamp type in symbol {}", stream_id);
+            util::check(std::holds_alternative<timestamp>(last_index), "Last index key is not of timestamp type in symbol {}", stream_id);
+            util::check(static_cast<bool>(last_row), "Unable to determine last row for symbol {}", stream_id);
+            return SymbolData {
+                stream_id,
+                std::get<timestamp>(first_index),
+                std::get<timestamp>(last_index),
+                last_row.value()
+            };
+        }));
+    }
+    auto symbols_data = folly::collect(output).get();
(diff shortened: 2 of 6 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1066,6 +1066,106 @@ VersionedItem LocalVersionedEngine::append_internal(
...
+    auto versions = batch_get_versions(store(), version_map(), stream_ids, version_queries);
+    std::vector<folly::Future<SymbolData>> output;
+    for (auto &&version : folly::enumerate(versions)) {
+        output.emplace_back(std::move(*version).thenValue([store = store()](auto maybe_index_key) -> ReadKeyOutput {
+            util::check(static_cast<bool>(maybe_index_key), "Version not found for stream");
+            return store->read_sync(maybe_index_key.value());
+        }).thenValue([store = store()](auto &&key_segment_pair) {
+            auto [index_key, index_segment] = std::move(key_segment_pair);
+            using FieldType = pipelines::index::Fields;
+            const auto first_index = pipelines::index::index_start_from_segment<SegmentInMemory, FieldType>(index_segment, 0);
```
mehertz, line 1089: This is intentionally not reading the descriptor right? So won't return timezone information?

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -281,6 +281,8 @@ class LocalVersionedEngine : public VersionedEngine {
 
     timestamp get_update_time_internal(const StreamId &stream_id, const VersionQuery &version_query);
 
+    VersionedItem refresh_symbol_cache();
```
mehertz, line 284: Even internal to the C++ I don't think `symbol_cache` is a clear name. In fact even though it's not necessary, I'm inclined to think in this case that `symbol_stats_cache` should be the consistent name used throughout both the Python and C++.

mehertz (changes requested):
Please can you write _detailed_ docstrings for both the generation method and the read method - and then pop those descriptions up on the PR description. 

Will mean anybody who is linked to the PR can quickly get to grips with what is being added and whether it'll be of value to them.

willdealtry:
These are the internal utc times

cpp/arcticdb/version/local_versioned_engine.cpp
willdealtry: We could add a timezone column to the output dataframe if required.

python/arcticdb/version_store/_store.py
```diff
@@ -2604,3 +2604,18 @@ def library(self):
 
     def library_tool(self) -> LibraryTool:
         return LibraryTool(self.library())
+
+    def generate_symbol_stats(self):
```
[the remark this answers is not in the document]
willdealtry: On a method we intend to remove?

cpp/arcticdb/stream/protobuf_mappings.hpp
```diff
@@ -29,6 +29,16 @@ inline arcticdb::proto::descriptors::NormalizationMetadata make_timeseries_norm_
     return norm_meta;
 }
 
+inline arcticdb::proto::descriptors::NormalizationMetadata make_rowcount_norm_meta(const entity::StreamId& stream_id) {
```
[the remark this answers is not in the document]
willdealtry: No, it's generally applicable to any object requiring normalization metadata that is rowcount indexed, and this matches the make_timeseries_norm_meta above

cpp/arcticdb/version/local_versioned_engine.cpp
mehertz: But then we'll need to read the stream descriptor?

willdealtry: The encoded field changes will make that all faster, we can choose to decode just the protobuf part which is small and fast, or the whole descriptor including columns etc. The problem at the moment is it's all or nothing

python/arcticdb/version_store/_store.py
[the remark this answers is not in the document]
willdealtry: They are really just for the one particular limited case. Given the number of things currently in version store that are not in library.py, I would say that these are far down the list

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -325,6 +327,10 @@ class LocalVersionedEngine : public VersionedEngine {
         const WriteOptions& write_options
     );
 
+    StreamId symbol_cache_id() {
+        return "__symbol_cache__";
```
qc00, line 331: Why not call this `__symbol_stats[_cache]__` just to be more specific and more consistent with the method writing it?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1066,6 +1066,106 @@ VersionedItem LocalVersionedEngine::append_internal(
...
+    };
+
+    auto stream_ids = symbol_list().get_symbols(store());
+    std::vector<VersionQuery> version_queries;
+    version_queries.resize(stream_ids.size());
+    auto versions = batch_get_versions(store(), version_map(), stream_ids, version_queries);
+    std::vector<folly::Future<SymbolData>> output;
+    for (auto &&version : folly::enumerate(versions)) {
+        output.emplace_back(std::move(*version).thenValue([store = store()](auto maybe_index_key) -> ReadKeyOutput {
+            util::check(static_cast<bool>(maybe_index_key), "Version not found for stream");
```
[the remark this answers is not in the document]
willdealtry: I think in that edge case I would prefer to fail. What I want to do is to constrain the state space, so that we either move from a valid prior state to a valid end state, or remain obviously in the valid (if out-of-date) prior state. Introducing "resilience" in this sort of operation in my experience produces a sort of "grey failure", where specific components such as symbols repeatedly fail to update, but the problem is not evident because the majority are updated correctly.

Rather than adding premature resilience, sometimes it is better to assume that the real-world likelihood of this actually occurring is very low, and to intentionally design with a slightly brittle adherence to correctness. If it turns out that there are failure conditions we need to tolerate, it will be simple to do so at a later point.

I think I will add the stream id as it will aid in debugging. It does require significant extra lambda-capture though, which is why I avoided it in the first place.

cpp/arcticdb/version/local_versioned_engine.hpp
willdealtry: This is consistent with the pattern that we use elsewhere. Also, following Sausurre, we should recognise the arbitrary, conventional relation of signifier and signified - no-one's actually going to read it.

cpp/arcticdb/version/local_versioned_engine.cpp
willdealtry: So I don't think we can do it with this PR until the other changes are merged, or it will be unacceptably slow

willdealtry:
Probably not going to do this now, although we may revisit in the future

PR #500, opened 2023-06-21
Print the symbol-list-cache-out-of-date-warning in the right place.

mehertz:
Previously it would print after everything had been loaded

See [here](https://github.com/man-group/ArcticDB/pull/489#discussion_r1234213811) for more context.

### DIFF
--- a/cpp/arcticdb/version/symbol_list.cpp
+++ b/cpp/arcticdb/version/symbol_list.cpp
@@ -177,7 +177,6 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
             const folly::Range<KeyVectorItr>& keys) {
         SYMBOL_LIST_RUNTIME_LOG("Loading symbols from symbol list keys");
         bool read_compaction = false;
-        uint64_t uncompacted_keys_found = 0;
         CollectionType symbols{};
         for(const auto& key : keys) {
             if(key.id() == compaction_id) {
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -243,10 +232,25 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
...
+    SymbolList::KeyVector SymbolList::get_all_symbol_list_keys(const std::shared_ptr<StreamSource>& store) {
         std::vector<AtomKey> output;
+        uint64_t uncompacted_keys_found = 0;
         store->iterate_type(KeyType::SYMBOL_LIST, [&] (auto&& key) -> void {
-            output.push_back(to_atom(key));
+            auto atom_key = to_atom(key);
+            if(atom_key.id() != compaction_id) {
+                uncompacted_keys_found++;
+            }
+            if (uncompacted_keys_found == this->max_delta_ * 2 && !warned_expected_slowdown_) {
```
qc00, line 243: Since `warned_expected_slowdown_` is global, you might also want to include "this warning will only appear once" in the warning.

Though I doubt the value of it. I think consistent behaviour is better for this warning.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -243,10 +232,25 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
...
+                log::version().warn(
+                        "Warning - symbol list cache is significantly out of date. "
```
qc00, line 245: The `Warning -` bit seems redundant.

Also, "significantly out of date" may lead to confusion. If the library has very frequent writes, even if the cache has been compacted seconds before, this message can still be triggered.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -120,10 +120,11 @@ bool SymbolList::can_update_symbol_list(const std::shared_ptr<Store>& store,
 
             if (stream_ids.size() == this->max_delta_ * 2 && !warned_expected_slowdown_) {
                 log::version().warn(
-                        "Warning - no compacted symbol list cache found. "
+                        "No compacted symbol list cache found. "
                         "`list_symbols` may take longer than expected. \n\n"
                         "See here for more information: https://docs.arcticdb.io/technical/on_disk_storage/#symbol-list-caching\n\n"
-                        "To resolve, run `list_symbols` through to completion to compact the symbol list cache.");
+                        "To resolve, run `list_symbols` through to completion to compact the symbol list cache.\n"
+                        "Note: This warning will only appear once.\n");
```
qc00, line 127: Minor: Not a fan of line feeds in log lines, and I would definitely avoid having a line feed at the end.

qc00 (approved):

cpp/arcticdb/version/symbol_list.cpp
mehertz: That's fair. I think you are right about he newline at the end. 

That said, as I've got a green build and as I'd like this to be in for the next release + this is a minor suggestion, I'm going to merge this PR.

PR #501, opened 2023-06-21
Add requirements file

alexowens90:
Add a requirements.txt file to simplify pip installing packages


poodlewars (approved):

qc00:
Declaring dependencies in `setup.cfg` format is part of the PEP standard. `requirements.txt` is not.

Considering we already have two (if not three?) additional dependency files to maintain due to Conda, we shouldn't add the risk by introducing yet another one.

Where would `requirements.txt` be used but `setup.cfg` cannot?

Issue #497, opened 2023-06-16
ArcticDB stores the library endpoint, breaking interzone-replication


mehertz:
### Describe the bug

ArcticDB stores the following S3 connectivity details in the library configuration (so saved at configuration time):

* Endpoint used to connect to the bucket
* Bucket name
* Auth credentials for the bucket (if given)

If any of these justifiably change (e.g. endpoints because now we're connecting to a new region!) then even if you pass in a new endpoint in the URL, you'll still access the library via the old endpoint.

### Steps/Code to Reproduce

Create a library in zone A - replicated to Zone B.

Access zone A by connecting with the URL for zone A.
Access zone B by connecting the the URL for zone B.

Reads from zone A will be fast. Reads from zone B will be slow as it'll actually connect to A.

### Expected Results

If specifying the S3 endpoint in the Arctic connection URL, this should be the URL used by default when accessing the library.

### OS, Python Version and ArcticDB Version

All versions are effected

### Backend storage used

S3

mehertz:
Thoughts:

* We should leave the behaviour (i.e. bucket/credential filling) up to each library adaptor. This will require an additional hook in the library adaptor I think to enable the fields to be populated after retrieving from storage.
* For S3, we could remove credentials and the bucket name from the library config and have the library adapter populate from the URI.
  * Library creation should therefore be modified to no longer populate these details in the library config. 
* However - this would no longer enable storing libraries over multiple buckets as there would be no place to store the credentials.
  * This might be worth sacrificing for now - we could add this functionality back later if required.

I was thinking that this would break the multi-bucket functionality which is a problem but I'm coming round to thinking that actually that's fine for now.

willdealtry:
If you want a quick and dirty fix then you could have a flag that tells the Arctic object to use its own endpoint rather than the endpoint from the storage config, that way the limitation would only be applied to particular libraries. It would need to be applied only to primaries otherwise we'd break replication etc internally

willdealtry:
Longer term, rather than having a single storage config for a given storage name, we could (optionally) have a dictionary of region to storage configs, with a default if we don't get a match. Then the local region could either be supplied by the user or inferred from running a `HeadBucket` on the config bucket. This would allow for the selection of an appropriate region without comprimising our ability to do multi-storage operations

willdealtry:
Actually the default could be the current storage config, so if the multi-region flag is set we first try to determine an appropriate local region, then fall back to that one if we fail.

[...]

PR #502, opened 2023-06-21
WIP S3 credentials override

willdealtry:
Fixes https://github.com/man-group/ArcticDB/issues/497

Add a URI parameter that instructs the Arctic object to use its own credentials and endpoint for a storage config, so that we respect AWS interzone replication.

A StorageOverride variant type is exposed from C++ and set (or not) by the storage adapters. This class is a variant that at present has only one member type besides empty (monostate) which is an S3CredentialsOverride.


python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -72,6 +75,35 @@ def test_library_creation_deletion(arctic_client):
...
+    altered_ac = Arctic(moto_s3_uri_incl_bucket)
+    altered_ac._library_adapter = S3LibraryAdapter(wrong_uri)
+    # At this point the library_manager is still correct, so we can write
+    # and retrieve libraries, but the library_adapter has fake credentials
+    altered_ac.create_library("override_endpoint", LibraryOptions())
+    lib = altered_ac["override_endpoint"]
+    s3_storage = _get_s3_storage_config(lib)
+    assert s3_storage.endpoint == "otherhost:17988"
+    assert s3_storage.credential_name == "dog"
+    assert s3_storage.credential_key == "cat"
```
mehertz, line 96: There's no overide here - what does this return overridden values?

willdealtry: They're not override at this point, I've just reset the library manager so that they are different to the endpoint and credentials of the actual storage - this library as-is wouldn't work because it's got bogus endpoint and credentials in its storage config. This is just so that I can check that they're overridden correctly at l102

mehertz: Discussed. I read this wrong - the second run uses the correct settings and so does test overriding.

mehertz (approved):

PR #503, opened 2023-06-21
Use RFC 3986 URL encoding for S3 interactions

poodlewars:
Internal ref: AN-984

This helps to support symbols with special characters in their names on some storages, by using correct URI encoding for S3 interactions rather than legacy compat one in AWS SDK (built for compat with legacy internal Amazon services).

Concretely, symbols containing `:` characters were not working correctly on PURE storage.

At a code level, the impact of the flag is to make us enter this if-condition https://github.com/aws/aws-sdk-cpp/blob/0d214e8330bc1981917a764f5dff45dcfdf382fe/src/aws-cpp-sdk-core/source/http/URI.cpp#L33 and use standard URI encoding.

#### Any other comments?

The test I have added expresses the purpose of the change but actually passed anyway as moto s3 can cope with the bad URI encoding.

### DIFF
--- a/cpp/arcticdb/storage/s3/s3_api.cpp
+++ b/cpp/arcticdb/storage/s3/s3_api.cpp
@@ -21,6 +21,10 @@ namespace arcticdb::storage::s3 {
 S3ApiInstance::S3ApiInstance(Aws::Utils::Logging::LogLevel log_level) :
     log_level_(log_level),
     options_() {
+    // Use correct URI encoding rather than legacy compat one in AWS SDK. PURE S3 needs this to handle symbol names
+    // that have special characters (eg ':').
+    options_.httpOptions.compliantRfc3986Encoding = true;
+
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/storage/s3/s3_api.cpp
```diff
@@ -21,6 +21,10 @@ namespace arcticdb::storage::s3 {
 S3ApiInstance::S3ApiInstance(Aws::Utils::Logging::LogLevel log_level) :
     log_level_(log_level),
     options_() {
+    // Use correct URI encoding rather than legacy compat one in AWS SDK. PURE S3 needs this to handle symbol names
+    // that have special characters (eg ':').
+    options_.httpOptions.compliantRfc3986Encoding = true;
```
alexowens90, line 26: Tested the standard combos?

- New clients can read data written by old clients
- Old clients can read data written by new clients

willdealtry (approved):

PR #509, opened 2023-06-22
Use Man's compatibility layer in the arcticc namespace package

poodlewars:
This will only have an effect within Man Group systems.

In those systems, it installs a handler to redirect Python imports from our old internal structure,
where `arcticdb` was called `arcticc`, to the new structure that this project uses.

#### Reference Issues/PRs

Internal ref AN-994

### DIFF
--- a/python/arcticc/__init__.py
+++ b/python/arcticc/__init__.py
@@ -1,1 +1,9 @@
+# file from ArcticDB
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)
+
+# Use Man's API backwards compatibility layer if it is installed.
+# This will not do anything outside of Man Group systems.
+try:
+    from man.arcticdb.arcticdb_renamer import *
+except ImportError:
+    pass


python/arcticc/__init__.py
```diff
@@ -1 +1,9 @@
+# file from ArcticDB
```
qc00, line 1: This line is redundant

python/arcticc/__init__.py
```diff
@@ -1 +1,9 @@
...
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)
+
+# Use Man's API backwards compatibility layer if it is installed.
+# This will not do anything outside of Man Group systems.
+try:
+    from man.arcticdb.arcticdb_renamer import *
+except ImportError:
+    pass
```
qc00, line 9: I don't like the fact that this copy of the file behaves differently from the others.

There could be strange situations where `man.arcticdb` exists but something it needs doesn't and you only get the exception some of the time.

I wonder if it's possible to omit this file from the external repo and only have it in the internal bridge?

poodlewars: I don't know of a way to omit this file, since the `arcticc` namespace package exists in this repo and also internally.

python/arcticc/__init__.py
poodlewars: I've put a similar line in all the namespace package `__init__.py` files to aid debugging by identifying the source project.

python/arcticc/__init__.py
qc00: When I was playing with namespace packages, I noticed it's possible to have a package without `__init__.py` using the `sys.path` method.

I think it might not conflict with the `pkgutil` method, i.e. you might be able to remove this file.

The question then becomes if the protobuf files are compatible between `arcticc` and `arcticdb`.

python/arcticc/__init__.py
poodlewars: No, but you can still inspect the installed file:

```
import arcticc
arcticc.__file__
```

mehertz (approved):

PR #513, opened 2023-06-23
Remove requirements.txt

mehertz:

File not needed.


poodlewars (approved):
Fine with this as I think you've already adapted our dev container not to rely on this

PR #532, opened 2023-06-30
Refactor underlying segment write scheduling 

willdealtry:
The current version of batch_write which is called by a normal lib.write() to write the data segments collects on the encoded segments so that it can compare them to the dedup map. This in unecessary as the items in the dedup map already exist, and the by-hand batching functionality can be replaced with a folly::window to reduce tail latency. 

Enqueuing the (possible) writes to storage immediately after the compress and check for duplicates results in a 7x speedup in the high_entropy (i.e bandwidth-constrained) test case in test_stress_generated, which is a public benchmark from Wes McKinney's blog.

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -8,7 +8,7 @@
 #pragma once
 
 #include <arcticdb/async/task_scheduler.hpp>
-#include <arcticdb/util/clock.hpp>#
+#include <arcticdb/util/clock.hpp>
 #include <arcticdb/storage/store.hpp>
 #include <arcticdb/entity/performance_tracing.hpp>
 #include <arcticdb/storage/library.hpp>
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -8,7 +8,7 @@
 #pragma once
 
 #include <arcticdb/async/task_scheduler.hpp>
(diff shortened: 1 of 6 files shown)


alexowens90 (approved):

mehertz:
Under what circumstances would we expect to see performance improvements?

Is it only when writing many segments? Would writing one large segment see improved performance following this change?

cpp/arcticdb/async/async_store.hpp
```diff
@@ -404,65 +404,48 @@ class AsyncStore : public Store {
...
+            return async::submit_cpu_task(
+                EncodeAtomTask(std::move(key),
+                               ClockType::nanos_since_epoch(),
+                               std::move(seg),
+                               codec_,
+                               0, //TODO pointless, get rid
+                               encoding_version_));
+            }, write_count);
+
+            for(auto&& encode_fut : encode_futs) {
```
mehertz, line 422: The entire for loop is overly indented which makes it look like it's part of the windowing. I'm not going to make this as a required change comment but I do strongly suggest we fix the fomatting up as otherwise it's unnecessarily confusing.

mehertz (approved):

alexowens90 (approved):

mehertz:
After discussing with @willdealtry we are going to merge this PR despite the failing MacOS test.

This fixes known issues with the write logic that are considered more dangerous than what we believe the failing MacOS test is indicating.

PR #539, opened 2023-07-03
Extra update checks

willdealtry:
Add some extra checks to the update method to catch strange things happening due to out-of-order data. Unaffected + affected keys should equal the number of keys in the index segment reader. Also, the maximum number of output keys should be less than the unaffected keys + 2 times the affected keys (because they may be overlapping at both their beginning and end) + the number of new keys.

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -429,7 +429,7 @@ public:
                                encoding_version_));
         }, write_count);
 
-        for (folly::Future<VariantKey>&& encode_fut : encode_futs) {
+        for (folly::Future<storage::KeySegmentPair>& encode_fut : encode_futs) {
             futs.emplace_back(
                 std::move(encode_fut).thenValue([de_dup_map](auto &&ks) -> std::pair<VariantKey,
                                                                                      std::optional<Segment>> {
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -429,7 +429,7 @@ public:
                                encoding_version_));
         }, write_count);
 
-        for (auto &&encode_fut : encode_futs) {
+        for (folly::Future<VariantKey>&& encode_fut : encode_futs) {
             futs.emplace_back(
                 std::move(encode_fut).thenValue([de_dup_map](auto &&ks) -> std::pair<VariantKey,
                                                                                      std::optional<Segment>> {
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -330,6 +330,9 @@ VersionedItem update_impl(
                         std::end(affected_keys),
                         std::back_inserter(unaffected_keys));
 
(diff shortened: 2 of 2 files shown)


cpp/arcticdb/async/async_store.hpp
```diff
@@ -404,65 +404,47 @@ class AsyncStore : public Store {
...
-                                          }
-                                      }).thenValue([key_segs = key_segs, key_seg_mutex = key_seg_mutex](auto &&pk) {
-                    auto pair_key_write = std::forward<decltype(pk)>(pk);
-                    if (pair_key_write.second) {
-                        std::lock_guard kl{*key_seg_mutex};
-                        ARCTICDB_DEBUG(log::version(), "Enqueuing {} for write", pair_key_write.first.atom_key());
-                        key_segs->emplace_back(std::move(pair_key_write.first));
-                    }
-                }));
+        auto encode_futs = folly::window(key_segments, [*this] (auto&& ks) {
```
mehertz, line 411: This is more than just an addition of an `assert` 😆 

This is also in `batch_write` - how does this relate?

cpp/arcticdb/async/async_store.hpp
```diff
@@ -404,65 +404,47 @@ class AsyncStore : public Store {
         const std::shared_ptr<DeDupMap> &de_dup_map,
         const BatchWriteArgs &args) override {
```
alexowens90, line 405: I think we should get rid of `BatchWriteArgs`. This method is only called in one place, which doesn't provide the argument:
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/pipeline/write_frame.cpp#L108
It would make more sense to me to use `follw::window` for both the encoding and writing tasks, and to use `cpu_thread_count()` and `io_thread_count()` respectively as the window sizes. That should ensure maximum performance, while at the same time keeping the number of tasks in the queues short if there are multiple callers of this function at the same time.

alexowens90 (changes requested):

mehertz (changes requested):

cpp/arcticdb/async/async_store.hpp
```diff
@@ -404,65 +404,47 @@ class AsyncStore : public Store {
...
+                               0, //TODO pointless, get rid
+                               encoding_version_));
+            }, write_count);
+
+            for(auto&& encode_fut : encode_futs) {
+                futs.emplace_back(
+                    std::move(encode_fut).thenValue([de_dup_map](auto &&ks) -> std::pair<VariantKey,
+                                                                                         std::optional<Segment>> {
+                            auto key_seg = std::forward<decltype(ks)>(ks);
+                            if (auto de_dup_key = de_dup_map->get_key_if_present(key_seg.atom_key())) {
```
mehertz, line 427: Previously:

```
auto de_dup_key = de_dup_map ? de_dup_map->get_key_if_present(key_seg.atom_key()) : std::nullopt;
```

Now:

`auto de_dup_key = de_dup_map->get_key_if_present(key_seg.atom_key())`

That's a behavioural change, no? What if `de_dup_map` is not set?

mehertz (changes requested):

alexowens90 (approved):

PR #542, opened 2023-07-03
Fix compilation errors issued by clang-16.

vasil-pashov:
ArcticDB fails to build with clang-16 + lld. The changes are minor fixes for the compilation errors.
* Delete definitions of unused variables (in lambda captures and globals)
* Fix the way libatomic is linked to the tests

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -766,7 +766,6 @@ if(${TEST})
     else()
         set(COMMON_PUBLIC_TEST_LIBRARIES
                 arcticdb_core_static
-                libatomic.a  # Otherwise gtest_discover_tests cannot run the test if libatomic is not in a standard location
                 GTest::gtest
                 GTest::gmock
                 Python::Python # + pybind11::pybind11 (transitively included) = pybind11::embed, but latter is sometimes not found...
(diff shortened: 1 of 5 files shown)


cpp/arcticdb/CMakeLists.txt
```diff
@@ -766,7 +766,7 @@ if(${TEST})
     else()
         set(COMMON_PUBLIC_TEST_LIBRARIES
                 arcticdb_core_static
-                libatomic.a  # Otherwise gtest_discover_tests cannot run the test if libatomic is not in a standard location
+                atomic
```
mehertz, line 769: This is for linking surely? Why is clang affected?

vasil-pashov: The compilation error was because libatomic.a was not built with `-fPIC` so the linker failed (only with clang, though). There are two things I'm not quite sure about:
1. Why did we want to specify the library down to exact archive name? Usually in CMake you have specify only the name of the library (as I did in the PR) and it will handle the specifics of finding the library and linking (e.g. adding the `lib` prefix, potentially adding file extensions, etc.)
2. Why is it working now? My best guess is that it's dynamically linking, or CMake is doing some behind the scenes magic to find a suitable library to link against.

The canonical thing to do is to link against `atomic` not `libatomic.a`.

mehertz: If we do this we are dynamically linking. Maybe that's fine now we're using `auditwheel` to slurp up all the libraries. It wasn't before.

OK, I think this is a good change on reflection.

mehertz (approved):

PR #544, opened 2023-07-04
Enhancement/531/modify batch read behaviour if a version doesnt exist

### LINKED ISSUE #531: When querying multiple symbols from arcticdb using read_batch, the whole query fails if one of the symbols is invalid
Lets assume that I have saved three symbols s1, s2, and s3 to arcticdb. If I try to "read_batch" these symbols using the following it will fail
```
vesioned_items = lib.read_batch(["s1", "s2", "bogus"])
```
I get an exception: 
InternalException: E_ASSERTION_FAILURE Version not found for stream

There are two issues with this: (1) the exception does not tell me which symbol is invalid
(2) Ideally, we should get two valid versioned items with one error item.

A list of versioned items where first two items are valid and we can read "data" field from them. Third item may either throw if I try to access "data" or may indicate the error in some other way.

Because the exception is not very explicit, we have to either try each symbol separately or do a list_symbols () followed by some python gymnastics to eliminate invalid symbol names.

**Additional context**
We end up with situations where a user may query for several symbols without realizing some of them are incorrect or just fat-fingering some names. We should be more graceful and only throw exceptions for symbols that are incorrect.

alexowens90:
Closes #531 

The previous behaviour was to throw a `NoDataFoundException` if any of the requested symbols/versions did not exist. The new behaviour is to return a `None` in the list for any symbols/versions that do not exist.

The existing behaviour is maintained on `NativeVersionStore` for backwards compatibility.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -120,7 +120,7 @@ FrameAndDescriptor LocalVersionedEngine::read_column_stats_internal(
     return read_column_stats_impl(store(), versioned_item);
 }
 
-std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_column_stats_version_internal(
+ReadVersionOutput LocalVersionedEngine::read_column_stats_version_internal(
     const StreamId& stream_id,
     const VersionQuery& version_query) {
     auto versioned_item = get_version_to_read(stream_id, version_query);
@@ -130,7 +130,7 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_column_s
             stream_id
             );
     auto frame_and_descriptor = read_column_stats_internal(versioned_item.value());
-    return std::make_pair(versioned_item.value(), std::move(frame_and_descriptor));
+    return ReadVersionOutput{std::move(versioned_item.value()), std::move(frame_and_descriptor)};
 }
 
 ColumnStats LocalVersionedEngine::get_column_stats_info_internal(
@@ -315,7 +315,7 @@ IndexRange LocalVersionedEngine::get_index_range(
     return index::get_index_segment_range(version.value().key_, store());
 }
 
-std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_dataframe_version_internal(
+std::optional<ReadVersionOutput> LocalVersionedEngine::read_dataframe_version_internal(
     const StreamId &stream_id,
     const VersionQuery& version_query,
     ReadQuery& read_query,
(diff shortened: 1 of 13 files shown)


GemDot (approved):
As temporal solution for the temp_batch_read_internal_direct this change makes sense to me, so I approve. But as discussed we will need to do very similar changes inside batch_get_versions_async functions, which will be used by most of the batch methods, including read_batch. I also think that this PR covers all the testing regardless of which implementation we will use, which I think is great.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1008,33 +1015,44 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
...
-            return store->read_sync(maybe_index_key.value());
-        }).thenValue([store = store(), read_query = std::move(read_query), read_options](auto &&key_segment_pair) {
-            auto [index_key, index_segment] = std::move(key_segment_pair);
-            return async_read_direct(store,
-                                     std::move(index_key),
-                                     std::move(index_segment),
-                                     read_query,
-                                     std::make_shared<BufferHolder>(),
-                                     read_options);
+        output.emplace_back(std::move(*version).thenValue([store = store(), read_options](auto maybe_index_key) -> std::optional<ReadKeyOutput> {
```
qc00, line 1031: Minor: avoid copying the entire `read_options` for one boolean.

You can also choose between two different lambdas based on the option instead of passing the option value in for more efficiency.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1008,33 +1015,44 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
...
+            missing_data::check<ErrorCode::E_NO_SUCH_VERSION>(
+                    !(*read_options.batch_throw_on_missing_version_) || maybe_index_key.has_value(),
+                    "Version not found for symbol");
+            return maybe_index_key.has_value() ? std::make_optional<ReadKeyOutput>(store->read_sync(maybe_index_key.value())) : std::nullopt;
+        }).thenValue([store = store(), read_query = std::move(read_query), read_options](auto &&opt_key_segment_pair) {
```
qc00, line 1036: I don't understand the value of putting this in a separate lambda?

cpp/arcticdb/version/version_store_api.hpp
```diff
@@ -334,16 +334,16 @@ class PythonVersionStore : public LocalVersionedEngine {
...
-            item,
-            PythonOutputFrame{fd.frame_, fd.buffers_},
-            fd.desc_.proto().normalization(),
-            fd.desc_.proto().user_meta(),
-            fd.desc_.proto().multi_key_meta(),
+            read_version_output.versioned_item_,
+            PythonOutputFrame{read_version_output.frame_and_descriptor_.frame_, read_version_output.frame_and_descriptor_.buffers_},
+            read_version_output.frame_and_descriptor_.desc_.proto().normalization(),
+            read_version_output.frame_and_descriptor_.desc_.proto().user_meta(),
+            read_version_output.frame_and_descriptor_.desc_.proto().multi_key_meta(),
```
qc00, line 346: I would argue readability decreased due to high verbosity

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -327,15 +327,22 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
             log::version().warn("No index:  Key not found for {}, will attempt to use incomplete segments.", stream_id);
             identifier = stream_id;
         }
-        else
-            throw storage::NoDataFoundException(fmt::format("read_dataframe_version: version not found for symbol '{}'", stream_id));
+        else {
+            if (!read_options.batch_throw_on_missing_version_.has_value() || *read_options.batch_throw_on_missing_version_) {
```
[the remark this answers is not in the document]
alexowens90: `value_or`, but agreed, changed

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: `read_options` is needed by `async_read_direct`, not just for one bool

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Previously it didn't make sense, but I now use `read` instead of `read_sync` to get the index key

Issue #473, opened 2023-06-08
Warn if no credentials are in effect for S3


qc00:
**Is your feature request related to a problem? Please describe.**
When setting S3 credentials via environment variables, `aws_auth=true` also has to be set, but there's no explicit information about this.

**Describe the solution you'd like**
Possibilities:
* Require the URI to have a credential (aws_auth=true or credentials in URI)
* Print a warning if `AWS_*` environment variables are set but `aws_auth` is not.

mehertz:
I've been thinking about this.

> Print a warning if AWS_* environment variables are set but aws_auth is not.

They might not be using AWS env vars. I think this is quite hard to do actually.

Maybe we should focus our efforts on standardising the URI instead?

qc00:
Yeah, I don't want to tightly couple to any AWSSDK-specific stuff, either.

Maybe something like: if no explicit credentials or `aws_auth`, make a test connection to the bucket.

> Maybe we should focus our efforts on standardising the URI instead?

I am not sure that would necessarily help. My understanding of the user's issue is: the library appears to "open" successfully but the first call fails with an opaque error.

mehertz:
> if no explicit credentials or aws_auth, make a test connection to the bucket.

I think that's a good idea. Maybe just HEAD the bucket? Hope that would give a non-200 if you don't have any creds.

mehertz:
We need to be careful not to let this overload storages though. Would be adding a new request per object creation.

Issue #447, opened 2023-05-31
Use specific exception type for certain kinds of storage failures


qc00:
Per [this comment](https://github.com/man-group/ArcticDB/pull/393#discussion_r1200661617), we should make the storage throw our event-specific exceptions (e.g. Key/symbol/segment not found) instead of generic or storage framework's exception types.

We can then tighten the exception handling in the downstream code.

Related to #368, #365

qc00:
The less inconsistent sets of exceptions thrown **after** my PR is merged:

Legend:

- `d`=`DuplicateKeyException`
- `K`=`KeyNotFoundException` at the end, k=`KeyNotFoundException` on first missing key
- `p`=`PermissionException`/`PermissionSpecificException`
- `u`=unexpected exception (store-specific), `r`=retryable exception (store-specific)
- `>`=catching the inner exception and rethrowing as

Store\Op | write | update | read | exists | remove | iterate
-- | -- | -- | -- | -- | -- | --
LMDB | ud | ku d | K | - | K | -
Mongo | u | kur | ku^ | u | kur>K | -
S3 | u rp | &nbsp;urp | Kurp | urp | K | urp
Memory | &nbsp;d | k | k | - | k | -

^ mongo wanted to do ">K" but failed to catch the inner exceptions correctly…

qc00:
## Long form comparison:

<table style="border-collapse: collapse; width: 100.022%; height: 1080px;" border="1"><colgroup><col style="width: 6.8433%;"><col style="width: 13.1257%;"><col style="width: 22.7736%;"><col style="width: 17.9496%;"><col style="width: 6.50674%;"><col style="width: 32.8703%;"></colgroup>
<tbody>
<tr>
<td>Storage</td>
<td>Function</td>
<td>Throws</td>
<td>For situation</td>
<td>Key in error msg</td>
<td>Changed in this PR</td>
</tr>
<tr>
<td rowspan="8">LMDB</td>
<td>Common to all</td>
<td>lmdb::error</td>
<td>transaction/low-level failure</td>
<td>None</td>
<td>No</td>
</tr>
<tr>
<td rowspan="2">do_write<br><br></td>
<td>DuplicateKeyException</td>
<td>overwriting&nbsp;AtomKey</td>
<td>One</td>
<td>No</td>
</tr>
<tr>
<td>std::runtime_error</td>
<td>other write failures</td>
<td>One</td>
<td>No</td>
</tr>
<tr>
<td>do_update</td>
[...]

PR #551, opened 2023-07-04
#473 Warn if bucket cannot be accessed

qc00:
+ warn if AWS auth is being used and machine identity is disabled
+ Refactor S3 fixtures so various errors can be simulated
* Removed pytest-server-fixtures dependency which is a bit old and we only use function


cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -32,6 +34,48 @@ std::streamsize S3StreamBuffer::xsputn(const char_type* s, std::streamsize n) {
...
+            "EC2 metadata endpoint did not respond in time so was bypassed":
+            "AWS_EC2_METADATA_DISABLED environment variable is set to true";
+        log::storage().warn("{}. This means machine identity authentication will not work.", who_disabled);
+    }
+}
+
+bool S3Storage::check_creds_and_bucket() {
+    using namespace Aws::S3;
+    // We cannot easily change the timeouts of the s3_client_, so use async:
+    auto future = s3_client_.HeadBucketCallable(Model::HeadBucketRequest().WithBucket(bucket_name_.c_str()));
```
mehertz, line 50: Do we need to respect `path_prefix` here? What if we have access to the bucket but don't have permissions for the root path.

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -32,6 +34,48 @@ std::streamsize S3StreamBuffer::xsputn(const char_type* s, std::streamsize n) {
 }
 }
 
+void check_ec2_metadata_endpoint(const S3ApiInstance& s3_api) {
+    auto disabled_env = Aws::Environment::GetEnv("AWS_EC2_METADATA_DISABLED");
+    if (Aws::Utils::StringUtils::ToLower(disabled_env.c_str()) == "true") {
+        const char* who_disabled = s3_api.is_ec2_metadata_disabled_by_us() ?
+            "EC2 metadata endpoint did not respond in time so was bypassed":
```
mehertz, line 41: wrong way round?

python/tests/util/storage_fixtures.py
qc00, line 1: This is borrowed from the work for #447. It adds various ways to break a storage

cpp/arcticdb/storage/s3/s3_storage.cpp
qc00: Recall that S3 doesn't actually have folders, so we can only reliably `HEAD` bucket or a full object name.

mehertz: That is true. If this is advisory I think it's OK.

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -100,6 +100,13 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
                                             return none.cast<py::object>();
                                        });
         })
+        .def("check_primary_storage_is_accessible", [](Library& library) {
+            std::optional<bool> out;
+            library.storage_specific([&out](storage::s3::S3Storage& s3) {
+                out = s3.check_creds_and_bucket();
+            });
+            return out;
+        }, "Currently only implemented by S3 storage. Calling on other storage types does nothing.")
```
mehertz, line 109: I think we should document what it does as well as what it's implemented for. Something like:

`For S3 storages will _HEAD_ the bucket to test connectivity. Storage types other than S3 are not implemented - calling this will return _None_`.

Also, should we raise a `NotImplementedException` or something?

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -42,6 +86,7 @@ S3Storage::S3Storage(const LibraryPath &library_path, OpenMode mode, const Confi
 
     if (creds.GetAWSAccessKeyId() == USE_AWS_CRED_PROVIDERS_TOKEN && creds.GetAWSSecretKey() == USE_AWS_CRED_PROVIDERS_TOKEN){
         ARCTICDB_RUNTIME_DEBUG(log::storage(), "Using AWS auth mechanisms");
+        check_ec2_metadata_endpoint(*s3_api_);
```
mehertz, line 89: Why does this change necessitate this line?

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -32,6 +34,48 @@ std::streamsize S3StreamBuffer::xsputn(const char_type* s, std::streamsize n) {
...
+            auto& error = outcome.GetError();
+
+#define BUCKET_LOG(level, msg, ...) log::storage().level(msg "\nHTTP Status: {}. Server response: {}", \
+            ## __VA_ARGS__, int(error.GetResponseCode()), error.GetMessage().c_str()); break
+
+            // HEAD request can't return the error details, so can't use the more detailed error codes.
+            switch (error.GetResponseCode()) {
+            case Aws::Http::HttpResponseCode::UNAUTHORIZED:
+            case Aws::Http::HttpResponseCode::FORBIDDEN:
+                BUCKET_LOG(warn, "Unable to access bucket. Subsequent operations may fail.");
```
mehertz, line 64: Mention that we can't access the bucket because we don't have permission in the error message

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -32,6 +34,48 @@ std::streamsize S3StreamBuffer::xsputn(const char_type* s, std::streamsize n) {
...
+            "AWS_EC2_METADATA_DISABLED environment variable is set to true";
+        log::storage().warn("{}. This means machine identity authentication will not work.", who_disabled);
+    }
+}
+
+bool S3Storage::check_creds_and_bucket() {
+    using namespace Aws::S3;
+    // We cannot easily change the timeouts of the s3_client_, so use async:
+    auto future = s3_client_.HeadBucketCallable(Model::HeadBucketRequest().WithBucket(bucket_name_.c_str()));
+    auto wait = std::chrono::milliseconds(ConfigsMap::instance()->get_int("S3Storage.CheckBucketMaxWait", 1000));
```
mehertz, line 51: Document https://github.com/man-group/ArcticDB/blob/master/docs/mkdocs/docs/runtime_config.md 

I know there are lots of other that aren't documented, but going forwards we should document these.

cpp/arcticdb/storage/s3/s3_storage.hpp
```diff
@@ -232,6 +235,12 @@ inline Aws::Client::ClientConfiguration get_proxy_config(Aws::Http::Scheme endpo
     }
 }
 
+namespace {
+inline uint32_t fallback(uint32_t protobuf_config, const std::string& label, uint32_t default_val) {
```
mehertz, line 239: `protobuf_access_with_default`?

python/arcticdb/arctic.py
```diff
@@ -98,6 +98,8 @@ def __init__(self, uri: str, encoding_version: EncodingVersion = EncodingVersion
         self._library_manager = LibraryManager(self._library_adapter.config_library)
         self._uri = uri
 
+        self._library_adapter.check_storage_is_accessible()
```
mehertz, line 101: I think there should be a parameter in the URI that optionally disables this behaviour

mehertz (changes requested):

cpp/arcticdb/storage/s3/s3_storage.cpp
willdealtry: Is it true? I thought you absolutely could ACL specific directories in AWS S3, just not in the S3-like storages that we have (although it's on the roadmap for at least one of them)

cpp/arcticdb/storage/s3/s3_storage.cpp
willdealtry: Pelase don't use the preprocessor for this kind of stuff, if it's actually used enough to require something specific, then use a function. The only really legitimate use for the preprocessor is to toggle bits of code on and off based on compiler flags - this is regular code so it should use regular C++.

willdealtry (changes requested):

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -32,6 +34,48 @@ std::streamsize S3StreamBuffer::xsputn(const char_type* s, std::streamsize n) {
...
+#define BUCKET_LOG(level, msg, ...) log::storage().level(msg "\nHTTP Status: {}. Server response: {}", \
+            ## __VA_ARGS__, int(error.GetResponseCode()), error.GetMessage().c_str()); break
+
+            // HEAD request can't return the error details, so can't use the more detailed error codes.
+            switch (error.GetResponseCode()) {
+            case Aws::Http::HttpResponseCode::UNAUTHORIZED:
+            case Aws::Http::HttpResponseCode::FORBIDDEN:
+                BUCKET_LOG(warn, "Unable to access bucket. Subsequent operations may fail.");
+            case Aws::Http::HttpResponseCode::NOT_FOUND:
+                BUCKET_LOG(error, "The specified bucket {} does not exist.", bucket_name_);
```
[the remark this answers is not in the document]
qc00: Will raise

cpp/arcticdb/storage/s3/s3_storage.cpp
qc00: What are you referring to by "this change"?

cpp/arcticdb/storage/s3/s3_storage.cpp
willdealtry: Honestly for three usages I would just write it out. I'd rather the code was readable and robust rather than pretty, and VA_ARGS is a portability headache

willdealtry (dismissed):

cpp/arcticdb/storage/s3/s3_storage.cpp
mehertz: This PR. Checking the metadata endpoint seems unrelated?

python/arcticdb/arctic.py
mehertz: We may warn incorrectly if we can't `HEAD` the root path but can access other paths.

I also don't trust all S3 implementations to implement this correctly and uniformly - it wouldn't shock me if some implementations 404 HeadBucket operations but other ops work.

As a result I really think we should merge this with way to disable.

cpp/arcticdb/storage/s3/s3_storage.cpp
qc00: The metadata endpoint is used to fetch the machine identity, which is one of the AWS authentication methods.

We added logic to disable the metadata endpoint if it is not reachable because it's causing that delay with connection creation internally.

However, if the user is running in AWS and uses machine identity, then s/he might want to know we have disabled it. Otherwise, the authentication failure might become a mystery.

cpp/arcticdb/storage/s3/s3_storage.cpp
mehertz: I think this is the best we can do. The prefix isn't a directory and will give another error if it's yet to be created presumably.

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -32,6 +34,48 @@ std::streamsize S3StreamBuffer::xsputn(const char_type* s, std::streamsize n) {
 }
 }
 
+void check_ec2_metadata_endpoint(const S3ApiInstance& s3_api) {
+    auto disabled_env = Aws::Environment::GetEnv("AWS_EC2_METADATA_DISABLED");
+    if (Aws::Utils::StringUtils::ToLower(disabled_env.c_str()) == "true") {
+        const char* who_disabled = s3_api.is_ec2_metadata_disabled_by_us() ?
+            "EC2 metadata endpoint did not respond in time so was bypassed":
+            "AWS_EC2_METADATA_DISABLED environment variable is set to true";
```
mehertz, line 42: Is this the wrong way round?

mehertz (approved):

cpp/arcticdb/storage/s3/s3_storage.cpp
qc00: It is the correct way around.

When `is_ec2_metadata_disabled_by_us()` is true, how we disabled it (by setting `AWS_EC2_METADATA_DISABLED` internally) is an implementation detail that we can change at any time, so we only need to let the user know the relevant downside.

When the user disables it, s/he should be familiar with the environment name already, so we can be explicit.

mehertz:
Resolves #473

PR #553, opened 2023-07-05
Add environment variable controlling wheter AWS S3 should use SSL

### LINKED ISSUE #522: S3 connectivity does not verify SSL certificate validity
### Describe the bug

ArcticDB does not verify the certificate of SSL certificates.

You can connect to endpoints with expired certificates.

### Steps/Code to Reproduce

Connect to an S3 server with SSL enabled using an expired certificate.

### Expected Results

Certificate validity to be verified. 

All versions. 

### Backend storage used

S3

### Additional Context

_No response_

vasil-pashov:
#### Reference Issues/PRs
Fixes #522
Add environment variable controlling wheter AWS S3 should use SSL. If the environment is not set it will use whatever is written in the config protobuf.

### DIFF
--- a/cpp/arcticdb/storage/s3/s3_storage.hpp
+++ b/cpp/arcticdb/storage/s3/s3_storage.hpp
@@ -245,7 +245,9 @@ auto get_s3_config(const ConfigType& conf) {
     auto endpoint = conf.endpoint();
     util::check_arg(!endpoint.empty(), "S3 Endpoint must be specified");
     client_configuration.endpointOverride = endpoint;
-    client_configuration.verifySSL = false;
+    const bool use_ssl = ConfigsMap::instance()->get_int("S3Storage.UseSSL", conf.ssl());
+    log::storage().debug("Use ssl: {}", use_ssl);
+    client_configuration.verifySSL = use_ssl;
     client_configuration.maxConnections = conf.max_connections() == 0 ?
             ConfigsMap::instance()->get_int("VersionStore.NumIOThreads", 16) :
             conf.max_connections();
--- a/cpp/arcticdb/storage/s3/s3_storage.hpp
+++ b/cpp/arcticdb/storage/s3/s3_storage.hpp
@@ -245,9 +245,9 @@ auto get_s3_config(const ConfigType& conf) {
     auto endpoint = conf.endpoint();
     util::check_arg(!endpoint.empty(), "S3 Endpoint must be specified");
     client_configuration.endpointOverride = endpoint;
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/storage/s3/s3_storage.hpp
```diff
@@ -245,7 +245,9 @@ auto get_s3_config(const ConfigType& conf) {
     auto endpoint = conf.endpoint();
     util::check_arg(!endpoint.empty(), "S3 Endpoint must be specified");
     client_configuration.endpointOverride = endpoint;
-    client_configuration.verifySSL = false;
+    const bool use_ssl = ConfigsMap::instance()->get_int("S3Storage.UseSSL", conf.ssl());
```
mehertz, line 248: `VerifySSL` not `UseSSL`.

SSL will be used - we're just not verifying the certificate. Different things.

cpp/arcticdb/storage/s3/s3_storage.hpp
```diff
@@ -245,7 +245,9 @@ auto get_s3_config(const ConfigType& conf) {
...
+    log::storage().debug("Use ssl: {}", use_ssl);
```
mehertz, line 249: `ARCTICDB_RUNTIME_DEBUG`

cpp/arcticdb/storage/s3/s3_storage.hpp
mehertz, line 248: Will this cast the int to a bool?! What if I set a value that's neither 0 or 1?

mehertz (changes requested):

cpp/arcticdb/storage/s3/s3_storage.hpp
vasil-pashov: I think it's casting, if anything another number is passed it will be set to true. I didn't find a flag option. There were get_int, get_double, get_string. Should I make get bool?

vasil-pashov: It is indeed casting. But the whole thin is implemented via macro: https://github.com/man-group/ArcticDB/blob/da1d286e8a6840a93218ba3bede4e9fc661cb59d/cpp/arcticdb/util/configs_map.hpp#L34 even if I simply add bool it will cast it the same way. So the questions are: 

1. What do we want the functionality to be if neither 0 nor 1 is passed?
2. Do we want to add something as "check_flag" which only checks if the map contains the key without reading the value?

Also I really don't like the macro design of this class. IMO it can be replaced by template member functions.

mehertz (approved):

PR #559, opened 2023-07-05
Multi segment

willdealtry:
WIP Coalesce small segments into larger physical blocks on storage.


maxim-morozov:
Close for now, we have in the log and we can look at it when we resume working on it.

PR #571, opened 2023-07-10
Enhancement 569: Suppress Pandas warning about nanoseconds being disc…

### LINKED ISSUE #569: Suppress the warning about nanoseconds being ignored.
When querying arcticdb we often see the following warning:

```
arcticdb/version_store/_store.py:2252: UserWarning: Discarding nonzero nanoseconds in conversion
  _from_tz_timestamp(max_ts, "UTC").astimezone(pytz.timezone(tz)),
```
There is very little a user can do about it but it may spook a naive user.

There should be a way to suppress this warning.

None

alexowens90:
Closes #569

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -258,7 +258,7 @@ def _to_tz_timestamp(dt):
 
 def _from_tz_timestamp(ts, tz):
     # type: (int, Optional[str])->(datetime.datetime)
-    return pd.Timestamp(ts).tz_localize(tz).to_pydatetime()
+    return pd.Timestamp(ts).tz_localize(tz).to_pydatetime(warn=False)
 
 
 _range_index_props_are_public = hasattr(RangeIndex, "start")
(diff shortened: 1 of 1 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -258,7 +258,7 @@ def _to_tz_timestamp(dt):
 
 def _from_tz_timestamp(ts, tz):
     # type: (int, Optional[str])->(datetime.datetime)
-    return pd.Timestamp(ts).tz_localize(tz).to_pydatetime()
+    return pd.Timestamp(ts).tz_localize(tz).to_pydatetime(warn=False)
```
mehertz, line 261: https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html

I don't see that argument documented there?

alexowens90: Wrong method:
https://pandas.pydata.org/docs/reference/api/pandas.Timestamp.to_pydatetime.html

mehertz: 🙄 

Cool, thanks

mehertz (approved):

mehertz:
> LGTM, @alexowens90.
> 
> Should we add a non-regression test for #569?

It's a warning log, which would be pretty tricky to test. Have to redirect and capture logging output which is doable but a little tricky. I'm okay with merging this as is as a result, but I agree in a perfect world we would have a test for this that captures the log output and asserts that there is no warning.

jjerphan:
Ah, I should have mentioned that [`pytest` allows asserting the presence or absence of warnings](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#warns).

Anyway, we can merge it.

Issue #576, opened 2023-07-10
E_ASSERTION_FAILURE should be a catchable exception instead of a log message.


alokwatve:
**Is your feature request related to a problem? Please describe.**
If the symbol client is querying for is not available in arcticdb, we current see an error message:
"E_ASSERTION_FAILURE Version not found for stream"
While this is useful, it is not a catchable exception, so we cannot take some reasonable action on it. One possible action could be to catch the exception and print it in more user friendly way like "Symbol you are requesting is not available"

**Describe the solution you'd like**
Make this a catchable exception instead of just a log message.

**Describe alternatives you've considered**
None/

**Additional context**
n/a

mehertz:
With the release of 1.6.1rc0 is this still an issue?

PR #580, opened 2023-07-11
Enhancement/531/modify batch read behaviour if a version doesnt exist

mehertz:
Closes https://github.com/man-group/ArcticDB/issues/531

The previous behaviour was to throw a NoDataFoundException if any of the requested symbols/versions did not exist. The new behaviour is to return a None in the list for any symbols/versions that do not exist.

The existing behaviour is maintained on NativeVersionStore for backwards compatibility.

Note, this PR reapplies https://github.com/man-group/ArcticDB/pull/544 which was reverted to target 1.6.0.


alexowens90 (approved):

mehertz:
This will fix #576 as well

Issue #520, opened 2023-06-27
Segmentation fault (core dumped) with lmdb (read_batch)


prophetss:
### Steps/Code to Reproduce

Code1:
```
import pandas as pd
from arcticdb import Arctic

adb = Arctic(r"lmdb:///home/admin/arcticdb")
adb.create_library("test_lib")
lib = adb["test_lib"]

for i in range(1000):
    df = pd.DataFrame({"a": list(range(i, i + 1000))}, index=list(range(i, i + 1000)))
    lib.append(str(i), df, prune_previous_versions=True)

vers = lib.read_batch([str(i) for i in range(1000)]) # run ok
```
Code2:
```
from arcticdb import Arctic

adb = Arctic(r"lmdb:///home/admin/arcticdb")
lib = adb["test_lib"]

vers = lib.read_batch([str(i) for i in range(1000)]) # crash
```

Output:
```
*** Error in `/home/admin/anaconda3/bin/python': double free or corruption (fasttop): 0x00007ed50800ccb0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81329)[0x7f9dd5324329]
/home/admin/anaconda3/lib/python3.9/site-packages/arcticdb_ext.cpython-39-x86_64-linux-gnu.so(+0x15bb558)[0x7f9dcde26558]
/home/admin/anaconda3/lib/python3.9/site-packages/arcticdb_ext.cpython-39-x86_64-linux-gnu.so(+0x2a4463)[0x7f9dccb0f463]
/home/admin/anaconda3/lib/python3.9/site-packages/arcticdb_ext.cpython-39-x86_64-linux-gnu.so(+0x27291d)[0x7f9dccadd91d]
/home/admin/anaconda3/lib/python3.9/site-packages/arcticdb_ext.cpython-39-x86_64-linux-gnu.so(+0x27a7dd)[0x7f9dccae57dd]
/home/admin/anaconda3/lib/python3.9/site-packages/arcticdb_ext.cpython-39-x86_64-linux-gnu.so(+0x27a9b5)[0x7f9dccae59b5]
/home/admin/anaconda3/lib/python3.9/site-packages/arcticdb_ext.cpython-39-x86_64-linux-gnu.so(+0x270a62)[0x7f9dccadba62]
/home/admin/anaconda3/lib/python3.9/site-packages/arcticdb_ext.cpython-39-x86_64-linux-gnu.so(+0x40d4c1)[0x7f9dccc784c1]
[...]

PR #585, opened 2023-07-12
Only open library once in the interpreter lifetime

poodlewars:
#### Reference Issues/PRs

This addresses the issue reported here https://github.com/man-group/ArcticDB/issues/520#issuecomment-1611455485 in issue #520, but doesn't solve the OP's issue.

An LMDB env should never be opened twice in the same process. I posted some LMDB usage guidelines on issue #520. However we had many issues with our use of it.

- At library creation, a `LibraryIndex` object creates the storages, which creates LMDB envs for both the config library and the actual library, and holds those envs open.
- On calling `Arctic#get_library` a `LibraryManager` object then recreates the storages, regardless of whether they were already open or not. This is why the `get_library` call affected whether the user saw issues or not.

These break the rule of LMDB:

    _Do not have open an LMDB database twice in the same process at the same time. Not even from a plain open() call - close()ing it breaks flock() advisory locking._

and presumably leading to corruption in the memory-mapped file that LMDB holds open.

#### Alternatives

I also tried making the LMDB env global - https://github.com/man-group/ArcticDB/pull/579 - which I think would work too. However this approach seems cleaner and simpler and lets us manage the env with the RAII idioms that LMDB++ is designed for.

### DIFF
--- a/python/arcticdb/arctic.py
+++ b/python/arcticdb/arctic.py
@@ -212,8 +212,7 @@ class Arctic:
         >>> arctic.create_library('test.library')
         >>> my_library = arctic['test.library']
         """
-        already_open = self._open_libraries.get(name)
-        if already_open or self._library_manager.has_library(name):
+        if name in self._open_libraries or self._library_manager.has_library(name):
             raise ValueError(f"Library [{name}] already exists.")
 
         if library_options is None:
--- a/python/arcticdb/arctic.py
+++ b/python/arcticdb/arctic.py
@@ -5,7 +5,6 @@ Use of this software is governed by the Business Source License 1.1 included in 
 
 As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
 """
-import enum
 from typing import List, Optional
 
 from arcticdb.options import LibraryOptions
@@ -149,8 +148,13 @@ class Arctic:
         self._library_adapter = _cls(uri, self._encoding_version)
         self._library_manager = LibraryManager(self._library_adapter.config_library)
         self._uri = uri
+        self._open_libraries = dict()
 
     def __getitem__(self, name: str) -> Library:
+        already_open = self._open_libraries.get(name)
+        if already_open:
+            return already_open
+
         storage_override = self._library_adapter.get_storage_override()
         lib = NativeVersionStore(
             self._library_manager.get_library(name, storage_override),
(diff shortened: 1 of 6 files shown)


python/arcticdb/arctic.py
```diff
@@ -149,8 +148,13 @@ def __init__(self, uri: str, encoding_version: EncodingVersion = EncodingVersion
         self._library_adapter = _cls(uri, self._encoding_version)
         self._library_manager = LibraryManager(self._library_adapter.config_library)
         self._uri = uri
+        self._open_libraries = dict()
 
     def __getitem__(self, name: str) -> Library:
+        already_open = self._open_libraries.get(name)
```
qc00, line 154: `get(name, None)`

python/arcticdb/arctic.py
```diff
@@ -208,15 +212,18 @@ def create_library(self, name: str, library_options: Optional[LibraryOptions] =
         >>> arctic.create_library('test.library')
         >>> my_library = arctic['test.library']
         """
-        if self._library_manager.has_library(name):
-            raise ValueError(f"{name} already exists as a library. Please delete prior to re-creating.")
+        already_open = self._open_libraries.get(name)
+        if already_open or self._library_manager.has_library(name):
```
qc00, line 216: `name in self._open_libraries`

python/arcticdb/arctic.py
```diff
@@ -149,8 +148,13 @@ def __init__(self, uri: str, encoding_version: EncodingVersion = EncodingVersion
         self._library_adapter = _cls(uri, self._encoding_version)
         self._library_manager = LibraryManager(self._library_adapter.config_library)
         self._uri = uri
+        self._open_libraries = dict()
```
qc00, line 151: Should consider managing this with the "EnvironmentConfig.lib_by_path" dict

qc00 (changes requested):

python/arcticdb/arctic.py
poodlewars: That's the same as I've written https://docs.python.org/3/library/stdtypes.html?highlight=dict%20get#dict.get

python/arcticdb/arctic.py
poodlewars: I think that's quite a separate concept no? That should be for tracking all the libs that an environment _knows about_ (but which might not be "opened").

python/arcticdb/arctic.py
qc00: Ah, I've always thought the default is not optional...

PR #589, opened 2023-07-12
Support negative integers in `as_of`

alexowens90:
**What does this PR do?**

This PR enables negative indexing to select "the last-but-nth version". This is equivalent to [negative indexing in Python](https://www.tutorialspoint.com/what-is-a-negative-indexing-in-python).

For example, following this PR:

```
read(...as_of=0...) will select the first version
read(...as_of=-1...) will select the most recent version
read(...as_of=-2...) will select the version prior to the most recent version
```

**Engineering considerations**

This PR modifies `LocalVersionedEngine::get_specific_version` to do most of the heavy lifting. As a result, this change should take effect across the board for all read methods.

### DIFF
--- a/cpp/arcticdb/entity/types.hpp
+++ b/cpp/arcticdb/entity/types.hpp
@@ -71,6 +71,7 @@ using VariantId = std::variant<NumericId, StringId>;
 using StreamId = VariantId;
 using SnapshotId = VariantId;
 using VersionId = uint64_t;
+using SignedVersionId = int64_t;
 using GenerationId = VersionId;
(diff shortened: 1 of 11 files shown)


mehertz:
What happens if I pass in `-2` for a symbol with only one version?

alexowens90:
> What happens if I pass in `-2` for a symbol with only one version?

The behaviour is the same as for any non-existent integer version:

- throw for `read`
- throw for `NativeVersionStore.batch_read`
- return `None` for `Library.read_batch`

willdealtry (approved):

PR #592, opened 2023-07-13
maint: Rename `datetime64[ns]`-related fields and datatypes

jjerphan:

 - All occurrences of `MICROS_UTC64` have been renamed to `NANOSECONDS_UTC64`
 - All occurrences of `MICROS_UTC` have been renamed to `NANOSECONDS_UTC`
 - Context has been given near important places manipulating `NANOSECONDS_{UTC,UTC64}` and `datetime64[ns]`

For the full context, see the inline comments.

See [pandas 2.0 backward incompatible API changes](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#backwards-incompatible-api-changes).

#### Any other comments?

We could remove some duplication in Python (e.g. only use `DTN64_DTYPE` or `_NS_DTYPE`).

We can also remove some comments or make them shorter if this is too verbose.

:warning: Before merging, we must check what https://github.com/man-group/ArcticDB/pull/592#discussion_r1262344940 proposes is verified.


cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -20,7 +20,7 @@ message TypeDescriptor {
         FLOAT = 3;
 
         BOOL = 4;
-        MICROS_UTC = 5;
+        NANOSECONDS_UTC = 5;
```
mehertz, line 23: I think we've seen before that enum names can be changes without issue so I think this should be fine - but to be safe we should test that:

- This enables you to read data written prior to this change
- Older clients can read data written after this change has taken effect

mehertz: Ah, I see this is being run through the regressions tests. In which case we'd just need to test the latter case I think.

jjerphan: Reading [the documentation](https://protobuf.dev/programming-guides/encoding/#structure), it is changeable if the fields' ordering does not change:

> The binary version of a message just uses the field’s number as the key – the name and declared type for each field can only be determined on the decoding end by referencing the message type’s definition (i.e. the .proto file). Protoscope does not have access to this information, so it can only provide the field numbers.

If I am not mistaken, @poodlewars proposed that he runs extra checks to ensure that the _required verifications_ that you propose are verified.

poodlewars: Yep. @mehertz is still fixing the internal build. Once that's done we can validate these changes against old data.

jjerphan: Thanks for the update.

poodlewars (approved):
Happy with this. OK to wait a day or two until our internal build is fixed and we can check old data is still readable?

jjerphan:
> OK to wait a day or two until our internal build is fixed and we can check old data is still readable?

Yes, definitely we must wait for those checks to be verified.

PR #595, opened 2023-07-13
Fix in folly future chain for write batch with dedup

### LINKED ISSUE #593: Write_batch deadlocks when dedup option is enabled
### Describe the bug

write_batch operation can deadlock when depup=True option is used in the library. The reason is that check_reload function contain a get() for its futures, which breaks the entire chain of futures used by write_batch. Due to timing reasons, it cannot be reproduced in the current test infrastructure. So, it only can be reproduced when we use real storages in the test.

### Steps/Code to Reproduce

def test_write_batch_dedup(library_factory):
    ac = Arctic(<URI_REAL_STORAGE>)
    ac.create_library("test_write_batch_dedup", LibraryOptions(dedup=True))
    lib = ac["test_write_batch_dedup"]
    df1 = pd.DataFrame({'A': [1, 2, 3], 'B': [1, 2, 3]})                                                                                                                                              
    df2 = pd.DataFrame({'A': [10, 20, 30], 'B': [1, 2, 3]})     
    lib.write_batch([WritePayload("TEST_sym_1", df1), WritePayload("TEST_sym_2", df2)])

### Expected Results

No deadlock

Python: 3.6.8 (default, Jun 20 2023, 11:53:23) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]
OS: Linux-5.15.0-72-generic-x86_64-with-centos-7.9.2009-Core
ArcticDB: 1.4.1.dev0

### Backend storage used

_No response_

### Additional Context

_No response_

GemDot:
If we currently use the deduplication library option for write batching, the chain of folly::Futures may be broken due to deep get() calls within the generate_segments_from_keys and generate_keys_from_segments functions. These get() calls are unnecessary since they don't collect any set of futures and are called immediately after the asynchronous read method is executed. In these cases, the most suitable function to use is read_sync() instead of read_async() to ensure that the chain of futures is not broken.

By looking at this, I also observed that might be other potential deadlocks when trying to delete old data (because the call todo_backwards_compat_check, which is not completely sync and we still incur some get()/s there). This part has also been addressed by creating pure sync functions for this particular niche case.

closes #593

### DIFF
--- a/cpp/arcticdb/util/key_utils.hpp
+++ b/cpp/arcticdb/util/key_utils.hpp
@@ -41,10 +41,32 @@ inline void delete_keys_of_type_if(const std::shared_ptr<Store>& store, Predicat
     }
 }
 
+template<class Predicate>
+inline void delete_keys_of_type_if_sync(const std::shared_ptr<Store>& store, Predicate&& predicate, KeyType key_type, const std::string& prefix = std::string(), bool continue_on_error = false) {
+    try {
+        store->iterate_type(key_type, [predicate=std::forward<Predicate>(predicate), store=store](VariantKey &&key) {
+            if(predicate(key))
+                store->remove_key_sync(key);
+        }, prefix);
+    }
+    catch(const std::exception& ex) {
+        if(continue_on_error)
+            log::storage().warn("Caught exception {} trying to delete key, continuing", ex.what());
+        else
+            throw;
+    }
+}
+
 inline void delete_keys_of_type_for_stream(const std::shared_ptr<Store>& store, const StreamId& stream_id, KeyType key_type, bool continue_on_error = false) {
     auto prefix = std::holds_alternative<StringId>(stream_id) ? std::get<StringId>(stream_id) : std::string();
     auto match_stream_id =  [&stream_id](const VariantKey & k){ return variant_key_id(k) == stream_id; };
     delete_keys_of_type_if(store, std::move(match_stream_id), key_type, prefix, continue_on_error);
+}
+
+inline void delete_keys_of_type_for_stream_sync(const std::shared_ptr<Store>& store, const StreamId& stream_id, KeyType key_type, bool continue_on_error = false) {
+    auto prefix = std::holds_alternative<StringId>(stream_id) ? std::get<StringId>(stream_id) : std::string();
+    auto match_stream_id =  [&stream_id](const VariantKey & k){ return variant_key_id(k) == stream_id; };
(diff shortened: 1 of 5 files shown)


cpp/arcticdb/util/key_utils.hpp
```diff
@@ -41,12 +41,34 @@ inline void delete_keys_of_type_if(const std::shared_ptr<Store>& store, Predicat
     }
 }
 
+template<class Predicate>
+inline void delete_keys_of_type_if_sync(const std::shared_ptr<Store>& store, Predicate&& predicate, KeyType key_type, const std::string& prefix = std::string(), bool continue_on_error = false) {
```
willdealtry, line 45: Just delete_keys_of_type_sync would be better

willdealtry (approved):

cpp/arcticdb/util/key_utils.hpp
GemDot: as discussed, despite the name is getting a bit wierd, it is still consistent with the non-sync version

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -752,10 +752,12 @@ def test_write_pickle_batch_duplicate_symbols(arctic_library):
     assert not lib.list_symbols()
 
 
-def test_write_batch(library_factory):
+@pytest.mark.parametrize("dedup", [True, False])
```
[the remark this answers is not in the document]

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -781,6 +781,42 @@ def test_write_batch(library_factory):
         assert read_batch_result[sym].metadata == "great_metadata" + str(sym)
         assert_frame_equal(read_batch_result[sym].data, original_dataframe)
 
+    for sym in range(num_symbols):
+        data_key_version = lib._nvs.read_index("symbol_" + str(sym))["version_id"][0]
+        assert data_key_version == 0
```
alexowens90, line 786: Shouldn't these 3 lines be in the test below? And we should check all the values in the `version_id` column, not just the first one.

alexowens90 (changes requested):

python/tests/integration/arcticdb/test_arctic.py
GemDot: oh yeah. What a lapses.

changed it and also done the check for all the segments

alexowens90 (approved):

PR #597, opened 2023-07-13
Issue 520 Fix for LMDB DBI Handling

poodlewars:
**Only the HEAD and HEAD~ commits are for review here. The other two are under separate PR #585 that should be merged shortly.**

#### Reference Issues/PRs

Issue #520 this PR fixes the issue that `prophetss` reported related to parallelism in a read-only session with LMDB.

It depends on the other PR #585 and includes commits from it that will get rebased out once that is merged.

As described in the comments under issue #520 this fixes an issue with our use of LMDB database handles.

There are separate notions in LMDB:

- The environment
- The databases in an environment

The environment is similar to the ArcticDB library. Within the environment we use one LMDB "database" per `KeyType`. With LMDB you open a handle called a `dbi_handle` to each database.

The LMDB programming model insists that a database handle is only opened once per process and never closed. There is no synchronization in LMDB that handles the cleanup when a handle is closed and LMDB suggest that it is almost always a mistake to close a DBI handle. Their resources are freed when the environment shuts down.

Our use of LMDB was incorrect here as we were opening new DBI handles within each transaction. We had a model where we nested,

```
LMDB environment (many long-lived)
|_ LMDB transactions (many short-lived)
    |_ LMDB database handles (many short-lived, with races in their cleanup)
```

whereas the correct model is as implemented here:

```
LMDB environment (one long-lived)
|_ LMDB database handles (few long-lived)
    |_ LMDB transactions (many short-lived)
```

### DIFF
--- a/cpp/arcticdb/storage/lmdb/lmdb_storage-inl.hpp
+++ b/cpp/arcticdb/storage/lmdb/lmdb_storage-inl.hpp
@@ -215,7 +215,7 @@ bool LmdbStorage::do_fast_delete() {
         ARCTICDB_SUBSAMPLE(LmdbStorageOpenDb, 0)
         ARCTICDB_DEBUG(log::storage(), "dropping {}", db_name);
         ::lmdb::dbi& dbi = dbi_by_key_type_.at(db_name);
-        ::lmdb::dbi_drop(dtxn, dbi, true);
+        ::lmdb::dbi_drop(dtxn, dbi);
     });
 
     dtxn.commit();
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/storage/lmdb/lmdb_storage-inl.hpp
```diff
@@ -206,19 +206,16 @@ bool LmdbStorage::do_fast_delete() {
...
     foreach_key_type([&] (KeyType key_type) {
-        auto db_name = fmt::format("{}", key_type);
-        try {
-            ARCTICDB_SUBSAMPLE(LmdbStorageOpenDb, 0)
-            auto dbi = ::lmdb::dbi::open(dtxn, db_name.data());
-            ::lmdb::dbi_drop(dtxn, dbi, true);
-        }
-        catch (const ::lmdb::not_found_error &e) {
-            // Key type was not created, just skip.
+        if (key_type == KeyType::TOMBSTONE) {
```
willdealtry, line 210: Oh wow that's not good, they are supposed to be unique. Not your problem to fix, but we should loop round soon and fix it.

willdealtry (approved):

PR #600, opened 2023-07-14
Ensure content hash is copied during restore version + fixing timestamp-uniqueness-related flaky tests

### LINKED ISSUE #469: `test_restore_version` frequent `DuplicateKeyException` in Conda build
### Describe the bug

_No response_

### Steps/Code to Reproduce

```
_____________________________ test_restore_version _____________________________

version_store_factory = <function version_store_factory.<locals>.create_version_store at 0x14164e0e0>

    def test_restore_version(version_store_factory):
        lmdb_version_store = version_store_factory(col_per_group=2, row_per_segment=2)
        symbol = "test_restore_version"
        df1 = get_sample_dataframe(20, 4)
        df1.index = pd.DatetimeIndex([pd.Timestamp.now()] * len(df1))
        metadata = {"a": 43}
        lmdb_version_store.write(symbol, df1, metadata=metadata)
        df2 = get_sample_dataframe(20, 6)
        df2.index = df1.index + pd.Timedelta(hours=1)
        second_write_item = lmdb_version_store.write(symbol, df2, prune_previous_version=False)
        assert second_write_item.version == 1
>       restore_item = lmdb_version_store.restore_version(symbol, as_of=0)

tests/integration/arcticdb/version_store/test_basic_version_store.py:1444: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = NativeVersionStore: Library: local.test_535_2023-06-08T09_44_40_271374, Primary Storage: lmdb_storage.
symbol = 'test_restore_version', as_of = 0, kwargs = {}
version_query = <arcticdb_ext.version_store.PythonVersionStoreVersionQuery object at 0x14155fcf0>

    def restore_version(self, symbol: str, as_of: Optional[VersionQueryInput] = None, **kwargs) -> VersionedItem:
      ...
        """
>       read_result = ReadResult(*self.version_store.restore_version(symbol, version_query))
E       arcticdb_ext.storage.DuplicateKeyException: d:test_restore_version:2:0x0@1686217480302280000[1686217480276590000,1686217480276590001]

arcticdb/version_store/_store.py:1676: DuplicateKeyException
```

### Expected Results

This test don't seem to fail on other builds

Conda

### Backend storage used

_No response_

### Additional Context

_No response_

qc00:
Fixes #469

`restore_ version` doesn't copy the content hash in data keys. `restore_version` will create new versions of the data keys but as all the data keys will have a content hash of `0`, it's possible for there to be collisions if the timestamp happens to be the same. 

This change ensures that the content hash is copied over ensuring that there is no possibility of duplicated/colliding key names. As a result, this fixes the restore_version test being flaky. 

The first commit introduces a reliably failing test and the second commit fixes it

### DIFF
--- a/cpp/arcticdb/stream/stream_utils.hpp
+++ b/cpp/arcticdb/stream/stream_utils.hpp
@@ -403,22 +403,15 @@ storage::KeySegmentPair make_target_key(KeyType key_type,
     if (is_ref_key_class(key_type)) {
         return {RefKey{stream_id, key_type}, std::move(segment)};
     } else {
-        if (is_ref_key_class(variant_key_type(source_key))) {
-            auto return_segment = std::move(segment);
-            auto clone = return_segment;
-            auto decoded = decode_segment(std::move(clone));
-            auto index = index_type_from_descriptor(decoded.descriptor());
-            auto range = get_range_from_segment(index, decoded);
-            auto new_key = atom_key_builder().version_id(version_id).creation_ts(ClockType::nanos_since_epoch()).start_index(range.start_).end_index(
-                range.end_).build(stream_id, key_type);
+        util::check(!is_ref_key_class(variant_key_type(source_key)),
+            "Cannot convert ref key {} to {}", source_key, key_type);
+        auto& atom_source_key = to_atom(source_key);
+        auto new_key = atom_key_builder().version_id(version_id).creation_ts(ClockType::nanos_since_epoch())
+            .start_index(atom_source_key.start_index()).end_index(atom_source_key.end_index())
+            .content_hash(atom_source_key.content_hash())
+            .build(stream_id, key_type);
 
-            return {new_key, std::move(return_segment)};
-        } else {
-            auto new_key = atom_key_builder().version_id(version_id).creation_ts(ClockType::nanos_since_epoch()).start_index(
-                to_atom(source_key).start_index()).end_index(to_atom(source_key).end_index()).build(stream_id, key_type);
-
-            return {new_key, std::move(segment)};
-        }
+        return {new_key, std::move(segment)};
     }
 }
 
--- a/cpp/arcticdb/util/clock.hpp
+++ b/cpp/arcticdb/util/clock.hpp
@@ -50,15 +50,12 @@ struct ManualClock {
     inline static std::atomic<entity::timestamp> time_{0};
 
     static entity::timestamp nanos_since_epoch() {
-        return LinearClock::time_.load();
+        return time_.load();
     }
     static entity::timestamp coarse_nanos_since_epoch() {
-        return LinearClock::time_.load();
+        return time_.load();
     }
 };
 
 
 } // namespace arcticdb::util
-
-
-
--- a/cpp/arcticdb/util/clock.hpp
+++ b/cpp/arcticdb/util/clock.hpp
@@ -46,6 +46,17 @@ struct LinearClock {
     }
 };
 
+struct ManualClock {
+    inline static std::atomic<entity::timestamp> time_{0};
+
+    static entity::timestamp nanos_since_epoch() {
+        return LinearClock::time_.load();
+    }
+    static entity::timestamp coarse_nanos_since_epoch() {
+        return LinearClock::time_.load();
+    }
+};
+
 
 } // namespace arcticdb::util
 
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -24,11 +24,11 @@
 #include <arcticdb/python/gil_lock.hpp>
 
 namespace arcticdb::version_store {
-
+template<class ClockType>
 LocalVersionedEngine::LocalVersionedEngine(
         const std::shared_ptr<storage::Library>& library,
-        const std::optional<std::string>& license_key ARCTICDB_UNUSED) :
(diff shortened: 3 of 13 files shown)


cpp/arcticdb/stream/stream_utils.hpp
```diff
@@ -404,22 +404,15 @@ storage::KeySegmentPair make_target_key(KeyType key_type,
     if (is_ref_key_class(key_type)) {
         return {RefKey{stream_id, key_type}, std::move(segment)};
     } else {
-        if (is_ref_key_class(variant_key_type(source_key))) {
```
qc00, line 407: I currently cannot find an example where we copy a ref key to an atom key, so this feature is removed to avoid having to refactor out the content hash logic.

cpp/arcticdb/util/clock.hpp
```diff
@@ -46,6 +46,17 @@ struct LinearClock {
     }
 };
 
+struct ManualClock {
+    inline static std::atomic<entity::timestamp> time_{0};
+
+    static entity::timestamp nanos_since_epoch() {
+        return LinearClock::time_.load();
```
poodlewars, line 53: Very surprised that this uses the `time_` from the other clock `LinearClock`? What is `ManualClock::time_` used for?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -24,11 +24,11 @@
 #include <arcticdb/python/gil_lock.hpp>
 
 namespace arcticdb::version_store {
-
+template<class ClockType>
 LocalVersionedEngine::LocalVersionedEngine(
         const std::shared_ptr<storage::Library>& library,
-        const std::optional<std::string>& license_key ARCTICDB_UNUSED) :
-    store_(std::make_shared<async::AsyncStore<util::SysClock>>(library, codec::default_lz4_codec(), encoding_version(library->config()))),
+        const ClockType&) :
```
poodlewars, line 30: The `ClockType` param isn't necessary

cpp/arcticdb/version/version_store_api.hpp
```diff
@@ -50,9 +50,10 @@ namespace as = arcticdb::stream;
 class PythonVersionStore : public LocalVersionedEngine {
 
   public:
-    explicit PythonVersionStore(
-        const std::shared_ptr<storage::Library>& library,
-        const std::optional<std::string>& license_key = std::nullopt);
+    template<class ClockType = util::SysClock>
+    explicit PythonVersionStore(const std::shared_ptr<storage::Library>& library, const ClockType& ct = util::SysClock{}) :
```
poodlewars, line 54: ClockType parameter not necessary

poodlewars (approved):

poodlewars (commented):
Fine with this other than the use of the LinearClock `time_`

cpp/arcticdb/util/clock.hpp
qc00: This is a copy-and-paste fail. I did not expect LinearClock would quality the field name...

cpp/arcticdb/version/local_versioned_engine.cpp
qc00: It's used to allow the `ClockType` template variable to be inferred, as there's no syntax to do this explicitly.

poodlewars: OK. There are plenty of other types that inject a clock without using this pattern, I would prefer that we're consistent with them.

poodlewars (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
qc00: This case is unique in that the `ClockType` is only used in the constructor itself and never in the body of the class.

In other words, I only wanted a templated constructor function, not a templated class which requires too many changes. (Every place that used to refer to `LocalVersionedEngine` and its sub-type `PythonVersionStore` will have to become transitively templated.)

The most sensible API is to make the c'tor accept `shared_ptr<Stroages>`, which would allow the Clock to be picked external to the c'tor, but that would still involve a lot of refactoring.

willdealtry: We could make it a variant, that's the normal way to avoid transitive type parameterization

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -651,6 +653,13 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
              "Returns latest timestamp of a symbol")
         ;
 
+    py::class_<ManualClockVersionStore, PythonVersionStore>(version, "ManualClockVersionStore")
```
willdealtry, line 656: Did you know that there is a PilotedClock that just counts up monotonically from zero?

qc00: I forgot about it, but looking at it, it's equivalent to the existing `LinearClock` where the "time" increase by 1 after each call.

Whereas the behaviour I need to test is the time stays the same between calls (to match the low-resolution clock on non-Linux).

cpp/arcticdb/version/local_versioned_engine.cpp
qc00: I'll open a separate PR to fix this

Issue #496, opened 2023-06-16
Flaky tests


mehertz:
Tracking ticket for flaky tests

- [x] #469 
- [ ] #316 
- [x] [3.8 windows test_filter_categorical[lmdb_version_store]](https://github.com/man-group/ArcticDB/actions/runs/5281710778/jobs/9556935531)
- [x] [3.8 windows test_filter_categorical[lmdb_version_store_dynamic_schema]](https://github.com/man-group/ArcticDB/actions/runs/5281710778/jobs/9556935531)
- [x] [3.7 windows test_filter_categorical[lmdb_version_store_dynamic_schema]](https://github.com/man-group/ArcticDB/actions/runs/5310593194/jobs/9613617253)
    -> Through debugging, the underlying issue is #229
- [x] [linux 3.11 test_filter_numeric_isnotin_unsigned](https://github.com/man-group/ArcticDB/actions/runs/5288922390/jobs/9571531805?pr=490) -> #604
- [x] [linux 3.9 test_filter_numeric_isnotin_unsigned](https://github.com/man-group/ArcticDB/actions/runs/5422309116/jobs/9859134990)
- [x] [Windows 3.11 test_filter_numeric_isnotin_unsigned](https://github.com/man-group/ArcticDB/actions/runs/5453649236/jobs/9924117342)
- [x] [test_filtering.py::test_filter_numeric_isnotin_unsigned](https://github.com/man-group/ArcticDB/actions/runs/5313566049/jobs/9621422799)
- [x] [linux 3.11 test_hypothesis_mean_agg_dynamic](https://github.com/man-group/ArcticDB/actions/runs/5422309116/jobs/9859052796)
    -> Not trivially fixable/skippable. See #603
- [x] [Windows 3.7 test_basic_version_store.py::test_read_ts[EncodingVersion.V2]](https://github.com/man-group/ArcticDB/actions/runs/5464577790/jobs/9947764682)
    -> Fix attempt bundled with #604
- [ ] [Windows 3.8 `test_find_version`](https://github.com/man-group/ArcticDB/actions/runs/5556047996/jobs/10149716495?pr=604#step:9:633)
    -> Looks similar in nature to `test_read_ts` above
- [x] `test_append_with_cont_mem_problem ` below #985 
- [x] `test_append_with_defragmentation` below #985 
[...]

PR #604, opened 2023-07-14
Fix flaky test_filter_numeric_isnotin_unsigned (#496)

qc00:
The cause was type_arithmetic_promoted_type would return int64 as the common type for uint64 and any signed int. When we then do the static_cast<WideType>(*ptr++) before calling the Is[Not]InOperator, the uint64 is converted to int64 and the special overloads in the Operators are never used.

Additional changes@

+ Ability to only_test_encoding_version_v1 in a test/class/module
+ Attempt to fix another flaky test `test_read_ts`

### DIFF
--- a/cpp/arcticdb/processing/operation_types.hpp
+++ b/cpp/arcticdb/processing/operation_types.hpp
@@ -181,7 +181,7 @@ struct type_arithmetic_promoted_type {
                             // If the signed type is strictly larger than the unsigned type, then promote to the signed type
                             typename arithmetic_promoted_type::details::signed_width_t<max_width>,
                             // Otherwise, check if the unsigned one is the widest type we support
-                            std::conditional_t<std::is_same_v<LHS, uint64_t> || std::is_same_v<LHS, uint64_t>,
+                            std::conditional_t<std::is_same_v<LHS, uint64_t> || std::is_same_v<RHS, uint64_t>,
                                 // If so, there's no common type that can completely hold both arguments. We trigger operation-specific handling
                                 std::conditional_t<std::is_base_of_v<MembershipOperator, Func>,
                                     RHS, // Retains ValueSetBaseType in binary_membership()
@@ -383,12 +383,6 @@ bool operator()(int64_t t, uint64_t u) const {
 
 struct MembershipOperator {
 protected:
-    /** Returns the high bits beyond what can be held in the signed integer type I, including I's sign bit. */
-    template<typename I>
-    static constexpr uint64_t incomparable_bits() {
-        return std::numeric_limits<uint64_t>::max() - static_cast<uint64_t>(std::numeric_limits<I>::max());
-    }
-
     template<typename U>
     static constexpr bool is_signed_int = std::is_integral_v<U> && std::is_signed_v<U>;
 
@@ -412,8 +406,7 @@ bool operator()(T t, const std::unordered_set<U>& u) const {
 
 template<typename U, typename=std::enable_if_t<is_signed_int<U>>>
 bool operator()(uint64_t t, const std::unordered_set<U>& u, UInt64SpecialHandlingTag = {}) const {
-    auto incomparable = t & incomparable_bits<U>();
-    if (incomparable)
+    if (t > static_cast<uint64_t>(std::numeric_limits<U>::max()))
         return false;
     else
         return u.count(t) > 0;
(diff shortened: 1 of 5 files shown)


python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1010,6 +1010,25 @@ def test_filter_numeric_isnotin_hashing_overflow(lmdb_version_store):
     assert_frame_equal(df, result)
 
 
+_uint64_max = np.iinfo(np.uint64).max
+
+
+@pytest.mark.parametrize("op", ("in", "not in"))
+@pytest.mark.parametrize("signed_type", (np.int8, np.int16, np.int32, np.int64))
+@pytest.mark.parametrize("uint64_in", ("df", "vals") if PANDAS_VERSION >= Version("1.2") else ("vals",))
```
alexowens90, line 1018: Why the limitation with the Pandas version?

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -373,18 +381,44 @@ bool operator()(int64_t t, uint64_t u) const {
...
 template<typename T, typename U>
 bool operator()(T t, const std::unordered_set<U>& u) const {
     return u.count(t) > 0;
 }
-bool operator()(uint64_t t, const std::unordered_set<int64_t>& u) const {
-    if (comparison::msb_set(t))
+
+template<typename U, typename=std::enable_if_t<is_signed_int<U>>>
+bool operator()(uint64_t t, const std::unordered_set<U>& u, UInt64SpecialHandlingTag = {}) const {
+    auto incomparable = t & incomparable_bits<U>();
```
alexowens90, line 415: I don't think `incomparable_bits` conveys intent very well. I think this would be equivalent and more obvious:
`auto incomparable = t > static_cast<uint64_t>(std::numeric_limits<U>::max())`

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -179,8 +180,15 @@ struct type_arithmetic_promoted_type {
                         std::conditional_t<(std::is_signed_v<LHS> && sizeof(LHS) > sizeof(RHS)) || (std::is_signed_v<RHS> && sizeof(RHS) > sizeof(LHS)),
                             // If the signed type is strictly larger than the unsigned type, then promote to the signed type
                             typename arithmetic_promoted_type::details::signed_width_t<max_width>,
-                            // Otherwise, promote to a signed type wider than the unsigned type, so that it can be exactly represented
-                            typename arithmetic_promoted_type::details::signed_width_t<2 * max_width>
+                            // Otherwise, check if the unsigned one is the widest type we support
+                            std::conditional_t<std::is_same_v<LHS, uint64_t> || std::is_same_v<LHS, uint64_t>,
```
alexowens90, line 184: Should one of these be `RHS`?

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_filtering.py
qc00: See the similar change on line 971. Pandas older than 1.2.0 (e.g. 1.1.5, the last version for Python 3.6) exhibit the same buggy behaviour when signed and unsigned mix.

cpp/arcticdb/processing/operation_types.hpp
qc00: Oops...

alexowens90 (approved):

python/tests/conftest.py
```diff
@@ -253,9 +253,15 @@ class EncodingVersion(enum.IntEnum):
     V2 = 1
 
 
-@pytest.fixture(params=list(EncodingVersion))
-def encoding_version(request) -> EncodingVersion:
-    return request.param
+@pytest.fixture(scope="session")
+def only_test_encoding_version_v1():
```
willdealtry, line 257: I don't really want this to become a general pattern. If we're going to alternate between encodings then that's fine, but I don't think we should be skipping v2 encoding for a whole test module

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -373,18 +381,37 @@ bool operator()(int64_t t, uint64_t u) const {
 }
 };
 
-struct IsInOperator {
+struct MembershipOperator {
```
willdealtry, line 384: Any particular reason why this needs to be a base class rather than just two templates?

python/tests/conftest.py
qc00: I fully appreciate your point and definitely didn't intend for this to be widely used. Currently, this is only used by `test_filtering` which should not involve much encoding version handling.

My main motivation is the `test_filtering` "unit" test file has many Hypothesis tests and takes many minutes to complete. It might be better to move those to an integration test to keep the unit suite fast.

cpp/arcticdb/processing/operation_types.hpp
qc00: It started that way, but something didn't compile and it was late in the night so I went for this.

In hindsight, what didn't work was probably one of the function templates, not the class-level template, so I am happy to try again if you feel strongly about this?

PR #608, opened 2023-07-15
Update python line in issue-template so it works on windows

jamesmunro:


jjerphan (approved):
Thank you for spotting this, @jamesmunro.

PR #610, opened 2023-07-17
Windows is Beta

jamesmunro:


poodlewars (approved):

jamesmunro:
> Why is it "beta"?

We've got several bug reports open against Windows (mostly about LMDB backend), my own testing also showed issues.

PR #623, opened 2023-07-19
Fix to ensure batch prune previous methods clean up index and data keys as well as version keys

### LINKED ISSUE #230: batch_write, batch_append, and batch_write_metadata orphan data when prune_previous_version=True
The version map is updated correctly, but the index and data keys are never deleted. Steps to reproduce given a `NativeVersionStore` `lib`:
```
import pandas as pd
lib_tool = lib.library_tool()
sym = "test_symbol"
df0 = pd.DataFrame({"col_0": ["a", "b"]}, index=pd.date_range("2000-01-01", periods=2))
df1 = pd.DataFrame({"col_0": ["c", "d"]}, index=pd.date_range("2000-01-03", periods=2))

lib.write(sym, df0)

lib.batch_write([sym], [df1], prune_previous_version=True) # Or batch_append

print(lib.list_versions(sym)) # Correctly shows only version 1 is live
print(lib_tool.find_keys(KeyType.TABLE_INDEX)) # Shows index keys for versions 0 and 1
print(lib_tool.find_keys(KeyType.TABLE_DATA)) # Shows data keys for versions 0 and 1
```

GemDot:
So far, in the "prune previous version" option for batch methods, the version map is updated correctly, but the index and data keys are never deleted. This pull request addresses this issue by calling the corresponding delete_unreferenced_pruned_indexes method after updating the version map. This is not a trivial task since delete_unreferenced_pruned_indexes is currently not designed to be asynchronous, so it had to be converted to a purely synchronous method.

closes #230

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -44,7 +44,7 @@ LocalVersionedEngine::LocalVersionedEngine(
     }
 }
 
-void LocalVersionedEngine::delete_unreferenced_pruned_indexes(
+folly::Future<folly::Unit> LocalVersionedEngine::delete_unreferenced_pruned_indexes(
         const std::vector<AtomKey> &pruned_indexes,
         const AtomKey& key_to_keep
 ) {
@@ -56,12 +56,13 @@ void LocalVersionedEngine::delete_unreferenced_pruned_indexes(
                     pruned_indexes);
             in_snaps.insert(key_to_keep);
             PreDeleteChecks checks{false, false, false, false, in_snaps};
-            delete_trees_responsibly(not_in_snaps, {}, {}, checks);
+            return delete_trees_responsibly(not_in_snaps, {}, {}, checks);
         }
     } catch (const std::exception &ex) {
         // Best-effort so deliberately swallow
         log::version().warn("Could not prune previous versions due to: {}", ex.what());
     }
+    return folly::Unit();
 }
 
 void LocalVersionedEngine::create_column_stats_internal(
@@ -555,7 +556,7 @@ VersionedItem LocalVersionedEngine::update_internal(
                                           dynamic_schema);
         if (prune_previous_versions) {
             auto pruned_indexes = version_map()->write_and_prune_previous(store(), versioned_item.key_, update_info.previous_index_key_);
-            delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_);
+            delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_).get();
         } else {
             version_map()->write_version(store(), versioned_item.key_);
         }
@@ -602,7 +603,7 @@ VersionedItem LocalVersionedEngine::write_versioned_metadata_internal(
     if(prune_previous_versions) {
         auto versioned_item = VersionedItem{index_key};
         auto pruned_indexes = version_map()->write_and_prune_previous(store(), index_key, update_info.previous_index_key_);
-        delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_);
+        delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_).get();
         return versioned_item;
     }
     else {
@@ -638,7 +639,7 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
 
     if(prune_previous_versions) {
         auto pruned_indexes = version_map()->write_and_prune_previous(store(), versioned_item.key_, maybe_prev);
-        delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_);
+        delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_).get();
         return versioned_item;
     }
     else {
@@ -743,7 +744,7 @@ std::unordered_map<StreamId, VersionId> min_versions_for_each_stream(const std::
     return out;
 }
 
-void LocalVersionedEngine::delete_trees_responsibly(
+folly::Future<folly::Unit> LocalVersionedEngine::delete_trees_responsibly(
         const std::vector<IndexTypeKey>& orig_keys_to_delete,
         const arcticdb::MasterSnapshotMap& snapshot_map,
         const std::optional<SnapshotId>& snapshot_being_deleted,
@@ -790,19 +791,14 @@ void LocalVersionedEngine::delete_trees_responsibly(
     if (load_type != LoadType::NOT_LOADED) {
         std::unordered_map<StreamId, std::shared_ptr<VersionMapEntry>> entry_map;
         {
-            auto min_vers = min_versions_for_each_stream(orig_keys_to_delete);
-
-            async::submit_tasks_for_range(min_vers,
-                    [store=store(), version_map=version_map(), load_type](auto& sym_version) {
-                        auto load_param = load_type == LoadType::LOAD_DOWNTO
-                                ? LoadParameter{load_type, static_cast<SignedVersionId>(sym_version.second)}
-                                : LoadParameter{load_type};
-                        return async::submit_io_task(CheckReloadTask{store, version_map, sym_version.first,
-                                                                     load_param, true});
-                    },
-                    [&entry_map](auto& sym_version, auto&& entry) {
-                        entry_map.emplace(std::move(sym_version.first), entry);
-                    });
+            auto min_versions = min_versions_for_each_stream(orig_keys_to_delete);
+            for (const auto& min : min_versions) {
+                auto load_param = load_type == LoadType::LOAD_DOWNTO
+                        ? LoadParameter{load_type, static_cast<SignedVersionId>(min.second)}
+                        : LoadParameter{load_type};
+                const auto entry = version_map()->check_reload(store(), min.first, load_param, true, false, __FUNCTION__);
+                entry_map.emplace(std::move(min.first), entry);
+            }
         }
 
         keys_to_delete.remove_if([&entry_map, &check, &not_to_delete](const auto& key) {
@@ -841,38 +837,51 @@ void LocalVersionedEngine::delete_trees_responsibly(
     auto data_keys_not_to_be_deleted = get_data_keys_set(store(), *not_to_delete, read_opts);
     not_to_delete.clear();
     log::version().debug("Forbidden: {} total of data keys", data_keys_not_to_be_deleted.size());
-
     RemoveOpts remove_opts;
     remove_opts.ignores_missing_key_ = true;
+    
+    std::vector<entity::VariantKey> vks_column_stats;
+    std::transform(keys_to_delete->begin(),
+                    keys_to_delete->end(),
+                    std::back_inserter(vks_column_stats),
+                    [](const IndexTypeKey& index_key) {
+        return index_key_to_column_stats_key(index_key);
+    });
+    log::version().debug("Number of Column Stats keys to be deleted: {}", vks_column_stats.size());
 
-    std::vector<entity::VariantKey> vks;
-    // Delete any associated column stats keys first
-    if (!dry_run) {
-        std::transform(keys_to_delete->begin(),
-                       keys_to_delete->end(),
-                       std::back_inserter(vks),
-                       [](const IndexTypeKey& index_key) {
-            return index_key_to_column_stats_key(index_key);
-        });
-        store()->remove_keys(vks, remove_opts).get();
-    }
+    std::vector<entity::VariantKey> vks_to_delete;
+    std::copy(std::make_move_iterator(keys_to_delete->begin()),
+                std::make_move_iterator(keys_to_delete->end()),
+                std::back_inserter(vks_to_delete));
+    log::version().debug("Number of Index keys to be deleted: {}", vks_to_delete.size());
 
-    vks.clear();
-    if (!dry_run) {
-        std::copy(keys_to_delete->begin(), keys_to_delete->end(), std::back_inserter(vks));
-        store()->remove_keys(vks, remove_opts).get();
-    }
-
-    vks.clear();
+    std::vector<entity::VariantKey> vks_data_to_delete;
     std::copy_if(std::make_move_iterator(data_keys_to_be_deleted.begin()),
-                 std::make_move_iterator(data_keys_to_be_deleted.end()),
-                 std::back_inserter(vks),
-                 [&](const auto& k) {return !data_keys_not_to_be_deleted.count(k);});
-    log::version().debug("Index keys deleted. Proceeding with {} number of data keys", vks.size());
+                std::make_move_iterator(data_keys_to_be_deleted.end()),
+                std::back_inserter(vks_data_to_delete),
+                [&](const auto& k) {return !data_keys_not_to_be_deleted.count(k);});
+    log::version().debug("Number of Data keys to be deleted: {}", vks_data_to_delete.size());
+
+    folly::Future<folly::Unit> remove_keys_fut;
     if (!dry_run) {
-        store()->remove_keys(vks, remove_opts).get();
+        // Delete any associated column stats keys first
+        remove_keys_fut = store()->remove_keys(std::move(vks_column_stats), remove_opts)
+        .thenValue([this, vks_to_delete = std::move(vks_to_delete), remove_opts](auto&& ) mutable {
+            log::version().debug("Column Stats keys deleted.");
+            return store()->remove_keys(std::move(vks_to_delete), remove_opts);
+        })
+        .thenValue([this, vks_data_to_delete = std::move(vks_data_to_delete), remove_opts](auto&&) mutable {
+            log::version().debug("Index keys deleted.");
+            return store()->remove_keys(std::move(vks_data_to_delete), remove_opts);
+        })
+        .thenValue([](auto&&){
+            log::version().debug("Data keys deleted.");
+            return folly::Unit();
+        });
     }
+    return remove_keys_fut;
 }
+
 
 void LocalVersionedEngine::remove_incomplete(
     const StreamId& stream_id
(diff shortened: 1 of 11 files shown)


python/tests/unit/arcticdb/version_store/test_deletion.py
```diff
@@ -285,3 +285,28 @@ def test_delete_date_range_pickled_symbol(lmdb_version_store):
...
+    assert len(lib_tool.find_keys(KeyType.TABLE_INDEX)) == 1
+    assert len(lib_tool.find_keys(KeyType.TABLE_DATA)) == 1
+
+    lib.batch_write([sym1, sym2], [df0, df0])
+    lib.batch_write([sym1, sym2], [df1, df1], prune_previous_version=True)
+
+    assert len(lib.list_versions(sym1)) == 1
+    assert len(lib.list_versions(sym2)) == 1
+    assert len(lib_tool.find_keys(KeyType.TABLE_INDEX)) == 2
+    assert len(lib_tool.find_keys(KeyType.TABLE_DATA)) == 2
```
mehertz, line 312: As @jamesmunro points out, can we also assert that the structure of:

1. Version ref keys
2. Symbol list keys

Is as expected. We should do this more broadly, but no reason not to do it for this test.

mehertz (changes requested):

python/tests/unit/arcticdb/version_store/test_deletion.py
```diff
@@ -285,3 +285,28 @@ def test_delete_date_range_pickled_symbol(lmdb_version_store):
     assert lmdb_version_store.is_symbol_pickled(symbol)
     with pytest.raises(InternalException) as e_info:
         lmdb_version_store.delete(symbol, (idx[1], idx[2]))
+
+
+def test_write_batch_prune_previous_versions(lmdb_version_store):
+    """Should be able to write different size of batch of data."""
```
mehertz, line 291: Wrong description.

mehertz (changes requested):

python/tests/unit/arcticdb/version_store/test_deletion.py
```diff
@@ -285,3 +285,28 @@ def test_delete_date_range_pickled_symbol(lmdb_version_store):
     assert lmdb_version_store.is_symbol_pickled(symbol)
     with pytest.raises(InternalException) as e_info:
         lmdb_version_store.delete(symbol, (idx[1], idx[2]))
+
+
+def test_write_batch_prune_previous_versions(lmdb_version_store):
```
mehertz, line 290: Surely this isn't a unit test?!

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -752,6 +752,45 @@ def test_write_pickle_batch_duplicate_symbols(arctic_library):
     assert not lib.list_symbols()
 
 
+def test_write_batch_prune_previous_versions(library_factory):
+    """Should be able to write different size of batch of data."""
```
mehertz, line 756: Wrong test description.

mehertz (changes requested):

python/tests/unit/arcticdb/version_store/test_deletion.py
GemDot: I have also tested it with the VERSION and SYMBOL_LIST keys

python/tests/unit/arcticdb/version_store/test_deletion.py
GemDot: Changed the description

python/tests/integration/arcticdb/test_arctic.py
GemDot: Changed the description

python/tests/unit/arcticdb/version_store/test_deletion.py
GemDot: Moved to test_basic_version_store.py, where other tests verify the same option for non-batch methods

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -209,6 +209,66 @@ def test_prune_previous_versions_multiple_times(lmdb_version_store, symbol):
...
+    # When
+    lib.batch_write([sym1, sym2], [df0, df0])
+    lib.batch_write([sym1, sym2], [df1, df1], prune_previous_version=True)
+
+    # Then - only latest version and keys should survive
+    assert len(lib.list_versions(sym1)) == 1
+    assert len(lib.list_versions(sym2)) == 1
+    assert len(lib_tool.find_keys(KeyType.TABLE_INDEX)) == 2
+    assert len(lib_tool.find_keys(KeyType.TABLE_DATA)) == 2
+    # Then - we got 3 version keys per symbol: version 0, version 0 tumbstone, version 1
```
mehertz, line 232:
```suggestion
    # Then - we got 3 version keys per symbol: version 0, version 0 tombstone, version 1
```

mehertz (changes requested):

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -840,39 +836,53 @@ void LocalVersionedEngine::delete_trees_responsibly(
...
+        std::copy_if(std::make_move_iterator(data_keys_to_be_deleted.begin()),
+                    std::make_move_iterator(data_keys_to_be_deleted.end()),
+                    std::back_inserter(vks_data_to_delete),
+                    [&](const auto& k) {return !data_keys_not_to_be_deleted.count(k);});
+
+        log::version().debug("Index keys deleted. Proceeding with {} number of data keys", vks_data_to_delete.size());
+        folly::Future<folly::Unit> ret_fut;
+        if (!dry_run) {
+            ret_fut = store()->remove_keys(std::move(vks_data_to_delete), remove_opts)
+            .thenValue([](auto&&){
```
alexowens90, line 877: Avoid having calls to `thenValue` inside other calls to `thenValue` where possible, it makes things much harder to reason about.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -840,39 +836,53 @@ void LocalVersionedEngine::delete_trees_responsibly(
...
+    .thenValue([this, data_keys_to_be_deleted = std::move(data_keys_to_be_deleted), data_keys_not_to_be_deleted = std::move(data_keys_not_to_be_deleted), remove_opts, dry_run](auto&& ){
+        std::vector<entity::VariantKey> vks_data_to_delete;
+        std::copy_if(std::make_move_iterator(data_keys_to_be_deleted.begin()),
+                    std::make_move_iterator(data_keys_to_be_deleted.end()),
+                    std::back_inserter(vks_data_to_delete),
+                    [&](const auto& k) {return !data_keys_not_to_be_deleted.count(k);});
+
+        log::version().debug("Index keys deleted. Proceeding with {} number of data keys", vks_data_to_delete.size());
+        folly::Future<folly::Unit> ret_fut;
+        if (!dry_run) {
```
alexowens90, line 875: This can be removed once the return is one long `thenValue` chain starting on line 857

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -866,6 +866,46 @@ def test_prune_previous_versions_with_write(arctic_library):
     assert v3.empty
 
 
+def test_prune_previous_versions_with_write_batch(library_factory):
```
alexowens90, line 869: What does this test that isn't covered in `test_basic_version_store.py`?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -210,6 +210,66 @@ def test_prune_previous_versions_multiple_times(lmdb_version_store, symbol):
...
+    version_keys = lib_tool.find_keys(KeyType.VERSION)
+    keys_for_sym1 = [k for k in version_keys if k.id == sym1]
+    keys_for_sym2 = [k for k in version_keys if k.id == sym2]
+    assert len(keys_for_sym1) == 3
+    assert len(keys_for_sym2) == 3
+    # Then - we got 4 symbol keys: 2 for each of the writes
+    assert len(lib_tool.find_keys(KeyType.SYMBOL_LIST)) == 4
+
+
+def test_prune_previous_versions_append_batch(lmdb_version_store):
```
alexowens90, line 243: What about a similar test for batch write metadata?

alexowens90 (changes requested):

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -840,39 +836,53 @@ void LocalVersionedEngine::delete_trees_responsibly(
...
+
+        write_version_fut = store()->remove_keys(std::move(vks_column_stats), remove_opts)
+        .thenValue([this, vks_to_delete = std::move(vks_to_delete), remove_opts](auto&& ) mutable {
+            return store()->remove_keys(std::move(vks_to_delete), remove_opts);
+        })
+        .thenValue([](auto&&){
+            return folly::Unit();
+        });
     }
+    return std::move(write_version_fut)
```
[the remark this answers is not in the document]
GemDot: I propose preserving the current functionality of dry_run as we agreed with Qi. After the return std::move(write_version_fut) statement, we need to handle both cases, dry_run=True and dry_run=False, without duplicating code. When dry_run is True, we won't perform any actions before that statement, but we still want to execute some actions afterwards. So, I am not sure what you are exactly proposing to do differently.

python/tests/integration/arcticdb/test_arctic.py
GemDot: we also have tests for the same batch operations from both, library and store APIs

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
GemDot: sounds plausible

alexowens90: I don't think it should delete any index or data keys, but I want the expected behaviour to be codified in a test

python/tests/integration/arcticdb/test_arctic.py
alexowens90: We should try to only add similar tests of `NativeVersionStore` and `Library` if the behaviour is different. In this case it isn't, so remove this test

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
GemDot: If prune_previous_versions is enabled for the write_metadata operation and there are existing previous versions, the expected behavior would be to create a new version that includes the previous data along with the new metadata. Additionally, any previous versions that are no longer needed would be removed, along with any unused data or index keys.
In summary, the operation would create a new version with the combined data and metadata, while removing unnecessary previous versions and associated keys.

Is it not what you expect @alexowens90 ?

python/tests/integration/arcticdb/test_arctic.py
GemDot: agree

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -841,39 +837,56 @@ void LocalVersionedEngine::delete_trees_responsibly(
...
```
vasil-pashov, line 866: Returning rvalues is usually a bad idea as it prevents the compiler from doing copy elision optimizations. Especially in this case as folly::Future has its copy [constructor deleted](https://github.com/facebook/folly/blob/c409a5a381bfb88fe4aafb5270aaf24c8267cf99/folly/futures/Future.h#L1069).
```suggestion
    return write_version_fut
```

cpp/arcticdb/version/local_versioned_engine.cpp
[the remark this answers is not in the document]
GemDot: As agreed, in order to leave the code cleaner, we will increase the memory footprint by constructing the three vectors in advance

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: As agreed, in order to leave the code cleaner, we will increase the memory footprint by constructing the three vectors in advance

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: As agreed, in order to leave the code cleaner, we will increase the memory footprint by constructing the three vectors in advance

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: I agree in the general idea. However, note that in this specific case you commented we were concatenating the futures (it is why the line you suggested to change does not contain ";"), which they really need to be moved when we create such a chain. For example: return std::move(A).ThenValue(...).ThenValue(...);

nevertheless, this specific part of the code has been refactored in the current version

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
alexowens90: Yes, this is what I expect

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -210,6 +210,96 @@ def test_prune_previous_versions_multiple_times(lmdb_version_store, symbol):
...
+    lib.batch_write_metadata([sym1, sym2], [meta1, meta1], prune_previous_version=True)
+
+    # Then - only latest version and keys should survive
+    assert len(lib.list_versions(sym1)) == 1
+    assert len(lib.list_versions(sym2)) == 1
+    assert len(lib_tool.find_keys(KeyType.TABLE_INDEX)) == 2
+    assert len(lib_tool.find_keys(KeyType.TABLE_DATA)) == 2
+    # Then - we got 3 version keys per symbol: version 0, version 0 tombstone, version 1
+    version_keys = lib_tool.find_keys(KeyType.VERSION)
+    keys_for_sym1 = [k for k in version_keys if k.id == sym1]
```
alexowens90, line 265: nit: The library tool has a `find_keys_for_id` method that just gets keys of the specified type for one symbol

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -210,6 +210,96 @@ def test_prune_previous_versions_multiple_times(lmdb_version_store, symbol):
...
+    keys_for_sym2 = [k for k in version_keys if k.id == sym2]
+    assert len(keys_for_sym1) == 3
+    assert len(keys_for_sym2) == 3
+    # Then - we got 4 symbol keys: 2 for each of the writes
```
alexowens90, line 269: Comment isn't accurate

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -841,39 +837,49 @@ void LocalVersionedEngine::delete_trees_responsibly(
...
-                 std::make_move_iterator(data_keys_to_be_deleted.end()),
-                 std::back_inserter(vks),
-                 [&](const auto& k) {return !data_keys_not_to_be_deleted.count(k);});
-    log::version().debug("Index keys deleted. Proceeding with {} number of data keys", vks.size());
+                std::make_move_iterator(data_keys_to_be_deleted.end()),
+                std::back_inserter(vks_data_to_delete),
+                [&](const auto& k) {return !data_keys_not_to_be_deleted.count(k);});
+    log::version().debug("Number of Data keys to be deleted: {}", vks_data_to_delete.size());
+
+    folly::Future<folly::Unit> write_version_fut;
```
alexowens90, line 865: This is a poor variable name, no versions are being written

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -841,39 +837,49 @@ void LocalVersionedEngine::delete_trees_responsibly(
...
     if (!dry_run) {
-        store()->remove_keys(vks, remove_opts).get();
+        // Delete any associated column stats keys first
+        write_version_fut = store()->remove_keys(std::move(vks_column_stats), remove_opts)
+        .thenValue([this, vks_to_delete = std::move(vks_to_delete), remove_opts](auto&& ) mutable {
+            return store()->remove_keys(std::move(vks_to_delete), remove_opts);
```
alexowens90, line 870: The logging has become slightly less useful. It used to log after the index keys were deleted, which gave some indication of progress. If you move the "Number of * keys to be deleted" logs into the `thenValue`s for index and data keys then it will show progress in the logs again.

alexowens90 (changes requested):

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: What's the point of having the dry_run option then?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
GemDot: changed

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
GemDot: changed

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: agree, changed

cpp/arcticdb/version/local_versioned_engine.cpp
vasil-pashov: You have a deleted copy constructor, there is no need to move in this case. The compiler will do the right thing in this case. It is either going to move or elide.

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: As discussed, separate logs have been added to track the progress of key removals.

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: The line of code commented in this thread is not present anymore and as I commented it was applying to the future chain construction, which require moving. Is there any other line that concerns you? If so, please can you refer to it.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -210,6 +210,99 @@ def test_prune_previous_versions_multiple_times(lmdb_version_store, symbol):
...
+    keys_for_sym1 = lib_tool.find_keys_for_id(KeyType.VERSION, sym1)
+    keys_for_sym2 = lib_tool.find_keys_for_id(KeyType.VERSION, sym2)
+
+    assert len(keys_for_sym1) == 3
+    assert len(keys_for_sym2) == 3
+    # Then - we got 4 symbol keys: 2 for each of the writes
+    assert len(lib_tool.find_keys(KeyType.SYMBOL_LIST)) == 4
+
+
+def test_prune_previous_versions_write_batch_metadata(lmdb_version_store):
```
mehertz, line 244:
```suggestion
def test_prune_previous_versions_write_metadata_batch(lmdb_version_store):
```

Should match the method name (which I really hope is consistent with the other methods!)

GemDot: I agree to change it, but `test_prune_previous_versions_batch_write_metadata` instead

mehertz: Oh I see. This method doesn't exist in `library.py`. Okay, yes, then agreed.

PR #627, opened 2023-07-20
Add a more strict check for chars in the symbol names

G-D-Petrov:
Add a more strict check for the chars in the symbol names based on a whitelist.
Also added a way to turn this functionality off by setting an env variable STRICT_SYMBOL_CHECK to something other than 1.

#### Any other comments?
There are 2 tests added to test the new functionality and the STRICT_SYMBOL_CHECK env var.


poodlewars (commented):
The function that you've changed affects more than just writing new symbols. I think we should make it so that the new validation only kicks in when writing a totally new symbol (as opposed to writing a new version of an existing symbol).

Hence:

- write
- write_batch
- write_metadata
- write_metadata_batch
- append (when in "upsert" mode, the default, append writes the symbol if it does not already exist)

This will mean it's still possible to work with old symbols with names that would fail the new validation.

cpp/arcticdb/version/test/rapidcheck_version_map.cpp
```diff
@@ -235,6 +235,8 @@ struct GetAllVersions : rc::state::Command<Model, MapStorePair> {
 };
 
 RC_GTEST_PROP(VersionMap, Rapidcheck, ()) {
+    char env[] = "ARCTICDB_NO_STRICT_SYMBOL_CHECK=0";
+    putenv(env);
```
qc00, line 239: Should reset this environment variable after the tests to avoid affecting a subsequent test.

We also have a run-time `ConfigsMap` which can be configured from environment variables (via a Python hook) and is easier to reset in tests.

qc00 (changes requested):

cpp/arcticdb/version/test/test_version_store.cpp
```diff
@@ -118,6 +118,27 @@ TEST(PythonVersionStore, DeleteAllVersions) {
...
 
+TEST(PythonVersionStore, WriteBadStreamId) {
+    using namespace arcticdb;
+    using namespace arcticdb::storage;
+    using namespace arcticdb::stream;
+    using namespace arcticdb::pipelines;
+
+    auto [version_store, mock_store] = python_version_store_in_memory();
+
+    std::string stream_id(1, 0);
```
poodlewars, line 129: I find 

`stream_id{"."}`. Clearer.

I don't think we should validate against ".". We should permit at least all the "safe" characters in https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html and the "Characters that might require special handling" (Man uses some of them widely).

cpp/arcticdb/version/version_map.hpp
```diff
@@ -564,6 +566,25 @@ class VersionMapImpl {
...
+
+        if (std::getenv("ARCTICDB_NO_STRICT_SYMBOL_CHECK"))
+        {
+            ARCTICDB_DEBUG(log::version(), "Key with stream id {} will not be strictly checked because ARCTICDB_NO_STRICT_SYMBOL_CHECK env var is set.", stream_id);
+            return;
+        }
+
+        for (unsigned char c : std::get<std::string>(stream_id)) {
+            if (c < 32 || c > 127) {
+                throw UserInputException(
```
poodlewars, line 582: Yeah, we should add an error code for this (a new User Input error). See `error_messages.md` and have a look at how we implement them in `error_code.hpp`. Use `user_input::raise` to raise it.

python/arcticdb/version_store/_store.py
```diff
@@ -81,6 +80,12 @@
 
 # These chars are encoded by S3 and on doing a list_symbols they will show up as the encoded form eg. &amp
 UNSUPPORTED_S3_CHARS = {"*", "&", "<", ">"}
+
+# If we have enabled STRICT_SYMBOL_CHECK (enabled by default)
+# we will use the SUPPORTED_CHARS set for the check
+# To add more chars as supported, simply add them to the set, e.g. SUPPORTED_CHARS.add('c')
+SUPPORTED_CHARS = set([chr(c) for c in range(1, 128)])
```
poodlewars, line 87: Why do we need this?

poodlewars:
It would be good to do a simple benchmark of writing new symbols before/after this change to check we aren't destroying the performance somehow

cpp/arcticdb/version/test/test_version_store.cpp
G-D-Petrov: This test is for "\0', not ".". But I agree, I will change it to stream_id{"\0"}.

Also all of the "safe" characters have been added to a test to make sure that they are supported.
The test is test_special_chars in test_basic_version_store.py

python/arcticdb/version_store/_store.py
G-D-Petrov: This is a left over, I will remove it

cpp/arcticdb/version/version_map.hpp
```diff
@@ -490,6 +490,8 @@ class VersionMapImpl {
         if (!skip_compat && !has_stored_entry(store, stream_id))
             do_backwards_compat_check(store, stream_id);
 
+        verify_stream_id(store, stream_id);
```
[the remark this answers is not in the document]
G-D-Petrov: Sorry, my mistake, I will add the validation only when version 0 is being written.

cpp/arcticdb/version/test/test_version_store.cpp
poodlewars: But we also need the "Characters that might require special handling" to work

docs/mkdocs/docs/error_messages.md
```diff
@@ -65,6 +65,8 @@ For legacy reasons, the terms `symbol`, `stream`, and `stream ID` are used inter
 | Error Code | Cause                                                                  | Resolution                                                                                                                                          |
 |------------|------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------|
 | 7000       | The input provided by the user is invalid in some fashion.             | The resolution will depend on the nature of the incorrect input, and should be explained in the associated error message. |
+| 7001       | The input was expected to be a valid decimal string but it is not a valid decimal string.             | Pass a valid decimal string. |
+| 7002       | An unsupported character was found in a symbol name.             | We support only the ASCII characters between 32-127 inclussive. Change your symbols name so it contains only valid characters. **If you want to bypass this check, you can define an environment variable called - ARCTICDB_VersionStore_NoStrictSymbolCheck_int=1**. |
```
poodlewars, line 69:
```suggestion
| 7002       | An unsupported character was found in a symbol name.             | We support only the ASCII characters between 32-127 inclusive. Change your symbol name so it contains only valid characters. **If you want to bypass this check, you can define an environment variable called - ARCTICDB_VersionStore_NoStrictSymbolCheck_int=1**.
```

poodlewars, line 69: Worth a test that setting the env var also sets the config int correctly?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -97,7 +98,40 @@ def test_simple_flow(lmdb_version_store_no_symbol_list, symbol):
...
+        "~",
+        "#",
+        "|",
+        "!",
+        "-",
+        "_",
+        ".",
+        "(",
+        ")",
+    ],
```
qc00, line 133: Minor: making use of the fact a Python `str` is a `Sequence`, you can do `list("$@...")`, which is easier to review.

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -160,6 +160,23 @@ inline void check_is_version(const AtomKey &key) {
...
 }
 
+inline void verify_stream_id(const StreamId &stream_id) {
+    if (ConfigsMap::instance()->get_int("VersionStore.NoStrictSymbolCheck"))
+    {
+        ARCTICDB_DEBUG(log::version(), "Key with stream id {} will not be strictly checked because VersionStore.NoStrictSymbolCheck variable is set to 1.", stream_id);
+        return;
+    }
+
+    for (unsigned char c : std::get<std::string>(stream_id)) {
```
poodlewars, line 170: Won't this blow up when the `stream_id` variant is a `NumericId`? I'm not sure how a `NumericId` could get passed in from the Python layer, but since the code has 

```
using NumericId = int64_t;
using StringId = std::string;
using VariantId = std::variant<NumericId, StringId>;
using StreamId = VariantId;
```
we should probably treat `StreamId` as a non-trivial variant everywhere.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -107,6 +141,51 @@ def test_special_chars(s3_version_store, special_char):
...
+@pytest.mark.parametrize("unhandled_char", [chr(0)])
+def test_unhandled_chars_no_strict_check(s3_version_store, unhandled_char):
+    """Test that when we turn the STRICT_SYMBOL_CHECK off, the problematic \x00 is raising an exception"""
+    set_config_int("VersionStore.NoStrictSymbolCheck", 1)
+    sym = f"prefix{unhandled_char}postfix"
+    df = sample_dataframe()
+    s3_version_store.write(sym, df)
+    with pytest.raises(InternalException):
+        s3_version_store.list_symbols()
+    unset_config_int("VersionStore.NoStrictSymbolCheck")
```
qc00, line 176: There's an `arcticdb.util.test.config_context` which does this in an exception-safe way

docs/mkdocs/docs/error_messages.md
G-D-Petrov: yep, tested it, it works

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -160,6 +160,29 @@ inline void check_is_version(const AtomKey &key) {
...
 }
 
+inline void verify_stream_id(const StreamId &stream_id) {
+    if (ConfigsMap::instance()->get_int("VersionStore.NoStrictSymbolCheck"))
+    {
+        ARCTICDB_DEBUG(log::version(), "Key with stream id {} will not be strictly checked because VersionStore.NoStrictSymbolCheck variable is set to 1.", stream_id);
+        return;
+    }
+
+    if (std::holds_alternative<NumericId>(stream_id))
```
poodlewars, line 170: Minor: Slightly prefer `util::variant_match` for the variant handling so that we get a compilation error if `StreamId` expands to include another type. That seems vanishingly unlikely for this case though, so fine to leave as-is, up to you.

poodlewars (approved):

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -107,6 +109,107 @@ def test_special_chars(s3_version_store, special_char):
...
+
+
+@pytest.mark.parametrize("unhandled_char", [chr(0)])
+def test_unhandled_chars_no_strict_check(s3_version_store, unhandled_char):
+    """Test that when we turn the STRICT_SYMBOL_CHECK off, the problematic \x00 is raising an exception"""
+    with config_context("VersionStore.NoStrictSymbolCheck", 1):
+        sym = f"prefix{unhandled_char}postfix"
+        df = sample_dataframe()
+        s3_version_store.write(sym, df)
+        with pytest.raises(InternalException):
```
qc00, line 142: My understanding of the `NoStrictSymbolCheck` feature is to allow characters that may work just fine (on some storages) but we don't want to officially support.

However, this test reveals that `\0` should be another (always) unsupported character just like `&*<>`. I can't imagine any user would be happy with writing a symbol that will break `list_symbols()`.

G-D-Petrov: Yes, indeed, I am also in favor of putting it in the unsupported symbols like &*<>.
@poodlewars do you agree?

poodlewars: Yep totally agree, thanks Qi

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -160,6 +160,33 @@ inline void check_is_version(const AtomKey &key) {
...
+    if (ConfigsMap::instance()->get_int("VersionStore.NoStrictSymbolCheck"))
+    {
+        ARCTICDB_DEBUG(log::version(), "Key with stream id {} will not be strictly checked because VersionStore.NoStrictSymbolCheck variable is set to 1.", stream_id);
+        return;
+    }
+
+    util::variant_match(
+        stream_id,
+        [] (const NumericId& num_stream_id) {
+            (void)num_stream_id; // Suppresses -Wunused-parameter
```
qc00, line 173: Minor: if you remove the parameter name from the lambda (but leave the type), the compiler won't warn about it.

willdealtry: The way we generally handle this is to use the macro ARCTICDB_UNUSED, which gives you the appropriate attribute for the platform

PR #629, opened 2023-07-20
Enhancement/612/read batch to catch all exceptions

### LINKED ISSUE #612: read_batch should catch all exceptions from individual queries and return what data it can
#531 covers missing version-style errors

This should extend that implementation to cover other errors, such as missing index keys, missing data keys, etc

alexowens90:
Closes #612

The previous behaviour of `Library.read_batch` and `NativeVersionStore.batch_read` was to throw an exception if there was an issue retrieving **any** of the requested symbol/version pairs. For backwards compatibility reasons, this behaviour has been maintained for `NativeVersionStore.batch_read`.

For `Library.read_batch`, this has been changed. A `DataError` object will now be returned in the position in the list returned corresponding to the symbol/version pair there was an issue retrieving. This contains the symbol and version requested, and the exception string thrown. For two well-defined categories of error, the error category and specific error code are also included:

- ErrorCategory.MISSING_DATA - ErrorCode.E_NO_SUCH_VERSION: The version requested by the user does not exist
- ErrorCategory.STORAGE - ErrorCode.E_KEY_NOT_FOUND: At least one of the keys required to read the specified version does not exist in the storage.

Otherwise these fields of the `DataError` object are left as `None`.

### DIFF
--- a/cpp/arcticdb/entity/data_error.hpp
+++ b/cpp/arcticdb/entity/data_error.hpp
@@ -1,0 +1,121 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <string>
+
+#include <fmt/format.h>
+
+#include <arcticdb/pipeline/query.hpp>
+
+namespace arcticdb::entity {
+
+enum class VersionRequestType: uint32_t {
+    SNAPSHOT,
+    TIMESTAMP,
+    SPECIFIC,
+    LATEST
+};
+
+class DataError {
+public:
+    DataError(const StreamId& symbol,
+              const pipelines::VersionQueryType& version_query_type,
+              std::string&& exception_string,
+              std::optional<ErrorCode> error_code=std::nullopt) :
+        symbol_(symbol),
+        exception_string_(exception_string),
+        error_code_(error_code){
+        util::variant_match(
+                version_query_type,
+                [this] (const pipelines::SnapshotVersionQuery& query) {
+                    version_request_type_ = VersionRequestType::SNAPSHOT;
+                    version_request_data_ = query.name_;
+                },
+                [this] (const pipelines::TimestampVersionQuery& query) {
+                    version_request_type_ = VersionRequestType::TIMESTAMP;
+                    version_request_data_ = query.timestamp_;
+                },
+                [this] (const pipelines::SpecificVersionQuery& query) {
+                    version_request_type_ = VersionRequestType::SPECIFIC;
+                    version_request_data_ = query.version_id_;
+                },
+                [this] (const std::monostate&) {
+                    version_request_type_ = VersionRequestType::LATEST;
+                }
+        );
+    }
+
+    DataError() = delete;
+
+    ARCTICDB_MOVE_COPY_DEFAULT(DataError);
+
+    void set_error_code(ErrorCode error_code) {
+        error_code_ = error_code;
+    }
+
+    std::string symbol() const {
+        return fmt::format("{}", symbol_);
+    }
+
+    VersionRequestType version_request_type() const {
+        return version_request_type_;
+    }
+
+    std::variant<std::monostate, int64_t, std::string> version_request_data() const {
+        return version_request_data_;
+    }
+
+    std::optional<ErrorCode> error_code() const {
+        return error_code_;
+    }
+
+    std::optional<ErrorCategory> error_category() const {
+        return error_code_.has_value() ? std::make_optional<ErrorCategory>(get_error_category(*error_code_)) : std::nullopt;
+    }
+
+    std::string exception_string() const {
+        return exception_string_;
+    }
+
+    std::string to_string() const {
+        std::string version_request_explanation;
+        switch (version_request_type_) {
+            case VersionRequestType::SNAPSHOT:
+                version_request_explanation = fmt::format("in snapshot '{}'", std::get<std::string>(version_request_data_));
+                break;
+            case VersionRequestType::TIMESTAMP:
+                version_request_explanation = fmt::format("at time '{}'", std::get<int64_t>(version_request_data_));
+                break;
+            case VersionRequestType::SPECIFIC:
+                version_request_explanation = fmt::format("with specified version '{}'", std::get<int64_t>(version_request_data_));
+                break;
+            case VersionRequestType::LATEST:
+                version_request_explanation = fmt::format("with specified version 'latest'");
+                break;
+            default:
+                internal::raise<ErrorCode::E_ASSERTION_FAILURE>(
+                        "Unexpected enum value in DataError::to_string: {}",
+                        static_cast<uint32_t>(version_request_type_));
+        }
+        return fmt::format(
+                "Version {} of symbol '{}' unable to be retrieved: Error Category: '{}' Exception String: '{}'",
+                version_request_explanation,
+                symbol_,
+                error_category().has_value() ? get_error_category_names().at(*error_category()) : "UNKNOWN",
+                exception_string_);
+    }
+private:
+    StreamId symbol_;
+    VersionRequestType version_request_type_{VersionRequestType::LATEST};
+    // int64_t for timestamp and SignedVersionId
+    std::variant<std::monostate, int64_t, std::string> version_request_data_;
+    std::string exception_string_;
+    std::optional<ErrorCode> error_code_;
+};
+}
--- a/cpp/arcticdb/pipeline/query.hpp
+++ b/cpp/arcticdb/pipeline/query.hpp
@@ -110,8 +110,15 @@ struct SpecificVersionQuery {
     SignedVersionId version_id_;
 };
 
+using VersionQueryType = std::variant<
+        std::monostate, // Represents "latest"
+        SnapshotVersionQuery,
+        TimestampVersionQuery,
+        SpecificVersionQuery
+        >;
+
 struct VersionQuery {
-    std::variant<std::monostate, SnapshotVersionQuery, TimestampVersionQuery, SpecificVersionQuery> content_;
+    VersionQueryType content_;
     std::optional<bool> skip_compat_;
     std::optional<bool> iterate_on_failure_;
 
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -120,7 +120,7 @@ FrameAndDescriptor LocalVersionedEngine::read_column_stats_internal(
     return read_column_stats_impl(store(), versioned_item);
 }
 
-std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_column_stats_version_internal(
+ReadVersionOutput LocalVersionedEngine::read_column_stats_version_internal(
     const StreamId& stream_id,
     const VersionQuery& version_query) {
     auto versioned_item = get_version_to_read(stream_id, version_query);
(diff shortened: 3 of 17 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -344,16 +344,17 @@ std::pair<VersionedItem, FrameAndDescriptor> LocalVersionedEngine::read_datafram
         if(opt_false(read_options.incompletes_)) {
             log::version().warn("No index:  Key not found for {}, will attempt to use incomplete segments.", stream_id);
             identifier = stream_id;
+        } else {
+            missing_data::raise<ErrorCode::E_NO_SUCH_VERSION>(
+                    "read_dataframe_version: version query '{}' not found for symbol '{}'", version_query, stream_id);
```
mehertz, line 349: How can you not find a query? Does this mean no version matching the query or something like that?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1028,33 +1029,64 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
...
-std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::temp_batch_read_internal_direct(
+std::vector<std::variant<ReadVersionOutput, DataError>> LocalVersionedEngine::temp_batch_read_internal_direct(
     const std::vector<StreamId> &stream_ids,
     const std::vector<VersionQuery> &version_queries,
     std::vector<ReadQuery> &read_queries,
     const ReadOptions &read_options) {
     py::gil_scoped_release release_gil;
+    // This read option should always be set when calling batch_read
+    internal::check<ErrorCode::E_ASSERTION_FAILURE>(read_options.batch_throw_on_missing_version_.has_value(),
+                                                    "ReadOptions::batch_throw_on_missing_version_ should always be set here");
```
mehertz, line 1040: Wasn't this not going to be set from `_store`?

cpp/arcticdb/entity/data_error.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <string>
+
+#include <fmt/format.h>
+
+#include <arcticdb/pipeline/query.hpp>
+
+namespace arcticdb::entity {
+
+enum class VersionRequestType: uint32_t {
+    SNAPSHOT,
```
willdealtry, line 19: This is quite similar to the LoadType enum, do they need to be kept in sync with each other for ever? Might it be worth having a function that converts from one to the other, i.e. from the thing we tried to the class of operation that failed?

cpp/arcticdb/pipeline/query.hpp
```diff
@@ -110,8 +110,15 @@ struct SpecificVersionQuery {
     SignedVersionId version_id_;
 };
 
+using VersionQueryType = std::variant<
```
willdealtry, line 113: I like this, it is definitely better

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -273,18 +274,18 @@ class LocalVersionedEngine : public VersionedEngine {
...
         bool validate_index);
 
-    std::vector<std::pair<VersionedItem, FrameAndDescriptor>> batch_read_keys(
+    std::vector<ReadVersionOutput> batch_read_keys(
         const std::vector<AtomKey> &keys,
         const std::vector<ReadQuery> &read_queries,
         const ReadOptions& read_options);
 
-    std::vector<std::pair<VersionedItem, FrameAndDescriptor>> batch_read_internal(
+    std::vector<std::variant<ReadVersionOutput, DataError>> batch_read_internal(
```
willdealtry, line 282: Did you consider std::expected?

cpp/arcticdb/version/versioned_engine.hpp
```diff
@@ -23,6 +23,18 @@ namespace arcticdb::version_store {
 using namespace arcticdb::entity;
 using namespace arcticdb::pipelines;
 
+struct ReadVersionOutput {
```
willdealtry, line 26: Nicer as a class than pair I think, we should make this the general pattern

willdealtry (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1066,16 +1098,36 @@ std::vector<std::pair<VersionedItem, FrameAndDescriptor>> LocalVersionedEngine::
...
+    read_versions_or_errors.reserve(stream_ids.size());
     for (size_t idx=0; idx < stream_ids.size(); idx++) {
         auto version_query = version_queries.size() > idx ? version_queries[idx] : VersionQuery{};
         auto read_query = read_queries.size() > idx ? read_queries[idx] : ReadQuery{};
-        results_fut.push_back(read_dataframe_version_internal(stream_ids[idx],
-                                                              version_query,
-                                                              read_query,
-                                                              read_options));
+        // TODO: https://github.com/man-group/ArcticDB/issues/241
+        // Remove this try-catch in favour of the implementation in temp_batch_read_internal_direct as part of #241
```
mehertz, line 1107: 241 will remove temp_batch_read_internal_direct surely?

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -130,6 +132,20 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
...
+            .value("LATEST", VersionRequestType::LATEST);
+
+    py::class_<DataError, std::shared_ptr<DataError>>(version, "DataError")
+            .def_property_readonly("symbol", &DataError::symbol)
+            .def_property_readonly("version_request_type", &DataError::version_request_type)
+            .def_property_readonly("version_request_data", &DataError::version_request_data)
+            .def_property_readonly("error_code", &DataError::error_code)
+            .def_property_readonly("error_category", &DataError::error_category)
+            .def_property_readonly("exception_string", &DataError::exception_string)
+            .def("__repr__", &DataError::to_string);
```
mehertz, line 148: This isn't a `__repr__`. `__repr__` would print something like `DataError(error_code=..., error_category=...)`.

I think this should be `__str__`.

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -1235,6 +1236,163 @@ def test_read_batch_unhandled_type(arctic_library):
...
+    assert batch[1].error_category == ErrorCategory.MISSING_DATA
+
+    assert isinstance(batch[2], DataError)
+    assert batch[2].symbol == "s2"
+    assert batch[2].version_request_type == VersionRequestType.SPECIFIC
+    assert batch[2].version_request_data == 1
+    assert batch[2].error_code == ErrorCode.E_NO_SUCH_VERSION
+    assert batch[2].error_category == ErrorCategory.MISSING_DATA
+
+
```
mehertz, line 1395: What if the vref key exists but the version key doesn't?

cpp/arcticdb/version/local_versioned_engine.cpp
mehertz: oh, does that never call into here somehow?

cpp/arcticdb/entity/data_error.hpp
mehertz: I was a bit concerned about this as well. Not a blocker but good if we can keep them in sync somehow.

cpp/arcticdb/version/local_versioned_engine.hpp
poodlewars: That's c++ 23, no?

poodlewars (approved):
Assume the other batch read methods (metadata, description, yadayada, are going to change to match this pattern soon too?

python/arcticdb/version_store/library.py
```diff
@@ -876,8 +876,10 @@ def read_batch(
 
         Returns
         -------
-        List[VersionedItem]
+        List[Union[VersionedItem, DataError]]
```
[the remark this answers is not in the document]
alexowens90: I can't really think of a good name that isn't similar to `VersionedItemOrDataError`, at which point it doesn't help. Open to ideas though.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -189,13 +205,22 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
 
 
 
-    auto adapt_read_dfs = [](std::vector<ReadResult> && ret) -> py::list {
+    auto adapt_read_dfs = [](std::vector<std::variant<ReadResult, DataError>> && ret) -> py::list {
```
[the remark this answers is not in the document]
alexowens90: Same point as in the Python layer, I think a variant with two elements is clearer than a name that just includes both somehow

alexowens90:
> Assume the other batch read methods (metadata, description, yadayada, are going to change to match this pattern soon too?

Yes, @GemDot is picking those up after this is merged.

python/tests/integration/arcticdb/test_arctic.py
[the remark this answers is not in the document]
alexowens90: This isn't trivial. We don't have a way to either generate the S3 object path from the `AtomKey`, or write arbitrary objects from the Python layer.

alexowens90: Added a test for this and made a code change so that it correctly returns an `E_KEY_NOT_FOUND` error

cpp/arcticdb/version/python_bindings.cpp
alexowens90: Changed

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Yes, but the resulting code will have a `folly::collectAll` in it like `temp_batch_read_internal_direct`, rather than this messy loop-try-catch

cpp/arcticdb/version/local_versioned_engine.hpp
alexowens90: Yeh C++23, nice to know the patterns we're working around not existing are coming though.

cpp/arcticdb/entity/data_error.hpp
alexowens90: It looks more like the `VersionQueryType` I added. It would be better if all those classes were exposed to Python, instead of the `set_*` methods on `VersionQuery` right now, and we could just return one of those combining the `version_request_type` and `version_request_data` return parts right now.

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Its always set to true by `NativeVersionStore` right now

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Yes that's what it means

mehertz: ...can we change the wording to make that clearer?

python/tests/integration/arcticdb/test_arctic.py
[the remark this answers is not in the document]
mehertz: That's a shame. We'll need a way to do this in the future otherwise we can't test error conditions where the data is just plain broken, rather than missing.

cpp/arcticdb/version/local_versioned_engine.cpp
mehertz: Shouldn't this check also be part of the non temp-batch read flow?

mehertz: In fact, maybe this check should be just in `batch_read_internal`?

alexowens90: Yeh I noticed this as well, have moved it

python/arcticdb/version_store/library.py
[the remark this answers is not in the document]
mehertz: I'm not sure we need an alias to be honest. Makes it harder to read IMO.

mehertz:
Closes #531

PR #540, opened 2023-07-03
maint: pandas 2.0 forward compatible changes

jjerphan:
#### Reference Issues/PRs

Fixes #237.
Follow-up of #343.

#### What does this implement/fix? Explain your changes.

This PR brings support for pandas 2.0. by:
 - removing bound in environment specifications
 - adapting the normalization and the denormalization of `pandas.Series` and `pandas.DataFrame`
 - adapting tests for changes of behavior due to the new defaults of pandas 2.0 

#### Any other comments?

We need to discuss how to properly support some changes of behavior of pandas with respect to:
 - [the specification of the resolution for the `datetime64` or `timedelta64`](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#construction-with-datetime64-or-timedelta64-dtype-with-unsupported-resolution)
 - [`datetime64` and `timedelta64` conversion for their resolution](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#disallow-astype-conversion-to-non-supported-datetime64-timedelta64-dtypes) introduced by https://github.com/pandas-dev/pandas/issues/48928
 
Edges cases which I think can be treated subsequently:
 - https://github.com/man-group/ArcticDB/pull/540/files#r1265126115
 - https://github.com/man-group/ArcticDB/pull/540/files#r1288498933
 - https://github.com/man-group/ArcticDB/pull/540/files#r1273260829
 - https://github.com/man-group/ArcticDB/pull/540/files#r1288531578
 
See `NOTE` and `TODO` in the diff of this PR.

Issue #224, opened 2023-04-12
Writing 0-row DataFrame: correctness and usability


qc00:
As exposed by `test_normalization.py::test_store_df[empty_midx]`

Internal thread: $.slack.com/archives/C02NEG3V38X/p1678796229175589

Pandas (rightly) changed the default type of a 0-row index to `object` which (rightly) triggers a normalization error:

> ArcticNativeNotYetImplemented: coercing column type is required when empty column of object type, Column type=object for column=...

1. Should provide guidance in the error message to explicitly set the type of the index OR special handling to allow it

2. Due to #223, above exception actually cause 0-row DF to be pickled

phoebusm:
Is it related to https://github.com/man-group/ArcticDB/issues/90?

qc00:
Yes, #90 is a dup of point 2 above.

jjerphan:
> Should provide guidance in the error message to explicitly set the type of the index OR special handling to allow it

I think users should not have to set it manually, and ArcticDB rather has to identify such case and coerce those empty `pd.Series` and `pd.DataFrame` to a relevant representations.

For empty column, we could change the newly used `dtype=object` to the previously used `dtype=float`, yet we might or will be blocked if other items than `float` are appended to the column (see https://github.com/man-group/ArcticDB/pull/540#discussion_r1265126115 which discusses it).

Is there a way to change the type of an (empty) column?

qc00:
> Is there a way to change the type of an (empty) column?

That's the crux of the problem. The current stream descriptor handling does not allow column type change (apart from mixing numerical types under dynamic schema).

We would need to change the C++ to support a special undefined type that can be changed into any other type in a subsequent data key.

willdealtry:
The changes that @vasil-pashov are moving over to support Decimals, numpy arrays etc will introduce a new NoneType which is promotable to anything, I suggest we wait until those are merged and then see if they do what we need for this case.

jjerphan:
Empty `pandas.DataFrame` and empty `pandas.Series` columns types might be set differently by Pandas depending on its version. They are generally also reset by pandas if the users provide them at creation.

Shouldn't there be a way to have users explicitly provide types for a symbol's schema to one of ArcticDB's API, be it a backed by an empty `pandas.DataFrame` or an empty `pandas.Series`, rather than automatically assuming the types of the schema without being able to guess if they have been provided by users or changed by Pandas in between?

jjerphan:
[...]

PR #646, opened 2023-07-24
Add none type

### LINKED ISSUE #639: Add C++ support for None type
In order to complete #540 we would need to change the C++ to support a special undefined type that can be changed into any other type in a subsequent data key. See also the discussion in #240.

Add a none type in the C++ layer

**Additional context**

vasil-pashov:
#### Reference Issues/PRs
Resolves #639
Relates to: #540 #224 

Change the C++ to support a special undefined type that can be changed into any other type in a subsequent data key.

#### Any other comments?
Only the C++ layers is changed, it might need additional python layer integration.


mehertz (changes requested):
image

Don't leave blank please!

cpp/arcticdb/python/python_handlers.hpp
```diff
@@ -0,0 +1,21 @@
...
+        void handle_type(
+                const uint8_t*& data,
+                uint8_t* dest,
+                size_t dest_bytes,
+                const VariantField& encoded_field,
+                const entity::TypeDescriptor& type_descriptor,
+                std::shared_ptr<BufferHolder> buffers
+        );
+    };
+} //namespace arcticc
```
jjerphan, line 21:
```suggestion
} // namespace arcticdb
```

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -0,0 +1,24 @@
+#include <python/python_handlers.hpp>
```
jjerphan, line 1: @mehertz: what is Man Group's policy regarding BSL headers? Must they always be present in header and source files?

jjerphan (commented):
Thank you for working on it, @vasil-pashov.

Here is a first pass of suggestions.

cpp/arcticdb/pipeline/frame_utils.hpp
```diff
@@ -149,6 +149,8 @@ void aggregator_set_data(
                     }
                 }
             }
+        } else if constexpr (is_empty_type(dt)) {
```
alexowens90, line 152: I think the `else` represents all numeric types? In which case I think it would be more explicit to have `else if constexpr (is_numeric_type(dt))`

cpp/arcticdb/processing/test/test_has_valid_type_promotion.cpp
```diff
@@ -121,3 +121,14 @@ TEST(HasValidTypePromotion, FloatFloat) {
     EXPECT_FALSE(has_valid_type_promotion(float64, float32));
 }
 
+TEST(HasValidTypePromotion, EmptyToEverything) {
```
alexowens90, line 124: An equivalent `EverythingToEmpty` test would also be useful

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -0,0 +1,24 @@
...
+            size_t dest_bytes,
+            const VariantField& encoded_field_info ARCTICDB_UNUSED,
+            const entity::TypeDescriptor& type_descriptor ARCTICDB_UNUSED,
+            std::shared_ptr<BufferHolder> buffers ARCTICDB_UNUSED
+    ) {
+        ARCTICDB_SAMPLE(HandleEmpty, 0)
+        util::check(dest != nullptr, "Got null destination pointer");
+        py::none none = {};
+        const size_t num_rows = dest_bytes / get_type_size(DataType::EMPTYVAL);
+        const PyObject** ptr_dest = reinterpret_cast<const PyObject**>(dest);
```
alexowens90, line 17: There is an assumption here that `sizeof(PyObject**) == 8`, so we should at least `internal::check` that

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -76,10 +76,13 @@ NativeTensor obj_to_tensor(PyObject *ptr) {
                 empty = casted_col == obj + size;
                 sample = *casted_col;
             }
-            if (empty || is_unicode(sample)) {
+            if (is_unicode(sample)) {
                 val_type = ValueType::UTF_DYNAMIC;
-            } else if (PYBIND11_BYTES_CHECK(sample))
+            } else if(empty){
+                val_type = ValueType::EMPTY;
```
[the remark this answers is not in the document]
alexowens90: Agreed, it looks like the empty check should come first

alexowens90 (changes requested):

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -0,0 +1,24 @@
...
+        ScopedGILLock lock;
+        for(auto row = 0u; row < num_rows; ++row) {
+            none.inc_ref();
+            *ptr_dest++ = none.ptr();
+        }
+    }
+}
```
qc00, line 24: New line at the end of every file, please.

cpp/arcticdb/python/python_to_tensor_frame.cpp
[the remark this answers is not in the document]
qc00: Yeah, would prefer checking `empty` first and leave the two string branches close together.

cpp/arcticdb/entity/types.hpp
```diff
@@ -115,6 +115,8 @@ enum class ValueType : uint8_t {
 
     UTF_DYNAMIC = 11,
     ASCII_DYNAMIC = 12,
+    EMPTY = 13,
+    COUNT // Not a real value type, should not be added to proto descriptor. Used to count the number of items in the enum
```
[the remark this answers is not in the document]
vasil-pashov: Currently only in the tests.

cpp/arcticdb/python/python_module.cpp
```diff
@@ -253,6 +256,9 @@ void register_metrics(py::module && m){
         arcticdb::PrometheusConfigInstance::instance()->config.CopyFrom(config);
     });
 }
+void register_type_handlers() {
+    arcticdb::TypeHandlerRegistry::instance()->register_handler(arcticdb::DataType::EMPTYVAL, arcticdb::EmptyHandler());
+}
```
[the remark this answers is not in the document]
vasil-pashov: The idea of the type handler utility is to register only non-trivial types. Currently the EMPTYVAL (None) type is the only non-trivial. The decimal and the array type would be the other non-trivial types. While floats and ints are trivial to read from python.

cpp/arcticdb/python/python_to_tensor_frame.cpp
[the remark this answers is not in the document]
vasil-pashov: Will change it for readability purpose. I don't think that it will have any perf implications, branch prediction is pretty good nowadays.

cpp/arcticdb/entity/types.hpp
```diff
@@ -115,6 +115,8 @@ enum class ValueType : uint8_t {
 
     UTF_DYNAMIC = 11,
     ASCII_DYNAMIC = 12,
+    EMPTY = 13,
```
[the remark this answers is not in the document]
vasil-pashov: It is in fact equivalent to None type in Python. I will add a comment in code. @willdealtry used this notation (Empty type/EmptyVal) so I went along with it as well. Happy to change it if people think it would be more readable otherwise.

cpp/arcticdb/python/python_handlers.cpp
vasil-pashov: @jjerphan what is a "BSL header"?

cpp/arcticdb/processing/test/test_has_valid_type_promotion.cpp
vasil-pashov: Do we want everything to be convertible to None? I'm not sure how sensible is this.

cpp/arcticdb/python/python_handlers.cpp
jjerphan: BSL stands for "Business Software License", [the license used by ArcticDB](https://github.com/man-group/ArcticDB/blob/master/LICENSE.txt).

Generally each file has an header in their first lines. See for instance:
https://github.com/man-group/ArcticDB/blob/bf9a8310bfcb7aa484778fd1ba764a2250f9b429/cpp/arcticdb/pipeline/frame_utils.hpp#L1-L6

cpp/arcticdb/python/python_module.cpp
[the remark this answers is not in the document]
jjerphan: Could you add this indication as a (d0c) comment of `register_type_handlers`?

cpp/arcticdb/python/python_to_tensor_frame.cpp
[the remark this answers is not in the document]
jjerphan: (My original thought was more that `is_unicode` might be called needlessly; that's a nitpick).

cpp/arcticdb/python/python_handlers.cpp
vasil-pashov: Any particular reason to do so?

cpp/arcticdb/entity/types-inl.hpp
```diff
@@ -38,6 +38,7 @@ auto visit_dim(DataType dt, Callable &&c) {
         DT_CASE(ASCII_DYNAMIC64)
         DT_CASE(UTF_FIXED64)
         DT_CASE(UTF_DYNAMIC64)
+        DT_CASE(EMPTYVAL)
```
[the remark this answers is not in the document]
willdealtry: Although in the case of the others, it's mostly the separator between the data type and the size, which doesn't apply in this case

cpp/arcticdb/entity/types.hpp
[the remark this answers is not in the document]
willdealtry: Could be NoneType I suppose. The purpose is to indicate that there's a column there but it doesn't have a defined type in a particular segment. In the case of Python, it's probably that all of the elements are None, but it's intended to be more generally applicable.

cpp/arcticdb/entity/types.hpp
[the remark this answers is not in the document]
willdealtry: Kind of a common pattern to be able to say "Tell me how many things are in this enum"

cpp/arcticdb/pipeline/frame_utils.hpp
willdealtry: Are all of the none-string types numeric types, and is this a logical fact or just a coincidental thing at the moment?

cpp/arcticdb/processing/test/test_has_valid_type_promotion.cpp
willdealtry: I think so, because if a column has a type already in some other dataframe, but this one is complelely empty, then that's fine, it's just still the same type as it was before

cpp/arcticdb/python/python_handlers.cpp
willdealtry: Do it at compile time?

vasil-pashov: It's ratther `sizeof(PyObject*) == 8` I'll add a `static_assert`

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -0,0 +1,24 @@
...
+            const uint8_t*& data ARCTICDB_UNUSED,
+            uint8_t* dest,
+            size_t dest_bytes,
+            const VariantField& encoded_field_info ARCTICDB_UNUSED,
+            const entity::TypeDescriptor& type_descriptor ARCTICDB_UNUSED,
+            std::shared_ptr<BufferHolder> buffers ARCTICDB_UNUSED
+    ) {
+        ARCTICDB_SAMPLE(HandleEmpty, 0)
+        util::check(dest != nullptr, "Got null destination pointer");
+        py::none none = {};
```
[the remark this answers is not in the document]
vasil-pashov: I'll move this just after the line acquiring the lock just to avoid confusion.

cpp/arcticdb/processing/test/test_has_valid_type_promotion.cpp
alexowens90: Nothing should be promotable to None. `has_valid_common_type` is used if you want to check both ways round.

cpp/arcticdb/pipeline/frame_utils.hpp
alexowens90: Until now, yes, coincidentally all non-numeric types were string types. @vasil-pashov please do a scan of the code to see if this assumption is made elsewhere, particularly in the filtering/projection code.

vasil-pashov: The `bool` type is neither string nor numeric type. However up to now it was processed by the `else` statement in this snippet.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -411,68 +415,38 @@ class AsyncStore : public Store {
...
+                    })
+                    .via(&async::io_executor()).thenValue([lib = library_](auto &&item) {
+                        auto key_opt_segment = std::forward<decltype(item)>(item);
+                        if (key_opt_segment.second)
+                            lib->write(Composite<storage::KeySegmentPair>({VariantKey{key_opt_segment.first},
+                                                                           std::move(*key_opt_segment.second)}));
+
+                        return key_opt_segment.first;
+                    }));
         }
```
vasil-pashov, line 447: Not part of this commit, result of a rebase.

willdealtry (changes requested):

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -338,8 +352,8 @@ void decode_into_frame_static(
                         it.dest_col(),
                         m.frame_field_descriptor_.name());
 
-            util::check(data != end, "Reached end of input block with {} fields to decode", it.remaining_fields());
-            decode_or_expand(data, buffer.data() + m.offset_bytes_, encoded_field, m.source_type_desc_,  m.dest_bytes_, buffers);
+            util::check(data != end || remaining_fields_empty(it, context), "Reached end of input block with {} fields to decode", it.remaining_fields());
```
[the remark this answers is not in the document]
vasil-pashov: I've seen that they were not used at all in the current code and found that a bit confusing. I didn't know that we'll need them for the arrays. If it's OK I'll leave it as is and add them again when we need them.

cpp/arcticdb/util/buffer.hpp
```diff
@@ -213,36 +214,36 @@ struct Buffer : public BaseBuffer<Buffer, true> {
...
     }
 
   private:
-    inline void resize(size_t bytes) {
-        if (bytes == 0) {
-            reset();
+    inline void resize(size_t alloc_bytes) {
+        const size_t bytes = alloc_bytes - preamble_bytes_;
+#ifdef DEBUG_BUILD
+        util::check(alloc_bytes >= preamble_bytes_, "The requested size of a resizes call is less than the preamble bytes");
```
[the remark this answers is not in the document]
vasil-pashov: I agree. There is a function `Buffer::check_invariants` which wraps the checks in an #ifdef and I made this to be in line with it.

cpp/arcticdb/entity/type_utils.hpp
```diff
@@ -12,8 +12,13 @@
 namespace arcticdb {
 
 inline bool trivially_compatible_types(entity::TypeDescriptor left, entity::TypeDescriptor right) {
-    if(left == right)
+    if(left == right) {
```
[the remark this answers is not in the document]
vasil-pashov: While I don't want to enforce my personal preferences I strongly believe we need some sort of coding standard for this. Homogeneous code is easier to read. Currently if different people touch the same function it's possible to end up with this:
```c++
if(condition) return true;
...
if(other_condition)
    return true;
...
if(third_condition) {
    return true;
}
```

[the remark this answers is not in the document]
vasil-pashov: Here is a snippet from [our code](https://github.com/man-group/ArcticDB/blob/d6b8c2a0eb00163942c6e97c7d10a628e00bbbbf/cpp/arcticdb/column_store/memory_segment_impl.hpp#L287)

```c++
                if constexpr (T::DimensionTag::value == Dimension::Dim0) {
                    if constexpr (is_sequence_type(T::DataTypeTag::data_type)) {
                        // test only for now
                        throw std::runtime_error("string type not implemented");
                    } else {
                        if constexpr(std::is_same_v<RawType, S>)
                        val = that->parent_->scalar_at<RawType>(that->row_id_, col);
                        else
                            util::raise_rte("Type mismatch in scalar access");
                    }
                } else {
                    throw std::runtime_error("Scalar method called on multidimensional column");
                }
```
Here we have if/else wrapped in block, if else without block and to spice things up different indentation of the codepaths

vasil-pashov:
This is a list of functions which assume that if something is not string it's number of vice versa:

```c++
VariantData binary_membership(const ColumnWithStrings& column_with_strings, ValueSet& value_set, Func&& func)
VariantData binary_comparator(const Value& val, const ColumnWithStrings& column_with_strings, Func&& func)
VariantData binary_comparator(const ColumnWithStrings& left, const ColumnWithStrings& right, Func&& func)
VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const Value& val, Func&& func)
SegmentInMemoryImpl::visit_field(Callable &&c)
SegmentInMemoryImpl::operator==(const SegmentInMemoryImpl& left, const SegmentInMemoryImpl& right)
FrameSliceMap::FrameSliceMap(const std::shared_ptr<PipelineContext>& context, bool dynamic_schema)
std::shared_ptr<SegmentInMemoryImpl> filter(const util::BitSet& filter_bitset,
                                                       bool filter_down_stringpool=false,
                                                       bool validate=false) const {
Value(const Value& other)
void SumAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values)
void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column)
void MaxOrMinAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values)
VariantData binary_membership(const ColumnWithStrings& column_with_strings, ValueSet& value_set, Func&& func) {
VariantData binary_comparator(const Value& val, const ColumnWithStrings& column_with_strings, Func&& func) {
VariantData binary_comparator(const ColumnWithStrings& left, const ColumnWithStrings& right, Func&& func) {
VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const Value& val, Func&& func) {
inline StreamId stream_id_from_segment(const SegmentInMemory &seg, ssize_t row) {
folly::Future<arcticdb::entity::VariantKey> write_frame_data(const TestDataFrame &data_frame, WriterType &writer) {
```
they all have contain if/else case similar to:
```
if(is_sequence_type(dt)) {
...
} else {
...
}
```

or

```
if(is_numeric_type(dt)) {
...
} else {
...
}
```

Probably for some of them it's fine to leave it as is, but we might have problems down the road when none type and other types are added.

cpp/arcticdb/python/python_module.cpp
```diff
@@ -25,10 +25,12 @@
 #include <arcticdb/util/global_lifetimes.hpp>
 #include <arcticdb/util/configs_map.hpp>
 #include <arcticdb/util/error_code.hpp>
+#include <arcticdb/python/python_handlers.hpp>
 
 #include <pybind11/pybind11.h>
 #include <folly/system/ThreadName.h>
 #include <folly/portability/PThread.h>
+#include "util/type_handler.hpp"
```
jjerphan, line 33: For consistency, should this be?
```suggestion
#include <arcticdb/util/type_handler.hpp>
#include <arcticdb/python/python_handlers.hpp>

#include <pybind11/pybind11.h>
#include <folly/system/ThreadName.h>
#include <folly/portability/PThread.h>
```

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -68,19 +68,34 @@ NativeTensor obj_to_tensor(PyObject *ptr) {
...
+                    }
+                    ++current_object;
+                }
+                sample = *current_object;
             }
-            if (empty || is_unicode(sample)) {
+            if (empty) {
+                val_type = ValueType::EMPTY;
+            } else if(all_nans || is_unicode(sample)){
                 val_type = ValueType::UTF_DYNAMIC;
```
jjerphan, line 95: Could you document why `UTF_DYNAMIC` is used if all values are nans (`all_nans`)?

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -68,19 +68,34 @@ NativeTensor obj_to_tensor(PyObject *ptr) {
...
             if (sample == none.ptr() || is_py_nan(sample)) {
-                // Iterate till we find the first non null element, the frontend ensures there is at least one.
+                empty = true;
+                all_nans = true;
                 util::check(c_style, "Non contiguous columns with first element as None not supported yet.");
-                auto casted_col =
-                    std::find_if(obj, obj + size, [&none](auto val) { return val != none.ptr() && !is_py_nan(val); });
-                empty = casted_col == obj + size;
-                sample = *casted_col;
+                PyObject** current_object = obj;
```
jjerphan, line 77:
```suggestion
                // Iterating on the content of the NumPy array to verify if all
                // elements are NaNs or if the NumPy array is empty.
                PyObject** current_object = obj;
```

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -68,19 +68,34 @@ NativeTensor obj_to_tensor(PyObject *ptr) {
...
+                while(current_object < obj + size) {
+                    if(*current_object == none.ptr()) {
+                        all_nans = false;
+                    } else if(is_py_nan(*current_object)) {
+                        empty = false;
+                    } else {
+                        all_nans = false;
```
jjerphan, line 84:
```suggestion
                        // `current_object` is an usual scalar value.
                        all_nans = false;
```

cpp/arcticdb/processing/test/test_has_valid_type_promotion.cpp
jjerphan, line 1: Should `trivially_compatible_types` also be tested similarly?

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -0,0 +1,29 @@
...
+    ) {
+        ARCTICDB_SAMPLE(HandleEmpty, 0);
+        util::check(dest != nullptr, "Got null destination pointer");
+        const size_t num_rows = dest_bytes / get_type_size(DataType::EMPTYVAL);
+        static_assert(get_type_size(DataType::EMPTYVAL) == sizeof(PyObject*));
+        const PyObject** ptr_dest = reinterpret_cast<const PyObject**>(dest);
+        py::none none = {};
+        for(auto row = 0u; row < num_rows; ++row) {
+            none.inc_ref();
+            *ptr_dest++ = none.ptr();
```
jjerphan, line 26: If I follow correctly, a NumPy array or `None` is returned when a column is empty?

jjerphan (commented):
Thank you, @vasil-pashov.

Here are a few comments and questions.

William and you mentioned other PRs in the threads of reviews. Do you intend to submit subsequent PRs for the support of other types? :slightly_smiling_face:

cpp/arcticdb/pipeline/read_frame.cpp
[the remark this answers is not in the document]
vasil-pashov: I don't think that a comment will add any value. Moreover that nothing in the code uses the buffers param. It's created in one of the top-level functions and passed in each nested function. TBH I'm not sure this is the best way to handle this. Maybe it would be better adding some sort of global buffer manager singleton, I don't know. My point is that we're not 100% sure what shape will this take, so adding a comment "We will add x, y, z here" might raise more questions than it answers.

cpp/arcticdb/entity/type_utils.hpp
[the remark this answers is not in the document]
willdealtry: The indentation is just plain wrong, I would always fix this where I noticed it. Other than that the bracketing does follow a pattern, which is that where both if and else blocks are a single line it is permitted to omit the braces, otherwise they should be used.

One of the things that I think aids readibility is not having too much vertical sprawl. I really don't want to put braces around every one-line branch for that reason.

cpp/arcticdb/entity/types.hpp
```diff
@@ -152,12 +155,17 @@ constexpr bool is_utf_type(ValueType v) {
...
+    return v == ValueType::EMPTY;
+}
+
 enum class SizeBits : uint8_t {
     UNKNOWN_SIZE_BITS = 0,
     S8 = 1,
     S16 = 2,
     S32 = 3,
     S64 = 4,
+    COUNT
```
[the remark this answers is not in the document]
vasil-pashov: As William mentioned in one of the comments this is quite standard practice in C++. I don't see what value will this add.

cpp/arcticdb/pipeline/read_frame.cpp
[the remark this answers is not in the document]
willdealtry: On reflection, I'm not convinced that moving the type handling changes in piecemeal has been a huge success, given that it already exists in its entirety in BitBucket. Small PRs are nice, but IMO they should be as small as they can be to still make sense, and no smaller. Maybe something to consider for future planning.

cpp/arcticdb/pipeline/frame_utils.hpp
```diff
@@ -172,6 +169,8 @@ void aggregator_set_data(
                     agg.set_array(col, t);
                 }
             }
+        }  else if constexpr (!is_empty_type(dt)) {
+            static_assert(!sizeof(dt), "Unknown data type");
```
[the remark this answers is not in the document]
vasil-pashov: This is kind of doing the same thing isn't it? We should do nothing in case of empty type is passed and if `dt` is not empty then we have unhandled type and the code should not compile at all. This is a shorthand for:
```c++
...
else if constexpr (is_empty_type(dt)) {
} else {
            static_assert(!sizeof(dt), "Unknown data type");
}
```
and it's better than throwing since if we have type which is not handled we won't even compile.

cpp/arcticdb/pipeline/frame_utils.hpp
willdealtry: I think what we should do in the relatively short term is to move the string handling code into the type handlers. That way we will still have two branches: 1) thing that are simple numeric types and don't require any language-specific handling like ints and floats, and 2) everything else - strings, decimals, numpy arrays, boxed Python types that can also be None etc. 

Unfortunately I think we'll need to get the existing code to read and write these types moved over from the old project more or less as-is first, because it's blocking some users, and refactoring string handling etc is much too big a job to attempt at the same time. Integrating the new object types into the data processing framework is something that we will need to do later, and I'm not sure it will even make sense in all cases. We may end up (as is the case with pickled and custom normalized data) with things that we can store and retrieve, but not sensibly do numerical operations on, at least for the time being.

cpp/arcticdb/python/python_to_tensor_frame.cpp
vasil-pashov: I'm personally not a fan of such type of comments. There is not sophisticated algorithm. If anyone else believes it adds value I'll add it.

cpp/arcticdb/python/python_handlers.cpp
vasil-pashov: No, we should return an array filled with None values.

cpp/arcticdb/entity/types.hpp
```diff
@@ -357,6 +370,7 @@ DATA_TYPE_TAG(ASCII_FIXED64, std::uint64_t)
 DATA_TYPE_TAG(ASCII_DYNAMIC64, std::uint64_t)
 DATA_TYPE_TAG(UTF_FIXED64, std::uint64_t)
 DATA_TYPE_TAG(UTF_DYNAMIC64, std::uint64_t)
+DATA_TYPE_TAG(EMPTYVAL, uintptr_t)
```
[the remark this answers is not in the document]
vasil-pashov: Seems reasonable. I don't think we care about 32bit systems.

cpp/arcticdb/codec/codec.cpp
```diff
@@ -30,8 +30,13 @@ std::pair<size_t, size_t> ColumnEncoder::max_compressed_size(
...
             const auto nbytes = block.value().nbytes();
-            util::check(nbytes, "Zero-sized block");
-            uncompressed_bytes += nbytes;
+            // For the empty type the column will contain 0 size of user data however the encoder might need add some
+            // encoder specific data to the buffer, thus the uncompressed size will be 0 but the max_compressed_bytes
+            // might be non-zero.
+            if constexpr(!is_empty_type(TDT::DataTypeTag::data_type)) {
+                util::check(nbytes > 0, "Zero-sized block");
+                uncompressed_bytes += nbytes;
+            }
```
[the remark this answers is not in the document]
vasil-pashov: The idea is to explain why we might throw exception in case of type other than emtpy_type (i.e. why we have the if)

cpp/arcticdb/python/python_handlers.cpp
jjerphan: I made a typo and meant "a NumPy array _of_ `None`".

cpp/arcticdb/python/python_to_tensor_frame.cpp
vasil-pashov: To be fair I'm not 100% sure why UTF_DYNAMIC was chosen. Could be related to the size of `NaN` on disc. @alexowens90 @willdealtry can you confirm?

cpp/arcticdb/python/python_handlers.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <entity/types.hpp>
+#include <util/type_handler.hpp>
+
+namespace arcticdb {
+    namespace py = pybind11;
+    struct EmptyHandler {
+
+        EmptyHandler() = default;
+
+        void handle_type(
```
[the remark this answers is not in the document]
vasil-pashov: Added a comment with reference to where the type erasure is defined.

cpp/arcticdb/pipeline/read_frame.cpp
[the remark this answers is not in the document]
vasil-pashov: I have reverted the changes and brought the buffers back.

cpp/arcticdb/python/python_to_tensor_frame.cpp
vasil-pashov: I've added a comment about the way we treat NaN and None, which also explains why we need this loop

cpp/arcticdb/python/python_to_tensor_frame.cpp
vasil-pashov: I've added this explanation as a comment in the code

cpp/arcticdb/pipeline/frame_utils.hpp
vasil-pashov: I've made a list of all function which behave as if there are only numbers and strings. But since it's out of the scope of the PR I won't change them.

cpp/arcticdb/python/python_to_tensor_frame.cpp
willdealtry: This is unfortunately a historic thing where NaNs were used to indicate missing strings. I'm not going to defend it making sense but there's not a lot we can do about it now. The decision was made that this unfortunate fact should go no further, so any other missing Python type, like a decimal or an array, should be represented by None.

Might be worth a comment.

vasil-pashov: I think the question is not why we treat `NaN` as as string (I think we've already got that covered), but why we set the DataType to be `UTF_DYNAMIC` and not `ASCII_DYNAMIC` for example.

cpp/arcticdb/processing/test/test_has_valid_type_promotion.cpp
vasil-pashov: I've added the tests

cpp/arcticdb/processing/test/test_has_valid_type_promotion.cpp
```diff
@@ -121,3 +121,28 @@ TEST(HasValidTypePromotion, FloatFloat) {
...
+}
+
+TEST(HasValidTypePromotion, EverythingToEmpty) {
+    using namespace arcticdb;
+    using namespace arcticdb::entity;
+    const TypeDescriptor target(ValueType::EMPTY, SizeBits::S64, Dimension::Dim0);
+    for(int value_type = int(ValueType::UNKNOWN_VALUE_TYPE); value_type < int(ValueType::COUNT); ++value_type) {
+        for(int size_bits = int(SizeBits::UNKNOWN_SIZE_BITS); size_bits < int(SizeBits::COUNT); ++size_bits) {
+            const TypeDescriptor source(ValueType(value_type), SizeBits(size_bits), Dimension::Dim0);
+            ASSERT_EQ(has_valid_type_promotion(source, target), target);
```
alexowens90, line 144: This is wrong, `has_valid_type_promotion` should always return `std::nullopt` if the target is the empty type

alexowens90 (changes requested):

alexowens90:
> This is a list of functions which assume that if something is not string it's number of vice versa:
> 
> ```c++
> VariantData binary_membership(const ColumnWithStrings& column_with_strings, ValueSet& value_set, Func&& func)
> VariantData binary_comparator(const Value& val, const ColumnWithStrings& column_with_strings, Func&& func)
> VariantData binary_comparator(const ColumnWithStrings& left, const ColumnWithStrings& right, Func&& func)
> VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const Value& val, Func&& func)
> SegmentInMemoryImpl::visit_field(Callable &&c)
> SegmentInMemoryImpl::operator==(const SegmentInMemoryImpl& left, const SegmentInMemoryImpl& right)
> FrameSliceMap::FrameSliceMap(const std::shared_ptr<PipelineContext>& context, bool dynamic_schema)
> std::shared_ptr<SegmentInMemoryImpl> filter(const util::BitSet& filter_bitset,
>                                                        bool filter_down_stringpool=false,
>                                                        bool validate=false) const {
> Value(const Value& other)
> void SumAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values)
> void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column)
> void MaxOrMinAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values)
> VariantData binary_membership(const ColumnWithStrings& column_with_strings, ValueSet& value_set, Func&& func) {
> VariantData binary_comparator(const Value& val, const ColumnWithStrings& column_with_strings, Func&& func) {
> VariantData binary_comparator(const ColumnWithStrings& left, const ColumnWithStrings& right, Func&& func) {
> VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const Value& val, Func&& func) {
> inline StreamId stream_id_from_segment(const SegmentInMemory &seg, ssize_t row) {
> folly::Future<arcticdb::entity::VariantKey> write_frame_data(const TestDataFrame &data_frame, WriterType &writer) {
> ```
> 
> they all have contain if/else case similar to:
> 
> ```
> if(is_sequence_type(dt)) {
> ...
> } else {
> ...
> }
> ```
> 
> or
> 
> ```
> if(is_numeric_type(dt)) {
> ...
> } else {
> ...
> }
> ```
> 
> Probably for some of them it's fine to leave it as is, but we might have problems down the road when none type and other types are added.

Can you add 2 tickets to deal with these. One for everything `QueryBuilder` related:
```c++
VariantData binary_membership(const ColumnWithStrings& column_with_strings, ValueSet& value_set, Func&& func)
VariantData binary_comparator(const Value& val, const ColumnWithStrings& column_with_strings, Func&& func)
VariantData binary_comparator(const ColumnWithStrings& left, const ColumnWithStrings& right, Func&& func)
VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const Value& val, Func&& func)
std::shared_ptr<SegmentInMemoryImpl> filter(const util::BitSet& filter_bitset,
                                                       bool filter_down_stringpool=false,
                                                       bool validate=false) const {
void SumAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values)
void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column)
void MaxOrMinAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values)
VariantData binary_membership(const ColumnWithStrings& column_with_strings, ValueSet& value_set, Func&& func) {
VariantData binary_comparator(const Value& val, const ColumnWithStrings& column_with_strings, Func&& func) {
VariantData binary_comparator(const ColumnWithStrings& left, const ColumnWithStrings& right, Func&& func) {
VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const Value& val, Func&& func) {
```
and one for everything else. I can knock all the query builder ones off pretty quickly, I don't know off the top of my head about the others.

cpp/arcticdb/entity/type_utils.hpp
```diff
@@ -15,6 +15,11 @@ inline bool trivially_compatible_types(entity::TypeDescriptor left, entity::Type
     if(left == right)
         return true;
 
+    // Empty type is coercible to any type, and any type is coercible to the empty type.
+    if(is_empty_type(left.data_type()) || is_empty_type(right.data_type())) {
+        return true;
+    }
```
alexowens90, line 21: This change should be removed, `trivially_compatible_types` is used to `memcpy` buffers around, which doesn't make sense with `None`-type columns.
e.g. https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/version/version_core.cpp#L747

alexowens90 (changes requested):

alexowens90 (approved):

PR #648, opened 2023-07-24
Fix clang 16 compilation errors

vasil-pashov:
Tried to compile the codebase with clang-16. It promoted some warnings to errors due to the Werror flag. Mostly unused function arguments or lambda captures.

### DIFF
--- a/cpp/arcticdb/processing/operation_dispatch_binary.hpp
+++ b/cpp/arcticdb/processing/operation_dispatch_binary.hpp
@@ -56,15 +56,13 @@ VariantData binary_membership(const ColumnWithStrings& column_with_strings, Valu
         entity::details::visit_type(column_with_strings.column_->type().data_type(),[&column_with_strings, &value_set, &func, &output] (auto column_desc_tag) {
             using ColumnTagType = typename std::decay_t<decltype(column_desc_tag)>;
             using ColumnType = typename ColumnTagType::raw_type;
-            constexpr auto column_tag_data_type = ColumnTagType::data_type;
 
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -1299,7 +1299,7 @@ VersionedItem defragment_symbol_data_impl(
             pre_defragmentation_info.append_after.value(),
             std::nullopt);
     
-    store->remove_keys(delete_keys);
+    std::ignore = store->remove_keys(delete_keys);
```
vasil-pashov, line 1302: This one is bugging me a bit. `remove_keys` returns a future, don't we want to use it/pass it to the caller.

alexowens90 (approved):

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -57,7 +57,7 @@ VariantData binary_membership(const ColumnWithStrings& column_with_strings, Valu
             using ColumnTagType = std::decay_t<decltype(column_desc_tag)>;
             using ColumnType =  typename ColumnTagType::raw_type;
 
-            entity::details::visit_type(value_set.base_type().data_type(),[&column_with_strings, &value_set, &func, &output, &column_desc_tag] (auto value_set_desc_tag) {
```
qc00, line 60: I think this was necessary to make another compiler work, but I can't remember which one...

qc00 (approved):

cpp/arcticdb/processing/operation_dispatch_binary.hpp
vasil-pashov: Edit: The final solution to please the both compilers was to use catch all by ref

PR #659, opened 2023-07-26
Fix negative as_of for batch methods

alexowens90:
#### What does this implement/fix?

Fixes batch methods (read, read metadata, get description) when a negative integer is provided in the `as_of` argument.

For non-batch methods, logic was added to `get_specific_version` in `version_functions.hpp` to handle signed integers. The same logic needed to be applied to `get_key_for_version_query` in `version_map_batch_methods.hpp`.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_map_batch_methods.hpp
+++ b/cpp/arcticdb/version/version_map_batch_methods.hpp
@@ -189,8 +189,25 @@ inline std::optional<AtomKey> get_key_for_version_query(
     const std::shared_ptr<VersionMapEntry>& version_map_entry, 
     const pipelines::VersionQuery& version_query) {
     return util::variant_match(version_query.content_,
-        [&version_map_entry] (const pipelines::SpecificVersionQuery& specific_version) {
-            return find_index_key_for_version_id(specific_version.version_id_, version_map_entry);
+        [&version_map_entry] (const pipelines::SpecificVersionQuery& specific_version) -> std::optional<AtomKey> {
+            auto signed_version_id = specific_version.version_id_;
+            VersionId version_id;
+            if (signed_version_id >= 0) {
+                version_id = static_cast<VersionId>(signed_version_id);
+            } else {
+                auto opt_latest = version_map_entry->get_first_index(true);
+                if (opt_latest.has_value()) {
+                    auto opt_version_id = get_version_id_negative_index(opt_latest->version_id(), signed_version_id);
+                    if (opt_version_id.has_value()) {
+                        version_id = *opt_version_id;
+                    } else {
+                        return std::nullopt;
+                    }
+                } else {
+                    return std::nullopt;
+                }
+            }
+            return find_index_key_for_version_id(version_id, version_map_entry);
         },
         [&version_map_entry] (const pipelines::TimestampVersionQuery& timestamp_version) -> std::optional<AtomKey> {
             auto version_key = get_version_key_from_time_for_versions(timestamp_version.timestamp_, version_map_entry->get_indexes(false));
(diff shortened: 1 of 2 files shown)


mehertz (approved):
Ideally this functionality is shared across `version_map.hpp` and `version_map_batch_methods.hpp` but this can be said for many of the methods in this file. 

Approving this PR, we'll need to clean the interaction between these files later.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -189,8 +189,25 @@ inline std::optional<AtomKey> get_key_for_version_query(
...
-            return find_index_key_for_version_id(specific_version.version_id_, version_map_entry);
+        [&version_map_entry] (const pipelines::SpecificVersionQuery& specific_version) -> std::optional<AtomKey> {
+            auto signed_version_id = specific_version.version_id_;
+            VersionId version_id;
+            if (signed_version_id >= 0) {
+                version_id = static_cast<VersionId>(signed_version_id);
+            } else {
+                auto opt_latest = version_map_entry->get_first_index(true);
+                if (opt_latest.has_value()) {
+                    auto opt_version_id = get_version_id_negative_index(opt_latest->version_id(), signed_version_id);
```
qc00, line 200: Not strictly a fault of this PR, but the implementation of `get_version_id_negative_index` assumes the version IDs are contiguous which is not always the case.

jjerphan, line 200: Looking at the implementation of `get_version_id_negative_index`, no special treatment exist if `signed_version_id` is negative and strictly smaller than `static_cast<SignedVersionId>(latest) - 1`, (a negative value for will be returned via `opt_version_id`).

Should ArcticDB adopt python's list's behavior, i.e.:
```python
>>> a = [1, 2, 3]
>>> a[-2]  
2
>>> a[-7]  
IndexError: list index out of range
```

and err similarly in this case?

jjerphan (approved):
LGTM modulo a suggestion that can be addressed in another PR if this fix really needs to get in.

cpp/arcticdb/version/version_map_batch_methods.hpp
alexowens90: This is exactly how it behaves?
`// Given the latest version, and a negative index into the version map, returns the desired version ID or std::nullopt if it would be negative
inline std::optional<VersionId> get_version_id_negative_index(VersionId latest, SignedVersionId index) {
    internal::check<ErrorCode::E_ASSERTION_FAILURE>(index < 0, "get_version_id_negative_index expects a negative index, received {}", index);
    // +1 so that as_of=-1 returns the latest entry
    auto candidate_version_id = static_cast<SignedVersionId>(latest) + index + 1;
    return candidate_version_id >= 0 ? std::make_optional<VersionId>(static_cast<VersionId>(candidate_version_id)) : std::nullopt;
}`
If `candidate_version_id` is negative, it returns `std::nullopt`, which further up the stack either throws an exception, or returns a `DataError` object

jjerphan: I think that my remark was not clear: a `std::nullopt` is indeed returned in this case, eventually returning a `DataError` to users.

As of my comprehension of the implementation, the error will not be explicit and it might be hard for users to know for which reason a `DataError` is returned. Currently, `DataError` are documented as being returned when the version is missing:

https://github.com/man-group/ArcticDB/blob/664b7597c249f46cc2c0e14646046b6df7e3aff9/python/arcticdb/version_store/library.py#L877-L882

Anyway that's a UX suggestion, we could address later in another PX.

alexowens90: Ah, I understand now. This behaviour is consistent with positive `as_of` (i.e. the error is the same if there are only 3 versions, but you request version 4). You're right that we don't currently have a good way to distinguish between the symbol doesn't exist, the symbol exists but the version has been deleted, the symbol exists but the version never existed, and the symbol exists but the requested version number cannot exist (i.e. is negative).

PR #660, opened 2023-07-26
Minor dependency upgrades for curl and openssl

poodlewars:
Curl 7.86.0 -> 7.88.1#1
OpenSSL 1.1.1l -> 1.1.1n#1

### DIFF
--- a/cpp/vcpkg.json
+++ b/cpp/vcpkg.json
@@ -61,24 +61,10 @@
     "azure-storage-blobs-cpp"
   ],
   "overrides": [
-    {
-      "name": "openssl",
-      "version-string": "1.1.1n#1"
-    },
-    {
-      "name": "aws-sdk-cpp",
-      "version": "1.11.4#0",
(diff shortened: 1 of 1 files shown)


vasil-pashov:
Was there any specific reason to update e.g. a known bug, security vulnerability, significant performance improvement?

cpp/vcpkg.json
```diff
@@ -156,7 +156,7 @@
     { "name": "boost-vcpkg-helpers", "version": "1.80.0#1" },
     { "name": "boost-winapi", "version": "1.80.0#1" },
     { "name": "civetweb", "version": "1.15#2" },
-    { "name": "curl", "version": "7.86.0#1" },
+    { "name": "curl", "version": "7.88.1#1" },
```
qc00, line 159: This should now be moved above the comment on line 78

poodlewars: I'm just going to remove the comments, I don't think they add much value and it's a maintenance burden to keep them correct.

mehertz (approved):

Issue #458, opened 2023-06-05
Apple Silicon build for conda-forge


mehertz:

DerThorsten:
* ArcticDb already builds fine with apple silicon. The is just a lack of a conda-forge build
* All packages are already migrated
* we are doing the apple-silicon build here https://github.com/conda-forge/arcticdb-feedstock/pull/6
* this is a bit more complicated than the normal builds as we need to cross-compile arm64 builds on an x64 machine.
* We cannot test the apple-silicon build in the github-actions (without using complicated emulation / cross-compilation) as there are no apple-silicon runners

WeiLi1201:
Hi, I installed with command  `conda install -c conda-forge arcticdb` success, but still wrong when run examples.
I create a new conda env retest,  hope your update, thx. 
env :  m1  macOS 13.1
```
Traceback (most recent call last):
  File "/Users/****/pycharmProject/vnpy****/examples/alpha_platform/factor_research/arcticdb_learn.py", line 5, in <module>
    from arcticdb import Arctic, QueryBuilder
  File "/Users/****/anaconda3/envs/vnpy/lib/python3.9/site-packages/arcticdb/__init__.py", line 1, in <module>
    import arcticdb_ext as _ext
ImportError: dlopen(/Users/****/anaconda3/envs/vnpy/lib/python3.9/site-packages/arcticdb_ext.cpython-39-darwin.so, 0x0002): tried: '/Users/****/anaconda3/envs/vnpy/lib/python3.9/site-packages/arcticdb_ext.cpython-39-darwin.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64')), '/System/Volumes/Preboot/Cryptexes/OS/Users/****/anaconda3/envs/vnpy/lib/python3.9/site-packages/arcticdb_ext.cpython-39-darwin.so' (no such file), '/Users/****/anaconda3/envs/vnpy/lib/python3.9/site-packages/arcticdb_ext.cpython-39-darwin.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
```

DerThorsten:
I can reproduce this and will investigate, thanks for reporting!

chen-001:
Hi, I have macos 13.4.1 with M1 max. I have the same question to [WeiLi1201's issue](https://github.com/man-group/ArcticDB/issues/458#issuecomment-1632456545). I also get
 ```
[...]

PR #662, opened 2023-07-27
conda-build: Use default compilers for macOS

DerThorsten:
#### Reference Issues/PRs
* #458  
* https://github.com/conda-forge/arcticdb-feedstock/pull/35

use the proper compiler when on conda forge, without this, we use the wrong compiler and this then gives binaries for the wrong arch in case of cross-compiling (ie we get x64 instead of arm64 binaries)

#### Any other comments?
needed for the apple-silicon build for conda forge. The output of https://github.com/conda-forge/arcticdb-feedstock/pull/35 indicates that this PR fixes the problem (at least the linker is not complaining about the wrong arch anymore)
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [x] Is the PR labeled as enhancement or bug so it appears in autogenerated release notes?
</details>


cpp/CMakePresets.json
```diff
@@ -112,8 +112,8 @@
       "generator": "Unix Makefiles",
       "cacheVariables": {
         "CMAKE_MAKE_PROGRAM": "make",
-        "CMAKE_C_COMPILER": "clang",
-        "CMAKE_CXX_COMPILER": "clang",
+        "CMAKE_C_COMPILER": "$env{CC}",
+        "CMAKE_CXX_COMPILER": "$env{CXX}",
```
qc00, line 116: These are the [default](https://cmake.org/cmake/help/latest/envvar/CC.html), so you can simply delete these lines.

.github/workflows/build_with_conda.yml
```diff
@@ -47,6 +47,7 @@ jobs:
           python -m pip install --no-build-isolation --no-deps -v -e .
         env:
           ARCTICDB_USING_CONDA: 1
+          CMAKE_BUILD_PARALLEL_LEVEL: 1
```
jjerphan, line 50:
```suggestion
          # Using one thread to avoid swapping which freezes the CI machine.
          CMAKE_BUILD_PARALLEL_LEVEL: 1
```

.github/workflows/build_with_conda.yml
```diff
@@ -87,6 +88,7 @@ jobs:
...
```
jjerphan, line 91:
```suggestion
          # Using one thread to avoid swapping which freezes the CI machine.
          CMAKE_BUILD_PARALLEL_LEVEL: 1
```

jjerphan (approved):
LGTM.

Nit: I would just change the title of the PR to mention that it only changes the configuration for MacOS.

PR #704, opened 2023-08-08
Bugfix 703: Restrict supported string types to type equality rather than has isinstance

### LINKED ISSUE #703: Columns of objects that inherit from `str` in Python layer can segfault
e.g. Enums end up being treated as unicode strings, which then cannot be converted to our internal UTF=8 format as they are not valie unicode.

alexowens90:
#### Reference Issues/PRs

Closes #703 

When normalising columns of `dtype` `object`, all of the contained objects **must** be of a Python string type.

The existing implementation used `isinstance` to check for this condition, but this also picks up classes that have inherited from `str`. Instead, we must check that the `type` of the column elements is an exact match for `str` in order for lower level code to work correctly.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -109,7 +109,10 @@ if PY3:
     def _accept_array_string(v):
         # TODO remove this once arctic keeps the string type under the hood
         # and does not transform string into bytes
-        return isinstance(v, string_types) or isinstance(v, binary_type)
+        # string_types and binary_type can be a single type or a tuple
+        supported_string_types = string_types if isinstance(string_types, tuple) else (string_types,)
+        supported_binary_types = binary_type if isinstance(binary_type, tuple) else (binary_type,)
+        return type(v) in supported_string_types or type(v) in supported_binary_types
 
 else:
 


python/arcticdb/version_store/_normalization.py
```diff
@@ -109,7 +109,10 @@ def from_cpp(fd):
     def _accept_array_string(v):
         # TODO remove this once arctic keeps the string type under the hood
         # and does not transform string into bytes
-        return isinstance(v, string_types) or isinstance(v, binary_type)
+        # string_types and binary_type can be a single type or a tuple
+        supported_string_types = string_types if isinstance(string_types, tuple) else (string_types,)
+        supported_binary_types = binary_type if isinstance(binary_type, tuple) else (binary_type,)
+        return type(v) in supported_string_types or type(v) in supported_binary_types
```
qc00, line 115: Minor, might be easier to change `string_types -> str` and `binary_type -> bytes` instead of importing them from `six` given we don't support Python 2 anymore.

qc00 (approved):

mehertz (approved):

python/arcticdb/version_store/_normalization.py
alexowens90: We should add a ticket to strip out dead code that was only needed for Python 2 in general

PR #705, opened 2023-08-08
Ensure compact_incomplete and recursive normalization obey the library setting for pruning

### LINKED ISSUE #538: Ensure `compact_incomplete` and recursive normalization obey the library setting for pruning
Previously, 

* `NativeVersionStore.compact_incomplete` (`Library.finalize_staged_data` in the public API), and
* `write` with recursive normalizers

ignored the `prune_previous_version(s)` argument and library option and never removed the old version(s).

qc00:
Closes #538

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -983,8 +983,6 @@ VersionedItem LocalVersionedEngine::compact_incomplete_dynamic(
             store_, stream_id, user_meta, update_info,
             append, convert_int_to_float, via_iteration, sparsify, get_write_options());
 
-    version_map_->write_version(store_, versioned_item.key_);
-
     write_version_and_prune_previous_if_needed(
         prune_previous_versions, versioned_item.key_, update_info.previous_index_key_);
 
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -53,17 +53,22 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_unreferenced_pruned_inde
 ) {
     try {
         if (!pruned_indexes.empty() && !cfg().write_options().delayed_deletes()) {
+            // TODO: the following function will load all snapshots, which will be horrifyingly inefficient when called
+            // multiple times from batch_*
             auto [not_in_snaps, in_snaps] = get_index_keys_partitioned_by_inclusion_in_snapshots(
                     store(),
                     pruned_indexes.begin()->id(),
                     pruned_indexes);
             in_snaps.insert(key_to_keep);
-            PreDeleteChecks checks{false, false, false, false, in_snaps};
-            return delete_trees_responsibly(not_in_snaps, {}, {}, checks);
+            PreDeleteChecks checks{false, false, false, false, std::move(in_snaps)};
+            return delete_trees_responsibly(not_in_snaps, {}, {}, checks)
+                    .thenError(folly::tag_t<std::exception>{}, [](auto const& ex) {
+                        log::version().warn("Failed to clean up pruned previous versions due to: {}", ex.what());
+                    });
         }
     } catch (const std::exception &ex) {
         // Best-effort so deliberately swallow
-        log::version().warn("Could not prune previous versions due to: {}", ex.what());
+        log::version().warn("Failed to clean up pruned previous versions due to: {}", ex.what());
     }
     return folly::Unit();
 }
@@ -552,12 +557,8 @@ VersionedItem LocalVersionedEngine::update_internal(
                                           std::move(frame),
                                           get_write_options(),
                                           dynamic_schema);
-        if (prune_previous_versions) {
-            auto pruned_indexes = version_map()->write_and_prune_previous(store(), versioned_item.key_, update_info.previous_index_key_);
-            delete_unreferenced_pruned_indexes(pruned_indexes, versioned_item.key_).get();
-        } else {
-            version_map()->write_version(store(), versioned_item.key_);
-        }
+        write_version_and_prune_previous_if_needed(
+            prune_previous_versions, versioned_item.key_, update_info.previous_index_key_);
         return versioned_item;
     } else {
         if (upsert) {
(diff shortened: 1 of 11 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1004,6 +985,9 @@ VersionedItem LocalVersionedEngine::compact_incomplete_dynamic(
 
     version_map_->write_version(store_, versioned_item.key_);
```
mehertz, line 986: Should this not be replaced by the call to `write_version_and_prune_previous_if_needed`? **It was in the previous PR**

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -403,7 +408,8 @@ class LocalVersionedEngine : public VersionedEngine {
             bool append,
             bool convert_int_to_float,
             bool via_iteration,
-            bool sparsify) override;
+            bool sparsify,
+            bool prune_previous_versions);
```
mehertz, line 412: Intentionally removing `override`?

cpp/arcticdb/version/versioned_engine.hpp
```diff
@@ -121,7 +121,8 @@ class VersionedEngine {
         bool validate_index
     ) = 0;
 
-    virtual std::pair<VersionedItem, std::vector<AtomKey>> write_individual_segment(
+    /** Test-specific cut-down version of write_versioned_dataframe_internal */
```
mehertz, line 124: This is only used in tests?

mehertz (approved):

willdealtry (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
qc00: Oops. Pushed a fix.

cpp/arcticdb/version/local_versioned_engine.hpp
qc00: See [versioned_engine.hpp](https://github.com/man-group/ArcticDB/pull/705/files#diff-4dedfebe3ff32890e64ef100fdf464ead480ce0a704d071192d7af2e731d429d) line 150

cpp/arcticdb/version/versioned_engine.hpp
qc00: Yes, and it's not in `arcticc`, which makes me believe it was added because `write_versioned_dataframe_internal` picked up some new behaviour that we don't want in tests?

python/arcticdb/version_store/_store.py
```diff
@@ -1735,13 +1744,21 @@ def compact_incomplete(
             Convert data to sparse format (for tick data only)
         metadata : `Optional[Any]`, default=None
             Add user-defined metadata in the same way as write etc
+        prune_previous_version
+            Removes the previous versions from the database after a successful compaction
+            (unless the previous version is in a snapshot).
```
alexowens90, line 1749: The docstring for the other methods that take this parameter is consistently:
`Removes previous (non-snapshotted) versions from the database.`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -53,17 +53,22 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_unreferenced_pruned_inde
...
                     store(),
                     pruned_indexes.begin()->id(),
                     pruned_indexes);
             in_snaps.insert(key_to_keep);
-            PreDeleteChecks checks{false, false, false, false, in_snaps};
-            return delete_trees_responsibly(not_in_snaps, {}, {}, checks);
+            PreDeleteChecks checks{false, false, false, false, std::move(in_snaps)};
+            return delete_trees_responsibly(not_in_snaps, {}, {}, checks)
+                    .thenError(folly::tag_t<std::exception>{}, [](auto const& ex) {
+                        log::version().warn("Failed to clean up pruned previous versions due to: {}", ex.what());
```
alexowens90, line 66: Am I right in saying that for batch methods, this will result in an extra log, but a `VersionedItem` will still be returned rather than a `DataError`?

python/tests/unit/arcticdb/version_store/test_compaction.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+import pandas as pd
+
+from arcticdb.version_store import NativeVersionStore
+
+
+@pytest.mark.parametrize("append", (True, False))
+@pytest.mark.parametrize("mode", ("no_prune", "lib_config", "arg"))
+def test_compact_incomplete_prune_previous(mode: str, append: bool, version_store_factory):
```
alexowens90, line 9: These parametrizations don't cover the lib config setting being true and the arg being false

python/arcticdb/version_store/_store.py
```diff
@@ -1713,6 +1721,7 @@ def compact_incomplete(
         via_iteration: Optional[bool] = True,
         sparsify: Optional[bool] = False,
         metadata: Optional[Any] = None,
+        prune_previous_version: Optional[bool] = None,
```
alexowens90, line 1724: Don't we need the same change in `library.py`?

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_compaction.py
alexowens90: Also, is there an equivalent test for recursive normalizers?

cpp/arcticdb/version/local_versioned_engine.cpp
qc00: Yes

python/arcticdb/version_store/_store.py
qc00: I find the old documentation ambiguous

python/tests/unit/arcticdb/version_store/test_compaction.py
qc00: > These parametrizations don't cover the lib config setting being true and the arg being false

Will handle that case.

> Also, is there an equivalent test for recursive normalizers?

No. Recursive normalizer was not in scope to begin with and not officially documented anyway.

python/arcticdb/version_store/_store.py
qc00: On a related note: should the new `sort_merge` function prune as well?

python/arcticdb/version_store/_store.py
qc00: Couldn't be bothered to update the other ones, so will use the old one to be consistent.

python/tests/unit/arcticdb/version_store/test_compaction.py
alexowens90: Can you create tickets for those if they don't already exist then?

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_compaction.py
qc00: Raised #711

PR #696, opened 2023-08-07
maint: Use `Storage` class hierarchy instead of `VariantStorage`

jjerphan:
#### Reference Issues/PRs

Towards #633.

To be merged after #670.

#### What does this implement/fix? How does it work (high level)? Highlight notable design decisions.

This PR:
 - removes `arcticdb::storage::VariantStorage` and perform associated name changes (note that the `arcticdb::proto::storage::VariantStorage` is left unchanged)
 - removes the CRTP on `arcticdb::storage::Storage`
 - defines pure virtual protected template methods
 - uses implementations of concrete classes extending `arcticdb::storage::Storage` as overrides of those template methods
 - moves `arcticdb::storage::s3::{S3,NfsBacked}Storage`'s implementation details in a dedicated file
 - simplifies the logic of `storage_specific` down to what we need (this needs to be discussed IMO)

#### Any other comments?

We could:
 - add inline documentation, although I do not think this is needed

We could in another PR:
 - rename `arcticdb::proto::storage::VariantStorage` to `arcticdb::proto::storage::StorageConfig` #709

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings and documentation?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

Issue #633, opened 2023-07-24
Remove VariantStorage


alexowens90:

jjerphan:
The main pain is to remove the templates on `Visitor`. For this, we need to find what are the instantiations of `Visitor`.

jjerphan:
Completed by #696.

qc00:
Additional compilation time savings: #763

PR #709, opened 2023-08-10
maint: Rename protocol buffer `VariantStorage` to `StorageConfig` 

jjerphan:
#### Reference Issues/PRs

Towards #633.

In addition to #696, this removes the last mention of `VariantStorage`, `arcticdb::proto::storage::VariantStorage` which encapsulates one configuration opaquely via `google.protobuf.Any`.

I chose `StorageConfig` as a name to be consistent with the names of the actual Storage configuration protocol buffers (i.e. `arcticdb::proto::*_storage::Config`).

#### Any other comments?

Do we need to maintain a Python alias for backward compatibility?

:warning: In any case, I think we must perform checks (such as the ones mentioned in #592) for this renaming.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/proto/arcticc/pb2/storage.proto
+++ b/cpp/proto/arcticc/pb2/storage.proto
@@ -19,11 +19,11 @@ message EnvironmentConfigsMap {
 // TODO Add Task scheduler configuration
 
 message EnvironmentConfig {
-    map<string, VariantStorage> storage_by_id = 1;
+    map<string, StorageConfig> storage_by_id = 1;
     map<string, LibraryDescriptor> lib_by_path = 2;
 }
 
-message VariantStorage {
+message StorageConfig {
     google.protobuf.Any config = 2;
     string storage = 3;
 }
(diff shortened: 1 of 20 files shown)


cpp/proto/arcticc/pb2/storage.proto
```diff
@@ -184,7 +184,7 @@ message LibraryConfig {
     Portable library configuration which is sufficient to recreate a library
     */
     LibraryDescriptor lib_desc = 1;
-    map<string, VariantStorage> storage_by_id = 2;
+    map<string, StorageConfig> storage_by_id = 2;
```
willdealtry, line 187: Is this going to be backwards compatible? We have literally thousands of these stored

jjerphan: I think we must test (see my comment in the description).

Alternatively, we can simply ignore this PR.

alexowens90: We use `FieldMask` in some internal code, so I think we might just have to live with this legacy name

jjerphan:
Closing because this PR will break backward compatibility as mentioned by https://github.com/man-group/ArcticDB/pull/709#discussion_r1291346310.

cpp/proto/arcticc/pb2/storage.proto
willdealtry: Can we just alias it in the C++ code? Nobody cares what it's called in the protobuf, and in the C++ we can call it whatever we want. The place for that would be in protobuf.hpp

PR #712, opened 2023-08-10
Enhancement 558+529+556: Add option to treat date_range as a clause to reduce memory usage, and so that it works with other QueryBuilder clauses

### LINKED ISSUE #529: GroupBy does not work with date-range filtering
Currently it throws an exception in `NativeVersionStore._post_process_dataframe` due to the hacky nature of how `date_range` (and `row_range`) restricts what is shown to the user down from what is returned from the C++ layer, which includes all of the rows in the relevant segments read from storage.
However, if this post processing is omitted, the result is still wrong. This is because the groupby and aggregation has happened over the whole segments read from storage for the specified date range, even if some of the rows should have been ignored.

### LINKED ISSUE #556: Make date_range a clause so that it works with groupbys and filtering
In a pure read scenario (i.e. no `QueryBuilder`) date_range should work as it does now as this is zero-copy.

### LINKED ISSUE #558: Add an option to treat date_range as a clause even if no `QueryBuilder` is provided to reduce memory usage
By default, we do a zero-copy decompress of data segments onto contiguous buffers for numpy to improve performance.
If a very small data range is requested, this can have a prohibitively high memory footprint, with only a small time saving as we would only need to do a small `memcpy` on top of the current work.

alexowens90:
#### Reference Issues/PRs

Closes #558 
Closes #529
Closes #556 

This PR adds a `date_range` clause to the `QueryBuilder` class. This can be used in place of the `date_range` arg to `read`-like methods. In this case, the output data will be the same, but there are more `memcpy`s involved, so it will be slightly slower, but with the advantage that the memory footprint of the returned dataframe will be smaller.

In addition, this means that filtering, projections, groupbys and aggregations now all work with the `date_range` parameter.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/column_store/column.hpp
+++ b/cpp/arcticdb/column_store/column.hpp
@@ -851,6 +851,63 @@ public:
         return data_.buffer().blocks();
     }
 
+    // Only works if column is of numeric type and is monotonically increasing
+    // Returns the index such that if val were inserted before that index, the order would be preserved
+    // By default returns the lowest index satisfying this property. If from_right=true, returns the highest such index
+    template<class T, std::enable_if_t<std::is_integral_v<T> || std::is_floating_point_v<T>, int> = 0>
+    size_t search_sorted(T val, bool from_right=false) const {
+        // There will not necessarily be a unique answer for sparse columns
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(!is_sparse(),
+                                                        "Column::search_sorted not supported with sparse columns");
+        std::optional<size_t> res;
+        type().visit_tag([this, &res, val, from_right](auto type_desc_tag){
+            using TDT = decltype(type_desc_tag);
+            using RawType = typename TDT::DataTypeTag::raw_type;
+            if constexpr(std::is_same_v<T, RawType>) {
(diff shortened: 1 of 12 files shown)


cpp/arcticdb/column_store/column.hpp
```diff
@@ -851,6 +851,63 @@ class Column {
...
+                            // Narrow the search interval
+                            high = mid - 1;
+                        } else {
+                            // val is greater than the previous value, so we have found the right interval
+                            res = mid;
+                        }
+                    }
+                }
+            } else {
+                // TODO: Could relax this requirement using something like has_valid_common_type
```
willdealtry, line 901: I don't think we'd do exactly that would we? The type of the needle needs to be less than that of the haystack, or have a value that can be contained in it, so it would be a one-way promotion rather than a common type

alexowens90: Not exactly, e.g. searching for an `int16_t` in a `uint16_t` column is possible, it will just trivially return `0` if the signed int is negative. Not one we need to tackle right now anyway.

Issue #668, opened 2023-07-28
top-k


octogenary:
- [x] #650 
- [x] #674 
- [x] #715
- [x] #754
- [ ] #716
- [ ] #717

PR #734, opened 2023-08-14
top-k initial cpp implementation

### LINKED ISSUE #674: Initial top-k C++ implementation
Deliberately vague since I'm not sure how far 'initial' will get. Will make slightly more detailed issues when we have a clearer idea of the order in which things will happen.

octogenary:
#### Reference Issues/PRs

Closes #674, #650.

A `VectorDB` wraps round a library (it takes the library when initialised). It supports

1. upsertion of vectors (for fresh identifiers, insertion; for used identifiers, updating); and
2. top-_k_ queries (presently Euclidean-only).

In Python, the API itself is otherwise completely unchanged. There are new methods in cpp that are accessible only through a `VectorDB` in Python.

#### Any other comments?

This is not very fast and limited to datasets that fit in memory; see #668 for further details.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/processing/clause.hpp
+++ b/cpp/arcticdb/processing/clause.hpp
@@ -49,6 +49,8 @@ struct ClauseInfo {
     std::optional<std::string> new_index_{std::nullopt};
     // Whether this clause modifies the output descriptor
     bool modifies_output_descriptor_{false};
+    // Whether to reorder columns as in the original DF.
+    bool follows_original_columns_order_{true};
 };
 
 // Changes how the clause behaves based on information only available after it is constructed
--- a/cpp/arcticdb/vector_db/vector_db.cpp
+++ b/cpp/arcticdb/vector_db/vector_db.cpp
@@ -1,0 +1,102 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#include <arcticdb/processing/clause.hpp>
+#include <arcticdb/vector_db/vector_db.hpp>
+
+namespace arcticdb {
+    TopKClause::TopKClause(std::vector<double> query_vector, uint8_t k) :
+            query_vector_(std::move(query_vector)),
+            k_(k) {
+        clause_info_.modifies_output_descriptor_ = true;
+        clause_info_.follows_original_columns_order_ = false;
+        clause_info_.can_combine_with_column_selection_ = false;
+    }
+
+    Composite<ProcessingSegment> TopKClause::process(std::shared_ptr<Store> store,
+                                                     Composite<ProcessingSegment> &&p) const {
+        auto procs = std::move(p);
+        TopK top_k(k_);
+        // We expect a vector in each column.
+        // top_k_ is updated as we read more vectors. We want to efficiently remove
+        // the 'worst' (least similar) vector as we go along to ensure we have the
+        // top k, not the top k+1.
+        auto lp = 2;
+        // Let's pretend that only the lp-norms exist. We'll use p=2 for now as a default.
+
+        procs.broadcast(
+                [&store, &top_k, lp, this](const ProcessingSegment &proc) {
+                    for (const auto& slice_and_key: proc.data()) {
+                        std::vector <std::shared_ptr<Column>> columns = slice_and_key.segment(store).columns();
+                        for (auto&& [idx, col]: folly::enumerate(columns)) {
+                            internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+                                    static_cast<long unsigned>(col->row_count()) == this->query_vector_.size(),
+                                    "Expected vector of length {}, got vector of length {}.",
+                                    col->row_count(),
+                                    this->query_vector_.size());
+                            col->type().visit_tag([&top_k, &slice_and_key, &store, &col=col, idx=idx, lp, this](auto type_desc_tag) {
+                                using TypeDescriptorTag = decltype(type_desc_tag);
+                                using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
+
+                                if constexpr (is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)) {
+
+                                    ColumnData col_data = col->data();
+                                    double sum_differences_exponentiated = 0;
+                                    auto j = 0u;
+
+                                    while (auto block = col_data.next<TypeDescriptorTag>()) {
+                                        auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                                        for (auto progress_in_block = 0u; progress_in_block <
+                                                                          block.value().row_count(); ++progress_in_block, ++j, ++ptr) {
+                                            sum_differences_exponentiated += pow(
+                                                    abs(*ptr - this->query_vector_[j]), lp);
+                                        }
+                                    }
+
+                                    top_k.try_insert(NamedColumnSortedByDistance(
+                                            sum_differences_exponentiated,
+                                            col,
+                                            slice_and_key.segment(store).field(idx).name()));
+                                }
+                            });
+                        }
+                    }
+                }
+        );
+        // By this time we should have iterated through each block of each SegmentInMemory of each ProcessingSegment;
+        // and in each case read a vector and inserted it if more similar than the running top_k_. Now we write to a new
+        // segment.
+        SegmentInMemory seg;
+        seg.descriptor().set_index(IndexDescriptor(0, IndexDescriptor::ROWCOUNT));
+        seg.set_row_id(query_vector_.size() - 1 + 1);
+        // -1 because of zero-indexing, +1 because we want to add the distance on at the end.
+            for (const auto &column: top_k.top_k_) {
+            // todo: replace (mem)cpying of columns' contents with holding of pointers in top_k_ and addition of pointers to seg
+            // this doesn't work presently because the columns can have >1 block.
+            column.column_->type().visit_tag([&column, &seg, lp, this](auto type_desc_tag) {
+                using TypeDescriptorTag = decltype(type_desc_tag);
+                using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
+
+                if constexpr (is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)) {
+                    auto write_col = std::make_shared<Column>(column.column_->type(), this->query_vector_.size() + 1, true, false);
+
+                    auto col_data = column.column_->data();
+                    auto write_ptr = reinterpret_cast<RawType *>(write_col->ptr());
+                    while (auto block = col_data.next<TypeDescriptorTag>()) {
+                        auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                        std::memcpy(write_ptr, ptr, block.value().row_count() * sizeof(RawType));
+                        write_ptr += block.value().row_count();
+                    }
+                    *write_ptr = RawType(pow(column.distance_, 1 / static_cast<double>(lp)));
+                    write_col->set_row_data(this->query_vector_.size());
+                    seg.add_column(scalar_field(DataType::FLOAT64, column.name_), write_col);
+                    }
+            });
+        }
+        return Composite{ProcessingSegment{std::move(seg)}};
+    }
+}
--- a/cpp/arcticdb/vector_db/vector_db.cpp
+++ b/cpp/arcticdb/vector_db/vector_db.cpp
@@ -62,6 +62,11 @@ namespace arcticdb {
                                             col,
                                             slice_and_key.segment(store).field(idx).name()));
                                 }
+                                else {
+                                    internal::raise<ErrorCode::E_INVALID_ARGUMENT>(
+                                            "Vectors should exclusively comprise floats; the Python layer should otherwise raise an error and prevent upsertion of vectors containing non-float components."
+                                            );
+                                }
                             });
                         }
                     }
@@ -94,6 +99,11 @@ namespace arcticdb {
                     write_col->set_row_data(this->query_vector_.size());
                     seg.add_column(scalar_field(DataType::FLOAT64, column.name_), write_col);
                     }
+                else {
+                    internal::raise<ErrorCode::E_INVALID_ARGUMENT>(
+                            "Vectors should exclusively comprise floats; the Python layer should otherwise raise an error and prevent upsertion of vectors containing non-float components."
+                    );
+                }
             });
         }
         return Composite{ProcessingSegment{std::move(seg)}};
--- a/cpp/arcticdb/vector_db/vector_db.cpp
+++ b/cpp/arcticdb/vector_db/vector_db.cpp
@@ -31,7 +31,7 @@ namespace arcticdb {
         procs.broadcast(
                 [&store, &top_k, lp, this](const ProcessingSegment &proc) {
                     for (const auto& slice_and_key: proc.data()) {
-                        std::vector <std::shared_ptr<Column>> columns = slice_and_key.segment(store).columns();
+                        const std::vector<std::shared_ptr<Column>>& columns = slice_and_key.segment(store).columns();
                         for (auto&& [idx, col]: folly::enumerate(columns)) {
                             internal::check<ErrorCode::E_ASSERTION_FAILURE>(
                                     static_cast<long unsigned>(col->row_count()) == this->query_vector_.size(),
@@ -67,9 +67,8 @@ namespace arcticdb {
                     }
                 }
         );
-        // By this time we should have iterated through each block of each SegmentInMemory of each ProcessingSegment;
-        // and in each case read a vector and inserted it if more similar than the running top_k_. Now we write to a new
-        // segment.
+        // By this time we've read each vector, keeping a running count of the top-k vectors closest to the query vector
+        // of the vectors read so far. Now we write to a new segment.
         SegmentInMemory seg;
         seg.descriptor().set_index(IndexDescriptor(0, IndexDescriptor::ROWCOUNT));
         seg.set_row_id(query_vector_.size() - 1 + 1);
--- a/cpp/arcticdb/vector_db/vector_db.hpp
+++ b/cpp/arcticdb/vector_db/vector_db.hpp
@@ -39,11 +39,11 @@ namespace arcticdb {
     };
 
     struct TopK {
-        uint64_t _k;
+        uint64_t k_;
         std::set<NamedColumnSortedByDistance> top_k_;
-        double _furthest_distance{std::numeric_limits<double>::infinity()};
+        double furthest_distance_{std::numeric_limits<double>::infinity()};
 
-        explicit TopK(int k) : _k(k) {}
+        explicit TopK(int k) : k_(k) {}
 
         TopK() = delete;
 
@@ -64,27 +64,27 @@ namespace arcticdb {
             // NB We require <= so that the vector name acts as the tiebreaker when the distance is the same.
             // This is not guaranteed to the user but is useful for testing.
             bool inserted = false;
-            if (col.distance_ < _furthest_distance || (col.distance_ == _furthest_distance && col < *top_k_.rbegin())) {
+            if (col.distance_ < furthest_distance_ || (col.distance_ == furthest_distance_ && col < *top_k_.rbegin())) {
                 top_k_.insert(col);
                 inserted = true;
             }
-            // Obviously if col.distance_ < _furthest_distance then we've got a closer vector so we insert it.
-            // Where col.distance_ == _furthest_distance, the latter cannot be infinity, since in Python we require that
+            // Obviously if col.distance_ < furthest_distance_ then we've got a closer vector so we insert it.
+            // Where col.distance_ == furthest_distance_, the latter cannot be infinity, since in Python we require that
             // all vectors should be within the n-cube centred about the origin whose long diagonal has length
-            // std::numeric_limits<double>::max(). Therefore, _furthest_distance is finite. Therefore, it's been
+            // std::numeric_limits<double>::max(). Therefore, furthest_distance_ is finite. Therefore, it's been
             // altered, and so we have at least member of top_k_. So top_k_ is nonempty and we can compare
             // col < *top_k_.rbegin().
-            if (top_k_.size() > _k) {
+            if (top_k_.size() > k_) {
                 top_k_.erase(*top_k_.rbegin());
-                _furthest_distance = top_k_.rbegin()->distance_;
-            } else if (inserted && top_k_.size() == _k) {
-                _furthest_distance = top_k_.rbegin()->distance_;
+                furthest_distance_ = top_k_.rbegin()->distance_;
+            } else if (inserted && top_k_.size() == k_) {
+                furthest_distance_ = top_k_.rbegin()->distance_;
             }
             // There are three cases here.
             // - top_k.size() < k. Then we don't need to erase anything or refresh the furthest distance. When there are
             //   fewer than k elements in top_k_, we want to add any new vector we see; we may remove it later of course.
             // - top_k_.size() is exactly k. Then we don't need to erase anything. But we might have just inserted
-            //   something so we need to reload the _furthest_distance if we inserted something.
+            //   something so we need to reload the furthest_distance_ if we inserted something.
             // - top_k.size() = k+1. Then we do need to erase the furthest element. We also need to refresh the largest
             //   element.
             return inserted;
--- a/cpp/arcticdb/vector_db/vector_db.hpp
+++ b/cpp/arcticdb/vector_db/vector_db.hpp
@@ -1,0 +1,126 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <arcticdb/processing/clause.hpp>
+
+namespace arcticdb {
+
+    struct NamedColumnSortedByDistance {
+        double distance_;
+        std::shared_ptr<Column> column_;
+        std::string_view name_;
+
+        explicit NamedColumnSortedByDistance(double distance,
+                                             std::shared_ptr<Column> column,
+                                             std::string_view name) :
+                distance_(distance),
+                column_(column),
+                name_(name) {}
+
+        NamedColumnSortedByDistance() = delete;
+
+        ARCTICDB_MOVE_COPY_DEFAULT(NamedColumnSortedByDistance);
+
+        bool operator<(const NamedColumnSortedByDistance &that) const {
+            return distance_ < that.distance_ || (distance_ == that.distance_ && name_ < that.name_);
+            // name is a tiebreaker.
+        };
+
+        bool operator==(const NamedColumnSortedByDistance &that) const {
+            return distance_ == that.distance_ && column_ == that.column_ && name_ == that.name_;
+        };
+
+    };
+
+    struct TopK {
+        uint64_t _k;
+        std::set<NamedColumnSortedByDistance> top_k_;
+        double _furthest_distance{std::numeric_limits<double>::infinity()};
+
+        explicit TopK(int k) : _k(k) {}
+
+        TopK() = delete;
+
+        ARCTICDB_MOVE_COPY_DEFAULT(TopK);
+
+        bool try_insert(
+                double distance,
+                std::shared_ptr<Column> column,
+                std::string_view name
+        ) {
+            return try_insert(NamedColumnSortedByDistance(
+                    distance,
+                    column,
+                    name));
+        }
+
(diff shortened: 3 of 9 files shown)


cpp/arcticdb/processing/clause.hpp
```diff
@@ -110,6 +112,62 @@ struct PassthroughClause {
...
+    const std::string_view name;
+
+    explicit NamedColumnSortedByDistance(float_t similarity,
+                                         std::shared_ptr<Column> column,
+                                         std::string_view name) :
+            distance(similarity),
+            column(column),
+            name(name) {}
+
+    NamedColumnSortedByDistance() = delete;
```
alexowens90, line 158: This class can be `ARCTICDB_MOVE_COPY_DEFAULT` as well

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+    // We expect a vector in each column.
+    // top_k is updated as we read more vectors. We want to efficiently remove
+    // the 'worst' (least similar) vector as we go along to ensure we have the
+    // top k, not the top k+1.
+    auto lp = 2;
+    // Let's pretend that only the lp-norms exist. We'll use p=2 for now as a default.
+
+    procs.broadcast(
+            [&, lp] (ProcessingSegment &proc) { // todo: cleanup references
+                std::for_each(
```
alexowens90, line 684: Prefer range-based `for` loops for readability where possible

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+                        proc.data().begin(),
+                        proc.data().end(),
+                        [&] (pipelines::SliceAndKey s) {
+                            std::vector<std::shared_ptr<Column>> columns = s.segment(store).columns();
+                            for (auto column : folly::enumerate(columns)) {
```
alexowens90, line 689: I find:
`for (auto&& [idx, column] : folly::enumerate(columns))`
more readable personally

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+                                column.element->type().visit_tag([&, lp] (auto type_desc_tag) {
+                                    using TypeDescriptorTag = decltype(type_desc_tag);
+                                    using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
```
alexowens90, line 692: We should check that the column length and the query vector length match at this point.
Can use `internal::check<ErrorCode::E_INVALID_USER_ARGUMENT>(...)`

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+
+                                    ColumnData col_data = column.element->data();
+                                    double_t sum_differences_exponentiated = 0;
+
+                                    while (auto block = col_data.next<TypeDescriptorTag>()) {
+                                        auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                                        for (auto j = 0u; j < block.value().row_count(); ++j, ++ptr) {
```
alexowens90, line 699: `j` needs to be initialised outside the `while` loop, this will only currently be working because there is only one block.

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+                                            sum_differences_exponentiated += pow(abs(RawType(*ptr)-query_vector[j]), lp);
```
alexowens90, line 700: I don't think the cast to `RawType` is needed?

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+
+
+            auto col_data = column.column->data();
+            auto write_ptr = reinterpret_cast<RawType *>(write_col->ptr());
+            while (auto block = col_data.next<TypeDescriptorTag>()) {
+                auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                for (auto i = 0u; i < block.value().row_count(); ++i, ++ptr) {
+                    *write_ptr = RawType(*ptr);
+                    ++write_ptr;
+                }
```
alexowens90, line 751: I think we can replace this `for` loop with a `memcpy`?

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+            }
+            *write_ptr = RawType(pow(column.distance, 1/static_cast<float>(lp)));
+            write_col->set_row_data(query_vector.size());
+            seg.add_column(scalar_field(DataType::FLOAT64, column.name), write_col);
+        });
+        }
+    Composite<ProcessingSegment> output;
+    ProcessingSegment segment = ProcessingSegment{std::move(seg)};
```
alexowens90, line 759: `ProcessingSegment segment(std::move(seg))` is more idiomatic

cpp/arcticdb/processing/clause.cpp
```diff
@@ -658,4 +658,107 @@ std::string DateRangeClause::to_string() const {
...
+    auto descriptor = segment.data().at(0); // now empty
```
alexowens90, line 760: Dead code?

alexowens90 (changes requested):

cpp/arcticdb/processing/clause.cpp
octogenary: I believe last time I did this I had some very strange problem in which somehow I ended up with two blocks. I have no idea why but am trying it again.

octogenary: Well, whatever problem I managed to incur last time has disappeared.

cpp/arcticdb/processing/clause.cpp
octogenary: Yes, debugging stuff; sorry, should've caught this.

cpp/arcticdb/processing/clause.hpp
```diff
@@ -110,6 +112,62 @@ struct PassthroughClause {
...
+
+    [[nodiscard]] std::string to_string() const;
+};
+
+struct NamedColumnSortedByDistance {
+    const double_t distance;
+    const std::shared_ptr<Column> column;
+    const std::string_view name;
+
+    explicit NamedColumnSortedByDistance(float_t similarity,
```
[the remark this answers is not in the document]
octogenary: No.

cpp/arcticdb/processing/clause.hpp
octogenary: Seemingly not since this triggers the error in Mac OS compilation [here](https://github.com/man-group/ArcticDB/actions/runs/5893310940/job/15984573949?pr=734).

python/tests/integration/arcticdb/test_vector_db.py
```diff
@@ -0,0 +1,229 @@
...
+    vector_db.upsert("first_is_first", first_df)
+    vector_db.upsert("first_is_first", first_df)
+    vector_db.upsert("first_is_first", first_df)
+
+    assert_frame_equal(vector_db.read("first_is_first"), vector_db.read("second_is_first"), check_like=True)
+
+
+@use_of_function_scoped_fixtures_in_hypothesis_checked
+@given(df=data_frames([column("a", elements=numeric_type_strategies())], index=range_indexes()))
+def test_upsertion_idempotent(arctic_client, df):
```
alexowens90, line 228: Needs fleshing out

python/tests/integration/arcticdb/test_vector_db.py
```diff
@@ -0,0 +1,229 @@
...
+        WritePayload,
+        ArcticUnsupportedDataTypeException,
+        ReadRequest,
+        ReadInfoRequest,
+        ArcticInvalidApiUsageException,
+        StagedDataFinalizeMethod,
+    )
+
+
+def test_top_k(arctic_client):
```
alexowens90, line 51: This would be a good candidate for a hypothesis test. Hypothesis will generate random dataframes (see `test_filtering.py` for some examples). Then we can generate a random value for `k` that is less than the number of columns, and the rest of the test will stay the same. It will probably take a long time to run though, so will need to tune the maximum number of samples.

python/arcticdb/version_store/vector_db.py
```diff
@@ -0,0 +1,465 @@
...
+    """
+
+    def __init__(self, library: Library):
+        """
+        Parameters
+        ----------
+        library
+            The library in which vectors will be stored. See `Arctic.create_library`.
+        dimensions
+            The number of dimensions those vectors should have.
```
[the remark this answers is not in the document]
octogenary: I think this is from an outdated version? It's not in mine presently.

cpp/arcticdb/vector_db/vector_db.hpp
```diff
@@ -0,0 +1,70 @@
...
+    const std::string_view name_;
+
+    explicit NamedColumnSortedByDistance(double_t distance,
+                                         std::shared_ptr<Column> column,
+                                         std::string_view name) :
+            distance_(distance),
+            column_(column),
+            name_(name) {}
+
+    NamedColumnSortedByDistance() = delete;
```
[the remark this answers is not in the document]
octogenary: Ah righto.

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,106 @@
...
+        double furthest_distance = std::numeric_limits<double>::infinity();
+        // We expect a vector in each column.
+        // top_k is updated as we read more vectors. We want to efficiently remove
+        // the 'worst' (least similar) vector as we go along to ensure we have the
+        // top k, not the top k+1.
+        auto lp = 2;
+        // Let's pretend that only the lp-norms exist. We'll use p=2 for now as a default.
+
+        procs.broadcast(
+                [&query_vector=query_vector, &k=k, &store, &furthest_distance, &top_k, lp](const ProcessingSegment &proc) {
```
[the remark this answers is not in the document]
octogenary: No, it wasn't; my compiler complains when I try `&query_vector` because it `doesn't name a variable`.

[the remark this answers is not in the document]
octogenary: Something like that happens [here](https://github.com/man-group/ArcticDB/blob/be7e70df588416119d64656a26b6f32d7dcd67b1/cpp/arcticdb/pipeline/slicing.cpp#L114).

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,106 @@
...
+                            col->type().visit_tag([&query_vector=query_vector, &col=col, &idx=idx, &k, &furthest_distance, &top_k, &slice_and_key, &store, lp](auto type_desc_tag) {
+                                using TypeDescriptorTag = decltype(type_desc_tag);
+                                using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
+
+                                ColumnData col_data = col->data();
+                                double sum_differences_exponentiated = 0;
+
+                                while (auto block = col_data.next<TypeDescriptorTag>()) {
+                                    auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                                    auto j = 0u;
```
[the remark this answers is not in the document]
octogenary: argh yes

python/arcticdb/version_store/vector_db.py
```diff
@@ -0,0 +1,491 @@
...
+        library
+            The library in which vectors will be stored. See `Arctic.create_library`. In initialising the library, we
+            recommend using the `new_vector_db` method.
+        """
+        if type(library) is not Library:
+            raise ArcticUnsupportedDataTypeException(
+                f"Vector databases must be initialised on libraries. You tried to initialise from a {type(library)}."
+            )
+        if library.list_symbols():
+            raise ArcticUnsupportedDataTypeException(
```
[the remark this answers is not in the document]
octogenary: Oh drat, that would be precluded yes.

python/arcticdb/version_store/vector_db.py
```diff
@@ -0,0 +1,491 @@
...
+            )
+        self._max_vector_size = library._nvs._library.config.write_options.segment_row_size - 1
+        # we put distances at the bottom of columns so have to decrement by 1
+        self._lib = library
+
+    def __repr__(self):
+        return f"VectorDB({str(self._lib)})"
+
+    def __contains__(self, namespace: str):
+        return f"{VECTOR_DB_DISTINGUISHED_PREFIX}{namespace}" in self._lib
```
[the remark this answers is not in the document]
octogenary: Hm I suppose a distinguished suffix can do the same thing.

python/arcticdb/version_store/vector_db.py
```diff
@@ -0,0 +1,491 @@
...
+    A `pd.DataFrame` suitable for writing to an Arctic symbol internally.
+
+    Raises
+    ------
+    ArcticUnsupportedDataTypeException
+        if any of the values in the frame has a non-numeric type.
+    ValueError
+        if the length of the vectors supplied is unexpected.
+    """
+    if any([t not in numeric_types for t in vectors.dtypes.unique()]):
```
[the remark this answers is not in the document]
octogenary: I think faiss only supports floats (actually not even doubles) but we automatically cast to float anyway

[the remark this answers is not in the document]
octogenary: Oh I see the comment below.

python/tests/integration/arcticdb/test_vector_db.py
octogenary: Hm I might just get rid of it since `test_vector_db_upsertion` tets idempotence, or maybe I'll use the same setup as the hypothesis test.

cpp/arcticdb/vector_db/vector_db.cpp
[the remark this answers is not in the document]
vasil-pashov: In this case I would suggest to capture `this` instead.

cpp/arcticdb/vector_db/vector_db.hpp
```diff
@@ -0,0 +1,70 @@
...
+    bool operator== (const NamedColumnSortedByDistance& that) const {
+        return distance_ == that.distance_;
+    };
+
+};
+
+
+struct TopKClause {
+    ClauseInfo clause_info_;
+    std::vector<float> query_vector;
```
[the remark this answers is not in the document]
octogenary: Yes oops

python/arcticdb/version_store/vector_db.py
```diff
@@ -0,0 +1,491 @@
...
+                f"VectorDB {self} has maximum vector size {self._max_vector_size}, but you attempted to "
+                f"upsert vectors of size {data_frame_to_upsert.shape[0]}. To store vectors with more than "
+                f"{self._max_vector_size} vectors, you will have to initialise a new VectorDB on a "
+                "fresh library like so:\n\n"
+                ">>> arctic_client.create_library('nomen', LibraryOptions(rows_per_segment = k)\n\n"
+                "where k is the maximum vector size."
+            )
+        if symbol_name in self._lib:
+            warnings.warn("This is extremely memory-inefficient!")
+            old_df = self.read(namespace)
```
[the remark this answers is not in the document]
octogenary: I'm slightly worried that `NoSuchVersionException` will be broader than `symbol_name not in self._lib` but have changed this anyway.

python/arcticdb/version_store/vector_db.py
```diff
@@ -0,0 +1,491 @@
...
+        if qv.shape[0] != expected_dimensions:
+            raise ValueError(
+                "Query vectors must have the same number of components "
+                f"({expected_dimensions}) "
+                "as the vectors in the VectorDB. The vector given had "
+                f"{qv.shape[0]} components."
+            )
+        q = QueryBuilder()
+        q.clauses.append(_TopKClause(qv, k))
+        q._python_clauses.append(PythonTopKClause(qv, k))
```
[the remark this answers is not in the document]
octogenary: Shall I just leave it hidden? I'm trying to make it relatively inaccessible.

python/tests/integration/arcticdb/test_vector_db.py
```diff
@@ -0,0 +1,229 @@
...
+    dimensions, number_of_vectors, k = 15000, 200, 5
+    string_column_names = [
+        "".join(random.choices(string.ascii_uppercase + string.digits, k=10)) for _ in range(number_of_vectors)
+    ]
+
+    ac.create_library("pytest_test_top_k")
+    vector_db = VectorDB(ac["pytest_test_top_k"])
+
+    df = pd.DataFrame(np.random.rand(dimensions, number_of_vectors), columns=string_column_names)
+    qv = np.array([0] * dimensions)
```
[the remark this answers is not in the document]
octogenary: Yes all right. (Although the hypothesis test now does that too.)

python/tests/integration/arcticdb/test_vector_db.py
octogenary: (did the latter)

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,102 @@
...
+                                            col,
+                                            slice_and_key.segment(store).field(idx).name()));
+                                }
+                            });
+                        }
+                    }
+                }
+        );
+        // By this time we should have iterated through each block of each SegmentInMemory of each ProcessingSegment;
+        // and in each case read a vector and inserted it if more similar than the running top_k_. Now we write to a new
```
octogenary, line 71: After review I'll rephrase this comment (which surely can't cause CI problems...) to something like 'By this time we've read each vector, keeping a running count of the top-k vectors closest to the query vector of the vectors read so far. Now we write to a new segment.'

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,102 @@
...
+                                                                          block.value().row_count(); ++progress_in_block, ++j, ++ptr) {
+                                            sum_differences_exponentiated += pow(
+                                                    abs(*ptr - this->query_vector_[j]), lp);
+                                        }
+                                    }
+
+                                    top_k.try_insert(NamedColumnSortedByDistance(
+                                            sum_differences_exponentiated,
+                                            col,
+                                            slice_and_key.segment(store).field(idx).name()));
```
octogenary, line 63: Just noticed this is probably quite inefficient since we check the name of each vector even if we don't insert it, but I suppose I can refactor this when parallelising (just dealing iwth all obvious performance improvements).

willdealtry: It's just a very small reference class, nothing gets copied so unless it actually shows up in a performance testing I think it's probably okay

cpp/arcticdb/vector_db/vector_db.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+        bool operator==(const NamedColumnSortedByDistance &that) const {
+            return distance_ == that.distance_ && column_ == that.column_ && name_ == that.name_;
+        };
+
+    };
+
+    struct TopK {
+        uint64_t _k;
+        std::set<NamedColumnSortedByDistance> top_k_;
```
willdealtry, line 43: nit-picking, but underscores after the variable in C++

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,102 @@
...
+        // top_k_ is updated as we read more vectors. We want to efficiently remove
+        // the 'worst' (least similar) vector as we go along to ensure we have the
+        // top k, not the top k+1.
+        auto lp = 2;
+        // Let's pretend that only the lp-norms exist. We'll use p=2 for now as a default.
+
+        procs.broadcast(
+                [&store, &top_k, lp, this](const ProcessingSegment &proc) {
+                    for (const auto& slice_and_key: proc.data()) {
+                        std::vector <std::shared_ptr<Column>> columns = slice_and_key.segment(store).columns();
```
willdealtry, line 34: Could this be const ref instead of copying the vector here?

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,102 @@
...
+        seg.set_row_id(query_vector_.size() - 1 + 1);
+        // -1 because of zero-indexing, +1 because we want to add the distance on at the end.
+            for (const auto &column: top_k.top_k_) {
+            // todo: replace (mem)cpying of columns' contents with holding of pointers in top_k_ and addition of pointers to seg
+            // this doesn't work presently because the columns can have >1 block.
+            column.column_->type().visit_tag([&column, &seg, lp, this](auto type_desc_tag) {
+                using TypeDescriptorTag = decltype(type_desc_tag);
+                using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
+
+                if constexpr (is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)) {
```
willdealtry, line 84: Do we want to add an exception when the column is of some other type, or is silently skipping them okay?

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,102 @@
...
+                    auto col_data = column.column_->data();
+                    auto write_ptr = reinterpret_cast<RawType *>(write_col->ptr());
+                    while (auto block = col_data.next<TypeDescriptorTag>()) {
+                        auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                        std::memcpy(write_ptr, ptr, block.value().row_count() * sizeof(RawType));
+                        write_ptr += block.value().row_count();
+                    }
+                    *write_ptr = RawType(pow(column.distance_, 1 / static_cast<double>(lp)));
+                    write_col->set_row_data(this->query_vector_.size());
+                    seg.add_column(scalar_field(DataType::FLOAT64, column.name_), write_col);
```
willdealtry, line 96: std::move(write_col)

cpp/arcticdb/vector_db/vector_db.hpp
octogenary: Hm I think I had some notion of trying to indicate it'd be private and forgot I could simply _make_ it private were I bothered.

cpp/arcticdb/vector_db/vector_db.cpp
octogenary: Alex suggested this as a compiler optimisation I think? This is already checked in Python.

cpp/arcticdb/vector_db/vector_db.cpp
octogenary: As a replacement argument to add_column?

cpp/arcticdb/vector_db/vector_db.hpp
```diff
@@ -0,0 +1,126 @@
...
+        };
+
+        bool operator==(const NamedColumnSortedByDistance &that) const {
+            return distance_ == that.distance_ && column_ == that.column_ && name_ == that.name_;
+        };
+
+    };
+
+    struct TopK {
+        uint64_t _k;
```
alexowens90, line 42: k_

cpp/arcticdb/vector_db/vector_db.cpp
```diff
@@ -0,0 +1,102 @@
...
+                            internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+                                    static_cast<long unsigned>(col->row_count()) == this->query_vector_.size(),
+                                    "Expected vector of length {}, got vector of length {}.",
+                                    col->row_count(),
+                                    this->query_vector_.size());
+                            col->type().visit_tag([&top_k, &slice_and_key, &store, &col=col, idx=idx, lp, this](auto type_desc_tag) {
+                                using TypeDescriptorTag = decltype(type_desc_tag);
+                                using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
+
+                                if constexpr (is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)) {
```
alexowens90, line 45: Add an `else{internal::raise...}`, otherwise it will silently do nothing/fail later with non-float columns

cpp/arcticdb/vector_db/vector_db.hpp
```diff
@@ -0,0 +1,126 @@
...
+        std::set<NamedColumnSortedByDistance> top_k_;
+        double _furthest_distance{std::numeric_limits<double>::infinity()};
```
alexowens90, line 44: furthest_distance_
Can you check the rest of the diff for these as well?

cpp/arcticdb/vector_db/vector_db.cpp
octogenary: Yes.

cpp/arcticdb/vector_db/vector_db.cpp
alexowens90: Same as above, add the else to throw so it's super explicit

cpp/arcticdb/vector_db/vector_db.cpp
octogenary: Never mind clang-tidy points out `[p]assing result of std::move() as a const reference argument; no move will actually happen`

cpp/arcticdb/vector_db/vector_db.cpp
octogenary: Going to push some changes given William's comments below so I'll do this too.

python/arcticdb/version_store/vector_db.py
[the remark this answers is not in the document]
alexowens90: The reason is that we do not support transactions, and so `symbol_name in self._lib` being `True` does not guarantee that `self.read(namespace)` will not throw, as the data could be deleted in between.
This is also a general Pythonic principle, "its better to ask for forgiveness than permission"
https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

python/arcticdb/version_store/vector_db.py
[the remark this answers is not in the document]
alexowens90: Yes, hidden is fine

python/arcticdb/vector_db/vector_db.py
```diff
@@ -0,0 +1,521 @@
...
+import numpy as np
+
+from collections import namedtuple
+from typing import Union, Optional, Collection, List, Mapping
+
+from arcticdb import Arctic
+from arcticdb.options import LibraryOptions
+from arcticdb.version_store.library import Library, ArcticUnsupportedDataTypeException
+from arcticdb.version_store.processing import QueryBuilder
+from arcticdb_ext.version_store import TopKClause as _TopKClause, NoSuchVersionException
```
alexowens90, line 12: The clause should be an unused import now

cpp/arcticdb/vector_db/vector_db.hpp
octogenary: Yep just pushed

cpp/arcticdb/vector_db/vector_db.hpp
octogenary: Yes sorry - see comment to William above; should have just pushed a fix.

python/arcticdb/vector_db/vector_db.py
```diff
@@ -0,0 +1,521 @@
...
+                f"upsert vectors of size {data_frame_to_upsert.shape[0]}. To store vectors with more than "
+                f"{self._max_vector_size} vectors, you will have to initialise a new VectorDB on a "
+                "fresh library like so:\n\n"
+                ">>> arctic_client.create_library('nomen', LibraryOptions(rows_per_segment = k)\n\n"
+                "where k is the maximum vector size."
+            )
+        if data_frame_to_upsert.isnull().values.any():
+            raise ValueError("Vectors upserted must not contain NaNs.")
+        maximum_component = ((0.5 * np.finfo(np.float64).max) ** 0.5) / actual_dimensions
+        if (abs(data_frame_to_upsert).values > maximum_component).any():
```
alexowens90, line 353: The `isnull()` check and this one are likely to be quite expensive. Not for now, but worth bearing in mind when profiling.

alexowens90 (changes requested):

python/arcticdb/version_store/vector_db.py
```diff
@@ -0,0 +1,465 @@
...
+        """
+        Parameters
+        ----------
+        namespace
+            The namespace to which `vectors` should be upserted.
+        vectors
+            In the case of a mapping, we expect a mapping whose keys are strings
+            (taken as identifiers of vectors), and whose values are in turn mapping
+            minimally containing a key-value pair 'value' pointing to something that
+            yields a one-dimensional `np.ndarray` of numeric types corresponding to a
```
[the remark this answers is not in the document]
octogenary: Oh, well 'something that yields a one-dimensional `np.ndarray`' could be a one-dimensional, but that could be clearer.

python/arcticdb/vector_db/vector_db.py
octogenary: Yep

python/arcticdb/vector_db/vector_db.py
octogenary: Ok

alexowens90 (approved):

Issue #613, opened 2023-07-18
`read_metadata_batch` should catch all exceptions from individual queries and return what data it can


alexowens90:
See #612

PR #748, opened 2023-08-15
Read metadata batch to catch all exceptions

GemDot:
Closes https://github.com/man-group/ArcticDB/issues/613

The previous behaviour of Library.read_metadata_batch and NativeVersionStore.batch_read_metadata was to throw an exception if there was an issue related to a missing key. For backwards compatibility reasons, this behaviour has been maintained for NativeVersionStore.batch_read_metadata.

For Library.read_metadata_batch, this has been changed. A DataError object will now be returned in the position in the list returned corresponding to the symbol/version pair there was an issue retrieving. This contains the symbol and version requested, and the exception string thrown. For one well-defined category of error, the error category and specific error code are also included:

- ErrorCategory.STORAGE - ErrorCode.E_KEY_NOT_FOUND: The index key required to read the specified version does not exist in the storage.

- ErrorCategory.MISSING_DATA - ErrorCode.E_NO_SUCH_VERSION: The version requested by the user does not exist

Otherwise these fields of the DataError object are left as None.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1606,25 +1606,52 @@ folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobu
     }
 }
 
-folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::get_metadata_async(
-    folly::Future<std::optional<AtomKey>>&& version_fut){
+folly::Future<std::pair<VariantKey, std::optional<google::protobuf::Any>>> LocalVersionedEngine::get_metadata_async(
+    folly::Future<std::optional<AtomKey>>&& version_fut,
+    const StreamId& stream_id,
+    const VersionQuery& version_query
+    ) {
     return  std::move(version_fut)
-    .thenValue([this](std::optional<AtomKey>&& key){
+    .thenValue([this, &stream_id, &version_query](std::optional<AtomKey>&& key){
+        missing_data::check<ErrorCode::E_NO_SUCH_VERSION>(key.has_value(),
+        "Unable to retrieve  metadata. {}@{}: version not found", stream_id, version_query);
         return get_metadata(std::move(key));
+    })
+    .thenValue([](auto&& metadata){
+        auto&& [opt_key, meta_proto] = metadata;
+        return std::make_pair(std::move(*opt_key), std::move(meta_proto));
     });
 }
-
-std::vector<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::batch_read_metadata_internal(
+ 
+std::vector<std::variant<std::pair<VariantKey, std::optional<google::protobuf::Any>>, DataError>> LocalVersionedEngine::batch_read_metadata_internal(
     const std::vector<StreamId>& stream_ids,
     const std::vector<VersionQuery>& version_queries,
     const ReadOptions& read_options
     ) {
-    auto versions_fut = batch_get_versions_async(store(), version_map(), stream_ids, version_queries, read_options.read_previous_on_failure_);
-    std::vector<folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>>> fut_vec;
-    for (auto&& version: versions_fut){
-        fut_vec.push_back(get_metadata_async(std::move(version)));
+    auto version_futures = batch_get_versions_async(store(), version_map(), stream_ids, version_queries, read_options.read_previous_on_failure_);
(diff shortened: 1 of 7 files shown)


python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -61,112 +61,9 @@ def generate_dataframe(columns, dt, num_days, num_rows_per_day):
...
-    assert results_list[1].metadata == {"meta2": 5}
-    assert results_list[2].metadata == {"meta3": 6}
-    assert results_list[3].metadata == {"meta1": 7}
-    assert results_list[4].metadata == {"meta2": 8}
-
-
 def test_read_meta_batch_with_as_ofs(arctic_library):
     lib = arctic_library
-    num_symbols = 10
+    num_symbols = 2
```
alexowens90, line 66: This test can just be removed, it is superseded by the new one with the same name

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -241,6 +138,144 @@ def test_write_metadata_batch_with_none(arctic_library):
...
+    assert not isinstance(batch[2], DataError)
+    assert batch[2].metadata == {"meta3": 0}
+
+
+def test_read_metadata_batch_symbol_doesnt_exist(arctic_library):
+    lib = arctic_library
+
+    # Given
+    df = pd.DataFrame({"a": [3, 5, 7, 9]}, index=pd.date_range(start="1/1/2018", end="1/4/2018"))
+    df.index.rename("named_index", inplace=True)
```
alexowens90, line 238: Why?

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -241,6 +138,144 @@ def test_write_metadata_batch_with_none(arctic_library):
...
+    assert batch[1].error_code == ErrorCode.E_NO_SUCH_VERSION
+    assert batch[1].error_category == ErrorCategory.MISSING_DATA
+
+
+def test_read_metadata_batch_version_doesnt_exist(arctic_library):
+    lib = arctic_library
+
+    # Given
+    df1 = pd.DataFrame({"a": [3, 5, 7]}, index=pd.date_range(start="1/1/2018", end="1/3/2018"))
+    df1.index.rename("named_index", inplace=True)
```
alexowens90, line 261: Why?

python/arcticdb/version_store/_store.py
```diff
@@ -1015,34 +1015,29 @@ def batch_read_metadata(
...
+                    meta_items.append(
+                        VersionedItem(
+                            symbol=vitem.symbol,
+                            library=self._library.library_path,
+                            data=None,
+                            version=vitem.version,
+                            metadata=meta,
+                            host=self.env,
+                        )
+                    )
```
alexowens90, line 1034: This code is now duplicated in `library.py`. What I had in mind was for `_batch_read_metadata_or_error` to do the denormalization.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1607,24 +1607,47 @@ folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobu
 }
 
 folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> LocalVersionedEngine::get_metadata_async(
```
alexowens90, line 1609: I think this function should return
`folly::Future<std::pair<VariantKey, std::optional<google::protobuf::Any>>>`
as the `VariantKey` is never `std::nullopt` due to the check.
This will also simplify the logic in `PythonVersionStore::batch_read_metadata`

alexowens90 (changes requested):

python/tests/integration/arcticdb/test_arctic_batch.py
GemDot: Actually, it is my fault that when I did the grep I couldn't see this one

python/tests/integration/arcticdb/test_arctic_batch.py
GemDot: yes, no reason

python/tests/integration/arcticdb/test_arctic_batch.py
GemDot: No reason, removed

cpp/arcticdb/version/local_versioned_engine.cpp
GemDot: good point

python/arcticdb/version_store/_store.py
GemDot: AS discussed, we will add a flag to differentiate between both logics (library and _store.py)

alexowens90 (approved):

PR #751, opened 2023-08-15
Delete pytest timings file

willdealtry:
I don't know why we have this massive thing checked in at all


qc00 (changes requested):
The test runner uses this to split the work between two jobs.

Due to an unknown issue in our native extension, the normal way of running tests in parallel in Python, which forks the Python process, often deadlock.

So I had to use an external script which starts two independent pytest processes. The only way to distribute work between the two is to this external timing file.

See https://jerry-git.github.io/pytest-split/ for details

mehertz:
> The test runner uses this to split the work between two jobs.
> 
> Due to an unknown issue in our native extension, the normal way of running tests in parallel in Python, which forks the Python process, often deadlock.
> 
> So I had to use an external script which starts two independent pytest processes. The only way to distribute work between the two is to this external timing file.
> 
> See https://jerry-git.github.io/pytest-split/ for details

Doesn't that mean we need to keep this up to date?

qc00:
> Doesn't that mean we need to keep this up to date?

For the tests not in the file, they will be arbitrarily assigned to one of the groups. This is inefficient if the slow tests are assigned to the same group, which is where this file is meant to resolve.

willdealtry:
> The test runner uses this to split the work between two jobs.
> 
> Due to an unknown issue in our native extension, the normal way of running tests in parallel in Python, which forks the Python process, often deadlock.
> 
> So I had to use an external script which starts two independent pytest processes. The only way to distribute work between the two is to this external timing file.
> 
> See https://jerry-git.github.io/pytest-split/ for details

I feel like if we have deadlocks we should investigate them, not engage in byzantine workarounds. We have tests that check we can fork without incident, the question we should be asking is why do they not work in this situation?

PR #761, opened 2023-08-17
Add uri support for mongodb

phoebusm:
#### Reference Issues/PRs
https://github.com/man-group/arcticdb-man/issues/26

Add uri support for mongodb-backed storage
#### Any other comments?
Due to the backward compatibility mongodb options MinPoolSize, MaxPoolSize and SelectionTimeoutMs, a relatively complicated design for connection string handling is adopted. In the old fashioned way, the three options are added by the connection string as
```
return fmt::format("{}&minPoolSize={}&maxPoolSize={}&serverSelectionTimeoutMS={}",
            uri, min_pool_size, max_pool_size, selection_timeout_ms);
```
The three options are supported by mongodb connection string so user may add them in the uri, the way supported in this PR.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/build_tooling/parallel_test.sh
+++ b/build_tooling/parallel_test.sh
@@ -13,8 +13,11 @@ catch=`{ which catchsegv 2>/dev/null || echo ; } | tail -n 1`
 function worker() {
     group=${1:?Must pass the group id argument}
     shift
+    duration_file="$TEST_OUTPUT_DIR/pytest-durations.$group.json"
     new_root=$PARALLEL_TEST_ROOT/$group
     set -o xtrace -o pipefail
+
+    cp "$tooling_dir/pytest-durations.json" "$duration_file"
 
     # Build a directory that's just the test assets, so can't access other Python source not in the wheel
     # Each test also get a separate directory since there's a mystery lock somewhere preventing concurrent runs (even)
@@ -25,7 +28,7 @@ function worker() {
 
     $catch python -m pytest -v --show-capture=no --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
         --junitxml="$TEST_OUTPUT_DIR/pytest.$group.xml" \
-        --splits $splits --group $group \
+        --splits $splits --group $group --durations-path="$duration_file" --store-durations \
         --basetemp="$new_root/temp-pytest-output" \
         "$@" 2>&1 | sed -ur "s#^(tests/.*/([^/]+\.py))?#$group: \2#"
 }
--- a/build_tooling/parallel_test.sh
+++ b/build_tooling/parallel_test.sh
@@ -13,11 +13,8 @@ catch=`{ which catchsegv 2>/dev/null || echo ; } | tail -n 1`
 function worker() {
     group=${1:?Must pass the group id argument}
     shift
-    duration_file="$TEST_OUTPUT_DIR/pytest-durations.$group.json"
     new_root=$PARALLEL_TEST_ROOT/$group
     set -o xtrace -o pipefail
-
-    cp "$tooling_dir/pytest-durations.json" "$duration_file"
 
     # Build a directory that's just the test assets, so can't access other Python source not in the wheel
     # Each test also get a separate directory since there's a mystery lock somewhere preventing concurrent runs (even)
@@ -28,7 +25,7 @@ function worker() {
 
     $catch python -m pytest -v --show-capture=no --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
         --junitxml="$TEST_OUTPUT_DIR/pytest.$group.xml" \
-        --splits $splits --group $group --durations-path="$duration_file" --store-durations \
+        --splits $splits --group $group \
         --basetemp="$new_root/temp-pytest-output" \
         "$@" 2>&1 | sed -ur "s#^(tests/.*/([^/]+\.py))?#$group: \2#"
 }
--- a/python/arcticdb/adapters/azure_library_adapter.py
+++ b/python/arcticdb/adapters/azure_library_adapter.py
@@ -86,9 +86,7 @@ class AzureLibraryAdapter(ArcticLibraryAdapter):
         return lib._library
 
     def _parse_query(self, query: str) -> ParsedQuery:
-        if query and query.startswith("?"):
-            query = query.strip("?")
-        elif not query:
+        if not query:
             raise ValueError(f"Invalid Azure URI. Missing query parameter")
 
         parsed_query = re.split("[;]", query)
--- a/python/arcticdb/adapters/mongo_library_adapter.py
+++ b/python/arcticdb/adapters/mongo_library_adapter.py
@@ -1,0 +1,70 @@
+"""
+Copyright 2023 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+from arcticdb.options import LibraryOptions
+from arcticc.pb2.storage_pb2 import EnvironmentConfigsMap, LibraryConfig
+from arcticdb.version_store.helper import add_mongo_library_to_env
+from arcticdb.config import _DEFAULT_ENV
+from arcticdb.version_store._store import NativeVersionStore
+from arcticdb.adapters.arctic_library_adapter import ArcticLibraryAdapter, set_library_options
+from arcticdb_ext.storage import Library
+from arcticdb.encoding_version import EncodingVersion
+from arcticdb.exceptions import UserInputException
+from collections import namedtuple
+from dataclasses import dataclass, fields
+import pymongo
+
+PARSED_QUERY = namedtuple("PARSED_QUERY", ["region"])
+
+
+class MongoLibraryAdapter(ArcticLibraryAdapter):
+    @staticmethod
+    def supports_uri(uri: str) -> bool:
+        return uri.startswith("mongodb://")
+
+    def __init__(self, uri: str, encoding_version: EncodingVersion, *args, **kwargs):
+        try:
+            parameters = pymongo.uri_parser.parse_uri(uri)
+            self._endpoint = f"{parameters['nodelist'][0][0]}:{parameters['nodelist'][0][1]}"
+        except Exception as e:
+            raise UserInputException(
+                f"Invalid connection string format. {e} Correct format: mongodb://[HOST]/[DATABASE][?OPTIONS]"
+            )
+        self._uri = uri
+        self._encoding_version = encoding_version
+
+        super().__init__(uri, self._encoding_version)
+
+    def __repr__(self):
+        return f"mongodb(endpoint={self._endpoint})"
+
+    @property
+    def config_library(self):
+        env_cfg = EnvironmentConfigsMap()
+
+        add_mongo_library_to_env(cfg=env_cfg, lib_name=self.CONFIG_LIBRARY_NAME, env_name=_DEFAULT_ENV, uri=self._uri)
+
+        lib = NativeVersionStore.create_store_from_config(
+            env_cfg, _DEFAULT_ENV, self.CONFIG_LIBRARY_NAME, encoding_version=self._encoding_version
+        )
+
+        return lib._library
+
+    def create_library(self, name, library_options: LibraryOptions):
+        env_cfg = EnvironmentConfigsMap()
+
+        add_mongo_library_to_env(cfg=env_cfg, lib_name=name, env_name=_DEFAULT_ENV, uri=self._uri)
+        library_options.encoding_version = (
+            library_options.encoding_version if library_options.encoding_version is not None else self._encoding_version
+        )
+        set_library_options(env_cfg.env_by_id[_DEFAULT_ENV].lib_by_path[name], library_options)
+
+        lib = NativeVersionStore.create_store_from_config(
+            env_cfg, _DEFAULT_ENV, name, encoding_version=library_options.encoding_version
+        )
+
+        return lib
--- a/python/arcticdb/util/test.py
+++ b/python/arcticdb/util/test.py
@@ -35,7 +35,7 @@ from arcticdb_ext import set_config_int, get_config_int, unset_config_int
 # leave out Mongo as spinning up a Mongo instance in Windows CI is fiddly, and Mongo support is only
 # currently required for Linux for internal use.
 # We also skip it on Mac as github actions containers don't work with macos
-RUN_MONGO_TEST = True if sys.platform == "linux" else False
+RUN_MONGO_TEST = sys.platform == "linux"
 
 
 def maybe_not_check_freq(f):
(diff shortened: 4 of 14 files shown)


python/arcticdb/adapters/mongo_library_adapter.py
```diff
@@ -0,0 +1,102 @@
...
+
+        add_mongo_library_to_env(
+            cfg=env_cfg, lib_name=self.CONFIG_LIBRARY_NAME, env_name=_DEFAULT_ENV, uri=self._endpoint
+        )
+
+        lib = NativeVersionStore.create_store_from_config(env_cfg, _DEFAULT_ENV, self.CONFIG_LIBRARY_NAME)
+
+        return lib._library
+
+    def _parse_query(self, query: str) -> ParsedQuery:
```
poodlewars, line 83: Same problem as Azure, this is going to give confusing errors if the user gets the query format a bit wrong

python/arcticdb/version_store/helper.py
poodlewars, line 1: Changes in this file not required?

python/arcticdb/adapters/mongo_library_adapter.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class MongoLibraryAdapter(ArcticLibraryAdapter):
+    REGEX = r"(?P<endpoint>mongodb://[^?]+)(?P<options>\?.*)?"
+
+    @staticmethod
+    def supports_uri(uri: str) -> bool:
+        return uri.startswith("mongodb://")
+
+    def __init__(self, uri: str, *args, **kwargs):
```
poodlewars, line 41: Any way to use `pymongo` for this? Something like https://pymongo.readthedocs.io/en/stable/api/pymongo/uri_parser.html

Doing the query parsing well is harder than it looks, especially to give good validation and user facing errors...

python/arcticdb/adapters/mongo_library_adapter.py
```diff
@@ -0,0 +1,102 @@
...
+        self._query_params: ParsedQuery = ParsedQuery(**additional_options)
+        self._min_pool_size = self._query_params.min_pool_size
+        self._max_pool_size = self._query_params.max_pool_size
+        self._selection_timeout_ms = self._query_params.selection_timeout_ms
+
+        super().__init__(uri, EncodingVersion.V1)
+
+    def __repr__(self):
+        return (
+            f"mongodb(endpoint={self._endpoint}, min_pool_size={self._min_pool_size},"
```
poodlewars, line 67: `self._endpoint` may contain authentication info, which should be obfuscated in the repo. Again hopefully `pymongo` can help with this, we shouldn't have to write it ourselves.

poodlewars (commented):
What I'm worried about is the below, am I misunderstanding anything Phoebus?

**Session 1**

* Create AC instance
* Connect to mongodb instance db, specify SelectionTimeoutMs option
* Create library lib

*Session 2*

* Create AC instance
* Connect to db but do not specify any options
* Access library lib
* You will connect using the options from Session 1 since the config from that is persisted in the cfg library in the protobuf. I don't think this is the correct behaviour

I think the StorageOverride concept that William introduced would help us resolve the situation so that the connection options are ephemeral in the way that I would expect.

python/arcticdb/adapters/mongo_library_adapter.py
qc00: All that is being parsed are run-time config, so there's no need to parse the URI at the moment.

python/tests/conftest.py
```diff
@@ -146,14 +147,16 @@ def moto_s3_uri_incl_bucket(moto_s3_endpoint_and_credentials):
...
+@pytest.fixture(scope="function", params=["S3", "LMDB", "MONGO", "Azure"] if AZURE_SUPPORT else ["S3", "LMDB", "MONGO"])
 def arctic_client(request, moto_s3_uri_incl_bucket, tmpdir, encoding_version):
     if request.param == "S3":
         ac = Arctic(moto_s3_uri_incl_bucket, encoding_version)
     elif request.param == "Azure":
         ac = Arctic(request.getfixturevalue("azurite_azure_uri_incl_bucket"), encoding_version)
     elif request.param == "LMDB":
         ac = Arctic(f"lmdb://{tmpdir}", encoding_version)
+    elif request.param == "MONGO":
+        ac = Arctic(request.getfixturevalue("mongo_test_uri"), encoding_version)
```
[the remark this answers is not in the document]
qc00: I've already refactored this out in my PR

python/tests/conftest.py
```diff
@@ -686,7 +700,9 @@ def spawn_azurite(azurite_port):
 @pytest.fixture(
     scope="function",
     params=(
-        ["moto_s3_uri_incl_bucket", "azurite_azure_uri_incl_bucket"] if AZURE_SUPPORT else ["moto_s3_uri_incl_bucket"]
+        ["moto_s3_uri_incl_bucket", "azurite_azure_uri_incl_bucket", "mongo_test_uri"]
+        if AZURE_SUPPORT
+        else ["moto_s3_uri_incl_bucket", "mongo_test_uri"]
```
qc00, line 705: We don't have a simple way to start a Mongo server in all environments, so I wouldn't add it to test parameters.

It's tested internally, and we are not fully supporting it for external use at the moment.

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -178,18 +178,23 @@ def get_path_prefix_option(uri):
         return "&path_prefix"
 
 
-def test_separation_between_libraries_with_prefixes(object_storage_uri_incl_bucket):
+@pytest.mark.parametrize(
+    "lib_type",
+    ["moto_s3_uri_incl_bucket", "azurite_azure_uri_incl_bucket"] if AZURE_SUPPORT else ["moto_s3_uri_incl_bucket"],
+)
+def test_separation_between_libraries_with_prefixes(request, lib_type):
```
qc00, line 185: Please revert changes to this test as I have rewritten these test setup in my branch (and it's not relevant to mongo).

python/tests/conftest.py
```diff
@@ -397,7 +400,18 @@ def mongo_store_factory(request, lib_name):
...
         uri = f"mongodb://{mongo_server_sess.hostname}:{mongo_server_sess.port}"
 
-    cfg_maker = functools.partial(create_test_mongo_cfg, uri=uri)
+    yield uri
+
+    # mongodb does not support prefix to differenciate different tests and the server is session-scoped
+    mongo_client = pymongo.MongoClient(uri)
+    for db_name in mongo_client.list_database_names():
+        if db_name != "local" and db_name != "admin" and db_name != "apiomat" and db_name != "config":
+            mongo_client.drop_database(db_name)
```
qc00, line 409: This is not reliable if the same Mongo server is used to support concurrent tests.

Just use the `clear()` method.

qc00 (changes requested):

python/tests/conftest.py
poodlewars: We need to add it to the tests. The whole point of this change is to start supporting it for external use.

python/arcticdb/adapters/mongo_library_adapter.py
poodlewars: It's still good to validate it early. And I think there will be a need to parse it actually, to separate out the connection details from the options.

qc00: Not if we have to import an entire Python dependency to do it.

If we have to have this early check, it should call into C++ and use the C++ SDK we've already compiled in.

python/arcticdb/adapters/azure_library_adapter.py
```diff
@@ -84,9 +83,7 @@ def config_library(self):
         return lib._library
 
     def _parse_query(self, query: str) -> ParsedQuery:
-        if query and query.startswith("?"):
```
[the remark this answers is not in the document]
phoebusm: It was added for parsing the never-published, S3-like azure uri. The line should be deleted when we have moved on to Azure connection string

python/tests/conftest.py
phoebusm: Thanks for heads up. I do miss your points. However, I am not sure whether this is a clean solution either. 
The clear up is for handling checking `assert ac.list_libraries() == []`. I believe it is for ensuring the test environment is clear. Yet I don't believe the checking is still valid if we run tests parallelly.
I think checking the particular library used for testing should be enough.

phoebusm:
> What I'm worried about is the below, am I misunderstanding anything Phoebus?
> 
> **Session 1**
> 
> * Create AC instance
> * Connect to mongodb instance db, specify SelectionTimeoutMs option
> * Create library lib
> 
> _Session 2_
> 
> * Create AC instance
> * Connect to db but do not specify any options
> * Access library lib
> * You will connect using the options from Session 1 since the config from that is persisted in the cfg library in the protobuf. I don't think this is the correct behaviour
> 
> I think the StorageOverride concept that William introduced would help us resolve the situation so that the connection options are ephemeral in the way that I would expect.
> 
> https://github.com/search?q=repo%3Aman-group%2FArcticDB%20storageoverride&type=code

I agree that `min_pool_size`, `max_pool_size`, `selection_timeout_ms` are runtime config. They should not be in the proto nor persisted.
However, introducing `StorageOverride` will introduce an extra layer of configuration. That will bring extra complexity but achieve so little. As there is no user currently using mongo uri, a more simple design can be used instead, as long as it is well documented.
Suggested design: Ones in `ConfigsMap` are used by default. However, if those settings are available in the connecting string, they will be used instead.

This will be a non-breaking change. It requires only **new** user who starts using mongo uri to read the manual to make sure the setting in `ConfigsMap` will affect them. Old user should not be bothered.

python/arcticdb/version_store/helper.py
phoebusm: Not required

python/arcticdb/adapters/mongo_library_adapter.py
phoebusm: I just strip the credential in the function `__repr__`

python/tests/conftest.py
phoebusm: Using `clear()` needs to keep track of the libraries in use. It will be a bit tricky to do so for uri related test

python/tests/integration/arcticdb/test_arctic.py
phoebusm: It will be handled in the final rebase

python/arcticdb/adapters/mongo_library_adapter.py
phoebusm: Use pymongo for better connecting string syntax verification

python/arcticdb/adapters/azure_library_adapter.py
```diff
@@ -84,9 +83,7 @@ def config_library(self):
...
-            query = query.strip("?")
-        elif not query:
+        if not query:
             raise ValueError(f"Invalid Azure URI. Missing query parameter")
 
         parsed_query = re.split("[;]", query)
```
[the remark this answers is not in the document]
phoebusm: https://github.com/man-group/ArcticDB/issues/790

poodlewars (approved):

python/arcticdb/adapters/mongo_library_adapter.py
```diff
@@ -0,0 +1,63 @@
...
+class MongoLibraryAdapter(ArcticLibraryAdapter):
+    @staticmethod
+    def supports_uri(uri: str) -> bool:
+        return uri.startswith("mongodb://")
+
+    def __init__(self, uri: str, *args, **kwargs):
+        try:
+            parameters = pymongo.uri_parser.parse_uri(uri)
+            self._endpoint = f"{parameters['nodelist'][0][0]}:{parameters['nodelist'][0][1]}"
+        except Exception as e:
```
poodlewars, line 33: At first I thought `Exception` was too broad, but `pymongo` is rubbish at this so not much we can do:

```
>>> uri = "blah not a uri"
>>> pymongo.uri_parser.parse_uri(uri)
  File "/var/aseaton/tst1/lib/python3.8/site-packages/pymongo/uri_parser.py", line 433, in parse_uri
    raise InvalidURI("Invalid URI scheme: URI must "
pymongo.errors.InvalidURI: Invalid URI scheme: URI must begin with 'mongodb://' or 'mongodb+srv://'
>>> pymongo.uri_parser.parse_uri("mongodb://notgreat")
{'nodelist': [('notgreat', 27017)], 'username': None, 'password': None, 'database': None, 'collection': None, 'options': {}, 'fqdn': None}
>>> pymongo.uri_parser.parse_uri("mongodb://b:a:d:u:r:i")
  File "/var/aseaton/tst1/lib/python3.8/site-packages/pymongo/uri_parser.py", line 514, in parse_uri
    nodes = split_hosts(hosts, default_port=default_port)
  File "/var/aseaton/tst1/lib/python3.8/site-packages/pymongo/uri_parser.py", line 364, in split_hosts
    nodes.append(parse_host(entity, port))
  File "/var/aseaton/tst1/lib/python3.8/site-packages/pymongo/uri_parser.py", line 111, in parse_host
    raise ValueError("Reserved characters such as ':' must be "
ValueError: Reserved characters such as ':' must be escaped according RFC 2396. An IPv6 address literal must be enclosed in '[' and ']' according to RFC 2732.
```

python/tests/integration/arcticdb/test_arctic.py
phoebusm: Reverted

build_tooling/parallel_test.sh
```diff
@@ -13,12 +13,9 @@ catch=`{ which catchsegv 2>/dev/null || echo ; } | tail -n 1`
 function worker() {
     group=${1:?Must pass the group id argument}
     shift
-    duration_file="$TEST_OUTPUT_DIR/pytest-durations.$group.json"
     new_root=$PARALLEL_TEST_ROOT/$group
     set -o xtrace -o pipefail
 
-    cp "$tooling_dir/pytest-durations.json" "$duration_file"
```
poodlewars, line 20: Why don't we need this any more? Should remove the checked in `pytest-durations.json` if it is not required

python/arcticdb/util/test.py
```diff
@@ -31,6 +32,11 @@
 from arcticdb.config import _DEFAULT_ENVS_PATH
 from arcticdb_ext import set_config_int, get_config_int, unset_config_int
 
+# leave out Mongo as spinning up a Mongo instance in Windows CI is fiddly, and Mongo support is only
+# currently required for Linux for internal use.
+# We also skip it on Mac as github actions containers don't work with macos
+RUN_MONGO_TEST = True if sys.platform == "linux" else False
```
poodlewars, line 38:
```suggestion
RUN_MONGO_TEST = sys.platform == "linux"
```

poodlewars (approved):

build_tooling/parallel_test.sh
phoebusm: After having another thought of it, this should be kept. Will revert.

PR #763, opened 2023-08-17
Decouple storage headers from implementations & storage.hpp

qc00:
Since VariantStorage is turned into pointer-based polymorphism (#695), there's no need to include the *_storage[-inl].hpp headers everywhere.

No change in logic


cpp/arcticdb/storage/memory/memory_storage.cpp
qc00, line 1: This file is `memory_storage-inl.hpp` + `MemoryStorage::MemoryStorage()`.

I gave up doing it this way after this one because the git does not trace this kind of change well

mehertz (approved):
Approved based on:

> No change in logic

PR #766, opened 2023-08-17
Partially revert #763 for consistency

qc00:
`memory_storage.*` now use the same pattern as other storages

[No logic change](https://github.com/man-group/ArcticDB/compare/9699ffb2c79b...decouple_storage_headers#diff-b676d611e4a24bd9522a502fe97f2fba9a29da927b54dfa62fb1435a4810db5d)


poodlewars (approved):
I've checked that:
- The logic in the same before and after
- The `inl` file has the same methods in as other storages

PR #775, opened 2023-08-21
Enhancement 702: Add option to create library if it does not exist when calling get_library

### LINKED ISSUE #702: Should have a way to create and get libraries without so much boiler plate.

Our current API means you need write boiler plate code to create and get libraries every time you write a script.

```python

if 'example' not in arctic.list_libraries():
    arctic.create_library('example')
lib = arctic.get_library('example')
```

Notes:
- the if is a guard against trying to create a library when it already exists (which throws)
- there is no `has_library` method, so you need to look in the list of libraries.
- you need to separately fetch the library, as `create_library` doesn't return anything.

- A `has_library` method so you can check for existence.
- `get_library` should accept a `create=True` argument, and will seamlessly create a library if needed.
- `get_library` should potentially also accept LibraryOptions

What should happen if the library exists but the LibraryOptions don't match?

- `'example' in arctic` could check for existance in a pythonic way.
- `create_library` could return the library so you don't have to seperatly fetch it
- `create_library` could not throw exceptions if the library exists, and instead return the existing library
- create a seperate method `ensure_library` which does the desired behviour (one-line creation and fetching)

alexowens90:
#### Reference Issues/PRs

Closes #702 

- Makes the `AzureLibraryAdapter` behave in the same way as the others with regards to default encoding
- Allow the user to override the encoding specified on the `Arctic` instance being used to create a library by specifying the encoding in the `LibraryOptions` (arguably previously a bug, since `LibraryOptions.encoding_version` was unused)
- Documented the `encoding_version` argument to both the `Arctic` and `LibraryOptions` `__init__` methods
- Added the `encoding_version` to the `LibraryOptions` `__repr__`
- Add an `options()` method to the `Library` class to retrieve the `LibraryOptions` object this library was created with
- Add a `has_library` method to the `Arctic` class
- Add optional arguments to `Arctic.create_library` to allow the library to be created if it does not exist

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/arctic.py
+++ b/python/arcticdb/arctic.py
@@ -7,10 +7,10 @@ As of the Change Date specified in that file, in accordance with the Business So
 """
 from typing import List, Optional
 
-from arcticdb.options import LibraryOptions
+from arcticdb.options import DEFAULT_ENCODING_VERSION, LibraryOptions
 from arcticdb_ext.storage import LibraryManager, StorageOverride
-from arcticdb.exceptions import LibraryNotFound
-from arcticdb.version_store.library import Library
+from arcticdb.exceptions import LibraryNotFound, MismatchingLibraryOptions
+from arcticdb.version_store.library import ArcticInvalidApiUsageException, Library
 from arcticdb.version_store._store import NativeVersionStore
 from arcticdb.adapters.s3_library_adapter import S3LibraryAdapter
 from arcticdb.adapters.lmdb_library_adapter import LMDBLibraryAdapter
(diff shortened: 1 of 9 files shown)


jamesmunro (approved):

python/arcticdb/arctic.py
```diff
@@ -188,9 +209,21 @@ def get_library(self, library: str) -> Library:
         -------
         Library
         """
-        return self[library]
-
-    def create_library(self, name: str, library_options: Optional[LibraryOptions] = None) -> None:
+        try:
```
poodlewars, line 212: Blow up if `create_if_missing` is Falsey but `library_options` are provided?

alexowens90: Done

PR #781, opened 2023-08-22
Remove map from key lookup

willdealtry:
Draft PR removing the semi::static_map from the key data lookup, as it unexpectedly has runtime behaviour that is unsafe and undesirable


mehertz:
Can you keep the PR template please? Not much point having if people remove it even for code changes.

PR #782, opened 2023-08-22
Bugfix/nativeversionstore write metadata batch should never return dataerror objects

alexowens90:

The `NativeVersionStore` API (used internally at Man Group) would previously throw an exception if there was a key missing from the storage required to complete a `batch_write_metadata` operation. This was accidentally modified in PR #476, such that `DataError` objects were returned instead. This PR restores the previous behviour, for the `NativeVersionStore` API, while maintaining the new behaviour for the `Library` API.
In addition:

- `throw_on_missing_version` was renamed everywhere to `throw_on_error`, as missing versions are not the only case in which exceptions are raised
- tests were added to cover the behaviour of `batch_read` from the `NativeVersionStore` API when there are missing keys

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -948,9 +948,9 @@ class NativeVersionStore:
             Dictionary of symbol mapping with the versioned items
         """
         _check_batch_kwargs(NativeVersionStore.batch_read, NativeVersionStore.read, kwargs)
-        throw_on_missing_version = True
+        throw_on_error = True
         versioned_items = self._batch_read_to_versioned_items(
-            symbols, as_ofs, date_ranges, columns, query_builder, throw_on_missing_version, kwargs
+            symbols, as_ofs, date_ranges, columns, query_builder, throw_on_error, kwargs
         )
         check(
             all(v is not None for v in versioned_items),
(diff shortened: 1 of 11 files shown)


GemDot (approved):

python/arcticdb/version_store/_store.py
```diff
@@ -959,14 +959,14 @@ def batch_read(
...
-        self, symbols, as_ofs, date_ranges, columns, query_builder, throw_on_missing_version, kwargs=None
+        self, symbols, as_ofs, date_ranges, columns, query_builder, throw_on_error, kwargs=None
     ):
         if kwargs is None:
             kwargs = dict()
         version_queries = self._get_version_queries(len(symbols), as_ofs, **kwargs)
         read_queries = self._get_read_queries(len(symbols), date_ranges, columns, query_builder)
         read_options = self._get_read_options(**kwargs)
-        read_options.set_batch_throw_on_missing_version(throw_on_missing_version)
+        read_options.set_batch_throw_on_error(throw_on_error)
```
willdealtry, line 969: Is there a reason why it didn't go in _get_read_options?

alexowens90: I hadn't thought of it to be honest. It isn't a `kwarg` though, so we'd have to modify them in the calling functions, which feels like a bit of an odd pattern?

PR #793, opened 2023-08-24
Small API change to VectorDB

octogenary:

Small API change to return a list of vectors with identifiers and distances instead of a dataframe with a last row containing the distance.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/vector_db/vector_db.py
+++ b/python/arcticdb/vector_db/vector_db.py
@@ -3,12 +3,13 @@ import pandas as pd
 import numpy as np
 
 from collections import namedtuple
-from typing import Union, Optional, Collection, List, Mapping
+from typing import Union, Optional, Collection, List, Mapping, Dict
 
 from arcticdb import Arctic
 from arcticdb.options import LibraryOptions
 from arcticdb.version_store.library import Library, ArcticUnsupportedDataTypeException
 from arcticdb.version_store.processing import QueryBuilder
+from arcticdb.exceptions import ArcticNativeNotYetImplemented
 from arcticdb_ext.version_store import NoSuchVersionException
 from arcticdb_ext.storage import NoDataFoundException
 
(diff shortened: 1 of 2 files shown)


octogenary:
Hang on ignore this, not sure why the diff has so many files in it.

octogenary:
> Hang on ignore this, not sure why the diff has so many files in it.

Fixed it now.

python/arcticdb/vector_db/vector_db.py
```diff
@@ -434,6 +436,42 @@ def top_k(
...
+        k
+            The number of vectors to return from the search.
+        query_vector
+            The vector from which distances are taken.
+        norm
+            The distance norm based on which the top k will be computed. This is presently ignored; the default
+            is Euclidean.
+        index
+            The index based on which to compute the top k. This is also ignored.
+        include_vectors
```
alexowens90, line 465: Does pinecone have an `include_distance` parameter as well? Or is that always included?

python/arcticdb/vector_db/vector_db.py
```diff
@@ -255,6 +256,19 @@ def _generate_dataframe(
     return df_to_upsert
 
 
+def _process_raw_top_k_result(
+    raw_result: pd.DataFrame,
+    include_vectors: bool,
+) -> List[Dict[str, Union[str, pd.Series]]]:
```
alexowens90, line 262: I think a list of floats is a more natural return type, as the user doesn't need to know anything about Pandas to use the rest of the API.

alexowens90 (changes requested):

python/arcticdb/vector_db/vector_db.py
octogenary: No (https://docs.pinecone.io/reference/query)

python/arcticdb/vector_db/vector_db.py
octogenary: Hm all right, that makes the test a bit easier.

PR #795, opened 2023-08-25
Add dep on packaging

poodlewars:
Required for:

```
from arcticdb.util._versions import IS_PANDAS_TWO
```

but seems to have been missed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,6 +47,7 @@ install_requires =
     pyyaml
     decorator
     prometheus_client
+    packaging
 
 [flake8]
 # max line length for black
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -32,7 +32,7 @@ install_requires =
     # is also distributed for other languages like C++). Python packages' names on conda-forge
     # might be suffixed or preffixed by "python", and one better double check using
     # conda-forge's package and feedstock browser.
-    #
+    #kkkkk
```
jjerphan, line 35:
```suggestion
    #
```

jjerphan (commented):
Good catch!

`packaging` was already used before https://github.com/man-group/ArcticDB/pull/540. Until now, it might have been pulled by other dependencies, but it indeed need to be specified.

Could you also update `environment_unix.yml` to add it as a dependency?

jjerphan (approved):
Thanks, @poodlewars.

PR #796, opened 2023-08-25
Symbol list refactor

willdealtry:
Fixes issue #773

Design doc: https://github.com/man-group/arcticdb-bloomberg/blob/main/RFCs/00005-symbol-list-refactor.md

Depending on timestamps being accurate in the symbol list has proved to be troublesome. Instead, we should use the most recent version id known to a client as an indication of the client's view of the world at the time as symbol list entry is written. That way, we can identify and correct symbol list entries that refer to conflicting writes. 

The timestamp is still stored, but is now only used as a coarse heuristic to determine whether writes to the same symbol occurred close enough together to constitute a possible overlapping write. 

When a problematic symbol list entry is encountered, it is added to a list, and at the end of the scan a batch operation is performed to determine whether the problematic symbol/version pairs exist or not. 

Regarding backwards compatibility, older clients will be able to read the new symbol keys, both journalled and compacted, however they will not make any use of the information in compaction. If a newer client detects that there are keys produced by an older client for a particular symbol, that entry will not be flagged as problematic on this run, since the information to determine whether or not it is genuinely problematic is not present.

### DIFF
--- a/cpp/arcticdb/stream/merge_utils.hpp
+++ b/cpp/arcticdb/stream/merge_utils.hpp
@@ -1,0 +1,106 @@
+#pragma once
+
+#include <arcticdb/column_store/chunked_buffer.hpp>
+#include <arcticdb/column_store/string_pool.hpp>
+#include <arcticdb/column_store/memory_segment.hpp>
+
+namespace arcticdb {
+inline void merge_string_column(
+    ChunkedBuffer& src_buffer,
+    const std::shared_ptr<StringPool>& src_pool,
+    const std::shared_ptr<StringPool>& merged_pool,
+    CursoredBuffer<ChunkedBuffer>& output,
+    bool verify
+) {
+    using OffsetType = StringPool::offset_t;
+    constexpr auto offset_size =  sizeof(OffsetType);
+    auto num_strings = src_buffer.bytes() / offset_size;
+    for(auto row = 0ULL; row < num_strings; ++row) {
+        auto offset = get_offset_string_at(row, src_buffer);
+        StringPool::offset_t new_value;
+        if (offset != not_a_string() && offset != nan_placeholder()) {
+            auto sv = get_string_from_pool(offset, *src_pool);
+            new_value = merged_pool->get(sv).offset();
+        } else {
+            new_value = offset;
+        }
+        output.ensure<OffsetType>(1);
+        util::check(new_value >= 0, "Unexpected negative number {} in merge string column", new_value);
+        output.typed_cursor<OffsetType>() = new_value;
+        output.commit();
+    }
+    if (verify) {
+        const auto& out_buffer = output.buffer();
+        auto num_out = out_buffer.bytes() /offset_size;
+        util::check(num_strings == num_out, "Mismatch in input/output size {} != {}", num_strings, num_out);
+        for(auto row = size_t(0); row < num_out; ++row) {
+            auto offset = get_offset_string_at(row, out_buffer);
+            if (offset != not_a_string() && offset != nan_placeholder()) {
+                auto sv ARCTICDB_UNUSED = get_string_from_pool(offset, *merged_pool);
+                ARCTICDB_DEBUG(log::version(), "Accessed {} from string pool", sv);
+            }
+        }
+    }
+}
+
+inline void merge_string_columns(const SegmentInMemory& segment, const std::shared_ptr<StringPool>& merged_pool, bool verify) {
+    for (size_t c = 0; c < segment.descriptor().field_count(); ++c) {
+        auto &frame_field = segment.field(c);
+        const auto& field_type = frame_field.type();
+
+        if (!is_sequence_type(field_type.data_type_))
+            continue;
+
+        auto &src = segment.column(static_cast<position_t>(c)).data().buffer();
+        CursoredBuffer<ChunkedBuffer> cursor{src.bytes(), false};
+        merge_string_column(src, segment.string_pool_ptr(), merged_pool, cursor, verify);
+        std::swap(src, cursor.buffer());
+    }
+}
+
+inline void merge_segments(
+    std::vector<SegmentInMemory>& segments,
+    SegmentInMemory& merged,
+    bool is_sparse) {
+    ARCTICDB_DEBUG(log::version(), "Appending {} segments", segments.size());
+    timestamp min_idx = std::numeric_limits<timestamp>::max();
+    timestamp max_idx = std::numeric_limits<timestamp>::min();
+    for (auto &segment : segments) {
+        std::vector<SegmentInMemory> history{{segment}};
+        const auto& latest = *history.rbegin();
+        ARCTICDB_DEBUG(log::version(), "Appending segment with {} rows", latest.row_count());
+        for(const auto& field : latest.descriptor().fields()) {
+            if(!merged.column_index(field.name())){//TODO: Bottleneck for wide segments
+                auto pos = merged.add_column(field, 0, false);
+                if (!is_sparse){
+                    merged.column(pos).mark_absent_rows(merged.row_count());
+                }
+            }
+        }
+        if (latest.row_count() && latest.descriptor().index().type() == IndexDescriptor::TIMESTAMP) {
+            min_idx = std::min(min_idx, latest.begin()->begin()->value<timestamp>());
+            max_idx = std::max(max_idx, (latest.end() - 1)->begin()->value<timestamp>());
+        }
+        merge_string_columns(latest, merged.string_pool_ptr(), false);
+        merged.append(*history.rbegin());
+        merged.set_compacted(true);
+        util::print_total_mem_usage(__FILE__, __LINE__, __FUNCTION__);
+    }
+}
+
+inline pipelines::FrameSlice merge_slices(
+    std::vector<pipelines::FrameSlice>& slices,
+    const StreamDescriptor& desc) {
+    util::check(!slices.empty(), "Expected to merge non-empty slices_vector");
+
+    pipelines::FrameSlice output{slices[0]};
+    for(const auto& slice : slices) {
+        output.row_range.first = std::min(output.row_range.first, slice.row_range.first);
+        output.row_range.second = std::max(output.row_range.second, slice.row_range.second);
+    }
+
+    output.col_range.first = desc.index().field_count();
+    output.col_range.second = desc.field_count();
+    return output;
+}
+}
--- a/cpp/arcticdb/toolbox/library_tool.cpp
+++ b/cpp/arcticdb/toolbox/library_tool.cpp
@@ -12,14 +12,23 @@
 #include <arcticdb/entity/atom_key.hpp>
 #include <arcticdb/entity/protobufs.hpp>
 #include <arcticdb/entity/types.hpp>
+#include <arcticdb/pipeline/pipeline_utils.hpp>
 #include <arcticdb/storage/library.hpp>
 #include <arcticdb/util/clock.hpp>
 #include <arcticdb/util/key_utils.hpp>
 #include <arcticdb/util/variant.hpp>
+#include <arcticdb/version/version_utils.hpp>
 
 namespace arcticdb::toolbox::apy {
 
 using namespace arcticdb::entity;
+
+ReadResult LibraryTool::read(const VariantKey& key) {
+    auto segment = read_to_segment(key);
+    auto segment_in_memory = decode_segment(std::move(segment));
+    auto frame_and_descriptor = frame_and_descriptor_from_segment(std::move(segment_in_memory));
+    return pipelines::make_read_result_from_frame(frame_and_descriptor, to_atom(key));
+}
 
 Segment LibraryTool::read_to_segment(const VariantKey& key) {
     auto kv = std::visit([lib=lib_](const auto &k) { return lib->read(k); }, key);
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -175,26 +175,25 @@ std::set<StreamId> LocalVersionedEngine::list_streams_internal(
     std::optional<SnapshotId> snap_name,
     const std::optional<std::string>& regex,
     const std::optional<std::string>& prefix,
-    const std::optional<bool>& opt_use_symbol_list,
-    const std::optional<bool>& opt_all_symbols
+    const std::optional<bool>& use_symbol_list,
+    const std::optional<bool>& all_symbols
     ) {
     ARCTICDB_SAMPLE(ListStreamsInternal, 0)
     auto res = std::set<StreamId>();
-    bool use_symbol_list = opt_use_symbol_list.value_or(cfg().symbol_list());
 
     if (snap_name) {
-        res = list_streams_in_snapshot(store(), snap_name.value());
+        res = list_streams_in_snapshot(store(), *snap_name);
     } else {
-        if(use_symbol_list)
+        if(use_symbol_list.value_or(cfg().symbol_list()))
             res = symbol_list().get_symbol_set(store());
         else
-            res = list_streams(store(), version_map(), prefix, opt_false(opt_all_symbols));
+            res = list_streams(store(), version_map(), prefix, all_symbols.value_or(false));
     }
 
     if (regex) {
         return filter_by_regex(res, regex);
     } else if (prefix) {
-        return filter_by_regex(res, std::optional("^" + prefix.value()));
+        return filter_by_regex(res, std::optional("^" + *prefix));
     }
     return res;
 }
@@ -212,7 +211,7 @@ std::optional<VersionedItem> LocalVersionedEngine::get_latest_version(
         ARCTICDB_DEBUG(log::version(), "get_latest_version didn't find version for stream_id: {}", stream_id);
         return std::nullopt;
     }
-    return VersionedItem{std::move(key.value())};
+    return VersionedItem{std::move(*key)};
 }
 
 std::optional<VersionedItem> LocalVersionedEngine::get_specific_version(
@@ -278,7 +277,7 @@ std::optional<VersionedItem> LocalVersionedEngine::get_version_at_time(
         return std::nullopt;
     }
 
-    return VersionedItem(std::move(index_key.value()));
+    return VersionedItem(std::move(*index_key));
 }
 
 std::optional<VersionedItem> LocalVersionedEngine::get_version_from_snapshot(
@@ -291,7 +290,7 @@ std::optional<VersionedItem> LocalVersionedEngine::get_version_from_snapshot(
     }
     // A snapshot will normally be in a ref key, but for old libraries it still needs to fall back to iteration of
     // atom keys.
-    auto segment = opt_snapshot.value().second;
+    auto segment = opt_snapshot->second;
     for (size_t idx = 0; idx < segment.row_count(); idx++) {
         auto stream_index = read_key_row(segment, static_cast<ssize_t>(idx));
         if (stream_index.id() == stream_id) {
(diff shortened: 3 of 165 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -976,7 +979,7 @@ void LocalVersionedEngine::push_incompletes_to_symbol_list(const std::set<Stream
     auto existing = list_streams_internal(std::nullopt, std::nullopt, std::nullopt, cfg().symbol_list(), false);
     for(const auto& incomplete : incompletes) {
         if(existing.find(incomplete) == existing.end())
-            symbol_list().add_symbol(store_, incomplete);
+            symbol_list().add_symbol(store_, incomplete, VersionId(0));
```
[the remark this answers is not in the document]
willdealtry: Yeah why not. It's a tick-data thing, probably doesn't belong here

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -935,17 +932,28 @@ void PythonVersionStore::delete_all_versions(const StreamId& stream_id) {
...
+                       all_index_keys.size());
+        if (!cfg().write_options().delayed_deletes()) {
+            delete_tree({all_index_keys.begin(), all_index_keys.end()});
+        } else {
+            ARCTICDB_DEBUG(log::version(), "Not deleting data for {}", stream_id);
+        }
+
+        ARCTICDB_DEBUG(log::version(), "Delete of Symbol {} successful", stream_id);
+    } catch(const StorageException& ex) {
+        log::version().error("Got storage exception in delete - possible parallel deletion?: {}", ex.what());
```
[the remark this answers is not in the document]
willdealtry: There are no actual ill-effects from doing this, the deleted data will still be gone and the rest of the symbol will be readable, I just wanted to advise the user if possible as they are probably doing something inadvisable (for example as you point out if they combine writes with writes or writes with deletes they will get last-writer-wins).

I'm thinking about how we can have more robust concurrenct guarantees over single symbols generally, but there's no shortcuts, it will have to be done in quite a different way.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,703 @@
...
+    }
+    return map;
+}
+
+auto tail_range(const std::vector<SymbolEntryData>& updated) {
+    auto it = updated.crbegin();
+    const auto reference_id = it->reference_id_;
+    ++it;
+
+    while(it != updated.crbegin() && it->reference_id_ == reference_id)
```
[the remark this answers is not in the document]
willdealtry: The list is sorted by reference id, so we're sorting backwards from the most recent looking for things that have the same reference id (i.e. conflicts). We don't care about conflicts that aren't the most recent as we assume that they will have been superseded by what happens after - for example if I race a write and delete on version 2, then have a non-racy write on version 3 then there is no problem because the symbol with both exist and be in the list.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,703 @@
...
+    auto data_type = std::holds_alternative<StringId>(type_holder) ? DataType::ASCII_DYNAMIC64 : DataType::UINT64;
+    return stream_descriptor(stream_id, RowCountIndex(), {
+        scalar_field(data_type, "deleted_symbol"),
+        scalar_field(DataType::UINT64, "deleted_version_id"),
+        scalar_field(DataType::NANOSECONDS_UTC64, "deleted_timestamp")
+    });
+}
+
+bool SymbolList::needs_compaction(const LoadResult& load_result) const {
+    return load_result.symbol_list_keys_.size() > data_.max_delta_ || !load_result.maybe_last_compaction;
```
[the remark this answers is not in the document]
willdealtry: When there is no pre-existing compaction, regardless of how many keys there are

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,703 @@
...
+}
+
+bool timestamps_too_close(
+        const std::vector<SymbolEntryData>::const_reverse_iterator& first,
+        const std::vector<SymbolEntryData>& updated,
+        timestamp min_allowed_interval,
+        bool all_same_action) {
+    bool same_as_updates = all_same_action && updated.rbegin()->action_ == first->action_;
+
+    return updated.rbegin()->timestamp_ - first->timestamp_ < min_allowed_interval && !same_as_updates;
```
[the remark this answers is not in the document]
willdealtry: The keys are sorted first by reference id, and then by timestamp, so if there is a negative value it means that the timestamsp and the reference ids are out of order with each other, which is definitely problematic

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -8,163 +8,201 @@
...
-    void write_symbol(const std::shared_ptr<Store>& store, const StreamId& symbol) {
-        write_journal(store, symbol, AddSymbol);
+    [[nodiscard]] ActionType action() const {
+        return problem_->action_;
     }
 
-    void delete_symbol(const std::shared_ptr<Store>& store, const StreamId& symbol) {
-        write_journal(store, symbol, DeleteSymbol);
+    explicit operator bool() const {
+        return static_cast<bool>(problem_);
```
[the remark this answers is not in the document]
willdealtry: Unknown reference ids means an old client, which means that we assume that nothing is problematic (not that it necessarily isn't problematic, but there's no way to tell, and we can't check everything)

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,703 @@
...
+        return true;
+
+    const auto& latest = *updated.rbegin();
+    if(existing.action_ == latest.action_)
+        return false;
+
+    if(existing.reference_id_ == latest.reference_id_)
+        return latest;
+
+    if(latest.timestamp_ - existing.timestamp_ < min_allowed_interval)
```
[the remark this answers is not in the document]
willdealtry: Same as above, the updates are all expected to all be strictly after the previously compacted stuff, if they're not then we definitely have a conflict

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,703 @@
...
+    return symbols;
+}
+
+CollectionType load_from_symbol_list_keys(
+        const std::shared_ptr<VersionMap>& version_map,
+        const std::shared_ptr<Store>& store,
+        const std::vector<AtomKey>& keys,
+        const MaybeCompaction& maybe_compaction) {
+    ARCTICDB_RUNTIME_DEBUG(log::symbol(),"Loading symbols from symbol list keys");
+    util::check(maybe_compaction.has_value(), "Expected prior compaction when loading from symbol list");
```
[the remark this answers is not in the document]
willdealtry: Not sure. It's a very lightweight check, it documents the expectation of the function and it's robust to any future refactoring, since if someone were to call this from elsewhere and not check they'd get a std::bad_optional_access. Kind of on the fence here, will remove if it really bothers you

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,703 @@
...
+                                                 data_.type_holder_).get();
+                    delete_keys(store, load_result.detach_symbol_list_keys(), std::get<AtomKey>(written)).get();
+                }
+            } else {
+                ARCTICDB_RUNTIME_DEBUG(log::symbol(),"Not compacting the symbol list due to lock contention");
+            }
+        } catch (const storage::PermissionException& ex) {
+            // Note: this only reflects AN's permission check and is not thrown by the Storage
+            ARCTICDB_RUNTIME_DEBUG(log::symbol(),"Not compacting the symbol list due to lack of permission", ex.what());
+        } catch (const std::exception& ex) {
```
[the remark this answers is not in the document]
willdealtry: I think @qc00 has a PR to address the storage exceptions, after which this catch can be narrowed

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -8,163 +8,201 @@
...
+        SymbolEntryData(reference_id, reference_time, action),
+        stream_id_(std::move(stream_id)) {
     }
+};
+
+struct ProblematicResult {
+    std::optional<SymbolEntryData> problem_;
+    bool contains_unknown_reference_ids_ = false;
+
+    ProblematicResult(const SymbolEntryData& data) :
```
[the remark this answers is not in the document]
willdealtry: It was annoying for some reason, will try to fix

python/spark_storage_lock.py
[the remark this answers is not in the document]
willdealtry: Yep, both this and the spark version of the multi-process madness tests in test_symbol_list.py will be added as a different PR

cpp/arcticdb/version/symbol_list.cpp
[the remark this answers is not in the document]
poodlewars: It won't keep me up at night, I'd just tend to use the type system for this kind of check

cpp/arcticdb/version/symbol_list.cpp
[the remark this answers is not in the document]
poodlewars: Gotcha, think I might grok a name like `last_compaction_if_one_exists` better. Well in to personal preference territory though

cpp/arcticdb/version/symbol_list.cpp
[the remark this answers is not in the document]
willdealtry: Sorry, yes I had misunderstood what you were suggesting. I think that's a good idea

cpp/arcticdb/toolbox/library_tool.cpp
```diff
@@ -12,15 +12,24 @@
...
 #include <arcticdb/util/variant.hpp>
+#include <arcticdb/version/version_utils.hpp>
 
 namespace arcticdb::toolbox::apy {
 
 using namespace arcticdb::entity;
 
+ReadResult LibraryTool::read(const VariantKey& key) {
+    auto segment = read_to_segment(key);
+    auto segment_in_memory = decode_segment(std::move(segment));
```
alexowens90, line 28: @vasil-pashov looks like we will need the decode methods

cpp/arcticdb/version/test/test_symbol_list.cpp
alexowens90, line 1: In the `IsProblematic` tests, all of timestamps always seem to be increasing through the `SymbolEntryData` vector. Would be good to also cover the clock-skew case

alexowens90 (changes requested):

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,738 @@
...
+
+    const auto data_type = get_symbol_data_type(seg);
+
+    // Because we need to be backwards compatible with the old style symbol list, the additions and deletions are
+    // in separate columns. The first three columns are the symbol, version and timestamp for the additions, and the
+    // next three are the same for the deletions. Old-style symbol lists will ignore everything but the first column
+    // which will mean that they can't do any conflict resolution but will get the correct data.
+    for(auto i = 0L; i < seg.column(0).row_count(); ++i) {
+        util::check(seg.column(0).row_count() == seg.column(1).row_count() && seg.column(0).row_count() == seg.column(2).row_count(),
+                    "Column mismatch in symbol segment additions: {} {} {}", seg.column(0).row_count(), seg.column(1).row_count(), seg.column(2).row_count());
```
alexowens90, line 179: This check should be outside the for loop

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,738 @@
...
-                        "To resolve, run `list_symbols` through to completion to compact the symbol list cache.\n"
-                        "Note: This warning will only appear once.\n");
-
-                warned_expected_slowdown_ = true;
-            }
-        });
-        auto res = batch_get_latest_version(store, version_map_, stream_ids, false);
+    if(seg.descriptor().field_count() == 6) {
+        for (auto i = 0L; i < seg.column(3).row_count(); ++i) {
+            util::check(seg.column(3).row_count() == seg.column(4).row_count() && seg.column(3).row_count() == seg.column(5).row_count(),
```
alexowens90, line 190: This check should be outside the for loop

alexowens90 (changes requested):

cpp/arcticdb/stream/merge_utils.hpp
vasil-pashov, line 1: Is there any reason not to put the implementation in a `.cpp` file?

cpp/arcticdb/stream/merge_utils.hpp
```diff
@@ -0,0 +1,106 @@
...
+    ChunkedBuffer& src_buffer,
+    const std::shared_ptr<StringPool>& src_pool,
+    const std::shared_ptr<StringPool>& merged_pool,
+    CursoredBuffer<ChunkedBuffer>& output,
+    bool verify
+) {
+    using OffsetType = StringPool::offset_t;
+    constexpr auto offset_size =  sizeof(OffsetType);
+    auto num_strings = src_buffer.bytes() / offset_size;
+    for(auto row = size_t(0); row < num_strings; ++row) {
```
vasil-pashov, line 18: NIT: having both `auto` and the actual types seems redundant. We could just type `size_t row = 0`

cpp/arcticdb/version/test/test_symbol_list.cpp
willdealtry: I guess. It completely ignores the timestamps in terms of correctness though, just uses their closeness to flag potentially problematic cases, which is tested. Having a timestamp in the updated vector which is earlier than the existing (i.e. compacted one) counts as too close (i.e., it's not an absolute calculation but a signed one), this is tested in the RealTimestamps test.

cpp/arcticdb/stream/merge_utils.hpp
willdealtry: Think I relocated this code from elsewhere, I agree it is a bit odd. Although if we are initializing with literal types I would prefer to be specific, as the above actually says 'unsigned long row = int'

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -175,20 +175,19 @@ std::set<StreamId> LocalVersionedEngine::list_streams_internal(
     std::optional<SnapshotId> snap_name,
     const std::optional<std::string>& regex,
     const std::optional<std::string>& prefix,
-    const std::optional<bool>& opt_use_symbol_list,
-    const std::optional<bool>& opt_all_symbols
+    const std::optional<bool>& use_symbol_list,
```
vasil-pashov, line 178: Since we don't use `value_or` does it make sense for this to be optional anymore? `use_symbol_list.value_or(cfg().symbol_list())` will throw if there is no value so we're kind of expecting there to be one. Why not enforce it in the definition of the function?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,738 @@
...
+        const std::shared_ptr<StreamSource>& store,
+        SymbolListData& data) {
+    std::vector<AtomKey> output;
+    uint64_t uncompacted_keys_found = 0;
+    store->iterate_type(KeyType::SYMBOL_LIST, [&data, &output, &uncompacted_keys_found] (auto&& key) {
+        auto atom_key = to_atom(std::forward<decltype(key)>(key));
+        if(atom_key.id() != compaction_id) {
+            uncompacted_keys_found++;
+        }
+        if (uncompacted_keys_found == data.max_delta_ * 2 && !data.warned_expected_slowdown_) {
```
vasil-pashov, line 96: What is the idea of this check `uncompacted_keys_found == data.max_delta_ * 2`? Is this some sort of heuristic?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,738 @@
...
+        return output;
+
+    const auto data_type = get_symbol_data_type(seg);
+
+    // Because we need to be backwards compatible with the old style symbol list, the additions and deletions are
+    // in separate columns. The first three columns are the symbol, version and timestamp for the additions, and the
+    // next three are the same for the deletions. Old-style symbol lists will ignore everything but the first column
+    // which will mean that they can't do any conflict resolution but will get the correct data.
+    for(auto i = 0L; i < seg.column(0).row_count(); ++i) {
+        util::check(seg.column(0).row_count() == seg.column(1).row_count() && seg.column(0).row_count() == seg.column(2).row_count(),
```
vasil-pashov, line 178: Having an enum with the column names would be more readable than using "magic numbers"

cpp/arcticdb/stream/merge_utils.hpp
vasil-pashov, line 1: I think we should add a BSL header here.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,738 @@
...
-            const folly::Range<KeyVectorItr>& keys) {
-        SYMBOL_LIST_RUNTIME_LOG("Loading symbols from symbol list keys");
-        bool read_compaction = false;
-        CollectionType symbols{};
-        for(const auto& key : keys) {
-            if(key.id() == compaction_id) {
-                read_list_from_storage(store, key, symbols);
-                read_compaction = true;
+    if(!problematic_symbols.empty()) {
+        std::map<StreamId, VersionId> symbol_versions;
```
vasil-pashov, line 409: Why do we need ordered map?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -6,292 +6,738 @@
...
+        const SymbolListEntry& entry,
+        position_t column) {
+    util::variant_match(type_holder,
+        [&entry, &list_segment, column](const StringId&) {
+            util::check(std::holds_alternative<StringId>(entry.stream_id_), "Cannot write string symbol name, existing symbols are numeric");
+            list_segment.set_string(column, std::get<StringId>(entry.stream_id_));
+        },
+        [&entry, &list_segment, column](const NumericId&) {
+            util::check(std::holds_alternative<NumericId>(entry.stream_id_), "Cannot write numeric symbol name, existing symbols are strings");
+            list_segment.set_scalar(column, std::get<NumericId>(entry.stream_id_));
```
vasil-pashov, line 568: NIT: maybe another variant match would make this shorter

cpp/arcticdb/version/test/symbol_list_backwards_compat.hpp
vasil-pashov, line 1: Don't we need BSL headers for new files?

vasil-pashov, line 1: Can we move the implementation to a `.cpp`

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -17,6 +17,44 @@
...
+          return async::submit_io_task(CheckReloadTask{store, version_map, symbol, load_param});
+        },
+        [output](const auto& id, auto &&entry) {
+          auto index_key = entry->get_first_index(false);
+          if (index_key) {
+              output->insert(std::make_pair<StreamId, SymbolStatus>(StreamId{id}, {index_key->version_id(), true, index_key->creation_ts()}));
+          } else {
+              index_key = entry->get_first_index(true);
+              if (index_key)
+                  output->insert(std::make_pair<StreamId, SymbolStatus>(StreamId{id}, {index_key->version_id(), false, index_key->creation_ts()}));
```
alexowens90, line 51: Is this correct? `entry->get_first_index(true)` will fetch the first index key in the version chain, but that isn't necessarily the "reference version id" we're looking for. e.g. if you do
`write -> write -> delete_version(1) -> delete_version(0)`
then `index_key->version_id()` here will be 0, but the reference version id is 1.

PR #810, opened 2023-08-30
Bugfix: nativeversionstore read metadata batch and write batch should never return dataerror objects

alexowens90:

The NativeVersionStore API (used internally at Man Group) would previously throw an exception if there was a key missing from the storage required to complete a `batch_read_metadata` or `batch_write` operation. This was accidentally modified in PRs https://github.com/man-group/ArcticDB/pull/748 and https://github.com/man-group/ArcticDB/pull/757, such that DataError objects were returned instead. This was noticed prior to merging of those PRs, and so they were merged into an integration branch so that the `NativeVersionStore` API on `master` remained the same. This PR restores the previous behaviour, for the NativeVersionStore API, while maintaining the new behaviour for the Library API.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/integration/arcticdb/version_store/test_basic_version_store.py
+++ b/python/tests/integration/arcticdb/version_store/test_basic_version_store.py
@@ -1341,6 +1341,25 @@ def test_batch_write_then_list_symbol_without_cache(request, factory_name):
         assert set(lib.list_symbols()) == set(symbols)
 
 
+def test_batch_write_missing_keys_dedup(version_store_factory):
+    """When there is duplicate data to reuse for the current write, we need to access the index key of the previous
+    versions in order to refer to the corresponding keys for the deduplicated data."""
+    lib = version_store_factory(de_duplication=True)
+    assert lib.lib_cfg().lib_desc.version.write_options.de_duplication
+
+    df1 = pd.DataFrame({"a": [3, 5, 7]})
+    df2 = pd.DataFrame({"a": [4, 6, 8]})
+    lib.write("s1", df1)
+    lib.write("s2", df2)
+
+    lib_tool = lib.library_tool()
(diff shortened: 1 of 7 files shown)


python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -1573,6 +1592,42 @@ def test_batch_read_meta_with_missing(lmdb_version_store_tombstone_and_sync_pass
...
+    lib_tool = lib.library_tool()
+    s1_index_key = lib_tool.find_keys_for_id(KeyType.TABLE_INDEX, "s1")[0]
+    s2_version_keys = lib_tool.find_keys_for_id(KeyType.VERSION, "s2")
+    s2_key_to_delete = [key for key in s2_version_keys if key.version_id == 0][0]
+    lib_tool.remove(s1_index_key)
+    lib_tool.remove(s2_key_to_delete)
+
+    # The exception thrown is different for missing version keys to everything else, and so depends on which symbol is
+    # processed first
+    with pytest.raises((NoDataFoundException, StorageException)):
```
GemDot, line 1615: If one of the exception types we want to verify never gets raised, would the test still always pass?

GemDot (approved):

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
alexowens90: Turns out `NoDataFoundException` was never being raised. I've separated the test out now so missing index and version keys are verified separately.

GemDot: great!

PR #811, opened 2023-08-30
Fix/229 LMDB configurable map size

### LINKED ISSUE #229: LMDB map size is too small on Windows
To get the Windows tests to run without using too much disk, we changed the LMDB map size on Windows to about 8 MiB by default.

This won't be sufficient for many external users (within Man Group LMDB is pretty much just used for our automated tests and only S3 is used in production).

We should either:
- Bump up the default, we are cleaning up our LMDB test fixtures correctly now so might be able to get away with it
- Make the map size configurable as a LibraryOption in the external API
- Both of the above

poodlewars:
#### Reference Issues/PRs

Fixes #229 

#### API Changes

Backwards compatible.

You can now create an LMDB connection with a map size specified in the form `[0-9]+(KB|MB|GB|TB)`. See the Arctic `__init__.py` docstring for details. 

```
ac = Arctic("lmdb:///home/user/my_lmdb?map_size=2GB")
```

We allow users to set an LMDB mapsize in their Arctic URI.

We put this option in the URI since it must be set before the LMDB env is open. Since in our model the Arctic instance is what holds the LMDB env open, this fits the lifetimes well.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/arcticdb/adapters/lmdb_library_adapter.py
```diff
@@ -17,22 +20,93 @@
...
+        query = query.strip("?")
+    elif not query:
+        return ParsedQuery()
+
+    parsed_query = re.split("[;&]", query)
+    parsed_query = {t.split("=", 1)[0]: t.split("=", 1)[1] for t in parsed_query}
+
+    result = dict()
+    for key in parsed_query.keys():
+        field_dict = {field.name: field for field in fields(ParsedQuery)}
```
G-D-Petrov, line 80: its not a big deal but shouldn't this be outside of the for loop?
I don't think we need to recreate the field_dict on every iteration

cpp/arcticdb/storage/library_manager.hpp
```diff
@@ -128,12 +128,17 @@ namespace {
...
                 },
                 [&lib_cfg_proto] (const AzureOverride& credentials_override) {
                     for(auto& storage: *lib_cfg_proto.mutable_storage_by_id()){
-                        credentials_override.modify_storage_credentials(storage.second);
+                        credentials_override.modify_storage_config(storage.second);
+                    }
+                },
+                [&lib_cfg_proto] (const LmdbOverride& credentials_override) {
+                    for(auto& storage: *lib_cfg_proto.mutable_storage_by_id()){
+                        credentials_override.modify_storage_config(storage.second);
```
vasil-pashov, line 141: Those 3 lambdas seem identical to me. We can use a single generic lambda instead of copy-pasting the same code.

python/arcticdb/adapters/lmdb_library_adapter.py
```diff
@@ -17,22 +20,93 @@
...
+    """
+    suffix = size[-2:]
+    if suffix not in SUFFIX_TO_SCALE:
+        raise LmdbOptionsError(MAP_SIZE_OPTION_ERROR_TEMPLATE.format(size))
+    scale = SUFFIX_TO_SCALE[suffix]
+    mantissa_str = size[:-2]
+    try:
+        mantissa = int(mantissa_str)
+        if mantissa < 1:
+            raise ValueError()
```
vasil-pashov, line 63: Why do we raise ValueError in this case but LmdbbOptionsError int the other two cases?

poodlewars: It immediately gets caught and rethrown below

vasil-pashov (approved):

PR #822, opened 2023-08-31
Fix the dependencies for the can_merge step

G-D-Petrov:
It seems that when a step depends on another step and that step is skipped, the downstream one is also skipped by default. This can be fixed with a check as per this stack overflow thread: 
https://stackoverflow.com/questions/69354003/github-action-job-fire-when-previous-job-skipped


poodlewars:
This should help to fix the publish step since the graph is like:

image

poodlewars (approved):

PR #827, opened 2023-09-01
Bugfix 767: Fix segfault in batch write with string columns

### LINKED ISSUE #767: write batch produces segmentation fault when using string indexes and dynamic strings is enabled
### Describe the bug

When writing in batches and using string indexes with dynamic strings set to true, we call the function PyUnicode_AsUTF8String. However, this function is not thread-safe, and accessing it concurrently can cause segmentation faults. 

To prevent concurrent access to this function, we need to use mechanisms like the Python GIL locks.

### Steps/Code to Reproduce

```
def test_batch_write_segfault(s3_version_store_v1):
     lib = s3_version_store_v1
     syms = ["opt_hard_upper_posbounds", "estimated_wave_volume_usd"]

     index1 = [f"AST{i}" for i in range(586)]
     index2 = [f"AST{i}" for i in range(585)]

     vals1 = np.random.randn(586).astype('float64')
     vals2 = np.random.randn(585).astype('float64')

     data = [pd.Series(data=vals1, index=index1), pd.Series(data=vals2, index=index2)]

     # sometimes segfaults the first time, sometimes doesn't, so run in a loop and it is very likely to crash
     for i in range(5):
         lib.batch_write(syms, data)
```

### Expected Results

We do not expect segmentation faults.

Python: 3.6.8 (default, Jun 20 2023, 11:53:23) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]
OS: Linux-5.15.0-76-generic-x86_64-with-centos-7.9.2009-Core

### Backend storage used

_No response_

### Additional Context

_No response_

alexowens90:
#### Reference Issues/PRs

Closes #767 

The segfault was caused by the allocation of `PyObject`s in threads that were not holding the GIL. For ASCII and UTF-8 strings, we do not need to use the `PyUnicode_AsUTF8String` method performing the allocation under the hood. It is sufficient to take a view over the underlying buffer. This should speed up writes of ASCII and UTF-8 strings in general.

For other unicode strings, it was considered to re-implement the logic to produce a UTF-8 buffer, which would still involve an allocation, but without creating a `PyObject`. However, this logic is quite involved, and maintaining it would be challenging, and so this approach was not taken, although it could be in the future.

Instead, the `PyUnicode_AsUTF8String` method is still used for these strings, and so the GIL must be held while it is called. This is achieved by releasing the GIL when `batch_write` is called. If a unicode string that requires `PyUnicode_AsUTF8String` is found, the GIL is taken by that thread for the remainder of the processing of that column, under the assumption that if a column contains one such string, then it is likely to contain more. This avoids taking and releasing the GIL in quick succession for a column of unicode strings.

#### Any other comments?

Releasing the GIL on entry to batch write means that in theory the ndarrays we are writing could be modified by another thread during the writes. If this is the case, user's should be using a finer-grained lock than the GIL to control this modification. However, if necessary, we could set the ndarrays to read only on entry, and back to whatever they were before on exit using NPY_ARRAY_WRITEABLE.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/pipeline/frame_utils.hpp
+++ b/cpp/arcticdb/pipeline/frame_utils.hpp
@@ -15,6 +15,7 @@
 #include <arcticdb/pipeline/input_tensor_frame.hpp>
 #include <arcticdb/stream/protobuf_mappings.hpp>
 #include <arcticdb/entity/protobuf_mappings.hpp>
+#include <arcticdb/python/gil_lock.hpp>
 #include <arcticdb/python/python_types.hpp>
 #include <arcticdb/python/python_to_tensor_frame.hpp>
 #include <arcticdb/pipeline/string_pool_utils.hpp>
@@ -130,6 +131,11 @@ void aggregator_set_data(
                     ptr_data = flatten_tensor<PyObject*>(flattened_buffer, rows_to_write, tensor, slice_num, regular_slice_size);
 
                 auto none = py::none{};
+                // GIL will be acquired if there is a string that is not pure ASCII/UTF-8
+                // In this case a PyObject will be allocated by convert::py_unicode_to_buffer
+                // If such a string is encountered in a column, then the GIL will be held until that whole column has
+                // been processed, on the assumption that if a column has one such string it will probably have many.
+                std::optional<ScopedGILLock> scoped_gil_lock;
                 for (size_t s = 0; s < rows_to_write; ++s, ++ptr_data) {
                     if (*ptr_data == none.ptr()) {
                         agg.set_no_string_at(col, s, not_a_string());
@@ -137,7 +143,7 @@ void aggregator_set_data(
                         agg.set_no_string_at(col, s, nan_placeholder());
                     } else {
                         if constexpr (is_utf_type(slice_value_type(dt))) {
-                            auto wrapper= convert::py_unicode_to_buffer(*ptr_data);
+                            auto wrapper= convert::py_unicode_to_buffer(*ptr_data, scoped_gil_lock);
                             agg.set_string_at(col, s, wrapper.buffer_, wrapper.length_);
                         } else {
                             auto wrapper = convert::pystring_to_buffer(*ptr_data);
--- a/python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
+++ b/python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
@@ -241,3 +241,31 @@ def test_update_fixed_width_to_dynamic_strings(lib_type, request):
     expected_df = dynamic_strings_data
     read_df = lib.read(symbol).data
     assert_frame_equal(expected_df, read_df)
+
+
+# https://github.com/man-group/ArcticDB/issues/767
+# Batch write and append call aggregator_set_data from threads running in the CPU threadpool (i.e. not the main thread)
+# With unicode strings, a PyObject allocation is needed, and therefore the thread must be holding the GIL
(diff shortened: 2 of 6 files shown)


cpp/arcticdb/pipeline/frame_utils.hpp
```diff
@@ -130,14 +131,19 @@ void aggregator_set_data(
                     ptr_data = flatten_tensor<PyObject*>(flattened_buffer, rows_to_write, tensor, slice_num, regular_slice_size);
 
                 auto none = py::none{};
+                // GIL will be acquired if there is a string that is not pure ASCII/UTF-8
+                // In this case a PyObject will be allocated by convert::py_unicode_to_buffer
+                // If such a string is encountered in a column, then the GIL will be held until that whole column has
+                // been processed, on the assumption that if a column has one such string it will probably have many.
+                std::unique_ptr<ScopedGILLock> scoped_gil_lock;
```
vasil-pashov, line 138: I don't think we need this on the heap. Maybe std::optional would be a better choice.

python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
```diff
@@ -241,3 +241,31 @@ def test_update_fixed_width_to_dynamic_strings(lib_type, request):
...
     read_df = lib.read(symbol).data
     assert_frame_equal(expected_df, read_df)
+
+
+# https://github.com/man-group/ArcticDB/issues/767
+# Batch write and append call aggregator_set_data from threads running in the CPU threadpool (i.e. not the main thread)
+# With unicode strings, a PyObject allocation is needed, and therefore the thread must be holding the GIL
+# This test ensures that the correct thread is holding the GIL when performing these allocations, and that there is no
+# deadlock.
+# This is not an issue with non-batch methods as aggregator_set_data is called from the main thread, which pybind11
```
willdealtry, line 251: I'm not convinced this is the case, non-batch writes do the slicing in the main thread, but then write in parallel to the individual segments, calling aggregator_set_data. Is it worth copying and pasting this test into a non-batch version just to check that it doesn't deadlock?

alexowens90: I might be missing something, but this seems to be before the parallelisation begins?
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/pipeline/write_frame.cpp#L96

I did test that it didn't deadlock with `lmdb_version_store_tiny_segment` and unicode strings with both `write` and `append`. With the debugger I could see the `scoped_gil_lock` successfully acquiring the GIL, which must have meant it was already held by the calling thread?

willdealtry: Yeah you're correct, I had forgotten exactly where the parallelization started. That's not ideal actually, if there is work to be done in aggregator_set_data for other object types we might need to fix it, by moving the work into the parallelized bit. Probably some write-time equivalent of the read handlers would be useful both in achieving that and finishing the job of moving the Python-specific bits out of the core.

willdealtry (approved):

python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
alexowens90: Yeh, there's some non-trivial work to do in that loop for populating the string pools at least, so parallelisation could bring some benefit. Should be pretty trivial as well. I'll add a ticket.

Issue #772, opened 2023-08-21
Jupyter Notebook for ArcticDB Diagnosis


GemDot:
**Is your feature request related to a problem? Please describe.**
We seek an automated and user-friendly approach to detect common problems when performing various ArcticDB methods. The tool should identify prevalent performance and functional issues, and serve an educational purpose for users.

**Describe the solution you'd like**
We would like an easy and intuitive diagnosis API that can provide meaningful information about potential problems on a per-method basis. Additionally, this information should be supported with a significant amount of visual aids.

**Describe alternatives you've considered**
The alternative is to embed this tool directly in ArcticDB, but Jupyter notebooks are more commonly used and are more user-friendly.

**Additional context**
The context originates from all the user requests we've received so far regarding ArcticDB. We aim to automate the diagnosis of the most common problems.

alexowens90:
Still to do:

- [ ] 	Library objects should show LibraryOptions
- [ ] 	Review English
- [ ] 	Testing

PR #834, opened 2023-09-06
ArcticDB diagnostics

GemDot-neubla:
#### Reference Issues/PRs

closes https://github.com/man-group/ArcticDB/issues/772

Introducing our cutting-edge diagnostics notebook, tailored specifically for ArcticDB. This powerful tool offers a seamless and efficient way to analyze and understand your ArcticDB data, by providing in-depth insights and performance metrics.
With our diagnostics notebook, users can:

- Easily assess different options for each ArcticDB library

- Gain valuable insights into the structure and performance of their databases

- Receive personalized explanations for each option

- Make informed decisions about their database operations

- Enhance overall productivity and effectiveness

#### Any other comments?

This is part of an effort to enhance the debuggability of ArcticDB and improve its understanding for both users and developers.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
+++ b/python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
@@ -8,9 +8,9 @@ def calculate_and_add_segment_occupancy_info(df):
     """
     Calculates the number of data points, rows, and columns for each segment in the DataFrame.
     """
-    df["data_points"] = (df["end_row"] - df["start_row"] + 1) * (df["end_col"] - df["start_col"] + 1)
-    df["rows"] = df["end_row"] - df["start_row"] + 1
-    df["columns"] = df["end_col"] - df["start_col"] + 1
+    df["data_points"] = (df["end_row"] - df["start_row"]) * (df["end_col"] - df["start_col"])
+    df["rows"] = df["end_row"] - df["start_row"]
+    df["columns"] = df["end_col"] - df["start_col"]
     return df
 
 
@@ -34,13 +34,15 @@ def calculate_segments_occupancy_maximums(df, total_segment_size, segment_row_si
     return max_data_points, max_rows, max_columns
 
 
-def calculate_fragmentation(total_data_points, total_rows, total_columns, max_data_points, max_rows, max_columns):
+def calculate_fragmentation(
+    total_data_points, total_rows, total_columns, max_data_points, max_rows, max_columns, dynamic_schema
+):
     """
     Calculates the overall fragmentation level as well as row-wise and column-wise fragmentation for the data.
     """
     fragmentation = float(total_data_points) / max_data_points
     per_row_fragmentation = float(total_rows) / max_rows
-    per_column_fragmentation = float(total_columns) / max_columns
+    per_column_fragmentation = None if dynamic_schema else float(total_columns) / max_columns
     return fragmentation, per_row_fragmentation, per_column_fragmentation
 
 
@@ -76,7 +78,7 @@ def display_message(fragmentation, per_row_fragmentation, per_column_fragmentati
             "However, be careful with the latter option as it can have other implications."
         )
 
-    if per_column_fragmentation < 0.5:
+    if per_column_fragmentation is not None and per_column_fragmentation < 0.5:
         message += (
             f"\n- Your column utilization is only about {per_column_fragmentation * 100:.1f}%. This could occur when"
             " the number of columns in your data exceeds the column group size. For example, if your column group size"
@@ -97,7 +99,8 @@ def plot_data_points(df, total_segment_size):
     description = (
         "\nThe plot below represents the distribution of the number of data points per data segment. Understanding this"
         " distribution can help optimize data storage and retrieval operations. The dotted red line represents the"
-        " maximum number of data points per segment."
+        " maximum number of data points per segment. Thus, the further we are from the dotted line, the more fragmented"
+        " the data is."
     )
     display(Markdown(description))
 
@@ -107,6 +110,27 @@ def plot_data_points(df, total_segment_size):
     plt.xlabel("Number of Data Points per Segment")
     plt.ylabel("Frequency")
     plt.title("Histogram of Number of Data Points per Data Segment")
+    plt.show()
+
+
+def plot_per_row_fragmentation(df, segment_row_size):
+    """
+    Plots a histogram of the number rows per segment.
+    """
+    description = (
+        "\nThe plot below represents the distribution of the number of rows per data segment. Understanding this"
+        " distribution can help optimize data storage and retrieval operations. The dotted red line represents the"
+        " maximum number of rows per segment. Thus, the further we are from the dotted line, the more fragmented the"
+        " data is."
+    )
+    display(Markdown(description))
+
+    plt.figure(figsize=(16, 9))
+    plt.hist(df["rows"], bins=100, rwidth=10)
+    plt.axvline(segment_row_size, color="r", linestyle="--")
+    plt.xlabel("Number of rows per Segment")
+    plt.ylabel("Frequency")
+    plt.title("Histogram of Number of rows per Data Segment")
     plt.show()
 
 
@@ -148,6 +172,7 @@ def fragmentation_diagnosis(lib, symbol, as_of=None):
 
     segment_row_size = lib._lib_cfg.lib_desc.version.write_options.segment_row_size
     column_group_size = lib._lib_cfg.lib_desc.version.write_options.column_group_size
+    dynamic_schema = lib._lib_cfg.lib_desc.version.write_options.dynamic_schema
 
     # Calculate the actual amount of data stored across all the segments
     df = lib.read_index(symbol, as_of=as_of)
@@ -162,9 +187,11 @@ def fragmentation_diagnosis(lib, symbol, as_of=None):
 
     # Infer fragmentation
     fragmentation, per_row_fragmentation, per_column_fragmentation = calculate_fragmentation(
-        sum_data_points, sum_rows, sum_columns, max_data_points, max_rows, max_columns
+        sum_data_points, sum_rows, sum_columns, max_data_points, max_rows, max_columns, dynamic_schema
     )
 
     # Display Results
     display_message(fragmentation, per_row_fragmentation, per_column_fragmentation)
     plot_data_points(df, total_segment_size)
+    if not dynamic_schema:
+        plot_per_row_fragmentation(df, segment_row_size)
--- a/python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
+++ b/python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
@@ -1,0 +1,170 @@
+# import necessary libraries
+from IPython.display import display, Markdown
+import matplotlib.pyplot as plt
+from .library_utils import check_and_adapt_library, check_symbol_exists
+
+
+def calculate_and_add_segment_occupancy_info(df):
+    """
+    Calculates the number of data points, rows, and columns for each segment in the DataFrame.
+    """
+    df["data_points"] = (df["end_row"] - df["start_row"] + 1) * (df["end_col"] - df["start_col"] + 1)
+    df["rows"] = df["end_row"] - df["start_row"] + 1
+    df["columns"] = df["end_col"] - df["start_col"] + 1
+    return df
+
+
+def calculate_sum_segments_occupancy_info(df):
+    """
+    Calculates the total sum of data points, rows, and columns in the DataFrame.
+    """
+    total_data_points = df["data_points"].sum()
+    total_rows = df["rows"].sum()
+    total_columns = df["columns"].sum()
+    return total_data_points, total_rows, total_columns
+
+
+def calculate_segments_occupancy_maximums(df, total_segment_size, segment_row_size, column_group_size):
+    """
+    Calculates the maximum number of data points, rows, and columns that can be stored in the segments.
+    """
+    max_data_points = total_segment_size * df.shape[0]
+    max_rows = segment_row_size * df.shape[0]
+    max_columns = column_group_size * df.shape[0]
+    return max_data_points, max_rows, max_columns
+
+
+def calculate_fragmentation(total_data_points, total_rows, total_columns, max_data_points, max_rows, max_columns):
+    """
+    Calculates the overall fragmentation level as well as row-wise and column-wise fragmentation for the data.
+    """
+    fragmentation = float(total_data_points) / max_data_points
+    per_row_fragmentation = float(total_rows) / max_rows
+    per_column_fragmentation = float(total_columns) / max_columns
+    return fragmentation, per_row_fragmentation, per_column_fragmentation
+
+
+def display_message(fragmentation, per_row_fragmentation, per_column_fragmentation):
+    """
+    Displays a markdown message based on the calculated fragmentation levels.
+    """
+    message = ""
+    if fragmentation < 0.5:
+        message += (
+            f"\n**Warning!** Your overall fragmentation is about {fragmentation * 100:.1f}%. The data appears quite"
+            " fragmented. "
+        )
+    elif fragmentation < 0.75:
+        message += (
+            f"\n**Attention:** Your overall fragmentation is about {fragmentation * 100:.1f}%. The data appears to be"
+            " somewhat fragmented. "
+        )
+    else:
+        message += (
+            f"\n**Good news:** Your overall fragmentation is about {fragmentation * 100:.1f}%. No major fragmentation"
+            " issues detected. "
+        )
+        display(Markdown(message))
+        return
+
+    if per_row_fragmentation < 0.5:
+        message += (
+            f"\n- Your row utilization is only about {per_row_fragmentation * 100:.1f}%. This typically occurs when"
+            " updates or appends add a small number of rows per data segment, compared to the total row capacity of"
+            " the segment. You may consider updating or appending the data in larger chunks or reducing the number "
+            "of rows defined by `segment_row_size` in the library options to optimize data storage and retrieval."
+            "However, be careful with the latter option as it can have other implications."
+        )
+
+    if per_column_fragmentation < 0.5:
+        message += (
+            f"\n- Your column utilization is only about {per_column_fragmentation * 100:.1f}%. This could occur when"
+            " the number of columns in your data exceeds the column group size. For example, if your column group size"
+            " is 50 and your data contains 60 columns, it will result in one segment of 50 columns and another"
+            " inefficient segment of 10 columns. Unfortunately, this is less controllable as it depends on the nature"
+            " of your data. If this situation is frequently encountered, consider evaluating the column group size"
+            " setting in your library, but tread cautiously as this may have other implications and could affect other"
+            " symbols in the same library."
+        )
+
+    display(Markdown(message))
+
+
+def plot_data_points(df, total_segment_size):
+    """
+    Plots a histogram of the number of data points per segment.
+    """
+    description = (
+        "\nThe plot below represents the distribution of the number of data points per data segment. Understanding this"
+        " distribution can help optimize data storage and retrieval operations. The dotted red line represents the"
+        " maximum number of data points per segment."
+    )
+    display(Markdown(description))
+
+    plt.figure(figsize=(16, 9))
+    plt.hist(df["data_points"], bins=20, rwidth=100)
+    plt.axvline(total_segment_size, color="r", linestyle="--")
+    plt.xlabel("Number of Data Points per Segment")
+    plt.ylabel("Frequency")
+    plt.title("Histogram of Number of Data Points per Data Segment")
+    plt.show()
+
+
+def fragmentation_diagnosis(lib, symbol, as_of=None):
+    """
+    Analyzes and displays the fragmentation level for a specified symbol from a library.
+
+    The function checks the provided library's type, reads the index of the symbol, and calculates
+    various statistics such as the total segment size, fragmentation, row-wise and column-wise fragmentation.
+    Based on these statistics, it generates a markdown report and a histogram plot showing the distribution
+    of data points per segment.
+
+    Parameters:
+    -----------
+    lib : object
+        The library object in which the symbol is located.
+    symbol : str
+        The symbol for which the fragmentation analysis is to be performed.
+    as_of : int, optional
+        The version of the symbol for which the analysis is to be performed.
+        If not specified, the latest version is used.
+
+    Returns:
+    --------
+    None
+    Outputs a markdown report and a histogram plot detailing the fragmentation analysis.
+
+    Example:
+    --------
+    fragmentation_diagnosis(lib, 'my_symbol', 3)
+    """
+
+    lib = check_and_adapt_library(lib)
+    if lib is None:
+        return
+
+    if not check_symbol_exists(lib, symbol, as_of):
+        return
+
+    segment_row_size = lib._lib_cfg.lib_desc.version.write_options.segment_row_size
+    column_group_size = lib._lib_cfg.lib_desc.version.write_options.column_group_size
+
+    # Calculate the actual amount of data stored across all the segments
+    df = lib.read_index(symbol, as_of=as_of)
+    df_with_segment_info = calculate_and_add_segment_occupancy_info(df)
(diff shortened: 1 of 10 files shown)


python/arcticdb/diagnostics_engine/version_list_diagnosis.py
```diff
@@ -0,0 +1,342 @@
...
+def get_target_version(parsed_versions, parsed_keys, as_of):
+    """
+    Get the target version based on the 'as_of', or the latest version if 'as_of' is None.
+    """
+    if as_of is not None:
+        return as_of
+
+    alive_versions = parsed_versions.keys()
+    for key in parsed_keys:
+        if key in alive_versions and parsed_versions[key.version_id]["deleted"] == False:
```
alexowens90, line 208: `key in alive_versions`
I don't understand how this works. Isn't `key` an `AtomKey` and `alive_versions` a collection of integers?

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
```diff
@@ -0,0 +1,342 @@
...
+    """
+    steps = 0
+    is_present = False
+    for i, key in enumerate(parsed_keys):
+        steps += 1
+        if key.version_id == target:
+            is_present = True
+            break
+
+    return steps - 1, is_present
```
alexowens90, line 224: If you put this on line 221 you can get rid of the `break`

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
```diff
@@ -0,0 +1,342 @@
...
+        if parsed_versions[target]["deleted"] == False:
+            if present_in_version_list:
+                exist_in_version_list = True
+            else:
+                exists_as_compacted_version = True
+        else:
+            if parsed_versions[target]["snapshots"]:
+                exists_in_snaphots = True
+    elif present_in_version_list:
+        is_tombstoned = True
```
alexowens90, line 258: A version can be deleted but also present in snapshots

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
```diff
@@ -0,0 +1,342 @@
...
+
+def get_compacted_versions(parsed_versions, last_version_in_list):
+    """
+    Get all the compacted versions after the last node from the version list.
+    """
+    compacted_versions = []
+    for version in parsed_versions.keys():
+        if parsed_versions[version]["deleted"] == False and version <= last_version_in_list:
+            compacted_versions.append(version)
+    return compacted_versions
```
alexowens90, line 235: Is this right? Does version compaction always compact everything into one object, including previously compacted objects?

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+    else:
+        description = (
+            f"***Version {target} has been deleted for that symbol***. It takes {steps} extra I/O operations to"
+            " traverse the version chain until "
+        )
+        if is_tombstoned:
+            description += "finding the deletion marker (tombstone)"
+        else:
+            description += "reaching the end. "
```
alexowens90, line 50: In what circumstances do we take this code path?

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
```diff
@@ -0,0 +1,342 @@
...
+    # get non-deleted versions
+    versions = lib.list_versions(symbol)
+    parsed_versions = parse_versions(versions)
+
+    # get actual version keys chain
+    keys = lib_tool.find_keys_for_id(KeyType.VERSION, symbol)
+    parsed_keys = parse_keys(keys)
+
+    # get specific version we are looking for
+    target = get_target_version(parsed_versions, parsed_keys, as_of)
```
alexowens90, line 303: What if `as_of` is a timestamp? Or a snapshot name?

alexowens90 (changes requested):

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: To cut down the complexity of the task, it has been intentionally left unsupported.

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: It's possible that after a compaction, the tombstone has already been compacted. Thus, you will traverse the list until the end, meaning until you find the version that contains the remainder of the compacted keys.

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: Yes, I conducted some experiments myself and also confirmed with @willdealtry. All the compacted versions are included in one version, which is always the last one in the chain.

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: Yes, that's definitely a bug. I had been considering this scenario all the time, but it wasn't correctly reflected in the logic. Good catch!"

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: I would prefer not to do it, as I would still require this return to manage situations where the version isn't present.

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: I will enhance the code's clarity and also add a comment, as it's clear that the current implementation isn't obvious.

python/arcticdb/diagnostics_engine/sortedness_diagnosis.py
```diff
@@ -0,0 +1,181 @@
...
+            " or UNKNOWN state and is also a time series. Please note that the 'Update' method does not have a"
+            " user-settable validate_index flag."
+        )
+
+
+def get_read_description(sorted_state):
+    """
+    Provides a description for the read method based on the current sorted state.
+    """
+    if sorted_state != "ASCENDING" and sorted_state != "UNKNOWN":
```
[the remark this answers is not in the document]
GemDot: agree

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
```diff
@@ -0,0 +1,170 @@
...
+    # Calculate the maximum amount of data that can be stored across all segments
+    total_segment_size = segment_row_size * column_group_size
+    max_data_points, max_rows, max_columns = calculate_segments_occupancy_maximums(
+        df_with_segment_info, total_segment_size, segment_row_size, column_group_size
+    )
+
+    # Infer fragmentation
+    fragmentation, per_row_fragmentation, per_column_fragmentation = calculate_fragmentation(
+        sum_data_points, sum_rows, sum_columns, max_data_points, max_rows, max_columns
+    )
```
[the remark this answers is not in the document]
GemDot: that's a fair point

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
```diff
@@ -0,0 +1,170 @@
...
+        "\nThe plot below represents the distribution of the number of data points per data segment. Understanding this"
+        " distribution can help optimize data storage and retrieval operations. The dotted red line represents the"
+        " maximum number of data points per segment."
+    )
+    display(Markdown(description))
+
+    plt.figure(figsize=(16, 9))
+    plt.hist(df["data_points"], bins=20, rwidth=100)
+    plt.axvline(total_segment_size, color="r", linestyle="--")
+    plt.xlabel("Number of Data Points per Segment")
```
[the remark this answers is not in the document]
GemDot: No, it is an actual number

Example:

image

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
willdealtry: Compaction always compacts everything at the moment, from the second key in the version list. It doesn't have to - the version list would work fine if it was large chunks of data connected by keys that act as a pointer to the next one like a std::deque rather than a std::vector, but that's not implemented in the compaction at the moment.

https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/version/version_map.hpp#L374

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
```diff
@@ -0,0 +1,170 @@
...
+from .library_utils import check_and_adapt_library, check_symbol_exists
+
+
+def calculate_and_add_segment_occupancy_info(df):
+    """
+    Calculates the number of data points, rows, and columns for each segment in the DataFrame.
+    """
+    df["data_points"] = (df["end_row"] - df["start_row"] + 1) * (df["end_col"] - df["start_col"] + 1)
+    df["rows"] = df["end_row"] - df["start_row"] + 1
+    df["columns"] = df["end_col"] - df["start_col"] + 1
```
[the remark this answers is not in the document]
GemDot: We are all wrong here. We don't any of them actually. Good catch.

python/arcticdb/diagnostics_engine/index_key_structure.py
```diff
@@ -0,0 +1,76 @@
...
+
+    This function takes a DataFrame that contains data segment information, resets its index,
+    selects specific columns for display, and then visually presents this information in a
+    styled table.
+    """
+    # Reset index to convert 'start_index' to a regular column
+    df_reset = df.reset_index()
+
+    # Select the columns to be displayed
+    columns = ["version_id", "start_index", "end_index", "creation_ts", "start_row", "end_row", "start_col", "end_col"]
```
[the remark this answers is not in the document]
GemDot: yes, good point

python/arcticdb/diagnostics_engine/library_options.py
```diff
@@ -0,0 +1,286 @@
...
+        and updates response strings accordingly. The function is recursive for nested protobuf message types.
+        """
+        default_response_obj = type(obj)()
+        for field in obj.DESCRIPTOR.fields:
+            value = getattr(obj, field.name)
+            if field.message_type:
+                self.traverse_descriptor(value)
+            elif field.name in self.explanations:
+                default_value = getattr(default_response_obj, field.name)
+                explanation = self.explanations.get(field.name, "No explanation required.")
```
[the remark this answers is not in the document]
GemDot: I am unsure of what do you exactly mean in your comment:

1. `default_value` is used in line 240, in order to determine whether we changed the original value or not.
2. I intentionally want to control what options to show to the user. It is why `field.name` must be included in `self.explanations`.

python/arcticdb/diagnostics_engine/library_options.py
```diff
@@ -0,0 +1,286 @@
...
+                if value != default_value:
+                    if field.name in self.internal_default_values:
+                        default_value = self.internal_default_values[field.name]
+                    self.response_non_default_str += (
+                        f"\n### **{field.name}**: {value} (Default: {default_value})\n\n{explanation}\n\n"
+                    )
+                else:
+                    if field.name in self.internal_default_values:
```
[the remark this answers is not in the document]
GemDot: No, I can't do that. I've already considered this possibility. If the first check (line 241) passes, we set the default_value variable. If the second check (line 247) passes, we set the value variable. Note that options defined internally within the library typically have their default value set to '0'. Therefore, we need to preserve this default_value until it can be compared with the actual value. If it has changed, we restore the actual default value (as defined internally). If it hasn't changed, then we know the actual value is the one defined internally.

I acknowledge it's a bit off. But it's primarily due to the fact that we hardcode things internally, which I believe is the root of the problem.

python/arcticdb/diagnostics_engine/symbol_list_consistency.py
```diff
@@ -0,0 +1,94 @@
...
+            f" the symbol list are: \n\n```\n{diff_set}\n```.\n\n- The set of symbols cached in the symbol list"
+            f" are:\n\n```\n{cached_symbols}\n```."
+        )
+
+    deleted_symbols_list = list(set(get_all_symbols_from_library(lib)) - non_cached_symbol_list_set)
+    if deleted_symbols_list:
+        deleted_symbols = ", ".join(deleted_symbols_list)
+        response += f"\n\n- The set of deleted symbols in your library are: ```\n{deleted_symbols}\n```."
+
+    deleted_symbols_present_in_cache_list = get_non_existent_symbols(lib, cached_symbol_list)
```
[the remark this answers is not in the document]
GemDot: Yes, I've considered both approaches and faced a dilemma. If we keep it as it is, we're certainly performing computations more than once, as you correctly noted. However, reusing the same information introduces a problem: the APIs for the created functions aren't self-contained and lack generality.

Can we have another thought about your comment @alexowens90 ? I am happy to do the changes as you pointed out. But first, I wanted to explain clearly the reason why I did it as is now.

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+    # Infer fragmentation
+    fragmentation, per_row_fragmentation, per_column_fragmentation = calculate_fragmentation(
+        sum_data_points, sum_rows, sum_columns, max_data_points, max_rows, max_columns
+    )
+
+    # Display Results
+    display_message(fragmentation, per_row_fragmentation, per_column_fragmentation)
+    plot_data_points(df, total_segment_size)
```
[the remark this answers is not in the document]
GemDot: Fair. I can do it

python/arcticdb/diagnostics_engine/library_utils.py
```diff
@@ -0,0 +1,30 @@
...
+        return None
+
+
+def check_symbol_exists(lib, symbol, as_of=None):
+    """
+    Verifies that the symbol and version (in case it is specified) exist
+    """
+    if as_of == None:
+        string_version = "the latest version"
+    else:
```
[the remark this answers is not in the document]
GemDot: As pointed out here https://github.com/man-group/ArcticDB/pull/834#discussion_r1322588867, it has been intentionally unsupported. So, I will change it only if we finally decide to support those options as well.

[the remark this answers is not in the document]
GemDot: done the corresponding changes

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: Done the corresponding changes to support these cases in a separate (the last one) commit

python/arcticdb/diagnostics_engine/library_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    if isinstance(lib, NativeVersionStore):
+        return lib
+    else:
+        display(Markdown(f"the provided library is not supported"))
+        return None
+
+
+def get_string_version(as_of):
+    if as_of is not None:
+        if isinstance(as_of, pd.Timestamp):
```
alexowens90, line 20: We support some other time types as well:
https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/supported_types.py#L21

python/arcticdb/diagnostics_engine/library_options.py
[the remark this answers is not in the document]
alexowens90: 1. I meant the default value which is the second argument to `get`
2. We should at least print that we don't have a detailed explanation for the setting in this case

python/arcticdb/diagnostics_engine/library_options.py
```diff
@@ -0,0 +1,286 @@
...
+        self.response_default_str = (
+            "## Next, see the list of your library options that remains with the default values:\n\n"
+        )
+
+        # Default values defined internally
+        self.internal_default_values = {
+            "column_group_size": 127,
+            "segment_row_size": 100000,
+            "max_num_buckets": 150,
+        }
```
alexowens90, line 225: I don't think this list is complete. If you create a new library within Man, then the protobuf defaults are overridden for a lot of settings.

python/arcticdb/diagnostics_engine/library_options.py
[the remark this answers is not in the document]
alexowens90: Oh I see what you mean

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
alexowens90: Ah yeh I hadn't noticed that, ignore me

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
alexowens90: Cool, I wasn't certain of the behaviour, just wanted to make sure it had been tested

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
```diff
@@ -0,0 +1,413 @@
...
+    # When the target version is None, we need to return the most recent 'alive' version, which is the
+    # first non-tombstone one. The 'alive_versions' contains all the integers from the version list,
+    # but we must always respect the order established by 'parsed_keys' (i.e., the actual physical version chain).
+    # This is why we iterate across this strict order. Finally, we need to check the 'deleted' flag from
+    # the version list, as we still need to exclude the versions that are present in the version list
+    # solely because they are included in any snapshot.
+    alive_versions = [key for key in parsed_versions.keys() if parsed_versions[key]["deleted"] == False]
+    for key in parsed_keys:
+        if key.version_id in alive_versions:
+            return key.version_id, False
```
alexowens90, line 255: I think this might miss the corner case where the latest live version is in a compacted key?
I think we should just use the output of `list_versions` to determine this though, live version numbers should always be increasing through the version chain.

python/arcticdb/diagnostics_engine/symbol_list_consistency.py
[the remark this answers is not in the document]
alexowens90: There are only 3 sets of symbols that are interesting:

- Those alive in the cache
- Those alive in version chains
- Those deleted in version chains

If you have methods to generate these 3 sets, everything else should just be set union/difference operations

alexowens90 (changes requested):

python/arcticdb/diagnostics_engine/symbol_list_consistency.py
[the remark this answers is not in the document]
GemDot: fine

python/arcticdb/diagnostics_engine/library_options.py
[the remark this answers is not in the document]
GemDot: ok. I see your point. That's certainly confusing because the existing logic makes the "No explanation required." default message totally unnecessary. It was intentional not to show options that we don't provide explanations. But happy to show them anyway.

python/arcticdb/diagnostics_engine/version_list_diagnosis.py
GemDot: True

python/arcticdb/diagnostics_engine/library_options.py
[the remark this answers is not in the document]
GemDot: Created an exclusion_set because there are some protobuf fields that are not considered optional per se when they're traversed recursively.

python/arcticdb/diagnostics_engine/library_utils.py
GemDot: fine

python/arcticdb/diagnostics_engine/library_options.py
GemDot: Apart from the boolean types, which don't allow the internal setting trick by comparison with zero, I've found only one additional item to add to this list. This is max_blob_size, which is set in the Python layer. I am surprised that you said "a lot". Am I missing something?

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
```diff
@@ -0,0 +1,197 @@
...
+    return max_data_points, max_rows, max_columns
+
+
+def calculate_fragmentation(
+    total_data_points, total_rows, total_columns, max_data_points, max_rows, max_columns, dynamic_schema
+):
+    """
+    Calculates the overall fragmentation level as well as row-wise and column-wise fragmentation for the data.
+    """
+    fragmentation = float(total_data_points) / max_data_points
```
alexowens90, line 43: This is still wrong with dynamic schema

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
```diff
@@ -0,0 +1,197 @@
...
+    # Infer fragmentation
+    fragmentation, per_row_fragmentation, per_column_fragmentation = calculate_fragmentation(
+        sum_data_points, sum_rows, sum_columns, max_data_points, max_rows, max_columns, dynamic_schema
+    )
+
+    # Display Results
+    display_message(fragmentation, per_row_fragmentation, per_column_fragmentation)
+    plot_data_points(df, total_segment_size)
+    if not dynamic_schema:
+        plot_per_row_fragmentation(df, segment_row_size)
```
alexowens90, line 197: This logic is wrong. For dynamic schema, it only makes sense to plot the row fragmentation, not the total fragmentation.
With static schema we should plot both.

alexowens90 (changes requested):

python/arcticdb/diagnostics_engine/library_options.py
```diff
@@ -0,0 +1,309 @@
...
+            "job_name",
+            "prometheus_env",
+            "prometheus_model",
+            "event_type",
+            "host",
+            "port",
+            "allow_name_lookup",
+            "file_path",
+            "storage_fallthrough",
+            "enabled",
```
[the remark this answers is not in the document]
GemDot: done the changes

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
GemDot: As discussed, we will not calculate/show the total fragmentation for dynamic Schema. It is more sensible/easy to understand to show only the per row fragmentation. Done the changes.

python/arcticdb/diagnostics_engine/fragmentation_diagnosis.py
GemDot: Idem. Done the changes.

python/arcticdb/diagnostics_engine/library_options.py
```diff
@@ -0,0 +1,324 @@
...
+        default_response_obj = type(obj)()
+        for field in obj.DESCRIPTOR.fields:
+            value = getattr(obj, field.name)
+            full_name = path + "." + field.name if path else field.name
+            if field.message_type:
+                self.traverse_descriptor(value, full_name)
+            elif full_name not in self.exclusion_set:
+                default_value = (
+                    getattr(default_response_obj, field.name)
+                    if (not self.is_internal_api or field.name not in self.internal_default_values)
```
alexowens90, line 274: For the external API, we should be displaying everything in terms of the `LibraryOptions` class, not the entire protobuf:
https://docs.arcticdb.io/api/arcticdb/arcticdb.LibraryOptions

alexowens90 (changes requested):

python/arcticdb/diagnostics_engine/library_options.py
GemDot: Ok done. Given that the library options for the external and internal APIs are exposed and modified by the user in different ways, I have completely separated the functions that retrieve the library options for the two APIs. This separation will help avoid unnecessary and messy code. 

Basically, for the external API case, I create a default LibraryOptions object and then traverse all the variables (i.e., options). These are the only ones with which the user can interact. Subsequently, I need to traverse the protobuf object of the library starting from the lib._lib_cfg.lib_desc.version level, in order to find the corresponding value. Please note that not all user options necessarily fall under write_options. For instance, encoding_version is one level up, and there could be future options that we may decide to add.

PR #841, opened 2023-09-07
small documentation improvements

joe-iddon:

Just some small documentation corrections to get used to the PR process (@poodlewars suggested I trial the "getting started experience" and suggest a few improvements).

- Clarified that the same dataframe is used throughout the guide.
- Reworded a sentence to read a bit easier.
- The docs claimed that the `update` command would "insert a duplicate entry for 2000-01-01 07:00". When following this it overrides what was there, so I've changed this to read "modify".
- The starting date for the additional data in the `append` example is the same as the end date of what was there, but the comment says it is "after" which is confusing. The docs make it clear the expected behaviour is "Append only accepts data for which the index of the first row is equal to or greater than the index of the last row in the existing data."
- Spotted a couple of `docs.arcticdb.io` typos, so modified `library.py`.

Edit: See comments below (updated the append example to make it more straightforward; and added a unit test for re-generating the outputs as needed).

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [x ] Have you updated the relevant docstrings and documentation?
</details>


willdealtry (approved):

docs/mkdocs/docs/index.md
```diff
@@ -339,7 +341,7 @@ VersionedItem(symbol=test_frame,library=data,data=n/a,version=2,metadata=None,ho
 2000-01-02 09:00:00      2      1     20     47     47     16     14     48  ...
 ```
 
-The final 7 rows included the 5 rows we have just appended and the last two rows that were written previously. 
+The final 7 rows consist of the last two rows written previously followed by the 5 rows that we have just appended.
```
poodlewars, line 344: Could you check that this code example is really how ArcticDB works? Probably adding a unit test for it is good idea if one doesn't already exist.

I think this example, with the duplicate in the index, is a bit confusing and might have been a typo. I think our first append example should do what it originally and falsely promised to do:

```
 Note the starting date of this DataFrame is after the final row written previously!
```

poodlewars (changes requested):

docs/mkdocs/docs/index.md
joe-iddon: I've changed the docs example to append data that starts after the final row written previously, and also added a unit test to verify this. The unit test also prints out the outputs for the other examples so that it is easier if somebody in the future wants to modify something and re-generate all the outputs.

poodlewars (approved):

Issue #842, opened 2023-09-07
Fail to connect to S3-backed storage if AWS authorisation is required


phoebusm:
### Describe the bug

ArcticDB fails to connect to S3-backed storage if AWS authorisation is required. (For storage's connection string which `aws_auth=true` exists)

### Steps/Code to Reproduce

```
uri = "s3://localhost:test_bucket_0?aws_auth=true&port=4233"
expected = pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]})
sym = "test"
ac = Arctic(uri)
ac[LIB_NAME].write(sym, expected)

ac = Arctic(uri)
assert_frame_equal(expected, ac[LIB_NAME].read(sym).data)
```

### Expected Results

No error

### OS, Python Version and ArcticDB Version

All versions affected

### Backend storage used

AWS S3

phoebusm:
The issue can only repeated if the library configuration is loaded from storage

PR #843, opened 2023-09-07
Fix aws auth not used, even the option is set to true

phoebusm:
#### Reference Issues/PRs

https://github.com/man-group/ArcticDB/issues/842

It fixes the connection failure to AWS S3 storage which requires AWS authorisation. 
The fix is primarily setting the access and key of affected libraries ' storage overrider to `_RBAC_`. These will make sure the config loaded from storage will have the access and key properly set, to trigger the special handling for AWS authorisation.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/scripts/test_update_storage.py
+++ b/python/tests/scripts/test_update_storage.py
@@ -1,6 +1,8 @@
 import sys
 
 import pytest
+import pandas as pd
+from pandas.util.testing import assert_frame_equal
 
 from arcticdb import Arctic
 from arcticdb.scripts.update_storage import run
@@ -109,6 +111,14 @@ def test_upgrade_script_s3_rbac_ok(moto_s3_endpoint_and_credentials, monkeypatch
     assert s3_storage.bucket_name == bucket
     assert s3_storage.credential_name == USE_AWS_CRED_PROVIDERS_TOKEN
     assert s3_storage.credential_key == USE_AWS_CRED_PROVIDERS_TOKEN
+
+    expected = pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]})
+    sym = "test"
+    ac = Arctic(uri)
+    ac[LIB_NAME].write(sym, expected)
+
+    ac = Arctic(uri)  # Force load lib config from storage, to check "storage-overridened" config
+    assert_frame_equal(expected, ac[LIB_NAME].read(sym).data)
 
 
 # Side effect needed from "_create_container" fixture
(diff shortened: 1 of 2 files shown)


phoebusm:
The commit(s) in this PR will be cherry-picked to branch `v300` so it will be included in the `v3.0.0` release.

python/tests/scripts/test_update_storage.py
```diff
@@ -110,6 +112,14 @@ def test_upgrade_script_s3_rbac_ok(moto_s3_endpoint_and_credentials, monkeypatch
     assert s3_storage.credential_name == USE_AWS_CRED_PROVIDERS_TOKEN
     assert s3_storage.credential_key == USE_AWS_CRED_PROVIDERS_TOKEN
 
+    expected = pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]})
```
mehertz, line 115: Testing that we can create and access a library without `AWS_ACCESS_KEY_ID` and `AWS_ACCESS_KEY_ID` being explicitly given feels like it should be asserted in a dedicated test, rather than a test that asserts correctness of a script. 

Should we move this assertion to a simple, dedicated test?

phoebusm: Agreed. I was torn between whether putting the new lines in an existing test or having a new test earlier. 
Fix pushed

python/tests/scripts/test_update_storage.py
```diff
@@ -111,6 +120,19 @@ def test_upgrade_script_s3_rbac_ok(moto_s3_endpoint_and_credentials, monkeypatch
     assert s3_storage.credential_key == USE_AWS_CRED_PROVIDERS_TOKEN
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Test fixture issue with creds")
```
mehertz, line 123: Why are we skipping windows here 😢 

If this doesn't work on Windows it doesn't inspire too much confidence in the correctness of the Windows build?

If this is an existing test fixture issue we don't need to fix this in this PR, but these tests really should be running on Windows as well at some point.

mehertz (approved):

python/tests/scripts/test_update_storage.py
phoebusm: It's copied from the previous test so I thought the skip is already reviewed and approved....

phoebusm: Will do further test on Windows

Issue #849, opened 2023-09-11
Connection to Azure blob storage fails if SAS token is used


phoebusm:
### Describe the bug

Connection made to Azure blob storage fails if the connection strings uses SAS token for authorisation.
If access key authorisation is used, no problem arises.

### Steps/Code to Reproduce

ac = Arctic("azure://DefaultEndpointsProtocol=https;EndpointSuffix=core.windows.net;AccountName=arcticdbldctest;AccountKey=6TbE/isHYFHc6neQoGQ79it+1MLRl5iaO+1X2JKfG3i38F+YG1WwpeBkI67pm7/cRgNat/ZSX398+AStcEo8eA==;BlobEndpoint=https://arcticdbldctest.blob.core.windows.net/;FileEndpoint=https://arcticdbldctest.file.core.windows.net/;QueueEndpoint=https://arcticdbldctest.queue.core.windows.net/;TableEndpoint=https://arcticdbldctest.table.core.windows.net/;Container=dataset1;CA_cert_path=")
expected = pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]})
LIB_NAME = "testabc"
sym="abc"
ac.create_library(LIB_NAME) #create_library gives no error
ac[LIB_NAME].write(sym, expected)

### Expected Results

Azure status code:0 
Azure reason phrase is empty

By setting environment variable "AZURE_LOG_LEVEL=1", Azure reports `Problem with the SSL CA cert (path? access rights?)

### OS, Python Version and ArcticDB Version

Windows

### Backend storage used

Azure

PR #851, opened 2023-09-11
Make azure sdk stick to winhttp if possible

phoebusm:
#### Reference Issues/PRs
https://github.com/man-group/ArcticDB/issues/849

Azure SDK has an undocumented feature, in the situation which if CA path is set, even it is left empty, uses libcurl instead of the default winhttp as backend support on Windows. During the test, using libcurl as backend shows buggy behaviour on ssl ca cert verification. 
This fix is to not set `CAINFO` if ca path is not specified, to give user choice to use winhttp, to avoid the buggy behaviour.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/azure/azure_storage.cpp
+++ b/cpp/arcticdb/storage/azure/azure_storage.cpp
@@ -41,10 +41,12 @@ AzureStorage::AzureStorage(const LibraryPath &library_path, OpenMode mode, const
 }
 
 Azure::Storage::Blobs::BlobClientOptions AzureStorage::get_client_options(const Config &conf) {
-    Azure::Core::Http::CurlTransportOptions curl_transport_options;
-    curl_transport_options.CAInfo = conf.ca_cert_path();
     BlobClientOptions client_options;
-    client_options.Transport.Transport = std::make_shared<Azure::Core::Http::CurlTransport>(curl_transport_options);
+    if (!conf.ca_cert_path().empty()) {//WARNING: Setting ca_cert_path will force Azure sdk uses libcurl as backend support, instead of winhttp
+        Azure::Core::Http::CurlTransportOptions curl_transport_options;
+        curl_transport_options.CAInfo = conf.ca_cert_path();
+        client_options.Transport.Transport = std::make_shared<Azure::Core::Http::CurlTransport>(curl_transport_options);
+    }
     return client_options;
 }
 
--- a/python/arcticdb/adapters/azure_library_adapter.py
+++ b/python/arcticdb/adapters/azure_library_adapter.py
@@ -8,6 +8,7 @@ As of the Change Date specified in that file, in accordance with the Business So
 import re
 import time
 from typing import Optional
+import platform
 
 from arcticdb.options import LibraryOptions
 from arcticc.pb2.storage_pb2 import EnvironmentConfigsMap, LibraryConfig
@@ -54,6 +55,8 @@ class AzureLibraryAdapter(ArcticLibraryAdapter):
             ]
         )
         self._container = self._query_params.Container
+        if platform.system() == "Windows" and self._query_params.CA_cert_path:
+            raise ValueError(f"CA_cert_path cannot be set on Windows platform")
         self._ca_cert_path = self._query_params.CA_cert_path
         self._encoding_version = encoding_version
 
--- a/python/arcticdb/arctic.py
+++ b/python/arcticdb/arctic.py
@@ -90,7 +90,7 @@ class Arctic:
                 +---------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------+
                 | Path_prefix               | Path within Azure container to use for data storage                                                                                                           |
                 +---------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------+
-                | CA_cert_path              | Azure CA certificate path. If not set, default path will be used.                                                                                             |
+                | CA_cert_path              | (Non-Windows platform only) Azure CA certificate path. If not set, default path will be used.                                                                 |
                 |                           | Note: For Linux distribution, default path is set to `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`.                                                     |
                 |                           | If the certificate cannot be found in the provided path, an Azure exception with no meaningful error code will be thrown.                                     |
                 |                           | For more details, please see https://github.com/Azure/azure-sdk-for-cpp/issues/4738.                                                                          |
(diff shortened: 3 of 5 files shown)


cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -42,7 +42,8 @@ AzureStorage::AzureStorage(const LibraryPath &library_path, OpenMode mode, const
 
 Azure::Storage::Blobs::BlobClientOptions AzureStorage::get_client_options(const Config &conf) {
```
qc00, line 43: You have to do it like [this](https://github.com/man-group/ArcticDB/pull/776/files#diff-b6fda20fb3213402c5c12984968f4f35baeaf64014bb25c402330bb4dbc9261aL43).

What makes the difference is not to make the assignment on line 48, not setting the CA path.

phoebusm: Resolved

qc00 (approved):

phoebusm:
Pending https://github.com/man-group/ArcticDB/pull/853 for Azure support on conda build test

python/arcticdb/arctic.py
```diff
@@ -103,6 +103,8 @@ def __init__(self, uri: str, encoding_version: EncodingVersion = DEFAULT_ENCODIN
                 |                           | "/etc/pki/tls/cacert.pem"                             OpenELEC                                                                                                |
                 |                           | "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"   CentOS/RHEL 7                                                                                           |
                 |                           | "/etc/ssl/cert.pem"                                   Alpine Linux                                                                                            |
+                |                           | WARNING for WINDOWS USER: Not leaving this empty will switch the backend support of Azure SDK from winhttp to libcurl. If ca cert path is needed to be        |
```
poodlewars, line 106: This warning is a bit ungrammatical, and this docstring is getting massive!

But it makes me think, is it ever correct for a Windows user to set `CA_cert_path`? Shouldn't we leave it to Windows to determine (it doesn't have the Linux "many locations" problem). Should we throw if a Windows user sets this to a non-empty value (and document that)?

I suggest throw if non-empty rather than throw if present so that users can still use the same code across Linux/Windows, just with a bit of config for the cert path location that differs by OS.

python/tests/conftest.py
```diff
@@ -103,6 +103,23 @@ def azure_client_and_create_container(azurite_container, azurite_azure_uri):
     return client
 
 
+@pytest.fixture(scope="function")
```
poodlewars, line 106: `function` is the default scope, FWIW

python/tests/conftest.py
```diff
@@ -103,6 +103,23 @@ def azure_client_and_create_container(azurite_container, azurite_azure_uri):
...
+def azure_account_sas_token(azure_client_and_create_container, azurite_azure_test_connection_setting):
+    start_time = datetime.now(timezone.utc)
+    expiry_time = start_time + timedelta(days=1)
+    _, _, credential_name, credential_key, _ = azurite_azure_test_connection_setting
+
+    sas_token = generate_account_sas(
+        account_key=credential_key,
+        account_name=credential_name,
+        resource_types=ResourceTypes.from_string("sco"),
+        permission=AccountSasPermissions.from_string("rwdlacup"),
```
poodlewars, line 116: What do these special strings mean? `sco`, `rwdlacup`? Worth a comment

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -1050,12 +1050,27 @@ def test_get_uri(object_storage_uri_incl_bucket):
...
+    endpoint, container, credential_name, credential_key, _ = azurite_azure_test_connection_setting
+    Arctic(
         f"azure://DefaultEndpointsProtocol=http;AccountName={credential_name};AccountKey={credential_key};BlobEndpoint={endpoint}/{credential_name};Container={container}"
     )
 
 
+def test_azure_sas_token(azure_account_sas_token, azurite_azure_test_connection_setting):
+    endpoint, container, credential_name, _, _ = azurite_azure_test_connection_setting
+    ac = Arctic(
+        f"azure://DefaultEndpointsProtocol=http;SharedAccessSignature={azure_account_sas_token};BlobEndpoint={endpoint}/{credential_name};Container={container}"
```
poodlewars, line 1062: We should document how to do this in the website

poodlewars (changes requested):

python/arcticdb/arctic.py
phoebusm: One more, docstring will be massively reduced once the ca path and directory can be automatically assigned. The commit is coming soon.

python/tests/conftest.py
phoebusm: That is a miss. Will do.

python/tests/integration/arcticdb/test_arctic.py
poodlewars: Yeah exactly, a little example in the docs showing how to connect to Azure with SAS

python/arcticdb/arctic.py
poodlewars: OK, fair enough. Can I suggest we log a warning on Windows when this is set, then?

What is your comment regarding 

`using libcurl as backend shows buggy behaviour on ssl ca cert verification.` about? What was the buggy behaviour?

It would be good to explain the impact of the change of backend a bit in the log warning.

poodlewars: Text suggestion for this docstring if we keep it:

```
WARNING for Windows users: Setting this to anything besides an empty string will switch the backend support of Azure SDK from winhttp to libcurl. <EXPLANATION OF IMPACT>

If CA Cert path needs to be specified, set it in the Windows settings rather than using the option, so that winhttp will be used.
```

poodlewars: Could it be that Azure switches to libcurl backend here because it assumes that if `CA_Cert_Path` is set then it can't be on Windows, or something? It would be good to see if there is anything in the Azure git blame to understand why this happens at all.

phoebusm: > OK, fair enough. Can I suggest we log a warning on Windows when this is set, then?
> 
> What is your comment regarding
> 
> `using libcurl as backend shows buggy behaviour on ssl ca cert verification.` about? What was the buggy behaviour?
> 
> It would be good to explain the impact of the change of backend a bit in the log warning.

Verification fails if SAS Token is used with libcurl on Windows. The problem is gone once switched to winhttp.
I use the word `buggy` as the fail is weird, e.g. `list_libraries()` is fine but `read` gives error.
And probably it will be good idea to give user a warning regarding it.

python/tests/integration/arcticdb/test_arctic.py
phoebusm: Maybe the best is by adding the link of below official guidelines for user reference?
Create token:
https://learn.microsoft.com/en-us/azure/storage/common/storage-account-sas-create-python
Connection string:
https://learn.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string#create-a-connection-string-using-a-shared-access-signature

python/arcticdb/arctic.py
phoebusm: After discussion with @poodlewars, exception will be thrown if `CA_cert_path` or `CA_cert_dir` is being set. 
Further compacting of the docstring is handled in https://github.com/man-group/ArcticDB/pull/867

poodlewars (approved):

python/arcticdb/adapters/azure_library_adapter.py
```diff
@@ -54,6 +55,8 @@ def __init__(self, uri: str, encoding_version: EncodingVersion, *args, **kwargs)
             ]
         )
         self._container = self._query_params.Container
+        if platform.system() is "Windows" and self._query_params.CA_cert_path:
```
qc00, line 58: Use `==` for safety

PR #865, opened 2023-09-17
Add asv benchmarks

G-D-Petrov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


benchmarks/benchmarks.py
```diff
@@ -0,0 +1,101 @@
...
+import numpy as np
+
+def generate_pseudo_random_dataframe(n, freq="S", end_timestamp="1/1/2023"):
+    """
+    Generates a Data Frame with 2 columns (timestamp and value) and N rows
+    - timestamp contains timestamps with a given frequency that end at end_timestamp
+    - value contains random floats that sum up to approximately N, for easier testing/verifying
+    """
+    # Generate random values such that their sum is equal to N
+    values = np.random.dirichlet(np.ones(n), size=1) * n
```
poodlewars, line 19: Shall we save the team from reading about probability theory and do this instead?

```suggestion
    values = np.random.uniform(0, 2, size=n)
```

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,101 @@
...
+)
+
+import pandas as pd
+import numpy as np
+
+def generate_pseudo_random_dataframe(n, freq="S", end_timestamp="1/1/2023"):
+    """
+    Generates a Data Frame with 2 columns (timestamp and value) and N rows
+    - timestamp contains timestamps with a given frequency that end at end_timestamp
+    - value contains random floats that sum up to approximately N, for easier testing/verifying
```
poodlewars, line 16: We could write deterministic tests by seeding `numpy.random` with a raodom seed and recording the seed as metadata with the dataframe. Then the verification step could reconstruct the expected result from the seed in the metadata.

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    # Generate random values such that their sum is equal to N
+    values = np.random.dirichlet(np.ones(n), size=1) * n
+    # Generate timestamps
+    timestamps = pd.date_range(end=end_timestamp, periods=n, freq="S")
+    # Create dataframe
+    df = pd.DataFrame({"timestamp": timestamps, "value": values[0]})
+    return df
+
+def get_prewritten_lib_name(rows):
```
poodlewars, line 26: Inline?

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,101 @@
...
+    return f"prewritten_{rows}"
+
+class TimeSuite:
+    """
+    An example benchmark that times the performance of various kinds
```
poodlewars, line 31: Fix docstring

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,101 @@
...
+    of iterating over dictionaries in Python.
+    """
+    number = 5
+    params = ([1_000, 100_000], [500, 1000])
+    param_names = ['rows', 'num_symbols']
```
poodlewars, line 36: What do you think about explicitly specifying the `version` here? It's a maintenance burden, but by default the results will get invalidated each time this file changes. So worried we might miss regressions if `version` is left as `None`.

benchmarks/benchmarks.py
poodlewars, line 1: Great! After this PR, it would be good to also cover more advanced features, such as the below.

- Dynamic schema
- Deletes / appends / updates
- Staged data
- list_symbols (both with and without symbol list cache populated) , has_symbol
- list_versions
- QueryBuilder (at least a subset of QueryBuilder API)
- Data slicing (columns, date_range)
- Sparse columns

There are some good notes about the various features in https://github.com/man-group/ArcticDB/blob/master/docs/mkdocs/docs/technical/contributing.md

cpp/arcticdb/pipeline/python_output_frame.cpp
```diff
@@ -50,7 +50,7 @@ PythonOutputFrame::~PythonOutputFrame() {
             });
         }
     }
-    Allocator::instance()->trim();
+    // Allocator::instance()->trim();
```
poodlewars, line 53: Do not merge in this PR

hashes_to_benchmark.txt
poodlewars, line 1: What's this for? Can't find references to it in ASV docs

python/arcticdb/version_store/helper.py
```diff
@@ -272,6 +272,12 @@ def add_s3_library_to_env(
     use_virtual_addressing=False,
 ):
     env = cfg.env_by_id[env_name]
+    if with_prefix and isinstance(with_prefix, str) and (with_prefix.endswith("/") or "//" in with_prefix):
```
poodlewars, line 275: Do not merge in this PR

poodlewars:
Thanks for this. Looks good. Think I've gone through all the bits that aren't in your other PRs.

What's your plan for how we'll handle the results from the benchmarks? How will the team find out about regressions, how will the data be portrayed? Will it be easy to pin down the commit that caused a regression?

How can we make sure we run the analysis workflow before making releases?

.asv/results/LONVG6269/machine.json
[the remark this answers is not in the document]
G-D-Petrov: I general we don't need it, I just wasn't sure if we would need this in the build.

benchmarks/benchmarks.py
G-D-Petrov: yeah that makes more sense

hashes_to_benchmark.txt
G-D-Petrov: You can specify commit hashes to run over/benchmark - https://asv.readthedocs.io/en/stable/commands.html?highlight=hashes#asv-run

PR #870, opened 2023-09-18
Fix performance degradation in read_batch

G-D-Petrov:
#### Reference Issues/PRs
Fixes issue 869

Fixes performance degradation in read_batch when called with many symbols.

#### Any other comments?
Here is a table of the performance **before** the fix:

  | 1000 | 5000 | 10000 
-- | -- | -- | --
Duration | 4.62| 47.73 | 184.35
Avg Time per Symbol | 0.0046 | 0.00954 | 0.018435

And here is the performance **after** the fix:
  | 1000 | 5000 | 10000 
-- | -- | -- | --
Duration | 3.56 | 19.0 | 42.59
Avg Time per Symbol | 0.00356 | 0.0038 | 0.0042

As you can see, the performance per symbol is more-or-less linear and over 10000 symbols we have about 4-4.5 times speedup.
The speedup is more pronounced for higher segment counts (e.g. more rows for each symbol) so for bigger data frames the speedup can be as much as **10 times**.

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -956,6 +956,7 @@ class NativeVersionStore:
             all(v is not None for v in versioned_items),
             "Null value from _batch_read_to_versioned_items. NoDataFoundException should have been thrown instead.",
         )
+        self.version_store.trim()
         return {v.symbol: v for v in versioned_items}
 
     def _batch_read_to_versioned_items(
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -900,6 +900,13 @@ class NativeVersionStore:
             vitem = self._adapt_read_res(read_result)
             yield vitem
 
+    def trim(self) -> None:
+        """
+        Calls trim on the allocator of the underlying version_store
+        Should be called after gc.collect() to remove any reference cycles
+        """
+        self.version_store.trim()
+
     def batch_read(
         self,
         symbols: List[str],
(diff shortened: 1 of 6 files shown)


python/arcticdb/version_store/_store.py
```diff
@@ -956,6 +956,7 @@ def batch_read(
             all(v is not None for v in versioned_items),
             "Null value from _batch_read_to_versioned_items. NoDataFoundException should have been thrown instead.",
         )
+        self.version_store.trim()
```
willdealtry, line 959: I don't know how much difference trim() will make here, as the user will still be holding references to the allocated results. We should probably remove it here and add something to the documentation telling them that they can call trim themselves once they've released all references to the returned dataframes and done a gc.collect() to remove any reference cycles

poodlewars: I'm a bit confused what benefit the malloc_trim is giving us here. On your write up on the ticket @G-D-Petrov you write,

```
Removing malloc_trim doesn't seem to have affected the memory usage.
Even on my stress test that went from 1000 up to 60000 symbols, in a continuous loop, there were no noticeable memory issues.
```

Shall we just get rid of it completely?

I'm a bit hesistant to put such complicated semantics: 

```
they can call trim themselves once they've released all references to the returned dataframes and done a gc.collect() to remove any reference cycles
```

on the public API (and commit to supporting them). Especially considering our target users are statisticians as much as developers.

willdealtry (approved):

python/arcticdb/version_store/_store.py
poodlewars: William explained the history lesson to me about why the trim is there.

Still keen not to put it on the external API `Library`. Makes it hard to change our minds about it.

How about we keep it on the `NativeVersionStore` only for now? Then `Library` users can still use it in an emergency via the `_nvs` handle.

G-D-Petrov: Okay I will remove the method from the Library API.
Do we want to keep the documentation that I added about this?
And point people that they can use the _nvs handle?

poodlewars (approved):
Please squash the commits before merging

Issue #850, opened 2023-09-11
LMDB libraries are unreadable once moved


qc00:
### Describe the bug

The full path is stored in the library config protobuf

### Steps/Code to Reproduce

Create a library in a new folder using LMDB -> move/rename containing folder -> library not readable

### Expected Results

The library is readable.

### OS, Python Version and ArcticDB Version

All

### Backend storage used

LMDB

poodlewars:
This was raised by a user in Slack https://arcticdb.slack.com/archives/C047WF34V7E/p1694449570998269 . We should let him know when this is fixed.

PR #877, opened 2023-09-20
Issue #850 Ensure that LMDB libraries are readable after being moved to a different location

poodlewars:
#### Reference Issues/PRs

See issue #850 .

Ensure that after moving or copying an LMDB database holding arctic data, then the data is still readable when read from the _new_ location.

Previously we were persisting the LMDB path for the Arctic instance used to _create_ the library in a storage protobuf, and reading the path from that protobuf when accessing the library in a subsequent session. Instead now we just do not persist the path at all in the protobuf, and always use the path that the current Arctic instance was created with.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/adapters/lmdb_library_adapter.py
+++ b/python/arcticdb/adapters/lmdb_library_adapter.py
@@ -15,7 +15,6 @@ from arcticdb.log import storage as log
 
 from arcticdb.options import LibraryOptions
 from arcticc.pb2.storage_pb2 import EnvironmentConfigsMap, LibraryConfig
-from arcticc.pb2.lmdb_storage_pb2 import Config as LmdbConfig
 from arcticdb.version_store.helper import add_lmdb_library_to_env
 from arcticdb.config import _DEFAULT_ENV
 from arcticdb.version_store._store import NativeVersionStore
@@ -26,7 +25,7 @@ from arcticdb.exceptions import ArcticDbNotYetImplemented, LmdbOptionsError
 
 
 def _rmtree_errorhandler(func, path, exc_info):
-    log.warn("Error removing LMDB tree at path=[{}]", path, exc_info=exc_info)
+    log.warn("Error removing LMDB tree at path=[{}] error=[{}]", path, exc_info)
 
 
 @dataclass
@@ -162,14 +161,12 @@ class LMDBLibraryAdapter(ArcticLibraryAdapter):
 
         return lib
 
-    def cleanup_library(self, library_name: str, library_config: LibraryConfig):
-        for k, v in library_config.storage_by_id.items():
-            lmdb_config = LmdbConfig()
-            v.config.Unpack(lmdb_config)
-            shutil.rmtree(os.path.join(lmdb_config.path, library_name), onerror=_rmtree_errorhandler)
+    def cleanup_library(self, library_name: str):
+        shutil.rmtree(os.path.join(self._path, library_name), onerror=_rmtree_errorhandler)
 
     def get_storage_override(self) -> StorageOverride:
         lmdb_override = LmdbOverride()
+        lmdb_override.path = self._path
         if self._query_params.map_size:
             lmdb_override.map_size = self._query_params.map_size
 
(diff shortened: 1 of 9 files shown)


python/arcticdb/adapters/lmdb_library_adapter.py
```diff
@@ -162,14 +161,12 @@ def create_library(self, name, library_options: LibraryOptions):
 
         return lib
 
-    def cleanup_library(self, library_name: str, library_config: LibraryConfig):
-        for k, v in library_config.storage_by_id.items():
-            lmdb_config = LmdbConfig()
-            v.config.Unpack(lmdb_config)
-            shutil.rmtree(os.path.join(lmdb_config.path, library_name), onerror=_rmtree_errorhandler)
+    def cleanup_library(self, library_name: str):
```
jjerphan, line 164: Side note: this PR makes a breaking change to the public API (like here with removing a parameter to this function), so according to SemVer, this should _stricto sensus_ be part of 5.0.

I am not sure that we want to follow semantic versioning at all cost. Henry Schreiner mentions [realistic semantic versioning](https://iscinumpy.dev/post/bound-version-constraints/#semver). Might this be relevant for ArcticDB?

python/arcticdb/adapters/lmdb_library_adapter.py
```diff
@@ -162,14 +161,12 @@ def create_library(self, name, library_options: LibraryOptions):
...
+        shutil.rmtree(os.path.join(self._path, library_name), onerror=_rmtree_errorhandler)
```
jjerphan, line 165: Naive question: should we add validation on the value of `self._path` to make sure this can't be used to delete some folders on the filesystem?

python/tests/integration/arcticdb/test_arctic_move_storage.py
```diff
@@ -0,0 +1,199 @@
...
+    # When - we move the data
+    new_container = f"{azurite_container}-new"
+    client = azure_client_and_create_container
+    client.create_container(name=new_container)
+
+    container_client = client.get_container_client(container=azurite_container)
+    new_container_client = client.get_container_client(container=new_container)
+
+    for b in container_client.list_blobs():
+        print(b)
```
jjerphan, line 185: Was this used for debugging?
```suggestion
```

G-D-Petrov (approved):

python/tests/integration/arcticdb/test_arctic_move_storage.py
jjerphan, line 1: Is a non-regression test needed for #850?

jjerphan (approved):
LGTM. Thank you, @poodlewars!

python/tests/integration/arcticdb/test_arctic_move_storage.py
poodlewars: Yes, but this file includes such tests, no?

python/arcticdb/adapters/lmdb_library_adapter.py
poodlewars: I think we're regarding Arctic, Library and QueryBuilder as our public API. Those are the only documented parts of the API. https://docs.arcticdb.io/api/

python/arcticdb/adapters/lmdb_library_adapter.py
poodlewars: That is a good point isn't it, we could have logic like:

- Look at `self._path / library_name`
- Does it contain .mdb files?
** No? Stop.
- Delete the .mdb files
- Is the dir now empty?
** Yes? Delete the dir.
** No? Stop.

poodlewars: Raised #906 to do that, to avoid scope creep in this PR.

python/tests/integration/arcticdb/test_arctic_move_storage.py
jjerphan: I should have been clearer: `python/tests/nonreg` seem to contain non-regression tests. Must some of them be present there?

poodlewars: Ah I see. I think this should be OK as is. I'm not actually sure what the intent of `nonreg` is now. It used to be that we ran tests in there were to check backwards compat with old data, but they didn't survive the open sourcing.

python/npm-debug.log
poodlewars, line 1: REMOVE

PR #860, opened 2023-09-14
Memory backed api

joe-iddon:
#### What does this implement/fix? How does it work (high level)? Highlight notable design decisions.

Implements a memory-backed API for the top-level `Arctic()` instance.

Intended usage:

```python
>>> import pandas as pd
>>> ac = Arctic("mem://memory_db") # this name is passed through to the C++ storage, but ignored there
>>> ac.create_library("lib")
...
```

Test definition files which now also verify `mem://`:
- `test_arctic_batch.py` - added via extending `arctic_library`/`arctic_client`
- `test_arctic.py` - added again via `arctic_library`/`arctic_client`
- `test_basic_version_store.py` - parameterised all `lmdb_version_store` fixtures to test the memory adapter too*
- `test_stress_generated.py` - parameterised `lmdb_version_store_tiny_segment` to test the memory adapter too.

\* I did not extend the more specific test fixtures like `lmdb_version_store_tombstone_and_sync_passive`.

These add an extra 15 seconds on top of times around 3 mins for the large test files.

#### Any other comments?

The timings for the stress test (`test_stress_small_row` in `test_stress_generated.py`) are:

Adapter type | Write / sec | Read / sec
--- | --- | ---
lmdb | 97 | 5
mem | 11 | 10

The test is for 1MB of data. The results for the high entropy and low entropy times were within a few seconds of each other. (High entropy has completely random data, low entropy has blocks of repeated rows: aaacccbbb.)

At the moment the name given is unused.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [x ] Have you updated the relevant docstrings and documentation?
 - [x ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [x ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [x ] Are API changes highlighted in the PR description?
 - [x ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #895, opened 2023-09-22
Docs for memory-backed API.

joe-iddon:
**Note**: This PR is branched off the memory_backed_API branch, so already contains those changes which are not on master.

- Documents #860
- Wrote a guide on the use of LMDB and in-memory backends. Explores concurrent use of LMDB, and speed comparisons if used on tmpfs vs disk. Note that I could not get `fio` to give reliable hardware figures so defaulted on timing `np.save` and `np.write` operations. I have documented how far I got with `fio` and we can return to this in the future if needed.
- Fixes some of the sphinx docs -- mainly replacing single backticks with double backticks (single were creating `<cite>` tags, rather than `<code>` and therefore appearing as italics. [See the docs](https://docutils.sourceforge.io/docs/user/rst/quickref.html)).
- Fixed the main `arcticdb.html` API docs page by changing from headers with `-----` to bold with `**LMDB**` etc in the `python/arcticdb.py` docstring. (The parameters will now show up properly, rather than getting lots of warnings for incorrect placement of titles.*)

*To see what I mean, look at this run: https://github.com/man-group/ArcticDB/actions/runs/6138089677/job/16654614393 under sphinx there are lots of `CRITICAL: Unexpected section title.` errors. I've tested locally and now it compiles with no errors and looks better.

#### Any other comments?

Everyone will have the stubs in `/docs/sphinxdocs/source/arcticdb/` and `/docs/sphinxdocs/source/library/` in their local repos so pulling from master should hopefully be fine at deleting them but we should be careful nobody re-commits them.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [x ] Have you updated the relevant docstrings and documentation?
</details>


docs/mkdocs/docs/tutorials/lmdb_and_in_memory.md
```diff
@@ -0,0 +1,273 @@
...
+
+```py
+ac_mem = Arctic('mem://')
+```
+
+The storage returned is owned by the `ac_mem` instance and lives for the lifetime of this Python object. Behind the
+scenes, a [Folly Concurrent Hash Map](https://github.com/facebook/folly/blob/main/folly/concurrency/ConcurrentHashMap.h)
+is used as the key/value store. For test cases and experimentation, the in-memory backend is a good option.
+
+On the other hand, if the datasets are too large for RAM, or multiple processes want concurrent access to the same data
```
poodlewars, line 38:
```suggestion
If multiple processes want concurrent access to the same data, then we recommend using LMDB over a tmpfs.
```

docs/mkdocs/docs/tutorials/lmdb_and_in_memory.md
```diff
@@ -0,0 +1,273 @@
...
+then LMDB is recommended.
+
+## How to handle concurrent writers?
```
poodlewars, line 41: I think we should focus this page on using ArcticDB with in-memory backends. Hence below we should be showing how to set up a tmpfs and run ArcticDB over it, and demonstrating that concurrent access works. I think we can skip the section describing LMDB over an on-disk filesystem.

docs/mkdocs/docs/tutorials/lmdb_and_in_memory.md
```diff
@@ -0,0 +1,273 @@
...
+
+Again, it should be notes that ArcticDB achieves its highest performance and scale when configured with an object
```
poodlewars, line 43:
```suggestion
Again, it should be noted that ArcticDB achieves its highest performance and scale when configured with an object
```

docs/sphinxdocs/source/arcticdb/arcticdb.Arctic.rst
poodlewars, line 1: It would be good to add lines in `.gitignore` so that these files don't get mistakenly checked in again.

poodlewars (changes requested):

docs/sphinxdocs/source/conf.py
```diff
@@ -48,7 +48,7 @@
 
 autodoc_default_flags = ["members"]
 autodoc_member_order = "bysource"
-autosummary_generate = True
+autosummary_generate = False
```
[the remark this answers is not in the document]
joe-iddon: Thanks for getting me to double-check. There actually were side effects which I hadn't thoroughly investigated. These stub files do need to be generated so that sphinx explores secondary links (e.g. it would still generate the top-level summaries of `Arctic` etc, but not to the lower level ones without the stubs).

Solution was to set this back to `True`, and to add all the stubs to `.gitignore`.

docs/mkdocs/docs/tutorials/lmdb_and_in_memory.md
```diff
@@ -0,0 +1,273 @@
...
+ephemeral LMDB backend. We can also see that the in-memory store is significantly faster across the board as would
+be expected.
+
+![](/images/LMDBtmpfsDiskMemSpeeds.png)
+
+Note: the ranges are 95% confidence intervals based on five repeats for each data size. The hardware limits are for
+writing to disk (not tmpfs) using the `np.save` and `np.load` functions. No data was sought for the RAM's hardware
+limit.
+
+### Attempts to measure hardware limits using `fio`
```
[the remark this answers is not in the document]
joe-iddon: Moved to Wiki page here: https://github.com/man-group/ArcticDB/wiki/Performance-test-code-and-determining-theoretical-maximum-read-write-speeds

docs/mkdocs/docs/tutorials/lmdb_and_in_memory.md
joe-iddon: I've kept the snippet for how to connect to an `lmdb` endpoint because I think this is useful for also explaining the `mem://` syntax afterwards. I've changed it to the tmpfs case though so it fits the title of the guide. For most users they will likely visit this page and leave once they see that they can use `mem://`, I'd imagine.

poodlewars (approved):

PR #897, opened 2023-09-24
Move api docs from sphinx to mkdocs

jamesmunro:
Having docs split across sphinx and mkdocs isn't the best user experience.
A key issue is that the search is split across sphinx and mkdocs.
This will also let us version the docs better (one version across all docs).


jamesmunro:
Preview of combined docs: https://f59c5f27.arcticdb-docs.pages.dev/

G-D-Petrov (approved):

docs/mkdocs/docs/api/arctic.md
joe-iddon, line 1: Just noting that we discussed to fix the nested table issue in the Arctic documentation in a future PR.

joe-iddon (approved):

.github/workflows/docs.yml
```diff
@@ -15,66 +15,40 @@ jobs:
...
-      - run: npm install @actions/artifact
-        if: ${{!inputs.api_wheel}}
-      - id: download
-        name: Fetch wheel artifacts
+  
+      - id: download-wheel-artifact
+        name: Download wheel artifact from last successful build
         if: ${{!inputs.api_wheel}}
-        uses: actions/github-script@v6
+        uses: dawidd6/action-download-artifact@v2
```
qc00, line 29: This will be confusing on a branch or tag build and may break if the checkout and the wheel from `master` have incompatible APIs.

jamesmunro: The previous code (which didn't seem to work) was trying to get the most recent artifact from any build?  In which case this is slightly more reliable.  I don't see an obvious way to default to the current branch, maybe I can detect it.

jamesmunro: Do you have a suggestion?

qc00: The previous code works differently depending on how it's triggered:

- If it's called as part of `build.yml`, it will use the wheel from the current run.
- If invoked on its own, the wheel info has to be explicitly supplied because it's not trivial to work out.

It was working the last time I checked.

But, ideally, it should work like this:

* If running as part of a code build **and not releasing**, use the wheel from the current run.
* If running as part of a release build, checkout docs from `master` (instead of this tag) and use the latest released wheel (from PyPI)
* If invoked on its own, use the wheel from the last build (doesn't have to be successful) of the same branch, falling back to the latest release on PyPI.

joe-iddon:
I added a few lines to the bottom of `.gitignore` for some sphinx artifacts. Could get rid of these now.

.github/workflows/docs.yml
jamesmunro: Thanks.  The issue with the current approach is that it has often reverted docs, given you are doing a deployment and you may be building docs older than master (say from an RC) and sometimes doing a patch on much older previous releases), it also doesn't work outside of `build.yml` as you say and you can only use a previously released arcticdb in that instance.

Adjusting your ideal list a bit
- If running as part of a code build and not releasing, use the wheel from the current run.
- If running as part of a release build, **generate versioned docs.** using the tagged docs and build.
- Also allow for a 'dev'/'latest' deployment based on master and last good build.
- If invoked on it's own always specify the version you are rebuilding and deploying 'dev' or a tag, and use the appropriate artifact.

jamesmunro:
Updated docs build: https://baa380c1.arcticdb-docs.pages.dev/

PR #925, opened 2023-09-30
fix interface in ManualClockVersionStore getter

jamesmunro:
Detected when trying to build api docs in mkdocs way.


qc00 (approved):

jamesmunro:
Not really documented by pybind.  Here is a test that shows it expects self/class as a first argument: https://github.com/pybind/pybind11/blob/master/tests/test_methods_and_attributes.cpp#L326

PR #933, opened 2023-10-05
Enhancement 914: Improve error messaging when string column encoding fails due to the presence of a non-string object

alexowens90:
#### Reference Issues/PRs

#914 

Include the column name and row of the input dataframe in the error message when string encoding fails, usually due to the presence of a non-string object in the column.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/pipeline/frame_utils.hpp
+++ b/cpp/arcticdb/pipeline/frame_utils.hpp
@@ -91,7 +91,7 @@ RawType* flatten_tensor(
 }
 
 template<typename Tensor, typename Aggregator>
-void aggregator_set_data(
+std::optional<convert::StringEncodingError> aggregator_set_data(
     const TypeDescriptor &type_desc,
     Tensor &tensor,
     Aggregator &agg,
@@ -102,7 +102,7 @@ void aggregator_set_data(
     size_t regular_slice_size,
     bool sparsify_floats
 ) {
-    type_desc.visit_tag([&](auto &&tag) {
+    return type_desc.visit_tag([&](auto &&tag) {
         using RawType = typename std::decay_t<decltype(tag)>::DataTypeTag::raw_type;
         constexpr auto dt = std::decay_t<decltype(tag)>::DataTypeTag::data_type;
 
@@ -131,6 +131,7 @@ void aggregator_set_data(
                     ptr_data = flatten_tensor<PyObject*>(flattened_buffer, rows_to_write, tensor, slice_num, regular_slice_size);
 
                 auto none = py::none{};
+                std::variant<convert::StringEncodingError, convert::PyStringWrapper> wrapper_or_error;
                 // GIL will be acquired if there is a string that is not pure ASCII/UTF-8
                 // In this case a PyObject will be allocated by convert::py_unicode_to_buffer
                 // If such a string is encountered in a column, then the GIL will be held until that whole column has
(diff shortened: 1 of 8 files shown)


cpp/arcticdb/pipeline/frame_utils.hpp
```diff
@@ -142,12 +142,21 @@ void aggregator_set_data(
...
                         if constexpr (is_utf_type(slice_value_type(dt))) {
-                            auto wrapper= convert::py_unicode_to_buffer(*ptr_data, scoped_gil_lock);
-                            agg.set_string_at(col, s, wrapper.buffer_, wrapper.length_);
+                            wrapper_or_error = convert::py_unicode_to_buffer(*ptr_data, scoped_gil_lock);
                         } else {
-                            auto wrapper = convert::pystring_to_buffer(*ptr_data, false);
-                            agg.set_string_at(col, s, wrapper.buffer_, wrapper.length_);
+                            wrapper_or_error = convert::pystring_to_buffer(*ptr_data, false);
+                        }
+                        // Cannot use util::variant_match as only one of the branches would have a return type
```
willdealtry, line 151: You can use variant matching if you specify the return type I think, as in `[] (const auto&) -> std::optional<convert::StringEncodingError> ( util:raise("blah")); }`

cpp/arcticdb/pipeline/frame_utils.hpp
```diff
@@ -142,12 +142,21 @@ void aggregator_set_data(
                     } else if(is_py_nan(*ptr_data)){
                         agg.set_no_string_at(col, s, nan_placeholder());
                     } else {
+                        std::variant<convert::StringEncodingError, convert::PyStringWrapper> wrapper_or_error;
```
willdealtry, line 145: Do you think there will be a performance impact of constructing a variant a number of times that is linear in terms of the number of rows?

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -83,20 +83,26 @@ folly::Future<std::vector<SliceAndKey>> write_slices(
             auto rows_to_write = slice.row_range.second - slice.row_range.first;
             if (frame.desc.index().field_count() > 0) {
                 util::check(static_cast<bool>(frame.index_tensor), "Got null index tensor in write_slices");
-                aggregator_set_data(
+                auto opt_error = aggregator_set_data(
                     frame.desc.fields(0).type(),
                     frame.index_tensor.value(),
                     agg, 0, rows_to_write, offset_in_frame, slice_num_for_column, regular_slice_size, false);
+                if (opt_error.has_value()) {
```
willdealtry, line 90: Normally we just use the implicit bool conversion.

willdealtry (changes requested):

python/tests/unit/arcticdb/test_string.py
```diff
@@ -161,3 +163,42 @@ def test_write_fixed_coerce_dynamic(lmdb_version_store):
     lmdb_version_store.write("strings", df, dynamic_strings=False)
     vit = lmdb_version_store.read("strings", force_string_to_object=True)
     assert_frame_equal(df, vit.data)
+
+
+def test_string_encoding_error_message(lmdb_version_store_tiny_segment):
```
vasil-pashov, line 168: NIT: I'd make a test class and then put each commented block as a function so that pytest sees them as separate test cases.

cpp/arcticdb/pipeline/frame_utils.hpp
vasil-pashov: I don't see how variant match will improve things. We have different cases based on the contained type, there is no duck-typed branch here allowing to do the same thing for a few types.

alexowens90: What I mean is that we only return from `aggregator_set_data` if the variant holds `StringEncodingError`, otherwise we continue through the loop.

cpp/arcticdb/pipeline/write_frame.cpp
alexowens90: I don't think we're very consistent on this:
```
(dev) [root@d22c3e3750f4 arcticdb]# grep -rI "has_value()" | wc -l
172
```

cpp/arcticdb/pipeline/frame_utils.hpp
alexowens90: I like to think compiler optimisations would amortize it, but can instantiate before the `for` loop to make it explicit

cpp/arcticdb/pipeline/write_frame.cpp
willdealtry: Fair enough, I'm not that fussed. has_value is arguably clearer anyway

cpp/arcticdb/pipeline/frame_utils.hpp
willdealtry: Fair enough, I guess in the case where there's no common return type this way is probably better.

willdealtry (approved):

PR #948, opened 2023-10-11
Support `count` aggregator with groupby

### LINKED ISSUE #929: Add support for `count` aggregator with groupby

Hind-M:
#### Reference Issues/PRs
Fixes #929 

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/processing/aggregation.cpp
+++ b/cpp/arcticdb/processing/aggregation.cpp
@@ -170,17 +170,22 @@ void CountAggregatorData::aggregate(const std::optional<ColumnWithStrings>& inpu
 
             aggregated_.resize(unique_values);
 
-            auto col_data = input_column->column_->data();
-            while (auto block = col_data.next<TypeDescriptorTag>()) {
-                auto ptr = reinterpret_cast<const RawType *>(block.value().data());
-                for (auto i = 0u; i < block.value().row_count(); ++i) {
-                    if (!is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)
-                        || (is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)
-                        && !std::isnan(static_cast<double>(ptr[i])))) {
-                        auto group = groups[i];
-                        auto& val = aggregated_[group];
-                        ++val;
+            if constexpr (is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)) {
+                auto col_data = input_column->column_->data();
+                while (auto block = col_data.next<TypeDescriptorTag>()) {
+                    auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                    for (auto i = 0u; i < block.value().row_count(); ++i) {
+                        if (!std::isnan(static_cast<double>(ptr[i]))) {
+                            auto group = groups[i];
+                            auto& val = aggregated_[group];
+                            ++val;
+                        }
                     }
+                }
+            } else {
+                for (auto group: groups) {
+                    auto& val = aggregated_[group];
+                    ++val;
                 }
             }
         });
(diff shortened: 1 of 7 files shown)


cpp/arcticdb/processing/aggregation.cpp
```diff
@@ -161,4 +161,41 @@ SegmentInMemory MeanAggregatorData::finalize(const ColumnName& output_column_nam
...
+            using TypeDescriptorTag =  decltype(type_desc_tag);
+            using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
+
+            aggregated_.resize(unique_values);
+
+            auto col_data = input_column->column_->data();
+            while (auto block = col_data.next<TypeDescriptorTag>()) {
+                auto ptr = reinterpret_cast<const RawType *>(block.value().data());
+                for (auto i = 0u; i < block.value().row_count(); ++i) {
+                    if(!std::isnan(static_cast<double>(ptr[i]))) {
```
alexowens90, line 177: We can use `if constexpr(is_floating_point_type(TypeDescriptorTag::date_type))` to avoid doing the `NaN` check for integer columns.

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -137,6 +137,60 @@ def test_hypothesis_max_min_agg(lmdb_version_store, df):
     assert_frame_equal(expected, vit.data)
 
 
+@use_of_function_scoped_fixtures_in_hypothesis_checked
+@settings(deadline=None)
+@given(
+    df=data_frames(
+        [
+            column("grouping_column", elements=string_strategy, fill=string_strategy),
+            column("a", elements=non_zero_numeric_type_strategies()),
```
alexowens90, line 146: We should test aggregating a string column as well, the code paths internally are often quite different.

cpp/arcticdb/processing/aggregation.cpp
alexowens90: In fact, the result only depends on `col_data` at all for float and double columns. For integers and strings you can just use the `groups` vector.

alexowens90 (changes requested):

cpp/arcticdb/processing/aggregation.cpp
```diff
@@ -161,4 +161,43 @@ SegmentInMemory MeanAggregatorData::finalize(const ColumnName& output_column_nam
...
 }
+
+void CountAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values) {
+    if(input_column.has_value()) {
+        input_column->column_->type().visit_tag([&] (auto type_desc_tag) {
+            using TypeDescriptorTag =  decltype(type_desc_tag);
+            using RawType = typename TypeDescriptorTag::DataTypeTag::raw_type;
+
+            aggregated_.resize(unique_values);
+
```
alexowens90, line 172: I was more meaning something like:
```
if constexpr (is_floating_point_type(TypeDescriptorTag::DataTypeTag::data_type)) {
    // What you have below, without the is_floating_point_type checks in the if
} else {
    for (auto group: groups) {
        auto& val = aggregated_[group];
        ++val;
    }
}
```
The constexpr is important to make this a compile time rather than runtime check

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -137,6 +138,94 @@ def test_hypothesis_max_min_agg(lmdb_version_store, df):
...
+        ],
+        index=range_indexes(),
+    )
+)
+def test_hypothesis_count_agg_non_zero_numeric(lmdb_version_store, df):
```
alexowens90, line 172: We can remove this and the equivalent dynamic schema test. We only need `non_zero_numeric_type_strategies` to avoid dividing by zero in the filtering and projection tests.

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_aggregation.py
Hind-M: Should we replace the `non_zero_numeric_type_strategies` with `numeric_type_strategies` in `mean`, `sum` and `min/max` tests as well? I can do that in another PR since it's unrelated to the `count`.

alexowens90: Yes, we can just use `numeric_type_strategies ` for all of them since there's no division by the column values involved.

PR #959, opened 2023-10-13
Add asv benchmarks new

### LINKED ISSUE #481: Performance regression testing
ArcticDB recently saw a performance regression. `batch_read` was mistakenly serialised without changing the functionality. As a result, no tests broke.

We should take steps to avoid accidentally creating performance regressions. 

Ideally, I would like to use [airspeed velocity](https://asv.readthedocs.io/en/stable/) to identify the commits that have resulted in a performance regression. This could be run by CI off of commits to master.

Alternatively we could simply have a suite of tests with expected runtimes that could be run manually.

G-D-Petrov:
#### Reference Issues/PRs
Fixes #481 

Adds benchmarks for some of the basic functionalities.

#### Any other comments?
**Not 100% ready**
There is also documentation on the [Wiki](https://github.com/man-group/ArcticDB/wiki/Running-ASV-Benchmarks)


.github/workflows/analysis_workflow.yml
```diff
@@ -2,9 +2,11 @@ name: Build with analysis tools
 on:
   workflow_dispatch:
     inputs:
-      run:
+      run_all_benchmarks:
         type: boolean
 
+  push:
+    branches: ["add_asv_benchmarks_new"]
```
poodlewars, line 9: Update before merge

.github/workflows/analysis_workflow.yml
```diff
@@ -15,77 +17,243 @@ jobs:
         cibuildwheel_ver: "2.12.1"
         force_update: false
 
-  code_coverage:
-    needs: [cibw_docker_image]  
-    runs-on: "ubuntu-latest"
+  # code_coverage:
```
poodlewars, line 20: ?

.github/workflows/analysis_workflow.yml
```diff
@@ -101,102 +269,59 @@ jobs:
...
-          path: cpp/out/linux-debug-build/coverage.zip
-
-      - name: Upload Python Coverage
-        uses: actions/upload-artifact@v3
-        with:
-          name: python-coverage-artifact
-          path: python/python_cov.zip
+          python -m asv machine -v --yes --machine ArcticDB-Runner-Medium
+          git config --global --add safe.directory .
+          # if ["${{inputs.run_all_benchmarks}}"]; then
```
poodlewars, line 275: Commented out code here?

.github/workflows/analysis_workflow.yml
```diff
@@ -101,102 +269,59 @@ jobs:
...
-                  "type": "section",
-                  "text": {
-                    "type": "mrkdwn",
-                    "text": "The CPP Code Coverage from the current run(${{ env.CPP_COV_PERCENT }}%) is lower the previous one(${{ env.CPP_COV_PREV_PERCENT }}%)."
-                  }
-                }
-              ]
-            }
+          python -m asv gh-pages --rewrite
+          # if [[!"${{github.ref}}" -eq 'refs/heads/master']]; then
```
poodlewars, line 300: Commented out code?

.github/workflows/analysis_workflow.yml
```diff
@@ -101,102 +269,59 @@ jobs:
...
+          #   python -m asv gh-pages --rewrite
+          # fi
         env:
-          SLACK_WEBHOOK_URL: ${{ secrets.ARCTICDB_DEV_WEBHOOK_URL }}
-          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+          CMAKE_BUILD_PARALLEL_LEVEL: 8
```
poodlewars, line 304: Can you remind me how we're building the ArcticDB wheel that we're benchmarking? Is it definitely a `Release` build rather than a `Debug` build?

It actually looks like were benchmarking against the build that was instrumented for code coverage? That doesn't seem right?

.github/workflows/analysis_workflow.yml
```diff
@@ -101,102 +269,59 @@ jobs:
...
+          SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
+          VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
+          VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
+          VCPKG_MAN_NUGET_USER: ${{secrets.VCPKG_MAN_NUGET_USER}} # For forks to download pre-compiled dependencies from the Man repo
+          VCPKG_MAN_NUGET_TOKEN: ${{secrets.VCPKG_MAN_NUGET_TOKEN}}
+          CMAKE_C_COMPILER_LAUNCHER: sccache
+          CMAKE_CXX_COMPILER_LAUNCHER: sccache
+
+      # - name: Run benchmarks for latest head
```
poodlewars, line 313: Commented out code

benchmarks/basic_functions.py
poodlewars, line 1: Would it make sense for these to go in `python/benchmarks` rather than `$ROOT/benchmarks`? Keen to keep the top-level directory structure of the project clean. Would also make sure we apply the Black formatting against these.

benchmarks/basic_functions.py
```diff
@@ -0,0 +1,148 @@
...
+)
+
+from .common import *
+
+class BasicFunctions:
+    number = 5
+    timeout = 6000
+
+    params = ([50, 100], [500, 1000])
+    param_names = ['rows', 'num_symbols']
```
poodlewars, line 21: not many rows! Hundreds of thousands would be quite typical

benchmarks/basic_functions.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def peakmem_write(self, rows, num_symbols):
+        lib = self.get_fresh_lib()
+        for sym in range(num_symbols):
+            lib.write(f"{sym}_sym", self.df)
+
+    def time_write_staged(self, rows, num_symbols):
+        lib = self.get_fresh_lib()
+        for sym in range(num_symbols):
+            lib.write(f"{sym}_sym", self.df, staged=True)
```
poodlewars, line 66: And a compact incomplete to finalize the staged data?

benchmarks/common.py
```diff
@@ -0,0 +1,72 @@
+"""
+Copyright 2023 Man Group Operations Limited
```
poodlewars, line 2: Top marks for remembering these 🥇

benchmarks/common.py
```diff
@@ -0,0 +1,72 @@
...
+    # Generate timestamps
+    timestamps = pd.date_range(end=end_timestamp, periods=n, freq=freq)
+    # Create dataframe
+    df = pd.DataFrame({"value": values})
+    df.index = timestamps
+    return df
+
+def generate_benchmark_df(N, freq="T", end_timestamp="1/1/2023"):
+    timestamps = pd.date_range(end=end_timestamp, periods=N, freq=freq)
+    K = N // 10
```
poodlewars, line 52: Bit weird variable naming here, upper-case variables typically reserved for module-level constants

benchmarks/persistent_query_builder.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def setup(self, num_rows):
+        pass
+
+    # The names are based on the queries used here: https://duckdblabs.github.io/db-benchmark/
+    def time_query_1(self, num_rows):
+        lib = self.ac[self.lib_name]
+        q = QueryBuilder()
+        q = q.groupby("id1").agg({"v1": "sum"})
+        _ = lib.read(f"{num_rows}_rows", query_builder=q)
```
poodlewars, line 33: No need for the assignment is there?

poodlewars (changes requested):
Just minor comments on the Python parts, but looks good.

Bit confused by the build script parts while there is commented out code in there, and I'd like to make sure we're benchmarking using a release build.

Actually, is there any reason not to put the performance testing steps in the normal "build and test" workflow? And have them triggered only if a particular variable is set, or something.

IIRC we needed code cov in its own workflow as it required an instrumented build of the code.

benchmarks/common.py
G-D-Petrov: Good point, it was copy-pasted from the original benchmark, I will fix it

benchmarks/basic_functions.py
G-D-Petrov: I will look into this, more rows means slower benchmarks, so I was conservative to speed-up the iterations.

benchmarks/basic_functions.py
G-D-Petrov: I will look into this, should be possiblle

.github/workflows/analysis_workflow.yml
G-D-Petrov: The code coverage and the benchmarking and separate jobs, so interference there.
The benchmarking job is building `Release` builds.

PR #964, opened 2023-10-16
Fix emptyval

willdealtry:
We are writing a single byte of compressed data when there is a column of the type EMPTYVAL, but we aren't adviancing the read pointer in the python handler. This causes an LZ4 decode error when reading the subsequent column.

### DIFF
--- a/cpp/arcticdb/python/python_handlers.cpp
+++ b/cpp/arcticdb/python/python_handlers.cpp
@@ -6,43 +6,46 @@
  */
 #include <python/python_handlers.hpp>
 #include <python/gil_lock.hpp>
+#include <arcticdb/codec/codec.hpp>
 
 namespace arcticdb {
-    void EmptyHandler::handle_type(
-        const uint8_t*& input,
-        uint8_t* dest,
-        const VariantField& variant_field,
-        const entity::TypeDescriptor&,
-        size_t dest_bytes,
-        std::shared_ptr<BufferHolder>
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -6,25 +6,46 @@
...
+    auto ptr_dest = reinterpret_cast<const PyObject **>(dest);
+    py::none none = {};
+    for (auto row = 0u; row < num_rows; ++row) {
+        none.inc_ref();
+        *ptr_dest++ = none.ptr();
     }
+
+    util::variant_match(variant_field, [&input](const auto &field) {
+        using EncodedFieldType = std::decay_t<decltype(*field)>;
+        if constexpr (std::is_same_v<EncodedFieldType, arcticdb::EncodedField>)
```
vasil-pashov, line 34: Is this going to be correct always? Maybe we should propagate the encoding version from the segment and check on it.

vasil-pashov (approved):

cpp/arcticdb/python/python_handlers.cpp
willdealtry: Good question. The encoding version is really a write-time thing, it controls which encoding you would prefer to write in. Clients with a write preference for one or the other should still be able to read either version, and the generally-used proxy for that is exactly this type distinction.

PR #966, opened 2023-10-16
Fix storage lock implementation for Windows debug builds

poodlewars:
This avoids error messages complaining that mutexes are destroyed while still active.

This has also makes `lock` consistent with `lock_timeout`. `lock_timeout` _always_ used to release the mutex on-exit, whereas to be consistent with `lock`, it should only release when there actually is a timeout.


willdealtry (approved):

poodlewars:
In hindsight as WIlliam said we should really just unlock the mutex in the `StorageLock` destructor... This will do for now though...

PR #970, opened 2023-10-18
Add asv benchmarks new (#962)

### LINKED ISSUE #481: Performance regression testing

G-D-Petrov:
#### Reference Issues/PRs
Fixes #481 

Adds benchmarks for some of the basic functionalities.

#### Any other comments?
**Not 100% ready**
There is also documentation on the [Wiki](https://github.com/man-group/ArcticDB/wiki/Running-ASV-Benchmarks)


.github/workflows/analysis_workflow.yml
```diff
@@ -200,3 +200,152 @@ jobs:
...
+        with:
+          aws_access_key: "${{ secrets.AWS_S3_ACCESS_KEY }}"
+          aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"
+
+      - name: Configure sccache
+        uses: mozilla-actions/sccache-action@v0.0.3
+        with:
+          version: "v0.4.0"
+        
+      # We are chainging the python here because we want to use the default python to build (it is devel version)
```
poodlewars, line 236:
```suggestion
      # We are changing the python here because we want to use the default python to build (it is devel version)
```

python/benchmarks/common.py
```diff
@@ -0,0 +1,76 @@
...
+    df = pd.DataFrame({"value": values})
+    df.index = timestamps
+    return df
+
+
+def generate_benchmark_df(n, freq="T", end_timestamp="1/1/2023"):
+    timestamps = pd.date_range(end=end_timestamp, periods=n, freq=freq)
+    k = n // 10
+    # Based on https://github.com/duckdblabs/db-benchmark/blob/master/_data/groupby-datagen.R#L19
+    DT = pd.DataFrame()
```
poodlewars, line 57: Still have the weird capitalized `DT` variable name

poodlewars (approved):

PR #973, opened 2023-10-18
Add mutex to ensure only single thread at pybind->c++ layer

phoebusm:
#### Reference Issues/PRs

https://github.com/man-group/ArcticDB/issues/974
https://github.com/man-group/ArcticDB/pull/973#discussion_r1363967704
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/python_bindings.cpp
+++ b/cpp/arcticdb/storage/python_bindings.cpp
@@ -11,6 +11,7 @@
 #include <pybind11/stl.h>
 
 #include <arcticdb/util/variant.hpp>
+#include <arcticdb/util/pybind_mutex.hpp>
 
 #include <arcticdb/storage/library.hpp>
 #include <arcticdb/storage/library_manager.hpp>
@@ -139,10 +140,10 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
             LibraryPath lib_path{library_path, '.'};
             return library_manager.write_library_config(lib_cfg, lib_path, storage_override, validate);
         },
-             py::arg("lib_cfg"),
-             py::arg("library_path"),
-             py::arg("override") = StorageOverride{},
-             py::arg("test_only_validation_toggle") = false)
+            py::arg("lib_cfg"),
+            py::arg("library_path"),
+            py::arg("override") = StorageOverride{},
+            py::arg("test_only_validation_toggle") = false)
         .def("get_library_config", [](const LibraryManager& library_manager, std::string_view library_path, const StorageOverride& storage_override){
             return library_manager.get_library_config(LibraryPath{library_path, '.'}, storage_override);
         }, py::arg("library_path"), py::arg("override") = StorageOverride{})
@@ -151,7 +152,7 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
         }, py::arg("library_path"), py::arg("throw_on_failure") = true)
         .def("remove_library_config", [](const LibraryManager& library_manager, std::string_view library_path){
             return library_manager.remove_library_config(LibraryPath{library_path, '.'});
-        })
+        }, py::call_guard<SingleThreadMutexHolder>())
         .def("get_library", [](LibraryManager& library_manager, std::string_view library_path, const StorageOverride& storage_override){
             return library_manager.get_library(LibraryPath{library_path, '.'}, storage_override);
         }, py::arg("library_path"), py::arg("storage_override") = StorageOverride{})
--- a/cpp/arcticdb/version/python_bindings.cpp
+++ b/cpp/arcticdb/version/python_bindings.cpp
@@ -23,6 +23,7 @@
 #include <arcticdb/pipeline/value_set.hpp>
 #include <arcticdb/python/adapt_read_dataframe.hpp>
 #include <arcticdb/version/schema_checks.hpp>
+#include <arcticdb/util/pybind_mutex.hpp>
 
 namespace arcticdb::version_store {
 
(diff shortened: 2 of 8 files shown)


cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1122,4 +1125,11 @@ void PythonVersionStore::force_delete_symbol(const StreamId& stream_id) {
     delete_all_for_stream(store(), stream_id, true);
 }
 
+std::lock_guard<std::mutex> PythonVersionStore::ensure_single_thread_cpp_pybind_entry(){ 
+    py::gil_scoped_release release;
```
vasil-pashov, line 1129: Why is there a gil release call here? IMO this does a bit more than the name implies.

poodlewars:
Please could you,

- Explain why we need the GIL part? I assume this is something about what happens if we hold the GIL in the thread blocked by the new mutex, which might deadlock the thread holding the mutex if _that_ has previously dropped the GIL and is now waiting to re-acquire it.
- Confirm how this works with Python multiprocessing? Is there a new mutex for each sub-process?
- Add tests for this for both multiprocessing (should permit the concurrency) and multithreading (should forbid it)

phoebusm:
> Please could you,
> 
> * Explain why we need the GIL part? I assume this is something about what happens if we hold the GIL in the thread blocked by the new mutex, which might deadlock the thread holding the mutex if _that_ has previously dropped the GIL and is now waiting to re-acquire it.
> * Confirm how this works with Python multiprocessing? Is there a new mutex for each sub-process?
> * Add tests for this for both multiprocessing (should permit the concurrency) and multithreading (should forbid it)
> * Share any performance benchmarking showing the impact of this when not using Python multi-threading
> 
> I thought the plan was to add this to all the functions. Is the idea that that will be a follow up?

1. Exactly
2. Correct me if I am wrong, but I believe C++ mutex is intraprocess only. 
3. Yea it passes both multithreading and multiprocessing test
4. Yes this is something I should do

I think I have misunderstood the urgency of the fix. I thought it is an urgent task so I only fix the one being reported. That's why this PR is drafted.
Pending task for a complete PR:
1. Mutex for all pybind function
2. Test
3. Benchmark
If above tasks are required before merging, I will like to add WIP to this PR or simply close it until the above tasks are ready

poodlewars:
> > Please could you,
> > 
> > * Explain why we need the GIL part? I assume this is something about what happens if we hold the GIL in the thread blocked by the new mutex, which might deadlock the thread holding the mutex if _that_ has previously dropped the GIL and is now waiting to re-acquire it.
> > * Confirm how this works with Python multiprocessing? Is there a new mutex for each sub-process?
> > * Add tests for this for both multiprocessing (should permit the concurrency) and multithreading (should forbid it)
> > * Share any performance benchmarking showing the impact of this when not using Python multi-threading
> > 
> > I thought the plan was to add this to all the functions. Is the idea that that will be a follow up?
> 
>     1. Exactly
> 
>     2. Correct me if I am wrong, but I believe C++ mutex is intraprocess only. <- Ok just realise something. Let me do more tests
> 
>     3. Yea it passes both multithreading and multiprocessing test
> 
>     4. Yes this is something I should do
> 
> 
> I think I have misunderstood the urgency of the fix. I thought it is an urgent task so I only fix the functions being reported in this PR. That's why this PR is drafted. Pending task for a complete PR:
> 
>     1. Mutex for all pybind function
> 
>     2. Test
> 
>     3. Benchmark
> 
> 
> If above tasks are required before merging, I will like to add WIP to this PR or simply close it until the above tasks are ready

Thanks. It would be good to confirm the urgency with the users, I'm not sure whether they have a good workaround or not. If they're in a rush, I think it's OK just to add this to read and read batch, and we can follow up immediately for the other methods. Otherwise, let's do them all at once and save ourselves an extra release.

I think it would be good to have the multithreading and processing tests whether they are in a rush or not, we should be careful that we are not making things worse.

cpp/arcticdb/version/version_store_api.cpp
vasil-pashov: Can you also add this information as a comment above the function (preferably in doxygen style)?

phoebusm:
No deterioration of performance observed after adding the lock:
Average time taken for `read`
Before adding the lock: 0.027s
After adding the lock: 0.023s

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -1299,3 +1302,51 @@ def test_read_description_batch_empty_nat(arctic_library):
...
+
+    def read():
+        try:
+            for sym in range(num_symbols):
+                assert_frame_equal(lib.read("sym_" + str(sym)).data, list_dataframes[sym])
+        except Exception as e:
+            q.put(e)
+
+    read_th_count = 1
+    batch_read_th_count = 1
```
poodlewars, line 1345: Increase these thread counts?

What is the point of the queue (which isn't synchronised correctly), why not just let the exception bubble up and fail the test?

cpp/arcticdb/version/version_store_api.hpp
```diff
@@ -344,6 +344,8 @@ class PythonVersionStore : public LocalVersionedEngine {
         bool prune_previous_versions);
 
     void delete_snapshot_sync(const SnapshotId& snap_name, const VariantKey& snap_key);
+
+    [[nodiscard]] static std::lock_guard<std::mutex> ensure_single_thread_cpp_pybind_entry();
```
poodlewars, line 348: I'm confused, I can't see where this is defined, or used?

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -148,7 +149,7 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
         }, py::arg("library_path"), py::arg("throw_on_failure") = true)
         .def("remove_library_config", [](const LibraryManager& library_manager, std::string_view library_path){
             return library_manager.remove_library_config(LibraryPath{library_path, '.'});
-        })
+        }, py::call_guard<SingleThreadMutexHolder>())
```
poodlewars, line 152: Why does this method need the mutex, and not the other methods in this class?

python/tests/integration/arcticdb/test_arctic_batch.py
phoebusm: I can increase the thread count. Yet according to my test, current setting is sufficient to repeating the issue. I would like to keep the additional time needed for running the test as minimum as possible.

cpp/arcticdb/version/version_store_api.hpp
phoebusm: It is used as `py::call_guard<SingleThreadMutexHolder>()` while defining `pybind` function. 
```
m.def("foo", foo, py::call_guard<T>());
```
is equal to 
```
m.def("foo", [](args...) {
    T scope_guard;
    return foo(args...); // forwarded arguments
});
```

cpp/arcticdb/async/python_bindings.cpp
```diff
@@ -8,6 +8,7 @@
 #include <arcticdb/python/python_utils.hpp>
 #include <arcticdb/async/python_bindings.hpp>
 #include <arcticdb/async/task_scheduler.hpp>
+#include <arcticdb/util/pybind_mutex.hpp>
```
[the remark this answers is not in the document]
phoebusm: Oops sorry will be remove

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -405,55 +406,55 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
              py::arg("library"),
              py::arg("license_key") = std::nullopt)
         .def("write_partitioned_dataframe",
-             &PythonVersionStore::write_partitioned_dataframe,
-             "Write a dataframe to the store")
+             &PythonVersionStore::write_partitioned_dataframe, 
+             py::call_guard<SingleThreadMutexHolder>(), "Write a dataframe to the store")
```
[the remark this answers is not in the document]
phoebusm: I have thought about that but the `call_guard` is only needed for functions involves folly::future, which are not the majority of the case. So at the end of the day, developers still need to differentiate whether they need the guard or not. Then, in your case, assigning the new subclass of `py::class_`. To me, it is not a big different to adding the `py::call_guard`.
Or, if you agree adding the mutex to add functions, it will justify this idea, I think.

cpp/arcticdb/storage/python_bindings.cpp
phoebusm: I think I should have put the explanation in the description of this PR.
So basically, my understanding is, this will be short-lived patch. So my idea is keeping it safe for the short term:
1. No major rehaul
2. Add mutex to all functions in `PythonVersionStore`
3. If functions use folly::Future, which, in my eyes, can be converted async task easily. So 
`remove_key` is called by `remove_library_config`, which uses `folly::Future`, mutex is necessary

python/tests/integration/arcticdb/test_arctic_batch.py
poodlewars: Sounds good, the minimal amount to repro the issue is fine.

cpp/arcticdb/version/python_bindings.cpp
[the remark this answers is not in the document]
poodlewars: I was thinking just for `PythonVersionStore`, but I can see that my idea might add about as much complexity as it removes. Let's leave it as-is for now.

cpp/arcticdb/version/version_store_api.hpp
phoebusm: Oops I see your point now. Removing

poodlewars (approved):

poodlewars (changes requested):
Worried about forking? Then new process will get the mutex state from the static one in `pybind_mutex.hpp`, which may be locked. How will it ever get unlocked? Do we need some mechanism to have a per-process mutex? Does pybind have anything to help us here?

poodlewars (approved):

phoebusm:
> Worried about forking? Then new process will get the mutex state from the static one in `pybind_mutex.hpp`, which may be locked. How will it ever get unlocked? Do we need some mechanism to have a per-process mutex? Does pybind have anything to help us here?

Handling for multiprocess added

PR #975, opened 2023-10-18
Add a mechanism to extend storage transaction lifetime to lifetime of…

willdealtry:
We discovered that in some cases a zero-copy reference to an LMDB read transaction is maintained longer than the transaction itself. This change add a type-erased mechanism for RAII extension of transaction lifetime for storages where this is necessary

### DIFF
--- a/cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
+++ b/cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
@@ -101,7 +101,7 @@ void LmdbStorage::do_update(Composite<KeySegmentPair>&& kvs, UpdateOpts opts) {
 
 void LmdbStorage::do_read(Composite<VariantKey>&& ks, const ReadVisitor& visitor, storage::ReadKeyOpts) {
     ARCTICDB_SAMPLE(LmdbStorageRead, 0)
-    auto txn = ::lmdb::txn::begin(env(), nullptr, MDB_RDONLY);
+    auto txn = std::make_shared<::lmdb::txn>(::lmdb::txn::begin(env(), nullptr, MDB_RDONLY));
 
     auto fmt_db = [](auto &&k) { return variant_key_type(k); };
     ARCTICDB_SUBSAMPLE(LmdbStorageInTransaction, 0)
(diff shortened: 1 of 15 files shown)


poodlewars (approved):

cpp/arcticdb/version/version_map.hpp
```diff
@@ -802,7 +802,6 @@ class VersionMapImpl {
         auto entry = get_entry(stream_id);
         entry->clear();
         load_via_iteration(store, stream_id, entry, false);
-        fix_stream_ids_of_index_keys(store, stream_id, entry);
```
poodlewars, line 805: Do we need to keep this `scan_and_rewrite` and `fix_ref_key` functions at all? What are they for?

poodlewars (approved):

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
```diff
@@ -115,10 +115,13 @@ void LmdbStorage::do_read(Composite<VariantKey>&& ks, const ReadVisitor& visitor
...
 
-            if (::lmdb::dbi_get(txn, dbi.handle(), &mdb_key, &mdb_val)) {
+            if (::lmdb::dbi_get(*txn, dbi.handle(), &mdb_key, &mdb_val)) {
                 ARCTICDB_SUBSAMPLE(LmdbStorageVisitSegment, 0)
-                visitor(k, Segment::from_bytes(reinterpret_cast<std::uint8_t *>(mdb_val.mv_data),
-                                               mdb_val.mv_size));
+                auto segment = Segment::from_bytes(reinterpret_cast<std::uint8_t *>(mdb_val.mv_data),
+                                                   mdb_val.mv_size);
+                std::any keepalive;
+                segment.set_keepalive(std::any(std::move(txn)));
```
poodlewars, line 123: Do we have to now be careful that `segment` is not used by others threads, and that this thread does not open another transaction? See http://www.lmdb.tech/doc/group__mdb.html#gad7ea55da06b77513609efebd44b26920 `mdb_txn_begin` docs:

```
A transaction and its cursors must only be used by a single thread, and a thread may only have a single transaction at a time. If [MDB_NOTLS](http://www.lmdb.tech/doc/group__mdb__env.html#ga5dca84a576d14b4bfe2deddc2dc622d3) is in use, this does not apply to read-only transactions. 
    Cursors may not span transactions.
```

willdealtry: We're fine as long as they're read-only transactions. Also we set up the environment with MDB_NOTLS, so the reader locktable isn't tied to a particular thread.

joe-iddon (approved):

PR #625, opened 2023-07-20
maint: Remove `VariantStorageFactory` and its implementations

jjerphan:
#### Reference Issues/PRs

Towards https://github.com/man-group/ArcticDB/issues/624.

Discussed with @willdealtry, @alexowens90 and others.

#### What does this implement/fix? Explain your changes.

This removes unnecessary factories which are useless proxy encapsulating a convoluted logic to simply create instance of storage (i.e. instance of `VariantStorage`).

#### Any other comments?

Eventually we could remove a few file once variants are not used anymore for storages. What do you think?

PR #991, opened 2023-10-23
RocksDB Python API

### LINKED ISSUE #913: RocksDB Backend

joe-iddon:
#### Reference Issues/PRs

Closes #913 
Extends #945 (C++ tests) and #961 (Conda mac problems)

Implements the RocksDB Python API. **Update: We will not advertise this yet, since conda support is not there, and the documentation still needs writing.**

Since the C++ code for RocksDB is excluded on conda, a static readonly property is attached to `LibraryManager` in the `python_bindings.cpp` there so that the Python code knows whether to offer up the `rocksdb_library_adapter.py`, or not.

I.e. 
```python
>>> arcticdb_ext.storage.LibraryManager.rocksdb_support
False # On conda. On PyPi (windows + Linux), this will give True
```

Still to do:
- Add a warning if ever opened twice, like in #1000 . Unlike LMDB, RocksDB does not support multi-process access.
- Add a proper cleanup method, as was done with LMDB. See code of `rocksdb_library_adapter.py`. Should be similar to #918
- See below about switching from `Slice` to `PinnableSlice`. 

#### Any other comments?

Removed the `std::make_unique` anti-pattern that was introduced to `storage_factory.cpp` in #625 which required `Storage` implementations to be move-constructable. This was highlighted to me since I defined a custom destructor for the `RocksDBStorage` class which destroyed its default move constructor, so gave compiler errors.

#### Ideas for Profiling + Improving Performance / Configuring RocksDB

Profiling ideas:
- Need to compile in release not debug mode, since RocksDB should behave better in this mode
- Flamegraph+perf
- Enable performance tracing with `#define`. Note might want to add more `ARCTICDB_SAMPLE` macros for this to `rocksdb_storage.cpp`.
- Use py-spy with PID input and speedscope output using options `--native` to profile C++ code, and `--idle` to include time spent waiting on I/O operations etc
- Edit: We now have a Wiki page to [explain our profiling options](https://github.com/man-group/ArcticDB/wiki/ArcticDB-Performance-Profiling).
- Ideas for graphs/parameters to vary in comparing LMDB and RocksDB:
  - write+read for different data sizes, varying segment size (rows+columns) since RocksDB's LSM tree should out-perform LMDB's BTree when there are a large number of key/value pairs. 
  - `batch_write` and `batch_read` operations to compare parallel performance
  - `append` and `update` as these will produce smaller segments, so may favor RocksDB as more key/value pairs

Performance ideas
- As noted in the code, should switch from `Slice` to `PinnableSlice`. This is only possible once #975 has been merged, but will enable us to eliminate a copy in creating the `Segment` from the bytes during a read.
- Writes cannot be optimised in this way. There is a necessary copy, which unlike in LMDB, cannot be avoided.
- Optimise the [BloomFilter](https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter) (already enabled, but could be tuned).
- Add [block caching](https://github.com/facebook/rocksdb/wiki/Block-Cache)
- Configure number of SST files in L0, L1, etc.
- Configure how many I/O threads to give RocksDB. Currently set to 16 from default 1 using `IncreaseParallelism`, but don't want too many, because then might try too hard to distribute work and harm performance from context switching.
- Implement `WriteBatch` when deleting/putting multiple key/values, rather than `Put` and `Delete` one-by-one.
- Add a `WaitForCompact` before closing as this will mean the DB can be opened faster next time.

Documentation and reference material on RocksDB:
- Documentation of RocksDB is [on Facebook's GitHub Wiki](https://github.com/facebook/rocksdb/wiki).
- The [Basic Operations](https://github.com/facebook/rocksdb/wiki/Basic-Operations) section is a good starting point.
- A good guide to the internals can be found [here](https://artem.krylysov.com/blog/2023/04/19/how-rocksdb-works/).

Random other idea:
- The C++ test parameterization in `test_embedded.cpp` could be made neater in two ways: either just spread the list definition over multiple lines, then can literally `#define` out the one `RocksDB` if we are on conda; OR make the BackendGenerator class have a `constexpr` constructor and then should be able to use a `constexpr` version of `emplace_back` ... except not on C++20, so would have to write a `constexpr` version of `emplace_back`.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/storage/library_manager.hpp
```diff
@@ -38,6 +38,14 @@ namespace arcticdb::storage {
 
         [[nodiscard]] bool has_library(const LibraryPath& path) const;
 
+        [[nodiscard]] static inline bool rocksdb_support() {
```
poodlewars, line 41: We can just hang thing off the module like here https://github.com/man-group/ArcticDB/blob/263e843015f010b4fcd30acb47b05a600b3d7448/cpp/arcticdb/storage/python_bindings.cpp#L38 can't we, no need for this to be tied to the `library_manager`?

cpp/arcticdb/storage/rocksdb/rocksdb_storage.cpp
```diff
@@ -118,17 +131,26 @@ void RocksDBStorage::do_read(Composite<VariantKey>&& ks, const ReadVisitor& visi
...
 
+    std::vector<VariantKey> failed_reads;
     (fg::from(ks.as_range()) | fg::move | fg::groupBy(grouper)).foreach([&](auto &&group) {
         auto key_type_name = fmt::format("{}", group.key());
         auto handle = handles_by_key_type_.at(key_type_name);
         for (const auto &k : group.values()) {
             std::string k_str = to_serialized_key(k);
             std::string value;
+            // TODO: Once PR: 975 has been merged we can use ::rocksdb::PinnableSlice to avoid the copy in
+            //       the consturction of the segment
```
poodlewars, line 142:
```suggestion
            //       the construction of the segment
```

cpp/arcticdb/storage/storage_factory.cpp
```diff
@@ -27,39 +30,35 @@ std::unique_ptr<Storage> create_storage(
     if (type_name == s3::S3Storage::Config::descriptor()->full_name()) {
         s3::S3Storage::Config s3_config;
         storage_descriptor.config().UnpackTo(&s3_config);
-        storage = std::make_unique<s3::S3Storage>(
-                s3::S3Storage(library_path, mode, s3_config)
-        );
+        storage = std::make_unique<s3::S3Storage>(library_path, mode, s3_config);
```
poodlewars, line 33: Ah wow, nice cleanup, much better.

python/arcticdb/adapters/rocksdb_library_adapter.py
```diff
@@ -0,0 +1,78 @@
...
+Copyright 2023 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+import re
+import os
+
+# from dataclasses import dataclass
```
poodlewars, line 11: rm

python/arcticdb/adapters/rocksdb_library_adapter.py
```diff
@@ -0,0 +1,78 @@
...
+from arcticdb.version_store.helper import add_rocksdb_library_to_env
+from arcticdb.config import _DEFAULT_ENV
+from arcticdb.version_store._store import NativeVersionStore
+from arcticdb.adapters.arctic_library_adapter import ArcticLibraryAdapter, set_library_options
+from arcticdb.encoding_version import EncodingVersion
+from arcticdb_ext.storage import CONFIG_LIBRARY_NAME
+
+
+class RocksDBLibraryAdapter(ArcticLibraryAdapter):
+    """
```
poodlewars, line 25: Can you document in here that this is a work in progress, beta API and should not be relied upon?
And also log some warnings if anyone does actually use the RocksDB adapter so they see at runtime we're not making any promises about it.

PR #996, opened 2023-10-24
Fix ccache's non-existence exiting the workflow

joe-iddon:

After merging the initial rocksdb work, clearly some build machines do not have the `ccache` installed, so we need a fix for this to not exit the workflow.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -104,7 +104,7 @@ jobs:
 
       - name: Find and remove ccache # See PR: #945
         if: matrix.os == 'windows'
-        run: rm $(which ccache)
+        run: rm $(which ccache) || true
 
       - name: Prepare C++ compilation env
         if: inputs.job_type != 'follower'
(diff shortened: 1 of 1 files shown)


.github/workflows/build_steps.yml
```diff
@@ -177,7 +177,7 @@ jobs:
       # ========================= Common =========================
       - name: Disk usage
         if: always()
-        run: du -m . ${{matrix.build_dir}} ${{matrix.vcpkg_packages_dir}} | sort -n | tail -n 50; df -h
+        run: du -m . ${{matrix.build_dir}} ${{matrix.vcpkg_packages_dir}} | sort -n | tail -n 50 || true; df -h
```
qc00, line 180: What could fail here?

qc00 (approved):

.github/workflows/build_steps.yml
joe-iddon: If those directories haven't been created yet then the `du` will fail and we don't get to inspect the `df` output.

PR #997, opened 2023-10-25
941 self hosted runners for ci

### LINKED ISSUE #941: Self-hosted runners for CI
Set up self-hosted runners for our CI workflows, rather than using the free ones that Github provides.

We found the Github provided "large" runners prohibitively expensive (hundreds of thousands of dollars per year).

So we can look at:

- AWS or Azure hosted runners (either is fine, this is the first preference as long as cost not in to hundreds of thousands per year)
- On-prem hosted runners within Man Group (Linux quite easy, could reuse our "build servers"? Windows harder as those build agents are more locked down). Could even consider just buying some of our own boxes and running them ourselves rather than in a datacentre. Only do this if the cloud hosted versions are prohibitively expensive.

We want to do the above for Linux and Windows.

For Mac, we have some Mac Studio boxes in the post and we'll use those.

Relates to #481

G-D-Petrov:
#### Reference Issues/PRs
Fixes #941

Adds self-hosted Linux runners for the Leader job in build.yml and build_with_conda.yml.
Before this changes, just the compilation of a Linux leader takes more than 20 min.
With these changes the Linux leader build takes less than 20 min to compile and test against all cpp and almost all pytests.

This way, developers can get pretty good feedback on their changes, faster than before.

#### Other comments
Some of the pytests have been changed to make them more consistent so we can paralelize them with [pytest-xdist](https://pytest-xdist.readthedocs.io/en/latest/known-limitations.html).
Some have been disabled to make the runs faster.

### DIFF
--- a/.github/actions/run_local_pytest/action.yml
+++ b/.github/actions/run_local_pytest/action.yml
@@ -1,0 +1,27 @@
+name: 'Run Local Pytest'
+description: 'Installs Python dependencies and runs Pytest against locally built ArcticDB binaries'
+inputs:
+  build_type:        {default: 'release', type: string, description: Specifies whether the build that we are testing is release or debug}
+  threads:           {default: '1', type: string, description: Specifies the number of threads that will be used for tests }
+  fast_tests_only:   {default: '0', type: string, description: Specifies whether we want to run only the fast tests or not }
+  other_params:      {default: '', type: string, description: Other params that should be passed to the python command }
+runs:
+  using: "composite"
+  steps:
+    - name: Run local pytests
+      shell: bash
+      run: |
+        yum install nodejs npm -y
+        npm config set strict-ssl false
+        npm install -g azurite
+        
+        cd python
+        ln -s ../cpp/out/linux-${{ inputs.build_type }}-build/arcticdb/arcticdb_ext.cpython-36m-x86_64-linux-gnu.so
+        export ARCTICDB_RAND_SEED=$RANDOM
+        python ${{inputs.other_params}} -m pytest -n ${{ inputs.threads }} tests
+      env:
+        TEST_OUTPUT_DIR: ${{ runner.temp }}
+        ARCTICDB_FAST_TESTS_ONLY: ${{ inputs.fast_tests_only }}
+        # Use the Mongo created in the service container above to test against
+        CI_MONGO_HOST: mongodb
+        CI_MONGO_HOST_2: mongodb_2
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -7,6 +7,9 @@ on:
 
   schedule: # Schdeule the job to run at 12 a.m. daily
     - cron: '0 0 * * *'
+
+  pull_request:
+
 jobs:
   cibw_docker_image:
     uses: ./.github/workflows/cibw_docker_image.yml
@@ -26,13 +29,10 @@ jobs:
         ports:
           - 27017:27017
     env:
-      SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
       VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
       VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
       VCPKG_MAN_NUGET_USER: ${{secrets.VCPKG_MAN_NUGET_USER}} # For forks to download pre-compiled dependencies from the Man repo
       VCPKG_MAN_NUGET_TOKEN: ${{secrets.VCPKG_MAN_NUGET_TOKEN}}
-      CMAKE_C_COMPILER_LAUNCHER: sccache
-      CMAKE_CXX_COMPILER_LAUNCHER: sccache
       ARCTIC_CMAKE_PRESET: linux-debug
     steps:
       - uses: actions/checkout@v3
@@ -49,6 +49,7 @@ jobs:
           echo -e "VCPKG_BINARY_SOURCES=$VCPKG_BINARY_SOURCES
           VCPKG_ROOT=$PLATFORM_VCPKG_ROOT" | tee -a $GITHUB_ENV
           cmake -P cpp/CMake/CpuCount.cmake | sed 's/^-- //' | tee -a $GITHUB_ENV
+          echo "ARCTICDB_CODE_COVERAGE_BUILD=1" | tee -a $GITHUB_ENV
         env:
           CMAKE_BUILD_PARALLEL_LEVEL: ${{vars.CMAKE_BUILD_PARALLEL_LEVEL}}
 
@@ -63,7 +64,6 @@ jobs:
           configurePreset: ${{env.ARCTIC_CMAKE_PRESET}}
           buildPreset: ${{env.ARCTIC_CMAKE_PRESET}}
         env:
-          ARCTICDB_CODE_COVERAGE_BUILD: 1
           ARCTICDB_DEBUG_FIND_PYTHON: ${{vars.ARCTICDB_DEBUG_FIND_PYTHON}}
           python_impl_name: 'cp311'
         
@@ -80,36 +80,20 @@ jobs:
       # and this python for the rest of the testing
       - name: Select Python (Linux)
         run: echo /opt/python/cp36-cp36m/bin >> $GITHUB_PATH
-      
-      - name: Test with pytest
-        shell: bash -l {0}
+
+      - name: Install local dependencies with pip
+        shell: bash
         run: |
-          yum install nodejs npm -y
-          npm config set strict-ssl false
-          npm install -g azurite
           python -m pip install --upgrade pip
+          ARCTIC_CMAKE_PRESET=skip pip install -ve .[Testing]
 
-          python >build_tooling/requirements-analysis-workflow.txt <<END
-          from configparser import ConfigParser
-          cf = ConfigParser()
-          cf.read("setup.cfg");
-
-          # setup_requires is not needed
-          print(cf['options']['setup_requires'])
-          print(cf['options']['install_requires'])
-          print(cf['options.extras_require']['Testing'])
-          END
-
-          python -m pip install -r build_tooling/requirements-analysis-workflow.txt
-          python -m grpc_tools.protoc -Icpp/proto --python_out=./python cpp/proto/arcticc/pb2/*.proto
-          cd python
-          ln -s ../cpp/out/linux-debug-build/arcticdb/arcticdb_ext.cpython-36m-x86_64-linux-gnu.so
-          python -m coverage run -m pytest tests
-        env:
-          TEST_OUTPUT_DIR: ${{runner.temp}}
-          ARCTICDB_CODE_COVERAGE_BUILD: 1
-          # Use the Mongo created in the service container above to test against
-          CI_MONGO_HOST: mongodb
+      - name: Test with pytest
+        uses: ./.github/actions/run_local_pytest
+        with:
+          build_type: debug
+          threads: 1
+          fast_tests_only: 0
+          other_params: '-m coverage run '
 
       - name: Get python Coverage report
         shell: bash -l {0}
@@ -201,6 +185,7 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.ARCTICDB_DEV_WEBHOOK_URL }}
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+
   start_ec2_runner:
     uses: ./.github/workflows/ec2_runner_jobs.yml
     secrets: inherit
@@ -242,35 +227,20 @@ jobs:
           ls /opt/python  
           echo /opt/python/cp36-cp36m/bin >> $GITHUB_PATH
 
-      - name: Install dependencies
-        shell: bash -el {0}
+      - name: Install local dependencies with pip
+        shell: bash
         run: |
           python -m pip install --upgrade pip
-
-          python >build_tooling/requirements-analysis-workflow.txt <<END
-          from configparser import ConfigParser
-          cf = ConfigParser()
-          cf.read("setup.cfg");
-
-          # setup_requires is not needed
-          print(cf['options']['setup_requires'])
-          print(cf['options']['install_requires'])
-          print(cf['options.extras_require']['Testing'])
-          END
-
-          python -m pip install -r build_tooling/requirements-analysis-workflow.txt
-          python -m grpc_tools.protoc -Icpp/proto --python_out=./python cpp/proto/arcticc/pb2/*.proto
-          
-          git config --global --add safe.directory .
+          ARCTIC_CMAKE_PRESET=skip pip install -ve .[Testing]
 
       - name: Benchmark all commits
         if: inputs.run_all_benchmarks == true
         shell: bash -el {0}
         run: |
+          git config --global --add safe.directory .
           python -m asv machine -v --yes --machine ArcticDB-Runner-Medium
           python -m asv run -v --show-stderr HASHFILE:python/hashes_to_benchmark.txt
         env:
-          CMAKE_BUILD_PARALLEL_LEVEL: 8
           SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
           VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
           VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
@@ -280,14 +250,30 @@ jobs:
           CMAKE_CXX_COMPILER_LAUNCHER: sccache
 
       - name: Benchmark only latest commit
-        if: inputs.run_all_benchmarks == false
+        if: github.event_name != 'pull_request' && inputs.run_all_benchmarks == false
         shell: bash -el {0}
         run: |
+          git config --global --add safe.directory .  
           python -m asv machine -v --yes --machine ArcticDB-Runner-Medium
           python -m asv run -v --show-stderr HEAD^!
           git log HEAD^..HEAD --oneline -n1 --decorate=no | awk '{print $1;}' >> python/hashes_to_benchmark.txt
         env:
-          CMAKE_BUILD_PARALLEL_LEVEL: 8
+          SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
+          VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
+          VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
+          VCPKG_MAN_NUGET_USER: ${{secrets.VCPKG_MAN_NUGET_USER}} # For forks to download pre-compiled dependencies from the Man repo
+          VCPKG_MAN_NUGET_TOKEN: ${{secrets.VCPKG_MAN_NUGET_TOKEN}}
+          CMAKE_C_COMPILER_LAUNCHER: sccache
+          CMAKE_CXX_COMPILER_LAUNCHER: sccache
+
+      - name: Benchmark against master
+        if: github.event_name == 'pull_request' && inputs.run_all_benchmarks == false
+        shell: bash -el {0}
+        run: |
+          git config --global --add safe.directory .
+          python -m asv machine -v --yes --machine ArcticDB-Runner-Medium
+          python -m asv continuous -v --show-stderr origin/master HEAD -f 1.15
+        env:
           SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
           VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
           VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
@@ -297,6 +283,7 @@ jobs:
           CMAKE_CXX_COMPILER_LAUNCHER: sccache
 
       - name: Publish results
+        if: github.event_name != 'pull_request' 
         shell: bash -el {0}
         run: |
           git config --global user.email "arcticdb@man.com"
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -241,6 +241,7 @@ jobs:
           python -m asv machine -v --yes --machine ArcticDB-Runner-Medium
           python -m asv run -v --show-stderr HASHFILE:python/hashes_to_benchmark.txt
         env:
+          # this is potentially overflowing the cache, so should be looked into after we address issue #1057
           SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
           VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
           VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
@@ -258,6 +259,7 @@ jobs:
           python -m asv run -v --show-stderr HEAD^!
           git log HEAD^..HEAD --oneline -n1 --decorate=no | awk '{print $1;}' >> python/hashes_to_benchmark.txt
         env:
+          # this is potentially overflowing the cache, so should be looked into after we address issue #1057
           SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
           VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
           VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
@@ -274,6 +276,7 @@ jobs:
           python -m asv machine -v --yes --machine ArcticDB-Runner-Medium
           python -m asv continuous -v --show-stderr origin/master HEAD -f 1.15
         env:
+          # this is potentially overflowing the cache, so should be looked into after we address issue #1057
           SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
           VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
           VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,6 @@ jobs:
           # Please declare any key you added below in build_steps.yaml's dummy matrix as well to aid the linting tools
           - linux_matrix:
               - os: linux
-                distro: ubuntu-latest
                 cmake_preset_prefix: linux
                 cibw_build_suffix: manylinux_x86_64
                 build_dir: /tmp/cpp_build
@@ -64,7 +63,6 @@ jobs:
                     - /:/mnt
             windows_matrix:
               - os: windows
-                distro: windows-latest
                 cmake_preset_prefix: windows-cl
                 cibw_build_suffix: win_amd64
                 build_dir: C:/cpp_build
@@ -145,22 +143,8 @@ jobs:
       matrix: ${{toJson(matrix.matrix_override)}}
       python_deps_ids: ${{toJson(matrix.python_deps_ids)}}
 
-  leader-compile-linux:
-    # First do the C++ core compilation using one Python version to seed the compilation caches (and fail quicker)
+  cpp-test-linux:
     needs: [cibw_docker_image, common_config]
-    name: Linux Compile
-    uses: ./.github/workflows/build_steps.yml
-    secrets: inherit
-    permissions: {packages: write}
-    with:
-      job_type: leader-compile
-      cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
-      cibw_image_tag: ${{needs.cibw_docker_image.outputs.tag}}
-      matrix: ${{needs.common_config.outputs.linux_matrix}}
-
-  leader-cpp-test-linux:
-    # Compile and run the C++ tests separately concurrently with the following job
-    needs: [leader-compile-linux, cibw_docker_image, common_config]
     name: Linux C++ Tests
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
@@ -169,10 +153,11 @@ jobs:
       cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
       cibw_image_tag: ${{needs.cibw_docker_image.outputs.tag}}
       matrix: ${{needs.common_config.outputs.linux_matrix}}
+      distro: compile-on-ec2
 
-  follower-linux:
+  build-python-wheels-linux:
     # Then use the cached compilation artifacts to build other python versions concurrently in cibuildwheels
-    needs: [leader-compile-linux, cibw_docker_image, common_config]
+    needs: [cibw_docker_image, common_config]
     strategy:
       fail-fast: false
       matrix:
@@ -197,7 +182,7 @@ jobs:
     secrets: inherit
     permissions: {packages: write}
     with:
-      job_type: follower
+      job_type: build-python-wheels
       python3: ${{matrix.python3}}
       cibw_image_tag: ${{needs.cibw_docker_image.outputs.tag}}
       cibw_version: ${{needs.common_config.outputs.cibuildwheel_ver}}
@@ -205,20 +190,10 @@ jobs:
       matrix: ${{toJson(matrix.matrix_override)}}
       python_deps_ids: ${{toJson(matrix.python_deps_ids)}}
       persistent_storage: ${{inputs.persistent_storage}}
-  
-  leader-compile-windows:
-    needs: [common_config]
-    name: Windows Compile
-    uses: ./.github/workflows/build_steps.yml
-    secrets: inherit
-    permissions: {packages: write}
-    with:
-      job_type: leader-compile
-      cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
-      matrix: ${{needs.common_config.outputs.windows_matrix}}
+      distro: compile-on-ec2
 
-  leader-cpp-test-windows:
-    needs: [leader-compile-windows, common_config]
+  cpp-test-windows:
+    needs: [common_config]
     name: Windows C++ Tests
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
@@ -226,9 +201,10 @@ jobs:
       job_type: cpp-tests
       cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
       matrix: ${{needs.common_config.outputs.windows_matrix}}
+      distro: windows-latest
 
-  follower-windows:
-    needs: [leader-compile-windows, common_config]
+  build-python-wheels-windows:
+    needs: [common_config]
     strategy:
       fail-fast: false
       matrix:
@@ -238,15 +214,16 @@ jobs:
     secrets: inherit
     permissions: {packages: write}
     with:
-      job_type: follower
+      job_type: build-python-wheels
       python3: ${{matrix.python3}}
       cibw_version: ${{needs.common_config.outputs.cibuildwheel_ver}}
       cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
       matrix: ${{needs.common_config.outputs.windows_matrix}}
       persistent_storage: ${{ inputs.persistent_storage }}
+      distro: windows-latest
 
   persistent_storage_verify_linux:
-    needs: [common_config, follower-linux, follower-windows]
+    needs: [common_config, build-python-wheels-linux, build-python-wheels-windows]
     if: inputs.persistent_storage == true
     strategy:
       fail-fast: false
@@ -268,7 +245,7 @@ jobs:
       python_deps_ids: ${{toJson(matrix.python_deps_ids)}}
 
   persistent_storage_verify_windows:
-    needs: [common_config, follower-windows, follower-linux]
+    needs: [common_config, build-python-wheels-windows, build-python-wheels-linux]
     if: inputs.persistent_storage == true
     strategy:
       fail-fast: false
@@ -299,7 +276,7 @@ jobs:
       job_type: cleanup
 
   can_merge:
-    needs: [leader-cpp-test-linux, leader-cpp-test-windows, follower-linux, follower-windows, persistent_storage_verify_linux, persistent_storage_verify_windows]
+    needs: [cpp-test-linux, cpp-test-windows, build-python-wheels-linux, build-python-wheels-windows, persistent_storage_verify_linux, persistent_storage_verify_windows]
     if: |
       always() &&
       !failure() &&
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -5,9 +5,10 @@ on:
       job_type:          {required: true, type: string, description: Selects the steps to enable}
       cmake_preset_type: {required: true, type: string, description: release/debug}
       matrix:            {required: true, type: string, description: JSON string to feed into the matrix}
+      distro:            {required: true, type: string, description: Label of the runner that will be used for running the workflow}
       cibw_image_tag:    {required: false, type: string, description: Linux only. As built by cibw_docker_image.yml workflow}
-      cibw_version:      {required: false, type: string, description: Follower only. Must match the cibw_image_tag}
-      python_deps_ids:   {default: '[""]', type: string, description: Follower test matrix parameter. JSON string.}
+      cibw_version:      {required: false, type: string, description: build-python-wheels only. Must match the cibw_image_tag}
+      python_deps_ids:   {default: '[""]', type: string, description: build-python-wheels test matrix parameter. JSON string.}
       python3:           {default: -1, type: number, description: Python 3 minor version}
       persistent_storage:      {default: "false", type: string, description: Specifies whether the python tests should tests against real storages e.g. AWS S3  }
       # TODO: Maybe we should add these as GitHub variables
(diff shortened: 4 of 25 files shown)


.github/workflows/build.yml
```diff
@@ -29,7 +29,15 @@ concurrency:
...
 jobs:
+  start_linux_ec2_runner:
+    uses: ./.github/workflows/ec2_runner_jobs.yml
+    secrets: inherit
+    permissions: write-all
+    with:
+      job_type: start
+
   common_config:
+    needs: [start_linux_ec2_runner]
```
qc00: I presume the EC2 instance will take a while to start. Is it really necessary to block the entire build?

It's better to pass the `distro` as a separate `workflow_call.input` to `build_steps`. It will also avoid having to duplicate all the matrix settings below, which might cause confusion.

.github/workflows/build.yml
```diff
@@ -150,19 +172,22 @@ jobs:
...
+  #   uses: ./.github/workflows/build_steps.yml
+  #   secrets: inherit
+  #   with:
+  #     job_type: follower
+  #     cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
+  #     cibw_image_tag: ${{needs.cibw_docker_image.outputs.tag}}
+  #     matrix: ${{needs.common_config.outputs.linux_matrix}}
+  #     python3: '${{matrix.python3}}'
+  #     cibw_version: ${{needs.common_config.outputs.cibuildwheel_ver}}
+  #     persistent_storage: ${{inputs.persistent_storage}}
```
qc00: Do not commit commented-out code.

.github/workflows/build.yml
```diff
@@ -199,6 +224,17 @@ jobs:
       matrix: ${{toJson(matrix.matrix_override)}}
       python_deps_ids: ${{toJson(matrix.python_deps_ids)}}
       persistent_storage: ${{inputs.persistent_storage}}
+
+  stop-ec2-runner:
+    needs: [start_linux_ec2_runner, follower-linux]
```
qc00, line 224: You only need to wait until `leader-compile-linux` to finish, isn't it?

Also, if the user wants to rerun just the linux compile leader, it will never work because the server has been stopped

.github/workflows/build_steps.yml
```diff
@@ -120,22 +120,22 @@ jobs:
           buildPreset: ${{env.ARCTIC_CMAKE_PRESET}}
 
       - name: Compile C++ tests
-        if: inputs.job_type == 'cpp-tests'
+        if: inputs.job_type != 'follower'
```
qc00, line 123: As long as the Github cache space is still being overrun, it's not worth blocking all the followers to compile the tests in the leader. They will most likely end up being re-compiled by the cpp-tests follower again.

Actual compilation issues in the C++ tests are also rare, so the early feedback opportunity is probably not worth the wait.

.github/workflows/build_steps.yml
```diff
@@ -120,22 +120,22 @@ jobs:
...
-        run: |
-          cd $ARCTICDB_BUILD_DIR/*-build
-          du -m --max-depth=2 | sort -n | tail -n 50
-          nohup rm -rf * &  # Clean up while the test is running to save build time
+      # - name: Clean up build dir # To save disk space as we're close to the limit on Linux
+      #   if: inputs.job_type != 'follower' && matrix.os == 'linux'
+      #   run: |
+      #     cd $ARCTICDB_BUILD_DIR/*-build
+      #     du -m --max-depth=2 | sort -n | tail -n 50
+      #     nohup rm -rf * &  # Clean up while the test is running to save build time
```
qc00, line 131: Do not commit commented-out code.

You can add a condition to the `if` that's always false or remove it.

.github/workflows/build_steps.yml
```diff
@@ -146,6 +146,41 @@ jobs:
         env:
           ARCTICDB_memory_loglevel: INFO
 
+      - name: Select Python (Linux)
```
qc00, line 149: Why not call the actual `python_tests` job instead of hand-rolling another?

.github/workflows/build_steps.yml
```diff
@@ -120,22 +120,22 @@ jobs:
...
+      # - name: Clean up build dir # To save disk space as we're close to the limit on Linux
+      #   if: inputs.job_type != 'follower' && matrix.os == 'linux'
+      #   run: |
+      #     cd $ARCTICDB_BUILD_DIR/*-build
+      #     du -m --max-depth=2 | sort -n | tail -n 50
+      #     nohup rm -rf * &  # Clean up while the test is running to save build time
 
       - name: C++ Rapidcheck
-        if: inputs.job_type == 'cpp-tests'
+        if: inputs.job_type != 'follower'
```
qc00, line 134: Avoid bundling the tests with the compilation:

- If any of the tests fail spuriously and the user wants just to rerun the test, they'd end up having to rerun everything.
- If one test fails, it will fail the rest of the build, so the user will not have a chance to learn what other tests will fail, esp. for other Python versions, which actually makes the feedback cycle longer. (Note our tests tend to seg fault the entire test runner.)
- Also, some tests are meant to take some time (hypothesis, stress), so running them here will block the followers.

.github/workflows/ec2_runner_jobs.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         with:
           mode: start
           github-token: ${{ secrets.ARCTICDB_TEST_PAT }}
-          ec2-image-id: ami-07ee6ba755b7f58f6
+          ec2-image-id: ami-08b85498b04e33a4a
```
qc00, line 36: Please document all magic numbers

python/tests/integration/arcticdb/test_config.py
```diff
@@ -7,7 +7,7 @@
 _LEVELS = tuple(LogLevel.__entries)
 
 
-@pytest.mark.parametrize("level", _LEVELS)
+@pytest.mark.parametrize("level", list(_LEVELS))
```
qc00, line 10: Why? Typically, the argument only needs to be iterable?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -16,6 +17,8 @@
 from arcticdb_ext.exceptions import InternalException, NormalizationException, SortingException
 from arcticdb_ext import set_config_int
 
+random.seed(42)
```
qc00, line 20: It is more beneficial for the tests to be random even if they're a bit flaky. We found so many bugs by having random data.

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1017,9 +1017,16 @@ def test_filter_numeric_isnotin_hashing_overflow(lmdb_version_store):
...
-@pytest.mark.parametrize("uint64_in", ("df", "vals") if PANDAS_VERSION >= Version("1.2") else ("vals",))
+@pytest.mark.parametrize("op", ["in", "not in"])
+@pytest.mark.parametrize("signed_type", [np.int8, np.int16, np.int32, np.int64])
+@pytest.mark.parametrize(
+    "uint64_in",
+    ["df", "vals"]
+    if PANDAS_VERSION >= Version("1.2")
+    else [
+        "vals",
+    ],
```
qc00, line 1028: God, this black formatting is impossible to read. Could you not turn them into lists...

qc00 (changes requested):

python/tests/unit/arcticdb/version_store/test_append.py
G-D-Petrov: I added a check to seed them only in a specific case, by default we will not seed them.
I need the seed to use [pytest-xdist](https://pytest-xdist.readthedocs.io/en/latest/known-limitations.html) to parallelize the tests.

.github/workflows/build_steps.yml
G-D-Petrov: I am having some problems with running the python_tests job on the ec2 machines.
Also with the way it is design, it would run slower on the ec2 machines, than this solution.

qc00: Could you share what problems you saw?

> Also with the way it is design, it would run slower on the ec2 machines, than this solution.

I presume this is because each job has to wait in turn for the only runner instance. To resolve this, you can use Docker to start multiple Github Action runners on one machine.

Also, another thing that the existing `python_tests` job takes care of is to have a realistic environment that's fully isolated from the build directories and each other. For example, it will install the actual packaged wheel in a fresh venv.

.github/workflows/ec2_runner_jobs.yml
```diff
@@ -5,6 +5,10 @@ on:
       job_type:          {required: true, type: string, description: Selects the steps to enable}
       label:             {type: string, description: the label of the machine that was create by the start-runner job needed by the stop-runner job}
       ec2-instance-id:   {type: string, description: the ec2-instance-id of the machine that was create by the start-runner job needed by the stop-runner job}
+      image-id:          {default: ami-08b85498b04e33a4a, type: string, description: ID of the Amazon Image that should be used as a base for the runner}
```
qc00, line 8: My comment about the magic numbers is more about what's changed in the actual content, i.e. why do we have to switch from `ami-07ee6ba755b7f58f6` to `ami-08b85498b04e33a4a`, not what these fields mean.

python/tests/unit/arcticdb/version_store/test_append.py
qc00: I still failed to see why `pytest-xdist` would require you to use a fixed seed?

.github/workflows/build.yml
```diff
@@ -36,6 +36,13 @@ jobs:
       cibuildwheel_ver: "2.12.1"
       force_update: false
 
+  start_linux_ec2_runner:
+    uses: ./.github/workflows/ec2_runner_jobs.yml
+    secrets: inherit
+    permissions: write-all
```
qc00, line 42: Does it actually need this permission? This will be trusting any third-party action (e.g. `machulav/ec2-github-runner`) called to do *anything* to our repo. AFAI can see, it shouldn't need any repo-level permission.

.github/workflows/build.yml
```diff
@@ -92,6 +99,7 @@ jobs:
       cmake_preset_type_resolved: ${{inputs.cmake_preset_type != '-' && inputs.cmake_preset_type || env.CMAKE_PRESET_TYPE}}
       linux_matrix: ${{toJson(matrix.linux_matrix)}}
       windows_matrix: ${{toJson(matrix.windows_matrix)}}
+      linux_leader_matrix: ${{toJson(matrix.linux_leader_matrix)}}
```
qc00, line 102: Redundant

.github/workflows/build.yml
```diff
@@ -157,18 +165,20 @@ jobs:
...
     name: Linux C++ Tests
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
+    permissions: {packages: write}
     with:
       job_type: cpp-tests
       cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
       cibw_image_tag: ${{needs.cibw_docker_image.outputs.tag}}
       matrix: ${{needs.common_config.outputs.linux_matrix}}
+      distro: ubuntu-latest
```
qc00, line 181: Please also remove the now redundant `distro` fields from `common_config` and [here](https://github.com/man-group/ArcticDB/blob/64d5e738b275dfa243527f5723c62861daa39053/.github/workflows/build_steps.yml#L25) to avoid confusion.

python/tests/unit/arcticdb/version_store/test_append.py
G-D-Petrov: From the xdist documentation
```
Order and amount of test must be consistent[](https://pytest-xdist.readthedocs.io/en/latest/known-limitations.html#order-and-amount-of-test-must-be-consistent)
It is not possible to have tests that differ in order or their amount across workers.
```

Because the parameters are part of each individual test signature, xdist thinks that different workers are seeing different test cases.

.github/workflows/build_steps.yml
G-D-Petrov: I think that I was having some problems with building the wheels, I will try to reproduce later today.

Yeas, each job would have to wait, but also just building and installing the wheel takes around as much time as just running the tests like I have done here.
I am still in favor of doing it this way, as would be simpler and faster.

.github/workflows/build.yml
G-D-Petrov: It needs the permission to register the runner.
As each runner is created and destroyed for each build, it needs this permission to register it as a self-hosted runner for GHA.

.github/workflows/ec2_runner_jobs.yml
G-D-Petrov: The newer image has some new dependencies installed/upgrade for existing ones.

.github/workflows/build_steps.yml
```diff
@@ -119,17 +119,45 @@ jobs:
...
+          
+      - name: Test with pytest
+        if: inputs.job_type == 'leader-compile' && matrix.os == 'linux'
+        run: |
+          yum install nodejs npm -y
+          npm config set strict-ssl false
+          npm install -g azurite
+          python -m pip install --upgrade pip
+
+          python >build_tooling/requirements-analysis-workflow.txt <<END
```
poodlewars, line 134: This will be the third copy of this code :(

.github/workflows/build_steps.yml
poodlewars: I think Georgi's intent with this PR is to introduce a fast "canary" build. It builds on one Python and runs the tests on it as fast as it can. Once it passes, developers can be quite confident their changes are fine, and move on to something else, without waiting for all the other things in the build. I think it makes sense.

.github/workflows/build_steps.yml
G-D-Petrov: yeah if we agree to do it this way, I will refactor it to a local action, similar to the one for the persistent storage vars

.github/workflows/build_steps.yml
qc00: > I think Georgi's intent with this PR is to introduce a fast "canary" build

That is pretty clear from the code; however, this design only speeds up these discoveries:

1. The Python module doesn't load
2. Some test(s) fail for the chosen Python version

But it will slow down/delay the discovery of these:

3. Some test(s) fail in other Python version(s) jobs
4. Some test(s) become flaky


Based on my experience, most people will run the relevant tests locally during development, and it's rare that code changes in one area break a test in a completely unexpected area, so speeding up 2 provides limited benefits.

However, under this "single-job on the fast machine" design, if we get a flaky test, it's no longer possible to retry the tests without restarting the whole build (including the Windows parts) because of how the fast build machine is started/stopped. With a bit of refactoring, it should be possible to run one branch of the existing test jobs on it and retain the ability to retry just one kind of test.

python/tests/unit/arcticdb/version_store/test_append.py
qc00: So this is to make those `gen_param...` functions work?

That is a pretty bad pattern anyway, given it will break test framework integration in all kinds of tooling.

Could you log an issue to replace those with Hypothesis?

Also, xdist only needs these randomly generated test parameters to be the same across all forks, so you can set an environment variable to a timestamp and then seed with that in Python. This will at least allow us to get different test cases from different runs of the job.

It will also be more clear to only fix the seed in those `gen_param...` functions and reset afterwards. You can easily write a decorator to do that.

G-D-Petrov: yeah that makes sense, I will try it

python/tests/util/mark.py
```diff
@@ -28,6 +35,10 @@
 # !!!!!!!!!!!!!!!!!!!!!! Below mark (variable) names should reflect where they will be used, not what they do.
 # This is to avoid the risk of the name becoming out of sync with the actual condition.
 
+LINUX_CONDA_MARK = pytest.mark.skipif(LINUX_CONDA_BUILD, reason=_LINUX_CONDA_BUILD_SKIP_REASON)
+
+FAST_TESTS_MARK = pytest.mark.skipif(FAST_TESTS_ONLY, reason="Skipping test as it takes a ong time to run")
```
poodlewars, line 40:
```suggestion
FAST_TESTS_MARK = pytest.mark.skipif(FAST_TESTS_ONLY, reason="Skipping test as it takes a long time to run")
```

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -23,7 +23,10 @@
 from hypothesis import assume, given, settings
 from hypothesis.extra.pandas import column, data_frames, range_indexes
 
+from tests.util.mark import LINUX_CONDA_MARK
 
+
+@LINUX_CONDA_MARK
```
poodlewars, line 29: Any ideas what's going wrong with this one? We should definitely at least have a ticket to come back and fix it...

python/tests/util/mark.py
```diff
@@ -45,6 +56,27 @@
 Currently controlled by the ARCTICDB_PERSISTENT_STORAGE_TESTS and ARCTICDB_FAST_TESTS_ONLY env vars."""
 
 
+def random_seed_decorator():
```
poodlewars, line 59: Can't we get rid of a layer of nesting if you make `random_seed_decorator` take an argument `func`, and decorate methods not with `@random_seed_decorator` but with `@random_seed_decorator`?

.github/actions/run_local_pytest/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        python -m pip install --upgrade pip
+
+        python >build_tooling/requirements-analysis-workflow.txt <<END
+        from configparser import ConfigParser
+        cf = ConfigParser()
+        cf.read("setup.cfg");
+
+        # setup_requires is not needed
+        print(cf['options']['setup_requires'])
+        print(cf['options']['install_requires'])
```
qc00, line 24: Have you tried `pip install -v .`? You shouldn't need to do any of these by hand.

Also, `setup_requires` and `install_requires` are only used by `pip` internally and shouldn't appear in the "runtime" environment. I.e. a real user doing `pip install arcticdb` won't have those dependencies.

.github/actions/run_local_pytest/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        python -m pip install -r build_tooling/requirements-analysis-workflow.txt
+        python setup.py protoc --build-lib python
+
+    - name: Run local pytests
+      if: inputs.job_type == 'run'
+      shell: bash
+      run: |
+        yum install nodejs npm -y
+        npm config set strict-ssl false
+        npm install -g azurite
```
qc00, line 37: Above three lines should be baked into the container (if not already).

.github/actions/run_local_pytest/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        
+        cd python
+        ln -s ../cpp/out/linux-${{ inputs.build_type }}-build/arcticdb/arcticdb_ext.cpython-36m-x86_64-linux-gnu.so
+        export ARCTICDB_RAND_SEED=$RANDOM
+        python ${{inputs.other_params}}-m pytest -n ${{ inputs.threads }} tests
```
qc00, line 42: Minor: the shell doesn't mind if you use multiple spaces to separate arguments, so you could have a space before the `-m`

python/tests/unit/arcticdb/version_store/test_aggregation.py
G-D-Petrov: I have no idea, I try to debug it next week when I have better access to conda on the new laptops.
I will create an issue about this in the mean time.

.github/actions/run_local_pytest/action.yml
G-D-Petrov: wouldn't `pip install -v .` trigger a rebuild? the point is to only install packages that we depend on, without recompiling because, presumably, we have already compiled and with specific definitions/presets

I agree about `setup_requires` but I think that we need `install_requires`, those are the packages that we depend on, no?

.github/workflows/build_steps.yml
G-D-Petrov: The test step has been added below. and it delaying the execution of the wheel builds.
This is what we had agreed on when we last discussed it. because this way it would be the fastest (wouldn't require any more recompiles) and it would stop the build as fast as possible, if there are any failing tests.
And it is not actually slowing down the total build as that is bound by the windows builds.
 
Do you want me to add it as a separate, non blocking job, similar to the C++ tests?

.github/actions/run_local_pytest/action.yml
G-D-Petrov: Okay thanks, I will experiment with this

.github/actions/run_local_pytest/action.yml
G-D-Petrov: i don't think that this is in the scope of this task, I have opened a new issue to track this [here](https://github.com/man-group/ArcticDB/issues/1036)

qc00: I am not sure how Alma Linux would help? It certainly wouldn't have `azurite` out of the box.

You can just move these three lines [here](https://github.com/man-group/ArcticDB/blob/master/build_tooling/build_many_linux_image.sh#L47)

.github/workflows/ec2_runner_jobs.yml
```diff
@@ -2,9 +2,13 @@ name: __ec2_runner
...
-      ec2-instance-id:   {type: string, description: the ec2-instance-id of the machine that was create by the start-runner job needed by the stop-runner job}
+      job_type:           {required: true, type: string, description: Selects the steps to enable}
+      label:              {type: string, description: the label of the machine that was create by the start-runner job needed by the stop-runner job}
+      ec2-instance-id:    {type: string, description: the ec2-instance-id of the machine that was create by the start-runner job needed by the stop-runner job}
+      image-id:           {default: ami-0936679d3002e5be3, type: string, description: ID of the Amazon Image that should be used as a base for the runner}
+      instance-type:      {default: m6i.4xlarge, type: string, description: Instace type that will be used to create the runner VM}
+      subnet-id:          {default: subnet-04ee7771eea5d61d9, type: string, description: VPC Subnet to be used when crating the runner}
+      security-group-id:  {default: sg-08bd5ce9b15ba4f4b, type: string, description: Security Group to be used when crating the runner}
```
poodlewars, line 11:
```suggestion
      security-group-id:  {default: sg-08bd5ce9b15ba4f4b, type: string, description: Security Group to be used when creating the runner}
```

.github/workflows/ec2_runner_jobs.yml
```diff
@@ -2,9 +2,13 @@ name: __ec2_runner
...
-      label:             {type: string, description: the label of the machine that was create by the start-runner job needed by the stop-runner job}
-      ec2-instance-id:   {type: string, description: the ec2-instance-id of the machine that was create by the start-runner job needed by the stop-runner job}
+      job_type:           {required: true, type: string, description: Selects the steps to enable}
+      label:              {type: string, description: the label of the machine that was create by the start-runner job needed by the stop-runner job}
+      ec2-instance-id:    {type: string, description: the ec2-instance-id of the machine that was create by the start-runner job needed by the stop-runner job}
+      image-id:           {default: ami-0936679d3002e5be3, type: string, description: ID of the Amazon Image that should be used as a base for the runner}
+      instance-type:      {default: m6i.4xlarge, type: string, description: Instace type that will be used to create the runner VM}
+      subnet-id:          {default: subnet-04ee7771eea5d61d9, type: string, description: VPC Subnet to be used when crating the runner}
```
poodlewars, line 10:
```suggestion
      subnet-id:          {default: subnet-04ee7771eea5d61d9, type: string, description: VPC Subnet to be used when creating the runner}
```

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -24,6 +24,7 @@
 from hypothesis.extra.pandas import column, data_frames, range_indexes
 
 
+@pytest.mark.xfail(reason="Test fails on linux conda, should be fixed by issue #1035")
```
poodlewars, line 27: The `xfail` should specify the platform right? eg we don't expect this to fail on PyPi builds.

.github/actions/run_local_pytest/action.yml
```diff
@@ -0,0 +1,35 @@
+name: 'Run Local Pytest'
+description: 'Installs Python dependencies and runs Pytest against locally built ArcticDB binaries'
+inputs:
+  job_type:          {default: '', type: string, description: Specifies which steps should be executed}
```
qc00, line 4: Just curious: is this ever not `run`?

python/tests/util/mark.py
```diff
@@ -45,6 +49,17 @@
 Currently controlled by the ARCTICDB_PERSISTENT_STORAGE_TESTS and ARCTICDB_FAST_TESTS_ONLY env vars."""
 
 
+@contextmanager
+def random_seed_context():
```
poodlewars, line 53: Is `mark.py` really the right file to hold this function? It isn't anything to do with marks

.github/workflows/analysis_workflow.yml
```diff
@@ -63,7 +67,6 @@ jobs:
           configurePreset: ${{env.ARCTIC_CMAKE_PRESET}}
```
qc00, line 67: Since you're doing a debug build, please _do not_ use `sccache`. Otherwise, this build will end up wiping out the tiny cache space we have.

environment_unix.yml
```diff
@@ -68,6 +68,7 @@ dependencies:
   # See: https://github.com/man-group/ArcticDB/pull/291
   - hypothesis < 6.73
   - pytest-sugar
+  - pytest-xdist
```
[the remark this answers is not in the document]
poodlewars: Ah OK. Can you put a ticket on our queue for this (exclusively using pytest-xdist in favour of parallel_test.sh)? I'm hoping that once the test fixtures work introduces a prefix feature for Mongo storages, it should be possible.

poodlewars (approved):

.github/workflows/build.yml
```diff
@@ -169,6 +167,7 @@ jobs:
       cmake_preset_type: ${{needs.common_config.outputs.cmake_preset_type_resolved}}
       cibw_image_tag: ${{needs.cibw_docker_image.outputs.tag}}
       matrix: ${{needs.common_config.outputs.linux_matrix}}
+      distro: ec2
```
qc00, line 170: Given the current Mongo fixture can't support more than 2 cores, probably don't need to run all the Python versions on EC2.

.github/workflows/build_steps.yml
```diff
@@ -5,6 +5,7 @@ on:
       job_type:          {required: true, type: string, description: Selects the steps to enable}
       cmake_preset_type: {required: true, type: string, description: release/debug}
       matrix:            {required: true, type: string, description: JSON string to feed into the matrix}
+      distro:            {default: '', type: string, description: Label of the runner that will be used for running the workflow}
```
qc00, line 8: Make it required?

.github/actions/run_local_pytest/action.yml
```diff
@@ -0,0 +1,35 @@
...
+      shell: bash
+      run: |
+        yum install nodejs npm -y
+        npm config set strict-ssl false
+        npm install -g azurite
+        
+        cd python
+        ln -s ../cpp/out/linux-${{ inputs.build_type }}-build/arcticdb/arcticdb_ext.cpython-36m-x86_64-linux-gnu.so
+        export ARCTICDB_RAND_SEED=$RANDOM
+        python ${{inputs.other_params}} -m pytest -n ${{ inputs.threads }} tests
```
qc00, line 29: Why not use the built-in `-n auto`?

.github/workflows/build_steps.yml
```diff
@@ -174,7 +196,7 @@ jobs:
       - name: Discover test directory names
         if: inputs.job_type == 'follower'
         # There are so few nonreg tests, run them in the hypothesis runner
-        run: find python/tests/* -maxdepth 0 -type d ! -regex '.*\(__pycache__\|util\|nonreg\)' -printf '"%f",' |
+        run: find python/tests/* -maxdepth 0 -type d ! -regex '.*\(__pycache__\|util\|nonreg\|scripts\)' -printf '"%f",' |
               sed 's/^/test_dirs=[/ ; s/"hypothesis"/"{hypothesis,nonreg}"/ ; s/,$/]/' | tee -a $GITHUB_ENV
```
qc00, line 200:
```suggestion
              sed 's/^/test_dirs=[/ ; s/"hypothesis"/"{hypothesis,nonreg,scripts}"/ ; s/,$/]/' | tee -a $GITHUB_ENV
```

I.e. we still need to run the tests in the scripts folder.

.github/workflows/build_steps.yml
```diff
@@ -207,8 +229,24 @@ jobs:
...
+      job_type: stop
+      label: ${{ needs.start_ec2_runner.outputs.label }}
+      ec2-instance-id: ${{ needs.start_ec2_runner.outputs.ec2-instance-id }}
+
   python_tests:
-    if: inputs.job_type == 'follower'
+    if: |
+      always() &&
+      !failure() &&
+      !cancelled() &&
```
qc00, line 248: Use `continue-on-error: true` on the previous step instead of these.

You may also want to set an auto-shutdown on all EC2 instances we start, in case any kind of error causes an instance not to get killed.

.github/workflows/build_steps.yml
```diff
@@ -299,6 +337,9 @@ jobs:
           # Use the Mongo created in the service container above to test against
           CI_MONGO_HOST: mongodb
           CI_MONGO_HOST_2: mongodb_2
+          # On windows, we can allocate more threads for running the tests
+          # because we are not testing against mongo
```
qc00, line 341: This is an incorrect assumption/misleading comment.

We did not restrict the tests to two processes because of Mongo. Rather, the current mongo fixture only supports two processes because the default free runner only has two cores.

The four-core machine we sometimes get is a random free upgrade from GitHub and we can't rely on always getting that.

build_tooling/parallel_test.sh
```diff
@@ -33,7 +33,8 @@ function worker() {
     MSYS=winsymlinks:nativestrict ln -s "$(realpath "$tooling_dir/../python/tests")" $new_root/
     cd $new_root
 
-    $catch python -m pytest -v --show-capture=no --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
+    export ARCTICDB_RAND_SEED=$RANDOM
+    $catch python -m pytest -v -n $PYTEST_THREADS --show-capture=no --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
```
qc00, line 37: DO NOT use pytest-xdist (`-n`) with pytest-splits (`--splits`)!!

build_tooling/vcpkg_caching.sh
```diff
@@ -6,7 +6,7 @@ pushd $VCPKG_ROOT
 PLATFORM_VCPKG_ROOT=`cygpath -wa . 2>/dev/null || pwd`
 
 [[ -x vcpkg ]] || ./bootstrap-vcpkg.sh -disableMetrics
-nuget="`which mono 2>/dev/null` `./vcpkg fetch nuget | tail -n 1`"  # which mono will return empty on windows
+nuget="`which mono 2>/dev/null` `./vcpkg fetch nuget | tail -n 1`" # which mono will return empty on windows
```
qc00, line 9: ?

python/tests/util/mark.py
```diff
@@ -27,6 +30,7 @@
 
 # !!!!!!!!!!!!!!!!!!!!!! Below mark (variable) names should reflect where they will be used, not what they do.
 # This is to avoid the risk of the name becoming out of sync with the actual condition.
+FAST_TESTS_MARK = pytest.mark.skipif(FAST_TESTS_ONLY, reason="Skipping test as it takes a long time to run")
```
qc00, line 31: This name is very misleading. Tests with this mark are not fast but actually slow....

python/tests/unit/arcticdb/version_store/test_engine.py
```diff
@@ -14,32 +14,37 @@
...
-    p = [list(range(2, 5))]
-    # rownums
-    periods = 6
-    p.append(list(range(1, periods + 2)))
-    # cols
-    p.append(list(chain(*[list(combinations(["a", "b", "c"], c)) for c in range(1, 4)])))
-    # tsbounds
-    p.append([(j, i) for i in range(periods) for j in range(i)])
-    return random.sample(list(product(*p)), 1000)
+    with random_seed_context():
```
qc00, line 21: Sorry for not going into this detail in my previous comment, but you can use `@contextmanager` functions as decorators:

```python
@random_seed_context
def gen_params():
...
```

should work

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -23,7 +23,13 @@
 from hypothesis import assume, given, settings
 from hypothesis.extra.pandas import column, data_frames, range_indexes
 
+import sys
+from tests.util.mark import ARCTICDB_USING_CONDA
 
+
+@pytest.mark.xfail(
+    ARCTICDB_USING_CONDA and sys.platform == "linux", reason="Test fails on linux conda, should be fixed by issue #1035"
+)
```
qc00, line 32: Minor: might want to declare a wrapping `@xfail_on_linux_conda(reason=...)` function in `mark.py` so we can easily locate these.

.github/actions/run_local_pytest/action.yml
G-D-Petrov: Yeah in the [benchmarks job](https://github.com/man-group/ArcticDB/blob/941-self-hosted-runners-for-ci/.github/workflows/analysis_workflow.yml#L229) because the "run" part is done by ASV

.github/workflows/build.yml
G-D-Petrov: We don't run any tests on the ec2 runners.
We only compile the wheels on ec2 runners, all tests are done on the free GH runners.

.github/actions/run_local_pytest/action.yml
G-D-Petrov: On linux, I want to pass -n 0, which disables xdist entirely because of the mongo tests.

.github/workflows/build_steps.yml
G-D-Petrov: This is addressing a limitation of GH Actions. Essentially because compile depends on the stert_ec2 job, if that step is skipped, GHA skips **all** downstream jobs by default. So we need to add these to explicitly enable them, even if the ster_ec2 job is skipped, e.g. on Windows.

The step for stopping/deleting the ec2 runners is always executed, but I will look into, if there is a way to add some timeout just in case.

.github/workflows/build_steps.yml
G-D-Petrov: In my testing, **running test mongo tests in parallel makes them break consistently**.
That is what my comment is referring to and to the fact that we cannot truly run the tests in parallel before this issue is fixed.
I have opened an issue to fix this limitation.

build_tooling/parallel_test.sh
G-D-Petrov: Why not? The tests seems stable for the most part and this is addressing the problem that:
- we cannot run the test only with xdist (due to mongo) and splits seems to remedy this somewhat 
- xdist is helping speed up the integration tests on windows by bringing the time down from ~100 min to ~20 min

build_tooling/vcpkg_caching.sh
G-D-Petrov: I had to debug this line as it was breaking on the ec2 runners.
I guess that when cleaning up, I deleted a space

python/tests/unit/arcticdb/version_store/test_engine.py
G-D-Petrov: Yes, I know but I had some problem with it being a decorator and also I like this approach better.

.github/actions/run_local_pytest/action.yml
G-D-Petrov: As I have mentioned in the ticket, these dependencies should be added to the container, we don't expect them to be there out-of-the-box.
I don't want to do it in this PR, as it would have impact that is outside of the scope of this task and this PR has already become far too bloated.

.github/actions/run_local_pytest/action.yml
qc00: When the `job_type` is not `run`, this action only does two lines of bash (after our simplification).

It then becomes a bit misleading to call this action, since the caller don't actually want to 'Run Local Pytest'

.github/workflows/build.yml
qc00: So the `distro` is only used by part of the job, which may cause confusion.

I would either use another name for this, e.g. `compile-on-ec2`, or put a comment here.

.github/actions/run_local_pytest/action.yml
qc00: That's fine. I'll address this once my new Mongo fixture is merged.

.github/workflows/build_steps.yml
qc00: What I meant was if you put `continue-on-error: true` on a step that can be skipped, it will not skip subsequent steps anymore, which is much clearer in the intention compared to this `if` condition.

.github/workflows/build_steps.yml
qc00: >In my testing, running test mongo tests in parallel makes them break consistently.

Oh, that might be because you're not running it with `parallel-test.sh`?

The current Mongo fixture relies on [parsing a pytest-split-specific argument](https://github.com/man-group/ArcticDB/blob/master/python/tests/conftest.py#L562) to select one of the two Mongo servers to use.

My new fixture will replace this dodgy hack

build_tooling/parallel_test.sh
qc00: Running two test splitting/parallelisation plugins is certainly an unsupported scenario, and I wouldn't even be confident that it's correctly running all the tests and reporting all the test outcomes.

Also, the effect of the two plugins will multiply, and one might end up running way more tests in parallel by mistake.

For the mongo issue, see my other comment.

python/tests/unit/arcticdb/version_store/test_filtering_dynamic.py
```diff
@@ -242,7 +242,7 @@ def test_filter_numeric_isnotin_unsigned(lmdb_version_store_dynamic_schema, df,
 
 
 @use_of_function_scoped_fixtures_in_hypothesis_checked
-@settings(deadline=None)
+@settings(deadline=None, max_examples=10)
```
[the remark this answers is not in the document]
G-D-Petrov: Okay I will experiment with this

build_tooling/parallel_test.sh
G-D-Petrov: According to the logs, it is behaving as expected.
It first splits the tests and then parallelizes only over the selected tests.

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -81,7 +81,7 @@ def gen_params_append():
         p.append([(j, i) for i in [1, periods - 1] for j in range(i)])
         # append_point
         p.append([k for k in range(1, periods - 1)])
-        result = sorted(random.sample(list(product(*p)), 500))
+        result = sorted(random.sample(list(product(*p)), 10))
```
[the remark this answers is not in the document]
G-D-Petrov: [Done](https://github.com/man-group/ArcticDB/issues/1045)

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -987,17 +987,22 @@ def test_tail(arctic_library):
     )
 
 
-@pytest.mark.parametrize("dedup", [True, False])
-def test_dedup(arctic_client, dedup):
+def test_dedup(arctic_client):
```
poodlewars, line 990: Why this change?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -75,14 +75,19 @@ def test_simple_flow(basic_store_no_symbol_list, symbol):
     assert basic_store_no_symbol_list.list_symbols() == basic_store_no_symbol_list.list_versions() == []
 
 
-@pytest.mark.parametrize("special_char", list("$@=;/:+ ,?\\{^}%`[]\"'~#|!-_.()"))
-def test_special_chars(object_version_store, special_char):
+def test_special_chars(object_version_store):
```
poodlewars, line 78: Why this change?

poodlewars (commented):
I'm happy with this other than the unresolved conversations 👍

python/tests/integration/arcticdb/test_arctic.py
G-D-Petrov: makes the test a bit faster on windows, due to not having to reinit the arctic_client on every option

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
G-D-Petrov: makes the test a bit faster on windows, due to not having to reinit the arctic_client on every option

.github/workflows/build_steps.yml
G-D-Petrov: Looks like continue-on-error is not supported on a job with a uses clause

python/tests/conftest.py
```diff
@@ -56,13 +56,19 @@
...
 from azure.storage.blob import BlobServiceClient, generate_account_sas, ResourceTypes, AccountSasPermissions
 
-
 from tests.util.mark import (
     AZURE_TESTS_MARK,
     MONGO_TESTS_MARK,
     REAL_S3_TESTS_MARK,
 )
 
+import hypothesis as h
```
qc00, line 65: I don't see any value in aliasing `hypothesis` to `h`. This is not a commonly used alias and will just cause confusion.

python/tests/hypothesis/arcticdb/test_append.py
qc00, line 1: I wouldn't move the entire file. Half of the tests are real unit tests.

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
```diff
@@ -313,14 +314,15 @@ def test_batch_read_snapshot(self, snap: str):
...
 def test_stateful(lib_type, request):
     VersionStoreComparison._lib = request.getfixturevalue(lib_type)
     run_state_machine_as_test(
         VersionStoreComparison,
         settings=settings(  # Note: timeout is a legacy parameter
-            max_examples=int(os.getenv("HYPOTHESIS_EXAMPLES", 100)),
+            max_examples=int(os.getenv("HYPOTHESIS_EXAMPLES", 10)),
             deadline=None,
-            stateful_step_count=100,
+            stateful_step_count=10,
```
qc00, line 325: 10 is really too low. This test is designed to spend a few minutes exercising the code.

The last time I touched it, the test should finish in about 2 minutes under the default settings. If that's no longer the case, then we had a performance regression.

python/tests/integration/arcticdb/test_arctic.py
qc00: Creating `Arctic` shouldn't be slow at all. The fact that it's noticeable means we have a performance bug somewhere.

Please log an issue and put the link in a comment on this test.
Please also comment on the test that the manual parameterization is to workaround a performance problem and should not be emulated in other tests.

build_tooling/parallel_test.sh
qc00: If they appear to work, then I am okay with this setup.

After my fixtures PRs is merged, `pytest-split` can be removed.

python/tests/integration/arcticdb/test_arctic.py
G-D-Petrov: created issue #1053 to address this

.github/workflows/analysis_workflow.yml
```diff
@@ -280,14 +250,30 @@ jobs:
...
           git log HEAD^..HEAD --oneline -n1 --decorate=no | awk '{print $1;}' >> python/hashes_to_benchmark.txt
         env:
-          CMAKE_BUILD_PARALLEL_LEVEL: 8
+          SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
+          VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
+          VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
+          VCPKG_MAN_NUGET_USER: ${{secrets.VCPKG_MAN_NUGET_USER}} # For forks to download pre-compiled dependencies from the Man repo
+          VCPKG_MAN_NUGET_TOKEN: ${{secrets.VCPKG_MAN_NUGET_TOKEN}}
+          CMAKE_C_COMPILER_LAUNCHER: sccache
+          CMAKE_CXX_COMPILER_LAUNCHER: sccache
```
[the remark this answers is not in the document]
G-D-Petrov: they are separate steps, so it is not exactly applicable.
ASV builds multiple times in release, so the sccache is very important to speed up the builds.

python/tests/unit/arcticdb/version_store/test_append.py
poodlewars, line 1: I've verified that the stuff being deleted here has all been moved to the `hypothesis/` tree 👍

build_tooling/parallel_test.sh
poodlewars: I've just "unresolved" this thread so it's easy to find later

.github/workflows/analysis_workflow.yml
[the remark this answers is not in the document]
qc00: Please link the sccache issue here as well.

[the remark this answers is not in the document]
G-D-Petrov: Opened issue #1057 to address this

Issue #981, opened 2023-10-19
Make RocksDB and LMDB storages singletons by path.


joe-iddon:
Only a single storage connection to a particular path/URI should be created from a particular Python process for both RocksDB and LMDB. This is already documented [here](https://github.com/man-group/ArcticDB/blob/master/docs/mkdocs/docs/index.md?plain=1#L139), but we are not enforcing this.

This is particularly important for RocksDB where it is smart enough to notice that we are opening a second connection from the same process and will error.

We need to decide whether re-connecting to the same library should either: (a) throw an error in the Python, or should (b) just seamlessly return the same C++ `Storage` object.

If we decide option (a) we can just use RocksDB's already written internal checks to do this. However this is more difficult for LMDB which doesn't really notice.

If we decide option (b), then we need to keep track of the open `Storage` objects, and return back an already-created one to the user if they try access that particularly library again. This is of course most common for the configuration `_arctic_cfg` library which is accessed whenever the user creates an `Arctic()` connection, even if they then go on to access a different library from there. This is probably quite an obscure error that the user will encounter if we don't address this issue properly, since they are probably unaware that they are actually accessing `_arctic_cfg` when they call `Arctic()`.

Anyway, some solutions discussed for implementing option (b) are:
- Solve in Python layer. If we ever support more languages we will have to re-implement different management for each of these.
- Solve with some kind of weak map attached as a static member of the c++ library manager, and re-use rocksdb libraries based off their paths. We would need to add something to global lifetimes to delete any remaining references to storages at shutdown.
- Do something very similar to (2), but subclass the library manager (maybe create a generic library manager interface), so that rocksdb can store its own map of open storages inside their and use proper RAII.

As a temporary work-around, calling `del ac` where `ac` is the `Arctic` instance does successfully destroy the `RocksDBStorage` objects to the `_arctic_cfg` and potentially any other libraries; thus allowing us to re-connect to the same URI with another call to `Arctic()`. This is terrible UX though.

<hr />

Code to trigger error with RocksDB:
```
from arcticdb import Arctic

print('Make first connection')
ac = Arctic('rocksdb:///tmp/rocksdb_db')
ac.get_library('joe', create_if_missing=True)
print('Deleting connection')
del ac
print('Make second connection')
ac = Arctic('rocksdb:///tmp/rocksdb_db')
ac.get_library('joe', create_if_missing=True)
print('Make third connection')
ac = Arctic('rocksdb:///tmp/rocksdb_db')  # error
```
[...]

Issue #889, opened 2023-09-21
Fetching library again causes next write to fail with lmdb


jamesmunro:
### Describe the bug

Fetching library from Arctic instance a second time causes next write to that library to fail.

### Steps/Code to Reproduce

```python
import numpy as np
from arcticdb import Arctic
a = Arctic("lmdb://test.lmdb")
lib = a.get_library('l0', create_if_missing=True)
lib.write('symbol', np.arange(10))
lib = a.get_library('l0')
lib.write('symbol', np.arange(10))
```

Output:
```
[2023-09-21 21:27:45.219] [arcticdb] [warning] Loading versions from storage via ref key failed with error: mdb_txn_begin: Invalid argument for stream symbol. Retrying
Traceback (most recent call last):
  File "/home/james/workspace/arcticdb-cli/weirdbug.py", line 9, in <module>
    lib.write('symbol', pd.DataFrame({'a': [1,2,3]}))
  File "/home/james/miniforge3/envs/cli/lib/python3.11/site-packages/arcticdb/version_store/library.py", line 421, in write
    return self._nvs.write(
           ^^^^^^^^^^^^^^^^
  File "/home/james/miniforge3/envs/cli/lib/python3.11/site-packages/arcticdb/version_store/_store.py", line 571, in write
    vit = self.version_store.write_versioned_dataframe(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
arcticdb_ext.exceptions.InternalException: lmdb::runtime_error(mdb_txn_begin: Invalid argument)
```

### Expected Results

No error

### OS, Python Version and ArcticDB Version

Python: 3.11.5 | packaged by conda-forge | (main, Aug 27 2023, 03:34:09) [GCC 12.3.0]
OS: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
ArcticDB: 3.0.0

### Backend storage used

LMDB

poodlewars:
We should be caching the library instance, due to this LMDB limitation:

```
Threads and Processes

[...]

PR #1000, opened 2023-10-25
Warning for LMDB when library reopened

poodlewars:
#### Reference Issues/PRs

#889 #520 #981 

Also see PR #990 for which this is a followup.

It's only safe to open an LMDB database once from a given process. LMDB document this. It's because they rely on `flock` locking on the files, which is re-entrant within a process. This warns the user if they have multiple live Arctic instances over the same LMDB database.

We also:

- Remove the `psutil` Python dep (which I considered using as an alternative solution here)
- Put the `_arctic_cfg` special string down in the Native extension
- Fix an issue with the `Arctic` API where it needlessly created an extra `NativeVersionStore` when creating a library. The new warning flagged this up.


python/arcticdb/arctic.py
```diff
@@ -285,8 +285,8 @@ def create_library(self, name: str, library_options: Optional[LibraryOptions] =
         if library_options is None:
             library_options = LibraryOptions()
 
-        library = self._library_adapter.create_library(name, library_options)
-        self._library_manager.write_library_config(library._lib_cfg, name, self._library_adapter.get_masking_override())
+        cfg = self._library_adapter.get_library_config(name, library_options)
```
joe-iddon, line 288: Just noting that this will fix the PR for the RocksDB Python API #991

python/tests/integration/arcticdb/test_lmdb.py
```diff
@@ -149,3 +154,32 @@ def test_lmdb_options_unknown_option(options):
...
+    ac = Arctic(f"lmdb://{tmpdir}")
+
+
+def test_warnings_arctic_instance(tmpdir):
+    ac = Arctic(f"lmdb://{tmpdir}")
+    # should warn
+    ac = Arctic(f"lmdb://{tmpdir}")
+
+    del ac
+    # should not warn
```
joe-iddon, line 169: Checking the warning is actually produced is carried out manually then?

python/arcticdb/version_store/_store.py
```diff
@@ -250,14 +249,22 @@ def create_store_from_lib_config(cls, lib_cfg, env, open_mode=OpenMode.DELETE):
         lib = cls.create_lib_from_lib_config(lib_cfg, env, open_mode)
         return cls(library=lib, lib_cfg=lib_cfg, env=env, open_mode=open_mode)
 
-    @classmethod
-    def create_store_from_config(
-        cls, cfg, env, lib_name, open_mode=OpenMode.DELETE, encoding_version=EncodingVersion.V1
+    @staticmethod
+    def create_library_config(
```
joe-iddon, line 253: This makes `arctic.py` much clearer now. There was no need to create a temporary Library just to get its config and delete it.

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
```diff
@@ -316,7 +321,50 @@ LmdbStorage::LmdbStorage(const LibraryPath &library_path, OpenMode mode, const C
...
+}
+
+LmdbStorage::LmdbStorage(LmdbStorage&& other)  noexcept
+    : Storage(std::move(static_cast<Storage&>(other))),
+    write_mutex_(std::move(other.write_mutex_)),
+    env_(std::move(other.env_)),
+    dbi_by_key_type_(std::move(other.dbi_by_key_type_)),
+    lib_dir_(std::move(other.lib_dir_)) {
+    other.lib_dir_ = "";
+}
```
joe-iddon, line 358: Just noting that when Storages are made no move or copy in #991 I'll remove this move constructor. Is that going to be fine? It might mean the check of `!lib_dir_.empy()` in the destructor can go also, because `lib_dir` should always be passed something in the constructor.

joe-iddon (approved):
Presumably we'll have to have a warning for RocksDB too, so we should maybe create an issue to do that?

python/tests/integration/arcticdb/test_lmdb.py
poodlewars: Yeah, checking the logging was super difficult and didn't seem worth it

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
poodlewars: Proof of the pudding will be in the eating, but yeah that sounds good.

Issue #242, opened 2023-04-19
delete_library doesnt work on aliyun S3


markqiu:
`ac.delete_library('my_data1')`

```
[2023-04-19 15:55:42.656] [arcticdb] [warning] Failed to iterate key type with key 'g' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.658] [arcticdb] [warning] Failed to iterate key type with key 'G' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.661] [arcticdb] [warning] Failed to iterate key type with key 'd' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.664] [arcticdb] [warning] Failed to iterate key type with key 'i' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.666] [arcticdb] [warning] Failed to iterate key type with key 'V' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.669] [arcticdb] [warning] Failed to iterate key type with key 'v' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.671] [arcticdb] [warning] Failed to iterate key type with key 'M' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.674] [arcticdb] [warning] Failed to iterate key type with key 's' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.676] [arcticdb] [warning] Failed to iterate key type with key 'l' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.679] [arcticdb] [warning] Failed to iterate key type with key 'r' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.681] [arcticdb] [warning] Failed to iterate key type with key 'h' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.684] [arcticdb] [warning] Failed to iterate key type with key 'a' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.686] [arcticdb] [warning] Failed to iterate key type with key 'm' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.688] [arcticdb] [warning] Failed to iterate key type with key 'x' NoSuchKey: The specified key does not exist.
[2023-04-19 15:55:42.691] [arcticdb] [warning] Failed to iterate key type with key 't' NoSuchKey: The specified key does not exist.
[...]

Issue #133, opened 2023-03-21
delete_library doesnt work


RockScience:
On my side `store.delete_library` silently fails. Is that a known issue?

samratbatth:
I am having the same issue. Can’t delete a library in Python 3.6.

mehertz:
Hi @RockScience @sambatth 

Thanks for your bug report. Sounds like there are two seperate issues here.

@RockScience I cannot reproduce:

```
In [10]: ac.create_library('mehertz_test')

In [11]: ac['mehertz_test']
Out[11]: Library(Arctic(config=S3(bucket=arcticdb-tests)), path=mehertz_test, storage=s3_storage)

In [12]: ac.delete_library('mehertz_test')

In [13]: ac['mehertz_test']
[2023-03-22 09:01:08.289] [arcticdb] [warning] Failed to find segment for key 'C:mehertz_test' : No response body.
---------------------------------------------------------------------------
...
StorageException: Composite: C:mehertz_test,
```

Can you provide the following information:

1. What OS/Distribution are you running on?
2. What Python version are you using?
3. What storage are you using? S3 (If so, AWS?)? LMDB?

Also note it might take a small amount of time for the deletion to be reflected in the next read.

mehertz:
@sambatth I think this is a seperate issue. Could you please create a seperate issue ticket with more information about the code you are running please.

RockScience:
@mehertz thank you. Agreed this is not the same issue as @sambatth 

on my side:

1. What OS/Distribution are you running on?
Linux-5.4.0-1100-azure-x86_64-with-Ubuntu-20.04-focal

2. What Python version are you using?
Python 3.7.16
ArcticDB 1.0.1

3. What storage are you using? S3 (If so, AWS?)? LMDB?
Google S3

poodlewars:
> Getting an error for delete_library(“name”): AttributeError: 'Library' object has no attribute 'version_store'
> 
> Python 3.6. using lmdb

Sorry to muddy the thread as I agree the LMDB issue is separate. However I have reproduced it and have raised this PR to fix - https://github.com/man-group/ArcticDB/pull/139

sambatth:
I am getting the same error while using LMDB.Should I still raise it on a different thread?ThxOn Mar 22, 2023, at 11:29, Alex Seaton ***@***.***> wrote:﻿

Getting an error for delete_library(“name”): AttributeError: 'Library' object has no attribute 'version_store'
[...]

PR #1013, opened 2023-10-30
Only claim to support AWS S3

poodlewars:
Since eg #133 GCP S3 needs work for full support.
Also eg #242 for aliyun.


poodlewars:
I'm going to document this stuff more exhaustively, so I'll decline this PR for now

Issue #1017, opened 2023-10-31
Bug with LMDB re-opening warnings


poodlewars:
```
In [1]: from arcticdb import Arctic
ac = Arctic(f"lmdb:///tmp/one")

In [3]: lib = ac.get_library?

In [4]: lib = ac.get_library("test", create_if_missing=True)

In [5]: ac = Arctic(f"lmdb:///tmp/one")
[2023-10-31 11:09:58.710] [arcticdb] [warning] LMDB path at /tmp/one/ has already been opened in this process which is not supported by LMDB. You should only open a single Arctic instance over a given LMDB path. To continue safely, you should delete this Arctic instance and any others over the LMDB path in this process and then try again. Current process ID=[79057]

In [6]: lib = ac["test"]
[2023-10-31 11:10:09.197] [arcticdb] [warning] LMDB path at /tmp/one/test has already been opened in this process which is not supported by LMDB. To continue safely, you should delete all Library objects over the LMDB path in this process and then try again. This indicates a bug in ArcticDB. Please report at https://github.com/man-group/ArcticDB. Current process ID=[79057]
```

The second bit of logging here is misleading.

PR #1022, opened 2023-10-31
Issue #1017 Only warn if the "base" LMDB env is opened twice

poodlewars:
Otherwise, if a user ignores the first error, they are then spammed with misleading errors on their subsequent interactions.

See the Github issue for a couple of examples that this change seeks to avoid.

#### Reference Issues/PRs
See #1017 and the original PR where this was introduced #1000


joe-iddon:
LGTM. Only warns on opening `_arctic_cfg` libraries, which makes sense.

joe-iddon (approved):
LGTM. Only warns on opening `_arctic_cfg` libraries, which makes sense.

Issue #910, opened 2023-09-28
Remove pymongo dependency


poodlewars:
We use `pymongo` just to parse the MongoDB URI in `mongo_library_adapter`. This was my (bad) idea. We have too many Python dependencies and it makes it hard to install ArcticDB.

Instead we should either:

- Parse and validate the URI by hand in Python
- Hook in to the Mongo C++ SDK to parse

PR #1027, opened 2023-11-01
Make dependency of pymongo optional in running

phoebusm:
#### Reference Issues/PRs

https://github.com/man-group/ArcticDB/issues/910

Make dependency of pymongo optional in running. 
pymongo offers uri_parser which offers fantastic error message for incorrect uri syntax. However it is not necessary. So this PR makes the dependency optional. If user's environment does not come with pymongo, a simple regex will be used for extracting the endpoint. And mongo C++ driver will detect any incorrect syntax still, but with relatively simple error message.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/environment_unix.yml
+++ b/environment_unix.yml
@@ -70,9 +70,9 @@ dependencies:
   - hypothesis < 6.73
   - pytest-sugar
   - pytest-xdist
-  - pymongo
   # Python dependencies (for tests only)
   - azure-storage-blob
   - azure-identity
   - pyarrow
   - asv
+  - pymongo
--- a/python/arcticdb/adapters/mongo_library_adapter.py
+++ b/python/arcticdb/adapters/mongo_library_adapter.py
@@ -14,23 +14,31 @@ from arcticdb.adapters.arctic_library_adapter import ArcticLibraryAdapter, set_l
 from arcticdb_ext.storage import CONFIG_LIBRARY_NAME
 from arcticdb.encoding_version import EncodingVersion
 from arcticdb.exceptions import UserInputException
-from collections import namedtuple
-import pymongo
+import re
 
-PARSED_QUERY = namedtuple("PARSED_QUERY", ["region"])
+try:
+    from pymongo.uri_parser import parse_uri
+
+    _HAVE_PYMONGO = True
+except ImportError:
+    _HAVE_PYMONGO = False
 
 
 class MongoLibraryAdapter(ArcticLibraryAdapter):
     @staticmethod
     def supports_uri(uri: str) -> bool:
-        return uri.startswith("mongodb://")
+        return uri.startswith("mongodb://")  # mongo+srv:// support?
 
     def __init__(self, uri: str, encoding_version: EncodingVersion, *args, **kwargs):
         try:
-            parameters = pymongo.uri_parser.parse_uri(
-                uri
-            )  # also checks pymongo uri syntax, throw exception as early as possible if syntax is incorrect
-            self._endpoint = f"{parameters['nodelist'][0][0]}:{parameters['nodelist'][0][1]}"
+            if _HAVE_PYMONGO:
+                parameters = parse_uri(
+                    uri
+                )  # also checks pymongo uri syntax, throw exception as early as possible if syntax is incorrect
+                self._endpoint = f"{parameters['nodelist'][0][0]}:{parameters['nodelist'][0][1]}"
+            else:
+                match = re.search(r"\/\/(?P<endpoint>[^\/]*)", uri)
+                self._endpoint = match["endpoint"]
         except Exception as e:
             raise UserInputException(
                 f"Invalid connection string format. {e} Correct format: mongodb://[HOST]/[DATABASE][?OPTIONS]"
--- a/python/tests/integration/arcticdb/test_arctic.py
+++ b/python/tests/integration/arcticdb/test_arctic.py
@@ -16,7 +16,7 @@ import numpy as np
 from datetime import datetime, timezone
 from botocore.client import BaseClient as BotoClient
 
-from arcticdb_ext.exceptions import InternalException
+from arcticdb_ext.exceptions import InternalException, UserInputException
 from arcticdb_ext.storage import NoDataFoundException
 from arcticdb.exceptions import ArcticDbNotYetImplemented, LibraryNotFound, MismatchingLibraryOptions
 from arcticdb.arctic import Arctic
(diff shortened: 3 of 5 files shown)


python/arcticdb/util/mongo_uri_parser.py
```diff
@@ -0,0 +1,1620 @@
+"""
```
poodlewars, line 1: This is wrong, this would be licensed like it is in `pymongo`

environment_unix.yml
```diff
@@ -68,9 +68,9 @@ dependencies:
   # See: https://github.com/man-group/ArcticDB/pull/291
   - hypothesis < 6.73
   - pytest-sugar
-  - pymongo
   # Python dependencies (for tests only)
   - azure-storage-blob
   - azure-identity
   - pyarrow
   - asv
+  - pymongo
```
poodlewars, line 76: Shall we drop pymongo here? Bit dodgy but then at least we'll have PyPi CI run _with_ pymongo and Conda run _without_, so we can cover both code paths

poodlewars (approved):

environment_unix.yml
phoebusm: Wow that's hacky genius thought. Good idea.

phoebusm: I have come up with a better idea on the testing, with parametrised pytest + monkeypatch

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -567,15 +568,24 @@ def test_delete_date_range(arctic_library):
     assert lib["symbol"].version == 1
 
 
+@pytest.mark.parametrize("pymongo_patcher", ["disable_pymongo_in_adapter", None])
 @pytest.mark.skipif(not RUN_MONGO_TEST, reason="Mongo test on windows is fiddly")
-def test_mongo_repr(mongo_test_uri):
+def test_mongo_repr(request, pymongo_patcher, mongo_test_uri):
```
poodlewars, line 573: I think it's a bit cleaner to just have two separate tests, rather than the parameterized fixture, or trying to shoe-horn `monkeypatch` in to a fixture. And we should assert that when pymongo is installed, we get the good exception. This test doesn't belong in `test_mongo_repr`.

environment_unix.yml
jjerphan: [`environment_unix.yml`](https://github.com/man-group/ArcticDB/pull/1027/files/8c05b51c36ab79700ea0bbc80373f020eb91b921#diff-f557abdd91aaee31705408441600af9f7aa105eb9c599b9af12ec50ba868fd82) can include all sorts of dependencies, and we need not be as strict and only list the required dependencies.

We can keep `pymongo` here, and add a comment to mention that it is a testing dependency.

```suggestion
  # Python test dependencies
  - pymongo
```

What matters for ArcticDB's distribution on conda-forge is the removal of [`pymongo` as a python runtime dependency on ArcticDB's feedstock recipe](https://github.com/conda-forge/arcticdb-feedstock/blob/main/recipe/meta.yaml#L104).

python/tests/integration/arcticdb/test_arctic.py
phoebusm: Addressed

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -557,14 +557,32 @@ def test_delete_date_range(arctic_library):
...
+
+
+@MONGO_TESTS_MARK
+def test_mongo_repr_uri_no_pymongo(monkeypatch, mongo_test_uri):
+    import arcticdb.adapters.mongo_library_adapter
+
+    monkeypatch.setattr(arcticdb.adapters.mongo_library_adapter, "_HAVE_PYMONGO", False)
+
+    uri = f"{mongo_test_uri}/?maxPoolSize=10&minPoolSize=100&serverSelectionTimeoutMS=1000"
+    ac = Arctic(uri)
```
poodlewars, line 578: Makes sense to check we can construct with an OK URI, but we don't need the repr assertion here too

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -557,14 +557,32 @@ def test_delete_date_range(arctic_library):
...
     assert repr(ac) == f"Arctic(config=mongodb(endpoint={mongo_test_uri[len('mongodb://'):]}))"
 
+    # With pymongo, exception thrown in the uri_parser;
+    with pytest.raises(UserInputException):
+        uri = f"{mongo_test_uri}//"
+        ac = Arctic(uri)
+
+
+@MONGO_TESTS_MARK
+def test_mongo_repr_uri_no_pymongo(monkeypatch, mongo_test_uri):
```
poodlewars, line 572: `test_mongo_construction_no_pymongo`

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -557,14 +557,32 @@ def test_delete_date_range(arctic_library):
...
-    min_pool_size = 100
-    selection_timeout_ms = 1000
-    uri = f"{mongo_test_uri}/?maxPoolSize={max_pool_size}&minPoolSize={min_pool_size}&serverSelectionTimeoutMS={selection_timeout_ms}"
+def test_mongo_repr_and_uri(mongo_test_uri):
+    uri = f"{mongo_test_uri}/?maxPoolSize=10&minPoolSize=100&serverSelectionTimeoutMS=1000"
     ac = Arctic(uri)
     assert repr(ac) == f"Arctic(config=mongodb(endpoint={mongo_test_uri[len('mongodb://'):]}))"
 
+    # With pymongo, exception thrown in the uri_parser;
+    with pytest.raises(UserInputException):
```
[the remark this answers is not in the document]
phoebusm: It is for testing whether the extraction of endpoint is valid. 
With `pymongo`, `uri_parser` is used for the extraction.
Without `pymongo`, `re` is used. It is not really necessary but it will be easier to debug if something gets messed up.

python/tests/integration/arcticdb/test_arctic.py
phoebusm: Same reason for the above comment. The repr assertion is for making sure the endpoint extraction is correct or not.

python/tests/integration/arcticdb/test_arctic.py
phoebusm: done

poodlewars (approved):

python/arcticdb/adapters/mongo_library_adapter.py
```diff
@@ -14,23 +14,31 @@
...
-            )  # also checks pymongo uri syntax, throw exception as early as possible if syntax is incorrect
-            self._endpoint = f"{parameters['nodelist'][0][0]}:{parameters['nodelist'][0][1]}"
+            if _HAVE_PYMONGO:
+                parameters = parse_uri(
+                    uri
+                )  # also checks pymongo uri syntax, throw exception as early as possible if syntax is incorrect
+                self._endpoint = f"{parameters['nodelist'][0][0]}:{parameters['nodelist'][0][1]}"
+            else:
+                match = re.search(r"\/\/(?P<endpoint>[^\/]*)", uri)
+                self._endpoint = match["endpoint"]
```
jjerphan, line 41: Could you document any benefits `pymongo.uri_parser.parse_uri` has over using a regex, here?

jjerphan (approved):
Thank you, @phoebusm.

PR #1051, opened 2023-11-13
Correct data pipelining in async write_batch

willdealtry:
The batch_read method creates two sets of tasks on the cpu executor, but since it does them in batches one after another they end up on the queue in that order, and the windowing around the first async operation doesn't have the desired effect of pushing the initial blocks all the way through to the end before starting other writes. This causes a memory spile as compressed blocks build up in memory without being written, and wastes potential bandwidth as chunks aren't being written to the storage as early as they could be.

We should create the whole chain in one go, without using intermediate vectors and use the window around the complete operation.


willdealtry:
Subsumed into aggregator_set_data parallelization

Issue #791, opened 2023-08-24
CI for Mongo backend on Windows


poodlewars:
It's skipped atm as spinning up the Mongo server is a pain on Windows.

Issue #1014, opened 2023-10-30
Storage Fixtures framework


qc00:
(Separated out from #473)

Separated the test fixture setup logic from `conftest.py` to its own framework:

* Proper documented API
* Callable without pytest
* Full life-cycle management with proper shutdown and clean-up
* Can have multiple isolated fixtures open at the same time
* Permission control feature to support testing of #473 and #447

PR #1054, opened 2023-11-13
#1014 storage fixtures

qc00:
#### Reference Issues/PRs
#1014

Introducing the StorageFixtures framework

* Standard, not pytest-coupled API for storage fixture life cycles
* Methods for setting permissions and shutting down the fixtures independent of pytest for error handling testing
* Proper cleanup and exception handling to allow tests to run concurrently and independently of earlier failures
* PrefixingLibraryAdapterDecorator to support isolating Mongo fixture instances, but can be used with any other Storage type


build_tooling/parallel_test.sh
```diff
@@ -7,48 +7,28 @@ echo Saving results to ${TEST_OUTPUT_DIR:="$(realpath "$tooling_dir/../cpp/out")
...
+    mkdir -p $PARALLEL_TEST_ROOT
+    MSYS=winsymlinks:nativestrict ln -s "$(realpath "$tooling_dir/../python/tests")" $PARALLEL_TEST_ROOT/
+    cd $PARALLEL_TEST_ROOT
 
     export ARCTICDB_RAND_SEED=$RANDOM
-    $catch python -m pytest -v -n $PYTEST_THREADS --show-capture=no --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
+
+    $catch python -m pytest -n $splits -v --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
         --junitxml="$TEST_OUTPUT_DIR/pytest.$group.xml" \
-        --splits $splits --group $group --durations-path="$duration_file" --store-durations \
```
G-D-Petrov, line 39: can we still store the durations and the durations files?
they are be very beneficial

qc00: You can get that info from the JUnit XML. The durations were specifically used by the pytest-split plugin.

poodlewars:
Would be nice to add some evidence (sorted output of pytest discovery before and after and a diff, maybe?) to demonstrate that the test cases we execute are similar before and after this change.

qc00:
@poodlewars There's no change in this PR on what storage the tests will run against; that's in the next PR.

This PR is mainly moving the same fixtures to a new API framework and using the API in tests.

python/arcticdb/adapters/prefixing_library_adapter_decorator.py
```diff
@@ -0,0 +1,82 @@
...
+import inspect
+import re
+from typing import Optional, Tuple, Type, Callable, TYPE_CHECKING, Iterable, List
+
+from arcticdb import LibraryOptions, Arctic
+from arcticdb.adapters.arctic_library_adapter import ArcticLibraryAdapter
+from arcticdb.encoding_version import EncodingVersion
+
+
+class PrefixingLibraryAdapterDecorator(ArcticLibraryAdapter if TYPE_CHECKING else object):
```
poodlewars, line 17: I find this quite confusing, because:

- It is only used for Mongo, but that isn't obvious
- The `__getattr__` wrapper is complex
- The test fixture modifying `Arctic._LIBRARY_ADAPTERS` is surprising
- I don't understand the `PrefixingLibraryAdapterDecoratorChecked` concept

Why not implement the prefixing feature for Mongo directly in the non-test tree by modifying `mongo_library_adapter.py`? We can keep the prefix feature undocumented and warn if it is used so that we don't commit to it as an API feature.

python/arcticdb/arctic.py
```diff
@@ -34,6 +35,10 @@ class Arctic:
         InMemoryLibraryAdapter,
     ]
 
+    # For test fixture clean up
+    _created_lib_names: Optional[List[str]] = None
```
poodlewars, line 39: Why do we need this additional state? Can't we use the libraries (with prefixes) that we already have stored in the library manager, and strip the prefix on read, using the logic in `library_manager_names_to_user_facing` which we already need?

python/arcticdb/arctic.py
```diff
@@ -34,6 +35,10 @@ class Arctic:
...
+    _accessed_libs: Optional[List[NativeVersionStore]] = None
```
poodlewars, line 40: The interactions with this don't seem to do anything?

python/arcticdb/storage_fixtures/api.py
```diff
@@ -0,0 +1,182 @@
...
+        return self.value
+
+
+class _SaferContextManager(AbstractContextManager):
+    def __enter__(self):
+        try:
+            self._safe_enter()
+            return self
+        except Exception as e:
+            self.__exit__(type(e), e, None)
```
poodlewars, line 44: Why do we need this exception raising logic? The exception will be raised even without this as long as `__exit__` returns anything other than `True`.

python/arcticdb/storage_fixtures/api.py
```diff
@@ -0,0 +1,182 @@
...
+            raise
+
+    def _safe_enter(self):
+        """Setup that don't start external resources can be in __init__.
```
poodlewars, line 48:
```suggestion
        """Setup that doesn't start external resources can be in __init__.
```

python/arcticdb/storage_fixtures/api.py
```diff
@@ -0,0 +1,182 @@
...
+        Resources that needs `__exit__()` cleanup during exceptions should be here."""
```
poodlewars, line 49:
```suggestion
        Resources that need `__exit__()` cleanup during exceptions should be here."""
```

python/arcticdb/storage_fixtures/api.py
```diff
@@ -0,0 +1,182 @@
...
+
+        If implemented, the set value should affect subsequent calls to ``create_fixture()`` and, if explicitly
+        documented, existing fixtures. The base implementation is hard-wired to ``False``."""
+        return False
+
+    @contextmanager
+    def enforcing_permissions_context(self, set_to=True):
+        saved = self.enforcing_permissions
+        try:
+            self.enforcing_permissions = set_to
```
poodlewars, line 176: You need a setter for `enforcing_permissions` don't you? It's read-only with the current code.

python/arcticdb/storage_fixtures/lmdb.py
```diff
@@ -0,0 +1,78 @@
...
+            #  keeping the LMDB env open before they will release the lock and allow Windows to delete the LMDB files.
+            result.version_store = None
+            result._library = None
+        for a in self.arctic_instances:
+            a._library_manager = None
+            pass
+
+        shutil.rmtree(self.db_dir)
+
+    def create_test_cfg(self, lib_name: str, *, lmdb_config: Dict[str, Any] = _DEFAULT_CONFIG) -> EnvironmentConfigsMap:
```
poodlewars, line 51: `_DEFAULT_CONFIG` is mutable, which is dangerous for a kwarg

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -0,0 +1,275 @@
...
+    key: Key
+    _boto_bucket: Any = None
+
+    def __init__(self, factory: "BaseS3StorageFixtureFactory", bucket: str):
+        super().__init__()
+        self.factory = factory
+        self.bucket = bucket
+
+        if factory.enforcing_permissions:
+            self.key = factory._create_user_get_key(bucket + "_user")
```
poodlewars, line 46: `BaseS3StorageFixtureFactory` doesn't have `_create_user_get_key` so this confuses my linter. Slightly neater to just have a polymorphic `get_key` method on the factory? Similar issue for `factory._iam_admin` and `factory._RW_POLICY` and `factory._RO_POLICY` below

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -0,0 +1,275 @@
...
+    default_prefix: Optional[str] = None
+    clean_bucket_on_fixture_exit = True
+
+    def __str__(self):
+        return f"{type(self).__name__}[{self.default_bucket or self.endpoint}]"
+
+    @staticmethod
+    def from_environment_variables():
+        out = BaseS3StorageFixtureFactory()
+        out.endpoint = os.getenv("ARCTICDB_REAL_S3_ENDPOINT")
```
poodlewars, line 118: Bit confusing here since the class is to `Logic and fields common to real and mock S3` but these env vars are specific to real storages?

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -0,0 +1,275 @@
...
+
+            _reqs_till_rate_limit = -1
+
+            def get_backend_for_host(self, host):
+                return self._MAP.get(host, host)
+
+            def __call__(self, environ, start_response):
+                path_info: bytes = environ.get("PATH_INFO", "")
+
+                with self.lock:
```
poodlewars, line 181: I don't think we need this feature, since we already have the storage failure simulator in the C++ layer, and I'm reluctant for us to have multiple storage failure simulation mechanisms.

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -0,0 +1,275 @@
...
+
+        self._s3_admin = self._boto("s3", self.default_key)
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        GracefulProcessUtils.terminate(self._p)
+
+    def _create_user_get_key(self, user: str, iam=None):
+        iam = iam or self._iam_admin
+        user_id = iam.create_user(UserName=user)["User"]["UserId"]
```
poodlewars, line 222: `user_id` is unused, do you mean to include it in the return value?

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -0,0 +1,275 @@
...
+        self._s3_admin.create_bucket(Bucket=bucket)
+        self._bucket_id += 1
+
+        out = S3Bucket(self, bucket)
+        self._live_buckets.append(out)
+        return out
+
+    def cleanup_bucket(self, b: S3Bucket):
+        self._live_buckets.remove(b)
+        if not len(self._live_buckets):
```
poodlewars, line 270: Invert the if-else to avoid the double negative?

python/arcticdb/storage_fixtures/utils.py
```diff
@@ -0,0 +1,105 @@
...
+import socketserver
+import time
+import warnings
+from typing import Union, Any
+from contextlib import contextmanager
+
+_WINDOWS = platform.system() == "Windows"
+
+
+def get_ephemeral_port():  # https://stackoverflow.com/a/61685162/
```
poodlewars, line 24: This seems much the same as the old way? https://github.com/man-group/ArcticDB/blob/615edcb76216024d51f2ec5ede92eec6ea52e587/python/tests/conftest.py#L86

python/arcticdb/storage_fixtures/utils.py
```diff
@@ -0,0 +1,105 @@
...
+    with socketserver.TCPServer(("localhost", 0), None) as s:
+        return s.server_address[1]
+
+
+ProcessUnion = Union[multiprocessing.Process, subprocess.Popen]
+
+
+class GracefulProcessUtils:
```
poodlewars, line 32: (nit) No reason for this to be a class rather than a normal Python module?

poodlewars (commented):
--WIP review: I haven't read the diff in the `tests/` tree yet.--

Done now.

python/arcticdb/storage_fixtures/azurite_patching.py
```diff
@@ -0,0 +1,79 @@
...
+        fake_node.unlink()
+    match = re.search(r'^AZURITE=("?)(.+)\1$', proc.stdout, flags=re.MULTILINE)
+    assert match, "Running `azurite` with our fake `node` didn't return the expected text: " + proc.stdout
+    resolved = Path(match.group(2).replace("\\\\", "\\")).resolve(strict=True)
+    assert resolved.match("**/node_modules/azurite/dist/src/*"), f"Unusual npm path [{resolved}] from {proc.stdout}"
+    return resolved.parent
+
+
+class PatchedAzurite:
+    """Azurite does not implement some of the documented permission requirements, so we patch it."""
```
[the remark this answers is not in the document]
qc00: The problem is that many of the permissions in Azurite are implemented incorrectly, but I only need one fixed for my tests.

If I contribute a patch, then I'll probably need to fix all the similar ones to pass their code review, and it will have wider implications (breaking other user's tests).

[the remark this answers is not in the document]
poodlewars: Can we raise a discussion with them and see if they would be open to the change, at least?

python/arcticdb/adapters/prefixing_library_adapter_decorator.py
qc00: >It is only used for Mongo, but that isn't obvious
It _can_ be used for any storage type. I have plans to use it for LMDB to address the performance problem of repeatedly creating new storage on Windows.

>The __getattr__ wrapper is complex
This is for safety.

For example, say someone added (or changed) a method in `ArcticLibraryAdapter` that takes a library name and s/he didn't know this prefixing thing, then the real library name will "leak" in that function, which can be a very confusing bug.

This has actually happened twice since I started this work, which is why I wrote this.

>The test fixture modifying Arctic._LIBRARY_ADAPTERS is surprising
How so? I would expect in the future, with some modularization, all LibraryAdapters would be dynamically registered this way.

>I don't understand the PrefixingLibraryAdapterDecoratorChecked concept
It's a marker attribute to avoid having to inspect the called functions each and every time. Computing the `Signature` of a function is quite slow.

>Why not implement the prefixing feature for Mongo directly in the non-test tree by modifying mongo_library_adapter.py?
It's
* conceptually cleaner (using a standard, self-contained decorator pattern)
* more reusable (see my reply to your first point)
* safer (similar to the check in `__getattr__`, if someone added a new method in `ArcticLibraryAdapter` with a default implementation and didn't know the mongo adapter has a prefixing logic in it, then the unprefixed name will leak through that default implementation)

python/arcticdb/arctic.py
qc00: In the `real_s3` fixture, there might be existing libraries created by other tests.

python/arcticdb/arctic.py
qc00: Ah, the clean up was reimplemented and don't need this anymore. Will remove.

python/arcticdb/storage_fixtures/api.py
qc00: It is meant to be read-only in the base class. Sub-types will supply a setter if they support permissions.

python/arcticdb/storage_fixtures/lmdb.py
qc00: It's never going to be mutated and there's no frozendict...

python/arcticdb/storage_fixtures/s3.py
qc00: _If_ we one-day support permissions on `BaseS3StorageFixtureFactory`, then it will have to implement these as well. I can add a type check for now.

python/arcticdb/storage_fixtures/s3.py
qc00: I can move this out of this class.

python/arcticdb/storage_fixtures/s3.py
qc00: I can take the commit out.

However, I don't think testing the storage exceptions solely in C++ is sufficient because it's not easy to test if the correct exception type is bubbled up to Python.

python/arcticdb/storage_fixtures/s3.py
qc00: Useful for debug

python/arcticdb/storage_fixtures/utils.py
qc00: It's move from conftest

python/arcticdb/storage_fixtures/utils.py
qc00: It makes it clear the three methods should be used as a system. Otherwise, there's a higher risk that someone might call `terminate()` on a process not started by `start()` and cause a confusing mess.

Ideally, these functions become methods of a full process wrapper class, but we would end up having the `__getattr__` debate all over again.

python/tests/conftest.py
```diff
@@ -72,468 +38,163 @@
...
-
-@pytest.fixture(scope="session")
-def only_test_encoding_version_v1():
-    """Dummy fixture to reference at module/class level to reduce test cases"""
-
-
-def pytest_generate_tests(metafunc):
-    if "encoding_version" in metafunc.fixturenames:
-        if sys.platform == "win32":
-            metafunc.parametrize("encoding_version", [EncodingVersion.V2])
```
poodlewars, line 379: Is dropping this going to slow down the Windows builds? Thinking about it though, the old logic is really weird, since we still have `DEFAULT_ENCODING_VERSION = EncodingVersion.V1` on all platforms

python/arcticdb/storage_fixtures/mongo.py
```diff
@@ -0,0 +1,145 @@
...
+    dedicated instance on a random port."""
+    import requests
+
+    mongo_host = os.getenv("CI_MONGO_HOST", "localhost")
+    try:
+        res = requests.get(f"http://{mongo_host}:27017")
+        if res.status_code == 200 and "mongodb" in res.text.lower():
+            return ExternalMongoDBServer(f"mongodb://{mongo_host}:27017")
+    except requests.exceptions.ConnectionError:
+        return ManagedMongoDBServer()
```
poodlewars, line 145: Is the fallback desirable? I would think that if we have `CI_MONGO_HOST` specified it would be surprising behaviour to silently _not_ use it!

python/arcticdb/storage_fixtures/mongo.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def __init__(self, data_dir: Optional[str] = None, port=0, executable="mongod"):
+        self._data_dir = data_dir or tempfile.mkdtemp("ManagedMongoDBServer")
+        self._port = port or get_ephemeral_port()
+        self._executable = executable
+
+    def _safe_enter(self):
+        from pymongo import MongoClient
+
+        cmd = [self._executable, "--port", str(self._port), "--dbpath", self._data_dir, "--nojournal"]
```
poodlewars, line 104: Do we really need `--nojournal`? It's removed in modern Mongos, and is in fact always on with them, which caused me a bit of pain when setting up the new dev machine.

python/tests/conftest.py
poodlewars, line 1: Does this resolve #791 ?

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -556,55 +514,43 @@ def test_delete_date_range(arctic_library):
     assert lib["symbol"].version == 1
 
 
-@MONGO_TESTS_MARK
-def test_mongo_construction_with_pymongo(mongo_test_uri):
```
poodlewars, line 560: This test has been removed?

python/tests/scripts/test_update_storage.py
```diff
@@ -37,139 +39,97 @@ def _get_azure_storage_config(cfg):
...
-    uri, bucket = get_s3_aws_auth_uri_bucket_and_setup_server(moto_s3_endpoint_and_credentials, monkeypatch)
+        monkeypatch.setattr(os, "putenv", putenv_s)
+        monkeypatch.setattr(os, "unsetenv", lambda n: putenv_s(n, ""))
 
-    ac = Arctic(uri)
-    create_library_config(ac, LIB_NAME)
+    monkeypatch.setenv("AWS_ACCESS_KEY_ID", s3_storage.key.id)
+    monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", s3_storage.key.secret)
+
+    uri = s3_storage.replace_uri_field(s3_storage.arctic_uri, ArcticUriFields.USER, "aws_auth=true", start=1)
```
poodlewars, line 85: Shouldn't we have a fixture that gives us an `aws_auth=true` URI rather than doing surgery on an access key based URI?

python/tests/scripts/test_update_storage.py
```diff
@@ -37,139 +39,97 @@ def _get_azure_storage_config(cfg):
...
-    endpoint, port, bucket, aws_access_key, aws_secret_key = moto_s3_endpoint_and_credentials
-    monkeypatch.setenv("AWS_ACCESS_KEY_ID", aws_access_key)
-    monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", aws_secret_key)
-    return (
-        endpoint.replace("http://", "s3://").rsplit(":", 1)[0] + ":" + bucket + "?aws_auth=true" + "&port=" + port,
-        bucket,
-    )
+def test_upgrade_script_s3_rbac_ok(s3_storage: S3Bucket, monkeypatch):
+    """Just _RBAC_ as creds is a placeholder. Leave config with that alone."""
+    if os.name == "nt":
```
poodlewars, line 72: `sys.platform == "win32"` like everywhere else?

python/tests/scripts/test_update_storage.py
```diff
@@ -37,139 +39,97 @@ def _get_azure_storage_config(cfg):
...
+        if sys.version_info < (3, 9):
```
poodlewars, line 73: Awesome investigation here, thank you. This test was being skipped on Windows before.

python/tests/scripts/test_update_storage.py
```diff
@@ -37,139 +39,97 @@ def _get_azure_storage_config(cfg):
...
-    assert s3_storage.bucket_name == bucket
+    assert s3_storage.bucket_name == s3_storage.bucket_name
     assert s3_storage.credential_name == USE_AWS_CRED_PROVIDERS_TOKEN
     assert s3_storage.credential_key == USE_AWS_CRED_PROVIDERS_TOKEN
 
+    lib = ac[LIB_NAME]
+    assert lib.read("x").data == 123
 
-@pytest.mark.skipif(sys.platform == "win32", reason="Test fixture issue with creds")
-def test_aws_auth_with_storage_overrider(moto_s3_endpoint_and_credentials, monkeypatch):
```
poodlewars, line 123: This test is a bit messy, but it seems to be our only check with `aws_auth=True`, so I don't think we should delete it completely.

python/arcticdb/storage_fixtures/s3.py
poodlewars: The failure simulator has a way to configure itself with a protobuf object from the Python layer, allowing for an end to end test.

python/arcticdb/storage_fixtures/lmdb.py
poodlewars: The normal `if x is None: x = default` idiom would be fine

python/arcticdb/arctic.py
qc00: It will be set on the instance level when it's set. This `None` is the default (when used in production code).

python/arcticdb/arctic.py
qc00: Actually, it's still used by LMDB cleanup: https://github.com/man-group/ArcticDB/blob/1014-storage-fixtures/python/arcticdb/storage_fixtures/lmdb.py#L34,L39

poodlewars: The state in library manager will suffice for it, won't it?

python/arcticdb/storage_fixtures/s3.py
qc00: I meant is it testing pybind as well?

python/tests/conftest.py
```diff
@@ -72,468 +38,163 @@
...
+        yield f
 
+
+# endregion
+# region ==================================== `Arctic` API Fixtures ====================================
 @pytest.fixture(
     scope="function",
-    params=[pytest.param("REAL_S3", marks=REAL_S3_TESTS_MARK)],
+    params=[
+        "s3",
```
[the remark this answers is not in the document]
qc00: It's all going to be removed in the next PR anyway.

python/tests/conftest.py
qc00: It will be slower, but I felt not testing both encoding versions is wrong.

I was thinking about using `mem` storage for V1 testing on Windows, but once I implement the LMDB reuse logic, it should be roughly as fast as `mem`.

python/arcticdb/storage_fixtures/mongo.py
qc00: Good point. Will change

qc00: I'll make the change, but it won't be able to support the `CI_MONGO_HOST` defaults to `localhost` behaviour any more.

In the medium term, we can install the `mongod` binary into our Docker image (it's already installed in other runners) and then drop the `CI_MONGO_HOST`.

python/tests/conftest.py
qc00: It has the code to do so, but I intentionally didn't enable it on Windows because:
* Currently there's limited business case
* It will make the test even slower

python/tests/integration/arcticdb/test_arctic.py
qc00: It's merged into `test_mongo_repr`.

python/tests/scripts/test_update_storage.py
qc00: I thought about it, but it's only used in this place and will require external mocking and the Windows patching to be in the fixture itself. That will re-couple the fixture with pytest...

python/tests/scripts/test_update_storage.py
qc00: Its logic is merged into the `test_upgrade_script_s3_rbac_ok` except that I replaced the DF with a `write_pickle`.

python/arcticdb/storage_fixtures/s3.py
poodlewars: Yes

python/arcticdb/storage_fixtures/azurite_patching.py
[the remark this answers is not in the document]
qc00: Raised https://github.com/Azure/Azurite/issues/2299

I'll remove this code, but it will mean the upcoming PRs will not be able to work until Azurite fixes it.

python/arcticdb/arctic.py
qc00: No, because the `LibraryManager` don't keep track of `NativeVersionStore` instances and we need to set the `.version_store` in these instances to `None` in order to release the reference to the LMDB Storage in C++.

python/arcticdb/storage_fixtures/s3.py
poodlewars: OK after our big chat right now I think this is in fact the way to go, since we don't have a good stub S3 client implementation at the moment. Let's keep this feature in

qc00:
>The removed tests, and if there are any other removed tests that I've missed. Hopefully some pytest discovery output can put my mind at rest about this.

Please refer to the attachments.

Both files are obtained using `pytest --co`. `after.txt` has these string replacements to match the fixture naming of the original:

```
Regex: \[(s3|lmdb|mem)- => \U$0
[azurite- => [Azure-
[mongo- => [Mongo-
[real_s3- => [Real_S3-
```

The only meaningful difference is `test_separation_between_libraries_with_prefixes` no longer parameterizes Mongo because the test code used to skip on mongo.

python/arcticdb/storage_fixtures/azurite_patching.py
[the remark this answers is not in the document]
poodlewars: Great thank you. Can we keep this patching mechanism in the git history so we can reinstate it if the Azurite people aren't helpful?

python/arcticdb/arctic.py
poodlewars: But it should just be instance state, right?

poodlewars (approved):

python/arcticdb/storage_fixtures/azurite_patching.py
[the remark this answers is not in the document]
qc00: You can find all past commits in this PR's time line

python/arcticdb/arctic.py
qc00: This is a pretty standard way to set defaults in Python? `self._created_lib_names` would get this `None` unless it's set in the instance.

PR #1065, opened 2023-11-16
Parallelize aggregator_set_data

willdealtry:
It was originally envisaged that aggregator_set_data would be almost weightless, however we now do quite a bit of string processing in this function, which can potentially be quite slow. We should parallelize earlier in the write pipeline, releasing the GIL and only retaking it if we know that we need to do allocations in Python (because there are strings that are not currently in compact utf8 format)

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -16,6 +16,7 @@
 #include <arcticdb/async/tasks.hpp>
 #include <arcticdb/stream/stream_utils.hpp>
 #include <arcticdb/processing/clause.hpp>
+#include <arcticdb/storage/key_segment_pair.hpp>
 
 namespace arcticdb::async {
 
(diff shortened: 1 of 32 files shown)


cpp/arcticdb/async/async_store.hpp
```diff
@@ -75,278 +76,266 @@ class AsyncStore : public Store {
...
-                        auto key_opt_segment = std::forward<decltype(item)>(item);
-                        if (key_opt_segment.second)
-                            lib->write(Composite<storage::KeySegmentPair>({VariantKey{key_opt_segment.first},
-                                                                           std::move(*key_opt_segment.second)}));
-
-                        return key_opt_segment.first;
-                    }));
-        }
+    folly::Future<SliceAndKey> async_write(
+            folly::Future<std::tuple<PartialKey, SegmentInMemory, pipelines::FrameSlice>> &&input_fut,
```
alexowens90, line 309: A struct would avoid this tuple being redefined everywhere this is called from.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -75,278 +76,266 @@ class AsyncStore : public Store {
...
+        using KeyOptSegment = std::pair<VariantKey, std::optional<Segment>>;
+        return std::move(input_fut).thenValue([this] (auto&& input) {
+            auto [key, seg, slice] = std::move(input);
+            auto key_seg = EncodeAtomTask{
+                std::move(key),
+                ClockType::nanos_since_epoch(),
+                std::move(seg),
+                codec_,
+                encoding_version_}();
+            return std::make_tuple<VariantKey, Segment, pipelines::FrameSlice>(std::move(key_seg.variant_key()), std::move(key_seg.segment()), std::move(slice));
```
alexowens90, line 320: This could just be `std::pair<KeySegmentPair, FrameSlice>`, as the key segment pair is just reconstructed in the next lambda anyway.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -33,142 +33,195 @@ namespace arcticdb::pipelines {
...
-            res.emplace_back(SliceAndKey{slices[i], std::move(to_atom(keys[i]))});
-        }
-        return res;
-    });
+        if (slice.row_range.first == first_row.value())
+            slice_num_for_column = 0;
+
+        slice_and_rowcount.emplace_back(slice, slice_num_for_column);
+        ++slice_num_for_column;
+    }
```
alexowens90, line 148: I'm confused by this logic. `first_row` is always set to `slices[0].row_range.first` and then never modified. So `slice_num_for_column` will be zero for all slices in the first row-slice, then 1, 2, 3,... for subsequent slices?

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -33,142 +33,195 @@ namespace arcticdb::pipelines {
...
+
+    const auto write_window = ConfigsMap::instance()->get_int("VersionStore.BatchWriteWindow", 50);
```
alexowens90, line 150: Can we default to something less heurisitic, like `2 * io_thread_count()`? We also need to be careful of the max task queue size for the CPU executor in `batch_write` situations though.

cpp/arcticdb/pipeline/write_frame.cpp
willdealtry: It's quite niche. If the data is not contiguous and we need to call the FlattenHelper we multiply the slice_num by the regular_slice_size by the array strides, to work out where our starting point is. So if a uint64 array was backwards and had a stride of -8, and we were dividing it into slices of two rows each, the first slice would start at 0 (the origin pointer) and work backwards, and then next slice would start at -16.

I think in earlier times this logic was a bit closer together, it's not obvious where it goes these days. Not totally sure how to fix it though, we definitely have the information here, and we need it to go there.

cpp/arcticdb/async/async_store.hpp
willdealtry: I'm not 100% sure I like it were we do that elsewhere, it makes it more opaque when it's not used frequently enough for people to know what it is. At the moment this is only called from once place, if it starts to get more widely used maybe we can consider doing it then.

alexowens90 (approved):

PR #1087, opened 2023-11-20
Add xfail to flaky tests

G-D-Petrov:
#### Reference Issues/PRs
issue #496

As agreed today, I am adding xfail to the tests that need to be investigated as part of issue #496 .
The xfails will be removed when the tests are no longer flaky.

### DIFF
--- a/python/tests/hypothesis/arcticdb/test_append.py
+++ b/python/tests/hypothesis/arcticdb/test_append.py
@@ -135,6 +135,7 @@ def test_(initial: InputFactories, append: InputFactories, match, swap, lmdb_ver
     df_in_str=st.booleans(),
 )
 @SLOW_TESTS_MARK
+@pytest.mark.xfail(reason="Needs to be fixed with issue #496")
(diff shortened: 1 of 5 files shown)


poodlewars:
Seems to be at least one more,

```
 3: ============================== warnings summary ===============================
3: test_deletion.py::test_delete_version_with_append[real_s3_store_factory-0]
3:   C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\_pytest\unraisableexception.py:78: PytestUnraisableExceptionWarning: Exception ignored in: <finalize object at 0x26fca086720; dead>
```

poodlewars (approved):
Please can you make sure #496 lists these tests you're changing?

It would be good to follow up on the root causes of these as soon as we can. Can you try during support? Then next week I think we'll need to find someone to focus on whichever are still `xfail`-d.

poodlewars:
Is `xfail` right here? We don't actually expect these to reliably fail, right?

PR #1089, opened 2023-11-20
Support for `first` and `last` aggregators - numeric columns - static schema

### LINKED ISSUE #957: Add support for `first` and `last` aggregators - numeric columns

Hind-M:
#### Reference Issues/PRs
Fix #957 

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/processing/aggregation.cpp
+++ b/cpp/arcticdb/processing/aggregation.cpp
@@ -627,15 +627,17 @@ void LastAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input
                 using ColumnType =  typename ColumnTagType::raw_type;
                 auto groups_pos = 0;
                 auto grp_idx = groups[0];
+                bool is_first_group_el = true;
                 while (auto block = col_data.next<TypeDescriptorTag<ColumnTagType, DimensionTag<entity::Dimension::Dim0>>>()) {
                     auto ptr = reinterpret_cast<const ColumnType *>(block.value().data());
                     for (auto i = 0u; i < block.value().row_count(); ++i, ++ptr) {
                         auto& val = out_ptr[groups[groups_pos]];
                         if constexpr(std::is_floating_point_v<ColumnType>) {
                             const auto& curr = GlobalRawType(*ptr);
-                            if ((grp_idx == groups[groups_pos]) && !std::isnan(static_cast<ColumnType>(curr))) {
+                            if ((grp_idx == groups[groups_pos]) && (is_first_group_el || !std::isnan(static_cast<ColumnType>(curr)))) {
                                 groups_cache.insert(groups[groups_pos]);
                                 val = curr;
+                                is_first_group_el = false;
                             }
                             else if (grp_idx != groups[groups_pos]) { // Updating when changing group
                                 if (auto iter = groups_cache.find(groups[groups_pos]); iter == groups_cache.end()) { // Group not seen yet
--- a/cpp/arcticdb/processing/aggregation.cpp
+++ b/cpp/arcticdb/processing/aggregation.cpp
@@ -604,5 +604,75 @@ SegmentInMemory FirstAggregatorData::finalize(const ColumnName& output_column_na
     return res;
 }
 
+/***********************
+ * LastAggregatorData *
+ ***********************/
+
+void LastAggregatorData::add_data_type(DataType data_type) {
+    add_data_type_impl(data_type, data_type_);
+}
+
+void LastAggregatorData::aggregate(const std::optional<ColumnWithStrings>& input_column, const std::vector<size_t>& groups, size_t unique_values) {
+    if(data_type_.has_value() && *data_type_ != DataType::EMPTYVAL && input_column.has_value()) {
+        entity::details::visit_type(*data_type_, [&input_column, unique_values, &groups, that=this] (auto global_type_desc_tag) {
+            using GlobalInputType = decltype(global_type_desc_tag);
+            using GlobalTypeDescriptorTag =  typename OutputType<GlobalInputType>::type;
+            using GlobalRawType = typename GlobalTypeDescriptorTag::DataTypeTag::raw_type;
+            that->aggregated_.resize(sizeof(GlobalRawType)* unique_values);
+            auto col_data = input_column->column_->data();
+            auto out_ptr = reinterpret_cast<GlobalRawType*>(that->aggregated_.data());
+            std::unordered_set<size_t> groups_cache;
+            entity::details::visit_type(input_column->column_->type().data_type(), [&groups, &groups_cache, &out_ptr, &col_data] (auto type_desc_tag) {
+                using ColumnTagType = std::decay_t<decltype(type_desc_tag)>;
+                using ColumnType =  typename ColumnTagType::raw_type;
+                auto groups_pos = 0;
+                auto grp_idx = groups[0];
+                while (auto block = col_data.next<TypeDescriptorTag<ColumnTagType, DimensionTag<entity::Dimension::Dim0>>>()) {
+                    auto ptr = reinterpret_cast<const ColumnType *>(block.value().data());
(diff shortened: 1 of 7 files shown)


DrNickClarke:
Will first and last aggregators be able to share some code, like min and max?

Hind-M:
> Will first and last aggregators be able to share some code, like min and max?

The `min` and `max` aggregators were split in this [PR](https://github.com/man-group/ArcticDB/pull/1007). So I think it's going to be the same for `first` and `last` (although it may be subject to change later depending on the implementation details).

cpp/arcticdb/processing/clause.cpp
```diff
@@ -367,6 +380,11 @@ Composite<EntityIds> AggregationClause::process(Composite<EntityIds>&& entity_id
     // Work out the common type between the processing units for the columns being aggregated
     procs.broadcast([&aggregators_data, &aggregators=aggregators_](auto& proc) {
         for (auto agg_data: folly::enumerate(aggregators_data)) {
+            // Check that segments row ranges are the same
+            internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+                std::all_of(proc.row_ranges_->begin(), proc.row_ranges_->end(), [&] (const auto& row_range) {return row_range->start() == proc.row_ranges_->at(0)->start();}),
```
Hind-M, line 385: IIUC, checking on `start` only would be enough.

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -217,6 +217,148 @@ def test_count_aggregation(local_object_version_store):
...
+    )
+    q = QueryBuilder()
+    q = q.groupby("grouping_column").agg({"get_first": "first"})
+    symbol = "test_first_aggregation"
+    local_object_version_store.write(symbol, df)
+
+    res = local_object_version_store.read(symbol, query_builder=q)
+    res.data.sort_index(inplace=True)
+
+    df = pd.DataFrame({"get_first": [100.0, 2.7, 5.8, np.nan]}, index=["group_1", "group_2", "group_3", "group_4"])
```
alexowens90, line 266: Is including NaNs in the returned df the default pandas behaviour?

cpp/arcticdb/processing/clause.cpp
```diff
@@ -367,6 +380,11 @@ Composite<EntityIds> AggregationClause::process(Composite<EntityIds>&& entity_id
     // Work out the common type between the processing units for the columns being aggregated
     procs.broadcast([&aggregators_data, &aggregators=aggregators_](auto& proc) {
```
alexowens90, line 381: I think this works kind of by luck, in that `broadcast` iterates forward through the contained vector in the composite, which is in order because of the way `procs` was constructed. However, this is an implementation detail of `Composite`, and should not be relied upon. Replacing `procs.broadcast` with `for (auto& proc: procs_as_range)` is guaranteed to have the correct behaviour.

cpp/arcticdb/processing/aggregation.cpp
```diff
@@ -489,4 +509,157 @@ SegmentInMemory CountAggregatorData::finalize(const ColumnName& output_column_na
...
+                            }
+                            else if (grp_idx != groups[groups_pos]) { // Updating when changing group
+                                if (is_first_group_el) { // Group not seen yet
+                                    groups_cache_.insert(groups[groups_pos]);
+                                    val = curr;
+                                }
+                                grp_idx = groups[groups_pos];
+                            }
+                        }
+                        ++groups_pos;
```
alexowens90, line 571: Can this be moved to line 536, to make it clear it corresponds to the row we are in within this column?

cpp/arcticdb/processing/aggregation.cpp
```diff
@@ -489,4 +509,157 @@ SegmentInMemory CountAggregatorData::finalize(const ColumnName& output_column_na
...
+            using GlobalTypeDescriptorTag =  typename OutputType<GlobalInputType>::type;
+            using GlobalRawType = typename GlobalTypeDescriptorTag::DataTypeTag::raw_type;
+            aggregated_.resize(sizeof(GlobalRawType)* unique_values);
+            auto col_data = input_column->column_->data();
+            auto out_ptr = reinterpret_cast<GlobalRawType*>(aggregated_.data());
+            entity::details::visit_type(input_column->column_->type().data_type(), [this, &groups, &out_ptr, &col_data] (auto type_desc_tag) {
+                using ColumnTagType = std::decay_t<decltype(type_desc_tag)>;
+                using ColumnType =  typename ColumnTagType::raw_type;
+                auto groups_pos = 0;
+                auto grp_idx = groups[0];
```
alexowens90, line 533: I'm confused by why we need this?

cpp/arcticdb/processing/aggregation.cpp
```diff
@@ -489,4 +509,157 @@ SegmentInMemory CountAggregatorData::finalize(const ColumnName& output_column_na
...
+                while (auto block = col_data.next<TypeDescriptorTag<ColumnTagType, DimensionTag<entity::Dimension::Dim0>>>()) {
+                    auto ptr = reinterpret_cast<const ColumnType *>(block.value().data());
+                    for (auto i = 0u; i < block.value().row_count(); ++i, ++ptr) {
+                        auto& val = out_ptr[groups[groups_pos]];
+                        bool is_first_group_el = (groups_cache_.find(groups[groups_pos]) == groups_cache_.end());
+                        if constexpr(std::is_floating_point_v<ColumnType>) {
+                            const auto& curr = GlobalRawType(*ptr);
+                            if ((grp_idx == groups[groups_pos]) && (is_first_group_el || std::isnan(static_cast<ColumnType>(val)))) {
```
alexowens90, line 541: This logic looks like the output of will contain NaNs only if that is the only value in the group, is that correct? And is that the same behaviour as pandas?

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_aggregation.py
Hind-M: Yes.

cpp/arcticdb/processing/aggregation.cpp
Hind-M: > This logic looks like the output of will contain NaNs only if that is the only value in the group, is that correct?

Exactly.

> And is that the same behaviour as pandas?

Yes.

cpp/arcticdb/processing/aggregation.cpp
Hind-M: I'm not sure I understand your question. `grp_idx` is used to check the current group we're in.

cpp/arcticdb/processing/clause.cpp
alexowens90: For now yes, if the start is the same the end will be too

cpp/arcticdb/processing/aggregation.cpp
alexowens90: Looking at the logic in lines 559-569, the behaviour would be the same if this variable didn't exist and it was removed from the if statements. Truth table:
```
grp_idx == groups[groups_pos] | is_first_group_el | Behaviour
true  | true  | groups_cache_.insert(groups[groups_pos]) and val = curr
true  | false | No action
false | true  | groups_cache_.insert(groups[groups_pos]) and val = curr
false | false | No action
```
i.e. the behaviour only depends on `is_first_group_el`, not on `grp_idx `

alexowens90 (changes requested):

cpp/arcticdb/processing/aggregation.cpp
Hind-M: Yes indeed! I guess leftovers when implementing, thanks!

alexowens90 (approved):

PR #1093, opened 2023-11-22
Change tmpdir to tmp_path

G-D-Petrov:
#### Reference Issues/PRs
Fix for issue #496 

Change tmpdir to tmp_path because according to the pytest docs, the tmpdir is depreciated and tmp_path is the way to create temporary paths that are safe to use in multiprocessing setups such as pytest-split and  pytest-xdist


poodlewars (approved):
This would have been a lot easier to review without the Black changes. We're not enforcing Black compliance at the moment as we had issues getting the Black checks consistent across different platforms

PR #1095, opened 2023-11-22
DRAFT: Fix QB floats issue by updating the hash

### LINKED ISSUE #1035: Using QueryBuilder on grouping column with floats fails on Linux Conda
### Describe the bug

Using QueryBuilder on grouping column with floats fails on Linux Conda.
The bigger problem is that this doesn't reproduce on all versions.
It has been experienced only on Linux Conda on self-hosted ec2 runners and only on release builds.

Can be reproduced with the following test - test_group_on_float_column_with_nans

### Steps/Code to Reproduce

``` python 
ac = Arctic("lmdb://test_dir")
ac.create_library("some_lib")
lib = ac["some_lib"]
sym = "test_group_on_float_column_with_nans"
df = pd.DataFrame({"grouping_column": [1.0, 2.0, np.nan, 1.0, 2.0, 2.0], "agg_column": [1, 2, 3, 4, 5, 6]})
lib.write(sym, df)
expected = df.groupby("grouping_column").agg({"agg_column": "sum"})
q = QueryBuilder()
q = q.groupby("grouping_column").agg({"agg_column": "sum"})
received = lib.read(sym, query_builder=q).data
received.sort_index(inplace=True)
```

### Expected Results

The columns are correctly grouped like this:

1.0                       5
2.0                      13

Python: 3.10.13 | packaged by conda-forge | (main, Oct 26 2023, 18:07:37) [GCC 12.3.0]
OS: Linux-6.2.0-1015-aws-x86_64-with-glibc2.35
ArcticDB: 1.4.1.dev0

### Backend storage used

LMDB

### Additional Context

_No response_

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1035 

The problem in #1035 seems to be caused only on some GCC compilers when we use the maximum optimizations (O3).
In that case, the hashing functions doesn't work correctly and is producing bogus hashes.

This PR updates the hashing function to use the [XXH3 family of xxhash functions](https://github.com/Cyan4973/xxHash).
This should make the hashing both faster and more consistent. 

#### Other comments
**This might be a breaking change, so should be evaluated carefully.**

### DIFF
--- a/cpp/arcticdb/processing/test/test_clause.cpp
+++ b/cpp/arcticdb/processing/test/test_clause.cpp
@@ -58,10 +58,33 @@ TEST(Clause, Partition) {
     auto entity_ids = Composite<EntityIds>(push_entities(component_manager, std::move(proc_unit)));
     auto partitioned = gather_entities(component_manager, partition.process(std::move(entity_ids)));
 
-    std::vector<std::unordered_set<int8_t>> tags = {{1, 3}, {2}};
-    std::array<size_t, 2> sizes = {180, 90};
+    std::vector<std::unordered_set<int8_t>> tags = {{1}, {3}, {2}};
+    std::array<size_t, 3> sizes = {90, 90, 90};
+    ASSERT_EQ(tags.size(), partitioned.size());
     for (auto inner_seg : folly::enumerate(partitioned.as_range())){
         segment_scalar_assert_all_values_equal<int8_t>(*inner_seg, ColumnName("int8"), tags[inner_seg.index], sizes[inner_seg.index]);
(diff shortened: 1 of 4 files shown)


alexowens90:
This will not work, as `XXH32` generates 32 bit hashes, and we store hashes in keys:
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/entity/atom_key.hpp#L28
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/entity/key.hpp#L43

alexowens90:
Superseded by #1261

PR #1100, opened 2023-11-24
[Column] Move some definitions to cpp file

Hind-M:
With an aim of reducing build time and improve code readability, moving some definitions from header to cpp file.
Follow-up to #1092


Hind-M:
> I'm fine with this sort of lift and shift change. We should make sure blame ignores this change on github though: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

I added the `.git-blame-ignore-revs` file containing the commits to be ignored.

alexowens90 (approved):

alexowens90:
> > I'm fine with this sort of lift and shift change. We should make sure blame ignores this change on github though: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
> 
> I added the `.git-blame-ignore-revs` file containing the commits to be ignored.

OK, make sure you use the "Create a merge commit", not "Rebase and merge" option when merging, otherwise the hashes will change.

alexowens90:
> > > > I'm fine with this sort of lift and shift change. We should make sure blame ignores this change on github though: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
> > > 
> > > 
> > > I added the `.git-blame-ignore-revs` file containing the commits to be ignored.
> > 
> > 
> > OK, make sure you use the "Create a merge commit", not "Rebase and merge" option when merging, otherwise the hashes will change.
> 
> "Create a merge commit" is not enabled on this repository (I guess to keep a clean history). We could maybe "Squash and merge" and then update the `.git-blame-ignore-revs` file right away in a PR with the correct commit hash?

Yep looks like that's the best option @jjerphan @Klaim @JohanMabille please follow this methodology for any other PRs that are jus shifting implementations from hpp to cpp

PR #1127, opened 2023-12-05
Only build on pull request events

poodlewars:
We're currently spending about $1500 USD per month on EC2 runners for our CI.

It probably isn't all necessary because currently we're building every push.

This changes it so that we build automatically when pull requests are:

- Opened
- Synchronized (ie, commits are pushed to the pull request)
- Reopened

It will still be possible to build your branch manually without a PR using the Github Actions UI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,8 +1,10 @@
 name: Build and Test
 on:
   push:
+    tags: ["v**"]
+    branches: ["master"]
+  pull_request:
     branches: ["**"]
-    tags: [v**]
     paths:
       - .github/workflows/*.yml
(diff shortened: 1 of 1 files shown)


alexowens90 (commented):
Looks like this wouldn't build on new tags, which we'll want to be automatic for releases

jamesmunro (approved):

qc00 (commented):
There is already built-in feature to not use the EC2 runner. You just need to set `compile-override` to `ubuntu-latest` instead of `compile-on-ec2`.

It will be much better than not running the build on branches at all.

alexowens90 (approved):

PR #1128, opened 2023-12-05
Use same region for S3 and EC2 to avoid data transfer costs

poodlewars:
We paid about $800 on transfer costs in November this year, forming the huge majority of our S3 related costs.

The root cause seems to be data transfer costs between eu-west-2 (London), to eu-west-1 (Ireland).

I can only think that this would be transfers between our EC2 workers (eu-west-1) and our S3 storage (eu-west-2). Transfers in the same region should be free, so it's possible we can eliminate this cost completely.

### DIFF
--- a/.github/workflows/persistent_storage.yml
+++ b/.github/workflows/persistent_storage.yml
@@ -8,9 +8,9 @@ on:
       python3:           {default: -1, type: number, description: Python 3 minor version}
       arcticdb_version:  {default: "", type: string, description: The version of ArcticDB that will be installed and used for seed/verify }
       # TODO: Maybe we should add these as GitHub variables
-      bucket:            {default: 'arcticdb-ci-test-bucket-01', type: string, description: The name of the S3 bucket that we will test against}
-      endpoint:          {default: 's3.eu-west-2.amazonaws.com', type: string, description: The address of the S3 endpoint}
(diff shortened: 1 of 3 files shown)


jamesmunro (approved):

.github/workflows/persistent_storage.yml
```diff
@@ -8,9 +8,9 @@ on:
       python3:           {default: -1, type: number, description: Python 3 minor version}
       arcticdb_version:  {default: "", type: string, description: The version of ArcticDB that will be installed and used for seed/verify }
       # TODO: Maybe we should add these as GitHub variables
-      bucket:            {default: 'arcticdb-ci-test-bucket-01', type: string, description: The name of the S3 bucket that we will test against}
-      endpoint:          {default: 's3.eu-west-2.amazonaws.com', type: string, description: The address of the S3 endpoint}
-      region:            {default: 'eu-west-2', type: string, description: The S3 region of the bucket}
+      bucket:            {default: 'arcticdb-ci-test-bucket-02', type: string, description: The name of the S3 bucket that we will test against}
+      endpoint:          {default: 's3.eu-west-1.amazonaws.com', type: string, description: The address of the S3 endpoint}
+      region:            {default: 'eu-west-1', type: string, description: The S3 region of the bucket}
```
qc00, line 13: These are no longer used either, so should be removed as well.

PR #1134, opened 2023-12-06
Check TTL in do_lock

joe-iddon:

The symbol list is protected by a storage lock, so only one writer can compact it at a time. This is implemented using a key of type `LOCK` of which there is almost at most one of in the storage backend. Writers will fail early if they see this key.

The method `lock()` implements a check for the TTL of the storage lock in case it was left locked by a failed write (default TTL is one day). But the `try_lock()` method used in the `list_symbols()` operation does not bother to check the TTL. This PR fixes that and adds another unit test.

The consequence of the previous behaviour was quite bad: libraries could get stuck in a permanent state of an uncompacted symbol list. So for tens of thousands of symbols, or frequent writers (each write inserts a new `__add__` to the symbol list), the `list_symbols()` method would grind to a halt since it would never be able to compact the symbols.

Already tested internally.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [x ] Have you updated the relevant docstrings, documentation and copyright notice?
</details>

### DIFF
--- a/cpp/arcticdb/util/test/test_storage_lock.cpp
+++ b/cpp/arcticdb/util/test/test_storage_lock.cpp
@@ -112,7 +112,7 @@ TEST(StorageLock, Contention) {
     collect(futures).get();
 
     ASSERT_EQ(lock_data->atomic_, lock_data->vol_);
-    //ASSERT_EQ(lock_data->contended_, true); Alas Jenkins is too rubbish for this, uncomment for tests on headnode
+    ASSERT_EQ(lock_data->contended_, true);
 }
 
 struct PessimisticLockTask {
@@ -182,6 +182,46 @@ struct ForceReleaseLockTask {
     }
 };
 
+struct OptimisticForceReleaseLockTask {
+    std::shared_ptr<LockData> data_;
+    size_t timeout_ms_;
+    size_t retry_ms_;
+
+    OptimisticForceReleaseLockTask(std::shared_ptr<LockData> data, size_t timeout_ms, size_t retry_ms) :
+            data_(std::move(data)),
+            timeout_ms_(timeout_ms),
+            retry_ms_(retry_ms)
+    {
+    }
+
+    folly::Future<folly::Unit> operator()() {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        StorageLock<> lock{data_->lock_name_};
+        bool contended = true;
+        size_t total_wait = 0;
+        while (contended && total_wait < timeout_ms_) {
+            contended = !lock.try_lock(data_->store_);
+            std::this_thread::sleep_for(std::chrono::milliseconds(retry_ms_));
+            total_wait += retry_ms_;
+
+        }
+        if (contended) {
+            data_->timedout_ = true;
+            data_->contended_ = true;
+        } else {
+            ++data_->vol_;
+            // This should be unnecessary as we are already locked
+            std::lock_guard l{data_->mutex_};
+            ++data_->atomic_;
+            // Dont unlock
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(5));
+        // Clean up locks to avoid "mutex destroyed while active" errors on Windows debug build
+        lock.unlock(data_->store_);
+        return makeFuture(Unit{});
+    }
+};
+
 TEST(StorageLock, Wait) {
     SKIP_MAC("StorageLock is not supported");
     using namespace arcticdb;
@@ -215,15 +255,18 @@ TEST(StorageLock, Timeouts) {
 }
 
 TEST(StorageLock, ForceReleaseLock) {
+    // This test spawns jobs to wait for the lock with a large timeout.
+    // The lock is never released, but the TTL is configured to something reasonable so that
+    // they will all eventually acquire the lock from successive TTL expirations.
     SKIP_MAC("StorageLock is not supported");
     using namespace arcticdb;
 
     auto lock_data = std::make_shared<LockData>(4);
     folly::FutureExecutor<folly::CPUThreadPoolExecutor> exec{4};
-    // This is set in nanoseconds => 1ms
-    ConfigsMap::instance()->set_int("StorageLock.TTL", 1000 * 1000);
-
-    std::vector<std::shared_ptr<StorageLock<>>> locks;
+    // This is set in milliseconds => 50ms for the preempting check
+    ConfigsMap::instance()->set_int("StorageLock.WaitMs", 50);
+    // This is set in nanoseconds => 200ms for the TTL
+    ConfigsMap::instance()->set_int("StorageLock.TTL", 200 * 1000 * 1000);
 
     // Create a first lock that the others will have to force release
     auto first_lock = StorageLock<>(lock_data->lock_name_);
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -182,6 +182,46 @@ struct ForceReleaseLockTask {
...
+            total_wait += retry_ms_;
+
+        }
+        if (contended) {
+            data_->timedout_ = true;
+            data_->contended_ = true;
+        } else {
+            ++data_->vol_;
+            // This should be unnecessary as we are already locked
+            std::lock_guard l{data_->mutex_};
```
vasil-pashov, line 214: `LockData::atomic_` is of type `std::atomic<uint64_t>` (which is provides the strongest memory ordering (Sequentially-consistent ordering)). There is not need to guard touching this variable with a lock. It is guaranteed by the standard that no data race is happening.

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -242,3 +285,39 @@ TEST(StorageLock, ForceReleaseLock) {
...
+    using namespace arcticdb;
+
+    auto lock_data = std::make_shared<LockData>(4);
+    folly::FutureExecutor<folly::CPUThreadPoolExecutor> exec{4};
+    // This is set in milliseconds => 50ms for the preempting check
+    ConfigsMap::instance()->set_int("StorageLock.WaitMs", 50);
+    // This is set in nanoseconds => 200ms for the TTL
+    ConfigsMap::instance()->set_int("StorageLock.TTL", 200 * 1000 * 1000);
+
+    std::vector<std::shared_ptr<StorageLock<>>> locks;
```
vasil-pashov, line 304: Where is this used?

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -182,6 +182,46 @@ struct ForceReleaseLockTask {
...
+
+    OptimisticForceReleaseLockTask(std::shared_ptr<LockData> data, size_t timeout_ms, size_t retry_ms) :
+            data_(std::move(data)),
+            timeout_ms_(timeout_ms),
+            retry_ms_(retry_ms)
+    {
+    }
+
+    folly::Future<folly::Unit> operator()() {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
vasil-pashov, line 198: Why is there a sleep here?

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -182,6 +182,46 @@ struct ForceReleaseLockTask {
...
+            contended = !lock.try_lock(data_->store_);
+            std::this_thread::sleep_for(std::chrono::milliseconds(retry_ms_));
+            total_wait += retry_ms_;
+
+        }
+        if (contended) {
+            data_->timedout_ = true;
+            data_->contended_ = true;
+        } else {
+            ++data_->vol_;
```
vasil-pashov, line 212: Volatile does not protect from data races, moreover volatile variables are not atomic and must be protected by a lock when concurrent reads and writes are done. The only thing that volatile does is to prevent the compiler from optimizing the value away it has no relation to the (concurrent) memory model. In the test the check data_->vol_ == 4 will produce a flaky test.

cpp/arcticdb/util/test/test_storage_lock.cpp
joe-iddon: Ah thanks. I noticed it was unused and deleted only one of the two places.

cpp/arcticdb/util/test/test_storage_lock.cpp
joe-iddon: The idea is to have a volatile member that we *want* to be corrupted/not incremented correctly in the case that the storage lock is not doing its job properly.

The atomic one is then a comparison, since this should definitely be updated correctly due to the nature of atomic type in C++.

I agree we don't need to wrap the atomic in a mutex, that is just how it was always done, and I'm not sure why. I think it's a mistake. If you agree should I remove this mutex? I.e. all uses of `data_->mutex_`?

cpp/arcticdb/util/test/test_storage_lock.cpp
joe-iddon: Again I'm not sure, just how these unit tests were originally implemented. Agree we should remove then?

cpp/arcticdb/util/test/test_storage_lock.cpp
joe-iddon: Please see my comment below about scrapping this mutex.

cpp/arcticdb/util/test/test_storage_lock.cpp
vasil-pashov: > The idea is to have a volatile member that we want to be corrupted/not incremented correctly in the case that the storage lock is not doing its job properly.

I'm not following on this. In the test you're doing the same check for the vol_ as for atmoic_. If the lock is not doing the correct thing vol_ will be a random value between 1 and 4

>  If you agree should I remove this mutex?
Remove it in all places where it guards the atomic, just make sure it's not guarding anything else.

cpp/arcticdb/util/test/test_storage_lock.cpp
vasil-pashov: Seems unnecessary to me. Maybe someone with more context can take a look as well.

vasil-pashov:
@willdealtry I see you're the creator LockData. I have a few questions:
1. timedout_ and contended_ are non-volatile, non-atomic but in several places of this file we're doing concurrent writes. It's most definitely a data race (given the compiler does not omit them due to the lack of volatile). Was this indented? AFAIK on x86 writes are atomic there's no chance of writing half the bool, however I think it's UB to write to them from multiple threads even if they are used just as a flag which can only be set from false to true. Also I'm not quite sure when will the memory be sync-ed, with relaxed atomic the thread join will issue a barrier which will sync the variables. My suggestion is to just use atomic or relaxed atomic (if we care that much about performance)
2. vol_ is an ordinary volatile while atomic_ is atomic. There are a few places where we require vol_ to be the same as atomic_, however vol_ is not guarded by a lock. This does not seem right, AFAIK volatile only prevents the compiler from optimizing the variable away and does not provide any atomicity nor memory ordering. I don't think it's reasonable to compare both (nor to compare vol_ against a specific value as Joe does). Was this on purpose?

cpp/arcticdb/util/test/test_storage_lock.cpp
joe-iddon: I think fine to remove. The tests still pass and there's no reason to have them.

joe-iddon:
> @willdealtry I see you're the creator LockData. I have a few questions:
> 
> 1. timedout_ and contended_ are non-volatile, non-atomic but in several places of this file we're doing concurrent writes. It's most definitely a data race (given the compiler does not omit them due to the lack of volatile). Was this indented? AFAIK on x86 writes are atomic there's no chance of writing half the bool, however I think it's UB to write to them from multiple threads even if they are used just as a flag which can only be set from false to true. Also I'm not quite sure when will the memory be sync-ed, with relaxed atomic the thread join will issue a barrier which will sync the variables. My suggestion is to just use atomic or relaxed atomic (if we care that much about performance)
> 2. vol_ is an ordinary volatile while atomic_ is atomic. There are a few places where we require vol_ to be the same as atomic_, however vol_ is not guarded by a lock. This does not seem right, AFAIK volatile only prevents the compiler from optimizing the variable away and does not provide any atomicity nor memory ordering. I don't think it's reasonable to compare both (nor to compare vol_ against a specific value as Joe does). Was this on purpose?

After a discussion (Vasil/Joe):

1. We discussed that the `timeout_` and `contended_` members should be `std::atomic` so that multiple threads can fail the test at the same time and not get UB by all trying to report that at once.
2. After explaining that the whole motivation for these tests is to check the storage lock, we deliberately are checking that the volatile member *is* protected by the storage lock being held. If we get the UB that arises from multiple threads writing to it at the same time, then that is great as the test will hopefully fail and then we will know there's a problem with the storage lock implementation. We also check an atomic member is incremented correctly so that we have another way of debugging the test if it goes wrong.

I did discover that Alex's change to fix Windows complaining that the local storage lock mutex is destructed without being unlocked (deliberately), when compiling a debug build was actually invalidating the purpose of the TTL expiration tests since at the end of each test we were calling unlock on the storage lock anyway, not "forgetting to unlock it" as originally intended for the test. I've fixed this now as well. Detail: Added a test method to the storage lock to allow manually just unlocking the local mutex, but leaving the store's `LOCK` key in place, which is what we would like to happen to simulate, e.g. someone CTRL-C on the Python process whilst compacting the symbol list and leaving the `LOCK` key in place.

vasil-pashov (approved):

PR #1137, opened 2023-12-07
Clean and reorganize `OffsetString` and `StringPool`

Hind-M:
- [x] Move definitions to `cpp` files
- [x] Remove unnecessary headers inclusions
- [x] Remove `OffsetString::offset_t` and  `StringPool::offset_t` in favor of `entity::position_t` (same thing, different names). This will avoid including `offset_string.hpp` and `string_pool.hpp` just for these while `types.hpp` is already there
- [ ] Define types in a specific file to be included where needed instead of the whole `types.hpp` (to be done in an upcoming PR)

### DIFF
--- a/cpp/arcticdb/codec/codec.cpp
+++ b/cpp/arcticdb/codec/codec.cpp
@@ -5,6 +5,7 @@
  * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
  */
 #include <arcticdb/codec/codec.hpp>
+#include <arcticdb/column_store/string_pool.hpp>
 #include <arcticdb/stream/protobuf_mappings.hpp>
 #include <arcticdb/entity/performance_tracing.hpp>
 #include <arcticdb/codec/default_codecs.hpp>
--- a/cpp/arcticdb/column_store/column.cpp
+++ b/cpp/arcticdb/column_store/column.cpp
@@ -7,6 +7,7 @@
 
 #include <arcticdb/column_store/column.hpp>
 
+#include <arcticdb/column_store/string_pool.hpp>
 #include <arcticdb/util/offset_string.hpp>
 
 namespace arcticdb {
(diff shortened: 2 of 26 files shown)


alexowens90 (commented):
The hpp->cpp changes all look good. What is the motivation for changing `StringPool::offset_t` and `OffsetString::offset_t` to `entity::position_t` everywhere?

Hind-M:
> The hpp->cpp changes all look good. What is the motivation for changing `StringPool::offset_t` and `OffsetString::offset_t` to `entity::position_t` everywhere?

To simplify, and because using `StringPool::offset_t` and `OffsetString::offset_t` involves including the corresponding headers `offset_string.hpp` and `string_pool.hpp` while `types.hpp` is already included (which contains `entity::position_t` which is the same thing as `X::offset_t`).
A good next step would be to move the `position_t`-like types from `types.hpp` to a separate file to be included instead of shipping the whole `types.hpp`.

alexowens90:
> > The hpp->cpp changes all look good. What is the motivation for changing `StringPool::offset_t` and `OffsetString::offset_t` to `entity::position_t` everywhere?
> 
> To simplify, and because using `StringPool::offset_t` and `OffsetString::offset_t` involves including the corresponding headers `offset_string.hpp` and `string_pool.hpp` while `types.hpp` is already included (which contains `entity::position_t` which is the same thing as `X::offset_t`). A good next step would be to move the `position_t`-like types from `types.hpp` to a separate file to be included instead of shipping the whole `types.hpp`.

There might be a reason the separate types were originally created, maybe @willdealtry can weigh in?

willdealtry:
> > > The hpp->cpp changes all look good. What is the motivation for changing `StringPool::offset_t` and `OffsetString::offset_t` to `entity::position_t` everywhere?
> > 
> > 
> > To simplify, and because using `StringPool::offset_t` and `OffsetString::offset_t` involves including the corresponding headers `offset_string.hpp` and `string_pool.hpp` while `types.hpp` is already included (which contains `entity::position_t` which is the same thing as `X::offset_t`). A good next step would be to move the `position_t`-like types from `types.hpp` to a separate file to be included instead of shipping the whole `types.hpp`.
> 
> There might be a reason the separate types were originally created, maybe @willdealtry can weigh in?

I don't think it hugely matters. I wrote the string pool in the first or second week, it's possible that position_t didn't exist at that point. The purpose is similar enough (it's an offset in the buffer) that I think we can just use one.

alexowens90 (approved):

PR #1139, opened 2023-12-07
Use free runners for non-PR, non-release builds

qc00:
#### Reference Issues/PRs
Improvement over #1127 

* Build any branch even without PR using the slower free runners
* Not handling `pull_request` build for security


poodlewars:
I don't think we need to worry about the security aspect - PRs from non-maintainers need approval from us to run. Since the CI for this PR is also failing, I think we'll decline it.

PR #1143, opened 2023-12-08
Remove test parellism, and speed up test bottleneck

alexowens90:
Stops column stats tests from segfaulting by removing parallelism from the pytest runs in CI. This slows the build down from ~45 mins to ~1h, with the Windows integration tests the bottleneck at ~45 minutes. To alleviate some of this slowdown, the slowest test in a local run with 1 worker in each of the CPU and IO thread pools has been switched from S3 to lmdb, taking it from ~15 minutes to ~2 minutes locally.

### DIFF
--- a/build_tooling/parallel_test.sh
+++ b/build_tooling/parallel_test.sh
@@ -7,16 +7,6 @@ echo Saving results to ${TEST_OUTPUT_DIR:="$(realpath "$tooling_dir/../cpp/out")
 
 [[ -e ${PARALLEL_TEST_ROOT:=/tmp/parallel_test} ]] && rm -rf $PARALLEL_TEST_ROOT
 
-if [[ -n "${ARCTICDB_PERSISTENT_STORAGE_TESTS}" ]]; then
-  # We want to run the pytests sequentially to avoid races
-  splits=1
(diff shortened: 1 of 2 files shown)


build_tooling/parallel_test.sh
```diff
@@ -28,7 +18,7 @@ catch=`{ which catchsegv 2>/dev/null || echo ; } | tail -n 1`
 
     export ARCTICDB_RAND_SEED=$RANDOM
 
-    $catch python -m pytest -n $splits $PYTEST_XDIST_MODE -v --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
+    $catch python -m pytest $PYTEST_XDIST_MODE -v --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
```
poodlewars, line 21: Why do we still have XDIST_MODE, or the env var `PARALLEL_TEST_ROOT` ? In fact shouldn't we rename this file too?

poodlewars (approved):

build_tooling/parallel_test.sh
alexowens90: We might want to reinstate some parallelism later, but if not then yes, probably all of those things

PR #1147, opened 2023-12-11
Fix persistent tests

G-D-Petrov:
The persistent storages try to access logic in the pytest fixtures, which is not accessible.

This PR refactors some of the logic in the fixtures so it can be shared between the pytest and the storage_test files.

#### Any other comments?
There are also some other minor fixes.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/util/test.py
+++ b/python/arcticdb/util/test.py
@@ -5,6 +5,7 @@ Use of this software is governed by the Business Source License 1.1 included in 
 
 As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
 """
+
 import os
 from contextlib import contextmanager
 from typing import Mapping, Any, Optional, Iterable, NamedTuple, List, AnyStr
@@ -17,7 +18,6 @@ import time
 import attr
 from copy import deepcopy
 from functools import wraps
-import sys
 
 from arcticdb.config import Defaults
 from arcticdb.log import configure, logger_by_name
@@ -30,6 +30,16 @@ from arcticc.pb2.storage_pb2 import LibraryDescriptor, VersionStoreConfig
 from arcticdb.version_store.helper import ArcticFileConfig
 from arcticdb.config import _DEFAULT_ENVS_PATH
 from arcticdb_ext import set_config_int, get_config_int, unset_config_int
+
+
+def create_df(start=0, columns=1) -> pd.DataFrame:
+    data = {}
+    for i in range(columns):
+        col_name = chr(ord("x") + i)  # Generates column names like 'x', 'y', 'z', etc.
+        data[col_name] = np.arange(start + i * 10, start + (i + 1) * 10, dtype=np.int64)
+
+    index = np.arange(start, start + 10, dtype=np.int64)
+    return pd.DataFrame(data, index=index)
 
 
 def maybe_not_check_freq(f):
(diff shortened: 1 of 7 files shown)


python/arcticdb/util/test.py
```diff
@@ -90,22 +100,20 @@ def random_string(length: int):
...
-            "int8": random_integers(size, np.int8),
-            "int16": random_integers(size, np.int16),
-            "int32": random_integers(size, np.int32),
-            "int64": random_integers(size, np.int64),
-            "float32": np.random.randn(size).astype(np.float32),
-            "float64": np.arange(size, dtype=np.float64),
-            "bool": np.random.randn(size) > 0,
-        }
-    )
+    df = pd.DataFrame({
```
poodlewars, line 103: Please can you avoid making spurious formatting changes like this, as it increases the probability of conflicts with other changes and makes the code review harder

python/tests/util/dataframes.py
```diff
@@ -0,0 +1,12 @@
+import numpy as np
+import pandas as pd
+
+
+def create_df(start=0, columns=1) -> pd.DataFrame:
```
poodlewars, line 5: duplicated with `util/test.py`?

python/tests/util/storage_test.py
```diff
@@ -5,7 +5,42 @@
...
+    else:
+        path_prefix = os.getenv("ARCTICDB_PERSISTENT_STORAGE_UNIQUE_PATH_PREFIX")
+
+    clear = str(os.getenv("ARCTICDB_REAL_S3_CLEAR")).lower() in ("true", "1")
+
+    return endpoint, bucket, region, access_key, secret_key, path_prefix, clear
+
+
+def get_real_s3_uri(shared_path: bool = True):
+    # TODO: Remove this when the latest version that we support
```
poodlewars, line 29: Suggest linking in the comment to where "the latest version that we support" is defined?

poodlewars (changes requested):

python/arcticdb/util/test.py
G-D-Petrov: this is from the black formatter, I haven't touched this part of the code

qc00: Which version of `black` are you using and what's the exact command line?

python/tests/util/storage_test.py
```diff
@@ -5,7 +5,41 @@
 import re
 from datetime import datetime
 
-from arcticdb.storage_fixtures.s3 import real_s3_from_environment_variables
```
qc00, line 8: I just made these changes. Why are they being reverted without any meaningful explanation in the commit message?

qc00 (changes requested):

python/arcticdb/util/test.py
G-D-Petrov: latest `black`, with just `black --line-length 120 python/`

python/tests/util/storage_test.py
G-D-Petrov: I left a comment, essentially for this part of the code we are relying on installing arcticdb (both the latest version of the package and an older one, currently 3.0.0).
Because your changes are not part of the older version this is breaking the code.
As I have written in the comment, we will reincorporate your changes when they are part of the oldest version that we support.

python/arcticdb/util/test.py
G-D-Petrov: Some of the formatting was indeed fixed by using black==22.8.0 which is used by our docker image and the GHA hook

poodlewars (approved):

PR #1151, opened 2023-12-11
Support for `first` and `last` aggregators - string columns

### LINKED ISSUE #1105: Add support for `first` and `last` aggregators - string columns

Hind-M:
Fix #1105

### DIFF
--- a/cpp/arcticdb/pipeline/read_frame.cpp
+++ b/cpp/arcticdb/pipeline/read_frame.cpp
@@ -1061,7 +1061,7 @@ std::unique_ptr<StringReducer> get_string_reducer(
             const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, true);
             string_reducer = std::make_unique<UnicodeConvertingStringReducer>(column, context, frame, frame_field, alloc_width);
         } else {
-            const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, true);
+            const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, frame.fixed_str_as_dyn());
             string_reducer = std::make_unique<FixedStringReducer>(column, context, frame, frame_field, alloc_width);
         }
     } else {
--- a/cpp/arcticdb/pipeline/read_frame.cpp
+++ b/cpp/arcticdb/pipeline/read_frame.cpp
@@ -1061,7 +1061,7 @@ std::unique_ptr<StringReducer> get_string_reducer(
             const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, true);
             string_reducer = std::make_unique<UnicodeConvertingStringReducer>(column, context, frame, frame_field, alloc_width);
         } else {
-            const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, false);
+            const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, true);
             string_reducer = std::make_unique<FixedStringReducer>(column, context, frame, frame_field, alloc_width);
         }
     } else {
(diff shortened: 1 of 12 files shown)


python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -291,18 +333,47 @@ def test_first_agg_with_append(local_object_version_store):
...
-    df=data_frames(
-        [
-            column("grouping_column", elements=string_strategy, fill=string_strategy),
-            column("a", elements=numeric_type_strategies()),
-        ],
-        index=range_indexes(),
-    )
-)
-def test_hypothesis_last_agg_numeric(lmdb_version_store, df):
+def test_first_agg_strings_with_append_local(local_object_version_store):
```
Hind-M, line 336: This test is segfaulting while the same using `lmdb_version_store` below is working fine.

alexowens90 (changes requested):
Can we add some tests with `None` and `NaN` values in the aggregation columns?
Can we also make the `*_with_append` tests a bit more complicated, possibly using hypothesis and `lmdb_version_store_tiny_segment`?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -1013,7 +1013,7 @@ std::unique_ptr<StringReducer> get_string_reducer(
             const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, true);
             string_reducer = std::make_unique<UnicodeConvertingStringReducer>(column, context, frame, frame_field, alloc_width);
         } else {
-            const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, false);
+            const auto alloc_width = get_max_string_size_in_column(column.data().buffer(), context, frame, frame_field, slice_map, true);
```
alexowens90, line 1016: This will be prohibitively expensive for large dataframes with fixed-width strings that have not been aggregated, and so `false` would be sufficient. You need to pass the information in to this method of what this parameter should be.

alexowens90 (changes requested):

cpp/arcticdb/pipeline/read_frame.cpp
alexowens90: A hacky solution would be to set the orig_type in the aggregator/clause, so that was_coerced_from_dynamic_to_fixed is true in this case. Then add a comment that this is a big hack. Long term we want all reads to go through the `ComponentManager`, at which point it will decide what to do, so this is OK for now.

Hind-M: I first tried the hacky solution, but it's just bringing more complexity since the final dataframe to be compared with the expected one in the tests turns out to have different indexes and data regarding encoding, which involves additional hacking to do.
I think the first solution is fair enough (see this [commit](https://github.com/man-group/ArcticDB/pull/1151/commits/54a190e6a6f765ff7b88ecc0fda44bee015ddfd2#diff-b0ecf8c7fd4b7a415c9bb858bef66da846d90d1c7499a18057562c1375200972)).

Hind-M:
> Can we add some tests with `None` and `NaN` values in the aggregation columns? 

I added `None` values to the tests (cf. this [commit](https://github.com/man-group/ArcticDB/pull/1151/commits/4349d4021da85f5e59a57d04d3ae81a86ad4bd55)). Regarding `NaN` values, the tests were already including them, or are you thinking about something specific?

> Can we also make the `*_with_append` tests a bit more complicated, possibly using hypothesis and `lmdb_version_store_tiny_segment`?

Using hypothesis in  `*_with_append` tests gives an unexpected output when the given input dataframes are the following:
df1:
```
          grouping_column      a
0               0             0.0
```
df2:   
```
          grouping_column      a
0               0             0.0
```
df3:    
```
           grouping_column      a
0               0              0.0
1              00              0.0
```
using:
```
lib.write(symbol, df1)
lib.append(symbol, df2)
lib.append(symbol, df3)

```
Outputs are:
```
expected_df:                 
grouping_column     a
0                  0.0
00                 0.0
```
```
actual_dataframe:  
grouping_column     a
0                  0.0
0                  0.0
00                 0.0
```
I tried replicating this behavior with a basic test without using hypothesis, but it does give the right expected output dataframe. It seems that something is happening in the `PartitionClause` on the `repartition` level which makes it behave this way. I'm not sure what that could be yet...

Hind-M:
**Update**:  For some reason, all the values in the `grouping_column` corresponding to `0` don't get to be in the same bucket...

PR #1152, opened 2023-12-11
Build-time improvments: allocator.hpp, log.hpp, buffer.hpp

Klaim:
Reduces build-time of the whole project and most cpps in isolation.

I can't  make super reliable benchmarks but in general currently on Windows with MSVC/MSBuild with 12 jobs, hardware with 24 cores and 32GB and some other processes not active but still around, I get Build Insight build from-scratch:
- master: 6m33s
- this branch: 5m52s

Note that the difference stays stable but the actual numbers varies a lot, so it's just a general indication that build time have been reduced at least in my checks. For example, I had better times last week with the same computer but I did some system configuration tweaks on that computer in the weekend (it's my personal computer) and there migbht have been an impact.
I didn't check yet with 1 job as it's what's running on github's ci, neither did I check with ninja or on other platforms than Windows, but I can't reliably do these checks so we'll have to use some other reliable machine to be sure.
Note also that github's CI cannot be relied upon for build-time benchmark in my experience as depending on the machine performance can vary a lot (I once saw the same build on several branches take between 30 to 50 minutes, no stable build times.

We can see with Build Insights that allocator.hpp is not anymore a major source of build times:
master:
image

this branch:
image

So I think we can go with that PR and I'll start another PR for `data_erro.hpp`.

Visual Studio also have an experimental feature clarifying how much each header is used, it helped me a lot here. I'll try to do a pass in another PR for cleaning up includes that are not actually used.

### DIFF
--- a/cpp/arcticdb/codec/encode_v1.cpp
+++ b/cpp/arcticdb/codec/encode_v1.cpp
@@ -7,6 +7,7 @@
 #include <arcticdb/codec/encode_common.hpp>
 #include <arcticdb/codec/typed_block_encoder_impl.hpp>
 #include <arcticdb/column_store/memory_segment.hpp>
+#include <arcticdb/util/configs_map.hpp>
 
 namespace arcticdb {
     void add_bitmagic_compressed_size(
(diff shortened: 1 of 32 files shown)


Klaim:
Woops, missing keyword, not sure why msvc accept but ok.

jjerphan (approved):
Thank you, @Klaim.

Klaim:
@alexowens90 Looks like it's all green now

alexowens90 (approved):

PR #1155, opened 2023-12-11
Refactor: Move DataError method implementaitons into cpp

alexowens90:
Move method definitions for `DataError` class into `cpp` file

### DIFF
--- a/cpp/arcticdb/entity/data_error.hpp
+++ b/cpp/arcticdb/entity/data_error.hpp
@@ -9,8 +9,6 @@
 
 #include <string>
 
-#include <fmt/format.h>
-
 #include <arcticdb/pipeline/query.hpp>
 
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/entity/data_error.hpp
```diff
@@ -9,8 +9,6 @@
 
 #include <string>
 
-#include <fmt/format.h>
-
 #include <arcticdb/pipeline/query.hpp>
```
vasil-pashov, line 12: Looks like `<arcticdb/pipeline/query.hpp>` is not needed anymore.

alexowens90: It is, for `VersionQueryType`

vasil-pashov (approved):

PR #1159, opened 2023-12-12
Replace defragmentation logic with a simple read and write

joe-iddon:
#### Reference Issues/PRs

> [!WARNING]
> **It was decided that this is overly cautious and our C++ implementation is robust enough.**
> **Superseded by a by #1163 which just fixes up the C++ layer.**

Implements the suggestion in PR #1158 to replace the defragmentation logic with a simple read and write in the Python layer.

> [!WARNING]  
> ### API Changes 
> We no longer support the `segment_size` parameter. Please use the library option if you need to change the `segment_row_size` for a particular library.

#### Any other comments?

- We still return a `VersionedItem` without the `.data` property, so that the API will remain consistent when we restore the C++ implementation.
- This implementation in the Python layer will be temporary because in the C++ layer we can avoid reading `TABLE_DATA` keys at all by skipping those that are already defragmented. This is especially important for defragmenting a few appends to very large symbols which do not fit in memory, for instance.
- Dynamic schema and empty types should be considered as potential issues with the Python read+write implementation. E.g. if a user managed to write a mix of `int` and `nan` as an `int` column then reading and writing that via Pandas would convert to a `float` column. It doesn't seem possible with the current API to be able to get data into a symbol of this format though at the moment. We should tests sufficient hypothesis tests for this.
- We could have an `optimise_memory_use` parameter to allow use of both versions. Eventually we'd deprecate this and everyone would use the C++ implementation.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -2785,15 +2785,11 @@ class NativeVersionStore:
         if self._lib_cfg.lib_desc.version.write_options.bucketize_dynamic:
             raise ArcticDbNotYetImplemented(f"Support for library with 'bucketize_dynamic' ON is not implemented yet")
 
-        result = self.version_store.defragment_symbol_data(symbol, segment_size)
-        return VersionedItem(
-            symbol=result.symbol,
-            library=self._library.library_path,
-            version=result.version,
-            metadata=None,
-            data=None,
-            host=self.env,
-        )
+        proto_cfg = self._lib_cfg.lib_desc.version.write_options
+        if segment_size and segment_size != self.resolve_defaults("segment_row_size", proto_cfg):
+            raise ArcticDbNotYetImplemented(f"Support for segment_size parameter is not implemented yet")
+
+        return self.write(symbol, self.read(symbol).data)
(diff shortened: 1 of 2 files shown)


python/arcticdb/version_store/_store.py
```diff
@@ -2785,15 +2785,11 @@ def defragment_symbol_data(self, symbol: str, segment_size: Optional[int] = None
...
-            symbol=result.symbol,
-            library=self._library.library_path,
-            version=result.version,
-            metadata=None,
-            data=None,
-            host=self.env,
-        )
+        proto_cfg = self._lib_cfg.lib_desc.version.write_options
+        if segment_size and segment_size != self.resolve_defaults("segment_row_size", proto_cfg):
+            raise ArcticDbNotYetImplemented(f"Support for segment_size parameter is not implemented yet")
```
poodlewars, line 2790: Deprecation warning normal way to do the API upgrade path, rather than breaking?

poodlewars (approved):

PR #1168, opened 2023-12-17
Bug: Add TypeDescriptor type to dtype

8W9aG:

* dtype is not a string but in fact a TypeDescriptor
* Reflect this in the typing hints to allow for static analysers to find bugs based on this
assumption.


python/arcticdb/version_store/library.py
```diff
@@ -105,7 +106,7 @@ class NameWithDType(NamedTuple):
     """A named tuple. A name and dtype description pair."""
 
     name: str
```
joe-iddon, line 108: For the `index` property we might not have a name, in which case `None` is returned. See:

```python
>>> l.get_description('s').columns[0].name
'a'
>>> l.get_description('s').index.name
[None]
```

So I'd suggest we make this change:

```suggestion
    name: Optional[str]
```

python/arcticdb/version_store/library.py
```diff
@@ -105,7 +106,7 @@ class NameWithDType(NamedTuple):
...
-    dtype: str
+    dtype: TypeDescriptor
```
joe-iddon, line 109: This is great. Thank you for spotting this, and it looks correct to me. Please see the below though.

I noticed when testing your change that the `index` property of `SymbolDescription` actually returns a `NameWithDType` where the `dtype` is `List[TypeDescriptor]`. See below.

```python
>>> l
Library(Arctic(config=MEM()), path=l, storage=in_memory_storage)
>>> l.list_symbols()
['s']
>>> isinstance(l.get_description('s').index.dtype[0], arcticc.pb2.descriptors_pb2.TypeDescriptor)
True
>>> l.get_description('s').index.dtype
[value_type: INT
]
>>> isinstance(l.get_description('s').index.dtype[0], arcticc.pb2.descriptors_pb2.TypeDescriptor)
True
>>> isinstance(l.get_description('s').index.dtype, list)
True
```

This is due to in [`library.py`](https://github.com/man-group/ArcticDB/blob/904b0e5b88bdf9b7c3bcf430fc8dced32d4c05e0/python/arcticdb/version_store/library.py#L1537) we do:

```python
index = NameWithDType(info["col_names"]["index"], info["col_names"]["index_dtype"])
```

where `info["col_names"]["index_dtype"]` comes from [`_store.py`](https://github.com/man-group/ArcticDB/blob/904b0e5b88bdf9b7c3bcf430fc8dced32d4c05e0/python/arcticdb/version_store/_store.py#L2568) here:

```python
index_dtype = []
...
if input_type == "df":
    ...
    if index_type == "multi_index" or (index_type == "index" and index_metadata.is_not_range_index):
        ...
        index_dtype = [dtypes.pop(0)]
    else:
        # RangeIndex
        ...
        index_dtype = [dtype_desc]

    if index_type == "multi_index":
        for i in range(1, index_metadata.field_count + 1):
            ...
            index_dtype.append(dtypes.pop(0))
...
return {
    "col_names": {"columns": columns, "index": index, "index_dtype": index_dtype},
    ...
```

So therefore I'd suggest we keep the current API, but have the `dtype` optionally be a `List[TypeDescptor]` since to support the case of multi-level indexes, the `index` property of `SymbolDescriptor` is always a list.

```suggestion
    dtype: Union[TypeDescriptor, List[TypeDescriptor]]
```

joe-iddon (changes requested):
Thank you for taking the time to submit a PR to the codebase. I've left a few suggestions.

If you would like to have this merged, then please address these and squash your commits to a single one. Please include a declaration of acceptance of our Contributor License Agreement. Details for what we require can be found in the final step of [the agreement](https://github.com/man-group/ArcticDB/blob/master/Individual%20Contributor%20License%20Agreement.md).

I've noticed you have submitted a couple of [issues](https://github.com/search?q=author%3A8W9aG+repo%3Aman-group%2Farcticdb&type=issues&ref=advsearch&s=created&o=desc) to the repository too. We've had a particularly busy December, so apologies for not following up on these sooner.

We are grateful for your ideas and we would love to discuss them further with you as well as hearing about how you are finding ArcticDB.

If this sounds okay to you, then please do contact us at: arcticdb@man.com

Thank you again for your contributions. We are focused on making this software as good as it can be and that is made much easier with help from our users.

python/arcticdb/version_store/library.py
```diff
@@ -20,6 +20,7 @@
 from arcticdb.version_store._store import NativeVersionStore, VersionedItem, VersionQueryInput
 from arcticdb.version_store._normalization import denormalize_user_metadata
 from arcticdb_ext.exceptions import ArcticException
+from arcticdb_ext.types import TypeDescriptor
```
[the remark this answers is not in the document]
8W9aG: Done

joe-iddon:
Thank you for making the suggested changes. This looks good to me now.

If you could please squash the commits in this branch down to a commit with a commit message in accordance with the Contributor License Agreement (see below), then we can merge this.

If you need help with how to perform this, then please let me know.

> To make this document effective, you are required to include the following line as the last line of the commit message for each commit in your contribution. You must use your real name (no pseudonyms, and no anonymous contributions).
> “Signed-Off By Random J. Developer [random@developer.example.org](mailto:random@developer.example.org). By including this sign-off line I agree to the terms of the Contributor License Agreement”

python/arcticdb/version_store/library.py
```diff
@@ -104,8 +106,8 @@ def __repr__(self):
 class NameWithDType(NamedTuple):
     """A named tuple. A name and dtype description pair."""
 
-    name: str
-    dtype: str
+    name: Union[Optional[str], List[Optional[str]]]
+    dtype: Union[TypeDescriptor, List[TypeDescriptor]]
```
alexowens90, line 110: I believe the only reason these need to be `Union[<type>, List[<type>]]` is to handle multi-indexes? In which case I think it would be better to make `index` in the `SymbolDescription` class either `Tuple[NameWithDType]` where the tuple only has one element in the (more common) non-multi-index case, or `Union[NameWithDType, Tuple[NameWithDType]]` so that the most common case is less ugly.

PR #1170, opened 2023-12-18
Add a way to handle non-string values for index names

### LINKED ISSUE #1119: pybind cast error when writing with non-string index names
### Describe the bug

I receive the following error when writing a pandas dataframe to a library:

```
  File "/usr/local/lib/python3.10/site-packages/arcticdb/version_store/library.py", line 422, in write
    return self._nvs.write(
  File "/usr/local/lib/python3.10/site-packages/arcticdb/version_store/_store.py", line 578, in write
    vit = self.version_store.write_versioned_dataframe(
arcticdb_ext.exceptions.InternalException: pybind11::cast_error(Unable to cast Python instance of type <class 'list'> to C++ type 'std::vector<std::string, std::allocator<std::string> >')"
```

### Steps/Code to Reproduce

```python
from arcticdb import Arctic
import pandas as pd
import io

ac = Arctic(f"s3://XXX")
buf = io.StringIO("2023-11-27 00:00:00,0.73260,0.73260,0.73260,0.73260,7")
df = pd.read_csv(buf, parse_dates=[0], index_col=0, header=None)
df = df.rename(columns={0: "timestamp", 1: "Open", 2: "High", 3: "Low", 4: "Close", 5: "Volume"})
df.index = pd.to_datetime(df.index)
df.index = df.index.tz_localize('US/Eastern')
ac['XXX'].write('XXX', df)
```

### Expected Results

No pybind error is thrown and the dataframe is successfully written to arcticdb

Linux Ubuntu 20.04
Python 3.8.10
arcticdb 4.1.0

### Backend storage used

GCP S3

### Additional Context

_No response_

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1119 

Add a way to handle non-string index names


poodlewars (commented):
This will break when we try to round-trip indexes with integer names won't it? In the sense that users will write int-named-indexes, read the df out again, and get string-named-indexes.

I think the way to handle this is writing some extra info in `NormalizationMetadata` as well as doing this coercion here. Also needs a `pytest`.

python/arcticdb/version_store/_normalization.py
```diff
@@ -431,12 +436,14 @@ def _normalize_columns_names(columns_names, index_names, norm_meta, dynamic_sche
         else:
             if dynamic_schema and (counter[col] > 1):
                 raise ArcticNativeException("Same column names not allowed in dynamic_schema")
+            new_name = col_str
             if counter[col] > 1 or col in index_names:
                 new_name = "__col_{}__{}".format(col, 0 if dynamic_schema else idx)
+            if not isinstance(col, str) or new_name != col_str:
                 norm_meta.common.col_names[new_name].original_name = col_str
-                if isinstance(col, int):
-                    norm_meta.common.col_names[new_name].is_int = True
```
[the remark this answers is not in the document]
arcticdb-service-user: Yeah I will bring it back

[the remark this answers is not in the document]
G-D-Petrov: Yeah, I will bring it back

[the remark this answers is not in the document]
poodlewars: Thanks, it would be good to quickly test the interop here, ie can data written by this client be read by older client and vice versa, can data written by older client be read by this client?

poodlewars (commented):
I've been thinking a bit about this.

On reflection, Python pickles can be very problematic. We might for example have users write with one Python venv and then be unable to read their data on a different Python venv, if they write with a sufficiently odd index name.

I think it would instead be sensible to just restrict ourselves to string and int index names for now, and validate against other names. That should probably cover most use cases without introducing too much complexity. What do you think @G-D-Petrov ?

G-D-Petrov:
> I've been thinking a bit about this.
> 
> On reflection, Python pickles can be very problematic. We might for example have users write with one Python venv and then be unable to read their data on a different Python venv, if they write with a sufficiently odd index name.
> 
> I think it would instead be sensible to just restrict ourselves to string and int index names for now, and validate against other names. That should probably cover most use cases without introducing too much complexity. What do you think @G-D-Petrov ?

yeah that makes sense - should we also add doubles or only int and str, for now?

cpp/arcticdb/python/normalization_checks.cpp
```diff
@@ -153,7 +152,6 @@ bool check_pandas_like(bool is_append,
                                     stop);
                 }
 
-                normalization::check<ErrorCode::E_UPDATE_NOT_SUPPORTED>(is_append, "update() on ROWCOUNT index is not implemented");
```
poodlewars, line 156: Why is it OK to drop this check?

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -478,6 +478,59 @@ def test_no_inplace_index_array_modification(lmdb_version_store, sym, datetime64
...
 
 
+def test_index_names_roundtrip_datetime(lmdb_version_store, sym):
+    df = pd.DataFrame(data={"a": [1, 2, 3]}, index=pd.DatetimeIndex(["2020-01-01", "2020-01-02", "2020-01-03"]))
+    new_index = pd.Timestamp("2020-01-01")
+    df.index.rename(new_index, inplace=True)
+    lmdb_version_store.write(sym, df)
+    df.index.rename(str(new_index), inplace=True)
+    assert_frame_equal(lmdb_version_store.read(sym).data, df)
+    assert df.index.names[0] == str(new_index)
```
poodlewars, line 488: I thought we would just validate against these ones that we can't get to roundtrip properly, just with a better error message than the current:

```
InternalException: pybind11::cast_error(Unable to cast Python instance of type <class 'list'> to C++ type 'std::vector<std::string, std::allocator<std::string> >')
```

I think it will get quite hard to work with ArcticDB unless we keep the invariant that,

```
lib.write("s", df)
pandas.testing.assert_frame_equal(lib.read("s").data, df)
```

cpp/arcticdb/python/normalization_checks.cpp
G-D-Petrov: This check is not really meaningful because:
1. it is checking only the normalization data, so before my changes, if there is no normalization data, the appends would still go through. So the check is not really correct. And now we have cases with normalization data that previous passed but are now failing, which doesn't seem correct.
2. I have explicitly added tests that test appends/updates with ROWCOUNT and they seems to be working:
https://github.com/man-group/ArcticDB/blob/b085ec11202324644ee4141b8d201d269739e35e/python/tests/integration/arcticdb/version_store/test_update_with_date_range.py#L140

python/tests/unit/arcticdb/version_store/test_normalization.py
G-D-Petrov: We could do it like that, but for the column names, we already implicitly parse and use them as strings (unless they are ints).
If we add this verification, it could result in breaking code that worked before.

poodlewars (approved):

Issue #1082, opened 2023-11-20
Backport empty column handling fixes to `4.0.x` branch


alexowens90:

jjerphan:
Completed by #1079.

PR #1180, opened 2023-12-22
Forward port internal PR #1082

poodlewars:
Leave off flyweight optimization to avoid the boost dep.

The optimization can be done in the enterprise software that reads the op logs.

I also remove some dead code in `version_log.{c,h}pp`. `version_log.cpp` wasn't included in our build scripts. We can put the logic back in to the enterprise codebase.


joe-iddon:
Looks good to me. A few changes as clearly the symbol list has been refactored on master and not moved to 4.0.x, e.g. from William's PR https://github.com/man-group/ArcticDB/pull/796

joe-iddon:
Looks good. Agree with William's internal comment that the `op_log.cpp` logic to store either a full `key_` or its constituents was the wrong way of doing things.

joe-iddon (approved):

PR #1181, opened 2023-12-22
Sorting speed improvements in SegmentInMemory 

IvoDD:
This pr introduces 4 different changes each of which speeding up a particular part of the sorting.

Each of the changes is in a separate commit and can be looked at on it's own.

The changes are:
- Improve create_jive_table by precomputing all column values we need to sort by
- Improve create_jive_table by optimizing away unneeded sorting
- Improve sort_external by replacing an unneeded bm::bvector<> which is slow to set
- Rewrite sort_external for sparse columns to improve performance drastically

The total speed improvements are as follows (individual changes can be seen in commit messages):

Before this pr:
```
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BM_sort_shuffled/100000/100           169 ms          156 ms           10
BM_sort_shuffled/1000000/1            659 ms          608 ms           10
BM_sort_ordered/100000/100           36.6 ms         33.8 ms           10
BM_sort_sparse/100000/100             823 ms          760 ms           10
```

After this pr:
```
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
BM_sort_shuffled/100000/100_mean         92.1 ms         85.0 ms            3
BM_sort_shuffled/1000000/1_mean           173 ms          160 ms            3
BM_sort_ordered/100000/100_mean          7.26 ms         6.70 ms            3
BM_sort_sparse/100000/100_mean            110 ms          102 ms            3
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/column_store/column.cpp
+++ b/cpp/arcticdb/column_store/column.cpp
@@ -220,64 +220,60 @@ void Column::append(const Column& other, position_t at_row) {
                 row_count(), sparse_map().count());
 }
 
+void Column::physical_sort_external(std::vector<uint32_t> &&sorted_pos, size_t physical_rows) {
+    auto& buffer = data_.buffer();
+    type().visit_tag([&buffer, &sorted_pos, &physical_rows] (auto tdt) {
+        using TagType = decltype(tdt);
+        using RawType = typename TagType::DataTypeTag::raw_type;
+
+        for (auto i=0u; i<physical_rows; ++i){
+            if (i != sorted_pos[i]){
+                auto& current = buffer.cast<RawType>(i);
+                // Amortized O(1) complexity, because each iteration places an element where it's supposed to go
+                // and once an element is in it's sorted position we never move it.
+                while (i != sorted_pos[i]){
+                    auto move_to = sorted_pos[i];
+                    std::swap(sorted_pos[i], sorted_pos[move_to]);
+                    std::swap(current, buffer.cast<RawType>(move_to));
+                }
+            }
+        }
+    });
+}
+
 void Column::sort_external(const JiveTable& jive_table) {
     auto rows = row_count();
     if(!is_sparse()) {
         auto sorted_pos = jive_table.sorted_pos_;
-        auto& buffer = data_.buffer();
-        type().visit_tag([&buffer, &sorted_pos, &rows] (auto tdt) {
-            using TagType = decltype(tdt);
-            using RawType = typename TagType::DataTypeTag::raw_type;
-
-            for (auto i=0u; i<rows; ++i){
-                if (i != sorted_pos[i]){
-                    auto& current = buffer.cast<RawType>(i);
-                    // Amortized O(1) complexity, because each iteration places an element where it's supposed to go
-                    // and once an element is in it's sorted position we never move it.
-                    while (i != sorted_pos[i]){
-                        auto move_to = sorted_pos[i];
-                        std::swap(sorted_pos[i], sorted_pos[move_to]);
-                        std::swap(current, buffer.cast<RawType>(move_to));
-                    }
-                }
-            }
-        });
+        physical_sort_external(std::move(sorted_pos), rows);
     } else {
         const auto& sm = sparse_map();
-
         bm::bvector<>::enumerator en = sm.first();
         bm::bvector<>::enumerator en_end = sm.end();
         util::BitMagic new_map;
-
-        while (en < en_end) {
-            auto bv_index = *en;
-            new_map.set(jive_table.sorted_pos_[bv_index]);
-            ++en;
+        // The additional allocation is of the same size as the jive table
+        // and is needed for a significant speed improvement
+        auto sorted_logical_to_physical = std::vector<uint32_t>(jive_table.sorted_pos_.size());
+        for (auto physical=0u; en < en_end; ++physical, ++en) {
+            auto logical = *en;
+            auto sorted_logical = jive_table.sorted_pos_[logical];
+            new_map.set(sorted_logical);
+            sorted_logical_to_physical[sorted_logical] = physical;
         }
 
         util::check(new_map.count() == row_count(), "Mismatch between new bitmap size and row_count: {} != {}",
-            new_map.count(), row_count());
+                    new_map.count(), row_count());
 
-        auto new_buf = ChunkedBuffer::presized_in_blocks(data_.bytes());
+        auto physical_sort_pos = std::vector<uint32_t>(rows);
         en = new_map.first();
-        auto& buffer = data_.buffer();
-        auto rs = std::make_unique<util::BitIndex>();
-        sm.build_rs_index(rs.get());
+        for (auto sorted_physical=0u; sorted_physical<rows; ++sorted_physical, ++en){
+            auto sorted_logical = *en;
+            auto physical = sorted_logical_to_physical[sorted_logical];
+            physical_sort_pos[physical] = sorted_physical;
+        }
 
-        type().visit_tag([&jive_table, &sm, &rs, &buffer, rows, &en, &new_buf] (auto tdt) {
-            using TagType = decltype(tdt);
-            using RawType = typename TagType::DataTypeTag::raw_type;
-
-            for(auto i = 0u; i < rows; ++i) {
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/column_store/column.hpp
```diff
@@ -676,21 +688,16 @@ template <typename T>
 JiveTable create_jive_table(const Column& col) {
     JiveTable output(col.row_count());
     std::iota(std::begin(output.orig_pos_), std::end(output.orig_pos_), 0);
-    std::iota(std::begin(output.sorted_pos_), std::end(output.sorted_pos_), 0);
 
+    // Calls to scalar_at are expensive, so we precompute them to speed up the sort compare function.
+    auto values = col.template clone_scalars_to_vector<T>();
```
willdealtry, line 693: The thing that takes the time in scalar_at is the branching each time where we work out whether there's one contiguous block, a bunch of regular-sized ones of irregulat blocks. I think what we want to do eventually is return a variant column iterator that we can visit on which in the contiguous case just acts like a pointer + offset, in the regular case does a mask and only does the complicated stuff where blocks are irregular-sized (which is quite unusual). @JohanMabille will be looking at this sort of issue in due course, so I think this is fine for the moment

cpp/arcticdb/column_store/column.cpp
```diff
@@ -220,66 +220,65 @@ void Column::append(const Column& other, position_t at_row) {
...
+    util::check(sorted_pos.size() == physical_rows, "Mismatch between sorted_pos size and row_count: {} != {}",
+        sorted_pos.size(), physical_rows);
+
+    type().visit_tag([&buffer, &sorted_pos, &physical_rows] (auto tdt) {
+        using TagType = decltype(tdt);
+        using RawType = typename TagType::DataTypeTag::raw_type;
+
+        for (auto i=0u; i<physical_rows; ++i){
+            if (i != sorted_pos[i]){
+                auto& current = buffer.cast<RawType>(i);
```
willdealtry, line 236: You can eliminate the cost of the buffer.cast to i here by iterating over the physical blocks, since we need random access to the target but we know we're just forward-iterating the source. Alternatively we wait until we have a smarter iterator that would do this for us as described in the other comment.

cpp/arcticdb/column_store/column.cpp
```diff
@@ -220,66 +220,65 @@ void Column::append(const Column& other, position_t at_row) {
...
+                // Amortized O(1) complexity, because each iteration places an element where it's supposed to go
+                // and once an element is in it's sorted position we never move it.
+                while (i != sorted_pos[i]){
+                    auto move_to = sorted_pos[i];
+                    std::swap(sorted_pos[i], sorted_pos[move_to]);
```
willdealtry, line 241: Can you explain briefly how much difference this optimization makes, and why? It seems like it is introducing a fair bit of branching in the case where the numbers are randomly distributed

cpp/arcticdb/column_store/column.hpp
IvoDD: Agreed, if we have a fast random access iterator on the column data cloning to a vector will become obsolete. Should I return to the scalar_at calls now? (producing a significant slowdown (I think around 4x) which will be fixed later on with Johan's smarter iterators)

cpp/arcticdb/column_store/column.cpp
IvoDD: I think the cost of the buffer.cast to i is quite negligible and the added complexity to optimize it won't be worth it (and might make things slower).

The buffer.cast(i) will be called once for each permutation cycle (of length>1) in sorted_pos.

For random permutations the expected number of cycles is O(logn) see [here](https://en.wikipedia.org/wiki/Random_permutation_statistics#Expected_number_of_cycles_of_any_length_of_a_random_permutation).

So in the average case this would optimize O(logn) out of the O(n) calls to buffer.cast. In fact to optimize these O(logn) calls to cast we would need to do O(n) moves on the smarter iterator which might turn out to be slower.

cpp/arcticdb/column_store/column.hpp
willdealtry: No, I think leave it as it is for the moment, we'll just have to remember to go back and fix it later

cpp/arcticdb/column_store/column.cpp
```diff
@@ -220,66 +220,65 @@ void Column::append(const Column& other, position_t at_row) {
...
         util::BitMagic new_map;
-
-        while (en < en_end) {
-            auto bv_index = *en;
-            new_map.set(jive_table.sorted_pos_[bv_index]);
-            ++en;
+        // The additional allocation is of the same size as the jive table
+        // and is needed for a significant speed improvement.
+        // We could instead use a std::map and sacrifice some speed for smaller allocations.
+        auto sorted_logical_to_physical = std::vector<uint32_t>(jive_table.sorted_pos_.size());
```
vasil-pashov, line 261: I don't know how large is the jive table, if it's large this can have some overhead because the vector will call the default constructor of uint32_t setting everything to 0. Since we're not changing the size of the vector maybe a unique_ptr<uinr32_t[]> would be better. It the jive table is small this doesn't matter at all.

cpp/arcticdb/column_store/column.hpp
```diff
@@ -528,6 +525,18 @@ class Column {
         return *data_.buffer().ptr_cast<T>(bytes_offset(*physical_row), sizeof(T));
     }
 
+    // Copies all physical scalars to a std::vector<T>. This is useful if you require many random access operations
+    // and you would like to avoid the overhead of computing the exact location every time.
+    template<typename T>
+    std::vector<T> clone_scalars_to_vector() const {
+        auto values = std::vector<T>(row_count());
```
vasil-pashov, line 532: Maybe calling reserve + push_back could be faster

cpp/arcticdb/column_store/column.cpp
willdealtry: Right, I see. So since we often don't need the value of the current item at i, we'd end up having to calculate a random skip forward in the buffer anyway (i.e. the next time i != sorted_pos[i]), so we'd pretty much end up doing the same calculation buffer.cast does (or we'd end up doing a lot of branches to see if this was the end of the block, which as you say might be slower.

Cool, this looks good - nice work!

willdealtry (approved):

cpp/arcticdb/column_store/column.cpp
willdealtry: The jive table is the size of a column, but I think since there is likely to be only one jive table per segment we can handle this extra allocation. Ultimately it would be nice to allocated it as a buffer of regular-sized blocks from the slab allocator, but for that we need better iterator that can handle the contiguous memory vs evenly-sized blocks vs irregular blocks without branching on each iteration

IvoDD: A caveat with this allocation is that currently we are running the sort_external sequentially on each column of the output.
Also for this allocation we don't care what are the initial values for the allocated memory (because we set them afterwards).

So, we could also do exactly one allocation for the entire sort and reuse the memory in different calls of sort_external.
However I decided against that because this will block running the sort_external calls in parallel, which we could currently do (with the cost of extra memory).

cpp/arcticdb/column_store/column.hpp
IvoDD: I did the change. This is not part of the bottleneck so I don't expect benchmarks to change but still did it as it seems nicer.

cpp/arcticdb/column_store/column.cpp
willdealtry: Generally speaking we parallelize over segments, not over columns, so it's safe to assume that we only need to allocate the memory once.

IvoDD: Sounds good, I did the change. 
This will also likely improve the speed in case of many very sparse columns on which we call sort_external, for which I don't have a benchmark.
I reran the other benchmarks and results were unchanged.

PR #1188, opened 2024-01-02
Bugfix/1080/batch read with snapshots fix port

### LINKED ISSUE #1080: Port `batch_read` with snapshots fix to `4.0.x` branch

alexowens90:
Fixes #1080

### DIFF
--- a/cpp/arcticdb/async/task_scheduler.hpp
+++ b/cpp/arcticdb/async/task_scheduler.hpp
@@ -47,11 +47,11 @@ struct TaskSchedulerPtrWrapper{
         ptr_ = ptr;
     }
 
-    TaskScheduler* operator->() {
+    TaskScheduler* operator->() const {
         return ptr_;
     }
 
-    TaskScheduler& operator*() {
+    TaskScheduler& operator*() const {
         return *ptr_;
     }
 };
@@ -61,6 +61,7 @@ public:
     explicit InstrumentedNamedFactory(folly::StringPiece prefix) : named_factory_(prefix){}
 
     std::thread newThread(folly::Func&& func) override {
+        std::lock_guard lock{mutex_};
         return named_factory_.newThread(
                 [func = std::move(func)]() mutable {
                 ARCTICDB_SAMPLE_THREAD();
@@ -76,6 +77,7 @@ public:
 #endif
 
 private:
+    std::mutex mutex_;
     folly::NamedThreadFactory named_factory_;
 };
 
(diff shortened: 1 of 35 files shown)


cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -102,8 +104,7 @@ struct SchedulerWrapper : public SchedulerType {
 };
 
 inline int64_t get_cgroup_value(const std::string& cgroup_file) {
-    const auto path = std::filesystem::path{fmt::format("/sys/fs/cgroup/{}",cgroup_file)};
-    if(std::filesystem::exists(path)){
+    if(const auto path = std::filesystem::path{fmt::format("/sys/fs/cgroup/{}",cgroup_file)}; std::filesystem::exists(path)){
```
vasil-pashov, line 107: `std::filesystem::path` has an `operator/` which should be used for concatenating paths instead of manual formatting.

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -156,22 +155,27 @@ class TaskScheduler {
...
+        util::check(cpu_thread_count_ > 0 && io_thread_count_ > 0, "Zero IO or CPU threads: {} {}", io_thread_count_, cpu_thread_count_);
         ARCTICDB_RUNTIME_DEBUG(log::schedule(), "Task scheduler created with {:d} {:d}", cpu_thread_count_, io_thread_count_);
     }
 
     ~TaskScheduler() = default;
 
     template<class Task>
-    auto submit_cpu_task(Task &&task) {
+    auto submit_cpu_task(Task &&t) {
+        auto task = std::forward<decltype(t)>(t);
```
vasil-pashov, line 166: Why do we need to do this? Why not return `cpu_exec_.addFuture(std::move(t));`?

cpp/arcticdb/entity/types.hpp
```diff
@@ -124,7 +124,7 @@ enum class ValueType : uint8_t {
     // For instance, Pandas empty series whose types has not been specified is mapping to EMPTY.
     // When data is appended, the column type is inferred from the data and the column is promoted to the inferred type.
     EMPTY = 13,
-    COUNT // Not a real value type, should not be added to proto descriptor. Used to count the number of items in the enum
+    COUNT = 14 // Not a real value type, should not be added to proto descriptor. Used to count the number of items in the enum
```
vasil-pashov, line 127: NIT: I think this is a bit error prone. When adding new type people have to add value and update the count. Usually the count is left without a value so that it auto-increments

cpp/arcticdb/log/log.cpp
```diff
@@ -65,6 +65,14 @@ spdlog::logger &message() {
     return Loggers::instance()->message();
 }
 
+spdlog::logger &symbol() {
```
vasil-pashov, line 68: [[nodiscard]]?

vasil-pashov (approved):

PR #1211, opened 2024-01-08
Update git blame file for `OffsetString` and `StringPool` implementation move

Hind-M:
Cf. #1137


vasil-pashov (approved):

vasil-pashov:
This would need someone with push rights as it does not touch the source code and can_merge cannot finish. @alexowens90 @poodlewars

alexowens90 (approved):

PR #1212, opened 2024-01-08
build: Do not depend on protobuf-lite

jjerphan:

Depending on protobuf is sufficient.

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -33,8 +33,6 @@ if(NOT WIN32 AND ${SSL_LINK})
 endif()
 
 find_package(Protobuf REQUIRED)
-set(Protobuf_LIBRARY protobuf::libprotobuf)
-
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/CMakeLists.txt
```diff
@@ -33,7 +33,6 @@ if(NOT WIN32 AND ${SSL_LINK})
 endif()
 
 find_package(Protobuf REQUIRED)
-set(Protobuf_lite_LIBRARY protobuf::libprotobuf-lite)
 set(Protobuf_LIBRARY protobuf::libprotobuf)
```
vasil-pashov, line 36: While we're at it. find_package will define  `Protobuf_LIBRARY` no need to set it explicitly. In general I think that creating a variable to wrap a module name is an atipattern and we're using it in several places (protobuf, lz4, zstd...)

alexowens90 (approved):

PR #584, opened 2023-07-12
447 Consistently use `KeyNotFoundException`, `DuplicateKeyException` and `PermissionException` in `Storage`s

qc00:
## Error handling behaviour comparison

<table>
<tbody>
<tr>
<td>Storage</td>
<td>Function</td>
<td>Throws</td>
<td>For situation</td>
<td>Key in error msg</td>
<td>Changed in this PR</td>
</tr>
<tr>
<td rowspan="8">LMDB</td>
<td>Common</td>
<td>lmdb::error</td>
<td>transaction/low-level failure</td>
<td>None</td>
<td>No</td>
</tr>
<tr>
<td rowspan="2">write<br><br></td>
<td>DuplicateKeyException</td>
<td>overwriting&nbsp;AtomKey</td>
<td>One</td>
<td>No</td>
</tr>
<tr>
<td>std::runtime_error</td>
<td>other write failures</td>
<td>One</td>
<td>No</td>
</tr>
<tr>
<td>update</td>
<td>KeyNotFoundException</td>
<td>upsert=false on existing key</td>
<td>All</td>
<td>Added user-friendly msg<br>(Instead of "Composite: ")</td>
</tr>
<tr>
<td>read</td>
<td>KeyNotFoundException</td>
<td>key not found</td>
<td>All</td>
<td>No</td>
</tr>
<tr>
<td>key_exists</td>
<td>-</td>
<td>-</td>
<td>-</td>
<td>No</td>
</tr>
<tr>
<td>remove</td>
<td>KeyNotFoundException</td>
<td>keys already deleted</td>
<td>All</td>
<td>No</td>
</tr>
<tr>
<td>iterate</td>
<td>(None)</td>
<td>-</td>
<td>-</td>
<td>No</td>
</tr>
<tr>
<td rowspan="6">memory</td>
<td>write</td>
<td>util::check</td>
<td>overwriting AtomKey</td>
<td>One</td>
<td>DuplicateKeyException</td>
</tr>
<tr>
<td>update</td>
<td>util::check_rte</td>
<td>upsert=false on existing key</td>
<td>One</td>
<td>KeyNotFoundException</td>
</tr>
<tr>
<td>read</td>
<td>KeyNotFoundException</td>
<td>&nbsp;</td>
<td><strong>One</strong></td>
<td>No</td>
</tr>
<tr>
<td>key_exists</td>
<td>-</td>
<td>-</td>
<td>-</td>
<td>No</td>
</tr>
<tr>
<td>remove</td>
[...]

PR #1219, opened 2024-01-10
Add frequently used items at the top level of arcticdb

muhammadhamzasajjad:
- New imports added to `python/arcticdb/__init__.py`.
- One unit test added to `python/tests/unit/arcticdb/test_import.py` to check that the items are available at the top level.


DrNickClarke (approved):
Looks good

DrNickClarke (changes requested):
Please make the docs changes as part of the same PR

joe-iddon (changes requested):
Should remove this duplicate line:
```
```

Maybe add all exception types too? Some here: https://docs.arcticdb.io/latest/api/exceptions/ and also some names of the storage exceptions are enumerated in [this PR](https://github.com/man-group/ArcticDB/pull/584).

Things like "version not found" when you try and read a version which don't exist would be good to characterise and move to the top level imports e.g. `adb.NoSuchVersionException` or similar.

poodlewars (dismissed):

python/tests/unit/arcticdb/test_import.py
```diff
@@ -6,6 +6,36 @@
...
+        "WritePayload",
+        "ReadInfoRequest",
+        "ReadRequest",
+        "StagedDataFinalizeMethod",
+        "WriteMetadataPayload",
+    ]
+    find_imports(imports_list)
+
+
+def find_imports(list_imports):
```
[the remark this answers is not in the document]
muhammadhamzasajjad: I thought it would be better because `find_imports` is a general functionality. For example I could make this function slightly better like `def find_imports(library, imports_list)` and then you can test that for any given library, not just top-level-imports. But I can change it to your suggestion if you don't think it's helpful. Let me know your thoughts.

DrNickClarke (approved):

docs/mkdocs/docs/tutorials/fundamentals.md
```diff
@@ -7,10 +7,10 @@ This tutorial will walk through the fundamentals of ArcticDB:
 3. Reading data
 4. Modifying data
 
-To start, let's import Arctic:
+To start, let's import arcticdb:
```
joe-iddon, line 10: Prefer ArcticDB or `arcticdb` I think.

joe-iddon (approved):
Looks good to me if the CI passes. 👍

docs/mkdocs/docs/tutorials/fundamentals.md
muhammadhamzasajjad: Thank you. I shall fix that.

PR #1223, opened 2024-01-11
Bugfix arcticdb-man/62: Speedup reading wide dataframes that have no empty columns

alexowens90:
Covers off https://github.com/man-group/arcticdb-man/issues/62 for `4.0.x`

The modified for loop created a performance regression compared to the internal Man Group client when reading wide, shallow dataframes.

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -711,7 +711,6 @@ class DataFrameNormalizer(_PandasNormalizer):
         columns, denormed_columns, data = _denormalize_columns(item, norm_meta, idx_type, n_indexes)
 
         if not self._skip_df_consolidation:
-            columns_dtype = {} if data is None else {name: np_array.dtype for name, np_array in data.items()}
             df = DataFrame(data, index=index, columns=columns)
 
             # Setting the columns' dtype manually, since pandas might just convert the dtype of some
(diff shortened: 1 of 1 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -725,17 +724,17 @@ def denormalize(self, item, norm_meta):
...
+                possibly_empty_columns_names = [] if data is None else [name for name, np_array in data.items() if np_array.dtype in OBJECT_TOKENS]
+                for column_name in possibly_empty_columns_names:
+                    # TODO(jjerphan): Remove once pandas < 2 is not supported anymore.
+                    if len(df[column_name]) == 0:
+                        # Before Pandas 2.0, empty series' dtype was float, but as of Pandas 2.0. empty series' dtype became object.
+                        # See: https://github.com/pandas-dev/pandas/issues/17261
+                        # EMPTY type column are returned as pandas.Series with "object" dtype to match Pandas 2.0 default.
+                        # We cast it back to "float" so that it matches Pandas 1.0 default for empty series.
+                        # Pandas 0.X overrides the index of df to a RangeIndex if the index isn't explicitly provided here
+                        df[column_name] = pd.Series([], index=index dtype='float64')
```
jjerphan, line 737:
```suggestion
                        df[column_name] = pd.Series([], index=index, dtype='float64')
```

jjerphan (approved):
Thank you, @alexowens90.

I under-looked potential performance regression when adding this condition for supporting pandas 2.

Here is a single comment to fix the syntax error; with this changes all the tests of `python/tests/unit/arcticdb/version_store/test_normalization.py` pass.

I have opened https://github.com/man-group/ArcticDB/pull/1224 to suggest changes I could not provide easily via the GitHub's UI.

The changes LGTM if the CI checks pass.

python/arcticdb/version_store/_normalization.py
alexowens90: Good catch on the syntax. I agree that your implementation in #1224 is cleaner, I'll amend

PR #1225, opened 2024-01-12
Bugfix arcticdb-man/62: Speedup reading wide dataframes that have no empty columns

alexowens90:
Closes https://github.com/man-group/arcticdb-man/issues/62

The modified for loop created a performance regression compared to the internal Man Group client when reading wide, shallow dataframes.

Also applied to the `4.0.x` branch in #1223.

Also adds a simple ASV benchmark to prevent similar performance regressions in the future.


jjerphan (approved):
LGTM modulo a :green_circle: CI.

Issue #1056, opened 2023-11-13
`compact_incomplete` fails if one of the dataframes is of empty-type


alexowens90:
v1 API, behaviour almost certainly the same with `finalize_staged_data` in the v2 API. Minimal repro:
```
lib.write(sym, pd.DataFrame({"col": ["hello"]}, index=[pd.Timestamp("2000-01-01")]), parallel=True)
lib.write(sym, pd.DataFrame({"col": [None]}, index=[pd.Timestamp("2000-01-02")]), parallel=True)
lib.compact_incomplete(sym, False, False)
```
gives `InternalException: E_ASSERTION_FAILURE Allocate data called with zero size`

PR #1227, opened 2024-01-12
Bugfix: Empty type

### LINKED ISSUE #1107: Appending empty type to non-empty column and vice versa is not working =
### Describe the bug

When trying to append an empty segment to a column that already has data (or trying to append data to an empty segment) it fails due to type mismatch. Two things need to be done:

1. Make it possible to append empty segments to non-empty columns (and vice versa)
2. Perform filling of data. If appending empty column to column of int values fill the "empty" ones with zero, if appending to floats fill with NaN if appending to strings fill with None.

### Steps/Code to Reproduce

```python
from arcticdb import Arctic
from pandas.testing import assert_frame_equal

ac = Arctic("lmdb://test")
df_empty = pd.DataFrame({"col1": [None, None]})
lib.write("test_can_append_to_empty_column", df_empty)
df_non_empty = pd.DataFrame({"col1": [1, 2, 3]})
lib .append("test_can_append_to_empty_column", df_non_empty)
expected_result = pd.concat([df_empty, df_non_empty], ignore_index=True)
assert_frame_equal(lib .read("test_can_append_to_empty_column").data, expected_result)
```

### Expected Results

No error is thrown.

{"col1": [0,0,1,2,3]}

All

### Backend storage used

_No response_

### Additional Context

_No response_

vasil-pashov:
#### Reference Issues/PRs
Closes #1107

Fixes how empty type interacts with other types. In general we should be able to:
* Append other types to columns which were initially empty
* Append empty columns to columns of any other type
* Update values with empty type preserving the type of the column

### Changes:
* Each type handler now has a function to report the byte size of its elements.
* Each type handler now has a function to default initialize some memory
* Empty type handler now backfills the "empty" elements.
  * integer types -> 0 (Not perfect but in future we're planning to add default value argument)
  * float types -> NaN
  * string types -> None
  * bool -> False
  * Nullable boolean -> None
  * Date -> NaT
* The function which does default initialization up to now was used only in dynamic schema. Now the empty handler calls it as well to do the backfill. The function first checks the PoD types and if any of them matches uses the basic default initialization, otherwise it checks if there is a type handler and if so uses its default_initialize functionality
* Refactor how updating works. `Column::truncate` is used instead of copying segment rows one by one. This should improve the performance of update.
* Add a new python fixture `lmdb_version_store_static_and_dynamic` to cover all combinations {V1, V2} ecoding x {Static, Dynamic} schema
* Empty typed columns are now reported as dense columns. They don't have a sparse map and both physical and logical rows are left uninitialized (value `-1`)

**DISCUSS**
- [x] Should we add an option to support updating non-empty stuff with empty (Conclusion reached in a slack thread: yes, we should allow to update with None as long as the type of the output is the same as the type of the column.)
- [x] What should be the output for the following example (a column of none vs a column of 0) (Conclusion reached in a slack thread. The result should be [0,0], i.e. the output should have the same type as the type of the column.)
```python
lib.write("sym", pd.DataFrame({"col": [1,2,3]}))
lib.append("sym", pd.DataFrame({"col": [None, None]}))
lib.read("sym", row_range=[3:5]).data
```
- [ ] Do we need hypothesis testing random appends of empty and non-empty stuff to the same column?
**Dev TODO**
- [x] Verify the following throws
```python
lib.write("sym", pd.DataFrame({"col": [None, None]}))
lib.append("sym", pd.DataFrame({"col": [1, 2, 3]}))
lib.append("sym", pd.DataFrame({"col": ["some", "string"]}))
```
- [x] Appending to empty for dynamic schema
- [x] Fix appending empty to other types with static schema
- [x] Fix appending empty to other types with dynamic schema
- [x] Create a single function to handle backfilling of data and use it both in the empty handler and in reduce_and_fix
- [x] Change the name of PYBOOL type
- [x] Fix update e.g.
```python
lmdb_version_store_v2.write('test', pd.DataFrame([None, None], index=pd.date_range(periods=2, end=dt.now(), freq='1T')))
lmdb_version_store_v2.update('test', pd.DataFrame([None, None], index=pd.date_range(periods=2, end=dt.now(), freq='1T')))
```
- [ ] Add tests for starting with empty column list e.g. `pd.DataFrame({"col": []})`. Potentially mark it as xfail and fix with a later PR.
- [ ] Add tests for update when the update range is not entirely contained in the dataframe range index

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -53,33 +53,61 @@ namespace arcticdb {
...
         ARCTICDB_SAMPLE(HandleEmpty, 0)
         util::check(dest != nullptr, "Got null destination pointer");
-        const size_t num_rows = dest_bytes / get_type_size(DataType::EMPTYVAL);
-        static_assert(get_type_size(DataType::EMPTYVAL) == sizeof(PyObject *));
-        auto ptr_dest = reinterpret_cast<const PyObject **>(dest);
-        fill_with_none(ptr_dest, num_rows);
+        const size_t num_rows = dest_bytes / destination_type_descriptor.get_type_byte_size();
+        static_assert(get_type_size(DataType::EMPTYVAL) == sizeof(PyObject*));
+
+        // TODO: arcticdb::util::default_initialize does mostly the same thing
```
willdealtry, line 75: You could check if it's a nullable object type and do the Python-specific stuff, otherwise forward to default_initialize() to avoid the repetition.

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -5,40 +5,511 @@
...
+        )
+
+    def test_bool(self, lmdb_version_store_static_and_dynamic, boolean_dtype):
+        df_non_empty = pd.DataFrame({"col1": np.array([True, False, True], dtype=boolean_dtype)})
+        lmdb_version_store_static_and_dynamic.append("sym", df_non_empty)
+        expected_result = pd.DataFrame({"col1": np.array([None, None, True, False, True], dtype=boolean_dtype)})
+        assert_frame_equal(lmdb_version_store_static_and_dynamic.read("sym").data, expected_result)
+        assert_frame_equal(
+            lmdb_version_store_static_and_dynamic.read("sym", row_range=[0,2]).data,
+            pd.DataFrame({"col1": np.array([None, None], dtype=boolean_dtype)})
```
alexowens90, line 78: This is inconsistent with the int and float behaviours for non-nullable bools?

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -5,40 +5,511 @@
...
+        )
+        assert_frame_equal(
+            lmdb_version_store_static_and_dynamic.read("sym", row_range=[2,5]).data,
+            df_non_empty
+        )
+
+    def test_bool(self, lmdb_version_store_static_and_dynamic, boolean_dtype):
+        df_non_empty = pd.DataFrame({"col1": np.array([True, False, True], dtype=boolean_dtype)})
+        lmdb_version_store_static_and_dynamic.append("sym", df_non_empty)
+        expected_result = pd.DataFrame({"col1": np.array([None, None, True, False, True], dtype=boolean_dtype)})
```
alexowens90, line 74: This must be incorrect for non-nullable bools?

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
alexowens90, line 1: This is really nice and clear

cpp/arcticdb/column_store/column.cpp
```diff
@@ -513,65 +512,89 @@ std::vector<std::shared_ptr<Column>> Column::split(const std::shared_ptr<Column>
...
-        }
-        do {
-            set_input_bit = input_sparse_map.get_next(set_input_bit);
-            if (set_input_bit != 0 && set_input_bit < end_row) {
-                inserter = set_input_bit - start_row;
-            }
-        } while (set_input_bit != 0);
-        inserter.flush();
+        util::BitMagic output_sparse_map = is_empty_type(column->type().data_type())
+            ? util::BitMagic{}
```
alexowens90, line 596: Should this be removed based on the discussion this morning?

cpp/arcticdb/column_store/memory_segment_impl.cpp
```diff
@@ -428,24 +431,42 @@ std::shared_ptr<SegmentInMemoryImpl> SegmentInMemoryImpl::truncate(size_t start_
...
         metadata.CopyFrom(*metadata_);
         output->set_metadata(std::move(metadata));
     }
 
     for(const auto&& [idx, column] : folly::enumerate(columns_)) {
-        auto truncated_column = Column::truncate(column, start_row, end_row);
-        output->add_column(descriptor_->field(idx), truncated_column);
+        const DataType column_type = column->type().data_type();
+        const Field& field = descriptor_->field(idx);
+        if (is_sequence_type(column_type) && reconstruct_string_pool) {
```
alexowens90, line 447: I think this logic would be clearer if `Column::truncate` was called before the if statement. Then you can just iterate through `truncated_column` taking a reference to the offset in the column. If it fails the [is_a_string](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/util/offset_string.hpp#L54) check, leave it as-is. Otherwise, update the value to `output->string_pool().get(string_pool_.get_view(value)).offset()`

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
vasil-pashov: Pandas interprets `None` as `False` if you set the dtype to be boolean. I'm happy to add a separate test if you think it will be easier to understand.

cpp/arcticdb/column_store/column.cpp
vasil-pashov: yes

cpp/arcticdb/pipeline/write_frame.hpp
```diff
@@ -63,15 +62,19 @@ folly::Future<entity::AtomKey> append_frame(
...
         VersionId version_id,
-        bool before,
+        AffectedSegmentPart affected_part,
         const std::shared_ptr<Store>& store);
 
-std::vector<SliceAndKey> flatten_and_fix_rows(
-        const std::vector<std::vector<SliceAndKey>>& groups,
-        size_t& global_count);
+// TODO: Use std::span when C++20 is enabled
+std::vector<SliceAndKey> flatten_and_fix_rows(boost::span<const std::vector<SliceAndKey>> groups, size_t& global_count);
```
alexowens90, line 78: Why `boost::span` and not just `array`?

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
alexowens90: Ah OK, its sufficient to add comments to the tests saying this then

cpp/arcticdb/pipeline/write_frame.hpp
vasil-pashov: At the time of writing this I wasn't sure if we have variable length of elements. Looking at it now it might be the case that we always have 5 elements in it. Will update it.

alexowens90:
Can you add a test for #1056 as well? `xfail` it if these changes don't fix it, but good to understand the current behaviour

cpp/arcticdb/column_store/memory_segment_impl.cpp
```diff
@@ -428,24 +431,41 @@ std::shared_ptr<SegmentInMemoryImpl> SegmentInMemoryImpl::truncate(size_t start_
...
-        auto truncated_column = Column::truncate(column, start_row, end_row);
-        output->add_column(descriptor_->field(idx), truncated_column);
+    for (const auto&& [idx, column] : folly::enumerate(columns_)) {
+        const DataType column_type = column->type().data_type();
+        const Field& field = descriptor_->field(idx);
+        std::shared_ptr<Column> truncated_column = Column::truncate(column, start_row, end_row);
+
+        if (is_sequence_type(column_type) && reconstruct_string_pool) {
+            ChunkedBuffer& truncated_buffer = truncated_column->data().buffer();
+            for (position_t row = 0; row < truncated_column->row_count(); ++row) {
```
alexowens90, line 451: `get_offset_string_at` will be doing a random-access into the chunked buffer, which adds a bunch of unnecessary branching when we are just iterating forwards like this. You can use the new `Column::for_each` [method](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/column_store/column.hpp#L661-L668) to make this both more efficient and more readable.

alexowens90 (changes requested):

alexowens90:
> * Please run the new empty missing tests before merging this code
> * Please test that defrag works on a column containing mixed int + missing and bool + missing (ie that is does not compress the missing data and values into the same segment, which I understand would cause an issue)

I hadn't considered the impact on defrag before. We should merge this and open a new ticket to deal with it though, it will be a mostly orthogonal change.

vasil-pashov:
Related epic tracking all remaining things TODO https://github.com/man-group/ArcticDB/issues/1339

alexowens90 (changes requested):

cpp/arcticdb/column_store/memory_segment_impl.cpp
```diff
@@ -428,24 +431,40 @@ std::shared_ptr<SegmentInMemoryImpl> SegmentInMemoryImpl::truncate(size_t start_
...
+        const TypeDescriptor column_type = column->type();
+        const Field& field = descriptor_->field(idx);
+        std::shared_ptr<Column> truncated_column = Column::truncate(column, start_row, end_row);
+        column_type.visit_tag([&](auto tag) {
+            if constexpr (is_sequence_type(decltype(tag)::data_type())) {
+                Column::transform<decltype(tag), decltype(tag)>(
+                    *truncated_column,
+                    *truncated_column,
+                    [this, &output](auto string_pool_offset) {
+                        const std::string_view string = get_string_from_pool(string_pool_offset, *string_pool_);
```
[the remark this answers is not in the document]
vasil-pashov: True

PR #1228, opened 2024-01-12
Switch from `arcticdb` to `adb` in the demo Notebooks

muhammadhamzasajjad:
In all demo notebooks switched to using `import arcticdb as adb` from ` arcticdb`.


DrNickClarke (approved):

joe-iddon (approved):
Looks good to me. The other notebooks already use this format.

PR #1236, opened 2024-01-15
Switch from `arcticdb` to `adb` in python docstrings

muhammadhamzasajjad:
This was missed in the previous [PR #1219](https://github.com/man-group/ArcticDB/pull/1219).


DrNickClarke (approved):
LGTM

PR #1255, opened 2024-01-19
Bugfix 1248 and 1249: compact_incomplete reject incomplete segments that overlap each other, or existing segments in the case of append

### LINKED ISSUE #1248: `compact_inomplete` should reject incomplete segments with overlapping date ranges

### LINKED ISSUE #1249: `compact_incomplete` with `append=True` should throw if the earliest timestamp in the index of an incomplete segment is earlier than the latest timestamp in the index of the existing symbol

alexowens90:
Closes #1248 
Closes #1249

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -849,7 +849,7 @@ void check_incompletes_index_ranges_dont_overlap(const std::shared_ptr<PipelineC
             if (next_it != unique_timestamp_ranges.end()) {
                 sorting::check<ErrorCode::E_UNSORTED_DATA>(
                         next_it->first >= it->second,
-                        "Cannot append staged segments to existing data as incomplete segment index values overlap one another: ({}, {}) ∩ ({}, {}) ≠ ∅",
+                        "Cannot append staged segments to existing data as incomplete segment index values overlap one another: ({}, {}) intersects ({}, {})",
                         it->first, it->second - 1, next_it->first, next_it->second - 1);
             }
         }
@@ -1209,7 +1209,6 @@ VersionedItem sort_merge_impl(
     auto num_versioned_rows = pipeline_context->total_rows_;
 
     read_incompletes_to_pipeline(store, pipeline_context, read_query, ReadOptions{}, convert_int_to_float, via_iteration, sparsify);
-    check_incompletes_index_ranges_dont_overlap(pipeline_context);
 
     std::vector<entity::VariantKey> delete_keys;
     for(auto sk = pipeline_context->incompletes_begin(); sk != pipeline_context->end(); ++sk) {
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -819,6 +819,43 @@ void read_incompletes_to_pipeline(
     pipeline_context->total_rows_ = pipeline_context->calc_rows();
 }
 
+void check_incompletes_index_ranges_dont_overlap(const std::shared_ptr<PipelineContext>& pipeline_context) {
+    // Does nothing if the symbol is not timestamp-indexed
+    // Checks both that the index ranges of incomplete segments do not overlap with one another, and that the earliest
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -1172,6 +1209,7 @@ VersionedItem sort_merge_impl(
     auto num_versioned_rows = pipeline_context->total_rows_;
 
     read_incompletes_to_pipeline(store, pipeline_context, read_query, ReadOptions{}, convert_int_to_float, via_iteration, sparsify);
+    check_incompletes_index_ranges_dont_overlap(pipeline_context);
```
joe-iddon, line 1212: The docs for the top-level `sort_merge` API state: "sort_merge will sort and merge incomplete segments. The segments do not have to be ordered - incomplete segments can contain interleaved time periods but the final result will be fully ordered".

Does it make sense then to check they don't overlap therefore here if the method doesn't require that they are non-overlapping?

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -335,19 +334,18 @@ def test_parallel_overlapping_incomplete_segments(lmdb_version_store, append):
...
-@pytest.mark.xfail(reason="See https://github.com/man-group/ArcticDB/issues/1249")
 def test_parallel_append_overlapping_with_existing(lmdb_version_store):
     lib = lmdb_version_store
     sym = "test_parallel_append_overlapping_with_existing"
     df_0 = pd.DataFrame({"col": [1, 2]}, index=[pd.Timestamp("2024-01-01"), pd.Timestamp("2024-01-02")])
     lib.write(sym, df_0)
     df_1 = pd.DataFrame({"col": [3, 4]}, index=[pd.Timestamp("2024-01-01T12"), pd.Timestamp("2024-01-03")])
     lib.append(sym, df_1, incomplete=True)
-    with pytest.raises(Exception):
+    with pytest.raises(SortingException):
```
joe-iddon, line 348: Maybe we could add an assert to check the values included in the error message string are not out-by-one or anything? You can `with pytest.raises(...) as exception_info`.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -819,6 +819,43 @@ void read_incompletes_to_pipeline(
...
+                    it->slice_and_key().key().start_time(),
+                    *last_existing_index_value);
+            unique_timestamp_ranges.insert({it->slice_and_key().key().start_time(), it->slice_and_key().key().end_time()});
+        }
+        for (auto it = unique_timestamp_ranges.begin(); it != unique_timestamp_ranges.end(); it++) {
+            auto next_it = std::next(it);
+            if (next_it != unique_timestamp_ranges.end()) {
+                sorting::check<ErrorCode::E_UNSORTED_DATA>(
+                        next_it->first >= it->second,
+                        "Cannot append staged segments to existing data as incomplete segment index values overlap one another: ({}, {}) ∩ ({}, {}) ≠ ∅",
```
joe-iddon, line 852: Maybe change the notation to be a bit more friendly?

joe-iddon (approved):

cpp/arcticdb/version/version_core.cpp
alexowens90: I suppose enforcing all users to have basic set-theory to understand our error messages might be exclusionary

python/tests/unit/arcticdb/version_store/test_parallel.py
alexowens90: I've verified this manually, and any future PR changing these values by 1 should have the reviewers (correctly) asking why.
From a "the tests are the documentation" point of view, I dislike tests that rely on the content of strings, either in logs or in exception messages, as it restricts our ability to modify these strings in the future. The behaviour we are guaranteeing is that this type of error will always raise a `SortingException`, the exception message is an implementation detail that we are free to change, and shouldn't be relied upon by users.

PR #1261, opened 2024-01-22
Bugfix 1035

### LINKED ISSUE #1035: Using QueryBuilder on grouping column with floats fails on Linux Conda

alexowens90:
Closes #1035 

There is still an issue with non-deterministic hashing in the linux conda build, #1268 opened to investigate this further.
In addition:

- We were previously dynamically linking `xxhash` in the conda builds, despite using macros designed for static linking. Now it is linked in statically.
- We were not pinning the `xxhash` version in the conda builds, and so were on the latest version `0.8.2`. This is now pinned to `0.8.2`. This pin will presumably need to be replicated in the feedstock.
- We were using `xxhash.c` and `xxhash.h` files copied from version `0.6.5` of the project. We now pull in this dependency using `vcpkg`, pinned to version `0.8.2`.
- https://fastcompression.blogspot.com/2018/03/xxhash-for-small-keys-impressive-power.html claims that for small keys, if the size argument to the `xxhash` hashing functions is a compile-time constant, then performance improvements can be observed, so an overload of `hash` was added to do this. This is the method used for hashing aggregations, where the keys are at most 8 bytes long for non-string columns.


willdealtry (approved):

cpp/arcticdb/util/hash.hpp
```diff
@@ -10,31 +10,42 @@
...
 using HashedValue = XXH64_hash_t;
 constexpr std::size_t DEFAULT_SEED = 0x42;
 
 template<class T, std::size_t seed = DEFAULT_SEED>
-HashedValue hash(T *d, std::size_t count = 1) {
+HashedValue hash(T *d, std::size_t count) {
     return XXH64(reinterpret_cast<const void *>(d), count * sizeof(T), seed);
 }
 
+// size argument to XXH64 being compile-time constant improves performance
```
jjerphan, line 43:
```suggestion
// This is the specialization of hash for `count = 1` size having the
// `size` argument to `XXH64` being compile-time constant improves performance.
```

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -22,10 +22,7 @@
 from hypothesis import assume, given, settings
 from hypothesis.extra.pandas import column, data_frames, range_indexes
 
-from tests.util.mark import xfail_on_linux_conda
```
jjerphan, line 25: `xfail_on_linux_conda` can now be removed since it is not used anymore.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -524,12 +528,13 @@ endif()
 # on their origin (i.e. vendored source or packages on conda-forge).
 if(${ARCTICDB_USING_CONDA})
 
-    # We must use dynamic libraries.
+    # We must use dynamic libraries, except for xxHash, which must be statically linked.
+    # In non-conda builds, we inline the functions so it does not need linking at all.
```
jjerphan, line 532:
```suggestion
    # For all builds but linux conda's, we inline the functions so that
    # xxhash does not need linking at all.
```

cpp/arcticdb/CMakeLists.txt
```diff
@@ -90,6 +90,10 @@ else()
     list(REMOVE_ITEM NEW_FOLLY_DEPS "/usr/local/opt/openssl/include" )
     set_target_properties(Folly::folly_deps PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${NEW_FOLLY_DEPS}")
 
+    add_library(xxHash STATIC IMPORTED)
+    set_target_properties(xxHash PROPERTIES IMPORTED_LOCATION ${xxHash_LIBRARY})
+    set_target_properties(xxHash PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${xxHash_INCLUDE_DIR})
```
jjerphan, line 95: Shouldn't this be done for conda packages and for wheels (i.e. independently from `ARCTICDB_USING_CONDA`)?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -524,12 +528,13 @@ endif()
...
-    # We must use dynamic libraries.
+    # We must use dynamic libraries, except for xxHash, which must be statically linked.
+    # In non-conda builds, we inline the functions so it does not need linking at all.
     list (APPEND arcticdb_core_libraries
-        xxHash::xxHash
         pcre
         msgpack-c
         ${LMDB_LIBRARIES}
+        xxHash
     )
```
jjerphan, line 538: Was this change needed because of vcpkg's distribution?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -524,12 +528,13 @@ endif()
 # on their origin (i.e. vendored source or packages on conda-forge).
 if(${ARCTICDB_USING_CONDA})
 
-    # We must use dynamic libraries.
+    # We must use dynamic libraries, except for xxHash, which must be statically linked.
```
jjerphan, line 531:
```suggestion
    # We must use dynamic libraries, except for xxHash, which must be statically linked
    # since `XXH64_state_t` which is used ArcticDB is only available for static linkage.
```

cpp/arcticdb/CMakeLists.txt
alexowens90: `xxHash` isn't a target in the vcpkg build, so this won't work

cpp/arcticdb/CMakeLists.txt
alexowens90: Rephrased

cpp/arcticdb/CMakeLists.txt
alexowens90: No, you can use either, I'd just removed it and then put it back with the alternative name. Reverted now.

cpp/arcticdb/CMakeLists.txt
alexowens90: Rephrased

PR #1269, opened 2024-01-24
Skip workflow jobs when only .md docs are updated

muhammadhamzasajjad:
Previously GitHub workflows were unnecessarily running even when only docs were changed. This PR is to avoid running slow workflow jobs when only docs are changed.

The solution is implemented according to the guidelines on the following page 

https://blog.pantsbuild.org/skipping-github-actions-jobs-without-breaking-branch-protection/

For testing the changes you can try the following commits

b77f245: this is to test that if you change the files that you want to ignore, GitHub workflows skips all the checks. In this test we ignore all `.yml` files as these are the only files changed in this PR. In fact it skips all the checks

53407d6: this is to test that if you change files to ignore and other files too, then GitHub workflows will run all the required tests. e.g. if you change a `.md` and a `.py` file you would expect all the checks to run. For this test the paths to ignore are set to `*/build.yml`. As this PR changes multiple `.yml` files including `build.yml` we can test the expected behaviour.

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -9,7 +9,8 @@ on:
     - cron: '0 0 * * *'
 
   pull_request:
-
+    paths-ignore:
+      - "**/*.md"
 jobs:
   cibw_docker_image:
     uses: ./.github/workflows/cibw_docker_image.yml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,8 @@ concurrency:
   cancel-in-progress: true
 jobs:
   cibw_docker_image:
+    if: ${{ needs.docs_only_check.outputs.docs_only != 'DOCS_ONLY' }}
+    needs: [docs_only_check]
     uses: ./.github/workflows/cibw_docker_image.yml
     permissions: {packages: write}
     with:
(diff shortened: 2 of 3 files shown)


poodlewars (approved):
Thanks for this. It does seem quite complicated though, although I think you've done it in the simplest way that will actually work. I'm wondering if it would be more sensible to just run the build even for docs changes, rather than maintaining this added complexity? What do you think?

muhammadhamzasajjad:
> Thanks for this. It does seem quite complicated though, although I think you've done it in the simplest way that will actually work. I'm wondering if it would be more sensible to just run the build even for docs changes, rather than maintaining this added complexity? What do you think?

Thanks for the review. 
Depends on how likely we are to update the docs. If we often update the docs, passing all the checks has been taking me around 6-7hours. Whereas this change skips the docs and only takes a couple of minutes to merge the changes. @DrNickClarke probably knows how frequently we tend to update the docs. I have seen he has been publishing the docs a few times.

There is another problem that this PR fixes:  `build.yml` file contains `can_merge` which is a required action for merging a branch. But `build.yml` previously skipped any checks for `docs/` (`paths` field in `build.yml` didn't contain `docs/` folder) and because of that `can_merge`  was also skipped. This required force merge as `can_merge` check would never pass for docs. This PR now smartly skips all the actions dependent on `can_merge` and passes `can_merge` action so you don't have to force merge if you just update the docs.

Apart from that there aren't any significant advantages that this update brings.

PR #1271, opened 2024-01-25
Add a way to enable/disable silencing of errors when deleting a library

G-D-Petrov:
When trying to delete a library using delete_library(), if the user doesn't have write access to the lib, deleting the content fails, but this is not raising an exception.

This PR introduces a way to enable/disable this silencing of errors when deleting a library.


alexowens90 (approved):

poodlewars:
What's the use case for `continue_on_error=True`? `continue_on_error=False` seems like it would be what we always want?

G-D-Petrov:
> What's the use case for `continue_on_error=True`? `continue_on_error=False` seems like it would be what we always want?

`continue_on_error=True` preserves the current behavior to silence the errors.
We want to add a way to enable them, not to enforce them by default, as this might break existing usage.

PR #1276, opened 2024-01-29
Bugfix 902: Cannot filter on nans and nones

### LINKED ISSUE #902: `QueryBuilder` does not support checking string columns against `None` or `NaN` values
We support storing the values `None` and `NaN` in string columns, but in/equality and set membership of these values cannot be filtered on using the `QueryBuilder`.
Something like Pandas `isna/notna` would be useful

alexowens90:
#### Reference Issues/PRs
Closes #902 

Implement `isna/isnull/notna/notnull` operators in the `QueryBuilder`. Notable decisions:

- `isna` is equivalent to `isnull` and `notna` is equivalent to `notnull`. In particular, that means that in string columns, which can contain both `NaN` and `None` values, no distinction is made between `NaN` and `None`. This is the same behaviour as Pandas.
- When applied to columns with integer dtypes, `isna/isnull` will always return empty dataframes, and `notna/notnull` will always return the same result as a plain `lib.read` call with no `QueryBuilder` provided.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/entity/types.hpp
+++ b/cpp/arcticdb/entity/types.hpp
@@ -397,8 +397,6 @@ struct DataTypeTag<DataType::__DT__> : public DataTypeTagBase { \
 }; \
 using TAG_##__DT__ = DataTypeTag<DataType::__DT__>;
 
-using timestamp = int64_t;
-
 
 DATA_TYPE_TAG(UINT8, std::uint8_t)
 DATA_TYPE_TAG(UINT16, std::uint16_t)
--- a/cpp/arcticdb/processing/operation_dispatch_unary.hpp
+++ b/cpp/arcticdb/processing/operation_dispatch_unary.hpp
@@ -108,39 +108,37 @@ VariantData visit_unary_operator(const VariantData& left, Func&& func) {
 
 template <typename Func>
 VariantData unary_comparator(const Column& col, Func&& func) {
-    if (is_empty_type(col.type().data_type())) {
-        return EmptyResult{};
+    if (is_empty_type(col.type().data_type()) || is_integer_type(col.type().data_type())) {
+        if constexpr (std::is_same_v<Func, IsNullOperator &&>) {
+            return is_empty_type(col.type().data_type()) ? VariantData(FullResult{}) : VariantData(EmptyResult{});
+        } else if constexpr (std::is_same_v<Func, NotNullOperator &&>) {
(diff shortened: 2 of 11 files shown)


cpp/arcticdb/util/constants.hpp
```diff
@@ -11,6 +11,14 @@ namespace arcticdb {
 
 using timestamp = int64_t;
 
+using position_t = int64_t;
+
+constexpr auto string_none = std::numeric_limits<position_t>::max();
+
+constexpr auto string_nan = std::numeric_limits<position_t>::max() - 1;
+
+constexpr auto NaT = std::numeric_limits<timestamp>::min();
```
alexowens90, line 20: Note that the previous definition of `NaT` used `auto`, resulting in a `uint64_t`, whereas `timestamp` is an alias for `int64_t`. It was previously only used in `std::fill`, and so the behaviour was correct, but now we want to perform comparisons to `timestamp` values, so defining as a `timestamp` avoids unnecessary static casts in the comparison code while maintaining the correct behaviour in `std::fill`.

cpp/arcticdb/processing/operation_dispatch_unary.hpp
```diff
@@ -65,23 +64,23 @@ VariantData unary_operator(const Column& col, Func&& func) {
...
-                *ptr++ = func.apply(block[idx]);
-
+        auto col_data = col.data();
+        while(auto block = col_data.next<TDT>()) {
+            const auto row_count = block->row_count();
+            const auto nbytes = sizeof(TargetType) * row_count;
+            auto in_ptr = reinterpret_cast<const RawType*>(block->data());
+            auto out_ptr = reinterpret_cast<TargetType*>(output->allocate_data(nbytes));
+            for(auto idx = 0u; idx < row_count; ++idx) {
+                *out_ptr++ = func.apply(*in_ptr++);
```
alexowens90, line 82: This changes avoids using the slower random-access `block[idx]`

cpp/arcticdb/entity/types.hpp
```diff
@@ -397,8 +397,6 @@ struct DataTypeTag<DataType::__DT__> : public DataTypeTagBase { \
 }; \
 using TAG_##__DT__ = DataTypeTag<DataType::__DT__>;
 
-using timestamp = int64_t;
```
alexowens90, line 400: This is duplicated from above, and no not needed

vasil-pashov (approved):

PR #1285, opened 2024-01-31
#447 LMDB Exceptions Normalization

muhammadhamzasajjad:
Implementation of lmdb storage specific exceptions. This PR builds on #584.

This PR also implements a framework for testing storage exceptions. `GenericStorageTest` class allows testing of common exceptions which are triggered among all the storages. Some examples of these exceptions are `KeyNotFoundException` if you read an in-existent key, `DuplicateKeyException` if you attempt to overwrite a key that already exists.

`lmdb::map_full_error` is an example of exception specific to LMDB which is triggered when the remaining map size is not enough to write new keys/values.

### DIFF
--- a/cpp/arcticdb/storage/storage.hpp
+++ b/cpp/arcticdb/storage/storage.hpp
@@ -61,10 +61,12 @@ private:
     std::optional<VariantId> key_;
 };
 
-class KeyNotFoundException : public ArcticSpecificException<ErrorCode::E_DUPLICATE_KEY> {
+class KeyNotFoundException : public ArcticSpecificException<ErrorCode::E_KEY_NOT_FOUND> {
+    using format_string_t = fmt::format_string<Composite<VariantKey>>;
 public:
-    explicit KeyNotFoundException(Composite<VariantKey>&& keys) :
-        ArcticSpecificException<ErrorCode::E_DUPLICATE_KEY>(fmt::format("{}", keys)),
+    explicit KeyNotFoundException(Composite<VariantKey>&& keys,
+                                  format_string_t format = "Not found: {}") :
+        ArcticSpecificException<ErrorCode::E_KEY_NOT_FOUND>(fmt::format(format, keys)),
         keys_(std::make_shared<Composite<VariantKey>>(std::move(keys))) {
     }
 
--- a/cpp/arcticdb/storage/test/test_storage_exceptions.cpp
+++ b/cpp/arcticdb/storage/test/test_storage_exceptions.cpp
@@ -1,0 +1,164 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#include <gtest/gtest.h>
+
+#include <arcticdb/codec/codec.hpp>
+#include <arcticdb/storage/storage.hpp>
+#include <arcticdb/storage/lmdb/lmdb_storage.hpp>
+#include <arcticdb/util/buffer.hpp>
+
+#include <filesystem>
+#include <memory>
+
+namespace ac = arcticdb;
+namespace acs = arcticdb::storage;
+
+inline const fs::path TEST_DATABASES_PATH = "./test_databases";
+
+class StorageFactory {
+public:
+    virtual ~StorageFactory() = default;
+    virtual std::unique_ptr<acs::Storage> Create() = 0;
+};
+
+class LMDBStorageFactory : public StorageFactory {
+private:
+    uint64_t map_size;
+
+public:
+    LMDBStorageFactory() : map_size(128ULL * (1ULL << 20)) { }
+
+    explicit LMDBStorageFactory(uint64_t map_size) : map_size(map_size) { }
+
+    std::unique_ptr<acs::Storage> Create() override {
+        arcticdb::proto::lmdb_storage::Config cfg;
+
+        fs::path db_name = "test_lmdb";
+        cfg.set_path((TEST_DATABASES_PATH / db_name).generic_string());
+        cfg.set_map_size(map_size);
+        cfg.set_recreate_if_exists(true);
+        cfg.set_recreate_if_exists(true);
+
+        acs::LibraryPath library_path{"a", "b"};
+
+        return std::make_unique<acs::lmdb::LmdbStorage>(library_path, acs::OpenMode::WRITE, cfg);
+    }
+};
+
+// Generic tests that run with all types of storages
+
+class GenericStorageTest : public ::testing::TestWithParam<std::shared_ptr<StorageFactory>> {
+protected:
+    std::unique_ptr<acs::Storage> storage;
+
+    void SetUp() override {
+        if (!fs::exists(TEST_DATABASES_PATH)) {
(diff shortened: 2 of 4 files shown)


cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+namespace ac = arcticdb;
+namespace acs = arcticdb::storage;
+
+inline const fs::path TEST_DATABASES_PATH = "./test_databases";
+
+class StorageFactory {
+public:
+    virtual ~StorageFactory() = default;
+    virtual std::unique_ptr<acs::Storage> Create() = 0;
```
joe-iddon, line 26: Usually lowercase naming is used for methods.

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -0,0 +1,164 @@
...
+    explicit LMDBStorageFactory(uint64_t map_size) : map_size(map_size) { }
+
+    std::unique_ptr<acs::Storage> Create() override {
+        arcticdb::proto::lmdb_storage::Config cfg;
+
+        fs::path db_name = "test_lmdb";
+        cfg.set_path((TEST_DATABASES_PATH / db_name).generic_string());
+        cfg.set_map_size(map_size);
+        cfg.set_recreate_if_exists(true);
+        cfg.set_recreate_if_exists(true);
```
joe-iddon, line 45: duplicate line here

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+INSTANTIATE_TEST_SUITE_P(
+        AllStoragesCommonTests,
+        GenericStorageTest,
+        ::testing::Values(std::make_shared<LMDBStorageFactory>())
+);
+
+// LMDB Storage specific tests
+
+class LMDBStorageTestBase : public ::testing::TestWithParam<std::shared_ptr<LMDBStorageFactory>> {
```
joe-iddon, line 137: This is used in `TEST_F` suites, not `TEST_P`. It should be possible to remove the `::testing::TestWithParam` then.

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    ASSERT_TRUE(storage->key_exists(k));
+
+    acs::KeySegmentPair kv1(k);
+    kv1.segment().set_buffer(std::make_shared<ac::Buffer>());
+
+
+    ASSERT_THROW({
+        storage->write(std::move(kv1));
+    },  acs::DuplicateKeyException);
```
joe-iddon, line 91: Another nitpick with formatting, but you should aim to be consistent with these patterns. I'd have just one blank line before, and this formatting of the `ASSERT_THROW` with the 3rd line at the same indent, rather than indented a lot as you have done elsewhere is the best way I think.

joe-iddon: Otherwise everything LGTM.

cpp/arcticdb/storage/storage.hpp
```diff
@@ -61,10 +61,12 @@ class NoDataFoundException : public ArcticCategorizedException<ErrorCategory::MI
...
 };
 
-class KeyNotFoundException : public ArcticSpecificException<ErrorCode::E_DUPLICATE_KEY> {
+class KeyNotFoundException : public ArcticSpecificException<ErrorCode::E_KEY_NOT_FOUND> {
+    using format_string_t = fmt::format_string<Composite<VariantKey>>;
 public:
-    explicit KeyNotFoundException(Composite<VariantKey>&& keys) :
-        ArcticSpecificException<ErrorCode::E_DUPLICATE_KEY>(fmt::format("{}", keys)),
+    explicit KeyNotFoundException(Composite<VariantKey>&& keys,
+                                  format_string_t format = "Not found: {}") :
```
poodlewars, line 68: I think the API is easier to follow if we just pass the formatted string in here, rather than the `format_string` template

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -0,0 +1,163 @@
...
+    void TearDown() override {
+        if (fs::exists(TEST_DATABASES_PATH)) {
+            fs::remove_all(TEST_DATABASES_PATH);
+        }
+    }
+};
+
+TEST_F(LMDBStorageTestBase, WriteMapFullError) {
+    LMDBStorageFactory factory(32ULL * (1ULL << 10));
+    auto storage = std::unique_ptr<acs::lmdb::LmdbStorage>(dynamic_cast<acs::lmdb::LmdbStorage*>(factory.create().release()));
```
poodlewars, line 152: ?? You can just call `factory.create()` right? What's the point of all this type manipulation?

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+    void TearDown() override {
+        if (fs::exists(TEST_DATABASES_PATH)) {
+            fs::remove_all(TEST_DATABASES_PATH);
+        }
+    }
+};
+
+TEST_F(LMDBStorageTestBase, WriteMapFullError) {
```
poodlewars, line 150: Need tests for other kinds of write error

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
poodlewars, line 1: We also need to improve the situation where we raise `lmdb::error` for various internal errors, as raising those is a normalization failure.

That kind of testing is the motivation for Ivo's S3 storage refactor, I think.

We can do that as a follow-up PR, though.

poodlewars (changes requested):

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+inline const fs::path TEST_DATABASES_PATH = "./test_databases";
+
+class StorageFactory {
+public:
+    virtual ~StorageFactory() = default;
+    virtual std::unique_ptr<acs::Storage> create() = 0;
+};
+
+class LMDBStorageFactory : public StorageFactory {
```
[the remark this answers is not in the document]
poodlewars: We agreed to keep this for now as there are some bugs around library cleanup that prevent the composition based class design

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
poodlewars: We agreed to do these later

cpp/arcticdb/storage/storage.hpp
muhammadhamzasajjad: When a `KeyNotFoundException` occurs, shouldn't we always ensure that the key is printed too? If you just dump `std::string` given by the user you won't print the key. Or we could modify it as `std::string message = "Not found"`, then format it as `fmt::format(message + ": {}", keys)`. I am asking this because by default, exception output should contain the key to make sense of the exception. So it might be cleaner to use `string_format`.

poodlewars (approved):
Looks good! Please squash the commits before merging this

PR #1286, opened 2024-01-31
Fix benchmarks

G-D-Petrov:
#### Reference Issues/PRs
Fixes the benchmark tests as there are 2 problems that are being addressed:
- persistent QB benchmarks
    - before, these benchmarks were not writing anything, just reading, as everything was prewritten in a specific prefix
    - this made the benchmarks run faster, but they were fragile to changes in the region/bucket, and we had such a change in December which broke them
    - now these benchmarks do the writing during the setup to make sure that the data is there
-  basic operations benchmarks
    - there was a change that led to writing a lot of data, much more than we could handle (>50GB)
    - this was caused by testing read/writes on short, wide dfs against many symbols
    - we have agreed with Alex O, that we don't need to tests this case against many symbols
    - now we write a short, wide dfs to only one symbols and benchmark that
    
 P. S. : this PR also contains the disabling of code cov on the pytest, due to a seg fault there

### DIFF
--- a/python/benchmarks/basic_functions.py
+++ b/python/benchmarks/basic_functions.py
@@ -16,12 +16,12 @@ class BasicFunctions:
     number = 5
     timeout = 6000
 
-    params = ([10_000, 15_000], [500, 1000])
+    params = ([100_000, 150_000], [500, 1000])
     # params = ([1000, 1500], [500, 1000])
     param_names = ["rows", "num_symbols"]
 
     def setup_cache(self):
-        self.ac = Arctic("lmdb://basic_functions?map_size=5GB")
+        self.ac = Arctic("lmdb://basic_functions?map_size=10GB")
         num_rows, num_symbols = BasicFunctions.params
 
         self.dfs = {rows: generate_pseudo_random_dataframe(rows) for rows in num_rows}
@@ -41,7 +41,7 @@ class BasicFunctions:
             self.ac.delete_library(lib)
 
     def setup(self, rows, num_symbols):
-        self.ac = Arctic("lmdb://basic_functions?map_size=5GB")
+        self.ac = Arctic("lmdb://basic_functions?map_size=10GB")
         self.read_reqs = [ReadRequest(f"{sym}_sym") for sym in range(num_symbols)]
 
         self.df = generate_pseudo_random_dataframe(rows)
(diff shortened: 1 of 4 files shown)


poodlewars:
You should rebase commit 3557fcb787eada3f0d1bcf1763dc1480bc0ab113 out shouldn't you?

python/benchmarks/common.py
```diff
@@ -37,7 +37,7 @@ def generate_random_floats_dataframe(num_rows, num_cols):
     return pd.DataFrame(data, columns=columns)
 
 
-def generate_benchmark_df(n, freq="T", end_timestamp="1/1/2023"):
+def generate_benchmark_df(n, freq="min", end_timestamp="1/1/2023"):
```
poodlewars, line 40: This is the same, isn't it?

python/benchmarks/common.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 
 
-def generate_pseudo_random_dataframe(n, freq="S", end_timestamp="1/1/2023"):
+def generate_pseudo_random_dataframe(n, freq="s", end_timestamp="1/1/2023"):
```
poodlewars, line 12: What does changing "S" to "s" do?

python/benchmarks/common.py
G-D-Petrov: Yes, but T is apparently deprecated, and we get warnings from pandas which are quite annoying.

python/benchmarks/common.py
G-D-Petrov: "S" is apparently deprecated, and we should use "s"

python/benchmarks/basic_functions.py
```diff
@@ -15,12 +16,12 @@ class BasicFunctions:
     number = 5
     timeout = 6000
 
-    params = ([100_000, 150_000], [500, 1000])
+    params = ([10_000, 15_000], [500, 1000])
```
alexowens90, line 19: Why this change? IMO these row counts are already very low, and so not testing very representative use cases

G-D-Petrov: I just wanted to see what would be the time difference with the smaller rows.
The benchmarks runs around 20-25 min faster with the smaller values, which isn't that much faster (~140 vs ~160 min total run time).

I have brought the old values back.

alexowens90 (approved):

PR #1291, opened 2024-02-01
Utility to analyze the size of various key types in a library

poodlewars:
The existing `scan_object_sizes` was not working as it was never actually submitting any work. It always gave the result that the library contained zero bytes. I've fixed it up here.

I've also introduced `scan_object_sizes_by_stream` so that users can figure out which symbols are using up space in their library.

I considered a map-reduce style approach in `scan_object_sizes_by_stream` rather than the mutex but decided against it as I was considered about memory use (or the probably needless complexity of batching reduce steps).

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1796,20 +1796,34 @@ std::unordered_map<KeyType, std::pair<size_t, size_t>> LocalVersionedEngine::sca
 }
 
 std::unordered_map<StreamId, std::unordered_map<KeyType, size_t>> LocalVersionedEngine::scan_object_sizes_by_stream() {
+    std::mutex mutex;
     std::unordered_map<StreamId, std::unordered_map<KeyType, size_t>> sizes;
     auto streams = symbol_list().get_symbols(store());
-    std::vector<KeyType> key_types {KeyType::TABLE_DATA, KeyType::TABLE_INDEX, KeyType::VERSION};
 
-    for (const auto& stream : streams) {
-        auto& sizes_for_stream = sizes[stream];
-        for (const auto& key_type : key_types) {
-            auto& size_for_key_type = sizes_for_stream[key_type];
-            iterate_keys_of_type_for_stream(store(), key_type, stream, [&size_for_key_type, &store=store()] (const auto& vk) {
-                // TODO not sync
-                size_for_key_type += store->read_compressed_sync(vk, storage::ReadKeyOpts{}).segment().total_segment_size();
+    foreach_key_type([&store=store(), &sizes, &mutex](KeyType key_type) {
+        std::vector<std::pair<VariantKey, stream::StreamSource::ReadContinuation>> keys;
+
+        store->iterate_type(key_type, [&keys, &mutex, &sizes, key_type](const VariantKey&& k){
+            keys.emplace_back(std::forward<const VariantKey>(k), [key_type, &sizes, &mutex] (auto&& ks) {
+                auto key_seg = std::move(ks);
+                auto variant_key = key_seg.variant_key();
+                auto stream_id = variant_key_id(variant_key);
+                auto size = key_seg.segment().total_segment_size();
+
+                {
+                    std::lock_guard lock{mutex};
+                    sizes[stream_id][key_type] += size;
+                }
+
+                return variant_key;
             });
+
+        });
+
+        if (!keys.empty()) {
+            store->batch_read_compressed(std::move(keys), BatchReadArgs{}).get();
         }
-    }
+    });
 
     return sizes;
 }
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1768,20 +1768,61 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
-                std::forward<const VariantKey>(k),
-                [&pair](auto &&ks) {
-                    auto key_seg = std::move(ks);
-                    ++pair.first;
-                    pair.second += key_seg.segment().total_segment_size();
-                    return key_seg.variant_key();
-                });
+        std::vector<std::pair<VariantKey, stream::StreamSource::ReadContinuation>> keys;
+        store->iterate_type(key_type, [&keys, &sizes, &key_type](const VariantKey&& k) {
+            auto& pair = sizes[key_type];
```
willdealtry, line 1775: Could you assign first and second to named things via a structured binding?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1768,20 +1768,61 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
+            ++pair.first;
+            auto& size_counter = pair.second;
+            keys.emplace_back(std::forward<const VariantKey>(k), [&size_counter] (auto&& ks) {
```
willdealtry, line 1778: I found the collection that had the functors in it being called 'keys' a bit confusing. Took me a while to work out where the work was getting done.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1768,20 +1768,61 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
 }
 
 std::unordered_map<KeyType, std::pair<size_t, size_t>> LocalVersionedEngine::scan_object_sizes() {
-    std::unordered_map<KeyType, std::pair<size_t, size_t>> sizes;
+    std::unordered_map<KeyType, std::pair<size_t, std::atomic<size_t>>> sizes;
```
vasil-pashov, line 1771: I might be lacking some context. What should the values in `std::pair` above represent?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1768,20 +1768,61 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
-                    return key_seg.variant_key();
-                });
+        std::vector<std::pair<VariantKey, stream::StreamSource::ReadContinuation>> keys;
+        store->iterate_type(key_type, [&keys, &sizes, &key_type](const VariantKey&& k) {
+            auto& pair = sizes[key_type];
+            ++pair.first;
+            auto& size_counter = pair.second;
+            keys.emplace_back(std::forward<const VariantKey>(k), [&size_counter] (auto&& ks) {
+                auto key_seg = std::move(ks);
+                size_counter += key_seg.segment().total_segment_size();
```
willdealtry, line 1780: There's an element of 'no good deed goes unpunished', but since we're changing this (and adding a test, which should have been there from the outset - my bad), I believe we're scanning the encoded fields and working out both the uncompressed and the compressed sizes. Do you think it would be possible to record both? I'm very interested in the compression ratios we're getting as I think we may want to move away from block encoders like LZ4

willdealtry (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1768,20 +1768,61 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
+        store->iterate_type(key_type, [&keys, &mutex, &sizes, key_type](const VariantKey&& k){
+            keys.emplace_back(std::forward<const VariantKey>(k), [key_type, &sizes, &mutex] (auto&& ks) {
+                auto key_seg = std::move(ks);
+                auto variant_key = key_seg.variant_key();
+                auto stream_id = variant_key_id(variant_key);
+                auto size = key_seg.segment().total_segment_size();
+
+                {
+                    std::lock_guard lock{mutex};
+                    sizes[stream_id][key_type] += size;
```
vasil-pashov, line 1815: Does it make sense to use atomic here as well? Or maybe switch to a spinlock instead of a mutex.

poodlewars: This case is a bit different to the function above because we also need some synchronization guarantees over the map itself, so I don't think atomic on its own is safe. We could try spinlock, but really I'm not too worried - the locked operation should be lightning fast relative to the IO.

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: This is all recording compressed sizes. Is there a way to get the uncompressed size without actually doing the decompression of the segments in to `SegmentInMemory`?

cpp/arcticdb/version/local_versioned_engine.cpp
vasil-pashov: > need some synchronization guarantees over the map itself

Is this because there can be a race on creating `sizes[stream_id]`?

Issue #1278, opened 2024-01-29
Filtering does not work on sparse columns


alexowens90:
See also https://github.com/man-group/arcticdb-man/issues/82

PR #1301, opened 2024-02-05
Refactor 1278: Column data dense forward iterator

alexowens90:
#### Reference Issues/PRs
Refactor towards #1278 

Add an iterator class for `ColumnData`, and use this to remove the repeated pattern:
```
auto data = col.data()
while (auto block= data.next_block<TDT>()) {
    auto ptr =reinterpret_cast<const RawType *>(block->data());
    for (auto i = 0u; i < block->row_count(); ++i, ++ptr) {
        do_work(*ptr);
    }
}
```
in all of the filtering and projection code. Used this iterator to add static `for_each` and `transform` methods to `Column` with similar syntax to the standard library versions to make applying the same function to all values in a column/two columns simple.

#### Any other comments?

Also:

- Removed the variants of `binary_operator` and `binary_comparator` that have a value as the left operand and a column as the right operand, in favour of templating a single variant on an `arguments_reversed` bool`.
- Added `ScalarTypeInfo` to homogenise access to `TypeDescriptorTag`, "DataType", and `RawType` when visiting on all supported types.
- Updated `VariantData` to hold a `util::BitSet` rather than a shared pointer to a bitset, as shared ownership is not required.

Next steps - Add support for sparse data, and for `enumerate` style function signatures. This will allow the same pattern to be used in the aggregators.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/column_store/column.hpp
+++ b/cpp/arcticdb/column_store/column.hpp
@@ -19,6 +19,7 @@
 #include <arcticdb/util/sparse_utils.hpp>
 
 #include <folly/container/Enumerate.h>
+#include <folly/Function.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/numpy.h>
 
@@ -648,7 +649,8 @@ public:
 
     template<typename input_tdt>
     static void for_each(const Column& input_column,
-                          std::function<void(typename input_tdt::DataTypeTag::raw_type)>&& f) {
+                          folly::Function<void(typename input_tdt::DataTypeTag::raw_type)>&& f) {
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(f.heapAllocatedMemory() == 0, "Heap allocated {} bytes in Column::for_each", f.heapAllocatedMemory());
         auto input_data = input_column.data();
         std::for_each(input_data.cbegin<input_tdt>(), input_data.cend<input_tdt>(), [&f](auto input_value) {
             f(input_value);
@@ -658,17 +660,19 @@ public:
     template<typename input_tdt, typename output_tdt>
     static void transform(const Column& input_column,
                           Column& output_column,
-                          std::function<typename output_tdt::DataTypeTag::raw_type(typename input_tdt::DataTypeTag::raw_type)>&& f) {
+                          folly::Function<typename output_tdt::DataTypeTag::raw_type(typename input_tdt::DataTypeTag::raw_type)>&& f) {
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(f.heapAllocatedMemory() == 0, "Heap allocated {} bytes in Column::transform 1", f.heapAllocatedMemory());
         auto input_data = input_column.data();
         auto output_data = output_column.data();
-        std::transform(input_data.cbegin<input_tdt>(), input_data.cend<input_tdt>(), output_data.begin<output_tdt>(), f);
+        std::transform(input_data.cbegin<input_tdt>(), input_data.cend<input_tdt>(), output_data.begin<output_tdt>(), std::move(f));
     }
 
     template<typename left_input_tdt, typename right_input_tdt, typename output_tdt>
     static void transform(const Column& left_input_column,
                           const Column& right_input_column,
                           Column& output_column,
-                          std::function<typename output_tdt::DataTypeTag::raw_type(typename left_input_tdt::DataTypeTag::raw_type, typename right_input_tdt::DataTypeTag::raw_type)>&& f) {
+                          folly::Function<typename output_tdt::DataTypeTag::raw_type(typename left_input_tdt::DataTypeTag::raw_type, typename right_input_tdt::DataTypeTag::raw_type)>&& f) {
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(f.heapAllocatedMemory() == 0, "Heap allocated {} bytes in Column::transform 2", f.heapAllocatedMemory());
         auto left_input_data = left_input_column.data();
         auto right_input_data = right_input_column.data();
         auto output_data = output_column.data();
@@ -680,7 +684,8 @@ public:
     template<typename input_tdt>
     static void transform(const Column& input_column,
                           util::BitSet& output_bitset,
-                          std::function<bool(typename input_tdt::DataTypeTag::raw_type)>&& f) {
+                          folly::Function<bool(typename input_tdt::DataTypeTag::raw_type)>&& f) {
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(f.heapAllocatedMemory() == 0, "Heap allocated {} bytes in Column::transform 3", f.heapAllocatedMemory());
         auto input_data = input_column.data();
         util::BitSet::bulk_insert_iterator inserter(output_bitset);
         auto pos = 0u;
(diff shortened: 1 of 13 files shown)


cpp/arcticdb/column_store/column.hpp
```diff
@@ -646,6 +646,72 @@ class Column {
         return *res;
     }
 
+    template<typename input_tdt>
+    static void for_each(const Column& input_column,
```
vasil-pashov, line 650: I don't think we need `std::funciton` here. We can add `typename FunctorT&&` template parameter as we don't store the function in memory. `std::function` uses type erasure and can sometimes dynamically allocate memory and have additional indirection.

cpp/arcticdb/column_store/column.hpp
```diff
@@ -646,6 +646,72 @@ class Column {
         return *res;
     }
 
+    template<typename input_tdt>
```
vasil-pashov, line 649: I'm also wondering if Column has begin/end functions can't we use std::for_each instead of having our own version?

cpp/arcticdb/column_store/column.hpp
```diff
@@ -646,6 +646,72 @@ class Column {
...
+            if (f(input_value)) {
+                inserter = pos;
+            }
+            ++pos;
+        });
+        inserter.flush();
+    }
+
+    template<typename left_input_tdt, typename right_input_tdt>
+    static void transform(const Column& left_input_column,
```
vasil-pashov, line 697: I think a comment on this would be helpful as it's a bit different to a regular transform (since it takes more parameters)

cpp/arcticdb/column_store/column_data.hpp
```diff
@@ -133,6 +133,78 @@ struct ColumnData {
...
+        ColumnDataIterator() = delete;
+
+        // Used to construct [c]begin iterators
+        explicit ColumnDataIterator(ColumnData* parent):
+        parent_(parent)
+        {
+            opt_block_ = parent->next<TDT>();
+            if(opt_block_.has_value()) {
+                remaining_values_in_block_ = opt_block_->row_count();
+                if constexpr(constant) {
```
vasil-pashov, line 153: Isn't this redundant. It looks like we have already set the "const type" in the constructor
```c++
std::conditional_t<constant, typename TDT::DataTypeTag::raw_type const, typename TDT::DataTypeTag::raw_type>
```
can't we use `ColumnIterator::type` or whatever boost defined the result of the conditional to be?

cpp/arcticdb/column_store/segment_utils.hpp
```diff
@@ -17,29 +17,22 @@
...
-                const auto row_count = block->row_count();
-                for (auto i = 0u; i < row_count; ++i) {
-                    output.insert(*ptr++);
-                }
-            }
-            return output;
+    details::visit_type(column.type().data_type(), [&](auto col_desc_tag) {
+        using type_info = ScalarTypeInfo<decltype(col_desc_tag)>;
+        if constexpr(is_sequence_type(type_info::data_type)) {
+            Column::for_each<typename type_info::TDT>(column, [&output_set](auto value) {
```
vasil-pashov, line 28: This looks nice 🎉

cpp/arcticdb/column_store/column.hpp
alexowens90: It isn't clear in this PR, but I'm going to add some generalisations for sparse data that will be kept cleaner by hiding this detail inside the `Column` static methods.
In addition, hiding the iterators from the user makes it obvious that operations are happening over the whole column. We don't want users sub-selecting ranges inside columns, as this will not be efficient in general.

cpp/arcticdb/column_store/column.hpp
alexowens90: I like using this format, as it makes it explicit that you cannot pass in arbitrary functions, they must have a very specific format.
As a general rule, `std::function` will avoid the heap if the function it is wrapping is small enough, although there is no way to tell if this is the case.
`folly::Function`, however, has a `heapAllocatedMemory()` method. I'm currently running all the tests with an `internal::check` that the heap allocated memory is always zero. If this passes on all platforms, we can remove the checks and know that the existing uses do no heap allocation. The tricky thing is enforcing this going forwards.

cpp/arcticdb/column_store/column.hpp
```diff
@@ -646,6 +646,72 @@ class Column {
         return *res;
     }
 
+    template<typename input_tdt>
+    static void for_each(const Column& input_column,
+                          std::function<void(typename input_tdt::DataTypeTag::raw_type)>&& f) {
+        auto input_data = input_column.data();
+        std::for_each(input_data.cbegin<input_tdt>(), input_data.cend<input_tdt>(), [&f](auto input_value) {
```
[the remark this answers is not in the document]
alexowens90: This is always working on contiguous buffers of memory representing `n` of the same type. Any non-trivial struct would be stored in a Dremel encoding, so this will still hold at the `Column` level.

vasil-pashov (approved):

cpp/arcticdb/column_store/column_data.hpp
alexowens90: This one is needed, as it flips between `reinterpret_cast` and 'const_cast'.
I've updating the ``RawType definition to:
`using RawType = std::conditional_t<constant, const typename TDT::DataTypeTag::raw_type, typename TDT::DataTypeTag::raw_type>;`
though, which removes the need for a bunch of other `std::conditional_t`s

cpp/arcticdb/column_store/column.hpp
alexowens90: It is equivalent to (3) in https://en.cppreference.com/w/cpp/algorithm/transform

cpp/arcticdb/column_store/column.hpp
vasil-pashov: Concepts will help this case when we enable C++20. You can achieve the same thing with enable_if and company, but it's kind of ugly.

cpp/arcticdb/column_store/column_data.hpp
```diff
@@ -133,6 +133,78 @@ struct ColumnData {
...
+            } else {
+                opt_block_ = parent_->next<TDT>();
+                if(ARCTICDB_LIKELY(opt_block_.has_value())) {
+                    remaining_values_in_block_ = opt_block_->row_count();
+                    if constexpr(constant) {
+                        ptr_ = reinterpret_cast<RawType*>(opt_block_->data());
+                    } else {
+                        ptr_ = const_cast<RawType*>(opt_block_->data());
+                    }
+                }
```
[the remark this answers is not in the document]
alexowens90: Agreed, will factor out

cpp/arcticdb/column_store/column_data.hpp
```diff
@@ -133,6 +133,78 @@ struct ColumnData {
...
+        }
+
+        RawType& dereference() const {
+            return *ptr_;
+        }
+
+        ColumnData* parent_{nullptr};
+        std::optional<TypedBlockData<TDT>> opt_block_{std::nullopt};
+        std::size_t remaining_values_in_block_{0};
+        RawType* ptr_{nullptr};
```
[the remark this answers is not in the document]
alexowens90: For the sparse case, we'll still need a `RawType*` for retrieving the value. I think a `row_idx_` (or similar) member will be needed for the `enumerate `style methods, and that will be a `size_t` for dense columns, and a bit magic enumerator for sparse columns (as you did in the aggregators). The idea will be to have the `if` as high up as possible, so that it is only branched on once, rather than for every value.

PR #1304, opened 2024-02-06
#447 S3 Storage Exceptions Handling

muhammadhamzasajjad:
This is a refactor of #584 for S3 Storage Exceptions. This PR also implements the tests for S3 storage exceptions. Note that in some cases S3 Storage Exceptions are different from other storages. For instance if you write a path that already exists, other storages throw `DuplicateKeyException`. In this case, S3 just over-writes the key without any errors, therefore no Exception is thrown.

The function `handle_s3_error` handles other S3 Storage specific exceptions as follows:
- `E_PERMISSION`: raised when `ACCESS_DENIED`, `INVALID_ACCESS_KEY_ID` or `SIGNATURE_DOES_NOT_MATCH` errors are returned.
- `E_S3_RETRYABLE`: raised for all the other S3 exceptions that are retry-able.
- `E_UNEXPECTED_S3_ERROR`: raised for all the other S3 exceptions that are not retry-able.

These have been implemented following the PR #584.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/storage/s3/mock_s3_client.cpp
```diff
@@ -38,18 +38,19 @@ std::string operation_to_string(S3Operation operation){
 std::string MockS3Client::get_failure_trigger(
         const std::string& s3_object_name,
         S3Operation operation_to_fail,
-        Aws::S3::S3Errors error_to_fail_with) {
-    return fmt::format("{}#Failure_{}_{}", s3_object_name, operation_to_string(operation_to_fail), (int)error_to_fail_with);
+        Aws::S3::S3Errors error_to_fail_with, bool retryable) {
```
joe-iddon, line 41: Minor: Should fix this formatting

cpp/arcticdb/storage/storage.hpp
```diff
@@ -73,6 +73,13 @@ class KeyNotFoundException : public ArcticSpecificException<ErrorCode::E_KEY_NOT
             keys_(std::make_shared<Composite<VariantKey>>(std::move(keys))) {
     }
 
+    explicit KeyNotFoundException(const VariantKey& single_key):
+            KeyNotFoundException(Composite<VariantKey>{VariantKey{single_key}}) {}
```
joe-iddon, line 77: For reference this is a [delegating constructor](https://en.cppreference.com/w/cpp/language/constructor#:~:text=member%20initializer.-,Delegating%20constructor,-If%20the%20name).

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -44,50 +44,47 @@ namespace s3 {
...
 
             log::storage().error(error_message);
-            throw UnexpectedS3ErrorException(error_message);
-        }
-
-        inline bool is_expected_error_type(Aws::S3::S3Errors err) {
-            return err == Aws::S3::S3Errors::NO_SUCH_KEY
-                   || err == Aws::S3::S3Errors::NO_SUCH_BUCKET
-                   || err == Aws::S3::S3Errors::INVALID_ACCESS_KEY_ID
-                   || err == Aws::S3::S3Errors::ACCESS_DENIED
```
joe-iddon, line 83: So previously it would throw "key doesn't exist", but now we explicitly will throw a permissions error. 👍

cpp/arcticdb/storage/s3/mock_s3_client.hpp
```diff
@@ -48,7 +48,7 @@ class MockS3Client : public S3ClientWrapper {
     //
     // The returned name looks like "{s3_object_name}#Failure_{operation_to_fail}_{error_to_fail_with}".
```
joe-iddon, line 49: This comment should be updated to include the `_{retryable}` clause you've added.

cpp/arcticdb/util/error_code.hpp
```diff
@@ -79,6 +79,9 @@ inline std::unordered_map<ErrorCategory, const char*> get_error_category_names()
     ERROR_CODE(5001, E_DUPLICATE_KEY) \
     ERROR_CODE(5002, E_SYMBOL_NOT_FOUND) \
     ERROR_CODE(5003, E_LMDB_MAP_FULL) \
+    ERROR_CODE(5004, E_PERMISSION) \
```
joe-iddon, line 82: Remember to update the documentation for the error codes here: https://docs.arcticdb.io/latest/error_messages/#errors-with-numeric-error-codes

joe-iddon (changes requested):

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -44,50 +44,47 @@ namespace s3 {
         template<class It>
         using Range = folly::Range<It>;
 
-        class UnexpectedS3ErrorException : public std::exception {
-        public:
-            UnexpectedS3ErrorException(const std::string& message):message(message){}
-            const char* what() const noexcept override {
-                return message.c_str();
+        inline void handle_s3_error(const Aws::S3::S3Error& err){
```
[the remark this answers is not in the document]
muhammadhamzasajjad: I agree with the `inline` point. The function has to be in the header file as it is called by the functions declared in this file. Should we make the `detail-inl.hpp` file a source file rather than a header?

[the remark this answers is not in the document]
poodlewars: On second thought let's leave it as is

I think we should do a separate refactor to get rid of this -inl.hpp file completely

But it will make the PR too hard to review if we do that right now

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -44,50 +44,47 @@ namespace s3 {
...
+            std::string error_message;
+            auto type = err.GetErrorType();
+
+            if(type == Aws::S3::S3Errors::NO_SUCH_KEY || type == Aws::S3::S3Errors::RESOURCE_NOT_FOUND || type == Aws::S3::S3Errors::NO_SUCH_BUCKET) {
```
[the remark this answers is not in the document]
muhammadhamzasajjad: Both `RESOURCE_NOT_FOUND` and `NO_SUCH_BUCKET`  need to be conditionally skipped. If we always raise them, several python tests fail.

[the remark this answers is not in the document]
poodlewars: We've agreed to keep RESOURCE_NOT_FOUND (it is raised in some situations where the SDK would ideally raise a key not found error so we have no choice) and we will clean up the NO_SUCH_BUCKET situation subsequently as it is a behaviour change for several storages.

poodlewars (approved):

Issue #1295, opened 2024-02-02
`in` operator on Arctic does not work


poodlewars:
### Describe the bug

If we attempt to check whether a `str` name for a library is contained in an Arctic instance, we blow up with a confusing pybind error.

I consider this a bug since we support collection like subscripting and hence we have implemented a sort of failed attempt to make Arctic behave like a Python collection. The following works:

```python
lib = ac["tst"]
```

This blows up:

```python
"tst" in lib
```

### Steps/Code to Reproduce

```python
In [2]: from arcticdb import Arctic

In [3]: ac = Arctic("lmdb://blah2")

In [4]: "tst" in ac
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 "tst" in ac

File ~/venvs/310/lib/python3.10/site-packages/arcticdb/arctic.py:159, in Arctic.__getitem__(self, name)
    156 if already_open:
    157     return already_open
--> 159 if not self._library_manager.has_library(name):
    160     raise LibraryNotFound(name)
    162 storage_override = self._library_adapter.get_storage_override()

TypeError: has_library(): incompatible function arguments. The following argument types are supported:
    1. (self: arcticdb_ext.storage.LibraryManager, arg0: str) -> bool

Invoked with: <arcticdb_ext.storage.LibraryManager object at 0x7f93afbc3030>, 0
```

### Expected Results

For any Arctic instance `ac` and any string `lib_name`,

```python
lib_name in ac == ac.has_library(lib_name)
```

### OS, Python Version and ArcticDB Version

Tested with arcticdb 1.6.2 and 4.2.1.

PR #1309, opened 2024-02-07
Add __contains__ to Arctic class, to support lib in arctic

DrNickClarke:
TLDR: add ability to test for a library using

`lib in arctic`
and
`lib not in arctic`

A full description of the fix is in this issue

https://github.com/man-group/ArcticDB/issues/1295


poodlewars (approved):

poodlewars:
Nice! Thank you.

I've added a "bug" label to this PR, which helps make sure this appears in the next set of release notes.

PR #1274, opened 2024-01-26
Abstract an S3ClientWrapper out of details-inl.hpp

IvoDD:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #1315, opened 2024-02-08
Abstract AzureClientWrapper out of azure_storage.cpp

muhammadhamzasajjad:
This is a refactor which should not modify behavior.

The refactor abstracts away building the requests and parsing the responses from Azure behind an abstract class `AzureClientWrapper`. The refactor is written similar to how #1274 is for S3.

`RealAzureClient` (derives `AzureClientWrapper`) is implemented by moving the Azure Client logic from `azure_storage.cpp`. `AzureClientWrapper` will then be used to add `MockAzureClient` class which would be used to simulate azure erros.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/azure/azure_real_client.cpp
+++ b/cpp/arcticdb/storage/azure/azure_real_client.cpp
@@ -84,6 +84,11 @@ Segment RealAzureClient::read_blob(
 void RealAzureClient::delete_blobs(
         const std::vector<std::string>& blob_names,
         unsigned int request_timeout) {
+
+    util::check(blob_names.size() <= BATCH_SUBREQUEST_LIMIT,
+                "Azure delete batch size {} exceeds maximum permitted batch size of {}",
+                blob_names.size(),
+                BATCH_SUBREQUEST_LIMIT);
     auto batch = container_client.CreateBatch();
 
     for (auto& blob_name: blob_names) {
--- a/cpp/arcticdb/storage/azure/azure_real_client.cpp
+++ b/cpp/arcticdb/storage/azure/azure_real_client.cpp
@@ -1,0 +1,115 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+
+#include <azure/core/http/curl_transport.hpp>
+#include <azure/core.hpp>
+#include <azure/storage/blobs.hpp>
+
+#include <arcticdb/storage/storage_utils.hpp>
+#include <arcticdb/storage/azure/azure_real_client.hpp>
+#include <arcticdb/storage/azure/azure_client_wrapper.hpp>
+#include <arcticdb/storage/object_store_utils.hpp>
+
+namespace arcticdb::storage {
+using namespace object_store_utils;
+namespace azure {
+using namespace Azure::Storage;
+using namespace Azure::Storage::Blobs;
+
+Azure::Core::Context get_context(unsigned int request_timeout){
+    Azure::Core::Context requestContext; //TODO: Maybe can be static but need to be careful with its shared_ptr and ContextSharedState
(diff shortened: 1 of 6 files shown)


cpp/arcticdb/storage/azure/azure_real_client.cpp
```diff
@@ -0,0 +1,115 @@
...
+    auto batch = container_client.CreateBatch();
+
+    for (auto& blob_name: blob_names) {
+        ARCTICDB_RUNTIME_DEBUG(log::storage(), "Removing azure blob with key {}", blob_names);
+        batch.DeleteBlob(blob_name);
+    }
+
+    ARCTICDB_RUNTIME_DEBUG(log::storage(), "Submitting DeleteBlob batch");
+    ARCTICDB_SUBSAMPLE(AzureStorageDeleteObjects, 0)
+    container_client.SubmitBatch(batch, Azure::Storage::Blobs::SubmitBlobBatchOptions(), get_context(request_timeout));//To align with s3 behaviour, deleting non-exist objects is not an error, so not handling response
```
poodlewars, line 96: Haven't we lost the batching logic here? master has logic to submit deletes in batches of size no greater than `delete_object_limit`? Without the batching, deletes for blobs outside of the batch size get silently dropped (effectively breaking deletion).

cpp/arcticdb/storage/azure/azure_real_client.cpp
```diff
@@ -0,0 +1,115 @@
...
+    std::shared_ptr<Buffer> tmp;
+    auto hdr_size = segment.segment_header_bytes_size();
+    auto [dst, write_size] = segment.try_internal_write(tmp, hdr_size);
+    util::check(arcticdb::Segment::FIXED_HEADER_SIZE + hdr_size + segment.buffer().bytes() <= write_size,
+                "Size disparity, fixed header size {} + variable header size {} + buffer size {}  >= total size {}",
+                arcticdb::Segment::FIXED_HEADER_SIZE,
+                hdr_size,
+                segment.buffer().bytes(),
+                write_size);
+    ARCTICDB_SUBSAMPLE(AzureStorageUploadObject, 0)
```
poodlewars, line 58: Why is the section above here (calculating dst and write size) in the real vs mock client API? I would think we want to execute the segment logic above regardless of whether we are using mock or real azure client?

cpp/arcticdb/storage/azure/azure_real_client.cpp
poodlewars: Ah I see OK, so you've just lifted the batching logic out and this method assumes that `blob_names` is within the batch size, which makes sense :+1: I think it would be worth adding an assertion that `blob_names` is within the max batch size?

cpp/arcticdb/storage/azure/azure_real_client.cpp
muhammadhamzasajjad: Because I need `dst` and `write_size` for `blob_client.UploadFrom` in real client. For the mock client, I don't need `dst` and `write_size` I shall directly save the `segment` in a map. Have a look at Ivo's similar logic #1274.

cpp/arcticdb/storage/azure/azure_real_client.cpp
muhammadhamzasajjad: Done in [b9a6e99](https://github.com/man-group/ArcticDB/pull/1315/commits/b9a6e99b7738e1623c1ed4af80411744a86b0fa6).

poodlewars (approved):
Looks good! Please squash the commits when you merge this

cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -172,49 +141,42 @@ void do_read_impl(Composite<VariantKey> && ks,
...
+            try {
+                azure_client.delete_blobs(to_delete, request_timeout);
             }
-            catch (const Azure::Core::RequestFailedException& e){
+            catch (const Azure::Core::RequestFailedException& e) {
                 util::raise_rte("Failed to create azure segment batch remove request {}: {}",
                                     static_cast<int>(e.StatusCode),
                                     e.ReasonPhrase);
             }
-            batch = container_client.CreateBatch();
```
phoebusm, line 195: This line is madatory for clearing `batch`. The SDK won't do the job for us

PR #1316, opened 2024-02-09
Make benchmarks more stable

### LINKED ISSUE #1311: Address benchmarking issues
The current benchmarking setup has some deficiencies, such as:
1. It doesn't report a failure when one of the benchmarks fails
2. Runs take a very long time
    - also we accumulate a lot of commits that we follow
3. We produce a lot of noise because we push a new commit every night

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1311 

- Makes it so when a benchmark fails, asv will return a non 0 exit code, thus failing the build (this is done by adding the `--strict` flag)
- Explicitly add the code for creating a AWS URL to the benchmark code, so all versions can work with persistent storage
- Reduce the rows of the DFs used in benchmarking QB with AWS from [10_000_000, 100_000_000] to [1_000_000, 10_000_000], as there doesn't seem to be much benefit(performance scales linearly, as expected) but when testing with 100_000_000 rows, each run takes around **17m**, and also there is a large of cost to reading this much data, so frequently.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -246,20 +246,20 @@ jobs:
         if: inputs.run_all_benchmarks == true
         shell: bash -el {0}
         run: |
-          python -m asv run -v --show-stderr HASHFILE:python/hashes_to_benchmark.txt
+          python -m asv run -v --show-stderr --strict HASHFILE:python/hashes_to_benchmark.txt
 
       - name: Benchmark only latest commit
         if: github.event_name != 'pull_request' && inputs.run_all_benchmarks == false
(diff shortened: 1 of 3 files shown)


poodlewars:
Looks good but I don't see the code change for,

```
Makes it so when a benchmark fails, asv will return a non 0 exit code, thus failing the build (this is done by adding the --strict flag)
```

?

alexowens90 (approved):

poodlewars (approved):

G-D-Petrov:
No longer needed

PR #1326, opened 2024-02-13
Test benchmarking improvements

### LINKED ISSUE #1311: Address benchmarking issues

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1311 

TLDR: This PR aims to address some of the major bottlenecks that we have addressed when using ASV, namely:
- reduce the time that it takes to run/regenerate all  benchmarks - currently 17+ hours vs ~2 hours after this PR;
  - this is done by parallelizing the benchmarking of each commit on a separate GH **Free** Runner
- eliminate the need to push the results to git - ASV tracks the results of each benchmark as a json file, but this results in many unneeded commits and potential clashes
  - this is fixed by transforming the jsons to pandas DFs and storing them in an ArcticDB database in AWS S3
  
  
More detailed explanation of the various changes:
1. Removed the [ASV results jsons from git](https://github.com/man-group/ArcticDB/pull/1326/files#diff-5fe96e5ac9998c3baa62242a849c12213c3a1519e2d6b28338e6f4cd8e61b78a), only the [benchmarks.json](https://github.com/man-group/ArcticDB/pull/1326/files#diff-789c26163b19ea0ebd9d1b4f211c2d2e31f166a8c848f1e762f28dbbc75737d8) has been kept as it is needed to generate the HTML and it is a bit hard to generate on-the-fly
2. [build_tooling/transform_asv_results.py](https://github.com/man-group/ArcticDB/pull/1326/files#diff-e6cfb2b0545257e63d5b4d5b5d53af13b51a467c9d7b2ce2069c413955a834fa) file has been added which is responsible for the transformation between ASV jsons and DF and vice verse 
3. [build_tooling/get_commits_for_benchmark.py](https://github.com/man-group/ArcticDB/pull/1326/files#diff-1100dd99a115c31a48b4ea2cc167549ec0d9da872508dae86f665346c87f8f3a) file has been added which is encapsulates the logic for determining which commits should be used for benchmarking, currently supports 2 modes:
  - run_all_benchmars - which select all commits from which we have created new versions (after version 2.0.0), this is mainly for regenerating the benchmarks
  - default - which just select the commit hash for the current HEAD, this is the incremental benchmark
4. [ASV config](https://github.com/man-group/ArcticDB/pull/1326/files#diff-d18ceac2def9772dfdf24d7a83cbe688bf5717e334108eff2aeb581d9992d531) has been changed to use `conda `as env manager instead of `virtualenv `- this is because the free runners are not working well with `virtualenv`
  - I've kept the old config as [asv.conf_old.json](https://github.com/man-group/ArcticDB/pull/1326/files#diff-6052c98ef491566f9d9edea33859de3003f919e5e716c412797da56146401491) because that one is easier to use when testing locally
5. I've made some changes to the benchmarks themselves to make them run a bit faster and more consistent


asv.conf_old.json
alexowens90, line 1: Can this be deleted?

alexowens90 (changes requested):

build_tooling/transform_asv_results.py
```diff
@@ -0,0 +1,160 @@
...
+    results = pd.concat(results_list, ignore_index=True)
+    results["date"] = pd.to_datetime(data["date"], unit="ms")
+    return results
+
+
+def get_results_lib(arcticdb_client_override, arcticdb_library):
+    if arcticdb_client_override is not None:
+        ac = Arctic(arcticdb_client_override)
+    else:
+        factory = real_s3_from_environment_variables(shared_path=True)
```
poodlewars, line 83: Which s3 bucket are we using for this? Would be good to have a dedicated one for benchmarks (in the correct region!) in case we need to tear down the shared one or whatever.

python/benchmarks/basic_functions.py
```diff
@@ -32,16 +32,25 @@ def setup_cache(self):
             lib = self.ac[lib]
             for sym in range(num_symbols[-1]):
                 lib.write(f"{sym}_sym", self.dfs[rows])
+
+        lib = get_prewritten_lib_name(5_000)
+        self.ac.delete_library(lib)
+        self.ac.create_library(lib)
+        lib = self.ac[lib]
```
poodlewars, line 39: (minor) can do `lib = self.ac.create_library(lib)`

build_tooling/transform_asv_results.py
```diff
@@ -0,0 +1,160 @@
...
+
+def save_asv_results(lib, json_path):
+    for file_path in json_path.glob("**/*.json"):
+        if "benchmark" in file_path.name or "machine" in file_path.name:
+            continue
+        full_path = str(file_path)
+        commit_hash = file_path.name.split("-")[0]
+        print(f"Processing {full_path}")
+        df = asv_json_to_df(full_path)
+        lib.write(commit_hash, df)
```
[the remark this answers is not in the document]
G-D-Petrov: yes but we use only one machine type for all tests

.github/workflows/analysis_workflow.yml
```diff
@@ -13,13 +13,123 @@ on:
...
+        with:
+          aws_access_key: "${{ secrets.AWS_S3_ACCESS_KEY }}"
+          aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"
+
+      - name: Install ArcticDB[Testing]
+        shell: bash -el {0}
+        run: |
+          pip install arcticdb[Testing]
+
+      - name: Publish results to ArcticDB database
```
poodlewars, line 123: typo, should be `publish results from ArcticDB...`?

build_tooling/transform_asv_results.py
G-D-Petrov: it is using the one we use for the ci_test - arcticdb-ci-test-bucket-02.
But i agree, it is better to have it in a separate one.

build_tooling/transform_asv_results.py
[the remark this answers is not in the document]
poodlewars: Won't it be fragile if we change the machine that we run on or something?

.github/workflows/benchmark_commits.yml
```diff
@@ -0,0 +1,105 @@
...
+    secrets: inherit
+    with:
+      job_type: start
+
+  benchmark_commit:
+    needs: [start_ec2_runner]
+    if: |
+      always() &&
+      !cancelled()
+    # TODO: Do we need this?
```
poodlewars, line 20: TODO

build_tooling/transform_asv_results.py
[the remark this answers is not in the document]
G-D-Petrov: This is quite rare and it that case it would be better to regenerate the results for all of the versions in any case.

alexowens90 (approved):

build_tooling/transform_asv_results.py
```diff
@@ -0,0 +1,160 @@
...
+    new_df["date"] = (pd.to_datetime(results_df["date"]).astype(int) // 1000000).astype(
+        object
+    )
+    new_df["version"] = results_df["version"].astype(object)
+
+    metadata = {
+        "commit_hash": new_df["commit_hash"].iloc[0],
+        "env_name": new_df["env_name"].iloc[0],
+        "date": new_df["date"].iloc[0],
+        "params": eval(new_df["params"].iloc[0]),
```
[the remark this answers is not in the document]
G-D-Petrov: Added example.

`json.loads` doesn't work because they are not valid jsons list/dicts because they contain mixed types, e.g. lists with ints and float.
That is why we need to use eval to evaluate them to Python lists/dicts

poodlewars (approved):

PR #1329, opened 2024-02-15
S3 local delete failure raising meaningful error

IvoDD:
- Previously on a local delete failure we would have raised a KeyNotFoundException which is quite misleading
- Now we raise a E_UNEXCPETED_S3_ERROR with additional error messages

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -202,7 +202,7 @@ namespace s3 {
             ARCTICDB_SUBSAMPLE(S3StorageDeleteBatch, 0)
             auto fmt_db = [](auto &&k) { return variant_key_type(k); };
             std::vector<std::string> to_delete;
-            std::vector<VariantKey> failed_deletes;
+            std::vector<std::pair<VariantKey, std::string>> failed_deletes;
             static const size_t delete_object_limit =
                     std::min(DELETE_OBJECTS_LIMIT,
                              static_cast<size_t>(ConfigsMap::instance()->get_int("S3Storage.DeleteBatchSize", 1000)));
@@ -222,12 +222,13 @@ namespace s3 {
                                                            to_delete.size(),
                                                            variant_key_view(*k));
                                     for (auto& bad_key: delete_object_result.get_output().failed_deletes) {
-                                        auto bad_key_name = bad_key.substr(key_type_dir.size(),
-                                                                           std::string::npos);
+                                        auto bad_key_name = bad_key.s3_object_name.substr(key_type_dir.size(),
+                                                                                          std::string::npos);
                                         failed_deletes.emplace_back(
                                                 variant_key_from_bytes(
                                                         reinterpret_cast<const uint8_t *>(bad_key_name.data()),
-                                                        bad_key_name.size(), group.key()));
+                                                        bad_key_name.size(), group.key()),
+                                                bad_key.error_message);
                                     }
                                 } else {
                                     auto& error = delete_object_result.get_error();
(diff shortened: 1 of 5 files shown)


cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -202,7 +202,7 @@ namespace s3 {
             ARCTICDB_SUBSAMPLE(S3StorageDeleteBatch, 0)
             auto fmt_db = [](auto &&k) { return variant_key_type(k); };
             std::vector<std::string> to_delete;
-            std::vector<VariantKey> failed_deletes;
+            std::vector<std::pair<VariantKey, std::string>> failed_deletes;
```
poodlewars, line 205: Minor but a struct rather than the pair might be more readable?

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -240,8 +241,19 @@ namespace s3 {
 
             util::check(to_delete.empty(), "Have {} segment that have not been removed",
                         to_delete.size());
-            if (!failed_deletes.empty())
-                throw KeyNotFoundException(Composite<VariantKey>(std::move(failed_deletes)));
+            if (!failed_deletes.empty()){
+                auto failed_deletes_message = std::ostringstream();
+                for (auto i=0u; i<failed_deletes.size(); ++i){
+                    auto& failed = failed_deletes[i];
```
poodlewars, line 247: If we keep the pair then consider expanding it with a structured binding here rather than `.first` and `.second`

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -240,8 +241,19 @@ namespace s3 {
 
             util::check(to_delete.empty(), "Have {} segment that have not been removed",
                         to_delete.size());
-            if (!failed_deletes.empty())
-                throw KeyNotFoundException(Composite<VariantKey>(std::move(failed_deletes)));
+            if (!failed_deletes.empty()){
```
poodlewars, line 244: Space before `{` please :pray:

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -240,8 +241,19 @@ namespace s3 {
...
+                for (auto i=0u; i<failed_deletes.size(); ++i){
+                    auto& failed = failed_deletes[i];
+                    failed_deletes_message<<fmt::format("'{}' failed with '{}'", to_serialized_key(failed.first), failed.second);
+                    if (i != failed_deletes.size()){
+                        failed_deletes_message<<", ";
+                    }
+                }
+                auto error_message = fmt::format("Failed to delete some of the objects: {}.",
+                                                 failed_deletes_message.str());
+                raise<ErrorCode::E_UNEXPECTED_S3_ERROR>(error_message);
```
poodlewars, line 255: Some of the keys not being present might still be the root cause right? In which case `KeyNotFoundException` would still be suitable. I'm concerned how we will change the catch-es further up the stack from `std::exception` to `KeyNotFoundException` with this change. Is there any way to get more info from AWS SDK about the root cause of the failures?

poodlewars: I mean eventually we'll want to make the change to `version_map.hpp` here https://github.com/man-group/ArcticDB/pull/584/files#diff-d6219b28b5b4e3cb754fc3fd7b77f5faddb5c965ca27191157cd16018baf939d

poodlewars: Resolving - Ivo explained that AWS SDK specifically states that deleting a non-existent object does not raise

poodlewars (approved):

PR #1332, opened 2024-02-15
build: Use C++20

### LINKED ISSUE #1229: C++ 20 Upgrade
We're on 17 at the moment.

We should try to stay on the cutting edge.

C++ 20 modules might be very helpful in defining the C++ API around ArcticDB, for example for the arcticdb-enterprise project to consume.

jjerphan:
#### Reference Issues/PRs

To be merged after https://github.com/man-group/ArcticDB/pull/1333.

Fixes #1229.

Use C++20 globally to be able to use new features of the language instead of folly's and `xtensor/xparrow` which is developed using this standard.

Make the necessary adaptations to be able to compile the project.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


vasil-pashov:
Why does `fmt` require changes with C++20?

jjerphan:
The behavior of `fmt` changes with C++20, as briefly explained [in their documentation](https://fmt.dev/9.0.0/api.html#core-api).

Also, if we want to transition to using the interfaces in `std`, some more changes will be required, as explained in [this blog post](https://biowpn.github.io/bioweapon/2023/08/01/std-format.html).

Klaim (commented):
The only missing thing now is to figure out why cmake overrides our `/std:c++20` flag everywhere

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -34,6 +34,23 @@ static const char* failure_names[] = {
...
+struct formatter<arcticdb::FailureType> {
+    template<typename ParseContext>
+    constexpr auto parse(ParseContext &ctx) { return ctx.begin(); }
+
+    template<typename FormatContext>
+    auto format(const arcticdb::FailureType failure_type, FormatContext &ctx) const {
+        return fmt::format_to(ctx.out(), fmt::runtime(arcticdb::failure_names[int(failure_type)]));
+    }
+};
+}
```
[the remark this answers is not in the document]
Klaim: I think it would be better to add comments about this, so that future maintainers have a chance to learn this. Keeping the info close to the code.

Klaim:
> @Klaim
> 
> > The only missing thing now is to figure out why cmake overrides our /std:c++20 flag everywhere
> 
> Can you elaborate? I can't understand the issue.

CMake generates visual studio projects which do not build with C++20 using this pr, even if clearly specified. Well it depends on the target/project but most targets compile with C++17 instead. I've observed this on my computer few days ago and confirmed it today on CI, see for example: https://github.com/man-group/ArcticDB/actions/runs/7972577696/job/21764697724?pr=1332#step:15:1291
where we can find:
```
cl : Command line warning D9025 : overriding '/std:c++20' with '/std:c++17'
```

Since I found this issue I tried various alternative ways to specify C++20 with cmake in a separate "minimal changes for C++20" branch (see `klaim/c++20`) but got the same results. I also tried without using presets, with caching the `CMAKE_CXX_STANDARD` variable and various other "hijacking" hacks, nothing works.
The only thing that allowed me to check the changes properly on windows/msvc was to change the flag in the generated project locally. This is, of course, not an acceptable solution.

Looking for info about that warning in the logs we can find (this also affects warning level flags in arcticdb): https://stackoverflow.com/questions/58708772/cmake-project-in-visual-studio-gives-flag-override-warnings-command-line-warnin

Which suggests we might need to either enforce the previous cmake policy about this, OR find another "clean" way to do it.

Note that the changes here are compatible with C++17 (because we didnt use anything that C++20 added and fmt  just use the c++17 constructs in this case, which are still the same api) so that's why it does build at the moment.

Klaim:
@vasil-pashov If you have suggestions for another way to do it that could work, or find one on your side, I'm interested.

cpp/arcticdb/storage/lmdb/lmdb_storage.hpp
```diff
@@ -18,6 +18,10 @@
 #ifdef ARCTICDB_USING_CONDA
     #include <lmdb++.h>
 #else
+    // LMDB++ is using `std::is_pod` in `lmdb++.h`, which is deprecated as of C++20.
```
Klaim, line 21: Do deprecation warning impacts some of the build at the moment?

vasil-pashov:
> > @Klaim
> > > The only missing thing now is to figure out why cmake overrides our /std:c++20 flag everywhere
> > 
> > 
> > Can you elaborate? I can't understand the issue.
> 
> CMake generates visual studio projects which do not build with C++20 using this pr, even if clearly specified. Well it depends on the target/project but most targets compile with C++17 instead. I've observed this on my computer few days ago and confirmed it today on CI, see for example: https://github.com/man-group/ArcticDB/actions/runs/7972577696/job/21764697724?pr=1332#step:15:1291 where we can find:
> 
> ```
> cl : Command line warning D9025 : overriding '/std:c++20' with '/std:c++17'
> ```
> 
> Since I found this issue I tried various alternative ways to specify C++20 with cmake in a separate "minimal changes for C++20" branch (see `klaim/c++20`) but got the same results. I also tried without using presets, with caching the `CMAKE_CXX_STANDARD` variable and various other "hijacking" hacks, nothing works. The only thing that allowed me to check the changes properly on windows/msvc was to change the flag in the generated project locally. This is, of course, not an acceptable solution.
> 
> Looking for info about that warning in the logs we can find (this also affects warning level flags in arcticdb): https://stackoverflow.com/questions/58708772/cmake-project-in-visual-studio-gives-flag-override-warnings-command-line-warnin
> 
> Which suggests we might need to either enforce the previous cmake policy about this, OR find another "clean" way to do it.
> 
> Note that the changes here are compatible with C++17 (because we didnt use anything that C++20 added and fmt just use the c++17 constructs in this case, which are still the same api) so that's why it does build at the moment.

Hmmm this is really interesting. I have a few tasks with higher priority now but I'd definitely give it a try and see if I can help somehow. I've tested C++ 20 on a pet project of mine with CMake and MSVC and it's working. Obviously something in Arctic's environment is wrong. If I had to guess I'd say that VCPKG might have something to do with this.

Klaim:
@vasil-pashov We found that it's Folly that is injecting C++17 version flags into dependent targets which later is taken into consideration first instead of the C++20 flag we add. 🥲 
@jjerphan pointed that we could patch folly through the vcpkg thing so I'm attempting to make that work. Basically, any library out there must NOT force it's user to use a particular C++ version flag, otherwise a library in C++17 would force a C++20 usercode to use C++17. 
Anyway, for now I'm patching locally to see if that fixes the issue.

cpp/arcticdb/storage/memory/memory_storage.cpp
```diff
@@ -110,7 +110,7 @@ namespace arcticdb::storage::memory {
                 auto it = key_vec.find(k);
 
                 if(it != key_vec.end()) {
-                    ARCTICDB_DEBUG(log::storage(), "Read key {}: {}, with {} bytes of data", variant_key_type(k), variant_key_view(k));
+                    ARCTICDB_DEBUG(log::storage(), "Read key {}: {}, with {} bytes of data", variant_key_type(k), variant_key_view(k), k);
```
alexowens90, line 113:
```suggestion
                    ARCTICDB_DEBUG(log::storage(), "Removed key {}: {}", variant_key_type(k), variant_key_view(k));
```

alexowens90 (changes requested):

alexowens90:
> The behavior of `fmt` changes with C++20, as briefly explained [in their documentation](https://fmt.dev/9.0.0/api.html#core-api).
> 
> Also, if we want to transition to using the interfaces in `std`, some more changes will be required, as explained in [this blog post](https://biowpn.github.io/bioweapon/2023/08/01/std-format.html).

Looks like the changes required to move to `std::format` are minimal (I'm not aware of using dynamic format strings anywhere). We should add a chore ticket to swap them out at some point. In the same piece of work we should unify on a new standard file layout, since we can no longer use class definition then formatter consistently now that formatters need to be forward declared in some places.

vasil-pashov:
Can you also promote the warning related to overriding the language version flag to an error so that it's easier to spot if it happens again?

jjerphan:
I think we should first update folly to the same version that the one used for conda builds.

In this regards, I have opened https://github.com/microsoft/vcpkg/pull/36893.

cpp/arcticdb/util/test/generators.hpp
```diff
@@ -54,7 +54,7 @@ struct SinkWrapperImpl {
             SchemaPolicy{
                 index_.create_stream_descriptor(std::move(stream_id), fields), index_
             },
-            [=](
+            [=, this](
```
willdealtry, line 57: We only need sink, so we should either capture it minimally by value [sink=sink_] or just capture 'this'

cpp/arcticdb/entity/types.hpp
```diff
@@ -194,7 +194,7 @@ enum class DataType : uint8_t {
     UNKNOWN = 0,
 };
 
-std::string_view datatype_to_str(DataType dt);
+std::string_view datatype_to_str(const DataType dt);
```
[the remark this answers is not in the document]
willdealtry: I don't hugely mind, but my concern is that one of the most common mistakes I see is intending to pass by const reference but forgetting the reference. With const pass-by-value, for none-trivial types, I'm left wondering whether this is a 'for indication only' const or a missing reference. For POD types that would naturally be passed either by value or non-const reference the meaning is clearer, but I wouldn't want us to start doing it with non-trivial types as a general thing

cpp/arcticdb/util/test/generators.hpp
jjerphan: Shouldn't we capture `sink_` by reference?

jjerphan: For now, e8cb5f7d532e77c1679f69e924881cf73fcec9c7 adapts the lambda only to capture `this`.

alexowens90 (approved):

jjerphan:
@vasil-pashov: thank you, for your interjection.

I think it would be worth promoting such warnings to errors. When compiling with MSVC, it appeared as compiler errors, so we might want to promote both warnings.

vasil-pashov:
> @vasil-pashov: thank you, for your interjection.
> 
> I think it would be worth promoting such warnings to errors. When compiling with MSVC, it appeared as compiler errors, so we might want to promote both warnings.

I played with this for a while and it seems like it's not easy (maybe not possible at all) to turn this into a warning. There are a few problems.
1. D9025 is not a compiler warning, but a [Command Line Warning](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/command-line-errors-d8000-through-d9999?view=msvc-170) so it's not controllable by the usual compiler flags e.g. /we<error-code>. Not only that but the warning is not issued when building from Visual Studio. It's emitted only when build from the command line (invoking cl.exe followed by ninja as in the CI)
2. CMake doesn't seem to have a way to control these warnings. Visual Studio itself doesn't seem to have a way to control them.

I suggest we leave it as is.

Klaim:
> > @vasil-pashov: thank you, for your interjection.
> > I think it would be worth promoting such warnings to errors. When compiling with MSVC, it appeared as compiler errors, so we might want to promote both warnings.
> 
> I played with this for a while and it seems like it's not easy (maybe not possible at all) to turn this into a warning. There are a few problems.
> 
> 1. D9025 is not a compiler warning, but a [Command Line Warning](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/command-line-errors-d8000-through-d9999?view=msvc-170) so it's not controllable by the usual compiler flags e.g. `/we<error-code>`. Not only that but the warning is not issued when building from Visual Studio. It's emitted only when build from the command line (invoking cl.exe followed by ninja as in the CI)
> 2. CMake doesn't seem to have a way to control these warnings. Visual Studio itself doesn't seem to have a way to control them.
> 
> I suggest we leave it as is.

Interesting, thanks for exploring this, I didnt find the time to get into it yet. I had my suspicion that it is not controllable either but didnt remember why I had that gut feeling.

Note that we could add an actual check in the code to not compile if it's not at least c++20, something like
```c++
static_assert(__cplusplus >= 202002L);
```
See for details: https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros
Note that for msvc, [this flag](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170) has to be set for the macro to be correctly defined (because of legacy reasons) but that's pretty common, even folly injected it in it's users 😬 
Anyway if all the compilers we use do handle that macro correctly, that check alone can be a better showstopper than the warning we currently get, whatever the compiler and buildsystem.
What do you think?

cpp/arcticdb/entity/index_range.hpp
```diff
@@ -155,28 +144,44 @@ inline IndexRange universal_range(){ return IndexRange{std::numeric_limits<times
...
+    auto format(const arcticdb::entity::IndexRange& r, FormatContext& ctx) const {
         return fmt::format_to(ctx.out(), "{}-{}", r.start_, r.end_);
     }
 };
 
 } //namespace fmt
+
+namespace arcticdb::entity {
+
+inline bool IndexRange::accept(const IndexValue &index) {
```
Klaim, line 174: There is a comment missing about this definition needing the formatter before it's definition (as for the other functions)

Klaim (approved):
Other than some missing comments, LGTM!

cpp/CMakeLists.txt
```diff
@@ -1,7 +1,13 @@
 cmake_minimum_required(VERSION 3.21) # TARGET_RUNTIME_DLLS
+cmake_policy(SET CMP0048 NEW)
```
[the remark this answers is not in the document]
jjerphan: This line was already present in this file, but was moved at the top of the file by 36b45ba330a8a0105843b22f68d5baac91ebd51b.

Which comment would you add? :)

vasil-pashov:
> > > @vasil-pashov: thank you, for your interjection.
> > > I think it would be worth promoting such warnings to errors. When compiling with MSVC, it appeared as compiler errors, so we might want to promote both warnings.
> > 
> > 
> > I played with this for a while and it seems like it's not easy (maybe not possible at all) to turn this into a warning. There are a few problems.
> > 
> > 1. D9025 is not a compiler warning, but a [Command Line Warning](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/command-line-errors-d8000-through-d9999?view=msvc-170) so it's not controllable by the usual compiler flags e.g. `/we<error-code>`. Not only that but the warning is not issued when building from Visual Studio. It's emitted only when build from the command line (invoking cl.exe followed by ninja as in the CI)
> > 2. CMake doesn't seem to have a way to control these warnings. Visual Studio itself doesn't seem to have a way to control them.
> > 
> > I suggest we leave it as is.
> 
> Interesting, thanks for exploring this, I didnt find the time to get into it yet. I had my suspicion that it is not controllable either but didnt remember why I had that gut feeling.
> 
> Note that we could add an actual check in the code to not compile if it's not at least c++20, something like
> 
> ```c++
> static_assert(__cplusplus >= 202002L);
> ```
> 
> See for details: https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros Note that for msvc, [this flag](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170) has to be set for the macro to be correctly defined (because of legacy reasons) but that's pretty common, even folly injected it in it's users 😬 Anyway if all the compilers we use do handle that macro correctly, that check alone can be a better showstopper than the warning we currently get, whatever the compiler and buildsystem. What do you think?

Seems reasonable to me. How do you suggest we do this? I'm thinking about a separate TU inside arctic. I don't think we should care if deps are defining the flag in question. We're interested if deps are overriding Arctic's version. If we override the version of a dependency either nothing will happen or we'll get an error compiling the dependency, so in both cases we'll be notified.

Klaim:
> Seems reasonable to me. How do you suggest we do this? I'm thinking about a separate TU inside arctic. I don't think we should care if deps are defining the flag in question. We're interested if deps are overriding Arctic's version. If we override the version of a dependency either nothing will happen or we'll get an error compiling the dependency, so in both cases we'll be notified.

Yes a translation unit is best as this library has a binary anyway.
We can do it in a separate PR after this one, so that we know we are in C++20 already.

cpp/CMakeLists.txt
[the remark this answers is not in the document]
Klaim: I dont really know actually, I was wondering. It's best if documented but as it was already there then nevermind.

PR #1350, opened 2024-02-21
Removed pytest dependency from arcticdb

DrNickClarke:
* moved param_dict code from arcticdb/util/test.py to python/tests/util/mark.py
* removed xfail arg and if clause (if was not used and broken - it appended a tuple to a list)
* updated the only use case: test_normalization.py


jjerphan (commented):
Thank you, @DrNickClarke!

LGTM modulo one request: could you move the specification of `pytest` bellow line 77 in the conda environment specification so that we know `pytest` isn't a runtime dependency but only one for the tests?

https://github.com/man-group/ArcticDB/blob/e69297890b824c0c2a83b285e0cdb94b2e228452/environment_unix.yml#L66

jjerphan (approved):

Issue #340, opened 2023-05-05
Writing even very small pickled data can fail on low memory machines


poodlewars:
# Issue

When writing pickled data, we allocate a very large (4 GiB) buffer.

This can fail on some machines. For example, I have observed this on a Windows machine with 8GB memory (and the default 1.5GB page file).

```
l.write_pickle("sym", [1,2])
```

```
File C:\venvs\310_1\lib\site-packages\arcticdb\version_store\_normalization.py:852, in MsgPackNormalizer.normalize(self, obj, **kwargs)
    851 def normalize(self, obj, **kwargs):
--> 852     buffer = mmap(-1, self._size)
    853     try:

OSError: [WinError 1450] Insufficient system resources exist to complete the requested service

During handling of the above exception, another exception occurred:

ArcticException                           Traceback (most recent call last)
Cell In[13], line 1
----> 1 l.write_pickle("sym", [1,2])
```

The stack also includes,

```
OSError: [WinError 1455] The paging file is too small for this operation to complete

During handling of the above exception, another exception occurred:
```

In the example above we are writing _very_ little data, just `[1,2]`.

where `l` is an LMDB backed library.

## Notes

It is unclear how this issue affects small Linux boxes, which should also be tested.

In principle a similar issue affects writing metadata too, since that is pickled. However the buffer size is much smaller, so it should not cause an issue in practice.

# Workaround

You can adjust the buffer size by setting:

```
import arcticdb.version_store._normalization as n

In [22]: n.MsgPackNormalizer.MMAP_DEFAULT_SIZE
Out[22]: 4294968320  # the default, about 4GB

In [23]: n.MsgPackNormalizer.MMAP_DEFAULT_SIZE = 1 << 31  # adjust for your machine, this is 2GB
```

PR #1366, opened 2024-02-25
Bugfix #340: create allocate_buffer function with backoff for low-memory machines and Windows

jamesmunro:

Issue #340 

#### Any other comments?

Please consider whether this backoff approach is the best approach, and whether there are better ways to test this.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


willdealtry (approved):

python/arcticdb/version_store/_normalization.py
```diff
@@ -937,6 +938,31 @@ def _packb(*args, **kwargs):
...
+    assert attempts > 0
+    assert divisor > 0
+
+    while attempts > 0:
+        try:
+            return mmap(-1, max_size)
+        except OSError as oserror:
+            log.error("Failed to allocate buffer with size: {}B, error: {}".format(max_size, oserror))
+            if (oserror.errno == errno.ENOMEM) or \
+                (oserror.errno == errno.EINVAL and hasattr(oserror, 'winerror') and oserror.winerror != 1450):
```
jamesmunro, line 953: Bug here.  Should be `== 1450`.   I'll retest on windows.

jamesmunro: - [ ] Fix test and code for windows

python/arcticdb/version_store/_normalization.py
```diff
@@ -937,6 +938,31 @@ def _packb(*args, **kwargs):
     return packb(*args, **kwargs)
 
 
+def allocate_buffer(max_size: int , attempts: int = 4, divisor: int = 2):
+    """Allocate a buffer of max_size, retrying up to attempts times on out-of-memory errors, reducing the size by divisor each time."""
+    assert max_size > 0
+    assert attempts > 0
+    assert divisor > 0
```
alexowens90, line 945: https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/preconditions.py#L11
Generally use this instead of `assert`

python/arcticdb/version_store/_normalization.py
```diff
@@ -937,6 +938,31 @@ def _packb(*args, **kwargs):
     return packb(*args, **kwargs)
 
 
+def allocate_buffer(max_size: int , attempts: int = 4, divisor: int = 2):
```
alexowens90, line 941: This combination of parameters means that the smallest attempted allocation will be 500MB, I would add a couple more attempts

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -557,3 +560,55 @@ def test_pyarrow_error(lmdb_version_store):
...
+    b1 = allocate_buffer(100)
+    assert len(b1) == 100
+    assert hasattr(b1, "write") # Check that it's a buffer-like object
+    b2 = allocate_buffer(200)
+    assert len(b2) == 200
+    assert hasattr(b2, "write")
+
+
+def test_allocate_buffer_too_big():
+    try:
```
alexowens90, line 575: `with pytest.raises(OSError) as e`

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -557,3 +560,55 @@ def test_pyarrow_error(lmdb_version_store):
...
+    mmap.side_effect = OSError(errno.ENOMEM, "Out of memory")
+    try:
+        allocate_buffer(16 << 20)
+    except OSError as e:
+        assert e.errno == errno.ENOMEM        
+        assert mmap.call_count == 4
+        assert mmap.call_args_list[0][0] == (-1, 16 << 20)
+        assert mmap.call_args_list[1][0] == (-1, 16 << 19)
+        assert mmap.call_args_list[2][0] == (-1, 16 << 18)
+        assert mmap.call_args_list[3][0] == (-1, 16 << 17)
```
alexowens90, line 614: I would add one more test that if you specify a large size and a huge number of attempts that it eventually succeeds

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_normalization.py
jamesmunro: mocked out or not?  I was trying to avoid a test that would OMM the testing environment.

jamesmunro:
After further investigation, it seems to mmap'd buffer was originally created to limit the msgpack serialisation to 10MB, it was then progressively increased to now, 4GB.  4GB is roughly the limit for msgpack in our case (depending on the data stored): https://github.com/msgpack/msgpack/blob/master/spec.md#limitation

However this is now causing other issues, 
- always allocating 4GB regardless of data size stored
- failure when you cannot allocate 4GB, which we've seen reported on Windows

I'm closing this.  Unless I get feedback to the contrary, I'll replace with a PR to remove the mmap buffer for this case and deprecate `max_blob_size`.

The metadata buffer limit still seems useful so I intend to leave that.

PR #1375, opened 2024-02-26
Add metadata extraction functions in library_tool

IvoDD:
- Adds read_descriptor
- Adds read_timeseries_descriptor
- Adds TimeseriesDescriptor python binding
- Fixes StreamDescriptor fields() function
- Adds a test to showcase how lib_tool.read_timeseries_descriptor can be used for type promotion tests

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/stream/python_bindings.cpp
```diff
@@ -24,6 +24,14 @@ namespace py = pybind11;
 namespace arcticdb {
 using namespace arcticdb::python_util;
 
+std::vector<FieldRef> field_collection_to_ref_vector(const FieldCollection& fields){
+    auto result = std::vector<FieldRef>();
```
vasil-pashov, line 28: 1. You can reserve using `FieldCollection::size`
2. Given that `FieldCollection` I think we can use `std::copy`

cpp/arcticdb/toolbox/library_tool.cpp
```diff
@@ -37,6 +37,22 @@ Segment LibraryTool::read_to_segment(const VariantKey& key) {
     return kv.segment();
 }
 
+std::optional<google::protobuf::Any> LibraryTool::read_metadata(const VariantKey& key){
+    auto store = std::make_shared<arcticdb::async::AsyncStore<util::SysClock>>(lib_, codec::default_lz4_codec(), encoding_version(lib_->config()));
```
vasil-pashov, line 41: I'm just wondering why do we need a shared_ptr?

cpp/arcticdb/toolbox/library_tool.hpp
```diff
@@ -36,6 +36,12 @@ class LibraryTool {
 
     Segment read_to_segment(const VariantKey& key);
 
+    std::optional<google::protobuf::Any> read_metadata(const VariantKey& key);
+
+    StreamDescriptor read_descriptor(const VariantKey& key);
+
+    TimeseriesDescriptor read_timeseries_descriptor(const VariantKey& key);
```
vasil-pashov, line 43: NIT: I'd mark those as [[nodiscard]]

python/tests/unit/arcticdb/version_store/test_column_type_changes.py
```diff
@@ -90,4 +93,32 @@ def test_changing_fixed_string_width(version_store_factory, dynamic_schema, wide
 
     expected_update = pd.DataFrame({"col": ["aa", "d" * (1 if wider_strings_first else 3), "cc"]}, index=pd.date_range("2024-01-01", periods=3))
     received_update = lib.read(sym_update).data
-    assert_frame_equal(expected_update, received_update)
\ No newline at end of file
+    assert_frame_equal(expected_update, received_update)
+
+def test_type_promotion_stored_in_index_key(version_store_factory):
```
vasil-pashov, line 98: NIT: You can use the `lmdb_version_store_dynamic_schema` to cover for both encoding versions and avoid the call to  `version_store_factory(dynamic_schema=True)`. I'm not sure if you can call `library_tool` on it, though...

cpp/arcticdb/toolbox/library_tool.cpp
IvoDD: Good point, I just copied it from another method which needed the shared_ptr. 
Instead I created a commit to use a `store_` property to be used by all methods, instead of creating it every time.

vasil-pashov: Is there a reason to create a pointer at all?
1. Can we use const ref instead of a pointer?
2. If we can't use const ref I think `std::unique_ptr` is better.

cpp/arcticdb/stream/python_bindings.cpp
vasil-pashov: My personal preference is to use `vector::reserve` + `std::back_inserter` (instead of setting the size in the constructor of the vector) to avoid the call to the default constructor which will later be overridden anyways. I'm aware that it's a micro-optimization that probably won't change anything, so I'll leave it up to you to decide if you want to leave it as is.

python/arcticdb/toolbox/library_tool.py
```diff
@@ -136,3 +137,23 @@ def read_to_keys(
...
+
+
+    def read_descriptor(self, key: VariantKey) -> StreamDescriptor:
+        """
+        Gives the <StreamDescriptor> for a Variant key. The Stream Descriptor contains the <FieldRef>s for all fields in
+        the value written for that key.
+
+        E.g. an Index key will have fields like 'start_index', 'end_index', 'creation_ts', etc.
+        """
+        return super().read_descriptor(key)
```
alexowens90, line 149: You can add docstrings directly to the method definition in the Python bindings:
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/version/python_bindings.cpp#L151-L175

cpp/arcticdb/toolbox/library_tool.hpp
```diff
@@ -54,6 +61,7 @@ class LibraryTool {
 
 private:
     std::shared_ptr<storage::Library> lib_;
+    std::shared_ptr<arcticdb::async::AsyncStore<util::SysClock>> store_;
```
alexowens90, line 64: I think `store_` can probably do everything `lib_` can do, so could get rid of `lib_`, but this is only a debug/testing tool so not a requirement

cpp/arcticdb/stream/python_bindings.cpp
```diff
@@ -24,6 +24,12 @@ namespace py = pybind11;
 namespace arcticdb {
 using namespace arcticdb::python_util;
 
+std::vector<FieldRef> field_collection_to_ref_vector(const FieldCollection& fields){
+    auto result = std::vector<FieldRef>(fields.size());
```
alexowens90, line 28: This will call the default constructor of `FieldRef` `fields.size()` times, `reserve` avoids this

alexowens90 (changes requested):

cpp/arcticdb/toolbox/library_tool.cpp
IvoDD: As discussed offline:

We keep using a `shared_ptr<AsyncStore>` because it is required to pass in as an argument to `delete_all_keys_of_type`.

Instead left a TODO comment to remove the `shared_ptr` when `delete_all_keys_of_type` changes its argument to a const ref.

cpp/arcticdb/toolbox/library_tool.hpp
IvoDD: Cool, done

alexowens90 (approved):

PR #1379, opened 2024-02-28
Disable missing key warnings when expected

IvoDD:
- Sets `opts.dont_warn_about_missing_key=true` when missing keys are expected (For example when writing a previously missing symbol it is expected to try to read a missing ref key)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -153,10 +153,10 @@ namespace s3 {
                           ReadKeyOpts opts) {
             ARCTICDB_SAMPLE(S3StorageRead, 0)
             auto fmt_db = [](auto &&k) { return variant_key_type(k); };
-            std::vector<VariantKey> failed_reads;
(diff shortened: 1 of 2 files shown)


muhammadhamzasajjad:
@IvoDD, whatever fix you decide to use, please ensure that it is consistent between all the storages. Currently, you are removing the warning only from s3, but the same warning is also logged in other storages. If you decide to stick with this solution, please remove the warning from other storages too.

Instead if you decide to set `opts.dont_warn_about_missing_key` to true as a solution, then the behaviour should already be consistent between the storages.

poodlewars:
As discussed in Slack let's set `opts.dont_warn_about_missing_key=true` on `read_symbol_ref` and leave the storage backends untouched. I'm assuming that `read_symbol_ref` is the only additional place we wish to suppress the warning, but worth checking if we should add it elsewhere too.

poodlewars (approved):

PR #1381, opened 2024-02-29
Add a global timeout for pytests

G-D-Petrov:
We are experiencing tests in pytest that get stuck.
It is very hard to debug as they are just causing the overall CI step to timeout.
This change sets a global timeout for the pytests of 1 hour.
This way the stuck tests will fail faster (1 hours vs 6 hours) and it will be noted as a failed test se we can understand which tests exactly are stuck so we can fix them.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


poodlewars (approved):
Good idea! Do we have any tests that we expect to take so long? I wonder if a timeout of more like 15 minutes might be suitable and give a faster feedback loop.

vasil-pashov:
Is the limit 1h per test or it's 1h for the whole suite?

PR #1386, opened 2024-02-29
Example implementation of enumerate

willdealtry:
Example of internal implementation of enumerate

### DIFF
--- a/cpp/arcticdb/util/enumerate.hpp
+++ b/cpp/arcticdb/util/enumerate.hpp
@@ -4,6 +4,11 @@
 
 #include <iterator>
 #include <arcticdb/util/preprocess.hpp>
+#if defined(_MSC_VER)
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
 
 namespace arcticdb {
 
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/util/enumerate.hpp
```diff
@@ -0,0 +1,119 @@
...
+};
+
+template <class Iterator>
+class Enumerator {
+public:
+    constexpr explicit Enumerator(Iterator it) : it_(std::move(it)) {}
+
+    class Proxy {
+    public:
+        using difference_type = ssize_t;
```
vasil-pashov, line 39: What is `difference_type` supposed to represent? Is it difference between Enumerator::idx_? If so it seems a bit dangerous as idx_ is of unsigned type.

cpp/arcticdb/util/test/test_enumerate.cpp
```diff
@@ -0,0 +1,30 @@
+#include <gtest/gtest.h>
+#include <arcticdb/util/enumerate.hpp>
+#include <arcticdb/log/log.hpp>
+
+TEST(Enumerate, Vector) {
+    std::vector<int> vec = {1, 2, 3, 4};
```
vasil-pashov, line 6: nit: `static constexpr std::array = {1,2,3,4}`

PR #1331, opened 2024-02-15
Add a `MockAzureClient` which can simulate azure failures

muhammadhamzasajjad:
This is similar to #1281.
- Adds a config option to use the `MockAzureClient` instead of the `RealAzureClient`
- If a `MockAzureClient` is created, one can simulate failures by passing the appropriate symbol name
- Creates tests to ensure that `MockAzureClient` works properly for `read`, `write`, `update`, `delete` and `list` functions.

To do: Test exceptions thrown by `MockAzureClient`. This will be done in the next PR where we normalize Azure exception.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #1395, opened 2024-03-05
#447 Add a `MockMongoClient` which can simulate mongo failures

muhammadhamzasajjad:
Similar to #1331. The mock client will then be used to test exception normalization. This patch also wraps up the common logic between all Mock clients into `MockStorageClient` as mentioned in [this comment](https://github.com/man-group/ArcticDB/pull/1331#discussion_r1491276078)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
```diff
@@ -0,0 +1,201 @@
+/* Copyright 2023 Man Group Operations Limited
```
vasil-pashov, line 1:
```suggestion
/* Copyright 2024 Man Group Operations Limited
```

cpp/arcticdb/storage/mongo/mongo_mock_client.hpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    mongocxx::operation_exception get_exception() const {
+        return std::get<mongocxx::operation_exception>(failure);
+    }
+
+};
+
+// the key is composed of the database name, collection name and the variant key
+using mongo_document_key = std::tuple<std::string, std::string, std::string>;
```
IvoDD, line 33: What do you think about using a struct instead?

cpp/arcticdb/storage/azure/azure_mock_client.cpp
```diff
@@ -63,81 +52,62 @@ std::optional<Azure::Core::RequestFailedException> has_failure_trigger(const std
     }
 }
 
+Azure::Core::RequestFailedException MockAzureClient::missing_key_failure() const {
+    auto error_code = AzureErrorCode_to_string(AzureErrorCode::BlobNotFound);
+    std::string message = fmt::format("Simulated Error, message: Failed to find blob, {} {}", error_code,
+                                      (int) Azure::Core::Http::HttpStatusCode::NotFound);
```
poodlewars, line 58: While we're here I agree with Phoebus' comment in the other PR that we should replace these C-style casts with `static_cast`

cpp/arcticdb/storage/azure/azure_mock_client.hpp
```diff
@@ -14,18 +14,11 @@
...
-enum class AzureOperation{
-    READ,
-    WRITE,
-    DELETE, // delete_blobs will fail for all objects if one of them triggers a delete failure
-    LIST,
-    EXISTS,
-};
-
-class MockAzureClient : public AzureClientWrapper {
+class MockAzureClient : public AzureClientWrapper, public MockStorageClient<std::string, Azure::Core::RequestFailedException> {
```
poodlewars, line 21: Since AFAICT the MockStorageClient is only here for code reuse, not for runtime polymorphism, I think it would be cleaner to use composition over inheritance for it.

Here is a page on why to worry about multiple inheritance: https://stackoverflow.com/questions/406081/why-should-i-avoid-multiple-inheritance

And a more balanced one https://isocpp.org/wiki/faq/multiple-inheritance

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+    return fmt::format("{}#Failure_{}_{}", key, operation_to_string(operation_to_fail), static_cast<int>(error_code));
+}
+
+template <typename exception_type>
+MongoFailure create_failure(const std::string& message, MongoError error_code) {
+    static_assert(std::is_base_of<mongocxx::operation_exception, exception_type>::value, "exception_type must be a subclass of mongocxx::operation_exception");
+    if(error_code == MongoError::NoAcknowledge) return {std::monostate()};
+
+    bsoncxx::document::value empty_doc_value = bsoncxx::builder::basic::document{}.extract();
+    auto ec = std::error_code((int)error_code, std::generic_category());
```
poodlewars, line 33: C-style cast

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+        const std::string& key,
+        StorageOperation operation_to_fail,
+        MongoError error_code) {
+    return fmt::format("{}#Failure_{}_{}", key, operation_to_string(operation_to_fail), static_cast<int>(error_code));
+}
+
+template <typename exception_type>
+MongoFailure create_failure(const std::string& message, MongoError error_code) {
+    static_assert(std::is_base_of<mongocxx::operation_exception, exception_type>::value, "exception_type must be a subclass of mongocxx::operation_exception");
+    if(error_code == MongoError::NoAcknowledge) return {std::monostate()};
```
poodlewars, line 30: Avoid one line if

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    bsoncxx::document::value empty_doc_value = bsoncxx::builder::basic::document{}.extract();
+    auto ec = std::error_code((int)error_code, std::generic_category());
+
+    return {exception_type(ec, std::move(empty_doc_value), message)};
```
poodlewars, line 35: is the std::move necessary?

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+std::optional<MongoFailure> MockMongoClient::has_failure_trigger(
+        const mongo_document_key& key,
+        StorageOperation operation) const {
+    auto key_id = get<2>(key);
+    auto failure_string_for_operation = "#Failure_" + operation_to_string(operation) + "_";
+    auto position = key_id.rfind(failure_string_for_operation);
+    if (position == std::string::npos) return std::nullopt;
```
poodlewars, line 63: avoid one line if

(applies to several pleaces in this PR)

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    try {
+        auto start = position + failure_string_for_operation.size();
+        auto error_code = MongoError(stoi(key_id.substr(start)));
+        auto error_message = fmt::format("Simulated Error, message: operation {}, error code {}",
+                                         operation_to_string(operation), static_cast<int>(error_code));
+
+        return get_failure(error_message, operation, error_code);
+    } catch (std::exception&) {
```
poodlewars, line 72: What case does this catch cover? Do we really want to swallow these?

cpp/arcticdb/storage/mongo/mongo_mock_client.hpp
```diff
@@ -0,0 +1,92 @@
...
+#pragma once
+
+#include <arcticdb/storage/storage_utils.hpp>
+#include <arcticdb/storage/mongo/mongo_client_wrapper.hpp>
+#include <arcticdb/storage/storage_mock_client.hpp>
+#include <mongocxx/exception/operation_exception.hpp>
+
+namespace arcticdb::storage::mongo {
+
+struct MongoFailure {
```
poodlewars, line 17: An alias might help to make the code "self-documenting" here,

```
using NoAckFailure = std::monostate;
```

cpp/arcticdb/storage/mongo/mongo_storage.cpp
```diff
@@ -145,13 +146,19 @@ MongoStorage::MongoStorage(
...
-        ConfigsMap::instance()->get_int("MongoClient.MinPoolSize", 100),
-        ConfigsMap::instance()->get_int("MongoClient.MaxPoolSize", 1000),
-        ConfigsMap::instance()->get_int("MongoClient.SelectionTimeoutMs", 120000))
-        ) {
+    instance_(MongoInstance::instance()) {
+    if(config.use_mock_storage_for_testing()) {
+        ARCTICDB_RUNTIME_DEBUG(log::storage(), "Using Mock Mongo storage");
+        client_ = std::make_unique<MockMongoClient>();
+    }
+    else {
```
poodlewars, line 154: (nit) we usually have the else on the line above

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
muhammadhamzasajjad: This could happen if someone messes up with the special `#Failure_` string instead of using `get_failure()`. In that case for example, `stoi` could fail and throw an exception. We just swallow the exception and return no failure because it hasn't been generated using `get_failure()`.  This was also done for s3 and [azure mock clients](https://github.com/man-group/ArcticDB/pull/1331/files#diff-5cc7ef4386f48dc47564ce4f98e1ddd8aff872a4d37ac11ac6c4c9388a64ff73R55).

Note that potential exceptions returned by `get_failure` aren't swallowed because they are only thrown by the caller. This function just returns them and only swallows exceptions thrown during parsing of the error codes.

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
muhammadhamzasajjad: Yes. Please view `mongocxx::operation_exception` [constructor](https://mongocxx.org/api/mongocxx-3.4.0/classmongocxx_1_1operation__exception.html#afd5c7cf50479bc765df9ea5726e5bc8f). The other exceptions are just subclasses of this class and use the same constructor.

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
poodlewars: I would think we want to blow up loudly is someone messes up the special string?

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
muhammadhamzasajjad: Fixed in [ea60ab1](https://github.com/man-group/ArcticDB/pull/1395/commits/ea60ab11b0de9987e1146fa249374be97efd0934)

cpp/arcticdb/storage/s3/s3_mock_client.hpp
```diff
@@ -23,32 +24,34 @@
...
-    LIST
-};
-
+namespace arcticdb::storage::s3 {
 
 // A mock S3ClientWrapper which can simulate failures.
 // The MockS3Client stores the segments in memory to simulate regular S3 behavior for unit tests.
 // The MockS3Client can simulate storage failures by using the get_failure_trigger for s3_object_names.
-class MockS3Client : public S3ClientWrapper {
+class MockS3Client : public S3ClientWrapper, public MockStorageClient<std::pair<std::string, std::string>, Aws::S3::S3Error> {
```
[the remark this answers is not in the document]
muhammadhamzasajjad: Fixed in [ea60ab1](https://github.com/man-group/ArcticDB/pull/1395/commits/ea60ab11b0de9987e1146fa249374be97efd0934)

cpp/arcticdb/CMakeLists.txt
```diff
@@ -448,6 +448,10 @@ set(arcticdb_srcs
         storage/azure/azure_real_client.cpp
         storage/azure/azure_mock_client.hpp
         storage/azure/azure_mock_client.cpp
+        storage/mongo/mongo_client_wrapper.hpp
+        storage/storage_mock_client.hpp
```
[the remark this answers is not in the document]
muhammadhamzasajjad: fixed in [ea60ab1](https://github.com/man-group/ArcticDB/pull/1395/commits/ea60ab11b0de9987e1146fa249374be97efd0934)

cpp/arcticdb/storage/mongo/mongo_mock_client.hpp
muhammadhamzasajjad: Fixed in [ea60ab1](https://github.com/man-group/ArcticDB/pull/1395/commits/ea60ab11b0de9987e1146fa249374be97efd0934). This is actually a very good idea. I have now defined a `MongoDocumentKey` which saves the entire `VariantKey` and defines `id_string()` which will contain the failure trigger(if set using `get_failure_trigger()`). This overcomes the previous [limitation in `list_keys()`](https://github.com/man-group/ArcticDB/pull/1395/files/7896c47e4be34fe966fd6bae28fb27629a525f97#diff-c25a56cd300b6b66f5ec51696594f1e745d58755fe77b64d2e4c15bafc6d03e7R179), we can now return any key instead of just `TABLE_DATA`.

cpp/arcticdb/storage/azure/azure_mock_client.cpp
muhammadhamzasajjad: Fixed in [3283f32](https://github.com/man-group/ArcticDB/pull/1395/commits/3283f32c0a323b10940abd25d0b876ca8b57984e)

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
muhammadhamzasajjad: Fixed in [3283f32](https://github.com/man-group/ArcticDB/pull/1395/commits/3283f32c0a323b10940abd25d0b876ca8b57984e)

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
muhammadhamzasajjad: Fixed in [3283f32](https://github.com/man-group/ArcticDB/pull/1395/commits/3283f32c0a323b10940abd25d0b876ca8b57984e)

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
muhammadhamzasajjad: Fixed in [3283f32](https://github.com/man-group/ArcticDB/pull/1395/commits/3283f32c0a323b10940abd25d0b876ca8b57984e)

cpp/arcticdb/storage/mongo/mongo_mock_client.hpp
muhammadhamzasajjad: Fixed in [3283f32](https://github.com/man-group/ArcticDB/pull/1395/commits/3283f32c0a323b10940abd25d0b876ca8b57984e)

cpp/arcticdb/storage/mongo/mongo_storage.cpp
muhammadhamzasajjad: Fixed in [3283f32](https://github.com/man-group/ArcticDB/pull/1395/commits/3283f32c0a323b10940abd25d0b876ca8b57984e)

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
poodlewars: As discussed this is exceptions as control flow - we expect the throw and catch for normal symbols that aren't the special failure strings. I think it's fair to say that fixing that here is scope creep, so I'm fine with this as is.

cpp/arcticdb/storage/s3/s3_mock_client.hpp
```diff
@@ -67,9 +71,9 @@ class MockS3Client : public S3ClientWrapper {
             const std::string& prefix,
             const std::string& bucket_name,
             const std::optional<std::string> continuation_token) const override;
+
 private:
-    // Stores a mapping from pair<bucket_name, s3_name> to a Segment.
-    std::map<std::pair<std::string, std::string>, Segment> s3_contents;
+    std::map<S3Key, Segment> s3_contents;
```
poodlewars, line 76: Don't understand why this is an ordered map?

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
poodlewars: You still have,

```
    if (position == std::string::npos)
        return std::nullopt;
```

should be:

```
    if (position == std::string::npos) {
        return std::nullopt;
    }
```

poodlewars (approved):

cpp/arcticdb/storage/s3/s3_mock_client.hpp
muhammadhamzasajjad: 1. The minor advantage is that when you run a `list_objects` on it, it will return a sorted list where all the keys are already lexicographically sorted. In [places like this](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/storage/test/test_s3_storage.cpp#L281C1-L285C6) it can be handy if you decide to print the list. (You could sort the returned list manually though if you use `std::unordered_map`
2. [`std::unordered_map`](https://en.cppreference.com/w/cpp/container/unordered_map) would require defining a hash function, the `operator<` and `operator==` for `S3Key` so `std::map` is slightly easier to setup. My reasoning for sticking to `std::map` is that we are not aiming for testing mock client with a huge number of symbols where `std::unordered_map` would be advantageous from a performance point of view. The maximum number of symbols we have tested a mock client with is 25.

cpp/arcticdb/storage/mongo/mongo_mock_client.cpp
poodlewars: I agree . We have https://github.com/man-group/ArcticDB/wiki/Development-Practices but nothing about style. Let's discuss at the retro.

PR #1400, opened 2024-03-06
build: Use C++23

jjerphan:
#### Reference Issues/PRs

Follow-up on https://github.com/man-group/ArcticDB/pull/1332/.

This is a requirement to use the latest feature of C++ to be able to replace some elements of folly.

#### Checklist


jjerphan:
We need to upgrade the compilers on the CI first.

PR #1407, opened 2024-03-08
Fix crash during normalization of time zones

### LINKED ISSUE #1406: ArcticDB crashes when when trying to normalize timezone data with MsgPack
### Describe the bug

ArcticDB crashes when trying to normalize timestamp with time zone when MsgPack is used as a normalizer.
This problem occurs most often when trying to write metadata with a time stamp with a time zone.

N.B.: Time zones are already supported for indexes and **not** supported for the data in regular columns.
This is because regular columns are treated as numpy arrays and numpy doesn't support time zones.

### Steps/Code to Reproduce

``` python
import pandas as pd
from arcticdb import Arctic

ac = Arctic("lmdb://tz_fix")
lib = ac.get_library("tz_fix", create_if_missing=True)
df = pd.DataFrame([{"dt": pd.Timestamp("2021-01-01", tz="Europe/Amsterdam")}])
df.index = df["dt"]
lib.write(
    "blah",
    data=df,
    metadata={"index_start": pd.Timestamp("2024-03-04", tz="Europe/Amsterdam")},
)

print(lib.read("blah").data.index.tz)
print(lib.read("blah").metadata["index_start"].tz)
print(df.iloc[0].dt.tz)
```

### Expected Results

Should not crash and the time zone in the metadata should be preserved. 

Python: 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
OS: Linux-5.15.146.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
ArcticDB: dev

### Backend storage used

LMDB

### Additional Context

_No response_

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1406 

Converts time zone data to string when normalizing with MsgPack.
Adds tests to make sure that the crash doesn't reproduce in the future.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/docs/mkdocs/docs/faq.md
+++ b/docs/mkdocs/docs/faq.md
@@ -183,3 +183,16 @@ The handling of `NaN` in ArcticDB depends on the type of the column under consid
 * For string columns, `NaN`, as well as Python `None`, are fully supported.
 * For floating-point numeric columns, `NaN` is also fully supported.
 * For integer numeric columns `NaN` is not supported. A column that otherwise contains only integers will be treated as a floating point column if a `NaN` is encountered by ArcticDB, at which point [the usual rules](api/arctic.md#LibraryOptions) around type promotion for libraries configured with or without dynamic schema all apply as usual.
+
+### How does ArcticDB handle `time zone` information?
+
+ArcticDB takes a more strict approach to time zone handling than Pandas. While Pandas support multiple types for storing time zone information, ArcticDB coerces all time zone information to `datetime.timezone`. This way we can ensure that all time zone information is stored in a consistent manner, and that all time zone information is preserved when data is written to and read from ArcticDB.
+
+When writing data with time zone information to ArcticDB, we preserve the time zone offset and name. When reading data with time zone information from ArcticDB, this data is used to create a `datetime.timezone` object.
+
+!!! note
+
+    This means that regardles of the timezone types being written in ArcticDB, all time zone types are normalised to `datetime.timezone`.
+    If you would like a another time-zone type back then tzname can be used to recreate it:
+    - For `pytz` you can use: `pytz.gettz(dt.tzname())`
+    - For `ZoneInfo`, after Python 3.9+, you can use: `ZoneInfo(dt.tzname())`
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -7,9 +7,9 @@ As of the Change Date specified in that file, in accordance with the Business So
 """
 import copy
 import datetime
-from datetime import timedelta
+from datetime import timedelta, timezone
 import math
-
+import pytz
 import numpy as np
 import os
 import sys
@@ -21,7 +21,12 @@ from pandas.api.types import is_integer_dtype
 from arcticc.pb2.descriptors_pb2 import UserDefinedMetadata, NormalizationMetadata, MsgPackSerialization
 from arcticc.pb2.storage_pb2 import VersionStoreConfig
 from collections import Counter
-from arcticdb.exceptions import ArcticNativeException, ArcticDbNotYetImplemented, NormalizationException, SortingException
+from arcticdb.exceptions import (
+    ArcticNativeException,
+    ArcticDbNotYetImplemented,
+    NormalizationException,
+    SortingException,
+)
 from arcticdb.supported_types import DateRangeInput, time_types as supported_time_types
 from arcticdb.util._versions import IS_PANDAS_TWO, IS_PANDAS_ZERO
 from arcticdb.version_store.read_result import ReadResult
@@ -911,8 +916,15 @@ class MsgPackNormalizer(Normalizer):
 
     def _custom_pack(self, obj):
         if isinstance(obj, pd.Timestamp):
-            tz = obj.tz.zone if obj.tz is not None else None
-            return ExtType(MsgPackSerialization.PD_TIMESTAMP, packb([obj.value, tz]))
+            offset = obj.tzinfo.utcoffset(obj).total_seconds() if obj.tzinfo else 0
+            if obj.tz and hasattr(obj.tz, "zone"):
+                tz = obj.tz.zone
+            else:
+                tz = obj.tzname()
+
+            return ExtType(
+                MsgPackSerialization.PD_TIMESTAMP, packb([obj.value, tz, offset])
+            )
 
         if isinstance(obj, datetime.datetime):
             return ExtType(MsgPackSerialization.PY_DATETIME, packb(_to_tz_timestamp(obj)))
@@ -928,7 +940,24 @@ class MsgPackNormalizer(Normalizer):
     def _ext_hook(self, code, data):
         if code == MsgPackSerialization.PD_TIMESTAMP:
             data = unpackb(data, raw=False)
-            return pd.Timestamp(data[0], tz=data[1]) if data[1] is not None else pd.Timestamp(data[0])
+            if len(data) == 2:
+                # We used to store only the value and the timezone as a string
+                # This is covering this legacy case, where we don't have the offset
+                val, tz = data
+                offset = 0
+                if tz:
+                    # We need to convert the string tz to a timedelta to denormalize to a datetime.timezone
+                    obj = pd.Timestamp(val, tz=tz)
+                    offset = obj.tzinfo.utcoffset(obj).total_seconds()
+            else:
+                # This covers the current case where we store the value, the timezone and the offset
+                # Should be changed if the _custom_pack method is changed
+                val, tz, offset = data[:3]
+
+            if tz is None:
+                return pd.Timestamp(val)
+
+            return pd.Timestamp(val, tz=timezone(timedelta(seconds=offset), tz))
 
         if code == MsgPackSerialization.PY_DATETIME:
             data = unpackb(data, raw=False)
@@ -1248,8 +1277,7 @@ def restrict_data_to_date_range_only(data: T, *, start: Timestamp, end: Timestam
         if not getattr(data, "timezone", None):
             start, end = _strip_tz(start, end)
         data = data[
-            start.to_pydatetime(warn=False)
-            - timedelta(microseconds=1) : end.to_pydatetime(warn=False)
+            start.to_pydatetime(warn=False) - timedelta(microseconds=1) : end.to_pydatetime(warn=False)
             + timedelta(microseconds=1)
         ]
     return data
--- a/python/tests/unit/arcticdb/version_store/test_normalization.py
+++ b/python/tests/unit/arcticdb/version_store/test_normalization.py
@@ -5,8 +5,10 @@ Use of this software is governed by the Business Source License 1.1 included in 
 
 As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
 """
+
 import datetime
 import sys
+from datetime import timezone, timedelta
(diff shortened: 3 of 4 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -1016,7 +1023,7 @@ def _nested_msgpack_unpackb(buff, raw=False):
 
     def _custom_pack(self, obj):
         if isinstance(obj, pd.Timestamp):
-            tz = obj.tz.zone if obj.tz is not None else None
+            tz = str(obj.tz) if obj.tz is not None else None
```
poodlewars, line 1026: It seems dangerous to rely on `str` for this, as most libraries don't give API guarantees for it.

How about:

```
if obj.tz is None:
  tz = None
elif obj.tz.zone is None:
   tz = obj.tzname()
else:
   tz = obj.tz.zone
```
?

I also think we have the same problem elsewhere in this file, eg:

```
def _to_tz_timestamp(dt):
    # type: (datetime.datetime)->(int, str)
    ts = pd.Timestamp(
        year=dt.year,
        month=dt.month,
        day=dt.day,
        hour=dt.hour,
        minute=dt.minute,
        second=dt.second,
        microsecond=dt.microsecond,
    ).value
    tz = dt.tzinfo.zone if dt.tzinfo is not None else None
    return ts, tz
```

poodlewars (changes requested):

python/arcticdb/version_store/_normalization.py
```diff
@@ -1016,7 +1023,13 @@ def _nested_msgpack_unpackb(buff, raw=False):
...
     def _custom_pack(self, obj):
         if isinstance(obj, pd.Timestamp):
-            tz = obj.tz.zone if obj.tz is not None else None
+            if obj.tz and hasattr(obj.tz, "zone"):
+                tz = obj.tz.zone
+            else:
+                tz = obj.tzname()
+                # Edge case: Pandas tzname() returns "Coordinated Universal Time" instead of "UTC" on Windows
+                # Which is breaking the pytz deserialization used by msgpack
+                tz = "UTC" if tz == "Coordinated Universal Time" else tz
```
poodlewars, line 1032: Doesn't this mean that writes with tz == "Coordinated Universal Time" won't roundtrip correctly?

jamesmunro: According to the docs, it looks like Windows is passing `Coordinated Universal Time` into the `timezone` constructor.
https://docs.python.org/3/library/datetime.html#datetime.timezone.tzname

Indeed the name is arbitrary, and probably coming from a system local timezone database?

What's the use case for changing this?  Why not roundtrip, so the user gets back what they put in?

jamesmunro: If I understanding what's happening here, then it looks like we've hit a snag because it's not correctly round-tripping at the moment, because we didn't keep the offset.  The other possibility is that your testing is across OS environments, and the timezone is not usable in the other environment. 

If it's the first problem then I think there is a chance to fix it.   The de-serializer is reading just the first two elements of the tuple.  
https://github.com/man-group/ArcticDB/blame/5b6df4fdc7289fcd4fd0cb9601e4ddfc6f46b979/python/arcticdb/version_store/_normalization.py#L1056C9-L1056C9

We could serialize the offset as a third element, and then construct the `timezone` object ourselves.  De-serialisers from old versions will not be effected.

G-D-Petrov: This is a very specific edge case, which I think is a bug in pandas.
The scenario is (du == dateutil):
1. we create a `tz = tzwin('UTC')` with `du.tz.gettz("UTC")`
2. This is passed to pandas to create a pd.Timestamp - `ts = pd.Timestamp("2024-03-04", tz=tz)`
3. `ts.tzname()` returns `Coordinated Universal Time` instead of `UTC` which we have asked for
  - this happens **only on Windows and only with du.tz.gettz** , in any other case, trying to set the time zone to `UTC`, would results to it being `UTC` (actually there is no other way to create a tz with `Coordinated Universal Time` as it is not supported by pytz)
4. We save the result from `ts.tzname()` during the normalization/serialization
5. When we try to read the metadata, we get back `Coordinated Universal Time` and we try to create a pd.Timestamp with it
6. This fails because `Coordinated Universal Time` is `not a valid time zone name`

Actually,  `Coordinated Universal Time` is not a valid time zone name for `du.tz.gettz` as well, passing **that string to it also results in a crash**.
So the 2 ways for us to support this edge case would be to:
1. pickle the time zone data - which could cause other issues
2. store the `UTC` time zone during normalization, then during the denormalization, try to create the timezone with `du.tz.gettz("UTC")` and pass it to the pd.Timestamp, but only on Windows and we would need to know that it was `dateutil` to begin with

In short, I don't think that it is worth investing the time to support this edge case behavior that looks like a bug in pandas in the first place.

jamesmunro: You're explaining what's happening with `tzname` and this case.  My point is that we shouldn't be trying to interpret the `tzname`, we should be trying to get the original `pd.Timestamp` object back.  Which means we should be capturing `offset` and `name` for `timezone` and using e.g. ` pd.Timestamp('2024-03-04', tzinfo=timezone(timedelta(hours=0), name='Coordinate Universal time'))`, that is what the original `pd.Timestamp` contained.  What did the original `pd.Timestamp` contain?

jamesmunro: As discussed separately.  I understand now that we have a range of `tzinfo` subtypes in `pd.Timestamp` and that we're not going to support some of them, particularly `tzfile`.  So we do need to normalise the types coming out.

python/arcticdb/version_store/_normalization.py
```diff
@@ -1040,7 +1054,12 @@ def _custom_pack(self, obj):
     def _ext_hook(self, code, data):
         if code == MsgPackSerialization.PD_TIMESTAMP:
             data = MsgPackNormalizer._nested_msgpack_unpackb(data)
-            return pd.Timestamp(data[0], tz=data[1]) if data[1] is not None else pd.Timestamp(data[0])
+            val, tz, offset = data
```
jamesmunro, line 1057: It may be wise to keep using the numeric indexing so that we can continue to extend if needed.

Also this needs to be able to read old data which only had two fields.   So we should add a test that covers the legacy case as well.
```
if len(data) == 2:
  val, tz = data
  # old method...
else:
  val, tz, offset = data[:3]
  # new methods...
```

docs/mkdocs/docs/faq.md
```diff
@@ -155,3 +155,9 @@ The handling of `NaN` in ArcticDB depends on the type of the column under consid
 * For string columns, `NaN`, as well as Python `None`, are fully supported.
 * For floating-point numeric columns, `NaN` is also fully supported.
 * For integer numeric columns `NaN` is not supported. A column that otherwise contains only integers will be treated as a floating point column if a `NaN` is encountered by ArcticDB, at which point [the usual rules](api/arctic.md#LibraryOptions) around type promotion for libraries configured with or without dynamic schema all apply as usual.
+
+### How does ArcticDB handle `time zone` information?
+
+ArcticDB takes a more strict approach to time zone handling than Pandas. While Pandas support multiple types for storing time zone information, ArcticDB coerces all time zone information to `datetime.timezone`. This way we can ensure that all time zone information is stored in a consistent manner, and that all time zone information is preserved when data is written to and read from ArcticDB.
+
+When writing data with time zone information to ArcticDB, we preserve the time zone offset and name. When reading data with time zone information from ArcticDB, this data is used to create a `datetime.timezone` object.
```
jamesmunro, line 163: I would add a bit extra to help people get what they need of they want the richer un-serializable types back.

> This does mean that if you have any number of `tzinfo` subtypes being stored in ArcticDB, then they are all normalised to `datetime.timezone`.  This is so that ArcticDB does not do any time-zone interpretation itself, but can guarantee the time instant is unchanged.
> If you would like a full time-zone type back then `tzname` is preserved. 
> You could use the `pytz` module:
> `pytz.gettz(dt.tzname())`
> or in Python 3.9+ you can use:
> `ZoneInfo(dt.tzname())`

python/arcticdb/version_store/_normalization.py
G-D-Petrov: Indeed I had forgotten to push these changes.
As for testing the legacy case, we can do it only with the compatibility tests and I have added the needed changes [here](https://github.com/man-group/ArcticDB/pull/1407/files#diff-4df7402f5282307e8aa7d22634403c683d6ce0c08840a40eba4c47f12665b008)

python/arcticdb/version_store/_normalization.py
```diff
@@ -1040,7 +1054,16 @@ def _custom_pack(self, obj):
     def _ext_hook(self, code, data):
         if code == MsgPackSerialization.PD_TIMESTAMP:
             data = MsgPackNormalizer._nested_msgpack_unpackb(data)
-            return pd.Timestamp(data[0], tz=data[1]) if data[1] is not None else pd.Timestamp(data[0])
+            if len(data) == 2:
```
poodlewars, line 1057: Well worth a comment here to explain the old 2-element data format and that this is for backwards compat, as the need for the 2 element case is impossible to understand from the code itself.

poodlewars (dismissed):

python/arcticdb/version_store/_normalization.py
```diff
@@ -1040,7 +1054,18 @@ def _custom_pack(self, obj):
...
+                # We used to store only the value and the timezone as a string
+                # This is covering this legacy case, where we don't have the offset
+                val, tz = data
+                offset = 0
```
jamesmunro, line 1061: This is wrong, no?
Why would the offset be zero?

jamesmunro (changes requested):

python/arcticdb/version_store/_normalization.py
jamesmunro: How does the compatibility test change result in a test that exercises the `if len(data)==2:` block?

python/arcticdb/version_store/_normalization.py
```diff
@@ -1040,7 +1054,18 @@ def _custom_pack(self, obj):
...
+            else:
+                val, tz, offset = data[:3]
```
jamesmunro, line 1063: I'd also comment on the reason for doing `data[:3]` here.

python/arcticdb/version_store/_normalization.py
```diff
@@ -1040,7 +1054,24 @@ def _custom_pack(self, obj):
...
-            return pd.Timestamp(data[0], tz=data[1]) if data[1] is not None else pd.Timestamp(data[0])
+            if len(data) == 2:
+                # We used to store only the value and the timezone as a string
+                # This is covering this legacy case, where we don't have the offset
+                val, tz = data
+                offset = 0
+                if tz:
+                    # We need to convert the string tz to a timedelta to denormalize to a datetime.timezone
+                    obj = pd.Timestamp(val, tz=tz)
+                    offset = obj.tzinfo.utcoffset(obj).total_seconds()
```
jamesmunro, line 1065: Tests for this case, both asserting zero and non-zero offset please.

jamesmunro (changes requested):

jamesmunro:
This PR does change the stored and returned timezone data.  We should likely hold off for a major version, just as with #1410
Is the original "Coordinated Universal Time" windows example issue urgent in any way?

PR #1409, opened 2024-03-08
Add testing API to recreate a TaskScheduler

poodlewars:
Useful for testing as allows pytests to change thread pool sizes.

Used in enterprise repo.


poodlewars:
> Why not create a test case which calls `reinit_task_scheduler`? Or is it already tested somewhere else?

Yeah I'll do that. It's tested, but in the enterprise repo. Definitely safer to have a test with it here too.

PR #1411, opened 2024-03-08
#447 Mongo Exceptions Normalization

muhammadhamzasajjad:
This normalizes mongo exceptions as in #584. The logic for exception normalization is now moved to  `mongo_storage.cpp`  as we also support a `MockMongoClient` which can mock failures. The Permission exceptions are also normalized. Previously there were two types of permission exceptions. One was triggered when a storage API returned a permission failure, the second was triggered if an unpermitted storage operation was attempted as per the library `OpenMode`. The two permission have now been normalized to inherit from the same class.

Furthermore, the exception hierarchy in python has also been changed. `DuplicateKeyException` inherits from `StorageException` in C++, this should be reflected on python side too. Why? because if someone wanted they could catch all the storage exceptions as follows instead of having to catch `PermissionException`, `DuplicateKeyException` and `StorageException` separately:

```python
try:
  # an operation that calls storage from python
except StorageException:
  # log storage error
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/python/python_module.cpp
```diff
@@ -236,6 +236,9 @@ void register_error_code_ecosystem(py::module& m, py::exception<arcticdb::Arctic
             internal_exception(msg.c_str());
         }
     });
+    // DuplicateKeyException and PermissionException are subclasses of StorageException so keep the hierarchy in Python too
```
poodlewars, line 239: drop the comment?

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -34,7 +34,7 @@ std::shared_ptr<LibraryIndex> create_library_index(const std::string &environmen
     return std::make_shared<LibraryIndex>(EnvironmentName{environment_name}, mem_resolver);
 }
 
-void register_bindings(py::module& storage, py::exception<arcticdb::ArcticException>& base_exception) {
+void register_bindings(py::module& storage, py::exception<arcticdb::ArcticException>&) {
```
poodlewars, line 37: Remove the unused argument completely?

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
poodlewars, line 1: It's really great working like this where you've written the tests with the old behaviour in an earlier PR. Makes it much easier to see that the changes are tested, and what behaviour changes you're introducing :+1:

poodlewars (changes requested):
This looks great! Just minor comments

cpp/arcticdb/python/python_module.cpp
muhammadhamzasajjad: Done in [59436ce](https://github.com/man-group/ArcticDB/pull/1411/commits/59436ceca3bb044cf4e6412c2eceaebf9080880f)

cpp/arcticdb/storage/python_bindings.cpp
muhammadhamzasajjad: Done in [59436ce](https://github.com/man-group/ArcticDB/pull/1411/commits/59436ceca3bb044cf4e6412c2eceaebf9080880f)

poodlewars (approved):
Looks good! Please squash before you merge this in

Issue #1412, opened 2024-03-11
maint: Folly Replacement Plan


jjerphan:
## Motivation

Folly is admittedly:
 - [API and ABI unstable in between commits and not designed to be packaged as a shared library](https://github.com/facebook/folly?tab=readme-ov-file#build-notes)
 - not supporting Windows, and _untested_ on this platform
 - depending on a lot of other libraries

This makes ArcticDB hardly portable, packageable on many platforms under a shared library on conda-forge.

<details>

<summary>Test disablement</summary>

[Tests are disabled for Windows](https://github.com/facebook/folly/blob/f67559532c274fee8a84ab39d5711b422dc963a7/CMakeLists.txt#L889) for elements ArcticDB uses including but not limited to:
- `F14Set`
- `ConcurrentHashMap`
- `ThreadPoolExecutors`
- `Future`
- `ThreadName`
- `FBString`
- `Hash`
- `ThreadCachedInt`
- `ThreadLocal`

</details>

<details>
    <summary>folly's dependency graph</summary>

```
mamba repoquery depends -t -c conda-forge folly
```
    
```
folly[2023.10.30.00]
  ├─ jemalloc[4.4.0]
  ├─ libboost-headers[1.82.0]
  ├─ libgcc-ng[13.2.0]
  │  ├─ _openmp_mutex[4.5]
  │  │  ├─ _libgcc_mutex[0.1]
  │  │  └─ llvm-openmp[17.0.4]
  │  │     ├─ libzlib[1.2.13]
  │  │     └─ zstd[1.5.5]
  │  │        ├─ libzlib already visited
  │  │        └─ libstdcxx-ng[13.2.0]
  │  └─ _libgcc_mutex already visited
  ├─ libzlib already visited
  ├─ zstd already visited
  ├─ libstdcxx-ng already visited
  ├─ bzip2[1.0.8]
  │  └─ libgcc-ng already visited
  ├─ gflags[2.2.2]
  │  ├─ libgcc-ng already visited
  │  └─ libstdcxx-ng already visited
  ├─ lz4-c[1.9.4]
  │  ├─ libgcc-ng already visited
  │  └─ libstdcxx-ng already visited
  ├─ glog[0.6.0]
  │  ├─ libgcc-ng already visited
  │  ├─ libstdcxx-ng already visited
  │  └─ gflags already visited
  ├─ fmt[9.1.0]
  │  ├─ libgcc-ng already visited
  │  └─ libstdcxx-ng already visited
[...]

PR #1417, opened 2024-03-13
maint: Replace use of folly getCurrentThreadId with STL

ianthomas23:

This is work to assist @jjerphan in removing use of folly in the ArcticDB code base.

The change here is to remove use of `folly::getCurrentThreadID()` in `get_thread_id()`, replacing it with use of `std::this_thread::get_id()`.

This replaces #1408 and is submitted from a branch of the main repo rather than my fork so that CI runs automatically.

#### Any other comments?

The folly implementation is much more complicated as there are different implementation for the 3 major OSes:

https://github.com/facebook/folly/blob/69a1b93df285457cd0799a6898e2e5390a09cd73/folly/system/ThreadId.cpp#L29-L37

The different implementations are not required as the STL C++11 solution should work on all OSes. Following
 https://github.com/man-group/ArcticDB/pull/1408#discussion_r1521142399 the solution here returns a `std::thread::id` rather than a `uint64_t` and this is formatted correctly in log messages (the only use of thread ID) by including `<fmt/std.h>`.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/util/storage_lock.hpp
+++ b/cpp/arcticdb/util/storage_lock.hpp
@@ -14,9 +14,9 @@
 #include <arcticdb/util/exponential_backoff.hpp>
 #include <arcticdb/util/configs_map.hpp>
 
-#include <folly/system/ThreadId.h>
-
+#include <fmt/std.h>
 #include <mutex>
+#include <thread>
 
 namespace arcticdb {
 
@@ -66,8 +66,8 @@ struct StorageLockTimeout : public std::runtime_error {
     using std::runtime_error::runtime_error;
 };
 
-inline uint64_t get_thread_id() {
-    return folly::getCurrentThreadID();
+inline std::thread::id get_thread_id() noexcept {
+    return std::this_thread::get_id();
 }
 
 template <class ClockType = util::SysClock>


alexowens90 (approved):

ianthomas23:
Some CI builds are passing and some failing. The failures are 
```
error: invalid use of incomplete type ‘struct std::variant_size<long unsigned int>’
     struct variant_size<volatile _Variant> : variant_size<_Variant> {};
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I can reproduce this locally using a docker build whereas a conda build works fine (on Ubuntu 13.10).

I've tracked this down to fmtlib/fmt#3068 which was fixed in fmtlib/fmt#3072, dated 2022-09-03. The first release of `fmt` to include this fix is `10.0.0`. I note that
https://github.com/man-group/ArcticDB/blob/baf9eb99ee164c35f7063ff398a8535c7565854b/environment_unix.yml#L42-L43
indicates we need to stick with `fmt < 10`.

Apparently `gcc >= 11` doesn't need the `fmt` fix which is presumably why building using `conda` works OK (I have gcc 12.3.0 this way) whereas in the docker image I have gcc 10.2.1.

I suppose the easiest way to avoid these problems is to remove the `#include <fmt/std.h>` and instead locally implement a formatter for `std::thread::id`.

jjerphan:
We might be able to use fmt 10.

jjerphan (approved):
LGTM given :green_circle: CI checks.

Thank you for helping with #1412, @ianthomas23.

PR #1429, opened 2024-03-15
Implement empty index for 0-rowed columns

vasil-pashov:
#### Reference Issues/PRs
Closes: #1428 

Create an empty-index type. This required change in the Python and in the C++ layer.
* In the C++ layer an new index type was added. (IndexDescriptor::EMPTY). It does not allocate a filed in the storage (similar to how row range index does not allocate a field). The checks for index compatibility are relaxed, the empty index is compatible with all other index types and it gets overridden the first time a non-empty index is written (either through update or append). On write we check if the dataframe contains 0 rows and if so it gets assigned an empty index.
* The logic in the python layer is dodgy and needs discussion. In the current state the normalization metadata and the index descriptor are stored separately. There is one proto message describing both DateTime index and Ranged Index. The current change made it so that in case of 0 rows the python layer passes RowRange index to the C++ layer which checks if there are any rows in the DF. If there are rows Row range index is used, otherwise empty index is used.  Note the `is_not_range_index` proto field. IMO it needs some refactoring in further PRs. It's used in the python layer to check if the first column is index or not.
#### Any other comments?

Merge this after: https://github.com/man-group/ArcticDB/pull/1436

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/entity/types_proto.cpp
vasil-pashov, line 1: Moved from types_proto.hpp

cpp/arcticdb/stream/index.hpp
```diff
@@ -296,10 +293,47 @@ class RowCountIndex : public BaseIndex<RowCountIndex> {
...
+
+    static constexpr const char* name() {
+        return "empty";
+    }
+
+    static constexpr EmptyIndex default_index() {
+        return {};
+    }
+
+    [[nodiscard]] static IndexValue start_value_for_segment(const SegmentInMemory& segment) {
```
willdealtry, line 315: Why the cast to a timestamp?

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -220,21 +220,15 @@ std::shared_ptr<InputTensorFrame> py_ndf_to_frame(
                 "Number idx names {} and values {} do not match",
                 idx_names.size(), idx_vals.size());
 
-    if (idx_names.empty()) {
-        res->index = stream::RowCountIndex();
-        res->desc.set_index_type(IndexDescriptor::ROWCOUNT);
-    } else {
+    if (idx_names.empty() == false) {
```
willdealtry, line 223: `!index_names.empty()`?

cpp/arcticdb/version/schema_checks.hpp
```diff
@@ -29,42 +29,56 @@ struct StreamDescriptorMismatch : ArcticSpecificException<ErrorCode::E_DESCRIPTO
...
+            index_type_to_str(old_idx_kind)
+        );
     }
 }
 
-inline bool columns_match(const StreamDescriptor &left, const StreamDescriptor &right) {
-    if (left.fields().size() != right.fields().size())
+inline bool columns_match(const StreamDescriptor& left, const StreamDescriptor& right) {
+    const int left_index_is_empty = left.index().type() == IndexDescriptor::EMPTY;
+    const int right_fields_offset = right.index().field_count() * left_index_is_empty;
```
alexowens90, line 68: Multiplying by a bool implicitly cast to an int is hard to read. Prefer
`const int right_fields_offset = left.index().type() == IndexDescriptor::EMPTY ? right.index().field_count() : 0;`

cpp/arcticdb/version/schema_checks.hpp
```diff
@@ -29,42 +29,56 @@ struct StreamDescriptorMismatch : ArcticSpecificException<ErrorCode::E_DESCRIPTO
...
+            "Cannot append {} index to {} index",
+            index_type_to_str(new_idx_kind),
+            index_type_to_str(old_idx_kind)
+        );
     }
 }
 
-inline bool columns_match(const StreamDescriptor &left, const StreamDescriptor &right) {
-    if (left.fields().size() != right.fields().size())
+inline bool columns_match(const StreamDescriptor& left, const StreamDescriptor& right) {
```
alexowens90, line 66: It seems like this method should be symmetrical, but it does not handle the left non-empty/right empty case correctly

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -15,17 +15,18 @@
 class DtypeGenerator:
     """
     Class which can generate all dtypes which ArcticDB supports. It can generate them by type category e.g. int, float,
-    etc. It can also generate a list of all available dtypes
+    etc. It can also generate a list of all available dtypes. Not all supported scalar dtypes are generated as this
+    leads to a combinatorc explosion in test case count.
```
alexowens90, line 19: This description is misleading now we don't generate all types, I would say it just generates a representative subset of our supported types sufficient for testing empty column types

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -74,7 +75,7 @@ def empty_index(request):
     yield request.param
 
 
-class TestCanAppendToColumnWithNones:
+class TestCanAppendToColunWithNones:
```
alexowens90, line 78: typo

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -498,11 +499,11 @@ def test_float(self, lmdb_version_store_static_and_dynamic, float_dtype):
         )
         lmdb_version_store_static_and_dynamic.update(
             "sym",
-            pd.DataFrame({"col": [None, np.nan]}, index=self.update_index(), dtype=float_dtype)
+            pd.DataFrame({"col": [None, None]}, index=self.update_index())
```
alexowens90, line 502: Why?

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -498,11 +499,11 @@ def test_float(self, lmdb_version_store_static_and_dynamic, float_dtype):
...
         )
         assert_frame_equal(
             lmdb_version_store_static_and_dynamic.read("sym").data,
-            pd.DataFrame({"col": [1, np.nan, np.nan, 4]}, index=self.index(), dtype=float_dtype)
+            pd.DataFrame({"col": [1, float("NaN"), float("NaN"), 4]}, index=self.index(), dtype=float_dtype)
```
alexowens90, line 506: Why?

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -784,10 +789,31 @@ def test_date(self, lmdb_version_store_static_and_dynamic, date_dtype):
...
+    @pytest.mark.parametrize(
+            "incompatible_indexes",
+            [
+                (pd.RangeIndex(0,3), list(pd.date_range(start="1/1/2024", end="1/3/2024"))),
+                (list(pd.date_range(start="1/1/2024", end="1/3/2024")), pd.RangeIndex(0, 3))
+            ]
+    )
+    def test_cannot_append_different_index_type_after_first_non_empty(self, lmdb_version_store_static_and_dynamic, incompatible_indexes):
+        lmdb_version_store_static_and_dynamic.write("sym", pd.DataFrame({"col": []}))
+        lmdb_version_store_static_and_dynamic.append("sym", pd.DataFrame({"col": [1,2,3]}, index=incompatible_indexes[0]))
```
alexowens90, line 817: We should read the data back here and have some assertions about the `dtype` of `col` and the type of the index

alexowens90 (changes requested):

cpp/arcticdb/stream/index.hpp
vasil-pashov: The IndexValue::NumericType is typedefed to timestamp. I'll changing this to IndexValue::NumericType

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -16,15 +16,15 @@
 #include <pybind11/numpy.h>
 
 namespace arcticdb::convert {
-const char none_char[8] = {'\300', '\000', '\000', '\000', '\000', '\000', '\000', '\000'};
+constexpr const char none_char[8] = {'\300', '\000', '\000', '\000', '\000', '\000', '\000', '\000'};
 
 using namespace arcticdb::pipelines;
 
-bool is_unicode(PyObject *obj) {
+[[nodiscard]] static inline bool is_unicode(PyObject *obj) {
```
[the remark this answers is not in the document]
vasil-pashov: Code hygiene. This is used only in thus TU so static will exclude it from the linking phase, it's a one-liner thus it makes sense to be inlined and we shouldn't throw the result.

cpp/arcticdb/version/schema_checks.hpp
vasil-pashov: It often makes difference in performance critical parts. I agree this is not the case here.

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
vasil-pashov: Artifact from testing. Changed to `np.nan, float("NaN")` so that we cover both.

cpp/arcticdb/version/schema_checks.hpp
vasil-pashov: The right is not supposed to have empty index since modifying operations, except write, with 0 rows don't reach the C++ layer. Should I make it symmetric?

cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -57,6 +57,8 @@ message TypeDescriptor {
 message IndexDescriptor {
     enum Type {
         UNKNOWN = 0;
+        // Used then the dataframe has 0 rows. Convertible to any other type
+        EMPTY = 69; // 'E'
```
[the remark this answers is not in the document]
vasil-pashov: It worked:

```
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> import numpy as np
>>> from arcticdb import Arctic
>>> import arcticdb
>>> arcticdb.__version__
'dev'
>>> ac = Arctic("lmdb://test")
>>> lib = ac.get_library("test", create_if_missing=True)
>>> lib.write("rowrange", pd.DataFrame({"col": [1, 2, 3]})
... )
VersionedItem(symbol='rowrange', library='test', data=n/a, version=0, metadata=None, host='LMDB(path=C:\\Users\\vasil\\Documents\\ArcticDB\\python\\test)', timestamp=1711724920728818900)
>>> lib.write("datetime", pd.DataFrame({"col": [1, 2, 3]}, index=list(pd.date_range(start="1/1/2024", end="1/3/2024"))))
VersionedItem(symbol='datetime', library='test', data=n/a, version=0, metadata=None, host='LMDB(path=C:\\Users\\vasil\\Documents\\ArcticDB\\python\\test)', timestamp=1711725775655538800)


-----------------------------------------------------------------------------------

>>> import pandas as pd
>>> import numpy as np
>>> import arcticdb
>>> arcticdb.__version__
'4.2.1'
>>> from arcticdb import Arctic
>>> ac = Arctic("lmdb://python/test")
>>> lib = ac.get_library("test", create_if_missing=True)
>>> lib.read("rowrange").data
   col
0    1
1    2
2    3
>>> lib.read("datetime").data
            col
2024-01-01    1
2024-01-02    2
2024-01-03    3
```

cpp/arcticdb/version/schema_checks.hpp
alexowens90: It's fine for it to not be symmetric, I would just rename the parameters to make it more obvious they can't be swapped

alexowens90 (changes requested):

alexowens90 (approved):

PR #1437, opened 2024-03-18
Bug Fix Windows: Remove lmdb files when `delete_library` is called

### LINKED ISSUE #1266: delete_library does not delete the data file in the folder
### Describe the bug

Hi guys, I tried to create a new library then delete it. I find that even though the library seems to be dropped as if I run arctic.list_libraries() shows nothing in the list after the deletiong; However, the 'sample' folder which contains the data and lock file is still in the database folder. I doubt it is because the 'sample' folder is not available for deletion when I run delete_library, but I am not sure. Are there any other steps that need to be done before the library could be deleted? Thanks very much!

### Steps/Code to Reproduce

The code is simple as the following
arctic = adb.Arctic("lmdb://D:/arcticDB")
lib = arctic.get_library('sample', create_if_missing=True)
arctic.delete_library('sample')

### Expected Results

No error is shown but the files are still there.

Python: 3.8.18 (default, Sep 11 2023, 13:39:12) [MSC v.1916 64 bit (AMD64)]
OS: Windows-10-10.0.19045-SP0
ArcticDB: 4.2.1

### Backend storage used

LMDB

### Additional Context

_No response_

muhammadhamzasajjad:
#### Reference Issues/PRs
Fixes #1266. The cause of this error was that `LmdbStorage` member `env_` keeps a handle open to the files and windows doesn't delete files if they have been opened by another process. The previous fix was attempting to do `del lib` in python and assuming that this would automatically call the `LmdbStorage` destructor and remove the `env_` handle. `del lib` doesn't guarantee immediate deletion of the library. Therefore on windows delete_library doesn't work as other lmdb has opened the files.

 The fix moves the file removal logic in `LmdbStorage` by adding a new function `close`. The lmdb env handle is closed and this ensures no process has opened the lmdb files, so they are safe to delete.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/library_manager.cpp
+++ b/cpp/arcticdb/storage/library_manager.cpp
@@ -156,10 +156,10 @@ std::shared_ptr<Library> LibraryManager::get_library(const LibraryPath& path, co
     return lib;
 }
 
-void LibraryManager::close_library_if_open(const LibraryPath &path) {
+void LibraryManager::cleanup_library_if_open(const LibraryPath &path) {
     std::lock_guard<std::mutex> lock{open_libraries_mutex_};
     if (auto it = open_libraries_.find(path); it != open_libraries_.end()) {
-        it->second->close();
+        it->second->cleanup();
         open_libraries_.erase(it);
     }
 }
--- a/cpp/arcticdb/storage/library_manager.cpp
+++ b/cpp/arcticdb/storage/library_manager.cpp
@@ -158,7 +158,10 @@ std::shared_ptr<Library> LibraryManager::get_library(const LibraryPath& path, co
 
 void LibraryManager::close_library_if_open(const LibraryPath &path) {
     std::lock_guard<std::mutex> lock{open_libraries_mutex_};
-    open_libraries_.erase(path);
+    if (auto it = open_libraries_.find(path); it != open_libraries_.end()) {
+        it->second->close();
+        open_libraries_.erase(it);
+    }
 }
 
 std::vector<LibraryPath> LibraryManager::get_library_paths() const {
--- a/cpp/arcticdb/storage/lmdb/lmdb_storage.hpp
+++ b/cpp/arcticdb/storage/lmdb/lmdb_storage.hpp
@@ -46,7 +46,7 @@ class LmdbStorage final : public Storage {
 
     inline bool do_fast_delete() final;
 
-    void close() override;
+    void cleanup() override;
 
     void do_iterate_type(KeyType key_type, const IterateTypeVisitor& visitor, const std::string &prefix) final;
 
@@ -54,7 +54,8 @@ class LmdbStorage final : public Storage {
 
     ::lmdb::env& env() {
         if (!env_) {
-            raise<ErrorCode::E_UNEXPECTED_LMDB_ERROR>("Unexpected LMDB Error: Invalid operation: LMDB environment has been removed");
+            raise<ErrorCode::E_UNEXPECTED_LMDB_ERROR>("Unexpected LMDB Error: Invalid operation: LMDB environment has been removed. "
+                                                      "Possibly because the library has been deleted");
         }
         return *env_;
     }
--- a/cpp/arcticdb/storage/lmdb/lmdb_storage.hpp
+++ b/cpp/arcticdb/storage/lmdb/lmdb_storage.hpp
@@ -46,6 +46,8 @@ class LmdbStorage final : public Storage {
 
     inline bool do_fast_delete() final;
 
+    void close() override;
+
     void do_iterate_type(KeyType key_type, const IterateTypeVisitor& visitor, const std::string &prefix) final;
 
     bool do_key_exists(const VariantKey & key) final;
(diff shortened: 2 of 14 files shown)


cpp/arcticdb/storage/lmdb/lmdb_storage.hpp
```diff
@@ -46,11 +46,18 @@ class LmdbStorage final : public Storage {
...
+    void close() override;
+
     void do_iterate_type(KeyType key_type, const IterateTypeVisitor& visitor, const std::string &prefix) final;
 
     bool do_key_exists(const VariantKey & key) final;
 
-    ::lmdb::env& env() { return *env_;  }
+    ::lmdb::env& env() {
+        if (!env_) {
+            raise<ErrorCode::E_UNEXPECTED_LMDB_ERROR>("Unexpected LMDB Error: Invalid operation: LMDB environment has been removed");
```
poodlewars, line 57: Worth explaining that the root cause might be that the library has been deleted?

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -45,25 +45,30 @@ class LMDBStorageFactory : public StorageFactory {
...
     uint64_t map_size;
     bool use_mock;
+    fs::path db_path;
+    std::string lib_name;
 public:
-    explicit LMDBStorageFactory(bool use_mock = false) : map_size(128ULL * (1ULL << 20) /* 128MB */), use_mock(use_mock) { }
+    explicit LMDBStorageFactory(uint64_t map_size, bool use_mock = false) : map_size(map_size), use_mock(use_mock), db_path(TEST_DATABASES_PATH / "test_lmdb"), lib_name("test_lib") { }
 
-    explicit LMDBStorageFactory(uint64_t map_size, bool use_mock = false) : map_size(map_size), use_mock(use_mock) { }
+    explicit LMDBStorageFactory(bool use_mock = false) : LMDBStorageFactory(128ULL * (1ULL << 20), use_mock) { }
```
poodlewars, line 53: Would be good to keep the 128MB comment

python/arcticdb/adapters/lmdb_library_adapter.py
```diff
@@ -168,24 +168,7 @@ def get_library_config(self, name, library_options: LibraryOptions):
         )
 
     def cleanup_library(self, library_name: str):
```
poodlewars, line 170: Can just remove this now - this is how it's defined on the base `arctic_library_adapter`. In fact, I think we can drop the `cleanup_library` method completely now

cpp/arcticdb/storage/library_manager.cpp
```diff
@@ -158,7 +158,10 @@ std::shared_ptr<Library> LibraryManager::get_library(const LibraryPath& path, co
 
 void LibraryManager::close_library_if_open(const LibraryPath &path) {
     std::lock_guard<std::mutex> lock{open_libraries_mutex_};
-    open_libraries_.erase(path);
+    if (auto it = open_libraries_.find(path); it != open_libraries_.end()) {
+        it->second->close();
```
poodlewars, line 162: The naming is a bit weird, you have `close()` and `close_library_if_open` but these are both actually deleting files. Think a bit of renaming would be good.

poodlewars:
Have you done a bit of manual testing? Can we still create an lmdb library, closed the python interpreter and have that library still readable? Want to be sure we aren't deleting anything inadvertently. Our CI tests probably wouldn't catch it as they all deliberately delete the LMDB database at the end anyway!

cpp/arcticdb/storage/library_manager.cpp
muhammadhamzasajjad: I think `cleanup_library_if_open` might be a better name. Similarly, in storage it would be called `cleanup()` rather than `close()`. I am not calling it `delete_library` or similar for storages because it doesn't do anything for all the other storages except lmdb

cpp/arcticdb/storage/test/test_storage_exceptions.cpp
```diff
@@ -45,25 +45,30 @@ class LMDBStorageFactory : public StorageFactory {
 private:
     uint64_t map_size;
     bool use_mock;
+    fs::path db_path;
+    std::string lib_name;
 public:
-    explicit LMDBStorageFactory(bool use_mock = false) : map_size(128ULL * (1ULL << 20) /* 128MB */), use_mock(use_mock) { }
+    explicit LMDBStorageFactory(uint64_t map_size, bool use_mock = false) : map_size(map_size), use_mock(use_mock), db_path(TEST_DATABASES_PATH / "test_lmdb"), lib_name("test_lib") { }
```
poodlewars, line 51: I think the defaults here "test_lib" and TEST_DATABASES_PATH / "test_lmdb" are dangerous, remove them?

poodlewars (approved):

muhammadhamzasajjad:
>Have you done a bit of manual testing? Can we still create an lmdb library, closed the python interpreter and have that library still readable?
Yes, I have tested that and lmdb library isn't unexpectedly deleted if we close the interpreter.

PR #1438, opened 2024-03-19
Fix reading segments with no rows

alexowens90:
Fixes https://github.com/man-group/arcticdb-man/issues/102

### DIFF
--- a/cpp/arcticdb/column_store/chunked_buffer.hpp
+++ b/cpp/arcticdb/column_store/chunked_buffer.hpp
@@ -271,7 +271,11 @@ class ChunkedBufferImpl {
     }
 
     [[nodiscard]] const uint8_t* data() const {
-        util::check(blocks_.size() == 1, "Taking a pointer to the beginning of a non-contiguous buffer");
+        if (blocks_.empty()) {
+            return nullptr;
+        }
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(blocks_.size() == 1,
+                                                        "Taking a pointer to the beginning of a non-contiguous buffer");
         blocks_[0]->magic_.check();
         return blocks_[0]->data();
     }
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -271,7 +271,11 @@ class ChunkedBufferImpl {
     }
 
     [[nodiscard]] const uint8_t* data() const {
-        util::check(blocks_.size() == 1, "Taking a pointer to the beginning of a non-contiguous buffer");
+        if (blocks_.empty()) {
+            return nullptr;
+        }
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(blocks_.size() == 1,
```
vasil-pashov, line 277: Do we still need this?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -1947,6 +1947,13 @@ def test_get_index(object_version_store):
     assert idx.iloc[0]["version_id"] == 0
 
 
+def test_read_empty_index(lmdb_version_store_v1):
```
vasil-pashov, line 1950: Test for v2 encoding?

DrNickClarke (approved):

cpp/arcticdb/column_store/chunked_buffer.hpp
alexowens90: Yes, if there are multiple blocks this will not work as expected

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
alexowens90: This change sits far above the encoding layer, not everything needs to be tested with both

PR #1443, opened 2024-03-20
Fix #1384 by refactoring `VersionMap` to include all combinations of `LoadType`s and whether to `include_deleted`

### LINKED ISSUE #1384: Poor performance when reading as_of a date with many early versions deleted
### Describe the bug

There is a feature called _tombstone all_ that is supposed to prevent version search having to walk the entire historical version list when the early versions have all been deleted.

It works for as_of = a version number (it will return quickly not having found the version)

However when as_of = a date is used it can be slow. This is much more apparent using AWS where the latency is higher.

When there are thousands of versions the read can take several minutes.

### Steps/Code to Reproduce

import arcticdb as adb
import pandas as pd
import numpy as np
from datetime import datetime, timedelta

arctic = adb.Arctic(<AWS S3 uri>)
lib = arctic.get_library('adb_bugs', create_if_missing=True)

N = 3
df = pd.DataFrame(
    index=pd.date_range("20240101", periods=N),
    data={'col': np.arange(0., N)}
)

#### write 500 versions
sym1 = 'asof_slow_read'
for i in range(500):
    lib.write(sym1, df)
	
#### remove early versions
lib.delete(sym1)

#### add one more version
lib.write(sym1, df)

#### this is slow (12s in my test)
as_of = datetime.now() - timedelta(days=1)
lib.read(sym1, as_of=as_of)

#### this is fast (171ms in my test)
lib.read(sym1, as_of=499)

### Expected Results

Results are as expected. This is a performance issue. 

Python 3.10
Linux Linux version 5.15.133.1-microsoft-standard-WSL2
arcticdb 4.3.1

### Backend storage used

AWS S3

### Additional Context

This is possibly related to this issue (failure to observe tombstone correctly). It may be easier to solve the two together

https://github.com/man-group/ArcticDB/issues/1385

IvoDD:
#### Reference Issues/PRs
Fixes #1384 

The #1384 bug also applies to `as_of=<VERSION_ID>` it is just not seen in the repro because it is not looking for an old enough version.

The bug is that we perform the check on `TOMBSTONE_ALL` only for `LOAD_UNDELETED` and `LOAD_LATEST_UNDELETED`. However `LOAD_DOWNTO` and `LOAD_FROM_TIME` should also only work on undeleted versions when querying with `as_of`.

As discussed we would like to have the ability to both load deleted and undeleted with `LOAD_DOWNTO` and `LOAD_FROM_TIME`. This essentially leaves us with the full cartesian product of possible load types and whether to include deleted or not.

This the majority of this change is to do the refactor to extract the `include_deleted` out of the load type and ensure correctness.

The PR is split in several commits, each one of which should be reviewed separately. The commits have detailed descriptions but in a few words the commits do the following:
- Introduce a failing cpp test to demonstrate #1384 
- Moves entry updates when tombstoning inside `write_tombstone`
- Revamp version map caching logic to simplify the checks and to make less unnecessary cache misses
- Fix cache invalidation when tombstoning
- Refactor the `LoadType` into a `LoadStrategy` which encapsulates the `LoadType`, `include_deleted` and the timestamp or version for `LOAD_DOWNTO` and `LOAD_FROM_TIME`. This fixes the #1384 bug and provides new functionality 
- Small refactor with renames to make the logic inside `follow_version_chain` clearer.
- Decrease the default `unsync_tolerance` because previously we were ~always skipping the cache and deeming it outdated. Also fixes a bug where we don't clear the cache when clearing the storage.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/test/test_stream_version_data.cpp
+++ b/cpp/arcticdb/version/test/test_stream_version_data.cpp
@@ -12,8 +12,9 @@ TEST(StreamVersionData, SpecificVersion) {
     VersionQuery query_2{SpecificVersionQuery{VersionId(4)}, false};
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_DOWNTO);
-    ASSERT_EQ(stream_version_data.load_param_.load_until_version_, 4);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_DOWNTO);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_, 4);
 }
 
 TEST(StreamVersionData, SpecificVersionReversed) {
@@ -24,8 +25,9 @@ TEST(StreamVersionData, SpecificVersionReversed) {
     VersionQuery query_2{SpecificVersionQuery{VersionId(12)}, false};
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_DOWNTO);
-    ASSERT_EQ(stream_version_data.load_param_.load_until_version_, 4);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_DOWNTO);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_, 4);
 }
 
 TEST(StreamVersionData, Timestamp) {
@@ -38,8 +40,9 @@ TEST(StreamVersionData, Timestamp) {
     VersionQuery query_2{TimestampVersionQuery{timestamp(4)}, false};
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_FROM_TIME);
-    ASSERT_EQ(stream_version_data.load_param_.load_from_time_, 4);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_FROM_TIME);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_, 4);
 }
 
 TEST(StreamVersionData, TimestampUnordered) {
@@ -54,8 +57,9 @@ TEST(StreamVersionData, TimestampUnordered) {
     VersionQuery query_3{TimestampVersionQuery{timestamp(4)}, false};
     stream_version_data.react(query_3);
     ASSERT_EQ(stream_version_data.count_, 3);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_FROM_TIME);
-    ASSERT_EQ(stream_version_data.load_param_.load_from_time_, 3);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_FROM_TIME);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_, 3);
 }
 
 TEST(StreamVersionData, Latest) {
@@ -66,8 +70,9 @@ TEST(StreamVersionData, Latest) {
     VersionQuery query_1{std::monostate{}, false};
     stream_version_data.react(query_1);
     ASSERT_EQ(stream_version_data.count_, 1);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_LATEST_UNDELETED);
-    ASSERT_EQ(stream_version_data.load_param_.load_until_version_.has_value(), false);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_LATEST);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_.has_value(), false);
 }
 
 TEST(StreamVersionData, SpecificToTimestamp) {
@@ -80,9 +85,10 @@ TEST(StreamVersionData, SpecificToTimestamp) {
     VersionQuery query_2{TimestampVersionQuery{timestamp(3)}, false};
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_UNDELETED);
-    ASSERT_EQ(stream_version_data.load_param_.load_until_version_.has_value(), false);
-    ASSERT_EQ(stream_version_data.load_param_.load_from_time_.has_value(), false);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_ALL);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_.has_value(), false);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_.has_value(), false);
 }
 
 TEST(StreamVersionData, TimestampToSpecific) {
@@ -95,7 +101,8 @@ TEST(StreamVersionData, TimestampToSpecific) {
     VersionQuery query_2{SpecificVersionQuery{VersionId(12)}, false};
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_UNDELETED);
-    ASSERT_EQ(stream_version_data.load_param_.load_until_version_.has_value(), false);
-    ASSERT_EQ(stream_version_data.load_param_.load_from_time_.has_value(), false);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_ALL);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_.has_value(), false);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_.has_value(), false);
 }
--- a/cpp/arcticdb/version/test/test_stream_version_data.cpp
+++ b/cpp/arcticdb/version/test/test_stream_version_data.cpp
@@ -13,7 +13,7 @@ TEST(StreamVersionData, SpecificVersion) {
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_DOWNTO);
-    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_objective_, LoadObjective::UNDELETED);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_, 4);
 }
 
@@ -26,7 +26,7 @@ TEST(StreamVersionData, SpecificVersionReversed) {
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_DOWNTO);
-    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_objective_, LoadObjective::UNDELETED);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_, 4);
 }
 
@@ -41,7 +41,7 @@ TEST(StreamVersionData, Timestamp) {
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_FROM_TIME);
-    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_objective_, LoadObjective::UNDELETED);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_, 4);
 }
 
@@ -58,7 +58,7 @@ TEST(StreamVersionData, TimestampUnordered) {
     stream_version_data.react(query_3);
     ASSERT_EQ(stream_version_data.count_, 3);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_FROM_TIME);
-    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_objective_, LoadObjective::UNDELETED);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_, 3);
 }
 
@@ -71,7 +71,7 @@ TEST(StreamVersionData, Latest) {
     stream_version_data.react(query_1);
     ASSERT_EQ(stream_version_data.count_, 1);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_LATEST);
-    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_objective_, LoadObjective::UNDELETED);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_.has_value(), false);
 }
 
@@ -86,7 +86,7 @@ TEST(StreamVersionData, SpecificToTimestamp) {
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_ALL);
-    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_objective_, LoadObjective::UNDELETED);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_.has_value(), false);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_.has_value(), false);
 }
@@ -102,7 +102,7 @@ TEST(StreamVersionData, TimestampToSpecific) {
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_type_, LoadType::LOAD_ALL);
-    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.to_load_, ToLoad::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_objective_, LoadObjective::UNDELETED);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_until_version_.has_value(), false);
     ASSERT_EQ(stream_version_data.load_param_.load_strategy_.load_from_time_.has_value(), false);
 }
--- a/cpp/arcticdb/version/test/test_stream_version_data.cpp
+++ b/cpp/arcticdb/version/test/test_stream_version_data.cpp
@@ -12,8 +12,8 @@ TEST(StreamVersionData, SpecificVersion) {
     VersionQuery query_2{SpecificVersionQuery{VersionId(4)}};
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
-    ASSERT_EQ(stream_version_data.load_strategy_.load_type_, LoadType::LOAD_DOWNTO);
-    ASSERT_EQ(stream_version_data.load_strategy_.load_objective_, LoadObjective::UNDELETED);
+    ASSERT_EQ(stream_version_data.load_strategy_.load_type_, LoadType::DOWNTO);
+    ASSERT_EQ(stream_version_data.load_strategy_.load_objective_, LoadObjective::UNDELETED_ONLY);
     ASSERT_EQ(stream_version_data.load_strategy_.load_until_version_, 4);
 }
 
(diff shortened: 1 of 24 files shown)


cpp/arcticdb/version/version_functions.hpp
```diff
@@ -78,7 +78,7 @@ inline std::optional<AtomKey> get_specific_version(
         SignedVersionId signed_version_id,
         const pipelines::VersionQuery& version_query,
         bool include_deleted = false) {
-    LoadParameter load_param{LoadType::LOAD_DOWNTO, signed_version_id};
+    LoadParameter load_param{LoadType::LOAD_DOWNTO, true, signed_version_id};
```
IvoDD, line 81: TODO: Use include_deleted=false (This is where the as_of argument gets created)

cpp/arcticdb/version/version_functions.hpp
```diff
@@ -240,7 +240,7 @@ inline std::optional<AtomKey> load_index_key_from_time(
     const StreamId &stream_id,
     timestamp from_time,
     const pipelines::VersionQuery& version_query) {
-    LoadParameter load_param{LoadType::LOAD_FROM_TIME, from_time};
+    LoadParameter load_param{LoadType::LOAD_FROM_TIME, true, from_time};
```
IvoDD, line 243: TODO: Think whether we don't need false instead.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -556,13 +555,14 @@ class VersionMapImpl {
      *
      * @param stream_id symbol to check
      * @param load_param the load type
-     * @return whether we have a cached entry suitable for the load type, so do not need to go to storage
+     * @return whether we have a cached entry suitable for the load strategy, so do not need to go to storage
      */
-    bool has_cached_entry(const StreamId &stream_id, const LoadParameter& load_param) const {
-        LoadType requested_load_type = load_param.load_type_;
+    bool has_cached_entry(const StreamId &stream_id, const LoadStrategy& requested_load_strategy) const {
+        // TODO: This might not be needed
```
IvoDD, line 561: Remove this

cpp/arcticdb/version/version_map.hpp
```diff
@@ -580,64 +580,37 @@ class VersionMapImpl {
...
+                return cached_timestamp <= requested_load_strategy.load_from_time_.value();
+            }
             case LoadType::LOAD_ALL:
-                return true;
+                // We can have all versions cached only when cache was loaded by loading all versions.
+                // Also, we either need the cached entry to include delete or to not request for deleted versions.
+                if (cached_load_type==LoadType::LOAD_ALL){
+                    return entry->load_strategy_.include_deleted_ || !requested_load_strategy.include_deleted_;
+                }
+                break;
```
IvoDD, line 613: Better to just return false here.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -247,7 +247,7 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
 
     submit_tasks_for_range(wrapper, [store, version_map](auto sym_version) {
                 auto first_version = *std::min_element(std::begin(sym_version.second), std::end(sym_version.second));
-                LoadParameter load_param{LoadType::LOAD_DOWNTO, static_cast<SignedVersionId>(first_version)};
+                LoadParameter load_param{LoadType::LOAD_DOWNTO, true, static_cast<SignedVersionId>(first_version)};
```
IvoDD, line 250: Here as well

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -46,17 +44,17 @@ struct VersionDetails {
...
 };
 
-inline constexpr bool is_latest_load_type(LoadType load_type) {
-    return load_type == LoadType::LOAD_LATEST || load_type == LoadType::LOAD_LATEST_UNDELETED;
-}
-
-struct LoadParameter {
-    explicit LoadParameter(LoadType load_type) :
-        load_type_(load_type) {
+// The strategy used for loading entries from the version chain.
```
IvoDD, line 47: Rewrite this comment

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -83,6 +81,63 @@ struct LoadParameter {
     }
 };
 
+
+inline bool is_undeleted_strategy_subset(const LoadStrategy& left, const LoadStrategy& right){
+    switch (left.load_type_) {
+        case LoadType::NOT_LOADED:
+            return true;
+        case LoadType::LOAD_LATEST:
+            // We LOAD_LATEST is not a subset of LOAD_DOWNTO because LOAD_DOWNTO may not reach the latest undeleted version.
```
IvoDD, line 90: Remove "We"

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -155,18 +216,19 @@ struct VersionMapEntry {
...
+        // Sorting by creation_ts is safe from clock skew because we don't support parallel writes to the same symbol.
         std::sort(std::begin(keys_), std::end(keys_), [](const AtomKey &l, const AtomKey &r) {
             return l.creation_ts() > r.creation_ts();
         });
     }
 
     void clear() {
         head_.reset();
-        load_type_ = LoadType::NOT_LOADED;
         last_reload_time_ = 0;
```
IvoDD, line 227: Also clear the strategy.

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -277,28 +270,33 @@ static constexpr timestamp nanos_to_seconds(timestamp nanos) {
...
+
+    if (loaded_deleted_or_undeleted_timestamp > *load_strategy.load_from_time_)
         return false;
 
     ARCTICDB_DEBUG(log::version(),
                    "Exiting load from timestamp because request {} <= {}",
-                   *load_params.load_from_time_,
-                   load_progress.earliest_loaded_timestamp_);
+                   *load_strategy.load_from_time_,
+                   loaded_deleted_or_undeleted_timestamp);
```
IvoDD: Swap these timestamps in logging.

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -277,28 +270,33 @@ static constexpr timestamp nanos_to_seconds(timestamp nanos) {
...
 }
 
-inline bool load_latest_ongoing(const LoadParameter &load_params, const std::shared_ptr<VersionMapEntry> &entry) {
-    if (!(load_params.load_type_ == LoadType::LOAD_LATEST_UNDELETED && entry->get_first_index(false).first) &&
-        !(load_params.load_type_ == LoadType::LOAD_LATEST && entry->get_first_index(true).first))
+inline bool load_latest_ongoing(const LoadStrategy &load_strategy, const std::shared_ptr<VersionMapEntry> &entry) {
+    if (!(load_strategy.load_type_ == LoadType::LOAD_LATEST && entry->get_first_index(load_strategy.include_deleted_).first))
         return true;
 
     ARCTICDB_DEBUG(log::version(), "Exiting because we found a non-deleted index in load latest");
```
IvoDD: Move doc update from later commit into this commit.

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -83,6 +80,63 @@ struct LoadParameter {
...
+            return true;
+        case LoadType::LOAD_LATEST:
+            // LOAD_LATEST is not a subset of LOAD_DOWNTO because LOAD_DOWNTO may not reach the latest undeleted version.
+            return right.load_type_ != LoadType::NOT_LOADED && right.load_type_ != LoadType::LOAD_DOWNTO;
+        case LoadType::LOAD_DOWNTO:
+            if (right.load_type_ == LoadType::LOAD_ALL) {
+                return true;
+            }
+            if (right.load_type_ == LoadType::LOAD_DOWNTO && ((left.load_until_version_.value()>=0) == (right.load_until_version_.value()>=0))) {
+                // Left is subset of right only when the [load_until]s have same sign and left's version is >= right's version
```
alexowens90, line 96: I don't think this is right when they're both negative

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -83,6 +80,63 @@ struct LoadParameter {
...
+            util::raise_rte("Invalid load type: {}", left.load_type_);
+    }
+    return false;
+}
+
+// Returns a strategy which is guaranteed to load all versions requested by left and right.
+// Works only on strategies with include_deleted=false.
+inline LoadStrategy union_of_undeleted_strategies(const LoadStrategy& left, const LoadStrategy& right){
+    util::check(!left.include_deleted_, "Trying to produce a union of undeleted strategies but left strategy includes deleted.");
+    util::check(!right.include_deleted_, "Trying to produce a union of undeleted strategies but right strategy includes deleted.");
```
alexowens90, line 120: nit: Use `internal::check<ErrorCode::E_ASSERTION_ERROR>`

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -83,6 +80,63 @@ struct LoadParameter {
...
+    return LoadStrategy{LoadType::LOAD_ALL, false};
+}
+
+struct LoadParameter {
+    LoadParameter(LoadType load_type, bool include_deleted) : load_strategy_(load_type, include_deleted) {}
+    LoadParameter(LoadType load_type, bool include_deleted, int64_t load_from_time_or_until) :
+        load_strategy_(load_type, include_deleted, load_from_time_or_until) {}
+
+    LoadStrategy load_strategy_;
+    bool iterate_on_failure_ = false;
```
alexowens90, line 137: Do we need a separate struct just to add this bool? And is `iterate_on_faiure` ever set to true?

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -83,6 +80,63 @@ struct LoadParameter {
...
+    util::check(!left.include_deleted_, "Trying to produce a union of undeleted strategies but left strategy includes deleted.");
+    util::check(!right.include_deleted_, "Trying to produce a union of undeleted strategies but right strategy includes deleted.");
+    if (is_undeleted_strategy_subset(left, right)){
+        return right;
+    }
+    if (is_undeleted_strategy_subset(right, left)){
+        return left;
+    }
+    // If none is subset of the other then we should load all versions;
+    return LoadStrategy{LoadType::LOAD_ALL, false};
```
alexowens90, line 128: I'm wary of performance regressions with this, was the behaviour as conservative before?

cpp/arcticdb/version/version_map.hpp
```diff
@@ -580,68 +579,40 @@ class VersionMapImpl {
...
+                return loaded_as_far_as_version_id(*entry, requested_load_strategy.load_until_version_.value());
+            case LoadType::LOAD_FROM_TIME: {
+                // We check whether the cached (deleted or undeleted) timestamp is before or at the requested one
+                auto cached_timestamp = requested_load_strategy.include_deleted_ ?
+                                        entry->loaded_with_progress_.earliest_loaded_timestamp_ :
+                                        entry->loaded_with_progress_.earliest_loaded_undeleted_timestamp_;
+                return cached_timestamp <= requested_load_strategy.load_from_time_.value();
+            }
+            case LoadType::LOAD_ALL:
+                // We can have all versions cached only when cache was loaded by loading all versions.
```
alexowens90, line 607: I don't think this is strictly true, as you can load all of the information in the version chain without the load parameter having been `LOAD_ALL`
e.g. loading down to a timestamp before version 0 was written
Probably not worth worrying about here as it is just a cache, but worth correcting the comment.

alexowens90 (changes requested):

poodlewars (commented):
I've focused on the tests in this review as AlexO has done the rest.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -107,7 +107,7 @@ class VersionMapImpl {
      */
     using MapType =  std::map<StreamId, std::shared_ptr<VersionMapEntry>>;
 
-    static constexpr uint64_t DEFAULT_CLOCK_UNSYNC_TOLERANCE = ONE_SECOND * 2;
+    static constexpr uint64_t DEFAULT_CLOCK_UNSYNC_TOLERANCE = ONE_MILLISECOND * 200;
```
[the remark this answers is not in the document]
IvoDD: This is from the last commit. Pasting description below:
```
Before this change version_map's reload_interval was equal to the
unsync_tolerance. This means that we would ~always miss the cache.

The check for outdated cache is as follows:
- When reading into cache set `last_reload_time = now -
  unsync_tolerance`
- When checking if cache is outdated we check whether `now -
  last_reload_time > reload_interval

This change updates the unsync tolerance to 200ms. This will mean the
cache is valid for 1.8s

Tests uncovered that we were not flushing the version_map when clearing
the storage. This change fixes this as well.
```

cpp/arcticdb/version/version_map_entry.hpp
IvoDD: I think it should be correct. E.g. `-2 >= -3` and `-2` will load only 2 versions whereas the `-3` will load 3 versions which means `-2` will be a subset of `-3`.

cpp/arcticdb/version/version_map_entry.hpp
IvoDD: Thanks, done

cpp/arcticdb/version/version_map_entry.hpp
IvoDD: The `iterate_on_failure` is used [here](https://github.com/man-group/ArcticDB/blob/a2aa6864e5eb65a779363aee2f924e0860a21576/cpp/arcticdb/version/version_functions.hpp#L17) to use the `VersionQuery.iterate_on_failure`. 
That option of the version query seems like it could be useful though I'm not sure whether its used currently.

If you think it's not useful I'm happy to remove it as it will clean up code nicely.

cpp/arcticdb/version/version_map_entry.hpp
IvoDD: Updated the comment.

I think we cannot do better than this because the `LoadType`s are such that when none is subset of the other there isn't really a better thing to do than to LOAD_ALL.

Also yes, that's what the code was doing before. Code responsible for this checks used to be [here](https://github.com/man-group/ArcticDB/blob/a2aa6864e5eb65a779363aee2f924e0860a21576/cpp/arcticdb/version/version_map_batch_methods.cpp#L22-L80).

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -614,336 +645,76 @@ TEST(VersionMap, FollowingVersionChainEndEarly) {
         version_map->follow_version_chain(store, ref_entry, follow_result, load_params);
         // When loading with any of the specified load params we should end following the version chain early
         // at version 1 because that's when we encounter the TOMBSTONE_ALL.
-        EXPECT_EQ(follow_result->loaded_until_, VersionId{1});
+        EXPECT_EQ(follow_result->loaded_with_progress_.oldest_loaded_index_version_, VersionId{1});
     }
 }
 
-TEST(VersionMap, CacheInvalidationLoadDownTo) {
```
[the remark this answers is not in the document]
IvoDD: I kept this test and renamed it to `CacheInvalidationWithTombstoneAfterLoad`. I added some more tests there for the case where we have two consecutive undeleted versions.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -614,336 +645,76 @@ TEST(VersionMap, FollowingVersionChainEndEarly) {
...
-
-    auto entry = version_map->check_reload(
-            store,
-            id,
-            LoadParameter{LoadType::LOAD_DOWNTO, static_cast<SignedVersionId>(1)},
-            __FUNCTION__);
-
-    ASSERT_TRUE(version_map->has_cached_entry(id, LoadParameter{LoadType::LOAD_LATEST_UNDELETED}));
-
-    // When - we delete version 1 and reload
```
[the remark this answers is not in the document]
IvoDD: I kept this test and added more tests after the modification in the new upgraded version of this test: `CacheInvalidationWithTombstoneAfterLoad`.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -614,336 +645,76 @@ TEST(VersionMap, FollowingVersionChainEndEarly) {
...
+
+    auto check_all_caching = [&](const std::vector<LoadParameter>& to_load, const std::vector<LoadParameter>& to_check_if_cached, bool expected_result){
+        for (auto to_load_param : to_load) {
+            for (auto to_check_if_cached_param : to_check_if_cached){
+                check_caching(to_load_param, to_check_if_cached_param, expected_result);
+            }
+        }
+    };
 
-    ASSERT_TRUE(version_map->has_cached_entry(id, LoadParameter{LoadType::LOAD_LATEST_UNDELETED}));
```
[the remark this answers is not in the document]
IvoDD: I think this assertion should be covered by [this](https://github.com/man-group/ArcticDB/blob/1c63312614b8598711d868f47943a182bbe220df/cpp/arcticdb/version/test/test_version_map.cpp#L754) `LoadStrategy` which compares the LOAD_LATEST_UNDELETED to a bunch of other checks.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -614,336 +645,76 @@ TEST(VersionMap, FollowingVersionChainEndEarly) {
...
-    ASSERT_TRUE(version_map->has_cached_entry(id, LoadParameter{LoadType::LOAD_DOWNTO, static_cast<SignedVersionId>(1)}));
-
-    // Should not rely on the cache upon (eg) a LOAD_ALL here hence pick up an extra INDEX
-    auto reloaded = version_map->check_reload(
-            store,
-            id,
-            LoadParameter{LoadType::LOAD_ALL},
-            __FUNCTION__);
-
-    ASSERT_EQ(reloaded->get_indexes(true).size(), 3);
```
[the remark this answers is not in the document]
IvoDD: Added some assertions like this in this [here](https://github.com/man-group/ArcticDB/blob/1c63312614b8598711d868f47943a182bbe220df/cpp/arcticdb/version/test/test_version_map.cpp#L646C9-L648C82) which cover a lot more cases.

cpp/arcticdb/version/version_map.hpp
IvoDD: Agreed, updated the comment.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -622,7 +622,42 @@ void write_alternating_deleted_undeleted(std::shared_ptr<InMemoryStore> store, s
...
+    auto check_strategy_loads_to = [&](LoadStrategy load_strategy, VersionId should_load_to){
+        auto ref_entry = VersionMapEntry{};
+        read_symbol_ref(store, id, ref_entry);
+        auto follow_result = std::make_shared<VersionMapEntry>();
+
+        version_map->follow_version_chain(store, ref_entry, follow_result, load_strategy);
+        EXPECT_EQ(follow_result->loaded_with_progress_.oldest_loaded_index_version_, VersionId{should_load_to});
+    };
+
+    check_strategy_loads_to(LoadStrategy{LoadType::LOAD_DOWNTO, true, static_cast<SignedVersionId>(0)}, 0);
```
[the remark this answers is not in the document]
IvoDD: Nice suggestion. I did the suggested refactor and it's quite clearer now.

cpp/arcticdb/version/version_map.hpp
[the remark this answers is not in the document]
poodlewars: I don't understand the motivation for the "clock unsync tolerance" - it's being applied to times from this process' clock and only being held in memory. How can we have clock skew within a single process? Any memories @willdealtry?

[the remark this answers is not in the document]
poodlewars: AFAICT this unsync tolerance is somehow a tolerance WRT this process' own clock, which theoretically can jump around a little bit, but I don't think reducing the tolerance to 200ms is very worrying.

cpp/arcticdb/version/version_map_entry.hpp
poodlewars: I think this implementation is correct since we want these semantics:

left v2 right v3 => left not subset of right

left v3 right v2 => left subset of right

left v-1 right v-2 => left subset of right (left is HEAD, right is HEAD~)

left v-2 right v-1  => left not subset of right

poodlewars (changes requested):
Happy with this other than the remark about the tombstone being added twice

cpp/arcticdb/version/version_functions.hpp
```diff
@@ -202,13 +201,12 @@ inline version_store::TombstoneVersionResult tombstone_version(
                             stream_id, version_id);
             }
             // We will write a tombstone key even when the index_key is not found
-            tombstone = version_map->write_tombstone(store, version_id, stream_id, entry, creation_ts);
+            version_map->write_tombstone(store, version_id, stream_id, entry, creation_ts);
```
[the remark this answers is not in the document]
poodlewars: As discussed - the function that needs changing is actually in the enterprise repo, which I had checked out while doing this review. So this is fine - we just need to tweak the enterprise repo after this change.

poodlewars (approved):
Approved but please wait for AlexO and William to have a look too before merging

cpp/arcticdb/version/version_functions.hpp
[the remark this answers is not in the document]
IvoDD: Draft pull request to do this: https://github.com/man-group/arcticdb-enterprise/pull/86

cpp/arcticdb/version/version_map.hpp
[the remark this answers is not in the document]
willdealtry: Agreed

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -35,6 +33,12 @@ inline constexpr bool is_partial_load_type(LoadType load_type) {
     return load_type == LoadType::LOAD_DOWNTO || load_type == LoadType::LOAD_FROM_TIME;
 }
 
+// Used to specify whether we want to load all or only undeleted versions
+enum class ToLoad : uint32_t {
```
willdealtry, line 37: Slightly odd name. Maybe something like LoadObjective or LoadRequirement?

cpp/arcticdb/version/version_map_entry.hpp
willdealtry: Let's get rid of it. iterate_on_failure is a historical relic from when we first introduced the version ref keys, I think it's long past retirement age

cpp/arcticdb/version/version_map_entry.hpp
IvoDD: Thanks! `LoadObjective` is way better.

cpp/arcticdb/version/version_map_entry.hpp
IvoDD: Cool, removed it entirely. Made things quite cleaner.

cpp/arcticdb/version/version_map_entry.hpp
alexowens90: Agreed, ignore me

alexowens90 (changes requested):

python/arcticdb/version_store/_store.py
```diff
@@ -1968,7 +1967,6 @@ def list_versions(
         symbol: Optional[str] = None,
         snapshot: Optional[str] = None,
         latest_only: Optional[bool] = False,
-        iterate_on_failure: Optional[bool] = False,
```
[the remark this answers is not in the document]
IvoDD: Re-added and added a log that it's deprecated.

alexowens90 (approved):

cpp/arcticdb/version/test/test_stream_version_data.cpp
```diff
@@ -7,95 +7,102 @@ TEST(StreamVersionData, SpecificVersion) {
...
-    VersionQuery query_1{SpecificVersionQuery{VersionId(12)}, false};
+    VersionQuery query_1{SpecificVersionQuery{VersionId(12)}};
     stream_version_data.react(query_1);
-    VersionQuery query_2{SpecificVersionQuery{VersionId(4)}, false};
+    VersionQuery query_2{SpecificVersionQuery{VersionId(4)}};
     stream_version_data.react(query_2);
     ASSERT_EQ(stream_version_data.count_, 2);
-    ASSERT_EQ(stream_version_data.load_param_.load_type_, LoadType::LOAD_DOWNTO);
-    ASSERT_EQ(stream_version_data.load_param_.load_until_version_, 4);
+    ASSERT_EQ(stream_version_data.load_strategy_.load_type_, LoadType::LOAD_DOWNTO);
```
willdealtry, line 15: Should we lose the extra LOAD? as in LoadType::LATEST not LoadType::LOAD_LATEST? It's a bit smurf naming at the moment (not your fault) https://devcards.io/smurf-naming-convention

cpp/arcticdb/version/test/version_backwards_compat.hpp
```diff
@@ -49,7 +49,7 @@ std::vector<AtomKey> backwards_compat_write_and_prune_previous(std::shared_ptr<S
     log::version().debug("Version map pruning previous versions for stream {}", key.id());
 
     std::vector<AtomKey> output;
-    auto entry = version_map->check_reload(store, key.id(), LoadParameter{LoadType::LOAD_ALL},  __FUNCTION__);
+    auto entry = version_map->check_reload(store, key.id(), LoadStrategy{LoadType::LOAD_ALL, LoadObjective::ANY}, __FUNCTION__);
```
willdealtry, line 52: I find a LoadObjective of ANY a bit confusing. Any what?

cpp/arcticdb/version/version_map.hpp
```diff
@@ -702,27 +674,27 @@ class VersionMapImpl {
...
      * @param entry the version map state to check
      * @param load_param the load request to test for completeness
      * @return true if and only if entry already contains data at least as far back as load_param requests
      */
-    bool loaded_as_far_as_load_until(const VersionMapEntry& entry, const LoadParameter& load_param) const {
-        if (is_positive_version_query(load_param)) {
-            if (entry.loaded_until_ <= static_cast<VersionId>(load_param.load_until_version_.value())) {
+    bool loaded_as_far_as_version_id(const VersionMapEntry& entry, SignedVersionId requested_version_id) const {
+        if (requested_version_id >= 0) {
+            if (entry.loaded_with_progress_.oldest_loaded_index_version_ <= static_cast<VersionId>(requested_version_id)) {
```
willdealtry, line 685: entry.load_progress_?

willdealtry (approved):

PR #1446, opened 2024-03-20
maint: Remove use of folly/system/ThreadName.h

ianthomas23:
#### Reference Issues/PRs

Fixes one item of the folly replacement plan, issue #1412.

This removes use of `folly/system/ThreadName.h`. It was used in two places to obtain a `std::string` name of a thread, and here is replaced with conversion of unique thread ID to a string using `fmt` via
```c++
fmt::format("{}", arcticdb::get_thread_id())
```

#### Any other comments?

This is used in two places, the first in the termination handler in `python_module.cpp` which I have manually tested. The second use is in the [Remotery](https://github.com/Celtoys/Remotery) which, as far as I can tell, is not tested in CI so I am not sure of the status of the Remotery support.

### DIFF
--- a/cpp/arcticdb/entity/performance_tracing.cpp
+++ b/cpp/arcticdb/entity/performance_tracing.cpp
@@ -8,9 +8,9 @@
 
 #include <arcticdb/entity/performance_tracing.hpp>
 #include <arcticdb/log/log.hpp>
-#include <folly/system/ThreadName.h>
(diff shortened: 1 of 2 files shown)


jjerphan (approved):
LGTM. Thank you, @ianthomas23.

I am not sure whether Remotery should be kept: is it still used?

PR #1467, opened 2024-03-28
Bugfix/1250/compact incompletes rationalisation

### LINKED ISSUE #1250: `compact_incomplete` should work or raise useful error messages in all scenarios with dynamic schema
- Be restrictive with static schema to all columns being present and types not changeing
- Types changing, both between the existing symbol and the incompletes, and between the incompletes themselves
- Columns missing from all incomplete segments
- Columns missing from some incomplete segments
- Respect the lib config segment row size
Note there is a test called `CompactIncompleteDynamicSchema` in `test_version_store.cpp` that passes, and an `xfail` test in `test_parallel.py` that does not.

alexowens90:
Closes #1250 

Added `xfail` tests and opened #1466 for outstanding issues that require larger code changes

### DIFF
--- a/cpp/arcticdb/column_store/column.cpp
+++ b/cpp/arcticdb/column_store/column.cpp
@@ -540,6 +540,7 @@ void Column::change_type(DataType target_type) {
             });
         });
     }
+    buf.commit();
     type_ = TypeDescriptor{target_type, type_.dimension()};
     std::swap(data_, buf);
 }
--- a/cpp/arcticdb/column_store/column.cpp
+++ b/cpp/arcticdb/column_store/column.cpp
@@ -366,23 +366,25 @@ void Column::mark_absent_rows(size_t num_rows) {
 }
 
 void Column::default_initialize_rows(size_t start_pos, size_t num_rows, bool ensure_alloc) {
-    type_.visit_tag([that=this, start_pos, num_rows, ensure_alloc](auto tag) {
-        using T= std::decay_t<decltype(tag)>;
-        using RawType = typename T::DataTypeTag::raw_type;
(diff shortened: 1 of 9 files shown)


cpp/arcticdb/column_store/column.cpp
```diff
@@ -366,23 +366,25 @@ void Column::mark_absent_rows(size_t num_rows) {
...
+    if (num_rows > 0) {
+        type_.visit_tag([that = this, start_pos, num_rows, ensure_alloc](auto tag) {
+            using T = std::decay_t<decltype(tag)>;
+            using RawType = typename T::DataTypeTag::raw_type;
+            const auto bytes = (num_rows * sizeof(RawType));
 
-        if(ensure_alloc)
-            that->data_.ensure<uint8_t>(bytes);
+            if (ensure_alloc)
+                that->data_.ensure<uint8_t>(bytes);
```
willdealtry, line 377: Could get rid of that->, but as you didn't do it we can always catch it on a subsequent PR. Maybe fix it if you have to re-push anyway

willdealtry (approved):

cpp/arcticdb/stream/segment_aggregator.hpp
```diff
@@ -100,6 +109,7 @@ template<class Index, class Schema, class SegmentingPolicy = RowCountSegmentPoli
     std::vector<SegmentInMemory> segments_;
     std::vector<pipelines::FrameSlice> slices_;
     SliceCallBack slice_callback_;
+    std::optional<StreamDescriptor> stream_descriptor_;
```
willdealtry, line 112: How is the descriptor on the aggregator different to the one on the schema?

PR #1473, opened 2024-04-02
Fix SYMBOL_LIST Segment index out of bounds bug

IvoDD:
The check `row < row_count()` in `string_at` was problematic for symbol list entries because they have the exception that it is possible for a column to have more rows than segment rows.

That exception is because of the nature of the symbol list refactor in 4.2.0 which needed to remain backwards compatible with old readers.

This change bypasses the check for symbol list entries and directly
calls string_at or scalar_at of the Column.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/symbol_list.cpp
+++ b/cpp/arcticdb/version/symbol_list.cpp
@@ -128,17 +128,35 @@ MaybeCompaction last_compaction(const std::vector<AtomKey>& keys) {
     }
 }
 
+// The below string_at and scalar_at functions should be used for symbol list cache segments instead of the ones
+// provided in SegmentInMemory, because the symbol list structure is the only place where columns can have more entries
+// than the segment has rows. Hence, we need to bypass the checks inside SegmentInMemory's function and directly call the
+// Column's string_at and scalar_at.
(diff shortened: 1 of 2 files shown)


willdealtry (approved):

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -128,17 +128,32 @@ MaybeCompaction last_compaction(const std::vector<AtomKey>& keys) {
     }
 }
 
+// The below string_at and scalar_at functions should be used for symbol list cache segments instead of the ones
+// provided in SegmentInMemory, because the symbol list structure is the only place where columns can have more entries
+// than the segment has rows. Hence, we need to bypass the checks inside SegmentInMemory's function and directly call the
+// Column's string_at and scalar_at.
+std::string string_at(const SegmentInMemory& seg, position_t row, position_t col){
+    auto offset = seg.column(col).scalar_at<position_t >(row).value();
```
alexowens90, line 136: Can we add an `internal::check` that the optional contains a value? (below too)
Much easier to debug than a bad optional access

alexowens90 (approved):

PR #1475, opened 2024-04-05
Feature flag empty_index

vasil-pashov:
#### Reference Issues/PRs
Relates to: #1429, #1440

Feature flag the use of empty type. It uses the same library config option as the empty column type. In order to preserve behavior empty dataframes will use DataTimeIndex.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/schema_checks.hpp
+++ b/cpp/arcticdb/version/schema_checks.hpp
@@ -67,7 +67,7 @@ inline void check_normalization_index_match(
             );
         } else {
             // (old_idx_kind == IndexDescriptor::TIMESTAMP && new_idx_kind == IndexDescriptor::ROWCOUNT) is left to preserve
-            // pre-empty index behavior with pandas 2, see test_empty_writes.cpp::test_append_empty_series. Empty pd.Series
+            // pre-empty index behavior with pandas 2, see test_empty_writes.py::test_append_empty_series. Empty pd.Series
             // have Rowrange index, but due to: https://github.com/man-group/ArcticDB/blob/bd1776291fe402d8b18af9fea865324ebd7705f1/python/arcticdb/version_store/_normalization.py#L545
             // it gets converted to DatetimeIndex (all empty indexes except categorical and multiindex are converted to datetime index
             // in pandas 2 if empty index type is disabled), however we still want to be able to append pd.Series to empty pd.Series.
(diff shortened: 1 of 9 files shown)


python/tests/unit/arcticdb/version_store/test_empty_column_type.py
alexowens90, line 1: Don't we need to test the behaviour of the empty index both with and without the feature flag enabled? I don't see where that's happening right now

python/arcticdb/version_store/_store.py
```diff
@@ -332,6 +332,9 @@ def _try_normalize(
         dynamic_schema = self.resolve_defaults(
             "dynamic_schema", self._lib_cfg.lib_desc.version.write_options, False, **kwargs
         )
+        empty_types = self.resolve_defaults(
```
alexowens90, line 335: This isn't quite right, as the C++ layer only checks the lib config, not the kwargs. I don't think this needs to be a kwarg, so sufficient to pass an empty dict here

cpp/arcticdb/version/schema_checks.hpp
```diff
@@ -42,29 +42,57 @@ inline IndexDescriptor::Type get_common_index_type(const IndexDescriptor::Type&
...
 }
 
-inline bool columns_match(const StreamDescriptor& df_in_store_descriptor, const StreamDescriptor& new_df_descriptor) {
-    const int right_fields_offset = df_in_store_descriptor.index().type() == IndexDescriptor::EMPTY ? new_df_descriptor.index().field_count() : 0;
+inline bool columns_match(
+    const StreamDescriptor& df_in_store_descriptor,
+    const StreamDescriptor& new_df_descriptor,
+    bool empty_types
+) {
+    const int right_fields_offset = (empty_types && df_in_store_descriptor.index().type() == IndexDescriptor::EMPTY)
```
alexowens90, line 93: Why is `empty_types` needed in this check? If the index type of the in store descriptor is empty, then won't `right_fields_offset` always be `new_df_descriptor.index().field_count()`?

cpp/arcticdb/version/schema_checks.hpp
```diff
@@ -42,29 +42,57 @@ inline IndexDescriptor::Type get_common_index_type(const IndexDescriptor::Type&
...
+        if (empty_types) {
+            normalization::check<ErrorCode::E_INCOMPATIBLE_INDEX>(
+                common_index_type != IndexDescriptor::UNKNOWN,
+                "Cannot append {} index to {} index",
+                index_type_to_str(new_idx_kind),
+                index_type_to_str(old_idx_kind)
+            );
+        } else {
+            // (old_idx_kind == IndexDescriptor::TIMESTAMP && new_idx_kind == IndexDescriptor::ROWCOUNT) is left to preserve
+            // pre-empty index behavior with pandas 2, see test_empty_writes.cpp::test_append_empty_series. Empty pd.Series
```
alexowens90, line 70: typo, `test_empty_writes.py`

alexowens90 (changes requested):

python/arcticdb/version_store/_store.py
willdealtry: Agreed, we don't want anyone to be able to mix and match within a library, that would be bad news

alexowens90 (approved):

PR #1496, opened 2024-04-12
hot-fix: Use cmake 3.29.2 with `jwlawson/actions-setup-cmake@v2`

jjerphan:
#### Reference Issues/PRs
See: https://github.com/actions/runner-images/issues/9680#issuecomment-2051917949i

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #1497, opened 2024-04-15
maint: Revert #1496

jjerphan:
#### Reference Issues/PRs
This reverts commit 74f993285492f25875297aa9c09ebf6ea60d3f90 after [the publication of a new image containing CMake 3.29.2](https://github.com/actions/runner-images/releases/tag/win22/20240414.1).

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -73,13 +73,6 @@ jobs:
         with:
           submodules: recursive # Just in case a dep has its own third-party deps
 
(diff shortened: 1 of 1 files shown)


poodlewars (approved):
I still think that we should specify and pin a specific CMake across all platforms, but OK with merging this.

jjerphan:
> I still think that we should specify and pin a specific CMake across all platforms, but OK with merging this.

I agree, yet I do not know if this is possible without duplicating the installation of CMake.

PR #1501, opened 2024-04-17
Missing Empty tests: added ArcticDB read simulator

DrNickClarke:
This change modifies the test targets for the missing_empty tests.

Nones are replaced by:

0 for int
False for bool

To simulate the effect of the ArcticDB read.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_missing_empty.py
+++ b/python/tests/unit/arcticdb/version_store/test_missing_empty.py
@@ -125,6 +125,39 @@ def pd_delete_replace(df1, df2, date_range=None):
     # concat preserves types over other methods eg int -> float (due to temp NaN creation)
     return pd_restore_secondary_index_levels(pd.concat(to_concat), df1.index.names)
 
+
+def infer_type(s: pd.Series):
+    t = None
+    for v in s:
+        if not pd.isnull(v):
+            tv = type(v)
+            if t and tv != t:
+                raise ValueError(f"Mixed non-None types in column {s.name}: {t} and {tv}")
+            t = tv
+    return t
+
+
+def fill_value(t):
+    fill_values = {
+        bool: False,
+        int: 0
+    }
+    if t in fill_values.keys():
+        return fill_values[t]
+    return None
+
+
+def pd_arcticdb_read_sim(df):
+    read_df = df.copy(True)
+    for c in df.columns:
+        t = infer_type(df[c])
+        print(c, t, fill_value(t))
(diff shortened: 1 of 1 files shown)


python/tests/unit/arcticdb/version_store/test_missing_empty.py
```diff
@@ -125,6 +125,39 @@ def pd_delete_replace(df1, df2, date_range=None):
...
+
+
+def pd_arcticdb_read_sim(df):
+    read_df = df.copy(True)
+    for c in df.columns:
+        t = infer_type(df[c])
+        print(c, t, fill_value(t))
+        if t and fill_value(t) is not None:
+            fv = fill_value(t)
+            read_df[c] = df[c].fillna(fv)
```
muhammadhamzasajjad, line 157: nit: here it is enough to call `fill_value`  only once as follows
```python
fv = fill_value(t)
if fv: # if t is None fv will also be None 
    read_df[c] = df[c].fillna(fv)
```
In this case `fill_value` is not expensive but still, I would save the value in a variable and use it rather than re-calling the function.

python/tests/unit/arcticdb/version_store/test_missing_empty.py
```diff
@@ -125,6 +125,39 @@ def pd_delete_replace(df1, df2, date_range=None):
...
+    if t in fill_values.keys():
+        return fill_values[t]
+    return None
+
+
+def pd_arcticdb_read_sim(df):
+    read_df = df.copy(True)
+    for c in df.columns:
+        t = infer_type(df[c])
+        print(c, t, fill_value(t))
```
muhammadhamzasajjad, line 154: There shouldn't be a need to do this print as this function will be called by different tests and we don't want unnecessary logging.

python/tests/unit/arcticdb/version_store/test_missing_empty.py
```diff
@@ -125,6 +125,39 @@ def pd_delete_replace(df1, df2, date_range=None):
...
+            t = tv
+    return t
+
+
+def fill_value(t):
+    fill_values = {
+        bool: False,
+        int: 0
+    }
+    if t in fill_values.keys():
```
muhammadhamzasajjad, line 145: nit: even `t in fill_values` would work. No need to call `keys()`.

muhammadhamzasajjad (changes requested):

muhammadhamzasajjad (approved):
LGTM, given CI checks pass 🟢

PR #1508, opened 2024-04-19
Add version_chain ASV benchmarks

IvoDD:
- Will allow monitoring for regressions in version chain iteration and version map caching

#### Reference Issues/PRs
Should show improvement from #1443


python/.asv/results/ArcticDB-Runner-Medium/a35d2b7b-virtualenv-py3.6.json
G-D-Petrov, line 1: This file is no longer needed, as the data is stored in a ArcticDB in AWS.
OK to be deleted

G-D-Petrov (approved):

PR #1511, opened 2024-04-19
bugfix/1336: on s3 and azure add delimiter in iterate_type when prefix is empty to avoid KeyType::LOG_COMPACTED when KeyType::LOG is queried

### LINKED ISSUE #1336: KeyType::LOG and KeyType::LOG_COMPACTED share the same prefix in S3 and Azure
`KeyType::LOG`'s long name is `log` and `KeyType::LOG_COMPACTED`'s long name is `logc`.
And currently there is no "end delimiter" for key prefix. So if user is trying to iterate `KeyType::LOG`, the prefix for iteration will be `.../log`. As prefix of keys `KeyType::LOG_COMPACTED` is `.../logc`, the iteration will return both keys.

The issue currently has no impact but could be hazardous for future development.

muhammadhamzasajjad:
#### Reference Issues/PRs
Fixes #1336 

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/azure/azure_storage.cpp
+++ b/cpp/arcticdb/storage/azure/azure_storage.cpp
@@ -229,32 +229,33 @@ void do_remove_impl(Composite<VariantKey>&& ks,
         }
 }
 
-auto default_prefix_handler() {
-    return [] (const std::string& prefix, const std::string& key_type_dir, const KeyDescriptor& key_descriptor, KeyType) {
-        return !prefix.empty() ? fmt::format("{}/{}*{}", key_type_dir, key_descriptor, prefix) : key_type_dir;
-    };
+std::string prefix_handler(const std::string& prefix, const std::string& key_type_dir, const KeyDescriptor& key_descriptor, KeyType) {
+    return !prefix.empty() ? fmt::format("{}/{}*{}", key_type_dir, key_descriptor, prefix) : key_type_dir;
 }
 
-template<class KeyBucketizer, class PrefixHandler>
+template<class KeyBucketizer>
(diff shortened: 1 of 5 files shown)


cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -229,32 +229,33 @@ void do_remove_impl(Composite<VariantKey>&& ks,
...
 }
 
-template<class KeyBucketizer, class PrefixHandler>
+template<class KeyBucketizer>
 void do_iterate_type_impl(KeyType key_type,
     const IterateTypeVisitor& visitor,
     const std::string& root_folder,
     AzureClientWrapper& azure_client,
     KeyBucketizer&& bucketizer,
-    PrefixHandler&& prefix_handler = default_prefix_handler(),
```
alexowens90, line 244: I don't have the context for why this was previously templated, one to check with @willdealtry

muhammadhamzasajjad: I think this may have been due to a copy paste from S3 to Azure? In S3 it makes sense because two different prefix_handlers are used, a [default one](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/storage/s3/detail-inl.hpp#L268) and a custom on in [nfs_backend](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp#L161). But in Azure, the same definition has been used twice. But @willdealtry probably knows better.

PR #1517, opened 2024-04-22
Bugfix/1507/disallow appending row range to datetime index with dynamic schema

### LINKED ISSUE #1507: Dynamic schema can append rowrange to datetime index leading to datacorruption
### Describe the bug

When dynamic schema is enabled and the symbol contains empty dataframe it is possible to append a rowrange index to the symbol. This should be forbidden since the indexes are not matching. The append changes the index type to be rowrange, but still it is possible to append datetime to that as well leading to a data corruption.

### Steps/Code to Reproduce

```python
import arcticdb
import numpy as np
import pandas as pd

# 4.4.0
print(arcticdb.__version__)

opts = arcticdb.LibraryOptions(dynamic_schema=True)

lib.write("test_dyn", pd.DataFrame({"a": [], dtype="float64"}))
assert lib.read("test_dyn").data.index.equals(pd.DatetimeIndex([]))

# This should not be allowed
lib.append("test_dyn", pd.DataFrame({"a": [1.0]}, pd.RangeIndex(start=0, stop=1, step=1)))
# The index appears to be changed
assert lib.read("test_dyn").data.index.equals(pd.RangeIndex(start=0, stop=1, step=1))
# The output looks ok even though the index is messed up
#    a
# 0  1
print(lib.read("test_dyn").data)


# Even though the index appears to be changed the following will throw:
# lib.append("test_dyn", pd.DataFrame({"a": [1.0]}, pd.RangeIndex(start=1, stop=2, step=1)))

# Appending datetime to this is allowed only if validate index is false
lib.append("test_dyn", pd.DatetimeIndex(["01/01/2024"]), validate_index=False)

# The data is corrupted
#                         index
# 1                         NaT
# 682406825570164596 2024-01-01
```

### Expected Results

There are two cases:
1. The empty index feature is enabled (#1429, #1475)
Anything could be appended to symbol containing an empty dataframe. The type of the index is determined at the time of first append (or update). After the type of the index is determined index mixing indexes should not be allowed.

2. The empty index feature is disabled
The index is determined at write time. In case of Pandas 2 it is DatetimeIndex([]) for all empty DataFrames (in case of Pandas 1 it is not so consistent refer to test_empty_column_type.py for the expected behavior). Appending to a symbol containing empty dataframe is allowed only if the indexes are matching.

OS: Windows-10-10.0.22631-SP0
ArcticDB: 4.4.0
Numpy: 1.26.3
Pandas: 2.1.4

### Backend storage used

_No response_

### Additional Context

_No response_

alexowens90:
Fixes #1507

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -461,6 +461,7 @@ set(arcticdb_srcs
         storage/storage_factory.cpp
         stream/aggregator.cpp
         stream/append_map.cpp
+        stream/index.cpp
         stream/piloted_clock.cpp
         toolbox/library_tool.cpp
         util/allocator.cpp
--- a/cpp/arcticdb/entity/merge_descriptors.cpp
+++ b/cpp/arcticdb/entity/merge_descriptors.cpp
@@ -46,8 +46,13 @@ StreamDescriptor merge_descriptors(
     // Merge all the fields for all slices, apart from the index which we already have from the first descriptor.
     // Note that we preserve the ordering as we see columns, especially the index which needs to be column 0.
     for (const auto &fields : entries) {
-        if(has_index)
-            util::variant_match(index, [&fields] (const auto& idx) { idx.check(*fields); });
+        if (has_index) {
+            util::variant_match(index,
+                [](const EmptyIndex&) {},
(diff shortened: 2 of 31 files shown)


alexowens90:
Superseded by #1529

PR #1529, opened 2024-04-25
Bugfix 1507+1509: dynamic schema different index append fixes

### LINKED ISSUE #1507: Dynamic schema can append rowrange to datetime index leading to datacorruption

### LINKED ISSUE #1509: Dynamic schema allows appending non-matching RangeIndexes
### Describe the bug

When dynamic schema is enabled it is possible to append dataframe with ranged index to symbol even if the ranges are not matching. This makes it possible to
1. Append ranged index with start different than the stop of the symbol
2. Append ranged index with step different than the step of the symbol
Both actions change the index even for the old rows and are forbidden with static_schema.

### Steps/Code to Reproduce

```python
>>> import arcticdb
>>> import pandas as pd
>>> import numpy as np
>>> ac = arcticdb.Arctic("lmdb://test")
>>> opts = arcticdb.LibraryOptions(dynamic_schema=True)
>>> lib = ac.get_library("test", create_if_missing=True, library_options=opts)
>>> lib.write("test", pd.DataFrame({"a": [1, 2, 3]}))
VersionedItem(symbol='test', library='test', data=n/a, version=0, metadata=None, host='LMDB(path=...test)', timestamp=1713515824376569000)
>>> lib.read("test").data
   a
0  1
1  2
2  3
>>> lib.append("test", pd.DataFrame({"a": [1]}, pd.RangeIndex(start=5, stop=6, step=1)))
VersionedItem(symbol='test', library='test', data=n/a, version=1, metadata=None, host='LMDB(path=...test)', timestamp=1713515904432580000)
>>> lib.read("test").data
   a
5  1
6  2
7  3
8  1
>>> lib.append("test", pd.DataFrame({"a": [1]}, pd.RangeIndex(start=5, stop=6, step=10)))
VersionedItem(symbol='test', library='test', data=n/a, version=6, metadata=None, host='LMDB(path=...test)', timestamp=1713517240251682100)
>>> lib.read("test").data
    a
5   1
15  2
25  3
35  1
45  1
```

### Expected Results

Do not allow appending RangeIndex-ed dataframes if the ranges are not matching.

Python: 3.10.11 (tags/v3.10.11:7d4cc5a, Apr 5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
OS: Windows-10-10.0.22631-SP0
ArcticDB: 4.4.0
Numpy: 1.26.3
Pandas: 2.1.4

### Backend storage used

_No response_

### Additional Context

_No response_

alexowens90:
Closes #1507 
Closes #1509

While fixing, noticed that the added test `test_append_range_index_from_zero` would also not have passed, so fixed this too.

### DIFF
--- a/cpp/arcticdb/python/normalization_checks.cpp
+++ b/cpp/arcticdb/python/normalization_checks.cpp
@@ -124,12 +124,12 @@ void update_rowcount_normalization_data(
             size_t new_start = new_index->start();
             auto stop = old_index->start() + old_length * old_index->step();
             normalization::check<ErrorCode::E_INCOMPATIBLE_INDEX>(
-                new_start == stop,
-                "The appending data has a RangeIndex.start={} that is not contiguous with the {}"
-                "stop ({}) of",
-                error_suffix,
+                new_start == stop || (new_start == 0 && new_index->step() == 1),
+                "The appending data has a RangeIndex.start={} that is not contiguous with the "
+                "stop ({}) of {}",
                 new_start,
-                stop
+                stop,
+                error_suffix
             );
 
             new_pandas->get().mutable_index()->set_start(old_index->start());
--- a/cpp/arcticdb/python/normalization_checks.cpp
+++ b/cpp/arcticdb/python/normalization_checks.cpp
@@ -122,17 +122,15 @@ void update_rowcount_normalization_data(
             );
 
             size_t new_start = new_index->start();
-            if (new_start != 0) {
-                auto stop = old_index->start() + old_length * old_index->step();
-                normalization::check<ErrorCode::E_INCOMPATIBLE_INDEX>(
-                    new_start == stop,
-                    "The appending data has a RangeIndex.start={} that is not contiguous with the {}"
-                    "stop ({}) of",
-                    error_suffix,
-                    new_start,
-                    stop
-                );
-            }
+            auto stop = old_index->start() + old_length * old_index->step();
+            normalization::check<ErrorCode::E_INCOMPATIBLE_INDEX>(
+                new_start == stop,
+                "The appending data has a RangeIndex.start={} that is not contiguous with the {}"
+                "stop ({}) of",
+                error_suffix,
+                new_start,
+                stop
+            );
 
             new_pandas->get().mutable_index()->set_start(old_index->start());
         }
--- a/cpp/arcticdb/version/schema_checks.hpp
+++ b/cpp/arcticdb/version/schema_checks.hpp
@@ -121,12 +121,9 @@ inline void fix_descriptor_mismatch_or_throw(
     const auto &old_sd = existing_isr.tsd().as_stream_descriptor();
     check_normalization_index_match(operation, old_sd, new_frame, empty_types);
 
-    if (dynamic_schema)
-        return; // TODO: dynamic schema may need some of the checks as below
-
     fix_normalization_or_throw(operation == APPEND, existing_isr, new_frame);
 
-    if (!columns_match(old_sd, new_frame.desc)) {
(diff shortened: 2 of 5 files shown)


cpp/arcticdb/python/normalization_checks.cpp
```diff
@@ -122,17 +122,15 @@ void update_rowcount_normalization_data(
...
-                    "The appending data has a RangeIndex.start={} that is not contiguous with the {}"
-                    "stop ({}) of",
-                    error_suffix,
-                    new_start,
-                    stop
-                );
-            }
+            auto stop = old_index->start() + old_length * old_index->step();
+            normalization::check<ErrorCode::E_INCOMPATIBLE_INDEX>(
+                new_start == stop || (new_start == 0 && new_index->step() == 1),
```
vasil-pashov, line 127: is this going to allow something like this to happen
     a
0   0
1   1
2   2

Append index (0, 1, 2)

     a
0   0
1   1
2   2
0   3
1   4
2   5

Or is there some other check preventing it?

alexowens90: We have to allow appending index `[0, 1, 2` to index `[0, 1, 2]`, otherwise our default behaviour when you don't provide an index won't work. We only store the start and step though, so it will come out as:
```
0 0
1 1
2 2
3 3
4 4
5 5
```

vasil-pashov (approved):

PR #1530, opened 2024-04-25
Should speed up ASV benchmarks significantly

IvoDD:
- Separates non-batch and batch BasicFunctions. Uses multiple symbols only where it is required
- Use single symbol for ModificationFunctions
- Use less rows for batch benchmarks
- Decrease version chain length for IteratingVersionChain

### DIFF
--- a/python/benchmarks/basic_functions.py
+++ b/python/benchmarks/basic_functions.py
@@ -15,9 +15,12 @@ from .common import *
 # Common parameters between BasicFunctions and ModificationFunctions
 WIDE_DF_ROWS = 5_000
 WIDE_DF_COLS = 30_000
-# PARAMS = ([1_000, 1_500], [50, 100])
-PARAMS = ([100_000, 150_000], [500, 1000])
-PARAM_NAMES = ["rows", "num_symbols"]
+# We use larger dataframes for non-batch methods
+PARAMS = ([1_000_000, 1_500_000])
+PARAM_NAMES = ["rows"]
+BATCH_PARAMS = ([25_000, 50_000], [500, 1000])
+BATCH_PARAM_NAMES = ["rows", "num_symbols"]
+DATE_RANGE = pd.date_range("2022-12-31", "2023-01-01")
 
 
 class BasicFunctions:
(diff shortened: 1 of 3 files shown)


python/benchmarks/basic_functions.py
```diff
@@ -78,20 +77,87 @@ def get_fresh_lib(self):
...
+
+        self.df = generate_pseudo_random_dataframe(rows)
+        self.lib = self.ac[get_prewritten_lib_name(rows)]
+        self.fresh_lib = self.get_fresh_lib()
+
+    def get_fresh_lib(self):
+        self.ac.delete_library("fresh_lib")
+        return self.ac.create_library("fresh_lib")
+
     def time_write_staged(self, rows, num_symbols):
```
alexowens90, line 161: Staged writes aren't a batch method, move to `BasicFunctions` and make sure there are multiple incomplete segments to compact

python/benchmarks/basic_functions.py
```diff
@@ -78,20 +77,87 @@ def get_fresh_lib(self):
...
+
+    def peakmem_read_with_date_ranges(self, rows):
+        self.lib.read(f"sym", date_range=BasicFunctions.DATE_RANGE).data
+
+
+class BatchBasicFunctions:
+    number = 5
+    timeout = 6000
+    CONNECTION_STRING = "lmdb://batch_basic_functions?map_size=20GB"
+    DATE_RANGE = pd.date_range("2023-01-01", "2023-01-01")
```
alexowens90, line 125: I know this hasn't changed, but this date range will filter out all rows, which I'm guessing isn't the intended behaviour

alexowens90 (changes requested):

python/benchmarks/basic_functions.py
IvoDD: Agreed. Thanks for flagging!
Should be fixed now.

python/benchmarks/basic_functions.py
IvoDD: Thanks, moved it to `2022-12-31, 2023-01-01`, which should be non empty (last timestamp on all dataframes should be 2023-01-01)

alexowens90 (approved):

PR #1538, opened 2024-04-29
Fix crash in calling Library::tail when the tail contains empty typed columns

vasil-pashov:
This fixes two things:
1. When using dynamic schema and not using direct read the function copying the source column to the dest column now checks if the source is empty type. If it is empty type it will call default_initialize. Previously it tried to call memcpy and crashed because it tried to deref NULL
2. During decoding (both v1 and v2) the number of row in a segment is determined by taking the max number of rows in each column. Previously it was determined by the number of rows in the first column. However it the first column was of empty type it would report 0 rows, which is wrong. This lead to creating sparse maps for dense columns.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/codec/codec.cpp
```diff
@@ -348,13 +348,14 @@ void decode_v2(
...
         EncodedFieldCollection encoded_fields(std::move(encoded_fields_buffer));
-        const auto seg_row_count = fields_size ? ssize_t(encoded_fields.at(0).ndarray().items_count()) : 0L;
+        auto seg_row_count = fields_size ? ssize_t(encoded_fields.at(0).ndarray().items_count()) : ssize_t(0);
         res.init_column_map();
 
         for (std::size_t i = 0; i < fields_size; ++i) {
             const auto& encoded_field = encoded_fields.at(i);
             const auto& field_name = desc.fields(i).name();
             util::check(data!=end, "Reached end of input block with {} fields to decode", fields_size-i);
+            seg_row_count = std::max(seg_row_count, ssize_t(encoded_field.ndarray().items_count()));
```
alexowens90, line 358: I assume `items_count` is the number of values actually stored, ignoring sparsity? If so, this will work right now, but there is a potential future use case where we have sparse data with a row-count index, where the segment row count depends on the column sparse maps as well. e.g.
```
lib.write(sym, pd.DataFrame("col1": [np.nan, 0.0], "col2": [1.0, np.nan]), sparsify_floats=True)
```
would set `seg_row_count` to `1` here, when it should be `2`. Please add a comment detailing this.

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
```diff
@@ -209,75 +209,63 @@ class TestCanAppendColumnWithNonesToColumn:
...
+        assert_frame_equal(lmdb_version_store_static_and_dynamic.read("sym").data, expected_df)
+        assert_frame_equal(lmdb_version_store_static_and_dynamic.read("sym", row_range=[0,3]).data, df_initial)
         assert_frame_equal(
             lmdb_version_store_static_and_dynamic.read("sym", row_range=[3,5]).data,
-            pd.DataFrame({"col": np.array([0,0], dtype=int_dtype)})
+            pd.DataFrame({"col": np.array([0,0], dtype=int_dtype), "other": [4, 5]})
         )
+        # Cannot compare with expected_df.tail(n=1) due to issue #1537: https://github.com/man-group/ArcticDB/issues/1537
+        expected_tail = pd.DataFrame({"col": np.array([0], dtype=int_dtype), "other": [5]})
+        assert_frame_equal(lmdb_version_store_static_and_dynamic.tail("sym", n=1).data, expected_tail)
```
alexowens90, line 224: Tail and row range do the same thing under the hood here, so I don't think we need to test both (or if we do for some reason, should also test head)

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
vasil-pashov: Nope, the row row range is performing a `direct_read` while tails is performing `read_and_process`. I was thinking that we should add separate test suite testing the process pipeline, so I did the minimal change here. Should I add head for completeness?

cpp/arcticdb/codec/codec.cpp
vasil-pashov: Hmm, it's an interesting question. There is also EncodedField::values_count_ which I believe holds the value you're describing, but I'm not 100% sure. @willdealtry can you shed some light on this?

python/tests/unit/arcticdb/version_store/test_empty_column_type.py
alexowens90: You're right. Weird inconsistency, but yes, let's test `head` as well for completeness

cpp/arcticdb/codec/codec.cpp
alexowens90: I've discussed this with William, and we can actually do this properly now. After the iteration over fields, the result segment in memory has the columns and their sparse maps fully populated. Then we can just take the max over the last logical rows of all the columns to get the segment row count.

willdealtry (approved):

PR #1542, opened 2024-04-30
Revert to C++17 build

willdealtry:
Building with C++20 causes sigabrt when ray is imported first. Temporarily revert to C++17

### DIFF
--- a/cpp/arcticdb/log/log.cpp
+++ b/cpp/arcticdb/log/log.cpp
@@ -33,16 +33,12 @@ namespace arcticdb::log {
 
 static const char* DefaultLogPattern = "%Y%m%d %H:%M:%S.%f %t %L %n | %v";
 
-
 namespace {
-    std::shared_ptr<Loggers> loggers_instance_;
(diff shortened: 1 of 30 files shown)


cpp/arcticdb/log/log.cpp
```diff
@@ -72,7 +72,7 @@ struct Loggers::Impl
 
 
 constexpr auto get_default_log_level() {
-    return spdlog::level::info;
+    return spdlog::level::trace;
```
alexowens90, line 75: This needs backing out

alexowens90 (changes requested):

cpp/arcticdb/log/log.cpp
willdealtry: Oops, good spot

willdealtry: Keep meaning to write a test for this 'cos I do it all the time

alexowens90 (approved):

Issue #711, opened 2023-08-10
Make all modification methods respect `prune_previous_versions`


qc00:
**Is your feature request related to a problem? Please describe.**

Some methods in `library.py` do not respect the `prune_previous_versions` library option.
This makes the API inconsistent.

Discovered in #705 

**Describe the solution you'd like**

* Make `sort_merge_internal`, `sort_index`,  `delete_range_internal` and `defragment_symbol_data` C++ methods accept an extra argument and call `write_version_and_prune_previous_if_needed` instead of `write_version`
* Make the corresponding Python methods check the library option and pass it to C++. Also, consider adding a parameter to allow the user to override on each call.
* [x] Thoroughly check if there are other incorrect uses of `write_version`
* [x] Consider if we need to bump the major version

**Describe alternatives you've considered**

* Leave the methods unchanged, which might lead to the original cause of #705 where a user unknowingly leave thousands of versions behind.
* Improve the documentation to ask the user to call `prune_previous_versions` after such methods. This is inefficient as the version chain might need reloading by then.

phoebusm:
-  `sort_merge` -> `sort_merge_internal`: prune_previous not available
- `sort_index`: prune_previous not available
- `delete_range` -> `delete_range_internal`: prune_previous not available
- `defragment_symbol_data`: prune_previous not available

PR #1556, opened 2024-05-08
Make all modification methods respect prune_previous_versions

phoebusm:
#### Reference Issues/PRs

https://github.com/man-group/ArcticDB/issues/711
Add prune_previous option to APIs which still don't support it
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -49,6 +49,14 @@ struct KeySizesInfo {
     size_t uncompressed_size; // bytes
 };
 
+struct SortMergeOptions {
+    bool append_;
+    bool convert_int_to_float_;
+    bool via_iteration_;
+    bool sparsify_;
+    bool prune_previous_versions_;
+};
+
 class LocalVersionedEngine : public VersionedEngine {
 
 public:
@@ -81,7 +89,7 @@ public:
     VersionedItem delete_range_internal(
         const StreamId& stream_id,
         const UpdateQuery& query,
-        bool dynamic_schema) override;
+        const DeleteRangeOptions& option) override;
 
     void append_incomplete_segment(
         const StreamId& stream_id,
@@ -258,11 +266,7 @@ public:
     VersionedItem sort_merge_internal(
         const StreamId& stream_id,
         const std::optional<arcticdb::proto::descriptors::UserDefinedMetadata>& user_meta,
-        bool append,
-        bool convert_int_to_float,
-        bool via_iteration,
-        bool sparsify
-        );
+        const SortMergeOptions& option);
 
     std::vector<folly::Future<AtomKey>> batch_write_internal(
         const std::vector<VersionId>& version_ids,
@@ -329,7 +333,7 @@ public:
 
     bool is_symbol_fragmented(const StreamId& stream_id, std::optional<size_t> segment_size) override;
 
-    VersionedItem defragment_symbol_data(const StreamId& stream_id, std::optional<size_t> segment_size) override;
+    VersionedItem defragment_symbol_data(const StreamId& stream_id, std::optional<size_t> segment_size, bool prune_previous_versions) override;
     
     StorageLockWrapper get_storage_lock(const StreamId& stream_id) override;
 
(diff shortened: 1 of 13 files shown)


cpp/arcticdb/version/test/test_version_store.cpp
```diff
@@ -194,7 +194,7 @@ TEST_F(VersionStoreTest, SortMerge) {
         test_store_->append_incomplete_frame(symbol, std::move(frame.input_frame_));
     }
 
-    test_store_->sort_merge_internal(symbol, std::nullopt, true, false, false, false);
+    test_store_->sort_merge_internal(symbol, std::nullopt, true, false, false, false, false);
```
poodlewars, line 197: 5 bools in a row isn't a good API (nor was 4 before your change!). Refactor by pulling out a struct for the options?

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -81,7 +81,8 @@ class LocalVersionedEngine : public VersionedEngine {
     VersionedItem delete_range_internal(
         const StreamId& stream_id,
         const UpdateQuery& query,
-        bool dynamic_schema) override;
+        bool dynamic_schema,
+        bool prune_previous_versions) override;
```
poodlewars, line 85: Too many bools - pull out a struct for the options?

cpp/arcticdb/version/version_store_api.hpp
```diff
@@ -170,7 +171,8 @@ class PythonVersionStore : public LocalVersionedEngine {
             bool append,
             bool convert_int_to_float,
             bool via_iteration,
-            bool sparsify
+            bool sparsify,
+            bool prune_previous_versions = false
```
poodlewars, line 175: Why is this the only one with a default???

python/tests/conftest.py
```diff
@@ -238,6 +238,26 @@ def arctic_library(arctic_client, lib_name):
     return arctic_client.create_library(lib_name)
 
 
+@pytest.fixture(
+    scope="function",
+    params=[
```
poodlewars, line 243: Why is this parameterized? Isn't just LMDB fine? Isn't there a similar fixture already?

python/tests/integration/arcticdb/version_store/test_deletion.py
```diff
@@ -540,6 +540,37 @@ def test_delete_date_range_with_strings(version_store_factory, index_start):
     assert_frame_equal(vit.data, df)
 
 
+@pytest.mark.parametrize("prune_previous_versions", [True, False])
```
poodlewars, line 543: Most of this test is the same as the one above it?

poodlewars (changes requested):

cpp/arcticdb/version/version_store_api.hpp
phoebusm: Yes misread that. It is unnecessary

python/tests/conftest.py
phoebusm: I was trying to create a library fixture similar to existing `def basic_store()`
It is used by `test_write_parallel_sort_merge` which tests parallelism. So I think it's safer to test multiple storages.

python/tests/integration/arcticdb/version_store/test_deletion.py
phoebusm: Yes but I don't want to make `def test_delete_date_range_with_strings` more complicated. 
And I only need to test against lmdb

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -49,6 +49,14 @@ struct KeySizesInfo {
     size_t uncompressed_size; // bytes
 };
 
+struct SortMergeOptiona {
```
poodlewars, line 52: typo - should be `SortMergeOptions`

poodlewars (changes requested):

cpp/arcticdb/version/local_versioned_engine.hpp
phoebusm: 😿

poodlewars (approved):

PR #1560, opened 2024-05-09
Only write ref key once when writing with prune_previous_versions

poodlewars:

We should only write the version ref key once when we write with `prune_previous_versions=True`. Currently we are writing it twice - once after we write the tombstone all and once when we write the new version. This means that there is a period of time where the symbol is unreadable.

This was fixed a while ago with PR #1104 but regressed with PR #1355.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/build_tooling/parallel_test.sh
+++ b/build_tooling/parallel_test.sh
@@ -9,16 +9,16 @@ echo Saving results to ${TEST_OUTPUT_DIR:="$(realpath "$tooling_dir/../cpp/out")
 
 catch=`{ which catchsegv 2>/dev/null || echo ; } | tail -n 1`
 
-    set -o xtrace -o pipefail
+set -o xtrace -o pipefail
 
-    # Build a directory that's just the test assets, so can't access other Python source not in the wheel
-    mkdir -p $PARALLEL_TEST_ROOT
-    MSYS=winsymlinks:nativestrict ln -s "$(realpath "$tooling_dir/../python/tests")" $PARALLEL_TEST_ROOT/
-    cd $PARALLEL_TEST_ROOT
+# Build a directory that's just the test assets, so can't access other Python source not in the wheel
+mkdir -p $PARALLEL_TEST_ROOT
+MSYS=winsymlinks:nativestrict ln -s "$(realpath "$tooling_dir/../python/tests")" $PARALLEL_TEST_ROOT/
+cd $PARALLEL_TEST_ROOT
 
-    export ARCTICDB_RAND_SEED=$RANDOM
+export ARCTICDB_RAND_SEED=$RANDOM
 
-    $catch python -m pytest --timeout=3600 $PYTEST_XDIST_MODE -v --log-file="$TEST_OUTPUT_DIR/pytest-logger.$group.log" \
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/version/version_map.hpp
```diff
@@ -246,15 +246,6 @@ class VersionMapImpl {
             log_write(store, key.id(), key.version_id());
     }
 
-    AtomKey write_tombstone_all_key(
```
poodlewars, line 249: This was unused.

alexowens90: In the old internal project it was used in replication, not sure if you ripped it out there too though?

cpp/arcticdb/version/version_map.hpp
```diff
@@ -512,7 +506,7 @@ class VersionMapImpl {
         auto journal_key = to_atom(std::move(journal_single_key(store, key, entry->head_)));
         write_to_entry(entry, key, journal_key);
         auto previous_index = entry->get_second_undeleted_index();
-        write_symbol_ref(store, key, previous_index, journal_key);
+        return previous_index;
```
alexowens90, line 509: This seems logically wrong. `do_write` is called from a few places, some of which then call `write_symbol_ref`, and some of which don't.

cpp/arcticdb/version/version_map.hpp
poodlewars: Good memory! Enterprise uses,

```
bool delete_version_and_previous_versions(
        const std::shared_ptr <Store> &store,
        const std::shared_ptr <VersionMap> &version_map,
        const StreamId &stream_id,
        VersionId version_id) {
    ARCTICDB_SAMPLE(deleteVersionAndPreviousVersions, 0)
    auto dummy_key = atom_key_builder()
        .version_id(version_id)
        .build(stream_id, KeyType::VERSION);
    version_map->tombstone_from_key_or_all(store, stream_id, dummy_key);
    auto symbol_deleted = !has_undeleted_version(store, version_map, stream_id);
    return symbol_deleted;
}
```

now, so I think should survive this change.

poodlewars:
PR #1355 also added a `write_symbol_ref` call to `write_entry_to_storage`. Repeating the exercise above we have:

```
write_entry_to_storage
|_ compact_and_remove_deleted_indexes (unused, no impact)
  |_ writes the symbol ref a second time
|_ overwrite_symbol_tree
  |_ writes the symbol ref a second time
|_ rewrite_entry
  |_ recover_deleted
    |_ writes the symbol ref a second time
  |_ remove_and_rewrite_version_keys
    |_ writes the symbol ref a second time
  |_ scan_and_rewrite
    |_ writes the symbol ref a second time
```

build_tooling/parallel_test.sh
poodlewars, line 1: These changes are just formatting.

cpp/arcticdb/version/version_map.hpp
alexowens90: OK, you've convinced me 😄

alexowens90 (approved):

PR #1566, opened 2024-05-10
Remove non-lmdb stores from unit tests

### LINKED ISSUE #1564: Make sure that we can run unit tests locally
Unit tests should be easy to execute and shouldn't take too much time to execute

Remove / have a way to skip tests that require a mongo/azure simulator, make sure that all tests that require lmdb are passing locally

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1564 

It remove all simulators from the unit tests (e.g. S3 moto, azurite, mongo)

TODO - make sure that we don't need these tests, if we don - add them to the integration tests

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_aggregation_dynamic.py
+++ b/python/tests/unit/arcticdb/version_store/test_aggregation_dynamic.py
@@ -37,7 +37,10 @@ def assert_equal_value(data, expected):
     assert_frame_equal(received.astype("float"), expected)
 
 
-@pytest.mark.xfail(MACOS_CONDA_BUILD, reason="Conda Pandas returns nan instead of inf like other platforms")
+@pytest.mark.xfail(
+    MACOS_CONDA_BUILD,
+    reason="Conda Pandas returns nan instead of inf like other platforms",
+)
 @use_of_function_scoped_fixtures_in_hypothesis_checked
 @settings(deadline=None)
 @given(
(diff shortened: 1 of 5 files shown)


python/tests/unit/arcticdb/version_store/test_aggregation_dynamic.py
```diff
@@ -93,8 +96,8 @@ def test_hypothesis_mean_agg_dynamic(lmdb_version_store_dynamic_schema_v1, df):
         index=range_indexes(),
     )
 )
-def test_hypothesis_sum_agg_dynamic(s3_version_store_dynamic_schema_v2, df):
-    lib = s3_version_store_dynamic_schema_v2
+def test_hypothesis_sum_agg_dynamic(lmdb_version_store_dynamic_schema_v1, df):
```
vasil-pashov, line 99: Why was the encoding version changed from v2 to v1?

G-D-Petrov: typo, fixed

G-D-Petrov:
Moved the test_fork test to integration as it seems sensible to test this with a s3 simulation at least
@alexowens90 - do we need to test the QB aggregations with s3/azure simulations or is lmdb enough?

willdealtry:
Are we sure this is a good idea? Azurite and Mongo are quite irritating in that they mostly fail locally, but S3 is the storage that the vast majority of our users actually use. This means we're removing more than half the tests in the whole project from the one storage that actually matters to the majority of people, which seems like a high price to pay for knocking a few minutes off a CI run

Issue #399, opened 2023-05-18
Add a way to return just the index of a symbol when its index has no name 


aschonfeld:
Say we have a symbol which has a time series index with no name.

```
import pandas as pd

df = pd.DataFrame(dict(foo=[1,2,3], bar=[4,5,6]))
df = df.reindex([pd.Timestamp('20230101'), pd.Timestamp('20230102'), pd.Timestamp('20230103')])
df.index
```

It would be excellent if we can select that index using the `columns` parameter of the `read` function by doing something like the following:

```
lib.read('symbol', columns=['index'])
```

This would return: `DatetimeIndex(['2023-01-01', '2023-01-02', '2023-01-03'], dtype='datetime64[ns]', freq=None)`

jamesmunro:
Rather than a special column name, `index`, perhaps a new arg, or an attribute on the returned object?
```
lib.read('symbol').index
```

aschonfeld:
That would be great assuming the data will be lazy-loaded (meaning it doesn't load the entire dataframe just to pop off the index)

Naich:
> ps a new arg, or an attri

this would be same as `lib.read('symbol').data.index`, won't benefit from col segment

vasil-pashov:
Resolved by #1568

PR #1568, opened 2024-05-10
Read index API

vasil-pashov:
#### Reference Issues/PRs

relates to: #399
Add unit tests to showcase the proposed API for reading the index of a symbol. 

**Proposal**
Add new `arcticdb.Library` API
```python
def read_index(symbol: str, as_of: Optional[AsOf] = None) -> VersionedItem
```
`symbol` and `as_of` have the same meaning as in [arcticdb.Library.read](https://docs.arcticdb.io/latest/api/library/#arcticdb.version_store.library.Library.read)

`VersionedItem.data` will contain the index retrieved from the symbol. Its type will be one of `pandas.RangeIndex`, `pandas.DatetimeIndex`, `pandas.MultiIndex` (it can be expanded to all supported index types) .

**TBD**
* Is optional `date_range`/`row_range` argument needed? (I would add it only if someone requests it)
* Do we support categorical indexes should they be considered in the first iteration of the task.

**Alternatives**
Extend `arcticdb.Library.read` with additional parameter `index_only`.
Pros:
* Less code duplication
* Move all read related stuff to one function
Cons:
* `VersionedItem.data` will be of different type depending on the parameter which might lead to confusion and bugs
* `arcticdb.Library.read` might become huge and hard to understand and use
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


alexowens90:
Mostly looks good, few comments/ideas:

- We have code to support string indexes, although I'm not sure how widely used it is. As a minimum, we need to behave sensibly for string indexed data.
- What should the behaviour be for the empty index type?
- We definitely want `date_range`, and for completeness I think we should include `row_range` too.
- We do not support categorical indexes, so these do not need to be considered.
- I think `VersionedItem` is the correct return type, with the index in the `data` field.
- `NativeVersionStore` already has a method called `read_index`, so we should use something else for this. Maybe `get_index`?
- I think it is better as a separate method, rather than an argument to `read`. When selecting a subset of columns, you always get the index columns too without specifying them. We could have a special case, where `columns=[]` gets just the index, but it isn't intuitive, and would change the return type of `data` based on a parameter, which is nasty, particularly if the columns list is generated at runtime and might be empty.

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Designed to test all combinations between schema and encoding version for LMDB
+    """
+    yield request.getfixturevalue(request.param)
+
+@pytest.mark.skip(reason="Not implemented")
+class TestReadSingleIndex:
+    @pytest.mark.parametrize("index", [
+        pd.RangeIndex(start=0, stop=10),
+        pd.RangeIndex(start=0, stop=10, step=2),
```
alexowens90, line 32: I would test with a non-zero start value here.

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,82 @@
...
+        pd.date_range(start="01/01/2024",end="01/10/2024"),
+        pd.MultiIndex.from_arrays(
+            [pd.date_range(start="01/01/2024", end="01/10/2024"), pd.RangeIndex(start=0, stop=10)],
+            names=["datetime", "level"]
+        )
+    ])
+    def test_rowrange(self, lmdb_version_store_static_and_dynamic, index):
+        lmdb_version_store_static_and_dynamic.write("sym", pd.DataFrame({"col": range(0, len(index))}, index=index))
+        result = lmdb_version_store_static_and_dynamic.read_index("sym")
```
alexowens90, line 41: `lmdb_version_store_static_and_dynamic` is of type `NativeVersionStore`, which has a `read_index` method with different functionality.

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,82 @@
...
+            pd.MultiIndex.from_arrays(
+                [pd.date_range(start="01/11/2024", end="01/21/2024"), pd.RangeIndex(start=10, stop=21)],
+                names=["datetime", "level"]
+            )
+        ]
+    ])
+    def test_as_of(self, lmdb_version_store_static_and_dynamic, indexes):
+        data = [list(range(0, len(index))) for index in indexes]
+        
+        lmdb_version_store_static_and_dynamic.write("sym", pd.DataFrame({"col": data[0]}, index=indexes[0]))
```
alexowens90, line 72: Would be good to use a fixture with 2 columns per segment, and >2 columns to check that everything works with column slicing.

python/arcticdb/version_store/_store.py
```diff
@@ -1716,14 +1756,21 @@ def head(
...
         q = q._head(n)
         version_query, read_options, read_query = self._get_queries(
             symbol=symbol, as_of=as_of, date_range=None, row_range=None, columns=columns, query_builder=q, **kwargs
         )
         read_result = self._read_dataframe(symbol, version_query, read_query, read_options)
-        return self._post_process_dataframe(read_result, read_query, q)
+        index_type = read_result.norm.df.common.WhichOneof("index_type")
+        index_is_rowcount = index_type == "index" and not read_result.norm.df.common.index.is_physically_stored
+        if implement_read_index and columns == [] and index_is_rowcount:
+            return self._postprocess_df_with_only_rowcount_idx(read_result, (0, n))
```
alexowens90, line 1771: What if `n` is negative? (same question for tail)

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,413 @@
...
+    @pytest.mark.parametrize("dynamic_schema", [False, True])
+    def test_read_index_columns(self, lmdb_storage, index, lib_name, dynamic_schema):
+        ac = lmdb_storage.create_arctic()
+        lib = ac.create_library(lib_name, LibraryOptions(dynamic_schema=dynamic_schema))
+        df = pd.DataFrame({"col": range(0, len(index))}, index=index)
+        lib.write("sym", df)
+        result = lib.read("sym", columns=[])
+        assert isinstance(result, arcticdb.VersionedItem)
+        assert result.symbol == "sym"
+        assert result.version == 0
```
alexowens90, line 51: We can omit the 3 asserts on lines 49-51 in the other tests, they add a lot of verbosity without adding much value

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,413 @@
...
+        assert result.data.index.equals(index)
+        assert result.data.empty
+
+    @pytest.mark.parametrize("dynamic_schema", [False, True])
+    def test_read_index_columns_silce(self, lmdb_storage, index, lib_name, dynamic_schema):
```
alexowens90, line 56: Typo. This also tests row slicing

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,413 @@
...
+
+    @pytest.mark.parametrize("dynamic_schema", [False, True])
+    def test_recursive_throws(self, lmdb_storage, lib_name, dynamic_schema):
+        data = {"a": np.arange(5), "b": np.arange(8)}
+        ac = lmdb_storage.create_arctic()
+        lib = ac.create_library(lib_name, LibraryOptions(dynamic_schema=dynamic_schema))
+        lib._nvs.write("sym_recursive", data, recursive_normalizers=True)
+        with pytest.raises(InternalException) as exception_info:
+            lib.read("sym_recursive", columns=[])
+        assert "Reading the index column is not supported when recursive or custom normalizers are used." in str(exception_info.value)
```
alexowens90, line 323: Less specific exception message checking, and on line 335

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,413 @@
...
+        assert result.version == 0
+        assert result.data.index.equals(index)
+        assert result.data.empty
+
+    @pytest.mark.parametrize("dynamic_schema", [False, True])
+    def test_read_index_columns_head(self, lmdb_storage, index, lib_name, dynamic_schema):
+        ac = lmdb_storage.create_arctic()
+        lib = ac.create_library(lib_name, LibraryOptions(dynamic_schema=dynamic_schema))
+        lib.write("sym", pd.DataFrame({"col": range(0, len(index))}, index=index))
+        result = lib.head("sym", columns=[], n=3)
```
alexowens90, line 75: Test for negative `n` (and in tail test too)

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -0,0 +1,413 @@
...
+        lib._nvs.write("sym_custom", data)
+
+        with pytest.raises(InternalException) as exception_info:
+            lib.read("sym_custom", columns=[])
+        assert "Reading the index column is not supported when recursive or custom normalizers are used." in str(exception_info.value)
+
+
+class TestReadBatch:
+    @pytest.mark.parametrize("dynamic_schema", [False, True])
+    def test_read_batch(self, lmdb_storage, lib_name, dynamic_schema, index):
```
alexowens90, line 340: Need a test with batch read and row-range arguments

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_read_index.py
vasil-pashov: batch read's row_range is generally broken: #1692

python/tests/unit/arcticdb/version_store/test_read_index.py
alexowens90: `silce`

python/tests/unit/arcticdb/version_store/test_read_index.py
alexowens90: Line 346 is still v specific

cpp/arcticdb/pipeline/query.hpp
```diff
@@ -163,21 +156,24 @@ inline FilterQuery<index::IndexSegmentReader> create_static_col_filter(util::Bit
...
+        auto res = std::make_unique<util::BitSet>(static_cast<util::BitSetSizeType>(pipeline->overall_column_bitset_->size())
+        );
         util::check(isr.bucketize_dynamic(), "Expected column group in index segment reader dynamic column filter");
         auto hash_bucket = isr.column(index::Fields::hash_bucket).begin<stream::SliceTypeDescriptorTag>();
         auto num_buckets = isr.column(index::Fields::num_buckets).begin<stream::SliceTypeDescriptorTag>();
 
-        bm::bvector<>::enumerator col = sc.first();
-        bm::bvector<>::enumerator col_end = sc.end();
+        bm::bvector<>::enumerator col = pipeline->overall_column_bitset_->first();
+        bm::bvector<>::enumerator col_end = pipeline->overall_column_bitset_->end();
```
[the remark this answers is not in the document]
vasil-pashov: Yes, see `build_col_read_query_filters` it's the only place this is used.

python/arcticdb/version_store/_store.py
vasil-pashov: I've added tests for negative values in head and tail.

cpp/arcticdb/pipeline/query.hpp
```diff
@@ -312,15 +303,54 @@ inline std::vector<FilterQuery<ContainerType>> build_read_query_filters(
...
+            auto res = std::make_unique<util::BitSet>(static_cast<util::BitSetSizeType>(isr.size()));
+            auto start_row = isr.column(index::Fields::start_row).begin<stream::SliceTypeDescriptorTag>();
+            auto start_row_end = isr.column(index::Fields::start_row).end<stream::SliceTypeDescriptorTag>();
+            size_t index_segment_row = 0;
+            ankerl::unordered_dense::set<decltype(start_row)::value_type> requested_start_rows;
+            while (start_row != start_row_end) {
+                if (requested_start_rows.find(*start_row) == requested_start_rows.end()) {
+                    (*res)[index_segment_row] = true;
+                    requested_start_rows.insert(*start_row);
+                }
```
alexowens90, line 326: It is cheaper to call `insert` and check the second element of the returned pair than to find and then insert

cpp/arcticdb/pipeline/query.hpp
```diff
@@ -312,15 +303,54 @@ inline std::vector<FilterQuery<ContainerType>> build_read_query_filters(
...
+    if (pipeline_context->only_index_columns_selected() && pipeline_context->overall_column_bitset_->count() > 0) {
+        auto query = [pipeline = std::move(pipeline_context)](const index::IndexSegmentReader& isr, std::unique_ptr<util::BitSet>&&) mutable {
+            auto res = std::make_unique<util::BitSet>(static_cast<util::BitSetSizeType>(isr.size()));
+            auto start_row = isr.column(index::Fields::start_row).begin<stream::SliceTypeDescriptorTag>();
+            auto start_row_end = isr.column(index::Fields::start_row).end<stream::SliceTypeDescriptorTag>();
+            size_t index_segment_row = 0;
+            ankerl::unordered_dense::set<decltype(start_row)::value_type> requested_start_rows;
+            while (start_row != start_row_end) {
+                if (requested_start_rows.find(*start_row) == requested_start_rows.end()) {
+                    (*res)[index_segment_row] = true;
```
alexowens90, line 324: I think the bitset has a `set_bit` method that would make this more readable

alexowens90 (changes requested):

alexowens90 (approved):

PR #1572, opened 2024-05-15
build: Use C++20 and dynamically link the C++ standard library

jjerphan:
#### Reference Issues/PRs

Mitigation for https://github.com/ray-project/ray/issues/18506#issuecomment-2088161331.

[`manylinux` wheels (starting `manylinux1`) must depend on external libraries](https://peps.python.org/pep-0513/#the-manylinux1-policy), including the C++ standard library.

Currently, it does, and this might be the cause of problems at runtime with standard librairies shared objects being loaded with duplication as observed with importing ray and ArcticDB.

This PR makes ArcticDB depend on implementations of the standard library dynamically rather than statically. 

This removes the `STATIC_LINK_STD_LIB` option, which is thus not needed anymore.

#### Any other comments?

A few patches are necessary to have ArcticDB compile without using the Docker Image.

Once this is merged, we should be able to reintroduce C++20.


cpp/arcticdb/CMakeLists.txt
```diff
@@ -508,32 +506,18 @@ if (WIN32)
...
+    message("Dynamically linking standard libraries")
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        set(standard_libraries
+            stdc++
+        )
+    elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        set(standard_libraries
+            stdc++
+            gcc
+        )
```
vasil-pashov, line 518: A few questions
1. Why do we have explicitly specify the standard libraries? I think CMake should be able to find them on it's own.
2. What does `gcc` do? I've never seen it.

jjerphan: We need to figure this out (I only have dedented the code).

JohanMabille: - gcc provides a low-level runtime library, see [https://gcc.gnu.org/onlinedocs/gccint/Libgcc.html](https://gcc.gnu.org/onlinedocs/gccint/Libgcc.html)
- I think specifying these libraries was required in the case of static link (because by default cmake chooses the dynamic ones), and I guess it was added in the case of dynamic link for the sake of symmetry ;) This should be removed

alexowens90:
Have you tried switching back to C++20 with these changes and seeing if the issue with Ray is definitely fixed?

cpp/arcticdb/CMakeLists.txt
jjerphan: Yes, but:
 - Why do we need this low-level runtime library?
 - Do we need to stdc++ over libc++ for clang?

jjerphan:
I did, and I can confirm that it resolved the problem on my machine, but I think we should test those new wheels on more systems.

cpp/arcticdb/CMakeLists.txt
jjerphan: [`manylinux2014` specifies `stdlibc++` but does not mention `libc++`](https://peps.python.org/pep-0599/#the-manylinux2014-policy); would it be the explanation?

jjerphan: Resolved by f55ca51.

JohanMabille (approved):

PR #1585, opened 2024-05-28
Fix overflow bug on append and update

IvoDD:
Fixes int overflow bug when appending to a symbol with more than `MAX_INT` rows.

Involves two commits:
- First commit: Adds a stress test which writes `2**32` rows and demonstrates the bug.
- Second commit: Fixes the overflow by using the correct type. Test now passes.

### DIFF
--- a/cpp/arcticdb/pipeline/index_utils.cpp
+++ b/cpp/arcticdb/pipeline/index_utils.cpp
@@ -80,7 +80,7 @@ std::pair<index::IndexSegmentReader, std::vector<SliceAndKey>> read_index_to_vec
 }
 
 TimeseriesDescriptor get_merged_tsd(
-        int row_count,
+        size_t row_count,
         bool dynamic_schema,
         const TimeseriesDescriptor& existing_tsd,
         const std::shared_ptr<pipelines::InputTensorFrame>& new_frame) {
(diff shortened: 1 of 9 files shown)


cpp/arcticdb/pipeline/index_utils.cpp
```diff
@@ -80,7 +80,7 @@ std::pair<index::IndexSegmentReader, std::vector<SliceAndKey>> read_index_to_vec
 }
 
 TimeseriesDescriptor get_merged_tsd(
-        int row_count,
+        size_t row_count,
```
willdealtry, line 83: Can we make it a uint64_t? That's the type it will be in the v2 encoding and is totally unambiguous as to the capacity

willdealtry (approved):

cpp/arcticdb/pipeline/index_utils.cpp
IvoDD: Changed all types referring to a `row_count` to a `unit64_t`. 

This includes the `frame->offset` which used to be signed. As discussed offline, this should be fine since `offset >= 0`.

I did not change the `NativeTensor`'s `ssize_t` and `shape_t=int64_t` as it would have invloved a bigger change. Only left a TODO for these.

IvoDD: I undid the `uint64_t` as I realised I hadn't changed all of the required types ot `uint64_t` and it started breaking some tests because I had a lot of implicit conversions between the two. 

To do this correctly one would need to very carefully go through all places where we refer to a row_count and swap the `size_y` with `uint64_t` so we don't accidentally end up with lots of implicit conversions between the two which would end up making the change meaningless

PR #1606, opened 2024-06-08
Adapt Raise to use format string as message directly if no args are given

muhammadhamzasajjad:
Fixes https://github.com/man-group/arcticdb-man/issues/111

`raise<ErrorCode>()` function takes format string as the first argument. In some places this has been used incorrectly which leads errors like https://github.com/man-group/arcticdb-man/issues/111. 
- Adapt Raise to use format string as message directly if no args are given
- To test this, pass `sym{}` as symbol name
- Show symbol name as part of error message in mock clients. This will test whether `sym{}` is handled correctly.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/util/preconditions.hpp
+++ b/cpp/arcticdb/util/preconditions.hpp
@@ -23,12 +23,12 @@ struct Raise {
 
     template<typename...Args>
     [[noreturn]] void operator()(fmt::format_string<Args...> format, Args&&...args) const {
-        std::string combo_format = fmt::format(FMT_COMPILE("{} {}"), error_code_data<code>.name_, format);
         std::string msg;
         if constexpr(sizeof...(args) == 0) {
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/util/preconditions.hpp
```diff
@@ -24,7 +24,13 @@ struct Raise {
     template<typename...Args>
     [[noreturn]] void operator()(fmt::format_string<Args...> format, Args&&...args) const {
         std::string combo_format = fmt::format(FMT_COMPILE("{} {}"), error_code_data<code>.name_, format);
```
alexowens90, line 26: `combo_format` should be moved inside the else both here and below

alexowens90 (approved):

PR #1616, opened 2024-06-11
Bugfix 1615: Fix segfault in benchmark_clause.cpp

### LINKED ISSUE #1615: Segfault in `benchmark_clause.cpp`

alexowens90:
#### Reference Issues/PRs
Closes #1615 

After the refactor to speed up hash-based grouping, the C++ grouping benchmark needs updating to correctly set the last logical row on the input column, as this is now used to presize the vector mapping rows to buckets.

Also noticed that some benchmarks did not make sense (e.g. `int8_t` column with 100,000 unique grouping values), so corrected this as well.


poodlewars:
How long do the benchmarks take to run? Wonder if we should run them as part of the build to catch breaks. I did this recently in the enterprise repo.

poodlewars (approved):

alexowens90:
> How long do the benchmarks take to run? Wonder if we should run them as part of the build to catch breaks. I did this recently in the enterprise repo.

Seconds, no harm in including them

PR #1617, opened 2024-06-11
Fix sort_and_finalize_staged_data

willdealtry:
sort_and_finalize_staged_data lost some tests in the transition to open-source, and so when it was broken by a processing pipeline factor this was not recognised. This PR fixes it and adds some elementary tests to check that it's doing what it's supposed to

### DIFF
--- a/cpp/arcticdb/processing/clause.hpp
+++ b/cpp/arcticdb/processing/clause.hpp
@@ -109,9 +109,11 @@ struct IClause {
 
 using Clause = folly::Poly<IClause>;
 
-std::vector<std::vector<size_t>> structure_by_row_slice(std::vector<RangesAndKey>& ranges_and_keys,
-                                                        size_t start_from);
+std::vector<std::vector<size_t>> structure_by_row_slice(std::vector<RangesAndKey>& ranges_and_keys, size_t start_from);
+
 std::vector<std::vector<size_t>> structure_by_column_slice(std::vector<RangesAndKey>& ranges_and_keys);
+
+std::vector<std::vector<size_t>> structure_all_together(std::vector<RangesAndKey>& ranges_and_keys);
 
 Composite<ProcessingUnit> gather_entities(std::shared_ptr<ComponentManager> component_manager,
                                           Composite<EntityIds>&& entity_ids,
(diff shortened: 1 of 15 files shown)


DrNickClarke:
I built this locally and ran it against the same tests that I used to report the bug. It works on that use case now.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1285,9 +1285,10 @@ VersionedItem sort_merge_impl(
         [&](const stream::TimeseriesIndex &timeseries_index) {
             read_query.clauses_.emplace_back(std::make_shared<Clause>(SortClause{timeseries_index.name()}));
             read_query.clauses_.emplace_back(std::make_shared<Clause>(RemoveColumnPartitioningClause{}));
-            const auto split_size = ConfigsMap::instance()->get_int("Split.RowCount", 10000);
-            read_query.clauses_.emplace_back(std::make_shared<Clause>(SplitClause{static_cast<size_t>(split_size)}));
-            read_query.clauses_.emplace_back(std::make_shared<Clause>(MergeClause{timeseries_index, DenseColumnPolicy{}, stream_id, pipeline_context->descriptor()}));
+            //const auto split_size = ConfigsMap::instance()->get_int("Split.RowCount", 10000);
+            //read_query.clauses_.emplace_back(std::make_shared<Clause>(SplitClause{static_cast<size_t>(split_size)}));
+
+            read_query.clauses_.emplace_back(std::make_shared<Clause>(MergeClause{timeseries_index, SparseColumnPolicy{}, stream_id, pipeline_context->descriptor()}));
```
alexowens90, line 1291: Should we add a test case with sparse data using `sparsify_floats` if we're going to use `SparseColumnPolicy`?

cpp/arcticdb/processing/clause.hpp
```diff
@@ -614,8 +616,8 @@ struct MergeClause {
 
     [[nodiscard]] std::vector<std::vector<size_t>> structure_for_processing(
             std::vector<RangesAndKey>& ranges_and_keys,
-            size_t start_from) {
-        return structure_by_row_slice(ranges_and_keys, start_from);
+            size_t) {
+        return structure_all_together(ranges_and_keys);
```
alexowens90, line 620: Do we need this? `structure_for_processing` is only called on the first clause in the pipeline, which should never be `MergeClause`?

alexowens90 (approved):

cpp/arcticdb/processing/clause.hpp
willdealtry: It is conceivable that we might want to take a bunch of sorted things and merge them all together

cpp/arcticdb/version/version_core.cpp
willdealtry: I don't think there is a sparsify_floats with staged writes is there? The SparseColumnPolicy is about the output, not the input.

PR #1618, opened 2024-06-11
Bugfix 1612: Raise a meaningful exception type/message when an empty string is provided as a symbol/snapshot name

### LINKED ISSUE #1612: Useful error message when empty string provided as symbol name
Currently get logs like:
```
ASSERTION FAILURE: Empty string id in reference key
Loading versions from storage via ref key failed with error: Empty string id in reference key for stream . Retrying
ASSERTION FAILURE: Empty string id in reference key
Loading versions from storage via ref key failed with error: Empty string id in reference key for stream . Retrying
```

but the actual exception thrown is `ASSERTION FAILURE: Couldn't read via ref key even after multiple attempts`, which does not show the underlying cause.

Should just involve checking for empty strings in `verify_name` in `name_validation.cpp`, but need to carefully check that this will not break any other use cases of this function.

alexowens90:
#### Reference Issues/PRs
Closes #1612

### DIFF
--- a/python/tests/integration/arcticdb/version_store/test_basic_version_store.py
+++ b/python/tests/integration/arcticdb/version_store/test_basic_version_store.py
@@ -105,19 +105,6 @@ def test_s3_breaking_chars(object_version_store, breaking_char):
     assert sym not in object_version_store.list_symbols()
 
 
-def test_s3_breaking_chars_exception_compat(object_version_store):
-    """Test that chars that are not supported are raising the appropriate exception and that we fail on write without
-    corrupting the db
-    """
-    sym = "prefix*postfix"
-    df = sample_dataframe()
-    with pytest.raises(UserInputException) as e_info:
-        object_version_store.write(sym, df)
-
-    assert isinstance(e_info.value, UserInputException)
-    assert sym not in object_version_store.list_symbols()
-
-
 @pytest.mark.parametrize("prefix", ["", "prefix"])
 @pytest.mark.parametrize("suffix", ["", "suffix"])
 def test_symbol_names_with_all_chars(object_version_store, prefix, suffix):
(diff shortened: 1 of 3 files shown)


python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -105,19 +105,6 @@ def test_s3_breaking_chars(object_version_store, breaking_char):
     assert sym not in object_version_store.list_symbols()
 
 
-def test_s3_breaking_chars_exception_compat(object_version_store):
```
poodlewars, line 108: Why have you removed this?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -215,6 +202,66 @@ def test_unhandled_chars_already_present_update(object_version_store, unhandled_
     assert len(vitem.data) == len(new_vitem.data)
 
 
+# See AN-765 for why we need no_symbol_list fixture
+def test_large_symbols(basic_store_no_symbol_list):
+    # The following restrictions should be checked in the cpp layer's name_validation
+    MAX_SYMBOL_SIZE = 254
```
poodlewars, line 208: Why has this changed from 255?

poodlewars (approved):

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
alexowens90: It is a strict subset of the test above

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
alexowens90: This is the actual max symbol size:
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/entity/serialized_key.hpp#L87-L93
The old implementation of this test had to use `MAX_SYMBOL_SIZE - 1` as the max size to work

PR #1624, opened 2024-06-13
Enhancement/1610/method to force compact symbol list

### LINKED ISSUE #1610: Add method to forcibly compact the symbol list, and throw a useful error message if it fails
`list_symbols` compacts by default if there are too many symbol list cache keys and the user has write permissions on the library. If this compaction fails for any reason though, the call will still succeed and return the correct result, so there is no way to identify a failed compaction right now.

alexowens90:
#### Reference Issues/PRs
Closes #1610 

Considered augmenting `list_symbols` API to make this an option, but there are already a lot of arguments (particularly in the `NativeVersionStore` version) that this would not be compatible with, so added another method instead.

### DIFF
--- a/cpp/arcticdb/util/storage_lock.hpp
+++ b/cpp/arcticdb/util/storage_lock.hpp
@@ -116,7 +116,7 @@ class StorageLock {
         OnExit x{[that=this] () {
             that->mutex_.unlock();
         }};
-        if(!ref_key_exists(store) || ttl_expired(store)) {
+        if(!ref_key_exists(store) || !ttl_not_expired(store)) {
             ts_= create_ref_key(store);
             auto lock_sleep = ConfigsMap::instance()->get_int("StorageLock.WaitMs", 200);
             std::this_thread::sleep_for(std::chrono::milliseconds(lock_sleep));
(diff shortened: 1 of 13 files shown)


cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -116,7 +116,7 @@ class StorageLock {
         OnExit x{[that=this] () {
             that->mutex_.unlock();
         }};
-        if(!ref_key_exists(store) || ttl_expired(store)) {
+        if(!ref_key_exists(store) || !ttl_not_expired(store)) {
```
vasil-pashov, line 119: I find double negatives hard to read.

willdealtry: If I were refactoring I would probably replace both sides of the || with a single-line method whose name explains what is being checked `if(no_lock_object_present() || lock_object_within_ttl()`

willdealtry (approved):

python/arcticdb/version_store/library.py
```diff
@@ -1706,6 +1706,20 @@ def reload_symbol_list(self):
...
+
+
+        Raises
+        ------
+        PermissionException
+            Library has been opened in read-only mode
+        InternalException
+            Storage lock required to compact the symbol list could not be acquired
+        """
+        self._nvs.compact_symbol_list()
```
jamesmunro, line 1721: Can it return something to indicate whether it did anything to storage, or how much compaction it did?

alexowens90: The method will always compact the symbol list cache down to a single key or throw. I've updated to return the number of symbol list keys prior to the compaction though

PR #1640, opened 2024-06-24
Bugfix 1637: append does not respect prune previous

### LINKED ISSUE #1637: `append` does not respect prune previous
This is also the case with the legacy project inside Man Group.

Should check that all other modification methods respect this option as well.

alexowens90:
Fixes #1637 

Also fixes some other bugs and inconsistencies in our handling of prune previous:

- `update`, `delete` with a `date_range`, and `defragment_symbol_data` also did not respect the prune previous lib config setting/env var.
- `delete` with a `date_range` argument and `defragment_symbol_data` used `prune_previous_versions` (i.e. with a trailing `s`). All other methods use `prune_previous_version` (i.e. without the trailing `s`). For backwards compatibility, these two methods now support both variants.
- `write_metadata` and `delete_data_in_range` on the V2 API did not have a `prune_previous_versions` argument.
- `write_metadata_batch` on the V2 API defaulted the argument to `None` (implying respecting the lib config setting, but this is not a library option on the V2 API), where everything else defaults to `False`.
- Homogenised docstrings for `prune_previous_versions` on V2 API.

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -617,7 +617,7 @@ class NativeVersionStore:
         dataframe: TimeSeriesType,
         metadata: Optional[Any] = None,
         incomplete: bool = False,
-        prune_previous_version: bool = False,
+        prune_previous_version: Optional[bool] = None,
         validate_index: bool = False,
         **kwargs,
     ) -> Optional[VersionedItem]:
@@ -691,6 +691,11 @@ class NativeVersionStore:
         dynamic_strings = self._resolve_dynamic_strings(kwargs)
         coerce_columns = kwargs.get("coerce_columns", None)
 
+        proto_cfg = self._lib_cfg.lib_desc.version.write_options
+        prune_previous_version = self.resolve_defaults(
+            "prune_previous_version", proto_cfg, global_default=False, existing_value=prune_previous_version, **kwargs
+        )
+
         _handle_categorical_columns(symbol, dataframe)
 
         udm, item, norm_meta = self._try_normalize(
@@ -722,7 +727,7 @@ class NativeVersionStore:
         metadata: Any = None,
         date_range: Optional[DateRangeInput] = None,
         upsert: bool = False,
-        prune_previous_version: bool = False,
+        prune_previous_version: Optional[bool] = None,
         **kwargs,
     ) -> VersionedItem:
         """
@@ -785,10 +790,15 @@ class NativeVersionStore:
         """
         update_query = _PythonVersionStoreUpdateQuery()
         dynamic_strings = self._resolve_dynamic_strings(kwargs)
+        proto_cfg = self._lib_cfg.lib_desc.version.write_options
         dynamic_schema = self.resolve_defaults(
-            "dynamic_schema", self._lib_cfg.lib_desc.version.write_options, False, **kwargs
+            "dynamic_schema", proto_cfg, False, **kwargs
         )
         coerce_columns = kwargs.get("coerce_columns", None)
+
+        prune_previous_version = self.resolve_defaults(
+            "prune_previous_version", proto_cfg, global_default=False, existing_value=prune_previous_version, **kwargs
+        )
 
         if date_range is not None:
             start, end = normalize_dt_range_to_ts(date_range)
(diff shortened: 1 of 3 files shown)


python/arcticdb/version_store/_store.py
```diff
@@ -688,6 +688,11 @@ def append(
         dynamic_strings = self._resolve_dynamic_strings(kwargs)
         coerce_columns = kwargs.get("coerce_columns", None)
 
+        proto_cfg = self._lib_cfg.lib_desc.version.write_options
+        prune_previous_version = self.resolve_defaults(
```
vasil-pashov, line 692: Is this variable used?

python/arcticdb/version_store/library.py
```diff
@@ -1223,16 +1228,18 @@ def write_metadata(self, symbol: str, metadata: Any) -> VersionedItem:
             Symbol name for the item
         metadata
             Metadata to persist along with the symbol
+        prune_previous_versions : `Optional[bool]`, default=False
+            Removes previous (non-snapshotted) versions from the database.
```
jamesmunro, line 1232: Can we make it clear, perhaps warn, that this removes previous data, not just metadata.

python/arcticdb/version_store/library.py
```diff
@@ -1245,8 +1252,8 @@ def write_metadata_batch(
         ----------
         write_metadata_payloads : `List[WriteMetadataPayload]`
             Symbols and their corresponding metadata. There must not be any duplicate symbols in `payload`.
-        prune_previous_versions : `Optional[bool]`, default=None
-            Remove previous versions from version list. Uses library default if left as None.
+        prune_previous_versions : `Optional[bool]`, default=False
+            Removes previous (non-snapshotted) versions from the database.
```
jamesmunro, line 1256: As above,
> Can we make it clear, perhaps warn, that this removes previous data, not just metadata.

python/arcticdb/version_store/_store.py
alexowens90: Yes

python/arcticdb/version_store/_store.py
```diff
@@ -2805,10 +2830,20 @@ def defragment_symbol_data(self, symbol: str, segment_size: Optional[int] = None
...
+        if proto_cfg.bucketize_dynamic:
             raise ArcticDbNotYetImplemented(f"Support for library with 'bucketize_dynamic' ON is not implemented yet")
 
-        result = self.version_store.defragment_symbol_data(symbol, segment_size, prune_previous_versions)
+        # All other methods use prune_previous_version, but also support prune_previous_versions here in case
+        # anyone is relying on it
+        if prune_previous_versions is not None:
+            prune_previous_version = prune_previous_versions
+        else:
+            prune_previous_version = self.resolve_defaults(
```
poodlewars, line 2842: The `existing_value` arg on `resolve_defaults` seems to be designed for this purpose, rather than the `if is not None` logic up in its caller

python/arcticdb/version_store/library.py
```diff
@@ -948,7 +948,7 @@ def sort_and_finalize_staged_data(
             new timeseries. Append collects the staged data and appends them to the latest version.
 
         prune_previous_versions : `Optional[bool]`, default=False
```
poodlewars, line 950: The type hint is `Optional[bool]` on some methods and `bool` on others. I think it should be `bool` throughout.

python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
```diff
@@ -294,4 +296,117 @@ def test_date_range_multi_index(lmdb_version_store):
...
+    ("write", "append", "update", "write_metadata", "batch_write", "batch_append", "batch_write_metadata")
+)
+@pytest.mark.parametrize("lib_config", (True, False))
+@pytest.mark.parametrize("env_var", (True, False))
+@pytest.mark.parametrize("arg", (True, False, None))
+def test_prune_previous_general(version_store_factory, method, lib_config, env_var, arg):
+    lib = version_store_factory(prune_previous_version=lib_config, use_tombstones=True)
+    should_be_pruned = lib_config
+    if env_var:
+        os.environ["PRUNE_PREVIOUS_VERSION"] = "true"
```
poodlewars, line 313: `monkeypatch.setenv` to make sure we don't interfere with other tests

alexowens90: Done

python/arcticdb/version_store/library.py
alexowens90: Made `bool` everywhere

python/arcticdb/version_store/_store.py
alexowens90: Fair, I'd confused myself over `prune_previous_versions` vs `prune_previous_version`. Used `existing_value` now

poodlewars (approved):

Issue #1327, opened 2024-02-13
Storage Exceptions: Throw Storage Exception when a key_exists / read is attempted after clearing the storage.


muhammadhamzasajjad:
In `details-inl.hpp` S3Storage currently swallows `Aws::S3::S3Errors::NO_SUCH_BUCKET`  which happens when storage is cleared and a key_exists / read is attempted. `NO_SUCH_BUCKET` shouldn't be swallowed, S3 should throw an exception. This was attempted in #1304, but if an exception is thrown, `/python/tests/integration/arcticdb/test_arctic_move_storage.py::test_move_storage` test fails as lmdb and azure don't throw an exception in this case. This, therefore, requires changes in multiple storages and should be done after all the exceptions have been handled.

muhammadhamzasajjad:
Upon further investigation I found that the error `Aws::S3::S3Errors::NO_SUCH_BUCKET` and `AzureErrorCode::ContainerNotFound` actually have to be swallowed. When a new arctic instance is created, `ac.list_libraries()` should return an empty list. If we don't swallow this error, list_libraries calls iterate_type which would throw an exception and this would cause inconsistency with other storage + a incorrect behaviour. Therefore, the error is swallowed in `iterate_type`.

PR #1648, opened 2024-06-26
NO_SUCH_BUCKET and RESOURCE_UNAVAILABLE are not expected error types

willdealtry:
We should not treat resource unavailable or no such bucket as expected S3 errors, as we might incorrectly decide that an object doesn't exist and can be newly created when in fact we've just not managed to talk to the storage to ascertain whether it's there or not.

### DIFF
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -47,9 +47,15 @@ namespace s3 {
         inline void raise_s3_exception(const Aws::S3::S3Error& err){
             std::string error_message;
             auto type = err.GetErrorType();
-            // s3_client.HeadObject returns RESOURCE_NOT_FOUND if a key is not found.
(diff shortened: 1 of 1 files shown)


poodlewars (approved):

muhammadhamzasajjad:
We shouldn't remove `RESOURCE_NOT_FOUND`. As mentioned in [this issue](https://github.com/aws/aws-sdk-cpp/issues/345), `RESOURCE_NOT_FOUND` is returned when you call `HeadObject()` and the key is not found. You can also see this in failing tests in the CI. This came up during my exception normalization PRs (I guess I should've left a comment explaining why `RESOURCE_NOT_FOUND` is swallowed).

As per `NO_SUCH_BUCKET`, it is kept as per what is found [here](https://github.com/man-group/ArcticDB/issues/1327#issuecomment-2039523978). Although, we might be able remove this with some more work.

jamesmunro (changes requested):
Please can we have a representative test?  Something that shows the full issue and how this solves that.

Given the confusion about how these errors are used by AWS S3 SDK, I think it deserves a comment in the code as well please.  Thanks.

PR #1650, opened 2024-06-26
Make modification of snapshot atomic and corresponding tombstone handling changes

phoebusm:
#### Reference Issues/PRs

Originally only for forward porting https://mangit.maninvestments.com/projects/DATA/repos/arcticc/pull-requests/1095/overview
But corresponding test requires https://github.com/man-group/arcticdb-enterprise/issues/32 so part of changes is included too.

1. Modification of snapshot will not longer delete the original `SNAPSHOT_REF` key. It will be simply overwritten so atomicity can be achieved
2. As the changes will affect op logs as well and the test cannot be made without https://github.com/man-group/arcticdb-enterprise/issues/32 so some corresponding changes for that ticket as well. More details can be found there.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/async/tasks.hpp
+++ b/cpp/arcticdb/async/tasks.hpp
@@ -514,24 +514,17 @@ struct DecodeMetadataAndDescriptorTask : BaseTask {
             );
     }
 };
-
-template<typename ConstVarKeyGetter,
-        typename=std::enable_if_t<!std::is_reference_v<ConstVarKeyGetter> // Make obvious if iterator& is captured
-                        && std::is_same_v<decltype(*std::declval<ConstVarKeyGetter>()), const entity::VariantKey&>>>
 struct KeyExistsTask : BaseTask {
-    ConstVarKeyGetter key_;
+    const VariantKey key_;
     std::shared_ptr<storage::Library> lib_;
 
-    /**
-     * @param key Could be a vector iterator or a pointer to a const VariantKey.
-     */
-    KeyExistsTask(ConstVarKeyGetter key, std::shared_ptr<storage::Library> lib): key_(key), lib_(std::move(lib)) {
-        ARCTICDB_DEBUG(log::storage(), "Creating key exists task for key {}", variant_key_view(*key_));
+    KeyExistsTask(auto &&key, std::shared_ptr<storage::Library> lib): key_(std::forward<decltype(key)>(key)), lib_(std::move(lib)) {
+        ARCTICDB_DEBUG(log::storage(), "Creating key exists task for key {}",key_);
     }
 
     bool operator()() {
         ARCTICDB_SAMPLE(KeyExistsTask, 0)
-        return lib_->key_exists(*key_);
+        return lib_->key_exists(key_);
     }
 };
 
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -75,7 +75,7 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_unreferenced_pruned_inde
                     std::move(pruned_indexes));
             in_snaps.insert(key_to_keep);
             PreDeleteChecks checks{false, false, false, false, std::move(in_snaps)};
-            return delete_trees_responsibly(not_in_snaps, {}, {}, checks)
+            return delete_trees_responsibly(store(), version_map(), not_in_snaps, {}, {}, checks)
                     .thenError(folly::tag_t<std::exception>{}, [](auto const& ex) {
                         log::version().warn("Failed to clean up pruned previous versions due to: {}", ex.what());
                     });
@@ -818,7 +818,9 @@ std::unordered_map<StreamId, VersionId> min_versions_for_each_stream(const std::
     return out;
 }
 
-folly::Future<folly::Unit> LocalVersionedEngine::delete_trees_responsibly(
+folly::Future<folly::Unit> delete_trees_responsibly(
+        std::shared_ptr<Store> store,
+        std::shared_ptr<VersionMap> &version_map,
         const std::vector<IndexTypeKey>& orig_keys_to_delete,
         const arcticdb::MasterSnapshotMap& snapshot_map,
         const std::optional<SnapshotId>& snapshot_being_deleted,
@@ -870,7 +872,7 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_trees_responsibly(
                 auto load_param = load_type == LoadType::LOAD_DOWNTO
                         ? LoadParameter{load_type, static_cast<SignedVersionId>(min.second)}
                         : LoadParameter{load_type};
-                const auto entry = version_map()->check_reload(store(), min.first, load_param, __FUNCTION__);
+                const auto entry = version_map->check_reload(store, min.first, load_param, __FUNCTION__);
                 entry_map.try_emplace(std::move(min.first), entry);
             }
         }
@@ -904,11 +906,11 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_trees_responsibly(
 
     storage::ReadKeyOpts read_opts;
     read_opts.ignores_missing_key_ = true;
-    auto data_keys_to_be_deleted = recurse_index_keys(store(), *keys_to_delete, read_opts);
+    auto data_keys_to_be_deleted = recurse_index_keys(store, *keys_to_delete, read_opts);
     log::version().debug("Candidate: {} total of data keys", data_keys_to_be_deleted.size());
 
     read_opts.dont_warn_about_missing_key = true;
-    auto data_keys_not_to_be_deleted = recurse_index_keys(store(), *not_to_delete, read_opts);
+    auto data_keys_not_to_be_deleted = recurse_index_keys(store, *not_to_delete, read_opts);
     not_to_delete.clear();
     log::version().debug("Forbidden: {} total of data keys", data_keys_not_to_be_deleted.size());
     storage::RemoveOpts remove_opts;
@@ -939,14 +941,14 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_trees_responsibly(
     folly::Future<folly::Unit> remove_keys_fut;
     if (!dry_run) {
         // Delete any associated column stats keys first
-        remove_keys_fut = store()->remove_keys(std::move(vks_column_stats), remove_opts)
-        .thenValue([this, vks_to_delete = std::move(vks_to_delete), remove_opts](auto&& ) mutable {
+        remove_keys_fut = store->remove_keys(std::move(vks_column_stats), remove_opts)
+        .thenValue([store=store, vks_to_delete = std::move(vks_to_delete), remove_opts](auto&& ) mutable {
             log::version().debug("Column Stats keys deleted.");
-            return store()->remove_keys(std::move(vks_to_delete), remove_opts);
+            return store->remove_keys(std::move(vks_to_delete), remove_opts);
         })
-        .thenValue([this, vks_data_to_delete = std::move(vks_data_to_delete), remove_opts](auto&&) mutable {
+        .thenValue([store=store, vks_data_to_delete = std::move(vks_data_to_delete), remove_opts](auto&&) mutable {
             log::version().debug("Index keys deleted.");
-            return store()->remove_keys(std::move(vks_data_to_delete), remove_opts);
+            return store->remove_keys(std::move(vks_data_to_delete), remove_opts);
         })
         .thenValue([](auto&&){
             log::version().debug("Data keys deleted.");
@@ -955,7 +957,6 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_trees_responsibly(
     }
     return remove_keys_fut;
 }
-
 
 void LocalVersionedEngine::remove_incomplete(
     const StreamId& stream_id
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -57,6 +57,16 @@ struct SortMergeOptions {
     bool prune_previous_versions_;
 };
 
+folly::Future<folly::Unit> delete_trees_responsibly(
+    std::shared_ptr<Store> store,
+    std::shared_ptr<VersionMap> &version_map,
+    const std::vector<IndexTypeKey>& orig_keys_to_delete,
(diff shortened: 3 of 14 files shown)


phoebusm:
Depended by https://github.com/man-group/arcticdb-enterprise/pull/133

cpp/arcticdb/async/tasks.hpp
```diff
@@ -492,24 +492,54 @@ struct DecodeMetadataAndDescriptorTask : BaseTask {
...
+
+template<typename T, typename U>
+using is_same_cvref_removed = std::is_same<std::remove_cvref_t<T>, U>;
+
+template<typename T,
+        typename=std::enable_if_t< // Make obvious if iterator& is captured
+                        is_same_cvref_removed<T, entity::VariantKey>::value
+                        || is_iter_of<T, entity::VariantKey>::value
+                        || is_pointer_of<T, entity::VariantKey>::value>>
 struct KeyExistsTask : BaseTask {
```
poodlewars, line 517: The old implementation was going to great lengths to ensure that we are never copying the key in to the `KeyExistsTask`, why do we need to relax that constraint now??

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -400,6 +394,7 @@ class LocalVersionedEngine : public VersionedEngine {
     }
     void _test_set_store(std::shared_ptr<Store> store);
     std::shared_ptr<VersionMap> _test_get_version_map();
+    SymbolList& _test_get_symbol_list();
```
poodlewars, line 397: Return by value to avoid confusion if this is just for testing?

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -37,6 +37,16 @@ enum class BatchGetVersionOption {
     COUNT
 };
 
+inline std::optional<std::string> collect_futures_exceptions(auto futures) {
```
poodlewars, line 40: Move the futures in?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -371,10 +371,8 @@ void PythonVersionStore::add_to_snapshot(
         retained_keys.emplace_back(std::move(key));
 
     std::sort(std::begin(retained_keys), std::end(retained_keys));
-    if(variant_key_type(snap_key) == KeyType::SNAPSHOT_REF && cfg().write_options().delayed_deletes()) {
-        tombstone_snapshot(store(), to_ref(snap_key), std::move(snap_segment), version_map()->log_changes());
-    } else {
-        delete_trees_responsibly(deleted_keys, get_master_snapshots_map(store()), snap_name).get();
+    if(variant_key_type(snap_key) != KeyType::SNAPSHOT_REF || !cfg().write_options().delayed_deletes()) {
```
poodlewars, line 374: The arcticc version with `bool is_delete_keys_immediately` was clearer

poodlewars, line 374: I know it was here already, but this check,

```
variant_key_type(snap_key) != KeyType::SNAPSHOT_REF
```

is really confusing. Don't we already know what type `snap_key` is??

poodlewars:
This change requires a test in this repo. I think that I did something similar a while ago for vref atomicity - https://github.com/man-group/ArcticDB/blob/master/python/tests/stress/arcticdb/version_store/test_concurrent_read_and_write.py

poodlewars (changes requested):

cpp/arcticdb/version/local_versioned_engine.hpp
phoebusm: For the testing only in the other repo

cpp/arcticdb/async/tasks.hpp
phoebusm: In the old implementation, the task itself doesn't hold the memory during construction. This could be a ticking timebomb (at least I hit it) as the memory could no longer be valid by the time the task is run.
IMO in the worker-pool model, every task should be self-sufficient - not depending on memory resides outside the class

cpp/arcticdb/version/version_store_api.cpp
phoebusm: Hmmm got it.... Let me re-align my style

phoebusm: I had the same question when I was working on it. My guess is for backward compatibility with `KeyType::SNAPSHOT`:
https://github.com/man-group/ArcticDB/blob/3adbed459f6046d79b3a61264e2a5b5b62cc3ca8/cpp/arcticdb/version/snapshot.cpp#L214

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -358,23 +358,24 @@ void PythonVersionStore::add_to_snapshot(
...
+            }
         }
     }
 
     for(auto&& [id, key] : *specific_versions_index_map)
         retained_keys.emplace_back(std::move(key));
 
     std::sort(std::begin(retained_keys), std::end(retained_keys));
-    if(variant_key_type(snap_key) == KeyType::SNAPSHOT_REF && cfg().write_options().delayed_deletes()) {
-        tombstone_snapshot(store(), to_ref(snap_key), std::move(snap_segment), version_map()->log_changes());
```
alexowens90, line 375: When do we tombstone snapshots now?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -358,23 +358,24 @@ void PythonVersionStore::add_to_snapshot(
...
-    } else {
-        delete_trees_responsibly(deleted_keys, get_master_snapshots_map(store()), snap_name).get();
+    if(is_delete_keys_immediately) {
+        delete_trees_responsibly(store(), version_map(), deleted_keys, get_master_snapshots_map(store()), snap_name).get();
         if (version_map()->log_changes()) {
```
alexowens90, line 379: Shouldn't this be outside the `if(is_delete_keys_immediately)` block?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -402,20 +403,21 @@ void PythonVersionStore::remove_from_snapshot(
...
         }
     }
 
-    if(variant_key_type(snap_key) == KeyType::SNAPSHOT_REF && cfg().write_options().delayed_deletes()) {
-        tombstone_snapshot(store(), to_ref(snap_key), std::move(snap_segment), version_map()->log_changes());
-    } else {
-        delete_trees_responsibly(deleted_keys, get_master_snapshots_map(store()), snap_name).get();
+    if(is_delete_keys_immediately) {
+        delete_trees_responsibly(store(), version_map(), deleted_keys, get_master_snapshots_map(store()), snap_name).get();
         if (version_map()->log_changes()) {
```
alexowens90, line 421: Same questions as above on tombstoning and logging changes

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -57,6 +57,16 @@ struct SortMergeOptions {
     bool prune_previous_versions_;
 };
 
+folly::Future<folly::Unit> delete_trees_responsibly(
```
alexowens90, line 60: Why did this move?

cpp/arcticdb/version/local_versioned_engine.hpp
alexowens90: Agree with Alex, returning by reference is unnecessarily dangerous for test code

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -895,11 +897,11 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_trees_responsibly(
 
     storage::ReadKeyOpts read_opts;
     read_opts.ignores_missing_key_ = true;
-    auto data_keys_to_be_deleted = get_data_keys_set(store(), *keys_to_delete, read_opts);
+    auto data_keys_to_be_deleted = get_data_keys_set(store, *keys_to_delete, read_opts);
```
alexowens90, line 900: I've renamed this function in #1657, whoever merges second will need to rebase

cpp/arcticdb/async/tasks.hpp
```diff
@@ -492,24 +492,54 @@ struct DecodeMetadataAndDescriptorTask : BaseTask {
...
 
     /**
-     * @param key Could be a vector iterator or a pointer to a const VariantKey.
+     * @param key Could be a vector iterator, a pointer or a value to a const VariantKey.
      */
-    KeyExistsTask(ConstVarKeyGetter key, std::shared_ptr<storage::Library> lib): key_(key), lib_(std::move(lib)) {
-        ARCTICDB_DEBUG(log::storage(), "Creating key exists task for key {}", variant_key_view(*key_));
+    KeyExistsTask(T key, std::shared_ptr<storage::Library> lib): key_(key), lib_(std::move(lib)) {
+        // ARCTICDB_DEBUG(log::storage(), "Creating key exists task for key {}", variant_key_view(*key_));
+        ARCTICDB_DEBUG(log::storage(), "Creating key exists task for key {}",
```
alexowens90, line 526: This is broken

cpp/arcticdb/async/tasks.hpp
alexowens90: This is still dangerous, it is allowing iterators into vectors that can go out of scope and raw pointers. To ensure liveness of the key memory, only accept rvalue references and shared pointers

cpp/arcticdb/version/version_store_api.cpp
phoebusm: Because of https://github.com/man-group/arcticdb-enterprise/pull/133

cpp/arcticdb/version/local_versioned_engine.hpp
phoebusm: For https://github.com/man-group/arcticdb-enterprise/pull/133

cpp/arcticdb/version/local_versioned_engine.hpp
phoebusm: It is the last resort for https://github.com/man-group/arcticdb-enterprise/blob/2af2e32dc3321b28f03a9167bc0d40092f22e7a8/cpp/arcticdb-enterprise/background_delete/test/test_deletion.cpp#L74

cpp/arcticdb/version/version_store_api.cpp
phoebusm: I don't think so. Doing so will make the snapshot modification in replication process stays non-atomic

cpp/arcticdb/version/version_store_api.cpp
phoebusm: I don't think so. Doing so will make the snapshot modification in replication process stays non-atomic

cpp/arcticdb/version/local_versioned_engine.cpp
phoebusm: I'll rebase it once the PR is approved

cpp/arcticdb/async/tasks.hpp
phoebusm: I am not sure what do you mean?

cpp/arcticdb/async/tasks.hpp
phoebusm: I am happy to do so if there is no concern for performance. It can simplify the mad template parameter checking in this class.
I thought it is designed to accept iterator and shared pointer for a reason.
My changes is only for making the class store the key locally if a lvalue key is passed

alexowens90: Sorry, yes, copying lvalues is also acceptable from a correctness perspective. I haven't seen this level of madness in any similar methods though, how do they solve this problem? I've definitely seen overloaded constructors that copy lvalues or move rvalues, are there other approaches?

cpp/arcticdb/version/local_versioned_engine.hpp
alexowens90: Actually, why do we need this at all? Just call `symbol_list()` directly from the test?

cpp/arcticdb/async/tasks.hpp
phoebusm: The template has been removed as `KeyExistsTask` now store copy of key, per discussed

cpp/arcticdb/version/local_versioned_engine.hpp
phoebusm: The dependency is removed by turning off the `symbol_list` in the test

cpp/arcticdb/async/tasks.hpp
```diff
@@ -492,24 +492,18 @@ struct DecodeMetadataAndDescriptorTask : BaseTask {
...
-    ConstVarKeyGetter key_;
+    const VariantKey key_;
     std::shared_ptr<storage::Library> lib_;
 
-    /**
-     * @param key Could be a vector iterator or a pointer to a const VariantKey.
-     */
-    KeyExistsTask(ConstVarKeyGetter key, std::shared_ptr<storage::Library> lib): key_(key), lib_(std::move(lib)) {
-        ARCTICDB_DEBUG(log::storage(), "Creating key exists task for key {}", variant_key_view(*key_));
+    KeyExistsTask(auto &&key, std::shared_ptr<storage::Library> lib): key_(std::forward<decltype(key)>(key)), lib_(std::move(lib)) {
```
alexowens90, line 499: The advantage of `auto&&` is eliminated by the fact that all of the uses of this ctor pass by `const&`. Please go through the usages of `key_exists`, `key_exists_sync`, and `batch_key_exists` (including in the enterprise repo) and see if there are any usages which could take advantage of move semantics.

cpp/arcticdb/async/tasks.hpp
```diff
@@ -492,24 +492,18 @@ struct DecodeMetadataAndDescriptorTask : BaseTask {
...
+        // ARCTICDB_DEBUG(log::storage(), "Creating key exists task for key {}", variant_key_view(*key_));
```
alexowens90, line 500: Remove

alexowens90 (approved):

cpp/arcticdb/async/tasks.hpp
phoebusm: I have just done another sweep and replace as much as I can.

poodlewars (approved):

PR #1657, opened 2024-07-03
Refactor/1643/reduce prune previous memory footprint

### LINKED ISSUE #1643: Reduce memory footprint of `prune_previous_versions`
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/util/key_utils.hpp#L88-L115
Currently, generates a vector of all data keys that can potentially be deleted, then collapses that vector to a hash set.
In the case where a symbol is being appended to constantly, this results in a large amount of duplication in the vector.
Materialising `AtomKey`s is also generally memory hungry, and not needed in this case.

alexowens90:
Closes #1643 

Vastly reduces the runtime and memory footprint of pruning lots of old versions of a symbol. Needed because of #1637

Performance differences tested with:
```
def test_prune_previous_memory_usage(lmdb_version_store_very_big_map):
    lib = lmdb_version_store_very_big_map
    sym = "test_prune_previous_memory_usage"
    num_versions = 8000
    for idx in range(num_versions):
        lib.append(sym, pd.DataFrame({"col": np.arange(idx, idx+1)}), write_if_missing=True)
    assert len(lib.list_versions(sym)) == num_versions
    gc.collect()
    start = time.time()
    lib.prune_previous_versions(sym)
    print(f"Prune took {time.time() - start}s")
    assert len(lib.list_versions(sym)) == 1
``` 
Results:
```
                Max RSS      Time
Before change     8.7GB     25.9s
 After change     1.3GB      4.1s
```

### DIFF
--- a/cpp/arcticdb/column_store/key_segment.cpp
+++ b/cpp/arcticdb/column_store/key_segment.cpp
@@ -1,0 +1,153 @@
+/* Copyright 2024 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#include <arcticdb/column_store/key_segment.hpp>
+#include <arcticdb/pipeline/index_fields.hpp>
+
+namespace arcticdb {
+
+StreamId stream_id_from_column_entry(uint64_t column_entry, bool is_sequence_type, const StringPool& string_pool) {
+    if (is_sequence_type) {
+        // String columns are stored as uint64s (offsets into the string pool), but StringPool::get_const_view expects an int64
+        // Hence the static cast
+        return StringId(string_pool.get_const_view(static_cast<int64_t>(column_entry)));
+    } else {
+        return NumericId(column_entry);
+    }
+}
+
+IndexValue index_value_from_column_entry(int64_t column_entry, bool is_sequence_type, const StringPool& string_pool) {
+    if (is_sequence_type) {
+        // String columns are stored as uint64s, whereas timestamp index start/end values are stored as int64s (nanoseconds since epoch)
+        // We could add another if statement based on index_is_sequence_type, but as these types are the same width,
+        // we can just do some fancy casting to reinterpret the value in the iterator as a uint64.
+        // We then need to static cast back to int64 for the same reason as above RE StringPool::get_const_view
+        return StringId(string_pool.get_const_view(static_cast<int64_t>(*reinterpret_cast<const uint64_t*>(&(column_entry)))));
+    } else {
+        return NumericIndex(column_entry);
+    }
+}
+
+KeySegment::KeySegment(SegmentInMemory&& segment, SymbolStructure symbol_structure):
+        num_keys_(segment.row_count()),
+        string_pool_(segment.string_pool_ptr()),
+        symbol_structure_(symbol_structure)
+{
+    // Needed as the column map is not initialised at read time if the segment has no rows
+    segment.init_column_map();
+    stream_ids_ = segment.column_ptr(static_cast<uint32_t>(pipelines::index::Fields::stream_id));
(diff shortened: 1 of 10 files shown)


cpp/arcticdb/column_store/key_segment.cpp
```diff
@@ -0,0 +1,146 @@
...
+                    symbol_ = safe_convert_to_numeric_id(stream_ids_->template reference_at<uint64_t>(0));
+                }
+            }
+            break;
+        case SymbolStructure::UNIQUE:
+            debug::check<ErrorCode::E_ASSERTION_FAILURE>(check_symbols_all_unique(),
+                                                         "Expected all symbols to be unique in KeySegment");
+        case SymbolStructure::UNKNOWN:
+        default:
+            break;
```
willdealtry, line 45: What happens when we don't recognize the symbol structure? Is this expected?

cpp/arcticdb/column_store/key_segment.cpp
```diff
@@ -0,0 +1,146 @@
...
+        auto stream_id_data = stream_ids_->data();
+        auto stream_id_it = stream_id_data.template cbegin<uint64_TDT>();
+        const bool stream_id_is_sequence_type = is_sequence_type(stream_ids_->type().data_type());
+        const bool index_is_sequence_type = is_sequence_type(start_indexes_->type().data_type());
+        auto index_start_it = index_start_data.template cbegin<int64_TDT>();
+        auto index_end_it = index_start_data.template cbegin<int64_TDT>();
+        for (size_t row_idx = 0;
+             row_idx < num_keys_;
+             ++row_idx, ++stream_id_it, ++version_it, ++creation_ts_it, ++content_hash_it, ++index_start_it, ++index_end_it, ++key_types_it) {
+            // String columns are stored as uint64s (offsets into the string pool), but StringPool::get_const_view expects an int64
```
willdealtry, line 87: Can we change that? That seems quite weird and wrong

cpp/arcticdb/column_store/key_segment.cpp
```diff
@@ -0,0 +1,146 @@
...
+            // Hence the static cast
+            auto symbol_value = symbol_.value_or(
+                    stream_id_is_sequence_type ? StreamId(StringId(string_pool_->get_const_view(static_cast<int64_t>(*stream_id_it)))) : StreamId(NumericId(*stream_id_it))
+            );
+            // String columns are stored as uint64s, whereas timestamp index start/end values are stored as int64s (nanoseconds since epoch)
+            // We could add another if statement above based on index_is_sequence_type, but as these types are the same width,
+            // we can just do some fancy casting to reinterpret the value in the iterator as a uint64.
+            // We then need to static cast back to int64 for the same reason as above RE StringPool::get_const_view
```
willdealtry, line 95: That was historical because I didn't think people would want to store times pre-epoch. It's not ideal

cpp/arcticdb/column_store/key_segment.cpp
```diff
@@ -0,0 +1,146 @@
...
+            auto index_start_value = index_is_sequence_type ?
+                    IndexValue(StringId(string_pool_->get_const_view(static_cast<int64_t>(*reinterpret_cast<const uint64_t*>(&(*index_start_it)))))) :
```
willdealtry, line 97: This seems similar to the stuff in stream_utils.hpp, like get_range_for_segment, is there something in there that could be used, or maybe a well-named function that does this could be added. Probably should all be in index_utils.hpp as well to be fair

cpp/arcticdb/column_store/key_segment.cpp
alexowens90: That's fine, it just means optimisations available to the other 2 enum values can't be applied (not that they're generally implemented yet).
This might get replaced with inheritance at some point, but this is sufficient for now.

cpp/arcticdb/column_store/key_segment.cpp
alexowens90: Decided not to change as strong none/nan are defined in terms of `std::numeric_limits<int64_t>::max()`

willdealtry (approved):

PR #1660, opened 2024-07-05
Refactor log_event function and add retry logic

G-D-Petrov:
We will attempt to retry the writing of the op log up to `Storage.MaxOpLogWriteRetries` times.
This approach is similar to the one already used when [loading the ref key](https://github.com/man-group/ArcticDB/blob/66a5d19c4a2c3b2423b52a68f984ca8d9b52afce/cpp/arcticdb/version/version_map.hpp#L190)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_log.hpp
+++ b/cpp/arcticdb/version/version_log.hpp
@@ -25,8 +25,24 @@ namespace arcticdb {
     };
 
     inline void log_event(const std::shared_ptr<StreamSink>& store, const StreamId& id, std::string action, VersionId version_id=0) {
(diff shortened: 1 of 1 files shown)


jamesmunro:
Some backoff would be useful.  Apart from avoiding extra load when things go wrong it also helps reduce the chance of a retry hitting the same issue.  Why not use https://github.com/man-group/ArcticDB/blob/dc6764fb5ef2d159f9f2e35bcf9c68493d94b140/cpp/arcticdb/util/exponential_backoff.hpp ?

willdealtry (approved):

PR #1671, opened 2024-07-10
Fix sort index

willdealtry:
In the case where there's no sensible way to sort on start_index, sort_index should nevertheless start from zero, otherwise the data will be become unreadable


willdealtry:
wrong branch

PR #1672, opened 2024-07-10
Enhancement 1666: Do not fallback to iterating snapshots if if version as_of int or timestamp not found in version chain in V2 API

### LINKED ISSUE #682: `as_of=Timestamp(...)` searches in snapshots
### Describe the bug

Any operation (`read`, etc.) that accepts `Timestamp` as a way to locate a version will currently search through all the snapshots if there are no live versions available.

This is probably for backwards compatibility with Arctic Python, but it is slow.

This feature is probably unexpected for external users, so we should hide it behind a library option that's enabled for internal users only.

### LINKED ISSUE #1666: Reading `as_of=int` should not search snapshots if the version has been tombstoned in the V2 API

alexowens90:
#### Reference Issues/PRs
CLoses #1666 
Closes #682 

Only affects V2 API, V1 API will behave as it always has

⚠️  API change, do not merge until 5.0.0 ⚠️

### DIFF
--- a/cpp/arcticdb/version/test/test_version_map_batch.cpp
+++ b/cpp/arcticdb/version/test/test_version_map_batch.cpp
@@ -57,7 +57,7 @@ TEST_F(VersionMapBatchStore, SimpleVersionIdQueries) {
         auto stream = fmt::format("stream_{}", i);
         for(uint64_t j = 0; j < num_versions_per_stream; j++){
             stream_ids.emplace_back(stream);
-            version_queries.emplace_back(VersionQuery{SpecificVersionQuery{static_cast<SignedVersionId>(j)}});
+            version_queries.emplace_back(VersionQuery{SpecificVersionQuery{static_cast<SignedVersionId>(j), false}});
         }
     }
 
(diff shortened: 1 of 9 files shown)


python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -378,17 +378,16 @@ def test_non_existent_list_versions_latest_only(arctic_library):
 
 def test_delete_version_with_snapshot(arctic_library):
```
jamesmunro, line 379: Do we have VersionStore coverage for the ` iterate_snapshots_if_tombstoned=False` case?

alexowens90: The V2 API tests implicitly cover the V1 API doing the right thing of the `kwarg` is provided

cpp/arcticdb/version/test/test_version_map_batch.cpp
```diff
@@ -244,9 +244,9 @@ TEST_F(VersionMapBatchStore, CombinedQueries) {
         for(uint64_t j = 0; j < num_versions_per_stream; j++){
             uint64_t idx = i * num_versions_per_stream + j;
             stream_ids.emplace_back(stream);
-            version_queries.emplace_back(VersionQuery{SpecificVersionQuery{static_cast<SignedVersionId>(j)}, false});
+            version_queries.emplace_back(VersionQuery{SpecificVersionQuery{static_cast<SignedVersionId>(j), false}, false});
             stream_ids.emplace_back(stream);
-            version_queries.emplace_back(VersionQuery{TimestampVersionQuery{timestamp(versions[idx]->creation_ts())}, false});
+            version_queries.emplace_back(VersionQuery{TimestampVersionQuery{timestamp(versions[idx]->creation_ts()), false}, false});
```
willdealtry, line 249: Not for now, but I was thinking we should use enums instead of bools for these. We don't have quite as many as we used to, but do_something(symbol, true, false, false, true, false) is not terribly meaningful and also is not type safe. Something we can maybe refactor towards gradually.

willdealtry (approved):

jamesmunro (approved):

PR #1674, opened 2024-07-11
Sort merge starts from zero

willdealtry:
adjust_slice_rowcounts is sometimes used when there are pre-existing rows, but sort_index is always over a whole timeseries so the first row should be zero


willdealtry:
stuff

PR #1675, opened 2024-07-11
Sort index should always start from zero

willdealtry:
adjust_slice_rowcounts can handle preceding rows, but when it's used in sort_index it should always start from zero as it's always running over the whole timeseries

### DIFF
--- a/cpp/arcticdb/version/test/test_sort_index.cpp
+++ b/cpp/arcticdb/version/test/test_sort_index.cpp
@@ -1,0 +1,98 @@
+#include <gtest/gtest.h>
+
+#include <arcticdb/entity/atom_key.hpp>
+#include <arcticdb/pipeline/frame_slice.hpp>
+#include <arcticdb/pipeline/pipeline_common.hpp>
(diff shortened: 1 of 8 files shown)


cpp/arcticdb/version/test/test_sort_index.cpp
```diff
@@ -0,0 +1,98 @@
+#include <gtest/gtest.h>
+
+#include <arcticdb/entity/atom_key.hpp>
+#include <arcticdb/pipeline/frame_slice.hpp>
+#include <arcticdb/pipeline/pipeline_common.hpp>
+#include <arcticdb/pipeline/index_writer.hpp>
+#include "storage/test/in_memory_store.hpp"
+
+TEST(SortIndex, Basic) {
```
alexowens90, line 9: Would be good to have one more test v similar to this one, but where the start and end index of every data key is the same, as that was the use case that flagged the issue up in the first place

willdealtry: I'll add one. Although the result of that was that the first start_index was non-zero, which is what is tested in the second test

PR #1684, opened 2024-07-11
Release GIL on recursive normalizers (master)

willdealtry:
Check various use-cases of unicode strings, including recursive normalizers

### DIFF
--- a/cpp/arcticdb/version/version_store_api.cpp
+++ b/cpp/arcticdb/version/version_store_api.cpp
@@ -561,7 +561,6 @@ VersionedItem PythonVersionStore::write_versioned_composite_data(
     bool prune_previous_versions
     ) {
     ARCTICDB_SAMPLE(WriteVersionedMultiKey, 0)
-
     ARCTICDB_RUNTIME_DEBUG(log::version(), "Command: write_versioned_composite_data");
     auto [maybe_prev, deleted] = ::arcticdb::get_latest_version(store(), version_map(), stream_id);
     auto version_id = get_next_version_from_key(maybe_prev);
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -583,7 +582,10 @@ VersionedItem PythonVersionStore::write_versioned_composite_data(
     }
 
     auto frames = create_input_tensor_frames(sub_keys, items, norm_metas, user_metas, cfg().write_options().empty_types());
+    // Need to hold the GIL up to this point as we will call pb_from_python
```
alexowens90, line 585: `create_input_tensor_frames` only depends on the input parameters, so we can release the GIL right at the start if we move this call up above line 568

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -583,7 +582,10 @@ VersionedItem PythonVersionStore::write_versioned_composite_data(
...
+    auto release_gil = std::make_unique<py::gil_scoped_release>();
     auto index_keys = folly::collect(batch_write_internal(std::move(version_ids), sub_keys, std::move(frames), std::move(de_dup_maps), false)).get();
+    release_gil.reset();
```
alexowens90, line 588: Why do we need to retake the GIL here?

python/tests/integration/arcticdb/test_unicode_strings.py
alexowens90, line 1: Can we also please test:

- `batch_write/batch_append`
- `head/tail`
- `compact_incomplete`
- `batch_read` (both with and without a `QueryBuilder`, as the code paths are different)

alexowens90 (approved):

PR #1690, opened 2024-07-15
Sort index should always start from row zero

willdealtry:
Sort_index sorts the whole timeseries, so no matter what the sort order is, the first row slice needs to start from zero


alexowens90 (approved):
Same comment https://github.com/man-group/ArcticDB/pull/1675/files?diff=unified&w=0#r1677947058

PR #1694, opened 2024-07-16
Bugfix 1683: Add validate_index arg to staged data finalization in both V1 and V2 APIs

### LINKED ISSUE #1683: Add `validate_index` argument to `compact_incomplete` and `finalize_staged_data`
Should behave the same way as regular writes and appends in this respect.

Will also need backporting to `4.4.x`

alexowens90:
#### Reference Issues/PRs
Fixes #1683 

V2 API `finalize_staged_data` continues to validate that index values are sorted and non-overlapping by default, but now checking can be disabled by passing `validate_index=False` in line with other writing methods.
V1 API `compact_incomplete` now allows unsorted timeseries index data by default, and checking can be enabled by passing `validate_index=True` in line with other writing methods.

Also means that when writing incomplete segments, the `validate_index` argument is used to check sortedness or not, previously it was assumed that incompletes had to be `ASCENDING`.

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -855,22 +855,26 @@ void check_incompletes_index_ranges_dont_overlap(const std::shared_ptr<PipelineC
         for (auto it = pipeline_context->incompletes_begin(); it!= pipeline_context->end(); it++) {
             sorting::check<ErrorCode::E_UNSORTED_DATA>(
                     !last_existing_index_value.has_value() || it->slice_and_key().key().start_time() >= *last_existing_index_value,
-                    "Cannot append staged segments to existing data as incomplete segment contains index value < existing data: {} <= {}",
-                    it->slice_and_key().key().start_time(),
-                    *last_existing_index_value);
+                    "Cannot append staged segments to existing data as incomplete segment contains index value < existing data (in UTC): {} <= {}",
+                    date_and_time(it->slice_and_key().key().start_time()),
+                    date_and_time(*last_existing_index_value));
             auto [_, inserted] = unique_timestamp_ranges.insert({it->slice_and_key().key().start_time(), it->slice_and_key().key().end_time()});
             // This is correct because incomplete segments aren't column sliced
             sorting::check<ErrorCode::E_UNSORTED_DATA>(
                     inserted,
-                    "Cannot finalize staged data as incomplete segments overlap one another");
+                    "Cannot finalize staged data as 2 or more incomplete segments cover identical index values (in UTC): ({}, {})",
+                    date_and_time(it->slice_and_key().key().start_time()), date_and_time(it->slice_and_key().key().end_time()));
         };
         for (auto it = unique_timestamp_ranges.begin(); it != unique_timestamp_ranges.end(); it++) {
             auto next_it = std::next(it);
             if (next_it != unique_timestamp_ranges.end()) {
                 sorting::check<ErrorCode::E_UNSORTED_DATA>(
                         next_it->first >= it->second,
-                        "Cannot finalize staged data as incomplete segment index values overlap one another: ({}, {}) intersects ({}, {})",
-                        it->first, it->second - 1, next_it->first, next_it->second - 1);
+                        "Cannot finalize staged data as incomplete segment index values overlap one another (in UTC): ({}, {}) intersects ({}, {})",
+                        date_and_time(it->first),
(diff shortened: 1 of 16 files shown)


python/arcticdb/version_store/library.py
```diff
@@ -900,22 +899,28 @@ def finalize_staged_data(
...
         mode : `StagedDataFinalizeMethod`, default=StagedDataFinalizeMethod.WRITE
-            Finalise mode. Valid options are WRITE or APPEND. Write collects the staged data and writes them to a
-            new timeseries. Append collects the staged data and appends them to the latest version.
+            Finalize mode. Valid options are WRITE or APPEND. Write collects the staged data and writes them to a
+            new version. Append collects the staged data and appends them to the latest version.
         prune_previous_versions: bool, default=False
             Removes previous (non-snapshotted) versions from the database.
         metadata : Any, default=None
             Optional metadata to persist along with the symbol.
+        validate_index: bool, default=True
```
poodlewars, line 919: It would be good to describe why anyone would want to set this to False. Is it just for a performance boost?

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -260,6 +260,32 @@ def test_staged_data(arctic_library, finalize_method):
...
+    if finalize_method == StagedDataFinalizeMethod.APPEND:
+        df_0 = pd.DataFrame({"col": [1, 2]}, index=[pd.Timestamp("2024-01-01"), pd.Timestamp("2024-01-02")])
+        lib.write(sym, df_0)
+    df_1 = pd.DataFrame({"col": [3, 4]}, index=[pd.Timestamp("2024-01-03"), pd.Timestamp("2024-01-04")])
+    df_2 = pd.DataFrame({"col": [5, 6]}, index=[pd.Timestamp("2024-01-03T12"), pd.Timestamp("2024-01-05")])
+    lib.write(sym, df_2, staged=True)
+    lib.write(sym, df_1, staged=True)
+    if validate_index is None:
+        # Test default behaviour when arg isn't provided
+        with pytest.raises(SortingException):
```
poodlewars, line 277: Also need tests for cases where the validation should _pass_

cpp/arcticdb/version/version_core.hpp
```diff
@@ -29,6 +29,15 @@ namespace arcticdb::version_store {
 using namespace arcticdb::entity;
 using namespace arcticdb::pipelines;
 
+struct CompactIncompleteOptions {
```
poodlewars, line 32: Nice cleanup

cpp/arcticdb/version/version_core.cpp
```diff
@@ -858,14 +858,18 @@ void check_incompletes_index_ranges_dont_overlap(const std::shared_ptr<PipelineC
                     "Cannot append staged segments to existing data as incomplete segment contains index value < existing data: {} <= {}",
                     it->slice_and_key().key().start_time(),
                     *last_existing_index_value);
-            unique_timestamp_ranges.insert({it->slice_and_key().key().start_time(), it->slice_and_key().key().end_time()});
-        }
+            auto [_, inserted] = unique_timestamp_ranges.insert({it->slice_and_key().key().start_time(), it->slice_and_key().key().end_time()});
+            // This is correct because incomplete segments aren't column sliced
+            sorting::check<ErrorCode::E_UNSORTED_DATA>(
+                    inserted,
+                    "Cannot finalize staged data as incomplete segments overlap one another");
```
poodlewars, line 865: Should be possible to make the error message describe the overlap?

poodlewars (commented):
Merge conflicts

python/arcticdb/version_store/library.py
alexowens90: No, it is for if you want to allow unordered data. It is following the same pattern as the docstrings for `write` etc, but the phrasing has to be a bit more complex because you're creating the new version from incompletes.

python/tests/integration/arcticdb/test_arctic.py
alexowens90: Agree, I'll add it to `test_parralel.py` though since the v2 API just passes through anyway

alexowens90:
> Merge conflicts

Rebased

cpp/arcticdb/version/version_core.cpp
alexowens90: Added, and also made all index timestamps human readable

poodlewars (approved):

PR #1696, opened 2024-07-16
Add support for cgroup v2 cpu limits

### LINKED ISSUE #1695: Add support for cgroup v2
In ArcticDB we use files in `/sys/fs/cgroup` (or cgroup v1) to determine the number cpu/io threads (as seen [here](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/async/task_scheduler.hpp#L105)).
There is now [cgroup v2](https://www.kernel.org/doc/Documentation/cgroup-v2.txt), which introduces new files with a new format for this.

Extend the current logic for handling the cgroup to support the new interface

There are other ways of getting the number of cpu cores, but some of them don't work on older machines and this is the most universal approach.

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1695 

It extends the current cgroup handling logic to support the new interface.

### DIFF
--- a/cpp/arcticdb/async/task_scheduler.hpp
+++ b/cpp/arcticdb/async/task_scheduler.hpp
@@ -13,6 +13,8 @@
 #include <arcticdb/util/exponential_backoff.hpp>
 #include <arcticdb/util/configs_map.hpp>
 #include <arcticdb/util/home_directory.hpp>
+#include <arcticdb/util/string_utils.hpp>
+
 #include <arcticdb/async/base_task.hpp>
 #include <arcticdb/entity/performance_tracing.hpp>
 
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -101,28 +103,62 @@ struct SchedulerWrapper : public SchedulerType {
     }
 };
 
-inline int64_t get_cgroup_value(const std::string& cgroup_file) {
+struct CGroupValues {
+    int64_t cpu_quota;
+    int64_t cpu_period;
```
alexowens90, line 108: Given how these are used, `std::optional<double>` would make more sense

alexowens90 (approved):

PR #1697, opened 2024-07-16
Bugfix 1692: fix batch read with row range no query builder

alexowens90:
#### Reference Issues/PRs
Fixes 1692 on 4.4.x branch

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1034,15 +1034,17 @@ folly::Future<ReadVersionOutput> async_read_direct(
     const std::shared_ptr<Store>& store,
     const VariantKey& index_key,
     SegmentInMemory&& index_segment,
-    const ReadQuery& read_query,
+    ReadQuery& read_query,
     std::shared_ptr<BufferHolder> buffers,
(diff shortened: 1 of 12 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1150,14 +1154,14 @@ std::vector<std::variant<ReadVersionOutput, DataError>> LocalVersionedEngine::te
 std::vector<std::variant<ReadVersionOutput, DataError>> LocalVersionedEngine::batch_read_internal(
     const std::vector<StreamId>& stream_ids,
     const std::vector<VersionQuery>& version_queries,
-    std::vector<ReadQuery>& read_queries,
+    std::vector<std::shared_ptr<ReadQuery>>& read_queries,
```
willdealtry, line 1157: What was the problem that required a shared_ptr to the ReadQuery? Did we have lifetime problems with it in a more async context?

alexowens90: This function modifies the passed in read queries, and these modifications need to be visible in the Python layer for `_post_process_dataframe` to work properly when the non-query builder branch is taken

willdealtry (approved):

PR #1698, opened 2024-07-17
Parallelize string handling, enable python type handlers, prepare for Arrow

willdealtry:
Move the string handling to the type handlers, get rid of some of the duplicate methods of reading dataframes and do some groundwork for reading data in Arrow format instead of Pandas


vasil-pashov (commented):
Currently we're skipping the tests in test_nullable_boolean_column_type.py, test_empty_column_type.py, test_array_column_type.py

I think it's worth running them against this change just to be sure nothing is broken.

cpp/arcticdb/codec/segment_header.hpp
```diff
@@ -22,7 +22,7 @@ static constexpr std::array<std::string_view, 5> offset_names_ = {
     "COLUMN"
 };
 
-inline void write_fixed_header(std::uint8_t *dst, const FixedHeader& hdr) {
+inline void write_fixed_header(std::uint8_t *dst, FixedHeader hdr) {
```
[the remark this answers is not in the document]
willdealtry: It's the size of a uint64_t so pass-by-value is more efficient than pass by const ref

cpp/arcticdb/processing/test/benchmark_clause.cpp
```diff
@@ -124,7 +124,7 @@ void BM_hash_grouping_int(benchmark::State& state) {
     grouping::ModuloBucketizer bucketizer(num_buckets);
 
     for (auto _ : state) {
-        auto buckets = get_buckets(col_with_strings, grouper, bucketizer);
+        auto buckets ARCTICDB_UNUSED = get_buckets(col_with_strings, grouper, bucketizer);
```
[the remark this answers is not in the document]
willdealtry: We should probably just change them all at the same time

cpp/arcticdb/processing/operation_types.hpp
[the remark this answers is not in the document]
willdealtry: It's not wrong, it's just pointless given the size of the object

cpp/arcticdb/version/test/test_merge.cpp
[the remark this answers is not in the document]
willdealtry: Yep, it's gone

python/tests/stress/arcticdb/test_stress_strings.py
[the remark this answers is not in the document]
willdealtry: I'm pretty sure we have hypothesis tests for this. At some point we should make an ASV benchmark for just strings

python/tests/integration/storage_fixtures/test_fixture_import.py
```diff
@@ -1,15 +1,19 @@
...
     script = """
 import sys
 sys.path = [path for path in sys.path[1:]]
 import arcticdb.storage_fixtures.s3
-import arcticdb.storage_fixtures.azure
-import arcticdb.storage_fixtures.mongo
-import arcticdb.storage_fixtures.lmdb
+#import arcticdb.storage_fixtures.azure
+#import arcticdb.storage_fixtures.mongo
+#import arcticdb.storage_fixtures.lmdb
```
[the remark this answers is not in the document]
willdealtry: Yes but it's frequently failing with later versions of Python. I don't want to try to fix it in this PR as it's already huge. Will open an issue.

cpp/arcticdb/column_store/column_utils.hpp
```diff
@@ -93,17 +93,39 @@ inline py::array array_at(const SegmentInMemory& frame, std::size_t col_pos, py:
...
+            // cpp/arcticdb/pipeline/column_mapping.hpp::external_datatype_size
+            auto column_data = frame.column(col_pos).data();
+            auto none = py::none();
+            auto &api = py::detail::npy_api::get();
+            auto it = column_data.buffer().iterator(sizeof(PyObject*));
+            while(!it.finished()) {
+                auto* ptr = reinterpret_cast<PyObject*>(it.value());
+                util::check(ptr != nullptr, "Can't set base object on null item");
+                if(ptr != none.ptr())
+                    api.PyArray_SetBaseObject_(ptr, anchor.inc_ref().ptr());
```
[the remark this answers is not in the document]
willdealtry: When the datafame object goes out of scope, it will decref all the contained arrays, which will in turn decref the reference count on anchor. In the normal case the refcount of anchor will also fall to zero, immediately after the last array is released. 

However, if someone has a reference to one of the arrays individually, it will prolong the life of the anchor, which is what we want because otherwise the memory backing for the array will disappear while the array is still reachable, leading to a segfault

vasil-pashov (approved):

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -1109,56 +641,26 @@ void reduce_and_fix_columns(
...
-        }
-
-        if(!jobs.empty())
-            folly::collect(jobs).get();
-    } else {
-        for (size_t c = 0; c < static_cast<size_t>(frame.descriptor().fields().size()); ++c) {
-            ReduceColumnTask(frame, c, slice_map, context, unique_string_map, py_nan, spinlock, dynamic_schema, false)();
-        }
-    }
+    static const auto batch_size = ConfigsMap::instance()->get_int("ReduceColumns.BatchSize", 100);
```
alexowens90, line 646: Any reason for this not to just be 2 * CPU threadpool size?

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -20,67 +22,42 @@ namespace arcticdb {
...
-        delete reinterpret_cast<std::shared_ptr<Column> *>(obj);
-    }));
-    return arr.release().ptr();
+
+static inline PyObject** fill_with_none(PyObject** ptr_dest, size_t count, SpinLock& spin_lock) {
+    auto none = py::none();
+    for(auto i = 0U; i < count; ++i)
+        *ptr_dest++ = none.ptr();
+
+    spin_lock.lock();
```
alexowens90, line 32: Should we check we're holding the GIL here?

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -92,51 +69,113 @@ void EmptyHandler::handle_type(
...
+        ChunkedBuffer& dest_buffer,
+        size_t num_rows,
+        size_t offset_bytes,
+        arcticdb::entity::TypeDescriptor,
+        arcticdb::entity::TypeDescriptor,
+        const arcticdb::DecodePathData &,
+        std::any& any,
+        const std::shared_ptr<StringPool> &) {
+    const auto& sparse_map = source_column.opt_sparse_map();
+    const auto num_bools = sparse_map.has_value() ? sparse_map->count() : num_rows;
```
alexowens90, line 163: Do we ever write columns with no values? So sparse map doesn't have a value, but there is no data either?

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -149,90 +188,193 @@ int BoolHandler::type_size() const {
...
+                    );
+                    block_pos += block->shapes()[i];
+                    ++en;
+                }
+            }
+        });
+    } else {
+        column_data.type().visit_tag([&ptr_dest, &column_data, &py_dtype](auto type_desc_tag) {
+            using TDT = decltype(type_desc_tag);
+            while (auto block = column_data.next<TDT>()) {
```
alexowens90, line 356: Do we have exactly one block per array with >0 dimensional data?

cpp/arcticdb/python/python_handlers.hpp
```diff
@@ -1,79 +1,179 @@
...
+        TypeDescriptor dest_type_desc,
+        const DecodePathData& shared_data,
+        std::any& handler_data,
+        const std::shared_ptr<StringPool>& string_pool);
+};
+
+inline void register_array_types() {
+    using namespace arcticdb;
+    constexpr std::array<DataType, 5> array_data_types = {
+        DataType::INT64, DataType::FLOAT64, DataType::EMPTYVAL, DataType::FLOAT32, DataType::INT32};
```
alexowens90, line 159: What about small/unsigned integer types?

alexowens90 (approved):

cpp/arcticdb/pipeline/read_frame.cpp
willdealtry: Probably not, will change subsequently

cpp/arcticdb/python/python_handlers.cpp
willdealtry: Not at the moment, we will write empty-type columns but they should never get here.

willdealtry:
> Currently we're skipping the tests in test_nullable_boolean_column_type.py, test_empty_column_type.py, test_array_column_type.py
> 
> I think it's worth running them against this change just to be sure nothing is broken.

have done

cpp/arcticdb/python/python_handlers.cpp
willdealtry: No, but the block->row_count() will know how many arrays we have

cpp/arcticdb/python/python_handlers.cpp
willdealtry: I think we'll get away without it with none, as it is a global singleton, so there will never be any allocation

PR #1703, opened 2024-07-19
Enhancement 1750: lazy dataframe implementation

### LINKED ISSUE #1750: `LazyDataFrame` implementation

### LINKED ISSUE #1788: Cannot re-use QueryBuilder if date range also specified
### Describe the bug

If you make two read calls, both with a date range specified, and re-use the query builder for both calls, you'll get the error: "ArcticException: In QueryBuilder.prepend: Date range and Resample only supported as first clauses in the pipeline". This error has only started as of version 4.5.0.

### Steps/Code to Reproduce

```python
df = pd.DataFrame({"A": [0, 1, 2]}, index=pd.date_range("2024-01-01", "2024-01-03"))
lib.write("dummy", df)

q = QueryBuilder()
q = q[q["A"] > 1]

lib.read("dummy", date_range=(datetime(2024, 1, 2), datetime(2024, 1, 3)), query_builder=q)
lib.read("dummy", date_range=(datetime(2024, 1, 2), datetime(2024, 1, 3)), query_builder=q)
```

### Expected Results

The second call gives the error "ArcticException: In QueryBuilder.prepend: Date range and Resample only supported as first clauses in the pipeline". This doens't feel intuitive - the read call shouldn't modify the QueryBuilder passed by the user. Perhaps a solution would be to clone the existing QueryBuilder and prepend the DateRangeClause onto that?

For now, I can get around this issue by doing:

```python
q = QueryBuilder()
q = q[q["A"] >= 1]
q.prepend(QueryBuilder().date_range((datetime(2024, 1, 2), datetime(2024, 1, 3))))
```

and not specifying a date_range in the lib.read calls. But would be good if a fix was made. Thanks!

`test_lazy_collect_twice_with_date_range` in `test_lazy_dataframe.py` is xfailed on this issue, and should have the pytest mark removed if #1703 is merged before this is fixed.

Python: 3.9.17 (main, Jun  6 2023, 20:11:04) 
[GCC 9.4.0]
OS: Linux-5.4.0-51-generic-x86_64-with-glibc2.31
ArcticDB: 4.5.0

### Backend storage used

MINIO

### Additional Context

_No response_

alexowens90:
Closes #1750 

Adds a lazy dataframe implementation that allows a more natural syntax for querying operations, with actual read+processing deferred until `collect` is called.

Also fixes #1788

### DIFF
--- a/python/arcticdb/__init__.py
+++ b/python/arcticdb/__init__.py
@@ -10,8 +10,16 @@ import arcticdb.version_store.library as library
 from arcticdb.tools import set_config_from_env_vars
 from arcticdb_ext.version_store import DataError, VersionRequestType
 from arcticdb_ext.exceptions import ErrorCode, ErrorCategory
-from arcticdb.version_store.library import WritePayload, ReadInfoRequest, ReadRequest
-from arcticdb.version_store.library import StagedDataFinalizeMethod, WriteMetadataPayload
+from arcticdb.version_store.library import (
+    WritePayload,
+    ReadInfoRequest,
+    ReadRequest,
+    col,
+    LazyDataFrame,
+    LazyDataFrameCollection,
+    StagedDataFinalizeMethod,
+    WriteMetadataPayload
+)
 
 set_config_from_env_vars(_os.environ)
 
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -344,7 +344,7 @@ class LazyDataFrame(QueryBuilder):
         self.lib = lib
         self.read_request = read_request._replace(query_builder=None)
 
-    def to_read_request(self) -> ReadRequest:
+    def _to_read_request(self) -> ReadRequest:
         """
         Convert this object into a ReadRequest, including any queries applied to this object since the read call.
 
@@ -373,11 +373,11 @@ class LazyDataFrame(QueryBuilder):
         VersionedItem
             Object that contains a .data and .metadata element.
         """
-        return self.lib.read(**self.to_read_request()._asdict())
+        return self.lib.read(**self._to_read_request()._asdict())
 
     def __str__(self) -> str:
         query_builder_repr = super().__str__()
-        return "LazyDataFrame(" + self.read_request.__repr__() + (" | " if len(query_builder_repr) else "") + query_builder_repr + ")"
+        return f"LazyDataFrame({self.read_request.__repr__()}{' | ' if query_builder_repr else ''}{query_builder_repr})"
 
     # Needs to be explicitly defined for lists of these objects in LazyDataFrameCollection to render correctly
     def __repr__(self) -> str:
@@ -416,10 +416,10 @@ class LazyDataFrameCollection(QueryBuilder):
             self,
             lazy_dataframes: List[LazyDataFrame],
     ):
-        lib_set = set([lazy_dataframe.lib for lazy_dataframe in lazy_dataframes])
+        lib_set = {lazy_dataframe.lib for lazy_dataframe in lazy_dataframes}
         check(
             len(lib_set) in [0, 1],
-            f"LazyDataFrameCollection init requires all provided lazy dataframes to be referring to the same library, but received:\n{[lib for lib in lib_set]}"
+            f"LazyDataFrameCollection init requires all provided lazy dataframes to be referring to the same library, but received: {[lib for lib in lib_set]}"
         )
         super().__init__()
         self._lazy_dataframes = lazy_dataframes
@@ -445,14 +445,14 @@ class LazyDataFrameCollection(QueryBuilder):
         List[Union[VersionedItem, DataError]]
             See documentation on `Library.read_batch`.
         """
-        if self._lib is None:
+        if not len(self._lazy_dataframes):
             return []
         return self._lib.read_batch(self._read_requests())
 
     def _read_requests(self) -> List[ReadRequest]:
         # Combines queries for individual LazyDataFrames with the global query associated with this
         # LazyDataFrameCollection and returns a list of corresponding read requests
-        read_requests = [lazy_dataframe.to_read_request() for lazy_dataframe in self._lazy_dataframes]
+        read_requests = [lazy_dataframe._to_read_request() for lazy_dataframe in self._lazy_dataframes]
         if len(self.clauses):
             for read_request in read_requests:
                 if read_request.query_builder is None:
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -312,7 +312,7 @@ class ReadInfoRequest(NamedTuple):
 class LazyDataFrame(QueryBuilder):
     """
     Lazy dataframe implementation, allowing chains of queries to be added before the read is actually executed.
-    Returned by `Library.read` calls when `lazy=True`.
+    Returned by `Library.read`, `Library.head`, and `Library.tail` calls when `lazy=True`.
 
     See Also
     --------
@@ -346,7 +346,7 @@ class LazyDataFrame(QueryBuilder):
 
     def to_read_request(self) -> ReadRequest:
         """
-        Convert this object back into a ReadRequest, including any queries applied to this object since the read call.
+        Convert this object into a ReadRequest, including any queries applied to this object since the read call.
 
         Returns
         -------
@@ -387,7 +387,7 @@ class LazyDataFrame(QueryBuilder):
 class LazyDataFrameCollection(QueryBuilder):
     """
     Lazy dataframe implementation for batch operations. Allows the application of chains of queries to be added before
-    the actual reads are performed. Queries applied to this object will be applied to all of the symbols being read.
+    the actual reads are performed. Queries applied to this object will be applied to all  the symbols being read.
     If per-symbol queries are required, split can be used to break this class into a list of LazyDataFrame objects.
     Returned by `Library.read_batch` calls when `lazy=True`.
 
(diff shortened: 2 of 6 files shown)


python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -111,6 +111,52 @@ def test_reuse_querybuilder(lmdb_version_store_tiny_segment):
...
+    q = QueryBuilder()
+    q = q[q["col1"].isin(2, 3, 7)]
+
+    expected_0 = df.query("col1 in [2, 3]")
+    received_0 = lib.read(symbol, date_range=(None, pd.Timestamp("2000-01-06")), query_builder=q).data
+    assert_frame_equal(expected_0, received_0)
+
+    received_1 = lib.read(symbol, date_range=(None, pd.Timestamp("2000-01-06")), query_builder=q).data
+    assert_frame_equal(expected_0, received_1)
+
```
DrNickClarke, line 131: Would another test with a different date_range and the same query be worthwhile?

python/arcticdb/__init__.py
```diff
@@ -10,8 +10,16 @@
 from arcticdb.tools import set_config_from_env_vars
 from arcticdb_ext.version_store import DataError, VersionRequestType
 from arcticdb_ext.exceptions import ErrorCode, ErrorCategory
-from arcticdb.version_store.library import WritePayload, ReadInfoRequest, ReadRequest
-from arcticdb.version_store.library import StagedDataFinalizeMethod, WriteMetadataPayload
+from arcticdb.version_store.library import (
+    WritePayload,
+    ReadInfoRequest,
+    ReadRequest,
+    col,
```
DrNickClarke, line 17: perhaps a test for the new "col" feature? to show how to use it.

alexowens90: That's covered here:
https://github.com/man-group/ArcticDB/blob/71d917621b1aa9360df1c87fa3d426c4df592f80/python/tests/unit/arcticdb/version_store/test_lazy_dataframe.py#L123

python/arcticdb/version_store/library.py
```diff
@@ -288,6 +302,195 @@ class ReadInfoRequest(NamedTuple):
...
+        if read_request.query_builder is None:
+            super().__init__()
+        else:
+            self.clauses = read_request.query_builder.clauses
+            self._python_clauses = read_request.query_builder._python_clauses
+            self._optimisation = read_request.query_builder._optimisation
+        self.lib = lib
+        self.read_request = read_request._replace(query_builder=None)
+
+    def to_read_request(self) -> ReadRequest:
```
poodlewars, line 347: Should we make this private? Why would a user call this?

alexowens90: They wouldn't but it's used by `LazyDataFrameCollection`, and IDEs whinge when you use private methods

python/arcticdb/version_store/library.py
```diff
@@ -288,6 +302,195 @@ class ReadInfoRequest(NamedTuple):
...
+        Returns
+        -------
+        VersionedItem
+            Object that contains a .data and .metadata element.
+        """
+        return self.lib.read(**self.to_read_request()._asdict())
+
+    def __str__(self) -> str:
+        query_builder_repr = super().__str__()
+        return "LazyDataFrame(" + self.read_request.__repr__() + (" | " if len(query_builder_repr) else "") + query_builder_repr + ")"
```
poodlewars, line 380: Why `if len(query_builder_repr)` not `if query_builder_repr` ?

python/arcticdb/version_store/library.py
```diff
@@ -288,6 +302,195 @@ class ReadInfoRequest(NamedTuple):
...
+    >>> lazy_dfs = LazyDataFrameCollection(lazy_dfs)
+    >>> lazy_dfs = lazy_dfs[lazy_dfs["new_col"] > 0]
+    # Actual read and processing happens here
+    >>> res = lazy_dfs.collect()
+    """
+    def __init__(
+            self,
+            lazy_dataframes: List[LazyDataFrame],
+    ):
+        lib_set = set([lazy_dataframe.lib for lazy_dataframe in lazy_dataframes])
```
poodlewars, line 419: Use a set comprehension

python/arcticdb/version_store/library.py
```diff
@@ -288,6 +302,195 @@ class ReadInfoRequest(NamedTuple):
...
+        check(
+            len(lib_set) in [0, 1],
+            f"LazyDataFrameCollection init requires all provided lazy dataframes to be referring to the same library, but received:\n{[lib for lib in lib_set]}"
```
poodlewars, line 422: Remove the newline? (makes grepping logs hard)

python/arcticdb/version_store/library.py
```diff
@@ -288,6 +302,195 @@ class ReadInfoRequest(NamedTuple):
...
+        """
+        Read the data and execute any queries applied to this object since the read_batch call.
+
+        Returns
+        -------
+        List[Union[VersionedItem, DataError]]
+            See documentation on `Library.read_batch`.
+        """
+        if self._lib is None:
+            return []
```
poodlewars, line 449: What's the rationale not to raise in this case?

python/tests/unit/arcticdb/version_store/test_lazy_dataframe.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+def test_lazy_read(lmdb_library):
+    lib = lmdb_library
+    sym = "test_lazy_read"
+    df = pd.DataFrame(
+        {"col1": np.arange(10, dtype=np.int64), "col2": np.arange(100, 110, dtype=np.int64)}, index=pd.date_range("2000-01-01", periods=10)
+    )
+    lib.write(sym, df)
+    lib.write_pickle(sym, 1)
```
poodlewars, line 24: I know this is just to increment the version number, but how does the LazyDataFrame behave when reading pickled data? Probably worth some docs

python/tests/unit/arcticdb/version_store/test_lazy_dataframe.py
```diff
@@ -0,0 +1,456 @@
...
+
+    lazy_df = lib.read(sym, lazy=True)
+    lazy_df = lazy_df.resample("D").agg({"col1": "sum", "col2": "first"})
+    received = lazy_df.collect().data
+    expected = df.resample("D").agg({"col1": "sum", "col2": "first"})
+
+    assert_frame_equal(expected, received)
+
+
+def test_lazy_with_initial_query_builder(lmdb_library):
```
poodlewars, line 179: What's the motivation for this usage? Should we just forbid it and simplify things a bit?

python/tests/unit/arcticdb/version_store/test_lazy_dataframe.py
poodlewars, line 1: Does re-use work? I mean,

```
lazy_df2 = lazy_df
lazy_df = lazy_df[lazy_df["col1"].isin(0, 3, 6, 9)]
```

is lazy_df2 unaffected by the modification to lazy_df?

python/arcticdb/version_store/library.py
poodlewars: Probably better than having this on the versioned API forever?

alexowens90: Yeh fair, have made private

python/arcticdb/version_store/library.py
alexowens90: Done

python/arcticdb/version_store/library.py
alexowens90: Done

python/arcticdb/version_store/library.py
alexowens90: Done

python/arcticdb/version_store/library.py
alexowens90: The real condition for wanting to return `[]` is if `len(self._lazy_dataframes) == 0`, which is also when `self._lib is None`. Updated the if statement to reflect this

python/tests/unit/arcticdb/version_store/test_lazy_dataframe.py
alexowens90: Same as `QueryBuilder`, an exception is thrown. This is covered in the `QueryBuilder` docstring:
https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/version_store/processing.py#L384

python/tests/unit/arcticdb/version_store/test_lazy_dataframe.py
alexowens90: I guess it doesn't really add anything with `read`. With `read_batch` I can see it being used when you have one query you want to apply to all symbols before splitting to apply separate queries to different symbols, and I want to keep `read` and `read_batch` consistent with each other.

python/tests/unit/arcticdb/version_store/test_lazy_dataframe.py
alexowens90: `lazy_df2` was modified in this case, although the issue was subtle as `__getitem__` modified `self`. I've changed `__getitem__` to return a copy of itself instead. This will fix the example @jamesmunro gave in #1773. I don't agree that all the other methods should do the same though. The underlying issue is that it's called a `QueryBuilder`, but it doesn't really follow the builder pattern, in that there is no final `build()` method that returns a `Query` object, this is implicitly handled in `read` calls.

poodlewars: Need a test for commit e0bc2033d8fe9480db4b34607bd4057ad8117916

python/arcticdb/version_store/processing.py
```diff
@@ -390,6 +390,11 @@ class QueryBuilder:
     """
 
     def __init__(self):
+        # Note that these lists will be modified using + rather than append so that operations like this work:
+        # q = QueryBuilder()
+        # q_copy = q
```
poodlewars, line 395: Did you consider implementing 

```
__copy__
```
? Could help to make things less fragile

poodlewars (approved):

PR #1713, opened 2024-07-26
Change the default `dynamic_strings` to True

IvoDD:
This is in line with library creation.


poodlewars (changes requested):
We can't do this - the default for dynamic strings _used_ to be false. This change would break existing libraries.

poodlewars (approved):
Apparently the type promotion fixed -> dynamic works even when `dynamic_schema=False` 
```
we do have tests where appending with dynamic strings to a fixed string dataframe works by switching to dynamic strings [here](https://github.com/man-group/ArcticDB/blob/0a33f5a213f1be566bb98e6e6f88891f7256997f/python/tests/unit/arcticdb/version_store/test_column_type_changes.py#L51)
```

PR #1715, opened 2024-07-26
Add a better error message for bad variant access when taking buffer view

poodlewars:
Otherwise a `std::bad_variant_access` bubbles up with:

```
std::get: wrong index for variant
```

and it can be hard to see which `std::get<T>` is the one that's raising.

Motivation: https://github.com/man-group/arcticdb-enterprise/pull/139

### DIFF
--- a/cpp/arcticdb/util/buffer.hpp
+++ b/cpp/arcticdb/util/buffer.hpp
@@ -309,11 +309,17 @@ public:
     }
 
     [[nodiscard]] BufferView view() const {
-        if (std::holds_alternative<std::shared_ptr<Buffer>>(buffer_)) {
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/util/buffer.hpp
```diff
@@ -309,11 +309,17 @@ class VariantBuffer {
     }
 
     [[nodiscard]] BufferView view() const {
```
willdealtry, line 311: Have we got a repro for this? It seems like something that should absolutely never happen

poodlewars: It's a use after move in the enterprise replication tooling

willdealtry (approved):

PR #1729, opened 2024-07-31
Get version information from the symbol list when available

willdealtry:
WIP In the case where we don't need snapshot information or deleted versions, we can get the latest version information more quickly from the new-style symbol list


willdealtry:
closing

PR #1741, opened 2024-08-01
Refactor/1722/remove composite from processing pipeline

### LINKED ISSUE #1586: Further optimise billion row challenge performance on high core counts
While working on https://github.com/man-group/ArcticDB/pull/1584 it is clear there is a bottleneck calling `gather_entities` in `AggregationClause::process`. Due to the `Composite<EntityIds>` structure containing `num_input_row_slices`, which is 10,000 for a billion rows, and a mutex must be acquired for each element of the composite.

We should get rid of `Composite` entirely (#1722) in the processing pipeline, it hinders rather than helps at this stage. `process` should take `EntityIds == std::vector<EntityId>` as input and return the same as output. `repartition` should take `std::vector<EntityIds>` as input and return the same as output. This will allow us to flatten out to a single call to `gather_entities` in `AggregationClause::process`.

In addition, the only (and inherently) serial stage of the processing is the repartition after the bucketising. This takes ~700ms of the ~5s total time, or ~15% of the total runtime executing the billion row challenge, and scales inversely to the number of cores. This time is dominated by gathering and pushing entities from/to the `ComponentManager` due to map insertion/erasure, when in reality, the entities are not being modified by the repartitioning, so we should eliminate that bottleneck.

### LINKED ISSUE #1722: Remove `Composite` from processing pipeline
Since the introduction of the `ComponentManager`, restructuring entities on entry to/exit from clauses is trivial. `process` methods should just accept and return a collection of entity IDs. Repartition methods should accept and return a vector of such collections.

alexowens90:
#### Reference Issues/PRs
Closes #1722 
Closes #1586

Billion row challenge using 64 cores has gone from 5.4 to 3.5 seconds with this change

### DIFF
--- a/cpp/arcticdb/async/tasks.hpp
+++ b/cpp/arcticdb/async/tasks.hpp
@@ -395,18 +395,18 @@ struct SegmentFunctionTask : BaseTask {
 
 struct MemSegmentProcessingTask : BaseTask {
     std::vector<std::shared_ptr<Clause>> clauses_;
-    Composite<EntityIds> entity_ids_;
+    std::vector<EntityId> entity_ids_;
 
     explicit MemSegmentProcessingTask(
            std::vector<std::shared_ptr<Clause>> clauses,
-           Composite<EntityIds>&& entity_ids) :
+           std::vector<EntityId>&& entity_ids) :
         clauses_(std::move(clauses)),
         entity_ids_(std::move(entity_ids)) {
     }
 
     ARCTICDB_MOVE_ONLY_DEFAULT(MemSegmentProcessingTask)
 
-    Composite<EntityIds> operator()() {
+    std::vector<EntityId> operator()() {
         std::ranges::reverse_view reversed_clauses{clauses_};
         for (const auto& clause: reversed_clauses) {
             entity_ids_ = clause->process(std::move(entity_ids_));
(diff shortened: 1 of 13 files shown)


cpp/arcticdb/async/tasks.hpp
```diff
@@ -395,26 +395,26 @@ struct SegmentFunctionTask : BaseTask {
 
 struct MemSegmentProcessingTask : BaseTask {
     std::vector<std::shared_ptr<Clause>> clauses_;
-    Composite<EntityIds> entity_ids_;
+    EntityIds entity_ids_;
 
     explicit MemSegmentProcessingTask(
            std::vector<std::shared_ptr<Clause>> clauses,
-           Composite<EntityIds>&& entity_ids) :
+           EntityIds&& entity_ids) :
```
willdealtry, line 402: I'm never sure whether aliasing a std::vector is a good idea, I spent some time wondering if EntityIds was a class that did something super smart with multiple EntityIds.  Unless of course the plan is to swap it out for something like that in the future.

willdealtry (approved):

cpp/arcticdb/async/tasks.hpp
alexowens90: Agreed, I added it in originally because otherwise you would have had `std::vector<Composite<std::vector<EntityId>>>`, but now I'm ripping `Composite` out it'll be readable without the alias

Issue #1724, opened 2024-07-30
conda-build: undefined behaviors detected with `libcxx v18.1.8` on macOS


jjerphan:
First observed [here](https://github.com/man-group/ArcticDB/actions/runs/9889332532/job/27315155035#step:7:1043). Logs for posterity:
```
The following tests FAILED:
	  1 - Async.SinkBasic (ILLEGAL)
	  2 - Async.DeDupTest (ILLEGAL)
	 23 - Segment.RoundtripTimeseriesDescriptorWriteToBufferV2 (ILLEGAL)
	 27 - SegmentHeader.SerializeUnserializeV1 (ILLEGAL)
	323 - SparseTestStore.Compact (ILLEGAL)
	324 - SparseTestStore.CompactWithStrings (ILLEGAL)
	388 - VersionStoreTest.CompactIncompleteDynamicSchema (ILLEGAL)
	408 - TestEmbedded/SimpleTestSuite.Example/lmdb  # GetParam() = 24-byte object <08-6C 6D-64 62-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00> (ILLEGAL)
	412 - AllStoragesCommonTests/GenericStorageTest.WriteDuplicateKeyException/(ptr = 0x6000023e8018, value = 8-byte object <90-1F 99-0A 01-00 00-00>) (ILLEGAL)
```

Also observed on the feedstock with https://github.com/conda-forge/arcticdb-feedstock/pull/249.

Currently solved by using `libcxx<18` (see https://github.com/man-group/ArcticDB/pull/1680).

---

Other projects are having a similar problem with this build (see https://github.com/conda-forge/libcxx-feedstock/issues/162)

ianthomas23:
I can reproduce this on an M3 mac using master branch [environment_unix.yml](https://github.com/man-group/ArcticDB/blob/master/environment_unix.yml) but with `libcxx` unconstrained. I needed to manually change the use of `posix_memalign` rather than `memalign` (issue #1322).

For me the failed tests are:
```
The following tests FAILED:
	  1 - Async.SinkBasic (SIGTRAP)
	  2 - Async.DeDupTest (SIGTRAP)
	 25 - Segment.RoundtripTimeseriesDescriptorWriteToBufferV2 (SIGTRAP)
	 29 - SegmentHeader.SerializeUnserializeV1 (SIGTRAP)
	208 - LMDBStorageTestBase.WriteMapFullError (Failed)
[...]

PR #1745, opened 2024-08-05
Use EncodedFieldImpl.blocks() rather than casting member variable _blocks

ianthomas23:
#### Reference Issues/PRs

Fixes part of #1724.

This fixes one of the undefined behaviours that came to light in issue #1724 when building and testing ArcticDB on macOS ARM using a hardened build of `libcxx`. The C++ test it fixes is `SegmentHeader.SerializeUnserializeV1`.

#### Any other comments?

Previously `EncodedFieldImpl::shapes` was applying an offset into the `_blocks` member variable `std::array` before casting to an `EncodedBlock`. Other use of `_blocks` in this class uses the member function `blocks()` to do it the other way round, i.e. casting then offsetting. In practice this has been fine but in principle it is not as some hardware platform could have some interesting memory alignment to mess this up, which is what the hardened `libcxx` identifies.

The fix is to use the same approach in `shapes()` to that in `values()`, the function below it.

I have tested this locally against a hardened build on macOS ARM, and also against an unhardened build on Ubuntu. Hardened builds are currently not tested in CI and I suspect that we don't want to add one as it would be considered a poor use of CI resources.

### DIFF
--- a/cpp/arcticdb/codec/encoded_field.hpp
+++ b/cpp/arcticdb/codec/encoded_field.hpp
@@ -307,9 +307,9 @@ struct EncodedFieldImpl : public EncodedField {
     [[nodiscard]] const EncodedBlock& shapes(size_t n) const {
         util::check(shapes_count_ != 0, "No shape allocated");
         if(!is_old_style_shapes())
-            return *reinterpret_cast<const EncodedBlock*>(&blocks_[0]);
+            return blocks()[n];
         else
-            return *reinterpret_cast<const EncodedBlock*>(&blocks_[n * 2]);
+            return blocks()[n * 2];
     }
 
     [[nodiscard]] const EncodedBlock &values(size_t n) const {
--- a/cpp/arcticdb/codec/encoded_field.hpp
+++ b/cpp/arcticdb/codec/encoded_field.hpp
@@ -306,9 +306,10 @@ struct EncodedFieldImpl : public EncodedField {
 
     [[nodiscard]] const EncodedBlock& shapes(size_t n) const {
         util::check(shapes_count_ != 0, "No shape allocated");
-        if(!is_old_style_shapes())
-            return blocks()[n];
-        else
+        if(!is_old_style_shapes()) {
+            util::check(n == 0, "Block index must be 0 not {} if not using old style shapes", n);
+            return blocks()[0];
+        } else
             return blocks()[n * 2];
     }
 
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/codec/encoded_field.hpp
```diff
@@ -307,9 +307,9 @@ struct EncodedFieldImpl : public EncodedField {
     [[nodiscard]] const EncodedBlock& shapes(size_t n) const {
         util::check(shapes_count_ != 0, "No shape allocated");
         if(!is_old_style_shapes())
-            return *reinterpret_cast<const EncodedBlock*>(&blocks_[0]);
+            return blocks()[n];
```
ianthomas23, line 310: I think this should be an `n` but it was previously a `0`, so this needs to be confirmed.

vasil-pashov: This should be 0. In the old style shapes we had a block of shapes followed by a block of values. The new style has only a single block of shapes and it's always located at 0.

willdealtry: Maybe we should assert that it's zero, since asking for a shape block > n if it's not old style is always wrong

ianthomas23: Thanks, I've changed the index to 0 and added an assert.

cpp/arcticdb/codec/encoded_field.hpp
```diff
@@ -306,10 +306,11 @@ struct EncodedFieldImpl : public EncodedField {
...
     [[nodiscard]] const EncodedBlock& shapes(size_t n) const {
         util::check(shapes_count_ != 0, "No shape allocated");
-        if(!is_old_style_shapes())
-            return *reinterpret_cast<const EncodedBlock*>(&blocks_[0]);
-        else
-            return *reinterpret_cast<const EncodedBlock*>(&blocks_[n * 2]);
+        if(!is_old_style_shapes()) {
+            util::check(n == 0, "Block index must be 0 not {} if not using old style shapes", n);
+            return blocks()[0];
+        } else
```
vasil-pashov, line 312: (nit: formatting) usually when the if is wrapped in {} we wrap the else in {} as well even if it's single statement

ianthomas23: Done in commit e49d4ff711

willdealtry (approved):

PR #1746, opened 2024-08-05
Refactor - safer KeySegmentPair API

poodlewars:
The motivation for the change is to allow `arcticdb-enterprise` to copy blocks to NFS storages without a use-after-move. I explained this in https://github.com/man-group/arcticdb-enterprise/pull/139 but to have an open record:

CopyCompressedInterStoreTask has:

```
                        // Don't bother copying the key segment pair when writing to the final target
                        if (it == std::prev(target_stores_.end())) {
                            (*it)->write_compressed_sync(std::move(key_segment_pair));
                        } else {
                            auto key_segment_pair_copy = key_segment_pair;
                            (*it)->write_compressed_sync(std::move(key_segment_pair_copy));
                        }
```

KeySegmentPair has a shared_ptr to a KeySegmentPair, which we can think of here as just a `Segment`.

Therefore the old `key_segment_pair_copy` is shallow, the underlying Segment is the same. But the segment eventually gets passed as an rvalue reference further down the stack.

In `do_write_impl` we call `put_object` which calls `serialize_header`.

This modifies the segment in place and passes that buffer to the AWS SDK.

In the `NfsBackedStorage` we have:

```
void NfsBackedStorage::do_write(Composite<KeySegmentPair>&& kvs) {
    auto enc = kvs.transform([] (auto&& key_seg) {
      return KeySegmentPair{encode_object_id(key_seg.variant_key()), std::move(key_seg.segment())};
    });
    s3::detail::do_write_impl(std::move(enc), root_folder_, bucket_name_, *s3_client_, NfsBucketizer{});
}
```

where the segment gets moved from.

Subsequent attempts to use the segment (eg copying on to the next store) then fail.

https://github.com/man-group/arcticdb-enterprise/pull/139 fixed this issue by cloning the segment, but this approach avoids the (expensive) clone.

## HEAD~1 Commit

Refactor the KeySegmentPair API so that it is possible to re-use a `Segment` without cloning it.

Do not expose lvalue references to the segment.
Make the Segment const (ish) and pass it with a shared_ptr rather than moving when possible.

The main design decisions to review are in `key_segment_pair.hpp` and `segment.hpp`, which we now pass by const-ref whenever possible. Most of the other changes are just coping with those API changes.

## HEAD Commit

Move the CopyCompressedInterStoreTask down to ArcticDB from arcticdb-enterprise. Add a test for it on NFS storage. I've verified that the tests in this commit fail without the refactor in the HEAD~1 commit. The only changes to `CopyCompressedInterstoreTask` from enterprise are:

- Get rid of the optimisation where we skipped the `key_segment_pair_copy` for the last target. The `KeySegmentPair` is cheap to copy (especially considering we are about to copy an object across storages, likely with a network hop).
- We have adopted the new `set_key` API of `KeySegmentPair`:
```
if (key_to_write_.has_value()) {
            }
```

### DIFF
--- a/cpp/arcticdb/storage/key_segment_pair.hpp
+++ b/cpp/arcticdb/storage/key_segment_pair.hpp
@@ -9,6 +9,7 @@
 
 #include <arcticdb/entity/key.hpp>
 #include <arcticdb/codec/segment.hpp>
+#include <utility>
 
 namespace arcticdb::storage {
     using namespace entity;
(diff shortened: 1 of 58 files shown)


cpp/arcticdb/storage/key_segment_pair.hpp
```diff
@@ -90,14 +79,8 @@ namespace arcticdb::storage {
         }
 
     private:
-        std::shared_ptr<KeySegmentData> data_;
+        std::shared_ptr<VariantKey> key_ = std::make_shared<VariantKey>();
+        std::shared_ptr<Segment> segment_ = std::make_shared<Segment>();
     };
 
-    template<class T, std::enable_if_t<std::is_same_v<T, EnvironmentName> || std::is_same_v<T, StorageName>, int> = 0>
-    bool operator==(const T &l, const T &r) {
```
poodlewars, line 97: This shouldn't have been in this file. I've moved it to `common.hpp`.

PR #1756, opened 2024-08-07
Fix Segment use-after-move when replicating to NFS

poodlewars:
## Motivation

(this section copied from previous (closed) attempt - https://github.com/man-group/ArcticDB/pull/1746)

The motivation for the change is to allow `arcticdb-enterprise` to copy blocks to NFS storages without a use-after-move. I explained this in https://github.com/man-group/arcticdb-enterprise/pull/139 but to have an open record:

CopyCompressedInterStoreTask has:

```
                        // Don't bother copying the key segment pair when writing to the final target
                        if (it == std::prev(target_stores_.end())) {
                            (*it)->write_compressed_sync(std::move(key_segment_pair));
                        } else {
                            auto key_segment_pair_copy = key_segment_pair;
                            (*it)->write_compressed_sync(std::move(key_segment_pair_copy));
                        }
```

KeySegmentPair has a shared_ptr to a KeySegmentPair, which we can think of here as just a `Segment`.

Therefore the old `key_segment_pair_copy` is shallow, the underlying Segment is the same. But the segment eventually gets passed as an rvalue reference further down the stack.

In `do_write_impl` we call `put_object` which calls `serialize_header`.

This modifies the segment in place and passes that buffer to the AWS SDK.

In the `NfsBackedStorage` we have:

```
void NfsBackedStorage::do_write(Composite<KeySegmentPair>&& kvs) {
    auto enc = kvs.transform([] (auto&& key_seg) {
      return KeySegmentPair{encode_object_id(key_seg.variant_key()), std::move(key_seg.segment())};
    });
    s3::detail::do_write_impl(std::move(enc), root_folder_, bucket_name_, *s3_client_, NfsBucketizer{});
}
```

where the segment gets moved from.

Subsequent attempts to use the segment (eg copying on to the next store) then fail.

https://github.com/man-group/arcticdb-enterprise/pull/139 fixed this issue by cloning the segment, but this approach avoids the (expensive) clone.

## Logical Change

Copy the `KeySegmentPair`'s pointer to the `Segment` in `nfs_backed_storage.cpp` rather than moving from the segment.

### Copy Task

Move the CopyCompressedInterStoreTask down to ArcticDB from arcticdb-enterprise. Add a test for it on NFS storage. I've verified that the tests in this commit fail without the refactor in the HEAD~1 commit. The only changes to `CopyCompressedInterstoreTask` from enterprise are:

- Pass the `KeySegmentPair` by value in to `write_compressed{_sync}`. The `KeySegmentPair` is cheap to copy (especially considering we are about to copy an object across storages, likely with a network hop).
- We have adopted the new `set_key` API of `KeySegmentPair`:
```
if (key_to_write_.has_value()) {
            }
```
- We have namespaced the `ProcessingResult` struct in to the task

### KeySegmentPair

- Replace methods returning mutable lvalue references to keys with a `set_key` method.
- Remove the `release_segment` method as it dangerously leaves the `KeySegmentPair` pointing at a `Segment` object that has been moved from, and it is not actually necessary.

## Follow up work

The non-const `Segment& KeySegmentPair#segment()` API is still dangerous and error prone. I have a follow up change to remove it, but that API change affects very many files and will be best raised separately so that  it doesn't block this fix for replication.

A draft PR showing a proposal for that change is here - https://github.com/man-group/ArcticDB/pull/1757 .


cpp/arcticdb/storage/key_segment_pair.hpp
```diff
@@ -18,63 +18,53 @@ namespace arcticdb::storage {
...
-        }
-
         [[nodiscard]] const VariantKey& variant_key() const {
-            return data_->key_;
+            return *key_;
         }
 
         [[nodiscard]] const Segment &segment() const {
-            return data_->segment_;
+            return *segment_;
```
alexowens90, line 67: Check `segment_` holds something first? And same in `variant_key()`?

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -78,7 +78,7 @@ void tombstone_snapshot(
     }
     // Append a timestamp to the ID so that other snapshot(s) can reuse the same snapshot name before the cleanup job:
     std::string new_key = fmt::format("{}@{:x}", key_segment_pair.ref_key(), util::SysClock::coarse_nanos_since_epoch() / 1'000'000);
-    key_segment_pair.ref_key() = RefKey(new_key, KeyType::SNAPSHOT_TOMBSTONE);
+    key_segment_pair.set_key(RefKey(new_key, KeyType::SNAPSHOT_TOMBSTONE));
```
alexowens90, line 81: Move in `new_key`

cpp/arcticdb/async/tasks.cpp
```diff
@@ -36,8 +36,8 @@ namespace arcticdb::async {
     }
 
     pipelines::SegmentAndSlice DecodeSliceTask::decode_into_slice(storage::KeySegmentPair&& key_segment_pair) {
-        auto key = std::move(key_segment_pair.atom_key());
-        auto seg = std::move(key_segment_pair.release_segment());
+        auto key = key_segment_pair.atom_key();
+        auto& seg = key_segment_pair.segment();
```
alexowens90, line 40: Can this be `const&`?

cpp/arcticdb/async/tasks.hpp
```diff
@@ -315,6 +291,113 @@ struct CopyCompressedTask : BaseTask {
...
+    entity::VariantKey key_to_read_;
+    std::optional<entity::AtomKey> key_to_write_;
+    bool check_key_exists_on_targets_;
+    bool retry_on_failure_;
+    std::shared_ptr<Store> source_store_;
+    std::vector<std::shared_ptr<Store>> target_stores_;
+    std::shared_ptr<BitRateStats> bit_rate_stats_;
+
+    // Returns an empty set if the copy succeeds, otherwise the set contains the names of the target stores that failed
+    std::set<std::string, std::less<>> copy() {
```
alexowens90, line 352: Why specifying `std::less<>`?

alexowens90: And does it need to be ordered?

cpp/arcticdb/async/tasks.hpp
```diff
@@ -315,6 +291,113 @@ struct CopyCompressedTask : BaseTask {
     }
 };
 
+// Used in arcticdb-enterprise, do not remove without checking whether it is still used there
+struct CopyCompressedInterStoreTask : async::BaseTask {
+
+    struct AllOk {};
```
alexowens90, line 297: This isn't a pattern we use anywhere else to my knowledge. Could just use `std::optional`?

cpp/arcticdb/async/tasks.hpp
```diff
@@ -315,6 +291,113 @@ struct CopyCompressedTask : BaseTask {
...
+
+    struct FailedTargets {
```
alexowens90, line 299: Why wrapped in a struct?

alexowens90 (changes requested):

cpp/arcticdb/async/tasks.cpp
poodlewars: No, because of `decode_into_memory_segment`. Sorting that out forces us in to the "make Segment logically const" refactor that we declined.

cpp/arcticdb/async/tasks.hpp
poodlewars: The ordering seems just to be used for logging in enterprise so I'll make this unordered. The `std::less<>` was to resolve a Sonarlint warning that was probably spurious.

cpp/arcticdb/async/tasks.hpp
poodlewars: I've tidied this up a bit to,

```
    using AllOk = std::monostate;
    using FailedTargets = std::unordered_set<std::string>;
    using ProcessingResult = std::variant<AllOk, FailedTargets>;
```

which I prefer over `std::optional` as the names describe the meaning of empty vs populated optional. We have a similar pattern with `StorageResult`.

alexowens90 (approved):

cpp/arcticdb/async/tasks.cpp
willdealtry: What happens in decode_into_memory_segment?

PR #1762, opened 2024-08-12
Add Storage API to check for existence of a key matching a predicate

poodlewars:
Useful for enterprise (where we want to check if op logs exist on secondaries as a safety guard).

Remove `test_mongo_storage.cpp` - it just returned immediately and was a maintenance burden to keep it compiling.

The important API changes are in `storage.hpp`.

We add a new kind of `IterateTypeVisitor` - an `IterateTypePredicate`. We change the internal `do_iterate_type` functions to take the predicate instead of the visitor. We change the public `iterate_type` to coerce its `IterateTypeVisitor` in to a degenerate `IterateTypePredicate` that never matches anything, so this call still iterates all keys.

Finally we add `test_storage_operations.cpp` to test the new API across all the supported storage backends.


cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -270,9 +270,9 @@ namespace s3 {
             };
         }
 
-        template<class Visitor, class KeyBucketizer, class PrefixHandler>
-        void do_iterate_type_impl(KeyType key_type,
-                                  Visitor &&visitor,
+        template<class KeyBucketizer, class PrefixHandler>
+        bool do_iterate_type_impl(KeyType key_type,
```
willdealtry, line 274: Since we're thinking about code formatting, in the case where we have multiple arguments I would prefer the first argument to start on the next line with a single indent, otherwise the indent depends on the length of the function name and files as a whole end up looking a mess

willdealtry: Sorry to be boring about it, but my preference is also to have a double indent on parameter lists, which helps disambiguate them from the following code, without needing floating braces

cpp/arcticdb/storage/storage.hpp
```diff
@@ -158,8 +158,16 @@ class Storage {
         return do_key_exists(key);
     }
 
+    bool key_exists(KeyType key_type, const IterateTypePredicate& predicate) {
```
willdealtry, line 161: Is this potentially a source of confusion given the radically different performance characteristics of the two methods? One is constant and the other is worst-case linear in the number of keys of that type - a difference of less than a second vs potentially several hours. I think I'd prefer if the scanning version was more up-front about what it's doing - scan_for_matching_key or something like that

cpp/arcticdb/storage/test/test_embedded.cpp
```diff
@@ -16,182 +16,185 @@
...
 #include <stdexcept>
 #include <arcticdb/entity/atom_key.hpp>
-#include <arcticdb/util/buffer.hpp>
 #include <arcticdb/entity/types.hpp>
 #include <arcticdb/util/test/test_utils.hpp>
 #include <arcticdb/util/random.h>
 #include <arcticdb/stream/row_builder.hpp>
-#include <arcticdb/stream/aggregator.hpp>
+
+namespace test_embedded {
```
willdealtry, line 24: I know it's not your naming, but I'm a bit at a loss to explain what 'test_embedded' actually tests, or what a BackendGenerator is (a thing that makes storages, apparently). Also, I think that anonymous namespaces are probably fine for cpp files in test code, we don't need to give them unique namespace names.

willdealtry: It looks like this is the one that tests local storages, so the generator could be a LocalStorageGenerator, and in the tet below we have a RemoteStorageGenerator that make the other kinds

cpp/arcticdb/storage/s3/detail-inl.hpp
poodlewars: I've pushed a change with your preferred formatting for this one, but really it's a mess even within this file let alone across the project. Motivates the automated formatting checks, I suppose.

cpp/arcticdb/storage/test/test_embedded.cpp
poodlewars: Thanks, I've tidied this up quite a bit and put the Generator somewhere it can be shared across the two test files.

willdealtry (approved):

Issue #1737, opened 2024-08-01
Calling `sort_and_finalize_staged_data` throws non informative exception


vasil-pashov:
**Describe the bug**
If nothing is added to the staged segment and `sort_and_finalize_staged_data` an exception is thrown.

**Steps to reproduce**
```python
import pandas as pd
import numpy as np
import arcticdb as adb

ac = adb.Arctic("lmdb://test")
lib = ac.get_library("test", create_if_missing=True)
lib.sort_and_finalize_staged_data("sym")
```

Output:
```
[2024-08-01 12:48:34.144] [arcticdb] [error] E_ASSERTION_FAILURE Stream descriptor not found in pipeline context
Traceback (most recent call last):
  File "...\test.py", line 7, in <module>
    lib.sort_and_finalize_staged_data("sym")
  File "...arcticdb\version_store\library.py", line 959, in sort_and_finalize_staged_data
    self._nvs.version_store.sort_merge(symbol, None, mode == StagedDataFinalizeMethod.APPEND, prune_previous_versions=prune_previous_versions)
arcticdb_ext.exceptions.InternalException: E_ASSERTION_FAILURE Stream descriptor not found in pipeline context
```

**Expected result**
Improve the exception message.

**Related issues**
Possibly related to: #1735, #1736

**OS, Python, Arctic versions**
```
Python: 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
OS: Windows-10-10.0.22631-SP0
ArcticDB: 4.5.0rc1
```

Issue #1736, opened 2024-08-01
Calling `sort_and_finalize_staged_data` with staged segment of DataFrame with 0 rows fails


vasil-pashov:
**Describe the bug**
When the staging area is consisted of DataFrames without any rows calling `sort_and_finalize_staged_data` fails.

**Steps to reproduce**
```python
import pandas as pd
import numpy as np
import arcticdb as adb

ac = adb.Arctic("lmdb://test")
lib = ac.get_library("test", create_if_missing=True)
df = pd.DataFrame({"col": []}, index=pd.DatetimeIndex([]))
lib.write("sym", df, staged=True)
lib.sort_and_finalize_staged_data("sym")
```

Ouput:
```
Traceback (most recent call last):
  File "...test.py", line 8, in <module>
    lib.write("sym", df, staged=True)
  File "...arcticdb\version_store\library.py", line 461, in write
    return self._nvs.write(
  File "...arcticdb\version_store\_store.py", line 583, in write
    self.version_store.write_parallel(symbol, item, norm_meta, udm)
arcticdb_ext.exceptions.InternalException: E_ASSERTION_FAILURE Allocate data called with zero size
```

**Expected behavior**
This should succeed and be a noop.

**Related issues**
This could be related to: #1735, #1737

**OS, Python, Arctic versions**
```
Python: 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
OS: Windows-10-10.0.22631-SP0
ArcticDB: 4.5.0rc1
```

PR #1763, opened 2024-08-14
Fixes for sort and finalize

vasil-pashov:
#### Reference Issues/PRs
Fixes: #1737 #1736 #1735

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -285,6 +285,67 @@ def test_parallel_writes_and_appends_index_validation(arctic_library, finalize_m
...
+        lib = lmdb_library
+        index = pd.DatetimeIndex([pd.Timestamp(2024, 1, 1), pd.Timestamp(2024, 1, 3), pd.Timestamp(2024, 1, 4)])
+        df1 = pd.DataFrame({"col": [1, 2, 3]}, index=index)
+        df2 = pd.DataFrame({})
+        df3 = pd.DataFrame({"col": [4]}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 5)]))
+        lib.write("sym", df1, staged=True)
+        lib.write("sym", df2, staged=True)
+        lib.write("sym", df3, staged=True)
+        lib.finalize_staged_data("sym", mode=finalize_method)
+        assert_frame_equal(lib.read("sym").data, pd.concat([df1, df2, df3]).sort_index())
```
alexowens90, line 307: Why does this need `sort_index`?

python/tests/unit/arcticdb/version_store/test_sort_merge.py
alexowens90, line 1: Is there already a test if you call `sort_and_finalize_staged_data` with `StagedDataFinalizeMethod.APPEND` and the incomplete segments overlap with the existing data?

cpp/arcticdb/stream/append_map.cpp
```diff
@@ -147,17 +147,24 @@ SegmentInMemory incomplete_segment_from_frame(
...
         auto copy_prev_key = prev_key;
         auto timeseries_desc = index_descriptor_from_frame(frame, existing_rows, std::move(prev_key));
         util::check(!timeseries_desc.fields().empty(), "Expected fields not to be empty in incomplete segment");
         auto norm_meta = timeseries_desc.proto().normalization();
         auto descriptor = timeseries_desc.as_stream_descriptor();
+
+        SegmentInMemory output;
+        if (num_rows == 0) {
+            output = SegmentInMemory(FixedSchema{descriptor, index}.default_descriptor(), 0, false, false);
+            output.set_timeseries_descriptor(pack_timeseries_descriptor(descriptor, existing_rows + num_rows, std::move(copy_prev_key), std::move(norm_meta)));
```
alexowens90, line 164: `num_rows` here is always zero

cpp/arcticdb/version/version_core.cpp
```diff
@@ -847,11 +853,15 @@ void check_incompletes_index_ranges_dont_overlap(const std::shared_ptr<PipelineC
...
         for (auto it = pipeline_context->incompletes_begin(); it!= pipeline_context->end(); it++) {
+            if (it->slice_and_key().slice().rows().diff() == 0) {
+                continue;
+            }
             sorting::check<ErrorCode::E_UNSORTED_DATA>(
                     !last_existing_index_value.has_value() || it->slice_and_key().key().start_time() >= *last_existing_index_value,
                     "Cannot append staged segments to existing data as incomplete segment contains index value < existing data (in UTC): {} <= {}",
                     date_and_time(it->slice_and_key().key().start_time()),
-                    date_and_time(*last_existing_index_value));
+                    last_existing_index_value ? date_and_time(*last_existing_index_value) : "NaT"
```
alexowens90, line 863: This doesn't make sense. This exception will only be throw if `last_existing_index_value.has_value()` is true

cpp/arcticdb/version/version_core.cpp
```diff
@@ -789,22 +789,28 @@ void read_incompletes_to_pipeline(
     if(incomplete_segments.empty())
         return;
 
+    const auto* seg = &incomplete_segments.front().segment(store);
```
alexowens90, line 792: Why is this now a pointer? What was it before?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1301,10 +1315,10 @@ VersionedItem sort_merge_impl(
...
                 update_info.previous_index_key_->end_time() - 1 > std::get<timestamp>(TimeseriesIndex::start_value_for_segment(segments[0].segment_.value()))) {
                 store->remove_keys(delete_keys).get();
-                user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>(
-                    "Cannot append index starting at {} to data frame with index ending at {}",
-                    std::get<timestamp>(TimeseriesIndex::start_value_for_segment(segments[0].segment_.value())),
-                    update_info.previous_index_key_->end_time() - 1
+                sorting::raise<ErrorCode::E_UNSORTED_DATA>(
+                    "Cannot append staged segments to existing data as incomplete segment contains index value < existing data (in UTC): {} <= {}",
+                    date_and_time(update_info.previous_index_key_->end_time() - 1),
+                    date_and_time(std::get<timestamp>(TimeseriesIndex::start_value_for_segment(segments[0].segment_.value())))
```
alexowens90, line 1321: These arguments are formatted the wrong way round

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_sort_merge.py
vasil-pashov: Yes, `test_appended_df_interleaves_with_storage`

python/tests/integration/arcticdb/test_arctic.py
vasil-pashov: It's a leftover I had a case with overlapping segments. It's not needed anymore.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: That is true, but we're passing function arguments and by standard they are always evaluated and the order of evaluation is not defined. So even if the condition for throwing is false the rest of the arguments will be evaluated including the call to `*last_existing_index_value` which throws when the optional is empty.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: It was a reference pointing to the first segment in the list. However the first segment can be empty and then the normalization metadata is wrong (this is what caused the problem with wrong column names). Now I'm scanning until I find the first non-empty segment. It's easier to keep track using a pointer.

cpp/arcticdb/version/version_core.cpp
alexowens90: Ah I see. In that case I would replace the `NaT` with an empty string and add a comment about it being unreachable but needed for this reason

cpp/arcticdb/version/version_core.cpp
alexowens90: Ah I get it. What's the behaviour if all of the segment row counts are 0? Line 792 means we'll still have `seg` point at the first incomplete segment, which doesn't seem intentional. If it was initialised to `nullptr`, and then was still `nullptr` after the loop below, what should we do?

alexowens90 (changes requested):

cpp/arcticdb/version/version_core.cpp
vasil-pashov: This shouldn't be nullptr in any case. It's initialized with the first element of incomplete_segments. If incomplete segments is empty the function returns (one line above ln. 790). If all segments are empty it will get the default normalization metadata an empty dataframe has.

alexowens90: OK, this logic is extremely non-obvious, please add some comments to explain why it works

alexowens90 (approved):

PR #1775, opened 2024-08-19
Limit 3 item ref key bypass to only when loading undeleted versions

G-D-Petrov:
Change the logic for the 3 item ref key optimization/bypass to be used only when loading undeleted versions.
Otherwise we might skip versions that have been deleted but are part of a snapshot.

Also removes a catch for KeyNotFoundException when writing keys as this catch might cause us to silently ignore real problems.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/async/test/test_async.cpp
+++ b/cpp/arcticdb/async/test/test_async.cpp
@@ -404,7 +404,7 @@ TEST(Async, CopyCompressedInterStoreNoSuchKeyOnWrite) {
     auto targets = std::vector<std::shared_ptr<arcticdb::Store>>{
         create_store(library_path, library_index, user_auth, codec_opt),
         create_store(library_path, failed_library_index, user_auth, codec_opt),
-        create_store(library_path, failed_library_index, user_auth, codec_opt)
+        create_store(library_path, library_index, user_auth, codec_opt)
     };
 
     // When - we write a key to the source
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/version/version_map.hpp
```diff
@@ -159,7 +159,10 @@ class VersionMapImpl {
             cached_penultimate_index = ref_entry.keys_[1];
         }
 
-        if (key_exists_in_ref_entry(load_strategy, ref_entry, cached_penultimate_index)) {
+        // The 3 item ref key bypass can be used only when we are loading undeleted versions
+        // because otherwise it might skip versions that are deleted but part of snapshots
+        if (load_strategy.load_objective_ == LoadObjective::UNDELETED_ONLY
```
willdealtry, line 164: 2 things: Firstly I don't understand the scenario where this is a problem. Is it to do with the slow path where we allow looking for specific version ids that are deleted in the version list but are contained in snapshots?

Secondly, most of the machinery for deciding whether to use the ref-key only or proceed with loading the version list is in `key_exists_in_ref_entry`, wouldn't it be cleaner if this check just returned false from that method?

G-D-Petrov: yeah sorry, I should've given more details, this is a minimal repro/test for the scenario:
```
append_small_df(source_lib, "tmp")  # v0
append_small_df(source_lib, "tmp")  # v1
source_lib.snapshot("snap")
source_lib.delete_version("tmp", 1)
append_small_df(source_lib, "tmp")  # v2

library_replicator.process_library(syncing_libraries)
assert compare_libs(source_lib, target_libs)
```
Essentially, version 1 is being deleted but it should be replicated to the secondary because it is in a snapshot.
When using append to write v2, it writes the v2 index key and the v0 index key, because it is the previous undeleted version.
Then when trying to get the index key for v1 during replication, we fail to get it because:

- due to [this optimization of the DOWNTO load strategy](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/version/version_utils.hpp#L341) (as 1 => 0)
- we enter [this logic for loading the index keys](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/version/version_map.hpp#L163), which leads to loading only the index keys in the vref (so only 0 and 2)
- and we fail when we try to filter for v1 in the get_index_key function

As for the second point, yeah I was debating with myself where it would be better to put this check, I will move it to the `key_exists_in_ref_entry` tomorrow

G-D-Petrov:
- commit [6417b20](https://github.com/man-group/ArcticDB/pull/1775/commits/6417b2017a17004620ee744e42281624eaa88c74) - the actual fix for the PR
- commit [ed5d9ec](https://github.com/man-group/ArcticDB/pull/1775/commits/ed5d9ec57b495bb370db2e05e935cf43735905fb) - test for the fix
- commit [3c128d7](https://github.com/man-group/ArcticDB/pull/1775/commits/3c128d778d6bf4b1646af5a4877a7162f5b0de83) - unrelated to this PR, I improved another test a bit

willdealtry (approved):

PR #1793, opened 2024-08-23
Add interface to pass a cached entry to tombstone all

G-D-Petrov:
#### Reference Issues/PRs
Needed for enterprise issue https://github.com/man-group/ArcticDB/issues/151

There are use cases where we want to:
- add an index key
- add a version key for it (**don't add a ref key here**)
- tombstone them
- add a ref key for the tombstone

Currently, the tombstone all operation relies on the ref key load and reconstruct the version chain (through `check_reload`).
This PR adds a way to pass a `cached_entry` with the index key and version key (or any other keys) already written, so we don't need to rely on the ref key.

The added test illustrates the current behavior (without the cached entry) and the desired new behavior of the functionality.

#### Any other comments?
Another way to achieve this would be to use `load_via_iteration` but this seems like a more flexible approach to me, especially as we would have an entry already on hand, if we are writing any individual keys.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_map.hpp
+++ b/cpp/arcticdb/version/version_map.hpp
@@ -254,13 +254,19 @@ public:
     std::pair<VersionId, std::vector<AtomKey>> tombstone_from_key_or_all(
             std::shared_ptr<Store> store,
             const StreamId& stream_id,
-            std::optional<AtomKey> first_key_to_tombstone = std::nullopt
+            std::optional<AtomKey> first_key_to_tombstone = std::nullopt,
+            std::optional<std::shared_ptr<VersionMapEntry>> cached_entry = std::nullopt
             ) {
-        auto entry = check_reload(
-            store,
-            stream_id,
-            LoadStrategy{LoadType::ALL, LoadObjective::UNDELETED_ONLY},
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/version/version_map.hpp
```diff
@@ -254,13 +254,19 @@ class VersionMapImpl {
...
+            std::optional<AtomKey> first_key_to_tombstone = std::nullopt,
+            std::optional<std::shared_ptr<VersionMapEntry>> cached_entry = std::nullopt
             ) {
-        auto entry = check_reload(
-            store,
-            stream_id,
-            LoadStrategy{LoadType::ALL, LoadObjective::UNDELETED_ONLY},
-            __FUNCTION__);
+        std::shared_ptr<VersionMapEntry> entry;
+        if (cached_entry)
```
poodlewars, line 261: I'd rather not have the one line if as it's quite error prone in the face of merges / refactoring. Especially since here the statement in the `else` case spans multiple lines in the file.

poodlewars (approved):

PR #1796, opened 2024-08-27
Fix sort index

IvoDD:
Fixes the `sort_index` and `delete_range` wrong version bug.

Bug description:
We used to calculate the version to use for these two cases based on the last undeleted version which is wrong.
For example if we have the following version chain:
```
v0 <- v1 <- v2 <- tombstone (v2) <- tombstone (v1)
```
Calling `sort_index` or `delete_range` would result in us writing `v1` instead of the correct `v3`. This has various bad implications.

So this pr does the following:
1. One commit adds a test showcasing the bug and an assertion that all index keys should be increasing in version.
2. One commit which fixes the bug and uses the `UpdateInfo` construct as other modification methods.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_version_chain.py
+++ b/python/tests/unit/arcticdb/version_store/test_version_chain.py
@@ -1,0 +1,56 @@
+"""
+Copyright 2023 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+import pandas as pd
+import numpy as np
+from datetime import datetime
+import pytest
+
+from pandas import MultiIndex
+from arcticdb.version_store import NativeVersionStore
+from arcticdb_ext.exceptions import (
+    InternalException,
+    NormalizationException,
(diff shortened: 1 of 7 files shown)


python/tests/unit/arcticdb/version_store/test_version_chain.py
```diff
@@ -0,0 +1,54 @@
...
+        if operation == "update":
+            df_update = pd.DataFrame({"col": [4, 5]}, index=pd.date_range(start=pd.Timestamp(1), periods=2, freq='ns'))
+            lib.update(sym, df_update)
+        if operation == "append":
+            df_append = pd.DataFrame({"col": [4, 5]}, index=pd.date_range(start=pd.Timestamp(3), periods=2, freq='ns'))
+            lib.append(sym, df_append)
+        if operation == "sort_index":
+            lib.version_store.sort_index(sym, False, False)
+        if operation == "delete_range":
+            lib.delete(sym, date_range=(pd.Timestamp(1), pd.Timestamp(1)))
```
poodlewars, line 40: (minor) Raise if unknown `operation`

IvoDD: Thanks, fixed it.

PR #1799, opened 2024-08-29
More sort and finalize fixes

### LINKED ISSUE #1466: `compact_incomplete` with `append=True` should fail if dynamic schema is false and incomplete segment fields do not exactly match the existing fields

### LINKED ISSUE #1738: Allow unordered indexes in staged area when `sort_and_finalize_staged_data` is used
Currently `sort_and_finalize_staged_data` the indexes in all segments to be sorted. Or an exception is thrown.
```python
import pandas as pd
import numpy as np
import arcticdb as adb

ac = adb.Arctic("lmdb://test")
dates = [np.datetime64('2023-01-03'), np.datetime64('2023-01-01'), np.datetime64('2023-01-05')]
df = pd.DataFrame({"col": [2, 1, 3]}, index=dates)
lib.write("sym", df, staged=True)
lib.sort_and_finalize_staged_data("sym")
```

Output:
```
  File "...\test.py", line 9, in <module>
    lib.write("sym", df, staged=True)
  File "...\arcticdb\version_store\library.py", line 461, in write
    return self._nvs.write(
  File "...\arcticdb\version_store\_store.py", line 583, in write
    self.version_store.write_parallel(symbol, item, norm_meta, udm)
arcticdb_ext.exceptions.UnsortedDataException: E_UNSORTED_DATA When writing/appending staged data in parallel, input data must be sorted.
```

Allow unordered indexes in staged segments and sort then when `sort_and_finalize_staged_data` is called.

### LINKED ISSUE #1781: Sort and finalize staged data fails with pre-epoch dates
### Describe the bug

When the index in any of the staged segments contains pre-epoch dates calling `sort_and_finalize_staged_data` fails.

### Steps/Code to Reproduce

```python
import pandas as pd
import numpy as np
import arcticdb as adb

ac = adb.Arctic("lmdb://test")
df = pd.DataFrame({"b": [0]}, index=pd.DatetimeIndex([pd.Timestamp(1969, 12, 31, 12, 59, 59)]))
lib.write(symbol, df, staged=True)
lib.sort_and_finalize_staged_data(symbol)
```

Throws:
```
InternalException: E_INVALID_ARGUMENT timestamp decreasing, current val=0, candidate=-39601000000000
```

### Expected Results

Work just as append and write store pre-epoch dates.

OS: Windows-10-10.0.22631-SP0
ArcticDB: master

### Backend storage used

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1795: Sort and finalize can reorder the fields in a segment and crash with static schema
### Describe the bug

Sort and finalize uses merge_descriptors to generate the field descriptor for the newly added segment. After that during the merging phase it creates an Aggregator and strips the fields from the merged descriptor leaving the aggregator to create the field collection. This later leads to crash on write since the segment field descriptor is different than the one in the header.

In the example below merge_descriptors would order the fields in the field descriptor in order of appearance: index - 0, a - 1, b - 2. The final sorted segment adds rows one by one in the order of their index, thus column b will be reported first and it will have index 1 after that a will have index 2.

### Steps/Code to Reproduce

```python
import numpy as np
import pandas as pd
import arcticdb

idx1 = pd.DatetimeIndex([
    pd.Timestamp("2024-01-02")
])
df1 = pd.DataFrame({
     "a": np.array([1], dtype="float"),
     "b": np.array([22250], dtype="int64")
}, index=idx1)

b = np.array([-53979, -53973], dtype="int64")

idx = pd.DatetimeIndex([
    pd.Timestamp("2024-01-03"),
    pd.Timestamp("2024-01-01")
])

df2 = pd.DataFrame({"b": b}, index=idx)

lib.write("sym", df1, staged=True)
lib.write("sym", df2, staged=True)
lib.sort_and_finalize_staged_data("sym")
lib.read("sym")
```

### Expected Results

Create the right field descriptor and do not throw.

OS: Windows-10-10.0.22631-SP0
ArcticDB: dev

### Backend storage used

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1797: Sort and finalize and regular write/append handle NaT index values differently
### Describe the bug

When the index contains NaT regular append/write throw sorting exception while sort and finalize allows NaT values in the index.

### Steps/Code to Reproduce

```python
import arcticdb
import pandas as pd
import numpy as np

df1 = pd.DataFrame({"a": [1]}, index=pd.DatetimeIndex([pd.NaT]))
lib.write("sym", df1, staged=True)
lib.sort_and_finalize_staged_data("sym") # Does not throw

lib.write("sym_2", df) # Throws
```

### Expected Results

The two should behave the same way and not allow NaT values in the index

Python: 3.10.11 (tags/v3.10.11:7d4cc5a, Apr 5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
OS: Windows-10-10.0.22631-SP0
ArcticDB: dev

### Backend storage used

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1807: Sort and finalize throws when staged columns are of different but promotable types (dynamic schema)
### Describe the bug

If one column is of type int64 and the same column is in another segment but of type int32 sort and finalize will throw. However the stream descriptor will be of the largest type because merge descriptors uses `valid_common_type`.

### Steps/Code to Reproduce

```python
import pandas as pd
import numpy as np
import arcticdb as adb
from arcticdb.version_store.library import StagedDataFinalizeMethod
ac = adb.Arctic("lmdb://test")
opts = adb.LibraryOptions(dynamic_schema=True)
symbol = "test"
lib.write("sym", pd.DataFrame({"a": np.array([1], dtype="int32")}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 1)])), staged=True)
lib.write("sym", pd.DataFrame({"a": np.array([1], dtype="int64")}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 2)])), staged=True)
lib.sort_and_finalize_staged_data("sym", mode=StagedDataFinalizeMethod.WRITE)
```

Throws:
```
InternalException: E_ASSERTION_FAILURE Type mismatch in set_scalar, expected 4
```

Note the exception is thrown after all descriptors were resolved, during the merge phase. The exception is thrown when setting a value in a column.

`finalize_staged_data` does not allow having staged of different (but compatible types). The difference is due to [this](https://github.com/man-group/ArcticDB/blob/bdf44d6a0268c55fcd5d3da619812325e0e8b44f/cpp/arcticdb/processing/clause.cpp#L919)

### Expected Results

Two options:
 * Promote the "smaller" type and succeed
 * Act as `finalize_staged_segments` and require exactly the same types in all staged segments

OS: Windows-10-10.0.22631-SP0
ArcticDB: 4.5.0

### Backend storage used

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1828: date_and_time does not work with pre-epoch timestamps
### Describe the bug

[date_and_time](https://github.com/man-group/ArcticDB/blob/52cdc81b7f175ad410b4a3626b366e3ab1aec3a1/cpp/arcticdb/util/timer.hpp#L296) uses gmtime which does not work with pre-epoch dates in that case gmtime returns nullptr.

### Steps/Code to Reproduce

N/A

### Expected Results

Make it so that it works with pre-epoch dates or throw reasonable exception.

Python: 3.10.11 (tags/v3.10.11:7d4cc5a, Apr 5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
OS: Windows-10-10.0.22631-SP0
ArcticDB: 4.5.0

### Backend storage used

_No response_

### Additional Context

_No response_

vasil-pashov:
#### Reference Issues/PRs

Fixes #1738 
Fixes #1781 
Fixes #1466 
Fixes #1795 
Fixes #1797
Fixes #1807
Fixes #1828

A notable change is that staged writes no longer validate the index is sorted. The validation is done at the moment compact_incompletes/finalize_staged_data/sort_and_finalize_staged_data is called. This is because sort_and_finalize_staged_data does not require the segments to be sorted, but the call for adding a staged segment is the same. We should add a separate call for that.

Note also that all incomplete keys for a symbol are deleted if any of the finalize calls fail. The other option is to leave the segments. In that case the user will have the responsibility of calling `delete_staged_data`.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/version/version_core.cpp
```diff
@@ -1426,16 +1496,21 @@ VersionedItem sort_merge_impl(
                 },
                 DynamicSchema{pipeline_context->descriptor(), index},
                 [pipeline_context=pipeline_context, &fut_vec, &store](SegmentInMemory &&segment) {
-                    auto local_index_start = TimeseriesIndex::start_value_for_segment(segment);
-                    auto local_index_end = TimeseriesIndex::end_value_for_segment(segment);
+                    const auto local_index_start = TimeseriesIndex::start_value_for_segment(segment);
+                    const auto local_index_end = TimeseriesIndex::end_value_for_segment(segment);
+                    for (const std::shared_ptr<Column>& col : segment.columns()) {
+                        if (col->row_count() == 0) {
+                            col->default_initialize_rows(0, segment.row_count(), true);
```
vasil-pashov, line 1503: I think think that this should be handled by a sparse_map but I could find how to make it work. I tried calling set_sparse_map and append_sparse_map but it didn't work.

python/tests/hypothesis/arcticdb/test_sort_merge.py
```diff
@@ -8,64 +7,316 @@
...
+        lib.sort_and_finalize_staged_data(symbol, mode)
+
+def has_nat_in_index(segment_list):
+    return any(pd.NaT in segment.index for segment in segment_list)
+
+def merge_and_sort_segment_list(segment_list, int_columns_in_df=None):
+    merged = pd.concat(segment_list)
+    # pd.concat promotes dtypes. If there are missing values in an int typed column
+    # it will become float column and the missing values will be NaN.
+    int_columns_in_df = int_columns_in_df if int_columns_in_df else []
```
vasil-pashov, line 84: On a second thought this is not quite reliable especially with dynamic schema. Columns can start as int then get promoted to float. I think if we can make sparse maps work this check will work. Until then I'm thinking of replacing all nans with 0 in both source and expected.

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -142,8 +142,8 @@ def test_write_parallel_sort_merge(basic_arctic_library, prune_previous_versions
     dataframes = []
     df = pd.DataFrame()
 
+    cols = random.sample(columns, 4)
```
alexowens90, line 145: Assume this change is necessary because `basic_arctic_library` has static schema? In which case, just generate 4 random column names instead of generating 8 and sampling 4

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -312,8 +312,10 @@ def test_parallel_sortedness_checks_unsorted_data(lmdb_version_store, append, va
         with pytest.raises(SortingException):
             if append:
                 lib.append(sym, df_1, incomplete=True, validate_index=True)
+                lib.compact_incomplete(sym, True, False)
             else:
                 lib.write(sym, df_1, parallel=True, validate_index=True)
+                lib.compact_incomplete(sym, True, False)
```
alexowens90, line 318: These compactions won't be necessary when the change I requested in `_store.py` is made

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -135,35 +136,29 @@ def test_merge_strings_dynamic(lmdb_library):
...
     lib.sort_and_finalize_staged_data("sym")
-    expected_index = pd.DatetimeIndex([np.datetime64('2023-01-01'), np.datetime64('2023-01-01'), np.datetime64('2023-01-03'),
-                                       np.datetime64('2023-01-03'), np.datetime64('2023-01-05'), np.datetime64('2023-01-05')],
-                                      dtype="datetime64[ns]")
     data = lib.read("sym").data
-    assert data.index.equals(expected_index)
-    for i in range(0, len(data["col"])):
-        assert data["col"][i] == df1["col"][i // 2] or data["col"][i] == df2["col"][i // 2]
+    expected = pd.concat([df1, df2]).sort_index()
+    assert data.index.equals(expected.index)
```
alexowens90, line 157: We can also check that the values in `col` are correct. First value is 1 or 4, second value is 1 or 4 and not equal to the first value, etc

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+        assert "col_0" in str(exception_info.value)
+        assert "col_1" in str(exception_info.value)
+
+    def test_appending_reordered_column_set_throws(self, lmdb_library):
+        lib = lmdb_library
+
+        df1 = pd.DataFrame({"col_0": [1], "col_1": ["test"], "col_2": [1.2]}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 1)]))
+        lib.write("sym", df1, staged=True)
+
+        df2 = pd.DataFrame({"col_1": ["asd"], "col_2": [2.5], "col_1": [2]}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 2)]))
```
alexowens90, line 325: Duplicated `col_1` in the dict seems like a mistake

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+        lib.write("sym", df2, staged=True)
+
+        with pytest.raises(SchemaException) as exception_info:
+            lib.sort_and_finalize_staged_data("sym", mode=mode)
+        assert "col_0" in str(exception_info.value)
+        assert "FLOAT" in str(exception_info.value)
+        assert "INT" in str(exception_info.value)
+
+    @pytest.mark.parametrize("mode", [StagedDataFinalizeMethod.APPEND, StagedDataFinalizeMethod.WRITE])
+    def test_type_mismatch_in_staged_segments_throws_with_promotoable_types(self, lmdb_library, mode):
```
alexowens90, line 365: These types aren't promotable?

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+        assert "UTF_DYNAMIC" in str(exception_info.value)
+
+    @pytest.mark.parametrize("mode", [StagedDataFinalizeMethod.APPEND, StagedDataFinalizeMethod.WRITE])
+    def test_staged_segments_cant_be_reordered(self, lmdb_library, mode):
+        lib = lmdb_library
+
+        df1 = pd.DataFrame({"col_0": [1], "col_1": ["test"], "col_2": [1.2]}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 1)]))
+        lib.write("sym", df1, staged=True)
+        
+        df2 = pd.DataFrame({"col_1": ["asd"], "col_2": [2.5], "col_1": [2]}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 2)]))
```
alexowens90, line 386: Duplicated `col_1` in the dict seems like a mistake

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+        with pytest.raises(SchemaException) as exception_info:
+            lib.sort_and_finalize_staged_data("sym", mode=StagedDataFinalizeMethod.APPEND)
+        assert "col_1" in str(exception_info.value)
+        assert "col_2" in str(exception_info.value)
+
+    def test_cannot_append_reordered_columns(self, lmdb_library):
+        lib = lmdb_library
+        self.init_symbol(lib, "sym")
+        
+        df = pd.DataFrame({"col_1": [1.4], "col_0": [5], "col_2": ["val"]}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 1)]))
```
alexowens90, line 416: On Linux, `col_0` will have dtype `int64` here, force it to match that in `init_symbol`

alexowens90: Make sure all dtypes below are specified, so it is clear whether they are meant to match the `init_symbol` dtypes or not

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+        
+        df = pd.DataFrame({"col_0": np.array([1], dtype="int32"), "col_1": [5], "col_2": ["val"], "col_3": [1]}, index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 1)]))
+        lib.write("sym", df, staged=True)
+        with pytest.raises(SchemaException) as exception_info:
+            lib.sort_and_finalize_staged_data("sym", mode=StagedDataFinalizeMethod.APPEND)
+        assert "col_3" in str(exception_info.value)
+        
+        
+# This was a added as a bug repro for GH issue #1795.
+def test_two_columns_with_different_dtypes(lmdb_library_dynamic_schema):
```
alexowens90, line 456: More suitable as a nonreg test?

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+
+    idx1 = pd.DatetimeIndex([
+        pd.Timestamp("2024-01-02")
+    ])
+    df1 = pd.DataFrame({
+         "a": np.array([1], dtype="float"),
+         "b": np.array([22250], dtype="int64")
+    }, index=idx1)
+    
+    b = np.array([-53979, -53973], dtype="int64")
```
alexowens90, line 467: I assume the weirdly specific values `1/22250/-53979/-53973` aren't necessary to repro the original bug? Replace them with something more standard, otherwise it seems like these values are special somehow

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+
+    df2 = pd.DataFrame({"b": b}, index=idx)
+    
+    lib.write("sym", df1, staged=True)
+    lib.write("sym", df2, staged=True)
+    lib.sort_and_finalize_staged_data("sym")
+    lib.read("sym")
+
+@pytest.mark.parametrize("mode", [StagedDataFinalizeMethod.APPEND, StagedDataFinalizeMethod.WRITE])
+def test_nat_is_not_allowed_in_index(lmdb_library_static_dynamic, mode):
```
alexowens90, line 482: This isn't very comprehensive

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -222,50 +217,303 @@ def test_prune_previous(lmdb_library):
...
+    lib = lmdb_library_static_dynamic
+
+    df1 = pd.DataFrame({"a": [1]}, index=pd.DatetimeIndex([pd.NaT]))
+    lib.write("sym", df1, staged=True)
+    with pytest.raises(SortingException) as exception_info:
+        lib.sort_and_finalize_staged_data("sym", mode=mode)
+    assert "NaT" in str(exception_info.value)
+
+class TestSortMergeDynamicSchema:
```
alexowens90, line 491: Need some tests for type promotion logic as well

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -293,41 +293,125 @@ def test_parallel_writes_and_appends_index_validation(arctic_library, finalize_m
 
 @pytest.mark.parametrize("finalize_method", (StagedDataFinalizeMethod.APPEND, StagedDataFinalizeMethod.WRITE))
 class TestFinalizeWithEmptySegments:
-    def test_staged_segment_is_only_empty_dfs(self, lmdb_library, finalize_method):
-        lib = lmdb_library
+    def test_staged_segment_is_only_empty_dfs(self, arctic_library, finalize_method):
```
alexowens90, line 296: Why?

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -293,41 +293,125 @@ def test_parallel_writes_and_appends_index_validation(arctic_library, finalize_m
...
 
 @pytest.mark.parametrize("finalize_method", (StagedDataFinalizeMethod.APPEND, StagedDataFinalizeMethod.WRITE))
-def test_finalize_without_adding_segments(lmdb_library, finalize_method):
-    lib = lmdb_library
+def test_finalize_without_adding_segments(arctic_library, finalize_method):
+    lib = arctic_library
     with pytest.raises(UserInputException) as exception_info:
         lib.finalize_staged_data("sym", mode=finalize_method)
 
+class TestFinalizeStagedDataSchemaMismatch:
```
alexowens90, line 328: `finalize_staged_data` just passes arguments through to the V1 API, so the vast majority of the tests should be on the V1 API. So either move this to `test_parallel.py`, or remove it if it is duplicating tests already covered there

python/tests/hypothesis/arcticdb/test_sort_merge.py
alexowens90, line 1: This is quite an odd way to declaratively test this feature. Let's discuss on Monday when you're back

alexowens90 (changes requested):

cpp/arcticdb/processing/clause.cpp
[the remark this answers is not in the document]
willdealtry: sure

cpp/arcticdb/stream/merge.hpp
```diff
@@ -30,8 +44,33 @@ void do_merge(
             std::advance(val, IndexType::field_count());
             for(; val != next->row().end(); ++val) {
                 val->visit_field([&rb] (const auto& opt_v, std::string_view name, const TypeDescriptor& type_desc) {
-                    if(opt_v)
-                        rb.set_scalar_by_name(name, opt_v.value(), type_desc.data_type());
+                    if (opt_v) {
+                        const StreamDescriptor& descriptor = rb.descriptor();
+                        const std::optional<size_t> field_idx = descriptor.find_field(name);
+                        if (!field_idx || type_desc == descriptor.field(*field_idx).type()) {
```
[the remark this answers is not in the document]
willdealtry: It would be better to check the whole column as we load the segment, since within each segment all the elements in a column will be the same type.

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -601,7 +601,7 @@ struct ReduceColumnTask : async::BaseTask {
             } else {
                 column.default_initialize_rows(0, frame_.row_count(), false);
             }
-        } else {
+        } else if (column_data != slice_map_->columns_.end()) {
```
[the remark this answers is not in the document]
vasil-pashov: The slice map contains only the string typed columns. I'm adding the check taking this into account.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1407,13 +1475,15 @@ VersionedItem sort_merge_impl(
...
-                    date_and_time(update_info.previous_index_key_->end_time() - 1)
+            if (options.append_ && update_info.previous_index_key_ && !segments.empty()) {
+                const timestamp last_index_on_disc = update_info.previous_index_key_->end_time();
+                const timestamp incomplete_start =
+                    std::get<timestamp>(TimeseriesIndex::start_value_for_segment(segments[0].segment(store)));
+                sorting::check<ErrorCode::E_UNSORTED_DATA>(
+                    last_index_on_disc - 1 <= incomplete_start,
+                    "Cannot append staged segments to existing data as incomplete segment contains index value {} < existing data {}",
+                    incomplete_start,
+                    last_index_on_disc
```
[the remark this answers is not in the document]
vasil-pashov: date_and_time throws exception in MSVC. I couldn't exactly figure out why. The exception is because timeptr (which is some internal static for the standard library) is becomes nullptr.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1407,13 +1475,15 @@ VersionedItem sort_merge_impl(
...
             read_query.clauses_.emplace_back(std::make_shared<Clause>(MergeClause{timeseries_index, SparseColumnPolicy{}, stream_id, pipeline_context->descriptor()}));
             auto segments = read_and_process(store, pipeline_context, read_query, ReadOptions{}, pipeline_context->incompletes_after());
-            if (options.append_ && update_info.previous_index_key_.has_value() &&
-                update_info.previous_index_key_->end_time() - 1 > std::get<timestamp>(TimeseriesIndex::start_value_for_segment(segments[0].segment_.value()))) {
-                store->remove_keys(delete_keys).get();
-                sorting::raise<ErrorCode::E_UNSORTED_DATA>(
-                    "Cannot append staged segments to existing data as incomplete segment contains index value < existing data (in UTC): {} <= {}",
-                    date_and_time(std::get<timestamp>(TimeseriesIndex::start_value_for_segment(segments[0].segment_.value()))),
-                    date_and_time(update_info.previous_index_key_->end_time() - 1)
+            if (options.append_ && update_info.previous_index_key_ && !segments.empty()) {
```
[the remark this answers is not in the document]
vasil-pashov: This was before we decided that calling finalize without staged segments must throw. It's not needed anymore.

python/tests/integration/arcticdb/test_arctic.py
vasil-pashov: The whole file uses `arctic_library` as a fixture. I added `lmdb_library` by mistake.

cpp/arcticdb/version/version_core-inl.hpp
```diff
@@ -136,6 +137,13 @@ void do_compact(
             if (sk.slice().rows().diff() == 0) {
                 continue;
             }
+
+            const auto& segment = sk.segment(store);
+            sorting::check<ErrorCode::E_UNSORTED_DATA>(
+                !check_index_sorted || segment.descriptor().sorted() == SortedValue::ASCENDING,
```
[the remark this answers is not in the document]
vasil-pashov: Can we be sure that the old client wrote an ordered index?

cpp/arcticdb/stream/merge.hpp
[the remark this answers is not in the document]
vasil-pashov: This one is a bit tricky, since each staged segment can have different datatype and we're not processing segment by segment but row by row and consecutive rows can be from different segments. Anyway it's a valid concern, I'll think of some way to avoid it.

On a tangential line, can is possible to propagate if dynamic_schema is used to this function?

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
vasil-pashov: Edit: It's actually needed. Segments will be empty if a dataframe with 0 rows is staged and then finalize is called.

python/tests/unit/arcticdb/version_store/test_sort_merge.py
vasil-pashov: Adding one tests with only NaT and another with NaT and a valid data. If there's something else let me know.

python/tests/integration/arcticdb/test_arctic.py
vasil-pashov: Ah, I wasn't aware of `test_parallel.py` (when it comes to finalize.* methods the file naming is somewhat random). I thought that this is the place where `finalize_staged_data` is tested. I thing the correct place for this is `test_parallel.py`, moving it there.

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
alexowens90: OK, is `aggregator.commit();` happy and behave sanely if `segments` is empty?

cpp/arcticdb/version/version_core.cpp
alexowens90: Surely a staged segment with 0 rows always be compacted together with other incomplete segments? We're never going to write a data key with zero rows here, that would be pointless

cpp/arcticdb/version/version_core-inl.hpp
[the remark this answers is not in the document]
alexowens90: No, but we also don't know it's unordered, so we should allow the operation

cpp/arcticdb/version/version_core.cpp
vasil-pashov: Yeah, but imagine the situation where you have a dynamic schema.
Staged segment 1 has columns [a, b] and 0 rows
Staged segment 2 has columns[a] and 10 rows
It would column b into the descriptor and column b must have 10 rows. Currently they are default initialized but it would be cool if it could have sparse map (or empty type at some point)

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
vasil-pashov: It returns without calling any callback. That looks OK. However this means that [this ](https://github.com/man-group/ArcticDB/blob/52cdc81b7f175ad410b4a3626b366e3ab1aec3a1/cpp/arcticdb/version/version_core.cpp#L1432) won't be reached I'm not sure if not writing a TABLE_DATA key is breaking any invariant.

python/tests/hypothesis/arcticdb/test_sort_merge.py
```diff
@@ -1,71 +1,235 @@
...
+    sym = "test_sort_merge_static_schema_write"
+    for df in df_list:
+        lib.write(sym, df, staged=True, validate_index=False)
+    if len(df_list) == 0:
+        assert_cannot_finalize_without_staged_data(lib, sym, StagedDataFinalizeMethod.WRITE)
+        assert len(get_append_keys(lib, sym)) == 0
+        return
+    if not all(df_list[0].dtypes.equals(segment.dtypes) for segment in df_list):
+        assert_staged_columns_are_compatible(lib, sym, StagedDataFinalizeMethod.WRITE)
+        assert len(get_append_keys(lib, sym)) == 0
```
alexowens90, line 132: Could/should this assert (and the similar ones below) be rolled into the `assert_staged_columns_are_compatible`/`assert_nat_is_not_supported`/`assert_appended_data_does_not_overlap_with_storage` functions?

alexowens90 (changes requested):

python/tests/hypothesis/arcticdb/test_sort_merge.py
vasil-pashov: Yep, makes sense.

python/tests/hypothesis/arcticdb/test_sort_merge.py
```diff
@@ -1,71 +1,235 @@
...
+    column_infos = draw(st.lists(st.sampled_from(column_list), unique_by=lambda x: x.name, min_size=1))
+    columns = [hs_pd.column(name=ci.name, dtype=ci.dtype) if ci.dtype != 'object' else string_column_strategy(ci.name) for ci in column_infos]
+    if not IS_PANDAS_TWO:
+        # Due to https://github.com/man-group/ArcticDB/blob/7479c0b0caa8121bc2ca71a73e29769bbc41c66a/python/arcticdb/version_store/_normalization.py#L184
+        # we change the dtype of empty float columns. This makes hypothesis tests extremely hard to write as we must
+        # keep addional state about is there a mix of empty/non-empty float columns in the staging area, did we write
+        # empty float column (if so it's type would be object). These edge cases are covered in the unit tests.
+        index = hs_pd.indexes(dtype="datetime64[ns]", min_size=1 if min_size <= 0 else min_size).filter(lambda x: allow_nat_in_index or not pd.NaT in x)
+    else:
+        index = hs_pd.indexes(dtype="datetime64[ns]", min_size=min_size).filter(lambda x: allow_nat_in_index or not pd.NaT in x)
```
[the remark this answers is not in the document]
vasil-pashov: They are out of order. This is an example index this generated:
```python
DatetimeIndex(['1969-12-31 23:59:59.999999760',
               '1969-12-31 23:59:59.999999825',
               '1970-01-01 00:00:00.000004499',
               '1969-12-31 23:59:59.999999924',
               '2262-04-11 23:47:16.854775807',
               '1969-12-31 23:59:57.642695770',
               '1970-01-01 00:00:00.000047326']
 ```

File: python/tests/hypothesis/arcticdb/test_sort_merge.py [hunk 23]
```diff
@@ -1,71 +1,229 @@
...
+import string
+from arcticdb.util._versions import IS_PANDAS_TWO
+from pandas.api.types import is_numeric_dtype, is_integer_dtype, is_float_dtype
+from arcticdb.util.hypothesis import use_of_function_scoped_fixtures_in_hypothesis_checked
+from arcticdb.toolbox.library_tool import KeyType
 
 ColumnInfo = namedtuple('ColumnInfo', ['name', 'dtype'])
 
-COLUMN_DESCRIPTIONS = [ColumnInfo("a", "float"), ColumnInfo("b", "int64"), ColumnInfo("c", "str"), ColumnInfo("d", "datetime64[ns]")]
+COLUMNS = [f"col_{i}" for i in range(0, 5)]
```
willdealtry, line 20: is it possible to have an unsigned type and a few more columns? Five is very narrow.

vasil-pashov: Do you have any number on your mind? The only concern is that generating too many columns might slow the tests but we can play with it until we're happy.

alexowens90 (approved):

PR #1805, opened 2024-09-02
Pre arrow refactor

willdealtry:
Some changes to segment construction that allow a greater range of behaviours, moving these out from the main Arrow work as they are a pure refactoring at this point and touch a lot of files


cpp/arcticdb/CMakeLists.txt
```diff
@@ -501,7 +501,10 @@ set(arcticdb_srcs
         version/symbol_list.cpp
         version/version_map_batch_methods.cpp
         storage/s3/ec2_utils.cpp
-        util/buffer_holder.cpp util/native_handler.hpp)
+        util/buffer_holder.cpp
+        util/native_handler.hpp entity/output_type.hpp
+        entity/output_type.hpp
+        entity/output_type.hpp)
```
alexowens90, line 507: Duplicated, and out of order

cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -20,6 +22,7 @@ struct ReadOptions {
     std::optional<bool> set_tz_;
     std::optional<bool> optimise_string_memory_;
     std::optional<bool> batch_throw_on_error_;
+    OutputType output_type_ = OutputType::PANDAS;
```
alexowens90, line 25: Inconsistent with everything else being optional. I guess they all have implicit defaults, so I guess (not for now) the solution is to remove the optionality from all of those as well.
Unless there are fields where the default value depends on the code path?

cpp/arcticdb/stream/aggregator-inl.hpp
```diff
@@ -28,7 +28,7 @@ inline void Aggregator<Index, Schema, SegmentingPolicy, DensityPolicy>::commit_i
     if(final)
         return;
 
-    segment_ = SegmentInMemory(schema_policy_.default_descriptor(), segmenting_policy_.expected_row_size(), false, SparsePolicy::allow_sparse);
+    segment_ = SegmentInMemory(schema_policy_.default_descriptor(), segmenting_policy_.expected_row_size(), AllocationType::DYNAMIC, SparsePolicy::allow_sparse);
```
alexowens90, line 31: Is `SparsePolicy` now representing the same information as `Sparsity`?

cpp/arcticdb/stream/aggregator.hpp
```diff
@@ -131,12 +131,12 @@ class ListOfSegmentPolicies {
...
 class DenseColumnPolicy {
 public:
-    static constexpr bool allow_sparse = false;
+    static constexpr Sparsity allow_sparse = Sparsity::NOT_PERMITTED;
 };
 
 class SparseColumnPolicy {
 public:
-    static constexpr bool allow_sparse = true;
+    static constexpr Sparsity allow_sparse = Sparsity::PERMITTED;
```
alexowens90, line 139: What do these classes now achieve that templating directly on `Sparsity` would not?

cpp/arcticdb/pipeline/read_options.hpp
willdealtry: I don't think they're really the same kind of thing. The rest of the options are tri-state booleans, unset, set-true and set-false, where as you say the default might (potentially) depend on what operation is requested. 

We can prune the `optional<bool>`s, only about half of them are relevant.

I don't think the default output type will end up being used, it's really just here so it isn't left in an uninitialized state. We could potentially have it as an optional and assert that it's set - will have to see what the Python front-end ends up looking like.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -947,18 +947,16 @@ void check_incompletes_index_ranges_dont_overlap(const std::shared_ptr<PipelineC
...
-                "Cannot append staged segments to existing data as incomplete segment contains index value < existing data (in UTC): {} <= {}",
-                date_and_time(key.start_time()),
-                // Should never reach "" but the standard mandates that all function arguments are evaluated
-                last_existing_index_value ? date_and_time(*last_existing_index_value) : ""
-            );
-            auto [_, inserted] = unique_timestamp_ranges.insert({key.start_time(), key.end_time()});
+                    !last_existing_index_value.has_value() || key.start_time() >= *last_existing_index_value,
+                    "Cannot append staged segments to existing data as incomplete segment contains index value < existing data (in UTC): {} <= {}",
+                    date_and_time(it->slice_and_key().key().start_time()),
+                    date_and_time(*last_existing_index_value));
```
alexowens90, line 953: Vasil changed this recently, I think it's needed?

alexowens90 (changes requested):

alexowens90 (approved):

vasil-pashov (approved):

PR #1810, opened 2024-09-03
Fix regression in bool pickling

vasil-pashov:
#### Reference Issues/PRs

When nullable booleans (which are now disabled) were introduced bool pickling stopped working. This reverts the previous behavior.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -234,8 +234,6 @@ def _to_primitive(arr, arr_name, dynamic_strings, string_max_len=None, coerce_co
             casted_arr = coerce_string_column_to_fixed_length_array(arr, type(sample), string_max_len)
     elif dynamic_strings and sample is None:  # arr is entirely empty
         return arr
-    elif isinstance(sample, bool) or isinstance(sample, np.ndarray):
-        return arr
     else:
         raise ArcticDbNotYetImplemented(
             f"Failed to normalize column '{arr_name}' with dtype '{arr.dtype}'. Found first non-null value of type "
--- a/python/tests/unit/arcticdb/version_store/test_normalization.py
+++ b/python/tests/unit/arcticdb/version_store/test_normalization.py
@@ -620,3 +620,17 @@ def test_norm_failure_error_message(lmdb_version_store_v1):
                [write_exception, batch_write_exception])
     assert all("pickle_on_failure" not in str(e.value) for e in
                [append_exception, batch_append_exception, update_exception])
+
+def test_bools_are_pickled(lmdb_version_store_allows_pickling):
+    lib = lmdb_version_store_allows_pickling
+    sym = "test_bools_are_pickled"
+
+    df = pd.DataFrame({"a": [True, False]})
+    lib.write(sym, df)
+    lib.get_info(sym)['type'] == 'pickled'
+    assert_frame_equal(df, lib.read(sym).data)
+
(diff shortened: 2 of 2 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -234,8 +234,6 @@ def _to_primitive(arr, arr_name, dynamic_strings, string_max_len=None, coerce_co
             casted_arr = coerce_string_column_to_fixed_length_array(arr, type(sample), string_max_len)
     elif dynamic_strings and sample is None:  # arr is entirely empty
         return arr
-    elif isinstance(sample, bool) or isinstance(sample, np.ndarray):
```
poodlewars, line 237: Won't this lead to problems when we allow writing nullable bools or np arrays in dataframes? Is there a feature flag for writing those types?

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -620,3 +620,17 @@ def test_norm_failure_error_message(lmdb_version_store_v1):
                [write_exception, batch_write_exception])
     assert all("pickle_on_failure" not in str(e.value) for e in
                [append_exception, batch_append_exception, update_exception])
+
+def test_bools_are_pickled(lmdb_version_store_allows_pickling):
```
poodlewars, line 624: and numpy arrays?

Is there a test that Python bools and numpy arrays blow up like we expect when pickling is disabled?

python/arcticdb/version_store/_normalization.py
vasil-pashov: We currently don't have a feature flag for the nullable booleans. We have this line here and [this](https://github.com/man-group/ArcticDB/blob/2cad4622122e730032271a0e936a8c590f611cd4/cpp/arcticdb/python/python_to_tensor_frame.cpp#L52). While it's possible to add a flag this normalization function is a bit deep in the call chain. IMO it would be a fair amount of work adding the feature flag and (and later on removing it) compared to adding this PR and reverting it.

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -620,3 +620,17 @@ def test_norm_failure_error_message(lmdb_version_store_v1):
...
+    lib = lmdb_version_store_allows_pickling
+    sym = "test_bools_are_pickled"
+
+    df = pd.DataFrame({"a": [True, False]})
+    lib.write(sym, df)
+    lib.get_info(sym)['type'] == 'pickled'
+    assert_frame_equal(df, lib.read(sym).data)
+
+    df = pd.DataFrame({"a": [True, False, np.nan]})
```
alexowens90, line 633: What's the behaviour if the `NaN` is the first element in the column? Is it the same as `arcticc`?

vasil-pashov: If np.nan is first it gets converted to float True -> 1.0 False -> 0.0. It's the same in arcticc and ArcitcDB

python/tests/unit/arcticdb/version_store/test_normalization.py
vasil-pashov: Arrays are pickled. I added tests, just in case.

poodlewars (approved):

alexowens90 (approved):

PR #1831, opened 2024-09-12
Read as arrow backend

willdealtry:
Monday ref: 7855121013

Backend work for reading arrow. This PR should not change existing behavior in any way. Fully exposing reading as arrow to users will happen in a separate frontend PR.
Currently reading as arrow is only allowed on V1 API behind `_output_format` kwarg (only to be used for tests).

Arrow layout description:
When reading a symbol we produce a `pyarrow.Table` which consists of several `RecordBatch`es. The record batches correspond to our row sliced segments. We might want to add different layout options in the future.
String columns are returned very similar to our internal representation. They are returned as a `DictArray` from `int32` index in a `BigStringArray`. The difference is that arrow requires the string buffers to be part of the column unlike our `StringPool` which is shared between columns. The `ArrowStringHandler` handles this conversion.
Note that we use `int32` instead of `int64` dict keys. That should be fine because of the row group separation based on our data keys. We use `int64` string offsets to allow usage for longer strings.

Changes included in this PR:
- All sparrow code required to zero-copy construct arrow arrays from `SegmentInMemory`s. (contained in `cpp/arcticdb/arrow`)
- Adds a new Chunked buffer `AllocationType::DETACHABLE`. This allows detaching the memory blocks and moving the memory to sparrow. (mostly introduced in previous PR but refined in this one)
- Adds extra buffer hashtable to `Column`s. This is used to attach additional `ExtraBufferType::OFFSET` and `ExtraBufferType::STRING` buffers to specific column offsets. We also use `ExtraBufferType::BITMAP` for all sparse columns. These extra buffers will be used to zero-copy construct a dict string arrow array.
- Modifies the interface of type handlers to allow the type handler to insert the extra buffers when processing arrow strings
- Adds a `UTF_DYNAMIC32` type. It will only be used when outputting arrow (we should never save a column with such type)
- Applies `ColumnTruncation` logic to `OutputFormat::Arrow`. This way arrow output doesn't require post-processing. (contained in `read_frame.cpp` and chunked buffer logic). The logic is partially broken from previous PRs. I will write the fixes and more thorough testing in a separate PR.
- `ReadResult` change to include `output_format` and have a `variant<PandasOutputFrame, ArrowOutputFrame>`.

Commits:
The PR consists of two commits. The first one is squashed from William's initial work on the branch. It consists of the majority of the changes. 
The second one is including mostly bugfixes and refactors to make review easier.
Note to reviewers: It would be best to review the commits together unless you're already familiar with the first commit.

Known issues to be fixed in later PRs:
- No tests for arrow sparse data (+ several known issues with current sparse data marked with TODOs in this PR)
- Fixed string handling with `OutputFormat::Arrow`
- Truncation is very broken. Will be rewritten in a follow up PR


wuxianliang:
Dear friend, Great Job! I have saved all stocks data in ArcticDB, I would like to read the data from SSD directly into arrow table. Then we can query with DuckDB, LanceDB even KuzuDB in memory. As I understand finally maybe use ArcticDB zero-copy like this?
``` python
import arcticdb
import duckdb
import lancedb
import kuzudb

......

arrow_table = lib.read("Symbol", output_format=OutputFormat.ARROW).data

duckdb.sql("SELECT * FROM arrow_table")
lancedb.create_table("Symbol", arrow_table, schema=schema)
kuzudb.execute("COPY Symbol FROM arrow_table")
```

willdealtry:
> Dear friend, Great Job! I have saved all stocks data in ArcticDB, I would like to read the data from SSD directly into arrow table. Then we can query with DuckDB, LanceDB even KuzuDB in memory. As I understand finally maybe use ArcticDB zero-copy like this?
> 
> ```python
> import arcticdb
> import duckdb
> import lancedb
> import kuzudb
> 
> ......
> 
> arrow_table = lib.read("Symbol", output_format=OutputFormat.ARROW).data
> 
> duckdb.sql("SELECT * FROM arrow_table")
> lancedb.create_table("Symbol", arrow_table, schema=schema)
> kuzudb.execute("COPY Symbol FROM arrow_table")
> ```

Yes that's exactly right. I'm very pleased to hear that you are excited about this piece of work!

wuxianliang:
Does read_batch method support read_as_arrow too? Sometimes I wish to analysis all symbols in a daterange.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -213,7 +214,8 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
 
     py::enum_<OutputFormat>(version, "OutputFormat")
         .value("PANDAS", OutputFormat::PANDAS)
-        .value("ARROW", OutputFormat::ARROW);
+        .value("ARROW", OutputFormat::ARROW)
+        .value("PARQUET", OutputFormat::PARQUET);
```
jamesmunro, line 218: Drop this parquet option for now?

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -213,7 +214,8 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
 
     py::enum_<OutputFormat>(version, "OutputFormat")
         .value("PANDAS", OutputFormat::PANDAS)
```
jamesmunro, line 216: `PANDAS` is probably fine as a name, but worth considering that we'll want to add an arrow-backed pandas dataframe option later.

cpp/arcticdb/version/python_bindings.cpp
IvoDD: Done

python/arcticdb/version_store/_store.py
```diff
@@ -1712,6 +1712,7 @@ def _get_read_options(self, **kwargs):
         read_options = _PythonVersionStoreReadOptions()
         read_options.set_force_strings_to_object(_assume_false("force_string_to_object", kwargs))
         read_options.set_optimise_string_memory(_assume_false("optimise_string_memory", kwargs))
+        read_options.set_output_format(kwargs.get("output_format") or OutputFormat.PANDAS)
```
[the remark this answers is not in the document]
IvoDD: Will do these changes in frontend PR as discussed offline. Will have a new `RuntimeLibraryOptions` so we don't have to specify the output format on every request.

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -357,32 +371,31 @@ ColumnTruncation get_truncate_range(
...
-            }
-        }
+        util::variant_match(read_query.row_filter,
+            [&truncate_rows, &frame, &context, &index_field, index_field_offset, encoding_version] (const IndexRange& index_range) {
+                const auto& time_range = static_cast<const TimestampRange&>(index_range);
+                const auto& slice_time_range =  context.slice_and_key().key().time_range();
+                if(contains(slice_time_range, time_range.first) || contains(slice_time_range, time_range.second)) {
+                    if(context.fetch_index()) {
+                        const auto& index_column = frame.column(0);
+                        truncate_rows = get_truncate_range_from_index(index_column, time_range.first, time_range.second);
```
IvoDD, line 381: This change is mostly whitespace. Only removes the `current_row_range` and the arguments it passes to `get_truncate_range_from_index`.

This is to avoid a segfault in existing tests. There are many more issues with this logic (off by one errors etc.). This will be reworked in a future PR so no need to spend too much time on review on this change.

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -229,7 +228,7 @@ void decode_index_field(
             auto offset = sz * (slice_and_key.slice_.row_range.first - frame.offset());
             auto tot_size = sz * slice_and_key.slice_.row_range.diff();
 
-            SliceDataSink sink(buffer.data() + offset, tot_size);
+            SliceDataSink sink(buffer.bytes_at(offset, tot_size), tot_size);
```
IvoDD, line 231: This is required to handle the case where `offset` is not in the first block of the `buffer`

cpp/arcticdb/pipeline/pipeline_utils.hpp
```diff
@@ -27,9 +27,11 @@ inline void apply_type_handlers(SegmentInMemory seg, std::any& handler_data, Out
     for(auto i = 0U; i < seg.num_columns(); ++i) {
         auto& column = seg.column(i);
         if(auto handler = get_type_handler(output_format, column.type()); handler) {
+            // TODO: To support arrow output format we'll need to change the allocation logic for the dest_column.
+            // We'll need to consider what arrow layout we want to output the data in.
+            util::check(output_format == OutputFormat::PANDAS, "Only Pandas output format is supported for read_result_from_single_frame");
```
IvoDD, line 32: `read_result_from_single_frame` is only used for lib tool utilities and operations like `read_index`.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1507,43 +1502,43 @@ std::vector<std::variant<VersionedItem, DataError>> LocalVersionedEngine::batch_
...
+                            true);
+                    }
+                    return std::move(index_key_fut).thenValueInline([update_info = std::move(update_info)](auto&& index_key) mutable {
+                        return IndexKeyAndUpdateInfo{std::move(index_key), std::move(update_info)};
+                    });
+                })
+                .thenValue([this, prune_previous_versions](auto&& index_key_and_update_info) {
+                    auto&& [index_key, update_info] = index_key_and_update_info;
+                    return write_index_key_to_version_map_async(version_map(), std::move(index_key), std::move(update_info), prune_previous_versions);
+                })
```
IvoDD, line 1547: Only whitespace

cpp/arcticdb/pipeline/frame_slice.hpp
```diff
@@ -60,17 +60,9 @@ struct RowRange : AxisRange {
 };
 
 inline bool contains(const RowRange& range, size_t row) {
-    return row >= range.first && row < range.second;
+    return row >= range.first && row <= range.second;
```
IvoDD, line 63: `contains` is only used in `truncation` logic.
This makes more sense to revert to exclusive end. Will do in the follow up PR to fix truncation.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -538,7 +534,18 @@ set(arcticdb_srcs
...
+        arrow/arrow_utils.hpp
+        arrow/arrow_output_frame.hpp
+        arrow/arrow_utils.cpp
+        arrow/arrow_output_frame.cpp
+        arrow/arrow_handlers.hpp
+        arrow/array_from_block.hpp
+        arrow/arrow_handlers.cpp
+        pipeline/read_query.hpp
+        pipeline/read_query.cpp
+        pipeline/read_pipeline.cpp
```
alexowens90: nit: alphabetical order

cpp/arcticdb/codec/encode_v1.cpp
```diff
@@ -138,6 +138,8 @@ namespace arcticdb {
             ARCTICDB_TRACE(log::codec(), "Encoding fields");
             for (std::size_t column_index = 0; column_index < in_mem_seg.num_columns(); ++column_index) {
                 const auto& column = in_mem_seg.column(column_index);
+                auto data_type = column.data().type().data_type();
+                util::check(!is_arrow_output_only_type(data_type), "Attempts to encode an output only type {}", data_type);
```
alexowens90:
```suggestion
                util::check(!is_arrow_output_only_type(column.type()), "Attempts to encode an output only type {}", data_type);
```
and same in `encode_v2.cpp`

cpp/arcticdb/column_store/column.cpp
```diff
@@ -609,7 +609,7 @@ void Column::truncate_last_block(size_t row) {
 }
 
 void Column::truncate_single_block(size_t start_row, size_t end_row) {
-    if(!is_sparse()) {
+    if(!is_sparse()) { // TODO: We need to also do truncation on sparse data
```
alexowens90, line 612: Is there a consolidated list of remaining Arrow work somewhere? Would be good to track all things like this in one place so we don't forget anything

cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+    auto &strings = column.get_extra_buffer(offset, ExtraBufferType::STRING);
+    const auto strings_buffer_size = strings.block(0)->bytes();
+    util::check(
+        strings_buffer_size < std::numeric_limits<int32_t>::max(),
+        "Strings buffer too large for arrow string array. String buffer size: {}. Max buffer size: {}. Column: {}",
+        strings_buffer_size, std::numeric_limits<int32_t>::max(), name);
+    sparrow::u8_buffer<char> strings_buffer(reinterpret_cast<char *>(strings.block(0)->release()), strings_buffer_size);
+
+    // We use `int32_t` dictionary keys because pyarrow doesn't work with unsigned dictionary keys:
```
alexowens90, line 97: Discussing out of band using int64 for more resilience to massive strings, and adding an optimisation later to use int32 where possible (i.e. if a column would have <2^32 bytes of strings in every record batch)

cpp/arcticdb/entity/types.hpp
```diff
@@ -393,6 +398,7 @@ DATA_TYPE_TAG(ASCII_FIXED64, std::uint64_t)
 DATA_TYPE_TAG(ASCII_DYNAMIC64, std::uint64_t)
 DATA_TYPE_TAG(UTF_FIXED64, std::uint64_t)
 DATA_TYPE_TAG(UTF_DYNAMIC64, std::uint64_t)
+DATA_TYPE_TAG(UTF_DYNAMIC32, std::uint32_t)
```
alexowens90: Should this be signed?

cpp/arcticdb/pipeline/frame_slice.hpp
```diff
@@ -60,17 +60,9 @@ struct RowRange : AxisRange {
...
 }
 
-/*
- * This class is mostly (exclusively?) used in SliceAndKey objects, where the duplication of the StreamDescriptor
- * with the SegmentInMemory is confusing and error-prone. Where possible do not add more uses of this class and
- * SliceAndKey, prefer RangesAndKey or SegmentAndSlice depending on requirements.
- * FrameSlice stores the vertical (RowRange) and horizontal (ColRange) offsets for a table-subsection, as well as
- * (optionally) a descriptor for the source stream/index data. For dynamic_schema's bucketizing, it also stores
- * the hash bucket it represents and the number of total buckets that are present for the current calculation
- */
```
alexowens90: Seems still needed

cpp/arcticdb/pipeline/pandas_output_frame.cpp
```diff
@@ -40,29 +40,37 @@ PythonOutputFrame::~PythonOutputFrame() {
...
+                if constexpr (is_object_type(td)) {
                     if constexpr(is_array_type(td)) {
                         auto it = column_data.buffer().iterator(sizeof(PyObject*));
                         while(!it.finished()) {
-                            util::check(reinterpret_cast<PyObject*>(it.value()) != nullptr, "Can't delete null item");
-                            Py_DECREF(reinterpret_cast<PyObject*>(it.value()));
+                            if (reinterpret_cast<PyObject*>(it.value()) != nullptr) {
+                                Py_DECREF(reinterpret_cast<PyObject*>(it.value()));
+                            } else {
+                                log::version().error("Unexpected nullptr to DecRef in PandasOutputFrame destructor");
```
alexowens90, line 49: Was the check throwing in some cases before? Do we ever see this error?

cpp/arcticdb/util/bitset.hpp
```diff
@@ -32,4 +32,24 @@ using BitIndex = bm::bvector<>::rs_index_type;
 constexpr bm::bvector<>::size_type bv_size(uint64_t val) {
     return static_cast<bm::bvector<>::size_type>(val);
 }
+
+constexpr size_t bitset_unpacked_size(size_t size) {
+    return (size + 63) / 64 * sizeof(uint64_t);
```
alexowens90, line 37: Shouldn't this return 1 if `size` is 1-8, 2 if `size` is 9-16, etc? If not worth a comment as to why

alexowens90: Oh is it padded out to the nearest 8 bytes? Is that an Arrow thing, or an us thing?

cpp/arcticdb/util/bitset.hpp
```diff
@@ -32,4 +32,24 @@ using BitIndex = bm::bvector<>::rs_index_type;
...
+inline void bitset_to_packed_bits(const bm::bvector<>& bv, uint64_t* dest_ptr) {
+    std::memset(dest_ptr, 0, bitset_unpacked_size(bv.size()));
+
+    auto en = bv.first();
+    auto last = bv.end();
+    for (; en != last; ++en) {
+        size_t bit_pos = *en;
+
+        size_t word_idx = bit_pos / 64;
+        size_t bit_idx = bit_pos % 64;
```
alexowens90, line 49: These `size_t`s should all be `uint64_t`, and can use `std::div` here to get the quotient and remainder in one operation

cpp/arcticdb/util/bitset.hpp
```diff
@@ -32,4 +32,24 @@ using BitIndex = bm::bvector<>::rs_index_type;
...
+
+        dest_ptr[word_idx] |= (1ULL << bit_idx);
+    }
+}
```
alexowens90, line 53: Not for this PR, but on the TODO list of outstanding work, we should add a microbenchmark for this and make sure it's optimal. For example, `bit_pos`, and therefore `word_idx` are monotonically increasing, but the compiler can't know this. So it might be possible to improve the performance by building up each `uint64_t` and then just assigning once to `dest_ptr[word_idx]`.
There are also more efficient set-bit visitors possible, at the cost of a bit more code:
https://github.com/tlk00/BitMagic/tree/master/samples/bvsample25

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -289,21 +289,21 @@ struct VersionMapEntry {
...
+        strm << "Keys: " << "\n" << "\n";
         for(const auto& key: keys_)
-            strm << fmt::format("    {}", key) << std::endl;
+            strm << fmt::format("    {}", key) << "\n";
 
-        strm << "Tombstones: " << std::endl << std::endl;
+        strm << "Tombstones: " << "\n" << "\n";
         for(const auto& tombstone: tombstones_)
-            strm << fmt::format("    {} - {}", tombstone.first, tombstone.second) << std::endl;
+            strm << fmt::format("    {} - {}", tombstone.first, tombstone.second) << "\n";
```
alexowens90: nit: Could fold a lot of these newlines into adjacent strings

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -0,0 +1,211 @@
...
+@pytest.mark.parametrize("dynamic_strings", [
+    True,
+    pytest.param(False, marks=pytest.mark.xfail(reason="Arrow fixed strings are not normalized correctly"))
+])
+def test_strings_basic(lmdb_version_store_v1, dynamic_strings):
+    lib = lmdb_version_store_v1
+    df = pd.DataFrame({"x": ["mene", "mene", "tekel", "upharsin"]})
+    lib.write("arrow", df, dynamic_strings=dynamic_strings)
+    vit = lib.read("arrow", _output_format=OutputFormat.ARROW)
+    result = convert_pandas_categorical_to_str(vit.data.to_pandas())
```
alexowens90, line 99: I assume this means pyarrow to_pandas converts string columns to categoricals?

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -0,0 +1,211 @@
...
+    lib = lmdb_version_store_v1
+    # sample dataframe contains all dtypes + unicode strings
+    df = get_sample_dataframe()
+    lib.write("arrow", df)
+    vit = lib.read("arrow", _output_format=OutputFormat.ARROW)
+    result = convert_pandas_categorical_to_str(vit.data.to_pandas())
+    assert_frame_equal(result, df)
+
+
+@pytest.mark.parametrize("start_offset,end_offset", [(2, 3), (3, 75), (4, 32), (0, 99), (7, 56)])
```
alexowens90, line 133: Could also test with the first, second, and both offsets being `None`

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -0,0 +1,211 @@
...
+def test_date_range(lmdb_version_store_v1, start_offset, end_offset):
```
alexowens90, line 134: Would be good to run this with `lmdb_version_store_tiny_segment` as well (might need your truncate PR to pass)

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -0,0 +1,211 @@
...
+def test_with_querybuilder(lmdb_version_store_v1):
+    lib = lmdb_version_store_v1
+    df = pd.DataFrame({"x": np.arange(10), "y": np.arange(10.0, 20.0)})
+    q = QueryBuilder()
+    q = q[q["x"] < 5]
+    lib.write("arrow", df)
+    vit = lib.read("arrow", _output_format=OutputFormat.ARROW, query_builder=q)
+    expected = df[df["x"] < 5]
+    result = vit.data.to_pandas()
+    assert_frame_equal(result, expected)
```
alexowens90, line 163: Before making publicly available we should make sure all of the processing pipeline tests run with both Pandas and Arrow as the output formats

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def test_dynamic_schema(lmdb_version_store_dynamic_schema):
+    lib = lmdb_version_store_dynamic_schema
+    df1 = pd.DataFrame({"x": np.arange(10), "y": np.arange(10.0, 20.0)})
+
+    lib.write("arrow", df1)
+    df2 = pd.DataFrame({"y": np.arange(20.0, 30.0), "z": np.arange(10.0, 20.0)})
+    lib.append("arrow", df2)
+    vit = lib.read("arrow", _output_format=OutputFormat.ARROW)
```
alexowens90, line 173: Can we also assert the existence and correct population of the validity bitmaps in the returned record batches?

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def test_dynamic_schema_column_change(lmdb_version_store_dynamic_schema):
+    lib = lmdb_version_store_dynamic_schema
+    df1 = pd.DataFrame({"x": np.arange(10)})
+
+    lib.write("arrow", df1)
+    df2 = pd.DataFrame({"x": np.arange(20.0, 30.0)})
+    lib.append("arrow", df2)
+    vit = lib.read("arrow", _output_format=OutputFormat.ARROW)
```
alexowens90: Can we assert that the type of `x` in all of the returned record batches is `float64`?
Presumably this is a requirement of the pyarrow table format? i.e. we couldn't return a list of record batches where `x` is `int64` in the first and `float64` in the second and have pyarrow recognise it as a valid table? Just thinking out loud about potential performance optimisations with dynamic schema

cpp/arcticdb/python/adapt_read_dataframe.hpp
```diff
@@ -10,12 +10,13 @@
...
 #include <arcticdb/entity/read_result.hpp>
 #include <arcticdb/python/python_handler_data.hpp>
+#include <arcticdb/arrow/arrow_utils.hpp>
 
 namespace arcticdb {
 
-inline py::tuple adapt_read_df(ReadResult&& ret, std::pair<std::any&, OutputFormat>* const handler) {
-    if (handler) {
-        apply_global_refcounts(handler->first, handler->second);
+inline py::tuple adapt_read_df(ReadResult&& ret, std::any* const handler_data) {
```
alexowens90, line 17: Looks like `handler_data` is never `nullptr`, so can be passed by reference?

cpp/arcticdb/python/python_utils.hpp
```diff
@@ -239,28 +239,31 @@ class PyTimestampRange {
     timestamp end_;
 };
 
-inline py::list adapt_read_dfs(std::vector<std::variant<ReadResult, DataError>>&& r, std::pair<std::any&, OutputFormat>* const handler) {
+inline py::list adapt_read_dfs(std::vector<std::variant<ReadResult, DataError>>&& r, std::any* const handler) {
```
alexowens90, line 242:
```suggestion
inline py::list adapt_read_dfs(std::vector<std::variant<ReadResult, DataError>>&& r, std::any& handler) {
```

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -236,33 +237,42 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
...
 
     using FrameDataWrapper = arcticdb::pipelines::FrameDataWrapper;
     py::class_<FrameDataWrapper, std::shared_ptr<FrameDataWrapper>>(version, "FrameDataWrapper")
             .def_property_readonly("data", &FrameDataWrapper::data);
 
-    using PythonOutputFrame = arcticdb::pipelines::PythonOutputFrame;
-    py::class_<PythonOutputFrame>(version, "PythonOutputFrame")
+    using PandasOutputFrame = arcticdb::pipelines::PandasOutputFrame;
+    py::class_<PandasOutputFrame>(version, "PandasOutputFrame")
+        //.def(py::init<const SegmentInMemory&, OutputFormat output_format, std::shared_ptr<BufferHolder>>())
```
alexowens90: nit: remove

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -236,33 +237,42 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
...
-        .def_property_readonly("index_columns", &PythonOutputFrame::index_columns, py::return_value_policy::reference)
-        .def_property_readonly("row_count", [](PythonOutputFrame& self) {
+        .def_property_readonly("names", &PandasOutputFrame::names, py::return_value_policy::reference)
+        .def_property_readonly("index_columns", &PandasOutputFrame::index_columns, py::return_value_policy::reference)
+        .def_property_readonly("row_count", [](PandasOutputFrame& self) {
             return self.frame().row_count();
         });
 
+        py::class_<ArrowOutputFrame>(version, "ArrowOutputFrame")
+        .def_property_readonly("names", &ArrowOutputFrame::names)
```
alexowens90: Seems to be unused

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1439,8 +1439,7 @@ folly::Future<SegmentInMemory> prepare_output_frame(
         row.set_string_pool(row.slice_and_key().segment(store).string_pool_ptr());
     }
 
-    const auto allocation_type = read_options.output_format() == OutputFormat::ARROW ? AllocationType::DETACHABLE : AllocationType::PRESIZED;
-    auto frame = allocate_frame(pipeline_context, read_options.output_format(), allocation_type);
+    auto frame = allocate_frame(pipeline_context, read_options.output_format());
```
alexowens90, line 1447: We could optimise in future to move buffers from the processed segments into the output frame rather than memcpy them

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -262,6 +261,16 @@ void handle_truncation(
     handle_truncation(dest_column, mapping.truncate_);
 }
 
+void create_dense_bitmap(size_t offset, const util::BitSet& sparse_map, Column& dest_column, AllocationType allocation_type) {
+    auto& sparse_buffer = dest_column.create_extra_buffer(
+        offset,
+        ExtraBufferType::BITMAP,
+        bitset_unpacked_size(sparse_map.count() * sizeof(uint64_t)),
```
alexowens90, line 268:
```suggestion
        bitset_unpacked_size(sparse_map.size() * sizeof(uint64_t)),
```

cpp/arcticdb/column_store/column.hpp
```diff
@@ -70,6 +70,66 @@ struct JiveTable {
...
+    size_t offset_;
+    ExtraBufferType type_;
+};
+
+inline bool operator==(const ExtraBufferIndex& lhs, const ExtraBufferIndex& rhs) {
+    return (lhs.offset_ == rhs.offset_) && (lhs.type_ == rhs.type_);
+}
+
+struct ExtraBufferIndexHash {
+    static void hash_combine(std::size_t& seed, std::size_t value) {
```
alexowens90: nit: folly (and presumably boost) has hashing combinations to avoid hand-rolling our own

cpp/arcticdb/column_store/column.hpp
```diff
@@ -70,6 +70,66 @@ struct JiveTable {
...
+enum class ExtraBufferType : uint8_t {
+    OFFSET,
+    STRING,
+    ARRAY,
+    BITMAP
+};
+
+
+struct ExtraBufferIndex {
+    size_t offset_;
```
alexowens90: It took me a while to work out what this means, maybe a rename to `offset_bytes_` and a comment.
Using this struct as a map key also seems quite weird at the moment, maybe it will become clear by the time I finish reading the PR though

cpp/arcticdb/column_store/column.hpp
```diff
@@ -70,6 +70,66 @@ struct JiveTable {
     std::vector<uint32_t> sorted_pos_;
 };
 
+enum class ExtraBufferType : uint8_t {
+    OFFSET,
```
alexowens90, line 74: This word is so overloaded, I'm guessing this is the dict keys in a string column?
If we use int64 as the dict keys, there is a possible future optimisation where we don't have an extra buffer, and just reuse the original column buffer, overwriting our string pool offsets with the dict keys to avoid an extra allocation. Would only work for single-block columns, but that is most of our data

cpp/arcticdb/entity/read_result.hpp
```diff
@@ -79,7 +86,13 @@ inline ReadResult create_python_read_result(
         }
     }
 
-    auto python_frame = pipelines::PythonOutputFrame{result.frame_, output_format};
+    auto python_frame = [&]() -> OutputFrame {
+        if (output_format == OutputFormat::ARROW) {
+            return ArrowOutputFrame{segment_to_arrow_data(result.frame_), names_from_segment(result.frame_)};
+        } else {
+            return pipelines::PandasOutputFrame{result.frame_, output_format};
```
alexowens90: Is the output format really required in the `PandasOutputFrame` ctor? Seems like an error to pass anything but `OutputFormat::PANDAS` in?

cpp/arcticdb/entity/read_result.hpp
```diff
@@ -12,31 +12,38 @@
...
             norm_meta(norm_meta),
             user_meta(user_meta),
             multi_key_meta(multi_key_meta),
             multi_keys(std::move(multi_keys)) {
 
     }
     std::variant<VersionedItem, std::vector<VersionedItem>> item;
-    pipelines::PythonOutputFrame frame_data;
+    OutputFrame frame_data;
+    OutputFormat output_format;
```
alexowens90, line 46: We could get rid of this and use `isinstance` on the `frame_data` member in the Python layer to avoid duplicating information

cpp/arcticdb/util/type_handler.hpp
```diff
@@ -137,7 +138,7 @@ class TypeHandlerRegistry {
         return handlers_[static_cast<int>(output_format)];
     }
 
-    std::array<TypeHandlerMap, static_cast<size_t>(OutputFormat::COUNT)> handlers_;
+    std::array<TypeHandlerMap, 4> handlers_;
```
alexowens90: Previous impl seems better? (and on line 123 as well)

cpp/arcticdb/arrow/arrow_output_frame.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+ArrowOutputFrame::ArrowOutputFrame(
+    std::shared_ptr<std::vector<sparrow::record_batch>>&& data,
+    std::vector<std::string>&& names) :
+    data_(std::move(data)),
+    names_(std::move(names)) {
+}
+
+size_t ArrowOutputFrame::num_blocks() const {
+    if(data_->empty())
```
alexowens90:
```suggestion
    if(!data_ || data_->empty())
```

cpp/arcticdb/arrow/arrow_output_frame.cpp
```diff
@@ -0,0 +1,48 @@
...
+        return 0;
+
+    return data_->size();
+}
+
+std::vector<RecordBatchData> ArrowOutputFrame::extract_record_batches() {
```
alexowens90, line 27: I guess return an empty vector if `data_` is null

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -0,0 +1,73 @@
...
+    const auto num_columns = segment.num_columns();
+    const auto column_blocks = segment.column(0).num_blocks();
+
+    auto output = std::make_shared<std::vector<sparrow::record_batch>>();
+    output->reserve(total_blocks);
+
+    for(auto i = 0UL; i < column_blocks; ++i)
+        output->emplace_back(sparrow::record_batch{});
+
+    util::check(total_blocks == column_blocks * num_columns, "Expected regular block size");
```
alexowens90: nit: This can be checked as soon as the 3 variables are assigned

cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+#include <sparrow/sparrow.hpp>
+#include <sparrow/layout/primitive_layout/primitive_data_access.hpp>
+
+namespace arcticdb {
+
+inline std::optional<sparrow::validity_bitmap> create_validity_bitmap(size_t offset, const Column& column) {
+    if(column.has_extra_buffer(offset, ExtraBufferType::BITMAP)) {
+        auto &bitmap_buffer = column.get_extra_buffer(offset, ExtraBufferType::BITMAP);
+        const auto bitmap_size = bitmap_buffer.block(0)->bytes() / sizeof(uint32_t);
```
alexowens90, line 19: Why the division? Bitmap buffers seemed to be aligned to 8-bytes before?

cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -0,0 +1,139 @@
...
+        std::string_view name,
+        std::optional<sparrow::validity_bitmap> maybe_bitmap) {
+    const auto offset = block.offset();
+    // We use 32-bit offsets and keys because we use a layout where each row-segment has its own arrow array.
+    // By default, the row-segments are 100k rows.
+
+    // String offsets must be `int32_t` even though they are unsigned indices due to arrow spec.
+    // https://arrow.apache.org/docs/format/Columnar.html#variable-size-binary-layout
+    auto &string_offsets = column.get_extra_buffer(offset, ExtraBufferType::OFFSET);
+    const auto offset_buffer_size = string_offsets.block(0)->bytes() / sizeof(int32_t);
```
alexowens90:
```suggestion
    const auto offset_buffer_value_count = string_offsets.block(0)->bytes() / sizeof(int32_t);
```

cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -0,0 +1,139 @@
...
+    sparrow::u8_buffer<char> strings_buffer(reinterpret_cast<char *>(strings.block(0)->release()), strings_buffer_size);
+
+    // We use `int32_t` dictionary keys because pyarrow doesn't work with unsigned dictionary keys:
+    // https://github.com/pola-rs/polars/issues/10977
+    const auto block_size = block.row_count();
+    util::check(
+        block_size < std::numeric_limits<int32_t>::max(),
+        "Key buffer too large for arrow dictionary array. Key buffer size: {}. Max buffer size: {}. Column: {}",
+        block_size, std::numeric_limits<int32_t>::max(), name);
+    sparrow::u8_buffer<int32_t> dict_keys_buffer{reinterpret_cast<int32_t *>(block.release()), block_size};
```
alexowens90, line 100: Won't this `block` be int64 offsets into our stringpool? I'm clearly missing something about this implementation

alexowens90: Oh I think I get it, the conversions is taken care of by the `ArrowStringHandler`?

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -0,0 +1,115 @@
...
+void ArrowStringHandler::convert_type(
+    const Column& source_column,
+    Column& dest_column,
+    const ColumnMapping& mapping,
+    const DecodePathData&,
+    std::any&,
+    const std::shared_ptr<StringPool>& string_pool) const {
+    size_t bytes = 0;
+    using ArcticStringColumnTag = ScalarTagType<DataTypeTag<DataType::UTF_DYNAMIC64>>;
+    auto dest_ptr = reinterpret_cast<uint32_t*>(dest_column.bytes_at(mapping.offset_bytes_, source_column.row_count() * sizeof(uint32_t)));
```
alexowens90: This mixing of `uint32_t` and `int32_t` seems like a recipe for disaster. Can we stick a single `using` statement for the raw type we are using for dict keys in Arrow string columns somewhere and use it everywhere? That way we can also trivially flip to `int64_t` as well

cpp/arcticdb/arrow/arrow_handlers.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+inline void register_arrow_handler_data_factory() {
+    TypeHandlerRegistry::instance()->set_handler_data(OutputFormat::ARROW, std::make_unique<ArrowHandlerDataFactory>());
+}
+
+
+inline void register_arrow_string_types() {
+    using namespace arcticdb;
+    constexpr std::array<DataType, 4> dynamic_string_data_types = {
+        DataType::ASCII_DYNAMIC64, DataType::UTF_DYNAMIC64, DataType::ASCII_FIXED64, DataType::UTF_FIXED64};
```
alexowens90, line 61: This class has registered to handle all of our support internal string types, but in reality `ArrowStringHandler::convert_type` will only actually handle `UTF_DYNAMIC64`

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -0,0 +1,115 @@
...
+    auto begin_ptr = strv_ptr;
+    auto keys_ptr = reinterpret_cast<uint32_t*>(offsets_buffer.data());
+    for(auto unique_string : unique_offsets) {
+        const auto strv = string_pool->get_const_view(unique_string.first);
+        memcpy(strv_ptr, strv.data(), strv.size());
+        *keys_ptr = unique_string.second.bytes_pos_;
+        util::check(strv_ptr - begin_ptr == unique_string.second.bytes_pos_, "Mismatch in buffer offset");
+        strv_ptr += strv.size();
+        ++keys_ptr;
+    }
```
alexowens90, line 95: I don't think this is correct except in the (unlikely) case that iteration through `unique_offsets` matches the insertion order into it

alexowens90 (changes requested):

alexowens90:
Prior to releasing, we should ensure:

- we are definitely zero-copy on the direct-read path, and only incur one additional copy on the processing pipeline path
- we are correctly transferring ownership of the underlying buffers, so that there is no unexpected memory overhead or leak associated with reading an Arrow structure from ArcticDB vs creating it in pyarrow or similar
- we have profiled (and included microbenchmarks/ASV benchmarks for critical regions) and optimised as far as possible

cpp/arcticdb/column_store/column.cpp
IvoDD: Yes I've kept those as tasks on the arrow epic. This specific TODO corresponds to 9288960436 ticket.

cpp/arcticdb/arrow/array_from_block.hpp
IvoDD: Agreed, will do

cpp/arcticdb/entity/types.hpp
IvoDD: It is converted from unsigned to signed in the ArrowStringHandler with checks if it will overflow.

As per other discussion I think we should just stick to using the largest possible types and add optimizations later. I.e. this would mean we can reuse the `UTF_DYNAMIC64`.

cpp/arcticdb/pipeline/pandas_output_frame.cpp
IvoDD: We have not seen this error I think. But it is undefined behavior to raise in destructor. That's why replaced with logs.

cpp/arcticdb/util/bitset.hpp
IvoDD: Added a note to the ticket for benchmarking. Also noted there that we should include micro benchmarks for:
- ArrowStringHandler
- ColumnTruncation

python/tests/unit/arcticdb/version_store/test_arrow.py
IvoDD: yes, we'll need to decide whether we will want to correctly round trip with a plain `pyarrow.to_pandas`. 
I don't think we should as it will be very hard but happy to discuss.

python/tests/unit/arcticdb/version_store/test_arrow.py
IvoDD: Will be done in follow up truncation PR

python/tests/unit/arcticdb/version_store/test_arrow.py
IvoDD: yes, doing this in follow up truncation pr. Would not pass in this one.

python/tests/unit/arcticdb/version_store/test_arrow.py
IvoDD: Adding a note to the frontend ticket to do that

python/tests/unit/arcticdb/version_store/test_arrow.py
IvoDD: I suspect validity bitmaps are a bit broken so have refrained from thoroughly testing them. 

We should add such tests when doing the sparse data ticket 9288960436

cpp/arcticdb/python/adapt_read_dataframe.hpp
IvoDD: I think it could be (and there was a reason for doing it this way when discussing this with Vasil with respect to null handling). I'll look and come back

cpp/arcticdb/version/python_bindings.cpp
IvoDD: ah yes the `names` bit is unused

cpp/arcticdb/version/version_core.cpp
IvoDD: I'll add a ticket

cpp/arcticdb/column_store/column.hpp
IvoDD: Not exactly the dict keys but the offsets in the string pool. It is slightly different e.g if we have a column:
```
["ab", "c", "ab"]
```

It will get stored for arrow as:
```
column values:       [0, 1, 0]
OFFSET extra buffer: [0, 2, 3]
STRING extra buffer: ["a", "b", "c"]
```

I'm not sure I understand your proposed optimization. Let's chat offline

cpp/arcticdb/arrow/array_from_block.hpp
IvoDD: yeah the `ArrowStringHandler` attaches the required `ExtraBuffer`s

cpp/arcticdb/arrow/arrow_handlers.cpp
IvoDD: Ah that's correct. Will rework

cpp/arcticdb/arrow/array_from_block.hpp
alexowens90: I think I had misunderstood the buffer structure for Arrow dict columns, let's discuss

cpp/arcticdb/entity/types.hpp
alexowens90: Discussed in person, use int32_t

cpp/arcticdb/util/bitset.hpp
alexowens90: Park this until we fully implement sparse data

cpp/arcticdb/pipeline/read_frame.cpp
alexowens90: Park until we're fixing sparse in general

cpp/arcticdb/entity/read_result.hpp
alexowens90: `NATIVE` format used in some C++ tests to avoid a Python dependency, leave in
Add a comment to `PandasOutputFrame` ctor to avoid other people asking this question in future

cpp/arcticdb/arrow/array_from_block.hpp
alexowens90: Park until sparse implemented

cpp/arcticdb/arrow/arrow_handlers.hpp
alexowens90: Can come later

cpp/arcticdb/arrow/arrow_handlers.cpp
IvoDD: Used an `std::vector` to keep track of the correct order. Renamed quite a few variables to make logic more clear.

cpp/arcticdb/python/adapt_read_dataframe.hpp
IvoDD: There are some python calls like `restore_version` [here](https://github.com/man-group/ArcticDB/blob/becbe4000d61b72b10ffc31c51c9f1a35a213e4a/cpp/arcticdb/version/python_bindings.cpp#L779). There we construct a python read result without requiring any python handler allocations, so will leave as is.

cpp/arcticdb/python/python_utils.hpp
IvoDD: Similar to above for `batch_restore_version` which I slightly changed to use `adapt_read_dfs` to be consistent with other `batch` methods.

cpp/arcticdb/entity/read_result.hpp
IvoDD: Turned out the C++ tests did not rely on the NATIVE output format so I removed the `output_format` member of `PandasOutputFrame`

cpp/arcticdb/entity/read_result.hpp
IvoDD: I'll keep it here to differetiate between NATIVE and PANDAS because on the read result layer we see which handlers we need to use.

Still removing from the python read_result as you're right that it's not needed there.

alexowens90 (approved):

PR #1840, opened 2024-09-19
Don't override the https value for backwards compatibility

### LINKED ISSUE #1846: Libraries written by 4.5.0 are not backwards compatible for s3 https stоrages
### Describe the bug

Client using 4.5.0 are [no longer writing](https://github.com/man-group/ArcticDB/blob/5dd6a642c3ab70be0ead45665c8cfcfa8cffafa3/cpp/arcticdb/storage/storage_override.hpp#L118) the `https` flag in the library config, which is needed by the older clients to correctly for their http requests.

We need to change it, so:
- the `https`' value is written when creating the library - so the libraries are backwards compatible and can be read by older clients
- the `https`' flag is overridden when the config is read - so the newer clients can read libraries created by 4.5.0

### Steps/Code to Reproduce

```python

ac = Arctic("s3s://...")
lib = ac.create_library("test")
lib.write("sym", pd.DataFrame())

config = ac._library_manager.get_library_config("test")
s3_storage = get_s3_storage_config(config)
assert s3_storage.https
```

### Expected Results

The assert check should succeed as the `https` flag's value should reflect that the library was created with `s3s/https` in mind.

Python: 3.10.14 | packaged by conda-forge | (main, Mar 20 2024, 12:45:18) [GCC 12.3.0]
OS: Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35 ArcticDB: 4.5.0

### Backend storage used

S3

### Additional Context

_No response_

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1846 on master

Overriding the value also wipes it from the storage config.
Older clients have no way of overriding the value and they rely on the value in storage.
So don't override it for now, as it breaks backwards compatibility.

#### Any other comments?
I've tested the change manually and libraries created after the change can be read correctly by older versions.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/library_manager.cpp
+++ b/cpp/arcticdb/storage/library_manager.cpp
@@ -30,21 +30,23 @@ const std::string BAD_CONFIG_IN_ATTEMPTED_WRITE = "Attempting to write forbidden
 template<typename T>
 struct StorageVisitor {
     arcticdb::proto::storage::LibraryConfig& lib_cfg_proto;
+    bool override_https;
 
     void operator()(const T& storage_override) {
         for(auto& storage: *lib_cfg_proto.mutable_storage_by_id()){
-            storage_override.modify_storage_config(storage.second);
+            storage_override.modify_storage_config(storage.second, override_https);
         }
     }
 };
 
 void apply_storage_override(const StorageOverride& storage_override,
-                                   arcticdb::proto::storage::LibraryConfig& lib_cfg_proto) {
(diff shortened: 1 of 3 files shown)


poodlewars (approved):
Approving as this is the same logical change as https://github.com/man-group/ArcticDB/pull/1842/files which has already been approved

PR #1844, opened 2024-09-20
Support the ignore missing key read option in recurse_index_keys

poodlewars:
This is required for delayed deletes to work correctly - https://github.com/man-group/arcticdb-enterprise/pull/167 includes an end to end failing test to motivate this.

The issue is that if we have a snapshot being modified over time, replication might sync over its current state from the source storage before the index keys inside it. So delayed deletes needs to be tolerant to the fact that some of the keys it is asked to recurse may not actually exist.

I also noticed that `recurse_index_keys` requires all the index keys it is run over to be for the same symbol. I'm going to check that we are in fact doing that, and add some debug checks, as a follow up.

### DIFF
--- a/cpp/arcticdb/util/key_utils.hpp
+++ b/cpp/arcticdb/util/key_utils.hpp
@@ -163,36 +163,44 @@ inline ankerl::unordered_dense::set<AtomKey> recurse_index_keys(
     ankerl::unordered_dense::set<AtomKey> res;
     ankerl::unordered_dense::set<AtomKeyPacked> res_packed;
     for (const auto& index_key: keys) {
-        if (index_key.type() == KeyType::MULTI_KEY) {
-            // recurse_index_key includes the input key in the returned set, remove this here
-            auto sub_keys = recurse_index_key(store, index_key);
-            sub_keys.erase(index_key);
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/util/key_utils.hpp
```diff
@@ -163,36 +163,44 @@ inline ankerl::unordered_dense::set<AtomKey> recurse_index_keys(
...
-            util::variant_match(
+        try {
+            if (index_key.type() == KeyType::MULTI_KEY) {
+                // recurse_index_key includes the input key in the returned set, remove this here
+                auto sub_keys = recurse_index_key(store, index_key);
+                sub_keys.erase(index_key);
+                for (auto &&key : sub_keys) {
+                    res.emplace(std::move(key));
+                }
+            } else if (index_key.type() == KeyType::TABLE_INDEX) {
```
poodlewars, line 174: I considered pulling out the different behaviours based on type to functions as this is getting quite messy, but the increased indirection made things harder to follow rather than easier.

vasil-pashov (approved):

willdealtry (approved):

PR #1858, opened 2024-09-27
Use a new `E_NON_INCREASING_INDEX_VERSION` exception

IvoDD:
The new exception type can be used to catch repeating an op log in
replication.

Apart from replication writing a non-increasing index version is most
likely due to parallel writes to the same symbol. We now indicate that
in the error message.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


poodlewars (dismissed):

poodlewars:
I don't think we realised that this error would be user facing when we first added it, so it should have a proper user facing error type rather than raising an internal error.

poodlewars (commented):
Is it possible to write a pytest for it, maybe with the library tool? Would be good to test that the mapping to Python works

poodlewars (approved):

phoebusm (approved):

PR #1861, opened 2024-09-30
Change test for library emptiness to allow specifying key types to exclude

poodlewars:
It's unintuitive that the `empty` call excludes symbol list keys.

Both STORAGE_INFO and SYMBOL_LIST keys can legitimately be added to a library after it is cleared.

I've therefore changed the empty test to allow specifying an arbitrary list of key types to exclude from the checks.

This test is used when we restore from a backup - but is failing because STORAGE_INFO keys (representing replication timestamps) are present, since replication ran after tearing down the storages.


alexowens90 (approved):

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1118,11 +1118,14 @@ void PythonVersionStore::clear(const bool continue_on_error) {
 }
 
 bool PythonVersionStore::empty() {
+    return is_empty_excluding_key_types({});
```
poodlewars, line 1121: Wrong - should exclude SYMBOL_LIST

PR #1862, opened 2024-09-30
Fix library options compatibility across ArcticDB versions

IvoDD:
This PR consists of two commits:

The first commit introduces compatibility tests. They showcase an issue with modifying library options on libraries created with version before 3.0.0.

The second commit fixes the problem by moving the modification to the c++ layer and applying the protobuf change without applying a storage_override.

I recommend these to be reviewed separately. Also read the commit descriptions before reviewing:

[Introducing compatibility tests](https://github.com/man-group/ArcticDB/pull/1862/commits/05ad97b963627c64035186be646990afe7ccb12a)
The new compatibility tests are in `tests/compat`.
This includes the following:
- `old_venv` fixture which allows executing some arctic commands frow
  within a custom venv.
- 2 compatibility tests. One of which showcases an issue with modifying
  library options of old environments.
- Tweak to the Build and Test workflow to run the compat tests only on
  newer linux runs.

[Make modify_library_option backwards compatible](https://github.com/man-group/ArcticDB/pull/1862/commits/d5d1d46c1f52889cfa6b058ff597b2e1d0d38e85)
Previously modifying a library option with a new version (>=v3.0.0)
would break old libraries (<v3.0.0)

This was because of the storage overrides introduced in v3.0.0.

This change moves the logic to modify a library option to C++
and it does the protobuf modification without applying a storage
override.

This makes storage overrides work well with both old libraries (as
demonstrated by now passing compatibility tests) and with new libraries
(as confirmed by test_arctic_library_management.py)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -185,9 +185,17 @@ jobs:
 
       - name: Discover test directory names
         if: inputs.job_type == 'build-python-wheels'
-        # There are so few nonreg tests, run them in the hypothesis runner
-        run: find python/tests/* -maxdepth 0 -type d ! -regex '.*\(__pycache__\|util\|nonreg\|scripts\)' -printf '"%f",' |
+        # We only run the compat tests on for newer python versions.
+        # There are so few nonreg tests, run them in the hypothesis runner.
+        run: |
+          if [ ${{inputs.python3}} -gt 6 ]
+          then
+            find python/tests/* -maxdepth 0 -type d ! -regex '.*\(__pycache__\|util\|nonreg\|scripts\)' -printf '"%f",' |
               sed 's/^/test_dirs=[/ ; s/"hypothesis"/"{hypothesis,nonreg,scripts}"/ ; s/,$/]/' | tee -a $GITHUB_ENV
+          else
+            find python/tests/* -maxdepth 0 -type d ! -regex '.*\(__pycache__\|util\|nonreg\|scripts\|compat\)' -printf '"%f",' |
+              sed 's/^/test_dirs=[/ ; s/"hypothesis"/"{hypothesis,nonreg,scripts}"/ ; s/,$/]/' | tee -a $GITHUB_ENV
+          fi
 
       # ========================= Common =========================
       - name: Disk usage
--- a/cpp/arcticdb/storage/library_manager.cpp
+++ b/cpp/arcticdb/storage/library_manager.cpp
@@ -86,14 +86,18 @@ LibraryManager::LibraryManager(const std::shared_ptr<storage::Library>& library)
 
 void LibraryManager::write_library_config(const py::object& lib_cfg, const LibraryPath& path, const StorageOverride& storage_override,
                           const bool validate) const {
-    SegmentInMemory segment;
-    segment.descriptor().set_index({0UL, IndexDescriptor::Type::ROWCOUNT});
-
     arcticdb::proto::storage::LibraryConfig lib_cfg_proto;
-    google::protobuf::Any output = {};
     python_util::pb_from_python(lib_cfg, lib_cfg_proto);
 
     apply_storage_override(storage_override, lib_cfg_proto, false);
+
+    write_library_config_internal(lib_cfg_proto, path, validate);
+}
+
+void LibraryManager::write_library_config_internal(const arcticdb::proto::storage::LibraryConfig& lib_cfg_proto, const LibraryPath& path, bool validate) const {
+    SegmentInMemory segment;
+    segment.descriptor().set_index({0UL, IndexDescriptor::Type::ROWCOUNT});
+    google::protobuf::Any output = {};
 
     output.PackFrom(lib_cfg_proto);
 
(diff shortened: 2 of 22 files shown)


jjerphan:
Compilation on https://github.com/conda-forge/arcticdb-feedstock/pull/298 completes.

I would recommend flushing the `*micromamba*` caches on GitHub Action.

.github/workflows/build_steps.yml
poodlewars, line 1: I'm fine with skipping the compat tests on Python 3.6, but why can't we run them on Windows?

.github/workflows/build_with_conda.yml
poodlewars, line 1: Why can't we run the new tests on Conda?

python/tests/compat/arcticdb/test_compatibility.py
poodlewars, line 1: It would be good to have tests using all the different storage backends, and a mix of options for each backend, since the compatibility depends a lot on the exact backend being used.

python/tests/compat/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+import tempfile
+
+class ErrorInVenv(Exception):
+    """To signal errors occuring from within the venv"""
+
+def run_shell_commands(commands_list, cwd=None):
+    command = " && ".join(commands_list)
+    print("Executing command: ", command)
+    result = subprocess.run(command, cwd=cwd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if result.returncode != 0 or True: # Fix
```
poodlewars, line 14: This case needs to be sorted out
- The `or True`
- Silent failure

python/tests/compat/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+        return VenvArctic(self, uri)
+
+
+class VenvArctic:
+    def __init__(self, venv, uri):
+        self.venv = venv
+        self.uri = uri
+        self.init_storage()
+
+    def execute(self, python_commands, dfs={}):
```
poodlewars, line 57: A reference to the same empty dict default for `dfs` will be shared across invocations which might cause confusion or bugs. The normal way to do this in Python is,

```
def execute(dfs=None):
  if dfs is None:
    dfs = {}
  # stuff
```

python/tests/compat/conftest.py
poodlewars, line 1: Some type hints on the new APIs you're introducing would be helpful, I think

python/tests/compat/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+        if create_if_not_exists:
+            self.create_lib()
+
+    def create_lib(self):
+        self.arctic.execute([f"ac.create_library('{self.lib_name}')"])
+
+    def execute(self, python_commands, dfs={}):
+        python_commands = [
+            f"lib = ac.get_library('{self.lib_name}')",
+            "print(lib._nvs.lib_cfg())" # Remove
```
poodlewars, line 106: TODO

python/tests/compat/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@pytest.fixture(
+    params=("lmdb", "s3_bucket_versioning")
+)
+def arctic_uri(request):
+    """
+    arctic_uri uses either lmdb or s3 uri to be used for creating both old and current Arctic instances.
+
+    We use s3_bucket_versioning because it has ssl disabled which allows running tests for older versions like 1.6.2
```
poodlewars, line 139: I think it would be more robust to define some of your own fixtures for this, than re-using `s3_bucket_versioning`

cpp/arcticdb/storage/library_manager.cpp
```diff
@@ -115,14 +119,61 @@ void LibraryManager::write_library_config(const py::object& lib_cfg, const Libra
...
+                                           std::variant<ModifiableLibraryOption, ModifiableEnterpriseLibraryOption> option,
+                                           std::variant<bool, int64_t> new_value) const {
+    // We don't apply a storage override to keep modification backwards compatible.
+    // Before v3.0.0 we didn't use storage overrides and applying one when modifying options would break readers older
+    // than v3.0.0.
+    arcticdb::proto::storage::LibraryConfig config = get_config_internal(path, std::nullopt);
+    auto mutable_write_options = config.mutable_lib_desc()->mutable_version()->mutable_write_options();
+
+    auto get_bool = [&option](std::variant<bool, int64_t> value){
+        util::check(std::holds_alternative<bool>(value), "{} only supports bool values but received {}. Not changing library option.", option, value);
```
poodlewars, line 132: I think the `util::check` will bubble up to the user in Python as an `InternalException` - instead it should be one of the user-facing exception types

cpp/arcticdb/storage/library_manager.cpp
```diff
@@ -115,14 +119,61 @@ void LibraryManager::write_library_config(const py::object& lib_cfg, const Libra
     );
 }
 
+void LibraryManager::modify_library_option(const arcticdb::storage::LibraryPath &path,
+                                           std::variant<ModifiableLibraryOption, ModifiableEnterpriseLibraryOption> option,
+                                           std::variant<bool, int64_t> new_value) const {
```
poodlewars, line 124: A type alias for `std::variant<bool, int64_t>` will help

cpp/arcticdb/storage/library_manager.cpp
```diff
@@ -115,14 +119,61 @@ void LibraryManager::write_library_config(const py::object& lib_cfg, const Libra
...
+                    case ModifiableEnterpriseLibraryOption::REPLICATION:
+                        mutable_write_options->mutable_sync_passive()->set_enabled(get_bool(new_value));
+                        break;
+                    case ModifiableEnterpriseLibraryOption::BACKGROUND_DELETION:
+                        mutable_write_options->set_delayed_deletes(get_bool(new_value));
+                        break;
+                }
+            });
+
+    // We don't want to validate old pre v3.0.0 configs
```
poodlewars, line 165: Comment is a bit confusing, mention that it's documenting the `false` parameter?

cpp/arcticdb/storage/library_manager.cpp
```diff
@@ -115,14 +119,61 @@ void LibraryManager::write_library_config(const py::object& lib_cfg, const Libra
...
+    auto get_positive_int = [&option](std::variant<bool, int64_t> value){
+        util::check(std::holds_alternative<int64_t>(value) && std::get<int64_t>(value)>0, "{} only supports positive int values but received {}. Not changing library option.", option, value);
+        return std::get<int64_t>(value);
+    };
+
+    util::variant_match(
+            option,
+            [&](const ModifiableLibraryOption& option){
+                switch (option) {
+                    case ModifiableLibraryOption::DEDUP:
```
poodlewars, line 144: Raise if the option is one that isn't implemented here?

poodlewars (changes requested):
It would be good to have some tests that verify the actual contents of the config protobuf once we're done modifying it. It would also be a useful debugging tool - to have an API that returned the protobuf unadulterated.

python/tests/compat/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+import venv
+import tempfile
+
+class ErrorInVenv(Exception):
+    """To signal errors occuring from within the venv"""
+
+def run_shell_commands(commands_list, cwd=None):
+    command = " && ".join(commands_list)
+    print("Executing command: ", command)
+    result = subprocess.run(command, cwd=cwd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
alexowens90, line 13: Do we need `shell=True`? If not, `check_call` is more idiomatic

python/tests/compat/arcticdb/test_compatibility.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+import pandas as pd
+import pytest
+from arcticdb.arctic import Arctic
+from arcticdb.util.test import assert_frame_equal
+from arcticdb_ext import set_config_int
+from arcticdb.options import ModifiableEnterpriseLibraryOption
+
+def test_compat_write_read(old_venv, arctic_uri):
```
alexowens90, line 9: I think we discussed having 2 Arctic venvs as fixtures to avoid any dependency on the current version?

python/tests/compat/conftest.py
```diff
@@ -0,0 +1,142 @@
+import pytest
+import subprocess
+import os
+import venv
+import tempfile
+
+class ErrorInVenv(Exception):
+    """To signal errors occuring from within the venv"""
+
+def run_shell_commands(commands_list, cwd=None):
```
poodlewars, line 10: `commands_list` seems to always have a single element?

.github/workflows/build_steps.yml
IvoDD: Made it work on Windows as well.

.github/workflows/build_with_conda.yml
IvoDD: Only disabled it on conda macOS as discussed

python/create_version_chain.py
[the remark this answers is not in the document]
IvoDD: Sorry it was a leftover from an old experiment. Removed it.

python/tests/compat/arcticdb/test_compatibility.py
IvoDD: Added all backends (apart from mem which can't work across venvs). Note that mongo is not supported on 1.6.2 so it is skipped.

What storage options to you have in mind? I was thinking of s3 with ssl enabled but it also doesn't work on 1.6.2 so it is a bit pointless to enable it.

python/tests/compat/conftest.py
IvoDD: Sorry, fixed.

python/tests/compat/conftest.py
IvoDD: Thanks! Ugh python is weird.

python/tests/compat/conftest.py
IvoDD: Added type hints

python/tests/compat/conftest.py
IvoDD: Thanks

python/tests/compat/conftest.py
IvoDD: yes, declared new `s3_ssl_disabled`

cpp/arcticdb/storage/library_manager.cpp
IvoDD: Thanks, fixed.

cpp/arcticdb/storage/library_manager.cpp
IvoDD: Thanks, fixed.

cpp/arcticdb/storage/library_manager.cpp
IvoDD: Done

cpp/arcticdb/storage/library_manager.cpp
IvoDD: Done raises the `UnknownLibraryOption` as before this change.

python/tests/compat/conftest.py
IvoDD: Reworked to make it work on windows as well. Now `shell=True` is required for windows.

Also [python docs](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) say:
```
The recommended approach to invoking subprocesses is to use the [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) function for all use cases it can handle. For more advanced use cases, the underlying [Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) interface can be used directly.
```

python/tests/compat/conftest.py
IvoDD: Fixed

python/tests/compat/arcticdb/test_compatibility.py
IvoDD: We discussed having different Arctic venvs as different fixtures if for some reason we need to test some very specific issue where e.g. 3 different versions are involved. We can easily separate those out of the `old_venv` fixture if needed in the future but for sake of simplicity I'm keeping it as is as now we don't need it.

Dependening on current version is critical I think. E.g. if we just have a test running only 1.6.2 and 4.5.0 we will never catch a regression in the current version which is what these tests are for. We want to know if a change we make will break e.g. modifying a 1.6.2 library in some weird surprising way.

.github/workflows/build_with_conda.yml
IvoDD: Moved to a pytest mark to skip these tests on conda.

IvoDD:
> It would be good to have some tests that verify the actual contents of the config protobuf once we're done modifying it. It would also be a useful debugging tool - to have an API that returned the protobuf unadulterated.

Added a `LibraryTool.read_unaltered_lib_cfg` and used it in the test to verify.

poodlewars (approved):

python/tests/compat/conftest.py
alexowens90: Aha, my knowledge is out of date

Issue #1827, opened 2024-09-11
get_description timezone is incorrect


Shrimpy48:
### Describe the bug

When a time-series symbol has a non-UTC index, the `date_range` returned by `get_description` follows the same timezone as the index, however the datetimes are marked as UTC.
This means that, for example, the following code may not read the whole dataset:
```python
date_range = library.get_description(symbol).date_range
data = library.read(symbol, date_range=date_range).data
```

### Steps/Code to Reproduce

```python
from arcticdb import Arctic
import pandas as pd

index = pd.to_datetime(["2024-01-01 00:00", "2024-01-01 00:01"]).tz_localize("Asia/Shanghai")
df = pd.DataFrame({'A': [1, 2]}, index=index)

ac = Arctic('s3s://...')
lib = ac['mylibrary']

lib.write('mysymbol', df)
print(lib.get_description('mysymbol').date_range)  # (datetime.datetime(2024, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), datetime.datetime(2024, 1, 1, 0, 1, tzinfo=datetime.timezone.utc))
```

### Expected Results

```python
(Timestamp('2024-01-01 00:00:00+0800', tz='Asia/Shanghai'), Timestamp('2024-01-01 00:01:00+0800', tz='Asia/Shanghai'))
```
or if the result must be UTC:
```python
(datetime.datetime(2023, 12, 31, 16, 0, tzinfo=datetime.timezone.utc), datetime.datetime(2023, 12, 31, 16, 1, tzinfo=datetime.timezone.utc))
```

### OS, Python Version and ArcticDB Version

Python: 3.10.14 (main, Apr  6 2024, 18:45:05) [GCC 9.4.0]
OS: Linux-5.15.0-119-generic-x86_64-with-glibc2.31
ArcticDB: 4.4.3

### Backend storage used

MinIO S3

PR #1866, opened 2024-10-01
Fix time zone handling in index values and metadata

### LINKED ISSUE #1406: ArcticDB crashes when when trying to normalize timezone data with MsgPack

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1406 and #1827 

- Unify the way time zones are handled for indexes and metadata, as much as possible
- Extend the support for time zone in the indexes/metadata to cover all cases supported by pandas
- Update the documentation to be more explicit about how/where time zones are supported

This is done by extending the current approach to decompose the timestamp into:
- its value - int
- its time zone info - str

This approach is beneficial as it:
- keeps the interface the same
- yields the most consistent results

The main problem is that we need to do some post processing to coerce the time zone info to be consistent ([done here](https://github.com/man-group/ArcticDB/blob/a8d3bd7d12d22cc4211a38631974407dcd812df7/python/arcticdb/version_store/_normalization.py#L180))

#### Any other comments?
An alternative approach was considered in a [past PR here](https://github.com/man-group/ArcticDB/pull/1407), but that had a couple of limitation:
- required an interface break
- was inconsistent because:
  - getting the offset value in some cases was incorrect inputs of type `datetime `
  - still needed to do some post processing to make sure that the time zone info is in a consistent format

### DIFF
--- a/docs/mkdocs/docs/faq.md
+++ b/docs/mkdocs/docs/faq.md
@@ -183,3 +183,16 @@ The handling of `NaN` in ArcticDB depends on the type of the column under consid
 * For string columns, `NaN`, as well as Python `None`, are fully supported.
 * For floating-point numeric columns, `NaN` is also fully supported.
 * For integer numeric columns `NaN` is not supported. A column that otherwise contains only integers will be treated as a floating point column if a `NaN` is encountered by ArcticDB, at which point [the usual rules](api/arctic.md#LibraryOptions) around type promotion for libraries configured with or without dynamic schema all apply as usual.
+
+### How does ArcticDB handle `time zone` information?
+
+ArcticDB takes a more strict approach to time zone handling than Pandas. While Pandas support multiple types for storing time zone information, ArcticDB coerces all time zone information to `datetime.timezone`. This way we can ensure that all time zone information is stored in a consistent manner, and that all time zone information is preserved when data is written to and read from ArcticDB.
+
+When writing data with time zone information to ArcticDB, we preserve the time zone offset and name. When reading data with time zone information from ArcticDB, this data is used to create a `datetime.timezone` object.
+
+!!! note
+
+    This means that regardles of the timezone types being written in ArcticDB, all time zone types are normalised to `datetime.timezone`.
+    If you would like a another time-zone type back then tzname can be used to recreate it:
+    - For `pytz` you can use: `pytz.gettz(dt.tzname())`
+    - For `ZoneInfo`, after Python 3.9+, you can use: `ZoneInfo(dt.tzname())`
--- a/docs/mkdocs/docs/faq.md
+++ b/docs/mkdocs/docs/faq.md
@@ -186,13 +186,12 @@ The handling of `NaN` in ArcticDB depends on the type of the column under consid
 
 ### How does ArcticDB handle `time zone` information?
 
-ArcticDB takes a more strict approach to time zone handling than Pandas. While Pandas support multiple types for storing time zone information, ArcticDB coerces all time zone information to `datetime.timezone`. This way we can ensure that all time zone information is stored in a consistent manner, and that all time zone information is preserved when data is written to and read from ArcticDB.
+ArcticDB takes a more strict approach to time zone handling than Pandas. While Pandas support multiple types for storing time zone information, ArcticDB coerces all time zone information to `pytz.tzfile` as this is the default representation for Pandas as well. This way we can ensure that all time zone information is stored in a consistent manner, and that all time zone information is preserved when data is written to and read from ArcticDB.
 
-When writing data with time zone information to ArcticDB, we preserve the time zone offset and name. When reading data with time zone information from ArcticDB, this data is used to create a `datetime.timezone` object.
+When writing data with time zone information to ArcticDB, we preserve the time name as string (e.g. "Europe/Amsterdam"). When reading data with time zone information from ArcticDB, this data is used to create a `pytz.tzfile` object.
 
 !!! note
 
-    This means that regardles of the timezone types being written in ArcticDB, all time zone types are normalised to `datetime.timezone`.
-    If you would like a another time-zone type back then tzname can be used to recreate it:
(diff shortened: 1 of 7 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -317,16 +373,16 @@ def _normalize_single_index(index, index_names, index_norm, dynamic_strings=None
                 index_norm.fake_field_pos.append(0)
             log.debug("Index has no name, defaulting to 'index'")
         if isinstance(index, DatetimeIndex) and index.tz is not None:
-            index_tz = get_timezone(index.tz)
+            index_tz = index
```
poodlewars, line 376: I think the naming has gone a bit weird here, `index_tz` being `pd.Index or np.ndarray` is odd

python/tests/integration/arcticdb/test_persistent_storage.py
```diff
@@ -18,8 +18,10 @@
 
 
 @pytest.fixture(params=[pytest.param("REAL_S3", marks=REAL_S3_TESTS_MARK)])
-def shared_persistent_arctic_client(real_s3_storage_without_clean_up, encoding_version):
-    return real_s3_storage_without_clean_up.create_arctic(encoding_version=encoding_version)
+# DONT MERGE: there is a problem with interop tests and encoding_version 2
```
poodlewars, line 21: TODO

docs/mkdocs/docs/faq.md
```diff
@@ -183,3 +183,15 @@ The handling of `NaN` in ArcticDB depends on the type of the column under consid
...
 * For floating-point numeric columns, `NaN` is also fully supported.
 * For integer numeric columns `NaN` is not supported. A column that otherwise contains only integers will be treated as a floating point column if a `NaN` is encountered by ArcticDB, at which point [the usual rules](api/arctic.md#LibraryOptions) around type promotion for libraries configured with or without dynamic schema all apply as usual.
+
+### How does ArcticDB handle `time zone` information?
+
+ArcticDB takes a more strict approach to time zone handling than Pandas. While Pandas support multiple types for storing time zone information, ArcticDB coerces all time zone information to `pytz.tzfile` as this is the default representation for Pandas as well. This way we can ensure that all time zone information is stored in a consistent manner, and that all time zone information is preserved when data is written to and read from ArcticDB.
+
+When writing data with time zone information to ArcticDB, we preserve the time name as string (e.g. "Europe/Amsterdam"). When reading data with time zone information from ArcticDB, this data is used to create a `pytz.tzfile` object.
```
poodlewars, line 191: What happens if the writer uses a timezone file that is not in the reader's zone database? How does this work across operating systems? I assume the database files are quite different across Linux and Windows, but that may be a miscomprehension.

docs/mkdocs/docs/faq.md
```diff
@@ -183,3 +183,15 @@ The handling of `NaN` in ArcticDB depends on the type of the column under consid
...
+
+!!! note
+
+    This means that regardles of the timezone types being written in ArcticDB, all time zone types are normalised to `pytz.tzfile`.
```
poodlewars, line 195: I think it's important that we round-trip correctly - this could be quite a disruptive change otherwise. At least requiring a new major version - unless I'm misunderstanding? I notice the PR description says that this `keeps the interface the same`.

docs/mkdocs/docs/faq.md
poodlewars, line 1: I can't repro the crash in issue #1406 anymore

docs/mkdocs/docs/faq.md
```diff
@@ -183,3 +183,15 @@ The handling of `NaN` in ArcticDB depends on the type of the column under consid
 * For string columns, `NaN`, as well as Python `None`, are fully supported.
 * For floating-point numeric columns, `NaN` is also fully supported.
 * For integer numeric columns `NaN` is not supported. A column that otherwise contains only integers will be treated as a floating point column if a `NaN` is encountered by ArcticDB, at which point [the usual rules](api/arctic.md#LibraryOptions) around type promotion for libraries configured with or without dynamic schema all apply as usual.
+
+### How does ArcticDB handle `time zone` information?
+
+ArcticDB takes a more strict approach to time zone handling than Pandas. While Pandas support multiple types for storing time zone information, ArcticDB coerces all time zone information to `pytz.tzfile` as this is the default representation for Pandas as well. This way we can ensure that all time zone information is stored in a consistent manner, and that all time zone information is preserved when data is written to and read from ArcticDB.
```
poodlewars, line 189: Is `pytz` actually the default in Pandas? https://github.com/pandas-dev/pandas/blob/main/pyproject.toml#L33

PR #1872, opened 2024-10-03
Customize s3 timouts with ConfigsMap

IvoDD:
Now allows configuring custom:
- `S3.ConnectionTimeout`
- `S3.RequestTimeout`
- `S3.LowSpeedLimit`

This is useful for networking issues to allow failing faster with lower
timeout or for increasing the timeout if we expect to read large chunks.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #1875, opened 2024-10-06
Allow customizing the most useful AWS SDK settings.

poodlewars:
`ARCTICDB_S3Storage_HttpRequestTimeoutMs_int=3000`

suffices to resolve the slowness issues users have been reporting with VAST.

I've given the new settings a default in line with the SDK defaults - https://sdk.amazonaws.com/cpp/api/LATEST/aws-cpp-sdk-core/html/struct_aws_1_1_client_1_1_client_configuration.html .

Similar PR from @IvoDD  - https://github.com/man-group/ArcticDB/pull/1872/files

### DIFF
--- a/cpp/arcticdb/storage/s3/s3_storage.hpp
+++ b/cpp/arcticdb/storage/s3/s3_storage.hpp
@@ -235,11 +235,13 @@ auto get_s3_config(const ConfigType& conf) {
     client_configuration.maxConnections = conf.max_connections() == 0 ?
             ConfigsMap::instance()->get_int("VersionStore.NumIOThreads", 16) :
             conf.max_connections();
-    client_configuration.connectTimeoutMs = conf.connect_timeout() == 0 ?
-                                          ConfigsMap::instance()->get_int("S3Storage.ConnectTimeoutMs", 30000) :
-                                          conf.connect_timeout();
+
+
+    client_configuration.connectTimeoutMs = ConfigsMap::instance()->get_int("S3Storage.ConnectTimeoutMs",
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/storage/s3/s3_storage.hpp
```diff
@@ -235,8 +235,12 @@ auto get_s3_config(const ConfigType& conf) {
     client_configuration.maxConnections = conf.max_connections() == 0 ?
             ConfigsMap::instance()->get_int("VersionStore.NumIOThreads", 16) :
             conf.max_connections();
-    client_configuration.connectTimeoutMs = conf.connect_timeout() == 0 ? 30000 : conf.connect_timeout();
-    client_configuration.requestTimeoutMs = conf.request_timeout() == 0 ? 200000 : conf.request_timeout();
+    client_configuration.connectTimeoutMs = conf.connect_timeout() == 0 ?
+                                          ConfigsMap::instance()->get_int("S3Storage.ConnectTimeoutMs", 30000) :
+                                          conf.connect_timeout();
+    client_configuration.httpRequestTimeoutMs = ConfigsMap::instance()->get_int("S3Storage.HttpRequestTimeoutMs", 0);
```
IvoDD, line 241: I don't know if there was ever a way to set `conf.request_timeout()` on the library config but now we don't look at it at all.

Also for the connection timeout it feels like a better approach is to allow configsmap to override anything written in the library config (which is hard to change), i.e.
`ConfigsMap::instance()->get_int("S3Storage.ConnectTimeoutMs", conf.connect_timeout() == 0 ? 3000 : conf.connect_timeout())`

IvoDD (approved):

PR #1878, opened 2024-10-07
Bugfix 1841: Correctly roundtrip None and empty string pd.Series names

### LINKED ISSUE #1841: Series normalizer converts empty series name to None
### Describe the bug

When a series is stored into ArcticDB and the name of the series is empty string it gets converted to `None` on read.

### Steps/Code to Reproduce

```python
import pandas as pd
import numpy as np
import arcticdb as adb

ac = adb.Arctic("lmdb://test")

series = pd.Series(data=[1], index=pd.DatetimeIndex([pd.Timestamp(2024, 1, 1)]), name='')
print(type(series.name)) # <class 'str'>
lib.write("sym", series)
res = lib.read("sym").data
print(type(res.name)) # <class 'NoneType'>
```

### Expected Results

Keep the name the series an empty string

OS: Windows-10-10.0.22631-SP0
ArcticDB: 4.5.0

### Backend storage used

_No response_

### Additional Context

_No response_

alexowens90:
#### Reference Issues/PRs
Fixes #1841

### DIFF
--- a/cpp/proto/arcticc/pb2/descriptors.proto
+++ b/cpp/proto/arcticc/pb2/descriptors.proto
@@ -163,6 +163,8 @@ message NormalizationMetadata {
 
         // Pandas series support arbitrary objects but restricting to a string seems reasonable
         string name = 1;
+        // Backwards-compatible way of treating name as optional
+        bool has_name = 9;
 
         oneof index_type {
             PandasIndex index = 2;
--- a/cpp/proto/arcticc/pb2/descriptors.proto
+++ b/cpp/proto/arcticc/pb2/descriptors.proto
@@ -192,6 +192,7 @@ message NormalizationMetadata {
 
         map<string, ColumnName> col_names = 7;
         PandasIndex columns = 8;
+        // Note that 9 is taken above by has_name, so the next added field should be 10
     }
 
     message PandasDataFrame {
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -588,8 +588,10 @@ class SeriesNormalizer(_PandasNormalizer):
             empty_types=empty_types
         )
         norm.series.CopyFrom(norm.df)
-        if item.name:
+        if item.name is not None:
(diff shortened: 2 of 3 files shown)


cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -163,6 +163,8 @@ message NormalizationMetadata {
 
         // Pandas series support arbitrary objects but restricting to a string seems reasonable
         string name = 1;
+        // Backwards-compatible way of treating name as optional
+        bool has_name = 9;
```
poodlewars, line 167: Worth a comment below `columns = 8` to explain that there is a field with index `9`

python/arcticdb/version_store/_normalization.py
```diff
@@ -600,10 +602,19 @@ def denormalize(self, item, norm_meta):
 
         series = pd.Series() if df.columns.empty else df.iloc[:, 0]
 
-        if norm_meta.common.name:
-            series.name = norm_meta.common.name
+        if hasattr(norm_meta.common, "has_name"):
```
poodlewars, line 605: Is there a similar problem for dataframe index names, or is that handled correctly?

python/tests/unit/arcticdb/version_store/test_normalization.py
poodlewars, line 1: I think it would be worth having a test to check that an older client can work with series written by a newer client.

Might be better done after https://github.com/man-group/ArcticDB/pull/1862

poodlewars (approved):

python/arcticdb/version_store/_normalization.py
alexowens90: Turns out this already worked, added a test for it though

python/tests/unit/arcticdb/version_store/test_normalization.py
alexowens90: In general I would agree, you can tell by inspection that this is correct though. The `name` field of the protobuf is populated identically with the new client as with the old, so the old client will still be able to read it, just with the bug of not being able to distinguish Nones from empty strings

PR #1814, opened 2024-09-04
-Introduce aws_auth to proto

poodlewars:
#### Reference Issues/PRs

https://github.com/man-group/ArcticDB/issues/1883

#### What does this implement or fix?
Add support of AWS STS authentication

#### Any other comments?
STS is an AWS iam service which allows user account to gain access of resources by assuming role. 
First, the SDK will send assume role request to iam. If the authentication is successful, iam will reply with a temporary access key. With that access key, SDK will now have access to the resources. 
The above logic and the refresh of temporary access key are handled by the SDK.
P.S. the validity check of the temporary access key is made at the beginning of each IO only,

Maintainers of S3 C++ SDK has [refused](https://github.com/aws/aws-sdk-cpp/issues/150) to align this authentication method with other SDKs (e.g. boto3), which create 2 problems:
1. Role to be assumed is needed to specifed in the API but the role_arn, access key and id are needed to be specifed in the shared config file
2. The method is not added to the default credential provider chain. Although the SDK allows us to supply a customized chain with the STS authentication method, it has led to two drawbacks, which force me to switch to a dedicated option to switch on STS authentication:
* SDK will not auto refresh temporary credential. ArcticDB would be required to add a tedious key expiry detection and refresh logic
* Extra maintainece is required as the chain will need stay align with the default one to align the user experience with previous user experience

Detailed setup of the STS method can be referred to the content of this PR.
The test added in this PR requires real S3. Temporary user, policy and role are created for the test in the pipeline. 

The final caveat is seems the **config** file is loaded at the import of ArcticDB and not being refreshed at the entire lifecycle of the python interpreter. This has required a hack in the testing framework to always create the config file at the beginning of any tests. As it's real S3 test specific, it won't create any pain for day-to-day local development and general PR tests.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #1884, opened 2024-10-08
Add AWS STS authentication support

phoebusm:
Copy of https://github.com/man-group/ArcticDB/pull/1814 so @poodlewars can review

#### Reference Issues/PRs

https://github.com/man-group/ArcticDB/issues/1883

Add support of AWS STS authentication
Partially introduce static storage map along with the above

#### Any other comments?
STS is an AWS iam service which allows user account to gain access of resources by assuming role. 
First, the SDK will send assume role request to iam. If the authentication is successful, iam will reply with a temporary access key. With that access key, SDK will now have access to the resources. 
The above logic and the refresh of temporary access key are handled by the SDK.
P.S. the validity check of the temporary access key is made at the beginning of each IO only,

Maintainers of S3 C++ SDK has [refused](https://github.com/aws/aws-sdk-cpp/issues/150) to align this authentication method with other SDKs (e.g. boto3), which create 2 problems:
1. Role to be assumed is needed to specifed in the API but the role_arn, access key and id are needed to be specifed in the shared config file
2. The method is not added to the default credential provider chain. Although the SDK allows us to supply a customized chain with the STS authentication method, it has led to two drawbacks, which force me to switch to a dedicated option to switch on STS authentication:
* SDK will not auto refresh temporary credential. ArcticDB would be required to add a tedious key expiry detection and refresh logic
* Extra maintainece is required as the chain will need stay align with the default one to align the user experience with previous user experience

Detailed setup of the STS method can be referred to the content of this PR.
The test added in this PR requires real S3. Temporary user, policy and role are created for the test in the pipeline. 

The final caveat is seems the **config** file is loaded at the import of ArcticDB and not being refreshed at the entire lifecycle of the python interpreter. This has required a hack in the testing framework to always create the config file at the beginning of any tests. As it's real S3 test specific, it won't create any pain for day-to-day local development and general PR tests.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/common.hpp
+++ b/cpp/arcticdb/storage/common.hpp
@@ -13,6 +13,7 @@
 #include <arcticdb/entity/protobufs.hpp>
 #include <arcticdb/entity/variant_key.hpp>
 #include <arcticdb/entity/performance_tracing.hpp>
+#include <arcticdb/storage/s3/s3_settings.hpp>
 #include <sstream>
 
 namespace arcticdb::storage {
@@ -91,5 +92,17 @@ struct is_key_type<entity::VariantKey> : std::true_type {};
 template <typename T>
 inline constexpr bool is_key_type_v = is_key_type<T>::value;
 
+class NativeVariantStorage {
+public:
+    using VariantStorageConfig = std::variant<std::monostate, s3::S3Settings>;
+    explicit NativeVariantStorage(VariantStorageConfig config = std::monostate()) : config_(std::move(config)) {};
+    const VariantStorageConfig& variant() const {
+        return config_;
+    }
+    void update(const s3::S3Settings& config) {
+        config_ = config;
+    }
+private:
+    VariantStorageConfig config_;
+};
 }  //namespace arcticdb::storage
-
--- a/cpp/arcticdb/storage/config_cache.hpp
+++ b/cpp/arcticdb/storage/config_cache.hpp
@@ -41,20 +41,10 @@ class ConfigCache {
         return descriptor_map_.find(path) != descriptor_map_.end();
     }
 
-    void add_library_config(const LibraryPath &path, const arcticdb::proto::storage::LibraryConfig lib_cfg) {
-        add_library(path, decode_library_descriptor(lib_cfg.lib_desc()));
-        for(const auto& storage: lib_cfg.storage_by_id())
-            add_storage(StorageName{storage.first}, storage.second);
-    }
-
     void add_library(const LibraryPath &path, const LibraryDescriptor &desc) {
         config_resolver_->add_library(environment_name_, encode_library_descriptor(desc));
         std::lock_guard<std::mutex> lock{mutex_};
         descriptor_map_.emplace(path, desc);
-    }
-
-    void add_storage(const StorageName& storage_name, const arcticdb::proto::storage::VariantStorage& storage) {
-        config_resolver_->add_storage(environment_name_, storage_name, storage);
     }
 
     std::vector<LibraryPath> list_libraries(std::string_view prefix) {
@@ -70,7 +60,7 @@ class ConfigCache {
         return res;
     }
 
-    std::shared_ptr<Storages> create_storages(const LibraryPath &path, OpenMode mode) {
+    std::shared_ptr<Storages> create_storages(const LibraryPath &path, OpenMode mode, const NativeVariantStorage& native_storage_config) {
         auto maybe_descriptor = get_descriptor(path);
         if (!maybe_descriptor.has_value())
             throw std::runtime_error(fmt::format("Library {} not found", path));
@@ -92,7 +82,17 @@ class ConfigCache {
             if(storage_conf_pos != storage_configs_.end())
                 storage_conf = storage_conf_pos->second;
 
-            storages.emplace_back(create_storage(path, mode, storage_conf));
+            util::variant_match(native_storage_config.variant(),
+                [&storage_conf, &storages, &path, mode] (const s3::S3Settings& settings) {
+                    util::check(storage_conf.config().Is<arcticdb::proto::s3_storage::Config>(), "Only support S3 native settings");
+                    arcticdb::proto::s3_storage::Config s3_storage;
+                    storage_conf.config().UnpackTo(&s3_storage);
+                    storages.emplace_back(create_storage(path, mode, s3::S3Settings(settings).update(s3_storage)));
+                },
+                [&storage_conf, &storages, &path, mode](const auto &) {
+                    storages.emplace_back(create_storage(path, mode, storage_conf));
+                }
+            );
         }
         return std::make_shared<Storages>(std::move(storages), mode);
     }
@@ -110,13 +110,6 @@ class ConfigCache {
         auto storages = config_resolver_->get_storages(environment_name_);
         for(auto& [storage_name, config] : storages) {
             storage_configs_.try_emplace(StorageName(storage_name), config);
-        }
-        auto default_storages = config_resolver_->get_default_storages(environment_name_);
-        for(auto& [storage_name, config] : default_storages) {
-            if (storage_configs_.find(storage_name) == storage_configs_.end()) {
-                config_resolver_->add_storage(environment_name_, storage_name, config);
-                storage_configs_.try_emplace(StorageName(storage_name), config);
-            }
         }
     }
 
--- a/cpp/arcticdb/storage/config_resolvers.hpp
+++ b/cpp/arcticdb/storage/config_resolvers.hpp
@@ -27,7 +27,6 @@ class ConfigResolver {
     virtual void add_library(const EnvironmentName& environment_name, const arcticdb::proto::storage::LibraryDescriptor& library_descriptor) = 0;
     virtual void add_storage(const EnvironmentName& environment_name, const StorageName& storage_name, const arcticdb::proto::storage::VariantStorage& storage) = 0;
     virtual void initialize_environment(const EnvironmentName& environment_name) = 0;
-    virtual std::vector<std::pair<StorageName, arcticdb::proto::storage::VariantStorage>> get_default_storages(const EnvironmentName& environment_name) const = 0;
     virtual std::string_view resolver_type() const = 0;
 };
 
@@ -62,9 +61,6 @@ class InMemoryConfigResolver final : public ConfigResolver {
 
     void add_library(const EnvironmentName& environment_name, const arcticdb::proto::storage::LibraryDescriptor& library_descriptor) override;
     void add_storage(const EnvironmentName& environment_name, const StorageName& storage_name, const arcticdb::proto::storage::VariantStorage& storage) override;
-    std::vector<std::pair<StorageName, arcticdb::proto::storage::VariantStorage>> get_default_storages(const EnvironmentName&) const override {
-        return std::vector<std::pair<StorageName, arcticdb::proto::storage::VariantStorage>>();
-    }
 
     void initialize_environment(const EnvironmentName&) override { }
     std::string_view resolver_type()  const override { return "in_mem"; }
--- a/cpp/arcticdb/storage/python_bindings.cpp
+++ b/cpp/arcticdb/storage/python_bindings.cpp
@@ -18,6 +18,8 @@
 #include <arcticdb/storage/library_index.hpp>
 #include <arcticdb/storage/config_resolvers.hpp>
 #include <arcticdb/storage/constants.hpp>
+#include <arcticdb/storage/s3/s3_storage.hpp>
+#include <arcticdb/storage/s3/s3_settings.hpp>
 
 namespace py = pybind11;
 
@@ -93,6 +95,56 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
     py::register_exception<UnsupportedLibraryOptionValue>(storage, "UnsupportedLibraryOptionValue", base_exception.ptr());
 
     storage.def("create_library_index", &create_library_index);
+
+    
+    py::enum_<s3::AWSAuthMethod>(storage, "AWSAuthMethod")
+        .value("DISABLED", s3::AWSAuthMethod::DISABLED)
+        .value("DEFAULT_CREDENTIALS_PROVIDER_CHAIN", s3::AWSAuthMethod::DEFAULT_CREDENTIALS_PROVIDER_CHAIN)
+        .value("STS_PROFILE_CREDENTIALS_PROVIDER", s3::AWSAuthMethod::STS_PROFILE_CREDENTIALS_PROVIDER);
+
+    enum class S3SettingsPickleOrder : uint32_t {
+        AWS_AUTH = 0,
+        AWS_PROFILE = 1
+    };
+
+    py::class_<s3::S3Settings>(storage, "S3Settings")
+        .def(py::init<s3::AWSAuthMethod, const std::string&>())
+        .def(py::pickle(
+            [](const s3::S3Settings &settings) {
+                return py::make_tuple(settings.aws_auth(), settings.aws_profile());
+            },
+            [](py::tuple t) {
+                util::check(t.size() == 2, "Invalid S3Settings pickle objects");
+                s3::S3Settings settings(t[static_cast<uint32_t>(S3SettingsPickleOrder::AWS_AUTH)].cast<s3::AWSAuthMethod>(), t[static_cast<uint32_t>(S3SettingsPickleOrder::AWS_PROFILE)].cast<std::string>());
+                return settings;
+            }
+        ))
+        .def_property_readonly("aws_profile", [](const s3::S3Settings &settings) { return settings.aws_profile(); })
+        .def_property_readonly("aws_auth", [](const s3::S3Settings &settings) { return settings.aws_auth(); });
+
+    py::class_<NativeVariantStorage>(storage, "NativeVariantStorage")
+        .def(py::init<>())
+        .def(py::init<NativeVariantStorage::VariantStorageConfig>())
+        .def(py::pickle(
+            [](const NativeVariantStorage &settings) {
+                return util::variant_match(settings.variant(),
+                    [] (const s3::S3Settings& settings) {
+                        return py::make_tuple(settings.aws_auth(), settings.aws_profile());
+                    },
+                    [](const auto &) {
+                        util::raise_rte("Invalid native storage setting type");
+                        return py::make_tuple();
+                    }
+                );
+            },
+            [](py::tuple t) {
+                util::check(t.size() == 2, "Invalid S3Settings pickle objects");
+                s3::S3Settings settings(t[static_cast<uint32_t>(S3SettingsPickleOrder::AWS_AUTH)].cast<s3::AWSAuthMethod>(), t[static_cast<uint32_t>(S3SettingsPickleOrder::AWS_PROFILE)].cast<std::string>());
+                return NativeVariantStorage(std::move(settings));
+            }
+        ))
+        .def("update", &NativeVariantStorage::update);
+    py::implicitly_convertible<NativeVariantStorage::VariantStorageConfig, NativeVariantStorage>();
 
     storage.def("create_mem_config_resolver", [](const py::object & env_config_map_py) -> std::shared_ptr<ConfigResolver> {
         arcticdb::proto::storage::EnvironmentConfigsMap ecm;
@@ -192,12 +244,14 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
         .def("get_library", [](
                 LibraryManager& library_manager, std::string_view library_path,
                 const StorageOverride& storage_override,
-                const bool ignore_cache) {
-            return library_manager.get_library(LibraryPath{library_path, '.'}, storage_override, ignore_cache);
+                const bool ignore_cache,
+                const NativeVariantStorage& native_storage_config) {
+            return library_manager.get_library(LibraryPath{library_path, '.'}, storage_override, ignore_cache, native_storage_config);
         },
              py::arg("library_path"),
              py::arg("storage_override") = StorageOverride{},
-             py::arg("ignore_cache") = false
+             py::arg("ignore_cache") = false,
+             py::arg("native_storage_map") = std::nullopt
         )
         .def("cleanup_library_if_open", [](LibraryManager& library_manager, std::string_view library_path) {
             return library_manager.cleanup_library_if_open(LibraryPath{library_path, '.'});
@@ -228,10 +282,10 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
                 res.emplace_back(lp.to_delim_path());
             }
             return res;
-        } )
-        .def("get_library", [](LibraryIndex &library_index, const std::string &library_path, OpenMode open_mode = OpenMode::DELETE) {
+        })
+        .def("get_library", [](LibraryIndex &library_index, const std::string &library_path, OpenMode open_mode = OpenMode::DELETE, const NativeVariantStorage& native_storage_config = NativeVariantStorage()) {
             LibraryPath path = LibraryPath::from_delim_path(library_path);
-            return library_index.get_library(path, open_mode, UserAuth{});
+            return library_index.get_library(path, open_mode, UserAuth{}, native_storage_config);
         })
         ;
 }
--- a/cpp/arcticdb/storage/s3/s3_settings.hpp
+++ b/cpp/arcticdb/storage/s3/s3_settings.hpp
@@ -1,0 +1,142 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <string>
+#include <arcticdb/entity/protobufs.hpp>
+
+namespace arcticdb::storage::s3 {
+
+enum class AWSAuthMethod : uint32_t {
+    DISABLED = 0,
+    DEFAULT_CREDENTIALS_PROVIDER_CHAIN = 1,
+    STS_PROFILE_CREDENTIALS_PROVIDER = 2
+};
+
+class S3Settings {
+private:
+    std::string bucket_name_;
+    std::string credential_name_;
+    std::string credential_key_;
+    std::string endpoint_;
+    uint32_t max_connections_;
+    uint32_t connect_timeout_;
+    uint32_t request_timeout_;
+    bool ssl_;
+    std::string prefix_;
+    bool https_;
+    std::string region_;
+    bool use_virtual_addressing_;
+    bool use_mock_storage_for_testing_;
+    std::string ca_cert_path_;
+    std::string ca_cert_dir_;
+    bool use_raw_prefix_;
+    AWSAuthMethod aws_auth_;
+    std::string aws_profile_;
+
+public:
+    explicit S3Settings(AWSAuthMethod aws_auth, const std::string& aws_profile) :
+        aws_auth_(aws_auth),
+        aws_profile_(aws_profile) {
+    }
+    explicit S3Settings(const arcticc::pb2::s3_storage_pb2::Config& config) {
+        update(config);
+    }
+    S3Settings update(const arcticc::pb2::s3_storage_pb2::Config& config){
+        bucket_name_ = config.bucket_name();
+        credential_name_ = config.credential_name();
+        credential_key_ = config.credential_key();
+        endpoint_ = config.endpoint();
+        max_connections_ = config.max_connections();
+        connect_timeout_ = config.connect_timeout();
+        request_timeout_ = config.request_timeout();
+        ssl_ = config.ssl();
+        prefix_ = config.prefix();
+        https_ = config.https();
+        region_ = config.region();
+        use_virtual_addressing_ = config.use_virtual_addressing();
+        use_mock_storage_for_testing_ = config.use_mock_storage_for_testing();
+        ca_cert_path_ = config.ca_cert_path();
+        ca_cert_dir_ = config.ca_cert_dir();
+        use_raw_prefix_ = config.use_raw_prefix();
+        return *this;
+    }
+
+    std::string bucket_name() const {
+        return bucket_name_;
+    }
+
+    std::string credential_name() const {
+        return credential_name_;
+    }
+
+    std::string credential_key() const {
+        return credential_key_;
+    }
+
+    std::string endpoint() const {
+        return endpoint_;
+    }
+
+    uint32_t max_connections() const {
+        return max_connections_;
+    }
+
+    uint32_t connect_timeout() const {
+        return connect_timeout_;
+    }
+
+    uint32_t request_timeout() const {
+        return request_timeout_;
+    }
+
+    bool ssl() const {
+        return ssl_;
+    }
+
+    std::string prefix() const {
+        return prefix_;
+    }
+
+    bool https() const {
+        return https_;
+    }
+
+    std::string region() const {
+        return region_;
+    }
+
+    bool use_virtual_addressing() const {
+        return use_virtual_addressing_;
+    }
+
+    bool use_mock_storage_for_testing() const {
+        return use_mock_storage_for_testing_;
+    }
+
+    std::string ca_cert_path() const {
+        return ca_cert_path_;
+    }
+
+    std::string ca_cert_dir() const {
+        return ca_cert_dir_;
+    }
+
+    AWSAuthMethod aws_auth() const {
+        return aws_auth_;
+    }
+
+    std::string aws_profile() const {
+        return aws_profile_;
+    }
+
+    bool use_raw_prefix() const {
+        return use_raw_prefix_;
+    }
+};
+}
--- a/cpp/arcticdb/storage/s3/s3_storage.cpp
+++ b/cpp/arcticdb/storage/s3/s3_storage.cpp
@@ -9,6 +9,8 @@
 
 #include <locale>
 
+#include <aws/identity-management/auth/STSProfileCredentialsProvider.h>
+
 #include <arcticdb/log/log.hpp>
 #include <arcticdb/storage/s3/s3_api.hpp>
 #include <arcticdb/util/buffer_pool.hpp>
@@ -80,18 +82,26 @@ bool S3Storage::do_key_exists(const VariantKey& key) {
 
 namespace arcticdb::storage::s3 {
 
-S3Storage::S3Storage(const LibraryPath &library_path, OpenMode mode, const Config &conf) :
+S3Storage::S3Storage(const LibraryPath &library_path, OpenMode mode, const S3Settings &conf) :
     Storage(library_path, mode),
     s3_api_(S3ApiInstance::instance()),  // make sure we have an initialized AWS SDK
     root_folder_(object_store_utils::get_root_folder(library_path)),
     bucket_name_(conf.bucket_name()),
     region_(conf.region()) {
-
     auto creds = get_aws_credentials(conf);
 
     if (conf.use_mock_storage_for_testing()){
         ARCTICDB_RUNTIME_DEBUG(log::storage(), "Using Mock S3 storage");
         s3_client_ = std::make_unique<MockS3Client>();
+    }
+    else if (conf.aws_auth() == AWSAuthMethod::STS_PROFILE_CREDENTIALS_PROVIDER){
+        Aws::Config::ReloadCachedConfigFile(); // config files loaded in Aws::InitAPI; It runs once at first S3Storage object construct; reload to get latest
+        auto cred_provider = Aws::MakeShared<Aws::Auth::STSProfileCredentialsProvider>(
+                                "DefaultAWSCredentialsProviderChain", 
+                                conf.aws_profile(), 
+                                std::chrono::minutes(static_cast<size_t>(ConfigsMap::instance()->get_int("S3Storage.STSTokenExpiryMin", 60)))
+                            );
+        s3_client_ = std::make_unique<RealS3Client>(cred_provider, get_s3_config(conf), Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy::Never, conf.use_virtual_addressing());
     }
     else if (creds.GetAWSAccessKeyId() == USE_AWS_CRED_PROVIDERS_TOKEN && creds.GetAWSSecretKey() == USE_AWS_CRED_PROVIDERS_TOKEN){
         ARCTICDB_RUNTIME_DEBUG(log::storage(), "Using AWS auth mechanisms");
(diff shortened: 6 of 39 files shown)


cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -117,7 +118,22 @@ void register_bindings(py::module& storage) {
         .def_property("ca_cert_path", &S3Override::ca_cert_path, &S3Override::set_ca_cert_path)
         .def_property("ca_cert_dir", &S3Override::ca_cert_dir, &S3Override::set_ca_cert_dir)
         .def_property("https", &S3Override::https, &S3Override::set_https)
-        .def_property("ssl", &S3Override::ssl, &S3Override::set_ssl);
+        .def_property("ssl", &S3Override::ssl, &S3Override::set_ssl)
+        .def_property("aws_auth",
+            // pybind cannot smartly convert AWSAuthMethod defined in proto from C++ layer to Python layer
```
poodlewars, line 123: Remove the comment

cpp/proto/arcticc/pb2/s3_storage.proto
```diff
@@ -27,4 +29,6 @@ message Config {
     string ca_cert_dir = 15;
     // Whether to leave the prefix unchanged, or to translate the periods in it.
     bool use_raw_prefix = 16;
+    arcticc.pb2.storage_pb2.AWSAuthMethod aws_auth = 17; //default to be 0, DISABLED
```
poodlewars, line 32: I'm very worried about adding anything to these protobufs after all the issues recently with the `https` flag.

There is actually no reason for this to go in the protobuf at all is there? We never intend to serialize it?

We will need extensive compatibility testing (similar to https://github.com/man-group/ArcticDB/pull/1862 ) for this change either way.

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -86,13 +88,17 @@ S3Storage::S3Storage(const LibraryPath &library_path, OpenMode mode, const Confi
     root_folder_(object_store_utils::get_root_folder(library_path)),
     bucket_name_(conf.bucket_name()),
     region_(conf.region()) {
-
+
```
poodlewars, line 91: random whitespace

docs/mkdocs/docs/api/arctic_uri.md
```diff
@@ -22,11 +22,45 @@ Available options for S3:
...
 | secret                | S3 secret access key                                                                                                                                            |
 | path_prefix           | Path within S3 bucket to use for data storage                                                                                                                   |
-| aws_auth              | If true, authentication to endpoint will be computed via AWS environment vars/config files. If no options are provided `aws_auth` will be assumed to be true.   |
+| aws_auth              | AWS authentication method. If setting is `1` (or `true` for backward compatibility), authentication to endpoint will be computed via [AWS default credential provider chain](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/credproviders.html). If setting is `2`, AWS Security Token Service (STS) will be the authentication method used. If no options are provided AWS authentication will be assumed to be not used. More info is provided below |
```
poodlewars, line 25: Let's use human readable terms for this. `default` or `true` for the current behaviour, `sts` for this new behaviour?

poodlewars, line 25: `will be assumed to be not used` replace with, `will not be used and you should specify access and secret in the URI`

docs/mkdocs/docs/api/arctic_uri.md
```diff
@@ -22,11 +22,45 @@ Available options for S3:
...
+| aws_profile           | Only when `aws_auth` is set to be `2`. AWS profile to be used with AWS Security Token Service (STS). More info is provided below |
```
poodlewars, line 26: Why is this in the URI? Why don't we just use the `AWS_PROFILE` env var?

docs/mkdocs/docs/api/arctic_uri.md
```diff
@@ -22,11 +22,45 @@ Available options for S3:
...
 
 Note: When connecting to AWS, `region` can be automatically deduced from the endpoint if the given endpoint
 specifies the region and `region` is not set.
 
+### AWS Security Token Service (STS) setup
+
+STS allows users to assume specfic roles in order to gain temporary access to AWS resources. Please refer to [the offical website](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html) for more details
```
poodlewars, line 33: full stop

docs/mkdocs/docs/api/arctic_uri.md
```diff
@@ -22,11 +22,45 @@ Available options for S3:
...
+aws_secret_access_key = KEY
+```
+
+Use the configuration in ArcticDB:
+```
+>>> import arcticdb as adb
+>>> arctic = adb.Arctic('s3://s3.REGION.amazonaws.com:BUCKET?aws_auth=2&aws_profile=PROFILE')
+```
+
+For setting up resources with STS on iam, test function `real_s3_sts_from_environment_variables` in `s3.py` will be a good reference
```
poodlewars, line 61: I think remove this sentence. It will be too easy for us to break with a refactor.

python/arcticdb/adapters/s3_library_adapter.py
```diff
@@ -160,6 +174,15 @@ def _parse_query(self, query: str) -> ParsedQuery:
             if field_dict[key].type == Optional[bool] and field_dict[key] is not None:
                 parsed_query[key] = bool(strtobool(parsed_query[key][0]))
 
+            if key == "aws_auth" and field_dict[key] is not None:
```
poodlewars, line 177: Keep the existing pattern where we test based on `.type` rather than names

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -12,27 +12,35 @@
...
 
 # All storage client libraries to be imported on-demand to speed up start-up of ad-hoc test runs
 
-Key = NamedTuple("Key", [("id", str), ("secret", str), ("user_name", str)])
 _PermissionCapableFactory: Type["MotoS3StorageFixtureFactory"] = None  # To be set later
 
 logging.getLogger("botocore").setLevel(logging.INFO)
 
+S3_CONFIG_PATH = os.path.expanduser(os.path.join("~", ".aws", "config"))
+S3_BACKUP_CONFIG_PATH = os.path.expanduser(os.path.join("~", ".aws", "bk_config"))
```
poodlewars, line 36: What's this for?

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -12,27 +12,35 @@
...
+
+
+class Key:
+    def __init__(self, id: str, secret: str, user_name: str):
```
poodlewars, line 40: Make this kwargs only? Its invocations with three strings next to each other are hard to understand

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -217,6 +243,241 @@ def real_s3_from_environment_variables(*, shared_path: bool):
...
 
 
+def real_s3_sts_from_environment_variables(user_name: str, role_name: str, policy_name: str, profile_name: str):
+    out = real_s3_from_environment_variables(False)
+    iam_client = boto3.client("iam", aws_access_key_id=out.default_key.id, aws_secret_access_key=out.default_key.secret)
+    # Create IAM user
+    try:
+        iam_client.create_user(UserName=user_name)
+        out.sts_test_key = Key(None, None, user_name)
+        print("User created successfully.")
```
poodlewars, line 253: logging not print statements...

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -217,6 +243,241 @@ def real_s3_from_environment_variables(*, shared_path: bool):
...
+        print(f"Error creating policy: {e}")
+        raise e
+
+    # Attach the policy to the role
+    try:
+        iam_client.attach_role_policy(
+            RoleName=role_name,
+            PolicyArn=out.aws_policy_name
+        )
+        print("Policyattached to role successfully.")
```
poodlewars, line 327: typo Policyattached

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -217,6 +243,241 @@ def real_s3_from_environment_variables(*, shared_path: bool):
...
+        config_file.write(aws_credentials)
+
+
+def real_s3_sts_resources_ready(factory: BaseS3StorageFixtureFactory): 
+    sts_client = boto3.client(
+        "sts",
+        aws_access_key_id=factory.sts_test_key.id,
+        aws_secret_access_key=factory.sts_test_key.secret
+    )
+    for _ in range(20):
```
poodlewars, line 400: 20 retries seems excessive?

poodlewars (changes requested):

python/arcticdb/storage_fixtures/s3.py
phoebusm: A temporary config file for sts credential and profile used in the test

python/arcticdb/storage_fixtures/s3.py
phoebusm: Once it requires 13 seconds to get the resources available in S3

docs/mkdocs/docs/api/arctic_uri.md
phoebusm: And it aligns with the boto3 API: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts/client/assume_role.html
which role is specified in the API, rather than the environment variable

python/arcticdb/storage_fixtures/s3.py
phoebusm: Three strings as parameters I don't see them as a lot. And current implementation gives better checking, IMO

poodlewars: That's fine but you can still require kwargs for these

cpp/arcticdb/storage/common.hpp
```diff
@@ -91,5 +95,13 @@ struct is_key_type<entity::VariantKey> : std::true_type {};
...
 inline constexpr bool is_key_type_v = is_key_type<T>::value;
 
+using NativeVariantStorage = s3::S3Settings;
+using NativeVariantStorageMap = std::map<std::string, NativeVariantStorage>; //key: storage_id
+using EnvironmentNativeVariantStorageMap = std::map<std::string, NativeVariantStorageMap>; // key: env
 }  //namespace arcticdb::storage
 
+// to avoid pybind11 built-in cast as it will copy the containers
+// Below needs to be not within any namespace and before any usage of those
+PYBIND11_MAKE_OPAQUE(arcticdb::storage::NativeVariantStorageMap)
```
poodlewars, line 105: We shouldn't use pybind in this layer of the code - this will make it harder to add new language bindings that don't depend on Python. I don't understand why a copy here is so bad.

cpp/arcticdb/storage/config_cache.hpp
```diff
@@ -92,7 +82,16 @@ class ConfigCache {
             if(storage_conf_pos != storage_configs_.end())
                 storage_conf = storage_conf_pos->second;
 
-            storages.emplace_back(create_storage(path, mode, storage_conf));
+            //it assume that the storage always exists
```
poodlewars, line 85: Comment does not make sense

cpp/arcticdb/storage/config_cache.hpp
```diff
@@ -111,18 +110,16 @@ class ConfigCache {
...
-            }
+        auto native_storages = config_resolver_->get_native_storages(environment_name_);
+        for(auto& [storage_name, config] : native_storages) {
+            native_storage_configs_.try_emplace(StorageName(storage_name), config);
         }
     }
 
     EnvironmentName environment_name_;
     std::unordered_map<LibraryPath, LibraryDescriptor> descriptor_map_;
     std::unordered_map<StorageName, arcticdb::proto::storage::VariantStorage> storage_configs_;
```
poodlewars, line 121: Why can't we replace `storage_configs_` with `native_storage_configs_`? Why do we need both? Can't we just change the type of `storage_configs_`?

cpp/arcticdb/storage/config_resolvers.hpp
```diff
@@ -40,10 +41,12 @@ namespace arcticdb::storage::details {
 class InMemoryConfigResolver final : public ConfigResolver {
   public:
     typedef std::unordered_map<StorageName, arcticdb::proto::storage::VariantStorage> StorageMap;
+    typedef std::unordered_map<StorageName, NativeVariantStorage> NativeStorageMap;
     typedef std::unordered_map<LibraryPath, arcticdb::proto::storage::LibraryDescriptor> LibraryMap;
 
     struct MemoryConfig {
         StorageMap storages_;
+        NativeStorageMap native_storages_;
```
poodlewars, line 49: Similar question, why can't we use the existing `StorageMap` for this?

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -95,21 +97,77 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
...
     storage.def("create_library_index", &create_library_index);
 
-    storage.def("create_mem_config_resolver", [](const py::object & env_config_map_py) -> std::shared_ptr<ConfigResolver> {
+    
+    py::enum_<s3::AWSAuthMethod>(storage, "AWSAuthMethod")
+        .value("DISABLED", s3::AWSAuthMethod::DISABLED)
+        .value("DEFAULT_CREDENTIALS_PROVIDER_CHAIN", s3::AWSAuthMethod::DEFAULT_CREDENTIALS_PROVIDER_CHAIN)
+        .value("STS_PROFILE_CREDENTIALS_PROVIDER", s3::AWSAuthMethod::STS_PROFILE_CREDENTIALS_PROVIDER);
+
+    auto s3settings_def = py::class_<s3::S3Settings>(storage, "S3Settings")
```
poodlewars, line 106: Why do we need the name `s3settings_def` for this? Can't we leave it anonymous like the others?

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -95,21 +97,77 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
...
+    py::class_<NativeVariantStorageMap>(storage, "NativeVariantStorageMap")
+        .def(py::init<>())
+        .def("__getitem__", [](NativeVariantStorageMap &storage_map, const std::string &storage_id)  -> NativeVariantStorage&  {
+            return storage_map[storage_id];
+        }, py::return_value_policy::reference_internal)
+        .def("__setitem__", [](NativeVariantStorageMap &storage_map, const std::string &storage_id, const NativeVariantStorage &storage) {
+            storage_map[storage_id] = storage;
+        })
+        .def("__contains__", [](const NativeVariantStorageMap &storage_map, const std::string &storage_id) {
+            return storage_map.count(storage_id) != 0;
```
poodlewars, line 121: contains?

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -95,21 +97,77 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
...
+    py::class_<EnvironmentNativeVariantStorageMap>(storage, "EnvironmentNativeVariantStorageMap")
+        .def(py::init<>())
+        .def("__getitem__", [](EnvironmentNativeVariantStorageMap &env_map, const std::string &env) -> NativeVariantStorageMap& {
+            return env_map[env];
+        }, py::return_value_policy::reference_internal)
+        .def("__setitem__", [](EnvironmentNativeVariantStorageMap &env_map, const std::string &env, const NativeVariantStorageMap &storage_map) {
+            env_map[env] = storage_map;
+        })
+        .def("__contains__", [](const EnvironmentNativeVariantStorageMap &env_map, const std::string &env) {
+            return env_map.count(env) != 0;
```
poodlewars, line 140: contains not count?

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -80,19 +82,23 @@ bool S3Storage::do_key_exists(const VariantKey& key) {
...
     region_(conf.region()) {
-
     auto creds = get_aws_credentials(conf);
 
     if (conf.use_mock_storage_for_testing()){
         ARCTICDB_RUNTIME_DEBUG(log::storage(), "Using Mock S3 storage");
         s3_client_ = std::make_unique<MockS3Client>();
     }
+    else if (conf.aws_auth() == AWSAuthMethod::STS_PROFILE_CREDENTIALS_PROVIDER){
+        Aws::Config::ReloadCachedConfigFile(); // config files loaded in Aws::InitAPI; It runs once at first S3Storage object construct; reload to get latest
```
poodlewars, line 98: Why is this reload here? I expected this to only be necessary within tests

cpp/arcticdb/storage/storage_factory.hpp
```diff
@@ -15,7 +15,10 @@
 
 namespace arcticdb {
     namespace storage {
-
+        std::shared_ptr<Storage> create_storage(
+                const LibraryPath &library_path,
+                OpenMode mode,
+                const NativeVariantStorage& storage_descriptor);
         std::shared_ptr<Storage> create_storage(
```
poodlewars, line 22: I don't think we should have overloads here. There should be a single `create_storage` method that takes a variant that's either the native object (for S3) or a protobuf (for the other storages). I would also be fine with just biting the bullet now and moving the other storages over to native configuration within the code too if you feel that would help - I'm wondering whether doing this incrementally is adding more complexity than it's worth.

cpp/arcticdb/storage/storage_override.hpp
```diff
@@ -5,6 +5,7 @@
 
 #include <arcticdb/entity/protobufs.hpp>
 #include <arcticdb/util/pb_util.hpp>
+#include <arcticdb/storage/s3/s3_storage.hpp>
```
poodlewars, line 8: Why?

cpp/proto/arcticc/pb2/storage.proto
```diff
@@ -12,6 +12,12 @@ package arcticc.pb2.storage_pb2;
 import "google/protobuf/any.proto";
 import "arcticc/pb2/utils.proto";
 
+enum AWSAuthMethod {
```
poodlewars, line 15: Why is this here? We had agreed there would be no changes to the proto definitions whatsoever?

docs/mkdocs/docs/api/arctic_uri.md
poodlewars: OK. The sentence `Only when `aws_auth` is set to be `2`.` is no longer correct

docs/mkdocs/docs/api/arctic_uri.md
```diff
@@ -22,11 +22,43 @@ Available options for S3:
...
-| aws_auth              | If true, authentication to endpoint will be computed via AWS environment vars/config files. If no options are provided `aws_auth` will be assumed to be true.   |
+| aws_auth              | AWS authentication method. If setting is `default` (or `true` for backward compatibility), authentication to endpoint will be computed via [AWS default credential provider chain](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/credproviders.html). If setting is `sts`, AWS Security Token Service (STS) will be the authentication method used. If no options are provided AWS authentication will not be used and you should specify access and secret in the URI. More info is provided below |
+| aws_profile           | Only when `aws_auth` is set to be `2`. AWS profile to be used with AWS Security Token Service (STS). More info is provided below |
 
 Note: When connecting to AWS, `region` can be automatically deduced from the endpoint if the given endpoint
 specifies the region and `region` is not set.
 
+### AWS Security Token Service (STS) setup
```
poodlewars, line 31: I think this section should be in a new storage guide like https://docs.arcticdb.io/latest/aws/ rather than in this page describing the URI, which we want to keep as brief as we can.

python/arcticdb/adapters/s3_library_adapter.py
```diff
@@ -127,10 +141,13 @@ def config_library(self):
...
             ca_cert_dir=self._ca_cert_dir,
             ssl=self._ssl,
+            aws_auth=self._query_params.aws_auth,
+            aws_profile=self._query_params.aws_profile,
+            native_cfg=self.native_cfg,
         )
 
         lib = NativeVersionStore.create_store_from_config(
-            env_cfg, _DEFAULT_ENV, CONFIG_LIBRARY_NAME, encoding_version=self._encoding_version
+            env_cfg, _DEFAULT_ENV, CONFIG_LIBRARY_NAME, encoding_version=self._encoding_version, native_cfg=self.native_cfg
```
poodlewars, line 150: Similar here, can we just change the type of `env_cfg` rather than adding the new `native_cfg` parameter? (in a backwards compatible way)

python/arcticdb/storage_fixtures/api.py
```diff
@@ -87,12 +89,17 @@ def _factory_impl(
...
+            cfg, native_cfg = cfgs
+        else:
+            cfg = cfgs
+            native_cfg = None
         lib = cfg.env_by_id[Defaults.ENV].lib_by_path[name]
         # Use symbol list by default (can still be overridden by kwargs)
         lib.version.symbol_list = True
         apply_lib_cfg(lib, kwargs)
-        out = ArcticMemoryConfig(cfg, Defaults.ENV)[name]
+        out = ArcticMemoryConfig(cfg, Defaults.ENV, native_cfg)[name]
```
poodlewars, line 102: ArcticMemoryConfig is used in Man, so need a default to avoid a breaking change

python/tests/integration/arcticdb/version_store/test_symbol_list.py
```diff
@@ -48,9 +48,9 @@ def small_max_delta():
         unset_config_int("SymbolList.MaxDelta")
 
 
-def make_read_only(lib):
-    return NativeVersionStore.create_store_from_lib_config(
-        lib.lib_cfg(), Defaults.ENV, OpenMode.READ
+def make_read_only(lib, native_config):
+    return NativeVersionStore.create_store_from_lib_and_static_config(
```
poodlewars, line 52: Breaking change to create_store_from_lib_config?

Only used in our code in Man though. Similar to other comments, I think the clean way to do this is to extend the types allowed in the first argument rather than allowing a new argument.

poodlewars (changes requested):

python/arcticdb/version_store/_store.py
```diff
@@ -29,7 +29,7 @@
 from arcticdb.supported_types import DateRangeInput, ExplicitlySupportedDates
 from arcticdb.toolbox.library_tool import LibraryTool
 from arcticdb.version_store.processing import QueryBuilder
-from arcticdb_ext.storage import OpenMode as _OpenMode
+from arcticdb_ext.storage import OpenMode as _OpenMode, EnvironmentNativeVariantStorageMap
```
poodlewars, line 32: Unused import?

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -43,23 +53,38 @@ class S3Bucket(StorageFixture):
...
     }
 
     key: Key
     _boto_bucket: Any = None
 
-    def __init__(self, factory: "BaseS3StorageFixtureFactory", bucket: str):
+    def __init__(self, 
+                 factory: "BaseS3StorageFixtureFactory", 
+                 bucket: str, 
+                 native_config: Optional[EnvironmentNativeVariantStorageMap] = EnvironmentNativeVariantStorageMap()
```
poodlewars, line 66: I don't understand `EnvironmentNativeVariantStorageMap`. Why is it a map?

cpp/arcticdb/storage/common.hpp
```diff
@@ -91,5 +95,13 @@ struct is_key_type<entity::VariantKey> : std::true_type {};
 template <typename T>
 inline constexpr bool is_key_type_v = is_key_type<T>::value;
 
+using NativeVariantStorage = s3::S3Settings;
+using NativeVariantStorageMap = std::map<std::string, NativeVariantStorage>; //key: storage_id
+using EnvironmentNativeVariantStorageMap = std::map<std::string, NativeVariantStorageMap>; // key: env
```
poodlewars, line 100: Like in the other comment, I don't get this. Why is this a map

cpp/arcticdb/storage/config_cache.hpp
poodlewars, line 1: I don't understand the changes to this file. Why are you caching an object that we already have in memory? Just add the native config to the `LibraryIndex#get_library` signature?

cpp/arcticdb/storage/config_resolvers.cpp
```diff
@@ -70,4 +82,9 @@ void InMemoryConfigResolver::add_storage(const EnvironmentName& environment_name
     config.storages_.try_emplace(StorageName(storage_name), storage);
 }
 
+void InMemoryConfigResolver::add_native_storage(const EnvironmentName& environment_name, const std::string& storage_name, const arcticdb::storage::NativeVariantStorage& storage) {
```
poodlewars, line 85: Don't understand why this is necessary

cpp/arcticdb/storage/s3/s3_storage.cpp
phoebusm: I think it's better to be reloaded everytime when `Arctic` is called for the sake of user-friendliness. Or user is needed to restarted the interpreter whenever there is a change of config (or called an API-pending to reload the config)

cpp/arcticdb/storage/storage_factory.hpp
phoebusm: Yes it is for doing the shift from proto to native config incrementally. Storage config other than `S3` is still untouched. 
Eventually the proto version of `create_storages` will be removed

cpp/proto/arcticc/pb2/storage.proto
phoebusm: It's a relic of previous implementation. It's been removed

docs/mkdocs/docs/api/arctic_uri.md
phoebusm: Will do

cpp/arcticdb/storage/config_cache.hpp
```diff
@@ -92,7 +82,15 @@ class ConfigCache {
...
 
-            storages.emplace_back(create_storage(path, mode, storage_conf));
+            if (auto it = native_storage_map.find(storage_name.value); 
+                storage_conf.config().Is<arcticdb::proto::s3_storage::Config>() && it != native_storage_map.end()) {
+                arcticdb::proto::s3_storage::Config s3_storage;
+                storage_conf.config().UnpackTo(&s3_storage);
+                storages.emplace_back(create_storage(path, mode, s3::S3Settings(it->second).load_from_proto(s3_storage)));
+            }
+            else {
+                storages.emplace_back(create_storage(path, mode, storage_conf));
```
poodlewars, line 92: (nice to have) Assertion that the native config variant is monostate

cpp/arcticdb/storage/s3/s3_settings.hpp
```diff
@@ -0,0 +1,233 @@
...
+    bool use_mock_storage_for_testing_;
+    std::string ca_cert_path_;
+    std::string ca_cert_dir_;
+    bool use_raw_prefix_;
+    AWSAuthMethod aws_auth_;
+    std::string aws_profile_;
+
+public:
+    S3Settings() = default;
+    S3Settings(const arcticc::pb2::s3_storage_pb2::Config& config){
```
poodlewars, line 44: Is it possible to require this to be created with the protobuf and the native object at the same time and make this object immutable, so that the type system prevents it from ever being in an invalid state?

cpp/arcticdb/storage/s3/s3_settings.hpp
```diff
@@ -0,0 +1,233 @@
...
+    bool use_virtual_addressing_;
+    bool use_mock_storage_for_testing_;
+    std::string ca_cert_path_;
+    std::string ca_cert_dir_;
+    bool use_raw_prefix_;
+    AWSAuthMethod aws_auth_;
+    std::string aws_profile_;
+
+public:
+    S3Settings() = default;
```
poodlewars, line 43: Why do we need this?

python/arcticdb/version_store/helper.py
```diff
@@ -272,10 +292,22 @@ def get_s3_proto(
...
+    
+    if aws_auth is not None or aws_profile is not None:
+        if is_nfs_layout:
+            raise ArcticNativeException("aws_auth and aws_profile can only be set for S3")
+        if native_cfg is None:
+            raise ArcticNativeException("Missing native cfg")
+        native_s3_settings = NativeS3Settings()
+        if aws_auth is not None:
+            native_s3_settings.aws_auth = AWSAuthMethod(aws_auth)
+        if aws_profile is not None:
```
poodlewars, line 304: Validate against the case where `aws_auth=sts` and `aws_profile` is None

poodlewars (commented):
Happy other than the comments I've made.

It would be good to test that the STS auth works correctly with replication.

cpp/arcticdb/storage/s3/s3_settings.hpp
phoebusm: I have made the class immutable on the python layer but still required to be mutable on the C++ layer
For example, [here](https://github.com/man-group/ArcticDB/blob/1f936758ef0facc7abe859e37c569e3001e16c5a/cpp/arcticdb/storage/library_manager.cpp#L219), the new native settings object being passed not along with the existing protobuf object. The config read from storage could be different and should have priority over the native settings object.

Therefore it is needed to be mutable on C++ layer during this awkward transtion period, until protobuf settings is phased out, except serlization/derserilization to/from storage.

cpp/arcticdb/storage/common.hpp
```diff
@@ -13,8 +13,12 @@
 #include <arcticdb/entity/protobufs.hpp>
 #include <arcticdb/entity/variant_key.hpp>
 #include <arcticdb/entity/performance_tracing.hpp>
+#include <arcticdb/storage/s3/s3_settings.hpp>
 #include <sstream>
 
+#include <pybind11/pybind11.h>
```
poodlewars, line 19: Remove the pybind includes

cpp/arcticdb/storage/config_cache.hpp
poodlewars: This hasn't been done?

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -94,6 +96,36 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
...
+    };
+
+    py::class_<s3::S3Settings>(storage, "S3Settings")
+        .def(py::init<s3::AWSAuthMethod, const std::string&>())
+        .def(py::pickle(
+            [](const s3::S3Settings &settings) {
+                return py::make_tuple(settings.aws_auth(), settings.aws_profile());
+            },
+            [](py::tuple t) {
+                if (t.size() != 2) {
```
poodlewars, line 117: `util::check` that size is 2?

cpp/arcticdb/storage/s3/s3_settings.hpp
```diff
@@ -0,0 +1,142 @@
...
+    bool use_virtual_addressing_;
+    bool use_mock_storage_for_testing_;
+    std::string ca_cert_path_;
+    std::string ca_cert_dir_;
+    bool use_raw_prefix_;
+    AWSAuthMethod aws_auth_;
+    std::string aws_profile_;
+
+public:
+    S3Settings(AWSAuthMethod aws_auth, const std::string& aws_profile) :
```
poodlewars, line 43: explicit?

docs/mkdocs/docs/aws.md
```diff
@@ -79,3 +79,39 @@ s3.get_object(Bucket=bucket, Key='_arctic_check/check.txt')
...
+
+[profile BASE_PROFILE]
+aws_access_key_id = ID
+aws_secret_access_key = KEY
+```
+
+Use the configuration in ArcticDB:
+```
+>>> import arcticdb as adb
+>>> arctic = adb.Arctic('s3://s3.REGION.amazonaws.com:BUCKET?aws_auth=2&aws_profile=PROFILE')
```
poodlewars, line 110: `aws_auth=2` is the obsolete way to set this up?

docs/mkdocs/docs/aws.md
```diff
@@ -79,3 +79,39 @@ s3.get_object(Bucket=bucket, Key='_arctic_check/check.txt')
...
+```
+
+Due to the limitation of AWS C++ SDK, if ArcticDB fails to get the temporary token by assuming role, assertion error like below will happen:
```
poodlewars, line 113: Is there really no way for us to give a better error in this case?

docs/mkdocs/docs/aws.md
```diff
@@ -79,3 +79,39 @@ s3.get_object(Bucket=bucket, Key='_arctic_check/check.txt')
...
+```
+virtual void Aws::Auth::STSProfileCredentialsProvider::Reload(): Assertion `!profileIt->second.GetCredentials().IsEmpty()' failed.
+```
+It's usually due to incorrect iam account setup or incorrect config file given
```
poodlewars, line 117: The English of this sentence is incorrect

python/arcticdb/adapters/s3_library_adapter.py
```diff
@@ -244,3 +278,13 @@ def _configure_aws(self):
     @property
     def path_prefix(self):
         return self._query_params.path_prefix
+
+    
+    def copy_config_library_native_config_to_library(self, lib_cfg: LibraryDescriptor):
```
poodlewars, line 283: I think this is a sign that something is a bit wrong with the design. We only support a single native configuration per Arctic object, so why are we storing them in the `_native_cfg` map as if we support many of them? I still don't really see why we need the `_native_cfg` map at all.

python/arcticdb/storage_fixtures/s3.py
poodlewars: This hasn't been addressed

python/arcticdb/version_store/helper.py
```diff
@@ -272,10 +292,20 @@ def get_s3_proto(
         s3.ca_cert_dir = ca_cert_dir
     if ssl is not None:
         s3.ssl = ssl
+    
+    if aws_auth == AWSAuthMethod.STS_PROFILE_CREDENTIALS_PROVIDER or aws_profile:
+        if is_nfs_layout:
+            raise ArcticNativeException("aws_auth and aws_profile can only be set for S3")
```
poodlewars, line 298: These should be user input errors https://docs.arcticdb.io/latest/error_messages/#user-input-errors

python/tests/conftest.py
```diff
@@ -104,6 +107,11 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize("encoding_version", [EncodingVersion.V1] if only_v1 else list(EncodingVersion))
 
 
+@pytest.fixture(scope="session")
```
poodlewars, line 110: Why is this session scoped? Could lead to horrible problems if a test mutates it

docs/mkdocs/docs/aws.md
phoebusm: Miss it again...

docs/mkdocs/docs/aws.md
phoebusm: Unfortunately no: https://github.com/aws/aws-sdk-cpp/blob/b5153b8309526738d498ee86a73ccc386fee0979/src/aws-cpp-sdk-identity-management/source/auth/STSProfileCredentialsProvider.cpp#L305

cpp/arcticdb/storage/config_cache.hpp
phoebusm: `NativeVariantStorage` is not a variant but an alias to `S3Settings` for now. It was named to be "variant" so later addition of other storages' native config could be easier

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -114,7 +114,7 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
                 return py::make_tuple(settings.aws_auth(), settings.aws_profile());
             },
             [](py::tuple t) {
-                util::raise_rte(t.size() != 2, "Invalid S3Settings pickle objects");
+                util::check(t.size() != 2, "Invalid S3Settings pickle objects");
```
poodlewars, line 117: Now this is the wrong way round, we assert the state that we expect

cpp/arcticdb/storage/common.hpp
```diff
@@ -91,5 +92,5 @@ struct is_key_type<entity::VariantKey> : std::true_type {};
 template <typename T>
 inline constexpr bool is_key_type_v = is_key_type<T>::value;
 
-}  //namespace arcticdb::storage
-
+using NativeVariantStorage = s3::S3Settings;
```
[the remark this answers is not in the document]
phoebusm: `NativeVariantStorage` is just a breadcrumb that left for myself of where to pick up from. 
`std::variant` is definitely a better one but I prefer not spending on some provisional design atm, as I haven't really made up my mind about how to add the native config for the rest of the storage
Though this is probably the way

cpp/arcticdb/storage/common.hpp
```diff
@@ -91,5 +92,17 @@ struct is_key_type<entity::VariantKey> : std::true_type {};
 template <typename T>
 inline constexpr bool is_key_type_v = is_key_type<T>::value;
 
-}  //namespace arcticdb::storage
-
+class NativeVariantStorage {
+public:
+    using VariantStorageConfig = std::variant<std::monostate, s3::S3Settings>;
+    explicit NativeVariantStorage(VariantStorageConfig config) : config_(std::move(config)) {};
```
poodlewars, line 98: Might be slightly nicer to have a zero argument constructor that sets the `std::monostate` and then constructors for each other member of the variant type.

cpp/arcticdb/storage/storage_factory.hpp
```diff
@@ -18,7 +18,7 @@ namespace arcticdb {
...
                 const LibraryPath &library_path,
                 OpenMode mode,
-                const NativeVariantStorage& storage_descriptor);
+                const s3::S3Settings& storage_descriptor);
```
poodlewars, line 21: This seems wrong? This could create any type of storage, not just S3?

phoebusm: I have consolidated the rest of storages setup calling below function at line 22.

cpp/arcticdb/storage/common.hpp
phoebusm: Not a huge fan of default paramter but maybe it's a good use case here to conslidate the constructor. It should be just a very thin wrapper of the variant so can't forsee the default parameter will be a nightmare in the future

poodlewars (approved):
Approved! Please make sure to squash the commits when you merge

PR #1915, opened 2024-10-14
Make library manager an LRU cache

willdealtry:
We don't want to cache an infinite number of libraries, so make the map of libraries into an LRU cache


poodlewars:
The diff is broken - it shows 0 files changed.

willdealtry:
> The diff is broken - it shows 0 files changed.

Weird, the rebase seems to have erased the change entirely. And yet the builds don't pass ;-/

Will fix

PR #1928, opened 2024-10-16
Harden against object lifetime on exceptions

willdealtry:
Throw from pipeline and storages to check resiliency

### DIFF
--- a/cpp/arcticdb/async/tasks.hpp
+++ b/cpp/arcticdb/async/tasks.hpp
@@ -25,6 +25,7 @@
 #include <arcticdb/processing/processing_unit.hpp>
 #include <arcticdb/util/constructors.hpp>
 #include <arcticdb/codec/codec.hpp>
+#include <arcticdb/util/test/random_throw.hpp>
 
 #include <type_traits>
 #include <ranges>
@@ -90,6 +91,7 @@ struct EncodeAtomTask : BaseTask {
 
     storage::KeySegmentPair encode() {
         ARCTICDB_DEBUG(log::codec(), "Encoding object with partial key {}", partial_key_);
+        ARCTICDB_DEBUG_THROW(5)
         auto enc_seg = ::arcticdb::encode_dispatch(std::move(segment_), *codec_meta_, encoding_version_);
         auto content_hash = get_segment_hash(enc_seg);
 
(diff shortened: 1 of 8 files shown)


cpp/arcticdb/async/tasks.hpp
```diff
@@ -507,6 +510,7 @@ struct MemSegmentFunctionTask : BaseTask {
     ARCTICDB_MOVE_ONLY_DEFAULT(MemSegmentFunctionTask)
 
     folly::Unit operator()(std::pair<VariantKey, SegmentInMemory> &&seg_pair) {
+        ARCTICDB_DEBUG_THROW(5)
```
alexowens90, line 513: This task is unused, assume this should be in `MemSegmentProcessingTask`

cpp/arcticdb/processing/clause.cpp
```diff
@@ -100,6 +100,7 @@ std::vector<EntityId> PassthroughClause::process(std::vector<EntityId>&& entity_
 }
 
 std::vector<EntityId> FilterClause::process(std::vector<EntityId>&& entity_ids) const {
+    ARCTICDB_DEBUG_THROW(5)
```
alexowens90, line 103: Could just stick this in the base class to avoid the duplication
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/processing/clause.hpp#L59-L61
And could put one in the `structure_for_processing` overload that takes `entity_ids_vec` as well for good measure.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1435,10 +1435,10 @@ VersionedItem collate_and_write(
     });
 }
 
-void delete_incomplete_keys(PipelineContext* pipeline_context, Store* store) {
+void delete_incomplete_keys(PipelineContext& pipeline_context, Store& store) {
```
alexowens90, line 1438: Guessing `PipelineContext` can be const&

alexowens90 (changes requested):

cpp/arcticdb/version/version_core.cpp
willdealtry: A rational person would think so, but because of irritating reasons the function returning the iterator into the pipeline context can't be made easily const, and we want this class to just go away so I don't want to spend too much time (any time) polishing it.

alexowens90:
Superseded by https://github.com/man-group/ArcticDB/pull/2037

PR #1930, opened 2024-10-17
Make library manager an LRU cache

willdealtry:
At the moment as we open more and more libraries we never close them, meaning that we can run out of filehandles.

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -344,6 +344,7 @@ set(arcticdb_srcs
         util/spinlock.hpp
         util/key_utils.hpp
         util/lock_table.hpp
+        util/lru_cache.hpp
         util/magic_num.hpp
         util/movable_priority_queue.hpp
         util/movable_priority_queue.hpp
@@ -508,7 +509,7 @@ set(arcticdb_srcs
         version/symbol_list.cpp
         version/version_map_batch_methods.cpp
         storage/s3/ec2_utils.cpp
-        util/buffer_holder.cpp)
+        util/buffer_holder.cpp util/test/rapidcheck_lru_cache.cpp)
 
 add_library(arcticdb_core_object OBJECT ${arcticdb_srcs})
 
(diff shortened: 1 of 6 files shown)


poodlewars:
Will this help with the file descriptor exhaustion problem? Given https://github.com/man-group/ArcticDB/pull/1908 aren't we going to hold them open regardless of this change?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -1061,7 +1062,7 @@ if(${TEST})
             util/test/rapidcheck_string_pool.cpp
             util/test/rapidcheck_main.cpp
             version/test/rapidcheck_version_map.cpp
-    )
+            util/test/rapidcheck_lru_cache.cpp)
```
alexowens90, line 1065: Remove from other targets

cpp/arcticdb/util/lru_cache.hpp
```diff
@@ -0,0 +1,82 @@
...
+            return std::nullopt;
+        }
+
+        ARCTICDB_DEBUG(log::inmem(), "Key {} found", key);
+        list_.splice(list_.begin(), list_, it->second);
+        return it->second->value;
+    }
+
+    void remove(const KeyType& key) {
+        std::unique_lock lock(mutex_);
```
alexowens90, line 53: Why `unique_lock` over `lock_guard`?

cpp/arcticdb/util/lru_cache.hpp
```diff
@@ -0,0 +1,82 @@
...
+        ARCTICDB_DEBUG(log::inmem(), "Removing key {}", key);
+        auto it = cache_.find(key);
+        if (it == cache_.end())
+            return;
+
+        list_.erase(it->second);
+        cache_.erase(it);
+    }
+
+    void put(KeyType key, ValueType value) {
```
alexowens90, line 63: nit: Could pass both args by `const&`, and then only incur the copy cost of `key` in the `else` branch below

alexowens90 (changes requested):

cpp/arcticdb/util/lru_cache.hpp
willdealtry: It's a std::shared_mutex

alexowens90 (approved):

PR #1995, opened 2024-11-12
new bi sample benchmark tests

grusev:

NEW PR CREATED : #2019

        Sample test benchmark for using one opensource BI CSV source.
        The logic of a test is 
            - download if parquet file does not exists source in .bz2 format
            - convert it to parquet format
            - prepare library with it containing  several symbols that are constructed based on this DF
            - for each query we want to benchmark do a pre-check that this query produces SAME result on Pandas and arcticDB
            - run the benchmark tests

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/data/CityMaxCapita_1.parquet.gzip
G-D-Petrov, line 1: We will need to setup [git lfs](https://docs.github.com/en/repositories/working-with-files/managing-large-files/configuring-git-large-file-storage?platform=linux) for this file before merging the PR

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+    def get_query_groupaby_city_count_all(self, q):
+        return q.groupby("City").agg({"Keyword": "count"})
+               
+    def time_query_groupaby_city_count_all(self, times_bigger) -> pd.DataFrame:
+        q = QueryBuilder()
+        q = self.get_query_groupaby_city_count_all( q)
+        df = self.lib.read(f"{self.symbol}{times_bigger}", query_builder=q)
+        return df.data
+
+    def get_query_groupaby_city_count_isin_filter(self, q):
```
G-D-Petrov, line 127: Typo: should be `groupby`
Applies to all the methods with the same typo

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+        self.ac = Arctic(f"lmdb://opensource_datasets_{self.lib_name}?map_size=20GB")
+        self.lib = self.ac.get_library(self.lib_name)
+
+    def teardown(self, num_rows):
+        del self.ac
+
+    def time_query_readall(self, times_bigger):
+        self.lib.read(f"{self.symbol}{times_bigger}")
+
+    def get_query_groupaby_city_count_all(self, q):
```
G-D-Petrov, line 118: nit: looks like there is no need for these to be methods of the class, can be defined outside the class which will make it a bit more readable

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+        arctic_df = self.time_query_groupaby_city_count_isin_filter(BIBenchmarks.params[0])
+        _df = self.get_query_groupaby_city_count_isin_filter(_df)
+        arctic_df.sort_index(inplace=True)
+        self.assert_frame_equal(_df, arctic_df)
+
+        print("Pre-check correctness for query_groupaby_city_count_filter_two_aggregations")
+        _df = self.df.copy(deep=True)
+        arctic_df = self.time_query_groupaby_city_count_filter_two_aggregations(BIBenchmarks.params[0])
+        _df = self.get_query_groupaby_city_count_filter_two_aggregations(_df)
+        arctic_df.sort_index(inplace=True)
```
G-D-Petrov, line 94: Looks like the sorting of the index is already done in the `assert_frame_equal` function, so it is not needed here.

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+        self.assert_frame_equal(_df, arctic_df)
+
+        print("All pre-checks completed SUCCESSFULLY")
+
+        del self.ac
+    
+    def assert_frame_equal(self, pandas_df:pd.DataFrame, arctic_df:pd.DataFrame):
```
G-D-Petrov, line 101: nit: Looks like this doesn't need to be a method of the class

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+    def get_query_groupaby_city_count_filter_two_aggregations(self, q):
+        return q[q["Keyword"] == "faggot" ].groupby("City").agg({"Keyword": "count", "Number of Records" : "sum"})  
+    
+    def time_query_groupaby_city_count_filter_two_aggregations(self, times_bigger) -> pd.DataFrame:
+        q = QueryBuilder()
+        q = self.get_query_groupaby_city_count_filter_two_aggregations(q) 
+        df = self.lib.read(f"{self.symbol}{times_bigger}", query_builder=q)
+        return df.data
+
+    def peakmem_query_groupaby_city_count_filter_two_aggregations(self, times_bigger):
```
G-D-Petrov, line 145: Might be a good idea to add a peakmem_... for the other time_... variants as well

G-D-Petrov (changes requested):
Not necessary for this PR, but it might be nice to do these benchmarks also for the `read_batch` method.
But this will probably be more reasonable if we add more of these data sets in the future.

python/benchmarks/bi_benchmarks.py
grusev: done!

python/benchmarks/bi_benchmarks.py
grusev: done!

python/benchmarks/bi_benchmarks.py
grusev: ìndeed! late refactoring omission

python/benchmarks/bi_benchmarks.py
grusev: I was considing this also. I will add them later we can remove them easily

grusev: Now I added also to print info for large dataframes that are produced by this test. So our first test is with 1 GB DF while last is with 10 GB df

             OUTPUT -------->
             Parquet file exists!
             The procedure is creating N times larger dataframes
             by concatenating original DF N times
             DF for iterration xSize original ready:  10
             <class 'pandas.core.frame.DataFrame'>
             Index: 9126570 entries, 0 to 912656
             Data columns (total 31 columns):
              #   Column               Dtype
             ---  ------               -----
              0   City/Admin           object
              1   City/State           object
              2   City                 object
              3   Created Date/Time    float64
              4   Date Joined          float64
              5   FF Ratio             float64
              6   Favorites            int32
              7   First Link in Tweet  object
              8   Followers            int32
              9   Following            int32
              10  Gender               object
              11  Influencer?          int32
              12  Keyword              object
              13  LPF                  float64
              14  Language             object
              15  Lat                  float64
              16  Listed Number        int32
              17  Long Domain          object
              18  Long                 float64
              19  Number of Records    int32
              20  Region               object
              21  Short Domain         object
              22  State/Country        object
              23  State                object
              24  Tweet Text           object
              25  Tweets               int32
              26  Twitter Client       object
              27  User Bio             object
              28  User Loc             object
              29  Username 1           object
              30  Username             object
             dtypes: float64(6), int32(7), object(18)
             memory usage: 10.8 GB

grusev:
New PR was created to resolve issues #2019

PR #2001, opened 2024-11-14
first version of tests that reproduce mem leak

grusev:
#### Reference Issues/PRs

The tests contain logic that is able to catch large memory leaks by repeating certain operation N times and cleaning mem through garbage collection each time. As the mem leaks is iterrative process the success of approach will be determined by how large the mem leak is. For small mem leaks this process will not be able to capture them unless repeated many many times, which could be very costly in terms of time.

It is advised that when approach is taken and new test is prepared the check procedure parameters to be finetuned for that particular case.. To do so you need to execute the process from command line with "-s" option so that pytest does not hide the output of print statements. See more in the doc of the procedure..

The output that you should expect from precedure should be like:

tart check process for memory leaks
Num iterrations:  30
Maxumum memory growth/lost to the examined process:  4808662528
Maxumum machine memory utilization:  4808662528
Lets pause for 10 secs so GC to kick in
Available memory  19730.07MB/[20688478208] 
--------------------------------------------------------------------------------
Process initial RSS  8671.72MB/[9092956160] 
Starting watched code ...........
Lets pause for 7 secs so GC to kick in
Iter No[0]    : Process did added (or if - number means cleaned)  1014.14MB/[1063407616] .  Avail memory:  18793.03MB/[19705921536]  Used Mem: 61.0%
  Overall stats : Process growth since start  1014.14MB/[1063407616]  AVG growth per iter  1014.14MB/[1063407616.0] 
  Minimum growth so far:  1014.14MB/[1063407616] 
  Maximum growth so far:  1014.14MB/[1063407616] 
  Number of times there was 50% drop in memory: 0
Starting watched code ...........
Lets pause for 7 secs so GC to kick in
Iter No[1]    : Process did added (or if - number means cleaned)  990.93MB/[1039065088] .  Avail memory:  17372.98MB/[18216894464]  Used Mem: 63.9%
  Overall stats : Process growth since start  2005.07MB/[2102472704]  AVG growth per iter  1002.54MB/[1051236352.0] 
  Minimum growth so far:  1014.14MB/[1063407616] 
  Maximum growth so far:  2005.07MB/[2102472704] 
  Number of times there was 50% drop in memory: 0
  
  ..........................................................

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def proc_to_examine():
+        data = lib.read(symbol).data
+        del data
+
+    df = lib.read(symbol).data
+    print(df)
+    del df
+
+    check_process_memory_leaks(proc_to_examine, 30, 4808662528, 80.0)
```
G-D-Petrov, line 157: I would make 4808662528 into a variable and break down how it was calculated so it is easier to see the value that is tests.
Same goes for 808662528 below.

G-D-Petrov (changes requested):

poodlewars:
I think that a tool like this will make these tests properly automated https://github.com/bloomberg/memray

python/tests/conftest.py
```diff
@@ -304,11 +304,25 @@ def arctic_client_no_lmdb(request, encoding_version):
     assert not ac.list_libraries()
     return ac
 
+@pytest.fixture(
+    scope="function",
+    params=[
+        "lmdb"
+    ],
+)
+def arctic_client_lmdb(request, encoding_version):
```
poodlewars, line 313: Don't parameterize this fixture, just pass in a `lmdb_storage` rather than the `request`

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+import pandas as pd
+import pytest
+import arcticdb as adb
+from pympler import muppy, summary
+
+from arcticdb.util.test import get_sample_dataframe
+from arcticdb.version_store._store import NativeVersionStore
+
+
+def get_process_by_id(id: int | None) -> psutil.Process:
```
poodlewars, line 23: We don't need this, just do
```
process = psutil.Process()
```

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+def nice_bytes_str(bytes) :
+   return f" {bytes / (1024 * 1024):.2f}MB/[{bytes}] "
+
+def lets_collect_some_garbage(time_sec:int = 7):
+    """
+       Do a garbage collection
+    """
+    gc.collect()
+    print(f"Lets pause for {time_sec} secs so GC to kick in") 
+    time.sleep(time_sec)
```
poodlewars, line 41: Why do you need a sleep?

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+        Quickly inflates the dataframe by doubling its size for each iteration
+        Thus the final size will be 'num_times_xx2' power of two
+    """
+    for n in range(1,num_times_xx2):
+        df_prev = df_to_grow.copy(deep=True)
+        df_to_grow = pd.concat([df_to_grow, df_prev])
+    return df_to_grow
+
+@pytest.mark.xfail(reason = "ArcticDBPR#1998 String memory leaks while read")
```
poodlewars, line 139: PR 1998 has been merged, so we shouldn't xfail

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    count = 0 
+    for i in range(1, len(numbers)): 
+        if numbers[i] < percentage_drop_0_1 * numbers[i - 1]: count += 1 
+    return count    
+
+
+def check_process_memory_leaks(process_func , number_iterations, max_total_mem_lost_treshold, max_machine_memory_percentage) -> np.int64:
+    """
+        This check accepts a function which will be called iterativly 'number_iterations'. During this
```
poodlewars, line 59: typo iterativly

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+        utilization of memory is not above certain percentage 'max_machine_memory_percentage' to prevent 
+        crash of the process
+
+        Specify more than 3 iterations (advised min 30), and be sure that there will be some initial growth in the first 1-2 
+        iterations, thus your expected growth threshold should be above that
+
+        IMPORTANT: It is strongly advised during creation of future tests to they are executed first 
+        through command line to determine the optimal iteration  and 'max_total_mem_lost_treshold' for the
+        python process. During the execution you will see lots of information on the screen including maximum growth
+        of memory since iterations started as weel as minumum growth of memory. That along with information
```
poodlewars, line 71: typo `weel`

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+        how many times the growth of mem droped 50% could provide insight for presence or lack of memory issues
+
+        NOTE: This test is NOT CONCLUSIVE for absence of memory leacks. It would catch LARGE leaks.
```
poodlewars, line 74: typo `leacks`

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+        of memory since iterations started as weel as minumum growth of memory. That along with information
+        how many times the growth of mem droped 50% could provide insight for presence or lack of memory issues
+
+        NOTE: This test is NOT CONCLUSIVE for absence of memory leacks. It would catch LARGE leaks. 
+        Small leaks will not be caught. If the procedure fails to sport problems it will return 2 numbers
+        for further investigation by the test:
+         - how much the process has grown during execution in bytes
+         - how many times the process memory dropped more than 50% (indicating some C++ garbage collection)
+
+        NOTE 2: Currently this code controlls python GC only.
```
poodlewars, line 80: typo `controlls`

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+    print ("-" * 80)
+    print ("Start check process for memory leaks")
+    print ("Num iterations: ", number_iterations)
+    print ("Maxumum memory growth/lost to the examined process: ", max_total_mem_lost_treshold)
+    print ("Maxumum machine memory utilization: ", max_total_mem_lost_treshold)
+    
+    lets_collect_some_garbage(10)
+
+    mem_per : np.int64 = psutil.virtual_memory().percent
+    assert max_machine_memory_percentage >- mem_per , "Machine has less memory available the check cannot start"
```
poodlewars, line 94: typo? `>-` ?

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+        p_iter_mem_start : np.int64  = p.memory_info().rss
+        
+        print("Starting watched code ...........")
+        process_func()
+        lets_collect_some_garbage()
+
+        p_iter_mem_end : np.int64  = p.memory_info().rss
+        process_growth : np.int64 = p.memory_info().rss - process_initial_memory
+        mem_growth_each_iter.append(process_growth)
+        process__avg_growth = process_growth / (n + 1)
```
poodlewars, line 112: `process__avg_growth` should be `process_avg_growth`

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+        p_iter_mem_notcleaned : np.int64  = p_iter_mem_end - p_iter_mem_start
+        mem_avail : np.int64 = psutil.virtual_memory().available
+        mem_per : np.int64 = psutil.virtual_memory().percent
+        print (f"Iter No[{n}]    : Process did added (or if - number means cleaned) {nice_bytes_str(p_iter_mem_notcleaned)}.  Avail memory: {nice_bytes_str(mem_avail)} Used Mem: {mem_per}%")
```
poodlewars, line 116: typo `Process did added`?

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+        print (f"  Overall stats : Process growth since start {nice_bytes_str(process_growth)} AVG growth per iter {nice_bytes_str(process__avg_growth)}")
+        print (f"  Minimum growth so far: {nice_bytes_str(min(mem_growth_each_iter))}")
+        print (f"  Maximum growth so far: {nice_bytes_str(max(mem_growth_each_iter))}")
+        print (f"  Number of times there was 50% drop in memory: {count_drops(mem_growth_each_iter, 0.5)}")
```
poodlewars, line 120: How do you think people will use the stats about the number of drops?

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+
+        assert max_total_mem_lost_treshold >= process_growth , f"Memory of the process grew more than defined threshold: {nice_bytes_str(process_growth)} (specified: {nice_bytes_str(max_total_mem_lost_treshold)} )"
+        assert max_machine_memory_percentage >= mem_per , f"Machine utilized more memory than specified treshold :{mem_per}% (specified {max_machine_memory_percentage}%)"
+        
+    print ("The process assesment finished within expectations. Total consumed additionaml mem is bellow threshold: ", process_growth)
```
poodlewars, line 125: typos assesment additionaml

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+
+    symbol = "test"
+    lib.write(symbol, df)
+
+    def proc_to_examine():
+        data = lib.read(symbol).data
+        del data
+
+    df = lib.read(symbol).data
+    print(df)
```
poodlewars, line 154: This doesn't print all 32M rows hopefully?

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+    del df
+
+    '''
+        To determine optimal max_mem first set the procedure to a very high number like 10 GB
```
poodlewars, line 158: These docs probably make more sense in `check_process_memory_leaks`

python/tests/stress/arcticdb/test_mem_leaks.py
grusev: yes. gc.collect() seems to be asynchronous - in other words the more you wait the better is the chance to go as low as possible. That is the reason I needed to introduce waiting. Without it sometimes almost nothing is collected. And yes it will add some latency to the test for sure ...

Memray tells similar story

https://pytest-memray.readthedocs.io/en/latest/usage.html
.....
Warning

As the Python interpreter has its own [object allocator](https://docs.python.org/3/c-api/memory.html) it’s possible that memory is not immediately released to the system when objects are deleted, so tests using this marker may need to give some room to account for this.

python/tests/stress/arcticdb/test_mem_leaks.py
grusev: great! will remove it!

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+
+        NOTE 2: Currently this code controlls python GC only.
+
+    """
+    p = get_process_by_id(os.getpid())
+    avail_mem_start : np.int64 = psutil.virtual_memory().available
+    print ("-" * 80)
+    print ("Start check process for memory leaks")
+    print ("Num iterations: ", number_iterations)
+    print ("Maxumum memory growth/lost to the examined process: ", max_total_mem_lost_treshold)
```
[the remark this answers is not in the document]
grusev: btw, I have installed vscode spellcheck so my comments will improve in future!

python/tests/stress/arcticdb/test_mem_leaks.py
grusev: The number of drops means there is garbage collection going on. Thus you might want to extend with more repeats if garbage collection goes but it is not enough ... Potentially this metric is here to get you oriented what is going on during execution. Thus finetuning of this is important by running pytest with -s and only this test from command line. I will create also a wiki page with advises if needed

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,201 @@
...
+
+         if both conditions are met then the process is likely to not leak big chunks of mem
+         and you can take the max mem growth number add 20 percent safety margin and you will
+         arrive at reasonable  max_mem
+
+         Then put for number of iterations 3x or 5x and this mem limit and 
+         run the test from command line again to assure it runs ok before commit 
+
+    '''
+    max_mem = 4808662528
```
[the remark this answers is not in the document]
grusev: this is bytes ...

Now scaling down is indeed possible, and perhaps I will do that now when things are fixed. Before I wanted to give the process maximum possible iterrations (7) on my machine to survive

python/tests/stress/arcticdb/test_mem_leaks.py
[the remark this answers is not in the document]
grusev: they will run few minutes only. The point with those tests is that they need to execute both on our machines and on runners. Thus it will be always subjective how much time to give them. If they atart to take too much time we can move them in new jobs with special mark

python/tests/stress/arcticdb/test_mem_leaks.py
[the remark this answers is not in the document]
poodlewars: `153623040` is still about 153GB isn't it? Surely we don't expect this to use so much memory?

poodlewars:
Please could you rebase this on to latest master to drop the commits from other people?

python/tests/stress/arcticdb/test_mem_leaks.py
```diff
@@ -0,0 +1,190 @@
...
+        del data
+
+    df = lib.read(symbol).data
+    print(df)
+    del df
+
+    """ 
+        See comment in previous test
+    """
+    max_mem_bytes = 608662528
```
poodlewars, line 188: Surely we can have a tighter threshold than 600GB!

python/tests/stress/arcticdb/test_mem_leaks.py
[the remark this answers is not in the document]
grusev: I have started the execution couple of times it was 116MB the first the the max and less than 100MB second execution This means that we cannot set fix number, therefore I assigned a litle bit of safety margin. As we execute it many times the idea is that it would stay lower that number. If we try to optimize it I think we could end in a situation where the test becaomes flaky. Sometimes it will go above that number and when you run it all will be well the next time. Thus I recommend staying that way

python/tests/stress/arcticdb/test_mem_leaks.py
grusev: one execution was 430 MB another little over 100MB. Not sure It is good idea to optimized more see my previous comment

grusev: So it is 600MB the limit here....

python/tests/stress/arcticdb/test_mem_leaks.py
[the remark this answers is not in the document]
poodlewars: I did the units conversion wrong. This looks good.

python/tests/stress/arcticdb/test_mem_leaks.py
poodlewars: Maths fail from me :)

PR #2013, opened 2024-11-22
7898685792 Finalize staged data memory use improvements

poodlewars:
Discussion: https://chat-man.slack.com/archives/C02NEG3V38X/p1732181354814409

Testing in https://github.com/man-group/ArcticDB/commit/0cc7090bfe0fac4aaf0ad232f95f14f7b0d9d989 shows that with Georgi's repro script with 1000 incomplete segments, the peak memory use went from 3200MB to 1200MB.

Running the repro script with 18_000 incompletes, the peak memory now stays at 1200GB. Previously, running this would OOM crash on a machine with 64GB of memory.

Future work:

- Backpressure - https://github.com/man-group/ArcticDB/pull/2013/files#diff-07407837d69849c0f3084af0394134f44646daae2e34ba042122d7bdc8a44d12R151 . There are still theoretical ways where memory use could grow, I think (if we have too many incomplete "write" futures live at once, since each holds a segment, and the IO executor has an unbounded queue size).

### DIFF
--- a/cpp/arcticdb/pipeline/frame_slice.hpp
+++ b/cpp/arcticdb/pipeline/frame_slice.hpp
@@ -173,11 +173,13 @@ private:
 
 // Collection of these objects is the input to batch_read_uncompressed
 struct RangesAndKey {
-    explicit RangesAndKey(const FrameSlice& frame_slice, entity::AtomKey&& key):
+    explicit RangesAndKey(const FrameSlice& frame_slice, entity::AtomKey&& key, bool is_incomplete):
     row_range_(frame_slice.rows()),
     col_range_(frame_slice.columns()),
-    key_(std::move(key)) {
+    key_(std::move(key)),
+    is_incomplete_(is_incomplete) {
     }
+
     explicit RangesAndKey(const RowRange& row_range, const ColRange& col_range, const entity::AtomKey& key):
             row_range_(row_range),
             col_range_(col_range),
@@ -203,6 +205,10 @@ struct RangesAndKey {
         return key_.end_time() - 1;
     }
 
+    bool is_incomplete() const {
+        return is_incomplete_;
+    }
+
     friend bool operator==(const RangesAndKey& left, const RangesAndKey& right) {
         return left.row_range_ == right.row_range_ && left.col_range_ == right.col_range_ && left.key_ == right.key_;
     }
@@ -214,6 +220,7 @@ struct RangesAndKey {
     RowRange row_range_;
     ColRange col_range_;
     entity::AtomKey key_;
+    bool is_incomplete_{false};
 };
 
 /*
--- a/cpp/arcticdb/version/version_core-inl.hpp
+++ b/cpp/arcticdb/version/version_core-inl.hpp
@@ -13,6 +13,7 @@
 #include <arcticdb/stream/merge.hpp>
 #include <arcticdb/pipeline/index_utils.hpp>
 #include <arcticdb/stream/segment_aggregator.hpp>
+#include <arcticdb/version/schema_checks.hpp>
 
 namespace arcticdb {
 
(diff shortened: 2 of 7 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -890,25 +890,18 @@ bool read_incompletes_to_pipeline(
         ensure_timeseries_norm_meta(*pipeline_context->norm_meta_, pipeline_context->stream_id_, sparsify);
     }
 
-    if (!dynamic_schema) {
-        const auto first_different_seg = std::adjacent_find(
```
poodlewars, line 894: This check was the cause of the high memory use. It calls `.segment()` on every `SliceAndKey` which loads the segment from storage, and then holds it live in memory for the rest of the processing pipeline.

This PR moves this check so that it is done inline during the aggregation of the incomplete segments during compaction.

I also implement changes to make sure that we clean up any data segments that we wrote during a compaction that eventually failed. This change also applies to `defragment_symbol_data` since it also uses the `do_compact` function. This seems correct.

cpp/arcticdb/version/version_core-inl.hpp
```diff
@@ -99,37 +100,64 @@ void merge_frames_for_keys(
...
     std::vector<pipelines::FrameSlice>& slices,
     const std::shared_ptr<Store>& store,
     bool convert_int_to_float,
     std::optional<size_t> segment_size,
-    bool validate_index){
+    bool validate_index,
+    compaction_details::StaticSchemaCompactionChecks&& checks) {
+        using namespace compaction_details;
+        CompactionResult result;
         auto index = stream::index_type_from_descriptor(pipeline_context->descriptor());
```
poodlewars, line 136: aa

cpp/arcticdb/version/version_core-inl.hpp
```diff
@@ -99,37 +100,64 @@ void merge_frames_for_keys(
...
     const std::shared_ptr<pipelines::PipelineContext>& pipeline_context,
-    std::vector<folly::Future<VariantKey>>& fut_vec,
     std::vector<pipelines::FrameSlice>& slices,
     const std::shared_ptr<Store>& store,
     bool convert_int_to_float,
     std::optional<size_t> segment_size,
-    bool validate_index){
+    bool validate_index,
+    compaction_details::StaticSchemaCompactionChecks&& checks) {
+        using namespace compaction_details;
```
poodlewars, line 134: k

cpp/arcticdb/version/version_core-inl.hpp
```diff
@@ -99,37 +100,64 @@ void merge_frames_for_keys(
...
+        CompactionResult result;
         auto index = stream::index_type_from_descriptor(pipeline_context->descriptor());
+
```
poodlewars, line 137: k

cpp/arcticdb/pipeline/frame_slice.hpp
```diff
@@ -214,6 +220,7 @@ struct RangesAndKey {
     RowRange row_range_;
     ColRange col_range_;
     entity::AtomKey key_;
+    bool is_incomplete_{false};
```
poodlewars, line 223: It feels a bit dirty to add this here, but since the main point of this struct is an input parameter to reads (to which we may or may not need to chain the extra validation) it actually seems fairly suitable.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -669,15 +669,54 @@ std::shared_ptr<std::unordered_set<std::string>> columns_to_decode(const std::sh
     return res;
 }
 
-std::vector<RangesAndKey> generate_ranges_and_keys(const std::vector<SliceAndKey>& slice_and_keys) {
+std::vector<RangesAndKey> generate_ranges_and_keys(PipelineContext* const pipeline_context) {
```
willdealtry, line 672: We don't generally use pointers unless the result can be null, is it possible to use a reference? A reference sort of implies that the lifetime is managed by the enclosing functions, whereas a pointer implies that the memory is entirely elsewhere (and might  be null or invalid)

The one place that we break the rule is with classes, where sometimes reference initialization of members can be annoying for PODs, as they are no longer trivially copyable etc

cpp/arcticdb/version/version_core.hpp
```diff
@@ -209,5 +210,101 @@ folly::Future<ReadVersionOutput> read_frame_for_version(
...
+            else
+                return *it;
+        }();
+        if (sk.slice().rows().diff() == 0) {
+            continue;
+        }
+
+        const SegmentInMemory& segment = sk.segment(store);
+
+        if(validate_index && segment.descriptor().sorted() != SortedValue::ASCENDING && segment.descriptor().sorted() != SortedValue::UNKNOWN) {
```
willdealtry, line 278: I'd be tempted here to move the conditions (besides validate _index) into a short well-named function.

cpp/arcticdb/pipeline/frame_slice.hpp
willdealtry: Yeah I'm fine with it. The implication is of not incomplete is that it's been indexed before somewhere, so that even in the case where we were merging timeseries together it is at least correct wherever it comes from (and so any schema mismatch errors would ideally happen at a higher level). Whereas if it's incomplete then all bets are off, it's just some isolated data that a user wrote and might contain anything

willdealtry (approved):

PR #2014, opened 2024-11-25
Reliable storage lock

IvoDD:
Introduces a reliable storage lock

Introduces a new `ReliableStorageLock` and `ReliableStorageLockGuard` to
be used as a slower but more reliable alternative to the existing
`StorageLock`.

It uses the new If-None-Match atomic put operations in S3.

First commit (Reliable storage lock):
- Upgrades the aws-sdk-cpp in vcpkg (which needed a few additions
  because of some problematic dependencies)
- Adds `write_if_none_match` capability to `AsyncStore`'s S3 and to
  `InMemoryStore`
- Logic for `ReliableStorageLock`
- C++ tests using the `InMemoryStore`

Second commit (Real S3 storage python tests for ReliableStorageLock)
- Adds a `ReliableStorageLockManager` which exposes functions to aquire and free the lock in python. (The guard structure is unusable for python)
- Adds a new storage lock test with the existing `real_s3_version_store` fixture.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/entity/key.cpp
```diff
@@ -60,6 +60,7 @@ KeyData get_key_data(KeyType key_type) {
     STRING_REF(KeyType::APPEND_REF, aref, 'a')
     STRING_KEY(KeyType::MULTI_KEY, mref, 'm')
     STRING_REF(KeyType::LOCK, lref, 'x')
+    STRING_REF(KeyType::SLOW_LOCK, lref, 'x')
```
poodlewars, line 63: This clashes with the character for `LOCK` above which is very odd. Is this intentional?

cpp/arcticdb/stream/piloted_clock.hpp
```diff
@@ -14,4 +16,11 @@ struct PilotedClock {
     }
 };
 
+struct PilotedClockNoAutoIncrement {
```
poodlewars, line 19: I think there's already a clock type in the codebase that works like this

cpp/arcticdb/stream/piloted_clock.hpp
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
poodlewars, line 1: Yuck

cpp/arcticdb/util/reliable_storage_lock.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace lock {
+
+using Epoch = uint64_t;
+
+// The ReliableStorageLock is a storage lock which relies on atomic If-None-Match Put and ListObject operations to
+// provide a slower but more reliable lock than the StorageLock. It should be completely consistent unless a process
+// holding a lock get's paused for times comparable to the lock timeout.
+// It lock follows the algorithm described here:
+// https://www.morling.dev/blog/leader-election-with-s3-conditional-writes/
```
poodlewars, line 28: Maybe we should save of copy of this blog somewhere, will suck to have no docs of how this works if it vanishes

cpp/arcticdb/util/test/test_reliable_storage_lock.cpp
```diff
@@ -0,0 +1,127 @@
...
+        }
+        // std::cout<<"Freeing a lock with "<<value_before_sleep<<std::endl;
+        cnt_ = value_before_sleep + 1;
+    }
+};
+
+
+TEST(ReliableStorageLock, StressMultiThreaded) {
+    // It is hard to use a piloted clock for these tests because the folly::FunctionScheduler we use for the lock
+    // extensions doesn't support a custom clock. Thus this test will need to run for about 2 minutes.
```
poodlewars, line 105: It's useful to keep `test_unit_arcticdb` running fast. Might be better to add this stress test to a new cmake target. Can be done as a follow up.

python/tests/integration/arcticdb/test_storage_lock.py
```diff
@@ -0,0 +1,49 @@
...
+    ]
+    for p in processes:
+        p.start()
+
+    for p in processes:
+        p.join()
+
+    read_df = lib.read(symbol).data
+    expected_df = pd.DataFrame({"col": [num_processes]})
+    assert_frame_equal(read_df, expected_df)
```
poodlewars, line 49: Also assert on the version number after all these writes?

poodlewars (changes requested):

cpp/arcticdb/entity/key.cpp
willdealtry: They all have to be unique, otherwise bad things will happen. Also, I'm not sure why we're characterising this as a slow lock, since if anything it will be faster than the other lock which required variable (and ultimately unknowable) wait periods to ensure that it really was the holder of the lock.

My preference would be that the eventual lock key (the copy destination) is just a normal KeyType::LOCK, and the one that gets copied to it (the first one written is called a PENDING_LOCK or a LOCK_ATTEMPT or something like that

cpp/arcticdb/storage/s3/s3_client_wrapper.hpp
```diff
@@ -57,7 +57,8 @@ class S3ClientWrapper {
     virtual S3Result<std::monostate> put_object(
             const std::string& s3_object_name,
             Segment&& segment,
-            const std::string& bucket_name) = 0;
+            const std::string& bucket_name,
+            bool if_none_match = false) = 0;
```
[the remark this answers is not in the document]
willdealtry: also bools tend to multiply out of control, it's preferable to use an enum, that way callers have to have a name on them (WriteConditions::IF_NONE_MATCH is way more self-explanatory at the call site than 'true')

cpp/arcticdb/util/reliable_storage_lock.hpp
```diff
@@ -0,0 +1,69 @@
...
+    std::optional<Epoch> aquired_epoch_;
+    folly::Func on_lost_lock_;
+    folly::FunctionScheduler extend_lock_heartbeat_;
+};
+
+}
+
+}
+
+#include "arcticdb/util/reliable_storage_lock.tpp"
```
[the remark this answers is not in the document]
willdealtry: The other alternative is to put them in reliable_storage_lock-inl.hpp, which is how we usually indicate that really this is implementation but it's templated implementation. You can look at any of the other -inl.hpp files for the usual mechanism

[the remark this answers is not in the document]
willdealtry: As in normally we include them at the bottom of the header file, and ensure that they're not included anywhere else, that way you get file separation, but you don't need to know about the additional file in your cpp files, you just include the header as normal

cpp/arcticdb/util/reliable_storage_lock.tpp
```diff
@@ -0,0 +1,242 @@
...
+
+namespace arcticdb {
+
+namespace lock {
+
+const auto SEPARATOR = '*';
+const auto EXTENDS_PER_TIMEOUT = 5u;
+
+inline StreamDescriptor lock_stream_descriptor(const StreamId &stream_id) {
+    return StreamDescriptor{stream_descriptor(
```
willdealtry, line 26: I don't think it needs the other constructor as the function already returns a stream descriptor

cpp/arcticdb/util/reliable_storage_lock.tpp
```diff
@@ -0,0 +1,233 @@
...
+    SegmentInMemory output{lock_stream_descriptor(name)};
+    output.set_scalar(0, expiration);
+    output.end_row();
+    return output;
+}
+
+template <class ClockType>
+ReliableStorageLock<ClockType>::ReliableStorageLock(const std::string &base_name, const std::shared_ptr<Store> store, timestamp timeout) :
+    base_name_(base_name), store_(store), timeout_(timeout) {
+    auto s3_timeout = ConfigsMap::instance()->get_int("S3Storage.RequestTimeoutMs", 200000) * ONE_MILLISECOND;
```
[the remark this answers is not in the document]
willdealtry: Yeah we'll want to use it for azure etc which already has similar operations, it would be better to pass it in as a policy from the storage where we just expect a function that is void() that does whatever we need

cpp/arcticdb/util/reliable_storage_lock.tpp
```diff
@@ -0,0 +1,242 @@
...
+
+inline Epoch get_next_epoch(std::optional<Epoch> maybe_prev) {
+    if (maybe_prev.has_value()) {
+        return maybe_prev.value() + 1;
+    }
+    return 0;
+}
+
+template <class ClockType>
+StreamId ReliableStorageLock<ClockType>::get_stream_id(Epoch e) const {
```
willdealtry, line 64: e is commonly used for error, no harm in calling it 'epoch'

cpp/arcticdb/storage/s3/s3_client_wrapper.hpp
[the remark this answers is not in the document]
IvoDD: Thanks, will use a `PutHeader::IF_NONE_MATCH` enum

cpp/arcticdb/util/reliable_storage_lock.hpp
IvoDD: Given that we'll soon replace the locking with If-Match based locks, we can probably skip this.

cpp/arcticdb/util/reliable_storage_lock.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <arcticdb/util/configs_map.hpp>
+
+#include <mutex>
+#include <folly/experimental/FunctionScheduler.h>
+
+namespace arcticdb {
+
+namespace lock {
+
+using Epoch = uint64_t;
```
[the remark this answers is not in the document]
IvoDD: Thanks, used `AcquiredLockId`

cpp/arcticdb/util/reliable_storage_lock.tpp
```diff
@@ -0,0 +1,233 @@
...
+    auto [existing_locks, latest_epoch] = get_all_locks();
+    util::check(latest_epoch.has_value() && latest_epoch.value() >= held_lock_epoch,
+                "We are trying to free a newer epoch than the existing one in storage. Free epoch: {}, Existing epoch: {}",
+                held_lock_epoch, latest_epoch);
+    if (latest_epoch.value() != held_lock_epoch) {
+        // Lock is already lost
+        return;
+    }
+    auto lock_stream_id = get_stream_id(held_lock_epoch);
+    auto expiration = ClockType::nanos_since_epoch(); // Write current time to mark lock as expired as of now
```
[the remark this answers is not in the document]
IvoDD: Yeah you're right. The difference is not big though. Just 2x the clock skew (if we e.g. expect clock skew to be 100ms for a process this would enstill a 200ms total skew between the two processes, otherwise if we use the s3 time as a source of truth we would suffer only the 100ms skew).

Let's do this together with the If-Match rework (which will also require passing in the ETag from s3 to the `Store` API and we can do a similar change to use the last-modified time)

cpp/arcticdb/util/reliable_storage_lock.tpp
```diff
@@ -0,0 +1,233 @@
...
+timestamp ReliableStorageLock<ClockType>::get_expiration(RefKey lock_key) const {
+    auto kv = store_->read_sync(lock_key, storage::ReadKeyOpts{});
+    return kv.second.template scalar_at<timestamp>(0, 0).value();
+}
+
+template <class ClockType>
+void ReliableStorageLock<ClockType>::clear_old_locks(const std::vector<Epoch>& epochs) const {
+    auto now = ClockType::nanos_since_epoch();
+    auto to_delete = std::vector<VariantKey>();
+    // We only clear locks that have expired more than a timeout (we assume a process can't be paused for more than the timeout) ago.
```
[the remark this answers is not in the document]
IvoDD: Yes, ok done.

cpp/arcticdb/util/test/test_reliable_storage_lock.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    // We take the first lock at 0 and it should not expire until 20
+    Clock::time_ = 0;
+    ASSERT_EQ(lock1.try_take_lock(), std::optional<uint64_t>{0});
+    Clock::time_ = 5;
+    ASSERT_EQ(lock2.try_take_lock(), std::nullopt);
+    Clock::time_ = 10;
+    ASSERT_EQ(lock2.try_take_lock(), std::nullopt);
+    Clock::time_ = 19;
+    ASSERT_EQ(lock1.try_take_lock(), std::nullopt);
```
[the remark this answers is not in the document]
IvoDD: Thanks, agreed done

cpp/vcpkg.json
```diff
@@ -31,16 +31,53 @@
...
     },
     "mongo-cxx-driver",
-    {
-      "name": "aws-c-io",
-      "$version reason": "To pick up https://github.com/awslabs/aws-c-io/pull/515"
-    },
     {
       "name": "aws-sdk-cpp",
-      "$version reason": "Minimum version in the baseline that works with aws-c-io above.",
+      "version>=": "1.11.405",
```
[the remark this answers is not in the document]
IvoDD: I needed to enforce fairly new versions of all aws packages because otherwise newer versions of `aws-sdk-cpp` would not compile.

To do this I used `version>=` so in the future if aws packages fix their vcpkg dependencies we would not have to update all aws overrides when we want o upgrade the `aws-sdk-cpp`.

I think if we use an override we don't allow vcpkg to resolve dependency versions for us, but given it's currently broken either way I'll just add an override only for `aws-sdk-cpp`. I'll leave the other aws packages with `version>=` to not require upgrading all overrides when we only want to upgrade the `aws-sdk-cpp`

cpp/arcticdb/util/test/test_reliable_storage_lock.cpp
IvoDD: I agree it's useful to separate slower stress tests. I rememember other stress tests e.g. some of the tests in `test_slab_alloc.cpp` were also quite slow, we should probably move them together.

Adding a note to fix in the If-match cleanup.

python/tests/integration/arcticdb/test_storage_lock.py
IvoDD: I've noticed when running these that the persistent storage might not get cleaned up after a previous test run especially when running locally. So I think it might be possible to do these writes on a library which already has the `counter` written with some arbitrary version. I.e. it's possible we get a version 200 because the test was run before.

I think the counter being equal to 100 is essentially the same assertions as writing 100 versions, so there isn't much benefit and it makes local testing slightly more annoying. This ifine for the CI I think. What do you think?

cpp/arcticdb/util/reliable_storage_lock.tpp
IvoDD: Used `lock_id` because I renamed `Epoch` to `AcquiredLockId`

IvoDD:
Some of the discussed follow up fixes will be done in https://github.com/man-group/ArcticDB/issues/2028

cpp/arcticdb/util/reliable_storage_lock.tpp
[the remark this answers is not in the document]
poodlewars: I think that's reasonable on the basis that clock drift can still cause locking failures even if we make the change I suggested.

python/tests/integration/arcticdb/test_storage_lock.py
poodlewars: Would be simple to do a force delete symbol at the start of the test?

poodlewars (approved):

python/tests/integration/arcticdb/test_storage_lock.py
IvoDD: Ah didn't know about the `force_delete_symbol` API. Done, thanks!

cpp/CMakeLists.txt
```diff
@@ -84,7 +84,9 @@ if(WIN32)
     endif()
 else()
     if(${ARCTICDB_USING_CONDA})
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wextra")
+        # Required to be able to include headers from glog since glog 0.7
+        # See: https://github.com/google/glog/pull/1030
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wextra -DGLOG_USE_GLOG_EXPORT")
```
vasil-pashov, line 89: You don't need to change this but setting the CMAKE_CXX_FLAGS is an outdated approach. Unfortunately most of our code is setup like this :(

The correct thing to do is to use `target_compile_definitions`

PR #2015, opened 2024-11-25
Finalize stage data tests

grusev:
#### Reference Issues/PRs

A stress test to check finilize_staged_data() under stress - iterating through rounds of different finilization setups with different sizes of chunked data. Perhaps the test will take 1-2hrs. Overall should stretch the product and examine its weks points like leaking memories etc.

ASV test for time and peak memory to benchmark the performance of finilize_staged_data()  NOTE: The test will be moved to a different PR after the approval of this PR, and removed from here, as ASV cannot produce results when dependent modules of the tests are not avail in repo. In this PR there are 2 utility modules which are used both by stress and the ASV test test.py and utils.py. 

Added several functional tests

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/arcticdb/util/utils.py
```diff
@@ -0,0 +1,325 @@
+"""
+Copyright 2024 Man Group Operations Limited
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+from arcticdb import Arctic
```
poodlewars, line 5: Remove

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+
+class CacheParts:
+
+    def __init__(self, model_df:pd.DataFrame):
+        self.dataframe = copy_dataframe_structure(model_df)
+
+    def cache_samples_from(self, data_frame: pd.DataFrame):
+        self.dataframe = pd.concat([self.dataframe, data_frame.head(1), data_frame.tail(1)])
+
+    def verify_finalized_data(self, lib: Library, symbol: str):
```
poodlewars, line 29: I don't understand why we need sampling, can't we just compare every row?

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+        Verification of column by type. Especially when dynamic schema was 
+        used and new columns were added later to original dataseries
+    """
+    row = 0
+    print ("EXPECTED NUMBERS:", array)
+    print ("ACTUAL   NUMBERS:", df[row_name].to_list())
+    for number in array:
+        arctic = df.iloc[row][row_name]
+
+        if ( (pd.isna(number)) or (number is None) ):
```
poodlewars, line 85: You don't need so many brackets

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+            else:
+                assert arctic is None , f"When adding str/object column, previous missing values should be None for row {row}"
+        else:
+            if ('str' in str(max_type)):
+                assert (number == arctic), f"Number {number} is not same in arcticdb {arctic} for row {row}"
+            else:
+                # When we upcast we first try to evaluate against original number
+                # if that fails we try to upcast and evaluate
+                assert ((number == arctic) or (max_type(number) == max_type(arctic))), f"Number {number} is not same in arcticdb {arctic} for row {row}"
+        row = row + 1
```
poodlewars, line 102: `row += 1`

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+def verify_dataframe_column(df:pd.DataFrame, row_name, max_type, array):
+    """
+        Verification of column by type. Especially when dynamic schema was 
+        used and new columns were added later to original dataseries
+    """
+    row = 0
+    print ("EXPECTED NUMBERS:", array)
+    print ("ACTUAL   NUMBERS:", df[row_name].to_list())
+    for number in array:
+        arctic = df.iloc[row][row_name]
```
poodlewars, line 83: arctic is a confusing name

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+        date_range = pd.date_range(start=start_date, periods=5, freq='D') 
+        df = pd.DataFrame({ 'Column1': [10, 20, 30, 40, 50] }, index=date_range)
+        df.index.name = 'timestamp'
+        return df
+
+    lib = arctic_library_lmdb
+    symbol = "symbol"
+    
+    df = small_dataframe("2023-01-01")
+    print(">>>> INDEX: ", df.index)
```
poodlewars, line 139: Remove print statements throughout the PR. Debug logging would be suitable.

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+    if (new_version):
+        # There is no problem to append empty dataframe
+        # However later this will result in problem in finalization
+        # which will be detected. For some reason the empty dataframe name of the index is 
+        # accepted as 'index' although it is 'timstamp'
+        lib.append(symbol, validate_index=True, data=empty_df)
+    else:
+        # The problem when we start with an empty dataframe the new version
+        # and then we try to append is more severe. 
+        # Segmentation fault is the result
```
poodlewars, line 154: If we segfault then you need to skip, not xfail, the test

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+        lib.write(symbol, validate_index=True, data=empty_df)
+        assert False, "REMOVE THIS WHEN FIXED!" #If we continue in this case python is DEAD!
+    #df = cachedDF.generate_dataframe_timestamp_indexed(2, 40, cachedDF.TIME_UNIT)
```
poodlewars, line 157: ?

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+    total_number_rows = total_number_rows + num_rows_initially
+    lib.write(symbol, data=df, prune_previous_versions=True)
+    cachedParts = CacheParts(df)
+    cachedParts.cache_samples_from(df)
+    chunk_list = generate_chunk_sizes(2, 900, 1100)
+    for chunk_size in chunk_list:
+        df = cachedDF.generate_dataframe_timestamp_indexed(chunk_size, total_number_rows, cachedDF.TIME_UNIT)
+        lib.write(symbol, data=df, validate_index=True, staged=True)
+        ## Unfortunately there is a bug with empty dataframe with index
+        ## Uncomment when this is solved
```
poodlewars, line 412: An `xfail`'d test with a reference to a bug ticket is more suitable than commented out code

python/tests/integration/arcticdb/test_arctic_additional.py
```diff
@@ -0,0 +1,421 @@
...
+
+    num_rows_initially = 99
+    print(f"Writing to symbol initially {num_rows_initially} rows")
+    df = cachedDF.generate_dataframe_timestamp_indexed(num_rows_initially, total_number_rows, cachedDF.TIME_UNIT)
+    dataframe_dump_to_log("df", df)
+    empty_df = df.drop(df.index)
+    dataframe_dump_to_log("empty", empty_df)
+    total_number_rows = total_number_rows + num_rows_initially
+    lib.write(symbol, data=df, prune_previous_versions=True)
+    cachedParts = CacheParts(df)
```
poodlewars, line 405: I don't understand the need for `CacheParts`. Can't we just create a dataframe, stage chunks from it, finalize it, and check that the result matches the original dataframe?

python/tests/stress/arcticdb/version_store/test_stress_finalize_stage_data.py
```diff
@@ -0,0 +1,111 @@
...
+def test_finalize_monotonic_unique_chunks(arctic_library_lmdb):
+
+    options = [
+        {"chunks_descending" : False, "finalization_mode" : StagedDataFinalizeMethod.APPEND},
+        {"chunks_descending" : True, "finalization_mode" : StagedDataFinalizeMethod.APPEND},
+        {"chunks_descending" : False, "finalization_mode" : StagedDataFinalizeMethod.WRITE},
+        {"chunks_descending" : True, "finalization_mode" : StagedDataFinalizeMethod.WRITE},
+        ]
+
+    # Will hold the results after each iteration:
```
poodlewars, line 38: Defining a small `dataclass` for the objects you put in to results will help to make this clearer without the need for the comment

python/tests/stress/arcticdb/version_store/test_stress_finalize_stage_data.py
```diff
@@ -0,0 +1,111 @@
...
+            total_number_rows = INITIAL_TIMESTAMP + sum(chunk_list)
+
+
+        print("--" * 50)
+        print(f"STAGED ROWS {total_number_rows.get_value()} after iteration {cnt}")
+        print(f"SYMBOL ACTUAL ROWS before finalization - {lib._nvs.get_num_rows(symbol)} ")
+        start_time = time.time()
+        lib.finalize_staged_data(symbol=symbol, mode=options[cnt % 4]["finalization_mode"])
+        finalization_time = time.time() - start_time
+        gc.collect()
```
poodlewars, line 89: What's the point of the gc collect calls? You aren't doing anything with memory statistics?

python/tests/stress/arcticdb/version_store/test_stress_finalize_stage_data.py
poodlewars, line 1: How long does this test take to run?

I don't quite see its value given that it doesn't do any assertions on the result, and we already have ASV for performance and memory measurement.

Similar to an earlier comment, I think we need to make CI changes before we add extremely slow tests like this. A little bit like the current "real s3 tests" pytest mark that only gets enabled on certain CI runs, we could add a mark that increases the size of the test cases being executed, and only run that step in the CI on demand or in a nightly build.

poodlewars (changes requested):
Thanks - the tests themselves look good, most of the comments are about Python style. We do need to make sure we don't slow down the tests that run on every PR, so I think the test cases here need to be adjusted to run on smaller data until we have some kind of slow nightly build in the CI. That might be a good thing to work on next.

python/arcticdb/util/utils.py
grusev: Will remove - I thought tests also need this header .... Will not put it anymore

python/arcticdb/util/utils.py
```diff
@@ -0,0 +1,325 @@
...
+    TIME_ZERO : pd.Timestamp = pd.Timestamp(0)
+
+    def __init__(self, value:np.int64, type:SupportedFreqTypes=DEFAULT_FREQ) -> None: 
+        self.init_value:np.int64 = value
+        self.value:np.int64 = value
+        self.__type:TimestampNumber.SupportedFreqTypes = type
+
+
+    def get_type(self) -> SupportedFreqTypes:
+        return self.__type
```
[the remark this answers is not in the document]
grusev: __ is also for private fields - ie fields that you would not expose externally (like read only fields). As far as I know python makes them harder to access by adding something like "self.__private__type" to simulte private fields. Thus in this particular case if we allow to access and modify __type this will jeopordize the object as if we assign directly here "h" and it was "s" the TimeStmapNumber now is totally different. It would require full conversion of the whole object, which is possible but not needed at this point of usage of this class

python/arcticdb/util/utils.py
```diff
@@ -0,0 +1,325 @@
...
+        '''
+            Returns the initial value of the number.
+            This allows you to serve like a reference
+        '''
+        return self.init_value
+
+    
+    @classmethod
+    def calculate_timestamp_after_n_periods(cls, periods:int, freq:SupportedFreqTypes='s', 
+            start_time: pd.Timestamp = TIME_ZERO) -> Union[pd.Timestamp, Tuple[pd.Timestamp, pd.Timestamp]]:
```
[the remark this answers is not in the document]
grusev: will fix!

python/arcticdb/util/utils.py
```diff
@@ -0,0 +1,325 @@
...
+    def __init__(self, max_size:int=1500000, size_string_flds_array=[25,1,5,56]):
+        """
+            Define the number of rows for the cached dataframe
+        """
+        self.__cached_xlarge_dataframe:pd.DataFrame = None
+        self.max_size = max_size
+        self.size_string_flds_array = size_string_flds_array
+
+    def get_dataframe(self) -> pd.DataFrame: 
+        return self.__cached_xlarge_dataframe
```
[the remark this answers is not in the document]
grusev: :-) will fix ...

python/arcticdb/util/utils.py
```diff
@@ -0,0 +1,325 @@
...
+        else:
+            raise Exception("Not supported frequency")
+
+        if (add):
+            return end_time, (start_time, end_time)
+        else:
+            return end_time, (end_time , start_time)
+
+
+    @classmethod
```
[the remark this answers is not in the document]
grusev: The point of staticmethod is to make it object independent but at the same time add it to class that provides utilties.

From there is much easier to access relevant utils and not have to browse through hundreds of methods ot find one you need, example test.py ...

I would think that such kind of grouping utilities by classes would much more speed up work as this forces us to combine similar functions under single class thus would not spread here and there.

At least that is how things are organized in other OOP languages liek C#, Java ... 

Thus perhaps this is a topic for discussion then?

python/tests/integration/arcticdb/test_arctic_additional.py
grusev: yes, indeed that portion is extraxcted in "test_finalize_empty_dataframe". Ideally when fixed it could be added also back to the tests for completeness ... As now we cannot leave it there.

python/tests/integration/arcticdb/test_arctic_additional.py
grusev: We could .... if all we want to stage is small dataframes that fit in memory. Once we produce more DFs that we store in symbol than what we can store potentially in memory cache that will not be possible.

Thus if we leave the decision of what parts of DF to cache only and later check to such class it will be easier to create test with much more data. 

While this might not have be the concrete case - in many other occasions it could be.

I have no problem to remove this now and use direct dataframes cached and appeneded in huge one this time, and use this approach when we really lack enough mem

python/tests/stress/arcticdb/version_store/test_stress_finalize_stage_data.py
grusev: Will do that! thanks!

python/tests/stress/arcticdb/version_store/test_stress_finalize_stage_data.py
grusev: The test run arroun 90 mins on my machine. Thus it is long running. And it will be slow. We have to marke them indeed in a way so that it gets executed perhaps not that regularly. Is there already such mark?

As for the value - we currently assume that asv tests will capture regression like that we had. But we do not know that.

The stress test was capable of reproducing the issue, we know that for sure. Thus its value although that might seem questional is proven. Also the asv tests in their nature are different from this stress test. This stress tests adds all things in one symbol. The asv tests prepare data for each run because of certain limitation. see my comment there in setup() method

Therefore I cannot say that this stress test is redundant

python/arcticdb/util/utils.py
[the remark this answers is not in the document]
poodlewars: The convention is to have a single underscore for private fields, not two

python/arcticdb/util/utils.py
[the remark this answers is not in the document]
poodlewars: We don't normally write code like this. Static methods in Java are a thing because (until very recently) Java forced you to always define code inside of a class. Python and C++ have no such constraint.

I do take the point that it's useful for organisation, so I'm OK with this as-is. But please try to avoid static methods unless there is a good reason

python/benchmarks/finalize_staged_data.py
```diff
@@ -0,0 +1,85 @@
...
+        Check and benchmark performance of finalize_staged_data().
+        Due to specifics of this procedure we tune asv to make single measurement
+        which would be over a relatively big staged data.
+    '''
+    number = 1
+    rounds = 1
+    repeat = 1
+    min_run_count = 1
+
+    timeout = 1200
```
poodlewars, line 27: This is still 20 minutes which is too long to include in a normal CI run

python/tests/stress/arcticdb/version_store/test_stress_finalize_stage_data.py
poodlewars: Have a look at `tests/util/mark.py` for some example marks. I think we could perhaps rename the "REAL_S3_TESTS_MARK" to be a more general "on demand only" mark for tests that we trigger manually rather than on every CI run. Or we could create a new mark and CI option. @G-D-Petrov can help with this

poodlewars (approved):

PR #2017, opened 2024-11-25
Fix release issues

willdealtry:
Fix issues with backwards compatibility arising from 5.1 branch

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "cpp/vcpkg"]
 	path = cpp/vcpkg
 	url = https://github.com/microsoft/vcpkg.git
+[submodule "cpp/third_party/entt"]
+	path = cpp/third_party/entt
+	url = https://github.com/skypjack/entt.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,6 +16,3 @@
 [submodule "cpp/vcpkg"]
 	path = cpp/vcpkg
 	url = https://github.com/microsoft/vcpkg.git
-[submodule "cpp/third_party/entt"]
-	path = cpp/third_party/entt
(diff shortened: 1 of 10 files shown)


.gitmodules
```diff
@@ -16,6 +16,3 @@
 [submodule "cpp/vcpkg"]
 	path = cpp/vcpkg
 	url = https://github.com/microsoft/vcpkg.git
-[submodule "cpp/third_party/entt"]
```
poodlewars, line 19: We still need this, don't we?

python/arcticdb/version_store/_normalization.py
```diff
@@ -737,10 +738,40 @@ def denormalize(self, item, norm_meta):
 
         columns, denormed_columns, data = _denormalize_columns(item, norm_meta, idx_type, n_indexes)
 
-        if index is not None:
```
poodlewars, line 740: Noting original PR that is being reverted here https://github.com/man-group/ArcticDB/pull/1561 

We used to have `self._nvs._normalizer.df._skip_df_consolidation = True` in `library.py` - are we deliberately turning it back on for `Library` objects here?

`test_flattener.py` used to have `assert lib.read("sym").data[0].equals(test_data[0])` - we should change it to `assert_equal` in this PR shouldn't we? It still has `assert_series_equal` on master.

willdealtry: I think it was just accidental that the only equals test happened to be in that test, given that it passes the more stringent test we might as well leave that one as I've changed it, and check that both assert_frame_equal and equals() work for a wide variety of dataframes, as I've done in test_basic_version_store

poodlewars (approved):

PR #2019, opened 2024-11-25
new bi sample benchmark tests v2 (fromPR #1995)

grusev:
#### Reference Issues/PRs

From PR #1995 (all comments addressed)

    Sample test benchmark for using one opensource BI CSV source.
    The logic of a test is 
        - download if parquet file does not exists source in .bz2 format
        - convert it to parquet format
        - prepare library with it containing  several symbols that are constructed based on this DF
        - for each query we want to benchmark do a pre-check that this query produces SAME result on Pandas and arcticDB
        - run the benchmark tests

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/data/CityMaxCapita_1.parquet.gzip
alexowens90, line 1: parquet is already a compressed format, so there probably isn't any additional benefit to gzip

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,168 @@
...
+
+        print("The procedure is creating N times larger dataframes")
+        print("by concatenating original DF N times")
+        for num in BIBenchmarks.params:
+            _df = self.df.copy(deep=True)
+            if (num > 1):
+                # lets create N times bigger DF
+                dfcum = self.df.copy(deep=True)
+                for i in range(1, (BIBenchmarks.params[-1])):
+                    dfcum = pd.concat([dfcum, self.df])
```
alexowens90, line 89: This is much less efficient that calling concat once with a longer list of dataframes

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,168 @@
...
+
+    number = 2
+    timeout = 6000
+    LIB_NAME = "BI_benchmark_lib"
+    # We use dataframe in this file
+    CITY_BI_FILE = "data/CityMaxCapita_1.csv.bz2"
+    CITY_BI_FILE2 = "data/CityMaxCapita_1.parquet.gzip"
+
+    #Defines how many times bigger the database is
+    params = [1, 10]
```
alexowens90, line 55: How big is the original data set in terms of number of rows? How long does this whole thing take to run?

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,168 @@
...
+import os
+from pathlib import Path
+from arcticdb import Arctic
+from arcticdb.version_store.processing import QueryBuilder
+
+from .common import *
+
+
+def get_query_groupby_city_count_all(q):
+    return q.groupby("City").agg({"Keyword": "count"})
```
alexowens90, line 17: nit: it would be better programming practice to construct the `QueryBuilder` in these methods and then return it, rather than taking a parameter.

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,168 @@
...
+    params = [1, 10]
+
+    def __init__(self):
+        self.lib_name = BIBenchmarks.LIB_NAME
+        self.symbol = self.lib_name
+
+    def setup_cache(self):
+
+        file = os.path.join(Path(__file__).resolve().parent.parent, BIBenchmarks.CITY_BI_FILE2)
+        if (not os.path.exists(file)) :
```
alexowens90, line 64: Why is this needed since the file is checked in?

python/data/CityMaxCapita_1.parquet.gzip
grusev: size of parquet file is 800MB when gzipped it is 95MB

python/benchmarks/bi_benchmarks.py
grusev: Agree, but in this particular case I use those methods to pass QueryBuilder or DataFrame, so that I can use same methods to work with arctic and dataframes (pandas (see the setup cache)

grusev: But you have point as this si not obvious, I will add type hints here

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,168 @@
...
+
+def get_query_groupby_city_count_filter_two_aggregations(q):
+    return q[q["Keyword"] == "faggot" ].groupby("City").agg({"Keyword": "count", "Number of Records" : "sum"})  
+
+def assert_frame_equal(pandas_df:pd.DataFrame, arctic_df:pd.DataFrame):
+    arctic_df.sort_index(inplace=True)
+    pd.testing.assert_frame_equal(pandas_df,
+                                    arctic_df, 
+                                    check_column_type=False, 
+                                    check_dtype=False)
```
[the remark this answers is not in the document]
grusev: The one that we have is not suitable for this and I had to come up with one that will not fail due to not relevant things

python/benchmarks/bi_benchmarks.py
grusev: 912k rows. 

25 secs generating and doing functional checks over the original and large 10x dataframe (9120k rows)

See more details for the test:

OUTPUT -------->
             Parquet file exists!
             The procedure is creating N times larger dataframes
             by concatenating original DF N times
             Size of original Dataframe:  912657
             DF for iterration xSize original ready:  10
             <class 'pandas.core.frame.DataFrame'>
             Index: 9126570 entries, 0 to 912656
             Data columns (total 31 columns):
              #   Column               Dtype
             ---  ------               -----
              0   City/Admin           object
              1   City/State           object
              2   City                 object
              3   Created Date/Time    float64
              4   Date Joined          float64
              5   FF Ratio             float64
              6   Favorites            int32
              7   First Link in Tweet  object
              8   Followers            int32
              9   Following            int32
              10  Gender               object
              11  Influencer?          int32
              12  Keyword              object
              13  LPF                  float64
              14  Language             object
              15  Lat                  float64
              16  Listed Number        int32
              17  Long Domain          object
              18  Long                 float64
              19  Number of Records    int32
              20  Region               object
              21  Short Domain         object
              22  State/Country        object
              23  State                object
              24  Tweet Text           object
              25  Tweets               int32
              26  Twitter Client       object
              27  User Bio             object
              28  User Loc             object
              29  Username 1           object
              30  Username             object
             dtypes: float64(6), int32(7), object(18)
             memory usage: 10.8 GB
             If pandas query produces different dataframe than arctic one stop tests!
             This will mean query problem is there most likely
             Pre-check correctness for query_groupby_city_count_all
             Pre-check correctness for query_groupby_city_count_isin_filter
             Pre-check correctness for query_groupby_city_count_filter_two_aggregations
             All pre-checks completed SUCCESSFULLY. Time:  25.181268453598022
                                                                            ok
[31.25%] ··· Running (bi_benchmarks.BIBenchmarks.time_query_groupby_city_count_all--)....
[56.25%] ··· bi_benchmarks.BIBenchmarks.peakmem_query_groupby_city_count_all                                        ok
[56.25%] ··· ======== =======
              param1
             -------- -------
                1       372M
                10     2.77G
             ======== =======

[62.50%] ··· bi_benchmarks.BIBenchmarks.peakmem_query_groupby_city_count_filter_two_aggregations                    ok
[62.50%] ··· ======== =======
              param1
             -------- -------
                1       363M
                10     2.73G
             ======== =======

[68.75%] ··· bi_benchmarks.BIBenchmarks.peakmem_query_groupby_city_count_isin_filter                                ok
[68.75%] ··· ======== =======
              param1
             -------- -------
                1       361M
                10     2.69G
             ======== =======

[75.00%] ··· bi_benchmarks.BIBenchmarks.peakmem_query_readall                                                       ok
[75.00%] ··· ======== =======
              param1
             -------- -------
                1       936M
                10     8.06G
             ======== =======

[81.25%] ··· bi_benchmarks.BIBenchmarks.time_query_groupby_city_count_all                                           ok
[81.25%] ··· ======== ==========
              param1
             -------- ----------
                1      19.8±3ms
                10     163±4ms
             ======== ==========

[87.50%] ··· bi_benchmarks.BIBenchmarks.time_query_groupby_city_count_filter_two_aggregations                       ok
[87.50%] ··· ======== ==========
              param1
             -------- ----------
                1      21.3±3ms
                10     155±4ms
             ======== ==========

[93.75%] ··· bi_benchmarks.BIBenchmarks.time_query_groupby_city_count_isin_filter                                   ok
[93.75%] ··· ======== ==========
              param1
             -------- ----------
                1      20.0±2ms
                10     154±3ms
             ======== ==========

[100.00%] ··· bi_benchmarks.BIBenchmarks.time_query_readall                                                          ok
[100.00%] ··· ======== ============
               param1
              -------- ------------
                 1       365±20ms
                 10     3.54±0.02s
              ======== ============

python/benchmarks/bi_benchmarks.py
grusev: Originally I implemented only downloading, but it was slow. Then decided to leave this here, so that in future thise code can be copied and run even without having the data with you

python/benchmarks/bi_benchmarks.py
grusev: True!!!

python/benchmarks/bi_benchmarks.py
```diff
@@ -0,0 +1,168 @@
...
+            _df = self.df.copy(deep=True)
+            if (num > 1):
+                # lets create N times bigger DF
+                dfcum = self.df.copy(deep=True)
+                for i in range(1, (BIBenchmarks.params[-1])):
+                    dfcum = pd.concat([dfcum, self.df])            
+                _df = dfcum
+                print("DF for iterration xSize original ready: ", num)
+                _df.info(verbose=True,memory_usage='deep')
+            self.lib.write(f"{self.symbol}{num}", _df)
```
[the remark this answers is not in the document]
alexowens90: Ah OK. In that case this can still be tidied up, the copying is unnecessary:
```
for num in BIBenchmarks.params:
    _df = pd.concat([self.df] * num)
    print("DF for iterration xSize original ready: ", num)
    self.lib.write(f"{self.symbol}{num}", _df)
```

python/benchmarks/bi_benchmarks.py
alexowens90: And how about the runtime including the actual benchmarking? ASV runs each benchmark multiple times to reduce the standard deviation of the output, so it isn't clear what the total runtime would be from these numbers

python/benchmarks/bi_benchmarks.py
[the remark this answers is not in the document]
alexowens90: It just forwards to `pd.testing.assert_frame_equal`, so how is it not suitable?

python/benchmarks/bi_benchmarks.py
alexowens90: Ah I see, makes sense

python/benchmarks/common.py
alexowens90, line 1: Any reason these changes are in `common.py`? They're only used by `bi_benchmarks.py`, so should live in there, no?

python/benchmarks/bi_benchmarks.py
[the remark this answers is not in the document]
grusev: done

python/benchmarks/bi_benchmarks.py
grusev: total setup + tests <50secs

python/benchmarks/bi_benchmarks.py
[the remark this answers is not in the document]
grusev: done

PR #2022, opened 2024-11-26
Fix nfs test setup fail

phoebusm:
Seg fault in `moto` start up during fixture creation
Test affected: test_nfs_backed_s3_storage, test_read_path_with_dot
#### Any other comments?
For both test test_nfs_backed_s3_storage and test_read_path_with_dot , both needs to start up a new moto server respectively:
* test_nfs_backed_s3_storage
Sole user of nfs_backed_s3_storage
The moto server for nfs_backed_s3_storage will always be started at the spot
* test_read_path_with_dot
MotoNfsBackedS3StorageFixtureFactory and MotoS3StorageFixtureFactory will created moto server for themselves at the spot

Probably due to some underlying changes of the running image and version incompatibility, moto will seg fault at places related file descriptor, e.g. cannot read a json file packed in the moto wheel, or seg fault at importing werkzeug. I think it's due to file descriptor not being handled properly while using multiprocess fork to start up the server. The issue is gone once switched the startup method to spawn
The issue will be gone as well if the fixture is set autouse , to get the server started along with other storage fixture. e.g. s3_storage . I think that could be explained by different states of file descriptor at the time. So in another word, the issue should affect any late moto server start up.
As it only happens at py3.7, which is deprecated, I don't think it's worth opening a ticket for my observation.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/storage_fixtures/s3.py
+++ b/python/arcticdb/storage_fixtures/s3.py
@@ -12,8 +12,6 @@ import json
 import os
 import re
 import sys
-import trustme
-import subprocess
 import platform
 from tempfile import mkdtemp
 
@@ -339,7 +337,8 @@ class MotoS3StorageFixtureFactory(BaseS3StorageFixtureFactory):
             self.client_cert_file = ""
         self.client_cert_dir = self.working_dir
         
-        self._p = multiprocessing.Process(
+        spawn_context = multiprocessing.get_context('spawn')
+        self._p = spawn_context.Process(
             target=self.run_server,
             args=(
                 port,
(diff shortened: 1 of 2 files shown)


jamesmunro:
Could we have a comment explaining what the `spawn_context` is for and how it avoids the issue?

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -339,7 +337,8 @@ def _start_server(self):
             self.client_cert_file = ""
         self.client_cert_dir = self.working_dir
         
-        self._p = multiprocessing.Process(
+        spawn_context = multiprocessing.get_context('spawn')
```
jamesmunro, line 340: Could we have a comment explaining what the spawn_context is for and how it avoids the issue?

poodlewars (approved):

jamesmunro (approved):

Issue #2048, opened 2024-12-06
Add support for Numpy 2


G-D-Petrov:
**Is your feature request related to a problem? Please describe.**
`ArcticDB` currently support `Numpy<2`

**Describe the solution you'd like**
`ArcticDB` should support `Numpy>=2 `and the pin to it should be removed.
N.B. this involves and upgrade to `pybind to >=2.12` as well

PR #2050, opened 2024-12-06
2048 add support for numpy 2

G-D-Petrov:
#### Reference Issues/PRs
#2048 

There quite a few distinct changes as most of the work was in the CI and testing.
I will try to separate the different changes here:
- Numpy 2 Support: 
  - [support for new c++ interface](https://github.com/man-group/ArcticDB/pull/2050/files#diff-d23d2e3a46aa783e6c27de2ff06123f6149cda2727b09eb48e0755bc29a3ede2)
  - [small change for how == and is are evaluated](https://github.com/man-group/ArcticDB/pull/2050/files#diff-8d6759b6a2a11f86f70ee321e943d6c8e7898d4419ed281ad5a7358426a49a9a)
  - [added some more checks during normalization](https://github.com/man-group/ArcticDB/pull/2050/files#diff-fa52ad937e8dc0daf708927a99994e7264fd9cb62557326919c023696808859eL121)
  - changes to tests to support new Numpy conventions (all of the changes to the pytests)
- Pybind11 2.13 Support (Needed for numpy2):
  - removed python 3.6 and added 3.12 (relevant changes [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L30), [here
](https://github.com/man-group/ArcticDB/pull/2050/files#diff-ea07d77332dfd2427efe1d2148bec3d99c7dd3c06160f7975d5ad53069b9083f), and [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-ea07d77332dfd2427efe1d2148bec3d99c7dd3c06160f7975d5ad53069b9083f) 
  - Added new version of pybind11 and made changes to support the new interface:
    - [changed PythonUtils](https://github.com/man-group/ArcticDB/pull/2050/files#diff-57acdaae55064408141a79c555b1fe7c4f2d4a6ce7f7b291e814b36abe4bdbce) to add more info about the Python executable/libs neded for the newer version of pybind11
    - Changes to support new pybind11 interfaces (relevant changes [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-9357e7e6906b72682dbae796b1a310682b3f7b334e3bba0cfee71c08f2e8cf57), [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-dc9806fb1c292ea747ea6f0d333206c084159dcf6fc94919b98bfce3fd90d8c8), [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-02770184046bcf181939ea70f927d1f835687e57fbcc843a71a500684674b8e9),  [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-87b188a9b8a3a1ab16c10f975faa03e7f33e42347eb3019505b240432a4502e6), [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-6caec3c44be32d3a881cccbff1f03bbf8aedcdbf65e6e0c2c963e34bc675a246), [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-6c2d76c22148e822d651baf5e8c3c8e23528d7360afe98d3ba085104e46237db), [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-c7f1753eda666a2b8fd8c840f0ed2be21b809132b58ea6adbb14f49afb34039f) and [here](https://github.com/man-group/ArcticDB/pull/2050/files#diff-ed781e679ed39eaca3b7b13549967958654ab47eb42f920bfa5535933a1c097b)
- Changes for testing: (also added tickets for these): 
  - [Changed the persistent tests to use http instead of https, because the sts is broken on windows ](https://github.com/man-group/ArcticDB/pull/2050/files#diff-9c13e1c7a3ca0bbad7e24e48944e815de7bfe7ed57584ce31b0549eb7be6c94dL3) - #2080 
  - [change for persistent tests to use only encoding version 1, so as to not write the encoding_version to the cfg, because older versions cannot read it](https://github.com/man-group/ArcticDB/pull/2050/files#diff-11d84f04de731d90230614eaaeb91eb376a0b00061ce90dacbef195db6125005) - #2077 
  - [skip for storage lock test because it is flaky](https://github.com/man-group/ArcticDB/pull/2050/files#diff-f0c82624218a7a1ad65345a6e6ae054e689288ded7b0670999c5eabb9a1c89d9) - #2079 
  - [change finalize stage data to not be run against real s3 because it is too slow and it times out](https://github.com/man-group/ArcticDB/pull/2050/files#diff-67a5a1e0b68a1442a89e31f917eeaf651954e0b389d15c0ce0a6e7fd80706574) - #2078 
 
I've also tested in against the full suite of tests incl the persistent ones (run [here](https://github.com/man-group/ArcticDB/actions/runs/12388677078/job/34581688721))
**N.B.:** We need to continue using the submodule on pybind11 instead of getting it from vcpkg, because the latest version of pybind11 are not available on its vcpkg distribution.
Specifically, the versions that introduce the Numpy 2 Support.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,7 +104,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
         arcticdb_version: ["oldest", "latest"]
         include:
           - python_deps_ids: [""]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -225,7 +225,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
         arcticdb_version: ["oldest", "latest"]
         include:
           - python_deps_ids: [""]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,12 +170,24 @@ jobs:
               - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
               - python_deps_id: -compat38
                 python_deps: requirements-compatibility-py38.txt
+          - python3: 10
+            python_deps_ids: ["", -compat310]
+            matrix_override:
+              - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
+              - python_deps_id: -compat310
+                python_deps: requirements-compatibility-py310.txt
           - python3: 11
             python_deps_ids: ["", -compat311]
             matrix_override:
               - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
               - python_deps_id: -compat311
                 python_deps: requirements-compatibility-py311.txt
+          - python3: 12
+            python_deps_ids: ["", -compat312]
+            matrix_override:
+              - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
+              - python_deps_id: -compat312
+                python_deps: requirements-compatibility-py312.txt
     name: 3.${{matrix.python3}} Linux
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,24 +170,12 @@ jobs:
               - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
               - python_deps_id: -compat38
                 python_deps: requirements-compatibility-py38.txt
-          - python3: 10
-            python_deps_ids: ["", -compat310]
-            matrix_override:
-              - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
-              - python_deps_id: -compat310
-                python_deps: requirements-compatibility-py310.txt
           - python3: 11
             python_deps_ids: ["", -compat311]
             matrix_override:
               - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
               - python_deps_id: -compat311
                 python_deps: requirements-compatibility-py311.txt
-          - python3: 12
-            python_deps_ids: ["", -compat312]
-            matrix_override:
-              - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
-              - python_deps_id: -compat312
-                python_deps: requirements-compatibility-py312.txt
     name: 3.${{matrix.python3}} Linux
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
     uses: ./.github/workflows/cibw_docker_image.yml
     permissions: {packages: write}
     with:
-      cibuildwheel_ver: "2.12.1"
+      cibuildwheel_ver: "2.21.3"
       force_update: false
 
   common_config:
@@ -104,7 +104,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
         arcticdb_version: ["oldest", "latest"]
         include:
           - python_deps_ids: [""]
@@ -159,24 +159,23 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[7, 8, 9, 10, 11, 12]')}}
         include:
           - python_deps_ids: [""]
             matrix_override: ${{fromJson(needs.common_config.outputs.linux_matrix)}}
             pytest_xdist_mode: "--dist worksteal"
-          - python3: 6
-            python_deps_ids: ["", -compat36]
-            matrix_override:
-              - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
-              - python_deps_id: -compat36
-                python_deps: requirements-compatibility-py36.txt
-            pytest_xdist_mode: ""  # worksteal Not supported on Python 3.6
           - python3: 8
             python_deps_ids: ["", -compat38]
             matrix_override:
               - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
               - python_deps_id: -compat38
                 python_deps: requirements-compatibility-py38.txt
+          - python3: 11
+            python_deps_ids: ["", -compat311]
+            matrix_override:
+              - ${{fromJson(needs.common_config.outputs.linux_matrix)[0]}}
+              - python_deps_id: -compat311
+                python_deps: requirements-compatibility-py311.txt
     name: 3.${{matrix.python3}} Linux
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
@@ -207,7 +206,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[7, 8, 9, 10, 11, 12]')}}
     name: 3.${{matrix.python3}} Windows
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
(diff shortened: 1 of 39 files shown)


.github/workflows/build.yml
```diff
@@ -100,7 +100,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
```
vasil-pashov, line 103: Why is 12 missing here?

cpp/CMake/PythonUtils.cmake
```diff
@@ -125,13 +124,16 @@ if(ARCTICDB_FIND_PYTHON_DEV_MODE)
     if(NOT Python_EXECUTABLE AND NOT Python_ROOT_DIR AND NOT Python_LIBRARY)
         # FindPython searches the PATH environment last, but that's arguably the only correct place it should look
         find_program(Python_EXECUTABLE NAMES python3 python NAMES_PER_DIR REQUIRED NO_CMAKE_SYSTEM_PATH)
+        set(PYTHON_EXECUTABLE ${Python_EXECUTABLE} CACHE FILEPATH "Python executable found by FindPython")
```
vasil-pashov, line 127: I don't think this needs to be cached variable.

python/arcticdb/util/_versions.py
```diff
@@ -6,9 +6,11 @@
...
 """
 import pandas as pd
+import numpy as np
 from packaging import version
 
 PANDAS_VERSION = version.parse(pd.__version__)
 CHECK_FREQ_VERSION = version.Version("1.1")
 IS_PANDAS_ZERO = PANDAS_VERSION < version.Version("1.0")
 IS_PANDAS_TWO = PANDAS_VERSION >= version.Version("2.0")
+IS_NUMPY_TWO = version.parse(np.__version__) >= version.Version("2.0")
```
vasil-pashov, line 16: I know this is mimicking IS_PANDAS_TWO but I think it would be a better idea to check if version >=2 and version < 3.

python/arcticdb/version_store/_normalization.py
```diff
@@ -260,15 +260,16 @@ def _to_primitive(arr, arr_name, dynamic_strings, string_max_len=None, coerce_co
...
-    # Pick any unwanted data conversions (e.g. np.NaN to 'nan') or None to the string 'None'
+    # Pick any unwanted data conversions (e.g. np.nan to 'nan') or None to the string 'None'
     if np.array_equal(arr, casted_arr):
         return casted_arr
     else:
         if None in arr:
             raise ArcticDbNotYetImplemented(
                 "You have a None object in the numpy array at positions={} Column type={} for column={} "
-                "which cannot be normalized.".format(np.where(arr is None)[0], arr.dtype, arr_name)
+                "which cannot be normalized.".format(np.where(arr == None), arr.dtype, arr_name)
```
vasil-pashov, line 272: A comment that numpy have overridden `==` but not `is` would avoid confusion.

python/arcticdb/version_store/_store.py
```diff
@@ -1699,11 +1702,16 @@ def _postprocess_df_with_only_rowcount_idx(self, read_result, row_range):
...
             metadata=meta,
             host=self.env,
-            timestamp=read_result.version.timestamp
+            timestamp=read_result.version.timestamp,
         )
 
     def _resolve_empty_columns(self, columns, implement_read_index):
-        if not implement_read_index and columns == []:
+        is_columns_empty = (
+            columns is None
```
vasil-pashov, line 1710: This is used in order to implement [this](https://github.com/man-group/ArcticDB/blob/f9d955f71fbb7439ddce8724dda5ec8e23f6c171/python/arcticdb/version_store/library.py#L1445) in the NativeVersionStore [here](https://github.com/man-group/ArcticDB/blob/f9d955f71fbb7439ddce8724dda5ec8e23f6c171/python/arcticdb/version_store/_store.py#L1754)

None should mean read all columns while empty list should mean read only the index. I don't think this will work correctly but the tests are passing. Either I'm missing something or we have bad tests.

.github/workflows/build.yml
G-D-Petrov: This steps is setting up the older versions and writing data with them.
We haven't published python 3.12 for the older versions of arcticdb so it can't be added here yet.

cpp/CMake/PythonUtils.cmake
G-D-Petrov: Looks like it needs to be or the pybind11 cmake can't resolve to the correct Python (see [this run](https://github.com/man-group/ArcticDB/actions/runs/12413039790/job/34654113561))

python/tests/util/storage_test.py
```diff
@@ -137,7 +137,7 @@ def verify_library(ac):
 
 
 def is_strategy_branch_valid_format(input_string):
-    pattern = r"^(linux|windows)_cp3(6|7|8|9|10|11).*$"
+    pattern = r"^(linux|windows)_cp3(6|7|8|9|10|11|12|13).*$"
```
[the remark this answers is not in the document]
G-D-Petrov: No, we still want to install 3.6 in the persistent tests with the oldest version that we support for backwards compatibility.

I've also added 13 because we should add it soon, so better to save whoever is implementing it some headaches.

python/arcticdb/version_store/_store.py
G-D-Petrov: No idea about that, my change is mainly for supporting the case where `columns` is `ndarray` (`isinstance(columns, np.ndarray) and columns.size == 0)`) because some of the tests pass a `numpy array` to `columns` and `Numpy 2` is not happy with the old check.

cpp/CMake/PythonUtils.cmake
vasil-pashov: Oh I didn't see that it's not part of the CMakeLists.txt files but a separate file. Probably set with [parent scope](https://cmake.org/cmake/help/latest/command/set.html#normal) could work but it's too annoying to get it right.

build_tooling/requirements-compatibility-py311.txt
```diff
@@ -0,0 +1,2 @@
+# Makes sure we are able to use Numpy 1
+numpy<2
```
alexowens90, line 2: So what matrix of Python (minor) and Pandas/numpy (major) versions are now covered by the CI?

cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -119,14 +119,16 @@ struct NativeTensor {
     [[nodiscard]] auto expanded_dim() const { return expanded_dim_; }
     template<typename T>
     const T *ptr_cast(size_t pos) const {
+        util::check(ptr != nullptr, "Unexpected null ptr in NativeTensor");
```
alexowens90, line 122: Why is this no longer needed?

G-D-Petrov: It has been added. It is a bit of a sanity check because of there is a incompatibility of the numpy version used at runtime vs compile time, the whole numpy object is misread and this pointer is null

build_tooling/requirements-compatibility-py311.txt
G-D-Petrov: The following is the matrix on Linux based on the latest run in this branch:

| Python Version      | NumPy Version | Pandas Version |
|---------------------|---------------|----------------|
| Python 3.7          | 1.21.6        | 1.3.5          |
| Python 3.8          | 1.24.4        | 2.0.3          |
| Python 3.8 (Compat) | 1.24.4        | 1.5.3          |
| Python 3.9          | 2.0.2         | 2.2.3          |
| Python 3.10         | 2.2.1         | 2.2.3          |
| Python 3.11         | 2.2.1         | 2.2.3          |
| Python 3.11 (Compat)| 1.26.4        | 2.2.3          |
| Python 3.12         | 2.2.1         | 2.2.3          |

There is no Numpy 2 for python 3.8 and below.
Python 3.6 backwards/forwards compat can still be tested when the persistent tests are run.

G-D-Petrov:
> Are you sure that the encoding_version error has been properly understood? The values for the encoding_version are actually 0 (the default) and 1, even though we refer to it as version 1 and 2. It looks like it's complaining it has been given an invalid encoding version

@willdealtry yes, the older version are performing [this check](https://github.com/man-group/ArcticDB/commit/3bc665d930671586fa640c62f33ff07354de2e7c#diff-21f07ddd0b2b8ab0b82fd8392af726ab2a0b5ccafedab67d86497177d4f381a2L24) that is checking specifically for version 0 aka 1

python/arcticdb/version_store/_store.py
G-D-Petrov: `columns is None` was added because in the beginning I wanted to use just the `len` function to check for size but this wasn't working for `ndarray`s. Now it is more of a circuit breaker for the check, no need to check the types and so on if it is none, so it is not strictly needed but it is a bit of on optimization.

As for the tests, I am not sure why they shouldn't pass, they are passing an empty list(`[]`) to the function and that part of the logic is not changed, just now it is covered by `or (isinstance(columns, list) and len(columns) == 0)`

vasil-pashov: Ah, I missed the check bellow
```python
if not implement_read_index and is_columns_empty:
  columns = None
```
it makes sense now

vasil-pashov (approved):

build_tooling/requirements-compatibility-py311.txt
alexowens90: OK cool. Would be nice to have all 4 of numpy 1.x/2.x and pandas 1.x/2.x tested on all Python versions where that makes sense, but that way combinatorial explosions lie, so this seems like a good compromise

willdealtry (approved):

PR #2054, opened 2024-12-10
Query Builder tests for OOM and memory limit (from PR #2053)

grusev:
#### Checklist

<details>
  <summary>
Query Builder tests for OOM and memory limit
  </summary>

The PR contains code that introduces memray and pytest-memray libraries. With their help it is possible to create 2 types of tests:

 - memory leaks test - the memory leaks tests will accept one or two parameters:
   - mem leak threshold - a threshold of mem in KB/MB that will trigger failure if above
   - (optional) filter function - the function can be passed to memray to filterout code that we do not want to be accounted as mem leak. In future we might get some code that we KNOW has memory leaks but is outside of our scope. Thus we can use this function to filter it out.. Currently the code accepts that mem leaks under certain treshold will not lead to failure. If that threshold is vialoted in future this should mean immediate development investigative task. In the log there will be information what is the code that triggered that. IMPORTANT: code source will be if the test is executed against release build - only the size of leak will be correct. To know the source of the code the test must be executed against build with debug information

 - memory limit test - the memory limit test is about enforcing memory efficiency of code. Currently I have set CURRENT memory limit as limit for the this QueryBuilder code. If test fails in future then that would mean that our code is LESS memory efficient in that code. Why that has happened needs to be investigated Also, the number currently set accepts situation AS IS. It does not makes claim that we are memory efficient. No such study has been done it is outside of the test purpose. If such thing is needed then an investigative task need to be created
 
</details>


python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+
+def construct_df_querybuilder_tests(size: int) -> pd.DataFrame:
+    df = get_sample_dataframe(size)
+    date_range = pd.date_range(start="2000-1-1", periods=size, freq="s")
+    df.index = pd.Index(date_range)
+    df.index.name = "timestamp"
+    return df
+
+
+def query_group() -> QueryBuilder:
```
alexowens90, line 177: This name is a bit misleading, as the generated query will do a filter, then a groupby

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+    """
+        groupby composite aggregation query for QueryBuilder memory tests.
+        The query basically will do aggregation of half of dataframe
+    """
+    q = QueryBuilder()
+    return (
+        q[q["bool"] == True]
```
alexowens90, line 184: This line can just be `q[q["bool"]]`

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+
+
+def query_group_bigger_result() -> QueryBuilder:
+    """
+        groupby composite aggregation query for QueryBuilder memory tests.
+        The query basically will do aggregation of whole dataframe
+    """
+    q = QueryBuilder()
+    return (
+        q[q["strings"] != "QASDFGH"]
```
alexowens90, line 197: The chances of this filtering anything out are very low, is that deliberate?

vasil-pashov (changes requested):

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+def query_apply(str:str) -> QueryBuilder:
+    """
+        Apply query for QueryBuilder memory tests.
+        This version of apply does couple of nested operations
+        with columns that were just added
+    """
+    q = QueryBuilder()
+    q = q[q["strings"] != str].apply("NEW 1", q["uint32"] * q["int32"] / q["float32"])
+    q = q.apply("NEW 2", q["float32"] + q["float64"] + q["NEW 1"])
+    q = q.apply("NEW 3", q["NEW 2"] - q["NEW 1"] + q["timestamp"])
```
alexowens90, line 217: It is a bug that this works, we shouldn't allow adding timestamp columns to numeric columns

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -182,16 +310,61 @@ def proc_to_examine():
          run the test from command line again to assure it runs ok before commit 
 
     """
-    max_mem_bytes = 450_623_040
+    max_mem_bytes = 295_623_040
+
+    check_process_memory_leaks(proc_to_examine, 20, max_mem_bytes, 80.0)
+
+@pytest.mark.skipif(WINDOWS, reason="Not enough storage on Windows runners")
+@pytest.mark.skipif(MACOS, reason="Problem on MacOs")
```
alexowens90, line 318: More descriptive reason please

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -182,16 +310,61 @@ def proc_to_examine():
...
+    size = df.shape[0]
+    start_date = df.iloc[0].name
+    end_date = df.iloc[0].name
+
+    symbol = "test"
+    lib.write(symbol, df)
+    del df
+    gc.collect()
+
+    queries = [query_group(), query_group_bigger_result(), query_apply(random_string(10)), query_resample()]
```
alexowens90, line 339: Seems to be unused?

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -182,16 +310,61 @@ def proc_to_examine():
...
+
+    def proc_to_examine():
+        queries = [query_group(), 
+                    query_group_bigger_result(), 
+                    query_apply(random_string(10)), 
+                    query_resample(),
+                    query_row_range(size),
+                    query_date_range(start_date, end_date)]
```
alexowens90, line 347: Because of the randomness involved, the memory usage of `query_row_range` and `query_date_range` will not be stable. They could be modified to return a fixed percentage of the input dataframe to solve this.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+        for frame in stack.frames:
+            # do something to check if we need this to be added
+            # as mem leak
+            # print(f"SAMPLE >>> {frame.filename}:{frame.function}[{frame.lineno}]")
+            pass
+        return True
+
+    loc_limit="25 KB"
+    if (MACOS):
+        loc_limit = "35 KB"
```
alexowens90, line 492: Odd lack of newlines

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+    def test_mem_leak_querybuilder_read_memray(library_with_symbol):
+        """
+            Test to capture memory leaks >= of specified number
+
+            NOTE: we could filter out not meaningful for us stackframes
+            in future if something outside of us start to leak using
+            the argument "filter_fn" - just add to the filter function
+            what we must exclude from calculation
+        """
+        (lib, df) = library_with_symbol
```
alexowens90, line 504: This is an odd style. The pattern we use elsewhere is to use the `arctic_library_lmdb` fixture in this test, and then something like:
```
lib = arctic_library_lmdb
lib.write(symbol, construct_df_querybuilder_tests(size=2000000))
```

alexowens90: If it interferes with `memray`, then this could be a separate function that is excluded by `is_relevant`

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+        """
+            As memray instruments memory, we need to take out 
+            everything not relevant from mem leak measurement out of 
+            test, so it works as less as possible
+        """
+        lib: Library = arctic_library_lmdb
+        df = generate_big_dataframe(300000)
+        lib.write(symbol, df)
+        del df
+        yield lib
```
alexowens90, line 561: Prefer using `get_sample_dataframe` everywhere than these custom fixtures

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+        """
+
+        print("Test starting")
+        st = time.time()
+        lib: Library = library_with_big_symbol
+        data = lib.read(symbol).data
+        del data
+        print("Test took :", time.time() - st)
+
+        print("Sleeping for 10 secs")
```
alexowens90, line 579: Why?

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+        As memray instruments memory, we need to take out 
+        everything not relevant from mem leak measurement out of 
+        test, so it works as less as possible
+    """
+    lib: Library = arctic_library_lmdb
+    df = construct_df_querybuilder_tests(size=2000000)
+    lib.write(symbol, df)
+    yield (lib, df)
+
+def mem_query(lib: Library, df: pd.DataFrame, num_repetitions:int=1, read_batch:bool=False):
```
alexowens90, line 409: `pytest-memray` docs recommend running in a loop, should we set `num_repetitions>1` for fast tests?

alexowens90 (changes requested):

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
         df_to_grow = pd.concat([df_to_grow, df_prev])
     return df_to_grow
 
+def generate_big_dataframe(rows:int=1000000):
+    print("Generating big dataframe")
+    st = time.time()
+    df = get_sample_dataframe(rows)
+    df = grow_exp(df, 5)
```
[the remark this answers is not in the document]
grusev: Generating milions of lines with get_sample_dataframe() is perhaps slower than generating 10k, then appending same numbur and to result appending the result from previous op. I think that was the reason I tried to speed up things.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: yes, this is deliberate to impose attempt to filter by impossible condition which should still be evaluated. And as we know with strings things are always buggier than with numbers

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: Then that is tricky thing - if that is released it is a feature :-) especially if someone lik me has found that it works. So now we could have potentially lots of usages (or none)

Anyway I can and will change that, but perhaps I should create an issue on that to discuss this ?

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: generate_big_dataframe(100000)
Generation took : 6.429168701171875
df = get_sample_dataframe(16000000) #This is equivalent to generate_big_dataframe(100000)
Time:  49.78162121772766

Quite huge difference. 8x times and for larger DFs would be more. Thus for mem leaks etc time wize generation should be preferred

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: nope it is not related with is_relevant ... It is related with the way to use fixtures.

In arcticdb you actually have fixtures that setup environment. But you have accepted that environment you setup is simple one thing. In many cases the environment will be more than one thing. In this case BOTH enormaous dataframe and empty library.

In python it is common and OK to return more than one value. In our case it is needed. We have to prepare both library and dataframe before the actual execution of the test which memray will instrument. Thus both things must hapen in the pre-setup and this is nice way both to happen then. Can they be separate fixtures? Yes, pytest allows that ...

But the test logic behind test fixture is different. They provide a way to do _both_ setup and cleanup afterwards. That is the reason this approach was chosen. This way the test is linked exactly with this fixture which intentionaly if not yet part of library with other fixtures. It is very specialized fixture that in future might evolve.

I hope this justifies ;eaving that fixture as is.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: I do not think gc.collect() is synchronous task. Or at least my observations with psutil library are different. Memory is not released synchronous to the call or at least not all of it. Thus the decision for flocking sleep. 
That gives oppertunity additional release process to kick in. 
Again this is from observation with other approach (with psutils, prior to memray) where I could not make it relativly stable unless sleep is introduced. Again this sleep is only for those kind of tests it is not clobal one, thus it can stay. If removed I do not know what the results could be

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: mem leaks will be revealed over long running repeated processes indeed. As our dafarame is >16million all queries are likely to trigger extreme number of repetitions in C++ that should effectivley achieve that.

Alternative is to have smaller dataframe 30-50k symbols and repeat queries many times. This test will capture other types of mem leaks perhaps. 

But without knowing the code my inclination would be that more serious would be the leaks that current test will capture.

But at the same time it is really cheap to add now additional tests with more repetitions. That will only cost couple of more minutes time if no problems arise

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -201,6 +374,7 @@ def proc_to_examine():
         del data
 
     df = lib.read(symbol).data
+    print(df)
```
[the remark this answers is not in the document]
grusev: this for debug pruposes - removed now

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+symbol = "test"
+
+
+@pytest.fixture
+def library_with_symbol(arctic_library_lmdb) -> Generator[Tuple[Library, pd.DataFrame], None, None]:
+    """
+        As memray instruments memory, we need to take out 
+        everything not relevant from mem leak measurement out of 
+        test, so it works as less as possible
+    """
```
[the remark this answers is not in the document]
grusev: """
    As memray instruments memory, we need to take out 
    everything not relevant from mem leak measurement out of 
    test, and place it in setup of the test - in other words in the fixture 
    Otherwise, memray instruments that code and it results in much slower execution
    as well as mixing results of memory leaks - are they in what we test - for example
    read() or in construct_df_querybuilder_tests() ? or in other code?
    """

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+
+
+    @pytest.fixture
+    def library_with_big_symbol(arctic_library_lmdb) -> Generator[Library, None, None]:
+        """
+            As memray instruments memory, we need to take out 
+            everything not relevant from mem leak measurement out of 
+            test, so it works as less as possible
+        """
+        lib: Library = arctic_library_lmdb
```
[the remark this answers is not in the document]
grusev: ommissions, or somewhere not needed

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -182,16 +310,61 @@ def proc_to_examine():
...
     """
-    max_mem_bytes = 450_623_040
+    max_mem_bytes = 295_623_040
+
+    check_process_memory_leaks(proc_to_examine, 20, max_mem_bytes, 80.0)
+
+@pytest.mark.skipif(WINDOWS, reason="Not enough storage on Windows runners")
+@pytest.mark.skipif(MACOS, reason="Problem on MacOs")
+@SKIP_CONDA_MARK # Conda CI runner doesn't have enough storage to perform these stress tests
+def test_mem_leak_querybuilder_standard(arctic_library_lmdb):
```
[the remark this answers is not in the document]
grusev: That was intentionally - there are certains specifics between memray and standard approach and thus sharing same code could be problematic. Therefore intentionally I keep this code duplicated as the life of 2 approaches would be different

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
         df_to_grow = pd.concat([df_to_grow, df_prev])
     return df_to_grow
 
+def generate_big_dataframe(rows:int=1000000):
+    print("Generating big dataframe")
```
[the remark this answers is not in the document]
grusev: done for functions and tests related to memray

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
alexowens90: My issue is that the condition is not impossible, just unlikely. Making it actually impossible is fine (e.g. use a different character set than the generated dataframe), but we have optimisations in place for when nothing is filtered out, which could then skew the results if you get unlucky and the dataframe does contain `QASDFGH`

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
alexowens90: Sure, add a low-priority bug to Monday

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
alexowens90: I think both a single iteration with a large dataframe, and many iterations with smaller dataframes have value, as they could expose different styles of leak.
e.g. if we leak 1kb of memory every time we do a read with a filter, this will only be caught by running many filters

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
alexowens90: There's nothing in the [docs](https://docs.python.org/3/library/gc.html#gc.collect) suggesting this is asynchronous. What `psutil` stats were you comparing to?

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
alexowens90: Ah, I see that `filter_fn` is only an argument to `limit_leaks`, not `limit_memory`. I guess this is OK if fixtures are the only way to get memray to not track that memory then

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: Yes. It is very unlikely. Thus making the results 99.99% of the dataframe (note that by default I think we generate str of 10 chars array and this is 7 chars. So perhaps really impossible to achieve unless bug

Anyway for our purposes we do not need exactness at all. Thus actual returned result if it is 99 or 100% of rows really does not matter

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: makes sense - I will modify the code to have only start row/date random between first 40% of the rows and then return some persenatge of rows approx 57 for instance

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: will change!

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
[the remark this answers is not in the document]
grusev: I will ... btw ... actually in many cases accept that reasoning comes from experimentation. Like in that case You cannot predict how much memory is going to leak thus you have to make something large enough to both catch a problem that we had with massive leakage, and smaller ones, but at the same time keep time for execution reasonable :-)

Thus this code is needed for not memray approach based on repetitions, measurments with psutil (which turn out to be relative because gc.collect() will not do garbage collection, or even if it do it it will be in python code only and have no control on C++ part) Thus those types of tests heuristics based on knowledge and observation of the creator.

But at the same time as memray does not support windows having this approach avail is nice. Even that our Windows runners have less HDD memory than needed for the test to execute. :-(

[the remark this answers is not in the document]
grusev: def generate_big_dataframe(rows:int=1000000, num_exp_time_growth:int=5) -> pd.DataFrame:
    """
    A quick and time efficient wat to generate very large dataframe.
    The first parameter will be passed to get_sample_dataframe() so that a dataframe
    of that number of rows is generated. Later this df will be used to grow exponentially (power of 2)
    the number of rows on the final dataframe by concatenating N times the result dataframe to itself
    That process if >6-8 times more time efficient. The down side is that you have N**2 repetitions of
    same original table. But for many usages size is what we need not uniqueness
    """

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: https://man312219.monday.com/boards/7852509418/pulses/8065794446

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: Will add test

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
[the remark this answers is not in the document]
grusev: there is an issue opened already on that.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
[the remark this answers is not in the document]
grusev: for now yes - for future we do not know. Thus leaveing that in code should be ok

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -182,16 +310,61 @@ def proc_to_examine():
...
+
+    def proc_to_examine():
+        queries = [query_group(), 
+                    query_group_bigger_result(), 
+                    query_apply(random_string(10)), 
+                    query_resample(),
+                    query_row_range(size),
+                    query_date_range(start_date, end_date)]
+        for q in queries:
+            data: pd.DataFrame = lib.read(symbol, query_builder=q).data
```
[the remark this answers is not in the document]
grusev: I am hesitant to mix those 2 topics, I will add some kind of check that it is relatively correct, ie like we get enough rows but I do not want to mix those 2 types of tests especially with huge symbols/dataframes like this

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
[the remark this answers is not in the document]
grusev: rem oved

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+                                            ) for query in queries]
+                results_read = lib.read_batch(read_requests)
+                cnt = 0
+                for result in results_read:
+                    assert not result.data is None
+                    print_info(result.data, queries[cnt])
+                    cnt += 1
+                del read_requests, results_read
+            else:
+                print ("RUN read_batch() tests")
```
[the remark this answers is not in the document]
grusev: fixed

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -182,16 +310,61 @@ def proc_to_examine():
...
+            del data
+            gc.collect()
+
+    max_mem_bytes = 550_623_040
+
+    check_process_memory_leaks(proc_to_examine, 10, max_mem_bytes, 80.0)
+
+    del lib, queries
+    gc.collect()
+    time.sleep(10)
```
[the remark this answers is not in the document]
grusev: See how psutils sees memory without sleep and with sleep. Ovviously there are processes that require 'calm' process in order for external tools on which psutility is based to measure properly memory.

Thus having NO sleep it is rather road to flaky tests - I have observed that already.

-----------------------------------------------------------------------------
  WITHOUT WAITS 10 SEC
-----------------------------------------------------------------------------

Process initial RSS  1700.16MB/[1782743040] 
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[0]    : Process added (or if number is negative - process released)  -36.97MB/[-38768640] .  Avail memory:  43992.38MB/[46129352704]  Used Mem: 8.7%
  Overall stats : Process growth since start  -36.97MB/[-38768640]  AVG growth per iter  -36.97MB/[-38768640.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -36.97MB/[-38768640] 
  Number of times there was 50% drop in memory: 0
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[1]    : Process added (or if number is negative - process released)  0.17MB/[180224] .  Avail memory:  44074.61MB/[46215573504]  Used Mem: 8.5%
  Overall stats : Process growth since start  -36.80MB/[-38588416]  AVG growth per iter  -18.40MB/[-19294208.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -36.80MB/[-38588416] 
  Number of times there was 50% drop in memory: 1
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[2]    : Process added (or if number is negative - process released)  13.77MB/[14442496] .  Avail memory:  44088.59MB/[46230241280]  Used Mem: 8.5%
  Overall stats : Process growth since start  -23.03MB/[-24145920]  AVG growth per iter  -7.68MB/[-8048640.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -23.03MB/[-24145920] 
  Number of times there was 50% drop in memory: 2
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[3]    : Process added (or if number is negative - process released)  3.79MB/[3977216] .  Avail memory:  44066.72MB/[46207307776]  Used Mem: 8.5%
  Overall stats : Process growth since start  -19.23MB/[-20168704]  AVG growth per iter  -4.81MB/[-5042176.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -19.23MB/[-20168704] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[4]    : Process added (or if number is negative - process released)  43.75MB/[45879296] .  Avail memory:  44020.70MB/[46159044608]  Used Mem: 8.6%
  Overall stats : Process growth since start  24.52MB/[25710592]  AVG growth per iter  4.90MB/[5142118.4] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  24.52MB/[25710592] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[5]    : Process added (or if number is negative - process released)  103.92MB/[108969984] .  Avail memory:  43923.30MB/[46056919040]  Used Mem: 8.8%
  Overall stats : Process growth since start  128.44MB/[134680576]  AVG growth per iter  21.41MB/[22446762.666666668] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  128.44MB/[134680576] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[6]    : Process added (or if number is negative - process released)  -22.30MB/[-23379968] .  Avail memory:  43941.50MB/[46076002304]  Used Mem: 8.8%
  Overall stats : Process growth since start  106.14MB/[111300608]  AVG growth per iter  15.16MB/[15900086.857142856] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  128.44MB/[134680576] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[7]    : Process added (or if number is negative - process released)  -138.72MB/[-145457152] .  Avail memory:  44065.91MB/[46206459904]  Used Mem: 8.5%
  Overall stats : Process growth since start  -32.57MB/[-34156544]  AVG growth per iter  -4.07MB/[-4269568.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  128.44MB/[134680576] 
  Number of times there was 50% drop in memory: 4
  

-----------------------------------------------------------------------------
  WITH WAITS 10 SEC
-----------------------------------------------------------------------------
Process initial RSS  1734.86MB/[1819131904] 
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[0]    : Process added (or if number is negative - process released)  45.19MB/[47386624] .  Avail memory:  43974.38MB/[46110478336]  Used Mem: 8.7%
  Overall stats : Process growth since start  45.19MB/[47386624]  AVG growth per iter  45.19MB/[47386624.0] 
  Minimum growth so far:  45.19MB/[47386624] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 0
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[1]    : Process added (or if number is negative - process released)  -7.00MB/[-7344128] .  Avail memory:  43968.44MB/[46104248320]  Used Mem: 8.7%
  Overall stats : Process growth since start  38.19MB/[40042496]  AVG growth per iter  19.09MB/[20021248.0] 
  Minimum growth so far:  38.19MB/[40042496] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 0
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[2]    : Process added (or if number is negative - process released)  -82.82MB/[-86847488] .  Avail memory:  44054.60MB/[46194597888]  Used Mem: 8.6%
  Overall stats : Process growth since start  -44.64MB/[-46804992]  AVG growth per iter  -14.88MB/[-15601664.0] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 1
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[3]    : Process added (or if number is negative - process released)  88.26MB/[92545024] .  Avail memory:  43922.98MB/[46056579072]  Used Mem: 8.8%
  Overall stats : Process growth since start  43.62MB/[45740032]  AVG growth per iter  10.91MB/[11435008.0] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 1
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[4]    : Process added (or if number is negative - process released)  -82.00MB/[-85979136] .  Avail memory:  44019.06MB/[46157332480]  Used Mem: 8.6%
  Overall stats : Process growth since start  -38.38MB/[-40239104]  AVG growth per iter  -7.67MB/[-8047820.8] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 2
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[5]    : Process added (or if number is negative - process released)  42.20MB/[44249088] .  Avail memory:  43991.38MB/[46128308224]  Used Mem: 8.7%
  Overall stats : Process growth since start  3.82MB/[4009984]  AVG growth per iter  0.64MB/[668330.6666666666] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 2
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[6]    : Process added (or if number is negative - process released)  -58.56MB/[-61403136] .  Avail memory:  44048.85MB/[46188564480]  Used Mem: 8.6%
  Overall stats : Process growth since start  -54.73MB/[-57393152]  AVG growth per iter  -7.82MB/[-8199021.714285715] 
  Minimum growth so far:  -54.73MB/[-57393152] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 10 secs so GC to kick in

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: Well reallity is different.

See how psutils sees memory without sleep and with sleep. Ovviously there are processes that require 'calm' process in order for external tools on which psutility is based to measure properly memory.

Thus having NO sleep it is rather road to flaky tests - I have observed that already with standard approach. Is it going to repeat itself with memray tests I do not know, but it is better to pay the price of waiting. Remember patience is a virtue ;-) In our case it might produce lots of flakyness and for what?


-----------------------------------------------------------------------------
  WITHOUT WAITS 10 SEC
-----------------------------------------------------------------------------

Process initial RSS  1700.16MB/[1782743040] 
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[0]    : Process added (or if number is negative - process released)  -36.97MB/[-38768640] .  Avail memory:  43992.38MB/[46129352704]  Used Mem: 8.7%
  Overall stats : Process growth since start  -36.97MB/[-38768640]  AVG growth per iter  -36.97MB/[-38768640.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -36.97MB/[-38768640] 
  Number of times there was 50% drop in memory: 0
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[1]    : Process added (or if number is negative - process released)  0.17MB/[180224] .  Avail memory:  44074.61MB/[46215573504]  Used Mem: 8.5%
  Overall stats : Process growth since start  -36.80MB/[-38588416]  AVG growth per iter  -18.40MB/[-19294208.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -36.80MB/[-38588416] 
  Number of times there was 50% drop in memory: 1
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[2]    : Process added (or if number is negative - process released)  13.77MB/[14442496] .  Avail memory:  44088.59MB/[46230241280]  Used Mem: 8.5%
  Overall stats : Process growth since start  -23.03MB/[-24145920]  AVG growth per iter  -7.68MB/[-8048640.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -23.03MB/[-24145920] 
  Number of times there was 50% drop in memory: 2
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[3]    : Process added (or if number is negative - process released)  3.79MB/[3977216] .  Avail memory:  44066.72MB/[46207307776]  Used Mem: 8.5%
  Overall stats : Process growth since start  -19.23MB/[-20168704]  AVG growth per iter  -4.81MB/[-5042176.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  -19.23MB/[-20168704] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[4]    : Process added (or if number is negative - process released)  43.75MB/[45879296] .  Avail memory:  44020.70MB/[46159044608]  Used Mem: 8.6%
  Overall stats : Process growth since start  24.52MB/[25710592]  AVG growth per iter  4.90MB/[5142118.4] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  24.52MB/[25710592] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[5]    : Process added (or if number is negative - process released)  103.92MB/[108969984] .  Avail memory:  43923.30MB/[46056919040]  Used Mem: 8.8%
  Overall stats : Process growth since start  128.44MB/[134680576]  AVG growth per iter  21.41MB/[22446762.666666668] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  128.44MB/[134680576] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[6]    : Process added (or if number is negative - process released)  -22.30MB/[-23379968] .  Avail memory:  43941.50MB/[46076002304]  Used Mem: 8.8%
  Overall stats : Process growth since start  106.14MB/[111300608]  AVG growth per iter  15.16MB/[15900086.857142856] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  128.44MB/[134680576] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 0 secs so GC to kick in
Iter No[7]    : Process added (or if number is negative - process released)  -138.72MB/[-145457152] .  Avail memory:  44065.91MB/[46206459904]  Used Mem: 8.5%
  Overall stats : Process growth since start  -32.57MB/[-34156544]  AVG growth per iter  -4.07MB/[-4269568.0] 
  Minimum growth so far:  -36.97MB/[-38768640] 
  Maximum growth so far:  128.44MB/[134680576] 
  Number of times there was 50% drop in memory: 4
  

-----------------------------------------------------------------------------
  WITH WAITS 10 SEC
-----------------------------------------------------------------------------
Process initial RSS  1734.86MB/[1819131904] 
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[0]    : Process added (or if number is negative - process released)  45.19MB/[47386624] .  Avail memory:  43974.38MB/[46110478336]  Used Mem: 8.7%
  Overall stats : Process growth since start  45.19MB/[47386624]  AVG growth per iter  45.19MB/[47386624.0] 
  Minimum growth so far:  45.19MB/[47386624] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 0
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[1]    : Process added (or if number is negative - process released)  -7.00MB/[-7344128] .  Avail memory:  43968.44MB/[46104248320]  Used Mem: 8.7%
  Overall stats : Process growth since start  38.19MB/[40042496]  AVG growth per iter  19.09MB/[20021248.0] 
  Minimum growth so far:  38.19MB/[40042496] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 0
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[2]    : Process added (or if number is negative - process released)  -82.82MB/[-86847488] .  Avail memory:  44054.60MB/[46194597888]  Used Mem: 8.6%
  Overall stats : Process growth since start  -44.64MB/[-46804992]  AVG growth per iter  -14.88MB/[-15601664.0] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 1
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[3]    : Process added (or if number is negative - process released)  88.26MB/[92545024] .  Avail memory:  43922.98MB/[46056579072]  Used Mem: 8.8%
  Overall stats : Process growth since start  43.62MB/[45740032]  AVG growth per iter  10.91MB/[11435008.0] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 1
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[4]    : Process added (or if number is negative - process released)  -82.00MB/[-85979136] .  Avail memory:  44019.06MB/[46157332480]  Used Mem: 8.6%
  Overall stats : Process growth since start  -38.38MB/[-40239104]  AVG growth per iter  -7.67MB/[-8047820.8] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 2
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[5]    : Process added (or if number is negative - process released)  42.20MB/[44249088] .  Avail memory:  43991.38MB/[46128308224]  Used Mem: 8.7%
  Overall stats : Process growth since start  3.82MB/[4009984]  AVG growth per iter  0.64MB/[668330.6666666666] 
  Minimum growth so far:  -44.64MB/[-46804992] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 2
Starting watched code ...........
Lets pause for 10 secs so GC to kick in
Iter No[6]    : Process added (or if number is negative - process released)  -58.56MB/[-61403136] .  Avail memory:  44048.85MB/[46188564480]  Used Mem: 8.6%
  Overall stats : Process growth since start  -54.73MB/[-57393152]  AVG growth per iter  -7.82MB/[-8199021.714285715] 
  Minimum growth so far:  -54.73MB/[-57393152] 
  Maximum growth so far:  45.19MB/[47386624] 
  Number of times there was 50% drop in memory: 3
Starting watched code ...........
Lets pause for 10 secs so GC to kick in

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+
+def query_date_range(start:pd.Timestamp, end: pd.Timestamp) -> QueryBuilder:
+    """
+        Date range query for QueryBuilder memory tests
+        Will generate random date range query based on given
+        start and end timestamps/dates
+    """
+    q = QueryBuilder()
+    a = gen_random_date(start,end)
+    b = gen_random_date(start,end)
```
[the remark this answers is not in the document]
grusev: code is different now

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+
+    queries = [query_group(), 
+                query_group_bigger_result(), 
+                query_apply(random_string(10)), 
+                query_resample(),
+                query_row_range(size),
+                query_date_range(start_date, end_date)]
+    
+    for rep in range(num_repetitions):
+            if (read_batch):
```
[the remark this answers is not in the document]
grusev: read() and read_batch() have common way of doing things. When we plan to add more tests for other functionalities we can apply suggested approach if necesssary, currently is not needed to do changes as we just speculate that there will be additional tests and they will reuse same function. Therefore I think this comment should be done when actual overusage occur, not now

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+#region TESTS pytest-memray type for memory limit and leaks
+
+## NOTE: Currently tests can be executed on Python >= 3.8 only
+
+
+symbol = "test"
+
+
+@pytest.fixture
+def library_with_symbol(arctic_library_lmdb) -> Generator[Tuple[Library, pd.DataFrame], None, None]:
```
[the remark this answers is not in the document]
grusev: fixed using different approach fixture is a setup for a test, thus it is responsible to return what has been setup, including the name of symbol

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: And actually there seems to be a problem https://man312219.monday.com/boards/7852509418/pulses/8067881190
I will xfail and push the code

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
alexowens90: The name still doesn't mention filtering

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
alexowens90: It does matter. Returning 100% and 99.999% of a dataframe can take totally different code paths

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +410,249 @@ def proc_to_examine():
...
+        of queries against a large dataframe. Later this
+        function will be used for memory limit and memory leaks 
+        tests
+    """
+    size = df.shape[0]
+    start_date = df.index[0]
+    end_date = df.index[-1]
+
+    symbol = "test"
+    queries = []
```
alexowens90, line 464: Unnecessary

alexowens90 (changes requested):

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
vasil-pashov: Reading the comment above I don't understand the purpose of this. What do we want to achieve? Do we want to

1. Filter nothing at all and get an empty data frame?
2. Filter small part of the data frame?

If 1 is the desired option what Alex suggested i.e. using a different charset seems like a better idea.
If 2 is the desired option we should use a string that we're sure filters small amount of data.

Given that randomness is also involved I think there are some unknowns in the tests. @alexowens90 what is your opinion?

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +154,132 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+    return df
+
+
+def query_group() -> QueryBuilder:
+    """
+        groupby composite aggregation query for QueryBuilder memory tests.
+        The query basically will do aggregation of half of dataframe
+    """
+    q = QueryBuilder()
+    return (
```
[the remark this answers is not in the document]
vasil-pashov: I don't think you need then in this case. The function quoted above `query_resample` doesn't use brackets. We should keep the style homogeneous.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +160,158 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+    q = q[q["strings"] != strng].apply("NEW 1", q["uint32"] * q["int32"] / q["float32"])
+    q = q.apply("NEW 2", q["float32"] + q["float64"] + q["NEW 1"])
+    q = q.apply("NEW 3", q["NEW 2"] - q["NEW 1"] + q["timestamp"])
+    return q
+
+
+def query_resample() -> QueryBuilder:
+    """
+        Resample query for QueryBuilder memory tests
+    """
```
vasil-pashov, line 243: Indentation
```suggestion
    """
    Resample query for QueryBuilder memory tests
    """
```

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +160,158 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+        )
+
+
+def query_row_range(size: int) -> QueryBuilder:
+    """
+    Row range query for QueryBuilder memory tests
+    Pass size of dataframe and it will generate random row range
+    """
+    percentage_rows_returned = 0.57
+    start_percentage = random.uniform(0.01, 1.0 - percentage_rows_returned - 0.01)
```
vasil-pashov, line 264: Why are you subtracting 0.01? FYI it is safe to pass row_range(0, size) because the row_range is inclusive of the low end and exclusive of the upper end. https://docs.arcticdb.io/latest/api/library/#arcticdb.version_store.library.Library.read
Subtracting one on row 267 is not necessary

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +160,158 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+
+
+def query_date_range(start: pd.Timestamp, end: pd.Timestamp) -> QueryBuilder:
+    """
+    Date range query for QueryBuilder memory tests
+    Will generate random date range that will return 
+    always the specified percentage rows
+    """
+    percentage_rows_returned = 0.57
+    start_percentage = random.uniform(0.01, 1.0 - (percentage_rows_returned + 0.01))
```
vasil-pashov, line 280: Same as https://github.com/man-group/ArcticDB/pull/2054/files#r1889962631

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -182,16 +341,58 @@ def proc_to_examine():
...
+                    reason="Not enough storage on Windows runners, due to large Win OS footprint and less free mem")
+@pytest.mark.skipif(MACOS, reason="Problem on MacOs most probably similar to WINDOWS")
+@SKIP_CONDA_MARK # Conda CI runner doesn't have enough storage to perform these stress tests
+def test_mem_leak_querybuilder_standard(arctic_library_lmdb):
+    """
+        This test uses old approach with iterations.
+        It is created for comparison with the new approach
+        with memray
+        (If memray is good in future we could drop the old approach)
+    """
```
vasil-pashov, line 358:
```suggestion
    """
    This test uses old approach with iterations.
    It is created for comparison with the new approach
    with memray
    (If memray is good in future we could drop the old approach)
    """
```

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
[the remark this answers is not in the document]
vasil-pashov: I think we're laying bad foundations with this and I'm against it. Please see if @alexowens90 is OK with this, if so we'll merge it.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: Lets accept the way it is. If you want me to add additional query I will no problem, but the way I have constructed this is meant to remain

vasil-pashov: Let's see what @alexowens90 thinks

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -209,3 +383,201 @@ def proc_to_examine():
...
+    def library_with_big_symbol(arctic_library_lmdb) -> Generator[Library, None, None]:
+        """
+            As memray instruments memory, we need to take out 
+            everything not relevant from mem leak measurement out of 
+            test, so it works as less as possible
+        """
+        lib: Library = arctic_library_lmdb
+        df = generate_big_dataframe(300000)
+        lib.write(symbol, df)
+        del df
```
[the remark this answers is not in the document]
vasil-pashov: I think we should have good understanding of the code we're putting both in theory and in practice. It's unusual to delete stuff in python in such way. We should prove there is a reason to do this and document why it is needed. If not we should omit it.

Let's see what @alexowens90's take on this as well.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: added aditional query also without any filter. Thus we have now 3 queries that have groupby agg., with filter(50%) filter(100%) effectivly and no filter ...

there is additional functional test also that covers the case if there is something fishy with queries

alexowens90: @grusev the query needs to be changed to one of:

- Return 100% of the dataframe
- Return <100% of the dataframe

And it has to be consistent. We have optimisations in place so that a filter that doesn't remove anything is much more efficient than a filter that removes even 1 row.

grusev: only the comment is misleading but all is ok

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +160,174 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+        )
+
+
+def query_row_range_57percent(size: int) -> QueryBuilder:
+    """
+    Row range query for QueryBuilder memory tests
+    Pass size of dataframe and it will generate random row range
+    """
+    percentage_rows_returned = 0.57
+    start_percentage = random.uniform(0.01, 1.0 - percentage_rows_returned)
```
vasil-pashov, line 280: This will probably disable starting from 0-th row. Is this the desired behavior?

The same question applies to `query_date_range_57percent`

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
```diff
@@ -146,13 +160,174 @@ def grow_exp(df_to_grow: pd.DataFrame, num_times_xx2: int):
...
+    result_size_rows = int(0.57 * size)
+    q = QueryBuilder()
+    a = random.randint(0,int((size-1) * start_percentage))
```
vasil-pashov, line 283: 1. As the row range is closed on the left side and opened on the right side it is safe to pass size here. It's also the correct thing to do because the current thing represents a the dataframe without it's last row.
2. Why is the conversion to int needed. Looking at the type hint size is already an integer.

alexowens90 (approved):

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: There is no need to be exact in testing, thus nothing will change in both cases start or not from 0

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
grusev: We do not want exactness or correctness here, we want randomness, nothing changes for the test if it includes or not the last row. If we wanted to cover exactly last row we would not use at all random.

python/tests/stress/arcticdb/version_store/test_mem_leaks.py
vasil-pashov: I have the same comment as: https://github.com/man-group/ArcticDB/pull/2054/files#r1907343843

vasil-pashov (approved):

PR #2058, opened 2024-12-10
Create slack notification step for master builds

jamesmunro:
From here: https://github.com/integrations/slack/issues/1563#issuecomment-1588009077

Uses this integration: https://github.com/ravsamhq/notify-slack-action

### DIFF
--- a/.github/workflows/failure_notification.yaml
+++ b/.github/workflows/failure_notification.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.workflow_run.conclusion == 'failure' || github.event.workflow_run.conclusion == 'timed_out'
     steps:
-      - uses: ravsamhq/notify-slack-action@v2
+      - uses: ravsamhq/notify-slack-action@be814b201e233b2dc673608aa46e5447c8ab13f2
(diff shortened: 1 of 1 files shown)


maxim-morozov (approved):

.github/workflows/failure_notification.yaml
```diff
@@ -0,0 +1,20 @@
+name: Check for master failure
+on:
+  workflow_run:
+    workflows: [Build and Test]
```
poodlewars, line 4: Why not the Conda or Analysis builds?

poodlewars:
Needs test evidence showing a notification for an example branch

.github/workflows/failure_notification.yaml
jamesmunro: Yes, can do.  Just wanted to start with something.

poodlewars (approved):

jamesmunro:
> Needs test evidence showing a notification for an example branch
It's not possible to test this type of trigger outside the default branch.
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run
image

PR #2060, opened 2024-12-11
Fix flaky asv append test

grusev:
#### Checklist

<details>
  <summary>
   FIX Flaky asv append test 
  </summary>
  
The append dataframes are now generated in setup_cache() function, Each new append dataframe has now start segment that is after the previous one. 

As the append dfs are now generated only once the totoal time is lower. Each test now runs 3 instead of 2 times and eventually this should flatten the graphics also

NOTE: Due to the fact that setup_cache() now returns result, there is one additional parameter that (that result) that is added to signature of other functions. Only append functions are using it for now.

ASV time testing in more details with examples for ArcticDB: https://github.com/man-group/ArcticDB/wiki/Running-ASV-Benchmarks

raceback (most recent call last):

File "/opt/python/cp36-cp36m/lib/python3.6/site-packages/asv/benchmark.py", line 1293, in main_run_server

main_run(run_args)

File "/opt/python/cp36-cp36m/lib/python3.6/site-packages/asv/benchmark.py", line 1167, in main_run

result = benchmark.do_run()

File "/opt/python/cp36-cp36m/lib/python3.6/site-packages/asv/benchmark.py", line 573, in do_run

return self.run(*self._current_params)

File "/opt/python/cp36-cp36m/lib/python3.6/site-packages/asv/benchmark.py", line 673, in run

min_run_count=self.min_run_count)

File "/opt/python/cp36-cp36m/lib/python3.6/site-packages/asv/benchmark.py", line 726, in benchmark_timing

timing = timer.timeit(number)

File "/opt/python/cp36-cp36m/lib/python3.6/timeit.py", line 178, in timeit

timing = self.inner(it, self.timer)

File "", line 6, in inner

File "/opt/python/cp36-cp36m/lib/python3.6/site-packages/asv/benchmark.py", line 627, in

func = lambda: self.func(*param)

File "/__w/ArcticDB/ArcticDB/python/benchmarks/basic_functions.py", line 309, in time_append_large

self.lib.append(f"sym", self.df_append_large)

File "/__w/ArcticDB/ArcticDB/python/.asv/env/541b6fbd95fdb429e84621bc981fbd19/lib/python3.6/site-packages/arcticdb/version_store/library.py", line 744, in append

validate_index=validate_index,

File "/__w/ArcticDB/ArcticDB/python/.asv/env/541b6fbd95fdb429e84621bc981fbd19/lib/python3.6/site-packages/arcticdb/version_store/_store.py", line 711, in append

[...]

PR #2065, opened 2024-12-13
Fix asv problems

G-D-Petrov:
#### Reference Issues/PRs
Fixes problems with publishing of ASV benchmarks

This PR implements the following:
- Upgrade the Python analysis flow and C++ tests to Python 3.11 - this is needed because some of the benchmarks need at least 3.10
- Fixes the problems with publishing of the benchmarks - because they can't evaluate np.inf correctly
- Fixes LFS for BI benchmarks
- Incorporates changes from https://github.com/man-group/ArcticDB/pull/2060

See successful run [here](https://github.com/man-group/ArcticDB/actions/runs/12317667688/job/34380720877) (had to be started manually due to the changes to the workflow files)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/benchmarks/finalize_staged_data.py
+++ b/python/benchmarks/finalize_staged_data.py
@@ -13,12 +13,14 @@ from arcticdb.util.utils import CachedDFGenerator, TimestampNumber, stage_chunks
 from arcticdb.version_store.library import Library, StagedDataFinalizeMethod
 from .common import *
 
+
 class FinalizeStagedData:
-    '''
-        Check and benchmark performance of finalize_staged_data().
-        Due to specifics of this procedure we tune asv to make single measurement
-        which would be over a relatively big staged data.
-    '''
+    """
+    Check and benchmark performance of finalize_staged_data().
+    Due to specifics of this procedure we tune asv to make single measurement
+    which would be over a relatively big staged data.
+    """
+
     number = 1
     rounds = 1
     repeat = 1
@@ -27,7 +29,7 @@ class FinalizeStagedData:
     timeout = 600
     LIB_NAME = "Finalize_Staged_Data_LIB"
 
-    #Define the number of chunks
+    # Define the number of chunks
     params = [1000, 2000]
 
     def __init__(self):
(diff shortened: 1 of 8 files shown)


python/benchmarks/finalize_staged_data.py
```diff
@@ -74,32 +74,36 @@ def peakmem_finalize_staged_data(self, cache:CachedDFGenerator, param:int):
     def teardown(self, cache:CachedDFGenerator, param:int):
         self.ac.delete_library(self.lib_name)
 
-class FinalizeStagedDataWiderDataframeX3(FinalizeStagedData):
-    '''
-        The test is meant to be executed with 3 times wider dataframe than the base test
-    '''
+if sys.version_info >= (3, 8):
```
IvoDD, line 77: I know this is from #2060 but since we upgrade the benchmarks to 3.11 we can remove this check?

IvoDD (approved):

python/benchmarks/finalize_staged_data.py
```diff
@@ -61,45 +65,51 @@ def setup(self, cache:CachedDFGenerator, param:int):
...
-    def peakmem_finalize_staged_data(self, cache:CachedDFGenerator, param:int):
-        if (not SLOW_TESTS):
-            raise SkipNotImplemented ("Slow tests are skipped")
-        super().peakmem_finalize_staged_data(cache,param)
-
-    def teardown(self, cache:CachedDFGenerator, param:int):
-        if (SLOW_TESTS):
+
+    def setup(self, cache: CachedDFGenerator, param: int):
+        if not SLOW_TESTS:
```
poodlewars, line 98: Not your code but wouldn't it be cleaner to lift the `SLOW_TESTS` conditions up to control whether we define the class at all? Rather than having them on each method?

poodlewars (approved):

python/benchmarks/finalize_staged_data.py
G-D-Petrov: I am keen on merging this as is because I want to rebase my numpy branch with these changes, so I've created a separate issue (#2070) to track this and fix it in the future.

PR #2076, opened 2024-12-17
Performance regression when requesting a timestamp before the earliest version

### LINKED ISSUE #1772: Performance regression when requesting a timestamp before the earliest version
### Describe the bug

In the ASV benchmark time_read_from_epoch, we preload the cache back to the beginning, but then ask for an as_of time that is earlier than any versions. I think previously given that the cache hasn't expired we would detect that the prior load was a LOAD_ALL and not reload it, because LOAD_ALL trumps everything. However now we're just looking at the earliest loading time, seeing that the requested time is before that and ignoring the fact that we loaded everything recently, so there's no point doing it again.

### Steps/Code to Reproduce

Why do I have to fill in all these boxes?

### Expected Results

It is less slow

more boxes

### Backend storage used

Seriously?

### Additional Context

_No response_

ognyanstoimenov:
#### Reference Issues/PRs
Fixes #1772

Version Map entry now stores whether the whole chain has been loaded, so we don't need to reload again in certain scenarios
*  load_strategy_ is removed from version_map_entry and replaced with a flag is_earliest_version_loaded in LoadProgress as we only need to know whether the entire version chain has been loaded previously. Previously we only kept whether the cache has been loaded with an ALL strategy but that missed the cases where we load via DOWN_TO 0 or FROM_TIME that is before earliest
*  if we haven't loaded the entire version chain, we additionally check if we have loaded all undeleted versions. This is in cases where we have tombstone_all on a certain version and an undeleted version is requested. See https://github.com/man-group/ArcticDB/pull/2076#discussion_r1893755748
    * this means that tombstone_all should not invalidate cache. see https://github.com/man-group/ArcticDB/pull/2076#discussion_r1903937601

* Performance has improved
 image

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -750,48 +767,58 @@ TEST(VersionMap, CacheInvalidation) {
...
-        // ALL and LOAD_UNDELETED because they both load to v0 (UNDELETED_ONLY will load v0 because only there it will load
-        check_all_caching({load_all_param, load_all_undeleted_param}, should_load_to_v[i], true);
-        check_all_caching(should_load_to_v[i], {load_all_param, load_all_undeleted_param}, false);
+        // LOAD_ALL loads to version 0
+        check_all_caching({load_all_param}, should_load_to_v[i], true);
+        // Load_all_undeleted loads to version 1 because of the tombstone_all in ver 1
+        check_all_caching({load_all_undeleted_param}, should_load_to_v[i], i != 0);
+
+        // If we have loaded to version 0 all load requests are true
+        check_all_caching(should_load_to_v[i], {load_all_param, load_all_undeleted_param}, i == 0);
```
IvoDD, line 821: I think we can make `check_all_caching(should_load_to_v[1], {load_all_undeleted_param}, true)` pass?

It will need the comment about `LOAD_ALL` `UNDELETED_ONLY` caching changed to pass.

IvoDD: Actually even better, with your changes we can remove the separate testing for `load_all` and `load_all_undeleted`.

Whu don't just place the `LOAD_ALL` in `should_load_to_v[0]`
and `LOAD_ALL_UNDELETED` in `should_load_to_v[1]`?

cpp/arcticdb/version/version_map.hpp
```diff
@@ -606,14 +608,9 @@ class VersionMapImpl {
                 return cached_timestamp <= requested_load_strategy.load_from_time_.value();
             }
             case LoadType::ALL:
```
IvoDD, line 610: Sorry I didn't think about this before but currently if we do: `{LoadType::ALL, UNDELETED_ONLY}` repeatedly it would never get cached.

That is because if we have the following version chain:
```
v0 <- v1 <- tall1 <- v2
```

the load_all_undeleted would reach the tall0 and terminate there.

I think we already have the information we need. I.e. if `earliest_loaded_version <= tombstone_all.value()` then we know that we have all undeleted versions.

It would also be good to test this.

IvoDD (changes requested):
Main thing to change is the `LOAD_ALL_UNDELETED` regression + testing.

Others are some small refactors.

Also please update the PR description to describe better what are the core changes (i.e. the changes to the version map entry) and the behavior changes.

And I think you might need to rebase the PR to get the ASV fixes. I would really like to see the logs from ASV saying we have improved the benchmark `from_epoch`

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -547,23 +547,36 @@ TEST(VersionMap, StorageLogging) {
...
-std::shared_ptr<VersionMapEntry> write_two_versions(
-    std::shared_ptr<InMemoryStore> store,
-    std::shared_ptr<VersionMap> version_map,
-    const StreamId& id) {
+std::shared_ptr<VersionMapEntry> write_versions(
+    const std::shared_ptr<InMemoryStore>& store,
+    const std::shared_ptr<VersionMap>& version_map,
+    const StreamId& id,
+    size_t number_of_versions,
+    std::vector<VersionId> tombstones_versions = {},
```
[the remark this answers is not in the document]
ognyanstoimenov: That's a good idea, will do it

cpp/arcticdb/version/version_map.hpp
ognyanstoimenov: Ah, good catch! Didn't think about that case

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -547,56 +547,65 @@ TEST(VersionMap, StorageLogging) {
...
+                version_map->do_write(store, key, entry);
+                write_symbol_ref(store, key, std::nullopt, entry->head_.value());
+                break;
+            }
+            case VersionChainOperation::Type::TOMBSTONE: {
+                version_map->write_tombstone(store, version_id, id, entry);
+                break;
+            }
+            case VersionChainOperation::Type::TOMBSTONE_ALL: {
+                version_map->delete_all_versions(store, id);
```
IvoDD, line 584: Nit: Theoretically we can still have a tombstone_all to a version_id different than the last.
You can instead use the `write_tombstone_all_internal` + `write_symbol_ref` to do that.

If we decide against that it would be good to document that `VersionChainOperation{TOMBSTONE_ALL, 3}` might not do what we expect.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -547,56 +547,65 @@ TEST(VersionMap, StorageLogging) {
...
-    // Write version 1
-    version_map->do_write(store, key2, entry);
-    write_symbol_ref(store, key2, std::nullopt, entry->head_.value());
-
-    // Write version 2
-    version_map->do_write(store, key3, entry);
-    write_symbol_ref(store, key3, std::nullopt, entry->head_.value());
+    using Type = VersionChainOperation::Type;
+    write_versions(store, version_map, id, {{Type::WRITE, 0},
+                                                         {Type::TOMBSTONE_ALL},
```
IvoDD, line 597: Nit: These extra whitespaces seem off. Maybe align all the operations the same number of spaces to the right.

Hopefully @poodlewars's formatting will get merged and we can all use the same formatting throughout.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -669,8 +682,12 @@ TEST(VersionMap, FollowingVersionChainWithCaching){
 
     // LATEST should still be cached, but the cached entry now needs to have no undeleted keys
     check_loads_versions(LoadStrategy{LoadType::LATEST, LoadObjective::INCLUDE_DELETED}, 2, 0);
+    EXPECT_FALSE(version_map->has_cached_entry(id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(-1)}));
     // FROM_TIME UNDELETED_ONLY should no longer be cached even though we used the same request before because the undeleted key it went to got deleted. So it will load the entire version chain
     check_loads_versions(LoadStrategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, static_cast<timestamp>(10)}, 3, 0);
+    // We have the full version chain loaded, so has_cached_entry should always return true (even when requesting timestamp before earliest version)
+    EXPECT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, static_cast<timestamp>(-1)}));
+    EXPECT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(-1)}));
```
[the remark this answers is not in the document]
IvoDD: Looks like we missed this one?
I mean something like:
```
EXPECT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}));
```

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -726,13 +739,20 @@ TEST(VersionMap, FollowingVersionChainEndEarlyOnTombstoneAll) {
...
-    // Set up the version chain v0(tombstone_all) <- v1 <- v2(tombstoned)
+    // Set up the version chain v0 <- v1(tombstone_all) <- v2 <- v3(tombstoned)
     auto store = std::make_shared<InMemoryStore>();
     auto version_map = std::make_shared<VersionMap>();
     StreamId id{"test"};
-    write_alternating_deleted_undeleted(store, version_map, id);
+    using Type = VersionChainOperation::Type;
+    std::vector<VersionChainOperation> version_chain = {{Type::WRITE, 0},
+                         {Type::WRITE, 1},
+                            {Type::TOMBSTONE_ALL},
```
IvoDD, line 751: Nit: some more weird whitespaces

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -16,6 +16,8 @@
 #include <deque>
 #include <vector>
 
+#include "async/tasks.hpp"
```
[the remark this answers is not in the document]
IvoDD: Looks like we missed this?

IvoDD (approved):
Left a few more very minor comments. Otherwise all looks good! Thanks! Approved

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -902,19 +907,11 @@ TEST(VersionMap, CacheInvalidationWithTombstoneAllAfterLoad) {
...
-            ASSERT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, static_cast<timestamp>(1)}));
-            ASSERT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, static_cast<timestamp>(0)}));
-        }
-        else {
-            // If we haven't loaded everything we should invalidate cached undeleted checks
-            ASSERT_FALSE(version_map->has_cached_entry(id, LoadStrategy{LoadType::LATEST, LoadObjective::UNDELETED_ONLY}));
-            ASSERT_FALSE(version_map->has_cached_entry(id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, static_cast<timestamp>(1)}));
-            ASSERT_FALSE(version_map->has_cached_entry(id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, static_cast<timestamp>(0)}));
-        }
+        // Tombstone All should not invalidate cache as it deletes everything so all undeleted versions have been loaded.
```
ognyanstoimenov, line 910: This is something I caught when testing. Previously writing tombstone_all would cause any subsequent UNDELETED_ONLY lookups in cache to return false. However, I think this was wrong because writing tombstone_all deletes all previous versions so it is impossible for an undeleted version to exist in the version chain and not in the cache. Thus all undeleted versions have been loaded in cache.

poodlewars:
Please tidy up the git commit history

cpp/arcticdb/version/test/test_version_map.cpp
poodlewars: Add tests where the tombstone_all version ID is not the latest version?

poodlewars (commented):
Add a benchmark to demonstrate the performance improvement?
Is this functionality covered by an end to end (Python) test?

ognyanstoimenov:
> Please tidy up the git commit history

Will squash before merging

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -2704,3 +2704,71 @@ def test_missing_first_version_key_batch(basic_store):
     vits = lib.batch_read(symbols, as_ofs=write_times)
     for x in range(num_items):
         assert_equal(vits[symbols[x]].data, expected[x])
+
+@pytest.mark.parametrize("use_caching", [True, False])
+def test_version_chain_cache(basic_store, use_caching):
+    timeout = sys.maxsize if use_caching else 0
+    lib = basic_store
+    symbol = "test"
+    dataframes = [sample_dataframe()] * 10
```
IvoDD, line 2713: This way all dataframes are equal (in fact maybe pointing to the same dataframe object). It would be useful for your tests where you compare the dataframes to make them different.

Instead use something like:
```
[sample_dataframe() for _ in range(10)]
```

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -2704,3 +2704,71 @@ def test_missing_first_version_key_batch(basic_store):
...
+                assert_equal(lib.read(symbol, as_of=timestamps[i].after).data, dataframes[i-1])
+            else:
+                assert_equal(lib.read(symbol, as_of=i).data, dataframes[i])
+                assert_equal(lib.read(symbol, as_of=timestamps[i].after).data, dataframes[i])
+
+            if i == 0:
+                with pytest.raises(NoSuchVersionException):
+                    lib.read(symbol, as_of=timestamps[i].before)
+            else:
+                assert_equal(lib.read(symbol, as_of=timestamps[i].before).data, dataframes[i-1])
```
IvoDD, line 2761: Here is an example where the same dataframes bites us. If we request `as_of=timestamps[4].before` we should get `dataframes[2]` because 3 is deleted.

IvoDD (changes requested):

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -863,59 +873,99 @@ TEST(VersionMap, CacheInvalidationWithTombstoneAllAfterLoad) {
...
 
-    // Use a clean version_map
-    version_map = std::make_shared<VersionMap>();
+    for (const auto& load_strategy : {LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(1)},
+                                    LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(0)}}) {
+        store = std::make_shared<InMemoryStore>();
+        version_map = std::make_shared<VersionMap>();
 
+        // Version chain is v0 <- v1 <- v2
+        write_versions(store, version_map, id, 3);
```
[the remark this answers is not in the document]
IvoDD: But the store is cleared every time so this is fine.

IvoDD (approved):

poodlewars (approved):

PR #2083, opened 2024-12-18
Diagnostics for finalize staged data

poodlewars:
Draft


willdealtry:
> Draft

Some of these will need to be ARCTICDB_RUNTIME_DEBUG or they will be compiled out in release builds.  Beware of side-effects in frequently-called functions though, as the arguments will be evaluated

PR #2084, opened 2024-12-18
Add debug logging for finalize_staged_data

poodlewars:


cpp/arcticdb/async/async_store.hpp
```diff
@@ -93,7 +93,7 @@ folly::Future<entity::VariantKey> write(
                 "Descriptor id mismatch in atom key {} != {}",
                 stream_id,
                 segment.descriptor().id());
-
+    ARCTICDB_DEBUG(log::version(), "Creating write future key_type {} stream_id {} start_index {} end_index {}");
```
willdealtry, line 96: Do you want ARCTICDB_RUNTIME_DEBUG? Otherwise they're going to get compiled out

poodlewars: They're so noisy that I think I want them compiled out

willdealtry: Cool, but I thought you wanted some stuff you could switch on in a customer environment? Bear in mind that in frequently-called functions the parameters will get evaluated, so we need to be careful not to slow anything down

poodlewars: I was planning to give them a debug wheel (because I want to run the memory profiler too, which needs debug symbols). I didn't want to worry about the cost of the parameter evaulation in release builds, so thought safest to just compile them all out in release builds.

PR #2087, opened 2024-12-19
Improve the performance of update by parallelising reads. Implement интернал async update method.

vasil-pashov:
#### Reference Issues/PRs

Implement `async_update_impl` function which returns a future. The synchronous version for update just calls it and waits for the future just like append does.

This keeps most of the code for update the same, however instead of calling `.get` on futures it will chain then and return a future. In the process of doing this the reads needed by update were made in parallel. Thus the regular update will have improved performance.

Slight refactor of C++ unit tests, using std::array instead of std::vector for fixed size collections and placing const and constexpr specifiers. No functional changes.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/version/version_core.cpp
```diff
@@ -304,92 +324,137 @@ void check_update_data_is_sorted(InputTensorFrame& frame, const index::IndexSegm
...
+    const UpdateInfo& update_info,
+    const UpdateQuery& query,
+    const std::shared_ptr<InputTensorFrame>& frame,
+    WriteOptions&& options,
+    bool dynamic_schema,
+    bool empty_types) {
+    auto version_key = async_update_impl(store, update_info, query, frame, std::move(options), dynamic_schema, empty_types).get();
     auto versioned_item = VersionedItem(to_atom(std::move(version_key)));
-    ARCTICDB_DEBUG(log::version(), "updated stream_id: {} , version_id: {}", stream_id, update_info.next_version_id_);
+    ARCTICDB_DEBUG(log::version(), "updated stream_id: {} , version_id: {}", frame->desc.id(), update_info.next_version_id_);
```
alexowens90, line 457: Is this set on the input frame? If not, can use the stream ID from the `version_key`

cpp/arcticdb/pipeline/index_segment_reader.cpp
```diff
@@ -20,8 +20,13 @@ using namespace arcticdb::proto::descriptors;
 namespace arcticdb::pipelines::index {
 
 IndexSegmentReader get_index_reader(const AtomKey &prev_index, const std::shared_ptr<Store> &store) {
-    auto [key, seg] = store->read_sync(prev_index);
-    return index::IndexSegmentReader{std::move(seg)};
+    return async_get_index_reader(prev_index, store).get();
```
alexowens90, line 23: I think we should leave this implementation using `read_sync`, so that the scheduling overhead can be avoided if necessary

cpp/arcticdb/pipeline/read_pipeline.hpp
```diff
@@ -61,8 +61,7 @@ void foreach_active_bit(const util::BitSet &bs, C &&visitor) {
     }
 }
 
-template<typename ContainerType>
-std::vector<SliceAndKey> filter_index(const ContainerType &container, std::optional<CombinedQuery<ContainerType>> &&query) {
+inline std::vector<SliceAndKey> filter_index(const index::IndexSegmentReader& container, std::optional<CombinedQuery<index::IndexSegmentReader>> &&query) {
```
alexowens90, line 64: I assume the template parameter always resolved to `IndexSegmentReader`? Maybe rename `container` to match as well now?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -304,92 +324,137 @@ void check_update_data_is_sorted(InputTensorFrame& frame, const index::IndexSegm
...
-    auto affected_keys = filter_index(index_segment_reader, std::move(combined));
-    std::vector<SliceAndKey> unaffected_keys;
-    std::set_difference(std::begin(index_segment_reader),
-                        std::end(index_segment_reader),
-                        std::begin(affected_keys),
-                        std::end(affected_keys),
-                        std::back_inserter(unaffected_keys));
-
-    util::check(affected_keys.size() + unaffected_keys.size() == index_segment_reader.size(), "Unaffected vs affected keys split was inconsistent {} + {} != {}",
+    return index::async_get_index_reader(*(update_info.previous_index_key_), store).thenValue([=](index::IndexSegmentReader&& index_segment_reader) {
```
alexowens90, line 341: Why are we capturing everything by copy?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -304,92 +324,137 @@ void check_update_data_is_sorted(InputTensorFrame& frame, const index::IndexSegm
...
+        check_update_data_is_sorted(*frame, index_segment_reader);
+        const bool bucketize_dynamic = index_segment_reader.bucketize_dynamic();
+        (void)check_and_mark_slices(index_segment_reader, dynamic_schema, false, std::nullopt, bucketize_dynamic);
+        fix_descriptor_mismatch_or_throw(UPDATE, dynamic_schema, index_segment_reader, *frame, empty_types);
+        frame->set_bucketize_dynamic(bucketize_dynamic);
+        return slice_and_write(
+            frame,
+            get_slicing_policy(options, *frame),
+            IndexPartialKey{stream_id, update_info.next_version_id_}, store
+        ).thenValue([
```
alexowens90, line 357: `slice_and_write` finishes on the IO executor, but I think we want to be on CPU for the next task

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -252,40 +253,41 @@ static RowRange partial_rewrite_row_range(
...
-    auto fut_key = store->write(
-        key.type(),
-        version_id,
-        key.id(),
-        start_ts,
-        end_ts,
-        std::move(output)
-    );
-    return SliceAndKey{std::move(new_slice), std::get<AtomKey>(std::move(fut_key).get())};
+    return store->read(existing.key()).thenValue([=](std::pair<VariantKey, SegmentInMemory>&& key_segment) {
```
alexowens90, line 262: Capture by copy?

alexowens90: Could also be a `thenValueInline`

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -252,40 +253,41 @@ static RowRange partial_rewrite_row_range(
...
+        const auto& key = existing.key();
+        const SegmentInMemory& segment = key_segment.second;
+        const RowRange affected_row_range = partial_rewrite_row_range(segment, index_range, affected_part);
+        const int64_t num_rows = affected_row_range.end() - affected_row_range.start();
+        if (num_rows <= 0) {
+            return folly::Future<std::optional<SliceAndKey>>{std::nullopt};
```
alexowens90, line 268: I don't think this needs `folly::Future`, if you specify the return type of the lambda can probably just `return std::nullopt`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -304,92 +324,137 @@ void check_update_data_is_sorted(InputTensorFrame& frame, const index::IndexSegm
...
 }
 
-VersionedItem update_impl(
+struct UpdateRanges {
+    IndexRange front;
+    IndexRange back;
+    IndexRange original;
+};
+
+folly::Future<AtomKey> async_update_impl(
```
alexowens90, line 333: This methods a bit of a monster now, if there's a clean way to break it up a bit it should get more readable

alexowens90 (changes requested):

cpp/arcticdb/version/version_core.cpp
vasil-pashov: The whole implementation is inside `async_get_index_reader` and we need to propagate the input params. I used it as a shorthand. In theory, options can be moved but it's consisted only of PODs so it won't do anything.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: I couldn't think way to split things in logically rather than pulling them for the sake of splitting. I'll give it another try, now that everything compiles and all tests are passing there will be less unknowns.

cpp/arcticdb/pipeline/write_frame.cpp
vasil-pashov: Same as https://github.com/man-group/ArcticDB/pull/2087#discussion_r1894190555 the implementation is in the lambda and it uses all variables. So I need to pass them to the future. Can't capture by ref as they'll die when this is put in the queue and the function returns.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: Yes, it is. `PythonVersionStore::update` passes the result of `convert::py_ndf_to_frame` to `update_internal`

alexowens90 (approved):

PR #2088, opened 2024-12-19
Bugfix 4897570890562900007: Fix decoding of fields with >2^16 blocks

alexowens90:
#### Reference Issues/PRs
Fixes 4897570890562900007 on `5.1.x` branch

Allows for there being more than `2^16` blocks in an encoded field, observed with a large compacted symbol list key of a library with 10 million symbols.

### DIFF
--- a/cpp/arcticdb/storage/memory_layout.hpp
+++ b/cpp/arcticdb/storage/memory_layout.hpp
@@ -101,15 +101,15 @@ enum class BitmapFormat : uint8_t {
 // pointer to the first block
 struct EncodedField {
     EncodedFieldType type_ = EncodedFieldType::UNKNOWN;
-    uint16_t shapes_count_ = 0u;
-    uint16_t values_count_ = 0u;
+    uint32_t shapes_count_ = 0u;
+    uint32_t values_count_ = 0u;
     uint32_t sparse_map_bytes_ = 0u;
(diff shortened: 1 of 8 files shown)


cpp/arcticdb/storage/memory_layout.hpp
```diff
@@ -101,15 +101,15 @@ enum class BitmapFormat : uint8_t {
 // pointer to the first block
 struct EncodedField {
     EncodedFieldType type_ = EncodedFieldType::UNKNOWN;
-    uint16_t shapes_count_ = 0u;
-    uint16_t values_count_ = 0u;
+    uint32_t shapes_count_ = 0u;
+    uint32_t values_count_ = 0u;
```
poodlewars, line 105: Could add some checks that what we are decoding will fit in the `uint32_t` so we fail in a more obvious way?

poodlewars (approved):

cpp/arcticdb/storage/memory_layout.hpp
alexowens90: Yeh, one to discuss with @willdealtry when he's back I think, as the same logic applies to any of these ints that are <64 bits wide

willdealtry: How did we end up storing more than 65k blocks?

PR #2089, opened 2024-12-20
Bugfix 4897570890562900007: Fix decoding of fields with >2^16 blocks

alexowens90:
Reference Issues/PRs
Fixes 4897570890562900007 on master

What does this implement or fix?
Allows for there being more than 2^16 blocks in an encoded field, observed with a large compacted symbol list key of a library with 10 million symbols.


poodlewars (approved):
https://github.com/man-group/ArcticDB/pull/2088#discussion_r1893855175

PR #2098, opened 2025-01-02
Compatibility test may trigger segfaults

### LINKED ISSUE #1979: Comatibility test may trigger segfaults
We've seen azure storage compat tests trigger a segfault [here](https://github.com/man-group/ArcticDB/actions/runs/11666782192/job/32489194323):
```
FAILED tests/compat/arcticdb/test_compatibility.py::test_modify_old_library_option_with_current[4.5.0-azurite] - tests.compat.conftest.ErrorInVenv: Executing ['from arcticdb import Arctic', 'import pandas as pd', 'import numpy as np', "ac = Arctic('azure://DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10047/devstoreaccount1;Container=container3;CA_cert_path=/tmp/tmpos9yrqltAzuriteStorageFixtureFactory/client.pem')", "expected_df = pd.read_parquet('/tmp/tmp2dkdtq4p/expected_df.parquet')", "lib = ac.get_library('test_modify_old_library_option.269_2024-11-04T17_39_52_672424')", "read_df = lib.read('sym').data", 'pd.testing.assert_frame_equal(read_df, expected_df)'] failed with return code -11
```

We've seen such segfaults previously with `lmdb` and `mongo` which both had race conditions at desctruction time. We believe we've fixed the `lmdb` and `mongo` segfaults from `4.5.1` going forward.

Due to the nature of the compat tests they are very good at reproducing these type of desctructor segfaults because they run many short lived processes which end with many desctructions.

I suspect we might have a similar desctructor related bug with azure.

G-D-Petrov:
#### Reference Issues/PRs
Fixes #1979 

Removes the LMDB fixture from the compatibility tests because there is a bug on versions 1.6..2 and 4.5.1 that cause a Seg Fault with it.

Also adds a way to manually pass a string to pytest to change the tests that are executed.
Very useful in cases such as this change were you want to test only a specific subset of tests and you want to run them many times in parallel.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/compat/conftest.py
+++ b/python/tests/compat/conftest.py
@@ -15,13 +15,18 @@ from packaging.version import Version
 
 logger = logging.getLogger("Compatibility tests")
 
+
 class ErrorInVenv(Exception):
     """To signal errors occuring from within the venv"""
 
-def is_running_on_windows():
-    return os.name == 'nt'
 
-def run_shell_command(command : List[Union[str, os.PathLike]], cwd : Optional[os.PathLike] = None) -> subprocess.CompletedProcess:
+def is_running_on_windows():
+    return os.name == "nt"
+
+
+def run_shell_command(
+    command: List[Union[str, os.PathLike]], cwd: Optional[os.PathLike] = None
+) -> subprocess.CompletedProcess:
     logger.info(f"Executing command: {command}")
     result = None
     if is_running_on_windows():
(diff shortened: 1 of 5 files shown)


grusev (approved):
Very useful

python/tests/compat/conftest.py
```diff
@@ -126,33 +160,34 @@ def __init__(self, arctic, lib_name, create_if_not_exists=True):
...
-            "pd.testing.assert_frame_equal(read_df, expected_df)"
+            "print(read_df)",
+            "pd.testing.assert_frame_equal(read_df, expected_df)",
         ]
         return self.execute(python_commands, {"expected_df": df})
 
 
 @pytest.fixture(
-    scope="session",
+    # scope="session",
```
G-D-Petrov, line 182: The changes here are so we can support parallel executions with `pytest-xdist`

python/tests/compat/conftest.py
```diff
@@ -161,7 +196,9 @@ def old_venv(request):
 
 @pytest.fixture(
     params=[
-        "lmdb",
+        # Don't include lmdb because there is a bug in 1.6.2/4.5.1
+        # which causes the tests to segfault intermittently
+        # "lmdb",
```
G-D-Petrov, line 201: This is the main changes, as :
- there is a bug on version 1.6.2 and 4.5.1 with lead to flaky seg faults with lmdb storage
- 1.6.2 doesn't support Mongo

IvoDD: Instead of removing lmdb entirely what do you think about keeping lmdb but skipping if:
```
Version(old_venv.version) < Version("5.0.0") and arctic_uri.startswith("lmdb")
```

And discussing offline it sounds like you already tested that lmdb works fine with 5.0.0, what do you think about just adding the 5.0.0 version in the old_venv fixture? Should be quite simple and will help with any potential regressions between 5.0.0 and going forward.

IvoDD (approved):

python/tests/compat/conftest.py
poodlewars: Remove the commented out code? As I assume this change will be permanent

poodlewars (commented):
Please squash the commits when you merge

poodlewars (approved):

PR #2103, opened 2025-01-07
Enhancement/7992967434/filters and projections ternary operator

alexowens90:
#### Reference Issues/PRs
Implements [7992967434](https://man312219.monday.com/boards/7852509418/pulses/7992967434)

Implements a ternary operator equivalent to `numpy.where`, primarily for projecting new columns based on some condition, although it can also be used for filtering. Semantically the same as `left if condition else right`, although this Pythonic syntax cannot be made to work due to limitations of the language.

#### Any other comments?
See `test_ternary.py` for a plethora of examples and the expected behaviour in each case.
Example benchmark output with annotations below.
The first parameter to all benchmarks is the number of rows (100k for all of them right now), so the single-threaded per-row time can be calculated by dividing by 100,000.
e.g. projecting a new column of 100k rows by choosing from 2 dense columns (likely a common use case) takes 424us, or just over 4ns per row.
Other parameters are explained for each individual benchmark.
```
Run on (20 X 2918.4 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x10)
  L1 Instruction 32 KiB (x10)
  L2 Unified 1280 KiB (x10)
  L3 Unified 24576 KiB (x1)
Load Average: 4.23, 6.56, 6.73
--------------------------------------------------------------------------------------------------
Benchmark                                                        Time             CPU   Iterations
--------------------------------------------------------------------------------------------------
BM_ternary_bitset_bitset/100000                               13.1 us         13.1 us        58099
# Second arg is whether the boolean argument is true or false, third is whether the arguments are swapped
BM_ternary_bitset_bool/100000/1/1                             2.00 us         2.00 us       363634
BM_ternary_bitset_bool/100000/1/0                             7.43 us         7.43 us       101700
BM_ternary_bitset_bool/100000/0/1                             7.28 us         7.28 us        88907
BM_ternary_bitset_bool/100000/0/0                             2.45 us         2.45 us       307832
BM_ternary_numeric_dense_col_dense_col/100000                  424 us          424 us         1276
BM_ternary_numeric_sparse_col_sparse_col/100000               3548 us         3548 us          185
# Second arg is whether the arguments are swapped
BM_ternary_numeric_dense_col_sparse_col/100000/1              2555 us         2555 us          258
BM_ternary_numeric_dense_col_sparse_col/100000/0              2800 us         2800 us          262
# Second arg is the number of unique strings in each string column, third is whether the columns have the same string pool or not
BM_ternary_string_dense_col_dense_col/100000/100000/1          438 us          438 us         1534
BM_ternary_string_dense_col_dense_col/100000/100000/0        16257 us        16258 us           43
BM_ternary_string_dense_col_dense_col/100000/2/1               441 us          441 us         1603
BM_ternary_string_dense_col_dense_col/100000/2/0              4219 us         4219 us          186
BM_ternary_string_sparse_col_sparse_col/100000/100000/1       3854 us         3854 us          191
BM_ternary_string_sparse_col_sparse_col/100000/100000/0      10753 us        10754 us           67
BM_ternary_string_sparse_col_sparse_col/100000/2/1            3655 us         3655 us          183
BM_ternary_string_sparse_col_sparse_col/100000/2/0            4592 us         4592 us          123
BM_ternary_string_dense_col_sparse_col/100000/100000/1        2957 us         2957 us          236
BM_ternary_string_dense_col_sparse_col/100000/100000/0       13980 us        13980 us           50
BM_ternary_string_dense_col_sparse_col/100000/2/1             2967 us         2966 us          237
BM_ternary_string_dense_col_sparse_col/100000/2/0             5179 us         5179 us          160
# Second arg  is whether the arguments are swapped
BM_ternary_numeric_dense_col_val/100000/1                      360 us          359 us         1871
BM_ternary_numeric_dense_col_val/100000/0                      388 us          388 us         1692
BM_ternary_numeric_sparse_col_val/100000/1                    2244 us         2244 us          292
BM_ternary_numeric_sparse_col_val/100000/0                    2385 us         2385 us          283
# Second arg  is whether the arguments are swapped, third is the number of unique strings in the column
BM_ternary_string_dense_col_val/100000/1/100000               8259 us         8258 us           82
BM_ternary_string_dense_col_val/100000/0/100000               7683 us         7683 us           93
BM_ternary_string_dense_col_val/100000/1/2                    2578 us         2578 us          261
BM_ternary_string_dense_col_val/100000/0/2                    2385 us         2385 us          297
BM_ternary_string_sparse_col_val/100000/1/100000              6302 us         6302 us          129
BM_ternary_string_sparse_col_val/100000/0/100000              5792 us         5792 us          115
BM_ternary_string_sparse_col_val/100000/1/2                   2903 us         2903 us          249
BM_ternary_string_sparse_col_val/100000/0/2                   3095 us         3095 us          232
# Second arg  is whether the arguments are swapped
BM_ternary_numeric_dense_col_empty/100000/1                   1269 us         1269 us          584
BM_ternary_numeric_dense_col_empty/100000/0                   1354 us         1354 us          512
BM_ternary_numeric_sparse_col_empty/100000/1                  1363 us         1363 us          572
BM_ternary_numeric_sparse_col_empty/100000/0                  1374 us         1374 us          484
# Second arg  is whether the arguments are swapped, third is the number of unique strings in the column
BM_ternary_string_dense_col_empty/100000/1/100000             1217 us         1217 us          587
BM_ternary_string_dense_col_empty/100000/0/100000             1343 us         1343 us          577
BM_ternary_string_dense_col_empty/100000/1/2                  1287 us         1287 us          574
BM_ternary_string_dense_col_empty/100000/0/2                  1363 us         1363 us          518
BM_ternary_string_sparse_col_empty/100000/1/100000            1413 us         1413 us          524
BM_ternary_string_sparse_col_empty/100000/0/100000            1343 us         1343 us          517
BM_ternary_string_sparse_col_empty/100000/1/2                 1293 us         1293 us          540
BM_ternary_string_sparse_col_empty/100000/0/2                 1235 us         1235 us          480
BM_ternary_numeric_val_val/100000                              368 us          368 us         2039
BM_ternary_string_val_val/100000                               376 us          376 us         1862
# Second arg  is whether the arguments are swapped
BM_ternary_numeric_val_empty/100000/1                         40.7 us         40.7 us        16491
BM_ternary_numeric_val_empty/100000/0                         36.7 us         36.7 us        17836
BM_ternary_string_val_empty/100000/1                          40.8 us         40.8 us        17892
BM_ternary_string_val_empty/100000/0                          58.2 us         58.2 us        13825
# Second arg is whether the left argument is true or false, third is whether the right argument is true or false
BM_ternary_bool_bool/100000/1/1                               1.43 us         1.43 us       518204
BM_ternary_bool_bool/100000/1/0                               1.99 us         1.99 us       378598
BM_ternary_bool_bool/100000/0/1                               4.52 us         4.52 us       157505
BM_ternary_bool_bool/100000/0/0                              0.020 us        0.020 us     37060921
```

### DIFF
--- a/cpp/arcticdb/pipeline/frame_utils.cpp
+++ b/cpp/arcticdb/pipeline/frame_utils.cpp
@@ -108,7 +108,7 @@ void adjust_slice_ranges(const std::shared_ptr<pipelines::PipelineContext>& pipe
             // Hash groupings produce slices that all have row ranges starting at 0 and the same col ranges, hence the
             // second condition
             increment_row_slice = next_slice.row_range.first != slice.row_range.first ||
-                                  next_slice.col_range.first == slice.col_range.first;
+                                  next_slice.col_range.first <= slice.col_range.first;
         } else {
             increment_row_slice = true;
         }


cpp/arcticdb/processing/operation_dispatch_ternary.cpp
```diff
@@ -0,0 +1,582 @@
...
+        output_bitset = condition;
+    } else if (right) {
+        // left is false
+        output_bitset = ~condition;
+    } // else left and right are false, the resize below will correctly initialise all bits to zero
+    output_bitset.resize(condition.size());
+    return output_bitset;
+}
+
+VariantData visit_ternary_operator(const VariantData& condition, const VariantData& left, const VariantData& right) {
```
poodlewars, line 420: This function has lost me - I'll pick through it some more and we can chat about it tomorrow

cpp/arcticdb/processing/operation_dispatch_ternary.cpp
```diff
@@ -0,0 +1,582 @@
...
+                return transform_to_placeholder(result);
+            },
+            [&c](const std::shared_ptr<Value> &l, const ColumnWithStrings &r) -> VariantData {
+                auto result = ternary_operator<true>(c, r, *l);
+                return transform_to_placeholder(result);
+            },
+            [&c](const ColumnWithStrings &l, FullResult) -> VariantData {
+                user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+                        is_bool_type(l.column_->type().data_type()),
+                        "Ternary operator cannot combine column '{}' of type {} with a FullResult. This can be caused by dynamic schema when a row-slice has a necessary column missing.",
```
poodlewars, line 498: These error messages will be hard for users to understand. What does "necessary column" mean? The condition column? We should be able to fill the selected column with default values for slices where it is not present, shouldn't we?

cpp/arcticdb/processing/operation_dispatch_ternary.cpp
```diff
@@ -0,0 +1,582 @@
...
+                auto result = ternary_operator(c, l, bitset);
+                return transform_to_placeholder(result);
+            },
+            [&c](const ColumnWithStrings &l, const util::BitSet &r) -> VariantData {
+                auto bitset = std::get<util::BitSet>(transform_to_bitset(l));
+                auto result = ternary_operator(c, bitset, r);
+                return transform_to_placeholder(result);
+            },
+            [&c](const util::BitSet &l, const std::shared_ptr<Value> &r) -> VariantData {
+                user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(is_bool_type(r->data_type_),
```
poodlewars, line 454: IIUC this is because the operator needs to resolve to a filter - might be worth explaining that in the error message or in a comment as the reason for the constraint wasn't obvious to me till I stepped through the tests

poodlewars (commented):
Would be good to give more helpful error messages when users misuse the API, eg,

```
def test_filter_ternary_column_full_and_empty_results(lmdb_version_store_v1):
    lib = lmdb_version_store_v1
    symbol = "test_filter_ternary_column_full_and_empty_results"
    df = pd.DataFrame(
        {
            "conditional": [True, False, False, True, False, True],
            "col1": ["a", "b"] * 3,
            "col2": [0] * 6,
        },
        index=pd.date_range("2024-01-01", periods=6)
    )
    lib.write(symbol, df)

    q = QueryBuilder()
    q = q[where(q["conditional"], q["col1"], q["col1"])]
    received = lib.read(symbol, query_builder=q).data
    # blows up, E       arcticdb_ext.exceptions.InternalException: E_ASSERTION_FAILURE Cannot convert column '(COND ? col1 : col1)' of type TD<type=STRING, dim=0> to a bitset
```

python/tests/unit/arcticdb/version_store/test_ternary.py
```diff
@@ -0,0 +1,1134 @@
...
+        index=pd.date_range("2024-01-01", periods=6)
+    )
+    lib.write(symbol, df)
+
+    expected = df
+    expected["new_col"] = np.where(df["conditional"].to_numpy(), 0, 1)
+    q = QueryBuilder()
+    q = q.apply("new_col", where(q["conditional"], 0, 1))
+    received = lib.read(symbol, query_builder=q).data
+    assert_frame_equal(expected, received, check_dtype=False)
```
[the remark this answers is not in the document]
alexowens90: Agreed. I've done this by coercing the dtypes on the expected dataframes and removing `check_dtype=False`, except for in the hypothesis test, where until Vasil merges the `modify_schema` change, the output type can by either float64 or uint16 depending on the condition

python/tests/unit/arcticdb/version_store/test_ternary.py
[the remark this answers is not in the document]
alexowens90: "Would there be any value..."
Not really, by the time they hit the ternary operator it is just a bitset regardless of how the bitset was generated
"Do we support arbitrary expressions..."
Yes, see `test_project_ternary_sparse_col_val` for an example. Again, there isn't any advantage to testing more complex expressions here, as recursive AST traversal of a complex expression to generate a bitset is tested elsewhere (e.g. `test_filter_complex_expression`)

cpp/arcticdb/processing/test/test_arithmetic_type_promotion.cpp
```diff
@@ -11,552 +11,683 @@
...
+    static_assert(std::is_same_v<binary_operation_promoted_type<int32_t, double, DivideOperator>::type, double>);
+    static_assert(std::is_same_v<binary_operation_promoted_type<int64_t, double, DivideOperator>::type, double>);
+
+    static_assert(std::is_same_v<binary_operation_promoted_type<double, int8_t,  DivideOperator>::type, double>);
+    static_assert(std::is_same_v<binary_operation_promoted_type<double, int16_t, DivideOperator>::type, double>);
+    static_assert(std::is_same_v<binary_operation_promoted_type<double, int32_t, DivideOperator>::type, double>);
+    static_assert(std::is_same_v<binary_operation_promoted_type<double, int64_t, DivideOperator>::type, double>);
+}
+
+TEST(ArithmeticTypePromotion, Ternary) {
```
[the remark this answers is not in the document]
alexowens90: They are very close, the only difference is when you have one signed integer and a uint64. In this case the IsIn/IsNotIn operators have special handling to keep everything exact, whereas the ternary operator promotes to a double.
There is currently a static assertion testing gap of IsIn/IsNotIn though, I'll add that in.
I think I prefer keeping them separate (and adding in the tests as you suggested) rather than reusing `binary_operation_promoted_type` with some special bool handling and a tag type `Ternary` struct which would only be used as the `Func` template parameter, purely for readability. I would say that `ternary_operation_promoted_type` is much easier to grok than `binary_operation_promoted_type`, and so we can be more confident in it's correctness.

cpp/arcticdb/processing/operation_dispatch_ternary.cpp
alexowens90: I've tried to improve the error messages.
This stuff is a bit messy. `FullResult` is only really creatable by applying the `NOT` operator to an empty result, hence jumping through hoops like [this](https://github.com/man-group/ArcticDB/blob/enhancement/7992967434/filters-and-projections-ternary-operator/python/tests/unit/arcticdb/version_store/test_ternary.py#L843-L848) to produce one in tests. However, `EmptyResults` can be created either by trying to [get](https://github.com/man-group/ArcticDB/blob/enhancement/7992967434/filters-and-projections-ternary-operator/cpp/arcticdb/processing/processing_unit.cpp#L69) a column from a row-slice it is missing from, or by performing a filter that results in [zero set bits](https://github.com/man-group/ArcticDB/blob/424cd56e295afafd64444420b92fcf89a82dd1ea/cpp/arcticdb/processing/operation_dispatch.cpp#L22). At this point, we can't tell how we got a `FullResult`, so can't really do anything sensible.

alexowens90:
> Would be good to give more helpful error messages when users misuse the API, eg,
> 
> ```
> def test_filter_ternary_column_full_and_empty_results(lmdb_version_store_v1):
>     lib = lmdb_version_store_v1
>     symbol = "test_filter_ternary_column_full_and_empty_results"
>     df = pd.DataFrame(
>         {
>             "conditional": [True, False, False, True, False, True],
>             "col1": ["a", "b"] * 3,
>             "col2": [0] * 6,
>         },
>         index=pd.date_range("2024-01-01", periods=6)
>     )
>     lib.write(symbol, df)
> 
>     q = QueryBuilder()
>     q = q[where(q["conditional"], q["col1"], q["col1"])]
>     received = lib.read(symbol, query_builder=q).data
>     # blows up, E       arcticdb_ext.exceptions.InternalException: E_ASSERTION_FAILURE Cannot convert column '(COND ? col1 : col1)' of type TD<type=STRING, dim=0> to a bitset
> ```

That is actually coming from the filter after the ternary operator, you'd get the same with `q = q[q["col1"]]`. The only way to make it a bit more explanatory would be to attach expression names to bitsets in `VariantData`, similarly to how we do with columns that are functions of other columns. That's quite a big refactor though, if we decide it's worth it then I'd do it in a separate PR.

poodlewars (approved):

cpp/arcticdb/processing/operation_dispatch_ternary.cpp
```diff
@@ -0,0 +1,590 @@
...
+                                if (cond) {
+                                    // This is a bit faster than ColumnWithStrings::string_at_offset in this case as
+                                    // none of the additional checks/code are useful
+                                    if (is_a_string(left_val)) {
+                                        return string_pool->get(col.string_pool_->get_const_view(left_val)).offset();
+                                    } else {
+                                        return left_val;
+                                    }
+                                } else {
+                                    return right_val;
```
vasil-pashov, line 213: This snippet is repeated a few times. Can we abstract it into a function?

cpp/arcticdb/processing/test/benchmark_ternary.cpp
```diff
@@ -0,0 +1,507 @@
...
+    for (size_t idx = 0; idx < num_rows; ++idx) {
+        if (dis(gen) == 0) {
+            inserter = idx;
+        }
+    }
+    inserter.flush();
+    return bitset;
+}
+
+Value generate_numeric_value() {
```
vasil-pashov, line 37: It looks like we generate `std::mt19937` on each invocation of the call. Usually the pattern is to create `std::mt19937` only once and use its values to seed different distributions. That's because generating `std::mt19937` is slow and it's not worth it using it for a single value. Given this is not a production code I think it's not a big deal but ideally we can make it static.

cpp/arcticdb/processing/test/benchmark_ternary.cpp
vasil-pashov, line 1: @IvoDD mentioned that there's a way to make ASV take into account google benchmarks. Should we do it for this case?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1429,12 +1429,8 @@ folly::Future<SegmentInMemory> prepare_output_frame(
         std::any& handler_data) {
     pipeline_context->clear_vectors();
     pipeline_context->slice_and_keys_ = std::move(items);
-	ranges::sort(pipeline_context->slice_and_keys_, [] (const auto& left, const auto& right) {
```
vasil-pashov, line 1432: Why is this not needed anymore?

cpp/arcticdb/column_store/column.hpp
```diff
@@ -909,7 +909,12 @@ class Column {
         } else if (left_input_column.is_sparse() && right_input_column.is_sparse()) {
             // Both sparse, only check the intersection of on-bits from both sparse maps
             auto bits_to_check = left_input_column.sparse_map() & right_input_column.sparse_map();
-            initialise_output_bitset(bits_to_check, sparse_missing_value_output, output_bitset);
+            if (sparse_missing_value_output) {
+                output_bitset = bits_to_check;
+                output_bitset.flip();
```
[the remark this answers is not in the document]
alexowens90: `sparse_missing_value_output` being true means that in rows where either the left or right column do not have values, the bitset should have a 1 rather than a 0, which is what is guaranteed here.
We initialise bits where both columns have values to zero, and then the `inserter` below flips only the bits where `f(left_accessor.at(*set_bit), right_accessor.at(*set_bit))` is true

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -226,8 +233,55 @@ struct type_arithmetic_promoted_type {
     >;
 };
 
+template <class LHS, class RHS>
+struct ternary_operation_promoted_type {
```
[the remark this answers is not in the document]
alexowens90: I just tested it, and compiling a program that generates all of these type combinations is the same speed as compiling hello world

cpp/arcticdb/processing/clause.cpp
```diff
@@ -178,14 +176,37 @@ std::vector<EntityId> ProjectClause::process(std::vector<EntityId>&& entity_ids)
...
-                            const std::string_view name = output_column_;
-
-                            proc.segments_->back()->add_column(scalar_field(data_type, name), col.column_);
-                            auto new_col_range = std::make_shared<ColRange>(*proc.col_ranges_->back());
-                            ++new_col_range->second;
-                            proc.col_ranges_->back() = std::move(new_col_range);
+                            add_column(proc, col);
+                            output = push_entities(*component_manager_, std::move(proc));
+                        },
+                        [&proc, &output, this](const std::shared_ptr<Value>& val) {
```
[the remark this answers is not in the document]
alexowens90: Yes, it handles (for example) cases like
```
lib.write(sym, pd.DataFrame({"cond": [True], "col": [0]}))
lib.append(sym, pd.DataFrame({"col": [1]}))
q = QueryBuilder()
q["new_col"] = adb.where(q["cond"], q["col"], 5)
lib.read(sym, columns=["new_col"], query_builder=q).data
     new_col
0              0
1              5
```

cpp/arcticdb/processing/operation_dispatch_ternary.cpp
```diff
@@ -0,0 +1,582 @@
...
+    std::unique_ptr<Column> output_column;
+    std::shared_ptr<StringPool> string_pool;
+
+    details::visit_type(col.column_->type().data_type(), [&](auto col_tag) {
+        using col_type_info = ScalarTypeInfo<decltype(col_tag)>;
+        if constexpr (is_dynamic_string_type(col_type_info::data_type) ||
+                             is_numeric_type(col_type_info::data_type) ||
+                             is_bool_type(col_type_info::data_type)) {
+            if constexpr(is_dynamic_string_type(col_type_info::data_type)) {
+                // We do not need to construct a new string pool, as all the strings in the output column come from the
```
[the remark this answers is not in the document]
alexowens90: Yes that can happen, and will probably be the case most of the time. There isn't any value in trimming it at this stage, as the "old" string pool will still be kept around. Trimming happens (if `optimise_for_memory()` has been called on the input query builder) in `SegmentInMemoryImpl::filter`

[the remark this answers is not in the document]
alexowens90: Also worth pointing out that here, and particularly in the [column/column](https://github.com/man-group/ArcticDB/blob/3021be1e919be4b4f407ef553ad431fe939366b3/cpp/arcticdb/processing/operation_dispatch_ternary.cpp#L96) case, it is much faster if we do not need to construct a new string pool, by factors of x10-x40 in the benchmarks in the PR description

cpp/arcticdb/processing/ternary_utils.hpp
```diff
@@ -0,0 +1,425 @@
...
+    } else if (left_input_column.is_sparse()) {
+        // right_input_column is dense
+        output_sparse_map = (condition & left_input_column.sparse_map()) | ~condition;
+    } else if (right_input_column.is_sparse()) {
+        // left_input_column is dense
+        output_sparse_map = (~condition & right_input_column.sparse_map()) | condition;
+    } else {
+        // Both input columns are dense
+        // Bit vectors default initialise all bits to zero
+        output_sparse_map.flip();
```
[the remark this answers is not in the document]
alexowens90: Agreed, I didn't know about that API

cpp/arcticdb/processing/operation_dispatch_ternary.cpp
alexowens90: The other version is slightly different

cpp/arcticdb/processing/test/benchmark_ternary.cpp
alexowens90: Sure, initialised it once for the whole file

cpp/arcticdb/processing/test/benchmark_ternary.cpp
alexowens90: There's a [ticket](https://man312219.monday.com/boards/7852509418/pulses/8407446495?term=benchm&termColumns=XQAAAAIBAAAAAAAAAABAQfv____gAAAA) for it, but it isn't implemented yet

cpp/arcticdb/version/version_core.cpp
alexowens90: It was never really needed, the slices already come out of the processing pipeline in the right order. And it straight up broke `test_querybuilder_groupby_then_project` due to the changes to how the output segments are produced in `ProjectClause` now. Previously the new column was just tacked onto the last segment in the row-slice. Now it needs it's own segment, as it might have it's own stringpool.

vasil-pashov (approved):

PR #2105, opened 2025-01-07
Record stats on write

willdealtry:
Calculate min/max and unique count on write, roundtrip into storage for queries and as a first-pass for adaptive encodings


cpp/arcticdb/codec/test/test_codec.cpp
```diff
@@ -512,6 +512,86 @@ TEST(Segment, RoundtripTimeseriesDescriptorWriteToBufferV2) {
     ASSERT_EQ(decoded, copy);
 }
 
+TEST(Segment, RoundtripStatisticsV1) {
```
poodlewars, line 515: Do you think that there's an argument for only saving the new stats in the v2 encoding? And leaving v1 frozen as much as we can?

cpp/arcticdb/column_store/statistics.hpp
```diff
@@ -0,0 +1,211 @@
...
+    size_t get_unique_count() const {
+        return unique_count_;
+    }
+
+    FieldStatsImpl(FieldStats base) {
+        min_ = base.min_;
+        max_ = base.max_;
+        unique_count_ = base.unique_count_;
+        unique_count_precision_ = base.unique_count_precision_;
+        set_ = base.set_;
```
poodlewars, line 93: I don't quite get the value of a bitset here over separate booleans for each statistic

cpp/arcticdb/column_store/statistics.hpp
```diff
@@ -0,0 +1,211 @@
...
+        while (auto block = column_data.next<TagType>()) {
+            const RawType* ptr = block->data();
+            const size_t count = block->row_count();
+            if constexpr (is_numeric_type(TagType::DataTypeTag::data_type)) {
+                auto local_stats = generate_numeric_statistics<RawType>(std::span{ptr, count});
+                stats.compose<RawType>(local_stats);
+            } else if constexpr (is_dynamic_string_type(TagType::DataTypeTag::data_type)) {
+                auto local_stats = generate_string_statistics(std::span{ptr, count});
+                stats.compose<RawType>(local_stats);
+            }
```
poodlewars, line 206: else raise, like in the single block path?

cpp/arcticdb/column_store/statistics.hpp
```diff
@@ -0,0 +1,211 @@
...
+        FieldStatsImpl stats;
+        while (auto block = column_data.next<TagType>()) {
+            const RawType* ptr = block->data();
+            const size_t count = block->row_count();
+            if constexpr (is_numeric_type(TagType::DataTypeTag::data_type)) {
+                auto local_stats = generate_numeric_statistics<RawType>(std::span{ptr, count});
+                stats.compose<RawType>(local_stats);
+            } else if constexpr (is_dynamic_string_type(TagType::DataTypeTag::data_type)) {
+                auto local_stats = generate_string_statistics(std::span{ptr, count});
+                stats.compose<RawType>(local_stats);
```
poodlewars, line 205: The idea is that the string pool is shared across the whole column so we can just consider offsets in to it?

cpp/arcticdb/pipeline/frame_utils.hpp
poodlewars, line 1: This is just a refactor right? No logical changes intended?

poodlewars (approved):
I can't see the `HyperLogLog` implementation here, perhaps best to drop the enum value for it until it exists?

cpp/arcticdb/codec/test/test_codec.cpp
willdealtry: I thought about that (not least since it would be a chunk less work), but it's only a field in the protobuf header, it's opt-in (and off by default) and potentially when you have a use-case where you write once and read lots of times it's going to make a massive difference, so I didn't want to have to force people to switch encoding in order to get the benefit of it.

cpp/arcticdb/column_store/statistics.hpp
willdealtry: Just trying to keep things as small as possible for the binary encoding - I agree it doesn't make a lot of difference in V1

cpp/arcticdb/pipeline/frame_utils.hpp
willdealtry: Apart from adding the calls to calculate the stats, it's just breaking up that function that had become really huge and difficult to follow.

cpp/arcticdb/column_store/statistics.hpp
willdealtry: Yes. One of the performance advantages we'll be able to get is that at the moment when we want a unique string list for a segment (which is the precursor to various kinds of processing) we don't know how many unique items there are, whereas with stats we'll be able to just add that many offsets and then return. Which in the case where there's a low cardinality (like ['BID', 'ASK'] will make a huge difference.

alexowens90 (approved):

PR #2107, opened 2025-01-08
Prechecks for asv

grusev:

There are couple of checks that need to be done before making sure asv benchmark tests are ok to be merged:

Check that code of tests is ok and can run
> asv check --python=same

Checks that versions of benchmark tests are up to date in asv,conf.json
> asv run --bench just-discover --python=same

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/benchmarks/asv_checks.py
+++ b/python/benchmarks/asv_checks.py
@@ -1,0 +1,70 @@
+import os
+import subprocess
+import sys
+from typing import List
+
+
+def run_command(command: List[str], cwd) -> int:
+    result = subprocess.run(command, capture_output=True, text=True, cwd=cwd)
+
+    output = result.stdout
+    err_output = result.stderr
+    error_code = result.returncode
+
+    if error_code != 0:
+        print(f"Error Code: {error_code}") 
+
+    if not output is None:
+        print("Output:")
+        print(output)
+
+    if not err_output is None:
+        print("Error:", file=sys.stderr)
+        print(err_output, file=sys.stderr)
+
+    return error_code
+
+
+def file_unchanged(filepath, last_check_time):
+  try:
+    modified_time = os.path.getmtime(filepath)
+    return modified_time == last_check_time
+  except FileNotFoundError:
+    return False
+
+
+def get_project_root():
+    file_location = os.path.abspath(__file__)
+    return file_location.split("/python/benchmarks")[0]
+
+
+def perform_asv_checks():
+
+    path = get_project_root()
+
+    asv_config = f"{path}/asv.conf.json"
+    time = os.path.getatime(asv_config)
+
+    print("_" * 80)
+    print("""IMPORTANT: The tool checks CURRENT versions of asv tests along with asv.conf.json")
+            That means that if there are files that are not submitted yet,
+            they would need to be in order for completion of current PR""")
+    print("_" * 80)
+
+    print("\n\nCheck 1: Executing check for python cod of asv tests")
+    if run_command(["asv", "check", "--python=same"], path) != 0:
+        print("Please address all reported errors and submit code in the PR.", file=sys.stderr)
+    else:
+        print("Relax, no worries. Code is fine!")
+
+
+    print("\n\nCheck 2: Check that asv.conf.json has up to date latest versions of tests.")
+    if run_command(["asv", "run", "--bench", "just-discover", "--python=same"], path) != 0:
+        if not file_unchanged(asv_config, time):
+            print(f"""\n\n There are changes in asv test versions. 
+                  Make sure you submit the file {asv_config} in git repo""", file=sys.stderr)
+    else:
+        print("Great, there are no new versions of asv test either!")
+
+
+perform_asv_checks()


IvoDD (commented):
Left some minor comments. Apart from that I see the point in not running the asv checks on every build since it takes 35 minutes.

However couldn't we run it in parallel with the asv benchmarks and make it block merging? This shouldn't increase time to run the tests, since benchmarks take way longer. However I'm not sure if it's worth the extra CPU (will leave up to you or maybe @G-D-Petrov who knows more about our CI pipelines)

python/utils/asv_checks.py
```diff
@@ -0,0 +1,123 @@
...
+  try:
+    modified_time = os.path.getmtime(filepath)
+    return modified_time == last_check_time
+  except FileNotFoundError:
+    return False
+
+
+def get_project_root():
+    file_location = os.path.abspath(__file__)
+    return file_location.split("/python/")[0]
```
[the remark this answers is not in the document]
grusev: changed to increase confidence :-)

.github/workflows/asv_checks.yml
```diff
@@ -0,0 +1,79 @@
...
+      - 'asv.conf.json'
+
+  workflow_dispatch:
+    inputs:
+      do_nothing:
+        type: boolean
+        default: true
+
+jobs:
+  run-asv-check-script:
```
G-D-Petrov, line 27: I agree with @IvoDD,we don't need this to be a separate flow, just a separate job in the analysis_flow.
This way it will be easier for people, as they would have to check only 1 flow.
Let's move this job to the analysis_flow.yml, [similarly to the code_coverage job there](https://github.com/man-group/ArcticDB/blob/master/.github/workflows/analysis_workflow.yml#L107)

G-D-Petrov (changes requested):

.github/workflows/asv_checks.yml
```diff
@@ -0,0 +1,79 @@
+name: Run ASV Tests Check Python Script
```
poodlewars, line 1: I think if you call this "ASV Linting" or something it will be more obvious to people that it doesn't actually run the benchmarks

.github/workflows/asv_checks.yml
```diff
@@ -0,0 +1,79 @@
...
+  run-asv-check-script:
+    timeout-minutes: 120
+    runs-on: ubuntu-latest
+    container: ghcr.io/man-group/arcticdb-dev:latest
+    env:
+      SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
+      VCPKG_NUGET_USER: ${{secrets.VCPKG_NUGET_USER || github.repository_owner}}
+      VCPKG_NUGET_TOKEN: ${{secrets.VCPKG_NUGET_TOKEN || secrets.GITHUB_TOKEN}}
+      CMAKE_C_COMPILER_LAUNCHER: sccache
+      CMAKE_CXX_COMPILER_LAUNCHER: sccache
```
poodlewars, line 36: Don't get why we need these compiler settings given that the whole point is we don't need to build the wheel to run these linting checks

python/utils/asv_checks.py
```diff
@@ -0,0 +1,124 @@
+import hashlib
+import os
+import re
+import subprocess
+import sys
+from typing import List
+
+
+def error(mes):
```
poodlewars, line 9: Use `logging` not print statements in all PRs please

python/utils/asv_checks.py
```diff
@@ -0,0 +1,124 @@
...
+    result = subprocess.run(command, capture_output=True, text=True, cwd=cwd)
+
+    output = result.stdout
+    err_output = result.stderr
+    error_code = result.returncode
+
+    if not err_output is None:
+        error(err_output)
+        if error_code == 0:
+            print("ABOVE ERRORS DOES NOT AFFECT FINAL ERROR CODE = 0")
```
poodlewars, line 31: "DO NOT" not "DOES NOT"

python/utils/asv_checks.py
```diff
@@ -0,0 +1,124 @@
...
+    path = get_project_root()
+
+    sys.path.insert(0,f"{path}/python")
+    sys.path.insert(0,f"{path}/python/tests")
+
+    benchmark_config = f"{path}/python/.asv/results/benchmarks.json"
+    orig_hash = compute_file_hash(benchmark_config)
+
+    print("_" * 80)
+    print("""IMPORTANT: The tool checks CURRENT versions of asv tests along with asv.conf.json")
```
poodlewars, line 89: I don't understand this

poodlewars (commented):
-

.github/workflows/asv_checks.yml
grusev: to do the checks arcticdb library needs to be installed ... And installing released arcticdb does not help either as in benchmarks module we use libs from tests package (tested already as initally we wanted to be part of asv main worflow action) ... thus I need to invoke "pip install -ve ." which would do builds hence I coppied all things that would be needed for that from another workflow

.github/workflows/asv_checks.yml
grusev: agree makes lots of sence

.github/workflows/asv_checks.yml
grusev: ok, but note that this does something additional as check of versions of benchmark tests ... see bellow

python/utils/asv_checks.py
grusev: Changed to this:

    print("""IMPORTANT: The tool checks CURRENT ACTUAL versions of asv benchmark tests along with the one in benchmarks.json file.
            That means that if there are files that are not submitted yet (tests and benchmark.json),
            they would need to be in order for completion of current PR.
            benchmarks.json is updated with a version number calculated as a hash
            of the python test method. Thus any change of this method triggers different
            version. Hence you would need to update json file also.
            It happens automatically if you run following commandline:
             > asv run --bench just-discover --python=same  """)

poodlewars (approved):

G-D-Petrov (approved):

PR #2108, opened 2025-01-08
Notimplemented handling

G-D-Petrov:
#### Reference Issues/PRs
#203 in enterprise

Add a way to differentiate between a NotImplemented error and a regular StorageException.
This way we can properly handle S3 storages that haven't implemented the write_if_none operation.
Also updated the ReliableStorageLock to use the new behaviour.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -76,7 +76,7 @@ using Range = folly::Range<It>;
         raise<ErrorCode::E_S3_RETRYABLE>(fmt::format("Retry-able error: {}",
                                                      error_message_suffix));
     }
-
+    
     // We create a more detailed error explanation in case of NETWORK_CONNECTION errors to remedy #880.
     if (type == Aws::S3::S3Errors::NETWORK_CONNECTION) {
         error_message = fmt::format("Unexpected network error: {} "
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -68,6 +68,10 @@ using Range = folly::Range<It>;
                                                    error_message_suffix));
     }
 
+    if(err.GetExceptionName().find("NotImplemented") != std::string::npos) {
+        raise<ErrorCode::E_NOT_IMPLEMENTED>(fmt::format("Operation is not implemented for storage: {}", error_message_suffix));
+    }
+
     if (err.ShouldRetry()) {
         raise<ErrorCode::E_S3_RETRYABLE>(fmt::format("Retry-able error: {}",
                                                      error_message_suffix));
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -69,14 +69,14 @@ using Range = folly::Range<It>;
     }
 
     if(err.GetExceptionName().find("NotImplemented") != std::string::npos) {
-        raise<ErrorCode::E_NOT_IMPLEMENTED>(fmt::format("Operation is not implemented for storage: {}", error_message_suffix));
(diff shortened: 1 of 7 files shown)


cpp/arcticdb/storage/s3/nfs_backed_storage.hpp
```diff
@@ -36,7 +36,7 @@ class NfsBackedStorage final : public Storage {
     void do_write(Composite<KeySegmentPair>&& kvs) final;
 
     void do_write_if_none(KeySegmentPair&& kv [[maybe_unused]]) final {
-        storage::raise<ErrorCode::E_UNSUPPORTED_ATOMIC_OPERATION>("Atomic operations are only supported for s3 backend");
```
G-D-Petrov, line 39: The changes for NFS backed storages were made to make testing easier and to add the same general behavior as for S3 storages.
This can be reverted, if needed.

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -74,6 +74,10 @@ namespace s3 {
                 raise<ErrorCode::E_ATOMIC_OPERATION_FAILED>(fmt::format("Atomic operation failed: {}", error_message_suffix));
             }
 
+            if(err.GetExceptionName().find("NotImplemented") != std::string::npos) {
```
poodlewars, line 77: I assume there's no more robust way to figure this out? Not enough info the the `type`?

cpp/arcticdb/storage/s3/nfs_backed_storage.hpp
```diff
@@ -54,7 +54,7 @@ class NfsBackedStorage final : public Storage {
     }
 
     bool do_supports_atomic_writes() const final {
-        return false;
+        return true;
```
poodlewars, line 57: This should still be `false` shouldn't it? Is this to make the testing easier like you mention above? I think it's too confusing having this `true` and then always blowing up when this backend does an atomic write.

poodlewars (commented):
Needs tests.

I think this looks sensible. I discussed with Ivo that we should perhaps add some logic to test a conditional write by actually writing an object to the backend, as the different S3 providers implement this very differently. But we'll probably want this logic anyway.

cpp/arcticdb/storage/s3/nfs_backed_storage.hpp
willdealtry: Please don't change those before I merge the storage refactor, it will make a huge amount of work for me.

I was considering getting rid of write_if_none in favour of having it as an option to the write method, as the API is getting quite crowded

cpp/arcticdb/storage/s3/detail-inl.hpp
G-D-Petrov: The type of the error is UNKNOWN(100) is a but of a catch-all and might result in false positives.
I prefer to handle only this particular case.

cpp/arcticdb/storage/s3/nfs_backed_storage.hpp
G-D-Petrov: Added a way to configure this.
Having the ability to enable these writes for the nfs storage makes it easy to write an integration test for the ReliableLock in the enterprise repo

cpp/arcticdb/util/error_code.hpp
```diff
@@ -86,6 +86,7 @@ inline std::unordered_map<ErrorCategory, const char*> get_error_category_names()
     ERROR_CODE(5020, E_UNEXPECTED_S3_ERROR) \
     ERROR_CODE(5021, E_S3_RETRYABLE) \
     ERROR_CODE(5022, E_ATOMIC_OPERATION_FAILED) \
+    ERROR_CODE(5023, E_NOT_IMPLEMENTED) \
```
IvoDD, line 89: Nit: This sounds a bit like something that we did not implement yet. It's more that the storage (PURE) does not support it.
What do you think about `E_NOT_IMPLEMENTED_BY_STORAGE` or `E_NOT_SUPPORTED_BY_STORAGE`?

IvoDD (approved):

PR #2117, opened 2025-01-14
Chunk up incomplete segments by rows when they are staged

poodlewars:
This change applies row slicing when incomplete segments are staged, so by default if we staged an incomplete with 1M rows we now write 10 segments each with 100k rows. The motivation for this is to put some bound on the memory used when compacting incompletes, as the compaction step has to hold an entire segment in memory at a time.

I also refactored the code to separate things out better. Previously the staging code was shared for 3 quite different features:

- Staging an incomplete
- Sorting and staging an incomplete
- Appending an incomplete for tick streaming, forming a linked list of APPEND_DATA keys

I've separated these out to 3 separate functions now. The sorting case is quite different to the case without sorting, because to sort we need to transform the `InputTensorFrame` in to a `Segment` first, whereas when not sorting we can happily work with the incomplete still as an `InputTensorFrame`.

The staging an incomplete path is now much more similar to the normal write path.

We do not apply column slicing here as that makes it harder for compaction to infer the schema.

I renamed `append_map.hpp` to `incompletes.hpp` as I think that's more meaningful to most of us.

Monday: 8066284387


cpp/arcticdb/version/test/test_version_store.cpp
```diff
@@ -290,6 +290,155 @@ TEST_F(VersionStoreTest, CompactIncompleteDynamicSchema) {
...
+    auto read_result = test_store_->read_dataframe_version(symbol, VersionQuery{}, read_query, ReadOptions{}, handler_data);
+    const auto& seg = read_result.frame_data.frame();
+
+    ASSERT_EQ(seg.row_count(), num_rows_per_incomplete * num_incompletes);
+
+    count = 0;
+    auto col1_pos = seg.column_index( "thing1").value();
+    auto col2_pos = seg.column_index( "thing2").value();
+    auto col3_pos = seg.column_index( "thing3").value();
+    auto col4_pos = seg.column_index( "thing4").value();
```
alexowens90, line 356: nit: weird whitespace

cpp/arcticdb/version/test/test_version_store.cpp
```diff
@@ -290,6 +290,155 @@ TEST_F(VersionStoreTest, CompactIncompleteDynamicSchema) {
...
+
+    for (size_t i = 0; i < num_incompletes; ++i) {
+        for(size_t j = 0; j < num_rows_per_incomplete; ++j ) {
+            auto idx = seg.scalar_at<uint64_t>(count, 0);
+            log::version().info("Index value {} count {}", idx.value(), count);
```
alexowens90, line 361: Probably don't want to leave this in?

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -115,6 +115,22 @@ def test_parallel_write(basic_store):
     assert len(get_append_keys(basic_store, sym)) == 0
 
 
+def test_parallel_write_chunking(lmdb_version_store_tiny_segment):
+    lib = lmdb_version_store_tiny_segment # row size 2 column size 2
+    lib_tool = lib.library_tool()
+    sym = "sym"
+    df = pd.DataFrame({"col_1": np.arange(10), "col_2": np.arange(10), "col_3": np.arange(10)}, pd.date_range("2024-01-01", periods=10))
```
alexowens90, line 122: I don't see any tests of slicing string data, which is inherently more complicated than numeric data

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -214,72 +217,195 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
+    WriteOptions write_options = options.write_options;
+    write_options.column_group_size = std::numeric_limits<size_t>::max(); // column slicing not supported yet (makes it hard
+    // to infer the schema we want after compaction)
+
+    auto slicing_policy = FixedSlicer{write_options.column_group_size, write_options.segment_row_size};
+    auto slices = slice(*frame, slicing_policy);
+
+    if (slices.empty()) {
+        // We still write in this case because a user might only stage empty segments. After the user finalizes
+        // they will just get an empty dataframe.
```
alexowens90, line 320: From the looks of it this is just matching the previous behaviour, is it already tested somewhere?

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -214,72 +217,195 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
+        size_t existing_rows = 0;
+        auto timeseries_desc = index_descriptor_from_frame(frame, existing_rows, std::nullopt);
+        util::check(!timeseries_desc.fields().empty(), "Expected fields not to be empty in incomplete segment");
```
alexowens90, line 323: Can a user trigger this by staging `pd.DataFrame()`?

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -214,72 +217,195 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
+    auto bucketize_dynamic = frame->bucketize_dynamic;
+    bool sparsify_floats{false};
+
+    TypedStreamVersion typed_stream_version{stream_id, VersionId{0}, KeyType::APPEND_DATA};
+    return folly::collect(folly::window(std::move(slice_and_rowcount),
+        [frame, slicing_policy, key = std::move(key),
+         store, sparsify_floats, typed_stream_version = std::move(typed_stream_version),
+            bucketize_dynamic, de_dup_map, desc, norm_meta, user_meta](
+            auto&& slice) {
+            auto typed_stream_version_copy = typed_stream_version;
```
alexowens90, line 360: Unneeded, `get_partial_key_gen` copies second arg already

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -214,72 +217,195 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
+    } else {
+        do_sort(segment, *options.sort_columns);
+    }
+
+    auto timeseries_desc = index_descriptor_from_frame(frame, 0, std::nullopt);
+    auto norm_meta = timeseries_desc.proto().normalization();
+    auto tsd = pack_timeseries_descriptor(frame->desc, frame->num_rows, std::nullopt, std::move(norm_meta));
+    segment.set_timeseries_descriptor(tsd);
+
+    bool is_sorted = options.sort_on_index || (options.sort_columns && options.sort_columns->at(0) == segment.descriptor().field(0).name());
```
alexowens90, line 263: I think we only use sortedness in the context of timeseries? So if column 0 isn't a timestamp index this wouldn't be right?

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -214,72 +217,195 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
+            if (is_sorted) {
+                seg.descriptor().set_sorted(SortedValue::ASCENDING);
             }
-            do_sort(mutable_seg, cols);
-        } else {
-            do_sort(mutable_seg, *sort_columns);
+            const auto local_index_start = IdxType::start_value_for_segment(seg);
+            const auto local_index_end = IdxType::end_value_for_segment(seg);
+            stream::StreamSink::PartialKey pk{KeyType::APPEND_DATA, 0, stream_id, local_index_start, local_index_end};
+            auto fut = store->write(pk, std::move(seg))
```
alexowens90, line 278: Shouldn't these writes be windowed like in the non-sorting case?

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -214,72 +217,195 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
+    }
+
+    auto timeseries_desc = index_descriptor_from_frame(frame, 0, std::nullopt);
+    auto norm_meta = timeseries_desc.proto().normalization();
+    auto tsd = pack_timeseries_descriptor(frame->desc, frame->num_rows, std::nullopt, std::move(norm_meta));
+    segment.set_timeseries_descriptor(tsd);
+
+    bool is_sorted = options.sort_on_index || (options.sort_columns && options.sort_columns->at(0) == segment.descriptor().field(0).name());
+
+    auto segments = segment.split(options.write_options.segment_row_size);
```
alexowens90, line 265: This could potentially be expensive as it involves memcopying the columns to shorter columns. Not for now, but we should have a ticket to be able to do a zero-copy `slice_and_write` based on a `SegmentInMemory`.

cpp/arcticdb/version/test/test_version_store.cpp
poodlewars: I don't see it?

cpp/arcticdb/stream/incompletes.cpp
poodlewars: This is covered by `test_staged_segment_is_only_empty_dfs` - there's always a field representing an index even for empty DFs. Added a case for `pd.DataFrame()` to that test.

cpp/arcticdb/stream/incompletes.cpp
poodlewars: Yeah, `test_staged_segment_is_only_empty_dfs`

cpp/arcticdb/stream/incompletes.cpp
poodlewars: ty, refactor SNAFU

cpp/arcticdb/stream/incompletes.cpp
poodlewars: This is the current behaviour. Our checks on the sortedness all seem to also check whether the index is a timeseries,

```
void sorted_data_check_append(const InputTensorFrame& frame, index::IndexSegmentReader& index_segment_reader){
    if (!index_is_not_timeseries_or_is_sorted_ascending(frame)) {
        sorting::raise<ErrorCode::E_UNSORTED_DATA>("When calling append with validate_index enabled, input data must be sorted");
    }
    sorting::check<ErrorCode::E_UNSORTED_DATA>(
        !std::holds_alternative<stream::TimeseriesIndex>(frame.index) ||
        index_segment_reader.tsd().sorted() == SortedValue::ASCENDING,
        "When calling append with validate_index enabled, the existing data must be sorted");
}
```

so frame what I can tell, they're separate concepts at the moment.

cpp/arcticdb/stream/incompletes.cpp
poodlewars: Good point - 8264207371

python/tests/unit/arcticdb/version_store/test_sort.py
```diff
@@ -169,6 +171,29 @@ def test_stage_finalize_sort_index(arctic_library):
...
+
+    lib.stage(symbol, df1_shuffled, validate_index=False, sort_on_index=True, sort_columns=None)
+
+    lib_tool = lib.library_tool()
+    data_keys = lib_tool.find_keys_for_symbol(KeyType.APPEND_DATA, symbol)
+    # We don't apply column slicing when staging incompletes, do apply row slicing
+    assert len(data_keys) == 25
+
+    ref_keys = lib_tool.find_keys_for_symbol(KeyType.APPEND_REF, symbol)
+    assert not ref_keys
```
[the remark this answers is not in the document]
poodlewars: Thanks, this caught a bug fixed in c044a935820925ab0577b22df45ece4823cbd02a

cpp/arcticdb/version/test/test_version_store.cpp
alexowens90: Me neither now

cpp/arcticdb/stream/incompletes.cpp
poodlewars: We spoke about this. I've kept the current behaviour but we might as well fix it while we're here. We should only apply sortedness when column 0 is a timestamp index, anything else is a (pre-existing) bug. I'll fix in this PR.

willdealtry (approved):

poodlewars:
I'm going to merge this. There's this small change in the benchmarks but I think that's to be expected:

```
| Change   | Before [f92a5ac7]    | After [268a601d]    |   Ratio | Benchmark (Parameter)                                     |
|----------|----------------------|---------------------|---------|-----------------------------------------------------------|
| +        | 133±3ms              | 226±0.3ms           |    1.7  | basic_functions.BasicFunctions.time_write_staged(1500000) |
| +        | 93.1±3ms             | 157±0.9ms           |    1.69 | basic_functions.BasicFunctions.time_write_staged(1000000) |
```

poodlewars:
| Change   | Before [f92a5ac7]    | After [268a601d]    |   Ratio | Benchmark (Parameter)                                     |
|----------|----------------------|---------------------|---------|-----------------------------------------------------------|
| +        | 133±3ms              | 226±0.3ms           |    1.7  | basic_functions.BasicFunctions.time_write_staged(1500000) |
| +        | 93.1±3ms             | 157±0.9ms           |    1.69 | basic_functions.BasicFunctions.time_write_staged(1000000) |

PR #2118, opened 2025-01-14
Handle very old normalization metadata RangeIndexes

alexowens90:
See comment in diff for context


willdealtry (approved):

poodlewars (approved):
Ideally would have a test (maybe a C++ unit test simulating the old style PandasIndex structure) for this

PR #2119, opened 2025-01-14
Vector comparisons

willdealtry:
Implement some portable vectorized comparisons

### DIFF
--- a/cpp/arcticdb/column_store/chunked_buffer.cpp
+++ b/cpp/arcticdb/column_store/chunked_buffer.cpp
@@ -68,7 +68,7 @@ std::vector<ChunkedBufferImpl<BlockSize>> split(const ChunkedBufferImpl<BlockSiz
 }
 
 template std::vector<ChunkedBufferImpl<64>> split(const ChunkedBufferImpl<64>& input, size_t nbytes);
-template std::vector<ChunkedBufferImpl<3968>> split(const ChunkedBufferImpl<3968>& input, size_t nbytes);
+template std::vector<ChunkedBufferImpl<4032ul>> split(const ChunkedBufferImpl<4032ul>& input, size_t nbytes);
 
 // Inclusive of start_byte, exclusive of end_byte
 template <size_t BlockSize>
@@ -112,6 +112,6 @@ ChunkedBufferImpl<BlockSize> truncate(const ChunkedBufferImpl<BlockSize>& input,
 }
 
 template ChunkedBufferImpl<64> truncate(const ChunkedBufferImpl<64>& input, size_t start_byte, size_t end_byte);
-template ChunkedBufferImpl<3968> truncate(const ChunkedBufferImpl<3968>& input, size_t start_byte, size_t end_byte);
+template ChunkedBufferImpl<4032ul> truncate(const ChunkedBufferImpl<4032ul>& input, size_t start_byte, size_t end_byte);
 
 } //namespace arcticdb
--- a/cpp/arcticdb/util/min_max_float.hpp
+++ b/cpp/arcticdb/util/min_max_float.hpp
@@ -1,0 +1,118 @@
+#include <cstdint>
+#include <limits>
+#include <type_traits>
+#include <cstddef>
+#include <algorithm>
+
+#include <arcticdb/util/vector_common.hpp>
+
+namespace arcticdb {
+
+template<typename T>
+using vector_type __attribute__((vector_size(64))) = T;
+
+template<typename T>
+class FloatMinFinder {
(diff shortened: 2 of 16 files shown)


cpp/arcticdb/util/min_max_integer.hpp
```diff
@@ -0,0 +1,133 @@
...
+            T* lanes = reinterpret_cast<T*>(&vector_max);
+            for (size_t i = 0; i < lane_count; i++) {
+                lanes[i] = init_max;
+            }
+        }
+
+        size_t num_vectors = n / lane_count;
+        const VectorType* vdata = reinterpret_cast<const VectorType*>(data);
+        for (size_t i = 0; i < num_vectors; i++) {
+            VectorType v = vdata[i];
```
alexowens90, line 57: Compiler should take care of it, but in a `-O0` world, `vdata++` would be more efficient

cpp/arcticdb/util/min_max_integer.hpp
```diff
@@ -0,0 +1,133 @@
...
+    static ReturnType find(const T* data, size_t n) {
+        util::check(n > 0, "Empty array provided");
+        static_assert(is_supported_int<T>::value, "Type must be integer");
+        static_assert(std::is_integral_v<T>, "Type must be integral");
+
+        using VectorType = vector_type<T>;
+        constexpr size_t lane_count = sizeof(VectorType) / sizeof(T);
+
+        T init_min = std::numeric_limits<T>::max();
+        T init_max = std::numeric_limits<T>::min();
```
alexowens90, line 38: It smells a bit that there are max related variables when only min is being computed and vice versa, but I can't see a way to refactor that removes them without a lot of code duplication

cpp/arcticdb/util/min_max_integer.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+        VectorType vector_min, vector_max;
+        if constexpr (ComputeMin) {
+            T* lanes = reinterpret_cast<T*>(&vector_min);
```
alexowens90, line 42: I don't think this is needed, `vector_min[i]` should do the obvious thing
https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html

cpp/arcticdb/util/test/test_min_max_integer.cpp
```diff
@@ -0,0 +1,486 @@
...
+    auto std_duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+    std::cout << "std::minmax_element time: " << std_duration.count() << "ms\n";
+
+    EXPECT_EQ(result.min, *std_result.first);
+    EXPECT_EQ(result.max, *std_result.second);
+}
+
+#ifdef HAS_VECTOR_EXTENSIONS
+
+class MinMaxStressTest : public ::testing::Test {
```
alexowens90, line 284: Could move to a `benchmark`

cpp/arcticdb/util/test/test_min_max_integer.cpp
alexowens90, line 1: There's a lot of duplication in here, could we parametrize across supported types, and then move the random tests to rapidcheck?

cpp/arcticdb/column_store/chunked_buffer.cpp
```diff
@@ -68,7 +68,7 @@ std::vector<ChunkedBufferImpl<BlockSize>> split(const ChunkedBufferImpl<BlockSiz
 }
 
 template std::vector<ChunkedBufferImpl<64>> split(const ChunkedBufferImpl<64>& input, size_t nbytes);
-template std::vector<ChunkedBufferImpl<3968>> split(const ChunkedBufferImpl<3968>& input, size_t nbytes);
+template std::vector<ChunkedBufferImpl<4032ul>> split(const ChunkedBufferImpl<4032ul>& input, size_t nbytes);
```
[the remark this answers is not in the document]
alexowens90: If there's any data that has been written into storage in 3968 byte blocks, will they still be readable without further modifications?

cpp/arcticdb/util/min_max_float.hpp
```diff
@@ -0,0 +1,93 @@
...
+T find_float_min(const T* data, size_t n) {
+    return FloatExtremumFinder<T, FloatMinComparator<T>>::find(data, n);
+}
+
+template<typename T>
+T find_float_max(const T* data, size_t n) {
+    return FloatExtremumFinder<T, FloatMaxComparator<T>>::find(data, n);
+}
+
+#else
```
alexowens90, line 76: Same comments below the `#else` as for `min_max_integer.hpp`

PR #2137, opened 2025-01-23
asv comparison tests pandas dataframe memory creation vs pandas.dataframe read/write and arcticdb_lmdb read/writes

grusev:
#### Reference Issues/PRs

Those asv tests aim to compare memory usage between above mentioned operations. Seeing all of them on one graph could tell us how efficient we are in memory management.

If we consider that we will have also the ability to make same tests for Polars and Amazon S3 with time, such benchmarks could provide insights for our performance

             OUTPUT -------->
             DF generated 88.47616171836853                                                                                                 ok
[20.00%] ··· comparison_benchmarks.ComparisonBenchmarks.peakmem_create_dataframe                                                              3.16G
[40.00%] ··· comparison_benchmarks.ComparisonBenchmarks.peakmem_read_dataframe_arctic                                                         4.95G
[60.00%] ··· comparison_benchmarks.ComparisonBenchmarks.peakmem_read_dataframe_parquet                                                         5.4G
[80.00%] ··· comparison_benchmarks.ComparisonBenchmarks.peakmem_write_dataframe_arctic                                                        3.45G
[100.00%] ··· comparison_benchmarks.ComparisonBenchmarks.peakmem_write_dataframe_parquet                                                       2.77G

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


G-D-Petrov (approved):
LGTM

python/benchmarks/comparison_benchmarks.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    number = 5
+    timeout = 60000
+
+    LIB_NAME = "compare"
+    URL = "lmdb://compare"
+    SYMBOL = "dataframe"
+    NUMBER_ROWS = 3000000
+
+    def setup_cache(self):
```
poodlewars, line 37: I think this should just return the `df` given that the df and the dict are really the same thing

python/benchmarks/comparison_benchmarks.py
```diff
@@ -0,0 +1,97 @@
...
+            "region" : str_col(10, size),
+            "last_published_date" : random_dates(size),
+            "model_snapshot_id" :  random_integers(size, np.int64), 
+            "period" : str_col(20, size), 
+            "observation_type" : str_col(10, size),
+            "ric" : str_col(10, size),
+            "dtype" : str_col(10, size),
+        }       
+    
+    def peakmem_create_dataframe(self, tpl):
```
poodlewars, line 80: I get the idea with this but it's odd for us to have a benchmark that we have no control over - would be very odd for a Pandas regression to block our PRs.

poodlewars (approved):

python/benchmarks/comparison_benchmarks.py
grusev: Creation of the dataframe is basis for comparison how effective we will be on other graphs. Also it shows that although create and write done separetly have certain mem requirements, the mem requirements of both of them are not equal to the sum of each one. (same for read/write ets.

I would not say block, but warn and require action if the memory usage suddently grows. Why it grew? IS it ok? wtc. Such questions will pop in our heads which otherwise will not ... So see this as a feedback loop, that will remind us to thinl when something unusual happens.

Overall seing all processes in one test and graph will make a difference over time as this coul improve our knowledge and ways of management performance and requirements

python/benchmarks/comparison_benchmarks.py
grusev: The return of dict is needed for dataframe creation test, which is important for understanging all later numbers and their dynamics, thus this cannot be avoided

PR #2152, opened 2025-01-29
DBA tools for ReliableStorageLock

IvoDD:
Adds the options to:
- inspect latest state of the lock
- force take or free the lock. This is done by not respecting the expiration and jumping a few ids forward.

Also adds a C++ test to verify the new APIs work as expected.

The APIs will be exposed for use in arcticdb-enterprise.

#### Reference Issues/PRs
Related to 8013103642 monday ticket

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/util/reliable_storage_lock-inl.hpp
+++ b/cpp/arcticdb/util/reliable_storage_lock-inl.hpp
@@ -56,9 +56,19 @@ inline AcquiredLockId get_next_id(std::optional<AcquiredLockId> maybe_prev) {
     return 0;
 }
 
+inline AcquiredLockId get_force_next_id(std::optional<AcquiredLockId> maybe_prev) {
+    // When taking a lock with force we use a higher lock id so we can force acquire it even if there is high lock contention.
+    return maybe_prev.value_or(0) + 10;
+}
+
 template <class ClockType>
 StreamId ReliableStorageLock<ClockType>::get_stream_id(AcquiredLockId lock_id) const {
     return fmt::format("{}{}{}", base_name_, SEPARATOR, lock_id);
+}
+
+template <class ClockType>
+RefKey ReliableStorageLock<ClockType>::get_ref_key(AcquiredLockId lock_id) const {
+    return RefKey{get_stream_id(lock_id), KeyType::ATOMIC_LOCK};
 }
 
 inline AcquiredLockId extract_lock_id_from_stream_id(const StreamId& stream_id) {
@@ -99,7 +109,7 @@ void ReliableStorageLock<ClockType>::clear_old_locks(const std::vector<AcquiredL
     // 5. Process B decides to clear lock 7 since it's not the latest
     // 6. Process A succeeds in taking lock 7
     for (auto lock_id : lock_ids) {
-        auto lock_key = RefKey{get_stream_id(lock_id), KeyType::ATOMIC_LOCK};
+        auto lock_key = get_ref_key(lock_id);
         if (get_expiration(lock_key) + REMOVE_AFTER_TIMEOUTS * timeout_ < now) {
             to_delete.emplace_back(lock_key);
         }
@@ -111,13 +121,14 @@ template <class ClockType>
 ReliableLockResult ReliableStorageLock<ClockType>::try_take_lock() const {
     auto [existing_locks, latest] = get_all_locks();
     if (latest.has_value()) {
-        auto expires = get_expiration(RefKey{get_stream_id(latest.value()), KeyType::ATOMIC_LOCK});
+        auto expires = get_expiration(get_ref_key(latest.value()));
         if (expires > ClockType::nanos_since_epoch()) {
             // An unexpired lock exists
             return LockInUse{};
         }
     }
-    return try_take_next_id(existing_locks, latest);
+    auto next_id = get_next_id(latest);
+    return try_take_id(existing_locks, next_id);
 }
 
 template<class ClockType>
@@ -153,7 +164,8 @@ ReliableLockResult ReliableStorageLock<ClockType>::try_extend_lock(AcquiredLockI
         // We have lost the lock while holding it (most likely due to timeout).
         return LockInUse{};
     }
-    return try_take_next_id(existing_locks, latest);
+    auto next_id = get_next_id(latest);
+    return try_take_id(existing_locks, next_id);
 }
 
 template <class ClockType>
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/util/reliable_storage_lock-inl.hpp
```diff
@@ -56,11 +56,21 @@ inline AcquiredLockId get_next_id(std::optional<AcquiredLockId> maybe_prev) {
...
+}
+
 template <class ClockType>
 StreamId ReliableStorageLock<ClockType>::get_stream_id(AcquiredLockId lock_id) const {
     return fmt::format("{}{}{}", base_name_, SEPARATOR, lock_id);
 }
 
+template <class ClockType>
+RefKey ReliableStorageLock<ClockType>::get_ref_key(AcquiredLockId lock_id) const {
+    return RefKey{get_stream_id(lock_id), KeyType::ATOMIC_LOCK};
```
willdealtry, line 71: Isn't there already a lock KeyType? We want to have as few key types as we can get away with ideally.

cpp/arcticdb/codec/encoded_field_collection.hpp
```diff
@@ -68,7 +68,10 @@ class EncodedFieldCollection {
     EncodedFieldCollection() = default;
 
     [[nodiscard]] EncodedFieldCollection clone() const {
-        return {data_.clone(), offsets_.clone()};
+        auto output = EncodedFieldCollection{data_.clone(), offsets_.clone()};
```
willdealtry, line 71: Is this constructor for EncodedFieldCollection used for anything other than passing clones in? If not then it would be better to change the constructor to require the offset and count, otherwise it seems like someone will make this mistake again

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -162,7 +162,7 @@ KeySegmentPair do_read_impl(
         VariantKey&& variant_key,
         const std::string& root_folder,
         const std::string& bucket_name,
-        const S3ClientInterface& s3_client,
+        S3ClientInterface& s3_client,
```
willdealtry, line 165: Would it have been easier to make the mutex mutable in the mock store?

cpp/arcticdb/util/reliable_storage_lock.hpp
```diff
@@ -45,12 +54,21 @@ class ReliableStorageLock {
     ReliableLockResult try_extend_lock(AcquiredLockId acquired_lock) const;
     void free_lock(AcquiredLockId acquired_lock) const;
     timestamp timeout() const;
+    // Below APIs are for admin management of the lock
+    std::optional<ActiveLock> inspect_latest_lock() const;
+    // Forcefully takes a new lock without waiting for any timeouts. If custom_timeout is negative the resulting lock
```
willdealtry, line 59: Some empty lines might make it easier to read

cpp/arcticdb/util/reliable_storage_lock.hpp
```diff
@@ -26,6 +26,15 @@ using LockInUse = std::monostate;
...
+struct ActiveLock {
+    AcquiredLockId lock_id;
+    timestamp expiration;
+
+    bool operator == (const ActiveLock& other) const {
+        return lock_id == other.lock_id && expiration == other.expiration;
+    }
+};
+
 // The ReliableStorageLock is a storage lock which relies on atomic If-None-Match Put and ListObject operations to
```
willdealtry, line 38: Is it worth putting in a line or two about what the id is for? It's not obvious why you couldn't just have a lock object in the storage that writes either wrote or didn't

cpp/arcticdb/storage/mock/s3_mock_client.cpp
```diff
@@ -109,32 +113,42 @@ S3Result<std::monostate> MockS3Client::put_object(
...
-    if (header == PutHeader::IF_NONE_MATCH && s3_contents.contains({bucket_name, s3_object_name})) {
+    auto pos = s3_contents_.find({bucket_name, s3_object_name});
+    if (header == PutHeader::IF_NONE_MATCH && pos != s3_contents_.end() && pos->second.has_value()) {
         return {precondition_failed_error};
     }
 
-    s3_contents.insert_or_assign({bucket_name, s3_object_name}, std::move(segment));
+    // When we write Segments we occasionally don't fill in the size_. This is fine as it's not needed for writing.
+    // However, it is required when reading so we need to calculate it. It's easier to do on write.
+    auto size[[maybe_unused]] = segment.calculate_size();
```
poodlewars, line 123: the attribute is in a weird place, I would expected it at the start of the line?

cpp/arcticdb/async/tasks.hpp
```diff
@@ -396,9 +397,18 @@ struct CopyCompressedInterStoreTask : async::BaseTask {
                 key_segment_pair.set_key(*key_to_write_);
             }
 
-            for (auto & target_store : target_stores_) {
+            for (auto i=0u; i<target_stores_.size(); ++i) {
+                auto& target_store = target_stores_[i];
                 try {
-                    target_store->write_compressed_sync(key_segment_pair);
+                    auto key_segment_pair_copy = key_segment_pair;
+                    if (i < target_stores_.size()-1) {
```
poodlewars, line 404: Could just iterate with an iterator and check for the end with `std::prev(target_stores_.end())`

cpp/arcticdb/async/tasks.hpp
```diff
@@ -396,9 +397,18 @@ struct CopyCompressedInterStoreTask : async::BaseTask {
...
+                        // If this is not the last target we want to do a deep copy because write_compressed_sync will
+                        // move out of the underlying segment.
+                        auto key_copy = key_segment_pair.variant_key();
+                        auto segment_copy = key_segment_pair.segment().clone();
```
poodlewars, line 408: Worried about the performance impact of this. Have you done any measurement of that?

cpp/arcticdb/util/reliable_storage_lock-inl.hpp
IvoDD: yeah there is. I initially decided to use a new key type mostly for easier inspection e.g. via library tool and to not get confused with the locks used for symbol list compaction.

I think changing these two will make sense but we will need to figure out a rollout story because this lock is already in use. I think it will be sensible to tackle alongside the migration to IfMatch based lock. (ref 7963752098)

cpp/arcticdb/codec/encoded_field_collection.hpp
IvoDD: It is used [here](https://github.com/man-group/ArcticDB/blob/958a6ffd4b83c725b6da4a99b3efadde102aa4cd/cpp/arcticdb/codec/segment_header.cpp#L101). It seemed non-obvious how to derive the `offset` and `count` there, so I left it unchanged.

cpp/arcticdb/async/tasks.hpp
IvoDD: The last commit adds benchmarks for this.

In short, the impact should not be significant because `Segment::clone` is quite fast compared to the follow up `write_compressed_sync`. For lmdb on an SSD it's `5ms` vs `35ms`. For S3 and other slower storages the difference would be much greater. 

Also from the benchmark results it looks like the compiler will optimize away some of the cost of cloning.

poodlewars (approved):

G-D-Petrov (approved):

PR #2156, opened 2025-01-30
Document how we serialize metadata and warn when it is pickled

poodlewars:

PR #2160, opened 2025-01-31
Simplify our unpickling logic and add tests for Python 2 pickles

poodlewars:
This is fixing the issue that Alex pointed out here https://github.com/man-group/ArcticDB/pull/2156#issuecomment-2624547655 .

We do not need the `str` case - the stuff we are decoding is always bytes - otherwise we would not be in the pickling fallback. Nor do we need to specify a `bytes` encoding when decoding Python 2 objects. The "latin-1" fallback case is handled by `pd.read_pickle` so it's redundant for us to have it too.

The tests I added also pass without my changes to `_normalization.py` and I think suffice to show there is no behaviour change here.

### Old Logic

Old logic for pickles inside the msgpack was:

- If pickled in python 2, unpack msgpack with raw=True, pickle.loads with encoding="ascii".
- If pickled in python 3, unpack msgpack with raw=False, pickle.loads with encoding="bytes".

Where raw means:

```
        """    :param bool raw:
                If true, unpack msgpack raw to Python bytes.
                Otherwise, unpack to Python str by decoding with UTF-8 encoding (default).

        """
```

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -1050,11 +1050,11 @@ class MsgPackNormalizer(Normalizer):
             # If stored in Python2 we want to use raw while unpacking.
             # https://github.com/msgpack/msgpack-python/blob/master/msgpack/_unpacker.pyx#L230
             data = unpackb(data, raw=True)
-            return Pickler.read(data, pickled_in_python2=True)
+            return Pickler.read(data)
 
         if code == MsgPackSerialization.PY_PICKLE_3:
             data = unpackb(data, raw=False)
-            return Pickler.read(data, pickled_in_python2=False)
+            return Pickler.read(data)
 
         return ExtType(code, data)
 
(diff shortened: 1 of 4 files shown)


jamesmunro (approved):

python/arcticdb/version_store/_normalization.py
```diff
@@ -1070,21 +1070,8 @@ def _msgpack_unpackb(self, buff, raw=False):
 
 class Pickler(object):
     @staticmethod
-    def read(data, pickled_in_python2=False):
-        if isinstance(data, str):
-            return pickle.loads(data.encode("ascii"), encoding="bytes")
-        elif isinstance(data, str):
```
poodlewars, line 1076: This else case was clearly wrong https://github.com/man-group/ArcticDB/pull/2156#issuecomment-2624547655

python/arcticdb/version_store/_normalization.py
```diff
@@ -1070,21 +1070,8 @@ def _msgpack_unpackb(self, buff, raw=False):
...
-            if not pickled_in_python2:
-                # Use the default encoding for python2 pickled objects similar to what's being done for PY2.
-                return pickle.loads(data, encoding="bytes")
-
-        try:
-            # This tries normal pickle.loads first then falls back to special Pandas unpickling. Pandas unpickling
-            # handles Pandas 1 vs Pandas 2 API breaks better.
-            return pd.read_pickle(io.BytesIO(data))
-        except UnicodeDecodeError as exc:
```
poodlewars, line 1085: `pd.read_pickle` has this fallback already

IvoDD (approved):

PR #2161, opened 2025-01-31
Read and write snapshot metadata from the correct place

poodlewars:
Between v4.5.0 and v5.2.1 we changed to save snapshot metadata in the segment header in a `TimeseriesDescriptor` field. This was changed in the binary encoding change 516d16968f0 .

Before v4.5.0 we wrote it directly in a `UserDefinedMetadata` field.

This means that the current code falsely thinks that snapshots written before `v4.5.0` have no snapshot metadata.

This PR:

- Changes readers to check both places for any metadata, so we can read snapshot metadata written before `v4.5.0` again
- Changes writers back to writing directly in a `UserDefinedMetadata` field

This means that clients with versions between v4.5.0 and v5.2.1 will not see snapshot metadata written by modern versions (they will say it is `None`) and will need to upgrade.

Since the metadata gets saved in the `SegmentHeader` in both schemes, it is not possible to write data in a way that is compatible with both early readers and readers between `v4.5.0` and `v5.2.1`. We could make the alternative decision and break readers older than `v4.5.0`.

We should apply a similar change to `arcticc`'s reading code so it can understand snapshot metadata written by newer versions.

### DIFF
--- a/cpp/arcticdb/version/snapshot.cpp
+++ b/cpp/arcticdb/version/snapshot.cpp
@@ -39,11 +39,21 @@ void write_snapshot_entry(
         ARCTICDB_DEBUG(log::snapshot(), "Adding key {}", key);
         snapshot_agg.add_key(key);
     }
+
     // Serialize and store the python metadata in the journal entry for snapshot.
     if (!user_meta.is_none()) {
-        TimeseriesDescriptor timeseries_descriptor;
-        python_util::pb_from_python(user_meta, *timeseries_descriptor.mutable_proto().mutable_user_meta());
-        snapshot_agg.set_timeseries_descriptor(timeseries_descriptor);
+        arcticdb::proto::descriptors::UserDefinedMetadata user_meta_proto;
+        google::protobuf::Any output = {};
+        python_util::pb_from_python(user_meta, user_meta_proto);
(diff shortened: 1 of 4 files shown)


willdealtry (approved):

IvoDD (approved):
LGTM, only minor not required comments

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -39,11 +39,21 @@ void write_snapshot_entry(
...
-        TimeseriesDescriptor timeseries_descriptor;
-        python_util::pb_from_python(user_meta, *timeseries_descriptor.mutable_proto().mutable_user_meta());
-        snapshot_agg.set_timeseries_descriptor(timeseries_descriptor);
+        arcticdb::proto::descriptors::UserDefinedMetadata user_meta_proto;
+        google::protobuf::Any output = {};
+        python_util::pb_from_python(user_meta, user_meta_proto);
+        output.PackFrom(user_meta_proto);
+        snapshot_agg.set_metadata(std::move(output));
+
+        // Bewared: Between v4.5.0 and v5.2.1 we only saved this metadata on the
```
poodlewars, line 51: typo bewared

PR #2165, opened 2025-02-04
Real Storage asv perf POC

grusev:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -9,6 +9,10 @@ on:
         description: Tag of the ArcticCB development image to use for benchmark and code coverage flows
         type: string
         default: latest
+      lmdb_suite_to_run:
+        description: Run LMDB suite (true) and real storage (false)
+        type: boolean
+        default: true
 
   schedule: # Schedule the job to run at 12 a.m. daily
     - cron: '0 0 * * *'
@@ -50,6 +54,7 @@ jobs:
       run_all_benchmarks: ${{ inputs.run_all_benchmarks || false }}
       run_on_pr_head: ${{ github.event_name == 'pull_request_target' }}
       dev_image_tag: ${{ inputs.dev_image_tag || 'latest' }}
+      lmdb_suite_to_run: ${{ github.event_name == 'schedule' && true || github.event.inputs.lmdb_suite_to_run }}
 
   publish_benchmark_results_to_gh_pages:  
     name: Publish benchmark results to gh-pages
--- a/.github/workflows/benchmark_commits.yml
+++ b/.github/workflows/benchmark_commits.yml
@@ -6,6 +6,7 @@ on:
       commit:             {required: true, type: string, description: commit hash that will be benchmarked}
       run_on_pr_head:     {required: false, default: false, type: boolean, description: Specifies if the benchmark should run on PR head branch}
       dev_image_tag:      {required: false, default: 'latest', type: string, description: Tag of the ArcticDB development image}
+      lmdb_suite_to_run:  {required: true, type: boolean, default: true, description: Default LMDB benchmark (if false will run real storage)}
 jobs:
   start_ec2_runner:
     uses: ./.github/workflows/ec2_runner_jobs.yml
@@ -76,7 +77,11 @@ jobs:
         shell: bash -el {0}
         run: | 
           git config --global --add safe.directory .
-          python -m asv run -v --show-stderr ${{ inputs.commit }}^!
+          if [ "${{ github.event.inputs.lmdb_suite_to_run }}" == "true" ]; then
+            SUITE='^(?!real_).*'
+          else
+            SUITE='^(real_).*'
+          python -m asv run -v --show-stderr --bench $SUITE ${{ inputs.commit }}^!
         
       - name: Benchmark against master
         if: github.event_name == 'pull_request_target' && inputs.run_all_benchmarks == false
--- a/python/arcticdb/util/utils.py
+++ b/python/arcticdb/util/utils.py
@@ -4,6 +4,9 @@ Use of this software is governed by the Business Source License 1.1 included in 
 As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
 """
 
+from datetime import timedelta
+import random
+import string
 from typing import List, Tuple, Union
 import sys 
 if sys.version_info >= (3, 8): 
@@ -15,6 +18,10 @@ import pandas as pd
 
 from arcticdb.util.test import create_datetime_index, get_sample_dataframe
 from arcticdb.version_store.library import Library
+
+# Type definitions - supported by arctic
+ArcticIntType = Union[np.uint8, np.uint16, np.uint32, np.uint64, np.int8, np.int16, np.int32, np.int64]
+ArcticFloatType = Union[np.float64, np.float32]
 
 class  TimestampNumber:
     """
@@ -331,4 +338,156 @@ def stage_chunks(lib: Library, symbol:str, cachedDF:CachedDFGenerator, start_ind
         num_rows_staged = num_rows_staged + chunk_size
         iter= iter + 1
         total = total + chunk_size
+
+
+class RandomStringPool:
+    """
+    A class that will return a list of randomly selected strings from a string pool
+    with certain sized of each string and limited number of strings in the pool
+    """
+
+    def __init__(self, str_length: int, pool_size: int):
+        self.__pool = ListGenerators.generate_random_string_pool(str_length, pool_size)
+
+    def get_list(self, size: int) -> List[str]:
+        return [random.choice(self.__pool) for _ in range(size)]
+
+
+class ListGenerators:
+    """
+    Specialized class of utility functions for creating random lists of various type
+    """
+
+    @classmethod
+    def generate_random_floats(cls, dtype: ArcticFloatType, 
+                      size: int, minV: float = None, maxV: float = None, round_to: int = None
+                      ) -> List[ArcticFloatType]:
+        finfo = np.finfo(dtype)
+        if minV is None:
+            minV = max(finfo.min, -sys.float_info.max)
+            maxV = min(finfo.max, sys.float_info.max)
+        if round_to is None:
+            return np.random.uniform(minV, maxV, size).astype(dtype)
+        else :
+            return np.round(np.random.uniform(minV, maxV, size), round_to).astype(dtype)
+        
+    @classmethod
+    def generate_random_string_pool(cls, str_length: int, pool_size: int) -> List[str]:
+        unique_values = set()
+        while len(unique_values) < pool_size:
+            unique_values.add(ListGenerators.random_string(str_length))
+        return list(unique_values)
+
+    @classmethod
+    def generate_random_strings(cls, str_size: int, length: int) -> List[str]:
+        return [ListGenerators.random_string(str_size) for _ in range(length)]
+    
+    @classmethod
+    def generate_random_ints(cls, dtype: ArcticIntType, 
+                            size: int, minV: int = None, maxV: int = None
+                            ) -> List[ArcticIntType]:
+        # We do not generate integers outside the int64 range
+        platform_int_info = np.iinfo("int_")
+        iinfo = np.iinfo(dtype)
+        if minV is None:
+            minV = max(iinfo.min, platform_int_info.min)
+            maxV = min(iinfo.max, platform_int_info.max)
+        else:
+            minV = max(iinfo.min, platform_int_info.min, minV)
+            maxV = min(iinfo.max, platform_int_info.max, maxV)
+        return np.random.randint(
+            minV,
+            maxV,
+            size=size,
+            dtype=dtype
+        )
+    
+    @classmethod
+    def generate_random_bools(cls, size: int) -> List[bool]:
+        return np.random.choice([True, False], size=size) 
+    
+    @classmethod
+    def random_string(cls, length: int):
+        return "".join(random.choice(string.ascii_uppercase 
+                                       + string.digits + string.ascii_lowercase + ' ') for _ in range(length))
+
+
+class DFGenerator:
+    """
+    Easy generation of DataFrames, via fluent interface
+    """
+
+    def __init__(self, size: int):
+        self.__size = size
+        self.__data = {}
+        self.__types = {}
+        self.__df = None
+        self.__index = None
+
+    def generate_dataframe(self) -> pd.DataFrame:
+        if self.__df is None:
+            if self.__size == 0:
+                # Apply proper types for dataframe when size is 0
+                self.__df = pd.DataFrame({col: pd.Series(dtype=typ) for col, typ in self.__types.items()})
+            else:
+                self.__df = pd.DataFrame(self.__data)
+            if self.__index is not None:
+                self.__df.index = self.__index
+        return self.__df
+
+    def add_int_col(self, name: str, dtype: ArcticIntType = np.int64) -> 'DFGenerator':
+        list = ListGenerators.generate_random_ints(dtype, self.__size)
+        self.__data[name] = list
+        self.__types[name] = dtype
+        return self
+    
+    def add_int_col_ex(self, name: str, min: int, max: int, dtype: ArcticIntType = np.int64) -> 'DFGenerator':
+        list = ListGenerators.generate_random_ints(dtype, self.__size, min, max)
+        self.__data[name] = list
+        self.__types[name] = dtype
+        return self
+    
+    def add_float_col(self, name: str, dtype: ArcticFloatType = np.float64) -> 'DFGenerator':
+        list = ListGenerators.generate_random_floats(dtype, self.__size)
+        self.__data[name] = list
+        self.__types[name] = dtype
+        return self
+
+    def add_float_col_ex(self, name: str, min: float, max: float, round_at: int, dtype: ArcticFloatType = np.float64
+                         ) -> 'DFGenerator':
+        list = ListGenerators.generate_random_floats(dtype, self.__size, min, max, round_at)
+        self.__data[name] = list
+        self.__types[name] = dtype
+        return self
+
+    def add_string_col(self, name: str, str_size: int, num_unique_values: int = None) -> 'DFGenerator':
+        """
+        Generates a list of strings with length 'str_size', and if 'num_unique_values' values is None
+        the list will be of unique values if 'num_unique_values' is a number then this will be the length
+        pf the string pool of values
+        """
+        list = []
+        if num_unique_values is None:
+            list = ListGenerators.generate_random_strings(str_size, self.__size)
+        else:
+            list = RandomStringPool(str_size, num_unique_values).get_list(self.__size)
+        self.__data[name] = list
+        self.__types[name] = str
+        return self
+    
+    def add_bool_col(self, name: str) -> 'DFGenerator':
+        list = np.random.choice([True, False], size=self.__size) 
+        self.__data[name] = list
+        self.__types[name] = bool
+        return self
+    
+    def add_timestamp_indx(self, name_col:str, freq:Union[str , timedelta , pd.Timedelta , pd.DateOffset], 
+                            start_time: pd.Timestamp = pd.Timestamp(0)) -> 'DFGenerator':
+        self.__index = pd.date_range(start=start_time, periods=self.__size, freq=freq, name=name_col)
+        return self
+    
+    def add_range_indx(self, name_col:str, start:int = 0, step:int = 1, dtype: str = 'int') -> 'DFGenerator':
+        stop = (self.__size + start) * step
+        self.__index = pd.Index(range(start, stop, step), dtype=dtype, name=name_col)
+        return self
 
--- a/python/benchmarks/real_list_operations.py
+++ b/python/benchmarks/real_list_operations.py
@@ -1,0 +1,236 @@
+
+
+import time
+from typing import List
+import numpy as np
+import pandas as pd
+
+from arcticdb.options import LibraryOptions
+from arcticdb.util.utils import DFGenerator
+from arcticdb.version_store.library import Library
+from benchmarks.real_storage.libraries_creation import LibrariesBase, Storage
+
+
+#region Setup classes
+
+class SymbolLibraries(LibrariesBase):
+
+    def __init__(self, type:Storage = Storage.LMDB):
+        super().__init__(type)
+        self.ac = self.get_arctic()
+        self.df = self.generate_df(10)
+
+    def get_parameter_list(self):
+        return [500, 1000]
+    
+    def get_library_options(self):
+        return LibraryOptions(rows_per_segment=20, columns_per_segment=20)
+    
+    def get_library_names(self, num_symbols) -> List[str]:
+        return [f"PERM_LIST_OPS_SYMBOL_{num_symbols}", f"MOD_LIST_OPS_SYMBOL_{num_symbols}"]  
+    
+    def generate_df(self, row_num:int) -> pd.DataFrame:
+        """
+        Dataframe that will be used in list
+        """
+        st = time.time()
+        print("Dataframe generation started.")
+        df = (DFGenerator(row_num)
+            .add_int_col("int8", np.int8)
+            .add_int_col("int16", np.int16)
+            .add_int_col("int32", np.int32)
+            .add_int_col("int64")
+            .add_int_col_ex("uint8", 1, 20, np.uint64)
+            .add_int_col_ex("uint64", 100, 1999, np.uint64)
+            .add_float_col("float16",np.float32)
+            .add_float_col_ex("float64",-100.0, 200.0, 4)
+            .add_string_col("string100", 100, 200)
+            .add_string_col("string2", 2)
+            .add_bool_col("bool")
+            .add_timestamp_indx("time", 's', pd.Timestamp("2000-1-1"))
+            ).generate_dataframe()
+        print(f"Dataframe {row_num} rows generated for {time.time() - st} sec")
+        return df
+
+    def setup_library_with_symbols(self, num_syms):
+        """
+        Sets up required number of symbols per that library
+        """
+        self.lib = self.get_library(num_syms) 
+        print(f"Started creating library for {num_syms} symbols.")
+        st = time.time()
+        for num in range(num_syms):
+            sym = self.get_symbol_name(num)
+            self.lib.write(sym, self.df, metadata=self.df)
+            print(f"created {sym}")
+        print(f"Library '{self.lib}' created {num_syms} symbols COMPLETED for :{time.time() - st} sec")
+
+    def reset_symbols_cache(self):
+        pass #TBD
+
+    def setup_all(self):
+        st = time.time()
+        for sym in self.get_parameter_list():
+            self.setup_library_with_symbols(sym)
+        print(f"Total time {time.time() - st}")
+
+    def check_ok(self) -> bool:
+        for num in self.get_parameter_list():
+            lib = self.get_library(num)
+            list = lib.list_symbols()
+            print(f"number symbols {len(list)} in library {lib}")
+            if len(list) != num:
+                return False
+        return True
+        
+
+class VersionLibraries(SymbolLibraries):
+    """
+    A library for list versions testings is composed of X symbols.
+    Symbols are sym_name_[0,X)
+    Each symbol has as many versions as the index in the name
+    Thus first symbol will have 1, 2nd - 2, 3rd - 3 etc
+    Total version expected for N symbols:
+        Sn = n(n+1)/2
+        which is SUM(1..N)
+    """
+
+    def __init__(self, type:Storage = Storage.LMDB):
+        super().__init__(type)
+        self.ac = self.get_arctic()
+
+    def get_parameter_list(self):
+        return [5, 6]
+    
+    def get_library_names(self, num_symbols) -> List[str]:
+        return [f"PERM_LIST_OPS_VERSION_{num_symbols}", f"MOD_LIST_OPS_VERSION_{num_symbols}"]  
+
+    def setup_symbols_with_versions(self, num_symbols):
+        self.lib = self.get_library(num_symbols)
+        for sym_num in range(num_symbols):
+            sym = self.get_symbol_name(sym_num)
+            print(f"Generating {sym_num} versions for symbol [{sym}].")
+            st = time.time()
+            if sym_num > 0:
+                for num in range(sym_num):
+                    self.lib.write(sym, self.df, metadata=self.df)
+            print(f"Generating {sym_num} versions for symbol [{sym}]. COMPLETED for :{time.time() - st} sec")
+    
+    def setup_all(self):
+        st = time.time()
+        for sym in self.get_parameter_list():
+            self.setup_library_with_symbols(sym)
+            self.setup_symbols_with_versions(sym)
+        print(f"Total time {time.time() - st}")
+        
+    def check_ok(self) -> bool:
+
+        def expected_version(n):
+            return  int((n*(n+1)) /2)
+        
+        if not super().check_ok():
+            return False
+        for num in self.get_parameter_list():
+            lib = self.get_library(num)
+            vers = len(lib.list_versions())
+            #for v in lib.list_versions():
+            #    print(v)
+            exp = expected_version(num)
+            print(f"There are {vers} in library {lib} , expected {exp}")
+            if vers != exp:
+                return False
+        return True
+
+#endregion
+
+class AWS_ListSymbols:
+    """
+    This class is responsible for all checks on AWS storage
+    """
+
+    rounds = 1
+    number = 1 
+    repeat = 3 
+    min_run_count = 1
+    warmup_time = 0
+
+    timeout = 12000
+
+    SETUP_CLASS = SymbolLibraries(Storage.AMAZON)
+
+    params = SETUP_CLASS.get_parameter_list()
+    param_names = ["num_syms"]
+
+    def get_creator(self):
+        return AWS_ListSymbols.SETUP_CLASS
+
+    def setup_cache(self):
+        aws = self.get_creator()
+        if not aws.check_ok():
+            aws.setup_all()
+        else:
+            aws.reset_symbols_cache()
+
+    def setup(self, num_syms):
+        self.aws = self.get_creator()
+        self.lib = self.aws.get_library(num_syms)
+
+    def time_list_symbols(self, num_syms):
+        self.lib.list_symbols()
+
+    def time_has_symbol_nonexisting(self, num_syms):
+        self.lib.has_symbol("250_sym")        
+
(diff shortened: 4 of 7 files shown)


python/benchmarks/real_storage/libraries_creation.py
```diff
@@ -0,0 +1,179 @@
...
+        SetupConfig()
+    
+    def remove_libraries(self):
+        ## Remove only LMDB libs as they need to
+        if self.type == Storage.LMDB:
+            ac = self.get_arctic()
+            for lib in self.__libraries:
+                ac.delete_library(lib)
+
+    def get_arctic(self):
```
G-D-Petrov, line 72: I'd call this `get_arctic_client`

python/benchmarks/real_storage/libraries_creation.py
```diff
@@ -0,0 +1,179 @@
...
+import numpy as np
+import pandas as pd
+
+
+from arcticdb.arctic import Arctic
+from arcticdb.version_store.library import Library
+
+
+## Amazon s3 storage URL
+AWS_URL_TEST = 's3s://s3.eu-west-1.amazonaws.com:arcticdb-asv-real-storage?aws_auth=true'
```
G-D-Petrov, line 22: I'd prefer if you use the s3 fixture that we have https://github.com/man-group/ArcticDB/blob/master/build_tooling/transform_asv_results.py#L94

python/benchmarks/real_storage/libraries_creation.py
```diff
@@ -0,0 +1,179 @@
...
+    def get_modifyable_library(self, num_symbols: int = 1) -> Library:
+        """
+        Returns library to read write and delete after done.
+        """
+        ac = self.get_arctic()
+        lib_opts = self.get_library_options()
+        lib_name = self.get_library_names(num_symbols)[1]
+        self.__libraries.add(lib_name)
+        if lib_opts is None:
+            return ac.get_library(lib_name, create_if_missing=True)
```
G-D-Petrov, line 124: There is no need for the if here, can be just:
```
return ac.get_library(lib_name, create_if_missing=True, 
                                  library_options=lib_opts)
```

python/benchmarks/real_storage/libraries_creation.py
```diff
@@ -0,0 +1,179 @@
...
+            self.ac = Arctic(self.arctic_url)
+        return self.ac
+    
+    def get_library_options(self):    
+        """
+        Override to create non-default lib opts
+        """
+        return None
+    
+    def get_library_names(self, num_symbols: int = 1) -> List[str]:
```
G-D-Petrov, line 93: shouldn't this be an abstractmethod?

python/benchmarks/real_storage/libraries_creation.py
```diff
@@ -0,0 +1,179 @@
...
+            self.type = type
+        self.ac = None
+        self.__libraries = set()
+        SetupConfig()
+    
+    def remove_libraries(self):
+        ## Remove only LMDB libs as they need to
+        if self.type == Storage.LMDB:
+            ac = self.get_arctic()
+            for lib in self.__libraries:
```
G-D-Petrov, line 69: I am not sure what is the point of __libraries?
Wouldn't it be easier to just call ac.list_libraries()?

python/benchmarks/real_storage/libraries_creation.py
```diff
@@ -0,0 +1,179 @@
...
+                return False
+        return True
+    
+    def delete_libraries(self, nameStarts: str, confirm=False):
+        """
+        Deletes all libraries starting with specified string
+        """
+        assert confirm, "deletion not confirmed!"
+        ac = self.get_arctic()
+        list = ac.list_libraries()
```
G-D-Petrov, line 170: nit: library_list ot lib_list would be a better name.
Also this can be written as lib_list = [lib for lib in ac.list_libraries() if lib.startswith(nameStarts)]
Then you won't need the other check

python/benchmarks/real_read_write.py
```diff
@@ -0,0 +1,182 @@
...
+    repeat = 3 # On LMDB we repeat at least 3 times
+    min_run_count = 1
+    warmup_time = 0
+
+    timeout = 12000
+
+    SETUP_CLASS: ReadBenchmarkLibraries = ReadBenchmarkLibraries(Storage.LMDB)
+
+    params = SETUP_CLASS.get_parameter_list()
+    param_names = ["num_rows"]
```
G-D-Petrov, line 117: if you are going to have `get_parameter_list()`, it might be more consistent to also have `get_parameter_names_list()`

python/benchmarks/real_list_operations.py
```diff
@@ -0,0 +1,236 @@
...
+
+from arcticdb.options import LibraryOptions
+from arcticdb.util.utils import DFGenerator
+from arcticdb.version_store.library import Library
+from benchmarks.real_storage.libraries_creation import LibrariesBase, Storage
+
+
+#region Setup classes
+
+class SymbolLibraries(LibrariesBase):
```
G-D-Petrov, line 16: I am confused as to why we need these 2 classes SymbolLibraries and VersionLibraries.
They seem to do the same thing, but the versions one just writes more version and has a different set of parameter.

I think this can be refactored to be more concise.

G-D-Petrov (changes requested):

python/arcticdb/util/utils.py
```diff
@@ -332,3 +339,155 @@ def stage_chunks(lib: Library, symbol:str, cachedDF:CachedDFGenerator, start_ind
...
+        self.__types[name] = dtype
+        return self
+    
+    def add_float_col(self, name: str, dtype: ArcticFloatType = np.float64) -> 'DFGenerator':
+        list = ListGenerators.generate_random_floats(dtype, self.__size)
+        self.__data[name] = list
+        self.__types[name] = dtype
+        return self
+
+    def add_float_col_ex(self, name: str, min: float, max: float, round_at: int, dtype: ArcticFloatType = np.float64
```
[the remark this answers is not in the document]
grusev: you cannot have in python 2 functions with different parameter sets, hence there are 2 main scenarios for usafe float and integer function - one just get random vals (implicitly it means whole set of vals from -min to +max) and the other scenario is I need only between ceratin interval , and for  floats additional opt that you do not care for more than X digits, which could be important.

for each scenario there is a separate function to limit options only to those that matters. Usually in C# etc functions that have most arguments are with "_ex" ending, here I was not exactly sure what to add. But definately from suportability perspective going away from standard python pattern where you go and handle all in code is perhaps not best

>>>> Also we already have similar functionality [here]
What I could I reused. What I should not touch I tried not to touch, instead it is more clean approach to start from clean start, especially if you look at floats ...

Overall, what is is there is something used here and there and should be change with quite much care ...

At the other hand I think we need new more systematic approach which is aligned with intelisense and grouping operations in related classes of utility functions. This would be more evident and readily available and prevent mistakes.

Therefore going forward in more complex scenarios this approach will show its benefits through faster writing tests with more steps

python/benchmarks/real_list_operations.py
grusev: Because the versions currently depend on number of symbols.
While number of symbols has linear growth, the number of versions grow in a symbol grow as sum(1..n) where n is the number of symbol. With just 50 symbols you get > 300 version, 100 > 5000
For test of list symbols you need more symbols usually thousands.
Creating version library for 1000 is very time consuming and expensive, we could do that once and perhaps will do that as we have permanent storage, but initial generation will be huge.

Why I have done the versions in this model?

  - it is close to real life - each symbol has different number of versions
  - you have symbol with max number of versions which is another test, it can be easily created anoter test for n/2 versions for symbol

Still we have not decided yet to use the persistance feature or not and regenerate each test data when we need it (I think we should use persistance feature)

python/benchmarks/real_storage/libraries_creation.py
grusev: Reason for using separate fixture is that there will be cases when we delete libraries. In accident  we might delete library for asv results. Thus keeping the set separate other data is best. There could be times when we could wipe all  data. If we combine those we will have to take care of what is used and for what reason which complicates things.

python/benchmarks/real_storage/libraries_creation.py
grusev: True - old Java habbits

python/benchmarks/real_storage/libraries_creation.py
grusev: Keep track of what libraries are created. To be deleted at the end of asv tests by asv evntually when I find the right [;ace to do that, not to clutter shared storages (or those on our WSLs)
Currently libs are not deleted when LMDB is used which is ok for Transient GH runner but not optimal yet for us.
And for shares storages we have to take care of libs we do not need, first moment is when they are not needed at end of asv tests. That however might be tricky to do ...

Option 1: The class wipes data after tests are done - this example ... yet due to asv neture to open many processes there is no hook yet known to me "at end of all tests". So that work is still on progress if there is any way to do it will do it

Option 2: If no solution is founf we might need to have a new job invoked after this job to wipe out all "MOD_" and "TEMP_" libraries, or the keep only "PERM_" libs, last one is best IMO.  (It might not be a job but a python script in current job activated when real storage is used)
Option 2a: for personal WSL wipe out /tmp on regular bases should be practiced but I will leave that engineers for now

python/benchmarks/real_storage/libraries_creation.py
grusev: Renamed the var but prefer to keep the current way of implementation. Reason:

- delete is destructive operation and must always print what is deleted. That is the reason I added first line 'asser confirm' and I will add also additional check for lenth of string required > 4. So that by accident we do not delete anything that we do not want. 

Yet this function is not used yet, but will be at proper time if I find solution to above problem

grusev: fixed

python/benchmarks/real_list_operations.py
```diff
@@ -0,0 +1,272 @@
...
+
+import time
+from typing import List
+import numpy as np
+import pandas as pd
+
+from arcticdb.options import LibraryOptions
+from arcticdb.util.utils import DFGenerator
+from arcticdb.version_store.library import Library
+from benchmarks.real_storage.libraries_creation import LibrariesBase, Storage
```
G-D-Petrov, line 17: It is strange that the classes here are extending LibrariesBase which is in benchmarks.real_storage.libraries_creation
Maybe it needs to be here?

python/benchmarks/real_read_write.py
```diff
@@ -0,0 +1,203 @@
...
+
+    rounds = 1
+    number = 3 # invokes 3 times the test runs between each setup-teardown 
+    repeat = 1 # defines the number of times the measurements will invoke setup-teardown
+    min_run_count = 1
+    warmup_time = 0
+
+    timeout = 12000
+
+    SETUP_CLASS: ReadBenchmarkLibraries = ReadBenchmarkLibraries(Storage.LMDB)
```
G-D-Petrov, line 118: I don't understand why you are calling these SETUP_CLASS, their actual names are ...Libraries.
It would be clearer if they are called Something like ReadBenchmarkSetup

G-D-Petrov (changes requested):

python/arcticdb/util/utils.py
```diff
@@ -332,3 +339,138 @@ def stage_chunks(lib: Library, symbol:str, cachedDF:CachedDFGenerator, start_ind
         iter= iter + 1
         total = total + chunk_size
 
+
+class RandomStringPool:
+    """
+    A class that will return a list of randomly selected strings from a string pool
+    with certain sized of each string and limited number of strings in the pool
```
poodlewars, line 346: s/sized/size

python/arcticdb/util/utils.py
```diff
@@ -332,3 +339,138 @@ def stage_chunks(lib: Library, symbol:str, cachedDF:CachedDFGenerator, start_ind
...
+    """
+
+    def __init__(self, str_length: int, pool_size: int):
+        self.__pool = ListGenerators.generate_random_string_pool(str_length, pool_size)
+
+    def get_list(self, size: int) -> List[str]:
+        return [random.choice(self.__pool) for _ in range(size)]
+
+
+class ListGenerators:
```
poodlewars, line 356: These should not be in a class but instead be free functions in their own module. The methods should have an optional seed parameter so we can deterministically reproduce problems or recreate data

python/benchmarks/real_list_operations.py
poodlewars: I agree with Georgi P, we can have a utility that creates a library with mean MS symbols, mean MV versions per symbol, stddev SDS symbols and stddev SDV versions per symbol and that should be able to meet all use-cases?

python/benchmarks/real_list_operations.py
```diff
@@ -0,0 +1,272 @@
...
+    def setup_cache(self):
+        '''
+        Always provide implementation of setup_cache in
+        the child class
+
+        And always return the arctic url which should 
+        be first parameter for setup, tests and teardowns
+        '''
+        aws = AWS_ListSymbols.SETUP_CLASS
+        if not aws.check_ok():
```
poodlewars, line 193: invert the if statement to avoid double negative

python/benchmarks/real_list_operations.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def setup_cache(self):
+        '''
+        Always provide implementation of setup_cache in
+        the child class
+
+        And always return the arctic url which should 
+        be first parameter for setup, tests and teardowns
+        '''
+        aws = AWS_ListSymbols.SETUP_CLASS
```
poodlewars, line 192: Remove this variable `aws`? It's confusing given we also have `self.aws`

python/benchmarks/real_read_write.py
```diff
@@ -0,0 +1,203 @@
...
+    def get_library_names(self, num_symbols=1) -> List[str]:
+        return ["PERM_READ", "MOD_READ"]        
+
+    def get_parameter_list(self):
+        """
+            We might need to have different number of rows per different types of storages depending on the
+            speed of storage. LMDB as fastest might be the only different than any other
+        """
+        # Initially tried with those values but turn out they are too slow for Amazon s3
+        # Still they are avail in the storage
```
poodlewars, line 44: s/avail/available

python/benchmarks/real_read_write.py
```diff
@@ -0,0 +1,203 @@
...
+        # rows = [5_000_000, 10_000_000, 20_000_000]
+        rows = [1_000_000, 2_000_000]
+        if self.type == Storage.LMDB:
+            rows = [2_500_000, 5_000_000]
```
poodlewars, line 48: `if-else` clearer than overwriting `rows` in the LMDB case

python/benchmarks/real_read_write.py
poodlewars: Agree, `ReadBenchmarkSettings`?

python/benchmarks/real_storage/libraries_creation.py
```diff
@@ -0,0 +1,332 @@
...
+        if type is not None:
+            self.type = type
+        self.arctic_url = arctic_url
+        if self.arctic_url is not None:
+            if type is None:
+                if "lmdb" in self.arctic_url.lower():
+                    self.type = Storage.LMDB
+                if "s3s://" in self.arctic_url.lower() or "s3://" in self.arctic_url.lower():
+                    self.type = Storage.AMAZON
+                raise Exception(f"unsupported storage type: {type}")
```
poodlewars, line 106: Why do you allow specifying the `type` separately to the URI? Doesn't the URI always imply the type?

poodlewars: Using logic like the `Arctic` constructor to figure out the backend may be cleaner,

```
        for adapter_cls in self._LIBRARY_ADAPTERS:
            if adapter_cls.supports_uri(uri):
                _cls = adapter_cls
                break
```

python/benchmarks/real_storage/libraries_creation.py
poodlewars: Agree with GPetrov, just use `ac.list_libraries()`

python/benchmarks/real_read_write_wide.py
```diff
@@ -0,0 +1,189 @@
...
+    def setup(self, storage_info, params):
+        '''
+        This setup method for read and writes can be executed only once
+        No need to be executed before each test. That is why we define 
+        `repeat` as 1
+        '''
+        ## Construct back from arctic url the object
+        self.lmdb = VaryingSizeSymbolLibrary.fromStorageInfo(storage_info)
+        sym = self.lmdb.get_symbol_name(params)
+        print("STORAGE INFO: ", storage_info)
```
poodlewars, line 168: Remove print statements throughout. Replace any that you think are essential with log statements

python/benchmarks/real_storage/libraries_creation.py
poodlewars, line 1: I think this is over-abstracted at the moment, and it will be hard for the next person to follow through the layers of abstraction to figure out how to add the next benchmark. I don't understand why we can't just use normal ASV benchmarks, and some free functions that they can all use to do their setup and teardown? ie preferring composition over inheritance.

Then once we have a few benchmarks we can see what is really getting duplicated across them and have a better chance of finding the right abstractions.

In general we should try not to write our own wrappers around open source libraries like ASV or Pandas unless there is a huge benefit to us since we will then start to lose out on improvements they make to their APIs and their superior documentation.

poodlewars (changes requested):

python/benchmarks/real_storage/libraries_creation.py
poodlewars: Is the results data in its own bucket at the moment? If it isn't, it should be

python/arcticdb/util/utils.py
```diff
@@ -332,3 +339,138 @@ def stage_chunks(lib: Library, symbol:str, cachedDF:CachedDFGenerator, start_ind
         iter= iter + 1
         total = total + chunk_size
 
+
+class RandomStringPool:
```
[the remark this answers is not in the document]
grusev: In terms of code perhaps not, but in terms of clarity it provides much more value when used in tests, for creating dataframes easily and in more compact and understandable manner. As rule of thumb the e2e, scenario tests should be understood by anyone, even not knowing the programming language. Right now our tests are more close to unit tests.

What I try to do with such small steps is to build a kind of DSL (domain specific language) but with programming terms. I do not aim to for real DSL, but more like DSL readable code. Thus items that tend to repeat itself lots of time and have more simple explanation and are building blocks for other (in our case DFGenerator would be translated into such building blocks DSL elements so that the test intention is clear even without looking at specific implementations.

As an example: 

`       gen = DFGenerator(rows) 
        pool = RandomStringPool(100, 2000)
        for i in range(int(cols/10)):
            gen = (gen.add_bool_col(f"bool_{i}")
                .add_float_col(f"float64_{i}", np.float64)
                .add_float_col(f"float32_{i}", np.float32,)
                .add_string_col(f"str4_{i}", 4)
                .add_string_enum_col(f"enum100_{i}", pool)
                )
        ............  
`
Yes we could scrap the class but the result code will be less meaninful and harder to read (if we agree that the test should be understood by almost anyone despite their knowledge in programming)

python/benchmarks/real_storage/libraries_creation.py
grusev: The method was not used, but could be useful in the future in this form:

```
    def remove_all_created_libraries(self, confirm=False):
        """
        Removes all libraries that this instance created
        """
        assert confirm, "Deletion of all libraries must be confirmed"
        ac = self.get_arctic_client()
        for lib in self.__libraries:
            ac.delete_library(lib)
```

As we are on shared storage(s) ac.list_libraries() will not work well in this case. The instance must be able to remove only its own created libraries.

python/benchmarks/real_list_operations.py
grusev: Now looking back to the overall design I see room for making it more clean ... Will make changes to make it cleaner and easier to follow

python/benchmarks/real_storage/libraries_creation.py
grusev: Hi Alex,
Thank you for review comments. It is very helpful to get comments at early stage so that I can address all concerns properly and make the tests cleaner, maintainable and reliable.

Having the code for setup and cleanup tightly coupled with ASV is not the way to go. Yes it is easier to do, but much far from effective and reliable. If you look back into old test you will see copy and pastes from here and there which makes each test unique in all ways. And while the tests have to be unique - the setupup does not need to be unique (or parts of it)

Setup / cleanup must be unbundeled from ASV and be able to be executed on its own.

The overabstraction I will address, indeed now when I look back I see many ways to change it make it what it should be without overcomplication, but yet some abstraction will remain, but in more cleaner way. Your comment 'preferring composition over inheritance.' is right on the spot!

And you are correct that once I go over more tests the abstraction will become better and more duplication will be removed. 

On the comment for wtiting or now own wrappers - I do see value in them in test designs later. It will improve understanding of tests that are more e2e and make things more close to actual usage fromusers. Users will always abstract the domain to better manage complexity and put it into more understandable and more maintainable for their needs form. Thus making that step I would achieve that too .... We cannot tell users - you must not make any absraction. In fact while making those abstractions they will provide their own value further up the chain - a solution on top of arctic, but abstracted in certain way. And only when you do that step - do an abstraction to achieve something you start to reveal limitations of the underlying library ... That can be bugs, which are harder to find, or might be the way underlying libworks, but it is far from common and thus you have not understood that, because we always make assumptions on what we read. And if we what we read is only API specs, without examples and notes what will not work, the assumption is that it will work. And that is exactly what we face - we have library  which posseses certain functionality it has certain bugs which we will adress but it has also limitations that we would need to document, or in other words we can give users know how to avoid pain. But in order to give that know how we must also walt their path, not ours ... Thus the overall value of those abstractions is much more than you can see at first glance. Those abstractions indeed aim to help in future productivity - through some better organization over common functionalities and prevent also errors during test developement, but also by doing them the person that does it faces his/assumptions which could be right/wrong etc

python/benchmarks/real_storage/libraries_creation.py
grusev: Will see and use

grusev: >> Why do you allow specifying the type separately to the URI? Doesn't the URI always imply the type?

You have many implementations of S3 and Amazon is one implementation. Thus perhaps it is more correct to refer to this as a storage ... as we can have amazon s3, minio s3 etc. And from performance point of view we measure against something concrete - ie an actual implementation rather than against simply S3 ... as it is an abstraction

python/arcticdb/util/utils.py
grusev: Classes provide way of grouping similar functions together. They are used to provide coherent experience for users and also increase the speed of work. Instead of having thousands mixed functions in a library we group them by function inclasses. Thus at upper level you need to remomber just several classes. Type the class name and inteli sense will provide you with all relevant options. Much easier to maintain and provide coherency and same expreience via 10-20 similar functions than to search huge module for function you need. And then how to modify it to provide coherence as similar functions are scattered all over? We even do not think for others we solve our current problem with that function and soon we have 10-20 functions doing similar things in very different way.

I would actually continue with that approach to gather things in coherent utility classes. Over time the benefits will become apparent with its usage over older way. especially for scenarios where we hve sevral steps

python/benchmarks/real_storage/libraries_creation.py
poodlewars: Have each test work in its own `prefix` defined in the Arctic URI?

python/arcticdb/util/utils.py
poodlewars: We get the same advantages my putting them in their own Python module, without the complexity of the class wrapping them (which may or may not be stateless). The alternative to static classes isn't "functions scattered all over"

python/benchmarks/real_list_operations.py
grusev: fixed

python/benchmarks/real_list_operations.py
grusev: moved to utils

python/benchmarks/real_read_write.py
grusev: naming is adressed now

.github/workflows/analysis_workflow.yml
```diff
@@ -9,6 +9,10 @@ on:
         description: Tag of the ArcticDB development image to use for benchmark and code coverage flows
         type: string
         default: latest
+      lmdb_suite_to_run:
```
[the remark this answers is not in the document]
grusev: This is something I am not sure how to fix ... Experimented with 2 approaches but I get syntax error always. And things look ok.

image

python/arcticdb/util/test.py
```diff
@@ -351,13 +351,17 @@ def populate_db(version_store):
     version_store.write("rec_norm", data={"a": np.arange(5), "b": np.arange(8), "c": None}, recursive_normalizers=True)
 
 
-def random_integers(size, dtype):
+def random_integers(size, dtype, minV: int = None, maxV: int = None):
```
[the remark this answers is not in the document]
grusev: fixed

python/arcticdb/util/utils.py
```diff
@@ -13,9 +16,13 @@
 import numpy as np
 import pandas as pd
 
-from arcticdb.util.test import create_datetime_index, get_sample_dataframe
+from arcticdb.util.test import create_datetime_index, get_sample_dataframe, random_integers
 from arcticdb.version_store.library import Library
 
+# Type definitions - supported by arctic
```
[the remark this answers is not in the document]
grusev: fixed

python/benchmarks/real_read_write_wide.py
grusev: done. Usage of logging is now made possible, but always advised to be used from method logger() of setup utility

python/benchmarks/real_storage/libraries_creation.py
poodlewars: Then we should surely use the existing fixture

python/arcticdb/util/environment_setup.py
```diff
@@ -0,0 +1,719 @@
...
+logger.addHandler(console_handler)
+
+## Amazon s3 storage bucket dedicated for ASV performance tests
+AWS_S3_DEFAULT_BUCKET = 'arcticdb-asv-real-storage'
+
+## The tests will write in 3 different prefixes
+## First - Where all persistent libraries are going to be stored
+PERSISTENT_LIBS_PREFIX = None 
+## Second - When a test requires to write something to measure performance
+MODIFYABLE_LIBS_PREFIX = 'MODIFYABLE_LIBRARIES'
```
poodlewars, line 42: `s/MODIFYABLE/MODIFIABLE`

python/arcticdb/util/environment_setup.py
```diff
@@ -0,0 +1,719 @@
...
+            cls._instance.value = "Initialized"
+
+            ## AWS S3 variable setup
+            ## Copy needed environments variables from GH Job env vars
+            ## Will only throw exceptions if real tests are executed
+            cls._aws_default_factory = real_s3_from_environment_variables(shared_path=True)
+            cls._aws_default_factory.default_prefix = None
+            cls._aws_default_factory.default_bucket = AWS_S3_DEFAULT_BUCKET
+            cls._aws_default_factory.clean_bucket_on_fixture_exit = False
+        return
```
poodlewars, line 68: no need for the `return`

python/arcticdb/util/environment_setup.py
```diff
@@ -0,0 +1,719 @@
...
+        """
+        Constructs actual prefix based on enforced via 
+        environment variable + defined in the module type prefixes + defined by tests prefixes
+        all that makes unique combination
+        """
+        result = None
+        if SetupConfig.get_aws_enforced_prefix() is not None:
+            result = SetupConfig.get_aws_enforced_prefix()
+        if type_prefix is not None:
+            if result is not None:
```
poodlewars, line 168: (nit) invert the if statement to avoid double negative (this is a pet hate of mine :) )

python/arcticdb/util/environment_setup.py
```diff
@@ -0,0 +1,719 @@
...
+                result = f"{result}_{type_prefix}"
+        if self.__storage_test_env_prefix is not None:
+                result = f"{result}_{self.__storage_test_env_prefix}"
+        return result
+
+    def set_test_mode(self):
+        """
+        Makes the setup run in test mode. Will write data in separate storage space. Instead on persistent 
+        storage space on test storage space.
+        Has effect on persistent storages
```
poodlewars, line 180: I don't follow what

```
Will write data in separate storage space. Instead on persistent 
        storage space on test storage space.
        Has effect on persistent storages
```

means

python/benchmarks/real_storage/libraries_creation.py
grusev: fixed

python/benchmarks/real_storage/libraries_creation.py
grusev: currently the new version supports that - there are 3 levels of prefixes :
  - persistent storage - non-modifyable only for read data
  - modifyable test storage: {env_var} + modifyable_prefix + test_prefix (this makes it unique for a test and for environment)
  - test storage - again common storage to be used for tests of the framework which simulates persistent (shared) storage

.github/workflows/analysis_workflow.yml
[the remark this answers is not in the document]
grusev: fixed now - here it is string and the upper workflow that cals it is choice

.github/workflows/benchmark_commits.yml
```diff
@@ -76,7 +77,13 @@ jobs:
         shell: bash -el {0}
         run: | 
           git config --global --add safe.directory .
-          python -m asv run -v --show-stderr ${{ inputs.commit }}^!
+          if [ "${{ github.event.inputs.suite_to_run }}" == "REAL" ]; then
+            SUITE='^(?!real_).*'
+          else
+            SUITE='^(real_).*'
+          fi
+          ARCTICDB_PERSISTENT_STORAGE_SHARED_PATH_PREFIX="_github_"
```
G-D-Petrov, line 85: Better to move this to the step for setting up the env variables.
```yaml
-name: Set persistent storage variables
````

**Review by G-D-Petrov — CHANGES_REQUESTED** [review]

File: python/arcticdb/util/environment_setup.py (hunk 27 above)
Comment (grusev) [reply python/arcticdb/util/environment_setup.py:181]: True!

```
"""
        Makes the setup run in test mode. Will write data instead on the default persistent storage
        in a separate test persistent storage space. This allows you to do test of production code 
        and not pollute the default persistent store with unnecessary information and also protect data
        there from deletion or corruption.
        Always use test mode for new tests!
"""
```

File: python/arcticdb/util/environment_setup.py [hunk 29]
```diff
@@ -0,0 +1,720 @@
...
+## Amazon s3 storage bucket dedicated for ASV performance tests
+AWS_S3_DEFAULT_BUCKET = 'arcticdb-asv-real-storage'
+
+## The tests will write in 3 different prefixes
+## First - Where all persistent libraries are going to be stored
+PERSISTENT_LIBS_PREFIX = None 
+## Second - When a test requires to write something to measure performance
+MODIFYABLE_LIBS_PREFIX = 'MODIFYABLE_LIBRARIES' 
+## Third - This will be used for prefix instead of persistent storage when doing framework tests
+TEST_LIBS_PREFIX = 'TESTS_LIBRARIES'
```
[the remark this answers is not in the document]
grusev: Yep. It needs careful documentation and I will take care of. It makes sense also discussing this with whole team so that we all understand both opportunities but also the limitations. And to search for additional ideas also while on this task

python/arcticdb/util/environment_setup.py
```diff
@@ -0,0 +1,720 @@
...
+            return None
+
+    @classmethod
+    def get_aws_s3_arctic_uri(cls, prefix: str = MODIFYABLE_LIBS_PREFIX, confirm_persistent_storage_need: bool = False) -> str:
+        """
+        Persistent libraries should be created in `prefix` None!
+        Temporary should be created in prexies other than None
+        """
+        assert cls._aws_default_factory, "Environment variables not initialized (ARCTICDB_REAL_S3_ACCESS_KEY,ARCTICDB_REAL_S3_SECRET_KEY)"
+        if (PERSISTENT_LIBS_PREFIX is None and prefix is None) or prefix == PERSISTENT_LIBS_PREFIX:
```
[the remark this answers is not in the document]
grusev: actually we need such thing but to assert any of prexies are not null at very begining

python/arcticdb/util/environment_setup.py
grusev: good point ... old Java habbits makes me still think that way

python/benchmarks/real_read_write.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def time_read_with_date_ranges_last20_percent_rows(self, storage_info, num_rows):
+        sym = self.setup_env.get_symbol_name(num_rows, None)
+        self.setup_env.get_library().read(symbol=sym, date_range=self.last_20).data
+
+    def peakmem_read_with_date_ranges_last20_percent_rows(self, storage_info, num_rows):
+        sym = self.setup_env.get_symbol_name(num_rows, None)
+        self.setup_env.get_library().read(symbol=sym, date_range=self.last_20).data
+
+class AWSReadWrite(LMDBReadWrite):
```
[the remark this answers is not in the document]
G-D-Petrov: As we discussed, let's leave it as is otherwise AWV won't work properly.

python/benchmarks/real_list_operations.py
```diff
@@ -0,0 +1,115 @@
...
+        .set_params([25, 50]))
+
+    '''
+    # FOR TEST PURPOSES
+    .set_mean_number_versions_per_sym(3)
+    .set_max_number_versions(5) # number versions is approx = num_syms * mean_number_versions
+    .set_with_metadata_for_each_version()
+    .set_with_snapshot_for_each_version()
+    .set_params([5, 6]))
+    '''
```
G-D-Petrov, line 85: nit: I'd add the suggested values for testing as comments next to the actual code like:
``` python
    SETUP_CLASS = (GeneralSetupSymbolsVersionsSnapshots(storage=Storage.AMAZON, prefix="LIST_VERSIONS")
        .set_mean_number_versions_per_sym(35) # 3 for testing
        .set_max_number_versions(50) # number versions is approx = num_syms * mean_number_versions, 5 for testing
        .set_with_metadata_for_each_version()
        .set_with_snapshot_for_each_version()
        .set_params([25, 50])) # [5, 6] for testing

```

G-D-Petrov (approved):

poodlewars (approved):

PR #2166, opened 2025-02-04
8066268338 Key segment pair refactor - do not move the segment out of the KeySegmentPair as it is a shared resource

poodlewars:
This allows us to re-enable some tests in `arcticdb-enterprise` https://github.com/man-group/arcticdb-enterprise/pull/new/aseaton/keysegmentpair-refactor-enterprise . 

The changes are:

- Remove the mutable `Segment& segment()` in the `KeySegmentPair` - this was often used to move out of the `Segment` owned by the shared pointer in `KeySegmentPair`
- Pass `KeySegmentPair` by lvalue rather than rvalue reference where possible. The `Task` layer still needs rvalue references as that is what Folly expects (rightly, the Folly executors need to have ownership).
- Refactor `lookup_match_in_dedup_match` to return a variant
- Unrelated tidy up to `DecodeMetadataTask`
- `clone()` segments when adding them to the in in-memory and mock storages

In particular this lets `CopyCompressedInterstore` safely copy the same `KeySegmentPair` to several targets, whereas before the `Segment` could be moved from during the copy to each target.

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -24,9 +24,13 @@ namespace arcticdb::toolbox::apy{
 
 namespace arcticdb::async {
 
-std::pair<VariantKey, std::optional<Segment>> lookup_match_in_dedup_map(
+using ExistingObject = entity::VariantKey;
+using NewObject = storage::KeySegmentPair;
+using DeDupLookupResult = std::variant<ExistingObject, NewObject>;
+
+DeDupLookupResult lookup_match_in_dedup_map(
     const std::shared_ptr<DeDupMap> &de_dup_map,
-    storage::KeySegmentPair&& key_seg);
+    storage::KeySegmentPair& key_seg);
 
 template <typename Callable>
 auto read_and_continue(const VariantKey& key, std::shared_ptr<storage::Library> library, const storage::ReadKeyOpts& opts, Callable&& c) {
@@ -161,7 +165,7 @@ folly::Future<folly::Unit> write_compressed(storage::KeySegmentPair ks) override
 }
 
 void write_compressed_sync(storage::KeySegmentPair ks) override {
-    library_->write(std::move(ks));
+    library_->write(ks);
 }
 
 folly::Future<entity::VariantKey> update(const entity::VariantKey &key,
@@ -363,7 +367,6 @@ std::vector<folly::Future<bool>> batch_key_exists(
 folly::Future<SliceAndKey> async_write(
             folly::Future<std::tuple<PartialKey, SegmentInMemory, pipelines::FrameSlice>> &&input_fut,
             const std::shared_ptr<DeDupMap> &de_dup_map) override {
-        using KeyOptSegment = std::pair<VariantKey, std::optional<Segment>>;
         return std::move(input_fut).thenValue([this] (auto&& input) {
             auto [key, seg, slice] = std::forward<decltype(input)>(input);
             auto key_seg = EncodeAtomTask{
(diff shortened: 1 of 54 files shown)


cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -225,7 +225,7 @@ void do_read_impl(
     KeyDecoder&& key_decoder,
     ReadKeyOpts opts) {
     auto key_seg = do_read_impl(std::move(variant_key), root_folder, bucket_name, s3_client, std::forward<KeyBucketizer>(bucketizer), std::forward<KeyDecoder>(key_decoder), opts);
-    visitor(key_seg.variant_key(), std::move(key_seg.segment()));
+    visitor(key_seg.variant_key(), *key_seg.segment_ptr());
```
IvoDD, line 228: IMO perhaps when reading with a visitor it does make sense to move.

I.e. makes sense to move the ownership to the visitor instead of giving it a reference to something which will soon get destructed.

cpp/arcticdb/storage/mock/azure_mock_client.cpp
```diff
@@ -54,7 +54,7 @@ std::optional<Azure::Core::RequestFailedException> has_failure_trigger(const std
 
 void MockAzureClient::write_blob(
         const std::string& blob_name,
-        arcticdb::Segment&& segment,
+        arcticdb::Segment& segment,
```
IvoDD, line 57: We should clone for all mock storages. Looks like we've forgotten it for azure.

cpp/arcticdb/storage/s3/detail-inl.hpp
IvoDD: If we decide to undo this we should do it for all storages.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -161,7 +165,7 @@ folly::Future<folly::Unit> write_compressed(storage::KeySegmentPair ks) override
 }
 
 void write_compressed_sync(storage::KeySegmentPair ks) override {
```
IvoDD, line 167: Should we change this to take a `KeySegmentPair&` to be consistent with the others and to make clear that it does not copy the underlying segment?

IvoDD: Same goes for non-sync variant

cpp/arcticdb/async/tasks.hpp
```diff
@@ -532,16 +525,15 @@ struct DecodeTimeseriesDescriptorTask : BaseTask {
...
     DecodeTimeseriesDescriptorTask() = default;
 
-    std::pair<VariantKey, TimeseriesDescriptor> operator()(storage::KeySegmentPair &&ks) const {
+    std::pair<VariantKey, TimeseriesDescriptor> operator()(storage::KeySegmentPair &&key_seg) const {
         ARCTICDB_SAMPLE(DecodeTimeseriesDescriptorTask, 0)
-        auto key_seg = std::move(ks);
         ARCTICDB_DEBUG(log::storage(), "DecodeTimeseriesDescriptorTask decoding segment with key {}", variant_key_view(key_seg.variant_key()));
 
-        auto maybe_desc = decode_timeseries_descriptor(key_seg.segment());
+        auto maybe_desc = decode_timeseries_descriptor(*key_seg.segment_ptr());
```
IvoDD, line 532: Maybe again a bit out of scope but I think `decode_timeseries_descriptor` `decode_metadata_and_descriptor_fields` etc. can be changed to just take a `const Segment&`?

poodlewars: They can't be, they modify the segment in place

cpp/arcticdb/storage/s3/s3_client_interface.hpp
```diff
@@ -85,7 +85,7 @@ class S3ClientInterface {
 
     virtual S3Result<std::monostate> put_object(
         const std::string& s3_object_name,
-        Segment&& segment,
+        Segment& segment,
```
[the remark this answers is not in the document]
poodlewars: Yeah this is where we chose to draw the line and stop refactoring, making all the consumers of the `Segment` const will be an even bigger change. But isn't it implied that this shouldn't move data out of the segment, given the segment is passed by lvalue reference?

cpp/arcticdb/async/async_store.hpp
poodlewars: I think we have to keep it like this, this is what this bit of the PR description is referring to,

```
Pass KeySegmentPair by lvalue rather than rvalue reference where possible. The Task layer still needs rvalue references as that is what Folly expects (rightly, the Folly executors need to have ownership).
```

cpp/arcticdb/storage/s3/detail-inl.hpp
poodlewars: Great point thank you, I'll try this

IvoDD (approved):

PR #2167, opened 2025-02-05
8234437666 Bound the number of data segments kept live in memory during compaction and sort_merge

poodlewars:
Monday: 8234437666

This is a simpler alternative to PR #2136 . The idea is to restrict the number of `write_futures` that we allow to accumulate during `compact_incomplete` or `sort_and_finalize_staged_data` since each write task (writing a table data key) holds its segment live in memory, so if a queue of work builds up we can use an unbounded amount of memory.

This behaviour in memory use is only observable with relatively high latency storages and large amounts of work. I've simulated it here with only 1000 incomplete segments, using a faked 3000ms write latency.

We often use `folly::window` to achieve this kind of thing, but that doesn't fit in well with the `SegmentAggregator` (because the number of futures we'll end up with is not known ahead of time).

Profiling scripts are available here https://github.com/man-group/ArcticDB/commit/5e6a3e54fce49b6a9ff01db31b6c20d4d8cd870d

The conclusions from the profiling are:

```
All on my 20 core laptop (so IO pool size 30). All with 100k rows per incomplete. All with 1000 incompletes.

## Memory analysis: With 10sec sleep on writes

Using debug build.

Without fix 3.7GiB at peak
10 concurrent workers 125 MiB at peak
100 concurrent workers 755 MiB at peak
1000 concurrent workers 3.7GiB at peak

## Performance Analyis with 3sec sleep on writes

Using release build.

Without fix 120s
10 concurrent workers 325s
60 concurrent workers (proposed default for my machine) 120s
100 concurrent workers 119s
1000 concurrent workers 121s
```

Also tested with a single IO and CPU thread and the semaphore only having size 1 (to check we don't deadlock).

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -1674,6 +1674,7 @@ VersionedItem sort_merge_impl(
 
     std::vector<FrameSlice> slices;
     std::vector<folly::Future<VariantKey>> fut_vec;
+    auto semaphore = std::make_shared<folly::NativeSemaphore>(n_segments_live_during_compaction());
     auto index = stream::index_type_from_descriptor(pipeline_context->descriptor());
     util::variant_match(index,
         [&](const stream::TimeseriesIndex &timeseries_index) {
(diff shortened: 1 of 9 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -2082,4 +2082,14 @@ CheckOutcome check_schema_matches_incomplete(const StreamDescriptor& stream_desc
     return std::monostate{};
 }
 
+size_t n_segments_live_during_compaction() {
+    int64_t default_count = 2 * async::TaskScheduler::instance()->io_thread_count();
+    int64_t res = ConfigsMap::instance()->get_int("VersionStore.NumSegmentsLiveDuringCompaction", default_count);
```
poodlewars, line 2087: Not documented yes as I think the default will probably always be fine.

IvoDD (commented):
LGTM. What do you think about adding some correctness python tests like the ones you performed in the benchmarking commit?

E.g. the `NumSegmentsLiveDuringCompaction=1` can live in our automated tests (probably with smaller segments and a smaller `RowsPerSegment`?)

It would also be nice to somehow have an automated test which triggers a write failure via the mock storage simulator, however I can't think of a good way to trigger the failure on just a few of the data keys... so this is probably for a future PR

IvoDD (approved):

PR #2173, opened 2025-02-10
8371297134 Type promotion between integral and floating types

poodlewars:
To fix the bug reported in 8371297134

- Write an int64 column
- Append a float32 to it
- Read it -> We blow up at read time

I've added logic so that when we merge descriptors we combine:

```
any integer + float 64 -> float 64
integer up to 16 bits + float 32 -> float 32
integer above 16 bits + float 32 -> float 64
```

This is because a 16 bit integer can safely fit in a 32 bit float without loss of precision, whereas a 32 bit integer cannot. We could instead always promote up to float 64 which would be more wasteful but simpler.

Separately, for query builder pipelines I added changes so that when combining a float and an integer we always promote up to float64. This misses the cases where it is actually safe to promote to float32, but is simpler and matches Pandas. This is the cause of the test change in `test_sort_merge.py`.

I reworked `has_valid_type_promotion`, introducing a new function `is_valid_type_promotion_to_target` instead. This is because the `has_valid_type_promotion` signature was dangerous - it returned an optional type that was often interpreted only as a bool (the actual type inside it was ignored). I've replace the call sites that only tested the bool with calls to the new `is_valid_type_promotion_to_target` that only returns a bool.

### DIFF
--- a/cpp/arcticdb/entity/type_utils.cpp
+++ b/cpp/arcticdb/entity/type_utils.cpp
@@ -66,32 +66,29 @@ namespace arcticdb {
         return std::make_optional<entity::TypeDescriptor>(combine_data_type(entity::ValueType::FLOAT, target_size), dimension);
     }
 
-    std::optional<entity::TypeDescriptor> has_valid_type_promotion(
+    bool is_valid_type_promotion_to_target(
         const entity::TypeDescriptor& source,
         const entity::TypeDescriptor& target
     ) {
-
         if (source.dimension() != target.dimension()) {
             // Empty of dimension 0 means lack of any given type and can be promoted to anything (even if the dimensions
             // don't match), e.g. empty type can become int or array of ints. Empty type of higher dimension is used to
             // specify an empty array or an empty matrix, thus it cannot become any other type unless the dimensionality
             // matches
-            if (is_empty_type(source.data_type()) && source.dimension() == entity::Dimension::Dim0)
-                return target;
-            return std::nullopt;
+            return is_empty_type(source.data_type()) && source.dimension() == entity::Dimension::Dim0;
         }
 
         if (source == target)
-            return target;
+            return true;
 
         // Empty type is coercible to any type
         if (is_empty_type(source.data_type())) {
-            return target;
+            return true;
         }
 
         // Nothing is coercible to the empty type.
         if (is_empty_type(target.data_type())) {
-            return std::nullopt;
+            return false;
         }
 
         auto source_type = source.data_type();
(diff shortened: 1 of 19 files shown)


cpp/arcticdb/entity/type_utils.cpp
```diff
@@ -39,32 +39,56 @@ namespace arcticdb {
         return false;
     }
 
-    std::optional<entity::TypeDescriptor> has_valid_type_promotion(
+    constexpr bool is_mixed_float_and_integer(entity::DataType left, entity::DataType right) {
+        return (is_integer_type(left) && is_floating_point_type(right)) || (is_floating_point_type(left) && is_integer_type(right));
+    }
+
+    std::optional<entity::TypeDescriptor> common_type_float_integer(const entity::TypeDescriptor& left, const entity::TypeDescriptor& right) {
```
vasil-pashov, line 46: nit: If used only in this file mark it as static
```suggestion
    static std::optional<entity::TypeDescriptor> common_type_float_integer(const entity::TypeDescriptor& left, const entity::TypeDescriptor& right) {
```

cpp/arcticdb/entity/type_utils.cpp
```diff
@@ -73,101 +97,101 @@ namespace arcticdb {
...
-            return std::nullopt;
+            return false;
         } else {
             // Non-numeric source type
-            return std::nullopt;
+            return false;
+        }
+    }
+
+    std::optional<entity::TypeDescriptor> common_type_mixed_sign_ints(const entity::TypeDescriptor& left, const entity::TypeDescriptor& right) {
```
vasil-pashov, line 151:
```suggestion
    static std::optional<entity::TypeDescriptor> common_type_mixed_sign_ints(const entity::TypeDescriptor& left, const entity::TypeDescriptor& right) {
```

python/tests/unit/arcticdb/version_store/test_column_type_changes.py
```diff
@@ -122,4 +125,270 @@ def get_type_of_column():
...
+    lib = lmdb_version_store_dynamic_schema
+
+    if int_first:
+        original_data = pd.DataFrame({"a": np.array([1, 2, 3], int_type)}, index=[0, 1, 2])
+        first_append = pd.DataFrame({"a": np.array([4.0, 5.0, 6.0], float_type)}, index=[3, 4, 5])
+    else:
+        original_data = pd.DataFrame({"a": np.array([1, 2, 3], float_type)}, index=[0, 1, 2])
+        first_append = pd.DataFrame({"a": np.array([4.0, 5.0, 6.0], int_type)}, index=[3, 4, 5])
+
+    second_append = pd.DataFrame({"a": np.array([7, 8, 9], second_append_type)}, index=[5, 6, 7])
```
vasil-pashov, line 146: What does the second_append achieve?

cpp/arcticdb/entity/type_utils.cpp
```diff
@@ -73,101 +97,101 @@ namespace arcticdb {
...
+            internal::check<ErrorCode::E_ASSERTION_FAILURE>(left_size <= right_size,
+                                                            "Expected left_size <= right_size in has_valid_common_type");
+        } else {
+            // is_signed_type(right_type)
+            internal::check<ErrorCode::E_ASSERTION_FAILURE>(right_size <= left_size,
+                                                            "Expected right_size <= left_size in has_valid_common_type");
         }
 
-        return target;
+        auto target_size = entity::SizeBits(uint8_t(std::max(left_size, right_size)) + 1);
```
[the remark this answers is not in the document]
poodlewars: This is just the old logic moved around, so I'd rather not change it

python/tests/unit/arcticdb/version_store/test_column_type_changes.py
poodlewars: I wanted to have some kind of check that we haven't messed up the type descriptors after the first append and the promotion, so subsequent promotions behave the same as the first promotion.

vasil-pashov (approved):

willdealtry (approved):

cpp/arcticdb/pipeline/execution.hpp
```diff
@@ -105,7 +105,7 @@ struct ColumnExpression {
     ColumnExpression(const IterableContainer& left, const IterableContainer& right) :
         left_(left),
         right_(right) {
-        auto promoted_type = has_valid_type_promotion(left_->type(), right_->type());
+        auto promoted_type = has_valid_common_type(left_->type(), right_->type());
         util::check(static_cast<bool>(promoted_type), "Cannot promote from type {} and type {} in column expression", left_->type(), right_->type());
```
alexowens90, line 109: nit: `promoted_type` is now a `bool` so can get rid of the static cast

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -139,11 +139,10 @@ struct type_arithmetic_promoted_type {
...
                 >,
-                // Otherwise, if only one type is floating point, promote to this type
-                std::conditional_t<std::is_floating_point_v<LHS>,
-                    LHS,
-                    RHS
-                >
+                // Otherwise, if only one type is floating point, always promote to double
+                // For example when combining int32 and float32 the result can only fit in float64 without loss of precision
+                // Special cases like int16 and float32 can fit in float32, but we always promote up to float64 (as does Pandas)
+                double
```
alexowens90, line 145: This is technically a semver change of behaviour

python/tests/hypothesis/arcticdb/test_resample.py
```diff
@@ -107,4 +106,28 @@ def test_resample(lmdb_version_store_v1, df, rule, origin, offset):
                 if str(pandas_error) != "Values falls before first bin":
                     raise pandas_error
                 else:
-                    return
\ No newline at end of file
+                    return
+
+
+def test_resample_mean_large_arithmetic_error_repro(lmdb_version_store_v1):
```
alexowens90, line 112: Could move to nonreg tests

python/tests/unit/arcticdb/version_store/test_column_type_changes.py
```diff
@@ -122,4 +125,270 @@ def get_type_of_column():
...
+    lib.append("test", second_append)
+
+    # When
+    data = lib.read("test").data
+
+    # Then
+    expected_result = pd.concat([original_data, first_append, second_append])
+    assert_frame_equal(data, expected_result)
+
+    adb_lib = Library("desc", lib)
```
alexowens90, line 158: Can use `get_info` on `NativeVersionStore` to avoid this

poodlewars: I think I prefer using the v2 API for this so we can check the nicely mapped `DataType` objects rather than the `TypeDescriptor` objects the v1 API gives you (that probably shouldn't be on a public API at all).

python/arcticdb/util/test.py
```diff
@@ -860,7 +860,7 @@ def generic_resample_test(
     but it cannot take parameters such as origin and offset.
     """
     # Pandas doesn't have a good date_range equivalent in resample, so just use read for that
-    expected = lib.read(sym, date_range=date_range).data
+    original_data = lib.read(sym, date_range=date_range).data
```
poodlewars, line 863: This change is just to help debugging tests

alexowens90 (approved):

PR #2174, opened 2025-02-11
Automated Release scripts

ognyanstoimenov:
Currently versions are released by manually running the tag.yml workflow, which releases selected branch with a specified version.
This extends that with scripts that call the tag.yml with version to release and branch to release it from.
### Flow: 
* **Main entry point** -  `automated_release.full_version_cron` - on a schedule (or manual trigger) releases the latest release branch. Release branches follow the naming convention `X.Y.Z` so the release produced will be `vX.Y.Z`
* **When a release is created** - `automated_release.create_release_branch` is triggered which calculates what the next release version should be based on commits in `master` and create that release branch from which would be the base of the next release. Immediately releases it as `rc0` (calls `tag.yml` to release `<new_branch_name>rc0`)
* **On every commit in release branch** - `automated_release.rc_version` releases next rc version of that branch (calls `tag.yml` to release `<release_branch_name>rc1` or rc2, etc)
 
### Marking PR's as `patch`, `minor`, `major`
* **New PR template**:
```
## Change Type (Required)
- [ ] **Patch** (Bug fix or non-breaking improvement)
- [ ] **Cherry pick**
```
* **On every PR opened/edited to master** `automated_release.parse_pr_template` - parses new template and adds apropriate label which in future gets parsed by `automated_release.create_release_branch` specified above

## Proof of working/demo (~7min, watch in 1.5x speed)
https://github.com/user-attachments/assets/94705eb1-dd76-4397-8e2e-44828fc01ceb

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,6 +3,12 @@
 
 #### What does this implement or fix?
 
+## Change Type (Required)
+- [ ] **Patch** (Bug fix or non-breaking improvement)
+- [ ] **Minor** (New feature, but backward compatible)
+- [ ] **Major** (Breaking changes)
+- [ ] **Cherry pick**
+
 #### Any other comments?
 
 #### Checklist
--- a/.github/workflows/automated_release.create_release_branch.yml
+++ b/.github/workflows/automated_release.create_release_branch.yml
@@ -1,0 +1,92 @@
+name: "Automated Release: Create next release branch"
+
+on:
+  release:
+    types: [released]
+env:
+  VERSION_REGEX: ^v\d+\.\d+\.\d+$
+run-name: "Automated release: ${{ github.event.release.tag_name }} released. Cutting the branch for the next version"
+jobs:
+  create-next-release-branch:
+    name: Parse PRs labels and create next release branch
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      repository-projects: write
+    outputs:
+      new_branch: ${{ steps.get-next-version.outputs[steps.get_increment_type.outputs.increment_type] }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: master
+      
+      - name: Get default branch sha
+        id: get_sha
+        run: |
+          BASE_SHA=$(git rev-parse master)
+          echo "BASE_SHA=$BASE_SHA" >> $GITHUB_ENV
+          echo "Release branch will be cut from master at $BASE_SHA"
+      
+      - name: Get latest released tag
+        id: latest_tag
+        uses: oprypin/find-latest-tag@v1
+        with:
+          repository: ${{ github.repository }}
+          regex: ${{ env.VERSION_REGEX }}
+          sort-tags: true
+          releases-only: true
+
+      - name: Get PRs between master (${{ env.BASE_SHA }}) and ${{ steps.latest_tag.outputs.tag }}
+        uses: VanOns/get-merged-pull-requests-action@master
+        id: pull_requests
+        with:
+          github_token: ${{ github.token }}
+          current_tag: ${{ env.BASE_SHA }}
+          previous_tag: ${{ steps.latest_tag.outputs.tag }}
+          return_type: "all"
+
+      - name: Get labels of PRs and determine increment type
+        id: get_increment_type
+        env:
+          PULL_REQUESTS: ${{ steps.pull_requests.outputs.pull_requests }}
+          MAJOR_LABEL: major
+          MINOR_LABEL: minor
+          PATCH_LABEL: patch
+        run: |
+          if [ -z "$PULL_REQUESTS" ]; then
+            echo "Error: No PRs found between branches" && exit 1
+          fi
+          labels=$(echo "$PULL_REQUESTS" | jq -r '.[].labels[].name')
+          echo Labels: $labels
+          increment_type="patch"
+          case "$labels" in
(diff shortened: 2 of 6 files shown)


.github/workflows/automated_release.create_release_branch.yml
```diff
@@ -0,0 +1,92 @@
...
+  VERSION_REGEX: ^v\d+\.\d+\.\d+$
+run-name: "Automated release: ${{ github.event.release.tag_name }} released. Cutting the branch for the next version"
+jobs:
+  create-next-release-branch:
+    name: Parse PRs labels and create next release branch
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      repository-projects: write
```
G-D-Petrov, line 16: Do we really need all of these permissions?

.github/workflows/automated_release.full_version_cron.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      next_version: ${{ steps.calculate-next-version.outputs.next_version }}
+    steps:
+      - name: Get latest RC tag
+        id: latest_rc_tag
+        uses: oprypin/find-latest-tag@v1
+        with:
+          repository: ${{ github.repository }}
+          regex: ^v\d+\.\d+\.\d+rc\d+$
```
G-D-Petrov, line 22: This is called with different params than the ones in `create-next-release-branch`.
Maybe equalize them

G-D-Petrov (commented):
Looks pretty good, I just have a couple of questions

.github/workflows/automated_release.create_release_branch.yml
ognyanstoimenov: Thought they are needed for the downstream jobs, but just tested and seems like they are not necessary, will remove

.github/workflows/automated_release.full_version_cron.yml
ognyanstoimenov: They are slightly different in meaning. This one searches for the latest rc tag, and later strips 'rc'. while the other one searches for the latest full release and later increments the version

G-D-Petrov (approved):

.github/workflows/automated_release.create_release_branch.yml
```diff
@@ -0,0 +1,88 @@
...
+          PATCH_LABEL: patch
+        run: |
+          if [ -z "$PULL_REQUESTS" ]; then
+            echo "Error: No PRs found between branches" && exit 1
+          fi
+          labels=$(echo "$PULL_REQUESTS" | jq -r '.[].labels[].name')
+          echo Labels: $labels
+          increment_type="patch"
+          case "$labels" in
+            *"${{ env.MAJOR_LABEL }}"*) increment_type="major"; echo "Major changes found! Should increase major version." ;;
```
maxim-morozov, line 61: How this would work when we have mix of major, minor and patch between releases? Say something like when we cut the release candidate branch, master had only minor changes. But then during the week, master had major version changes. I wonder if we need to think about it?

.github/workflows/automated_release.full_version_cron.yml
```diff
@@ -0,0 +1,38 @@
+name: "Automated Release: Release next full version from release branch"
+
+on:
+  # schedule:
```
maxim-morozov, line 4: should we enable this? Or you want to test it manually first?

ognyanstoimenov: Let's test manually first and if all is ok, I will enable it

maxim-morozov (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,6 +3,12 @@
 
 #### What does this implement or fix?
 
+## Change Type (Required)
```
poodlewars, line 6: Can we just enforce these with labels on the PR rather than these checkboxes?

.github/workflows/automated_release.rc_version.yml
```diff
@@ -0,0 +1,44 @@
+name: "Automated Release: Release next RC from release branch"
+
+on:
+  push:
+    branches:
+      - '[0-9]+.[0-9]+.[0-9]+'
```
poodlewars, line 6: I think we need a stricter naming convention for these branches - otherwise someone will create a branch 65.11.22 and it will get released

.github/PULL_REQUEST_TEMPLATE.md
ognyanstoimenov: It is possible yes, maybe even clearer. I will enforce it so we can't merge without the labels. Also I think we can use the same set of labels for releases and release notes, so that we don't have e.g. both `minor` and `feature` but just the one that would be used for both release versions and release notes

.github/workflows/automated_release.full_version_cron.yml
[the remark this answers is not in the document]
ognyanstoimenov: I don't understand what you mean. Why would we patch an older version? If we have v5.5.5rc3 or rc4 etc, when it is time we release v5.5.5 and that's it for that version. If we find something is wrong after the full release and we want to patch it would be in the next patch version 5.5.6

.github/workflows/automated_release.rc_version.yml
ognyanstoimenov: Yes, I agree. I thought we could use `x.y.z-<branch cut sha>` as naming convention but it looked ugly. So blocking such creations should be fine. I will try to configure that as well as some merging restrictions in release branch

PR #2176, opened 2025-02-12
GCP Support over the S3 XML API

poodlewars:
This adds support for GCP over the S3-compatible XML API.

GCP support did not work out of the box because Google does not support any of the S3 "batch" APIs. We used batch delete for all our deletion operations, which broke. Instead I use the unary delete operation here, using the AWS SDK's async API to reduce the performance impact of sending single object deletes.

Users connect to GCP like:

```
ac = Arctic("gcpxml://storage.googleapis.com:aseaton_bucket?access=...&secret=...")
ac = Arctic("gcpxmls://storage.googleapis.com:aseaton_bucket?access=...&secret=...")
# Kept aws_auth naming as this will use ~/.aws/credentials
ac = Arctic("gcpxml://storage.googleapis.com:aseaton_bucket?aws_auth=true&aws_profile=blah")
ac = Arctic("gcpxml://storage.googleapis.com:aseaton_bucket?aws_auth=true&prefix=my_prefix")
```

We have a URL of `gcpxml` so that if we do a full GCP implementation later, users can get it with `gcp://`. I've added a GCP storage proto object so that users on the `gcpxml` API will later be able to move to a `gcp://` API without updating any library configs. The proto is tiny, just storing the library's prefix, which is the only information we need to serialize.

This is implemented as a small sublass of our existing S3 implementation, with its deletion functionality overridden. The Python library adapter creates the `NativeVersionStore` with a `GCPXMLSettings` in memory object (in the `NativeVariantStorage`) which makes us create a `GCPXMLStorage` C++ adapter rather than the normal `S3Storage`.

The interesting part of this PR is the threading model in `do_remove_no_batching_impl`. This is called from an IO executor as part of the `RemoveTask` so it is not safe (due to the risk of deadlocks) to submit any work to the IO executor from it. Instead:
- The storage client `s3_client_impl.cpp` submits delete operations with the S3 API. The S3 SDK has its own event loop to handle these. When they complete, they resolve a `folly::Promise`, and the client returns a `folly::Future` out of that promise.
- Our storage adapter (`detail-inl.hpp`) collects these futures, on an inline executor. This is fine because there is no work for this executor to do, other than waiting to be notified that promises have been resolved by the AWS SDK.

For testing, I created a `moto` based test fixture that returns errors when batch delete operations are invoked. Also tested manually against a real GCP backend.

There's some duplication between the `gcpxml` and the `s3` Python adapters, but since hopefully the `gcpxml` adapter will be temporary until a full GCP implementation, I didn't see the value in refactoring it away.
 
Monday: 8450684276

### DIFF
--- a/cpp/arcticdb/storage/common.hpp
+++ b/cpp/arcticdb/storage/common.hpp
@@ -75,14 +75,26 @@ inline std::vector<char> stream_to_vector(std::iostream& src) {
 
 class NativeVariantStorage {
 public:
-    using VariantStorageConfig = std::variant<std::monostate, s3::S3Settings>;
+    using VariantStorageConfig = std::variant<std::monostate, s3::S3Settings, s3::GCPXMLSettings>;
     explicit NativeVariantStorage(VariantStorageConfig config = std::monostate()) : config_(std::move(config)) {};
     const VariantStorageConfig& variant() const {
         return config_;
     }
+
     void update(const s3::S3Settings& config) {
         config_ = config;
     }
+
+    std::string to_string() {
+        return util::variant_match(config_, [](std::monostate) {
+            return "empty";
+        }, [](s3::S3Settings) {
+            return "s3";
+        }, [](s3::GCPXMLSettings) {
+            return "gcpxml";
+        });
+    }
+
 private:
     VariantStorageConfig config_;
 };
--- a/cpp/arcticdb/storage/config_cache.hpp
+++ b/cpp/arcticdb/storage/config_cache.hpp
@@ -89,6 +89,12 @@ class ConfigCache {
                     storage_conf.config().UnpackTo(&s3_storage);
                     storages.emplace_back(create_storage(path, mode, s3::S3Settings(settings).update(s3_storage)));
                 },
+                [&storage_conf, &storages, &path, mode] (const s3::GCPXMLSettings& settings) {
+                    util::check(storage_conf.config().Is<arcticdb::proto::gcp_storage::Config>(), "Only support GCP native settings");
+                    arcticdb::proto::gcp_storage::Config gcp_storage;
+                    storage_conf.config().UnpackTo(&gcp_storage);
+                    storages.emplace_back(create_storage(path, mode, s3::GCPXMLSettings(settings).update(gcp_storage)));
+                },
                 [&storage_conf, &storages, &path, mode](const auto &) {
                     storages.emplace_back(create_storage(path, mode, storage_conf));
                 }
--- a/cpp/arcticdb/storage/python_bindings.cpp
+++ b/cpp/arcticdb/storage/python_bindings.cpp
@@ -35,6 +35,77 @@ std::shared_ptr<LibraryIndex> create_library_index(const std::string &environmen
     return std::make_shared<LibraryIndex>(EnvironmentName{environment_name}, mem_resolver);
 }
 
+enum class S3SettingsPickleOrder : uint32_t {
+    TYPE = 0,
+    AWS_AUTH = 1,
+    AWS_PROFILE = 2,
+    USE_INTERNAL_CLIENT_WRAPPER_FOR_TESTING = 3
+};
+
+enum class GCPXMLSettingsPickleOrder : uint32_t {
+    TYPE = 0,
+    AWS_AUTH = 1,
+    CA_CERT_PATH = 2,
+    CA_CERT_DIR = 3,
+    SSL = 4,
+    HTTPS = 5,
+    PREFIX = 6,
+    ENDPOINT = 7,
+    SECRET = 8,
+    ACCESS = 9,
+    BUCKET = 10,
+};
+
+s3::GCPXMLSettings gcp_settings(const py::tuple& t) {
+    util::check(t.size() == 11, "Invalid GCPXMLSettings pickle objects, expected 11 attributes but was {}", t.size());
+    return s3::GCPXMLSettings{
+        t[static_cast<uint32_t>(GCPXMLSettingsPickleOrder::AWS_AUTH)].cast<s3::AWSAuthMethod>(),
+        t[static_cast<uint32_t>(GCPXMLSettingsPickleOrder::CA_CERT_PATH)].cast<std::string>(),
+        t[static_cast<uint32_t>(GCPXMLSettingsPickleOrder::CA_CERT_DIR)].cast<std::string>(),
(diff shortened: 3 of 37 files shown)


python/tests/conftest.py
```diff
@@ -368,9 +384,11 @@ def mem_storage() -> Generator[InMemoryStorageFixture, None, None]:
 @pytest.fixture(
     scope="function",
     params=[
```
poodlewars, line 386: Do not merge the commented out fixtures

cpp/arcticdb/storage/common.hpp
```diff
@@ -75,14 +75,26 @@ inline std::vector<char> stream_to_vector(std::iostream& src) {
...
     explicit NativeVariantStorage(VariantStorageConfig config = std::monostate()) : config_(std::move(config)) {};
     const VariantStorageConfig& variant() const {
         return config_;
     }
+
     void update(const s3::S3Settings& config) {
         config_ = config;
     }
+
+    std::string to_string() {
```
poodlewars, line 88: This is just slightly helpful for debugging in the Python layer.

G-D-Petrov (approved):
LGTM I just have a couple of comments that don't need to be addressed in this PR:
1. As far as I can see, it is still possible to connect to GCP via the `S3Storage` directly, rather than `GCPXMLStorage`, this would be a user error ofc but an annoying one. I think that we should either:
- have some logic that forbids this 
- add a fall back in the `S3Storage` when the batch delete fails
2. We should add some tests with a real GCP, not just a simulator

cpp/arcticdb/storage/config_cache.hpp
```diff
@@ -89,6 +89,12 @@ class ConfigCache {
                     storage_conf.config().UnpackTo(&s3_storage);
                     storages.emplace_back(create_storage(path, mode, s3::S3Settings(settings).update(s3_storage)));
                 },
+                [&storage_conf, &storages, &path, mode] (const s3::GCPXMLSettings& settings) {
+                    util::check(storage_conf.config().Is<arcticdb::proto::gcp_storage::Config>(), "Only support S3 native settings");
```
phoebusm, line 93: "S3" -> "GCP"

python/tests/integration/storage_fixtures/test_s3.py
```diff
@@ -27,3 +29,44 @@ def test_rate_limit(s3_storage_factory: MotoS3StorageFixtureFactory):  # Don't n
 
     # Then working again
     requests.head(s3.endpoint, verify=s3.client_cert_file).raise_for_status()
+
+
+def test_gcp_no_batch_delete(gcp_storage_factory: MotoGcpS3StorageFixtureFactory):
+    # Given
+    s3 = gcp_storage_factory
+    bucket = s3.create_fixture()
```
phoebusm, line 37: "s3" to "gcp" maybe to lessen the confusion?

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -130,6 +130,30 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
...
         });
 
+    py::class_<s3::GCPXMLSettings>(storage, "GCPXMLSettings")
+        .def(py::init<>())
+        .def(py::pickle(
+            [](const s3::GCPXMLSettings &settings) {
+                return py::make_tuple(settings.aws_auth());
+            },
+            [](py::tuple t) {
+                util::check(t.size() == 1, "Invalid GCPXMLSettings pickle objects");
```
phoebusm, line 140: If I have read the code correctly, gcpxml settings will be stored in this native class, instead of in protobuf, like we use to do.
If so, I think the rest of settings in the class are needed to be pickled as well.
Nice to have it test on `test_fork.py`

cpp/proto/arcticc/pb2/gcp_storage.proto
```diff
@@ -0,0 +1,14 @@
...
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+*/
+syntax = "proto3";
+
+package arcticc.pb2.gcp_storage_pb2;
+
+message Config {
+    string prefix = 1;
```
phoebusm, line 13: I wonder why prefix is the special kid, gets to stay in the proto, instead of living in native config like its sibilings 🤔

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -490,6 +505,97 @@ def mock_s3_with_error_simulation():
...
+            response_body = (
+                b'<?xml version="1.0" encoding="UTF-8"?>'
+                b'<Error>'
+                    b'<Code>NotImplemented</Code>'
+                    b'<Message>A header or query you provided requested a function that is not implemented.</Message>'
+                    b'<Details>POST ?delete is not implemented for objects.</Details>'
+                b'</Error>'
+            )
+            start_response(
+                "503 Slow Down", [("Content-Type", "text/xml"), ("Content-Length", str(len(response_body)))]
```
phoebusm: I think the http code is not `503 Slow Down` for the response of not-implemented batch delete right?

cpp/proto/arcticc/pb2/gcp_storage.proto
poodlewars: Because the path of a library is determined when the library is created, adding on the creation timestamp

python/arcticdb/storage_fixtures/s3.py
poodlewars: Yeah, fixed in the Windows fixes commit

poodlewars:
> LGTM I just have a couple of comments that don't need to be addressed in this PR:
> 
>     1. As far as I can see, it is still possible to connect to GCP via the `S3Storage` directly, rather than `GCPXMLStorage`, this would be a user error ofc but an annoying one. I think that we should either:
> 
> 
>     * have some logic that forbids this
> 
>     * add a fall back in the `S3Storage` when the batch delete fails
> 
> 
>     2. We should add some tests with a real GCP, not just a simulator

Thanks!

Agreed on 2, we'll do that as a follow up. I have done a lot of manual testing with real GCP though.

On 1, to start with this will be a bit of a "hidden" feature and we can talk people through how to use it. I'd rather not implement any fancy "is it GCP" detection or fallbacks until we know that people get bitten by this.

phoebusm (changes requested):
LGTM apart from the pickle handling
Nice async handling

phoebusm (approved):

PR #2183, opened 2025-02-17
Improve segment serialization for v1 segments

G-D-Petrov:
#### Reference Issues/PRs
Ref Monday Ticket: 7852509418

This PR is fixing a segment fault that occurs in the `CopyCompressedInterStoreTask` when used in enterprise.
The problem is a bit hard to reproduce as it involves:
1. Having some version V being written with an older version of ArcticDB (e.g. 4.4.5) to some primary
2. Trying to replicate this version V to some secondary with a version of enterprise that is using a newer version of ArcticDB (e.g. 5.2.3) and that version should have a changed structure of the Segments (e.g. what happened [here](https://github.com/man-group/ArcticDB/compare/e3b5f537bae293ac11f73d84ac861ab5ce01e718...757b4bd29e3a3de87a2265e2c021b9fb717854bc#diff-a17e4a0ad7760267af8c570795b362ec900136e414782a4b883485d36d9a8f84R167))
3. Because the newer version is trying to write more data than allocated, we are getting a seg fault

The bug is that `CopyCompressedInterStoreTask` is trying to copy the Segment but the information about its size is stale because it was saved/reserialized based on the information from the older version.

The fix involves:
- Add null pointer checks in `decode_header_and_fields` and `serialize_v1_header_to_buffer`
  - the check in `decode_header_and_fields` is not needed for the fix but it is a good change to prevent silly seg faults
- Simplify `calculate_size()` by removing unnecessary conditional
- Enhance buffer allocation and copying logic in `serialize_v1_header_to_buffer`
- Add explicit size and pointer validation checks to prevent potential memory issues

Technically, only [this change](https://github.com/man-group/ArcticDB/pull/2183/files#diff-d35d87cb4a49c502a6519b27dc94590231057f633894d26c6c34f506fee1980fR154) is needed for the fix of this bug.
But I think that the change in the [segment.cpp](https://github.com/man-group/ArcticDB/pull/2183/files#diff-21f07ddd0b2b8ab0b82fd8392af726ab2a0b5ccafedab67d86497177d4f381a2R283) make the code there easier to read/reason about + I've added some more checks to make it more robust.

**Because the change is quite hard to test here, I've added the relevant test for it in the enterprise repo with [this PR](https://github.com/man-group/ArcticDB/pull/2183).**

## Change Type (Required)
- [ ] **Cherry pick**

#### Any other comments?

The fix is resolving seg faults like:
``` bash
==40129== Thread 123 IOPool0:
==40129== Invalid write of size 8
==40129==    at 0x4C2E8BE: memcpy@@GLIBC_2.14 (vg_replace_strmem.c:1035)
==40129==    by 0xC22DBCF1: arcticdb::Segment::serialize_v1_header_to_buffer(unsigned long) (segment.cpp:301)
==40129==    by 0xC22DBED6: arcticdb::Segment::serialize_header_v1() (segment.cpp:320)
==40129==    by 0xC22DBF5D: arcticdb::Segment::serialize_header() (segment.cpp:326)
==40129==    by 0xC3BD7A30: arcticdb::storage::s3::S3ClientImpl::put_object(std::string const&, arcticdb::Segment&, std::string const&, arcticdb::storage::s3::PutHeader) (s3_client_impl.cpp:192)
==40129==    by 0xC3BF1289: void arcticdb::storage::s3::detail::do_write_impl<arcticdb::storage::object_store_utils::FlatBucketizer>(arcticdb::storage::KeySegmentPair&, std::string const&, std::string const&, arcticdb::storage::s3::S3ClientInterface&, arcticdb::storage::object_store_utils::FlatBucketizer&&) (detail-inl.hpp:140)
==40129==    by 0xC3BE98A9: arcticdb::storage::s3::S3Storage::do_write(arcticdb::storage::KeySegmentPair&) (s3_storage.cpp:50)
==40129==    by 0xE608111: void arcticdb::storage::Storage::write<arcticdb::storage::KeySegmentPair&>(arcticdb::storage::KeySegmentPair&) (storage.hpp:48)
==40129==    by 0xE5F5C8B: arcticdb::storage::Storages::write(arcticdb::storage::KeySegmentPair&) (storages.hpp:46)
==40129==    by 0xE5F8245: arcticdb::storage::Library::write(arcticdb::storage::KeySegmentPair&) (library.hpp:90)
==40129==    by 0xE698D85: arcticdb::async::AsyncStore<arcticdb::util::SysClock>::write_compressed_sync(arcticdb::storage::KeySegmentPair) (async_store.hpp:180)
==40129==    by 0xE70ABDF: arcticdb::async::CopyCompressedInterStoreTask::copy() (tasks.hpp:402)
==40129==  Address 0x1416413b0 is 0 bytes after a block of size 1,151,856 alloc'd
==40129==    at 0x4C29F73: malloc (vg_replace_malloc.c:309)
==40129==    by 0xC373EE45: arcticdb::AllocatorImpl<arcticdb::NullTracingPolicy, arcticdb::util::LinearClock>::internal_alloc(unsigned long) (allocator.cpp:194)
==40129==    by 0xC373F143: arcticdb::AllocatorImpl<arcticdb::NullTracingPolicy, arcticdb::util::LinearClock>::aligned_alloc(unsigned long, bool) (allocator.cpp:303)
==40129==    by 0xC1DAD983: arcticdb::Buffer::resize(unsigned long) (buffer.hpp:232)
==40129==    by 0xC1DAD78A: arcticdb::Buffer::ensure(unsigned long) (buffer.hpp:177)
==40129==    by 0xC22DBBC5: arcticdb::Segment::serialize_v1_header_to_buffer(unsigned long) (segment.cpp:283)
==40129==    by 0xC22DBED6: arcticdb::Segment::serialize_header_v1() (segment.cpp:320)
==40129==    by 0xC22DBF5D: arcticdb::Segment::serialize_header() (segment.cpp:326)
==40129==    by 0xC3BD7A30: arcticdb::storage::s3::S3ClientImpl::put_object(std::string const&, arcticdb::Segment&, std::string const&, arcticdb::storage::s3::PutHeader) (s3_client_impl.cpp:192)
==40129==    by 0xC3BF1289: void arcticdb::storage::s3::detail::do_write_impl<arcticdb::storage::object_store_utils::FlatBucketizer>(arcticdb::storage::KeySegmentPair&, std::string const&, std::string const&, arcticdb::storage::s3::S3ClientInterface&, arcticdb::storage::object_store_utils::FlatBucketizer&&) (detail-inl.hpp:140)
==40129==    by 0xC3BE98A9: arcticdb::storage::s3::S3Storage::do_write(arcticdb::storage::KeySegmentPair&) (s3_storage.cpp:50)
==40129==    by 0xE608111: void arcticdb::storage::Storage::write<arcticdb::storage::KeySegmentPair&>(arcticdb::storage::KeySegmentPair&) (storage.hpp:48)
==40129== 
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ docs/mkdocs/docs/notebooks/.ipynb_checkpoints/
 # Ignore asv results
 python/.asv/html/
 python/.asv/results/
+
+# Ignore azurite files
+__azurite_*
--- a/python/arcticdb/util/venv.py
+++ b/python/arcticdb/util/venv.py
@@ -42,8 +42,9 @@ def run_shell_command(
             stdin=subprocess.DEVNULL,
         )
     if result.returncode != 0:
-        logger.warning(
-            f"Command failed, stdout: {str(result.stdout)}, stderr: {str(result.stderr)}"
+        logger.error(
+            f"Command '{command_string}' failed with return code {result.returncode}\n"
+            f"stdout:\n{result.stdout.decode('utf-8')}\nstderr:\n{result.stderr.decode('utf-8')}"
         )
     return result
 
@@ -145,6 +146,10 @@ class VenvArctic:
     def get_library(self, lib_name: str) -> "VenvLib":
         return VenvLib(self, lib_name, create_if_not_exists=False)
 
+    def cleanup(self):
+        ac = Arctic(self.uri)
+        for lib in ac.list_libraries():
+            ac.delete_library(lib)
 
 class VenvLib:
     def __init__(self, arctic, lib_name, create_if_not_exists=True):
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -1161,4 +1161,6 @@ def old_venv_and_arctic_uri(old_venv, arctic_uri):
             "LMDB storage backed has a bug in versions before 5.0.0 which leads to flaky segfaults"
         )
 
-    return old_venv, arctic_uri
+    yield old_venv, arctic_uri
+
+    old_venv.create_arctic(arctic_uri).cleanup() 


cpp/arcticdb/codec/segment.hpp
```diff
@@ -151,8 +151,7 @@ class Segment {
     }
 
     [[nodiscard]] std::size_t calculate_size() {
-        if(!size_.has_value())
-            size_ = FIXED_HEADER_SIZE + segment_header_bytes_size() + buffer_bytes();
+        size_ = FIXED_HEADER_SIZE + segment_header_bytes_size() + buffer_bytes();
```
IvoDD, line 154: Just leaving a note as I had to think about this. This should be fine performance-wise.

Calling `segment_header_bytes_size()` might be expensive if we end up constructing the protobuf, but we will do that only if it is not already constructed.

There is the very niche edge case of `segment.clone().calculate_size()` which will now be slower because the clone doesn't get the protobuf constructed, but I don't think we ever do that, so it should be fine.

IvoDD (changes requested):
Only the suggested use of `total_size` instead of `calculate_size()` is required, others are optional / out of scope.

cpp/arcticdb/codec/segment.cpp
```diff
@@ -279,13 +280,36 @@ std::pair<uint8_t*, size_t> Segment::serialize_v1_header_in_place(size_t total_h
 
 std::tuple<uint8_t*, size_t, std::unique_ptr<Buffer>> Segment::serialize_v1_header_to_buffer(size_t hdr_size) {
     auto tmp = std::make_unique<Buffer>();
-    ARCTICDB_TRACE(log::storage(), "Header doesn't fit in internal buffer, needed {} bytes but had {}, writing to temp buffer at {:x}", hdr_size, buffer_.preamble_bytes(),uintptr_t(tmp->data()));
-    tmp->ensure(calculate_size());
+    auto bytes_to_copy = buffer().bytes();
+    auto offset = FIXED_HEADER_SIZE + hdr_size;
+
+    auto total_size = offset + bytes_to_copy;
```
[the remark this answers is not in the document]
G-D-Petrov: I've fixed the first suggestion.

As for the question, it is probably better to not have the `hdr_size` at all.
But is is used in `serialize_header_v1` and I think that it makes that function more readable by having it.
In any case, I'd leave it like this for now.

cpp/arcticdb/codec/segment.hpp
G-D-Petrov: Yeah `segment.clone().calculate_size()` will be slower but I think that this will be the correct behavior because we could have the exact problem that I'm trying to fix here where:
1. replication reads a version written by an older writter
2. it tries to calculate the size but it gets a stale value for the current version used by replication

IvoDD (approved):

cpp/arcticdb/codec/segment.cpp
```diff
@@ -279,14 +280,40 @@ std::pair<uint8_t*, size_t> Segment::serialize_v1_header_in_place(size_t total_h
...
+    if (src != nullptr) {   
+        std::memcpy(final_dst,
+                    src,
+                    bytes_to_copy);
+    } else {
+        util::check(bytes_to_copy == 0, "Expected bytes_to_copy to be 0 when src is nullptr");
+        ARCTICDB_DEBUG(log::codec(), "src is nullptr, skipping memcpy");
+    }
+
+    return std::make_tuple(tmp->preamble(), total_size, std::move(tmp));
```
vasil-pashov, line 316: Order of evaluation of parameters in C++ is not specified. Thus the compiler is allowed to first perform std::move(tmp) and then tmp->preamble(). The joys of C++...

cpp/arcticdb/codec/segment.cpp
```diff
@@ -279,14 +280,40 @@ std::pair<uint8_t*, size_t> Segment::serialize_v1_header_in_place(size_t total_h
...
+
+    // Verify we have enough space for everything
+    tmp->ensure(total_size);
+    util::check(tmp->available() >= total_size,
+        "Buffer available space {} is less than required size {}", 
+        tmp->available(), 
+        total_size);
+
+    // This is both a sanity check and a way to populate the segment with the correct size
+    util::check(total_size == calculate_size(), "Expected total size {} to be equal to calculated size {}", total_size, calculate_size());
```
vasil-pashov, line 296: This will call `calculate_size()` twice do we mind it?

vasil-pashov (changes requested):

vasil-pashov:
While this solves the issue I'm wondering if it won't leave the door opened for more similar issues. When I debugged this in the context of the StorageMover I found out we're calling Segment::from_buffer which creates a segment passing the wrong size. So basically if we don't call calculate_size() on that object it will still hold the wrong header size. So I'm wondering if this is the only place that needs fixing?

G-D-Petrov:
> While this solves the issue I'm wondering if it won't leave the door opened for more similar issues. When I debugged this in the context of the StorageMover I found out we're calling Segment::from_buffer which creates a segment passing the wrong size. So basically if we don't call calculate_size() on that object it will still hold the wrong header size. So I'm wondering if this is the only place that needs fixing?

AFAIK there should be no other problems for v1:
- `Segment::from_buffer` actually sets the correct size, as it parses the proto header that is written to storage and takes its size, so the size based on the older version
- the problem comes when we try to serialize the segment with a new header and its size is bigger

This is fixed in this PR when serializing it to a buffer.
When serializing in place, there is no problem as it makes similar checks to the ones implemented in this PR.

I agree that we should debug a bit further and maybe fix the logic around the Segment size.
But I'd like to make this as a separate PR.

cpp/arcticdb/codec/segment.cpp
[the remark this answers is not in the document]
willdealtry: I think the idea is that calculating the header size is not necessarily particularly cheap, because it includes a protobuf and getting the protobuf serialized size requires traversing all the fields in the structure, which in the case of V1 is really quite a lot of them, because there are repeated fields in there for the encoded fields and stream descriptor fields, and the have strings so there's a lot of working out to do including rather scattered memory accesses.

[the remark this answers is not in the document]
willdealtry: We could cache value in the proto_size() just like we cache the header itself, then we could get rid of the hdr_size that is passed around, at the moment it's weird because the header is cached but the size isn't.

cpp/arcticdb/codec/segment.hpp
willdealtry: I don't fully understand exactly why this issue comes about. Is the problem that we're cloning a segment, and some things are being copied but others are being recreated?

vasil-pashov (approved):

cpp/arcticdb/codec/segment.hpp
G-D-Petrov: @willdealtry what you are proposing is what is already happening. 
What Ivo is saying is that if we:
1. clone a segment 
2. try get the size with `calculate_size()`
3. it will need to regenerate the proto the first time, because the proto is not being cloned and we can't be sure that the size of the segment is the correct size (e.g. we could've read a segment written by an older version, cloned it and now we are trying to serialize it)

Which is the correct behavior, I think.

In general, we are not cloning segments much, so this shouldn't be a problem.
And we still have the option of calling `size() `which will contain the size of the original segment and will not need to regenerate the proto.

PR #2191, opened 2025-02-19
Fix off by one errors in staged segment validation

poodlewars:
Our validation when finalizing is too strict at the moment. The index start and end on the APPEND_DATA key is `[start_time, end_time+1]` of data contained within its segment. Our validation logic is not always substracting one from the end time on the key to get the true date range in the segment.

We need a similar change when detecting duplicate staged segments. If two staged segments, both covering a single duplicated index value, are staged, we should allow the write.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_parallel.py
+++ b/python/tests/unit/arcticdb/version_store/test_parallel.py
@@ -27,9 +27,10 @@ from arcticdb.util.test import (
 )
 from arcticdb.util._versions import IS_PANDAS_TWO
 from arcticdb.version_store.library import Library
+from arcticdb_ext.exceptions import UnsortedDataException
 from arcticdb_ext.storage import KeyType
 
-from arcticdb import util
+from arcticdb import util, LibraryOptions
 
 from arcticdb.util.test import config_context_multi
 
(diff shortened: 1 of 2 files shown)


python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -1437,3 +1438,140 @@ def test_writing_wide_segment_over_sliced_data(
...
+        pd.Timestamp(1000),
+        pd.Timestamp(1000),
+    ]
+
+    data = pd.DataFrame({"a": len(idx)}, index=idx)
+    lib.write("test", data, staged=True)
+
+    lt = lib._nvs.library_tool()
+    append_keys = lt.find_keys_for_id(KeyType.APPEND_DATA, "test")
+    assert len(append_keys) == 2
```
IvoDD, line 1470: Would be nice to also assert the start_index and end_index are what we expect:
```
assert sorted([key.start_index for key in append_keys]) == [0, 1000]
assert [key.end_index for key in append_keys] == [1001, 1001]
```

IvoDD: And maybe similar for `test_chunks_are_the_same`

IvoDD (approved):

PR #2194, opened 2025-02-20
Skip the compat tests in Conda

poodlewars:
## Change Type (Required)
- [ ] **Cherry pick**

They aren't reliable at the moment, see https://github.com/conda-forge/arcticdb-feedstock/pull/396/checks?check_run_id=37543268460 

Since they rely on pip anyway to setup the venv, it's probably best to just skip them in Conda.


ognyanstoimenov (approved):

IvoDD (approved):
Yeah we were previously explicitly skipping them just on the conda feedstock CI [here](https://github.com/conda-forge/arcticdb-feedstock/blob/51d337d37f304638d6d40059e70514fa12417cf6/recipe/meta.yaml#L158)

But that's harder to maintain, so skipping here even though they pass on our CI makes sense.

PR #2201, opened 2025-02-25
Resampling using dynamic schema [Part 1]

vasil-pashov:
#### Reference Issues/PRs
This enables using the QueryBuilder to resample libraries where dynamic schema is enabled. This requires handling two cases which are different in comparison to static schema.

1. Type propagation. It turns out that it works without any functional changes in the code.
2. Missing columns. Dynamic schema allows for segments to contain only a subset of all columns in the symbol. This case needed additional care. 

*Handling missing columns*
Two changes make this possible. First is in `std::vector<EntityId> ResampleClause<closed_boundary>::process(std::vector<EntityId>&& entity_ids)` it now builds a bitmap per column. The bitmap has length equal to the number index segments (as we have it as invariant that the index cannot be sparse) and has a bit set to 1 if the column appears in a particular segment. The vector of segments passed for aggregation has length equal to the number of set bits in the bitset. If all bits are set to 1 we go down the static schema way which is a bit more efficient otherwise we go down the path of "not all segments have the column"

The second change is in `Column SortedAggregator<aggregation_operator, closed_boundary>::aggregate`. Now it has two overloads. One taking a dense vector of columns. Where the i-th column corresponds to the i-th index segment. And a second overload that takes a vector of columns whose size is less than the number of index segments and a bitmap. The length of the bitmap is equal to the count of the index segments. We then iterate over the set bits in the bitmap.

The initial idea used std::vector<std::optional<Column>> but this makes the "dense" codepath a bit better.

*Code refactoring*
There's some code refactoring going on. Mostly moving things from `sorted_aggregation.hpp` to `sorted_aggregation.cpp`. Some of the functions did not require the state of `SortedAggregator` and were made free functions.

*Missing behavior*
The current implementation has a flaws related to using date_range

1. If `date_range` is passed to the QueryBuilder and the date range does not contain some column (which is otherwise in the data frame) we will return either an empty dataframe (if no other columns are being aggregated) or a dataframe not containing that column. More sensible behavior would be to backfill the missing buckets with a default value based on the dtype and the aggregator.
2. Type propagation takes into account only the segments using in the aggregation and the dtypes from the segment descriptor. This means that different date ranges can result in dataframes whose columns have different dtypes. More sensible behavior would be to use the timeseries descriptor.

@alexowens90 is working on making the TSD easily available. Both of the above can be implemented after that.

## Change Type (Required)
- [ ] **Patch** (Bug fix or non-breaking improvement)
- [ ] **Cherry pick**

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/column_store/column_data.hpp
+++ b/cpp/arcticdb/column_store/column_data.hpp
@@ -166,16 +166,18 @@ public:
         ssize_t idx_{0};
         RawType* ptr_{nullptr};
 
-        [[nodiscard]] inline ssize_t idx() const {
+        Enumeration() = default;
+
+        [[nodiscard]] ssize_t idx() const {
             return idx_;
         }
 
-        inline RawType& value() {
+        RawType& value() {
             debug::check<ErrorCode::E_ASSERTION_FAILURE>(ptr_ != nullptr, "Dereferencing nullptr in enumerating ColumnDataIterator");
             return *ptr_;
         };
 
-        inline const RawType& value() const {
+        const RawType& value() const {
             debug::check<ErrorCode::E_ASSERTION_FAILURE>(ptr_ != nullptr, "Dereferencing nullptr in enumerating ColumnDataIterator");
             return *ptr_;
         };
@@ -188,14 +190,14 @@ public:
     };
 
     template <class T, IteratorType iterator_type>
-    using IteratorValueType_t = typename std::conditional_t<
+    using IteratorValueType_t = std::conditional_t<
             iterator_type == IteratorType::ENUMERATED,
             Enumeration<T>,
             PointerWrapper<T>
     >;
 
     template <class T, IteratorType iterator_type, bool constant>
-    using IteratorReferenceType_t = typename std::conditional_t<
+    using IteratorReferenceType_t = std::conditional_t<
         iterator_type == IteratorType::ENUMERATED,
             std::conditional_t<constant, const Enumeration<T>, Enumeration<T>>,
             std::conditional_t<constant, const T, T>
@@ -217,8 +219,7 @@ public:
         using base_type = base_iterator_type<TDT, iterator_type, iterator_density, constant>;
         using RawType = typename TDT::DataTypeTag::raw_type;
     public:
-        ColumnDataIterator() = delete;
-
+        ColumnDataIterator() = default;
         // Used to construct [c]begin iterators
         explicit ColumnDataIterator(ColumnData* parent):
         parent_(parent)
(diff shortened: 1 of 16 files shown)


cpp/arcticdb/column_store/column_data.hpp
```diff
@@ -217,8 +219,7 @@ struct ColumnData {
         using base_type = base_iterator_type<TDT, iterator_type, iterator_density, constant>;
         using RawType = typename TDT::DataTypeTag::raw_type;
     public:
-        ColumnDataIterator() = delete;
-
+        ColumnDataIterator() = default;
```
vasil-pashov, line 222: This was intentional change. If `ColumnDataIterator` has a default constructor it implements the concept needed by most of the algorithms and classes in `std::ranges`. I used this in particular to create a `std::ranges::subrange` in `sorted_aggregation.cpp` and avoid passing iterators around.

cpp/arcticdb/entity/types.cpp
vasil-pashov, line 1: No functional changes here. Copy/paste from types.hpp

cpp/arcticdb/entity/types.hpp
vasil-pashov, line 1: Moved to types.cpp

cpp/arcticdb/processing/sorted_aggregation.hpp
```diff
@@ -29,40 +30,48 @@ struct ISortedAggregator {
...
+            const bm::bvector<>& existing_columns)>(&T::aggregate)
+    >;
 };
 
 using SortedAggregatorInterface = folly::Poly<ISortedAggregator>;
 
 template<ResampleBoundary closed_boundary>
 class Bucket {
 public:
-    Bucket(timestamp start, timestamp end):
```
vasil-pashov, line 49: Moved to sorted_aggregation.cpp

cpp/arcticdb/processing/sorted_aggregation.hpp
```diff
@@ -345,130 +354,44 @@ class SortedAggregator
...
-                return LastAggregatorSorted<typename scalar_type_info::RawType, true>();
-            } else if constexpr (is_numeric_type(scalar_type_info::data_type) || is_bool_type(scalar_type_info::data_type)) {
-                return LastAggregatorSorted<typename scalar_type_info::RawType>();
-            } else if constexpr (is_sequence_type(scalar_type_info::data_type)) {
-                return LastAggregatorSorted<std::optional<std::string_view>>();
-            }
-        } else if constexpr (aggregation_operator == AggregationOperator::COUNT) {
-            return CountAggregatorSorted();
-        }
-    }
```
vasil-pashov, line 439: Moved to sorted_aggregation.cpp

alexowens90 (changes requested):

PR #2202, opened 2025-02-25
Conversion from pyarrow `Expression`s to `QueryBuilder` expressions

IvoDD:
#### Reference Issues/PRs
Monday ref: 8471524478

Introduces `ExpressionNode._from_pyarrow_expression_str`.

This is required for predicate pushdown integration with polars. E.g. when we do:
```
lf = polars.scan_arcticdb(arctic_identifier)
lf = lf.filter(pl.col("float_col") <= 40.1)
lf.collect()
```
When calling `collect` the filter will get pushed down to our (to be implemented) `polars.scan_arcticdb` via a callback. The filter is given as a string which can be evaluated to a pyarrow expression. For reference string construction happens [here](https://github.com/pola-rs/polars/blob/4e286d8c83b0fcb56f0a7ea06d2eb731f179e01e/crates/polars-plan/src/plans/python/pyarrow.rs#L25).

I've decided to keep this conversion code in core arcticdb instead of polars for a few reasons:
- Gives us more flexibility if we decide to update our `QueryBuilder` expressions
- We'll have less code to maintain in a repository which is not ours
- Automated tests allow us to not accidentally break polars predicate pushdown to arcticdb

For additional reference see how pyiceberg handles predicate pushdown from polars [here](https://github.com/pola-rs/polars/blob/4e286d8c83b0fcb56f0a7ea06d2eb731f179e01e/py-polars/polars/io/iceberg.py#L197-L200)

And how this might fit in the prototype `polars.scan_arcticdb` implementation [here](https://github.com/IvoDD/polars/pull/1/files)

## Change Type (Required)
- [ ] **Cherry pick**

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/processing.py
+++ b/python/arcticdb/version_store/processing.py
@@ -257,7 +257,7 @@ class ExpressionNode:
 
 
     @classmethod
-    def from_pyarrow_expression_str(cls, expression_str : str, function_map : Optional[Dict[str, Callable]] = None) -> "ExpressionNode":
+    def _from_pyarrow_expression_str(cls, expression_str : str, function_map : Optional[Dict[str, Callable]] = None) -> "ExpressionNode":
         """
         Builds an ExpressionNode from a pyarrow expression string.
 
--- a/python/arcticdb/version_store/processing.py
+++ b/python/arcticdb/version_store/processing.py
@@ -7,6 +7,7 @@ As of the Change Date specified in that file, in accordance with the Business So
 """
 from collections import namedtuple
 import copy
+from collections.abc import Callable
 from dataclasses import dataclass
 import datetime
 from math import inf
@@ -15,7 +16,12 @@ import numpy as np
 import pandas as pd
 from pandas.tseries.frequencies import to_offset
 
-from typing import Dict, NamedTuple, Optional, Tuple, Union
+from typing import Dict, NamedTuple, Optional, Tuple, Union, Any
+
+import sys
+import ast
+
+from functools import singledispatch
 
 from arcticdb.exceptions import ArcticDbNotYetImplemented, ArcticNativeException, UserInputException
 from arcticdb.version_store._normalization import normalize_dt_range_to_ts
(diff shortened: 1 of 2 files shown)


python/arcticdb/version_store/processing.py
```diff
@@ -250,6 +256,155 @@ def get_name(self):
...
+
+@_ast_to_expression.register(ast.UnaryOp)
+def _(a: ast.UnaryOp, function_map) -> Any:
+    operand = _ast_to_expression(a.operand, function_map)
+    if isinstance(a.op, ast.Invert):
+        return ~operand
+    if isinstance(a.op, ast.USub):
+        # pyarrow expressions don't support unary subrtract, so this branch will not be reached.
+        # Leaving as future-proofing in case they ever introduce it.
+        return -operand
```
alexowens90, line 313: Abs operator?

IvoDD: pyarrow expressions don't support `abs`. E.g. the following:
```
>>> import pyarrow.compute as pc
>>> abs(pc.field("asdf"))
TypeError: bad operand type for abs(): 'pyarrow._compute.Expression'
```

Do you think it's worth explicitly adding it just to future-proof in case pyarrow starts to support it? 
Also currently `abs` will work out of the box if you pass function_map=`{"abs": abs}` as well.

python/tests/unit/arcticdb/version_store/test_query_builder_parse_pyarrow.py
alexowens90, line 1: Will projections work with this code? If so we should add some tests.
How are other clauses going to work? I'm mainly thinking about groupbys and aggregations, but maybe resampling too.

python/arcticdb/version_store/processing.py
```diff
@@ -250,6 +256,155 @@ def get_name(self):
...
+    if isinstance(op, ast.Div):
+        return lhs / rhs
+    raise ValueError(f"Unexpected BinOp: {op}")
+
+
+@_ast_to_expression.register(ast.Compare)
+def _(a: ast.Compare, function_map) -> Any:
+    # Compares in pyarrow Expression contain exactly one comparison (i.e. 1 < field("asdf") < 3 is not supported)
+    assert len(a.ops) == 1
+    assert len(a.comparators) == 1
```
alexowens90, line 381: We should make these a `check` rather than asserts

python/arcticdb/version_store/processing.py
```diff
@@ -250,6 +256,155 @@ def get_name(self):
...
+    if isinstance(op, ast.Lt):
+        return lhs < rhs
+    if isinstance(op, ast.LtE):
+        return lhs <= rhs
+    raise ValueError(f"Unknown comparison: {op}")
+
+
+@_ast_to_expression.register(ast.List)
+def _(a: ast.List, function_map) -> Any:
+    return [_ast_to_expression(e, function_map) for e in a.elts]
```
alexowens90, line 405: We support a bunch of iterables for set membership operations, looks like we should register for `ast.Tuple` and `ast.Set` at least as well

python/arcticdb/version_store/processing.py
```diff
@@ -250,6 +256,155 @@ def get_name(self):
...
+        return ~operand
+    if isinstance(a.op, ast.USub):
+        # pyarrow expressions don't support unary subrtract, so this branch will not be reached.
+        # Leaving as future-proofing in case they ever introduce it.
+        return -operand
+    raise ValueError(f"Unexpected UnaryOp: {a.op}")
+
+
+@_ast_to_expression.register(ast.Call)
+def _(a: ast.Call, function_map) -> Any:
```
alexowens90, line 318: I guess this will be immediately useful for the `abs` case discussed in the other comment. Presumably it's best practice to register this even if the returned node will almost certainly raise a `ValueError` with whatever node it is passed to as an operand?

python/arcticdb/version_store/processing.py
```diff
@@ -250,6 +256,155 @@ def get_name(self):
...
+    raise ValueError(f"Unexpected function call: {f}")
+
+
+@_ast_to_expression.register(ast.Attribute)
+def _(a: ast.Attribute, function_map) -> Any:
+    value = _ast_to_expression(a.value, function_map)
+    attr = a.attr
+    if isinstance(value, ExpressionNode):
+        # Handles expression function attributes like (<some expression>).isin([1, 2, 3])
+        if attr == "isin":
```
alexowens90, line 335: We also support `isnotin`.
For both isin and isnotin, we also support equality syntax, like `(<some expression>) == [1, 2, 3]`. Would this currently be handled correctly?

python/tests/unit/arcticdb/version_store/test_query_builder_parse_pyarrow.py
```diff
@@ -0,0 +1,136 @@
...
+        "bool_col": [i%2 == 0 for i in range(num_rows)],
+        "datetime_col": pd.date_range(start=pd.Timestamp(2025, 1, 1), periods=num_rows)
+    }
+    index = pd.date_range(start=pd.Timestamp(2025, 1, 1), periods=num_rows)
+    return pd.DataFrame(data=data, index=index)
+
+
+def compare_against_pyarrow(pyarrow_expr_str, expected_adb_expr, lib, function_map = None, expect_equal=True):
+    adb_expr = ExpressionNode.from_pyarrow_expression_str(pyarrow_expr_str, function_map)
+    assert str(adb_expr) == str(expected_adb_expr)
```
alexowens90, line 26: I think it might be impossible to modify `ExpressionNode.__eq__` so that we could use `adb_expr == expected_adb_expr`
Could add a separate method, but this is fine for now

python/tests/unit/arcticdb/version_store/test_query_builder_parse_pyarrow.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        assert len(adb_result) != len(pa_result)
+
+
+def test_basic_filters(lmdb_version_store_v1):
+    lib = lmdb_version_store_v1
+
+    # Filter by boolean column
+    expr = f"pc.field('bool_col')"
+    expected_expr = ExpressionNode.column_ref('bool_col')
```
alexowens90, line 54: Open question, would these tests be easier to read if we used the `QueryBuilder` directly? Here it would be:
```
q = QueryBuilder()
q["bool_col"] # Or equivalently q.bool_col
```
and then in `compare_against_pyarrow` just check the first clause?

python/arcticdb/version_store/processing.py
IvoDD: polars doesn't produce strings which contain `Tuple`s or `Set`s. See how they always construct lists for their `Series` [here](https://github.com/pola-rs/polars/blob/4e286d8c83b0fcb56f0a7ea06d2eb731f179e01e/crates/polars-plan/src/plans/python/pyarrow.rs#L41-L65)

There is the general question of whether we should support every possible pyarrow expression or only the ones generated by polars predicate pushdown. 
On one hand I imagine this code will only ever be used for the polars integration and doesn't need to be more complex than it should be. On the other hand it is a very clear abstraction from polars to just say "this supports" all pyarrow expressions.

python/arcticdb/version_store/processing.py
IvoDD: The pyarrow expression doesn't have `isnotin` (see [docs](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Expression.html)). I think we probably shouldn't aim to support arbitrary string pyarrow-like expressions, but only support what we know pyarrow expressions can have.

As for the `pc.field("column") == [1, 2, 3]` I think it might have different semantics. I.e. I'd find it quite surprising if it means `isin` and not `column element equals to the array [1, 2, 3]`

python/tests/unit/arcticdb/version_store/test_query_builder_parse_pyarrow.py
IvoDD: yeah we want `ExpressionNode.__eq__` reserved for stuff like `q["col"]*2 == q["col2"]`, where we actually want to filter on the comparison between values in these columns.

I'll add a separate method to make clearer

python/arcticdb/version_store/processing.py
IvoDD: Testing with pyarrow we can construct the expression: `pc.field("column") == [1, 2, 3]` but it doesn't work when trying to use it because it tries to match for array equality:
```
pyarrow.lib.ArrowNotImplementedError: Function 'equal' has no kernel matching input types (string, list<item: string>)
```

python/arcticdb/version_store/processing.py
IvoDD: Adding those because I like the idea of us supporting arbitrary pyarrow expressions instead of only what polars would pushdown to us. Might turn out useful for something else also

alexowens90: Agreed, it's low-effort, and if pyarrow add stuff before Polars supports it, or Polars change their implementation, then we're not being restrictive

python/arcticdb/version_store/processing.py
alexowens90: Yeh makes sense, we'll have to rethink that syntax when we support array elements

alexowens90 (approved):

IvoDD:
After talking to polars we understand there is a new way to add io bindings to polars. [IO plugins](https://docs.pola.rs/user-guide/plugins/io_plugins/) are the way to go and this code should live inside the io plugin instead of inside ArcticDB.

PR #2185, opened 2025-02-18
Additional asv tests

grusev:
#### Reference Issues/PRs

More tests migrated to asv with S3 and new framework that allows LMDB, Amazon S3 and others:
- batch tests
- modification tests - update, append, delete
- query benchamarks
- finalized test data.

ASV run on S3 without errors here: https://github.com/man-group/ArcticDB/actions/runs/13414624562/job/37472454850
New link after allmodifications: https://github.com/man-group/ArcticDB/actions/runs/13703208873

## Change Type (Required)
- [x] **Patch** (Bug fix or non-breaking improvement)
- [ ] **Minor** (New feature, but backward compatible)
- [ ] **Major** (Breaking changes)
- [ ] **Cherry pick**

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #2135, opened 2025-01-23
Compare memory needs for dataframe vs read/write dataframe in lmdb and aws s3

grusev:
#### Reference Issues/PRs
<!--Example: Fixes #1234. See also #3456.-->

    Purpose of this test is to compare read and write operations memory
    efficieny over large dataframes with pandas and between storage.
        comparisons
            pandas_dataframe to read_lmdb
            pandas_dataframe to read_aws
            pandas_dataframe to write_lmdb
            pandas_dataframe to write_aws
            write_lmdb write_aws
            read_lmdb read_aws

During experiments it was determined that we can relate creation of a dataframe with read/write processes with lmdb and aws s3 store.

If a dataframe construction tooks X ammount mem the read/write processes might tak from 1.x to 2.5 timex X ammount of that memory for dataframes . The bigger the number of rows the bigger the multiplier.

The test tries to showcase an approach for creating tests that could address some concerns in future and the example below shows that this was an actual case.

NOTE: my s3 Wifi connection reported by windows is approx 1 MBytes/s
NOTE 2: The approach uses memory_profiler which I added to needed libs. 

LINK TO GITHUB EXECUTION: https://github.com/man-group/ArcticDB/actions/runs/12946758440/job/36112967427 


Log from a failed execution  which explains perhaps more than words
======================================
========================================================================================================= test session starts =========================================================================================================
platform linux -- Python 3.10.12, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/grusev/source/dependencies_fix
configfile: pyproject.toml
plugins: memray-1.7.0, cpp-2.6.0, xdist-3.6.1, hypothesis-6.72.4, timeout-2.3.1
collected 1 item                                                                                                                                                                                                                      

python/tests/stress/arcticdb/version_store/test_mem_comparison.py INFO:MemCompare:Creating dataframe with 2000000 rows
INFO:MemCompare:START: create_dataframe 
INFO:MemCompare:Time took: 1.589540958404541 
INFO:MemCompare:create_dataframe 1820.5390625MB
INFO:MemCompare:START: mem_write_dataframe_arctic_lmdb 
INFO:MemCompare:Time took: 2.4767887592315674 
INFO:MemCompare:mem_write_dataframe_arctic_lmdb 2754.63671875MB
INFO:MemCompare:START: mem_read_dataframe_arctic_lmdb 
INFO:MemCompare:Time took: 2.5483663082122803 
[...]

PR #2204, opened 2025-02-26
Comparison of memory efficiency LMDB, S3, Pandas for read/write operations on dataframes (new approach on making sure we accurately measure peakmem with ASV)

grusev:
#### Reference Issues/PRs

NOTE: Part of the review are only the the tests the utility.py and environment_setup.py are part of https://github.com/man-group/ArcticDB/pull/2185 they are here to support the creation of this PR only

    The test aims to compare efficiency of read and write operations
    in terms of memory on one graph
        - creation of dataframe
        - read/write dataframe
        - read/write dataframe to arcticdb LMDB
        - read/write dataframe to arcticdb Amazon S3

    The above problem is interesting in 2 ways:
     a) how to measure accurately the memory of certain operation with asv
     b) how to represent information of comparison effectively    

    To accurately measure the peakmem of any operation with ASV we have to exclude
    to base memory that is the object memory space from the calculation.
    
    To be able to do that we need to introduce memory baseline - a run of the test 
    which actually does nothing. It will produce result X
    
    Then each new run of the same test with operation that does anything will produce
    memory which is X + Z, where Z will be the actual memory of that specific operation

    The only logical way to represent then the information is to put all in one graph:
      - the base line
      - operation A
      - operation B

    Sample measurement:

             ============================ =======
                     backend_type
             ---------------------------- -------
                  no-operation-load        2.66G
              create-df-pandas-from_dict   3.17G
                    pandas-parquet          5.6G
                    arcticdb-lmdb          5.01G
             ============================ =======

IMPOTANT READ:
from: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory

Note
The peak memory benchmark also counts memory usage during the setup routine, which may confound the benchmark results. One way to avoid this is to use setup_cache instead.

My takeaways: 
We have many tests which have no way unless they make heavy use of setup() to setup some preconditions which are not possible to be done at setup_cache as they run in different processes. Therefore :
   = we cannot assume that number we get there is a number related only with the process inside peakmem
   - the process peak mem should be significant compared to base memory (have this in mind during planning tests)
   - if we get a signal for 10% increase of memory it might not be directly related to what is the actual test .. but also to operations in setup
   - peakmem can give us only a clue for the trend but perhaps not the actual numbers unless baseline as above is established. (this makes memray approach OR memory_profiler better suited for actual analysis (see https://github.com/man-group/ArcticDB/pull/2135)

## Change Type (Required)
- [ ] **Cherry pick**

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/benchmarks/real_comparison_benchmarks.py
```diff
@@ -0,0 +1,209 @@
...
+        df, dict = tpl
+        if btype == NO_OPERATION:
+            # What is the tool mem load?
+            return
+        if btype == CREATE_DATAFRAME:
+            df = pd.DataFrame(dict)
+        elif btype == PANDAS_PARQUET:
+            pd.read_parquet(self.path_to_read)
+        elif btype == ARCTICDB_LMDB:
+            self.lib.read(RealComparisonBenchmarks.SYMBOL)
```
G-D-Petrov, line 165: Why are we using `RealComparisonBenchmarks.SYMBOL` here?
Isnn't this for `lmdb`?

python/arcticdb/util/environment_setup.py
```diff
@@ -664,6 +665,322 @@ def clear_symbols_cache(self):
...
+        """
+        Returns next timestamp after the last timestamp in passed sequence of 
+        indexed dataframes.
+        """
+        last = self.get_first_and_last_timestamp(sequence_df_list)[1]
+        next = TimestampNumber.from_timestamp(last, freq) + 1
+        return next
+    
+
+class GeneralSetupOfLibrariesWithSymbols(EnvConfigurationBase):
```
G-D-Petrov, line 830: Why do we even need `GeneralSetupOfLibrariesWithSymbols` and `GeneralAppendSetup`?
I don't them used anywhere

python/benchmarks/real_comparison_benchmarks.py
```diff
@@ -0,0 +1,209 @@
...
+        self.lib = self.ac[RealComparisonBenchmarks.LIB_NAME]
+        self.path = f"{tempfile.gettempdir()}/df.parquet"
+        self.path_to_read = f"{tempfile.gettempdir()}/df_to_read.parquet"
+        self.delete_if_exists(self.path)
+        df.to_parquet(self.path_to_read, index=True)
+        self.pid = os.getpid()
+        # With shared storage we create different libs for each process
+        # therefore we initialize the symbol here also not in setup_cache
+        self.s3_lib = RealComparisonBenchmarks.REAL_STORAGE_SETUP.get_modifiable_library(self.pid)
+        self.s3_symbol = RealComparisonBenchmarks.REAL_STORAGE_SETUP.get_symbol_name_template(self.pid)
```
G-D-Petrov, line 130: This is very hard to read 
`RealComparisonBenchmarks.REAL_STORAGE_SETUP.get_symbol_name_template`

either:
- define this as a variable and use it that way
- refactor the setup to be functions rather than a class, given that they are just abstracting  pretty simple calls to the arctic client

G-D-Petrov (changes requested):
I am becoming less and less convinced of setup-as-classes approach.
It is starting to generate a lot of boilerplate code due to unnecessary abstractions.
Maybe we should have a separate task to refactor the approach and to make it more maintainable/easier to use

python/benchmarks/real_comparison_benchmarks.py
```diff
@@ -0,0 +1,209 @@
...
+        self.ac = Arctic(RealComparisonBenchmarks.URL)
+        self.lib = self.ac[RealComparisonBenchmarks.LIB_NAME]
+        self.path = f"{tempfile.gettempdir()}/df.parquet"
+        self.path_to_read = f"{tempfile.gettempdir()}/df_to_read.parquet"
+        self.delete_if_exists(self.path)
+        df.to_parquet(self.path_to_read, index=True)
+        self.pid = os.getpid()
+        # With shared storage we create different libs for each process
+        # therefore we initialize the symbol here also not in setup_cache
+        self.s3_lib = RealComparisonBenchmarks.REAL_STORAGE_SETUP.get_modifiable_library(self.pid)
```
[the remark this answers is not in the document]
grusev: All asv code must allways assume tests are are executed in separate processes. That prevents errors that will arise if and when this happens, intentianally or unintentionally (change in asv execution policy).

So actually I would always the opposite comment and actually require code change in any future tests that does convey to this policy.

In other words that assumption is for safety reasons and allows engineers to always write code that will  work despite any changes that may occur in future.

This will also become part of the wiki I started to write for the asv test that would run on many types of storages

python/arcticdb/util/environment_setup.py
grusev: as mentioned there is another PR in review where those are used in actual tests. So helper modules are taken from there. There those are used along. So this file is "shared" and any modifications to it will fo with the other PR. https://github.com/man-group/ArcticDB/pull/2185.

This means that this PR will be merged only after previous one as it will cary the needed functionality.

I chose to separete this test from others as it carries new approach to ASV tests, while the other PR is for tests that we know well and exists in LMDB only version.

python/benchmarks/real_comparison_benchmarks.py
grusev: yes. I beleive it is used exatly for LMDB

grusev:
This is already part of : https://github.com/man-group/ArcticDB/pull/2249

PR #2206, opened 2025-02-26
Tolerate keys being deleted while scanning object sizes

poodlewars:
Monday: 8560764974

### DIFF
--- a/cpp/arcticdb/async/batch_read_args.hpp
+++ b/cpp/arcticdb/async/batch_read_args.hpp
@@ -11,7 +11,7 @@
 
 namespace arcticdb {
 struct BatchReadArgs {
-    BatchReadArgs() {}
+    BatchReadArgs() = default;
 
     explicit BatchReadArgs(size_t batch_size) :
         batch_size_(batch_size) {}
--- a/cpp/arcticdb/async/batch_read_args.hpp
+++ b/cpp/arcticdb/async/batch_read_args.hpp
@@ -11,27 +11,11 @@
 
 namespace arcticdb {
 struct BatchReadArgs {
-    // The below enum controls where (IO or CPU thread pool) decoding and data processing tasks are executed.
-    // If IO is used, all work will be done in the IO thread pool which may prevent context switches - this is therefore
-    // suitable for work that is CPU light.
-    enum Scheduler {
-        IO,
-        CPU
-    };
-
-    BatchReadArgs() :
-        batch_size_(ConfigsMap::instance()->get_int("BatchRead.BatchSize", 200)),
-        scheduler_(Scheduler::CPU) {}
+    BatchReadArgs() {}
 
     explicit BatchReadArgs(size_t batch_size) :
-        batch_size_(batch_size),
-        scheduler_(Scheduler::CPU) {}
+        batch_size_(batch_size) {}
 
-    explicit BatchReadArgs(Scheduler scheduler) :
-        batch_size_(ConfigsMap::instance()->get_int("BatchRead.BatchSize", 200)),
-        scheduler_(scheduler) { }
-
-    size_t batch_size_;
-    Scheduler scheduler_;
+    size_t batch_size_ = ConfigsMap::instance()->get_int("BatchRead.BatchSize", 200);
 };
 }
--- a/cpp/arcticdb/toolbox/storage_mover.hpp
+++ b/cpp/arcticdb/toolbox/storage_mover.hpp
@@ -122,14 +122,18 @@ struct BatchCopier {
                                 ++skipped_;
                             }
                         }
-                        auto collected_kvs = source_store_->batch_read_compressed(std::move(keys_to_copy), BatchReadArgs{}).get();
-                        if (!collected_kvs.empty()) {
+
+                        size_t n_keys = keys_to_copy.size();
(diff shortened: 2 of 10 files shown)


cpp/arcticdb/async/batch_read_args.hpp
```diff
@@ -11,27 +11,11 @@
...
-    // suitable for work that is CPU light.
-    enum Scheduler {
-        IO,
-        CPU
-    };
-
-    BatchReadArgs() :
-        batch_size_(ConfigsMap::instance()->get_int("BatchRead.BatchSize", 200)),
-        scheduler_(Scheduler::CPU) {}
+    BatchReadArgs() {}
```
vasil-pashov, line 14: nit: 
```suggestion
    BatchReadArgs() = default;
```

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1701,15 +1701,33 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
 };
 
+using KeySizeCalculators = std::vector<std::pair<VariantKey, stream::StreamSource::ReadContinuation>>;
+
+static void read_ignoring_key_not_found(Store* store, KeySizeCalculators&& calculators) {
+    if (calculators.empty()) {
+        return;
+    }
+
+    std::vector<folly::Future<folly::Unit>> res;
```
vasil-pashov, line 1716: I think it'll be slightly more efficient to pull out `store->batch_read_compressed(std::move(calculators), BatchReadArgs{})` in a variable and then pre-allocate res using reserve.

vasil-pashov (approved):

cpp/arcticdb/toolbox/storage_mover.hpp
```diff
@@ -122,14 +122,18 @@ struct BatchCopier {
                                 ++skipped_;
                             }
                         }
-                        auto collected_kvs = source_store_->batch_read_compressed(std::move(keys_to_copy), BatchReadArgs{}).get();
-                        if (!collected_kvs.empty()) {
+
+                        size_t n_keys = keys_to_copy.size();
+                        auto collected_kvs = folly::collect(source_store_->batch_read_compressed(std::move(keys_to_copy), BatchReadArgs{}))
```
phoebusm, line 127: It looks like `collected_kvs` is not being used?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1701,15 +1701,33 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
 struct AtomicKeySizesInfo {
-    size_t count;
+    size_t count{0};
     std::atomic<size_t> compressed_size;
     std::atomic<size_t> uncompressed_size;
 };
 
+using KeySizeCalculators = std::vector<std::pair<VariantKey, stream::StreamSource::ReadContinuation>>;
+
+static void read_ignoring_key_not_found(Store* store, KeySizeCalculators&& calculators) {
```
alexowens90, line 1711: Pass store by const ref if null case is not handled

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1701,15 +1701,33 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
+    if (calculators.empty()) {
+        return;
+    }
+
+    std::vector<folly::Future<folly::Unit>> res;
+    for (auto&& fut : store->batch_read_compressed(std::move(calculators), BatchReadArgs{})) {
```
alexowens90, line 1717: `folly::collectAll` is designed for this sort of case

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1741,10 +1756,10 @@ std::unordered_map<StreamId, std::unordered_map<KeyType, KeySizesInfo>> LocalVer
     auto streams = symbol_list().get_symbols(store());
 
     foreach_key_type([&store=store(), &sizes, &mutex](KeyType key_type) {
-        std::vector<std::pair<VariantKey, stream::StreamSource::ReadContinuation>> keys;
+        std::vector<std::pair<VariantKey, stream::StreamSource::ReadContinuation>> key_size_calculators;
```
alexowens90, line 1759: Can use `KeySizeCalculators` here as well

alexowens90 (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: There's a reference in their docs to,

```
  The Futures are moved in, so your copies are invalid. If you need to
  chain further from these Futures, use the variant with an output iterator.
```

which sounds like what I want, but I can't find the actual function anywhere (perhaps stale docs). Without that, I can't find a way to use `collectAll` here that's clearer than what I already have.

cpp/arcticdb/toolbox/storage_mover.hpp
poodlewars: No, it has continuations that do all the work

PR #2208, opened 2025-02-27
Improve CI compile times by fixing the sccache mechanism

G-D-Petrov:
#### Reference Issues/PRs

Monday ref: 8590725039

Currently, the sscache mechanism doesn't work properly because:
1. there is not enough space in GHA to store all of the data
2. for windows specifically, the build is using precompiled headers and sccache doesn't support them and so can't cache the objects for them

This PR changes the flows to:
- use S3 as a backend for the cache for "internal" PRs (still falls back to GHA for PRs by external contributors)
- removes the precompiled headers for windows builds

**This results in 3-4x speed up for compilation (10-15 min now vs 30-45 min before) for all builds.**
Now the Linux builds + tests take less than an hour ( vs ~90 min before)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -48,8 +48,8 @@ jobs:
     runs-on: ${{ needs.start_ec2_runner.status != 'failure' && needs.start_ec2_runner.outputs.label || matrix.distro}}
     container: ${{ (matrix.os == 'linux' && inputs.job_type != 'build-python-wheels') && matrix.container || null}}
     env:
-      # 0 - uses S3 Cache, 1 - does uses GHA cache
-      # this was extract PRs can use the GHA cache
+      # 0 - uses S3 Cache, 1 - uses GHA cache
+      # this way the external PRs can use the GHA cache
       SCCACHE_GHA_VERSION: ${{secrets.AWS_S3_ACCESS_KEY && 0 || 1}}
       SCCACHE_BUCKET: arcticdb-ci-sccache-bucket
       SCCACHE_ENDPOINT: http://s3.eu-west-1.amazonaws.com
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -48,7 +48,15 @@ jobs:
     runs-on: ${{ needs.start_ec2_runner.status != 'failure' && needs.start_ec2_runner.outputs.label || matrix.distro}}
     container: ${{ (matrix.os == 'linux' && inputs.job_type != 'build-python-wheels') && matrix.container || null}}
     env:
-      SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
+      # 0 - uses S3 Cache, 1 - does uses GHA cache
+      # this was extract PRs can use the GHA cache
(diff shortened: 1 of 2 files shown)


.github/workflows/build_steps.yml
```diff
@@ -101,16 +110,12 @@ jobs:
           CMAKE_BUILD_PARALLEL_LEVEL: ${{vars.CMAKE_BUILD_PARALLEL_LEVEL}}
 
       # ========================= Leader steps =========================
-      - name: Remove GitHub default packages (baseimage) # To save space
```
G-D-Petrov, line 104: This is no longer needed as the cache is reducing the compilation load, i.e. no as much disk space is needed.

.github/workflows/build_steps.yml
```diff
@@ -48,7 +48,15 @@ jobs:
     runs-on: ${{ needs.start_ec2_runner.status != 'failure' && needs.start_ec2_runner.outputs.label || matrix.distro}}
     container: ${{ (matrix.os == 'linux' && inputs.job_type != 'build-python-wheels') && matrix.container || null}}
     env:
-      SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}} # Setting this env var enables the caching
+      # 0 - uses S3 Cache, 1 - does uses GHA cache
```
poodlewars, line 51: typo `does uses` should be `uses`

.github/workflows/build_steps.yml
```diff
@@ -48,7 +48,15 @@ jobs:
...
+      # this was extract PRs can use the GHA cache
```
poodlewars, line 52: Think there's a typo in this sentence, I don't understand it

.github/workflows/build_steps.yml
```diff
@@ -48,7 +48,15 @@ jobs:
...
+      SCCACHE_GHA_VERSION: ${{secrets.AWS_S3_ACCESS_KEY && 0 || 1}}
+      SCCACHE_BUCKET: arcticdb-ci-sccache-bucket
+      SCCACHE_ENDPOINT: http://s3.eu-west-1.amazonaws.com
```
poodlewars, line 55: Do you know what region our CI runners run in? Allegedly they're all in the US, so that might be a better location for the cache.

.github/workflows/build_steps.yml
poodlewars: May still need it if we ever tear down (or invalidate) the whole cache?

poodlewars (approved):

.github/workflows/build_steps.yml
G-D-Petrov: Doesn't seem like it, the wheels build was able to run even when an empty cache (when I changed to S3)
The step below is still needed for C++ tests jobs.

.github/workflows/build_steps.yml
G-D-Petrov: I will test this next week

vasil-pashov (approved):

G-D-Petrov:
Will be resolved with https://github.com/man-group/ArcticDB/pull/2215

PR #2209, opened 2025-02-27
Return correct timestamp in the VersionItem returned by write when recursive normalizers are used

vasil-pashov:
#### Reference Issues/PRs
The issue was reported by users using recursive normalizers. The write method
```python
v2lib._nvs.write(..., recursive_normalizers=True)
```
return a VersionedItem whose timestamp is 0.

This PR is split into two commits.
1. First commit moves all recursive normalizers unit tests from `python/tests/integration/arcticdb/version_store/test_basic_version_store.py` to `python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py`
2. The second commit implements the actual fix:
 
    * Python code change: set the timestamp from the VersionedItem returned
      by the C++ layer. The C++ layer uses the MULTI_KEY to create a
      VersionItem
    
    * This also changes the read side when reucrsive normalizers are used.
      Prior this change when a read with recursive normalizers was done it
      used one of the TABLE_INDEX keys to create versioned item. There wasn't
      any particular reasoning for whose TABLE_INDEX key to select. Since
      both index keys are created at different times this desn't reflect well
      the write timestamp. Now it uses the MULTI_KEY to create the
      VersionedItem. This way both read and write will return the same
      VersionedItem (except the data as write does not return data) and will
      also have a single deterministic timestamp for all symbols.
    
    * Add minor improvements such as reserving space for vectors with known
      sizes.
    
    * Remove `test_simple_recursive_normalizer` as it is repetitive

#### Any other comments?
Interesting remark related to creating versioned items. When doing regular read and write we create a versioned item based on the INDEX_DATA key and not based on the VERSION key. I find this a bit misleading mainly due to the name `VersionedItem`.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/version/version_core.cpp
```diff
@@ -500,21 +500,25 @@ VersionedItem update_impl(
...
     VersionedItem versioned_item{std::move(dup)};
     TimeseriesDescriptor multi_key_desc{index_key_seg.index_descriptor()};
 
     return read_frame_for_version(store, versioned_item, std::make_shared<ReadQuery>(), ReadOptions{}, handler_data)
-    .thenValue([multi_key_desc=std::move(multi_key_desc), keys=std::move(keys)](ReadVersionOutput&& read_version_output) mutable {
+    .thenValue([multi_key_desc=std::move(multi_key_desc), keys=std::move(keys), key=std::move(key)](ReadVersionOutput&& read_version_output) mutable {
         multi_key_desc.mutable_proto().mutable_normalization()->CopyFrom(read_version_output.frame_and_descriptor_.desc_.proto().normalization());
         read_version_output.frame_and_descriptor_.desc_ = std::move(multi_key_desc);
         read_version_output.frame_and_descriptor_.keys_ = std::move(keys);
+        read_version_output.versioned_item_ = VersionedItem(std::move(key));
```
vasil-pashov, line 521: Actual fix 1: Create the VersionedItem from the MULTI_KEY on read.

alexowens90 (approved):

poodlewars (approved):
I know your change doesn't change how we write these structures, but it would be good to add a compat test for reading and writing recursively normalised data with different ArcticDB versions while we're here

PR #2148, opened 2025-01-28
Draft query stat API

phoebusm:
#### What does this implement or fix?
Draft query stat API for discussion and as placeholder

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #2210, opened 2025-02-27
Query Stats framework

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/8297768017
* Add basic C++ framework to stat query 
* Add stat to list_symbols on s3 to demostrate the framework

#### Any other comments?
For python layer, the change is minimum, only serve the purpose of outputing something that can be verified in the test
The missing functions will be handled in later PRs.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/arcticdb/toolbox/stats_query.py
```diff
@@ -0,0 +1,101 @@
...
+        self._is_context_manager = False
+        StatsQuery.register_new_query_stat_tool()
+
+    def __del__(self):
+        StatsQuery.unregister_query_stat_tool()
+
+    def __sub__(self, other):
+        return self._populate_stats(other._create_time)
+
+    def _populate_stats(self, other_time):
```
phoebusm, line 21: Boilerplate code to beautify the output for now
Pending changes and improvement in later PRs

cpp/arcticdb/util/stats_query.hpp
```diff
@@ -0,0 +1,118 @@
...
+        stats.emplace_back(stats_instance->info_, std::make_pair(col_name, fmt::format("{}", std::forward<T>(value))));
+    }
+
+    StatsOutputFormat get_stats();
+    bool is_enabled();
+    void register_new_query_stat_tool();
+    void deregister_query_stat_tool();
+    static StatsQuery& instance();
+
+    // TODO: Support enable in the work of a particualr python thread only, need to
```
poodlewars, line 56: I don't understand what this means

cpp/arcticdb/util/stats_query.hpp
```diff
@@ -0,0 +1,118 @@
...
+    static StatsQuery& instance();
+
+    // TODO: Support enable in the work of a particualr python thread only, need to 
+    // 1. Convert this to python_thread_local variable; Check Py_tss_t in python.h
+    // 2. Remove atomic
+    std::atomic<bool> stats_query_enabled = false;
+private:
+    std::atomic<int32_t> query_stat_tool_count = 0;
+    std::mutex stats_mutex_;
+    //TODO: Change to std::list<std::pair<GroupableStats, std::pair<std::string, std::variant<std::string, xxx>>>
```
poodlewars, line 63: This desperately needs some type aliases or structs pulling out. And should implement the TODO now if you want it.

cpp/arcticdb/util/stats_query.hpp
```diff
@@ -0,0 +1,118 @@
...
+        }
+    private:
+        std::shared_ptr<StatsInstance> stats_instance_;
+        std::chrono::time_point<std::chrono::high_resolution_clock> start_;
+        bool log_time_;
+    };
+
+}
+
+#define GROUPABLE_STAT_NAME(x) stats_query_info##x
```
poodlewars, line 94: I don't understand how I'm meant to use these APIs. A C++ unit test suite would help. How does the grouping work? Am I able to specify a grouping on a composite like, `increment the counter for objects of this key type seen during this storage operation during this arcticdb operation`?

python/tests/integration/toolbox/test_stats_query.py
```diff
@@ -0,0 +1,19 @@
...
+def test_query_stats(s3_version_store_v1):
+    s3_version_store_v1.write("a", 1)
+    query_stats_tools_start = StatsQueryTool()
+    s3_version_store_v1.list_symbols()
+    query_stats_tools_end = StatsQueryTool()
+    stats = query_stats_tools_end - query_stats_tools_start
+    """
+    Expected output; time values are not deterministic
+    arcticdb_call stage key_type     storage_op parallelized count  time_count_20  time_count_510
+    0  list_streams  None     None           None         None  None              0               1
```
poodlewars, line 12: `list_streams` isn't a Python API method so won't mean anything to the user. How has it ended up in this output?

cpp/arcticdb/util/stats_query.hpp
poodlewars, line 1: I can see why it has this, but this design has a kind of "no schema" approach to the stats, they're calculated dynamically based on the macro invocations. I think it may be better to have a defined schema (just as a C++ class) for the stats. This would add an extra chore when adding a new stat, but I think would make the whole thing clearer to people who don't use these APIs all the time (people may add new stats a couple of times a year so won't be familiar with this framework).

poodlewars (changes requested):
Nice job figuring out how to pass this stuff through Folly. Can I suggest merging a PR to start with that just introduces the custom Folly executors (with a suite of tests to show that the stats calculation works with both our task based APIs and normal `Folly::.via`) and then we can figure out the other discussions after.

It would have been helpful if your PR description had explained your design.

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -174,11 +175,73 @@ inline auto get_default_num_cpus([[maybe_unused]] const std::string& cgroup_fold
...
  */
+
+class CustomIOThreadPoolExecutor : public folly::IOThreadPoolExecutor{
+public:
+    template<typename... Args>
+    CustomIOThreadPoolExecutor(Args&&... args) : folly::IOThreadPoolExecutor(std::forward<Args>(args)...){}
+
+    // The first overload function will call the second one in folly. Have to override both as they are overloading
+    // Called by the submitter when submitted to a executor
+    void add(folly::Func func) override {
```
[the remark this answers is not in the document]
phoebusm: It's a C++ syntax. To override a parent function which is overloaded, it is needed to override all

cpp/arcticdb/util/stats_query.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <arcticdb/util/stats_query.hpp>
+#include <arcticdb/util/preconditions.hpp>
+#include <arcticdb/log/log.hpp>
+
+namespace arcticdb::util::stats_query {
+
+std::shared_ptr<StatsInstance> StatsInstance::instance(){
+    if (!instance_) {
+        // TODO: Add checking for non python threads must have instance_ as non-nullptr
```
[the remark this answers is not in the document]
phoebusm: Added

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -194,17 +257,27 @@ class TaskScheduler {
         auto task = std::forward<decltype(t)>(t);
         static_assert(std::is_base_of_v<BaseTask, std::decay_t<Task>>, "Only supports Task derived from BaseTask");
         ARCTICDB_DEBUG(log::schedule(), "{} Submitting CPU task {}: {} of {}", uintptr_t(this), typeid(task).name(), cpu_exec_.getTaskQueueSize(), cpu_exec_.kDefaultMaxQueueSize);
+        // Executor::Add will be called before below function
```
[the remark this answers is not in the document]
phoebusm: Here `copy_instance` is needed, as the instance is needed to copied for each worker from caller.
The one in executor calls `pass_instance`, as the instance just needs to be passed along the pipeline

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -428,6 +428,7 @@ bool do_iterate_type_impl(
     if (prefix.empty()) {
         key_type_dir += "/";
     }
+    STATS_QUERY_ADD_GROUPABLE_STAT(key_type, key_type)
```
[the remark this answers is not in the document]
phoebusm: In `STATS_QUERY_ADD_GROUPABLE_STAT`, `arcticdb::util::stats_query::GroupableStat` is created if query stat is on. It's the key of the design as this class uses RAII to pop the groupable stat and calculate the time taken for calling a particular API/function
To pack the object declare and creation, marco is used here

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -207,6 +207,7 @@ VariantKey get_ref_key(const SnapshotId& snap_name) {
 
 std::optional<VariantKey> get_snapshot_key(const std::shared_ptr<Store>& store, const SnapshotId &snap_name) {
     ARCTICDB_SAMPLE(getSnapshot, 0)
+    STATS_QUERY_ADD_GROUPABLE_STAT(parallelized, "False")
```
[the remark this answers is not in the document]
phoebusm: (It's a bit sad to discuss the output schema in the implementation PR, as it has been shared previously: https://github.com/man-group/ArcticDB/pull/2148)
> I think let's not track this parallelized concept
My idea was the user may can get confused to see, e.g. an arcticdb call will have 20 KeyExists Task. But the total wall time for the call is not the sum of the time needed for running KeyExists Task. 
But for some other API calls, KeyExist task is not parallelized.

>we should instead measure elapsed wall clock time at a sensible level
This feature is in this PR already

python/tests/integration/toolbox/test_stats_query.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_query_stats(s3_version_store_v1):
+    s3_version_store_v1.write("a", 1)
+    query_stats_tools_start = StatsQueryTool()
+    s3_version_store_v1.list_symbols()
+    query_stats_tools_end = StatsQueryTool()
+    stats = query_stats_tools_end - query_stats_tools_start
+    """
+    Expected output; time values are not deterministic
+    arcticdb_call stage key_type     storage_op parallelized count  time_count_20  time_count_510
```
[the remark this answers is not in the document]
phoebusm: The titles will be more intuitive in the json format

python/tests/integration/toolbox/test_stats_query.py
phoebusm: It's manually named. Will be updated

cpp/arcticdb/util/stats_query.hpp
phoebusm: User may use arcticdb in a python multithread fashion, i.e. `threading.Thread`. Query stats should support being enabled in a particular thread only, though it's not the case atm. The switch is global for now.

Spoken to Maxim about it. This is not necessary so will remove this comment

cpp/arcticdb/util/stats_query.hpp
```diff
@@ -0,0 +1,118 @@
...
+    static void copy_instance(std::shared_ptr<StatsInstance>& ptr);
+    static void pass_instance(std::shared_ptr<StatsInstance>&& ptr);
+
+    GroupableStats info_;
+private:
+    thread_local inline static std::shared_ptr<StatsInstance> instance_;
+};
+
+// process-global stats entry list
+class GroupableStat;
```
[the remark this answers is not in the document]
phoebusm: It's for "groupable stats", e.g. `arcticdb_call`, `stage`, etc

cpp/arcticdb/util/stats_query.hpp
phoebusm: Added the explaination to the description of this PR

phoebusm:
>How are you calculating the histogram buckets?

They are hardcoded 10ms buckets

cpp/arcticdb/util/stats_query.hpp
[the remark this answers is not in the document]
poodlewars: I still think `StatsGroup` is clearer, it's defining a grouping axis for statistics

python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -0,0 +1,64 @@
...
+def test_query_stats(s3_version_store_v1):
+    s3_version_store_v1.write("a", 1)
+    query_stats_tools_start = QueryStatsTool()
+    s3_version_store_v1.list_symbols()
+    query_stats_tools_end = QueryStatsTool()
+    stats = query_stats_tools_end - query_stats_tools_start
+    """
+    Sample output:
+    {
+        "list_symbols": {
```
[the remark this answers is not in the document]
phoebusm: Oops I have missed it in the conversion from df to map. Let me add it back

cpp/arcticdb/util/query_stats.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+    return result;
+}
+
+bool QueryStats::is_enabled(){
+    return query_stats_enabled.load(std::memory_order_relaxed);
+}
+
+void QueryStats::register_new_query_stat_tool() {
+    auto new_stat_tool_count = query_stat_tool_count.fetch_add(1, std::memory_order_relaxed) + 1;
```
poodlewars, line 71: I don't really understand all these counters, what are these for? And the count isn't correct is it (adding one in this thread after the atomic fetch add)

python/tests/integration/toolbox/test_query_stats.py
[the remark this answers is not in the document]
poodlewars: Also discussed including how many symbols are in the `list_symbols` result set

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include <map>
+#include <atomic>
+#include <type_traits>
+#include <string>
+#include <ctime>
+#include <chrono>
+#include <fmt/format.h>
+
+namespace arcticdb::util::query_stats {
+using StatsGroups = std::vector<std::shared_ptr<std::pair<std::string, std::string>>>;
```
poodlewars, line 22: a type alias for the pair would help me read this

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,120 @@
...
+    void register_new_query_stat_tool();
+    void deregister_query_stat_tool();
+    static QueryStats& instance();
+
+    std::atomic<bool> query_stats_enabled = false;
+private:
+    std::atomic<int32_t> query_stat_tool_count = 0;
+    std::mutex stats_mutex_;
+    //TODO: Change to std::list<std::pair<StatsGroups, std::pair<std::string, std::variant<std::string, xxx>>> 
+    std::list<std::pair<StatsGroups, std::pair<std::string, std::string>>> stats;
```
poodlewars, line 61: pull out structs and type alias for these :pray:

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,120 @@
...
+    auto stats_instance = StatsInstance::instance();
+    StatsGroup stats_group(stats_instance, false, "exec_time", std::chrono::high_resolution_clock::now().time_since_epoch().count());
+    QueryStats::instance().add_stat(stats_instance, col_name, value);
+}
+
+}
+
+#define STATS_GROUP_NAME(x) query_stats_info##x
+
+#define QUERY_STATS_ADD_STATS_GROUP_IMPL(log_time, col_name, value) \
```
poodlewars, line 101: rename `QUERY_STATS_ADD_GROUP_IMPL`

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,120 @@
...
+    std::optional<arcticdb::util::query_stats::StatsGroup> STATS_GROUP_NAME(col_name); \
+    if (arcticdb::util::query_stats::QueryStats::instance().is_enabled()) { \
+        auto stats_instance = arcticdb::util::query_stats::StatsInstance::instance(); \
+        STATS_GROUP_NAME(col_name).emplace(stats_instance, log_time, #col_name, value); \
+    }
+#define QUERY_STATS_ADD_STATS_GROUP(col_name, value) QUERY_STATS_ADD_STATS_GROUP_IMPL(false, col_name, value)
```
poodlewars, line 107: rename `QUERY_STATS_ADD_GROUP`

cpp/arcticdb/util/query_stats.cpp
phoebusm: >I don't really understand all these counters, what are these for?

The counter is for keeping track how many python object `QueryStatsTool` is in-use: https://github.com/man-group/ArcticDB/blob/da3168991b6fd23929a7d3ec48d15a02946cd41c/python/arcticdb/toolbox/query_stats.py#L13
If the counter > 0, query stats is ON
If counter reaches 0, query stats is OFF and stats are cleared.

> And the count isn't correct is it (adding one in this thread after the atomic fetch add)

It fetches the value before adding. So without the `+ 1` at the end, that will be the old value

Test https://github.com/man-group/ArcticDB/blob/da3168991b6fd23929a7d3ec48d15a02946cd41c/python/tests/integration/toolbox/test_query_stats.py#L86 to demostrate the idea

cpp/arcticdb/util/stats_query.hpp
poodlewars: We spoke about this a lot and a stricter schema has its own big downsides, so OK sticking with this. The strict schema forces context to be passed between layers of the stack, which is painful

cpp/arcticdb/util/stats_query.hpp
poodlewars: This must have a C++ unit test suite

cpp/arcticdb/async/test/test_async.cpp
```diff
@@ -141,6 +141,90 @@ TEST(Async, CollectWithThrow) {
    ARCTICDB_DEBUG(log::version(), "Collect returned");
 }
 
+TEST(Async, StatsQueryDemo) {
```
phoebusm, line 144: Demo for how to add/use query stats

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -165,21 +169,53 @@ inline auto get_default_num_cpus([[maybe_unused]] const std::string& cgroup_fold
...
+        // Called by the submitter when submitted to a executor
+        // Could be called by worker thread to submit the subsequent task back to executor
+        void add(folly::Func func) override {
+            T::add(std::move(func));
+        }
+
+        void add(folly::Func func,
+            std::chrono::milliseconds expiration,
+            folly::Func expireCallback) override {
+            if (arcticdb::util::query_stats::QueryStats::instance().is_enabled_) {
```
poodlewars, line 189: Why not a function for whether it's enabled?

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -165,21 +169,53 @@ inline auto get_default_num_cpus([[maybe_unused]] const std::string& cgroup_fold
...
+                auto func_with_stat_query_wrap = [layer = util::query_stats::QueryStats::instance().current_layer(), func = std::move(func)](auto&&... vars) mutable{
```
poodlewars, line 190: Just call this `wrapped_func`

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -194,19 +230,45 @@ class TaskScheduler {
         auto task = std::forward<decltype(t)>(t);
         static_assert(std::is_base_of_v<BaseTask, std::decay_t<Task>>, "Only supports Task derived from BaseTask");
         ARCTICDB_DEBUG(log::schedule(), "{} Submitting CPU task {}: {} of {}", uintptr_t(this), typeid(task).name(), cpu_exec_.getTaskQueueSize(), cpu_exec_.kDefaultMaxQueueSize);
+        // Executor::Add will be called before below function
+        // TODO: Add checking to make sure QUERY_STATS_ADD_GROUP is called once before folly tasks, once all query stats entries are added
+        // so the query stats could be presented
         std::lock_guard lock{cpu_mutex_};
-        return cpu_exec_.addFuture(std::move(task));
+        if (arcticdb::util::query_stats::QueryStats::instance().is_enabled_) {
+            auto task_with_stat_query_instance = [&parent_thread_local_var = util::query_stats::QueryStats::instance().thread_local_var_, task = std::move(task)]() mutable{
```
poodlewars, line 238: Call this `wrapped_task`

python/arcticdb/toolbox/query_stats.py
```diff
@@ -0,0 +1,89 @@
+from contextlib import contextmanager
+import numpy as np
+
+from arcticdb_ext.tools import QueryStats
+from arcticdb_ext.tools.QueryStats import StatsGroupName, StatsName
+
+class QueryStatsTool:
```
poodlewars, line 7: Like I said in the other comment, we don't need a class for this. I think these can all be free functions defined in this module

python/arcticdb/toolbox/query_stats.py
```diff
@@ -0,0 +1,89 @@
...
+        
+        # Process each layer
+        for layer in raw_stats:
+            if layer:
+                cls._process_layer(layer, result)
+            
+        return result
+    
+    @classmethod
+    def _process_layer(cls, layer, current_dict):
```
poodlewars, line 37: Type hint for the return type please, I have no idea what this does

python/arcticdb/toolbox/query_stats.py
```diff
@@ -0,0 +1,89 @@
...
+    def context_manager(cls):
+        @contextmanager
+        def _func():
+            cls.enable()
+            yield
+            cls.disable()
+        return _func()
+
+    @classmethod
+    def get_query_stats(cls):
```
poodlewars, line 22: Type hints (throughout these APIs)

python/arcticdb/toolbox/query_stats.py
```diff
@@ -0,0 +1,89 @@
...
+        # Process stats array
+        stats_array = layer.stats
+        for stat_enum in cls._STATS_NAME_VALUES:
+            stat_idx = int(stat_enum)
+            if stats_array[stat_idx] > 0:
+                stat_name = _get_enum_name(stat_enum)
+                if stat_name not in current_dict:
+                    current_dict[stat_name] = stats_array[stat_idx]
+                else:
+                    current_dict[stat_name] += stats_array[stat_idx]
```
poodlewars, line 50: This is fine while the stats are all counters, just document the limitation carefully

python/arcticdb/toolbox/query_stats.py
```diff
@@ -0,0 +1,89 @@
...
+        def _get_enum_name(enum_value):
+            return str(enum_value).split('.')[-1]
+        
+        # Process stats array
+        stats_array = layer.stats
+        for stat_enum in cls._STATS_NAME_VALUES:
+            stat_idx = int(stat_enum)
+            if stats_array[stat_idx] > 0:
+                stat_name = _get_enum_name(stat_enum)
+                if stat_name not in current_dict:
```
poodlewars, line 47: Invert the if statement to avoid double negative

python/arcticdb/toolbox/query_stats.py
poodlewars, line 37: I would find a comment here with an example showing an example of a `stats_array` and the layers quite useful

python/arcticdb/toolbox/query_stats.py
```diff
@@ -0,0 +1,89 @@
...
+        next_layer_maps = layer.next_layer_maps
+        for group_enum in cls._STATS_GROUP_NAME_VALUES:
+            group_idx = int(group_enum)
+            
+            if not next_layer_maps[group_idx]:
+                continue
+                
+            next_layer_map = next_layer_maps[group_idx]
+            
+            # top level
```
poodlewars, line 62: I think you should have a check in this function that the `arcticdb_call` is indeed at the top level of the stats object we're processing

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,160 @@
...
+ * - StatsGroup: RAII wrapper that temporarily extends the layer chain during its lifetime
+ *   When created, it adds a new layer and when destroyed, it restores the previous layer state
+ * 
+ * Note:
+ * To make the query stats model works, there are two requirements:
+ * 1. All calls from python layer must mark "QUERY_STATS_ADD_GROUP...." at least once in the call stack
+ *    so the stats logged in folly threads will be aggregated to the master map
+ *    (Checking will be added after all log entries are added)
+ * 2. All folly tasks must be submitted through the TaskScheduler::submit_cpu_task/submit_io_task
+ * 3. All folly tasks must complete ("collected") before last StatsGroup object is destroyed in the call stack
```
poodlewars, line 59: What happens if a task fails? Should add testing in your C++ test for that

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,160 @@
...
+ *                                                                                                      
+ * Structure:
+ * - QueryStats: Singleton manager class holding the stats collection framework
+ * - ThreadLocalQueryStatsVar: Thread-local storage for stats layers and management
+ * - StatsGroupLayer: Hierarchical node containing stats and references to child layers
+ * - StatsGroup: RAII wrapper that temporarily extends the layer chain during its lifetime
+ *   When created, it adds a new layer and when destroyed, it restores the previous layer state
+ * 
+ * Note:
+ * To make the query stats model works, there are two requirements:
```
poodlewars, line 54: work not works
three

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,160 @@
...
+    ~StatsGroup();
+private:
+    std::shared_ptr<StatsGroupLayer> prev_layer_;
+    std::chrono::time_point<std::chrono::high_resolution_clock> start_;
+    bool log_time_;
+};
+
+}
+
+#define STATS_GROUP_VAR_NAME(x) query_stats_info##x
```
poodlewars, line 145: I still don't think there's any need for these to be macros rather than normal functions? I get that you'd have to hold the `StatsGroup` alive for the RAII to work, but that should be fine.

python/tests/integration/toolbox/test_query_stats.py
poodlewars, line 1: Does `list_symbols` release the GIL? As soon as you instrument a function that does, should add tests to check how this all behaves with Python multi-threading

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -165,21 +169,53 @@ inline auto get_default_num_cpus([[maybe_unused]] const std::string& cgroup_fold
     #endif
 }
 
-/*
```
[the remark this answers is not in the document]
phoebusm: That's a mistake

cpp/arcticdb/toolbox/python_bindings.cpp
```diff
@@ -148,6 +150,52 @@ void register_bindings(py::module &m, py::exception<arcticdb::ArcticException>&
...
+        .def(py::init<>())
+        .def_readwrite("stats", &StatsGroupLayer::stats_)
+        .def_property_readonly("next_layer_maps", [](const StatsGroupLayer& self) {
+            return self.next_layer_maps_;
+        });
+    
+    query_stats_module.def("current_layer", []() { 
+        return QueryStats::instance().current_layer(); 
+    });
+    query_stats_module.def("root_layers", []() {
```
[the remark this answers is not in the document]
phoebusm: Yea I want to move the formatting to python level as much as I can.

cpp/arcticdb/util/query_stats.hpp
phoebusm: It will still work. All I need is the task to complete, failed or not, to avoid race condition
It will add a test to prove

cpp/arcticdb/util/query_stats.hpp
phoebusm: Yea I am referencing existing implementation, e.g. `ARCTICDB_SAMPLE`.
Declaring a variable is a bit weird for logging. And folding the ON/OFF into `StatsGroup` class's constructor/destructor is also weird IMO
And I want to make adding groupable stat and non-groupable stat unifrom as well, from the user's perspective

python/arcticdb/toolbox/query_stats.py
phoebusm: Good point. But I rather add the checking at the C++ layer

python/tests/integration/toolbox/test_query_stats.py
phoebusm: It doesn't release the GIL. 
I simulate the python multi-threading in the C++ test `Async.StatsQueryDemo`

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -166,20 +170,60 @@ inline auto get_default_num_cpus([[maybe_unused]] const std::string& cgroup_fold
...
+* Possible areas of inprovement in the future:
+* 1/ Task/op decoupling: push task and then use strategy to implement smart batching to
+* amortize costs wherever possible
+* 2/ Worker thread Affinity - would better locality improve throughput by keeping hot structure in
+* hot cachelines and not jumping from one thread to the next (assuming thread/core affinity in hw too) ?
+* 3/ Priority: How to assign priorities to task in order to treat the most pressing first.
+* 4/ Throttling: (similar to priority) how to absorb work spikes and apply memory backpressure
+*/
+template <typename T>
+class ExecutorWithStatsInstance : public T{
```
willdealtry, line 182: I have mixed feeling about attaching the stats to the task scheduler. If you look at the existing timings framework that allows a dump to stdout or a Remotery connection, which is covering many of the same areas that this one is, it is timing logical stages in the execution of a query, rather than the activities of the scheduler. That's because the way that things are scheduled is an implementation detail that can and does change - some logical activities are composed of multiple scheduling steps, some only one, and some things that we might want to time aren't scheduled at all but are executed inline in the main thread. I would hate to see a day where we were asynchronously scheduling things that didn't benefit from it just because we wanted to time them.

The scheduler is also the place where it's possible to have the worst impact in terms of imposing cost on people who don't want to use the query stats. As long as the access to the global variable  is moved outside the scheduling mutex, this doesn't look too bad, however we'll need to be sure - we can use the existing timings mechanism to see the overhead of the additional work.

In this case it doesn't matter particularly, since some additional cost is to be expected when the query stats is enabled, but traversing a tree of shared_ptr is not an inherently fast design - I would not want to see it in a feature that was going to be enabled by default. I think it stems in part from the fact that the framework is very open-ended, perhaps more so that we need. Ultimately in a database there are a limited number of things that we want to time, so a set number of (perhaps) thread-cached counters that could be manually added to as required would probably suffice, and be more performant.

In terms of the thread-caching, the specific async framework we use will schedule work almost (from the point of view of the stats framework) randomly across different threads, as it maintains a thread-pool and dispatched tasks to an available or lightly-loaded thread - is this something that you've taken into account?

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,185 @@
...
+    std::chrono::time_point<std::chrono::high_resolution_clock> start_;
+    bool log_time_;
+};
+
+std::string format_group_value(GroupName col_value, auto&& value) {
+    constexpr bool is_value_key_type = std::is_same_v<std::remove_cv_t<std::remove_reference_t<decltype(value)>>, arcticdb::entity::KeyType>;
+    check(
+        col_value != util::query_stats::GroupName::key_type || is_value_key_type, 
+        "key type query stats needs to have key_type value"
+    );
```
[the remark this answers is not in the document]
phoebusm: Good reminder. I have thought about it but didn't implement it

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,185 @@
...
+};
+
+enum class StatsName : size_t {
+    result_count = 0,
+    total_time_ms = 1,
+    count = 2
+};
+
+struct GroupingLevel {
+    std::array<int64_t, 3> stats_ = {0}; // sizeof(StatsName)
```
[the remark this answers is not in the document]
phoebusm: I think the current way is slightly neater for the conversion of from the internal map to python json format as it allows the copy of the values by iterating the array (no boiler plate code) and the enum that comes with the current design is handy for the naming of node in json

cpp/arcticdb/util/query_stats.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include <arcticdb/util/preconditions.hpp>
+#include <arcticdb/log/log.hpp>
+
+namespace arcticdb::util::query_stats {
+
+std::shared_ptr<GroupingLevel> QueryStats::current_level(){
+    // current_level_ != nullptr && root_level_ != nullptr -> stats has been setup; Nothing to do
+    // current_level_ == nullptr && root_level_ == nullptr -> clean slate; Need to setup
+    // current_level_ != nullptr && root_level_ == nullptr -> Something is off
+    // current_level_ == nullptr && root_level_ != nullptr -> Something is off
```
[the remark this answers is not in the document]
phoebusm: It's checked below (line 22)

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -194,19 +238,53 @@ class TaskScheduler {
         auto task = std::forward<decltype(t)>(t);
```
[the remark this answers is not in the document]
phoebusm: Thanks. The query stats ON performance will be addressed a bit later in a separate ticket

phoebusm:
More simple design has been preferred so this PR has been closed after some discussion

PR #2212, opened 2025-02-28
Fix regression when writing combination of np.NaN and None

vasil-pashov:
Early version of ArcticDB allowed for writing a column containing only none and nan values. We assumed both of those are placeholders for an empty string. At some point mixing only np.nan and None in a column stopped working on write side.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -163,29 +161,34 @@ NativeTensor obj_to_tensor(PyObject *ptr, bool empty_types) {
...
                 }
                 if(current_object != end)
                     sample = *current_object;
             }
-            if (empty && kind == 'O') {
+            // TODO: If empty types are brought back to life consider merging empty_string_placeholder and empty into a
+            // single thing as they have similar meaning. This would mean that [None, np.nan] would be considered empty
+            if (empty_string_placeholder) {
+                val_type = ValueType::UTF_DYNAMIC;
+            } else if (empty && kind == 'O') {
```
alexowens90, line 191: Is `empty_string_placeholder` really necessary? The only case it seems to solve (`dtype == None`) seems like it would also be solved by dropping the `&& kind == 'O'` from this line?

vasil-pashov: `empy` is badly named it should have been `all_nones`. It is also not needed. I though it might break empty types  but it appears we're still running tests for the empty types. What we need to check if it's an empty string placeholder and we also need to check the kind/dtype. If a column is consisted only np.nan we need to check if the dtype is object or a numeric dtype.

alexowens90 (approved):

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -145,14 +145,10 @@ NativeTensor obj_to_tensor(PyObject *ptr, bool empty_types) {
...
 
-        // If Numpy has type 'O' then get_value_type above will return type 'BYTES'
-        // If there is no value, and we can't deduce a type then leave it that way,
-        // otherwise try to work out whether it was a bytes (string) type or unicode
         if (!is_fixed_string_type(val_type) && element_count > 0) {
             auto none = py::none{};
             auto obj = reinterpret_cast<PyObject **>(arr->data);
-            bool empty = false;
-            bool all_nans = false;
+            bool empty_string_placeholder = false;
```
poodlewars, line 151: Maybe `array_has_empty_string` is a clearer name?

python/tests/unit/arcticdb/version_store/test_write.py
```diff
@@ -132,4 +133,26 @@ def test_write_non_timestamp_index(lmdb_version_store, index_type, sorted, valid
     info = lib.get_info(symbol)
     assert info["sorted"] == "UNKNOWN"
 
+@pytest.mark.parametrize("dtype", [None, object])
```
poodlewars, line 136: Parameterizing with `dtype=None` means we use the Pandas inferred type right? It might be clearer (to me at least) to state the type explicitly (eg `np.float64` for this test).

PR #2217, opened 2025-03-05
Resampling with dynamic schema

vasil-pashov:
#### Reference Issues/PRs
This is a rework of https://github.com/man-group/ArcticDB/pull/2201

This enables using the QueryBuilder to resample libraries where dynamic schema is enabled. This requires handling two cases which are different in comparison to static schema.

1. Type propagation. It turns out that it works without any functional changes in the code.
2. Missing columns. Dynamic schema allows for segments to contain only a subset of all columns in the symbol. This case needed additional care.

*Handling missing columns*
The aggregate method takes as an input an array of `std::optional` columns and empty optional marks a missing column. The first step of the aggregation is to determine the common type (which handles type propagation), `generate_common_input_type` skips missing columns and computes the common type for all available columns. The result is a `std::optional` type. If the optional is empty this means that all columns were skipped (because they were missing), at this point the aggregate function can terminate returning empty std::optional for the resulting column. The resampling clause will ignore empty optional columns returned by aggregate.

**Missing behavior**
The current implementation has a flaws related to using date_range

1. If `date_range` is passed to the QueryBuilder and the date range does not contain some column (which is otherwise in the data frame) we will return either an empty dataframe (if no other columns are being aggregated) or a dataframe not containing that column. More sensible behavior would be to backfill the missing buckets with a default value based on the dtype and the aggregator.
2. Type propagation takes into account only the segments using in the aggregation and the dtypes from the segment descriptor. This means that different date ranges can result in dataframes whose columns have different dtypes. More sensible behavior would be to use the timeseries descriptor.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/processing/clause.cpp
+++ b/cpp/arcticdb/processing/clause.cpp
@@ -774,8 +774,11 @@ std::vector<EntityId> ResampleClause<closed_boundary>::process(std::vector<Entit
                                 }
             );
         }
-        auto aggregated_column = std::make_shared<Column>(aggregator.aggregate(input_index_columns, input_agg_columns, bucket_boundaries, *output_index_column, string_pool));
-        seg.add_column(scalar_field(aggregated_column->type().data_type(), aggregator.get_output_column_name().value), aggregated_column);
+        std::optional<Column> aggregated_column = aggregator.aggregate(input_index_columns, input_agg_columns, bucket_boundaries, *output_index_column, string_pool);
+        if (aggregated_column) {
+            auto aggregated_column_ptr = std::make_shared<Column>(std::move(*aggregated_column));
+            seg.add_column(scalar_field(aggregated_column_ptr->type().data_type(), aggregator.get_output_column_name().value), std::move(aggregated_column_ptr));
+        }
     }
     seg.set_row_data(output_index_column->row_count() - 1);
     return push_entities(*component_manager_, ProcessingUnit(std::move(seg), std::move(output_row_range)));
--- a/python/tests/unit/arcticdb/version_store/test_resample.py
+++ b/python/tests/unit/arcticdb/version_store/test_resample.py
@@ -6,6 +6,8 @@ Use of this software is governed by the Business Source License 1.1 included in 
 As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
 """
 from functools import partial
+from typing import Union, Optional
+
 import numpy as np
 import pandas as pd
 import datetime as dt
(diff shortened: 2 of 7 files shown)


cpp/arcticdb/processing/clause.cpp
alexowens90, line 1: Comment on line 666 is now outdated, and reserve on line 657 is less useful

python/tests/unit/arcticdb/version_store/test_resample.py
```diff
@@ -17,14 +19,90 @@
...
+    """
+    Get the default value for an aggregation operation. This is used to back-fill Pandas dataframes when they are
+    compared against dynamic schema libraries. Dynamic schema allows for missing columns but Pandas does not. ArcticDB
+    will put these default values everywhere where we have an index value in a bucket but the segment does not contain
+    column data.
+    """
+
+    assert aggregation in ALL_AGGREGATIONS
+    if is_float_dtype(dtype):
+        return 0 if aggregation == "count" else np.nan
```
alexowens90, line 77: Doesn't `sum` also default to zero?

alexowens90, line 77: Doesn't `sum` default to 0 for floats as well?

python/tests/unit/arcticdb/version_store/test_resample.py
```diff
@@ -843,4 +903,116 @@ def test_min_with_one_infinity_element(lmdb_version_store_v1):
...
+        expected_slice_with_missing_column = pd.DataFrame({
+            f"to_{agg}": [default_aggregation_value(agg, dtype) for _ in expected_slice_with_missing_column_idx] for agg in all_aggregations(dtype)
+        }, index=expected_slice_with_missing_column_idx)
+        expected_slice_containing_column = df_to_append.resample('10ns').agg(None, **all_aggregations_dict("b", dtype))
+        expected = pd.concat([expected_slice_with_missing_column, expected_slice_containing_column])
+        expected = expected.reindex(columns=sorted(expected.columns))
+        assert_frame_equal(arctic_resampled, expected, check_dtype=False)
+
+    @pytest.mark.parametrize("dtype", [np.float32, np.float64, np.uint32, np.int32, np.int64, bool, "datetime64[ns]", str])
+    def test_bucket_spans_two_segments(self, lmdb_version_store_dynamic_schema_v1, dtype):
```
alexowens90, line 941: Also worth having a test where there are 3 row-slices in a bucket, with the first and last containing the aggregation column, but the middle one missing it

cpp/arcticdb/processing/clause.cpp
alexowens90, line 1: Are both branches of `responsible_for_first_overlapping_bucket` covered in testing?

PR #2218, opened 2025-03-05
Fix and test arcticdb reading streaming data

IvoDD:
#### Reference Issues/PRs
Fixes monday ref: 7855342201

Fixes:
- Column filter in static schema
- Column ordering when introducing a new column with an incomplete segment

Tests:
- Columns filter in static and dynamic schema
- Reading diffrent schema incompletes
- Compatibility test for reading incompletes from an old env

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -1085,9 +1085,17 @@ bool read_incompletes_to_pipeline(
     // Mark the start point of the incompletes, so we know that there is no column slicing after this point
     pipeline_context->incompletes_after_ = pipeline_context->slice_and_keys_.size();
 
-    // If there are only incompletes we need to add the index here
     if(pipeline_context->slice_and_keys_.empty()) {
+        // If there are only incompletes we need to do the following (typically done when reading the index key):
+        // - add the index columns to query
+        // - in case of static schema: populate the descriptor and column_bitset
         add_index_columns_to_query(read_query, seg.index_descriptor());
+        if (!dynamic_schema) {
+            pipeline_context->desc_ = seg.descriptor();
+            get_column_bitset_in_context(
+                read_query,
+                pipeline_context);
+        }
     }
     pipeline_context->slice_and_keys_.insert(std::end(pipeline_context->slice_and_keys_), incomplete_segments.begin(), incomplete_segments.end());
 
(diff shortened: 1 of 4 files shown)


python/tests/compat/arcticdb/test_compatibility.py
```diff
@@ -171,3 +172,39 @@ def test_compat_snapshot_metadata_read(old_venv_and_arctic_uri, lib_name):
...
+lib._nvs.append("sym", df_2, incomplete=True)
+            """
+        ], dfs={"df_1": df_1, "df_2": df_2})
+
+
+    with CurrentVersion(arctic_uri, lib_name) as curr:
+        read_df = curr.lib._nvs.read(sym, date_range=(None, None), incomplete=True).data
+        assert_frame_equal(read_df, df)
+
+        read_df = curr.lib._nvs.read(sym, date_range=(None, None), incomplete=True, columns=["float_col"]).data
```
poodlewars, line 209: Would be good to check that filtering down to several columns (rather than just one) works, and that the `date_range` filtering works

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1116,9 +1124,9 @@ bool read_incompletes_to_pipeline(
         pipeline_context->staged_descriptor_ =
             merge_descriptors(seg.descriptor(), incomplete_segments, read_query.columns);
         if (pipeline_context->desc_) {
-            const std::array fields_ptr = {pipeline_context->desc_->fields_ptr()};
+            const std::array staged_fields_ptr = {pipeline_context->staged_descriptor_->fields_ptr()};
```
poodlewars, line 1127: Could you explain why we need this change?

python/tests/unit/arcticdb/version_store/test_incompletes.py
```diff
@@ -80,3 +81,47 @@ def test_read_incompletes_no_chunking(lmdb_version_store_tiny_segment):
 
     ref_keys = lib_tool.find_keys_for_symbol(KeyType.APPEND_REF, sym)
     assert len(ref_keys) == 1
+
+@pytest.mark.parametrize("dynamic_schema", [True, False])
+def test_read_incompletes_columns_filter(version_store_factory, dynamic_schema):
```
poodlewars, line 86: A few ideas for extra tests:

- Tests where we append incompletes after a compaction (like you do in the test below)
- Tests with a chain of more than two incomplete segments (eg I could imagine messing up the linked list traversal)
- Filtering down to more than one column
- Filters on `date_range`

cpp/arcticdb/version/version_core.cpp
IvoDD: So that if we call `merge_descriptors(descriptor_from_index_key, new_columns_from_staged_data)` instead of the other way round.

This way if we have staged data which is missing columns it won't completely reorder the columns. Previously if e.g. we had index with columns `col_1, col_2, col_3` and had an incomplete with `col_2, col_3` reading them both would result in `col_2, col_3, col_1`, which seems quite odd. The dynamic schema tests this

python/tests/compat/arcticdb/test_compatibility.py
IvoDD: Added more assertions to the test

python/tests/unit/arcticdb/version_store/test_incompletes.py
IvoDD: Added more assertions to the two tests

poodlewars (approved):

PR #2220, opened 2025-03-07
Release the GIL during parallel writes to avoid deadlock

poodlewars:
Normal writes release the GIL, but until this change staged writes did not.

We recently (v5.2.0) changed staged writes to behave more like normal writes, in particular they now use `WriteToSegmentTask` which needs to take the GIL to allocate data backed by Python unicode strings. This lead to a deadlock.

This change simply releases the GIL in the "main" thread when we start a staged write, just like we already do for a normal write.

I will make a subsequent change to add more test coverage based on a real example, and to check how we handle unicode across our other API methods.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_write.py
+++ b/python/tests/unit/arcticdb/version_store/test_write.py
@@ -8,8 +8,9 @@ As of the Change Date specified in that file, in accordance with the Business So
 import numpy as np
 import pandas as pd
 import pytest
-from arcticdb.exceptions import SortingException, NormalizationException
+from arcticdb.exceptions import SortingException
 from arcticdb.util._versions import IS_PANDAS_TWO
+from arcticdb.util.test import assert_frame_equal
 from pandas import MultiIndex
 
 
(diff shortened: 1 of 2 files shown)


python/tests/unit/arcticdb/version_store/test_write.py
```diff
@@ -133,3 +134,28 @@ def test_write_non_timestamp_index(lmdb_version_store, index_type, sorted, valid
...
+
+def test_write_parallel_unicode(lmdb_version_store):
+    symbol = "test_write_parallel_unicode"
+    uc = "\u0420\u043e\u0441\u0441\u0438\u044f"
+
+    df1 = pd.DataFrame(
+        index=[pd.Timestamp("2018-01-02"), pd.Timestamp("2018-01-03")],
+        data={"a": ["123", uc]},
+    )
+    lmdb_version_store.write(symbol, df1, parallel=True)
```
poodlewars, line 158: This would deadlock before this fix

willdealtry (approved):

IvoDD (approved):
We probably need more complex unicode strings tests, with mixtures of reads/writes/staged writes/updates etc. Not for this PR.

G-D-Petrov (approved):

PR #2224, opened 2025-03-07
Testing for unicode handling across our API

poodlewars:
This moves our various tests for unicode handling in to a single file `test_unicode.py` and adds coverage for our other API methods where I can imagine unicode possibly making a difference, as well as testing for unicode column names and unicode index values (in a multi-index).

I also made a cross cutting testing change to `random_strings_of_length` which is used by many tests so that on average a third of the strings it generates include a unicode character. I kept it returning some pure ASCII strings too so we don't lose coverage of that case either.

### DIFF
--- a/python/arcticdb/util/test.py
+++ b/python/arcticdb/util/test.py
@@ -244,8 +244,16 @@ def assert_frame_equal_rebuild_index_first(expected: pd.DataFrame, actual: pd.Da
     assert_frame_equal(left=expected, right=actual)
 
 
+unicode_symbol = "\u00A0"  # start of latin extensions
+unicode_symbols = "".join([chr(ord(unicode_symbol) + i) for i in range(100)])
+
+
 def random_string(length: int):
-    return "".join(random.choice(string.ascii_uppercase + string.digits) for _ in range(length))
+    if random.randint(0, 3) == 0:
+        # (probably) Give a unicode string one time in three, we have special handling in C++ for unicode
+        return "".join(random.choice(string.ascii_uppercase + unicode_symbols) for _ in range(length))
+    else:
+        return "".join(random.choice(string.ascii_uppercase + string.digits) for _ in range(length))
 
 
 def get_sample_dataframe(size=1000, seed=0, str_size=10):
@@ -433,7 +441,15 @@ def get_pickle():
     )[np.random.randint(0, 2)]
 
 
-def random_strings_of_length(num, length, unique):
+def random_ascii_strings(count, max_length):
+    result = []
+    for _ in range(count):
+        length = random.randrange(max_length + 1)
+        result.append("".join(random.choice(string.ascii_letters) for _ in range(length)))
+    return result
+
+
+def random_strings_of_length(num, length, unique=False):
     out = []
     for i in range(num):
         out.append(random_string(length))
--- a/python/tests/unit/arcticdb/version_store/test_unicode.py
+++ b/python/tests/unit/arcticdb/version_store/test_unicode.py
@@ -1,0 +1,343 @@
+"""
+Copyright 2025 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+
+We have special handling in the codebase when working with unicode Python strings, since we need to take the GIL
+to handle them. This file checks that our APIs work even when passed unicode string."""
+import datetime
+
+import numpy as np
+import pandas as pd
+import pytest
+
+from arcticdb.util.test import assert_frame_equal, random_strings_of_length
+
+from arcticdb.version_store.library import Library
+from arcticdb.version_store.library import UpdatePayload
+from arcticdb_ext.storage import NoDataFoundException
+
+unicode_str = "\u0420\u043e\u0441\u0441\u0438\u044f"
+copyright = "My Thing Not Your's \u00A9"
+trademark = "My Word Not Your's \u2122"
+metadata = {copyright: trademark}
+symbol = "sym"
+
+
+def unicode_strs_df(start_date: pd.Timestamp, num_rows: int) -> pd.DataFrame:
+    index = [start_date + datetime.timedelta(days=i) for i in range(num_rows)]
+    df = pd.DataFrame(
+        index=index,
+        data={"a": random_strings_of_length(num_rows, 10), trademark: np.arange(num_rows), copyright: [unicode_str] * num_rows},
+    )
+    return df
+
+
+@pytest.mark.parametrize("parallel", (True, False))
+@pytest.mark.parametrize("multi_index", (True, False))
+def test_write(lmdb_version_store_tiny_segment, parallel, multi_index):
+    lib = lmdb_version_store_tiny_segment
+    start = pd.Timestamp("2018-01-02")
+    num_rows = 100
+    if multi_index:
+        index = pd.MultiIndex.from_arrays([[start + datetime.timedelta(days=i) for i in range(num_rows)], [unicode_str] * num_rows])
+    else:
+        index = pd.date_range(start=start, periods=num_rows)
+
+    df = pd.DataFrame(
+        index=index,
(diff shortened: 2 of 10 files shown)


IvoDD (approved):

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+@pytest.mark.parametrize("parallel", (True, False))
+@pytest.mark.parametrize("multi_index", (True, False))
+def test_write(lmdb_version_store_tiny_segment, parallel, multi_index):
+    lib = lmdb_version_store_tiny_segment
+    start = pd.Timestamp("2018-01-02")
+    num_rows = 100
+    if multi_index:
+        index = pd.MultiIndex.from_arrays([[start + datetime.timedelta(days=i) for i in range(num_rows)], [unicode_str] * num_rows])
+    else:
+        index = [start + datetime.timedelta(days=i) for i in range(num_rows)]
```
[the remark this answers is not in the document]
poodlewars: Fair, I always find that a bit hard to read as you have to "just now" the size of the period that it will use

python/arcticdb/util/test.py
```diff
@@ -244,8 +244,16 @@ def assert_frame_equal_rebuild_index_first(expected: pd.DataFrame, actual: pd.Da
...
 
 
+unicode_symbol = "\u00A0"  # start of latin extensions
+unicode_symbols = "".join([chr(ord(unicode_symbol) + i) for i in range(100)])
+
+
 def random_string(length: int):
-    return "".join(random.choice(string.ascii_uppercase + string.digits) for _ in range(length))
+    if random.randint(0, 3) == 0:
+        # Give a unicode string one time in three, we have special handling in C++ for unicode
```
alexowens90, line 253: nit: it has a chance of giving a unicode string, might still pick ascii

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
+"""We have special handling in the codebase when working with unicode Python strings, since we need to take the GIL
```
alexowens90, line 1: Missing copyright

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+        data={"a": random_strings_of_length(num_rows, 10), trademark: np.arange(num_rows), copyright: [unicode_str] * num_rows},
+    )
+
+    if parallel:
+        lib.write(symbol, df, parallel=True)
+        lib.compact_incomplete(symbol, append=False, convert_int_to_float=False, metadata=metadata)
+    else:
+        lib.write(symbol, df, metadata=metadata)
+
+    lib.create_column_stats(symbol, column_stats={trademark: {"MINMAX"}})
```
alexowens90, line 53: No need to drop this, but for reference, there's no special GIL handling for column stats, it's all C++ layer and string type agnostic

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+    vit = lib.read(symbol)
+    assert_frame_equal(vit.data, df)
+    assert vit.metadata == metadata
+
+
+def test_write_incompletes(lmdb_version_store_tiny_segment):
```
alexowens90, line 59: This is already covered by the test above?

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+    assert_frame_equal(vit.data, df1)
+
+    expected = [df1]
+    for i in range(100):
+        start = pd.Timestamp("2018-01-02") + datetime.timedelta(days=(i + 1) * 2)
+        df = pd.DataFrame(
+            index=[start, start + datetime.timedelta(days=1)],
+            data={copyright: ["123", unicode_str]},
+        )
+        lib.write(symbol, df, parallel=True)
```
alexowens90, line 157: It's the same code path now, but could use `append` with `incomplete=True`

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+    df1 = pd.DataFrame(
+        index=[pd.Timestamp("2018-01-02"), pd.Timestamp("2018-01-03")],
+        data={copyright: ["123", unicode_str]},
+    )
+    lib.write(symbol, df1)
+    vit = lib.read(symbol)
+    assert_frame_equal(vit.data, df1)
+
+    expected = [df1]
+    for i in range(100):
```
alexowens90, line 151: 100 seems excessive

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+    vit = lib.read(symbol)
+    assert_frame_equal(vit.data, df1)
+
+    df2 = unicode_strs_df(start + datetime.timedelta(days=100), 100)
+    lib.update(symbol, df2)
+    vit = lib.read(symbol)
+    expected = pd.concat([df1, df2])
+    assert_frame_equal(vit.data, expected)
+
+    df1_new = unicode_strs_df(start, 100)
```
alexowens90, line 180: If you use `start + datetime.timedelta(days=1)` here then there will be segments that need to be split so it's a slightly better test

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+        index=[pd.Timestamp("2018-01-04"), pd.Timestamp("2018-01-05")],
+        data={copyright: ["123", unicode_str]},
+    )
+    df3 = pd.DataFrame(
+        index=[pd.Timestamp("2018-02-04"), pd.Timestamp("2018-02-05")],
+        data={copyright: ["123", unicode_str]},
+    )
+
+    update_payloads = [UpdatePayload(sym_1, df2), UpdatePayload(sym_2, df3)]
+    adb_lib.update_batch(update_payloads)
```
alexowens90, line 210: This is basically an append, better test if the dates overlap

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -0,0 +1,234 @@
...
+    start = pd.Timestamp("2018-01-02")
+    index = [start + datetime.timedelta(days=i) for i in range(4)]
+
+    df = pd.DataFrame(
+        index=index,
+        data={"a": ["123", unicode_str, copyright, trademark], trademark: [1, 2, 3, 4], copyright: [unicode_str] * 4},
+    )
+
+    lmdb_version_store.write(symbol, df, metadata=metadata)
+    lmdb_version_store.snapshot(copyright)
```
alexowens90, line 230: If we support unicode in snapshot names then presumably we also do in symbol names?

alexowens90: Would this work on S3?

python/tests/unit/arcticdb/version_store/test_unicode.py
poodlewars: I don't think our testing should always be tied to how things are implemented right now

python/arcticdb/util/test.py
poodlewars: I'll update the comment, but since there are 100 unicode symbols and 27 ASCII symbols in the bag, the change of only getting ASCII with length 5 is less than 1 in 1000

python/tests/unit/arcticdb/version_store/test_unicode.py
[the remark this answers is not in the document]
poodlewars: I'll update with this change

python/tests/unit/arcticdb/version_store/test_unicode.py
poodlewars: We don't in symbol names (deliberately). There are bugs where we do allow them in snapshot names, and in fact for new symbols written with `compact_incomplete`. I've raised tickets to fix those and added tests for snapshots with unicode names in the meantime.

alexowens90 (approved):

PR #2232, opened 2025-03-12
Accept strings for the mode when finalizing staged data

ognyanstoimenov:
Monday ref 8645841959
Support strings for finalize staged data `mode` argument in addition to enum
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -1281,7 +1281,7 @@ class Library:
     def finalize_staged_data(
         self,
         symbol: str,
-        mode: Optional[StagedDataFinalizeMethod] = StagedDataFinalizeMethod.WRITE,
+        mode: Optional[Union[StagedDataFinalizeMethod, str]] = StagedDataFinalizeMethod.WRITE,
         prune_previous_versions: bool = False,
         metadata: Any = None,
         validate_index = True,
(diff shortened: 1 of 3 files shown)


python/tests/unit/arcticdb/version_store/test_api.py
```diff
@@ -89,3 +91,38 @@ def test_get_num_rows_pickled(lmdb_version_store):
...
+        ("WRITE", False),
+        ("APPEND", True),
+        (None, False),
+        (StagedDataFinalizeMethod.APPEND, True),
+        (StagedDataFinalizeMethod.WRITE, False),
+        ([], None),
+        {"write", None},
+    ],
+)
+def test_finalize_staged_data(arctic_library_lmdb, input_mode, expected_append):
```
poodlewars, line 108: This is a good unit test but I think we should also have some integration tests that check we are actually doing the right thing.

The `expected_append==None` case should probably be split to its own test, rather than parameters in this one.

python/arcticdb/version_store/library.py
```diff
@@ -1310,9 +1310,9 @@ def finalize_staged_data(
         symbol : `str`
             Symbol to finalize data for.
 
-        mode : `StagedDataFinalizeMethod`, default=StagedDataFinalizeMethod.WRITE
-            Finalize mode. Valid options are WRITE or APPEND. Write collects the staged data and writes them to a
-            new version. Append collects the staged data and appends them to the latest version.
+        mode : Union[`StagedDataFinalizeMethod`, str], default=StagedDataFinalizeMethod.WRITE
```
poodlewars, line 1313: I think lowercase `write, append` will match both Pandas and our `QueryBuilder` APIs better.

python/arcticdb/version_store/library.py
```diff
@@ -1383,9 +1383,12 @@ def finalize_staged_data(
         2024-01-03    3
         2024-01-04    4
         """
+        if not (mode is None or isinstance(mode, StagedDataFinalizeMethod) or mode == "WRITE" or mode == "APPEND"):
```
poodlewars, line 1386: Should validate against `mode==None` shouldn't you, what does it mean?

poodlewars, line 1386: `if not isinstance(mode, StagedDataFinalizeMethod) or mode not in ("write", "append")` is clearer to me, as it's closer to English

poodlewars, line 1386: Should validate that the enum value is either `WRITE` or `APPEND` as protection against someone changing the enum and us silently treating it as `WRITE`

poodlewars (changes requested):

python/arcticdb/version_store/library.py
ognyanstoimenov: I don't get what you mean here, I am validating if mode is None?

poodlewars (approved):

IvoDD (approved):

PR #2233, opened 2025-03-12
Enhancement/8561507350/precompute output schema from processing

alexowens90:
#### Reference Issues/PRs
[Closes 8561507350](https://man312219.monday.com/boards/7852509418/pulses/8561507350)

Adds a method to clauses to produce the correct `StreamDescriptor` and `NormalizationMetadata` given the same types as input, or throw an exception if the clause is not compatible with the input columns/types.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/processing/expression_node.hpp
```diff
@@ -86,6 +91,23 @@ struct ExpressionNode {
     ExpressionNode(VariantNode left, OperationType op);
 
     VariantData compute(ProcessingUnit& seg) const;
+
+    std::variant<BitSetTag, DataType> compute(
```
IvoDD, line 95: Sorry I'm not super familiar with this part of the code but what is the difference in principal between `BitSetTag` and `DataType::BOOL8`?

E.g. if we have `q = [q["bool_col"]]`, the expression used is just `ExpressionNode{ColumnName{"bool_col"}, IDENTITY}`, correct?
I think `compute` now will return a `DataType::BOOL8`, but for Filter clauses we require a `BitSetTag` to be returned [here](https://github.com/man-group/ArcticDB/blob/8ed95600d302b9f49273c5542a19763fd2bb54b1/cpp/arcticdb/processing/clause.cpp#L162).

Can we only return `DataType` where `BitSet` is inferred from `BOOL8`? Or at least we should change the assert in `FilterClause`?

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -70,4 +70,171 @@ VariantData ExpressionNode::compute(ProcessingUnit& seg) const {
...
+                if (right_value_set_state == ValueSetState::NON_EMPTY_SET) {
+                    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+                            (is_sequence_type(std::get<DataType>(left_type)) && is_sequence_type(std::get<DataType>(right_type))) || (is_numeric_type(std::get<DataType>(left_type)) && is_numeric_type(std::get<DataType>(right_type))),
+                            "Incompatible data types provided in set membership operator {}, {}",
+                            std::get<DataType>(left_type), std::get<DataType>(right_type));
+                } // else - Empty value set compatible with all data types
+                break;
+            case OperationType::AND:
+            case OperationType::OR:
+            case OperationType::XOR:
```
IvoDD, line 194: Related to my other comment about difference between `BitSetTag` and `DataType::BOOL8`.

Shouldn't `AND`, `OR`, `XOR`, `NOT` operations return `DataType::BOOL8`, not a `BitSetTag`?

E.g. if we have nested `(something AND (NOT something_else))` the `NOT` clause should either return a bool or the `AND` clause should accept a `BitSetTag`

cpp/arcticdb/processing/expression_node.hpp
IvoDD: Actually the assertion in `FilterClause` is correct because `Indentity` will convert it to `BitSet`. Still it's a bit confusing when something should return a bool and when a bitset (as per other comment)

cpp/arcticdb/processing/clause.cpp
```diff
@@ -170,6 +200,16 @@ std::vector<EntityId> ProjectClause::process(std::vector<EntityId>&& entity_ids)
     return output;
 }
 
+OutputSchema ProjectClause::modify_schema(OutputSchema&& output_schema) const {
+    check_column_presence(output_schema, *clause_info_.input_columns_, "Project");
+    auto root_expr = expression_context_->expression_nodes_.get_value(expression_context_->root_node_name_.value);
+    std::variant<BitSetTag, DataType> return_type = root_expr->compute(*expression_context_, output_schema.column_types());
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(std::holds_alternative<DataType>(return_type), "ProjectClause AST produces a column, not a bitset");
+    output_schema.stream_descriptor_.add_scalar_field(std::get<DataType>(return_type), output_column_);
+    output_schema.column_types().emplace(output_column_, std::get<DataType>(return_type));
```
IvoDD, line 209: Should we have a method `OutputSchema::add_field` to take responsibility of keeping things consistent?

cpp/arcticdb/processing/clause.cpp
```diff
@@ -432,6 +472,31 @@ std::vector<EntityId> AggregationClause::process(std::vector<EntityId>&& entity_
...
+        const auto& output_column_name = agg.get_output_column_name().value;
+        const auto& input_column_type = output_schema.stream_descriptor_.field(*output_schema.stream_descriptor_.find_field(input_column_name)).type().data_type();
+        auto agg_data = agg.get_aggregator_data();
+        agg_data.add_data_type(input_column_type);
+        const auto& output_column_type = agg_data.get_output_data_type();
+        stream_desc.add_scalar_field(output_column_type, output_column_name);
+    }
+
+    output_schema.stream_descriptor_ = std::move(stream_desc);
+    output_schema.clear_column_types();
```
IvoDD, line 492: Also similar to above `OutputSchema::set_stream_descriptor` can clear the `column_types` cache?

cpp/arcticdb/processing/test/test_output_schema.cpp
```diff
@@ -0,0 +1,1004 @@
...
+    ec = ExpressionContext();
+    ec.root_node_name_ = ExpressionName("new col");
+    ec.add_expression_node("new col", node);
+    project_clause = ProjectClause{{"col4", "col5"}, "new col", ec};
+    EXPECT_THROW(project_clause.modify_schema({stream_desc.clone(), norm_meta}), SchemaException);
+}
+
+// Type promotion rules are tested exhaustively in test_arithmetic_type_promotion.cpp, so we can just exercise the major
+// code paths once each here
+class AstParsingOutputTypesTest : public testing::Test {
```
IvoDD, line 173: Also maybe one or two deeper nested exseptions to cover `x And (y Or z)` from other comment.

cpp/arcticdb/processing/clause.cpp
```diff
@@ -133,6 +155,14 @@ std::vector<EntityId> FilterClause::process(std::vector<EntityId>&& entity_ids)
     return output;
 }
 
+OutputSchema FilterClause::modify_schema(OutputSchema&& output_schema) const {
+    check_column_presence(output_schema, *clause_info_.input_columns_, "Filter");
+    auto root_expr = expression_context_->expression_nodes_.get_value(expression_context_->root_node_name_.value);
+    std::variant<BitSetTag, DataType> return_type = root_expr->compute(*expression_context_, output_schema.column_types());
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(std::holds_alternative<BitSetTag>(return_type), "FilterClause AST produces a column, not a bitset");
```
willdealtry, line 162: do you mean "produced"?

cpp/arcticdb/processing/expression_node.cpp
willdealtry: Internally the result of a predicate operation is always a bitset and never a BOOL. The user can't get bools out of it, just filtered types of whatever column the predicate is being applied to, so I think a type BOOL is confusing here.

willdealtry (approved):

cpp/arcticdb/processing/clause.cpp
alexowens90: Yeh good idea, I added `column_types_` after everything else so the interface has got a bit confused

cpp/arcticdb/processing/clause.cpp
alexowens90: Agreed as above

cpp/arcticdb/processing/clause.cpp
alexowens90: I guess "would produce" is most accurate

cpp/arcticdb/processing/clause.cpp
```diff
@@ -463,6 +528,45 @@ void ResampleClause<closed_boundary>::set_component_manager(std::shared_ptr<Comp
     component_manager_ = std::move(component_manager);
 }
 
+template<ResampleBoundary closed_boundary>
+OutputSchema ResampleClause<closed_boundary>::modify_schema(OutputSchema&& output_schema) const {
```
vasil-pashov, line 532: One thing that's not clear to me is when the type promotion happens for resampling. In case we have multiple segments with different types are we going to know the output type prior reading all segments.

cpp/arcticdb/entity/stream_descriptor.hpp
```diff
@@ -271,6 +273,33 @@ struct StreamDescriptor {
...
+            }
+        }
+        return *column_types_;
+    }
+
+    void clear_column_types() {
+        column_types_ = std::nullopt;
+    }
+private:
+    std::optional<ankerl::unordered_dense::map<std::string, DataType>> column_types_;
```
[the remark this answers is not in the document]
alexowens90: In theory yes, although in the (admittedly niche) case where you have no columns, you would repeatedly see that the map is empty and regenerate it, after which it would remain empty.
Semantically though, I think an optional makes it clearer if it has been generated or not

cpp/arcticdb/processing/clause.cpp
```diff
@@ -133,6 +155,14 @@ std::vector<EntityId> FilterClause::process(std::vector<EntityId>&& entity_ids)
     return output;
 }
 
+OutputSchema FilterClause::modify_schema(OutputSchema&& output_schema) const {
```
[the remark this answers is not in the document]
alexowens90: Some clauses (e.g. projections) do something that looks like modifying the existing schema (adding one field).
Others (e.g. resampling) completely throw away the existing schema and create a new one from scratch, so I think this approach makes it clearer that the returned value is not necessarily closely related to the input value.

cpp/arcticdb/processing/clause.cpp
alexowens90: The `StreamDescriptor` in the input `OutputSchema` object is from the `TimeSeriesDescriptor` for the whole symbol-version pair. When dynamic schema is enabled and the types are changing for a given column, by definition, the type of the column in the `TimeSeriesDescriptor` is capable of representing all of the types of the individual row-slices of that column.

vasil-pashov (approved):

IvoDD (approved):

PR #2235, opened 2025-03-12
Fix convert_int_to_float in compact_incomplete V1 library API method

vasil-pashov:
This fixes the behavior of `convert_int_to_float` exposed via the V1 Library API's `compact_incomplete` method.

Convert int to float will change the type of all integer columns (both signed and unsigned) to float64. In contrast to dynamic schema conversion this will static cast the values of all segments to float64 prior writing. This functionality was broken when we made the stricter type checks for staged segments.

Additional bugfix was added for dynamic schema. Dynamic schema has a different code path as it calls merge descriptors when it collects all incomplete segments thus the option was not properly applied even in arcticc. Dynamic schema used to throw exception when unsigned and signed int were mixed together (in the same column) but this should not matter as they are both going to be cast to double.

There are two main points:
* columns_match was changed to take convert int to float as a parameter
* merge_descriptors was changed to take convert int to float as a parameter
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/entity/merge_descriptors.cpp
+++ b/cpp/arcticdb/entity/merge_descriptors.cpp
@@ -11,11 +11,15 @@
 #include <arcticdb/entity/type_utils.hpp>
 
 namespace arcticdb {
+/// @param convert_int_to_float This will convert all integer types (both signed and unsigned) to FLOAT64 without
+///   performing any type checks. This is available only via the V1 Library API and is used by tick collectors. It can
+///   be true only if descriptors are merged during a compact_incomplete call. Otherwise, it must be false.
 StreamDescriptor merge_descriptors(
     const StreamDescriptor &original,
     std::span<const std::shared_ptr<FieldCollection>> entries,
-    const std::unordered_set<std::string_view> &filtered_set,
-    const std::optional<IndexDescriptorImpl>& default_index) {
+    const std::unordered_set<std::string_view>& filtered_set,
+    const std::optional<IndexDescriptorImpl>& default_index,
+    bool convert_int_to_float) {
     using namespace arcticdb::stream;
     std::vector<std::string_view> merged_fields;
     std::unordered_map<std::string_view, TypeDescriptor> merged_fields_map;
@@ -55,10 +59,10 @@ StreamDescriptor merge_descriptors(
             );
         }
 
-        for (size_t idx = has_index ? 1u : 0u; idx < static_cast<size_t>(fields->size()); ++idx) {
+        for (size_t idx = has_index ? 1u : 0u; idx < fields->size(); ++idx) {
             const auto& field = fields->at(idx);
             const auto& type_desc = field.type();
-            if (filtered_set.empty() || (filtered_set.find(field.name()) != filtered_set.end())) {
+            if (!filtered_set.contains(field.name())) {
                 if(auto existing = merged_fields_map.find(field.name()); existing != merged_fields_map.end()) {
                     auto existing_type_desc = existing->second;
                     if(existing_type_desc != type_desc) {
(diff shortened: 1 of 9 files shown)


IvoDD (approved):
Your tests pretty much cover all cases with `convert_int_to_float=True`.

What do you think about also adding tests that we get schema exseptions if `convert_int_to_float=False`? (e.g. to ensure we don't accidentally pass `convert_int_to_float=True` somewhere deep in C++ where we shouldn't). It is possible we already have a bunch of these tests somewhere else, haven't looked through them

cpp/arcticdb/entity/merge_descriptors.cpp
```diff
@@ -68,17 +72,22 @@ StreamDescriptor merge_descriptors(
                                 "New type descriptor                     : {}",
                                 field.name(), existing_type_desc, type_desc
                         );
-                        auto new_descriptor = has_valid_common_type(existing_type_desc, type_desc);
-                        if(new_descriptor) {
-                            merged_fields_map[field.name()] = *new_descriptor;
+                        if (convert_int_to_float && is_integer_type(existing_type_desc.data_type()) && is_integer_type(type_desc.data_type())) {
```
alexowens90, line 75: What if the existing type is float64? Or float32?

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -1593,3 +1593,68 @@ def test_staging_in_chunks_default_settings(lmdb_storage, lib_name):
...
+        df1 = pd.DataFrame({"a": np.array([1, 2, 3], dtype=dtype)}, index=pd.date_range(pd.Timestamp(0), periods=3, freq="ns"))
+        df2 = pd.DataFrame({"a": np.array([4, 5, 6], dtype=dtype)}, index=pd.date_range(pd.Timestamp(3), periods=3, freq="ns"))
+        lib.write(sym, df1, parallel=True)
+        lib.write(sym, df2, parallel=True)
+        lib.compact_incomplete(sym, append=False, convert_int_to_float=True)
+        expected = pd.DataFrame({"a": np.arange(1, 7, dtype=np.double)}, index=pd.date_range(pd.Timestamp(0), periods=6, freq="ns"))
+        assert_frame_equal(expected, lib.read(sym).data, check_dtype=True)
+
+    @pytest.mark.parametrize("dtype1", [np.int32, np.uint16, np.int8, np.int64, np.uint64, np.float64])
+    @pytest.mark.parametrize("dtype2", [np.int32, np.uint16, np.int8, np.int64, np.uint64, np.float64])
```
alexowens90, line 1612: Include float32 in the tested dtypes?

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -1593,3 +1593,68 @@ def test_staging_in_chunks_default_settings(lmdb_storage, lib_name):
...
+        sym = "sym"
+        df1 = pd.DataFrame({"a": np.array([1, 2, 3], dtype=dtype1)}, index=pd.date_range(pd.Timestamp(0), periods=3, freq="ns"))
+        df2 = pd.DataFrame({"a": np.array([4, 5, 6], dtype=dtype2)}, index=pd.date_range(pd.Timestamp(3), periods=3, freq="ns"))
+        lib.write(sym, df1, parallel=True)
+        lib.write(sym, df2, parallel=True)
+        lib.compact_incomplete(sym, append=False, convert_int_to_float=True)
+        expected = pd.DataFrame({"a": np.arange(1, 7, dtype=np.double)}, index=pd.date_range(pd.Timestamp(0), periods=6, freq="ns"))
+        assert_frame_equal(expected, lib.read(sym).data, check_dtype=True)
+
+    @pytest.mark.parametrize("dtype2", [np.int32, np.uint16, np.int8, np.int64, np.uint64, np.float64])
```
alexowens90, line 1625: Include float32?

alexowens90 (approved):

PR #2239, opened 2025-03-12
Fix test_chunks_the_same flakiness

poodlewars:
Fix for flaky test test_chunks_the_same
    
This would fail sometimes on Windows / Mac because the clocks used there are less precise and the two [1000, 1001] segments would collide on exactly the same key with LMDB. I just change the test so their content hashes would be different. The test was written to cover an issue with identical index ranges, so it still meets its original purpose.
    
This just changes the tests so that their content hashes differ. Theoretically, this could be a real problem that comes up, and we ought to somehow retry when there is a key clash. But that seems very improbable outside of unit testing, so we leave it as is for now.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_parallel.py
+++ b/python/tests/unit/arcticdb/version_store/test_parallel.py
@@ -1531,17 +1531,17 @@ def test_chunks_match_at_ends(lmdb_storage, lib_name):
     assert result["a"][-1] == 6
 
 
-def test_chunks_the_same(lmdb_storage, lib_name):
+@pytest.mark.parametrize("n_runs", range(10))
+def test_chunks_the_same(lmdb_storage, lib_name, n_runs):
     """Given - we stage chunks with indexes:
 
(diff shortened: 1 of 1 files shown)


python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -1531,17 +1531,17 @@ def test_chunks_match_at_ends(lmdb_storage, lib_name):
     assert result["a"][-1] == 6
 
 
-def test_chunks_the_same(lmdb_storage, lib_name):
+@pytest.mark.parametrize("n_runs", range(10))
```
IvoDD, line 1534: Remove before merging?

IvoDD (approved):

python/tests/unit/arcticdb/version_store/test_parallel.py
poodlewars: I think it's good to keep as there is some non-determinism

PR #2248, opened 2025-03-14
Get object sizes based on S3's ListObjects output (just with scan_object_sizes() to start with)

poodlewars:
Monday: 8560764974

Limited to just `scan_object_sizes` to start with, to show the idea.

Calculate object sizes using functionality in the storage backend itself when possible. This PR calculates compressed sizes on S3 based on the `ListObjectsV2` output, so we don't need to read all the keys.

For storages where we haven't implemented anything fancy, just read all the (compressed) keys and check the compressed size on their header.

Remaining work:

- Use this approach for `scan_object_sizes_by_stream`
- Make sure we have testing against all backends especially NFS
- Implement `scan_object_sizes_for_stream`
- Native size calculations for LMDB and Azure
- Native size calculation for library size (LMDB has an API to do this "in one")
- `AdminTools` Python API on the v2 API, including search by regex. Docs page to explain the output.

### DIFF
--- a/cpp/arcticdb/storage/s3/s3_client_impl.cpp
+++ b/cpp/arcticdb/storage/s3/s3_client_impl.cpp
@@ -299,11 +299,13 @@ S3Result<ListObjectsOutput> S3ClientImpl::list_objects(
         next_continuation_token = {result.GetNextContinuationToken()};
 
     auto s3_object_names = std::vector<std::string>();
+    auto s3_object_sizes = std::vector<uint64_t>();
     for (const auto &s3_object: result.GetContents()) {
         s3_object_names.emplace_back(s3_object.GetKey());
+        s3_object_sizes.emplace_back(s3_object.GetSize());
     }
 
-    ListObjectsOutput output = {s3_object_names, next_continuation_token};
+    ListObjectsOutput output = {s3_object_names, s3_object_sizes, next_continuation_token};
     return {output};
 }
 
--- a/cpp/arcticdb/storage/s3/s3_client_impl.cpp
+++ b/cpp/arcticdb/storage/s3/s3_client_impl.cpp
@@ -305,8 +305,7 @@ S3Result<ListObjectsOutput> S3ClientImpl::list_objects(
         s3_object_sizes.emplace_back(s3_object.GetSize());
     }
 
-    ListObjectsOutput output = {s3_object_names, s3_object_sizes, next_continuation_token};
-    return {output};
+    return {ListObjectsOutput{std::move(s3_object_names), std::move(s3_object_sizes), next_continuation_token}};
 }
 
 }
--- a/cpp/arcticdb/storage/storage.hpp
+++ b/cpp/arcticdb/storage/storage.hpp
@@ -19,6 +19,21 @@
 
 namespace arcticdb::storage {
 
+struct ObjectSizes {
+    ObjectSizes(KeyType key_type, size_t count, size_t compressed_size_bytes) :
+        key_type_(key_type), count_(count), compressed_size_bytes_(compressed_size_bytes) {
+
+    }
+
+    ObjectSizes(KeyType key_type) : ObjectSizes(key_type, 0, 0) {
+
+    }
+
+    KeyType key_type_;
+    size_t count_;
+    size_t compressed_size_bytes_;
+};
+
 enum class SupportsAtomicWrites {
     NO,
     YES,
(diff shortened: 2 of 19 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1701,56 +1701,15 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
+        sizes.push_back(store->get_object_sizes(key_type, ""));
     });
 
-    std::unordered_map<KeyType, KeySizesInfo> result;
-    for (const auto& [k, v] : sizes) {
-        result[k] = KeySizesInfo{v.count, v.compressed_size, v.uncompressed_size};
-    }
-    return result;
+    folly::QueuedImmediateExecutor inline_executor;
+    return folly::collect(sizes_futs).via(&inline_executor).get();
```
alexowens90, line 1712: Is this how we're meant to be collecting in general?

cpp/arcticdb/storage/storages.hpp
```diff
@@ -184,6 +188,16 @@ class Storages {
         }
     }
 
+    ObjectSizes get_object_sizes(KeyType key_type, const std::string& prefix) {
+        ObjectSizes res{key_type, 0, 0};
+        for (const auto& storage : storages_) {
```
alexowens90, line 193: I see what you're going for here, but AFAIK none of our other methods check anything other than the first storage

cpp/arcticdb/storage/storage.hpp
```diff
@@ -218,3 +249,19 @@ class Storage {
...
+namespace fmt {
+using namespace arcticdb::storage;
+
+template<> struct formatter<ObjectSizes> {
+
+    template<typename ParseContext>
+    constexpr auto parse(ParseContext &ctx) { return ctx.begin(); }
+
+    template<typename FormatContext>
+    auto format(const ObjectSizes &srv, FormatContext &ctx) const {
```
alexowens90, line 262: srv?

cpp/arcticdb/stream/stream_source.hpp
```diff
@@ -78,6 +85,22 @@ struct StreamSource {
...
+        if (calculators.empty()) {
+            return;
+        }
+
+        std::vector<folly::Future<folly::Unit>> res;
+        for (auto&& fut : batch_read_compressed(std::move(calculators), BatchReadArgs{})) {
+            // Ignore some exceptions, someone might be deleting while we scan
+            res.push_back(std::move(fut)
+                              .thenValue([](auto&&) {return folly::Unit{};})
+                              .thenError(folly::tag_t<storage::KeyNotFoundException>{}, [](auto&&) { return folly::Unit{}; }));
```
alexowens90, line 99: Can use `collectAll` for this

cpp/arcticdb/storage/s3/s3_client_impl.cpp
```diff
@@ -299,11 +299,13 @@ S3Result<ListObjectsOutput> S3ClientImpl::list_objects(
...
 
     auto s3_object_names = std::vector<std::string>();
+    auto s3_object_sizes = std::vector<uint64_t>();
     for (const auto &s3_object: result.GetContents()) {
         s3_object_names.emplace_back(s3_object.GetKey());
+        s3_object_sizes.emplace_back(s3_object.GetSize());
     }
 
-    ListObjectsOutput output = {s3_object_names, next_continuation_token};
+    ListObjectsOutput output = {s3_object_names, s3_object_sizes, next_continuation_token};
```
alexowens90, line 308: Looks like first 2 arguments should be moved in

alexowens90 (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: I think it makes more sense, rather than worrying about which executors the futures themselves use and whether deadlocks are possible. Happy to change to the normal idiom and add some tests with single thread IO and CPU pools if you feel strongly though.

cpp/arcticdb/storage/storages.hpp
poodlewars: I'll add a check that there is exactly one storage, as there are a few different semantics you might want for this with multiple storages, and doesn't seem worth guessing while the multiple storages thing is theoretical.

There's a case for ripping out this multiple storages stuff until we implement it properly, as I imagine most of the implementations here would change when we do that anyway.

cpp/arcticdb/storage/storage.hpp
poodlewars: Cut and paste fail

cpp/arcticdb/stream/stream_source.hpp
poodlewars: I'm not sure it would simplify things - we could `collectAll` on the `batch_read_compressed` result but we would still need a for loop to chain the continuations on to the results. And short-circuiting if there is an un-expected exception seems desirable.

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: Actually I'll add those tests anyway.

willdealtry: I don't think there actually is any risk of deadlock with the CPU and IO executors. This pattern seems fine, but I can't see a reason to go around changing all the existing usages - we should settle on one pattern and use it everywhere

cpp/arcticdb/storage/storages.hpp
willdealtry: Why not make it obey the primary_only as the other read methods do? This problem has been thought about and solved already

PR #2249, opened 2025-03-17
Asv v2 s3 tests (Refactored)

grusev:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/arcticdb/util/environment_setup.py
```diff
@@ -1,82 +1,83 @@
...
+    MODIFIABLE = "MODIFIABLE_LIBRARIES"
+    TEST = "TESTS_LIBRARIES"
+
+
+class LibraryType(Enum):
+    PERSISTENT = "PERMANENT"
+    MODIFIABLE = "MODIFIABLE"
+
+
+class StorageSetup:
```
G-D-Petrov, line 88: The StorageSetup class can easily be refactored to be more readable like this:
``` python
def aws_default_factory() -> BaseS3StorageFixtureFactory:
    return real_s3_from_environment_variables(shared_path=True)

def get_machine_id() -> str:
    """
    Returns machine id, or id specified through environments variable (for github)
    """
    return os.getenv("ARCTICDB_PERSISTENT_STORAGE_SHARED_PATH_PREFIX", socket.gethostname())


def create_prefix(storage_space: StorageSpace, add_to_prefix: str) -> str:
    def is_valid_string(s: str) -> bool:
        return bool(s and s.strip())

    mandatory_part = storage_space.value
    optional = add_to_prefix if is_valid_string(add_to_prefix) else ''
    return f"{mandatory_part}/{optional}" if optional else mandatory_part


def check_persistence_access(storage_space: StorageSpace, confirm_persistent_storage_need: bool = False):
    assert aws_default_factory(), "Environment variables not initialized (ARCTICDB_REAL_S3_ACCESS_KEY,ARCTICDB_REAL_S3_SECRET_KEY)"
    if storage_space == StorageSpace.PERSISTENT:
        assert confirm_persistent_storage_need, "Use of persistent store not confirmed!"


def get_arctic_uri(storage: Storage, storage_space: StorageSpace, add_to_prefix: str = None, confirm_persistent_storage_need: bool = False) -> str:
    check_persistence_access(storage_space, confirm_persistent_storage_need)
    prefix = create_prefix(storage_space, add_to_prefix)
    if storage == Storage.AMAZON:
        factory = aws_default_factory()
        factory.default_prefix = prefix
        return factory.create_fixture().arctic_uri
    elif storage == Storage.LMDB:
        return f"lmdb://{tempfile.gettempdir()}/benchmarks_{prefix}"
    else:
        raise Exception("Unsupported storage type:", storage)
````

File: python/arcticdb/util/environment_setup.py [hunk 2]
```diff
@@ -99,491 +100,369 @@ def __new__(cls, *args, **kwargs):
...
+            arctic_url = StorageSetup._get_aws_s3_arctic_uri(storage_space, add_to_prefix, confirm_persistent_storage_need)
+        elif storage == Storage.LMDB:
+            arctic_url = StorageSetup._get_lmdb_arctic_uri(storage_space, add_to_prefix, confirm_persistent_storage_need)
+        else:
+            raise Exception("Unsupported storage type :", storage)
+        return arctic_url
 
 
-class StorageInfo:    
+class LibraryManager:
```
G-D-Petrov: As we spoke yesterday, this `LibraryManager` is unnecessary because it duplicates a lot of the logic of Arctic's internal LibraryManager.
The only needed functionality is:
- having a function to create the persistent/modifiable Arctic client with the correct URIs
- having a function that constructs the correct names for the libraries
- some helper function for cleaning up modifiable libraries, which can just iterate over the libraries in the modifiable Arctic client   

Every thing else here is can easily be handled though the arctic clients directly

python/benchmarks/real_modification_functions.py
```diff
@@ -0,0 +1,292 @@
...
+from arcticdb.util.utils import TimestampNumber
+from arcticdb.version_store.library import Library
+from benchmarks.common import AsvBase
+
+
+#region Setup classes
+
+WIDE_DATAFRAME_NUM_COLS = 30_000
+
+class LargeAppendDataModifyCache:
```
G-D-Petrov, line 23: This name is not very descriptive  and the comments seems a bit misleading.
AFAICS this is a cache for the expected results thought the run.
The name/comment should reflect that.

python/benchmarks/real_modification_functions.py
```diff
@@ -0,0 +1,292 @@
...
+
+    number_columns = 30
+
+    def get_library_manager(self) -> LibraryManager:
+        return AWSLargeAppendTests.library_manager
+    
+    def get_population_policy(self):
+        pass
+    
+    def get_index_info(self):
```
G-D-Petrov, line 64: should be renamed to something like `index_start`, then it doesn't need the comment

python/benchmarks/real_modification_functions.py
```diff
@@ -0,0 +1,292 @@
...
+    def setup_cache(self):
+        return self.initialize_cache(AWSLargeAppendTests.warmup_time,
+                                     AWSLargeAppendTests.params,
+                                     AWSLargeAppendTests.number_columns,
+                                     AWSLargeAppendTests.number)
+    
+    def initialize_cache(self, warmup_time, params, num_cols, num_sequential_dataframes):
+        # warmup will execute tests additional time and we do not want that at all for write
+        # update and append tests. We want exact specified `number` of times to be executed between
+        assert warmup_time == 0, "warm up must be 0"
```
G-D-Petrov, line 79: If it can be only 1 value, why do we even have it as a parameter?

python/benchmarks/real_modification_functions.py
```diff
@@ -0,0 +1,292 @@
...
+
+        num_sequential_dataframes = num_sequential_dataframes + 1
```
G-D-Petrov, line 81: num_sequential_dataframes += 1

python/benchmarks/real_modification_functions.py
```diff
@@ -0,0 +1,292 @@
...
+            self.initialize_update_dataframes(num_rows=num_rows, num_cols=num_cols, cached_results=cache, 
+                                              generator=generator)
+    
+        self.get_library_manager().clear_all_modifiable_libs()
+
+        self.get_library_manager().log_info()
+
+        return cache
+    
+    def initialize_update_dataframes(self, num_rows: int, num_cols: int, cached_results: LargeAppendDataModifyCache,
```
G-D-Petrov, line 101: This function is a bit hard to follow, consider refactoring it to something like:
``` python
def initialize_update_dataframes(self, num_rows: int, num_cols: int, cached_results: LargeAppendDataModifyCache, 
                                 generator: SequentialDataframesGenerator):
    logger = self.get_logger()
    initial_timestamp, freq = self.get_index_info()
    timestamp_number = TimestampNumber.from_timestamp(initial_timestamp, freq)
    
    def log_time_range(update_type: str, df_key: int):
        time_range = generator.get_first_and_last_timestamp([cached_results[update_type][df_key]])
        logger.info(f"Time range {update_type.upper()} update {time_range}")

    def generate_and_log(update_type: str, num_rows: int, start_ts: pd.Timestamp):
        df = generator.df_generator.get_dataframe(number_rows=num_rows, number_columns=num_cols, start_timestamp=start_ts, freq=freq)
        cached_results[update_type][num_rows] = df
        log_time_range(update_type, num_rows)

    logger.info(f"Frame START-LAST Timestamps {timestamp_number} == {timestamp_number + num_rows}")

    # Full update
    generate_and_log('update_full_dict', num_rows, initial_timestamp)

    # Half update
    half = num_rows // 2
    timestamp_number.inc(half - 3)
    generate_and_log('update_half_dict', half, timestamp_number.to_timestamp())

    # Upsert update
    generate_and_log('update_upsert_dict', num_rows, timestamp_number.to_timestamp())

    # Single update
    timestamp_number.inc(half)
    generate_and_log('update_single_dict', 1, timestamp_number.to_timestamp())

    # Single append
    next_timestamp = generator.get_next_timestamp_number(cached_results.write_and_append_dict[num_rows], freq)
    generate_and_log('append_single_dict', 1, next_timestamp.to_timestamp())
```

G-D-Petrov (changes requested):

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +111,387 @@ def __new__(cls, *args, **kwargs):
...
-    def get_library(self, library_suffix: Union[str, int] = None) -> Library:
-        """
-        Returns one time setup library (permanent)
-        """
-        return self._get_lib(self.get_arctic_client_persistent(), 
-                             self.get_library_names(library_suffix)[0])
-    
-    def get_modifiable_library(self, library_suffix: Union[str, int] = None) -> Library:
+
+class LibraryPopulationPolicy:
```
G-D-Petrov, line 336: As we have discussed, this can be greatly simplified by decoupling the configuration for the population from the logic of executing the populating.
This can be done with a  refactor like:
``` python
@dataclass
class LibraryPopulationConfig:
    """Immutable configuration for library population."""
    parameters: List[int]
    parameters_are_rows: bool = True
    fixed_rows: int = 1
    fixed_columns: int = 1
    symbol_prefix: str = ""
    use_auto_increment: bool = False
    with_metadata: bool = False
    versions_count: int = 1
    versions_mean: float = 1.0
    with_snapshots: bool = False

    def symbol_name(self, index: int) -> str:
        """Get the symbol name based on configuration."""
        prefix = f"symbol_{self.symbol_prefix}_" if self.symbol_prefix else "symbol_"
        return f"{prefix}{index}"
    
    def create_metadata(self) -> Dict[str, Any]:
        """Create metadata for symbols and snapshots."""
        if not self.with_metadata:
            return {}
        return DFGenerator.generate_random_dataframe(rows=3, cols=10).to_dict()


class LibraryPopulator:
    """
    Handles the actual population of a library based on a configuration.
    Separates the configuration from the execution.
    """
    def __init__(self, config: LibraryPopulationConfig, logger: logging.Logger, 
                 df_generator: DataFrameGenerator = None):
        self.config = config
        self.logger = logger
        self.df_generator = df_generator or VariableSizeDataframe()
    
    def populate(self, library):
        """Populate the library according to the configuration."""
        start_time = time.time()
        
        for i, param in enumerate(self.config.parameters):
            # Determine symbol index
            symbol_index = i if self.config.use_auto_increment else param
            symbol_name = self.config.symbol_name(symbol_index)
            
            # Determine rows and columns
            rows = param if self.config.parameters_are_rows else self.config.fixed_rows
            columns = self.config.fixed_columns if self.config.parameters_are_rows else param
            
            # Generate dataframe
            df = self.df_generator.generate_dataframe(rows, columns)
            
            # Create symbol
            symbol = library.create_symbol(symbol_name, df)
            
            # Add metadata if configured
            if self.config.with_metadata:
                symbol.set_metadata(self.config.create_metadata())
            
            # Create versions if configured
            if self.config.versions_count > 1:
                versions_list = self._generate_versions_list(len(self.config.parameters))
                for v in range(1, min(versions_list[i], self.config.versions_count) + 1):
                    version_df = self.df_generator.generate_dataframe(rows, columns)
                    version = symbol.create_version(version_df)
                    
                    # Add metadata if configured
                    if self.config.with_metadata:
                        version.set_metadata(self.config.create_metadata())
                    
                    # Create snapshot if configured
                    if self.config.with_snapshots:
                        snapshot = library.create_snapshot(f"snapshot_{symbol_name}_{v}")
                        if self.config.with_metadata:
                            snapshot.set_metadata(self.config.create_metadata())
        
        self.logger.info(f"Population completed in: {time.time() - start_time:.2f}s")
    
    def _generate_versions_list(self, number_symbols: int) -> List[np.int64]:
        """Generate a list of version counts for each symbol."""
        # Implementation would depend on your specific requirements
        # This is a placeholder based on the original code
        versions_list = np.random.poisson(self.config.versions_mean, number_symbols)
        versions_list = np.clip(versions_list, 1, self.config.versions_count)
        return versions_list.astype(np.int64)
```
The code is just a example that I got from a pass though Claude and can be simplified further e.g:
- there are parameter in the Policy that can be removed
- we probably don't need a LibraryPopulator class, as some helper functions that take a policy should suffice

python/arcticdb/util/environment_setup.py
grusev: I do not agree. There is a value of separating responsibility and working TestLibraryManager.

It provides better isolation and management. So I disagree with making those changes

python/arcticdb/util/environment_setup.py
grusev: BTW, It is TestLibrary manager now

python/benchmarks/real_modification_functions.py
grusev: The parameters of the tests are defined as such in each test case. Thus test case A parameters are memebers of class A not to the instance. Test case B parameters are for test case B. The function checks parameters of different classes

class AWSLargeAppendTests(AsvBase)
class AWS30kColsWideDFLargeAppendTests(AWSLargeAppendTests)

python/benchmarks/real_modification_functions.py
grusev: but it is also correct in current form right?

python/benchmarks/real_modification_functions.py
grusev: makes sence!

python/arcticdb/util/environment_setup.py
grusev: As we discussed we also agreed that current code is not conflicting any recommendations. Additionally there is no real benefit of making that rewrite other than to pass Claude's suggestions. The code is just ok in the form that is now. No need to changed it

python/benchmarks/real_modification_functions.py
grusev: I think the naming is ok. It returns both start and the index frequence - hence the name get_index_info(), also the comment there suggests that:

    def get_index_info(self):
        """
        Returns initial timestamp and index frequency
        """
        return (pd.Timestamp("2-2-1986"), 's')

python/arcticdb/util/environment_setup.py
IvoDD: I don't have a strong opinion whether these live inside a class or as separate functions. I think in this case the class only provides namespacing, which can also be done with a module but I don't think it matters hugely.

I think @G-D-Petrov 's suggestion provides a few simplifications which I find valueable:
- The main purpose of this being a class seems to be the caching of `_aws_default_factory`. I don't think creating it is expensive? Why bother caching it then? This way we need to worry about forgetting to initialize it the first time. I.e. we can just replace the `__new__` with a `aws_default_factory`.
- The `create_prefix` is rewritten in a much more concise way.

To sum up, I don't mind leaving this a class for the namespacing but it would be nice to still simplify the logic we can

python/arcticdb/util/environment_setup.py
```diff
@@ -1,94 +1,126 @@
...
     '''
     Defined special one time setup for real storages.
     Place here what is needed for proper initialization
     of each storage
+
+    Abstracts storage space allocation from how user access it
     '''
     _instance = None
     _aws_default_factory: BaseS3StorageFixtureFactory = None
     _fixture_cache = {}
```
IvoDD, line 119: `fixture_cache` is now unused?

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +131,421 @@ def __new__(cls, *args, **kwargs):
...
-            if result is None:
-                result = type_prefix
-            else:
-                result = f"{result}_{type_prefix}"
-        if self.__storage_test_env_prefix:
-                result = f"{result}_{self.__storage_test_env_prefix}"
-        return result
-
-    def set_test_mode(self):
+    def __init__(self, storage: Storage, name_benchmark: str, library_options: LibraryOptions = None) :
```
IvoDD, line 231: Should all libraries have the same `LibraryOptions`?

Doesn't it make more sense to pass `LibraryOptions` to `get_library`? This way we can create two libraries with different options which might be useful for some tests.

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +131,421 @@ def __new__(cls, *args, **kwargs):
...
+            self._get_arctic_client_modifiable() # Forces uri generation
+        mes = f"{self} arcticdb URI information for this test: \n"
+        for key in self._ac_cache.keys():
+            mes += f"arcticdb URI: {key}"
+        logger.info(mes) 
+
+    # Currently we're using the same arctic client for both persistant and modifiable libraries.
+    # We might decide that we want different arctic clients (e.g. different buckets) but probably not needed for now.
+    def _get_arctic_client_persistent(self) -> Arctic:
+        lib_type = StorageSpace.PERSISTENT
```
IvoDD, line 255: Isn't this `storage_type` not `lib_type`?

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +131,421 @@ def __new__(cls, *args, **kwargs):
...
-        """
-        Returns library to read write and delete after done.
-        """
-        return self._get_lib(self.get_arctic_client_modifiable(), 
-                             self.get_library_names(library_suffix)[1])
+    def clear_all_modifiable_libs_from_this_process(self):
+        ac = self._get_arctic_client_modifiable()
+        lib_names = set(ac.list_libraries())
+        to_deletes = [lib_name for lib_name in lib_names 
+                      if (f"_{os.getpid()}_" in lib_name) and (f"_{self.name_benchmark}_" in lib_name)]
```
IvoDD, line 308: Why not just the simpler `lib_name.startswith(f"{LibraryType.Modifiable}_{self.name_benchmark}_{os.getpid()}_")`?

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +131,421 @@ def __new__(cls, *args, **kwargs):
...
+        for to_delete in to_deletes:
+            ac.delete_library(to_delete)
 
-    def delete_modifiable_library(self, library_suffix: Union[str, int] = None):
+    def clear_all_benchmark_libs(self):
```
IvoDD, line 312: All these `clear` functions can use a common `clear_modifiable_with_prefix(self, prefix)`
And `clear_all_modifiable_from_this_process` can pass the prefix with the pid wheras `clear_all_benchmark_libs` can pass the prefix without the pid.

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +131,421 @@ def __new__(cls, *args, **kwargs):
...
-        return self._params
 
-    def is_multi_params_list(self):
-        """
-        Checks if the parameters are list of 2+ parameter lists
-        """
-        return all(isinstance(i, list) for i in self._params)
+# It is quite clear what is this responsible for: only dataframe generation
+# Using such an abstraction can help us deduplicate the dataframe generation code between the different `EnvironmentSetup`s
+# Note: We use a class instead of a generator function to allow caching of dataframes in the state
```
IvoDD, line 354: This doc is a leftover from the refactor proposal. We can probably just remove it, from the name it is clear enough what this does.

python/benchmarks/real_comparison_benchmarks.py
```diff
@@ -0,0 +1,200 @@
...
+    timeout = 60000
+
+    LIB_NAME = "COMPARISON"
+    URL = "lmdb://compare"
+    SYMBOL = "dataframe"
+    # NOTE: Change of parameters will trigger failure as original library must also be deleted manually.
+    #       Therefore if you plan changes to those numbers make sure to delete old library manually 
+    NUMBER_ROWS = 2_000_000 #100_000
+
+    params = [NO_OPERATION, CREATE_DATAFRAME, PANDAS_PARQUET, ARCTICDB_LMDB, ARCTICDB_AMAZON_S3]
```
IvoDD, line 51: What is the purpose of `NO_OPERATION`?

python/benchmarks/real_comparison_benchmarks.py
```diff
@@ -0,0 +1,200 @@
...
+            "period" : str_col(20, size), 
+            "observation_type" : str_col(10, size),
+            "ric" : str_col(10, size),
+            "dtype" : str_col(10, size),
+        }       
+
+    def peakmem_read_dataframe(self, tpl, btype):
+        df, dict = tpl
+        if btype == NO_OPERATION:
+            # What is the tool mem load?
```
IvoDD, line 144: Remove this comment?

python/benchmarks/real_comparison_benchmarks.py
```diff
@@ -0,0 +1,200 @@
...
+            return
+        if btype == CREATE_DATAFRAME:
+            df = pd.DataFrame(dict)
+        elif btype == PANDAS_PARQUET:
+            pd.read_parquet(self.parquet_to_read )
+        elif btype == ARCTICDB_LMDB:
+            self.lib.read(self.SYMBOL)
+        elif btype == ARCTICDB_AMAZON_S3:
+            self.s3_lib_read.read(self.s3_symbol) 
+            pass
```
IvoDD, line 154: Unneeded pass

python/benchmarks/real_list_operations.py
```diff
@@ -6,59 +6,81 @@
...
-        self.lib = self.aws.get_library(num_syms)
-
-    def time_list_symbols(self, storage_info, num_syms):
+        num_rows = AWSListSymbols.number_rows
+        manager = self.get_library_manager()
+        assert AWSListSymbols.number == 1, "There must be always one test between setup and tear down"
+        policy = self.get_population_policy()
+        for number_symbols in AWSListSymbols.params:
+            start = time.time()
+            policy.set_parameters([num_rows] * number_symbols)
```
IvoDD, line 51: Would it be better if `get_population_policy` took an argument `num_symbols` and constructs it in place instead of modifying the same instance many times?

python/benchmarks/real_list_operations.py
```diff
@@ -6,59 +6,81 @@
...
+            if not manager.has_library(AWSListSymbols.library_type, number_symbols):
+                populate_library(manager, policy, AWSListSymbols.library_type, number_symbols)
+                self.get_logger().info(f"Generated {number_symbols} with {num_rows} each for {time.time()- start}")
+            else:
+                self.get_logger().info(f"Library already exists, population skipped")
+        manager.log_info() # Always log the ArcticURIs 
+    
+    def setup(self, num_syms):
+        self.population_policy = self.get_population_policy()
```
IvoDD, line 60: We don't need the population policy in setup because we only use `Persistent` lib

python/benchmarks/real_list_operations.py
```diff
@@ -6,59 +6,81 @@
...
+        self.lib = self.get_library_manager().get_library(AWSListSymbols.library_type, num_syms)
+        self.test_counter = 1
+        symbols_list = self.lib.list_symbols()
+        assert num_syms == len(symbols_list), f"The library contains expected number of symbols {symbols_list}"
+        self.lib._nvs.version_store._clear_symbol_list_keys() # clear cache
+
+    def time_list_symbols(self, num_syms):
+        assert self.test_counter == 1, "Test executed only once in setup-teardown cycle" 
         self.lib.list_symbols()
+        self.test_counter += 1
```
IvoDD, line 70: I think using this `test_counter` is a bit too much. The assertion that number is 1 in `setup_cache` is enough.

Also it doesn't make any sense to do this only once if we are using a persistant storage. What are you trying to achieve? Do you want to avoid the symbol list compaction? If we want to measure symbol list compaction we should use a modifiable library. If we want to measure symbol list traversal without compaction we should pass `compact=False`. Probably doing both is actually a good idea.

python/benchmarks/real_list_operations.py
```diff
@@ -6,59 +6,81 @@
...
 
-    def time_list_symbols_first_snapshot(self, storage_info, num_syms):
-        self.lib.list_symbols(self.aws.first_snapshot)
 
-    def time_list_symbols_last_snapshot(self, storage_info, num_syms):
-        self.lib.list_symbols(self.aws.last_snapshot)
-
-class AWSVersionSymbols:
+class AWSVersionSymbols(AsvBase):
```
IvoDD, line 83: Is this also testing listing? Maybe include it in the name: `AWSListVersionSymbols`

python/benchmarks/real_list_operations.py
```diff
@@ -68,39 +90,90 @@ class AWSVersionSymbols:
 
     timeout = 1200
 
-    SETUP_CLASS = (GeneralSetupSymbolsVersionsSnapshots(storage=Storage.AMAZON, prefix="LIST_VERSIONS")
-        .set_mean_number_versions_per_sym(35) # change to lower for testing
-        .set_max_number_versions(50) # number versions is approx = num_syms * mean_number_versions
-        .set_with_metadata_for_each_version()
-        .set_with_snapshot_for_each_version()
-        .set_params([25, 50])) # for test purposes: .set_params([5, 6]))
+    library_manager = TestLibraryManager(storage=Storage.AMAZON, name_benchmark="LIST_SYMBOLS")
```
IvoDD, line 93: Name should be `LIST_VERSIONS` to not collide with the above

python/benchmarks/real_list_operations.py
```diff
@@ -68,39 +90,90 @@ class AWSVersionSymbols:
...
+                self.get_logger().info(f"Library already exists, population skipped")
+            # Getting one snapshot - the last
+            lib = self.get_library_manager().get_library(AWSVersionSymbols.library_type, number_symbols)
+            snapshot_name = lib.list_snapshots(load_metadata=False)[-1]
+            last_snapshot_names_dict[number_symbols] = snapshot_name
+        manager.log_info() # Always log the ArcticURIs 
+        return last_snapshot_names_dict
+    
+    def setup(self, last_snapshot_names_dict, num_syms):
+        self.population_policy = self.get_population_policy()
```
IvoDD, line 139: Unused

python/benchmarks/real_modification_functions.py
```diff
@@ -0,0 +1,285 @@
...
+        generate_and_log(cached_results.append_single_dict, 'append_single_dict', 1, next_timestamp.to_timestamp())
+
+    def setup(self, cache: CacheForModifiableTests, num_rows):
+        manager = self.get_library_manager()
+        self.cache = cache
+        writes_list = self.cache.write_and_append_dict[num_rows]
+        
+        self.lib = manager.get_library(LibraryType.MODIFIABLE)
+
+        self.symbol = f"symbol-{os.getpid()}"
```
IvoDD, line 144: pid is already in the library, fine to use a more generic name

python/benchmarks/real_modification_functions.py
```diff
@@ -0,0 +1,285 @@
...
+        self.get_library_manager().clear_all_modifiable_libs_from_this_process()
+
+    def time_update_single(self, cache, num_rows):
+        self.lib.update(self.symbol, self.cache.update_single_dict[num_rows])
+
+    def time_update_half(self, cache, num_rows):
+        self.lib.update(self.symbol, self.cache.update_half_dict[num_rows])
+
+    def time_update_full(self, cache, num_rows):
+        #self.lib.update(self.symbol, self.cache.update_full)
```
IvoDD, line 159: Remove comment?

python/benchmarks/real_read_write.py
```diff
@@ -89,126 +55,147 @@ class LMDBReadWrite:
...
+        self.read_lib = self.get_library_manager().get_library(LibraryType.PERSISTENT)
+        self.write_lib = self.get_library_manager().get_library(LibraryType.MODIFIABLE)
+        # We could also populate the library like so (we don't need )
+        # populate_library(self.write_lib, )
+
+    def teardown(self, num_rows):
+        # We could clear the modifiable libraries we used
+        self.get_library_manager().clear_all_modifiable_libs_from_this_process()
+
+    def get_last_x_percent_date_range(self, num_rows, percents):
```
IvoDD, line 102: That's why I wanted to move this function in utils. A very identiacal function was used in `read_batch_functions.py` iirc

IvoDD (changes requested):
I added quite a few small changes suggestions.

A few more imporant things to discuss maybe offline:
- random seeding
- list symbol tests with respect to symbol list compaction

python/benchmarks/real_modification_functions.py
grusev: yep we may choose to use or not :-)

python/benchmarks/real_list_operations.py
grusev: correct!

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +131,421 @@ def __new__(cls, *args, **kwargs):
...
-
-    def set_mean_number_versions_per_sym(self, mean) -> 'GeneralSetupSymbolsVersionsSnapshots':
+        self.versions_max = versions_max
         self.mean = mean
         return self
-
-    def set_max_number_versions(self, versions_max) -> 'GeneralSetupSymbolsVersionsSnapshots':
-        self.versions_max = versions_max
+    
+    def generate_snapshots(self) -> 'LibraryPopulationPolicy':
```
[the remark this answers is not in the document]
grusev: will think of better name

python/arcticdb/util/environment_setup.py
grusev: get_library will return existing library in most of cases. Only if does not exist will create new library. Thus the library options is possible but not quite determined in the case when library exists what to do - to check if lib is having same options or skip and disregard. Still confusing.

The hole TestLibraryManager is thin wrapper over Arctic, which goal is to "hide" details of internal management associated with how exactly placement of libraries happen and where and how to exactly locate and remove them. The argument that we can live without classes is not good as if we could we would have not had fixtures as classes at first place. And exactly this is  this class. It provides extension to functionality of Arcitc class which is suited to do jobs better in very specific environment. 

Thus a possible and better options in this case are:

 - define create_library method ... now or when we actually need it.

 - workarround with what we have currently: 
    tlm.library_options = LibraryOptions(...)
    lib = lib.get_library(...)
    
First proposal is clean why, second is workarround. But yet we do not have scenario that requires that, hence create_library was not implemented. 

That is something that specialized class provides over free floating functions - encapsulation, emergent design (extend when you need, change easily something when new info arrises without breaking tests etc)

Classes are the natural way to grow in complexity and need, as they require use of the whole contract not only parts of it function by function, which in time would errode the overall specialization and will lead to conflicts of requirements.

python/arcticdb/util/environment_setup.py
grusev: fixed

grusev: The for methods serve very different purposes. Their usage scenarios is now documented as their use might not be that obvious

python/benchmarks/real_batch_functions.py
```diff
@@ -0,0 +1,158 @@
...
+        manager = self.get_library_manager()
+        policy = self.get_population_policy()
+        logger = self.get_logger()
+        number_symbols_list, number_rows_list = AWSBatchBasicFunctions.params
+        for number_symbols in number_symbols_list:
+            lib_suffix = number_symbols
+            if not manager.has_library(LibraryType.PERSISTENT, lib_suffix):
+                start = time.time()
+                for number_rows in number_rows_list:
+                    policy.set_parameters([number_rows] * lib_suffix)
```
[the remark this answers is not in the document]
grusev: it really does not matter where exactly this is done. Nothing will change for good or bad, as population policy allows you to add/change parameters get_population_policy() is making sense to hold only common properties

python/benchmarks/real_comparison_benchmarks.py
grusev: yes good to add comment ...

    # NO_OPERATION measures class memory allocation. This is the actual memory that
    # is used by the tools and code that does the measurement. Thus any other measurement
    # number should be deducted with NO_OPERATION number to receive actual number.
    # The whole discussion is available at: 
    # https://github.com/man-group/ArcticDB/wiki/ASV-Benchmarks:-Running,-designing-and-implementing#understanding-and-implementing-peakmem-benchmarks

python/benchmarks/real_comparison_benchmarks.py
grusev: bettered:
            # measures base memory which need to be deducted from 
            # any measurements with actual operations
            # see discussion above

python/benchmarks/real_list_operations.py
grusev: no. if done it will be confusing.
Now it is clear that number of symbols is always the length of the of the list.
Thus num_symbols is always calculated. 

The `LibraryPopulationPolicy` doc was misplaced at constructor init ... it explains that well with examples

python/benchmarks/real_list_operations.py
grusev: as discussed we move population in setup() due to usage of self.lib._nvs.version_store._clear_symbol_list_keys() # clear cache
so we will actually use it there.

python/benchmarks/real_list_operations.py
grusev: As dicussed there is a common problem that there is no documentation on menchamrks what exactly is measure. We agreed that this will be this case:

    """
    The primary purpose of this test is to measure the complete time 
    the list_symbol takes to complete on a symbol without a cache.
    That would be maximum time the user would wait, and we have to track it
    """

self.test_counter could be too much indeed but it does not hurt at all the numbers, but strenthens the checks against accidental modifications. So leaving it as is should be ok

python/benchmarks/real_list_operations.py
grusev: nope only versions and snapshots and as original version of the benchamrk is populated already in DB we should keep current name

python/arcticdb/util/utils.py
```diff
@@ -386,17 +397,23 @@ def generate_random_floats(cls, dtype: ArcticFloatType,
             return np.round(np.random.uniform(min_value, max_value, size), round_to).astype(dtype)
         
     @classmethod
-    def generate_random_string_pool(cls, str_length: int, pool_size: int, seed = 1) -> List[str]:
+    def generate_random_string_pool(cls, str_length: int, pool_size: int,
```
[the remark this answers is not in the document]
grusev: Very good suggestion! Currently all functions that accept seed will take also none as parameter and will not do anything in this case leaving the test creator full control when needed to set the proper seed.

python/arcticdb/util/environment_setup.py
```diff
@@ -594,159 +556,152 @@ def get_versions_list(self, number_symbols: int) -> List[np.int64]:
                 seed=365)
```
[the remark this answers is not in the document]
grusev: There is no self.seed on the population list class, thus no way to use it, so random number prefered to set (makes no difference here exactly)

python/arcticdb/util/environment_setup.py
grusev: done

python/benchmarks/real_modification_functions.py
grusev: fixed

python/arcticdb/util/environment_setup.py
grusev: The current implementation is ok it goes for fluent pattern which is widely used in python and even in arcticdb (queries). The requested change would not add any additional value nor it will fix any error. Therefore it is not needed at this moment.

python/arcticdb/util/environment_setup.py
grusev: TestLibraryManager does not duplicate any work of Arctic Library manager. It provides very needed isolation for creating tests from where and how Arctic will create them. More info can be found in the new documentation of the class:
https://github.com/man-group/ArcticDB/blob/f053c6afdb52a88a6e0bff5b6167d463d33c0aa4/python/arcticdb/util/environment_setup.py#L192

As a conclusion I do not argue that eventually a version without use of class and only functions can be created. I argue its value and need. There are many many arguments why not take this approach in the description there and in fact that Arctic is class and in order to override class you need a class not functions.

I do find enough reasons that current implementation is better - one I can name is the fact that no one uses Arctic in man db to directly create libraries - there is and UI library manager. In all real implementation when you have certain specifics for management the infrastructure you override what is avail. Current implementation is exactly this. That is why it is ok andy change is rather waste of resources and eventual result is more likely to force implementation of similar thing at some point.

python/benchmarks/real_comparison_benchmarks.py
G-D-Petrov: Might be better to call this BASE_MEMORY or something like that

python/arcticdb/util/environment_setup.py
```diff
@@ -99,491 +131,421 @@ def __new__(cls, *args, **kwargs):
...
+        index = 0
+        for param_value in self.parameters:
+            versions = versions_list[index]
+            
+            if self.index_is_auto_increment:
+                symbol = self.get_symbol_name(index)
+            else:
+                symbol = self.get_symbol_name(param_value)
+
+            if self.parameters_is_number_rows_list:
```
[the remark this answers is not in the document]
grusev: For the scenarios that are currently implemented we do not need new dataframe on each write. That is the reason this was not implemented. And yes there is value of having new dataframes on each write. But that is expensive operation and should be approached carefully.

Still there is already a class that covers the scenario with new dataframes - SequentialDataframesGenerator, it is used in append scenarios.

Once a real case with Library Populator arises an optoion can be easily added for a new DF on each new write.

python/arcticdb/util/environment_setup.py
```diff
@@ -594,159 +556,152 @@ def get_versions_list(self, number_symbols: int) -> List[np.int64]:
...
-        setup.setup_environment()
-        assert setup.check_ok()
-
-        #Changing parameters should trigger not ok for setup
-        setup.set_with_metadata_for_each_version()
-        setup.set_with_snapshot_for_each_version()
-        setup.set_params([2, 3])
-        assert not setup.check_ok()
-        
+    def test_modifiable_access(cls):
```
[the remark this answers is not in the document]
grusev: done

python/benchmarks/real_batch_functions.py
```diff
@@ -0,0 +1,158 @@
...
+        return lpp
+    
+    def setup_cache(self):
+        manager = self.get_library_manager()
+        policy = self.get_population_policy()
+        logger = self.get_logger()
+        number_symbols_list, number_rows_list = AWSBatchBasicFunctions.params
+        for number_symbols in number_symbols_list:
+            lib_suffix = number_symbols
+            if not manager.has_library(LibraryType.PERSISTENT, lib_suffix):
```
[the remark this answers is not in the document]
grusev: The logic here needs to be more flexible and cannot be that fixed. In such occasions (rare) is ok not to use the function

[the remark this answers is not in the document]
grusev: (observe thta between has_library() and populate_library there is code

python/benchmarks/real_comparison_benchmarks.py
grusev: yes. That is better naming

python/benchmarks/real_read_write.py
```diff
@@ -89,126 +55,144 @@ class LMDBReadWrite:
...
-                                             # as well as name of unique storage prefix
-                                             prefix="READ_WRITE").set_params([1_000_000, 2_000_000])
+    library_manager = TestLibraryManager(storage=Storage.AMAZON, name_benchmark="READ_WRITE_WIDE",
+                                     library_options=LibraryOptions(rows_per_segment=1000, columns_per_segment=1000))
 
-    params = SETUP_CLASS.get_parameter_list()
-    param_names = LMDBReadWrite.param_names
+    param_names = ["num_cols"]
+    # NOTE: Change of parameters will trigger failure as original library must also be deleted manually.
+    #       Therefore if you plan changes to those numbers make sure to delete old library manually
```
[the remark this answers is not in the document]
grusev: Work with shared persistence storage requires indeed a handbook for what to do and what not to do. And I will create a wiki page for that. 

Here is what is quick explanation why this is need for that test.

When a test create a library on the shared persistent space, that library is considered to stay. Consequent test executions check that this library is there and continue the test or create it once again. Library structure is not checked (previous version contained structural checks of library and not just having checked its name, which was considered as overly complicating thing)

Now if the parameters of the test change, that would automatically trigger different code paths in test, but since the library was created once with old parameters the check will pass and  it will be considered created, thus tests done on the library with those new params will produce either misleading results, or they will simply do not work. Both outcomes are possible.

Therefore there are 2 ways to battle this problem:

 - having comments like this to assure whoever is changing parameters does clean the previous library first.
 - having a more complicated name where library has the parameters encoded as suffix. Then a change in parameters will simply mean a new library created on the persistent store to stay. The end effect could be many libraries created and persistent storage bloated, which actually defeats its purpose. (For that there is already test storage space which should be used)
 
-----------------------------------------------------------

In other words persistent storage comes with logistics complications and therefore management should be applied. That is counterintuitive to most developers as usually tests are considered isolated from each other, and thus errors are possible. But in fact tests can and should be created considering also such persistence factor. Still they require some additional management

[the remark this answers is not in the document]
grusev: Docs will be prepared, for now the comment is changed

python/arcticdb/util/environment_setup.py
```diff
@@ -359,13 +359,26 @@ def remove_all_test_libs(cls, storage_type: Storage):
...
             ac.delete_library(to_delete)  
             get_console_logger().info(f"Delete library [{to_delete}] from storage space having [{uri_str_to_confirm}]")          
         assert len(ac.list_libraries()) == 0, f"All libs for storage space [{uri_str_to_confirm}] deleted"        
 
+    def remove_all_persistent_libs_for_this_test(self):
+        """
+        This will remove all persistent libraries for this test from the persistent storage
+        Therefore use wisely only when needed (like change of parameters for tests)
+        """
+        name_prefix = f"{LibraryType.value}_{self.name_benchmark}"
```
IvoDD, line 374: Should be `LibraryType.PERSISTENT.value`

python/benchmarks/real_read_write.py
```diff
@@ -56,8 +56,15 @@ class AWSReadWrite(AsvBase):
...
-    #       Therefore if you plan changes to those numbers make sure to delete old library manually 
+    # NOTE: If you plan to make changes to parameters, consider that library already may exist created with different number
+    #       of rows. Therefore:
+    #  
+    #       1. make sure to use add set_test_mode() to instantiation of TestLibraryManager during tests. Like:
+    #              library_manager = TestLibraryManager(storage=Storage.AMAZON, name_benchmark="READ_WRITE").set_test_mode()
+    #       2. Delete library from persistent storage once new parameters are ready to be committed 
+    #          and remove "set_test_mode()". To delete all libs for current test use:
+    #              library_manager = TestLibraryManager(storage=Storage.AMAZON, name_benchmark="READ_WRITE")
+    #              library_manager.remove_all_persistent_libs_for_this_test()
```
IvoDD, line 67: I think this comment is confusing. Are you suggesting to commit a change to delete the library inside the github runners. That seems like an odd process, because then we would have to revert it.

I think it would be better if instructions describe how to do it locally instead of on github runner.
What do you think about the following:
```
If you plan to make changes to parameters, consider that library already may exist created with different number
of rows. Therefore, you need to either:
- Rename the library by changing `name_benchmark` or by adding a `lib_suffix`. Note that this will keep the old library around.
- Clean up the old library with (which needs to be run with the same environment variables as the github runner):
       library_manager = TestLibraryManager(storage=Storage.AMAZON, name_benchmark="READ_WRITE")           
       library_manager.remove_all_persistent_libs_for_this_test()
```

grusev: actually the process is much simpler and will leave it to the wiki which I currently work on: https://github.com/man-group/ArcticDB/wiki/ASV-Benchmarks:-Real-storage-tests

as we are talking about persistent tests which are accessed by all no specific vars are needed just simple 2-3 step process which will be explained there and the comment will have a link to it

IvoDD (approved):

G-D-Petrov (approved):

poodlewars (approved):
Please update the commit message to something meaningful when you merge this

PR #2252, opened 2025-03-19
ci: Add workflow for Windows

jjerphan:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_with_conda.yml
+++ b/.github/workflows/build_with_conda.yml
@@ -40,10 +40,10 @@ jobs:
         uses: SimenB/github-actions-cpu-cores@v1.1.0
         id: cpu-cores
 
-      - name: Install Conda environment from environment_unix.yml
+      - name: Install Conda environment from environment-dev.yml
         uses: mamba-org/setup-micromamba@v2.0.0
         with:
-          environment-file: environment_unix.yml
+          environment-file: environment-dev.yml
           init-shell: >-
             bash
           cache-environment: true
@@ -124,10 +124,10 @@ jobs:
         uses: SimenB/github-actions-cpu-cores@v1.1.0
         id: cpu-cores
 
-      - name: Install Conda environment from environment_unix.yml
+      - name: Install Conda environment from environment-dev.yml
         uses: mamba-org/setup-micromamba@v2.0.0
         with:
-          environment-file: environment_unix.yml
+          environment-file: environment-dev.yml
           environment-name: arcticdb
           init-shell: >-
             bash
@@ -175,3 +175,65 @@ jobs:
         env:
           ARCTICDB_USING_CONDA: 1
 
+  windows:
+    runs-on: windows-latest
+    env:
+      ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+    steps:
+      - uses: actions/checkout@v3.3.0
+        # DONT use recursive submodules checkout to simulate conda feedstock build
+        # with:
+        #   submodules: recursive
+
+      - name: Get number of CPU cores
+        uses: SimenB/github-actions-cpu-cores@v1.1.0
+        id: cpu-cores
+
+      - name: Install Conda environment from environment-dev.yml
+        uses: mamba-org/setup-micromamba@v2.0.0
+        with:
+          environment-file: environment-dev.yml
+          init-shell: >-
+            cmd
+          cache-environment: true
+          post-cleanup: 'all'
+
+      - name: Build ArcticDB with conda (ARCTICDB_USING_CONDA=1)
+        shell: cmd {0}
+        run: |
+          # Protocol buffers compilation require not using build isolation.
+          python -m pip install --no-build-isolation --no-deps -v -e .
+        env:
+          ARCTICDB_USING_CONDA: 1
+          ARCTICDB_BUILD_CPP_TESTS: 1
+          ARCTIC_CMAKE_PRESET: windows-conda-release-build
+
+      - name: Build C++ Tests
+        shell: cmd {0}
+        run: |
+          cd cpp\out\windows-conda-release-build
+          msbuild arcticdb_rapidcheck_tests.vcxproj /p:Configuration=Release /p:Platform=x64 /m
+          msbuild test_unit_arcticdb.vcxproj /p:Configuration=Release /p:Platform=x64 /m
+
+      - name: Run C++ Tests
+        shell: cmd {0}
+        run: |
+          cd cpp\out\windows-conda-release-build
+          ctest -C Release -T test
+
+      - name: Install npm
+        uses: actions/setup-node@v3.3.0
+        with:
+          node-version: '16'
+
+      - name: Install azurite
+        shell: cmd {0}
+        run: |
+          npm install -g azurite
+
+      - name: Test with pytest
+        shell: cmd {0}
+        run: |
+          cd python
+          set ARCTICDB_RAND_SEED=%RANDOM%
+          python -m pytest --timeout=3600
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -82,6 +82,8 @@ dependencies:
   - trustme
   - entt
   - psutil
-  - memray 
-  - pytest-memray
-
+  # Use memray and pytest-memray only on Linux and MacOS
+  - sel(linux): memray
+  - sel(linux): pytest-memray
+  - sel(osx): memray
+  - sel(osx): pytest-memray
--- a/docs/mkdocs/docs/technical/contributing.md
+++ b/docs/mkdocs/docs/technical/contributing.md
@@ -94,10 +94,10 @@ due to a linkage problems with libprotobuf.
 See: https://github.com/man-group/ArcticDB/pull/449**
 
  - [Install `mamba`](https://mamba.readthedocs.io/en/latest/installation.html)
- - Create the `arcticdb` environment from its specification (`environment_unix.yml`):
+ - Create the `arcticdb` environment from its specification (`environment-dev.yml`):
 
   ```bash
-  mamba env create -f environment_unix.yml
+  mamba env create -f environment-dev.yml
   ```
 
  - Activate the `arcticdb` environment (you will need to do this for every new shell session):
@@ -113,7 +113,7 @@ See: https://github.com/man-group/ArcticDB/pull/449**
 For instance:
 
  - for debug build on Linux with mamba and conda-forge, use:
- 
+
   ```bash
   export ARCTICDB_USING_CONDA=1
   cmake -DTEST=off --preset linux-conda-debug cpp
@@ -395,7 +395,7 @@ If possible, a version number should also be added such that future changes in t
 ### API
 
 Any changes to the API (including when exceptions are thrown and the type of the exception) must be weighed up against the change breaking behaviour for existing users.
-Please make this as clear to reviewers as possible by ensuring **API changes are clearly described in the PR description**. 
+Please make this as clear to reviewers as possible by ensuring **API changes are clearly described in the PR description**.
 If the change is breaking, please also ensure that that is appropriately highlighted in the PR description as well.
 This is particularly true of the `NativeVersionStore` API, as this has many users inside Man Group.
 


cpp/CMakeLists.txt
```diff
@@ -74,20 +77,82 @@ if(ARCTICDB_SANITIZER_FLAGS)
     message(STATUS "Building ArcticDB with sanitizers. Compiler flags: ${ARCTICDB_SANITIZER_FLAGS}")
 endif()
 
+if(${ARCTICDB_USING_CONDA})
+    # Required to be able to include headers from glog since glog 0.7
+    # See: https://github.com/google/glog/pull/1030
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DGLOG_USE_GLOG_EXPORT")
```
vasil-pashov, line 83: I think this is for setting C macro. I prefer using `add_compile_definitions` or even better `target_compile_definitions` rather than changing the CMAKE_CXX_FLAGS

cpp/CMakeLists.txt
```diff
@@ -74,20 +77,82 @@ if(ARCTICDB_SANITIZER_FLAGS)
...
+    endif()
 
     # We always generate the debug info on Windows since those are not in the final .dll.
     # sccache needs the /Z7 format, but the linker will still produce .pdb files.
     # Guide to MSVC compilation warnings https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c4000-through-c4199?view=msvc-170
     add_compile_options(
-        /DWIN32 /D_WINDOWS /GR /EHsc /bigobj /Z7 /wd4244 /wd4267
-        "$<$<CONFIG:Debug>:/Od;/MTd>"
-        "$<$<CONFIG:Release>:/MT;/Ox>"
+        /DWIN32 /D_WINDOWS /GR /EHsc /bigobj /Z7 /wd4244 /wd4267 /MP
```
vasil-pashov, line 139: Let's move the `/MP` flag in a preset. Arctic requires quite a lot of RAM to compile on multiple threads (it OOMs on my personal 16GiG 16 core machine). Leaving it in the main CMake would mean that people would have to change the CMakeLists.txt just to make it compile. Most open source project I've seen leave this outside of the `CMakeLists.txtx`

cpp/arcticdb/CMakeLists.txt
```diff
@@ -160,11 +160,19 @@ if (NOT MSVC)
     add_compile_options("-fno-omit-frame-pointer")
 else()
     add_compile_options(
-        /WX
         /wd4250 # Suppress inherits via dominance warning
-        /we4189 # Unused local variable is an error
         /we4100 # Unused function parameter is an error
+        # For now, we do not want to treat warnings as errors for those builds.
```
vasil-pashov, line 165: Why? What is the difference between vcpkg and conda in this regard?

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -79,7 +79,7 @@ VariantData binary_membership(const ColumnWithStrings& column_with_strings, Valu
 
     util::BitSet output_bitset;
     constexpr auto sparse_missing_value_output = std::is_same_v<std::remove_reference_t<Func>, IsNotInOperator>;
-    details::visit_type(column_with_strings.column_->type().data_type(),[&] (auto col_tag) {
+    details::visit_type(column_with_strings.column_->type().data_type(),[&, sparse_missing_value_output] (auto col_tag) {
```
vasil-pashov, line 82: Was this because of a compiler error? Shouldn't the `&` capture it?

jjerphan: It was required because it is not captured by value by default unless explicitly specified (in the scope of the lambda functions, `sparse_missing_value_output` was not declared nor defined).

cpp/CMakeLists.txt
jjerphan: Yes, I agree. For now, I just minimalistically moved code around to have it work, but eventually all the complexity of `CMakeLists.txt` must be removed.

cpp/arcticdb/CMakeLists.txt
```diff
@@ -52,6 +52,9 @@ if(NOT ${ARCTICDB_USING_CONDA})
     find_package(Libevent CONFIG REQUIRED)
     set(ARCTICDB_MONGO_LIBS $<IF:$<TARGET_EXISTS:mongo::mongocxx_static>,mongo::mongocxx_static,mongo::mongocxx_shared>)
     set(Zstd_LIBRARY $<IF:$<TARGET_EXISTS:zstd::libzstd_shared>,zstd::libzstd_shared,zstd::libzstd_static>)
+    # Libraries baked into the internal and external Linux images are found by searching various paths
+    # On Windows, vcpkg will provide a "Config" which takes precedence
+    find_library(Pcre_LIBRARY NAMES pcre PATHS /usr/local/lib/libpcre.a REQUIRED)
```
jjerphan, line 57: Is this an old workaround, or is it still required?

cpp/arcticdb/CMakeLists.txt
jjerphan: This must only be set for the conda-forge build, and we need to understand the origin of warning for them.

cpp/CMakePresets.json
```diff
@@ -63,8 +63,8 @@
       "generator": "Ninja",
       "environment": { "cmakepreset_expected_host_system": "Windows" },
       "cacheVariables": {
-        "VCPKG_OVERLAY_TRIPLETS": "custom-triplets",
-        "VCPKG_TARGET_TRIPLET": "x64-windows-static-msvc"
```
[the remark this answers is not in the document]
jjerphan: > I see that his was moved the windows-cl-debug but this will break the VCPKG PyPi releases.

I do not think this would break the VCPKG PyPI releases since `windows-cl-release` derives from `windows-cl-debug` whose definition is identical in practice: from `windows` and to `windows-cl-debug`, `CMAKE_{C,CXX}_COMPILER` have been pushed up, and `VCPKG_OVERLAY_TRIPLETS` and `VCPKG_TARGET_TRIPLET` have been pushed down.

I might have missed something.

The inheritance or composition could be reworked if this could improve the understanding.

cpp/CMakeLists.txt
```diff
@@ -14,8 +14,11 @@ project(arcticdb VERSION 0.0.1)
 
 enable_testing()
 
+# Temporary debug
+set(CMAKE_VERBOSE_MAKEFILE ON CACHE BOOL "ON" FORCE)
+
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-set(ARCTICDB_MSVC_OMIT_RUNTIME_CHECKS OFF CACHE BOOL "")
```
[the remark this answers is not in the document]
jjerphan: I need to remember.

.github/workflows/build_with_conda.yml
```diff
@@ -175,3 +175,89 @@ jobs:
...
+            -DCMAKE_C_COMPILER_LAUNCHER=sccache
+          cmake --build --preset windows-cl-conda-release --parallel --target arcticdb_rapidcheck_tests
+          cmake --build --preset windows-cl-conda-release --parallel --target test_unit_arcticdb
+          cd out/windows-cl-conda-release-build/
+          CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --target test
+
+      - name: Build Python target for conda
+        shell: cmd /C call {0}
+        run: |
+          python -m pip install --no-build-isolation --no-deps -v -e .
```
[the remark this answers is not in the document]
jjerphan: Currently, it seems that it rebuilds binaries, even though `sccache` makes it faster.

cpp/CMakeLists.txt
```diff
@@ -75,20 +78,82 @@ if(ARCTICDB_SANITIZER_FLAGS)
...
+
+            # Required to link against azure-storage-common.dll
+            # See: https://github.com/Azure/azure-sdk-for-cpp/blob/c914c0933fa4835c6618a42bb276d6077423b170/sdk/storage/azure-storage-common/inc/azure/storage/common/dll_import_export.hpp#L19
+            AZ_STORAGE_COMMON_DLL=1
+
+            # Required to link against azure-storage-blobs.dll
+            # See: https://github.com/Azure/azure-sdk-for-cpp/blob/bf36db827f5ea1ef2e38953875c6c335b4242b2e/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/dll_import_export.hpp
+            AZ_STORAGE_BLOBS_DLL=1
+        )
+    endif()
```
jjerphan, line 136: I needed those at a time (I cannot remember why) but this really should be handled by packages instead.

We might need to repackage some of those dependencies on conda-forge, this needs to be inspected further again.

.github/workflows/build_with_conda.yml
```diff
@@ -175,3 +175,89 @@ jobs:
...
+          environment-file: environment-dev.yml
+          init-shell: bash cmd.exe
+          cache-environment: true
+          post-cleanup: 'all'
+
+      - name: Build ArcticDB C++ target (arcticdb_ext) for conda
+        shell: cmd /C call {0}
+        run: |
+          set CMAKE_GENERATOR_PLATFORM=
+          set CMAKE_GENERATOR_TOOLSET=
```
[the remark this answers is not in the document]
jjerphan: This is required for Ninja to be usable with MSVC.

Adding them as part of the `windows-cl-conda-*` CMake preset does not work.

I do not know enough about MSVC's toolchain to be understand the reason behind it.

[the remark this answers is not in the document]
jjerphan: I just have added a comment.

If you think there's a better way to solve it, I would be glad to improve this workaround.

environment-dev.yml
```diff
@@ -31,7 +31,10 @@ dependencies:
   - cyrus-sasl
   - aws-sdk-cpp >=1.11.405
   - prometheus-cpp
-  - libprotobuf <7
+    # We have to use the build of libprotobuf starting from this PR:
+    # https://github.com/conda-forge/libprotobuf-feedstock/pull/277
+  - sel(win): libprotobuf==5.29.3=h*_3
```
jjerphan, line 36: This is a known installable and patched version of `libprotobuf 5`.

`libprotobuf 6` is not as installable for Windows in my experience, and also would need to be patched (see https://github.com/conda-forge/libprotobuf-feedstock/issues/275#issuecomment-3554088930).

alexowens90 (approved):

.github/workflows/build_with_conda.yml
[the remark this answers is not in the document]
vasil-pashov: I find it odd that this is needed only for the windows conda build. The other builds don't set this and still use ninja. How is conda different? What's the error that's raised if these are not added?

.github/workflows/build_with_conda.yml
```diff
@@ -288,3 +292,82 @@ jobs:
...
+  windows:
+    runs-on: windows-latest
+    env:
+      ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+      SCCACHE_GHA_VERSION: ${{vars.SCCACHE_GHA_VERSION || 1}}
+    steps:
+      - uses: actions/checkout@v3.3.0
+        # DONT use recursive submodules checkout to simulate conda feedstock build
+        # with:
+        #   submodules: recursive
```
vasil-pashov, line 303: nit: Why not just remove these?

.github/workflows/build_with_conda.yml
```diff
@@ -288,3 +292,82 @@ jobs:
...
+        shell: cmd /C call {0}
+        # Empty values for `CMAKE_GENERATOR_{PLATFORM,TOOLSET}` is necessary
+        # for Ninja to be used as a generator with MSVC.
+        run: |
+          set CMAKE_GENERATOR_PLATFORM=
+          set CMAKE_GENERATOR_TOOLSET=
+          cd cpp
+          cmake -DTEST=ON --preset windows-cl-conda-release
+          cmake --build --preset windows-cl-conda-release --parallel --target arcticdb_rapidcheck_tests
+          cmake --build --preset windows-cl-conda-release --parallel --target test_unit_arcticdb
```
vasil-pashov: Why are those lines needed? The test targets should be built by default when TEST=ON

cpp/CMakeLists.txt
vasil-pashov: I agree this doesn't look right. We can merge it like this, though I'd appreciate if you can investigate it at some point.

jjerphan: I can confirm that some packages on conda-forge would need to be rebuilt with a patch on Windows to avoid setting this.

The maintenance cost would be much more significant that having to set those compile definitions.

jjerphan: I can confirm that some packages on conda-forge would need to be rebuilt with a patch on Windows to avoid setting this.

The maintenance cost would be much more significant that having to set those compile definitions.

.github/workflows/build_with_conda.yml
jjerphan: Mainly to keep the style of this workflow file (this comment exists for Linux and Mac and is informative I think).

This file needs to be reformated and improved, and this better be done in another PR IMO.

.github/workflows/build_with_conda.yml
jjerphan: IIRC, I had to specify them explicitly for them to be built. But I can explore it again.

jjerphan (commented):
Replying to Vasil's review.

.github/workflows/build_with_conda.yml
vasil-pashov: Yes, please

jjerphan:
```suggestion
```

.github/workflows/build_with_conda.yml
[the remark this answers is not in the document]
jjerphan: I do not know. I can revert this change to see the error but (it is apparently something with Ninja which cannot be used for x64).

.github/workflows/build_with_conda.yml
jjerphan: https://github.com/man-group/ArcticDB/pull/2798 was opened for it.

.github/workflows/build_with_conda.yml
```diff
@@ -288,3 +288,106 @@
...
+            eval "$command"
+          else
+            cd python
+            python -m pytest --timeout=3600 -v -n logical --dist worksteal tests $ARCTICDB_PYTEST_ARGS
+          fi
+        env:
+          ARCTICDB_USING_CONDA: 1
+          COMMANDLINE: ${{ inputs.run_commandline }}
+          ARCTICDB_PYTEST_ARGS: ${{ inputs.run_custom_pytest_command }}
+          NODE_OPTIONS: --openssl-legacy-provider
```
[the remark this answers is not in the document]
jjerphan: Resolved with https://github.com/man-group/ArcticDB/pull/2252/files#diff-d7c8fbdebe6fe3a891aedff67e2e5b7461a6579af6a50f40b0d99f2784424952R2-R3.

.github/workflows/build_with_conda.yml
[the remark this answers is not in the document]
jjerphan: For context, see https://github.com/man-group/ArcticDB/pull/2252#issuecomment-3617897290.

cpp/CMakeLists.txt
jjerphan: I added a comment in the code in this regards.

PR #2254, opened 2025-03-21
Fix for STS support

poodlewars:
The change to support GCP #2176 accidentally broke STS support with S3.

The issue was that the native settings,

```
def native_config
```

just returned a dummy object. This was fine for the config library as it got overridden later correctly (so listing libraries worked) but was not overridden for normal libraries (so actually working with a library failed).

I've just removed all this complexity of overriding and made the `native_config` call always return something valid.

This call in `helper.py` is needless (and confusing) now, so I've removed it.

```
        native_cfg.update(NativeS3Settings(aws_auth, aws_profile, use_internal_client_wrapper_for_testing))
```

I've also added a set of unit tests, which caught a surprising number of small issues that are hopefully clear from the changeset, and some better Python reprs for debugging the settings.

This commit https://github.com/man-group/ArcticDB/commit/5f5addf0294f6066e2467fdb073cffa3bf7e0903 shows my notes on the STS setup and the repro script I used for an end to end test of STS support.

### DIFF
--- a/cpp/arcticdb/storage/common.hpp
+++ b/cpp/arcticdb/storage/common.hpp
@@ -86,13 +86,23 @@ public:
     }
 
     std::string to_string() {
-        return util::variant_match(config_, [](std::monostate) {
+        return util::variant_match(config_, [](std::monostate) -> std::string {
             return "empty";
-        }, [](s3::S3Settings) {
-            return "s3";
-        }, [](s3::GCPXMLSettings) {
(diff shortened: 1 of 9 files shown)


phoebusm (approved):

poodlewars:
Persistent storage tests look good, so merging https://github.com/man-group/ArcticDB/actions/runs/14031881508/job/39281814815

PR #2255, opened 2025-03-21
GCP real storage tests

grusev:
#### Reference Issues/PRs

( AWS S3 tests link - started from actions https://github.com/man-group/ArcticDB/actions/runs/14172795087/job/39700720982 )
(local tests only started from new branch: https://github.com/man-group/ArcticDB/actions/runs/14188770022)

At initial stage of work ...

Introducing 2 markers:
 - storage - for real storage tests (approx 390 out of 1400 tests)
 - authentication - for authentication tests (they are executed againsy real storage, but are different category). Main purpose is to provide quick check for actual state of the authentication only.

The markers now allow more targeted execution of tests - like only storage, authentication tests or all withou a group or more.

Added several new fixtures for GCP type of storage

marked test that use following fixtures as storage tests (potentially they will be executed against all _real_ storages). There will be also need to have exclude mark for certain storage type from some tests TBD)
- basic_store
- basic_store_factory
- version_store_and_real_s3_basic_store_factory
- object_store_factory
- object_version_store
- object_version_store_prune_previous
- real_s3_version_store
- real_s3_version_store_dynamic_schema
- real_s3_sts_store_factory
- basic_arctic_library
- basic_arctic_client
- arctic_library* (many)
- basic_store* (many)
- arctic_client_no_lmdb
- arctic_client_v1
- arctic_client
- version_store_and_real_s3_basic_store_factory

The PR now defines 3 groups of tests which can be locally controlled:
- local storage tests - all tests with local storage (including simulated shared storages)
  - To constorl use env var:   LOCAL_STORAGE_TESTS_ENABLED=(0|1) 
- real AWS S3 tests - tests that are executed agains actual Amazaon storage. 
  - To constorl use env vars:   
    - ARCTICDB_PERSISTENT_STORAGE_TESTS=(0|1) 
    - ARCTICDB_STORAGE_AWS_S3=(0|1) 
- reals GCP tests - tests that are executed against actual Google cloud storage.
  - To constorl use env vars:   
    - ARCTICDB_PERSISTENT_STORAGE_TESTS=(0|1) 
    - ARCTICDB_STORAGE_GCP=(0|1) 

Has a new folder with tests (enduser) which purpose is to be "no wrapping fixture" tests to cover real user experience

There are also 2 new custom marks added to help manage tests:
- @pytest.mark.only_fixture_params([<fixture_param_names_list>], ) - this mark allows limitting the number of of parameters that the fixture will generate for this test only to specified in the parameters list of the mark. Useful for overriding general behavor of a general fixture and producing custom fixture experience for this test
- @pytest.mark.skip_fixture_params([<fixture_param_names_list>], ) - this mark allows skipping generation of specific value(s) of a fixture, Useful when there is a problem with that test and a specific fixture from list. 

Additional useful custom mark is @pytest,mark.bug_ids([<list of bug ids>]) - usful to associate an issue with a test - in case it produced bug or the test is because of specific bugs logged. This mark does not have efect over execution, only helps traceability and provides context, Such marks can and should stay in code unlike xfail marks

With authentication we have potentially unsafe situation to log accidently credential or other sensitive information.
Therefore 2 things need to be introduced:
  - Sanitizing message exception
  - Sanitizing logger
  
 Similar is the problem with pytest parameter - avoiding passing secrets in clear form could be done through masking

Changes to the test execution workflow:

Because tests now can be executed for 3 different types of storages, the checkbox for persistance tests is now a drop down allowing selection of:
  - tests without real storage
  - real AWS S3 storage tests only (no local storage tests)
  - real GCP storage tests (no local test)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/storage_fixtures/s3.py
+++ b/python/arcticdb/storage_fixtures/s3.py
@@ -210,30 +210,34 @@ class GcpS3Bucket(S3Bucket):
         bucket: str,
         native_config: Optional[NativeVariantStorage] = None,
     ):
-        StorageFixture.__init__(self)
-        self.factory = factory
-        self.bucket = bucket
-        self.native_config = native_config
+        if any(sub in factory.endpoint for sub in ["http:", "https:"])  :
+            super().__init__(factory, bucket, native_config=native_config)
+            self.arctic_uri = self.arctic_uri.replace("s3", "gcpxml", 1)
+        else: 
+            StorageFixture.__init__(self)
+            self.factory = factory
+            self.bucket = bucket
+            self.native_config = native_config
 
-        host, port = re.match(r"(?:gcpxml://)?([^:/]+)(?::(\d+))?", factory.endpoint).groups()
-        self.arctic_uri = f"gcpxml://{host}:{self.bucket}?"
+            host, port = re.match(r"(?:gcpxml://)?([^:/]+)(?::(\d+))?", factory.endpoint).groups()
+            self.arctic_uri = f"gcpxml://{host}:{self.bucket}?"
 
-        self.key = factory.default_key
+            self.key = factory.default_key
 
-        if factory.aws_auth == None or factory.aws_auth == AWSAuthMethod.DISABLED:
-            self.arctic_uri += f"access={self.key.id}&secret={self.key.secret}"
-        else:
-            self.arctic_uri += "aws_auth=default"
-        if port:
-            self.arctic_uri += f"&port={port}"
-        if factory.default_prefix:
-            self.arctic_uri += f"&path_prefix={factory.default_prefix}"
-        if factory.ssl:
-            self.arctic_uri += "&ssl=True"
-        if platform.system() == "Linux":
-            if factory.client_cert_file:
-                self.arctic_uri += f"&CA_cert_path={self.factory.client_cert_file}"
-            # client_cert_dir is skipped on purpose; It will be tested manually in other tests
+            if factory.aws_auth == None or factory.aws_auth == AWSAuthMethod.DISABLED:
+                self.arctic_uri += f"access={self.key.id}&secret={self.key.secret}"
+            else:
+                self.arctic_uri += "aws_auth=default"
+            if port:
+                self.arctic_uri += f"&port={port}"
+            if factory.default_prefix:
+                self.arctic_uri += f"&path_prefix={factory.default_prefix}"
+            if factory.ssl:
+                self.arctic_uri += "&ssl=True"
+            if platform.system() == "Linux":
+                if factory.client_cert_file:
+                    self.arctic_uri += f"&CA_cert_path={self.factory.client_cert_file}"
+                # client_cert_dir is skipped on purpose; It will be tested manually in other tests
 
     def create_test_cfg(self, lib_name: str) -> EnvironmentConfigsMap:
         cfg = EnvironmentConfigsMap()
@@ -249,7 +253,6 @@ class GcpS3Bucket(S3Bucket):
             with_prefix=with_prefix
         )  # client_cert_dir is skipped on purpose; It will be tested manually in other tests
         return cfg, self.native_config
-
 
 
 class BaseS3StorageFixtureFactory(StorageFixtureFactory):
--- a/python/arcticdb/storage_fixtures/s3.py
+++ b/python/arcticdb/storage_fixtures/s3.py
@@ -36,7 +36,7 @@ from .utils import (
     get_ca_cert_for_testing,
 )
 from arcticc.pb2.storage_pb2 import EnvironmentConfigsMap
-from arcticdb.version_store.helper import add_s3_library_to_env
+from arcticdb.version_store.helper import add_gcp_library_to_env, add_s3_library_to_env
 from arcticdb_ext.storage import AWSAuthMethod, NativeVariantStorage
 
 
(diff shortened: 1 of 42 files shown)


python/tests/conftest.py
```diff
@@ -308,8 +325,9 @@ def real_s3_storage(real_s3_storage_factory) -> Generator[S3Bucket, None, None]:
 
 
 @pytest.fixture(scope="session")
-def real_s3_library(real_s3_storage, lib_name) -> Library:
```
G-D-Petrov, line 311: This probably shouldn't be deleted

python/tests/conftest.py
```diff
@@ -288,6 +297,14 @@ def real_s3_storage_factory() -> BaseS3StorageFixtureFactory:
     )
 
 
+@pytest.fixture(scope="session")
+def real_gcp_storage_factory() -> BaseGCPStorageFixtureFactory:
```
G-D-Petrov, line 301: Might be good to make a shared_path=True equivalent of this

G-D-Petrov (changes requested):
There is a fixture called `local_object_store_factory`, which is a way to easily iterate over all local simulators for testing.
It might be nice to create something similar for the real storages (e.g. `real_object_store_factory` and maybe use it in place of where we have only real_s3 at the moment.

python/tests/conftest.py
grusev: mistake - thanks!

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -263,6 +305,40 @@ def cleanup_bucket(self, b: S3Bucket):
             b.slow_cleanup(failure_consequence="The following delete bucket call will also fail. ")
 
 
+class BaseGCPStorageFixtureFactory(StorageFixtureFactory):
```
G-D-Petrov, line 308: I think that you can just extend `BaseS3StorageFixtureFactory` here and only overwrite the `create_fixture` method

python/tests/integration/arcticdb/test_persistent_storage.py
```diff
@@ -46,44 +60,71 @@ def test_real_s3_storage_write(shared_persistent_arctic_client):
...
     # We have another test, that tests with "deeper" data frames
-    ac = persistent_arctic_client
-    ac.create_library("test_persistent_storage_read_write_many_syms")
-    lib = ac["test_persistent_storage_read_write_many_syms"]
+    ac: Arctic = persistent_arctic_client
+
+    lib_name = "test_persistent_storage_read_write_many_syms"
+    print(f"Storage type: {persistent_test_type()}")
+    ac.create_library(lib_name)
+    lib = ac[lib_name]
```
G-D-Petrov, line 127: We should probably make a persistent storage equivalent of the `arctic_library` fixture and use that here instead of creating /deleting the library manually in the test

python/tests/util/mark.py
```diff
@@ -36,28 +36,73 @@
...
 SLOW_TESTS_MARK = pytest.mark.skipif(
     FAST_TESTS_ONLY or DISABLE_SLOW_TESTS, reason="Skipping test as it takes a long time to run"
 )
 
-AZURE_TESTS_MARK = pytest.mark.skipif(FAST_TESTS_ONLY or MACOS_CONDA_BUILD, reason=_MACOS_CONDA_BUILD_SKIP_REASON)
+AZURE_TESTS_MARK = pytest.mark.skipif(FAST_TESTS_ONLY or MACOS_CONDA_BUILD or not LOCAL_STORAGE_TESTS_ENABLED, 
+                                      reason=_MACOS_CONDA_BUILD_SKIP_REASON)
 """Mark to skip all Azure tests when MACOS_CONDA_BUILD or ARCTICDB_FAST_TESTS_ONLY is set."""
 
 MONGO_TESTS_MARK = pytest.mark.skipif(
```
G-D-Petrov, line 53: Worth adding a comment that running locally against a real mongo process/service

G-D-Petrov (changes requested):

python/arcticdb/storage_fixtures/s3.py
grusev: will not be correct thing to do due to fact that they may an perhaps will have different methods long term. Boto library is not for management gcp only aws s3, thus at least method _boto is not proper for BaseGCPStorageFixtureFactory.

The only current common class is indeed StorageFixtureFactory

G-D-Petrov (approved):

python/pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+markers =
+    storage: marks a test as a test against real storage (deselect with '-m "not storage"')
+    authentication: marks a test for authentication group (deselect with '-m "not authentication"')
+    pipeline: Pipeline tests (deselect with '-m "not pipeline"')
+    skip_fixture_params: will instruct fixture that supports excluding fixture values, which values to be excluded
+    only_fixture_params: will instruct fixture supporting that to include only parameters from the list
```
alexowens90, line 7: `bug_ids` too?

python/pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+markers =
+    storage: marks a test as a test against real storage (deselect with '-m "not storage"')
```
alexowens90, line 3: This mark is needed against almost all of our tests, since the vast majority of our Python tests can be executed against a real storage if desired. Would it be better to invert it? So have a `no_storage` mark, and mark tests that should not be run against real storages that way? My gut feeling is that this will result in easier to read code. It will also be less error prone, as a dev adding a new test can forget to add the mark. If they forget to add the `storage` mark as implemented, that test will never run against real storages, but if they forget to add a `no_storage` mark, then that test will (pointlessly) run against real storages, so we won't lose coverage, we will just execute some unnecessary tests.

python/pytest.ini
grusev: actually I have to delete pytest.ini, this was initial attempt to find right place which was not successfull. pytest algorythm for pytest.ini is not optimal and is missied in some cases - so all marks have been moved to toml file. There seems to work for all cases and pytest finds them wherever is executed from

python/pytest.ini
grusev: actually only about 390 tests out of 1400 have fixtures that did had real_s3. I started from there - all s3 tests to be able to be executed aboinst gcp also. Pontentially the other 1k tests can also be executed but they do not have right fixtures to do so. And perhaps that is ok. Tests against real storages take much longer time. So those approx 400 do take 2h 30 min to execute with new improvements on the process from GeorgiP ... Which means it does not make sense to add many more tests to be executed against realstorages.

So having the storage mark is not only right but it is effective.

On the fact that storage mark can be missed - I can write simple check to check that in GitHub as a separate task, which would avoid inverting the storage mark.

alexowens90 (approved):

PR #2258, opened 2025-03-24
debug: Use verbose mode for build systems

jjerphan:
This is used to able to see the commands executed by make or ninja, or msbuild and diagnose issues.


jjerphan:
The issue I faced is only observed when `msbuild` is used: it seems that `CMAKE_BUILD_TYPE` is not properly translated to `/p:Configuration=Debug` for `msbuild`: is this a CMake and msbuild integration problem?

PR #2260, opened 2025-03-24
Simple API to remove several staged symbols at once.

poodlewars:
Private in NativeVersionStore for now so we don't need to give API guarantees till we know whether it's useful.

The proper "batch" API where we give back `Union[VersionedItem, DataError]` is nicer - and we'll have to do it if we promote to the Library API - but would need changes to all the storage backends.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_parallel.py
+++ b/python/tests/unit/arcticdb/version_store/test_parallel.py
@@ -70,42 +70,92 @@ def test_staging_doesnt_write_append_ref(lmdb_version_store_v1):
     assert not len(lib_tool.find_keys_for_symbol(KeyType.APPEND_REF, sym))
 
 
-def test_remove_incomplete(basic_store):
-    lib = basic_store
-    lib_tool = lib.library_tool()
-    assert lib_tool.find_keys(KeyType.APPEND_DATA) == []
-    assert lib.list_symbols_with_incomplete_data() == []
+@pytest.mark.parametrize("batch", (True, False))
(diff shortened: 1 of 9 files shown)


python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -70,42 +70,92 @@ def test_staging_doesnt_write_append_ref(lmdb_version_store_v1):
...
+    with config_context_multi({"Storage.DeleteBatchSize": batch_size, "S3Storage.DeleteBatchSize": 2 * batch_size}):
+        lib = basic_store
+        lib_tool = lib.library_tool()
+        assert lib_tool.find_keys(KeyType.APPEND_DATA) == []
+        assert lib.list_symbols_with_incomplete_data() == []
+
+        n_symbols = 20
+        n_to_delete = 9
+        syms = [f"sym_{i}" for i in range(n_symbols)]
+        other_syms = [f"other_prefix_{i}" for i in range(n_symbols)]
```
alexowens90, line 132: `common_prefix` is (almost always) going to be an empty string here, could also delete just a subset of `syms`?

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_parallel.py
poodlewars: We're always deleting symbols that start with `sym_` here, but the inverse point stands, I'll add a test that deletes symbols with no shared prefix.

PR #2266, opened 2025-03-25
Symbol sizes API

poodlewars:
Add Python APIs to get sizes of symbols, in a new `AdminTools` class. Add documentation for this feature to our website.

You can access the new tools with:

```
lib: Library
lib.admin_tools(): AdminTools
```

Refactor the existing symbol scanning APIs to a visitor pattern so they can all share as much of the implementation as possible.

Monday: 8560764974

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -226,7 +226,7 @@ folly::Future<folly::Unit> visit_object_sizes(
     KeyType type, const std::optional<StreamId>& stream_id_opt, storage::ObjectSizesVisitor visitor) override {
     std::string prefix;
     if (stream_id_opt) {
-        auto stream_id = *stream_id_opt;
+        const auto& stream_id = *stream_id_opt;
         prefix = std::holds_alternative<StringId>(stream_id) ? std::get<StringId>(stream_id) : std::string();
     }
 
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -222,28 +222,49 @@ void iterate_type(
     library_->iterate_type(type, func, prefix);
 }
 
-folly::Future<storage::ObjectSizes> get_object_sizes(KeyType type, const std::string& prefix) override {
+folly::Future<folly::Unit> visit_object_sizes(
+    KeyType type, const std::optional<StreamId>& stream_id_opt, storage::ObjectSizesVisitor visitor) override {
+    std::string prefix;
+    std::optional<std::function<bool(const VariantKey&)>> match_stream_id = std::nullopt;
+    if (stream_id_opt) {
+        auto stream_id = *stream_id_opt;
+        prefix = std::holds_alternative<StringId>(stream_id) ? std::get<StringId>(stream_id) : std::string();
+        match_stream_id = [stream_id](const VariantKey& k) { return variant_key_id(k) == stream_id; };
+    }
+
     if (library_->supports_object_size_calculation()) {
         // The library has native support for some kind of clever size calculation, so let it take over
-        return async::submit_io_task(ObjectSizesTask{type, prefix, library_});
+        return async::submit_io_task(VisitObjectSizesTask{type, prefix, std::move(visitor), library_});
     }
 
     // No native support for a clever size calculation, so just read keys and sum their sizes
-    auto counter = std::make_shared<std::atomic_uint64_t>(0);
-    auto bytes = std::make_shared<std::atomic_uint64_t>(0);
     KeySizeCalculators key_size_calculators;
-    iterate_type(type, [&key_size_calculators, &counter, &bytes](const VariantKey&& k) {
-        key_size_calculators.emplace_back(std::forward<const VariantKey>(k), [counter, bytes] (auto&& ks) {
-            counter->fetch_add(1);
-            auto key_seg = std::forward<decltype(ks)>(ks);
-            auto compressed_size = key_seg.segment().size();
-            bytes->fetch_add(compressed_size);
+    iterate_type(type, [&key_size_calculators, &match_stream_id, &visitor](const VariantKey&& k) {
+        key_size_calculators.emplace_back(std::forward<const VariantKey>(k), [visitor, match_stream_id] (auto&& key_seg) {
+            if (!match_stream_id || match_stream_id.value()(key_seg.variant_key())) {
+                auto compressed_size = key_seg.segment().size();
+                visitor(key_seg.variant_key(), compressed_size);
+            }
             return key_seg.variant_key();
         });
     }, prefix);
 
     read_ignoring_key_not_found(std::move(key_size_calculators));
-    return folly::makeFuture(storage::ObjectSizes{type, *counter, *bytes});
+    return folly::makeFuture();
+}
+
+folly::Future<std::shared_ptr<storage::ObjectSizes>> get_object_sizes(KeyType type, const std::optional<StreamId>& stream_id_opt) override {
+    auto counter = std::make_shared<std::atomic_uint64_t>(0);
+    auto bytes = std::make_shared<std::atomic_uint64_t>(0);
+    storage::ObjectSizesVisitor visitor = [counter, bytes](const VariantKey&, storage::CompressedSize size) {
+        counter->fetch_add(1);
+        bytes->fetch_add(size);
+    };
+
+    return visit_object_sizes(type, stream_id_opt, std::move(visitor))
+    .thenValueInline([counter, bytes, type](folly::Unit&&) {
+        return std::make_shared<storage::ObjectSizes>(type, *counter, *bytes);
+    });
 }
 
 bool scan_for_matching_key(
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -225,11 +225,9 @@ void iterate_type(
 folly::Future<folly::Unit> visit_object_sizes(
     KeyType type, const std::optional<StreamId>& stream_id_opt, storage::ObjectSizesVisitor visitor) override {
     std::string prefix;
-    std::optional<std::function<bool(const VariantKey&)>> match_stream_id = std::nullopt;
     if (stream_id_opt) {
         auto stream_id = *stream_id_opt;
         prefix = std::holds_alternative<StringId>(stream_id) ? std::get<StringId>(stream_id) : std::string();
-        match_stream_id = [stream_id](const VariantKey& k) { return variant_key_id(k) == stream_id; };
     }
 
     if (library_->supports_object_size_calculation()) {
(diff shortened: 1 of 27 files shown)


cpp/arcticdb/async/async_store.hpp
```diff
@@ -223,28 +223,49 @@ void iterate_type(
...
     KeySizeCalculators key_size_calculators;
-    iterate_type(type, [&key_size_calculators, &counter, &bytes](const VariantKey&& k) {
-        key_size_calculators.emplace_back(std::forward<const VariantKey>(k), [counter, bytes] (auto&& ks) {
-            counter->fetch_add(1);
-            auto key_seg = std::forward<decltype(ks)>(ks);
-            auto compressed_size = key_seg.segment().size();
-            bytes->fetch_add(compressed_size);
+    iterate_type(type, [&key_size_calculators, &match_stream_id, &visitor](const VariantKey&& k) {
+        key_size_calculators.emplace_back(std::forward<const VariantKey>(k), [visitor, match_stream_id] (auto&& key_seg) {
+            if (!match_stream_id || match_stream_id.value()(key_seg.variant_key())) {
```
G-D-Petrov, line 245: I think that calling this `match_stream_id_func` or something similar will make this a bit clearer when read in isolation.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1701,48 +1701,88 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
+    KeyType::SYMBOL_LIST,
+};
+
 std::vector<storage::ObjectSizes> LocalVersionedEngine::scan_object_sizes() {
     using ObjectSizes = storage::ObjectSizes;
-    std::vector<folly::Future<ObjectSizes>> sizes_futs;
-    foreach_key_type([&store=store(), &sizes=sizes_futs](KeyType key_type) {
-        sizes.push_back(store->get_object_sizes(key_type, ""));
-    });
+    std::vector<folly::Future<std::shared_ptr<ObjectSizes>>> sizes_futs;
```
G-D-Petrov, line 1720: nit: we can do a `sizes_futs.reserve(TYPES_FOR_SIZE_CALCULATION.size());` like is done below

python/tests/integration/arcticdb/version_store/test_symbol_sizes.py
```diff
@@ -224,3 +253,46 @@ def test_symbol_sizes_concurrent(reader_store, writer_store):
         reader.terminate()
 
     assert exceptions_in_reader.empty()
+
+
+@pytest.mark.parametrize("store", ["s3_storage", "nfs_backed_s3_storage"])
```
G-D-Petrov, line 258: nit: can we add the `real_s3` as well?

Also does gcp support object sizes?

G-D-Petrov (commented):
Might be good to add some documentation on how to use the new API, either in the docs or in the gh wiki

cpp/arcticdb/version/local_versioned_engine.cpp
vasil-pashov: If we know the size at compile time it would be better to use either std::array or std::pmr::vector as both won't allocate heap memory.

vasil-pashov (changes requested):

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: The array is a pain because the Future is not default constructible. This is all a premature optimisation, so I'll just remove the reserving.

cpp/arcticdb/async/async_store.hpp
```diff
@@ -223,28 +223,49 @@ void iterate_type(
...
-    auto bytes = std::make_shared<std::atomic_uint64_t>(0);
     KeySizeCalculators key_size_calculators;
-    iterate_type(type, [&key_size_calculators, &counter, &bytes](const VariantKey&& k) {
-        key_size_calculators.emplace_back(std::forward<const VariantKey>(k), [counter, bytes] (auto&& ks) {
-            counter->fetch_add(1);
-            auto key_seg = std::forward<decltype(ks)>(ks);
-            auto compressed_size = key_seg.segment().size();
-            bytes->fetch_add(compressed_size);
+    iterate_type(type, [&key_size_calculators, &match_stream_id, &visitor](const VariantKey&& k) {
+        key_size_calculators.emplace_back(std::forward<const VariantKey>(k), [visitor, match_stream_id] (auto&& key_seg) {
```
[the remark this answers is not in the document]
poodlewars: Yeah, because the futures need ownership of all the objects they consume

poodlewars:
> Might be good to add some documentation on how to use the new API, either in the docs or in the gh wiki

Good idea, I'll push a change to add to the API docs,

image

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1708,48 +1708,86 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
+    KeyType::VERSION_REF,
+    KeyType::VERSION,
+    KeyType::TABLE_INDEX,
+    KeyType::TABLE_DATA,
+    KeyType::APPEND_DATA,
+    KeyType::SNAPSHOT_REF,
+    KeyType::LOG,
+    KeyType::LOG_COMPACTED,
+    KeyType::SYMBOL_LIST,
+};
```
alexowens90, line 1723: I think this would be better as an exclusion list. That way devs don't have to know to update it when they add a new key type

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1708,48 +1708,86 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
+    return res;
+}
+
+static constexpr std::array<KeyType, 5> TYPES_FOR_SIZE_BY_STREAM_CALCULATION = {
+    KeyType::VERSION_REF,
+    KeyType::VERSION,
+    KeyType::TABLE_INDEX,
+    KeyType::TABLE_DATA,
+    KeyType::APPEND_DATA,
+};
```
alexowens90, line 1747: `MULTI_KEY` too

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1708,48 +1708,86 @@ timestamp LocalVersionedEngine::latest_timestamp(const std::string& symbol) {
...
     using ObjectSizes = storage::ObjectSizes;
-    std::vector<folly::Future<ObjectSizes>> sizes_futs;
-    foreach_key_type([&store=store(), &sizes=sizes_futs](KeyType key_type) {
-        sizes.push_back(store->get_object_sizes(key_type, ""));
-    });
+    std::vector<folly::Future<std::shared_ptr<ObjectSizes>>> sizes_futs;
 
-    return folly::collect(sizes_futs).via(&async::cpu_executor()).get();
+    for (const auto& key_type : TYPES_FOR_SIZE_CALCULATION) {
+        sizes_futs.push_back(store()->get_object_sizes(key_type, std::nullopt));
```
alexowens90, line 1730: Do we want to be doing this scan over all of the key types in parallel? on S3 this will be doing a lot of listing which could DOS the storage

docs/mkdocs/docs/api/admin_tools.md
alexowens90, line 1: Might be worth including a basic example, either here or in the "Guides" section

G-D-Petrov (approved):

alexowens90 (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: I'm not sure, I think this has to be an explicit decision every time - it's a tradeoff between giving complete information vs confusing the output with irrelevant things. And we need to document every key type that we add, so someone would have to modify at least the Python layer every time anyway.

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: I'm not too worried since there's a fixed bound on the number of parallel operations

docs/mkdocs/docs/api/admin_tools.md
poodlewars: Good idea! Will add this and a test

image

cpp/arcticdb/async/async_store.hpp
```diff
@@ -222,28 +222,47 @@ void iterate_type(
     library_->iterate_type(type, func, prefix);
 }
 
-folly::Future<storage::ObjectSizes> get_object_sizes(KeyType type, const std::string& prefix) override {
+folly::Future<folly::Unit> visit_object_sizes(
+    KeyType type, const std::optional<StreamId>& stream_id_opt, storage::ObjectSizesVisitor visitor) override {
+    std::string prefix;
+    if (stream_id_opt) {
+        auto stream_id = *stream_id_opt;
```
vasil-pashov, line 229: I think we can do
```suggestion
        const auto& stream_id = *stream_id_opt;
```
to avoid copying the stream id

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -505,14 +505,16 @@ bool do_iterate_type_impl(
...
     KeyBucketizer&& bucketizer,
-    const PrefixHandler& prefix_handler = default_prefix_handler(),
-    const std::string& prefix = std::string{}) {
+    const PrefixHandler& prefix_handler,
+    const std::string& prefix,
+    const ObjectSizesVisitor& visitor
+    ) {
     ARCTICDB_SAMPLE(S3StorageCalculateSizesForType, 0)
 
     auto path_info = calculate_path_info(root_folder, key_type, prefix_handler, prefix, std::move(bucketizer));
```
vasil-pashov, line 520: I know it's not part of the PR but it would be technically more correct to std::forward the bucketizer as it's a forwarding reference.

cpp/arcticdb/storage/storage.hpp
```diff
@@ -234,9 +251,10 @@ class Storage {
     // the predicate.
     virtual bool do_iterate_type_until_match(KeyType key_type, const IterateTypePredicate& visitor, const std::string & prefix) = 0;
 
-    [[nodiscard]] virtual ObjectSizes do_get_object_sizes([[maybe_unused]] KeyType key_type, [[maybe_unused]] const std::string& prefix) {
+    virtual void do_visit_object_sizes([[maybe_unused]] KeyType key_type, [[maybe_unused]] const std::string& prefix,
```
vasil-pashov, line 254: nit: if we remove the parameter names we can also omit `[[maybe_unused]]` which makes it a bit shorter.

vasil-pashov (approved):

PR #2268, opened 2025-03-26
Fix appending with compact incomplete to an empty dataframe

vasil-pashov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -34,7 +34,15 @@ namespace arcticdb::version_store {
 
 namespace ranges = std::ranges;
 
-void modify_descriptor(const std::shared_ptr<pipelines::PipelineContext>& pipeline_context, const ReadOptions& read_options) {
+struct ReadIncompletesFlags {
+    bool convert_int_to_float{false};
+    bool via_iteration{false};
+    bool sparsify{false};
+    bool dynamic_schema{false};
+    bool has_active_version{true};
+};    
+
+static void modify_descriptor(const std::shared_ptr<pipelines::PipelineContext>& pipeline_context, const ReadOptions& read_options) {
 
     if (opt_false(read_options.force_strings_to_object()) || opt_false(read_options.force_strings_to_fixed()))
         pipeline_context->orig_desc_ = pipeline_context->desc_;
@@ -949,7 +957,7 @@ void add_index_columns_to_query(const ReadQuery& read_query, const TimeseriesDes
 
         std::vector<std::string> index_columns_to_add;
         for(const auto& index_column : index_columns) {
-            if(std::find(std::begin(*read_query.columns), std::end(*read_query.columns), index_column) == std::end(*read_query.columns))
+            if(ranges::find(*read_query.columns, index_column) == std::end(*read_query.columns))
                 index_columns_to_add.emplace_back(index_column);
         }
         read_query.columns->insert(std::begin(*read_query.columns), std::begin(index_columns_to_add), std::end(index_columns_to_add));
@@ -969,27 +977,29 @@ FrameAndDescriptor read_index_impl(
     return read_segment_impl(store, version.key_);
 }
 
-std::optional<pipelines::index::IndexSegmentReader> get_index_segment_reader(
-    const std::shared_ptr<Store>& store,
+std::optional<index::IndexSegmentReader> get_index_segment_reader(
+    Store& store,
     const std::shared_ptr<PipelineContext>& pipeline_context,
     const VersionedItem& version_info) {
-    std::pair<entity::VariantKey, SegmentInMemory> index_key_seg;
-    try {
-        index_key_seg = store->read_sync(version_info.key_);
-    } catch (const std::exception& ex) {
-        ARCTICDB_DEBUG(log::version(), "Key not found from versioned item {}: {}", version_info.key_, ex.what());
-        throw storage::NoDataFoundException(fmt::format("When trying to read version {} of symbol `{}`, failed to read key {}: {}",
-                                                        version_info.version(),
-                                                        version_info.symbol(),
-                                                        version_info.key_,
-                                                        ex.what()));
-    }
+    std::pair<VariantKey, SegmentInMemory> index_key_seg = [&]() {
+        try {
+            return store.read_sync(version_info.key_);
+        } catch (const std::exception &ex) {
+            ARCTICDB_DEBUG(log::version(), "Key not found from versioned item {}: {}", version_info.key_, ex.what());
+            throw storage::NoDataFoundException(fmt::format(
+                "When trying to read version {} of symbol `{}`, failed to read key {}: {}",
+                version_info.version(),
+                version_info.symbol(),
+                version_info.key_,
+                ex.what()));
+        }
+    }();
 
     if (variant_key_type(index_key_seg.first) == KeyType::MULTI_KEY) {
-        pipeline_context->multi_key_ = index_key_seg.second;
+        pipeline_context->multi_key_ = std::move(index_key_seg.second);
         return std::nullopt;
     }
-    return std::make_optional<pipelines::index::IndexSegmentReader>(std::move(index_key_seg.second));
+    return std::make_optional<index::IndexSegmentReader>(std::move(index_key_seg.second));
 }
 
 void check_can_read_index_only_if_required(
@@ -1021,7 +1031,7 @@ void read_indexed_keys_to_pipeline(
         const VersionedItem& version_info,
         ReadQuery& read_query,
         const ReadOptions& read_options) {
-    auto maybe_reader = get_index_segment_reader(store, pipeline_context, version_info);
+    auto maybe_reader = get_index_segment_reader(*store, pipeline_context, version_info);
     if(!maybe_reader)
         return;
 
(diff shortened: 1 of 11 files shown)


python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -1577,3 +1577,44 @@ def test_single_segment(self, version_store, dtype, request, append):
         lib.compact_incomplete(sym, append=append, convert_int_to_float=True)
         expected = pd.DataFrame({"a": np.array([1, 2, 3], dtype=np.float64)}, index=pd.date_range(pd.Timestamp(0), periods=3, freq="ns"))
         assert_frame_equal(lib.read(sym).data, expected, check_dtype=True)
+
+class TestEmptyDataFrames:
```
vasil-pashov, line 1581: TODO: @vasil-pashov add same tests for sort merge

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1696,7 +1703,7 @@ VersionedItem sort_merge_impl(
     std::optional<SortedValue> previous_sorted_value;
     if(options.append_ && update_info.previous_index_key_.has_value()) {
         read_indexed_keys_to_pipeline(store, pipeline_context, *(update_info.previous_index_key_), *read_query, ReadOptions{});
-        if (!write_options.dynamic_schema) {
+        if (!write_options.dynamic_schema && !pipeline_context->slice_and_keys_.empty()) {
```
IvoDD, line 1706: It looks like `sort_merge` and `compact_incomplete` share the code for reading the indexed keys and incompletes into the pipeline and this check. Maybe we can abstract it out in a separate function?

It seems like we don't have a similar check for read. What would happen if we try to read column sliced symbol with `incomplete=True`? If it is problematic for read we should probably just include this check as part of `read_incompletes_to_pipeline`

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -1577,3 +1577,53 @@ def test_single_segment(self, version_store, dtype, request, append):
...
+    )
+    def test_appending_to_empty_with_differing_columns_fails(self, lmdb_version_store_v1, to_append):
+        lib = lmdb_version_store_v1
+        symbol = "symbol"
+        empty = pd.DataFrame({"a": np.array([], np.int64)}, index=pd.DatetimeIndex([]))
+        lib.write(symbol, empty)
+        lib.write(symbol, to_append, parallel=True)
+        with pytest.raises(SchemaException) as exception_info:
+            lib.compact_incomplete(symbol, append=True, convert_int_to_float=False)
+        assert "wrong_col" in str(exception_info.value)
```
IvoDD, line 1629: `pytest.raises` has an argument `match` so you can do something like `pytest.raises(SchemaException, match='wrong_col')`. Nice thing about it is that it gives a friendlier error message.

Can also use it above with a regex e.g. `pytest.raises(SchemaException, match=r'index.*my_initial_index.*my_new_index')` but this relies on knowing the order of the strings in the error message

vasil-pashov: That's cool I didn't know that. I'll use it in this case. But the one with multiple checks seems more readable the way it is so I plan to leave it unless there are any strong opinions on this.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1690,6 +1691,35 @@ std::optional<DeleteIncompleteKeysOnExit> get_delete_keys_on_failure(
         return std::nullopt;
 }
 
+static void check_compaction_slicing_policy_matches_disk(
```
IvoDD, line 1694: Nit: I think the main job of this function is to load the indexed keys not to do the check? What do you think about renaming to something like `read_indexed_keys_to_pipeline_for_compaction`?

IvoDD (approved):

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1053,22 +1063,19 @@ void read_indexed_keys_to_pipeline(
...
+static bool read_incompletes_to_pipeline(
     const std::shared_ptr<Store>& store,
     std::shared_ptr<PipelineContext>& pipeline_context,
     const ReadQuery& read_query,
     const ReadOptions& read_options,
-    bool convert_int_to_float,
-    bool via_iteration,
-    bool sparsify,
-    bool dynamic_schema) {
+    const ReadIncompletesFlags flags) {
```
alexowens90, line 1071: Deliberate pass by copy?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1683,6 +1691,35 @@ std::optional<DeleteIncompleteKeysOnExit> get_delete_keys_on_failure(
...
+    ReadQuery& read_query,
+    const ReadOptions& read_options
+) {
+    const bool append_to_existing = options.append_ && update_info.previous_index_key_.has_value();
+    if(append_to_existing) {
+        read_indexed_keys_to_pipeline(store, pipeline_context, *(update_info.previous_index_key_), read_query, read_options);
+        if (!write_options.dynamic_schema && !pipeline_context->slice_and_keys_.empty()) {
+            user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+                pipeline_context->slice_and_keys_.front().slice().columns() == pipeline_context->slice_and_keys_.back().slice().columns(),
+                "Appending using sort and finalize is not supported when existing data being appended to is column sliced."
```
alexowens90, line 1709: This error message would also be shown for `finalize_staged_data` on V2 and `compact_incompletes` on V1

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1683,6 +1691,35 @@ std::optional<DeleteIncompleteKeysOnExit> get_delete_keys_on_failure(
         return std::nullopt;
 }
 
+static void read_indexed_keys_and_validate_compaction_slicing_policy(
```
alexowens90, line 1694: This name suggests it should be two separate functions unless that's hard for some reason I can't see?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1691,30 +1728,26 @@ VersionedItem sort_merge_impl(
...
-        }
-        previous_sorted_value.emplace(pipeline_context->desc_->sorted());
-    }
+    read_indexed_keys_and_validate_compaction_slicing_policy(options, update_info, store, pipeline_context, write_options, read_query, ReadOptions{});
     const auto num_versioned_rows = pipeline_context->total_rows_;
-
+    const bool append_to_existing = options.append_ && update_info.previous_index_key_.has_value();
+    // Cache this before calling read_incompletes_to_pipeline as it changes the descripor
+    const std::optional<SortedValue> initial_index_sorted_status = append_to_existing ? std::optional{pipeline_context->desc_->sorted()} : std::nullopt;
+    const ReadIncompletesFlags read_incomplete_flags{
```
alexowens90, line 1738: Would this be more readable if `ReadIncompletesFlags` had a `CompactIncompleteOptions` member? Then wouldn't need updating if we add more flags

python/tests/unit/arcticdb/version_store/test_empty_writes.py
```diff
@@ -171,3 +171,29 @@ def test_entirely_empty_column(lmdb_version_store):
...
+        2. What should happen with multiindex data. The empty index has a different normalization metadata which does
+        not allow for multiple names. What should be done in this case? Should we keep all index names?
+    """
+    @pytest.mark.parametrize("index",[pd.DatetimeIndex([], name="my_empty_index"), pd.RangeIndex(0,0,1, name="my_empty_index")])
+    def test_single_index(self, lmdb_version_store_v1, index):
+        lib = lmdb_version_store_v1
+        df = pd.DataFrame({"col": []}, index=index)
+        lib.write("sym", df)
+        result_df = lib.read("sym").data
+        assert result_df.index.name == "my_empty_index"
```
alexowens90, line 191: I don't think we preserve the names of range indexes in general? So probably shouldn't in this case either, may cause other problems elsewhere

alexowens90 (changes requested):

cpp/arcticdb/version/version_core.cpp
vasil-pashov: Good point. I tried putting as much code in the function as possible and using it both in sort_merge and compact_incomletes. Using two functions would be better though.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: It's 5 bytes so it should be the same as const ref. Const ref is more future proof though, I don't mind changing it.

python/tests/unit/arcticdb/version_store/test_empty_writes.py
vasil-pashov: It looks like we store them
```python
import arcticdb
import pandas as pd

df = pd.DataFrame({"a": [1]}, index=pd.RangeIndex(1, 2, 1, name="my_index"))
lib.write("test", df)
assert lib.read("test").data.index.name == "my_index"
```

cpp/arcticdb/version/version_core.cpp
vasil-pashov: I wanted to split those as `read_incompletes_to_pipeline` is called from `read_frame_for_version` as well and this is not related to compact incompletes or sort merge.

alexowens90 (approved):

PR #2282, opened 2025-04-02
Use WinHTTP on Windows for the Azure SDK

jjerphan:
#### Reference Issues/PRs

See discussions in https://github.com/conda-forge/azure-core-cpp-feedstock/pull/23.

Have the Azure SDK on Windows only use WinHTTP instead of `libcurl`. 

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/storage/azure/azure_client_impl.cpp
```diff
@@ -32,16 +57,32 @@ container_client(BlobContainerClient::CreateFromConnectionString(conf.endpoint()
...
-            curl_transport_options.CAInfo = conf.ca_cert_path();
-        }
-        if (!conf.ca_cert_dir().empty()) {
-            curl_transport_options.CAPath = conf.ca_cert_dir();
-        }
-        client_options.Transport.Transport = std::make_shared<Azure::Core::Http::CurlTransport>(curl_transport_options);
+#if defined(_WIN32)
+    // On Windows, always use WinHTTP and ignore CA cert settings
+    if (!conf.ca_cert_path().empty() || !conf.ca_cert_dir().empty()) {
+        ARCTICDB_RUNTIME_DEBUG(log::storage(),
```
phoebusm, line 63: I prefer an exception here rather, similar to what we have done on https://github.com/man-group/ArcticDB/blob/957668278e138971e1b3b7df2359ba42a511f77c/python/arcticdb/adapters/azure_library_adapter.py#L61

jjerphan: Has `libcurl` on macOS shown to be not supporting specifications CA cert path or directory?

phoebusm: I don't think so but we don't have any test for it. Since the ca cert has caused so many platform-specific issues before, I would rather turn it off until it's proven to be ready

P.S. macOS uses `KeyChain` to centralize the cert management. It makes setting CA cert path/directory setting less useful IMO

cpp/arcticdb/storage/azure/azure_client_impl.cpp
```diff
@@ -30,18 +61,56 @@ Azure::Core::Context get_context(unsigned int request_timeout){
...
+    // On Linux, use libcurl with CA cert configuration
+    ARCTICDB_RUNTIME_DEBUG(log::storage(), "Using libcurl transport");
+    Azure::Core::Http::CurlTransportOptions curl_transport_options;
+    if (!conf.ca_cert_path().empty()) {
+        curl_transport_options.CAInfo = conf.ca_cert_path();
+    }
+    if (!conf.ca_cert_dir().empty()) {
+        curl_transport_options.CAPath = conf.ca_cert_dir();
+    }
+    client_options.Transport.Transport = std::make_shared<Azure::Core::Http::CurlTransport>(curl_transport_options);
```
jjerphan, line 111: This line has not changed. Is a new version of `azure-core-cpp` pulled via vcpkg?

cpp/arcticdb/storage/azure/azure_client_impl.cpp
```diff
@@ -5,15 +5,46 @@
...
+ *   - This provides better performance and integration with Windows security
+ * 
+ * - On macOS (__APPLE__):
+ *   - Always uses libcurl as the HTTP transport
+ *   - CA certificate settings are not supported
+ *   - Custom CA certificates must be installed via Keychain Access
+ * 
+ * - On Linux:
+ *   - Always uses libcurl as the HTTP transport
+ *   - Supports custom CA certificate configuration via ca_cert_path and ca_cert_dir
```
jjerphan, line 26: I think we need to add notes about the Linux and macOS builds on conda-forge since they use conda-forge's distribution of `libcurl` which uses specific CA certificates.

jjerphan: Done with dbdb38a.

poodlewars (commented):
What's the current behaviour on Windows? Is it using winhttp or libcurl at the moment? Do we need similar changes for the AWS SDK?

If this is making a change to the default transport being used, we also need some performance testing to show any changes before and after this change.

Have you done some end to end testing against a real Azure storage?

poodlewars:
> > What's the current behaviour on Windows? Is it using winhttp or libcurl at the moment?
> 
> Currently `libcurl` is used, but provided CA certificates path or directories are ignored.
> 
> > Do we need similar changes for the AWS SDK?
> 
> Yes, this was a question I asked privately and I think you would benefit having an identical handling of CA certificates for each SDK (but this is out of the scope of this PR).
> 
> > If this is making a change to the default transport being used, we also need some performance testing to show any changes before and after this change.
> 
> I also agree.
> 
> > Have you done some end to end testing against a real Azure storage?
> 
> I have not.
> 
> Note that this change is motivated by the conda-forge builds for Windows and by those discussions [conda-forge/azure-core-cpp-feedstock#23](https://github.com/conda-forge/azure-core-cpp-feedstock/pull/23). If distributing Windows builds on conda-forge is a priority, we can in a first time make this change for the conda-forge builds and keep the behavior unchanged for the PyPI wheels.

Thanks, I think we need the end to end testing and some performance testing before we can merge this

poodlewars (changes requested):
As discussed needs end to end and performance testing

jjerphan:
I think most of the time will be spent in networking rather than in ArcticDB delegating to WinHTTP or `libcurl`, but I might be wrong. I would like to verify that also.

How would you design an end to end and performance testing?

poodlewars:
> I think most of the time will be spent in networking rather than in ArcticDB delegating to WinHTTP or `libcurl`, but I might be wrong. I would like to verify that also.
> 
> How would you design an end to end and performance testing?

You could have a look at our suite of ASV tests for some inspiration, a lot of those run against real S3 storages. I think that if you cover reading, writing and deleting a dataframe with 10M rows and 100 columns (including some strings), and also listing (for example `list_versions` on 10k symbols) that would be good evidence.

jjerphan:
I have seen https://github.com/airspeed-velocity/asv/issues/1465, which needs fixing, yes.

jjerphan:
https://github.com/man-group/ArcticDB/pull/2327 fixed the failure with the benchmark workflow.

I think we need to test that this does not break anything unknowingly.

I still think that most of the time spent storing and loading the data will depend on the (impredictable) network latency, be it with `libcurl` or WinHTTP (for orders of magnitudes regarding latency see [this reminder](https://colin-scott.github.io/personal_website/research/interactive_latency.html)).

poodlewars (changes requested):
This needs some unit tests please.

Otherwise this looks really neat, thank you, and thanks for the careful error message.

I think you should also announce this change before merging in the ArcticDB #general slack channel to give users a chance to say if they are relying on the current support for specifying certs on Windows.

python/arcticdb/adapters/azure_library_adapter.py
```diff
@@ -59,13 +60,51 @@ def __init__(self, uri: str, encoding_version: EncodingVersion, *args, **kwargs)
             ]
         )
         self._container = self._query_params.Container
-        if platform.system() != "Linux" and (self._query_params.CA_cert_path or self._query_params.CA_cert_dir):
+
+        # Extract CA certificate settings directly from the URI to be robust to any quirks
+        # in how the query string is built (e.g. duplicated keys).
+        # We parse the full URI to get the last occurrence of each parameter (in case of duplicates).
+        def _get_param_from_uri(key: str) -> Tuple[str, bool]:
```
phoebusm: I am a bit confused by the purpose of it, as looks like the last occurrence has been taken from the connection string
``
In [6]: ac = adb.Arctic("azure://CA_cert_dir=first;DefaultEndpointsProtocol=https;AccountName=arcticdbalexstore;CA_cert_dir=second;Container=abc")

In [7]: ac._library_adapter._ca_cert_dir
Out[7]: 'second'
```

File: python/tests/integration/arcticdb/test_azure_transport.py [hunk 5]
```diff
@@ -0,0 +1,136 @@
...
+    # Verify that the transport is created successfully
+    assert azure_storage is not None
+
+    # Perform basic operations to verify transport works
+    lib.write("test_symbol", "test_data")
+    assert lib.read("test_symbol").data == "test_data"
+
+
+@pytest.mark.skipif(platform.system().lower() != "linux", reason="Test only runs on Linux")
+def test_azure_transport_linux_ca_cert_path(azurite_storage: AzureContainer):
```
phoebusm, line 87: It looks similar to `test_upgrade_script_dryrun_azure` + `test_azurite_ssl_verification`
Maybe you can extend those to cover yours? I don't have strong opinion regarding it though

python/tests/integration/arcticdb/test_azure_transport.py
```diff
@@ -0,0 +1,136 @@
...
+        lib.write("test_symbol", "test_data")
+        assert lib.read("test_symbol") == "test_data"
+    finally:
+        # Clean up the temporary CA cert file
+        if os.path.exists(ca_cert_path):
+            os.remove(ca_cert_path)
+
+
+@pytest.mark.skipif(platform.system().lower() != "linux", reason="Test only runs on Linux")
+def test_azure_transport_linux_ca_cert_dir(azurite_storage: AzureContainer):
```
phoebusm, line 116: https://github.com/man-group/ArcticDB/pull/2282/files#r2580881381

python/arcticdb/adapters/azure_library_adapter.py
jjerphan: In your opinion, should we stop at the first occurrence?

python/tests/integration/arcticdb/test_azure_transport.py
jjerphan: This test also asserts that the CA certificate path is correctly set in the Azure Storage configuration.

I guess we can move `test_upgrade_script_dryrun_azure` and `test_azurite_ssl_verification` here?

python/arcticdb/adapters/azure_library_adapter.py
phoebusm: I mean the current master commits already gets the last occurrence. Not sure about the purpose of `_get_param_from_uri`

jjerphan: Fixed with caca993.

phoebusm (approved):

IvoDD (commented):
LGTM. I've scheduled a [CI run](https://github.com/man-group/ArcticDB/actions/runs/19935929374) with the AZURE persistant storage tests enabled (they are not run on every PR because they are somewhat expensive to run). Once it finishes I'll approve

PR #2283, opened 2025-04-03
test: Skip tests using `pyarrow` if it is not available

jjerphan:

Upstream [patch present on the feedstock](https://github.com/conda-forge/arcticdb-feedstock/blob/main/recipe/patches/0001-Skip-pyarrow.patch).

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -732,6 +732,7 @@ def test_column_names_roundtrip_csv(lmdb_version_store, sym):
     not IS_PANDAS_TWO, reason="The full-support of pyarrow-backed pandas objects is pandas 2.0-specific."
 )
 def test_pyarrow_error(lmdb_version_store):
+    _ = pytest.importorskip("pyarrow")
```
phoebusm, line 735: Would it possible to move the skip tests in a similar way as in https://github.com/man-group/ArcticDB/blob/329bd8c648297d21649c47c08c8d6ab4a10822b4/python/tests/util/mark.py
We centralize the skip marks there so they could be managed more easily

phoebusm (approved):

poodlewars:
Why would we be running without pyarrow installed? We are heavily working on Arrow at the moment and having builds that skip any arrow tests are very undesirable

jjerphan:
For now, some conda-forge builds of ArcticDB (the ones which depends on `libevent==2.1.10`) cannot be installed with any builds of `arrow-cpp` and thus of `pyarrow` (because recent versions of projects on conda-forge aren't built with `libevent==2.1.10`).

Those cases are unusual, and mostly needed for the feedstock or if the test suite is run in environment using `libevent==2.1.10`. I would prefer not to add those checks, but it is required unfortunately (I do not see other way around for now, but if you do please mention them).

PR #2286, opened 2025-04-03
Upgrade gcc to 11.5.0

IvoDD:
This PR upgrades the manylinux image to use `gcc-11.5.0` and adds the vcpkg sparrow dependency.

We were using g++-10 for the manylinux2014 image. 

Discussions with QS lead to deciding that we should upgrade compiler to
gcc 11. But the version availaible in package manager (11.2) had issues
with sparrow. Thus we compile in the most recent version of gcc11 (11.5.0)
on top of the manylinux images.

Adding the sparrow dependency required updating the vcpkg baseline to pick up the new package.

The new baseline requires extra package version pinning to make sure everything is compatible with eachother. I've made sure to not upgrade any major sdks which already had pins. This PR only adds extra pinning to make sure everything compiles.

Successful CI run from this branch's workflow [here](https://github.com/man-group/ArcticDB/actions/runs/14790770615). It also uses the updated `arcticdb-dev` docker image for the c++ tests instead of the clang one

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -28,7 +28,7 @@
         - name: Setup build dependencies
           run: |
             sudo apt-get update
-            sudo apt-get install -y gcc-10 g++-10 make mono-complete libkrb5-dev libsasl2-dev
+            sudo apt-get install -y gcc-11 g++-11 make mono-complete libkrb5-dev libsasl2-dev
 
         - name: Setup VCPKG cache
           run: |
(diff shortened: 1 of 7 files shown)


IvoDD:
Upgrading to gcc 11.5 caused us to update `stdlibc++` which is a no go for manylinux wheel audit tool. Abandoning this PR and will reuse parts of it in another where sparrow 0.8.0 should work with gcc 11.2

PR #2292, opened 2025-04-04
First pass to remove dependency on Python in arcticdb_core

JohanMabille:
This PRs starts to remove inclusion of pybind11 headers in lower layers of arcticdb core.


JohanMabille:
Ok Let's close this one and open a new PR, the failure is totally unrelated;

PR #2298, opened 2025-04-07
Fix nfs storage tests and bugs in nfs storage

G-D-Petrov:
#### Reference Issues/PRs
Fixes 8873431775

A regression was introduced in https://github.com/man-group/ArcticDB/pull/2176, specifically the change in nfs_backed_storage.cpp, that resulted in the shard directory being calculated incorrectly for NFS-backed storages when deleting objects.
This manifested as failing to delete snapshots, which is particularly noticeable because they then still appear in the output of list_snapshots. This would likely have resulted in no data being deleted correctly from NFS-backed storages, but for other key types this would present as orphaned data rather than errors, as data being logically deleted is represented by a write operation to the version chain.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
+++ b/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
@@ -208,7 +208,7 @@ static std::string prefix_handler(const std::string& prefix, const std::string& 
 }
 
 // signature needs to match PrefixHandler in s3_storage.hpp
-static std::string iter_prefix_handler([[maybe_unused]] const std::string& prefix, const std::string& key_type_dir, [[maybe_unused]] const KeyDescriptor& key_descriptor, [[maybe_unused]] KeyType) {
+static std::string iter_prefix_handler(const std::string&, const std::string& key_type_dir, const KeyDescriptor&, KeyType) {
     // The prefix handler is not used for filtering (done in func below)
     // so we just return the key type dir
     return key_type_dir;
--- a/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
+++ b/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
@@ -208,16 +208,22 @@ static std::string prefix_handler(const std::string& prefix, const std::string& 
 }
 
 bool NfsBackedStorage::do_iterate_type_until_match(KeyType key_type, const IterateTypePredicate& visitor, const std::string& prefix) {
+    auto iter_prefix_handler = [] ([[maybe_unused]] const std::string& prefix, const std::string& key_type_dir, [[maybe_unused]] const KeyDescriptor& key_descriptor, [[maybe_unused]] KeyType) {
+        // The prefix handler is not used for filtering (done in func below)
+        // so we just return the key type dir
+        return key_type_dir;
+    };
+    
     const IterateTypePredicate func = [&v = visitor, prefix=prefix] (VariantKey&& k) {
         auto key = unencode_object_id(k);
-        if(prefix.empty() || variant_key_id(key) == StreamId{prefix}) {
+        if(prefix.empty() || variant_key_view(key).find(prefix) != std::string::npos) {
           return v(std::move(key));
         } else {
           return false;
         }
     };
 
-    return s3::detail::do_iterate_type_impl(key_type, func, root_folder_, bucket_name_, *s3_client_, NfsBucketizer{}, prefix_handler, prefix);
+    return s3::detail::do_iterate_type_impl(key_type, func, root_folder_, bucket_name_, *s3_client_, NfsBucketizer{}, iter_prefix_handler, prefix);
 }
 
 bool NfsBackedStorage::do_key_exists(const VariantKey& key) {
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
```diff
@@ -208,16 +208,22 @@ static std::string prefix_handler(const std::string& prefix, const std::string&
 }
 
 bool NfsBackedStorage::do_iterate_type_until_match(KeyType key_type, const IterateTypePredicate& visitor, const std::string& prefix) {
+    auto iter_prefix_handler = [] ([[maybe_unused]] const std::string& prefix, const std::string& key_type_dir, [[maybe_unused]] const KeyDescriptor& key_descriptor, [[maybe_unused]] KeyType) {
```
poodlewars, line 211: Do we have tests for this change? I assume they're how you discovered this issue in the first place?

cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
poodlewars, line 1: We should also test this class with some C++ unit tests I think - doesn't have to be this PR (I can do it when I'm back)

python/tests/conftest.py
```diff
@@ -209,7 +207,9 @@ def s3_ssl_disabled_storage_factory() -> Generator[MotoS3StorageFixtureFactory,
 
 @pytest.fixture(scope="session")
 def nfs_backed_s3_storage_factory() -> Generator[MotoNfsBackedS3StorageFixtureFactory, None, None]:
-    with MotoNfsBackedS3StorageFixtureFactory(use_ssl=False, ssl_test_support=False, bucket_versioning=False) as f:
```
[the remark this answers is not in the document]
G-D-Petrov: I think that adding the option to create a NFS storage through the Arctic client is still beneficial.
It allows us to tap into many integration tests e.g. the one that caught the filtering issue:
https://github.com/man-group/ArcticDB/blob/103ef9354fb4ce126fef3235eb4ca8d2a9693cd1/python/tests/unit/arcticdb/version_store/test_parallel.py#L128

cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
G-D-Petrov: There are some tests for it already e.g:
https://github.com/man-group/ArcticDB/blob/103ef9354fb4ce126fef3235eb4ca8d2a9693cd1/cpp/arcticdb/storage/test/test_s3_storage.cpp#L333

They are not extensive though

cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
G-D-Petrov: Yes, this was caught by https://github.com/man-group/ArcticDB/blob/103ef9354fb4ce126fef3235eb4ca8d2a9693cd1/python/tests/unit/arcticdb/version_store/test_parallel.py#L128

Maybe we have to make the same change for the object sizes - https://github.com/man-group/ArcticDB/pull/2298/files/4b02b8a14b5b769a357ae10eb57f92278f533cea#diff-9185760f0cca77749e6d44696c431b40b770864dd9d42e964408743fe7cdeebbR238

I guess that the assumption for both was that the "prefix" was always going to be a symbol name/StreamID but that is not really the case for `do_iterate_type_until_match`, not sure about `do_get_object_sizes`

python/arcticdb/adapters/s3_library_adapter.py
```diff
@@ -105,13 +107,21 @@ def __init__(self, uri: str, encoding_version: EncodingVersion, *args, **kwargs)
 
         self._ssl = self._query_params.ssl
 
+        print(self._query_params)
```
poodlewars, line 110: drop the print

python/tests/conftest.py
```diff
@@ -92,15 +92,13 @@
 
 @pytest.fixture()
 def sym(request: "pytest.FixtureRequest"):
-    return request.node.name + datetime.utcnow().strftime("%Y-%m-%dT%H_%M_%S_%f")
+    return request.node.name + datetime.utcnow().strftime("%Y-%m-%dT%H_%M_%S_%f") + f"{time.time_ns() % 1000:03d}"
```
poodlewars, line 95: This is milis but we already have `%f` for micros, so what's the point of this change?

G-D-Petrov: I was trying to get nanos resolution to avoid some flaky tests.
But Georgi R has added a uuid here so this change is no longer needed and I will just drop it

poodlewars:
Please squash on merge 🙏

poodlewars (approved):

cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
```diff
@@ -208,16 +208,22 @@ static std::string prefix_handler(const std::string& prefix, const std::string&
...
+        // The prefix handler is not used for filtering (done in func below)
+        // so we just return the key type dir
+        return key_type_dir;
+    };
+    
     const IterateTypePredicate func = [&v = visitor, prefix=prefix] (VariantKey&& k) {
         auto key = unencode_object_id(k);
-        if(prefix.empty() || variant_key_id(key) == StreamId{prefix}) {
+        if(prefix.empty() || variant_key_view(key).find(prefix) != std::string::npos) {
```
vasil-pashov, line 219: I think a comment here would be helpful as people often forget that we're handling these path's differently.

PR #2302, opened 2025-04-08
Bugfix 8873431775: Fix NFS-backed storage sharding when deleting objects in 5.3.3

alexowens90:
Reference Issues/PRs
Fixes 8873431775 on 5.3.3

What does this implement or fix?
A regression was introduced in https://github.com/man-group/ArcticDB/pull/2176, specifically the change in nfs_backed_storage.cpp, that resulted in the shard directory being calculated incorrectly for NFS-backed storages when deleting objects.
This manifested as failing to delete snapshots, which is particularly noticeable because they then still appear in the output of list_snapshots. This would likely have resulted in no data being deleted correctly from NFS-backed storages, but for other key types this would present as orphaned data rather than errors, as data being logically deleted is represented by a write operation to the version chain.

### DIFF
--- a/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
+++ b/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
@@ -184,7 +184,7 @@ KeySegmentPair NfsBackedStorage::do_read(VariantKey&& variant_key, ReadKeyOpts o
 
 void NfsBackedStorage::do_remove(VariantKey&& variant_key, RemoveOpts) {
     auto enc = encode_object_id(variant_key);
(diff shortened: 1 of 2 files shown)


G-D-Petrov (approved):

vasil-pashov (approved):

willdealtry (approved):

alexowens90:
5.3.3 branch is still on master, and would look the same as head of master after the corresponding fixing the same PR on master is merged, so will just fast-forward the 5.3.3 branch to the head of master after https://github.com/man-group/ArcticDB/pull/2301 is merged

PR #2304, opened 2025-04-08
Query Stat framework v3

phoebusm:
New query stat implemenation which its schema is static
The feature of linking arcticdb API calls to storage operations has been dropped. Now only storage operation stats will be logged. Therefore the schema of the stats is hardcoded and allow the summation of stats is logged, one statical object with numerous atomic ints is enough to do the job.
No fancy map nor modification of folly executor.

#### Any other comments?
Sample output:
```
{ // Stats
        "SYMBOL_LIST":  // std::array<std::array<OpStats, NUMBER_OF_TASK_TYPES>, NUMBER_OF_KEYS>
         { 
            "storage_ops": {
                "S3_ListObjectsV2": 
                { // OpStats
                    "result_count": 1,
                    "total_time_ms": 34
                }
            }
        }
    }
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -17,6 +17,7 @@
 #include <arcticdb/util/exponential_backoff.hpp>
 #include <arcticdb/util/configs_map.hpp>
 #include <arcticdb/util/composite.hpp>
+#include <arcticdb/toolbox/query_stats.hpp>
 
 #include <aws/s3/model/GetObjectRequest.h>
 #include <aws/s3/model/PutObjectRequest.h>
@@ -466,12 +467,12 @@ bool do_iterate_type_impl(
 
     auto continuation_token = std::optional<std::string>();
     do {
+        auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_ListObjectsV2);
         auto list_objects_result = s3_client.list_objects(path_info.key_prefix_, bucket_name, continuation_token);
         if (list_objects_result.is_success()) {
             auto& output = list_objects_result.get_output();
 
             ARCTICDB_RUNTIME_DEBUG(log::storage(), "Received object list");
-
             for (auto& s3_object_name : output.s3_object_names) {
                 auto key = s3_object_name.substr(path_info.path_to_key_size_);
                 ARCTICDB_TRACE(log::version(), "Got object_list: {}, key: {}", s3_object_name, key);
--- a/cpp/arcticdb/toolbox/python_bindings.cpp
+++ b/cpp/arcticdb/toolbox/python_bindings.cpp
@@ -10,12 +10,13 @@
 #include <arcticdb/python/adapt_read_dataframe.hpp>
 #include <arcticdb/storage/library.hpp>
 #include <arcticdb/storage/s3/s3_storage_tool.hpp>
-#include <arcticdb/toolbox/library_tool.hpp>
-#include <arcticdb/util/memory_tracing.hpp>
 #include <arcticdb/version/symbol_list.hpp>
+#include <arcticdb/util/memory_tracing.hpp>
 #include <arcticdb/util/pybind_mutex.hpp>
 #include <arcticdb/util/storage_lock.hpp>
 #include <arcticdb/util/reliable_storage_lock.hpp>
+#include <arcticdb/toolbox/library_tool.hpp>
+#include <arcticdb/toolbox/query_stats.hpp>
 #include <arcticdb/toolbox/storage_mover.hpp>
 
 namespace arcticdb::toolbox::apy {
@@ -151,6 +152,27 @@ void register_bindings(py::module &m, py::exception<arcticdb::ArcticException>& 
             .def("lock_timeout", &StorageLockWrapper::lock_timeout)
             .def("try_lock", &StorageLockWrapper::try_lock)
             ;
-}
 
+    using namespace arcticdb::query_stats;
+    auto query_stats_module = tools.def_submodule("query_stats", "Query stats functionality");
+    
+    py::class_<QueryStats::OperationStatsOutput>(query_stats_module, "OperationStatsOutput")
+        .def_readonly("stats", &QueryStats::OperationStatsOutput::stats_);
+
+    query_stats_module.def("reset_stats", []() { 
+        QueryStats::instance()->reset_stats(); 
+    });
+    query_stats_module.def("enable", []() { 
+        QueryStats::instance()->enable(); 
+    });
+    query_stats_module.def("disable", []() { 
+        QueryStats::instance()->disable(); 
+    });
+    query_stats_module.def("is_enabled", []() { 
+        return QueryStats::instance()->is_enabled(); 
+    });
+    query_stats_module.def("get_stats", [](){ 
+        return QueryStats::instance()->get_stats(); 
+    });
+}
 } // namespace arcticdb::toolbox::apy
--- a/cpp/arcticdb/toolbox/query_stats.hpp
+++ b/cpp/arcticdb/toolbox/query_stats.hpp
@@ -1,0 +1,98 @@
+/* Copyright 2023 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <atomic>
+#include <type_traits>
+#include <string>
+#include <chrono>
+#include <array>
+#include <memory>
+#include <folly/ThreadCachedInt.h>
+
+#include <arcticdb/entity/key.hpp>
+#include <arcticdb/util/constants.hpp>
+
+namespace arcticdb::query_stats{
+enum class TaskType : size_t {
+    S3_ListObjectsV2 = 0,
+    END
+};
+
+enum class StatType : size_t {
+    TOTAL_TIME_MS = 0,
+    COUNT,
+    END
+};
+
+class RAIIAddTime {
+public:
+    RAIIAddTime(folly::ThreadCachedInt<timestamp>& time_var);
+    ~RAIIAddTime();
+private:
+    folly::ThreadCachedInt<timestamp>& time_var_;
+    std::chrono::time_point<std::chrono::steady_clock> start_;
+};
+
+// Example output:
+// {
+//     "SYMBOL_LIST": { <- STATS_BY_KEY_TYPE
+//         "storage_ops": { <- STATS_BY_OP_TYPE
+//             "S3_ListObjectsV2": { <- OperationStats::stats_
+//                 "total_time_ms": 83,
+//                 "count": 3
+//             }
+//         }
+//     }
+// }
+
+
+
+
+
+class QueryStats {
+public:
+    struct OperationStats{
+        folly::ThreadCachedInt<timestamp> total_time_ns_;
+        folly::ThreadCachedInt<uint32_t> count_;
+        void reset_stats(){
+            total_time_ns_.set(0);
+            count_.set(0);
+        }
+        OperationStats(){
+            reset_stats(); 
+        } 
+    };
+    struct OperationStatsOutput {
+        std::map<std::string, uint32_t> stats_;
+    };
+    using QueryStatsOutput = std::map<std::string, std::map<std::string, std::map<std::string, OperationStatsOutput>>>;
+    using STATS_BY_OP_TYPE = std::array<OperationStats, static_cast<size_t>(TaskType::END)>;
+    using STATS_BY_KEY_TYPE = std::array<STATS_BY_OP_TYPE, static_cast<size_t>(entity::KeyType::UNDEFINED)>;
+
+    ARCTICDB_NO_MOVE_OR_COPY(QueryStats);
+    void reset_stats();
+    static std::shared_ptr<QueryStats> instance();
+    void enable();
+    void disable();
+    bool is_enabled() const;
+    void add(entity::KeyType key_type, TaskType task_type, uint32_t value);
+    [[nodiscard]] std::optional<RAIIAddTime> add_task_count_and_time(entity::KeyType key_type, TaskType task_type);
+    QueryStatsOutput get_stats() const;
+    QueryStats();
+
+    STATS_BY_KEY_TYPE stats_by_key_type_;
+private:
+    static std::once_flag init_flag_;
+    static std::shared_ptr<QueryStats> instance_;
+    std::atomic<bool> is_enabled_ = false;
+};
+
+void add(entity::KeyType key_type, TaskType task_type, uint32_t value);
+[[nodiscard]] std::optional<RAIIAddTime> add_task_count_and_time(entity::KeyType key_type, TaskType task_type);
+}
--- a/python/arcticdb/toolbox/query_stats.py
+++ b/python/arcticdb/toolbox/query_stats.py
@@ -1,0 +1,124 @@
+"""
+Query Statistics API for ArcticDB.
+
+This module provides utilities for collecting query statistics.
+
+.. warning::
+    This API is unstable and not governed by ArcticDB's semantic versioning.
+    It may change or be removed in future versions without notice.
+"""
+
+from contextlib import contextmanager
+from typing import Dict, Any, Iterator
+import arcticdb_ext.tools.query_stats as qs
+from arcticdb_ext.exceptions import UserInputException
+
+
+@contextmanager
+def query_stats() -> Iterator[None]:
+    """
+    Context manager for enabling query statistics collection within a specific scope.
+    
+    When entering the context, query statistics collection is enabled.
+    When exiting the context, it is automatically disabled.
+    
+    Raises:
+        UserInputException: If query stats is already enabled.
+        
+    Example:
+        >>> with query_stats():
+        ...     store.list_symbols()
+    
+    .. warning::
+        This API is unstable and not governed by semantic versioning.
+    """
+    if qs.is_enabled():
+        raise UserInputException("Query Stats is already enabled")
+    enable()
+    yield
+    disable()
+    
+
+def get_query_stats() -> Dict[str, Any]:
+    """
+    Get collected query statistics.
+    
+    Returns:
+        Dict[str, Any]: A dictionary containing statistics organized by key type,
+            operation group, and task type. Each task contains timing and count information.
+            
+    Example output:
+    {
+        "SYMBOL_LIST": {
+            "storage_ops": {
+                "S3_ListObjectsV2": {
+                    "total_time_ms": 83,
+                    "count": 3
+                }
+            }
+        },
+        "VERSION_REF": {
+            "storage_ops": {
+                "S3_ListObjectsV2": {
+                    "total_time_ms": 21,
+                    "count": 1
+                }
+            }
+        }
+    }
+    
+    .. warning::
+        This API is unstable and not governed by semantic versioning.
+    """
+    raw_stats = qs.get_stats()
+    result = {}
+    
+    for key_type, key_type_data in raw_stats.items():
+        result[key_type] = {}
+            
+        for op_group, op_group_data in key_type_data.items():
+            result[key_type][op_group] = {}
+            for task_type, task_data in op_group_data.items():                
+                result[key_type][op_group][task_type] = task_data.stats
+                
+    return result
+
+
+def reset_stats() -> None:
+    """
+    Reset all collected query statistics.
+    
+    This clears all statistics that have been collected since enabling
+    the query statistics collection.
+    
+    .. warning::
+        This API is unstable and not governed by semantic versioning.
+    """
+    qs.reset_stats()
+
+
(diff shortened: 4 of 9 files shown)


cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,94 @@
...
+    RAIIAddTime(std::atomic<uint64_t>& time_var);
+    ~RAIIAddTime();
+private:
+    std::atomic<uint64_t>& time_var_;
+    std::chrono::time_point<std::chrono::steady_clock> start_;
+};
+} //util::query_stats
+} //arcticdb
+
+#define QUERY_STATS_ADD(stat_name, value) \
```
poodlewars, line 77: I don't think these need to be macros except for the timer

cpp/arcticdb/util/query_stats.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+QueryStats QueryStats::instance_;
+
+QueryStats& QueryStats::instance() {
+    return instance_;
+}
+
+void QueryStats::reset_stats() {
+    check(!async::TaskScheduler::instance()->tasks_pending(), "Folly tasks are still running");
```
poodlewars, line 54: Doesn't this mean we'll blow up if there are multiple Python threads running? I don't think we should let the stats measurements interrupt a program

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,94 @@
...
+};
+namespace util::query_stats {
+enum class TaskType : size_t {
+    S3_ListObjectsV2 = 0
+};
+
+constexpr const size_t NUMBER_OF_KEYS = 29;
+constexpr const size_t NUMBER_OF_TASK_TYPES = 1;
+
+class OpStats{
```
poodlewars, line 30: I don't quite follow what `OpStats` vs `Stats` are and what the differences are between them. I think it might be clearer to lay these out in explicit structs rather than in a `std::array`.

python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -0,0 +1,67 @@
+import arcticdb.toolbox.query_stats as qs
+
+def verify_list_symbool_stats(count):
+    stats = qs.get_query_stats()
+    # """
+    # Sample output:
+    # {
+    #     "SYMBOL_LIST": {
+    #         "storage_ops": {
```
poodlewars, line 9: I think we can probably drop this level of testing (`storage_ops`) and either have everything grouped under the key type, or not have any grouping at all and just lay everything out in a flat structure

cpp/arcticdb/util/query_stats.hpp
phoebusm: Yea my idea was as the timer needs macro anyway so I have kept everything in macro to give a more consistent interface to user

poodlewars: We already have the `ScopedTimer`

cpp/arcticdb/util/query_stats.hpp
phoebusm: I have kept the current design per discussed but classes have been renamed to make it more comprehensible

cpp/arcticdb/util/query_stats.hpp
phoebusm: Macros have been removed

python/tests/integration/toolbox/test_query_stats.py
phoebusm: The structure will remain as it is but the naming of underlying data structure will be improved

phoebusm: That's possible by adding new variable in:
```
class StatsHolder{
private:
    std::array<std::array<OperationStats, static_cast<size_t>(TaskType::END)>, static_cast<size_t>(entity::KeyType::UNDEFINED)> op_stats_;
    folly::ThreadCachedInt<uint32_t> new_stat_;
```

The output could look like this:
```
{
    "keys": {
        "SYMBOL_LIST": {
            "storage_ops": {
                "S3_ListObjectsV2": {
                    "result_count": 2,
                    "total_time_ms": 83,
                    "count": 3
                }
            }
        }
    },
    "new_stat": 1,
}
```

phoebusm: But agree that if the current design isn't favored, it could be easily.

python/arcticdb/toolbox/query_stats.py
```diff
@@ -0,0 +1,80 @@
+from contextlib import contextmanager
+from typing import Dict, Any, Iterator
+import arcticdb_ext.tools.query_stats as qs
+from arcticdb_ext.exceptions import UserInputException
+from arcticdb_ext.storage import KeyType
+
+
+@contextmanager
+def query_stats() -> Iterator[None]:
```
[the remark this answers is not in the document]
phoebusm: Spoken to Alex and we agree things stay as it is and await user input for iteration

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+enum class StatType : size_t {
+    RESULT_COUNT = 0,
+    TOTAL_TIME_MS,
+    COUNT,
+    END
+};
+
+
+class OperationStats{
```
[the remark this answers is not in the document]
alexowens90: This also implies `O(num_task_types * num_key_types^2)` thread cached ints, all templated on the same underlying integer type. This has performance implications:
https://github.com/facebook/folly/blob/main/folly/ThreadCachedInt.h#L30-L33
so we may want to use tags to split them out.

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,107 @@
...
+    bool is_enabled() const;
+    const StatsHolder& get_stats() const;
+    QueryStats(const QueryStats&) = delete;
+    QueryStats() = default;
+
+    StatsHolder stats_;
+private:
+
+    static QueryStats instance_;
+    bool is_enabled_ = false;
```
alexowens90, line 105: Should this be atomic? What happens if multiple Python threads are calling `enable/disable` simultaneously?

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include <arcticdb/entity/key.hpp>
+
+namespace arcticdb::util::query_stats{
+enum class TaskType : size_t {
+    S3_ListObjectsV2 = 0,
+    END
+};
+
+enum class StatType : size_t {
+    RESULT_COUNT = 0,
```
alexowens90, line 26: This seems to be a bit of a niche stat to include on day 1

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -443,12 +444,14 @@ bool do_iterate_type_impl(
 
     auto continuation_token = std::optional<std::string>();
     do {
+        util::query_stats::add(key_type, util::query_stats::TaskType::S3_ListObjectsV2, util::query_stats::StatType::COUNT, 1);
+        auto query_stat_operation_time = util::query_stats::add_time(key_type, util::query_stats::TaskType::S3_ListObjectsV2, util::query_stats::StatType::TOTAL_TIME_MS);
```
alexowens90, line 448: I'm not sure RAII is the right pattern here. If an exception is thrown, either on lines 449 or 480, the stat will still be added

cpp/arcticdb/toolbox/python_bindings.cpp
```diff
@@ -148,6 +150,55 @@ void register_bindings(py::module &m, py::exception<arcticdb::ArcticException>&
...
+    query_stats_module.def("enable", []() { 
+        QueryStats::instance().enable(); 
+    });
+    query_stats_module.def("disable", []() { 
+        QueryStats::instance().disable(); 
+    });
+    query_stats_module.def("is_enabled", []() { 
+        return QueryStats::instance().is_enabled(); 
+    });
+    query_stats_module.def("get_stats", []() -> const StatsHolder& {
```
alexowens90, line 169: Returning a reference to a C++ object that can still be being mutated seems odd. It would mean that:
```
s = get_stats()
print(s)
lib.read(sym)
print(s)
```
would show two different outputs from the prints. This might just be me, but I would expected `get_stats` to give a frozen-in-time view of the stats

alexowens90: The more I think about it, the weirder this is. As stats can be being updated during all of the processing in `query_stats.py`, you could end up displaying (for example) the number of index keys read for one call to `read`, but the number of data keys read from two calls to `read`

python/arcticdb/toolbox/query_stats.py
alexowens90, line 1: This is quite a lot of processing in Python to produce nested maps from strings to strings/ints. Is there a reason not to push this down to the C++ layer? Then it would be trivially cross-language

alexowens90 (changes requested):

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,107 @@
...
+    OperationStats();
+    
+    friend void add(entity::KeyType key_type, TaskType task_type, StatType stat_type, uint64_t value);
+    friend std::optional<RAIIAddTime> add_time(entity::KeyType key_type, TaskType task_type, StatType stat_type);
+};
+
+
+using STATS_BY_OP_TYPE = std::array<OperationStats, static_cast<size_t>(TaskType::END)>;
+using STATS_BY_KEY_TYPE = std::array<STATS_BY_OP_TYPE, static_cast<size_t>(entity::KeyType::UNDEFINED)>;
+class StatsHolder{
```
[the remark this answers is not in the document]
phoebusm: I am fine in both ways. Just intended to separate the data and function to make the implementation slightly more readable

cpp/arcticdb/toolbox/python_bindings.cpp
phoebusm: That's a good point. Let me fix it

cpp/arcticdb/util/query_stats.hpp
phoebusm: e.g. if the stat shows high of `SYMBOL_LIST`'s list_object `RESULT_COUNT` vs `COUNT`, it suggests the symbol list compaction hasn't been done for a while, which leads to slow `list_symbols` call

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include <type_traits>
+#include <string>
+#include <chrono>
+#include <array>
+#include <folly/ThreadCachedInt.h>
+
+#include <arcticdb/entity/key.hpp>
+
+namespace arcticdb::util::query_stats{
+enum class TaskType : size_t {
```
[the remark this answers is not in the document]
phoebusm: CPU tasks and IO tasks will share the same enum class.
I am thinking a big enum class for all backends and CPU tasks. What's being measured and counted are similar across all tasks (for now) so I don't think breaking the big enum class down into smaller one will be meaningful here (except shredding the enum class)
If categorization is needed, doing it at the output (python) layer maybe a better choice

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,107 @@
...
+    COUNT,
+    END
+};
+
+class RAIIAddTime {
+public:
+    RAIIAddTime(folly::ThreadCachedInt<uint32_t>& time_var);
+    ~RAIIAddTime();
+private:
+    folly::ThreadCachedInt<uint32_t>& time_var_;
```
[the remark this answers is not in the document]
phoebusm: That's a good suggestion but that will make `TOTAL_TIME_MS` has a datatype difference from the rest. Let's discuss this offline

cpp/arcticdb/util/query_stats.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void QueryStats::reset_stats() {
+    stats_.reset_stats();
+}
+
+void QueryStats::enable() {
+    is_enabled_ = true;
+}
+
+void QueryStats::disable() {
```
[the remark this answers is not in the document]
phoebusm: I think it's a bit weird to couple two functions together, unless query stat will "memory-leak" if leaving ON

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,107 @@
...
+    S3_ListObjectsV2 = 0,
+    END
+};
+
+enum class StatType : size_t {
+    RESULT_COUNT = 0,
+    TOTAL_TIME_MS,
+    COUNT,
+    END
+};
```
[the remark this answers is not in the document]
phoebusm: I'll remove the parameter. There are only a few enum in the class. Further shredding maybe a bit too much

cpp/arcticdb/util/query_stats.hpp
phoebusm: Even if it's atomic, the ultimate value of it is still indeterministic in your case
Since the support of python multithread has been taken away, I'll make it clear in the manual in later PR

cpp/arcticdb/storage/s3/detail-inl.hpp
phoebusm: I think if exception is thrown in either of the lines, it will be unexpected error and exception will be thrown to the caller. At that point, I think the unexepcted exception is much more pressing issue than the stats. 
I will make it clear in the manual that if unexpected exception happens, stat won't be reliable.
Such design is a compromised for the sake of user friendliness.
Still happy to change if accuracy is preferred

cpp/arcticdb/util/query_stats.hpp
```diff
@@ -0,0 +1,107 @@
...
+    void disable();
+    bool is_enabled() const;
+    const StatsHolder& get_stats() const;
+    QueryStats(const QueryStats&) = delete;
+    QueryStats() = default;
+
+    StatsHolder stats_;
+private:
+
+    static QueryStats instance_;
```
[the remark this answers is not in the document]
phoebusm: That's a very good point. Will amend

python/arcticdb/toolbox/query_stats.py
phoebusm: I have moved major part of the process into c++ for the sake of other changes

cpp/arcticdb/storage/s3/detail-inl.hpp
alexowens90: There are cases where we use `ExponentialBackoff` to retry operations on certain failure modes and we use exceptions as control flow in these limited cases. In that case, both the failed call and the successful one would appear in the stats, which I think would be extremely counterintuitive.

cpp/arcticdb/util/query_stats.hpp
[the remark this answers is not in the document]
alexowens90: `TOTAL_TIME_MS` is just about what we show to the user, it shouldn't be coupled to the implementation of how we're recording the times.

cpp/arcticdb/util/query_stats.hpp
alexowens90: We haven't really removed Python multithreading support, just thread-specific stats. So `enable()/disable()` can be being called at the same time as `add_time` is calling `is_enabled()`

cpp/arcticdb/util/query_stats.hpp
alexowens90: But you can tell that just from `COUNT`? If it is 1, there are <1000 symbol list keys, if it is 2, there are between 1001 and 2000. I'd be in favour of keeping it as simple as possible on day one, and adding later if we find a use case for it

cpp/arcticdb/util/query_stats.hpp
phoebusm: Discussed with Alex. Will switch to atomic instead

cpp/arcticdb/toolbox/python_bindings.cpp
```diff
@@ -6,17 +6,20 @@
...
-#include <arcticdb/util/memory_tracing.hpp>
 #include <arcticdb/version/symbol_list.hpp>
+#include <arcticdb/util/memory_tracing.hpp>
 #include <arcticdb/util/pybind_mutex.hpp>
 #include <arcticdb/util/storage_lock.hpp>
 #include <arcticdb/util/reliable_storage_lock.hpp>
+#include <arcticdb/toolbox/library_tool.hpp>
+#include <arcticdb/toolbox/query_stats.hpp>
 #include <arcticdb/toolbox/storage_mover.hpp>
+#include <arcticdb/async/task_scheduler.hpp>
```
alexowens90, line 22: Are any of these new includes needed apart from `query_stats.hpp`?

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -443,12 +444,13 @@ bool do_iterate_type_impl(
 
     auto continuation_token = std::optional<std::string>();
     do {
+        util::query_stats::add(key_type, util::query_stats::TaskType::S3_ListObjectsV2, 1);
+        auto query_stat_operation_time = util::query_stats::add_time(key_type, util::query_stats::TaskType::S3_ListObjectsV2);
```
alexowens90, line 448: Suspect it will be a very common operation to call both `add` and `add_time` in the same place, could add a wrapper method that does both?

cpp/arcticdb/toolbox/query_stats.hpp
```diff
@@ -0,0 +1,106 @@
...
+    friend QueryStats;
+};
+
+
+using STATS_BY_OP_TYPE = std::array<OperationStats, static_cast<size_t>(TaskType::END)>;
+using STATS_BY_KEY_TYPE = std::array<STATS_BY_OP_TYPE, static_cast<size_t>(entity::KeyType::UNDEFINED)>;
+
+struct OperationStatsOutput {
+    std::map<std::string, uint32_t> stats_;
+    std::map<std::string, std::map<std::string, uint32_t>> key_type_;
```
alexowens90, line 84: Is this now also unused?

cpp/arcticdb/toolbox/query_stats.hpp
```diff
@@ -0,0 +1,106 @@
...
+};
+
+using QueryStatsOutput = std::map<std::string, std::map<std::string, std::map<std::string, OperationStatsOutput>>>;
```
alexowens90, line 87: I think this can now just be `std::map<std::string, std::map<std::string, std::map<std::string, std::map<std::string, uint32_t>>>>`?
Would be good to include an example here (the one above on line 45 is out of date) with what this looks like in practice

cpp/arcticdb/toolbox/query_stats.hpp
```diff
@@ -0,0 +1,106 @@
...
+//                 }
+//             }
+//         }
+//     }
+// }
+
+class QueryStats;
+class OperationStats{
+private:
+    std::array<folly::ThreadCachedInt<uint32_t>, static_cast<size_t>(entity::KeyType::UNDEFINED)> logical_key_counts_;
```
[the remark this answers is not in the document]
phoebusm: Ok I'll move it to another PR

cpp/arcticdb/toolbox/query_stats.hpp
```diff
@@ -0,0 +1,106 @@
...
+    folly::ThreadCachedInt<timestamp> total_time_ns_;
+    folly::ThreadCachedInt<uint32_t> count_;
+public:
+    uint32_t get_logical_key_count(entity::KeyType key_type) const;
```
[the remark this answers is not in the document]
phoebusm: Ok I'll move it to another PR

cpp/arcticdb/toolbox/query_stats.hpp
phoebusm: Ok I'll move it to another PR

cpp/arcticdb/toolbox/python_bindings.cpp
phoebusm: MIss that. Will remove

cpp/arcticdb/toolbox/query_stats.hpp
```diff
@@ -0,0 +1,106 @@
...
+    void reset_stats();
+    OperationStats();
+    
+    friend void add(entity::KeyType key_type, TaskType task_type, uint32_t value);
+    friend std::optional<RAIIAddTime> add_time(entity::KeyType key_type, TaskType task_type);
+    friend QueryStats;
```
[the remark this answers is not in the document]
phoebusm: That's a good trick!

alexowens90 (approved):

python/arcticdb/toolbox/query_stats.py
poodlewars, line 1: You need docstrings for these APIs.

Please document that this API is unstable and not governed by our semantic versioning.

Once you instrument the storages completely, you should write a docs page with some tutorials and examples.

poodlewars (changes requested):
Just this to sort out please https://github.com/man-group/ArcticDB/pull/2304/files#r2070143373

poodlewars (approved):

PR #2308, opened 2025-04-09
Fix compatibility test cleanup

IvoDD:
Previously in several places we were calling `old_ac.clear()` to clear all arctic/libraries created by an compatibility tests.

This however was producing issues when running pytest with multiprocessing as we do on the CI:
- Each pytest worker sets up its own instance of the session scoped fixtures
- When one process finishes with a fixture it runs the cleanup on an arctic instance which is potentially actively used by another pytest worker. Thus the spurious mongo library not found errors.

To mitigate this we introduce a new context manager `CompatLibrary`. It is responsible for setting up the `Venv`, `VenvArctic` and `VenvLib` and also creates the library on `__enter__` and cleans the library on `__exit__`.

So we rework all compat tests to use the new `CompatLibrary`

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/util/venv.py
+++ b/python/arcticdb/util/venv.py
@@ -215,3 +215,38 @@ class CurrentVersion:
         unset_config_int("VersionMap.ReloadInterval")
         del self.lib
         del self.ac
+
+
+class CompatLibrary:
+    """
+    Responsible for creating and cleaning up a library when writing compatibility tests.
+
+    Usually the library cleanup is managed by pytest fixtures, but it is hard to do so for compatibility tests because
+    they are responsible for maintaining their own arctic instances across several processes. As long as libraries
+    within the test are wrapped in a CompatLibrary cleanup will be dealt with.
+    """
+    def __init__(self, old_venv, uri, lib_name, create_with_current_version=False):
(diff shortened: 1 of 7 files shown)


grusev (approved):

python/arcticdb/util/venv.py
```diff
@@ -214,4 +211,53 @@ def __enter__(self):
...
+                ac.delete_library(lib_name)
+            except StorageException as e:
+                if self.uri.startswith("lmdb"):
+                    # For lmdb on windows sometimes the venv process keeps a dangling reference to the library for some
+                    # time. Since we use temporary directories for lmdb it's fine to not retry because the temporary
+                    # directory will be cleaned up in the end of the pytest process.
+                    logger.info(f"Failed to delete an lmdb library due to a Storage Exception: {e}")
+                else:
+                    raise
+        del ac
```
poodlewars, line 263: Why do we need the `del ac` calls in this PR, especially this last one? Python gives very loose guarantees about what `del` actually does, so it's a bit of a worry (especially when the things we're calling `del` on are about to go out of scope anyway)

PR #2309, opened 2025-04-09
Query Stats s3 v3

phoebusm:
#### Reference Issues/PRs

Continuing https://github.com/man-group/ArcticDB/pull/2304 for S3 query stats
https://man312219.monday.com/boards/7852509418/pulses/8417112339

Add S3 query stats support

#### Any other comments?
For S3 asynchronous calls, the timing measurement begins when the SDK API is initially called, although the actual processing may be delayed until the SDK executor processes the request. This differs from synchronous calls, where timing also starts at the API call, but since these calls are blocking, the requests are processed immediately within the same execution stack.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -141,13 +141,12 @@ void do_write_impl(
     auto s3_object_name = object_path(bucketizer.bucketize(key_type_dir, k), k);
     auto seg = key_seg.segment_ptr();
     auto segment_size = seg->calculate_size();
-    auto key_type = key_seg.key_type();
 
-    auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_PutObject);
+    auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_PutObject);
     auto put_object_result = s3_client.put_object(s3_object_name, *seg, bucket_name);
 
     if (put_object_result.is_success()) {
-        query_stats::add(key_type, query_stats::TaskType::S3_PutObject, query_stats::StatType::SIZE_BYTES, segment_size);
+        query_stats::add(query_stats::TaskType::S3_PutObject, query_stats::StatType::SIZE_BYTES, segment_size);
     }
     else{
         auto& error = put_object_result.get_error();
@@ -177,17 +176,16 @@ KeySegmentPair do_read_impl(
 	    KeyDecoder&& key_decoder,
         ReadKeyOpts opts) {
     ARCTICDB_SAMPLE(S3StorageRead, 0)
-    auto key_type = variant_key_type(variant_key);
-    auto key_type_dir = key_type_folder(root_folder, key_type);
+    auto key_type_dir = key_type_folder(root_folder, variant_key_type(variant_key));
     auto s3_object_name = object_path(bucketizer.bucketize(key_type_dir, variant_key), variant_key);
-    auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_GetObject);
+    auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_GetObject);
     auto get_object_result = s3_client.get_object(s3_object_name, bucket_name);
     auto unencoded_key = key_decoder(std::move(variant_key));
 
     if (get_object_result.is_success()) {
         ARCTICDB_SUBSAMPLE(S3StorageVisitSegment, 0)
         auto segment = std::move(get_object_result.get_output());
-        query_stats::add(key_type, query_stats::TaskType::S3_GetObject, query_stats::StatType::SIZE_BYTES, segment.calculate_size());
+        query_stats::add(query_stats::TaskType::S3_GetObject, query_stats::StatType::SIZE_BYTES, segment.calculate_size());
         return {VariantKey{unencoded_key}, std::move(segment)};
     } else {
         auto& error = get_object_result.get_error();
@@ -214,19 +212,17 @@ folly::Future<KeySegmentPair> do_async_read_impl(
     KeyBucketizer&& bucketizer,
     KeyDecoder&& key_decoder,
     ReadKeyOpts) {
-    auto key_type = variant_key_type(variant_key);
-    auto key_type_dir = key_type_folder(root_folder, key_type);
+    auto key_type_dir = key_type_folder(root_folder, variant_key_type(variant_key));
     auto s3_object_name = object_path(bucketizer.bucketize(key_type_dir, variant_key), variant_key);
     return s3_client.get_object_async(s3_object_name, bucket_name).thenValue([
             vk=std::move(variant_key), 
             decoder=std::forward<KeyDecoder>(key_decoder),
-            key_type,
             start = std::chrono::steady_clock::now()
         ] (auto&& result) mutable -> KeySegmentPair {
-            auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_GetObjectAsync, start);
+            auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_GetObjectAsync, start);
             if(result.is_success()) {
                 auto segment = std::move(result.get_output());
-                query_stats::add(key_type, query_stats::TaskType::S3_GetObjectAsync, query_stats::StatType::SIZE_BYTES, segment.calculate_size());
+                query_stats::add(query_stats::TaskType::S3_GetObjectAsync, query_stats::StatType::SIZE_BYTES, segment.calculate_size());
                 return KeySegmentPair(std::move(vk), std::move(segment));
             }
             else {
@@ -300,7 +296,7 @@ void do_remove_impl(
                 to_delete.emplace_back(std::move(s3_object_name));
 
                 if (to_delete.size() == delete_object_limit || k.index + 1 == group.size()) {
-                    auto query_stat_operation_time = query_stats::add_task_count_and_time(group.key(), query_stats::TaskType::S3_DeleteObjects);
+                    auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_DeleteObjects);
                     auto delete_object_result = s3_client.delete_objects(to_delete, bucket_name);
                     if (delete_object_result.is_success()) {
                         ARCTICDB_RUNTIME_DEBUG(log::storage(), "Deleted {} objects, one of which with key '{}'",
@@ -405,13 +401,12 @@ void do_write_if_none_impl(
             auto s3_object_name = object_path(bucketizer.bucketize(key_type_dir, k), k);
             auto& seg = *kv.segment_ptr();
             auto segment_size = seg.calculate_size();
-            auto key_type = kv.key_type();
 
-            auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_PutObject);
+            auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_PutObject);
             auto put_object_result = s3_client.put_object(s3_object_name, seg, bucket_name, PutHeader::IF_NONE_MATCH);
 
             if (put_object_result.is_success()) {
-                query_stats::add(key_type, query_stats::TaskType::S3_PutObject, query_stats::StatType::SIZE_BYTES, segment_size);
+                query_stats::add(query_stats::TaskType::S3_PutObject, query_stats::StatType::SIZE_BYTES, segment_size);
             }
             else {
                 auto& error = put_object_result.get_error();
@@ -492,7 +487,7 @@ bool do_iterate_type_impl(
 
     auto continuation_token = std::optional<std::string>();
     do {
-        auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_ListObjectsV2);
+        auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_ListObjectsV2);
         auto list_objects_result = s3_client.list_objects(path_info.key_prefix_, bucket_name, continuation_token);
         if (list_objects_result.is_success()) {
             auto& output = list_objects_result.get_output();
@@ -586,11 +581,10 @@ bool do_key_exists_impl(
     const S3ClientInterface& s3_client,
     KeyBucketizer&& b
 ) {
-    auto key_type = variant_key_type(key);
-    auto key_type_dir = key_type_folder(root_folder, key_type);
+    auto key_type_dir = key_type_folder(root_folder, variant_key_type(key));
     auto s3_object_name = object_path(b.bucketize(key_type_dir, key), key);
 
-    auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_HeadObject);
+    auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_HeadObject);
     auto head_object_result = s3_client.head_object(
         s3_object_name,
         bucket_name);
--- a/cpp/arcticdb/storage/s3/detail-inl.hpp
+++ b/cpp/arcticdb/storage/s3/detail-inl.hpp
@@ -189,7 +189,6 @@ KeySegmentPair do_read_impl(
         auto segment = std::move(get_object_result.get_output());
         query_stats::add(key_type, query_stats::TaskType::S3_GetObject, query_stats::StatType::SIZE_BYTES, segment.calculate_size());
         return {VariantKey{unencoded_key}, std::move(segment)};
-        ARCTICDB_DEBUG(log::storage(), "Read key {}: {}", variant_key_type(unencoded_key), variant_key_view(unencoded_key));
     } else {
         auto& error = get_object_result.get_error();
         raise_if_unexpected_error(error, s3_object_name);
@@ -413,7 +412,8 @@ void do_write_if_none_impl(
 
             if (put_object_result.is_success()) {
                 query_stats::add(key_type, query_stats::TaskType::S3_PutObject, query_stats::StatType::SIZE_BYTES, segment_size);
-            } {
+            }
+            else {
                 auto& error = put_object_result.get_error();
                 raise_s3_exception(error, s3_object_name);
             }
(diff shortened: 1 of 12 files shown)


willdealtry:
I honestly don't understand why this is being organized around key types. Key types are an implementation detail and exposing them to the user acheives the opposite of what I would think the point of this features is, i.e. to expose useful information about the performance of particular operations to the end user in a way that they can understand. So for example for a developer the number of IO operations required to find a particular version is synonymous with the number of reads of the VERSION key-type, but that's not evident to any not intimately familiar with the workings of the project (and I would argue, should not have to be).

Also, some of the most useful statistics like the proportion of the possible data keys that are read for a specific query (i.e. the ones that are not excluded by a column or date range filter or by some determination made based on the query stats) have nothing to do with key-types at all. If the whole feature relies on key-types we will forever be hammering square pegs into round holes. 

The current implementation is much too close to the existing performance tracing, which is intentionally compiled out because the run-time cost of having it on all the time is too high - I worry that this feature will be more costly for less utility, since the current instrumentation is the result of years of performance analysis that will not easily be replicated by this feature in the foreseeable future. For this PR to be useful as a unique feature it needs to be much more abstract, mapping just string or (ideally, for performance) enum-type descriptors to counters, gauges and (perhaps later on) histograms, in an abstract way that allows those stats to be periodically pushed to an external sink. 

I would suggest taking around three orthogonal statistics such as the number of IO operations required to get a version, the proportion of data keys that are read from the maximum available for a symbol/version pair, and the total time spent waiting for IO (maybe vs the wall-clock time since this is done in parallel). Wall-clock IO wait vs total query time would also be useful, as it would indicate the amount of time where we are simply waiting for data. Implemented in a simple but performant manner these would add significant utility in and of themselves, and also give a good basis to which other useful statistics could be added.

In its current form this PR seems like a solution in search of a problem, and needs a serious rethink.

willdealtry:
> We organized the stats by key types specifically to help diagnose performance issues. For example, if a user reports slow read operations and the query stats show a high number of VERSION key reads, we can identify that this is likely due to version compaction not being run recently. This organization provides actionable insights for both developers and advanced users.

I don't think we need stats for developers, as we already have much better and more detailed information than this will provide. I also think it's not possible to do both in the same feature - stats for users and developers will end up being useful to nobody. My point about the key types is that some performance considerations are quite closely related to the key type (as in the example above), some only tangentially and some not at all, so a statistics feature that is based off key type is giving key types an apparent centrality in the project that they don't have in reality.

> The current implementation already uses enums rather than string maps for performance reasons. The stats are stored in static arrays indexed by these enums, providing efficient access patterns. I agree that histograms would be valuable and have planned to implement them in the next phase of development. The current design provides a foundation that can be extended to support external monitoring systems in the future.

I'm aware that it uses the KeyType enum. I mean that we should have a flexible set of values that can serve as fast keys that are not tied to some existing implementation detail.

> We completely agree that tracking wall-clock IO wait time would be a valuable addition. Unfortunately, implementing this feature would require mapping Python threads to Folly worker threads, which adds significant complexity. We've stripped and deferred this enhancement to a future iteration to keep the current implementation focused and maintainable.

The existing instrumentation that is enabled by uncommenting a line in peformance_tracing.hpp tracks both total IO time and wallclock IO time, so I don't agree that it's impossible without mapping Python threads to worker threads (at least in the case of single-threaded Python use, which is 99% of all current use-cases). Even in the case of multi-threaded Python use, I'm not convinced that having a breakdown per top-level Python method will ever be worth the effort. Even were it to be desirable, the correct way to do it would be to pass a timings parameter object into the relevant methods that could carry a trace id, which is entirely orthogonal to the current design decisions being made.

If fine-grained stats like the existing performance tracing is what's required, why don't we use the existing mechanism? All that would be required would be to move it from compile-time to run-time switching and we would have the benefit of dozens of measurements that have been evolved through hundreds of hours of peformance work, rather than what appears to be a far less capable implemtation of essentially the same thing?

cpp/arcticdb/toolbox/query_stats.cpp
```diff
@@ -116,37 +136,54 @@ QueryStats::QueryStatsOutput QueryStats::get_stats() const {
     return result;
 }
 
-void QueryStats::add(entity::KeyType key_type, TaskType task_type, uint32_t value) {
+void QueryStats::add(entity::KeyType key_type, TaskType task_type, StatType stat_type, uint32_t value) {
     if (is_enabled()) {
```
poodlewars, line 140: IMO an `if (!is_enabled())` then return at the top is cleaner than this nesting

cpp/arcticdb/toolbox/query_stats.cpp
```diff
@@ -116,37 +136,54 @@ QueryStats::QueryStatsOutput QueryStats::get_stats() const {
...
+                stats.total_time_ns_.increment(value);
+                break;
+            case StatType::COUNT:
+                stats.count_.increment(value);
+                break;
+            case StatType::SIZE_BYTES:
+                stats.size_bytes_.increment(value);
+                break;
+            default:
+                throw std::invalid_argument("Invalid stat type");
```
poodlewars, line 153: Assertion error with our normal framework to bubble up to the Python layer cleanly?

cpp/arcticdb/toolbox/query_stats.cpp
```diff
@@ -116,37 +136,54 @@ QueryStats::QueryStatsOutput QueryStats::get_stats() const {
...
 
-void add(entity::KeyType key_type, TaskType task_type, uint32_t value) {
-    QueryStats::instance()->add(key_type, task_type, value);
+void add(entity::KeyType key_type, TaskType task_type, StatType stat_type, uint32_t value) {
+    QueryStats::instance()->add(key_type, task_type, stat_type, value);
 }
 
-[[nodiscard]] std::optional<RAIIAddTime> add_task_count_and_time(entity::KeyType key_type, TaskType task_type) {
-    return QueryStats::instance()->add_task_count_and_time(key_type, task_type);
+[[nodiscard]] std::optional<RAIIAddTime> add_task_count_and_time(
```
poodlewars, line 183: I don't understand what `start` is for?

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -205,16 +215,25 @@ folly::Future<KeySegmentPair> do_async_read_impl(
...
-	    auto unencoded_key = decoder(std::move(vk));	
-            raise_s3_exception(result.get_error(), fmt::format("{}", unencoded_key));
-	}
+    return s3_client.get_object_async(s3_object_name, bucket_name).thenValue([
+            vk=std::move(variant_key), 
+            decoder=std::forward<KeyDecoder>(key_decoder),
+            key_type,
+            start = std::chrono::steady_clock::now()
+        ] (auto&& result) mutable -> KeySegmentPair {
+            auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_GetObjectAsync, start);
```
poodlewars, line 227: Is this the only usecase for the `start` parameter? I think we can live without it in this case and just have a slightly less accurate timing

docs/mkdocs/docs/tutorials/query_stats.md
```diff
@@ -0,0 +1,138 @@
+# Query Statistics
+
+> **Warning**: The Query Statistics API is unstable and not governed by ArcticDB's semantic versioning. It may change or be removed in future versions without notice.
+
+ArcticDB provides a Query Statistics API that allows you to collect and analyze performance metrics for operations performed on your data stores.
+This can be useful for debugging, performance optimization.
+The Query Statistics feature uses a global container to store all measurements. This means statistics from all Python threads in your application will be collected in the same structure.
```
poodlewars, line 7: I think it would be fine to move this to a "caveats" section instead of leading with it in the introduction

docs/mkdocs/docs/tutorials/query_stats.md
```diff
@@ -0,0 +1,138 @@
...
+Currently only S3 backend is supported. Support for other storage backend is pending.
+
+## Basic Usage
+
+There are two ways to enable query statistics collection:
+
+### Using the Context Manager
+
+The context manager automatically enables statistics at the beginning of a block and disables them at the end.
+Please note that recursion is not supported:
```
poodlewars, line 19: replace "recursion" with "nested context managers"

docs/mkdocs/docs/tutorials/query_stats.md
```diff
@@ -0,0 +1,138 @@
...
+
+The statistics are returned as a nested dictionary organized by:
+- Key type (e.g., `SYMBOL_LIST`, `TABLE_DATA`, `VERSION_REF`)
+- Operation group (currently `storage_ops` only)
+- Task type (e.g., `S3_ListObjectsV2`, `S3_PutObject`)
+
+Each task contains measurements like:
+- `count`: Number of times the operation was performed
+- `total_time_ms`: Total execution time in milliseconds
+- For data operations, additional metric `size_bytes` for the size of data being transferred
```
poodlewars, line 83: Explain what this means, is it compressed or uncompressed?

docs/mkdocs/docs/tutorials/query_stats.md
```diff
@@ -0,0 +1,138 @@
...
+
+# Get new statistics
+stats = qs.get_query_stats()
+print(stats)
+
+# Disable statistics collection when done
+qs.disable()
+```
+
+## Output Structure
```
poodlewars, line 73: Has this section rotted now that you've added the extra layer of IO nesting?

python/arcticdb/toolbox/query_stats.py
```diff
@@ -33,7 +33,7 @@ def query_stats() -> Iterator[None]:
         This API is unstable and not governed by semantic versioning.
     """
     if qs.is_enabled():
-        raise UserInputException("Query Stats is already enabled")
+        raise UserInputException("Query Stats is already enabled") # This will prohibit recursion
```
poodlewars, line 36: Again, replace "recursion" with "nested context managers"

python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -64,3 +101,396 @@ def test_query_stats_clear(s3_version_store_v1, clear_query_stats):
...
+    qs.enable()
+    s3_version_store_v1.write("a", 1)
+    s3_version_store_v1.write("a", 2)
+    with config_context("VersionMap.ReloadInterval", 0):
+        s3_version_store_v1.read("a")
+        s3_version_store_v1.read("a")
+    stats = qs.get_query_stats()
+#     {
+#     "SYMBOL_LIST": {
+#         "storage_ops": {
```
poodlewars, line 228: The top level grouping shouldn't be key type as William described. `storage_ops` (possibly renamed to `storage_operations` or `io`) should be at the root level, rather than key types

python/tests/integration/toolbox/test_query_stats.py
poodlewars, line 1: Tests for staged data, showing the `APPEND_DATA` IO, would be good too

poodlewars (changes requested):

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -385,10 +405,15 @@ void do_write_if_none_impl(
...
 
+            auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_PutObject);
             auto put_object_result = s3_client.put_object(s3_object_name, seg, bucket_name, PutHeader::IF_NONE_MATCH);
 
-            if (!put_object_result.is_success()) {
+            if (put_object_result.is_success()) {
+                query_stats::add(key_type, query_stats::TaskType::S3_PutObject, query_stats::StatType::SIZE_BYTES, segment_size);
+            } {
                 auto& error = put_object_result.get_error();
                 raise_s3_exception(error, s3_object_name);
```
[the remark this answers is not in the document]
phoebusm: Rebase error. My bad

cpp/arcticdb/toolbox/query_stats.cpp
```diff
@@ -116,37 +136,54 @@ QueryStats::QueryStatsOutput QueryStats::get_stats() const {
...
         auto& stats = stats_by_key_type_[static_cast<size_t>(key_type)][static_cast<size_t>(task_type)];
         stats.count_.increment(1);
-        return std::make_optional<RAIIAddTime>(stats.total_time_ns_);
+        return std::make_optional<RAIIAddTime>(stats.total_time_ns_, start.value_or(std::chrono::steady_clock::now()));
     }
     return std::nullopt;
 }
 
-RAIIAddTime::RAIIAddTime(folly::ThreadCachedInt<timestamp>& time_var) :
+RAIIAddTime::RAIIAddTime(folly::ThreadCachedInt<timestamp>& time_var, std::chrono::time_point<std::chrono::steady_clock> start) :
```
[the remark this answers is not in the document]
phoebusm: If exception is being raised by the SDK, it's a very rare siutation which I don't think user will expect query stats should still work correctly, though it's better to make it clear in the manual
If the exception is raised by functions in `details-inl.hpp` because of a http error e.g. permission, not found, it's still an IO IMO

cpp/arcticdb/toolbox/query_stats.cpp
phoebusm: For async IO `do_async_read_impl`

cpp/arcticdb/storage/s3/detail-inl.hpp
phoebusm: It's just a extra variable needed in the function to get a more accurate result. I rather not having this API logged than having something inaccurate

docs/mkdocs/docs/tutorials/query_stats.md
phoebusm: It was not rotted but will be as the key type grouping will be gone

cpp/arcticdb/toolbox/query_stats.hpp
```diff
@@ -17,82 +17,87 @@
...
-//     "SYMBOL_LIST": { <- STATS_BY_KEY_TYPE
-//         "storage_ops": { <- STATS_BY_OP_TYPE
-//             "S3_ListObjectsV2": { <- OperationStats::stats_
-//                 "total_time_ms": 83,
-//                 "count": 3
-//             }
+//     "storage_ops": { <- STATS_BY_STORAGE_OP_TYPE
+//         "S3_ListObjectsV2": { <- OperationStats
+//             "total_time_ms": 32,
+//             "count": 2
```
alexowens90, line 55: Presumably a size in bytes as well? Can use [this](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/util/format_bytes.hpp) to format it nicely

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -573,10 +594,10 @@ bool do_key_exists_impl(
...
 
         ARCTICDB_DEBUG(log::storage(), "Head object returned false for key {} {} {}:{}",
-                       variant_key_view(key),
-                       int(error.GetErrorType()),
-                       error.GetExceptionName().c_str(),
-                       error.GetMessage().c_str());
+                        variant_key_view(key),
+                        int(error.GetErrorType()),
+                        error.GetExceptionName().c_str(),
+                        error.GetMessage().c_str());
```
alexowens90, line 600: formatting

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -207,14 +214,21 @@ folly::Future<KeySegmentPair> do_async_read_impl(
...
-        else {
-	    auto unencoded_key = decoder(std::move(vk));	
-            raise_s3_exception(result.get_error(), fmt::format("{}", unencoded_key));
-	}
+    return s3_client.get_object_async(s3_object_name, bucket_name).thenValue([
+            vk=std::move(variant_key), 
+            decoder=std::forward<KeyDecoder>(key_decoder),
+            start = std::chrono::steady_clock::now()
+        ] (auto&& result) mutable -> KeySegmentPair {
+            auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_GetObjectAsync, start);
```
alexowens90, line 222: What's the advantage of this, over creating `query_stat_operation_time` right before the`get_object_async` call, and getting rid of the `start` parameter? The result should be basically the same?

cpp/arcticdb/toolbox/query_stats.hpp
phoebusm: I prefer leaving it as it is as it'll be easier to be processed by script if the unit remains unchanged

cpp/arcticdb/storage/s3/detail-inl.hpp
phoebusm: `get_object_async` is an async call. So the call would return immediately and the time measurement will be meaningless.

alexowens90 (approved):

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -140,10 +140,16 @@ void do_write_impl(
...
+    auto key_type = key_seg.key_type();
 
+    auto query_stat_operation_time = query_stats::add_task_count_and_time(key_type, query_stats::TaskType::S3_PutObject);
     auto put_object_result = s3_client.put_object(s3_object_name, *seg, bucket_name);
 
-    if (!put_object_result.is_success()) {
+    if (put_object_result.is_success()) {
+        query_stats::add(key_type, query_stats::TaskType::S3_PutObject, query_stats::StatType::SIZE_BYTES, segment_size);
+    }
+    else{
```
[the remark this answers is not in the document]
poodlewars: This hasn't been addressed

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -381,14 +400,20 @@ void do_write_if_none_impl(
...
+            auto segment_size = seg.calculate_size();
 
+            auto query_stat_operation_time = query_stats::add_task_count_and_time(query_stats::TaskType::S3_PutObject, key_type);
             auto put_object_result = s3_client.put_object(s3_object_name, seg, bucket_name, PutHeader::IF_NONE_MATCH);
 
-            if (!put_object_result.is_success()) {
+            if (put_object_result.is_success()) {
+                query_stats::add(query_stats::TaskType::S3_PutObject, key_type, query_stats::StatType::SIZE_BYTES, segment_size);
+            }
+            else {
```
poodlewars, line 416: formatting

poodlewars (approved):

PR #2319, opened 2025-04-11
Timeseries defrag function

alexowens90:
Adds a (private) function to defragment timeseries data. See big list of caveats in code comments for limitations

### DIFF
--- a/python/arcticdb/util/defrag_timeseries.py
+++ b/python/arcticdb/util/defrag_timeseries.py
@@ -1,0 +1,29 @@
+"""
+Copyright 2025 Man Group Operations Limited
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+from typing import Union
+
+from arcticdb.version_store.library import Library
+from arcticdb.version_store._store import NativeVersionStore
+
+
+def _defrag_timeseries(lib: Union[Library, NativeVersionStore], sym: str):
+    if isinstance(lib, Library):
+        lib = lib._nvs
+    target_rows_per_slice = lib.lib_cfg().lib_desc.version.write_options.segment_row_size
+    if target_rows_per_slice == 0:
+        target_rows_per_slice = 100_000
+    index = lib.read_index(sym)
+    row_counts = index["end_row"] - index["start_row"]
+    idx = 0
+    for row_count in row_counts:
+        if row_count == target_rows_per_slice:
+            idx += 1
+        else:
+            break
+    if idx < len(index) and index["start_row"][idx] != index["start_row"][-1]:
+        df = lib.read(sym, row_range=(index["start_row"][idx], index["end_row"][-1])).data
+        lib.update(sym, df)
--- a/python/arcticdb/util/defrag_timeseries.py
+++ b/python/arcticdb/util/defrag_timeseries.py
@@ -10,6 +10,17 @@ from arcticdb.version_store.library import Library
 from arcticdb.version_store._store import NativeVersionStore
 
 
+# This is incredibly basic, but should work well on append-only libraries
+# Limitations:
+# - Will throw an exception during the call to update if the symbol is not a timeseries
+# - Requires all the fragmented data to fit into memory at once
+# - Assumes that if the data is fragmented at a given timestamp, that all subsequent data is also fragmented
+# - Has no tolerance, so a row-slice with 99,999 rows will be considered fragmented and compacted into a 100k row
+#   segment
+# - Cannot specify the number of rows per slice, just uses the library default
+# - Has no concept of changing the column slicing policy
+# - Creates a new version
+# - Can inefficiently traverse the entire index key when data is column sliced but not fragmented
 def _defrag_timeseries(lib: Union[Library, NativeVersionStore], sym: str):
     if isinstance(lib, Library):
         lib = lib._nvs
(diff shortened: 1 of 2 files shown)


python/arcticdb/util/defrag_timeseries.py
```diff
@@ -0,0 +1,40 @@
...
+    if isinstance(lib, Library):
+        lib = lib._nvs
+    target_rows_per_slice = lib.lib_cfg().lib_desc.version.write_options.segment_row_size
+    if target_rows_per_slice == 0:
+        target_rows_per_slice = 100_000
+    index = lib.read_index(sym)
+    row_counts = index["end_row"] - index["start_row"]
+    idx = 0
+    for row_count in row_counts:
+        if row_count == target_rows_per_slice:
```
IvoDD, line 34: What do you think about using `>=`? 
E.g. it can save some extra compaction work if compactor received a large burst >100k messages (I think arcticc doesn't do the incomplete slicing). Although there is value with the current setup to "fragment" the large chunks

python/arcticdb/util/defrag_timeseries.py
```diff
@@ -0,0 +1,40 @@
...
+
+from arcticdb.version_store.library import Library
+from arcticdb.version_store._store import NativeVersionStore
+
+
+# This is incredibly basic, but should work well on append-only libraries
+# Limitations:
+# - Will throw an exception during the call to update if the symbol is not a timeseries
+# - Requires all the fragmented data to fit into memory at once
+# - Assumes that if the data is fragmented at a given timestamp, that all subsequent data is also fragmented
```
IvoDD, line 17: I think this can be problematic for some tick streaming users, who need to be able to backfill. Some folks will occasionally want to update historic data (e.g. prepend more data before collection started, or update existing data because they notice something in their kafka producer was wrong). In this case people would trigger an "entire" symbol defragmentation which can be problematic.

What do you think about at least being able to pass in a point in time from where fragmented data should be considered? E.g. it will start looking for fragmented data say a year in advance

python/tests/unit/arcticdb/test_defrag_timeseries.py
```diff
@@ -0,0 +1,132 @@
...
+            {
+                "col0": np.arange(num_rows, dtype=np.int64),
+                "col1": np.arange(num_rows, 2*num_rows, dtype=np.float32),
+                "col2": np.arange(2*num_rows, 3*num_rows, dtype=np.uint16),
+            },
+            index=index,
+        )
+        lib.append(sym, df)
+        start_date += pd.Timedelta(days=num_rows)
+    generic_defrag_test(lib, sym)
```
IvoDD, line 83: Would be nice to assert that the latest version's first few data keys are with v0. 
I.e. to verify the defragmentation didn't include the first few rows in the update.
E.g. something like:
```
index_df = lib.read_index(sym)
assert (index_df.iloc[:18]["version_id"]==0).all()
```

IvoDD: Also do a similar test for dynamic_schema test

alexowens90: Agreed to both, will add

python/arcticdb/util/defrag_timeseries.py
alexowens90: This function wouldn't really work with anything ticking that fast anyway, would OOM on the read

python/arcticdb/util/defrag_timeseries.py
alexowens90: I agree we'd need for a general purpose defrag method, but this is only really designed to work for super slow ticking data like minutely, where the entire symbol should fit in memory comfortably, and so an entire symbol defrag is fine, and possibly even desirable if the prepend/update was a small amount of data

IvoDD (approved):

python/arcticdb/util/defrag_timeseries.py
poodlewars: I think an API that takes a `date_range` to defrag could be useful but this is fine as it is for now

python/arcticdb/util/defrag_timeseries.py
```diff
@@ -0,0 +1,40 @@
...
+# - Has no tolerance, so a row-slice with 99,999 rows will be considered fragmented and compacted into a 100k row
+#   segment
+# - Cannot specify the number of rows per slice, just uses the library default
+# - Has no concept of changing the column slicing policy
+# - Creates a new version
+# - Can inefficiently traverse the entire index key when data is column sliced but not fragmented
+def _defrag_timeseries(lib: Union[Library, NativeVersionStore], sym: str):
+    if isinstance(lib, Library):
```
poodlewars, line 25: I can't see a test for the case where this is `True`

python/tests/unit/arcticdb/test_defrag_timeseries.py
```diff
@@ -0,0 +1,135 @@
...
+        cols = random.sample(col_set, num_cols)
+        index = pd.date_range(start_date, periods=num_rows)
+        df = pd.DataFrame(
+            data=rng.integers(0, 1_000, size=(num_rows, num_cols)),
+            columns=cols,
+            index=index,
+        )
+        lib.append(sym, df)
+        start_date += pd.Timedelta(days=num_rows)
+    generic_defrag_test(lib, sym)
```
poodlewars, line 135: Could test an empty dataframe

python/arcticdb/util/defrag_timeseries.py
poodlewars: I think this should be `>=` shouldn't it, eg someone could change the row slicing policy after the library was created

PR #2320, opened 2025-04-11
Fix race conditions on the None object refcount during a multithreaded read

vasil-pashov:
**Bugfix**
Columns are handled in multiple threads during read calls. String columns can contain `None` values. `None` is a global static ref counted object and the refcount is not atomic. When ArcticDB places `None` objects in columns it must increment the refcount. Currently None objects are allocated only via type handlers. ArcticDB has a global spin-lock that is shared by all type-handlers. The bug is caused by [this line](https://github.com/man-group/ArcticDB/blob/300e121e1be47ecfbabba78f077851a9c3b0772c/cpp/arcticdb/python/python_utils.hpp#L117) the spin-lock is wrapped in a `std::lock_guard` but there is a call to `unlock`. When `unlock` is called another thread will take the lock and start calling `Py_INCREF(Py_None)` but when the function exists the `std::scope_guard` will call unlock again allowing another thread to start calling `Py_INCREF(Py_None)` in parallel.

**Refactoring**
- Remove GIL safe py none. It was created because pybind11 wraps `Py_None` in an object and calls `Py_INCREF(Py_None)` and we must hold the GIL when incrementing the refcount. The wrapper we have was used only to get the pointer to the `Py_None` object. We don't need pybind11 to do that. Using the C API we can directly get `Py_None` which is global object
- Add function to check if a python object is `None`
- Remove uses of py::none{} in places where we don't hold the GIL (most of those were just to get the `Py_None` object that's inside `py:none` 

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/python/python_types.hpp
```diff
@@ -59,5 +59,13 @@ inline bool is_py_nan(PyObject* obj) {
     return false;
 }
 
+inline bool is_py_none(PyObject* obj) {
+#if PY_MAJOR_VERSION >=3 && PY_MINOR_VERSION >= 10
```
alexowens90, line 63: nit: This would not handle the release of Python 4.0

cpp/arcticdb/pipeline/index_utils.hpp
alexowens90, line 1: This seems to have snuck in from the other PR

cpp/arcticdb/storage/python_bindings.cpp
```diff
@@ -277,10 +277,8 @@ void register_bindings(py::module& storage, py::exception<arcticdb::ArcticExcept
                                        [](const arcticdb::proto::storage::VersionStoreConfig & cfg){
                                            return pb_to_python(cfg);
                                        },
-                                       [](const std::monostate & ){
-                                            auto none = py::none{};
-                                            none.inc_ref();
-                                            return none.cast<py::object>();
+                                       [](const std::monostate & ) -> py::object {
+                                            return py::none{};
```
alexowens90, line 281: You've used `Py_None` everywhere else?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -581,6 +582,7 @@ VersionedItem PythonVersionStore::write_versioned_composite_data(
 
     auto write_options = get_write_options();
     auto de_dup_map = get_de_dup_map(stream_id, maybe_prev, write_options);
+    debug::check<ErrorCode::E_ASSERTION_FAILURE>(PyGILState_Check(), "py::none object creation increases the refcount and must hold the GIL");
```
alexowens90, line 585: Why is this here? This is for recursive normalizers, which seems unrelated to the rest of the change?

cpp/arcticdb/python/python_handlers.cpp
```diff
@@ -16,23 +16,12 @@
...
-
-    spin_lock.lock();
-    for(auto i = 0U; i < count; ++i)
-        Py_INCREF(none->ptr());
-    spin_lock.unlock();
+    std::lock_guard lock{spin_lock};
+    for(auto i = 0U; i < count; ++i) {
+        *ptr_dest++ = Py_None;
+        Py_INCREF(Py_None);
+    }
```
alexowens90, line 24: This is now also locking around assigning to `ptr_dest`, we only want to lock around the incref

alexowens90: If we made `increment_none_refcount` a free function that takes a reference to a spin lock as well, then we could use that here too. Should be less error prone, as we can force all None increfs through that method.

cpp/arcticdb/python/python_utils.hpp
```diff
@@ -104,17 +103,14 @@ inline void prefill_with_none(
...
-    for (auto i = 0U; i < num_rows; ++i)
-        *ptr_dest++ = none->ptr();
+    std::fill_n(ptr_dest, num_rows, Py_None);
 
     if(inc_ref_count == IncrementRefCount::ON) {
+        std::lock_guard lock(spin_lock);
         auto none_count = num_rows - sparse_count;
         for (auto j = 0U; j < none_count; ++j)
-            Py_INCREF(none->ptr());
+            Py_INCREF(Py_None);
```
alexowens90, line 112: Same comment as above RE `increment_none_refcount` being a free function

cpp/arcticdb/storage/python_bindings.cpp
vasil-pashov: This is in the python bindings and this gets returned to the python layer. At this point we're holding the GIL and we must increment the refcount so I left pybind to handle this

cpp/arcticdb/version/version_store_api.cpp
vasil-pashov: Leftover from debugging. Will remove

alexowens90 (approved):

poodlewars:
Is it possible to include a regression test that exposes the current issue and protects us against it coming back?

poodlewars:
Another naive question. How does decrementing the refcount work? I get that if we manage to expose the data up to Python we should be all good, but how about in error cases, what decrements the refcount then? I get that this doesn't matter too much, at worst we leave the `None` object alive for no reason, who cares, just trying to understand how this works. Also begs the question whether we could just set the `None` refcount to some huge value when we load the arcticdb module and not worry about tracking it

vasil-pashov:
@poodlewars 

> Is it possible to include a regression test that exposes the current issue and protects us against it coming back?

I'm working on a follow up PR that fixes a more general problem and it has a test that will catch both case (not holding the GIL and racing with other python threads and this problem that racing with C++ threads) https://github.com/man-group/ArcticDB/pull/2334

> Naive question. From the PR description it sounds like just removing this unlock call would fix the problem? Is that true?
>
> (not saying we should do that. Just checking understanding)

That is correct. I think I removed it, didn't I?

Using pybind's `py:none{}` without holding the GIL was also a but, even though we would rarely see the consequences of it and I removed the GIL safe py none because we don't really need it and I think it was confusing and slightly pessimistic.

> Another naive question. How does decrementing the refcount work? I get that if we manage to expose the data up to Python we should be all good, but how about in error cases, what decrements the refcount then? I get that this doesn't matter too much, at worst we leave the None object alive for no reason, who cares, just trying to understand how this works. Also begs the question whether we could just set the None refcount to some huge value when we load the arcticdb module and not worry about tracking it

I think it works as in any other class. When a None gets out of scope in Python it's refcount gets decremented. Since we're filling Python data with None only in read and we handle it to Python we don't have to decref it, however we must increment the refcount because when the dataframe we created gets out of scope it'll call decref for each None in it and if we didn't increment it enough times it'll crash. As to whether we should just make it some large number and stop caring, probably we can. There's even a [PEP](https://peps.python.org/pep-0683/) for CPython that proposes similar thing (I think CPython 3.12.0 has this). I'm not sure if we must do it though as None is global and other libraries could be using it in unpredictable ways. I don't really believe it'll cause an error but I'm being cautious.

poodlewars (approved):

PR #2321, opened 2025-04-11
Refactor to_atom and to_ref to properly use forwarding references

vasil-pashov:
This solves two problems
- Code duplication. to_atom had 3 overloads for value/ref/rval ref for the same thing. Forwarding references were invented to solve this problem.
- There were unnecessary copies. `to_atom` had an overload taking `VeriantKey` by value at some point some APIs have changed and started returning `AtomKey` instead of `VariantKey` due to the excessive use of `auto` nobody noticed the difference. Thus we ended up with calling `to_atom` on an atom key, that worked because `VariantKey` can be constructed from an `AtomKey` implicitly thus we ended up constructing `VariantKey` from an `AtomKey` only to extract the `AtomKey` from that. Forwarding references do not allow implicit conversions thus the compiler pointed out all places in the code where the above happens.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_utils.cpp
+++ b/cpp/arcticdb/version/version_utils.cpp
@@ -41,7 +41,7 @@ VariantKey write_multi_index_entry(
                                      std::forward<decltype(segment)>(segment)).wait();
     });
 
-    for (AtomKey& key : keys) {
+    for (const AtomKey& key : keys) {
         multi_index_agg.add_key(key);
     }
     TimeseriesDescriptor timeseries_descriptor;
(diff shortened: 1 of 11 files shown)


cpp/arcticdb/version/version_utils.cpp
```diff
@@ -41,8 +41,8 @@ VariantKey write_multi_index_entry(
                                      std::forward<decltype(segment)>(segment)).wait();
     });
 
-    for (auto &key : keys) {
-        multi_index_agg.add_key(to_atom(key));
+    for (AtomKey& key : keys) {
```
ognyanstoimenov, line 44:
```suggestion
    for (const AtomKey& key : keys) {
```

ognyanstoimenov (approved):

alexowens90 (approved):

PR #2316, opened 2025-04-10
Installation tests

grusev:
#### What does this implement or fix?

This is a POC ground for creating a test suite that will be able to be executed against installable arcticdb from pypi and conda with set of tests that do now use any of our internal test fixtures that do wrap arctic in any way. On the contrary the fixtures for those tests are directly using arcticdb objects.

The tests that are selected from our test set will be sligthly modified so that they can be executed with new fixtures and old fixtures. How? instead of being tests they are modified as functions which are called from the respective test having appropriate fixture.

There is a new github workflow that will run very fast - no compilation of our project just installing python, arcticdb and all needed dependencies (in order to run tests from our code we will polute this clean environment with our testing dependencies, not sure yet how that can be avoided if those tests stay in current repo, obvioysly if we use dedicated repo we can avoid that)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #2330, opened 2025-04-15
v3 Of installation Tests

grusev:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/execute_installation_tests.yml
+++ b/.github/workflows/execute_installation_tests.yml
@@ -1,0 +1,136 @@
+name: Run Installation Tests v3
+
+on:
+  push:
+    branches: 
+      - installation_tests
+  workflow_dispatch:
+    inputs:
+      os:
+        description: 'Operating System to test'
+        required: false
+        default: 'ubuntu-22.04'
+        type: choice
+        options:
+          - all
+          - ubuntu-22.04
+          - windows-latest
+          - macos-latest
+      python:
+        description: 'Python Version to test'
+        required: false
+        default: '3.13'
+        type: choice
+        options:
+          - all
+          - "3.7"
+          - "3.8"
+          - "3.9"
+          - "3.10"
+          - "3.11"
+          - "3.12"
+          - "3.13"
+jobs:
+
+  test:
+    runs-on: ${{ inputs.os != 'all' && inputs.os || matrix.os }}
+    strategy:
+      matrix:
+        #os: [ubuntu-22.04, windows-latest, macos-latest]
+        #python: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        #os: [ubuntu-22.04, windows-latest, macos-13]
+        #python: ["3.8", "3.12"]
+        include:
+          - os: ubuntu-22.04
+            python: "3.7"
+          - os: windows-latest
+            python: "3.8"
+          - os: macos-latest
+            python: "3.9"
+          - os: windows-latest
+            python: "3.10"
+          - os: macos-latest
+            python: "3.11"
+          - os: ubuntu-22.04
+            python: "3.12"
+          - os: ubuntu-22.04
+            python: "3.13"
+      fail-fast: false  
+
+    steps:
+
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python || inputs.python}}
+
+    - name: Delete python/arcticdb folder
+      shell: bash -l {0}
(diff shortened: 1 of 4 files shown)


.github/workflows/execute_installation_tests_v3.yml
```diff
@@ -0,0 +1,212 @@
...
+      if: ${{ contains(matrix.os, 'macos') || contains(inputs.os, 'macos')}}
+      with:
+        environment-file: ${{ env.conda_file }}
+        environment-name: arcticdb
+        init-shell: >-
+          bash
+        cache-environment: true
+        post-cleanup: 'all'
+
+    - name: Add arcticdb from conda-forge
```
poodlewars, line 151: We'll want to run these using the Linux Conda build too, not just Mac

python/tests/conftest.py
```diff
@@ -143,9 +143,29 @@ def lmdb_shared_storage(tmp_path_factory) -> Generator[LmdbStorageFixture, None,
 
 
 @pytest.fixture(scope="function")
-def lmdb_storage(tmp_path) -> Generator[LmdbStorageFixture, None, None]:
-    with LmdbStorageFixture(tmp_path) as f:
-        yield f
+def lmdb_storage(request, tmp_path) -> Generator[LmdbStorageFixture, None, None]:
```
poodlewars, line 146: I think it will be clearer to have a dedicated test suite for these installability tests because they will have to keep working with possibly very old ArcticDB versions, so will in a sense be "frozen" - whereas the core tests will change more over time.

I also don't think we should add this complexity to the core `lmdb_storage` fixture - a dedicated single purpose fixture would be clearer to me.

I understand why you've done it like this, it is quite appealing to just reuse existing tests, and the way you've done it is really neat, I just think it is best to be painfully obvious here (especially for the comprehensibility for new joiners etc).

poodlewars:
Regarding the open questions in the description,

1) I expect we can get most of the value by running against simulators (moto, azurite) rather than real storage backends. Seems to be the "goldilocks zone"
2) I think this is an argument for a dedicated suite, that can cope with any API changes by testing the installed arcticdb version, rather than putting this in to the existing test suites
3) Sounds good


Finally, it will be important to have some brief docs explaining this setup (and why we have it) for the benefit of future developers.

poodlewars:
I think an important part of this will be having a way to constrain the dependencies used by earlier ArcticDB versions - we don't want tests to fail on ArcticDB vPREHISTORIC when numpy 3 is released, for example

poodlewars:
What kinds of problems is this designed to detect? Would a simple cron that installs latest arcticdb from PyPi and Conda and runs a simple write + read do the job?

poodlewars:
Do we think it's valuable to do this on PyPi (where we bundle all our C++ deps) or perhaps only doing this on Conda would suffice?

python/tests/conftest.py
grusev: I have testsed so far with 4.4.7 version (lowest) and there were no problems. But I do understand what you mean. An approach with what you suggest was the original one - https://github.com/man-group/ArcticDB/pull/2316 

The idea there was to reuse as little as possible, and therefore only shared_tests.py was the thing that both current tests and new tests shared, and the idea was once a test started to diverge to be no more shared but have two versions - one for the original and one for the installation tests (it was ok since the tests files/suites were different )

.github/workflows/execute_installation_tests_v3.yml
grusev: Yes I havethought about it too and would propose to add as one of the combination (currently we have 3 linuxes + pypi, we can make one of linuxes not pypi)

maxim-morozov:
1. I don't know if it makes sense to reuse some of the storage tests for installation tests. I think it makes it a bit hard to make sure that we are running only the version of tests from the branch when the release was cut. I think the goal of the installation tests is to make sure that ArcticDB basic functionality is still working when it comes down to the underlying dependencies. It probably can be sufficient to create a basic set of tests like writing and reading from different storages that can be run against any ArcticDB.
2. This is exactly my point as in 1. With this proposal, we are basically using master tests to run against previous versions of ArcticDB. ArcticDB does not ensure forward compatibility, so it can easily be broken. I think the options are:
a. Do pure installation tests with very minimal functionality that we expect is not going to be changing, like reading and writing. Potentially have an option to specify the set of versions these tests are supposed to work with.
b. If we want to reuse our integration tests, we should check out tests from the same branch and version ArcticDB release was cut from.
4. Yes, this is why we need a separate test suite for installation tests. Also, installation tests should be quick and run multiple times a day as the underlying dependencies can change at any time.

PR #2334, opened 2025-04-16
Hold the GIL when incrementing None's refcount to prevent race conditions when there are multiple Python threads

vasil-pashov:
#### Reference Issues/PRs

None is a global static object in Python which is also refcounted. When ArcticDB creates `None` objects it must increase their refcount. It must acquire the GIL when the refcount is increased. Currently we don't acquire the GIL when we do this, we only hold a SpinLock protecting other ArcticDB threads from racing on the GIL refcount. With this change we add an atomic variable in the PythonHandler data which will accumulate the refcount. Then at the end of the operation when we reacquire the GIL we will increase the refcount. The same is done for the NaN refcount, note that we don't really need the GIL to increase NaN's refcount as we create it internally and don't handle it to Python until the read operation is done. Currently only read operations need to work with the `None` object.

`apply_global_refcounts` must be called at the very end before passing the dataframe to python to prevent something raising an exception in after the refcount is applied but before python receives the data. Increasing None's refcount but never decreasing it doesn't seem to be fatal but we're trying to be good citizens. The best place for that is `adapt_read_df` or `adapt_read_dfs` as they are called at the end of all read functions. The code is changed so that the type handler data is created always in the python bindings file as it's easier to track.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/python/adapt_read_dataframe.hpp
+++ b/cpp/arcticdb/python/adapt_read_dataframe.hpp
@@ -9,10 +9,14 @@
 
 #include <arcticdb/python/python_utils.hpp>
 #include <arcticdb/entity/read_result.hpp>
+#include <arcticdb/python/python_handler_data.hpp>
 
 namespace arcticdb {
 
-inline auto adapt_read_df = [](ReadResult && ret) -> py::tuple{
+inline py::tuple adapt_read_df(ReadResult&& ret, std::pair<std::any&, OutputFormat>* const handler) {
+    if (handler) {
+        apply_global_refcounts(handler->first, handler->second);
+    }
     auto pynorm = python_util::pb_to_python(ret.norm_meta);
     auto pyuser_meta = python_util::pb_to_python(ret.user_meta);
     auto multi_key_meta = python_util::pb_to_python(ret.multi_key_meta);
--- a/cpp/arcticdb/python/python_handler_data.hpp
+++ b/cpp/arcticdb/python/python_handler_data.hpp
@@ -2,7 +2,6 @@
 
 #include <pybind11/pybind11.h>
 #include <folly/ThreadCachedInt.h>
-#include <arcticdb/python/python_utils.hpp>
 
 #include <memory>
 
@@ -59,15 +58,6 @@ struct PythonHandlerData {
             Py_INCREF(py_nan_->ptr());
         }
     }
-
-    ~PythonHandlerData() {
-        internal::check<ErrorCode::E_ASSERTION_FAILURE>(none_refcount_, "None refcount must not be null");
-        const size_t none_count = none_refcount_->readFull();
-        internal::check<ErrorCode::E_ASSERTION_FAILURE>(none_count == 0, "None refcount not applied. {} more to be applied", none_count);
-        internal::check<ErrorCode::E_ASSERTION_FAILURE>(nan_refcount_, "None refcount must not be null");
-        const size_t nan_count = none_refcount_->readFull();
-        internal::check<ErrorCode::E_ASSERTION_FAILURE>(nan_count == 0, "NaN refcount not applied. {} more to be applied", nan_count);
-    }
 private:
     std::shared_ptr<folly::ThreadCachedInt<uint64_t>> none_refcount_ = std::make_shared<folly::ThreadCachedInt<uint64_t>>();
     std::shared_ptr<folly::ThreadCachedInt<uint64_t>> nan_refcount_ = std::make_shared<folly::ThreadCachedInt<uint64_t>>();
@@ -84,7 +74,7 @@ inline void apply_global_refcounts(std::any& handler_data, OutputFormat output_f
 
 struct PythonHandlerDataFactory  : public TypeHandlerDataFactory {
     std::any get_data() const override {
-        return std::any{PythonHandlerData()};
+        return {PythonHandlerData{}};
     }
 };
 
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -21,9 +21,6 @@
 #include <arcticdb/version/version_map_batch_methods.hpp>
 #include <arcticdb/util/container_filter_wrapper.hpp>
 #include <arcticdb/util/allocation_tracing.hpp>
-#include <bitset>
-
-#include "python/python_handler_data.hpp"
 
 namespace arcticdb::version_store {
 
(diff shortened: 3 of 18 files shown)


cpp/arcticdb/python/python_handler_data.hpp
```diff
@@ -24,19 +25,64 @@ struct PythonHandlerData {
...
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(none_refcount_, "None refcount must not be null");
+        const size_t none_count = none_refcount_->readFull();
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(none_count == 0, "None refcount not applied. {} more to be applied", none_count);
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(nan_refcount_, "None refcount must not be null");
+        const size_t nan_count = none_refcount_->readFull();
+        internal::check<ErrorCode::E_ASSERTION_FAILURE>(nan_count == 0, "NaN refcount not applied. {} more to be applied", nan_count);
+    }
+private:
+    std::shared_ptr<folly::ThreadCachedInt<size_t>> none_refcount_ = std::make_shared<folly::ThreadCachedInt<size_t>>();
+    std::shared_ptr<folly::ThreadCachedInt<size_t>> nan_refcount_ = std::make_shared<folly::ThreadCachedInt<size_t>>();
```
alexowens90, line 70: Why are these shared pointers? `uint64_t` would be more portable than `size_t` as well

cpp/arcticdb/python/python_utils.cpp
```diff
@@ -1,10 +1,32 @@
 #include <arcticdb/python/python_utils.hpp>
+#include <arcticdb/python/python_handler_data.hpp>
 
 namespace arcticdb::python_util {
-void increment_none_refcount(size_t amount, SpinLock& lock) {
-    std::lock_guard guard(lock);
+void increment_none_refcount(size_t amount) {
```
alexowens90, line 5: This is now only called from `PythonHandlerData`. I think it would make the intended use (i.e. that all `None` increfs go through the handler) clearer if we just move the implementation into that class and remove this free function.

cpp/arcticdb/python/python_utils.cpp
```diff
@@ -1,10 +1,32 @@
...
+    std::fill_n(ptr_dest, num_rows, Py_None);
+
+    if(inc_ref_count == IncrementRefCount::ON) {
+        const auto none_count = num_rows - sparse_count;
+        python_handler_data.increment_none_refcount(none_count);
+    }
+}
+
+PyObject** fill_with_none(PyObject** ptr_dest, size_t count, PythonHandlerData& handler_data) {
+    std::fill_n(ptr_dest, count, nullptr);
```
alexowens90, line 27: Why is this `nullptr`?

cpp/arcticdb/python/python_handler_data.hpp
vasil-pashov: This class is them wrapped in std::any which requires all members to be copy-assignable and  ThreadCachedInt is not.

alexowens90 (approved):

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -797,6 +797,7 @@ std::vector<std::variant<ReadResult, DataError>> PythonVersionStore::batch_read(
                 }
                 );
     }
+    apply_global_refcounts(handler_data, read_options.output_format());
```
poodlewars, line 800: Is this and the couple of places in `python_bindings.cpp` the only places where we create Python `None` objects in the native layer?

python/tests/stress/arcticdb/test_stress_strings.py
poodlewars, line 1: I think we should have tests for all our APIs that can allocate `None`s - like `batch_read` and the things in `python_bindings.cpp` that you touched as well

cpp/arcticdb/version/version_store_api.cpp
vasil-pashov: Yes, read methods are affected: `read`, `read_batch`, `batch_read_keys`, `read_index`, `read_column_stats_version`, `LibraryTool::read`

python/tests/stress/arcticdb/test_stress_strings.py
vasil-pashov: Good idea, `batch_read` is easy to test. I'm not sure how the other methods like: `read_index`, `batch_read_keys`, `read_column_stats_version` and `LibraryTool::read` can end up with None values. I suspect that only `read` and `batch_read` can contain user defined strings that can be None.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1109,14 +1111,20 @@ VersionedItem LocalVersionedEngine::defragment_symbol_data(const StreamId& strea
...
+        read_futures.reserve(keys.size());
+        for (const auto& index_key: keys) {
+            read_futures.emplace_back(read_frame_for_version(store(), {index_key}, std::make_shared<ReadQuery>(), ReadOptions{}, handler_data));
+        }
+        Allocator::instance()->trim();
+        result = folly::collect(read_futures).get();
     }
-    Allocator::instance()->trim();
-    return folly::collect(res).get();
+    apply_global_refcounts(handler_data, OutputFormat::PANDAS);
```
IvoDD, line 1125: Couldn't theoretically code calling `batch_read_keys` throw an exception before giving the resulting object to python?
In this case None's ref count will never reach zero, which I'm not sure how big of a deal is.

Ideally we will call `apply_global_refcounts` right before returing objects to python (similar to the python bindings).

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -678,8 +680,10 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
              py::call_guard<SingleThreadMutexHolder>(), "Write a specific  version of this dataframe to the store")
         .def("read_dataframe_version",
              [&](PythonVersionStore& v,  StreamId sid, const VersionQuery& version_query, const std::shared_ptr<ReadQuery>& read_query, const ReadOptions& read_options) {
-                auto handler_data = TypeHandlerRegistry::instance()->get_handler_data(read_options.output_format());
-                return adapt_read_df(v.read_dataframe_version(sid, version_query, read_query, read_options, handler_data));
+                 auto handler_data = TypeHandlerRegistry::instance()->get_handler_data(read_options.output_format());
+                 py::list result = adapt_read_df(v.read_dataframe_version(sid, version_query, read_query, read_options, handler_data));
+                 apply_global_refcounts(handler_data, read_options.output_format());
```
IvoDD, line 685: I feel like this approach to call `apply_global_refcounts` at the end of any python binding which might need to deal with `None` objects is error prone.

E.g. looking at the call hierarchy of `decode_or_expand` which the place we allocate the `None`s I see that `read_dataframe_from_file` python binding can allocate `None`s but doesn't call `apply_global_refcounts`.

What do you think about calling the `apply_global_refcounts` inside `adapt_read_df` and `adapt_read_dfs`? This will be a good place for it because we need to call the `adapt` functions every time when returning something dataframe like to python. It will mean we call it cases we don't need it like `restore_version` but applying 0 refcounts should be fine?

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -23,6 +23,8 @@
 #include <arcticdb/version/schema_checks.hpp>
 #include <arcticdb/util/pybind_mutex.hpp>
 
+#include "python/python_handler_data.hpp"
```
IvoDD, line 26: Nit: Use `<>` include?

IvoDD (commented):
Only important comments are with regards to when to call `apply_global_refcounts`

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -779,7 +783,9 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
                  std::vector<std::shared_ptr<ReadQuery>>& read_queries,
                  const ReadOptions& read_options){
                  auto handler_data = TypeHandlerRegistry::instance()->get_handler_data(read_options.output_format());
-                 return python_util::adapt_read_dfs(v.batch_read(stream_ids, version_queries, read_queries, read_options));
+                 py::list result = python_util::adapt_read_dfs(v.batch_read(stream_ids, version_queries, read_queries, read_options));
+                 apply_global_refcounts(handler_data, read_options.output_format());
```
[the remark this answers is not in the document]
IvoDD: Actually no, we apply it in `batch_read_keys`. But still highlights the point that it's hard to follow when we should apply it

cpp/arcticdb/python/adapt_read_dataframe.hpp
```diff
@@ -9,10 +9,14 @@
 
 #include <arcticdb/python/python_utils.hpp>
 #include <arcticdb/entity/read_result.hpp>
+#include <arcticdb/python/python_handler_data.hpp>
 
 namespace arcticdb {
 
-inline auto adapt_read_df = [](ReadResult && ret) -> py::tuple{
+inline py::tuple adapt_read_df(ReadResult&& ret, std::pair<std::any&, OutputFormat>* const handler) {
```
IvoDD, line 16: The `std::pair<std::any&, OutputFormat>*` feels a little complicated and hard to reason about what is expected.
I think it would be better to just have two arguments:
`std::optional<std::any&> handler_data, OutputFormat output_format`.

Anyway, I'll soon be including the `OutputFormat` inside the `ReadResult` in arrow work so this will get cleaner, so happy to leave as is.

vasil-pashov: At the moment OutputFormat doesn't make any sense without `handler_data` that's why they are paired together.

willdealtry (approved):

Issue #2251, opened 2025-03-17
Cannot connect to backblaze b2, got "Unexpected error: S3Error#100 MissingContentLength: Unable to parse ExceptionName: MissingContentLength Message: You must provide the Content-Length HTTP header. for object '_arctic_cfg/cref/*sUt*test'"


tddschn:
### Describe the bug

```python3
import arcticdb as adb

db = adb.Arctic(url := get_s3_url())
print(url)
db.create_library("test")
```
gives error `Unexpected error: S3Error#100 MissingContentLength: Unable to parse ExceptionName: MissingContentLength Message: You must provide the Content-Length HTTP header. for object '_arctic_cfg/cref/*sUt*test'`

### Steps/Code to Reproduce

I'm using a new backblaze bucket created today, which they claim to be 'S3-compatible', and got the keys from https://secure.backblaze.com/app_keys.htm .

.env:

```
# https://secure.backblaze.com/app_keys.htm
B2_ENDPOINT=https://s3.us-west-000.backblazeb2.com
# redacted keys
B2_KEY_ID=key_id 
B2_ACCESS_KEY=key

# https://github.com/man-group/ArcticDB/blob/8635f8c2fd70ef5693dbe094096d338e62f9ee5b/docs/mkdocs/docs/runtime_config.md?plain=1#L104
# debug logging
ARCTICDB_AWS_LogLevel_int=6
```

```python
def get_s3_url(
    env_var_prefix: str = "B2", bucket: str = 'my-bucket', use_s3s: bool = True
) -> str:
    import os
    from dotenv import load_dotenv

    load_dotenv()
    _access_key = os.getenv(f"{env_var_prefix}_ACCESS_KEY")
    _key_id = os.getenv(f"{env_var_prefix}_KEY_ID")
    _endpoint = os.getenv(f"{env_var_prefix}_ENDPOINT")
    return f"s3://{_endpoint}:{bucket}?access={_key_id}&secret={_access_key}"

import arcticdb as adb

db = adb.Arctic(url := get_s3_url())
print(url)
db.create_library("test")

```

Output:

```
s3://https://s3.us-west-000.backblazeb2.com:my-bucket?access=<>&secret=<> # redacted
20250317 19:08:12.307973 83073196 E arcticdb | Unexpected error: S3Error#100 MissingContentLength: Unable to parse ExceptionName: MissingContentLength Message: You must provide the Content-Length HTTP header. for object '_arctic_cfg/cref/*sUt*test'
[...]

PR #2337, opened 2025-04-22
Disable s3 checksumming

phoebusm:
#### Reference Issues/PRs

https://github.com/man-group/ArcticDB/issues/2251
Disable s3 checksumming by setting environment variable in the wheel.

#### Any other comments?
This will also unblock the upgrade of `aws-sdk-cpp` on vcpkg. 
The upgrade will not be made in this PR

One of the newly added test is needed to be skipped as `conda` CI has `aws-sdk-cpp` pinned at non-s3-checksumming version due the `libarrow` pin. 
`environment-dev.yml` doesn't align with the counterpart in the feedstock. Therefore the new version of `aws-sdk-cpp` is only used in the feedstock thus release wheel but not in local and CI build here. 
This will be addressed in separate ticket.

[Commit](https://github.com/man-group/ArcticDB/pull/2337/commits/245a02cd455e39fb8f976301ccd5409e6ae88b13) to remove `libarrow` pin so more updated `aws-sdk-cpp`, which support s3 checksumming is in used in conda
It's for verifying the change with the newly added the test. The [test](https://github.com/man-group/ArcticDB/actions/runs/14732394443/job/41349695905) is successful.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/s3/s3_storage.hpp
+++ b/cpp/arcticdb/storage/s3/s3_storage.hpp
@@ -250,7 +250,8 @@ inline Aws::Client::ClientConfiguration get_proxy_config(Aws::Http::Scheme endpo
     }
 }
 // Since aws-sdk-cpp >= 1.11.486, it has turned on checksum integrity check `x-amz-checksum-mode: enabled`
-// This feature is not supported in many s3 implementation and causes error
+// This feature is not supported in many s3 implementations. SDK with this feature ON will reject responses from
+// s3 implementations that do not provide checksum, which leads to storage exception in arcticdb
 // Update environment variable before import arcticdb_ext to disable checksum validation
 inline void configure_s3_checksum_validation() {
     const char* response_checksum = std::getenv("AWS_RESPONSE_CHECKSUM_VALIDATION");
@@ -258,7 +259,9 @@ inline void configure_s3_checksum_validation() {
     
     if ((response_checksum && std::string(response_checksum) == "when_supported") || 
         (request_checksum && std::string(request_checksum) == "when_supported")) {
-        log::storage().warn("Checksum validation has been specifically enabled by user, which the endpoint may not support and causes error.");
+        log::storage().warn("S3 Checksum validation has been specifically enabled by user"
+                            "If endpoint doesn't support it, its response will not contain checksum"
+                            "which will be rejected by SDK and lead to storage exception in arcticdb");
     }
     else {
 #ifdef _WIN32
--- a/cpp/arcticdb/storage/s3/s3_storage.hpp
+++ b/cpp/arcticdb/storage/s3/s3_storage.hpp
@@ -249,9 +249,32 @@ inline Aws::Client::ClientConfiguration get_proxy_config(Aws::Http::Scheme endpo
         return Aws::Client::ClientConfiguration();
     }
 }
+// Since aws-sdk-cpp >= 1.11.486, it has turned on checksum integrity check `x-amz-checksum-mode: enabled`
+// This feature is not supported in many s3 implementation and causes error
+// Update environment variable before import arcticdb_ext to disable checksum validation
+inline void configure_s3_checksum_validation() {
+    const char* response_checksum = std::getenv("AWS_RESPONSE_CHECKSUM_VALIDATION");
+    const char* request_checksum = std::getenv("AWS_REQUEST_CHECKSUM_CALCULATION");
+    
+    if ((response_checksum && std::string(response_checksum) == "when_supported") || 
+        (request_checksum && std::string(request_checksum) == "when_supported")) {
+        log::storage().warn("Checksum validation has been specifically enabled by user, which the endpoint may not support and causes error.");
+    }
+    else {
+#ifdef _WIN32
+        _putenv_s("AWS_RESPONSE_CHECKSUM_VALIDATION", "when_required");
+        _putenv_s("AWS_REQUEST_CHECKSUM_CALCULATION", "when_required");
+#else
+        setenv("AWS_RESPONSE_CHECKSUM_VALIDATION", "when_required", 1);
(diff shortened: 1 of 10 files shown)


python/arcticdb/__init__.py
```diff
@@ -1,5 +1,18 @@
-import arcticdb_ext as _ext
+import logging as _logging
 import os as _os
+
+# Since aws-sdk-cpp >= 1.11.486, it has turned on checksum integrity check `x-amz-checksum-mode: enabled`
+# This feature is not supported in many s3 implementation and causes error
+# Update environment variable before import arcticdb_ext to disable checksum validation
```
poodlewars, line 6: Is there any way to add these hooks when `arcticdb_ext` is imported? May be more robust - nothing stops a user from importing `arcticdb_ext` first themselves!

python/arcticdb/__init__.py
```diff
@@ -1,5 +1,18 @@
...
+if _os.getenv("AWS_RESPONSE_CHECKSUM_VALIDATION") == "when_supported" or _os.getenv("AWS_REQUEST_CHECKSUM_CALCULATION") == "when_supported":
+    _logging.getLogger(__name__).warning(
+        "Checksum validation has been specfically enabled by user, which the endpoint may not support and causes error."
```
poodlewars, line 9: Similar to the comment above, this doesn't make sense

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -1335,3 +1337,28 @@ def test_norm_failure_error_message(arctic_library):
...
+
+
+@pytest.mark.parametrize(
+    "multiprocess", [
+        "spawn",
+        pytest.param("fork", marks=FORK_SUPPORTED),
+        pytest.param("forkserver", marks=FORK_SUPPORTED),
+    ]
+)
+def test_s3_checksum_off_by_env_var(s3_storage, lib_name, multiprocess):
```
poodlewars, line 1355: I don't understand, this test doesn't have anything to do with the checksum?

poodlewars, line 1355: Should check that the fixture actually fails in the way you expect when the checksum is enabled

poodlewars (changes requested):
-

python/arcticdb/__init__.py
phoebusm: Such usage will be quite cursed as I doubt anyone would import `arcticdb_ext` only and able to make a successful connection to s3 endpoint. Basically it would require the user to have their own python layer I think.

phoebusm: To move the hooks to `arcticdb_ext` level, the env var magic needs move to c++ layer I think

phoebusm:
[Commit](https://github.com/man-group/ArcticDB/pull/2337/commits/245a02cd455e39fb8f976301ccd5409e6ae88b13) to remove `libarrow` pin so more updated `aws-sdk-cpp`, which support s3 checksumming is in used in conda
It's for verifying the change with the newly added the test. The [test](https://github.com/man-group/ArcticDB/actions/runs/14732394443/job/41349695905) is successful.

cpp/arcticdb/storage/s3/s3_storage.hpp
```diff
@@ -249,9 +249,35 @@ inline Aws::Client::ClientConfiguration get_proxy_config(Aws::Http::Scheme endpo
...
+// Update environment variable before import arcticdb_ext to disable checksum validation
+inline void configure_s3_checksum_validation() {
+    const char* response_checksum = std::getenv("AWS_RESPONSE_CHECKSUM_VALIDATION");
+    const char* request_checksum = std::getenv("AWS_REQUEST_CHECKSUM_CALCULATION");
+    
+    if ((response_checksum && std::string(response_checksum) == "when_supported") || 
+        (request_checksum && std::string(request_checksum) == "when_supported")) {
+        log::storage().warn("S3 Checksum validation has been specifically enabled by user"
+                            "If endpoint doesn't support it, its response will not contain checksum"
+                            "which will be rejected by SDK and lead to storage exception in arcticdb");
```
poodlewars, line 264: You're missing a space between `checksum` and `which`. We've actually seen worse failures than this where `PutObject` calls _succeed_ but write garbage to the storage, so the description of the impact isn't quite correct.

cpp/arcticdb/storage/s3/s3_storage.hpp
```diff
@@ -249,9 +249,35 @@ inline Aws::Client::ClientConfiguration get_proxy_config(Aws::Http::Scheme endpo
...
+#else
+        setenv("AWS_RESPONSE_CHECKSUM_VALIDATION", "when_required", 1);
+        setenv("AWS_REQUEST_CHECKSUM_CALCULATION", "when_required", 1);
+#endif
+    }
+}
 
 template<typename ConfigType>
 auto get_s3_config(const ConfigType& conf) {
+    configure_s3_checksum_validation();
```
poodlewars, line 279: It seems weird to me to introduce this side-effect in `get_s3_config` which I would expect to be a pure function, shouldn't we push it up a layer?

phoebusm: I prefer not pushing it by a level as the env var is mandatory for all s3 clients in our code space. I rather rename `get_s3_config` to clarify what it does now

cpp/arcticdb/storage/s3/s3_storage.hpp
phoebusm: I haven't run into the garbage issue. Will update

poodlewars (changes requested):

poodlewars (approved):

maxim-morozov (approved):

PR #2339, opened 2025-04-23
Installation tests v4

grusev:

Successful execution 5.2.6: https://github.com/man-group/ArcticDB/actions/runs/14641126753/job/41083591802
5.1.2: https://github.com/man-group/ArcticDB/actions/runs/14637571996
4.5.1: https://github.com/man-group/ArcticDB/actions/runs/14639124835/job/41077126258
1.6.2: https://github.com/man-group/ArcticDB/actions/runs/14701046721/job/41250511273

The PR contains workflow definition to execute tests on installed arcticdb it is combination of approaches:

https://github.com/man-group/ArcticDB/pull/2330
https://github.com/man-group/ArcticDB/pull/2316

Installation tests are now in separate folder (python/installation_tests) not part of tests. They have their own fixtures, making them independent from rest of code base

The tests are direct copy from originals with one modified to user ver 2 API. Otherwise now if there are changes in API each test in installation set can be addapted. As tests run very fast no need to use simulators, instead directly using S3 real storage

The tests are executed by a workflow. 

Currently each test is executed against LMDB and real S3. The moto simulated version is not available in this moment due to tight coupling with protobufs which differ for ach version as well as tight coupling with whole existing test code.

The workflow have 2 triggers:

 - manual trigger - allowing tests to be executed manually on demand
 - on schedule - the schedule execution is overnight. Each arcticdb version tests are executed within 1hr difference from the other. Thats is due to fact that executing all at once is likely to generate errors with real storages

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/installation_tests/environment_unix.yml
```diff
@@ -0,0 +1,9 @@
+name: arcticdb
+channels:
+  - conda-forge
+dependencies:
+  - pytest
+  - wheel
+  - arcticdb
+  - setuptools
```
G-D-Petrov, line 7: nit: do we really need wheel and setuptools here?

.github/workflows/execute_installation_tests_v4.yml
```diff
@@ -0,0 +1,210 @@
+name: Run Installation Tests v4
+
+on:
+  push:
+    branches: 
+      - installation_tests_v4
```
G-D-Petrov, line 6: I assume that this will run daily(?) and only on master(?) so a schedule trigger needs to be added for that.

.github/workflows/execute_installation_tests_v4.yml
```diff
@@ -0,0 +1,210 @@
...
+          - "3.11"
+          - "3.12"
+          - "3.13"
+      arcticdb_version:
+        description: 'ArcticDB version to test'
+        required: false
+        default: 'latest'
+        type: choice
+        options:
+          - "latest"
```
G-D-Petrov, line 39: nit: for both the python version and the arcticdb versions we will have to add new ones here manually.
It'd be nice to have a script "grep" the latest version on each major/minor version branch instead of hard coding them here.
Something similar is already done here: https://github.com/man-group/ArcticDB/blob/master/.github/workflows/analysis_workflow.yml#L48

.github/workflows/execute_installation_tests_v4.yml
```diff
@@ -0,0 +1,210 @@
...
+
+jobs:
+
+  test:
+    runs-on: ${{ inputs.os != 'all' && inputs.os || matrix.os }}
+    strategy:
+      matrix:
+        #os: [ubuntu-22.04, windows-latest, macos-latest]
+        #python: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        include:
```
G-D-Petrov, line 72: I'd suggest creating the matrix with lists as above.
The lists will probably need to be initialized by a previous step/job because inputs only work with manually triggered (aka `workflow_dispatch`) runs..

G-D-Petrov (approved):

.github/workflows/installation_tests.yml
```diff
@@ -0,0 +1,235 @@
+name: Installation Tests Execution
+
+on:
+  push:
+    branches: 
+      - installation_tests_v4
```
alexowens90, line 6: Guess this will need backing out before merge?

.github/workflows/installation_tests.yml
```diff
@@ -0,0 +1,235 @@
...
+          - os: ubuntu-22.04
+            use_conda: "no"
+            python: "3.13"
+      fail-fast: false  
+    env:
+      install_tests_folder: "python/tests/compat"
+      conda_file: "environment_unix.yml"
+      arcticdb_version: ${{ inputs.arcticdb_version || '1.6.2' }}
+      lmdb: ${{ inputs.run_on_lmdb || '1' }}
+      real_s3: ${{ inputs.run_on_real_s3 || '1' }}
```
alexowens90, line 111: Shouldn't this default to zero to match the behaviour described above on line 65?

.github/workflows/installation_tests.yml
```diff
@@ -0,0 +1,235 @@
...
+      if: github.event_name == 'schedule'
+      run: |
+        hour=$(date -u +"%H")
+        case $hour in
+          "01") echo "arcticdb_version='1.6.2'" >> $GITHUB_ENV;;
+          "02") echo "arcticdb_version='4.5.1'" >> $GITHUB_ENV;;
+          "03") echo "arcticdb_version='5.1.2'" >> $GITHUB_ENV;;
+          "04") echo "arcticdb_version='5.2.6'" >> $GITHUB_ENV;;
+          "05") echo "arcticdb_version='latest'" >> $GITHUB_ENV;;
+        esac
```
alexowens90, line 148: This is a very odd way to implement this. Can't each cron job be passed a different parameter?

python/tests/compat/requirements-1.6.2.yml
alexowens90, line 1: Presumably all of these requirements files have been copy-pasted from the relevant tag?

python/installation_tests/test_installation.py
```diff
@@ -0,0 +1,474 @@
...
+from arcticdb.util.test import (
+    assert_frame_equal,
+    random_strings_of_length,
+    random_floats,
+)
+from arcticdb.version_store.processing import QueryBuilder
+
+
+PRE_4_X_X = arcticdb.__version__ < "4.0.0"
+PRE_5_X_X = arcticdb.__version__ < "5.0.0"
```
alexowens90, line 36: https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/util/_versions.py
This is the Pythonic way to check version numbers

.github/workflows/installation_tests.yml
grusev: this is the easiest and cleanest way to do this although that it might look odd. The other way to separate into 2 workflows and have one to invoke other workflow via rest call is not only unreadable but also I could not make it run, which means that it is much harder to maintin.

I lack experience for GH so I cannot say if there is no better and cleaner alternative but this one is perhaps cleanest and in one workflow you can see everything and not have to jump here and there.

Thus I do not have an alternative solution. If there happens to be one in future we can replace current one

alexowens90 (approved):

PR #2340, opened 2025-04-24
increase file descriptors limits for tests

grusev:

This is needed for real S3 and GCP tests as file descriptors required are more than initial limit of 100k
Link to S3 run (manual):  https://github.com/man-group/ArcticDB/actions/runs/14643182856

A run with 1M file descriptors on Linux: https://github.com/man-group/ArcticDB/actions/runs/15393174400/job/43307772735
No errors because of filedescriptors. Compat testst fail due to fact that GCP symbol names cannot have certain characters. Similar issues exists and this will be added to it and test will be xfailed

A run with 1M filedescriptors + compat test xfail: https://github.com/man-group/ArcticDB/actions/runs/15409225380

After making changes only for GCP: https://github.com/man-group/ArcticDB/actions/runs/15435539023

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


poodlewars:
I really don't think we should do this. Why are we using so many descriptors in the first place?

poodlewars:
I'd rather someone digs in to the root cause. Concerned about the ulimit complexity hiding there and the added complexity in our CI. GCP is not heavily used - there's no need for this kind of hotfix IMO. Could also hide real problems if we start opening loads of file descriptors but our CI uses unusual settings that can handle them.

.github/workflows/build_steps.yml
```diff
@@ -361,14 +361,28 @@ jobs:
...
+          echo "Persistent Storage: ${{ inputs.persistent_storage }}"
+          echo "Running on OS: $RUNNER_OS"
+
+          if [[ "${{ inputs.persistent_storage }}" == "GCPXML" ]]; then
+            if [[ "$RUNNER_OS" == "Windows" ]]; then
+              echo "Running Windows-specific tasks"
+              build_tooling/parallel_test.sh tests/${{matrix.type}}
+            else
+              echo "Running tasks for other OS"
+              prlimit --nofile=1000000 -- build_tooling/parallel_test.sh tests/${{matrix.type}}
```
phoebusm, line 373: The default file descriptor soft limit of github runner is 1048576. Giving it a lower setting can only make things worse

grusev:
skip for now until time comes

grusev:
not needed for now

PR #2344, opened 2025-04-28
Delete STS roles regularly 

grusev:

Due to limitation of STS roles number we should constantly do cleaning of failed to delete roles. The PR contains a scheduled job that would do that every Sa. The python script can also be executed at any time and will delete only roles created prior of today, leaving all currently running jobs unaffected

As roles cannot be guaranteed to be cleaned after tests execution due to many factors, we should take them out on regular bases, and perhaps this is the quickest and most reliable approach

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/utils/s3_roles_delete.py
```diff
@@ -0,0 +1,59 @@
...
+        iam_client.remove_role_from_instance_profile(InstanceProfileName=profile['InstanceProfileName'], RoleName=role_name)
+        iam_client.delete_instance_profile(InstanceProfileName=profile['InstanceProfileName'])        
+
+    try:
+        iam_client.delete_role(RoleName=role_name)
+        print("Role deleted successfully.")
+    except Exception:
+        print("Error deleting role")  # Role could be non-existent as creation of it may fail
+
+prefix = "gh_sts_test_role_" # This is the template name prefix of roles that GH runners are creating
```
phoebusm, line 45: Maybe it's worth importing the prefix from/to the normal testing pipeline

python/utils/s3_roles_delete.py
```diff
@@ -0,0 +1,59 @@
...
+client = boto_client()
+roles = list_roles_by_prefix(client, prefix)
+print(f"Found {len(roles)} roles")
+
+for i, role in enumerate(roles):
+    if datetime.today().strftime('%Y-%m-%d') in role:
+        print(f"Role {role} is from today, skipping it.")
+    else:
+        print(f"{i} DELETE role {role}. An old role")
+        delete_role(client, role)
```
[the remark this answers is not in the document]
grusev: addressed! We had >3000 active users!

python/utils/s3_roles_delete.py
grusev: addressed

poodlewars (approved):
This looks low risk so approving. Needs Phoebus' approval too.

phoebusm (approved):

PR #2345, opened 2025-05-07
File operations

willdealtry:
Add a Python front-end and some basic tests for the file operations so that this functionality doesn't rot

### DIFF
--- a/cpp/arcticdb/util/memory_mapped_file.hpp
+++ b/cpp/arcticdb/util/memory_mapped_file.hpp
@@ -13,7 +13,7 @@ class MemoryMappedFile {
 private:
     HANDLE file_ = INVALID_HANDLE_VALUE;
     HANDLE map_ = nullptr;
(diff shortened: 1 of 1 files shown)


python/arcticdb/file.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+
+    return udm, item, norm_meta
+
+
+def _denormalize_stateless(item: Any, norm_meta: Any, normalizer: Any) -> Any:
+    return normalizer.denormalize(item, norm_meta)
+
+
+def to_file(symbol: str, data: Any, file_path: str, metadata: Optional[Any] = None, **kwargs) -> VersionedItem:
```
alexowens90, line 40: Shall we make these functions private to discourage their use?

alexowens90 (approved):

python/arcticdb/file.py
willdealtry: Good idea

PR #2351, opened 2025-05-13
Bugfix/9123099670/fix resampling of old updated data

alexowens90:
#### Reference Issues/PRs
Fixes [9123099670](https://man312219.monday.com/boards/7852509418/views/168855452/pulses/9123099670)

Fixes three separate resampling bugs:

1. Old versions of `update` (changed sometime between `4.1.0` and `4.4.0`, I haven't pinned down exactly where) had a behaviour in which the `end_index` value in the data key of the segment overlapping with the start of the date range provided to the `update` call was set to the first value of the date range in the `update` call. For all other modification methods, this is set to 1 nanosecond larger than the last index value in the contained segment. Resampling assumed this to be the case, and had an assertion verifying it. Relaxing this assertion is sufficient to fix the issue.
2. Providing a `date_range` argument with a resample where the provided date range did not overlap with the timerange covered by the index of the symbol led to trying to reserve a vector with a negative size. This now correctly returns an empty result.
3. Previously, checks that a symbol being resampled had a timestamp index occurred after some operations which also require this to be true, which could lead to the same vector reserve issue above. It is now checked in advance, and a suitable exception raised.

### DIFF
--- a/cpp/arcticdb/processing/clause.cpp
+++ b/cpp/arcticdb/processing/clause.cpp
@@ -739,9 +739,12 @@ std::vector<EntityId> ResampleClause<closed_boundary>::process(std::vector<Entit
     internal::check<ErrorCode::E_ASSERTION_FAILURE>(is_time_type(first_row_slice_index_col.type().data_type()),
                                                     "Cannot resample data with index column of non-timestamp type");
     auto first_ts = first_row_slice_index_col.scalar_at<timestamp>(0).value();
-    // We can use the last timestamp from the first row-slice's index column, as by construction (structure_for_processing) the bucket covering
-    // this value will cover the remaining index values this call is responsible for
-    auto last_ts = first_row_slice_index_col.scalar_at<timestamp>(first_row_slice_index_col.row_count() - 1).value();
+    // If there is only one row slice, then the last index value of interest is just the last index value for this row
+    // slice. If there is more than one, then the first index value from the second row slice must be used to calculate
+    // the buckets of interest, due to an old bug in update. See test_compatibility.py::test_compat_resample_updated_data
+    // for details
+    auto last_ts = row_slices.size() == 1 ? first_row_slice_index_col.scalar_at<timestamp>(first_row_slice_index_col.row_count() - 1).value():
+            row_slices.back().segments_->at(0)->column(0).scalar_at<timestamp>(0).value();
     auto bucket_boundaries = generate_bucket_boundaries(first_ts, last_ts, responsible_for_first_overlapping_bucket);
     if (bucket_boundaries.size() < 2) {
         return {};
--- a/cpp/arcticdb/processing/clause.cpp
+++ b/cpp/arcticdb/processing/clause.cpp
@@ -631,6 +631,10 @@ std::vector<std::vector<size_t>> ResampleClause<closed_boundary>::structure_for_
     if (ranges_and_keys.empty()) {
         return {};
     }
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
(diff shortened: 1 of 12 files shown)


vasil-pashov (approved):
I think it's worth adding a tests in update to make sure the `end_index` value will not regress again.

cpp/arcticdb/processing/clause.cpp
```diff
@@ -653,6 +653,7 @@ std::vector<std::vector<size_t>> ResampleClause<closed_boundary>::structure_for_
 
     bucket_boundaries_ = generate_bucket_boundaries_(date_range_->first, date_range_->second, rule_, closed_boundary, offset_, origin_);
     if (bucket_boundaries_.size() < 2) {
```
poodlewars, line 655: Are there any valid cases where we end up with a single element in `bucket_boundaries_`?

python/tests/compat/arcticdb/test_compatibility.py
```diff
@@ -313,3 +314,28 @@ def test_storage_mover_clone_old_library(old_venv_and_arctic_uri, lib_name):
...
+        # Write to library using old version
+        compat.old_lib.write(sym, df_0)
+        compat.old_lib.update(sym, df_1, '(pd.Timestamp("2025-01-03 00:01:00"), None)')
+        compat.old_lib.update(sym, df_2, '(pd.Timestamp("2025-01-04 00:01:00"), None)')
+
+        # Resample using current version
+        with compat.current_version() as curr:
+            q = QueryBuilder().date_range((pd.Timestamp("2025-01-01"), pd.Timestamp("2025-01-04 23:59"))).resample("1D").agg({"col": "sum"})
+            curr.lib.read(sym, query_builder=q)
+            # TODO: Check that result is correct
```
poodlewars, line 341: TODO

poodlewars (changes requested):
Alex and I spoke a lot about the first bug - since the broken ArcticDB clients would still write timestamps on the data keys that formed a covering set of the timeseries it should be bearable to just relax the restriction.

Marking as request changes just because of the TODO in the compat test.

cpp/arcticdb/processing/clause.cpp
alexowens90: Nope, index values have to be between 2 bucket boundaries to be included in the resample, so only having one doesn't make any sense

poodlewars (approved):

cpp/arcticdb/processing/clause.cpp
```diff
@@ -739,9 +739,12 @@ std::vector<EntityId> ResampleClause<closed_boundary>::process(std::vector<Entit
     internal::check<ErrorCode::E_ASSERTION_FAILURE>(is_time_type(first_row_slice_index_col.type().data_type()),
                                                     "Cannot resample data with index column of non-timestamp type");
     auto first_ts = first_row_slice_index_col.scalar_at<timestamp>(0).value();
-    // We can use the last timestamp from the first row-slice's index column, as by construction (structure_for_processing) the bucket covering
-    // this value will cover the remaining index values this call is responsible for
-    auto last_ts = first_row_slice_index_col.scalar_at<timestamp>(first_row_slice_index_col.row_count() - 1).value();
+    // If there is only one row slice, then the last index value of interest is just the last index value for this row
+    // slice. If there is more than one, then the first index value from the second row slice must be used to calculate
+    // the buckets of interest, due to an old bug in update. See test_compatibility.py::test_compat_resample_updated_data
```
poodlewars, line 744: I think you should describe, here or in the test, exactly what the old bug in update was

PR #2353, opened 2025-05-13
Patch for Real S3 library names

grusev:

Currently we create library names which are too long for real S3, this is a patch for the tests until the real bug is addressed

Manually triggered run: https://github.com/man-group/ArcticDB/actions/runs/15013824867

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -102,7 +102,8 @@ def lib_name(request: "pytest.FixtureRequest") -> str:
     name = re.sub(r"[^\w]", "_", request.node.name)[:30]
     pid = os.getpid()
     thread_id = threading.get_ident()
-    return f"{name}.{random.randint(0, 9999999)}_{pid}_{thread_id}_{datetime.utcnow().strftime('%Y-%m-%dT%H_%M_%S_%f')}_{uuid.uuid4()}"
+    # There is limit to the name length
+    return f"{name}_{pid}_{thread_id}_{datetime.utcnow().strftime('%Y-%m-%dT%H_%M_%S_')}_{uuid.uuid4()}"
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


python/tests/conftest.py
```diff
@@ -102,7 +102,8 @@ def lib_name(request: "pytest.FixtureRequest") -> str:
     name = re.sub(r"[^\w]", "_", request.node.name)[:30]
     pid = os.getpid()
     thread_id = threading.get_ident()
-    return f"{name}.{random.randint(0, 9999999)}_{pid}_{thread_id}_{datetime.utcnow().strftime('%Y-%m-%dT%H_%M_%S_%f')}_{uuid.uuid4()}"
+    # There is limit to the name length
+    return f"{name}_{pid}_{thread_id}_{datetime.utcnow().strftime('%Y-%m-%dT%H_%M_%S_')}_{uuid.uuid4()}"
```
G-D-Petrov, line 106: nit: We can remove the `_{pid}_{thread_id}` part as well

G-D-Petrov (approved):

python/tests/conftest.py
alexowens90: Or the UUID? That is a lot of characters just to add some randomness, which the random int will achieve in fewer characters.
Can always make it 8 random alphanumeric characters if we want to decrease the collision probability

grusev: The (.) removal was the problem - luckily yesterday I was looking at man group mongo tests and that iss why I know it. :-) Left a comment in the last commit also ...

UUID is for some special cases when random numbers are generated one and the same, that is why random number if removed

grusev: %f was the nanoseconds which should also help about 6 symbols and the random number which was 8 totals to 14 symbols. That should be ok for our S3 tests. We had at most 10 more than the quota of 254

grusev: Thus the UUID that is left really makes things unique. Other parts also meaningful when debugging (the time overall can be sacrifised further if we have to cut more)

alexowens90 (approved):

PR #2358, opened 2025-05-16
Add resolve_defaults back in as a static method of NativeVersionStore

alexowens90:
#### Reference Issues/PRs
Was removed in #2345 , but is needed at least by some internal tests, and technically constitutes an API break (although we don't expect anybody to be using it)


poodlewars (approved):
Would ideally have a test so that we don't repeat this in future

phoebusm (approved):

PR #2359, opened 2025-05-16
Symbol list and storage lock improvements

ognyanstoimenov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/failure_simulation.hpp
+++ b/cpp/arcticdb/storage/failure_simulation.hpp
@@ -23,6 +23,7 @@ namespace arcticdb {
 enum class FailureType : int {
     WRITE = 0,
     READ,
+    WRITE_SLOW,
     ITERATE,
     DELETE,
 };
@@ -64,8 +65,8 @@ struct FailureAction {
             description_(std::move(description)), proxy_(std::move(proxy)) {}
 
     template<typename Func>
-    FailureAction(Description description, Func func):
-            FailureAction(std::move(description), FunctionWrapper{func}.asSharedProxy()) {}
+    FailureAction(Description description, Func&& func):
+            FailureAction(std::move(description), FunctionWrapper{std::forward<Func>(func)}.asSharedProxy()) {}
 
     inline void operator()(FailureType type) const {
         proxy_(type);
@@ -77,29 +78,50 @@ inline std::ostream& operator<<(std::ostream& out, const FailureAction& action) 
     return out;
 }
 
-namespace action_factories { // To allow `using namespace`
+namespace action_factories {
+// To allow `using namespace`
 static inline const FailureAction no_op("no_op", [](FailureType){});
+
+static FailureAction::FunctionWrapper maybe_execute(double probability, FailureAction::FunctionWrapper&& func) {
+    util::check_arg(probability >= 0 && probability <= 1.0, "Bad probability: {}", probability);
+
+    return [probability, f = std::move(func)](FailureType type) mutable {
+        if (probability == 0) {
+            return;
+        }
+
+        if (probability == 1.0) {
+            std::invoke(std::move(f), type);
+            return;
+        }
+
+        thread_local std::once_flag flag;
+        std::call_once(flag, [seed = (uint64_t)(&type)]() { init_random(seed); });
+        auto nd = random_probability();
+        if (nd < probability) {
+            std::invoke(std::move(f), type);
+        }
+    };
+}
 
 /** Raises the given exception with the given probability. */
 template<class Exception = StorageException>
-static inline FailureAction fault(double probability = 1.0) {
-    util::check_arg(probability >= 0, "Bad probability: {}", probability);
-
-    if (probability >= 1.0) {
-        return {"raise",  [](FailureType failure_type) {
-            throw Exception(fmt::format("Simulating {} storage failure", failure_type));
-        }};
-    } else {
-        return {fmt::format("fault({})", probability),
-                [prob=probability](FailureType failure_type) {
-                    thread_local std::once_flag flag;
-                    std::call_once(flag, [seed = uint64_t(&failure_type)]() { init_random(seed); });
-                    if (random_probability() < prob) {
-                        throw Exception(fmt::format("Simulating {} storage failure", failure_type));
-                    }
-                }};
-    }
+static  FailureAction fault(double probability = 1.0) {
+    return {fmt::format("fault({})", probability), maybe_execute(probability, [](FailureType failure_type) {
+                throw Exception(fmt::format("Simulating {} storage failure", failure_type));
+    })};
 }
+
+static FailureAction slow_action(double probability, int slow_down_ms_min, int slow_down_ms_max) {
+    return {fmt::format("slow_down({})", probability), maybe_execute(probability, [slow_down_ms_min, slow_down_ms_max](FailureType) {
+        thread_local std::uniform_int_distribution<size_t> dist(slow_down_ms_min, slow_down_ms_max);
+        thread_local std::mt19937 gen(std::random_device{}());
+        int sleep_ms = dist(gen);
+        ARCTICDB_INFO(log::lock(), "Testing: Sleeping for {} ms", sleep_ms);
+        std::this_thread::sleep_for(std::chrono::milliseconds(sleep_ms));
+    })};
+}
+
 }
 
 /** Independent state for each FailureType. Thread-safe except for the c'tors. */
@@ -149,10 +171,16 @@ public:
 
     StorageFailureSimulator() : configured_(false) {}
 
-    void configure(const arcticdb::proto::storage::VersionStoreConfig::StorageFailureSimulator& cfg)  {
-        configure({{FailureType::WRITE, {action_factories::fault(cfg.write_failure_prob())}},
-                   {FailureType::READ, {action_factories::fault(cfg.read_failure_prob())}}});
-    }
+    void configure(const arcticdb::proto::storage::VersionStoreConfig::StorageFailureSimulator& cfg) {
+        using enum arcticdb::FailureType;
+        Params default_actions = {
+            {WRITE, {action_factories::fault(cfg.write_failure_prob())}},
+            {READ, {action_factories::fault(cfg.read_failure_prob())}},
+            {WRITE_SLOW, {action_factories::slow_action(cfg.write_slowdown_prob(), cfg.slow_down_min_ms(), cfg.slow_down_max_ms())}}
+        };
+
+        configure(default_actions);
+    };
 
     void configure(const Params& params) {
         log::storage().info("Initializing storage failure simulator");
--- a/cpp/arcticdb/util/storage_lock.hpp
+++ b/cpp/arcticdb/util/storage_lock.hpp
@@ -73,17 +73,17 @@ inline std::thread::id get_thread_id() noexcept {
 
 // This StorageLock is inherently unreliable. It does not use atomic operations and it is possible for two processes to acquire if the timing is right.
 // If you want a reliable alternative which is slower but uses atomic primitives you can look at the `ReliableStorageLock`.
-template <class ClockType = util::SysClock, class MutexType = std::mutex>
+template <class ClockType = util::SysClock>
 class StorageLock {
-    static constexpr int64_t DEFAULT_TTL_INTERVAL = ONE_MINUTE * 60 * 24; // 1 Day
-    static constexpr int64_t DEFAULT_WAIT_MS = 1000; // 1 Second
-    static constexpr int64_t DEFAULT_INITIAL_WAIT_MS = 10;
-
-    MutexType mutex_;
+    std::mutex mutex_;
     const StreamId name_;
     timestamp ts_ = 0;
 
   public:
+    static constexpr int64_t DEFAULT_TTL_INTERVAL = ONE_MINUTE * 60 * 24; // 1 Day
+    static constexpr int64_t DEFAULT_WAIT_MS = 1000; // 1 Second
+    static constexpr int64_t DEFAULT_INITIAL_WAIT_MS = 10;
+
     static void force_release_lock(const StreamId& name, const std::shared_ptr<Store>& store) {
         do_remove_ref_key(store, name);
     }
@@ -112,7 +112,7 @@ class StorageLock {
     }
 
     bool try_lock(const std::shared_ptr<Store>& store) {
-       ARCTICDB_DEBUG(log::lock(), "Storage lock: try lock {}", get_thread_id());
+        ARCTICDB_DEBUG(log::lock(), "Storage lock: try lock {}", get_thread_id());
         if(!mutex_.try_lock()) {
             ARCTICDB_DEBUG(log::lock(), "Storage lock: failed local lock {}", get_thread_id());
             return false;
@@ -122,33 +122,12 @@ class StorageLock {
             that->mutex_.unlock();
         }};
 
-        auto start = ClockType::coarse_nanos_since_epoch();
-        if(!ref_key_exists(store) || !ttl_not_expired(store)) {
-            ts_= create_ref_key(store);
-            auto lock_sleep_ms = ConfigsMap::instance()->get_int("StorageLock.WaitMs", DEFAULT_WAIT_MS);
-            std::this_thread::sleep_for(std::chrono::milliseconds(lock_sleep_ms));
-            auto read_ts = read_timestamp(store);
-            auto duration = ClockType::coarse_nanos_since_epoch() - start;
-            auto duration_in_ms = duration / ONE_MILLISECOND;
-            ARCTICDB_INFO(log::lock(), "Took {} ms", duration_in_ms);
-            ARCTICDB_INFO(log::lock(), "Max is {} ms", 1.5 * lock_sleep_ms);
-            if (duration > 1.5 * lock_sleep_ms * ONE_MILLISECOND) {
-                ARCTICDB_DEBUG(log::lock(), "Took too long to read and write the lock. Aborting.");
-                return false;
-            }
-            if(read_ts && *read_ts == ts_) {
-                x.release();
-                ARCTICDB_DEBUG(log::lock(), "Storage lock: succeeded {}", get_thread_id());
-                return true;
-            } else {
-                ARCTICDB_DEBUG(log::lock(), "Storage lock: pre-empted {}", get_thread_id());
-                ts_ = 0;
-                return false;
-            }
-        } else {
-            ARCTICDB_DEBUG(log::lock(), "Storage lock: failed {}", get_thread_id());
-            return false;
+        const bool try_lock = try_acquire_lock(store);
+        if (try_lock) {
+            x.release();
         }
+
+        return try_lock;
     }
 
     void _test_release_local_lock() {
@@ -162,32 +141,50 @@ class StorageLock {
         thread_local std::uniform_int_distribution<size_t> dist;
         thread_local std::minstd_rand gen(std::random_device{}());
         size_t total_wait = 0;
-        do_wait:
-        while (ref_key_exists(store)) {
+
+        while (!try_acquire_lock(store)) {
             wait_ms += dist(gen, decltype(dist)::param_type{0, wait_ms / 2});
             log::lock().info("Didn't get lock, waiting {}", wait_ms);
             sleep_ms(wait_ms);
             total_wait += wait_ms;
             wait_ms *= 2;
-            auto read_ts = ttl_not_expired(store);
-            if (!read_ts)
-                break;
             if (timeout_ms && total_wait > *timeout_ms) {
                 ts_ = 0;
                 log::lock().info("Lock timed out, giving up after {}", wait_ms);
                 mutex_.unlock();
-                throw StorageLockTimeout{fmt::format("Storage lock {} timeout out after {} ms. Lock held since {} (UTC)", name_, total_wait, date_and_time(*read_ts))};
+                throw StorageLockTimeout{fmt::format("Storage lock {} timeout out after {} ms. Lock held since {} (UTC)", name_, total_wait, date_and_time(ts_/**read_ts*/))};
             }
         }
-        ts_ = create_ref_key(store);
-        log::lock().info("{} Lock unlocked, trying to set lock", get_thread_id());
-        auto lock_sleep = ConfigsMap::instance()->get_int("StorageLock.WaitMs", DEFAULT_WAIT_MS);
-        std::this_thread::sleep_for(std::chrono::milliseconds(lock_sleep));
-        auto read_ts = read_timestamp(store);
-        if(!read_ts || *read_ts != ts_) {
-            log::lock().info("Lock preempted, expected timestamp {} but got {}", ts_, read_ts.value_or(0));
-            ts_ = 0;
-            goto do_wait;
+    }
+
+    bool try_acquire_lock(const std::shared_ptr<Store>& store) {
+        auto start = ClockType::coarse_nanos_since_epoch();
+        if(!ref_key_exists(store) || !ttl_not_expired(store)) {
+            ts_= create_ref_key(store);
+            auto lock_sleep_ms = ConfigsMap::instance()->get_int("StorageLock.WaitMs", DEFAULT_WAIT_MS);
+            ARCTICDB_DEBUG(log::lock(), "Waiting for {} ms, thread id: {}", lock_sleep_ms, std::this_thread::get_id());
+            std::this_thread::sleep_for(std::chrono::milliseconds(lock_sleep_ms));
+            ARCTICDB_DEBUG(log::lock(), "Waited for {} ms, thread id: {}", lock_sleep_ms, std::this_thread::get_id());
+            auto read_ts = read_timestamp(store);
+            auto duration = ClockType::coarse_nanos_since_epoch() - start;
+            auto duration_in_ms = duration / ONE_MILLISECOND;
+            ARCTICDB_INFO(log::lock(), "Took {} ms", duration_in_ms);
+            ARCTICDB_INFO(log::lock(), "Max is {} ms", 1.5 * lock_sleep_ms);
+            if (duration > 1.5 * lock_sleep_ms * ONE_MILLISECOND) {
+                ARCTICDB_DEBUG(log::lock(), "Took too long to read and write the lock. Aborting.");
+                return false;
+            }
+            if(read_ts && *read_ts == ts_) {
+                ARCTICDB_DEBUG(log::lock(), "Storage lock: succeeded {}, written_timestamp: {} current_timestamp: {}", get_thread_id(), ts_, read_ts);
+                return true;
+            } else {
+                ARCTICDB_DEBUG(log::lock(), "Storage lock: pre-empted {}, written_timestamp: {} current_timestamp: {}", get_thread_id(), ts_, read_ts);
+                ts_ = 0;
+                return false;
+            }
+        } else {
+            ARCTICDB_DEBUG(log::lock(), "Storage lock: failed {}", get_thread_id());
+            return false;
         }
     }
 
--- a/cpp/arcticdb/util/storage_lock.hpp
+++ b/cpp/arcticdb/util/storage_lock.hpp
@@ -12,6 +12,7 @@
 #include <arcticdb/storage/store.hpp>
 #include <arcticdb/stream/index.hpp>
 #include <arcticdb/util/exponential_backoff.hpp>
+#include <arcticdb/storage/failure_simulation.hpp>
 #include <arcticdb/util/configs_map.hpp>
 
 #include <fmt/std.h>
@@ -72,11 +73,13 @@ inline std::thread::id get_thread_id() noexcept {
 
 // This StorageLock is inherently unreliable. It does not use atomic operations and it is possible for two processes to acquire if the timing is right.
 // If you want a reliable alternative which is slower but uses atomic primitives you can look at the `ReliableStorageLock`.
-template <class ClockType = util::SysClock>
+template <class ClockType = util::SysClock, class MutexType = std::mutex>
 class StorageLock {
-    // 1 Day
-    static constexpr int64_t DEFAULT_TTL_INTERVAL = ONE_MINUTE * 60 * 24;
-    std::mutex mutex_;
+    static constexpr int64_t DEFAULT_TTL_INTERVAL = ONE_MINUTE * 60 * 24; // 1 Day
+    static constexpr int64_t DEFAULT_WAIT_MS = 1000; // 1 Second
+    static constexpr int64_t DEFAULT_INITIAL_WAIT_MS = 10;
+
+    MutexType mutex_;
     const StreamId name_;
     timestamp ts_ = 0;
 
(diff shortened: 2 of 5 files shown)


cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -77,29 +78,50 @@ inline std::ostream& operator<<(std::ostream& out, const FailureAction& action)
...
+            return;
+        }
+
+        if (probability == 1.0) {
+            std::invoke(std::move(f), type);
+            return;
+        }
+
+        thread_local std::once_flag flag;
+        std::call_once(flag, [seed = (uint64_t)(&type)]() { init_random(seed); });
```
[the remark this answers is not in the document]
ognyanstoimenov: Yeah, I am a bit confused by the logic here, we can discuss this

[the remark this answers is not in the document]
IvoDD: As Ognyan pointed out the `thread_local` assures that each thread will have exactly one instance of this

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
```diff
@@ -0,0 +1,81 @@
...
+def write_symbols_worker(real_s3_storage_factory, lib_name, result_queue, run_time, step_symbol_id, first_symbol_id):
+    fixture = real_s3_storage_factory.create_fixture()
+    lib = fixture.create_arctic()[lib_name]
+    df = pd.DataFrame({"col": [1, 2, 3]})
+    cnt = 0
+    start_time = time.time()
+    while time.time() - start_time < run_time:
+        id = cnt * step_symbol_id + first_symbol_id
+        lib.write(f"sym_{id}", df)
+        cnt += 1
```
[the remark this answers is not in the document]
ognyanstoimenov: Will just remove the loop and use FailureSimulator with some probability, so each process writes/compacts only once as it is more deterministic

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -145,36 +137,54 @@ class StorageLock {
...
         mutex_.lock();
-        size_t wait_ms = ConfigsMap::instance()->get_int("StorageLock.InitialWaitMs", 10);
+        size_t wait_ms = ConfigsMap::instance()->get_int("StorageLock.InitialWaitMs", DEFAULT_INITIAL_WAIT_MS);
         thread_local std::uniform_int_distribution<size_t> dist;
         thread_local std::minstd_rand gen(std::random_device{}());
         size_t total_wait = 0;
-        do_wait:
-        while (ref_key_exists(store)) {
+
+        while (!try_acquire_lock(store)) {
```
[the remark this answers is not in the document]
ognyanstoimenov: There already is logging here `log::lock().info("Didn't get lock, waiting {}", wait_ms);`

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -141,26 +162,41 @@ class StorageFailureSimulator {
...
+        using enum arcticdb::FailureType;
+        log::storage().info("Initializing storage failure simulator from proto config");
+
+        if (cfg.read_failure_prob() > 0) {
+            categories_.try_emplace(READ, ParamActionSequence{action_factories::fault(cfg.read_failure_prob())});
+        }
+        if (cfg.write_failure_prob() > 0) {
+            categories_.try_emplace(WRITE, ParamActionSequence{action_factories::fault(cfg.write_failure_prob())});
+        }
+        else if (cfg.write_slowdown_prob() > 0) {
```
IvoDD, line 191: Not for this PR but:

I can see how it can be useful to have both probability to fail and probability to slowdown. 
E.g. to test that some unnatural timing + failures are handled gracefully.

This would require changing the `categories_` to something like `unordered_map<FailureType, std::vector<FailureTypeState>>`

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -233,7 +238,7 @@ class StorageLock {
             // check TTL
             auto ttl = ConfigsMap::instance()->get_int("StorageLock.TTL", DEFAULT_TTL_INTERVAL);
             if (ClockType::coarse_nanos_since_epoch() - *read_ts > ttl) {
-                log::lock().warn("StorageLock {} taken for more than TTL (default 1 day). Force releasing", name_);
+                log::lock().warn("StorageLock {} taken since {}, which is more than TTL (default 1 day). Force releasing.", name_, *read_ts);
                 force_release_lock(name_, store);
```
IvoDD: Sorry for the context creep, but I feel like `force_release`-ing the lock inside `ttl_not_expired` is problematic.

On one hand is quite surprising for fuction called `ttl_not_expired` to actually delete lock keys.
On the other hand this makes reasoning about the lock way more complicated. E.g. it is possible for racing processes to:
```
Process A and process B are running `try_acquire_lock`
0. There exists an expired lock in storage
1. Process A reads timestamp of the expired lock
2. Process B reads timestamp of the expired lock
3. Process A deletes expired lock
4. Process A writes new lock
5. Process B now happens to issue the delete of the lock
```
This will not cause a lock failure due to the slow write/read/delete check we have added but makes proving that this works harder.

What do you think about:
- Removing the `force_release_lock` call here. When calling `create_ref_key` it would either way get overridden. We'll have better atomicity if we just to one overwrite rather than a delete + write.
- Changing the log to `info`. It is expected to occasionally have a process die while holding a lock and I don't think we should warn users about it.

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -183,7 +193,8 @@ class StorageLock {
     }
 
     timestamp create_ref_key(const std::shared_ptr<Store>& store) {
-        auto ts =  ClockType::nanos_since_epoch();
+        auto ts = ClockType::nanos_since_epoch();
+        StorageFailureSimulator::instance()->go(FailureType::WRITE);
```
IvoDD: Not for this PR:

I think we should have the storage failure simulator run on a lower level. Currently all code which runs `write_sync` needs to think about whether it needs to simulate a failure.

This should happen on the `Store` abstraction level (i.e. call it inside `write_sync`) in such a way that it doesn't slowdown regular operation but allows enabling it via environment variable or something similar.

cpp/arcticdb/util/storage_lock.hpp
IvoDD: Also the returned timestamp is not used anywhere. Maybe we can just return a `bool` and rename this to `exists_active_lock`?

cpp/arcticdb/util/storage_lock.hpp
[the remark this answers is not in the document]
IvoDD: Sorry not sure how I missed it. Makes sense

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -145,36 +137,54 @@ class StorageLock {
...
             wait_ms *= 2;
-            auto read_ts = ttl_not_expired(store);
-            if (!read_ts)
-                break;
             if (timeout_ms && total_wait > *timeout_ms) {
                 ts_ = 0;
                 log::lock().info("Lock timed out, giving up after {}", wait_ms);
                 mutex_.unlock();
-                throw StorageLockTimeout{fmt::format("Storage lock {} timeout out after {} ms. Lock held since {} (UTC)", name_, total_wait, date_and_time(*read_ts))};
+                throw StorageLockTimeout{fmt::format("Storage lock {} timeout out after {} ms. Lock held since {} (UTC)", name_, total_wait, date_and_time(ts_/**read_ts*/))};
```
IvoDD: The lock held since part of the error is wrong. We should just remove that part of the error message as there are now other ways we can fail to take a lock other then a lock already existing.

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -145,36 +137,54 @@ class StorageLock {
...
+            auto lock_sleep_ms = ConfigsMap::instance()->get_int("StorageLock.WaitMs", DEFAULT_WAIT_MS);
+            ARCTICDB_DEBUG(log::lock(), "Waiting for {} ms..", lock_sleep_ms);
+            std::this_thread::sleep_for(std::chrono::milliseconds(lock_sleep_ms));
+            ARCTICDB_DEBUG(log::lock(), "Waited for {} ms", lock_sleep_ms);
+            auto read_ts = read_timestamp(store);
+            auto duration = ClockType::coarse_nanos_since_epoch() - start;
+            [[maybe_unused]] auto duration_in_ms = duration / ONE_MILLISECOND;
+            ARCTICDB_DEBUG(log::lock(), "Took {} ms", duration_in_ms);
+            ARCTICDB_DEBUG(log::lock(), "Max is {} ms", 1.5 * lock_sleep_ms);
+            if (duration > 1.5 * lock_sleep_ms * ONE_MILLISECOND) {
```
IvoDD: Nit: Extract the `1.5 * lock_sleep_ms` in a variable `max_allowed_duration`, so if we ever change the `1.5` we'll only need to change it in one place.

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -85,6 +79,7 @@ struct OptimisticLockTask {
             else {
                 // As of C++20, '++' expression of 'volatile'-qualified type is deprecated.
```
IvoDD: This comment doesn't make much sense with the introduced sleep.

What do you think about instead adding a comment briefly explaining the difference between the volatile and atomic counters. Something like:
```
// Volatile counter would diverge from atomic counter in case the lock fails and allows two threads to think they hold the lock simultaneously.
```

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -104,19 +99,19 @@ TEST(StorageLock, Contention) {
...
     ASSERT_EQ(lock_data->contended_, true);
 }
 
-struct PessimisticLockTask {
+struct LockTaskWithRetry {
     std::shared_ptr<LockData> data_;
     std::optional<size_t> timeout_ms_;
 
-    PessimisticLockTask(std::shared_ptr<LockData> data, std::optional<size_t> timeout_ms = std::nullopt) :
+    LockTaskWithRetry(std::shared_ptr<LockData> data, std::optional<size_t> timeout_ms = std::nullopt) :
```
IvoDD: I think `LockTaskWithRetry` would also benefit from the `sleep` in the volatile increment in the `operator ()`. What do you think about extracting that logic in a common function e.g. `increment_counters_under_lock`

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -363,4 +358,130 @@ TEST(StorageLock, OptimisticForceReleaseLock) {
...
+    std::vector<Future<Unit>> futures;
+    auto lock_data = std::make_shared<LockData>(num_writers);
+    lock_data->store_ = std::make_shared<InMemoryStore>();
+
+    for (size_t i = 0; i < num_writers; ++i) {
+        futures.emplace_back(exec.addFuture(LockTaskWithoutRetry(lock_data)));
+    }
+    collect(futures).get();
+
+    ASSERT_EQ(lock_data->atomic_, expected_locks);
```
IvoDD, line 400: Also assert that `lock_data->atomic_ == lock_data->vol_`?

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -363,4 +358,130 @@ TEST(StorageLock, OptimisticForceReleaseLock) {
...
+    lock_data->store_ = std::make_shared<InMemoryStore>();
+    FutureExecutor<CPUThreadPoolExecutor> exec{num_writers};
+    StorageFailureSimulator::instance()->configure({{FailureType::WRITE, SLOW_ACTIONS}});
+    std::vector<Future<Unit>> futures;
+    for(size_t i = 0; i < num_writers; ++i) {
+        futures.emplace_back(exec.addFuture(LockTaskWithRetry(lock_data, 5000)));
+    }
+    collect(futures).get();
+
+    ASSERT_EQ(lock_data->atomic_, 1); // Retruying should eventually make one compaction succeed
```
IvoDD: Why just one? I feel like this is tied to the `5000ms` timeout because if it was bigger eventually both tasks would succeed, hopefully one after the other.
I kind of think that would be a more meaningful test showing that with retries both will succeed and there would have been no races.

cpp/arcticdb/util/test/test_storage_lock.cpp
IvoDD: Similar for all other such tests

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -363,4 +358,130 @@ TEST(StorageLock, OptimisticForceReleaseLock) {
...
+    const auto max_ms = current_lock_sleep_wait_ms * 2;
+    const StorageFailureSimulator::ParamActionSequence SLOW_WRITE = {
+        action_factories::slow_action(1, min_ms, max_ms)
+    };
+    StorageFailureSimulator::instance()->configure({{FailureType::WRITE, SLOW_WRITE}});
+    auto lock = StorageLock("test");
+    ASSERT_FALSE(lock.try_lock(std::make_shared<InMemoryStore>()));
+}
+
+TEST(StorageLock, DISABLED_LockSameTimestamp) { // Not yet implemented
```
IvoDD, line 468: I don't have a strong opinion but I think it would be useful to enable the test to show that this indeed fails.

I.e. to assert `vol_==1` and `atomic_==2` and write a todo comment there. This would show the power of our testing method to identify racing locks.

If that doesn't work it might show us that we need to increase the sleep in the volatile counter write.

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -363,4 +358,130 @@ TEST(StorageLock, OptimisticForceReleaseLock) {
...
+    FutureExecutor<CPUThreadPoolExecutor> exec{num_writers};
+
+    StorageFailureSimulator::instance()->configure({{FailureType::WRITE, SLOW_ACTIONS}});
+
+    std::vector<Future<Unit>> futures;
+    auto lock_data = std::make_shared<LockData>(num_writers);
+    lock_data->store_ = std::make_shared<InMemoryStore>();
+
+    for (size_t i = 0; i < num_writers; ++i) {
+        futures.emplace_back(exec.addFuture(LockTaskWithoutRetry(lock_data)));
```
IvoDD: What do you think about allowing the retry? It would make the test slower but it would allow more varied lock request timings instead of all locks trying to get added in the beginning. 

E.g. I think right now only one lock will succeed. It would be good if at least 10 locks succeed?

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
```diff
@@ -0,0 +1,81 @@
...
+    # Trigger symbol list compaction on every list_symbols call
+    set_config_int("SymbolList.MaxDelta", 1)
+    fixture = real_s3_storage_factory.create_fixture()
+    lib = fixture.create_arctic()[lib_name]
+
+    start_time = time.time()
+    while time.time() - start_time < run_time:
+        lib.list_symbols()
+
+#@REAL_S3_TESTS_MARK
```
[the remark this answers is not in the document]
IvoDD: Looks like you missed this?

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
```diff
@@ -0,0 +1,87 @@
...
+from arcticdb_ext import set_config_int
+
+from arcticdb.options import LibraryOptions, EnterpriseLibraryOptions
+from arcticdb.config import set_log_level
+from arcticdb_ext.storage import KeyType
+
+
+def write_symbols_worker(lib, result_queue):
+    df = pd.DataFrame({"col": [1, 2, 3]})
+    id = os.getpid()
```
IvoDD: Nit: I think theoretically it is possible for pids to be the same. E.g. one process starts and finishes before another one starts and gets assigned the same pid.

We could just pass an argument `id` to the worker and use that for the symbol. 
Then we can also drop the need for the `result_queue` (given that below we raise if a subprocess fails)

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
```diff
@@ -0,0 +1,87 @@
...
+
+#@REAL_S3_TESTS_MARK
+@pytest.mark.parametrize("num_writers, num_compactors", [(1,1), (2, 10), (10, 50)])
+def test_stress_compaction_many_writers(slow_writing_library, num_writers, num_compactors):
+    set_config_int("SymbolList.MaxDelta", 1) # Trigger symbol list compaction on every list_symbols call
+    set_log_level(specific_log_levels={"lock": "DEBUG"})
+    results_queue = Queue()
+
+    writers = [
+        Process(target=write_symbols_worker, args=(slow_writing_library, results_queue))
```
IvoDD: I think this will fork the library object including the s3 connection. We've seen this being problematic before. 
I think it's best if we pass only the s3 uri and construct the library in the child process.

cpp/arcticdb/util/test/test_storage_lock.cpp
ognyanstoimenov: Talked offline to modify this test to run without slowdowns and the second lock timing out

cpp/arcticdb/util/test/test_storage_lock.cpp
ognyanstoimenov: Will test with retries as well

cpp/arcticdb/util/storage_lock.hpp
ognyanstoimenov: 100% agree. Will open a task for refactoring the StorageFailureSimulator to benefit from the abstractions in the code

cpp/arcticdb/storage/failure_simulation.hpp
ognyanstoimenov: Will add this to the StorageFailureSimulator refactor task

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
ognyanstoimenov: Will leave it as-is as it is cleaner. If flakyness occurs on the CI I will implement your suggestion

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
ognyanstoimenov: Will leave it as-is as it is cleaner. If flakyness occurs on the CI I will implement your suggestion

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -77,29 +77,51 @@ inline std::ostream& operator<<(std::ostream& out, const FailureAction& action)
...
+            return;
+        }
+
+        if (probability == 1.0) {
+            std::invoke(std::move(f), type);
+            return;
+        }
+
+        thread_local std::uniform_int_distribution<size_t> dist(0.0, 1.0);
+        thread_local std::mt19937 gen(std::random_device{}());
```
ognyanstoimenov, line 101: Changed to this random generator as old one gave me only probabilities up to 0.5 so might be something wrong with it. We can address this as part of the future failure simulation refactor task.

cpp/arcticdb/util/configs_map.hpp
```diff
@@ -21,11 +21,11 @@ using namespace arcticdb::proto::config;
...
-
     static void init();
-    static std::shared_ptr<ConfigsMap> instance();
+    static std::shared_ptr<ConfigsMap> instance() { 
+        static auto instance_ = std::make_shared<ConfigsMap>();
+        return instance_;
+    }
 
     ConfigsMap() :
         mutex_(std::make_unique<std::mutex>()) {
```
ognyanstoimenov: I don't understand what good is this mutex here (it is not used anywhere AFAIC) and why is it wrapped in a pointer. Anyone more familiar with this please give some insight. Otherwise, I think I will just add a normal mutex (not ptr) and add it to all the operations here

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -77,29 +77,51 @@ inline std::ostream& operator<<(std::ostream& out, const FailureAction& action)
...
+        double rnd = dist(gen);
+        if (rnd < probability) {
+            std::invoke(std::move(f), type);
```
poodlewars: Don't see why we're using `std::invoke` instead of just `f(type)`

cpp/arcticdb/storage/failure_simulation.hpp
```diff
@@ -77,29 +77,51 @@ inline std::ostream& operator<<(std::ostream& out, const FailureAction& action)
...
+                throw Exception(fmt::format("Simulating {} storage failure", failure_type));
+    })};
 }
+
+static FailureAction slow_action(double probability, int slow_down_ms_min, int slow_down_ms_max) {
+    return {fmt::format("slow_down({})", probability), maybe_execute(probability, [slow_down_ms_min, slow_down_ms_max](FailureType) {
+        thread_local std::uniform_int_distribution<size_t> dist(slow_down_ms_min, slow_down_ms_max);
+        thread_local std::mt19937 gen(std::random_device{}());
+        int sleep_ms = dist(gen);
+        ARCTICDB_INFO(log::lock(), "Testing: Sleeping for {} ms", sleep_ms);
```
poodlewars: The `lock` logger is a weird choice?

cpp/arcticdb/storage/failure_simulation.hpp
poodlewars: I think we can just model the slowdowns as failures can't we? More like my PR. https://github.com/man-group/ArcticDB/pull/2407

cpp/arcticdb/util/configs_map.hpp
poodlewars: It's probably to make the `ConfigsMap` movable

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -145,36 +134,55 @@ class StorageLock {
...
+            auto max_allowed_duration_ms = MAX_DURATION_FACTOR * lock_sleep_ms;
+            ARCTICDB_DEBUG(log::lock(), "Waiting for {} ms..", lock_sleep_ms);
+            std::this_thread::sleep_for(std::chrono::milliseconds(lock_sleep_ms));
+            ARCTICDB_DEBUG(log::lock(), "Waited for {} ms", lock_sleep_ms);
+            auto read_ts = read_timestamp(store);
+            auto duration = ClockType::coarse_nanos_since_epoch() - start;
+            [[maybe_unused]] auto duration_in_ms = duration / ONE_MILLISECOND;
+            ARCTICDB_DEBUG(log::lock(), "Took {} ms", duration_in_ms);
+            ARCTICDB_DEBUG(log::lock(), "Max is {} ms", max_allowed_duration_ms);
+            if (duration > max_allowed_duration_ms * ONE_MILLISECOND) {
```
poodlewars: Just compare with `duration_ms`

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -66,21 +67,20 @@ struct StorageLockTimeout : public std::runtime_error {
...
-    static constexpr int64_t DEFAULT_TTL_INTERVAL = ONE_MINUTE * 60 * 24;
     std::mutex mutex_;
     const StreamId name_;
     timestamp ts_ = 0;
 
   public:
+    static constexpr int64_t DEFAULT_TTL_INTERVAL = ONE_MINUTE * 60 * 24; // 1 Day
+    static constexpr int64_t DEFAULT_WAIT_MS = 1000; // 1 Second
+    static constexpr int64_t DEFAULT_INITIAL_WAIT_MS = 10;
+    static constexpr double MAX_DURATION_FACTOR = 1.5;
```
poodlewars: This factor should be configurable, and have a comment explaining what it controls

How have you chosen this factor to be `1.5`? Should be possible to write a quick model to figure out a sensible factor - IIRC the idea here is that if we take a long time to take the lock than the chance of another process pre-empting us and taking the lock too is higher. I think it would be worth having some rigour around how we choose this factor.

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -145,36 +134,55 @@ class StorageLock {
...
+            ARCTICDB_DEBUG(log::lock(), "Waiting for {} ms..", lock_sleep_ms);
+            std::this_thread::sleep_for(std::chrono::milliseconds(lock_sleep_ms));
+            ARCTICDB_DEBUG(log::lock(), "Waited for {} ms", lock_sleep_ms);
+            auto read_ts = read_timestamp(store);
+            auto duration = ClockType::coarse_nanos_since_epoch() - start;
+            [[maybe_unused]] auto duration_in_ms = duration / ONE_MILLISECOND;
+            ARCTICDB_DEBUG(log::lock(), "Took {} ms", duration_in_ms);
+            ARCTICDB_DEBUG(log::lock(), "Max is {} ms", max_allowed_duration_ms);
+            if (duration > max_allowed_duration_ms * ONE_MILLISECOND) {
+                ARCTICDB_DEBUG(log::lock(), "Took too long to read and write the lock. Aborting.");
```
poodlewars, line 185: Comment to explain why we have this logic

cpp/arcticdb/util/storage_lock.hpp
poodlewars: The simulator is at the level of the `InMemoryStore` which already has failure handling for `write_sync`? We should never see it being called outside of `in_memory_store.hpp`. In particular `in_memory_store.hpp` is only linked for test builds

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -227,18 +230,16 @@ class StorageLock {
         }
     }
 
-    std::optional<timestamp> ttl_not_expired(const std::shared_ptr<Store>& store) {
-        auto read_ts = read_timestamp(store);
-        if (read_ts) {
+    bool exists_active_lock(const std::shared_ptr<Store>& store) const {
```
poodlewars, line 246: `active_lock_exists` is a more grammatical name

cpp/arcticdb/util/storage_lock.hpp
ognyanstoimenov: But it is good to also be able to attach simulator to the async store. Like it is the case here I attach it to the storage_lock regardless of the store it uses. That is why I believe the refactor should be leave this StorageLockSimulator class as-is and the calls to StorageFailureSimulator::Go to be made on a store level which can be configured as a e.g. StoreWithFailures in the tests (python and c++)

cpp/arcticdb/storage/failure_simulation.hpp
ognyanstoimenov: Ther _are_ modeled as failures just like your PR. The proto config part is so that it is available to the python tests as well. But see other comment for a proposed (future) refactor

cpp/arcticdb/util/configs_map.hpp
ognyanstoimenov: Unwrapping the mutex as this is a singleton so it shouldn't need to be movable

cpp/arcticdb/util/storage_lock.hpp
ognyanstoimenov: There is a convention to prefix boolean variables and functions with a verb so they are more easily spotted as booleans. But if we don't follow that and prefer the more grammatical naming I am happy to change it, don't have a strong preference.

cpp/arcticdb/storage/failure_simulation.hpp
ognyanstoimenov: Good spot

cpp/arcticdb/storage/failure_simulation.hpp
ognyanstoimenov: To look prettier with std::move(f), but it isn't a big deal to omit it so I will change it to f(type)

cpp/arcticdb/util/storage_lock.hpp
ognyanstoimenov: I've made it configurable.

As for the exact value, I've taken it from the design docs here. https://manwiki.maninvestments.com/display/AlphaTech/Symbol+list+race+conditions. from Ivo and William. It seems a reasonable factor as that allows up WaitMs/2 slowdown for both reads+write (excluding the wait). Lower factor means we will have more false positives that would fail. Higher than that means that we are at higher risk of not catching a race that would result in two writers thinking they have the lock.

It shows big improvements in the stress tests I've written (given the slowdown values I am testing against of course, although they seem reasonable).
I haven't thought much about the exact factor choice we can change it if something else seems more reasonable.

cpp/arcticdb/util/storage_lock.hpp
poodlewars: Yeah I see the point and agree that pushing this up to the `Store` in a future refactor would be fine. But I think how things are at the moment is not right because the failure simulator will be called twice for the same write operation.

cpp/arcticdb/util/storage_lock.hpp
poodlewars: Great - given you already have the stress tests it shouldn't be a big deal to run some experiments with different factors and be a bit more scientific about things?

ognyanstoimenov: Sure, will do

poodlewars (approved):
Similar to Ivo I'm approving before I go away so I don't block merging - we've spoken about the outstanding changes here.

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -78,7 +78,7 @@ class StorageLock {
   public:
     static constexpr int64_t DEFAULT_TTL_INTERVAL = ONE_MINUTE * 60 * 24; // 1 Day
     static constexpr int64_t DEFAULT_WAIT_MS = 1000; // After writing the lock, waiting this time before checking if the written lock is still ours.
-    static constexpr double DEFAULT_MAX_DURATION_FACTOR = 1; // Controls what factor of the wait time is the maximum time the lock shoud spend. (writing + waiting + reading).
+    static constexpr double DEFAULT_MAX_DURATION_FACTOR = 1.5; // Controls what factor of the wait time is the maximum time the lock shoud spend. (writing + waiting + reading).
```
IvoDD: IIRC we discussed to add a good comment explaining why we choose 1.5

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
```diff
@@ -71,15 +71,36 @@ def test_stress_compaction_many_writers(slow_writing_library, num_writers, num_c
...
+    set_config_int("SymbolList.MaxDelta", compact_threshold)
+    for i in range(num_symbols):
+        sym = f"sym_{i}"
+        real_s3_library.write(sym, df)
+        symbols = real_s3_library.list_symbols()
+
+    expected_symbol_list = { f"sym_{i}" for i in range(num_symbols) }
+
+    result_symbol_list = set(symbols)
+    assert len(result_symbol_list) == len(expected_symbol_list)
```
IvoDD: Nit: I think this `len` assertion is obsolete

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
```diff
@@ -71,15 +71,36 @@ def test_stress_compaction_many_writers(slow_writing_library, num_writers, num_c
     assert len(result_symbol_list) == len(expected_symbol_list)
     assert result_symbol_list == expected_symbol_list
```
IvoDD, line 76: Maybe add a small comment explaining why we don't verify at least one compaction happened? Something like:
```
# Here we only assert that we are not missing symbols from the symbol list (which could happen if the storage lock fails).
# We don't check the number of compactions because they are non-deterministic. In fact with bad timings it is possible 0 compactions will happen if the slowdowns are big enough.
# So we have the `test_compaction_produces_single_key` to verify that compaction works as expected without slowdowns.
```

PR #2360, opened 2025-05-17
Implement modify schema functionality

vasil-pashov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


alexowens90 (changes requested):

cpp/arcticdb/version/version_core.cpp
```diff
@@ -901,6 +899,26 @@ std::vector<folly::Future<pipelines::SegmentAndSlice>> generate_segment_and_slic
...
+        FieldCollection fields_to_use;
+        auto overall_fields_it = pipeline_context.overall_column_bitset_->first();
+        const auto overall_fields_end = pipeline_context.overall_column_bitset_->end();
+        while (overall_fields_it != overall_fields_end) {
+            fields_to_use.add(desc.field(*overall_fields_it).ref());
+            ++overall_fields_it;
+        }
+        return OutputSchema{
+            StreamDescriptor{desc.data_ptr(), std::make_shared<FieldCollection>(std::move(fields_to_use))},
+            *pipeline_context.norm_meta_
```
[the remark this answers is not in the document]
vasil-pashov: This is not well defined for shared_ptr. Doing `std::move(*ptr)` will move the data from the stored object, but since it's a shared pointer this means that other owners of the pointer will be affected and will use garbage after this call. I don't think we should do this.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -948,24 +966,28 @@ folly::Future<std::vector<EntityId>> read_and_schedule_processing(
...
-
-                if (std::ranges::any_of(read_query->clauses_,
-                                        [](const std::shared_ptr<Clause>& clause) {
-                                            return clause->clause_info().modifies_output_descriptor_;
-                                        })) {
-                    set_output_descriptors(proc, read_query->clauses_, pipeline_context);
+                    std::shared_ptr<RowRange>,
+                    std::shared_ptr<ColRange> >(*component_manager, processed_entity_ids);
+                OutputSchema schema = create_initial_output_schema(*pipeline_context);
+                if (pipeline_context->rows_ > 0) {
```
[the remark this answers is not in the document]
vasil-pashov: Testing leftover. Not needed.

cpp/arcticdb/entity/stream_descriptor.hpp
```diff
@@ -308,13 +307,27 @@ struct OutputSchema {
...
+    }
+
+    void set_default_value_for_column(const std::string_view name, const VariantRawValue& value) {
+        default_values_.emplace(std::string(name), value);
+    }
+
 private:
     StreamDescriptor stream_descriptor_;
     std::optional<ankerl::unordered_dense::map<std::string, DataType>> column_types_;
+    ankerl::unordered_dense::map<std::string, VariantRawValue> default_values_;
```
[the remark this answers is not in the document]
vasil-pashov: I didn't know about the `Value` class

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -106,11 +107,6 @@ std::variant<BitSetTag, DataType> ExpressionNode::compute(
                 user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
                         std::holds_alternative<DataType>(left_type),
                         "Unexpected bitset input to {}", operation_type_);
-                user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
-                        is_floating_point_type(std::get<DataType>(left_type)) || is_sequence_type(std::get<DataType>(left_type)) ||
-                        is_time_type(std::get<DataType>(left_type)),
-                        "Unexpected data type {} input to {}",
-                        std::get<DataType>(left_type), operation_type_);
```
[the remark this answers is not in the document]
vasil-pashov: Modify schema on the filter clause throws an exception [here](https://github.com/man-group/ArcticDB/blob/7c326235643bbfccd227adf3ea114a5310812ff8/cpp/arcticdb/processing/clause.cpp#L159) it this is present. It's because we want to get an empty column when we filter int columns using is `OperationType::ISNULL` and `OperationType::NOTNULL`. I remember we discussed this and decided to remove the assertion from here.

cpp/arcticdb/processing/clause.cpp
```diff
@@ -156,7 +165,9 @@ std::vector<EntityId> FilterClause::process(std::vector<EntityId>&& entity_ids)
 }
 
 OutputSchema FilterClause::modify_schema(OutputSchema&& output_schema) const {
-    check_column_presence(output_schema, *clause_info_.input_columns_, "Filter");
+    if (first_missing_column(output_schema, *clause_info_.input_columns_) != clause_info_.input_columns_->end()) {
```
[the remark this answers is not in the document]
vasil-pashov: Looking at [test_filter_column_not_present_dynamic](https://github.com/man-group/ArcticDB/blob/7c326235643bbfccd227adf3ea114a5310812ff8/python/tests/unit/arcticdb/version_store/test_filtering.py#L1210) filtering is supposed to work with dynamic schema when the column is missing, `check_column_presence` would throw, however. Since the column is missing the schema is not changed in any way and we can return it.

According to [test_filter_column_not_present](https://github.com/man-group/ArcticDB/blob/7c326235643bbfccd227adf3ea114a5310812ff8/python/tests/unit/arcticdb/version_store/test_filtering.py#L48) static schema is supposed to throw. Indeed it throws [here](https://github.com/man-group/ArcticDB/blob/7c326235643bbfccd227adf3ea114a5310812ff8/cpp/arcticdb/processing/processing_unit.cpp#L63)

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1352,6 +1410,8 @@ struct CopyToBufferTask : async::BaseTask {
     DecodePathData shared_data_;
     std::any& handler_data_;
     OutputFormat output_format_;
+    IntToFloatConversion int_to_float_conversion_;
```
alexowens90: Unused

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1275,7 +1303,9 @@ void copy_frame_data_to_buffer(
         const RowRange& row_range,
         DecodePathData shared_data,
         std::any& handler_data,
-        OutputFormat output_format) {
+        OutputFormat output_format,
+        IntToFloatConversion int_to_float_conversion,
```
alexowens90: Shouldn't be an argument as it is always set to permissive

cpp/arcticdb/processing/clause.cpp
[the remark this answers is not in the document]
vasil-pashov: As discussed later, throwing is the correct behavior

cpp/arcticdb/processing/aggregation_interface.hpp
```diff
@@ -24,10 +25,13 @@ struct IGroupingAggregatorData {
         [[nodiscard]] SegmentInMemory finalize(const ColumnName& output_column_name, bool dynamic_schema, size_t unique_values) {
             return folly::poly_call<3>(*this, output_column_name, dynamic_schema, unique_values);
         }
+        [[nodiscard]] std::optional<Value> get_default_value() {
```
alexowens90, line 32: https://github.com/man-group/ArcticDB/pull/2360#discussion_r2107341713
The meaning of returning nullopt needs a comment here. Or we could make it explicit and always return `Value`, might be clearer

cpp/arcticdb/util/format_date.cpp
```diff
@@ -0,0 +1,40 @@
+#define BOOST_DATE_TIME_POSIX_TIME_STD_CONFIG // Allows using nanoseconds in boost.
+#include <arcticdb/util/format_date.hpp>
+#include <boost/date_time/posix_time/time_formatters.hpp>
+#include <boost/date_time/posix_time/posix_time_duration.hpp>
+#include <boost/date_time/posix_time/conversion.hpp>
+#include <arcticdb/util/constants.hpp>
+
+namespace arcticdb::util {
+
+    std::string format_timestamp(const entity::timestamp ts) {
```
alexowens90, line 10: nit: const not in header version

cpp/arcticdb/util/format_date.cpp
```diff
@@ -0,0 +1,40 @@
...
+        if (ts == NaT) {
+            return "NaT";
+        }
+        // Use boost as it can handle nanoseconds both on all OS's.
+        // std::std::chrono::time_point<std::chrono::system_clock> does not handle nanoseconds on Windows and Mac
+
+        const timestamp seconds = ts / 1'000'000'000;
+        const timestamp ns_remainder = ts % 1'000'000'000;
```
alexowens90: nit: `std::div`

cpp/arcticdb/pipeline/query.hpp
```diff
@@ -281,8 +281,6 @@ inline void build_col_read_query_filters(
         };
         queries.push_back(std::move(query));
     } else if (pipeline_context->overall_column_bitset_) {
-        util::check(!dynamic_schema || column_groups, "Did not expect a column bitset with dynamic schema");
```
alexowens90, line 284: I don't follow why this is no longer needed? If it really isn't, then the unused bool arg should be removed

cpp/arcticdb/pipeline/read_pipeline.hpp
```diff
@@ -157,9 +157,7 @@ inline std::vector<FilterQuery<ContainerType>> get_column_bitset_and_query_funct
     bool dynamic_schema,
     bool column_groups) {
     using namespace arcticdb::pipelines::index;
-    if(!dynamic_schema || column_groups) {
```
alexowens90: I don't get this change either

cpp/arcticdb/pipeline/value.hpp
```diff
@@ -61,10 +62,8 @@ struct Value {
         return *this;
     }
 
-    Value& operator==(const Value& other) {
-        data_type_ = other.data_type_;
-        assign(other);
-        return *this;
+    bool operator==(const Value& other) const {
+        return data_type_ == other.data_type_ && memcmp(other.data_, data_, 8) == 0;
```
alexowens90: This won't be correct for long strings, or possibly even numeric values that are <8 bytes. Presumably this is unused though, as the old implementation is so obviously wrong?

python/test_projection.py
alexowens90, line 1: Looks accidentally added?

python/tests/hypothesis/arcticdb/test_resample.py
```diff
@@ -1,33 +1,70 @@
...
+        if pd.api.types.is_integer_dtype(dtype):
+            # There are two reasons to cap int size:
+            # 1. To avoid overflows (happens usually when the dtype is 64bit and sum aggregator is used)
+            # 2. When dynamic schema is used. If we use pd.concat([df1, df2]) on two dataframes which are dynamic
+            #    schema segments, the columns which are missing in either df will become float64 in the result (so that
+            #    the missing values can be NaN), however, if int64 is used some values won't be represented correctly
+            #    and the test will fail.
+            current_byte_size = np.dtype(dtype).itemsize
+            max_bits = min(current_byte_size, 4) * 8 - 1 if pd.api.types.is_signed_integer_dtype(dtype) else 0
+            min_value = 0 if pd.api.types.is_unsigned_integer_dtype(np.dtype(dtype)) else -2**max_bits
```
alexowens90: This logic is really hard to follow, and sets `max_value` to zero for unsigned integers?

cpp/arcticdb/entity/type_utils.cpp
```diff
@@ -67,9 +67,32 @@ namespace arcticdb {
...
+    ) {
+        debug::check<ErrorCode::E_ASSERTION_FAILURE>(
+            is_integer_type(source.data_type()) && is_floating_point_type(target.data_type()),
+            "Expected source to be int and target to be float got: {} {}", source, target
+        );
+        switch (int_to_to_float_conversion) {
+            case IntToFloatConversion::STRICT: return target.get_size_bits() == entity::SizeBits::S64 || source.get_size_bits() < entity::SizeBits::S32;
+            case IntToFloatConversion::PERMISSIVE: return true;
+            default: {
+                internal::check<ErrorCode::E_ASSERTION_FAILURE>(
```
alexowens90: There is a `raise` method that doesn't need the first argument

cpp/arcticdb/processing/aggregation_utils.cpp
```diff
@@ -12,8 +12,7 @@ namespace arcticdb {
 
 void add_data_type_impl(entity::DataType data_type, std::optional<entity::DataType>& current_data_type) {
     if (current_data_type.has_value()) {
-        auto common_type = has_valid_common_type(entity::TypeDescriptor(*current_data_type, 0),
-                                                 entity::TypeDescriptor(data_type, 0));
+        const auto common_type = promotable_type(entity::TypeDescriptor(*current_data_type, 0), entity::TypeDescriptor(data_type, 0));
```
alexowens90: nit: Can use `make_scalar_type`

python/arcticdb/util/test.py
```diff
@@ -949,3 +985,123 @@ def equals(x, y):
...
+def common_float_int_type(float_dtype, int_dtype):
+    # We don't support float16
+    float_dtype = np.dtype(float_dtype)
+    int_dtype = np.dtype(int_dtype)
+    assert float_dtype.itemsize >= 4
+    if int_dtype.itemsize <= 2:
+        return float_dtype
+    return np.float64
+
+def valid_common_type(left, right):
```
alexowens90, line 1017: Could this be simplified using `np.promote_types`? Would also make it clear where our behaviour is the same and where it differs

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1116,6 +1115,16 @@ def test_float32_binary_comparison(lmdb_version_store_v1):
...
+@pytest.mark.xfail(reason="""Fails on Pandas < 2 because of this logic:
+    https://github.com/man-group/ArcticDB/blob/fc9514f25712d8e86fbdbd2f7e37e64f3a10df40/python/arcticdb/version_store/_normalization.py#L230
+    The assumptions there however are not correct. The dtype of empty columns is not deterministic and varies between
+    Pandas versions, for example the test passes on the CI with Python > 3.8 because it uses pandas 2.3.0 and the dtype
+    of the column is float64 (note the if in the link above sets the dtype to object only for pandas < 2 because it
+    expects that int Pandas >= 2 it'll always be object). It fails for object dtype because in arcticdb that becomes
+    string type and string types cannot be filtered using < or >, thus modify_schema fails
+    """,
+    strict=False
+)
```
alexowens90, line 1127: I understand how we got here, but this is going to look like a regression to users, we should discuss

python/tests/unit/arcticdb/version_store/test_resample.py
```diff
@@ -884,3 +890,309 @@ def test_date_range_outside_symbol_timerange(lmdb_version_store_v1):
     received_df = lib.read(sym, query_builder=q).data
     assert not len(received_df)
     assert received_df.columns == df.columns
+
+
+class TestResampleDynamicSchema:
+
+    @pytest.mark.parametrize("label", ["left", "right"])
+    @pytest.mark.parametrize("closed", ["left", "right"])
+    @pytest.mark.parametrize("dtype", [np.int32, np.float32, np.uint16])
```
alexowens90, line 899: `bool` also worth testing

python/tests/unit/arcticdb/version_store/test_resample.py
```diff
@@ -884,3 +890,309 @@ def test_date_range_outside_symbol_timerange(lmdb_version_store_v1):
...
+            rule,
+            agg,
+            pd.concat(df_list),
+            origin="epoch",
+            offset=None,
+            closed=closed,
+            label=label,
+            # Must be int or uint column otherwise dropping of empty buckets will not work
+            drop_empty_buckets_for="_empty_bucket_tracker_",
+            expected_types=expected_types)
```
alexowens90, line 1006: nit: whitespace

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1314,26 +1344,54 @@ void copy_frame_data_to_buffer(
...
+               (src_column.type().data_type() == DataType::UINT64 && dst_column.type().data_type() == DataType::INT64)) {
+        // Arctic cannot contain both uint64 and int64 columns in the dataframe because there is no common type between
+        // these types. This means that the second condition cannot happen during a regular read. The processing
+        // pipeline, however, can produce a set of segments where some are int64 and other uint64. This can happen in
+        // the sum aggregation (both for unsorted aggregations and resampling). Because we promote the sum type to the
+        // largest type of the respective category. E.g., we have int8 and uint8. Dynamic schema will allow this, and
+        // the global type descriptor will be int16. However, when segments are processed on their own int8 -> int64 and
+        // uint8 -> int64. We have decided to allow this and assign a common type of int64 (done in the modify_schema
+        // procedure). This is what pyarrow does as well. Because of the above, we allow here copying uint64 buffer in
+        // an int64 buffer.
```
alexowens90, line 1388: I don't think this method is called anywhere except from the processing pipeline? So `int_to_float_conversion` is always permissive, and the second condition in the `if` isn't needed?

cpp/arcticdb/processing/clause.hpp
```diff
@@ -645,7 +631,7 @@ struct ColumnStatsGenerationClause {
     }
 
     OutputSchema modify_schema(ARCTICDB_UNUSED OutputSchema&& output_schema) const {
-        internal::raise<ErrorCode::E_ASSERTION_FAILURE>("ColumnStatsGenerationClause::modify_schema should never be called");
+        return OutputSchema{};
```
alexowens90, line 635: Probably worth a comment that this isn't used, and is just there for compatibility

cpp/arcticdb/processing/clause.cpp
```diff
@@ -473,7 +464,11 @@ std::vector<EntityId> AggregationClause::process(std::vector<EntityId>&& entity_
                                 opt_input_column.emplace(std::move(column_with_strings));
                             }
                         }
-                        agg_data->aggregate(opt_input_column, row_to_group, num_unique);
+                        if (opt_input_column) {
+                            // The column is missing from the segment. Do not perform any aggregation and leave it to
+                            // the NullValueReducer to take care of the default values.
+                            agg_data->aggregate(opt_input_column, row_to_group, num_unique);
```
alexowens90: This looks like we should refactor `aggregate` so that the first argument is non-optional?

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,53 @@
...
+    const auto common_input_type = generate_common_input_type(input_agg_columns);
+    if (!common_input_type.has_value()) {
+        return std::nullopt;
+    }
+    Column res(
+        TypeDescriptor(generate_output_data_type(*common_input_type), Dimension::Dim0),
+        output_index_column.row_count(),
+        AllocationType::DYNAMIC,
+        Sparsity::PERMITTED
+    );
```
alexowens90: In the worse case, this will allocate a buffer for `output_index_column.row_count()` (which may be large) values, and then only use one value. I don't think we need to solve it now, but worth leaving a comment at least

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -81,9 +82,9 @@ Column SortedAggregator<aggregation_operator, closed_boundary>::aggregate(const
                                         if (ARCTICDB_LIKELY(current_bucket.contains(*index_it))) {
                                             push_to_aggregator<input_type_info::data_type>(bucket_aggregator, *agg_it, agg_column);
                                             bucket_has_values = true;
-                                        } else if (ARCTICDB_LIKELY(index_value_past_end_of_bucket(*index_it, *bucket_end_it)) && output_it != output_end_it) {
+                                        } else if (ARCTICDB_LIKELY(index_value_past_end_of_bucket(*index_it, *bucket_end_it)) && row_to_write < output_index_column.row_count()) {
                                             if (bucket_has_values) {
-                                                *output_it++ = finalize_aggregator<output_type_info::data_type>(bucket_aggregator, string_pool);
+                                                res.set_scalar(row_to_write++, finalize_aggregator<output_type_info::data_type>(bucket_aggregator, string_pool));
```
alexowens90: This is going to tank the performance in the dense case. It should be possible to precompute the sparsity structure of the output column from `input_index_columns` and `input_agg_columns` (which will also allow us to presize `res` exactly when some `input_agg_columns` are `std::nullopt`) and then use the column data iterators like before to keep the performance high. Probably needs a discussion in person on what this looks like.

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -120,11 +121,46 @@ Column SortedAggregator<aggregation_operator, closed_boundary>::aggregate(const
                                 }
                             }
                         );
+                    } else {
+                        // The column does not contain the aggregation column. However, one of the columns in
```
alexowens90: Whut?

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -120,11 +121,46 @@ Column SortedAggregator<aggregation_operator, closed_boundary>::aggregate(const
...
+                        // Note: The output index can contain fewer buckets than the buckets represented by
+                        // bucket_boundaries. Because if no data falls inside a bucket, we omit if from the result.
+                        const Column& input_index = *input_index_columns[idx];
+                        const timestamp last_index_value = *input_index.scalar_at<IndexTDT::DataTypeTag::raw_type>(input_index.row_count() - 1);
+                        // Check how many rows of the index must be skipped
+                        auto next_output_index_row = row_to_write + (label == ResampleBoundary::LEFT);
+                        while (next_output_index_row < output_index_column.row_count() &&
+                               *output_index_column.scalar_at<timestamp>(next_output_index_row) < last_index_value + (closed_boundary == ResampleBoundary::LEFT)) {
+                            if (bucket_has_values) {
+                                res.set_scalar(row_to_write, finalize_aggregator<output_type_info::data_type>(bucket_aggregator, string_pool));
```
alexowens90: I don't think this is correct in the case where a bucket spans 3 segments, and the column is missing from the middle segment?

cpp/arcticdb/processing/sorted_aggregation.hpp
```diff
@@ -56,15 +66,21 @@ class Bucket {
...
             return ts > start_ && ts <= end_;
         }
     }
+    [[nodiscard]] timestamp start() const {
+        return start_;
+    }
 
+    [[nodiscard]]timestamp end() const {
+        return end_;
+    }
```
alexowens90, line 83: These new methods seems to be unused

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -297,59 +331,30 @@ namespace
...
-                            if constexpr(T == Extremum::MAX) {
-                                *it = in_ptr->written_ ? in_ptr->value_ : std::numeric_limits<typename col_type_info::RawType>::lowest();
-                            } else {
-                                // T == Extremum::MIN
-                                *it = in_ptr->written_ ? in_ptr->value_ : std::numeric_limits<typename col_type_info::RawType>::max();
-                            }
-                        }
+                const std::span<const MaybeValueType> group_values{reinterpret_cast<const MaybeValueType*>(aggregated.data()), aggregated.size() / sizeof(MaybeValueType)};
+                Column::for_each_enumerated<typename col_type_info::TDT>(*col, [&](auto row) {
+                    const bool has_value = group_values[row.idx()].written_;
```
alexowens90: Won't this always be true? The `written_` field is always set to true at the same time as the bit in the sparse map is set (they are really duplicated information, so one should be eliminated, probably `written_` since we need the sparse map)

alexowens90 (changes requested):

cpp/arcticdb/column_store/column.hpp
```diff
@@ -454,6 +453,7 @@ class Column {
     void mark_absent_rows(size_t num_rows);
 
     void default_initialize_rows(size_t start_pos, size_t num_rows, bool ensure_alloc);
+    void default_initialize_rows(size_t start_pos, size_t num_rows, bool ensure_alloc, const std::optional<Value>& default_value);
```
IvoDD, line 456: Nit: Why not just have `default_value` as an optional argument with `std::nullopt` as default?

cpp/arcticdb/entity/stream_descriptor.hpp
```diff
@@ -308,13 +307,27 @@ struct OutputSchema {
...
+    auto release() {
+        column_types_.reset();
+        return std::tuple{std::move(stream_descriptor_), std::move(norm_metadata_), std::move(default_values_)};
+    }
+
+    void clear_default_values() {
+        default_values_.clear();
+    }
+
+    void set_default_value_for_column(std::string name, const Value& value) {
```
IvoDD: Nit: Why not just take `const std::string& name`?

I think if we do (which we probably won't):
```
std::string col_name = "name";
set_default_value_for_column(name, 0)
set_default_value_for_column(name, 1)
```
with a reference can save itself doing the copy on the second call.

cpp/arcticdb/entity/type_utils.cpp
```diff
@@ -67,9 +67,32 @@ namespace arcticdb {
...
+                    false,
+                    "Unknown int to float conversion type {}",
+                    static_cast<std::underlying_type_t<IntToFloatConversion>>(int_to_to_float_conversion));
```
IvoDD, line 85: Not a strong opinion:
We could define a `formatter` for `IntToFloatConversion` similar to e.g. [LoadType](https://github.com/man-group/ArcticDB/blob/2c5bdbc43bde920f154d77b41e9af3365cbccb56/cpp/arcticdb/version/version_map_entry.hpp#L516)

cpp/arcticdb/entity/type_utils.hpp
```diff
@@ -17,18 +16,30 @@ namespace entity {
     struct TypeDescriptor;
 }
 
+/// Defines which static casts from int to float are permitted in is_valid_type_promotion_to_target
+enum class IntToFloatConversion {
+    /// Avoids lossy casting from int to float and tries to make sure that the integer can be represented exactly using
+    /// the specified float, (u)int8, (u)int16 can be represented exactly using float 32, (u)int32 can be represented
+    /// exactly via float64. Note this still allows casting (u)int64 to flaot64 even though it's a lossy cast.
```
IvoDD: Typo: `flaot64`

cpp/arcticdb/pipeline/value.hpp
IvoDD: We should also fix the `assign` which does `memcpy(..., 8)`

cpp/arcticdb/processing/clause.cpp
```diff
@@ -588,6 +589,10 @@ OutputSchema ResampleClause<closed_boundary>::modify_schema(OutputSchema&& outpu
         agg.check_aggregator_supported_with_data_type(input_column_type);
         auto output_column_type = agg.generate_output_data_type(input_column_type);
         stream_desc.add_scalar_field(output_column_type, output_column_name);
+        const std::optional<Value>& default_value = agg.get_default_value(input_column_type);
+        if (default_value) {
+            output_schema.set_default_value_for_column(output_column_name, *default_value);
+        }
```
IvoDD, line 595: Looks like there is a lot in common between `ResampleClause::modify_schema` and `AggregationClause::modify_schema`. 
I don't think we need to refactor this but just for my understanding what are the differences between them? Feels like they should change the schema in very identical ways.

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -295,6 +291,11 @@ std::variant<BitSetTag, DataType> ExpressionNode::compute(
             child,
             [&column_types] (const ColumnName& column_name) -> std::variant<BitSetTag, DataType> {
                 auto it = column_types.find(column_name.value);
+                if (it == column_types.end()) {
```
IvoDD, line 294: It looks we were bitten by this in another place as well in [clasuse.cpp](https://github.com/man-group/ArcticDB/pull/2360/files#diff-63402a4915a118740820a8d2037ac602e98baed0b661fb8e580374acf03fda0cR104). 
What do you think about introducing a new methods in `column_types` which search for both the regular and the mangled name?

cpp/arcticdb/pipeline/value.hpp
vasil-pashov: Fixed the comparison and added some tests.

> We should also fix the assign which does memcpy(..., 8)

I'm not sure. Just copying the full array will probably be easier that going through the hassle of using type visitation to get the raw type.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: Yes, `int_to_float_conversion` is always permissive, I'll change it.

> and the second condition in the if isn't needed

I don't get this part. The second condition is `(src_column.type().data_type() == DataType::UINT64 && dst_column.type().data_type() == DataType::INT64)` we need that in the case of sum aggregations mixing int and uint columns.

cpp/arcticdb/processing/sorted_aggregation.hpp
vasil-pashov: Yes but I find the useful during debugging. Sometimes in debugging I want to print the boundaries to stdout. I'd like to keep them unless you have strong opinion against.

cpp/arcticdb/entity/stream_descriptor.hpp
vasil-pashov: It's unlikely to happen but it's a good point. I'll change it.

cpp/arcticdb/processing/aggregation_interface.hpp
vasil-pashov: I wanted only aggregations whose default is different than the default of the underlying type to return a value. The minimalist in me didn't want to fill the default value map with things we already know about. It's not a hill I'm willing to die on.

cpp/arcticdb/processing/sorted_aggregation.cpp
vasil-pashov: There is already a test `TestResampleDynamicSchema::test_bucket_spans_three_segments` is that what you are thinking about?

cpp/arcticdb/column_store/column.hpp
vasil-pashov: No particular reason, I find it a bit easier to read but it's quite subjective.

cpp/arcticdb/entity/type_utils.cpp
vasil-pashov: I can add one for completeness, but I don't think it can be use in this particular case because we're not printing existing enum values, this is throwing on unknown value.

python/arcticdb/util/test.py
vasil-pashov: I didn't know about this function. It's not exactly the same as arctic but it indeed helped simplifying this function.

python/tests/unit/arcticdb/version_store/test_resample.py
vasil-pashov: Currently bool is not acting like pandas. E.g. summing bools results in True/False instead of integer.

Since this is not going to master and we plan to change the behavior of bools. I'd rather leave it as is and add tests for bools with the PR changing the bool behavior.

cpp/arcticdb/processing/aggregation_interface.hpp
alexowens90: The comment is fine, it just isn't obvious without more context why some aggregations would return `std::nullopt`

cpp/arcticdb/pipeline/read_pipeline.hpp
```diff
@@ -157,9 +157,7 @@ inline std::vector<FilterQuery<ContainerType>> get_column_bitset_and_query_funct
...
-        get_column_bitset_in_context(query, pipeline_context);
-    }
+    get_column_bitset_in_context(query, pipeline_context);
```
alexowens90: `dynamic_schema` and `column_groups` args no longer needed

cpp/arcticdb/pipeline/query.hpp
```diff
@@ -257,7 +257,7 @@ inline void build_row_read_query_filters(
 template <typename ContainerType>
 inline void build_col_read_query_filters(
     std::shared_ptr<PipelineContext> pipeline_context,
-    bool dynamic_schema,
+    bool,
```
alexowens90: Remove

cpp/arcticdb/pipeline/query.hpp
alexowens90: Discussed in person, needed to generate the output schema correctly whether dynamic schema is enabled or not

cpp/arcticdb/pipeline/read_pipeline.hpp
alexowens90: See https://github.com/man-group/ArcticDB/pull/2360#discussion_r2174969997

python/tests/unit/arcticdb/version_store/test_resample.py
alexowens90: Cool, there's already a ticket in Monday to do this in `6.0.0`

cpp/arcticdb/version/version_core.cpp
alexowens90: Ah I thought `is_valid_type_promotion_to_target` would return true for uint64 -> int64 if in permissive mode, this is fine too

cpp/arcticdb/processing/clause.cpp
alexowens90: I think you're right that they will do the same thing. The sorted and unsorted aggregators have different implementations, so it is easiest to duplicate their logic just for schema computation.
The sorted aggregators have a much saner implementation, eventually I want to shift hashed aggregations to use them as well, at which point this problem will go away.

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,161 @@
...
+    const std::span<const std::optional<ColumnWithStrings>> input_agg_columns
+) const {
+    OutputColumnInfo output_column_info;
+    for (const auto& opt_input_agg_column: input_agg_columns) {
+        if (opt_input_agg_column.has_value()) {
+            auto input_data_type = opt_input_agg_column->column_->type().data_type();
+            check_aggregator_supported_with_data_type(input_data_type);
+            add_data_type_impl(input_data_type, output_column_info.data_type_);
+        } else {
+            output_column_info.is_sparse_ = true;
```
alexowens90: I don't think this will always be true?
```
lib.write(sym, pd.DataFrame({"a": [0]}, index=[pd.Timestamp(0)]))
lib.append(sym, pd.DataFrame({"b": [1]}, index=[pd.Timestamp(1)]))
lib.append(sym, pd.DataFrame({"a": [2]}, index=[pd.Timestamp(2)]))
lib.read(sym, lazy=True).resample("min").agg({"a": "sum"}).collect()
```
Of course, we can always just drop the sparse map at the end if `count() == size()`. What is the performance difference between the sparse and dense case?

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,161 @@
...
+        }
+    } else if constexpr (aggregation_operator == AggregationOperator::COUNT) {
+        output_type = DataType::UINT64;
+    }
+    return output_type;
+}
 
 template<AggregationOperator aggregation_operator, ResampleBoundary closed_boundary>
-Column SortedAggregator<aggregation_operator, closed_boundary>::aggregate(const std::vector<std::shared_ptr<Column>>& input_index_columns,
+typename SortedAggregator<aggregation_operator, closed_boundary>::OutputColumnInfo SortedAggregator<aggregation_operator, closed_boundary>::generate_common_input_type(
```
alexowens90: Could extract OutputColumnInfo from the `SortedAggregator` class as it doesn't need any of the template parameters

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -170,25 +263,15 @@ void SortedAggregator<aggregation_operator, closed_boundary>::check_aggregator_s
...
-            output_type = DataType::FLOAT64;
-        }
-    } else if constexpr (aggregation_operator == AggregationOperator::COUNT) {
-        output_type = DataType::UINT64;
+        return details::visit_type(generate_output_data_type(common_input_data_type), [&](auto tag) -> std::optional<Value> {
+            using data_type_tag = decltype(tag);
+            return Value{typename data_type_tag::raw_type{0}, data_type_tag::data_type};
+        });
+    } else {
+        return {};
```
alexowens90, line 278: Should we make `min` and `max` `std::numeric_limits<raw_type>::blah()` to match the hashed aggregations implementation?

cpp/arcticdb/processing/sorted_aggregation.cpp
alexowens90: Ah I see this is handled correctly in `generate_resampling_output_column`, I guess `maybe_sparse_` could be a better name

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,161 @@
...
+
+    if (!type_info.is_sparse_) {
+        return Column(
+            make_scalar_type(generate_output_data_type(*type_info.data_type_)),
+            output_index_column.row_count(),
+            AllocationType::PRESIZED,
+            Sparsity::NOT_PERMITTED
+        );
+    }
+    auto output_index_at = [&output_index_column](const int64_t idx) {
```
alexowens90: Looks like you could use a (possibly enumerating) `ColumnDataIterator`? Or if you really need random access, the `ColumnDataRandomAccessor`?

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,161 @@
...
+        return *(output_index_column.begin<IndexTDT>() + idx);
+    };
+    util::BitSet sparse_map(output_index_column.row_count());
+    int64_t output_row = 0, output_row_prev = 0;
+
+    for (size_t col_index = 0; const std::shared_ptr<Column>& input_index_column : input_index_columns) {
```
alexowens90: `folly::enumerate`

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,161 @@
...
+            output_index_column.row_count(),
+            AllocationType::PRESIZED,
+            Sparsity::NOT_PERMITTED
+        );
+    }
+    auto output_index_at = [&output_index_column](const int64_t idx) {
+        return *(output_index_column.begin<IndexTDT>() + idx);
+    };
+    util::BitSet sparse_map(output_index_column.row_count());
+    int64_t output_row = 0, output_row_prev = 0;
```
alexowens90: style

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,161 @@
...
+
+    for (size_t col_index = 0; const std::shared_ptr<Column>& input_index_column : input_index_columns) {
+        // Skip all labels that come before the first index value in the input column
+        const timestamp first_index_value = *input_index_column->begin<IndexTDT>();
+        while (output_row < output_index_column.row_count() && value_past_bucket_start<closed>(output_index_at(output_row), first_index_value)) {
+            ++output_row;
+        }
+        // If label is left this means the "bucket" is represented by the start of the interval, thus the loop above
+        // skipped to the beginning of the next bucket.
+        output_row_prev = output_row = std::max(int64_t{0}, output_row - (label == ResampleBoundary::LEFT));
```
alexowens90: style

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -8,37 +8,161 @@
...
+
+        // Compute how many output index values does the column span
+        const timestamp last_index_value = *(input_index_column->begin<IndexTDT>() + (input_index_column->row_count() - 1));
+        while (output_row < output_index_column.row_count() && value_past_bucket_start<closed>(output_index_at(output_row), last_index_value)) {
+            ++output_row;
+        }
+        output_row = std::max(int64_t{0}, output_row - (label == ResampleBoundary::LEFT));
+
+        if (input_agg_columns[col_index]) {
+            if (output_row >= sparse_map.size()) {
```
alexowens90, line 131: I think this will always be false by construction?

alexowens90 (approved):

cpp/arcticdb/processing/sorted_aggregation.cpp
vasil-pashov: Prior this PR for unsorted aggregations min/max were promoted to float for dynamic schema and got NaNs by default. With this they are NaN by default for float and 0 by default for int (they have a sparse map so we need to test them in Arrow to make sure it's arrow's None value). So sorted and unsorted aggregations are acting the same way.

cpp/arcticdb/processing/sorted_aggregation.cpp
vasil-pashov: It can be true when a column has values in the last bucket and there are columns after that which are also in the last bucket. There's no need to iterate over the rest columns as we only need to know if there's something in the bucket.

PR #2362, opened 2025-05-20
Update installation_tests.yml with 5.3 and 5.4 final versions

grusev:
#### Any other comments?

Moved 5.2.6 to different timeslot to eliminate the possibility about failures being because timeslot. Although a manual execution shows this problem with 5.2.6. is most probably persisting https://github.com/man-group/ArcticDB/actions/runs/15139549472/job/42559651096)

Added:
 5.3.4 https://github.com/man-group/ArcticDB/actions/runs/15133764164/
 5.4.1 https://github.com/man-group/ArcticDB/actions/runs/15133923361

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/installation_tests.yml
+++ b/.github/workflows/installation_tests.yml
@@ -11,6 +11,8 @@ on:
     - cron: "0 3 * * *"
     - cron: "0 4 * * *"
     - cron: "0 5 * * *"      
+    - cron: "0 6 * * *"      
+    - cron: "0 7 * * *"      
   workflow_dispatch:
     inputs:
       arcticdb_version:
@@ -22,6 +24,8 @@ on:
         type: choice
         options:
           - "latest"
+          - "5.4.1"
+          - "5.3.4"
           - "5.2.6"
           - "5.1.2"
           - "4.5.1"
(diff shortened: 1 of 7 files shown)


alexowens90 (approved):

.github/workflows/installation_tests.yml
```diff
@@ -100,8 +104,10 @@ jobs:
           "01") echo "arcticdb_version=1.6.2" >> $GITHUB_ENV;;
           "02") echo "arcticdb_version=4.5.1" >> $GITHUB_ENV;;
           "03") echo "arcticdb_version=5.1.2" >> $GITHUB_ENV;;
-          "04") echo "arcticdb_version=5.2.6" >> $GITHUB_ENV;;
-          "05") echo "arcticdb_version=latest" >> $GITHUB_ENV;;
+          "04") echo "arcticdb_version=5.3.4" >> $GITHUB_ENV;;
+          "05") echo "arcticdb_version=5.2.6" >> $GITHUB_ENV;;
```
IvoDD, line 108: Nit: 5.2.6 and 5.3.4 are not in order

python/tests/compat/requirements-5.3.4.txt
```diff
@@ -0,0 +1,4 @@
+arcticdb==5.3.4
+numpy<2
```
IvoDD, line 2: For a future PR:

It would be good to have some installation tests which run on numpy==2 as well. A lot of our users on more recent python versions would use it. See [here](https://github.com/man-group/ArcticDB/pull/2050#discussion_r1895568329)

The only reason to keep these pinned for now is to be able to run on old python versions on CI.

IvoDD (approved):

ognyanstoimenov:
I am a bit confused by the use of these requirements.txt files for each version of arcticdb here. If the intention of these installation tests is to simulate installing as the end user would, why do we need requirements that are different from the packaged arcticdb in pypi/conda. IIUC these tests would run with multiple python versions so this would be a way to test different pins (e.g. the numpy one)

python/tests/compat/requirements-5.3.4.txt
grusev: that is good proposal I will add enhancement task for that

PR #2364, opened 2025-05-21
Try to fix flaky compat tests

grusev:

When we write dataframes we pass deepcopy of them
Instead of Parquet we write/read CSV (also transfer metainfo with small additional code)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/util/venv.py
+++ b/python/arcticdb/util/venv.py
@@ -124,17 +124,36 @@ class VenvArctic:
 
         with tempfile.TemporaryDirectory() as dir:
             df_load_commands = []
+            index_load_commands = []
             for df_name, df_val in dfs.items():
                 import pandas as pd
-                df_value : pd.DataFrame = df_val.copy(deep=True)
+                import numpy as np
+                # For each dataframe we are writing we create a deep copy which we pass
+                df: pd.DataFrame = df_val.copy(deep=True)
                 if self.parquet:
                     file = os.path.join(dir, f"{df_name}.parquet")
-                    df_value.to_parquet(file)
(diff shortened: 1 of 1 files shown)


python/arcticdb/util/venv.py
```diff
@@ -118,17 +122,48 @@ def execute(self, python_commands: List[str], dfs: Optional[Dict] = None) -> Non
...
+                            index_load_commands.append(f"{df_name}.index.freq = {repr(df.index.freqstr)}")
+                        elif isinstance(df.index, pd.RangeIndex):
+                            index_load_commands = [ 
+                                                   f"{df_name}.index.start = {df.index.start}",
+                                                   f"{df_name}.index.stop = {df.index.stop}",
+                                                   f"{df_name}.index.step = {df.index.step}",
+                                                ]
+                        else:
+                            index_load_commands
+                        command = f"{df_name} = pd.read_csv({repr(file)}, dtype={dtypes_dict}, parse_dates=True, index_col=0)"
```
IvoDD, line 163: This most certainly won't work with pandas multi-index, but no compat tests require this for now. We can fix or use parquet when we want a compat test which uses multi-index.

IvoDD (approved):

python/arcticdb/util/venv.py
grusev: yep, have not tried this on other dataframes, but for now we do use only one index, so this logic is not general but good enough for compat test where it is only used

alexowens90 (approved):

PR #2365, opened 2025-05-23
Bugfix 9209057536: Allow concatenation of uint64 columns with int* columns

alexowens90:
#### Reference Issues/PRs
Fixes [9209057536](https://man312219.monday.com/boards/7852509418/pulses/9209057536)

Allows concatenating columns of type uint64 with columns of type int*

### DIFF
--- a/cpp/arcticdb/entity/type_utils.cpp
+++ b/cpp/arcticdb/entity/type_utils.cpp
@@ -198,4 +198,16 @@ namespace arcticdb {
         }
     }
 
+    // The above method requires that both types are exactly representable by one type
+    // This is more permissive, allowing (for example) a uint64_t to be combined with an int* with the result being a double
+    std::optional<entity::TypeDescriptor> promotable_type(const entity::TypeDescriptor& left, const entity::TypeDescriptor& right) {
+        auto res = has_valid_common_type(left, right);
+        if (!res.has_value() &&
+            is_valid_type_promotion_to_target(left, make_scalar_type(entity::DataType::FLOAT64)) &&
+            is_valid_type_promotion_to_target(right, make_scalar_type(entity::DataType::FLOAT64))) {
+            res = make_scalar_type(entity::DataType::FLOAT64);
+        }
+        return res;
+    }
+
 }
--- a/cpp/arcticdb/processing/clause_utils.cpp
+++ b/cpp/arcticdb/processing/clause_utils.cpp
@@ -408,10 +408,10 @@ void inner_join(StreamDescriptor& stream_desc, std::vector<OutputSchema>& input_
                     // and if necessary modify the columns_to_keep value to a type capable of representing all
                     auto& current_data_type = columns_to_keep_it->second;
                     if (current_data_type.has_value()) {
-                        auto opt_common_type = has_valid_common_type(make_scalar_type(*current_data_type),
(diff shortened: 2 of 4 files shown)


cpp/arcticdb/processing/clause_utils.cpp
```diff
@@ -408,10 +408,10 @@ void inner_join(StreamDescriptor& stream_desc, std::vector<OutputSchema>& input_
...
-                        auto opt_common_type = has_valid_common_type(make_scalar_type(*current_data_type),
-                                                                     make_scalar_type(it->second));
-                        if (opt_common_type.has_value()) {
-                            current_data_type = opt_common_type->data_type();
+                        auto opt_promotable_type = promotable_type(make_scalar_type(*current_data_type),
+                                                                   make_scalar_type(it->second));
+                        if (opt_promotable_type.has_value()) {
+                            current_data_type = opt_promotable_type->data_type();
                         } else {
                             current_data_type.reset();
```
maxim-morozov, line 416: If the data type is not promotable between two columns, why do we stick to the original data type?

python/tests/unit/arcticdb/version_store/test_symbol_concatenation.py
```diff
@@ -64,6 +64,20 @@ def test_symbol_concat_basic(lmdb_library_factory, dynamic_schema, rows_per_segm
         assert version.metadata == (None if idx == 1 else idx)
 
 
+@pytest.mark.parametrize("first_type", ["uint8", "uint16", "uint32", "uint64", "int8", "int16", "int32", "int64", "float32", "float64"])
```
maxim-morozov, line 67: It would be good to have a test of non promotable types for inner and out joins.

cpp/arcticdb/processing/clause_utils.cpp
alexowens90: We can't throw here, as the column where the types do not match may not be in the output result set. e.g. if you concat
```
pd.DataFrame({"a": [0], "b": [1]})
pd.DataFrame({"a": [0], "b": ["hello"]})
pd.DataFrame({"a": [0]})
```
the types of column `b` are incompatible between the first and second symbols, but it isn't present in all of the symbols, so it won't be in the output columns

python/tests/unit/arcticdb/version_store/test_symbol_concatenation.py
alexowens90: These are covered in `test_join_schemas.cpp`

cpp/arcticdb/processing/clause_utils.cpp
maxim-morozov: I see. Makes sense. I guess in this case, we could also accumulate all column mismatches and provide error on all of them, instead of the first one. Not critical, but can be nicer user experience.

maxim-morozov (approved):

cpp/arcticdb/entity/type_utils.cpp
```diff
@@ -198,4 +198,16 @@ namespace arcticdb {
         }
     }
 
+    // The above method requires that both types are exactly representable by one type
```
vasil-pashov, line 201: Can we add the name of the method. It might become confusing if code gets reordered.

PR #2366, opened 2025-05-27
Extend test coverage for recursive normalizers

poodlewars:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/integration/arcticdb/test_unicode_strings.py
+++ b/python/tests/integration/arcticdb/test_unicode_strings.py
@@ -108,3 +108,21 @@ def test_recursive_normalizers_blns(lmdb_version_store):
     lib.write(symbol, dict, recursive_normalizers=True)
     vit = lib.read(symbol)
     assert_dicts_of_dfs_equal(dict, vit.data)
+
+
+def test_recursive_normalizers_blns_in_keys(lmdb_version_store):
+    lib = lmdb_version_store
+    strings = read_strings()
+    symbol = "blnd_recursive_in_keys"
+    df = pd.DataFrame({"a": [1, 2, 3]})
+
+    for s in strings:
+        dict = {s: df}
+        try:
+            lib.write(symbol, dict, recursive_normalizers=True)
+        except:
+            # We just want to check that we can read anything we can write, so just skip anything we can't write
+            continue
+        vit = lib.read(symbol)
+        assert s in vit.data
+        pd.testing.assert_frame_equal(dict[s], vit.data[s])
--- a/python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
+++ b/python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
@@ -1,3 +1,5 @@
+from collections import namedtuple
+
 import pandas as pd
 import pytest
 import numpy as np
@@ -10,6 +12,7 @@ from arcticdb.exceptions import ArcticDbNotYetImplemented
 from arcticdb.util.venv import CompatLibrary
 from arcticdb.util.test import assert_frame_equal
 from arcticdb_ext.storage import KeyType
+from arcticdb_ext.version_store import NoSuchVersionException
 
 
 class AlmostAList(list):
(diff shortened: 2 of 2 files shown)


python/tests/integration/arcticdb/test_unicode_strings.py
```diff
@@ -107,4 +109,23 @@ def test_recursive_normalizers_blns(lmdb_version_store):
...
+    strings = read_strings()
+    symbol = "blnd_recursive_in_keys"
+    df = pd.DataFrame({"a": [1, 2, 3]})
+
+    for s in strings:
+        dict = {s: df}
+        try:
+            lib.write(symbol, dict, recursive_normalizers=True)
+        except:
+            # We just want to check that we can read anything we can write, so just skip anything we can't write
```
alexowens90, line 127: When we get this test otherwise passing, we should also assert that a sensible exception is thrown for keys we fail to write

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1152,6 +1152,16 @@ def test_filter_date_range_pickled_symbol(request, lib_type):
...
 
 
+def test_filter_date_range_none_none(lmdb_version_store_v1):
+    lib = lmdb_version_store_v1
+    symbol = "sym"
+    idx = pd.date_range("2000-01-01", periods=4)
+    df = pd.DataFrame({"a": [1, 2, 3, 4]}, index=idx)
+    lib.write(symbol, df)
+    vit = lib.read(symbol, date_range=(None, None))
+    assert_frame_equal(vit.data, df)
```
alexowens90, line 1162: Was this added on purpose?

poodlewars: Yeah, I couldn't (easily) find tests for `date_range=(None, None)` so thought I'd add one here

poodlewars:
> LGTM. Do you think it is worth adding tests (if there aren't any) for the most basic cases like:
> 
>     * using `recursive_normalizers=True` for data that doesn't need to be normalized?
> 
>     * using `recursive_normalizers=False` (or omitting it) for when it is needed. Do we fail gracefully then?

Good ideas, done

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
```diff
@@ -169,6 +192,52 @@ def test_data_directly_msgpackable(basic_store):
...
+    lib.write("sym", df, recursive_normalizers=True)
+
+    # Then
+    assert lib.get_info("sym")["type"] == "pandasdf"
+    assert_frame_equal(lib.read("sym").data, df)
+    lt = lib.library_tool()
+    assert len(lt.find_keys(KeyType.MULTI_KEY)) == 0
+
+
+@pytest.mark.parametrize("pickle_on_failure", (True, False))
```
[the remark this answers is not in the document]
poodlewars: Ah great point. The `pickle_on_failure` logic is just this in the `CompositeNormalizer`,

```
        try:
            return self._normalize(
                item,
                string_max_len=string_max_len,
                dynamic_strings=dynamic_strings,
                coerce_columns=coerce_columns,
                empty_types=empty_types,
                **kwargs,
            )
        except Exception as ex:
            log.debug("Could not normalize item of type: {} with the default normalizer due to {}", type(item), ex)
            if pickle_on_failure:
                log.debug("pickle_on_failure flag set, normalizing the item with MsgPackNormalizer", type(item), ex)
                return self.fallback_normalizer.normalize(item)
            else:
                raise
```

and in this case there's a `fallback_normalizer` for `msgpack`. It seems like the idea is that in this case we're msgpack serializing the dict, with a Pickle serialized dataframe inside, but that doesn't count as "pickling on failure".



Our docs on `NativeVersionStore#write` seems to have the unsubtle idea of how it should work:

```
        Pandas DataFrames, Pandas Series and Numpy NDArrays will be normalised into a common structure suitable for
        storage. Data that cannot be normalised can be written by pickling the data, however pickled data
        consumes more storage space, is less performant for reads and writes and does not support advanced query
        features. Pickling is therefore only supported via the `pickle_on_failure` flag.
```

and misleadingly gives the impression that the Pandas types will never end up pickled.

PR #2371, opened 2025-05-28
Fix for scheduled workflow build failures to appear in slack

grusev:

The fix aims to include scheduled workflow failures which does not appear on Slack. The reason is that they might not carry head_branch info

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/failure_notification.yaml
+++ b/.github/workflows/failure_notification.yaml
@@ -7,6 +7,7 @@ on:
 jobs:
   on-failure:
     runs-on: ubuntu-latest
+    # This condition aims to include scheduled workflow failures which does not appear on Slack. The reason is that they might not carry head_branch info
     if: |
       (github.event.workflow_run.conclusion == 'failure' || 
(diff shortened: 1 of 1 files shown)


.github/workflows/failure_notification.yaml
```diff
@@ -3,12 +3,15 @@ on:
...
     workflows: ["Build and Test", "Build with conda", "Build with analysis tools", "Coverity Static Analysis", "Installation Tests Execution"]
     types: [completed]
-    branches: [master]
 
 jobs:
   on-failure:
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion == 'failure' || github.event.workflow_run.conclusion == 'timed_out'
+    if: |
+      (github.event.workflow_run.conclusion == 'failure' ||
```
G-D-Petrov, line 11: Can you please add a comment here for that the `schedule` workflow doesn't set the `head_branch` properly and that is why this check is needed

PR #2375, opened 2025-05-29
Change delete to load only undeleted versions

G-D-Petrov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -62,43 +63,44 @@ def verify_list_symbol_stats(list_symbol_call_counts):
...
             else:
-                assert storage_operations[key_ops][op]["size_bytes"] == 0
+                assert storage_operations[key_ops][op]["size_bytes"] == 0, storage_operations
 
 
 def test_query_stats(s3_version_store_v1, clear_query_stats):
     s3_version_store_v1.write("a", 1)
     qs.enable()
-    
+    qs.reset_stats()
```
phoebusm, line 90: I am not sure the purpose of it as fixture `clear_query_stats` should have cleared the stats

python/tests/integration/arcticdb/test_s3.py
```diff
@@ -222,3 +226,161 @@ def test_library_get_key_path(lib_name, storage_bucket, test_prefix):
...
+    total_diff = total2 - total1
+    diff_str = f"{total_diff:+d}" if total_diff != 0 else "0"
+    summary = f"Total:".ljust(50) + f"{total1:>10} {total2:>10} {diff_str:>10}"
+    output.append(summary)
+
+    return "\n".join(output)
+
+
+def test_delete_over_time(lib_name, storage_bucket):
+    qs.enable()
```
phoebusm, line 327: query stats won't be disabled automatically. `clear_query_stats` is needed to be used in the test so query stats can be reset

G-D-Petrov: Thanks, that makes sense, I've changed it to use `clear_query_stats`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -890,8 +890,8 @@ folly::Future<folly::Unit> delete_trees_responsibly(
             auto min_versions = min_versions_for_each_stream(orig_keys_to_delete);
             for (const auto& min : min_versions) {
                 auto load_strategy = load_type == LoadType::DOWNTO
-                        ? LoadStrategy{load_type, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(min.second)}
-                        : LoadStrategy{load_type, LoadObjective::INCLUDE_DELETED};
+                        ? LoadStrategy{load_type, LoadObjective::UNDELETED_ONLY, static_cast<SignedVersionId>(min.second)}
+                        : LoadStrategy{load_type, LoadObjective::UNDELETED_ONLY};
```
IvoDD, line 894: As discussed offline this is safe because delayed deletes doesn't use this codepath.

However when we transition delayed deletes to arcticdb and re-enable tombstone timestamps we'll need to think about this. Do you mind adding this note to the ticket for re-enabling tombstone timestamps? (If there is one maybe we should create it)
To me it is not immediately obvious that if delayed deletes encounters a version chain like:
```
v0 <- v1 <-v2 <- t_all2 <-v3
```
It should stop traversing after reaching `v2`. I.e. if `v1` is still not physically deleted `v2` might share data keys with it. Most probably that is fine as anything shared between `v1` and `v2` is logically deleted and doesn't matter on which version we will perform the physical deletion, but it will require some more careful thought.

IvoDD (approved):

python/benchmarks/basic_functions.py
```diff
@@ -361,3 +363,9 @@ def time_delete_multiple_versions(self, lad: LargeAppendDataModify, rows):
 
     def time_delete_short_wide(self, lad: LargeAppendDataModify, rows):
         self.lib_short_wide.delete("short_wide_sym")
+
+    def time_delete_over_time(self, lad: LargeAppendDataModify, rows):
```
IvoDD, line 367: What do you think about adding such a benchmark on real storage? I think on LMDB where latency is low this might not have a big impact with just a 100 versions.

cpp/arcticdb/version/local_versioned_engine.cpp
G-D-Petrov: Added a task here - 9303299968 and added a link to the comment

alexowens90 (approved):

PR #2378, opened 2025-05-30
Various small improvements to recursive normalizers

poodlewars:
The first commit 630de4 just shows how to fix an existing TODO in the code to make the Flattener non-recursive. It turns out that `msgpack.packb` which we call straight afterwards is itself recursive as it serializes nested objects, so this change doesn't add any value and I just revert it in 2b8db. The git history is like this so that the reverted commit is still available in the history for people.

b287372 introduces a clearer error message when we hit that msgpack recursion limit.

c872353 just fixes a comment that misunderstood what `pickle_on_failure` controls.

Recursive normalizers rely on serializing data using a separator of `__` which meant that dictionaries with keys containing `__` did not roundtrip correctly (typically the `__` part of the keys would vanish upon read). 9ff3e94 validates against writing dictionaries with such a key. Alex and I discussed a lot about introducing a new serialization format that could handle these and decided against it - we can always do so in future.

ea0b066 widens the set of dict keys that we can write. We were previously applying our symbol name validation to them, which is stronger than necessary.


python/arcticdb/flattener.py
```diff
@@ -129,8 +146,16 @@ def will_obj_be_partially_pickled(self, obj):
...
 
-    def _create_meta_structure(self, obj, sym, to_write):
-        # TODO: convert to non recursive and remove to_write from func arguments to not rely on external state.
+    def _create_meta_structure(self, obj, sym, to_write, depth=0, original_symbol=None):
+        if original_symbol is None:
+            original_symbol = sym  # just used for error messages
+
+        if depth > DEFAULT_RECURSE_LIMIT // 2:
+            raise DataTooNestedException(f"Symbol {original_symbol} cannot be recursively normalized as it contains more than "
+                                         f"{DEFAULT_RECURSE_LIMIT // 2} levels of nested dictionaries. This is a limitation of the msgpack serializer.")
```
alexowens90, line 156: I don't understand the factor of 2, so probably worth a comment

cpp/arcticdb/stream/python_bindings.hpp
```diff
@@ -19,20 +19,20 @@ namespace stream {
 void register_stream_bindings(py::module &m);
 
 inline void register_bindings(py::module &m) {
-    auto arcticxx_types = m.def_submodule("types", R"pydoc(
+    auto arcticdb_ext_types = m.def_submodule("types", R"pydoc(
```
alexowens90, line 22: Will the bridge automagically handle these changes?

cpp/arcticdb/util/error_code.hpp
```diff
@@ -160,6 +162,18 @@ struct ArcticSpecificException : public ArcticCategorizedException<get_error_cat
...
+struct Error {
+
+    explicit Error(folly::Function<void(std::string)> raiser, std::string msg);
+    void throw_error();
+
+    folly::Function<void(std::string)> raiser_;
+    std::string msg_;
+};
+
+using CheckOutcome = std::variant<Error, std::monostate>;
```
alexowens90, line 175: Not needed now, but this would be more generally useful if templated on the second variant type, so it could be used as the return type from non-void functions as well

cpp/arcticdb/stream/python_bindings.hpp
poodlewars: This is just renaming the variable - so no impact on our APIs

poodlewars:
I've rebased - only https://github.com/man-group/ArcticDB/pull/2378/commits/3757c4da547d16c7e8f191c0bfc9ca3133633ba4 is new since Alex's review

cpp/arcticdb/util/name_validation.cpp
```diff
@@ -105,7 +115,10 @@ void verify_symbol_key(const StreamId& symbol_key) {
 constexpr auto UNSUPPORTED_LMDB_MONGO_PREFIX = '/';
 
 void verify_library_path(const StringId& library_path, char delim) {
-    verify_name("library name", library_path, false, {}, {}, delim);
+    CheckOutcome  res = verify_name("library name", library_path, false, {}, {}, delim);
```
ognyanstoimenov, line 118: Nice pattern! nit: two spaces here

ognyanstoimenov (approved):

PR #2379, opened 2025-06-03
abalyze failures for storage locks

grusev:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/integration/arcticdb/test_storage_lock.py
+++ b/python/tests/integration/arcticdb/test_storage_lock.py
@@ -1,8 +1,10 @@
+import os
 import pandas as pd
 import numpy as np
 import pytest
 import sys
 
+from arcticdb.util.utils import get_logger
 from arcticdb_ext.tools import ReliableStorageLock, ReliableStorageLockManager
 from tests.util.mark import REAL_S3_TESTS_MARK
 
@@ -11,6 +13,7 @@ import time
 from arcticdb.util.test import assert_frame_equal
 from multiprocessing import Process
 
+logger = get_logger()
 
 one_sec = 1_000_000_000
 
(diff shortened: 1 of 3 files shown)


python/tests/integration/arcticdb/test_storage_lock.py
```diff
@@ -1,36 +1,54 @@
...
+    symbol_name = f"{symbol_prefix}{pid}"
+    lib.write(symbol_name, df)
+    logger.info(f"Process {pid}: wrote unique symbol {symbol_name}")
     lock_manager.free_lock_guard()
+    logger.info(f"Process {pid}: completed")
 
-
-@pytest.mark.parametrize("num_processes,max_sleep", [(100, 1), (5, 20)])
+# NOTE: Is there is not enough memory the number of actually spawned processes
+# will be lowe. The test counts the actual processes that did really got executed
```
IvoDD, line 50: Couple of typos: `Is` -> `If`, `lowe` -> `lower`

python/tests/integration/arcticdb/test_storage_lock.py
```diff
@@ -1,36 +1,54 @@
...
     df = lib.read(symbol).data
-    df["col"][0] = df["col"][0] + 1
+    logger.info(f"Process {pid}: previous value {df['col'][0]}")
+    df["col", 0] = df["col"][0] + 1
     time.sleep(sleep_time)
     lib.write(symbol, df)
+    logger.info(f"Process {pid}: incrementing and saving value {df['col'][0]}")
+    symbol_name = f"{symbol_prefix}{pid}"
+    lib.write(symbol_name, df)
+    logger.info(f"Process {pid}: wrote unique symbol {symbol_name}")
```
IvoDD, line 45: The principal way to communicate how many processes succeeded to the main thread is to use a [Queue](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue).

Each process can push an item to the queue when it succeeds. [Example usage from Ognyan](https://github.com/man-group/ArcticDB/blob/fbfd7dd3cc8ffc436b3e76fd11befcf961900040/python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py#L53-L58).

The current approach could theoretically fail if we succeed at incrementing the counter but fail with the follow up write due to e.g. a network error. This failure would be less likely with a `multiprocessing.Queue`

IvoDD: Happy to leave as is to save running the tests again and waiting a lot. If we encounter further flakyness in the future we can do as suggeted.

IvoDD (approved):

python/tests/integration/arcticdb/test_storage_lock.py
grusev: agree, but for now works. Let's push this version at the moment

PR #2383, opened 2025-06-04
Fix moto teardown rare problem

grusev:

Address teardown error: https://github.com/man-group/ArcticDB/actions/runs/13861408824/job/38790568264

```
ERROR tests/integration/arcticdb/test_arctic_batch.py::test_write_metadata_batch_with_none[nfs_backed_s3-0] - requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=13635): Max retries exceeded with url: /moto-api/reset (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x14ece9c10>: Failed to establish a new connection: [Errno 61] Connection refused'))
= 7609 passed, 1848 skipped, 82 xfailed, 7 xpassed, 96189 warnings, 1 error in 1210.77s (0:20:10) =
```

the fix is safe - the fixture is at session scope and no test will run after that, moreover this is not real storage but simulated one

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/storage_fixtures/s3.py
+++ b/python/arcticdb/storage_fixtures/s3.py
@@ -880,9 +880,13 @@ class MotoS3StorageFixtureFactory(BaseS3StorageFixtureFactory):
                 if e.response["Error"]["Code"] != "NoSuchBucket" and not is_win_37:
                     raise e
         else:
-            requests.post(
-                self._iam_endpoint + "/moto-api/reset", verify=False
-            )  # If CA cert verify fails, it will take ages for this line to finish
+            try:
+                requests.post(
+                    self._iam_endpoint + "/moto-api/reset", verify=False
+                )  # If CA cert verify fails, it will take ages for this line to finish
+            except Exception:
+                # We clean bucket at session level so failure here does not matter
+                pass            
             self._iam_admin = None
 
 


vasil-pashov (commented):
I'm not familiar with the issue. Can you give more information about:

1. What's the root cause
2. Why does this fix it
3. Can we end in situation where the bucket is not cleaned? If so what are the consequences of that?

grusev:
> I'm not familiar with the issue. Can you give more information about:
> 
> 1. What's the root cause
> 2. Why does this fix it
> 3. Can we end in situation where the bucket is not cleaned? If so what are the consequences of that?

> 1. What's the root cause - the log on top has captured an error at teardown process of moto fixtures. That can happen if there is connectivity issue or the moto simulator is already dead. ( there is already several repeats that timeoud out)

> 2. Why does this fix it - the fix assumes such things can happen, as opposed to code which incorrectly assumes such things will never happen. Thus the fix is fixing wrong assumption at the very begining. NOTE: a problem in teardown is never something that can and should be accounted as failure.

> 3. This is irrecoverable situation. There is notthing to be done. If we leave it we will suffer only wrong status when nothing is broken. Additionally as this is something that lives only until started in memory it is for sure that all things on next start should not be preserved. NOTE also that teardown is at session scope - in other words ALL TESTEST COMPLETED already.

vasil-pashov (approved):

PR #2384, opened 2025-06-04
Modify schema [v2] draft

vasil-pashov:
#### Change of how dtypes are computed

This PR ensures that the dtypes of the columns will always be deterministic when processing such as (filtering/aggregation/resampling/etc..) is performed. The previous behavior returned dtypes based on the segments that were read, e.g. given a dataframe with each row being a separate segment

index    column_a_dtype
0          int8
1          uint16
2          uin8
3          float32

reading row 0 will produce int8, row 1 uint16, etc.. Reading row 0 and row 1 will be int32, reading row 1 and row 2 will be uint16, reading all rows will be float32. In other words the returned dtype was determined by examining the dtypes of all segments and taking the largest type that can represent all of them. With this change the output dtype will always be the same and it will be the dtype able to represent all dtypes in the dataframe.

#### Breaking change in unsorted aggregation (groupby) and resampling

Previously min and max aggregations in groupby clauses were always promoted to float64 so that if a group is empty the result would be `NaN`. This is no longer the case as the dtype is stable. In case the dtype is (un)signed integer the value will be zero (when the numpy backend is used, in future when arrow is used it will come out as missing value).

#### Implementation of resampling when dynamic schema is used.

Prior this PR libraries using dynamic schema could not use the resampling functionality of the query builder. With this PR this is fully implemented.

#### Sum aggregation dtype
When sum aggregation is performed (unsorted or resampling) the result dtype is the largest type of the respective category, e.g. int -> int64, uint -> uint64, float -> float64, this is done to avoid overflows.

When the dataframe contains int and uint colums this leads to exception and the query could not be executed. The int segment would be promoted to int64, the uint segment will be promoted to uint64 and we do not support a type able to represent both of them. This is fixed with this PR and no longer throws an exception. The common type is decided to be int64 in this particular case. This is because ArcticDB does not allow having mixing uint64 with any int type (append/update will throw). Which means that the largest possible types are uint32 and int64, whose common type is int64, which exactly aligns with our algorithms for type promotion.

### Implementation details
1. Add option to initialize column using a custom default value.
2. Make the columns for the min/max unsorted aggregation sparse so that proper backfilling can be done.
3. Add map of default values in the OutputSchema that's later stored in the PipelineContext so that CopyToBufferTask can do backfilling. This is used by the sum aggregations as they both need to return 0.0 instead of NaN for float dtypes.
4. Add permissive mode of conversion from int to float. This new mode is used only in the processing pipeline after processed data is copied to output buffers. It allows casting int to float even when the the float cannot exactly represent the int type, e.g. casting int64 to float32
5. Change how unsorted aggregations handle the sum column of empty type. They used to add the column and fill it with zero. Now the column is not added and is handled by the null value reducer.
6. The column returned by sorted aggregation
6.1. It is now sparse. With dynamic schema there can be many segments and different columns could fill different buckets in those segments.
6.2. It is now optional because the aggregated column can be missing from all segments.
7. Handle sparse columns when copying trivially compatible types in `copy_frame_data_to_buffer`. The current implementation did not consider sparse columns and used `memcpy` to copy all blocks to the output buffer. This does not work with sparse columns. The new implementation first checks if the column is dense, if so uses the old logic, but if it's sparse it handles it differently to account for missing values.
8. Improve the hypothesis test for unsorted aggregation with dynamic schema. It used only 3 segments with predefined pattern that did not capture all possible issues. The new tests generates many segments with different sets of columns.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


vasil-pashov:
PR moved in https://github.com/man-group/ArcticDB/pull/2360

PR #2385, opened 2025-06-04
Query Builder regex filter support and upgrade PCRE to PCRE2

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/9238018056

This brings support
1. Regex support on query builder filtering. The new api `regex_match` only sypports filtering str columns.
This PR includes the implementation, docstrings and correponding tests.
2. Upgrade regex backend from `PCRE` to `PCRE2` for the sake of UTF32 support on conda
- Please be noted that although this is a breaking change, as `PCRE2` has a stricter check on the pattern
- The only API that being affected is the regex filtering in `def list_symbols()`
- Major version won't be bumped as the patterns being impacted are obscure and doubt anyone will use on `list_symbols()`
- For differences in details, please refer to https://stackoverflow.com/questions/70273084/regex-differences-between-pcre-and-pcre2/73767663#73767663

#### Any other comments?
Valid asv run: https://github.com/man-group/ArcticDB/actions/runs/15472639473
The default asv CI triggered by this PR is destined to fail as it will test the new API in the new asv test against the existing master wheel. 
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/arcticdb/version_store/processing.py
```diff
@@ -429,7 +439,7 @@ class QueryBuilder:
 
     Supported filtering operations:
 
-    * isna, isnull, notna, and notnull - return all rows where a specified column is/is not NaN or None. isna is
+    * isna, isnull, notna, notnull and regex_match - return all rows where a specified column is/is not NaN or None. isna is
```
alexowens90: This doesn't belong here

python/arcticdb/version_store/processing.py
```diff
@@ -232,6 +233,15 @@ def notna(self):
 
     def notnull(self):
         return ExpressionNode.compose(self, _OperationType.NOTNULL, None)
+    
+    def regex_match(self, pattern: str):
+        if isinstance(pattern, str):
+            _RegexPattern(pattern)  # Validate the regex pattern
```
alexowens90: This is a bit hacky, can we have a function that does this?

python/arcticdb/version_store/processing.py
```diff
@@ -450,6 +460,8 @@ class QueryBuilder:
 
         q.isin(1, 2, 3)
 
+    regex_match accepts string as pattern and can only filter string columns
```
alexowens90: Could mention that it is similar to Pandas `contains` (which is an awful name)

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1319,3 +1320,154 @@ def test_filter_unsupported_boolean_operators():
...
+    assert_frame_equal(lib.read(sym, query_builder=q_c).data, df[df.c.str.contains(pattern_c)])
+
+    pattern_c2 = r"1\d[a-zA-Z]"
+    q_c2 = QueryBuilder()
+    q_c2 = q_c2[q_c2["c"].regex_match(pattern_c2)]
+    assert_frame_equal(lib.read(sym, query_builder=q_c2).data, df[df.c.str.contains(pattern_c2)])
+
+    q = QueryBuilder()
+    q = q[q["a"].regex_match(pattern_a) & q["c"].regex_match(pattern_c)]
+    expected = df[df.a.str.contains(pattern_a) & df.c.astype(str).str.contains(pattern_c)]
```
alexowens90: How come the `astype` is needed here?

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1319,3 +1320,154 @@ def test_filter_unsupported_boolean_operators():
...
+    q = QueryBuilder()
+    q = q[q["a"].regex_match(pattern_a) & q["c"].regex_match(pattern_c)]
+    assert lib.read(sym, query_builder=q).data.empty
+
+    q2 = QueryBuilder()
+    q2 = q2[q2["a"].regex_match(pattern_a) & q2["b"].isin([0])]
+    assert lib.read(sym, query_builder=q2).data.empty
+
+
+def test_filter_regex_match_empty_symbol(lmdb_version_store_v1, sym):
```
alexowens90: I think this behaviour will change after the `modify_schema` branch is merged, can just remove this test

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1319,3 +1320,154 @@ def test_filter_unsupported_boolean_operators():
...
+    df = pd.DataFrame(
+            index=pd.date_range(pd.Timestamp(0), periods=3),
+            data={"a": ["abc", "abcd", "aabc"], "b": [1, 2, 3]}
+        )
+    lib.write(sym, df)
+
+    with pytest.raises(UserInputException):
+        q = QueryBuilder()
+        q = q[q["b"].regex_match(r"\d+")]
+        lib.read(sym, query_builder=q)
```
alexowens90, line 1441: This should probably be in a separate test, as it is about type compatibility, not invalid patterns

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1319,3 +1320,154 @@ def test_filter_unsupported_boolean_operators():
...
+    pattern = "^" + unicode_symbols
+    q = QueryBuilder()
+    q = q[q["a"].regex_match(pattern)]
+    expected = df[df.a.str.contains(pattern)]
+    received = lib.read(sym, query_builder=q).data
+    assert_frame_equal(expected, received)
+    assert not expected.empty
+
+
+def test_filter_regex_match_chaining(lmdb_version_store_v1, sym):
```
alexowens90: We already test chaining filters with groupings [here](https://github.com/man-group/ArcticDB/blob/9c2ec0cfe39c10c04f5a38148d8b0fe832af18f0/python/tests/unit/arcticdb/version_store/test_query_builder.py#L731), so this test can be removed

cpp/arcticdb/column_store/string_pool.cpp
```diff
@@ -240,4 +237,20 @@ ankerl::unordered_dense::set<position_t> StringPool::get_offsets_for_column(cons
     }
     return output;
 }
+
+ankerl::unordered_dense::set<position_t> StringPool::get_regex_match_offsets_for_column(std::string_view str, const Column& column) const {
+    auto unique_values = unique_values_for_string_column(column);
+    remove_nones_and_nans(unique_values);
+    
+    util::RegexPattern pattern{std::string(str)};
```
alexowens90: This will mean the regex is re-compiled for every row-slice, it should only be compiled once

cpp/arcticdb/processing/operation_dispatch_binary.cpp
```diff
@@ -168,4 +170,74 @@ VariantData dispatch_binary(const VariantData& left, const VariantData& right, O
...
+    if (is_empty_type(column_with_strings.column_->type().data_type())) {
+        return EmptyResult{};
+    }
+    util::BitSet output_bitset;
+    details::visit_type(column_with_strings.column_->type().data_type(), [&](auto col_tag) {
+        using col_type_info = ScalarTypeInfo<decltype(col_tag)>;
+        details::visit_type(val.type().data_type(), [&](auto val_tag) {
+            using val_type_info = ScalarTypeInfo<decltype(val_tag)>;
+            if constexpr(is_sequence_type(col_type_info::data_type) && is_sequence_type(val_type_info::data_type)) {
+                std::string value_string = get_string_from_value_type(column_with_strings, val);
```
alexowens90: This won't be correct for fixed-width string types. PCRE supports UTF-32:
https://www.pcre.org/original/doc/html/pcreunicode.html
so we can just use that natively

cpp/arcticdb/processing/operation_dispatch_binary.cpp
```diff
@@ -159,6 +159,8 @@ VariantData dispatch_binary(const VariantData& left, const VariantData& right, O
             return visit_binary_membership(left, right, IsInOperator{});
         case OperationType::ISNOTIN:
             return visit_binary_membership(left, right, IsNotInOperator{});
+        case OperationType::REGEX_MATCH:
+            return visit_regex_match_membership(left, right);
```
alexowens90: We discussed implementing this as a binary comparator, and adding a `Regex` type to `VariantData`. This will also make it simple to only compile the regex once (or I guess twice, once for UTF-8 and once for UTF-32)

python/tests/unit/arcticdb/version_store/test_filtering.py
alexowens90, line 1: Can we add a test for finding members of a comma-separated list of strings? We have a customer use case for this. e.g.
`df = pd.DataFrame({"col": ["this,is,a,comma,separated,list,of,strings", ...]})`
and matching on both a single element being in the lists, and on multiple elements being present

alexowens90, line 1: We should run all of these tests with both fixed-width and dynamic string columns once the implementation is fixed to handle UTF-32

python/arcticdb/version_store/processing.py
```diff
@@ -232,6 +233,15 @@ def notna(self):
 
     def notnull(self):
         return ExpressionNode.compose(self, _OperationType.NOTNULL, None)
+    
+    def regex_match(self, pattern: str):
+        if isinstance(pattern, str):
+            _RegexPattern(pattern)  # Validate the regex pattern
+            return self._apply(pattern, _OperationType.REGEX_MATCH)
```
IvoDD: I think it would be good if we `self.apply(pattern, REGEX_MATCH)` can have the pattern be only of type `RegexPattern`.

This way we will only construct the regex pattern once in python here and use it throughout.

Related to [this comment](https://github.com/man-group/ArcticDB/pull/2385/files#r2132204564) from Alex

python/benchmarks/local_query_builder.py
```diff
@@ -70,6 +70,15 @@ def peakmem_filtering_string_isin(self, num_rows):
         q = q[q["id1"].isin(string_set)]
         self.lib.read(f"{num_rows}_rows", columns=["v3"], query_builder=q)
 
+    def time_filtering_string_regex_match(self, num_rows):
+        # Selects about 1% of the rows
+        k = min(3, num_rows // 1000)
```
IvoDD: The choice of k is weird, 3 will always be `< num_rows//1000` (smallest num_rows is 1_000_000). Maybe you meant:
`min(3, int(math.log10(num_rows)) - 3)` or something similar?

Also the comment on 1% of rows is misleading. It will filter 3 digit numbers which are 1% for 1_000_000 case but .1% for 10_000_000 case. (Will be fixed if we apply the log10 change I think)

cpp/arcticdb/processing/operation_dispatch_binary.cpp
IvoDD: I aggree with this. 

Just need to be careful if we're passing `RegexPattern`s around as they [only hold a reference to the string](https://github.com/man-group/ArcticDB/blob/c46b93e0337d10c825f0f88549feef9316c45019/cpp/arcticdb/util/regex_filter.hpp#L29) and that's a recipe for holding a reference to a destructed object.

python/arcticdb/version_store/processing.py
phoebusm: Yes definitely. I have missed the fact that the operation is made per segment. As that is the case, putting the compiled regex pattern in the tree is a logical choice

cpp/arcticdb/processing/operation_dispatch_binary.cpp
phoebusm: Yes I still remember the discussion. I agree with putting `RegexPattern` in to `VariantData` to avoid multiple compilation. For putting it in `binary_membership`, I just thinking maybe making the `if` cases in `binary_membership` more complicated may not be the best idea? The datatype checks in regex match are quite different from binary comparing

python/tests/unit/arcticdb/version_store/test_filtering.py
phoebusm: Removed

python/benchmarks/local_query_builder.py
phoebusm: https://github.com/man-group/ArcticDB/blob/14e0fcc7270c5ba793510ff704e91073aeea0be0/python/benchmarks/common.py#L67
`dt["id1"] = np.random.choice([f"id{str(i).zfill(3)}" for i in range(1, k + 1)], n)`

Assume there is `n` rows, there will be `n/10` patterns.
Let's make `k = n/10`
If `k < 100`, `0` will be padded at the beginning to create the pattern, e.g. `id010`.
Therefore `num_rows // 1000` can selecting 1% of the rows

`3` is a typo

python/arcticdb/version_store/processing.py
```diff
@@ -426,10 +436,11 @@ class QueryBuilder:
 
     * Binary arithmetic: +, -, *, /
     * Unary arithmetic: -, abs
+    * regex match: regex_match
```
alexowens90: Not an arithmetic operation

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1319,3 +1320,149 @@ def test_filter_unsupported_boolean_operators():
...
+    pattern_a = "^abc"
+    q_a = QueryBuilder()
+    q_a = q_a[q_a["a"].regex_match(pattern_a)]
+    expected = df[df.a.str.contains(pattern_a, na=False)]
+    assert_frame_equal(lib.read(sym, query_builder=q_a).data, expected)
+
+    pattern_c = r"\d\d[a-zA-Z]"
+    q_c = QueryBuilder()
+    q_c = q_c[q_c["c"].regex_match(pattern_c)]
+    expected = df[df.c.astype(str).str.contains(pattern_c, na=False)]
```
alexowens90: `astype` needed?

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1319,3 +1320,149 @@ def test_filter_unsupported_boolean_operators():
...
+    q = QueryBuilder()
+    q = q[q["a"].regex_match(pattern_a) & q["c"].regex_match(pattern_c)]
+    assert lib.read(sym, query_builder=q).data.empty
+
+    q2 = QueryBuilder()
+    q2 = q2[q2["a"].regex_match(pattern_a) & q2["b"].isin([0])]
+    assert lib.read(sym, query_builder=q2).data.empty
+    
+
+def test_filter_regex_match_nans_nones(lmdb_version_store_v1, sym):
```
alexowens90, line 1399: Might as well parametrize this over `dynamic_strings` as well

cpp/arcticdb/processing/expression_context.hpp
```diff
@@ -55,9 +55,14 @@ struct ExpressionContext {
...
     }
 
+    void add_regex(const std::string& name, std::shared_ptr<util::RegexGeneric> regex) {
+        regex_sets_.set_value(name, std::move(regex));
+    }
+
     ConstantMap<ExpressionNode> expression_nodes_;
     ConstantMap<Value> values_;
     ConstantMap<ValueSet> value_sets_;
+    ConstantMap<util::RegexGeneric> regex_sets_;
```
alexowens90: `regex_matches_` would be a better name

cpp/arcticdb/column_store/string_pool.hpp
```diff
@@ -174,8 +176,9 @@ class StringPool {
 
     py::buffer_info as_buffer_info() const;
 
-    std::optional<position_t> get_offset_for_column(std::string_view str, const Column& column);
-    ankerl::unordered_dense::set<position_t> get_offsets_for_column(const std::shared_ptr<std::unordered_set<std::string>>& strings, const Column& column);
+    std::optional<position_t> get_offset_for_column(std::string_view str, const Column& column) const;
+    ankerl::unordered_dense::set<position_t> get_offsets_for_column(const std::shared_ptr<std::unordered_set<std::string>>& strings, const Column& column) const;
+    ankerl::unordered_dense::set<position_t> get_regex_match_offsets_for_column(const std::shared_ptr<util::RegexGeneric>& regex_generic, const Column& column, bool fix_string_type) const;
```
alexowens90: The `bool` argument is unnecessary, the `Column` contains it's own type

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -280,32 +280,66 @@ VariantData binary_comparator(const ColumnWithStrings& column_with_strings, cons
...
+            return transform_to_placeholder(result);
         },
-        [&] ([[maybe_unused]] const std::shared_ptr<Value>& l, [[maybe_unused]] const std::shared_ptr<Value>& r) ->VariantData  {
-        user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>("Two value inputs not accepted to binary comparators");
-        return EmptyResult{};
+        [&func](const ColumnWithStrings& l, const std::shared_ptr<util::RegexGeneric>& r) ->VariantData {
+            auto result =  binary_comparator<decltype(func)>(l, r, std::forward<decltype(func)>(func));
+            return transform_to_placeholder(result);
+        },
+        [&func] ([[maybe_unused]] const std::shared_ptr<Value>& l, [[maybe_unused]] const std::shared_ptr<Value>& r) ->VariantData  {
```
alexowens90: `func` captured but not used

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -280,32 +280,66 @@ VariantData binary_comparator(const ColumnWithStrings& column_with_strings, cons
...
+            Column::transform<typename col_type_info::TDT>(
+                    *column_with_strings.column_,
+                    output_bitset,
+                    false,
+                    [&offset_set, &func](auto input_value) {
+                auto offset = static_cast<entity::position_t>(input_value);
+                return func(offset, offset_set);
+            });
+        } else {
+            user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>("Invalid comparison {} ({}) {} RegexGeneric",
```
alexowens90: We can be a bit more explicit here, something like:
"Cannot perform regex_match with pattern {regex_pattern} on column {col name} as it has non-string type {type}"

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -280,32 +280,66 @@ VariantData binary_comparator(const ColumnWithStrings& column_with_strings, cons
     return VariantData{std::move(output_bitset)};
 }
 
+template <typename Func>
+VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const std::shared_ptr<util::RegexGeneric>& regex_generic, Func&& func) {
```
alexowens90:
```suggestion
VariantData binary_comparator(const ColumnWithStrings& column_with_strings, const util::RegexGeneric& regex_generic, Func&& func) {
```

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -280,32 +280,66 @@ VariantData binary_comparator(const ColumnWithStrings& column_with_strings, cons
...
+    if (is_empty_type(column_with_strings.column_->type().data_type())) {
+        return EmptyResult{};
+    }
+    internal::check<ErrorCode::E_ASSERTION_FAILURE>(std::is_same_v<std::remove_reference_t<Func>, RegexMatchOperator>, "Invalid operator {} for regex match", func);
```
alexowens90:
```suggestion
    if constexpr(std::is_same_v<std::remove_reference_t<Func>, RegexMatchOperator>) {
```
with an `else {internal::raise<ErrorCode::E_ASSERTION_FAILURE>(...)}` will mean we can remove the new throws in the `LessThanOperator` etc

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -393,6 +396,9 @@ bool operator()(uint64_t t, int64_t u) const {
 bool operator()(int64_t t, uint64_t u) const {
     return comparison::equals(t, u);
 }
+bool operator()(entity::position_t, ankerl::unordered_dense::set<position_t>) const {
```
alexowens90: Remove all of these implementations (see comment in `operation_dispatch_binary.hpp`)

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -280,32 +280,66 @@ VariantData binary_comparator(const ColumnWithStrings& column_with_strings, cons
...
+        using col_type_info = ScalarTypeInfo<decltype(col_tag)>;
+        if constexpr(is_sequence_type(col_type_info::data_type)) {
+            constexpr bool fix_string_type = is_fixed_string_type(col_type_info::data_type);
+            auto offset_set = column_with_strings.string_pool_->get_regex_match_offsets_for_column(regex_generic, *column_with_strings.column_, fix_string_type);
+            Column::transform<typename col_type_info::TDT>(
+                    *column_with_strings.column_,
+                    output_bitset,
+                    false,
+                    [&offset_set, &func](auto input_value) {
+                auto offset = static_cast<entity::position_t>(input_value);
```
alexowens90: I think `input_value` will always be a `position_t`?

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -513,6 +534,16 @@ bool operator()(int64_t t, uint64_t u) const {
 }
 };
 
+struct RegexMatchOperator {
+template<typename T, typename U>
+bool operator()(T, U) const {
+    util::raise_rte("RegexMatchOperator does not support {} and {}", typeid(T).name(), typeid(U).name());
+}
+bool operator()(entity::position_t offset, ankerl::unordered_dense::set<position_t> offset_set) const {
```
alexowens90:
```suggestion
bool operator()(entity::position_t offset, const ankerl::unordered_dense::set<position_t>& offset_set) const {
```

cpp/arcticdb/column_store/string_pool.cpp
```diff
@@ -51,6 +52,11 @@ std::string_view StringBlock::const_at(position_t pos) const {
     return {head->data(), head->size()};
 }
 
+std::u32string_view StringBlock::u32_const_at(position_t pos) const {
```
alexowens90: I'm not a fan of this addition. The `string_view` returned by `const_at` is just a buffer and a size, whether it is interpreted as UTF-8/32 is an external concern to the string pool

cpp/arcticdb/column_store/string_pool.cpp
```diff
@@ -240,4 +242,24 @@ ankerl::unordered_dense::set<position_t> StringPool::get_offsets_for_column(cons
...
     return output;
 }
+
+ankerl::unordered_dense::set<position_t> StringPool::get_regex_match_offsets_for_column(const std::shared_ptr<util::RegexGeneric>& regex_generic, const Column& column, bool fix_string_type) const {
+    auto unique_values = unique_values_for_string_column(column);
+    remove_nones_and_nans(unique_values);
+
+    ankerl::unordered_dense::set<position_t> output;
+    for(auto pos : unique_values) {
+        if (fix_string_type) {
```
alexowens90: The compiler will probably optimise this for you, but better to have the if statement outside the loop

cpp/arcticdb/util/regex_filter.hpp
```diff
@@ -61,38 +85,90 @@ class RegexPattern : private PcreRegex {
     ARCTICDB_NO_MOVE_OR_COPY(RegexPattern)
 private:
     void compile_regex() {
-        handle_ = ::pcre_compile2(text_.data(), options_, &error_, &help_, &offset_, table_);
-        util::check(handle_ != nullptr, "Error {} compiling regex {}: {}", error_, text_, help_);
+        handle_ = this->pcre_compile2_(reinterpret_cast<typename PcreRegexEncode::StringCastingType>(text_.data()), options_, &error_, &help_, &offset_, table_);
+        // Should be safe to assume regex pattern must be user input
+        user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
```
alexowens90: This is an API breaking change, since this code is accessible from `list_symbols`. Let's leave as `util::check` and add a ticket with release `6.0.0` to change it later

cpp/arcticdb/util/regex_filter.hpp
```diff
@@ -61,38 +85,90 @@ class RegexPattern : private PcreRegex {
...
+using RegexPatternUTF32 = RegexPattern<PcreRegexUTF32>;
+
+class RegexGeneric {
+private:
+    RegexPatternUTF8 pattern_utf8_;
+    RegexPatternUTF32 pattern_utf32_;
+    RegexUTF8 regex_utf8_;
+    RegexUTF32 regex_utf32_;
+public:
+    RegexGeneric(const std::string& pattern) :
```
alexowens90, line 178: Agree with Ivo that there are a lot of references relying on this string being kept alive elsewhere, let's just take a copy and keep it alive ourselves to keep it simple

cpp/arcticdb/util/regex_filter.hpp
```diff
@@ -13,31 +13,55 @@
 
 namespace arcticdb::util {
 
-class PcreRegex {
+template<typename T>
+[[nodiscard]] auto convert_to_utf8_if_needed(const T& input) {
+    if constexpr (std::is_same_v<T, std::string> || std::is_same_v<T, std::string_view>) {
+        return input; // Already UTF-8
+    } else if constexpr (std::is_same_v<T, std::u32string_view>) {
+        return std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t>().to_bytes(std::u32string(input));
```
alexowens90: I believe this is deprecated?

cpp/arcticdb/util/regex_filter.hpp
```diff
@@ -61,38 +85,90 @@ class RegexPattern : private PcreRegex {
...
+class RegexGeneric {
+private:
+    RegexPatternUTF8 pattern_utf8_;
+    RegexPatternUTF32 pattern_utf32_;
+    RegexUTF8 regex_utf8_;
+    RegexUTF32 regex_utf32_;
+public:
+    RegexGeneric(const std::string& pattern) :
+        pattern_utf8_(pattern),
+        pattern_utf32_(std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t>().from_bytes(pattern)),
```
alexowens90: Deprecated, just store the pattern as a `std::string_view`, since all PCRE cares about is a pointer and a size anyway

alexowens90:
Would be nice to have some `benchmark` microbenchmarks, both for iterating on the performance, and for tighter performance regression control when we add the C++ benchmarks into the ASV results

cpp/arcticdb/util/regex_filter.hpp
phoebusm: The pcre compile function doesn't have size parameter

cpp/arcticdb/util/regex_filter.hpp
phoebusm: `class RegexPattern` already takes a copy of the pattern. `class RegexGeneric` just passes it over

cpp/arcticdb/processing/operation_dispatch_binary.hpp
phoebusm: It's `unsigned long`

python/tests/unit/arcticdb/version_store/test_filtering.py
phoebusm: I don't think `dynamic_strings=False` can be used with that particular dataframe

python/arcticdb/version_store/processing.py
phoebusm: 🤦Totally miss the category

python/tests/unit/arcticdb/version_store/test_filtering.py
alexowens90: Yeh you're right, ignore me

cpp/arcticdb/processing/test/benchmark_common.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return res;
+}
+
+ColumnWithStrings generate_string_dense_column(const size_t num_rows, const size_t unique_strings, DataType dt) {
+    auto string_pool = std::make_shared<StringPool>();
+    std::vector<entity::position_t> offsets;
+    offsets.reserve(unique_strings);
+    for (size_t _ = 0; _ < unique_strings; _++) {
+        auto str = generate_string();
```
alexowens90, line 33: I don't think this is doing quite the right thing with fixed width strings? Need to use `boost` to convert the ASCII into UTF32

phoebusm: Good point. Will amend

alexowens90 (approved):

python/benchmarks/local_query_builder.py
phoebusm: This part is a bit cursed. I'll simplify the pattern

IvoDD (approved):

cpp/arcticdb/util/regex_filter.hpp
```diff
@@ -27,46 +28,56 @@ template<typename T>
...
 };
 
 template<typename PcreRegexEncode>
 class RegexPattern : protected PcreRegexEncode {
     typename PcreRegexEncode::StringType text_;
     typename PcreRegexEncode::HandleType handle_ = nullptr;
-    typename PcreRegexEncode::ExtraPtr extra_ = nullptr;
     const char* help_  = "";
     int offset_ = 0;
-    const unsigned char * table_ = nullptr;
```
IvoDD: Looks like we don't use the the `help_` and `offset_` anymore?

Also the `error_` code is only used to signal an error during compilation. What do you think about moving it as a local variable similar to `erroroffset`?

cpp/arcticdb/util/regex_filter.hpp
```diff
@@ -86,13 +97,22 @@ class RegexPattern : protected PcreRegexEncode {
...
+            PCRE2_ZERO_TERMINATED,
+            options_,
+            &error_,
+            &erroroffset,
+            nullptr
+        );
         util::check(
             handle_ != nullptr, 
-            "Error {} compiling regex {}: {}", 
+            "Error {} compiling regex {}: {} at position {}",
```
IvoDD: `help_` will be an empty string. It would be useful if we can extract some error message from pcre_compile. If not, let's remove it. If yes, let's keep it as a local variable.

IvoDD (approved):

PR #2387, opened 2025-06-05
Rework C++ truncation for arrow

IvoDD:
Monday ref: 9320469407

#### C++ truncation (not yet user facing):
Arrow requires truncation to happen in C++ layer.
Previous logic had many issues:
- Wrong truncated range calculation
- Wrong offsets used when first row offset != 0
- Wrong row range calculation
- Does not construct empty arrow table correctly

This PR fixes the issues by rewriting a lot of the column truncation logic and adds more tests.

#### Minor performance improvments:

- improves `Column::search_sorted` to be trully `O(logN)` and adds C++ benchmarks for it.
- Includes optimizations to load fewer table data keys when reading with `date_range` and when doing an `update`. It does so by fixing more off by one errors and clarifying logic. 
Adds thorough testing with `query_stats` to verify that indeed we're not doing unneeded reads.

#### Notes to reviewers
C++ truncation logic has changed a lot, so I suggest looking at the end state (not the diff) for files related to truncation:
- `chunked_buffer.hpp` truncation functions
- `read_frame.cpp` `handle_truncation` and `get_truncate_range*` functions
- Tests are in `test_arrow.py`

The optimizations to load only the data keys we require can be considered independently from the C++ truncation logic. 
They are in interleaved in the same PR as they are required for C++ truncation to work and Alex discovered we are loading unneeded chunks.
These changes mostly reside in:
- `query.hpp` and `query.cpp`
- `index_range.hpp`
- Tests are in `test_num_storage_oprations.py` which contains a few tests moved from `test_s3`

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


alexowens90 (approved):

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -14,16 +14,55 @@
...
+    auto res = type.visit_tag([](auto &&impl) {
+        using TagType = std::decay_t<decltype(impl)>;
+        using DataTagType = typename TagType::DataTypeTag;
+        using RawType = typename DataTagType::raw_type;
+        std::optional<sparrow::validity_bitmap> validity_bitmap;
+        if constexpr (is_sequence_type(TagType::DataTypeTag::data_type)) {
+            // TODO: This is super hacky. Our string column return type is dictionary encoded, and this should be
+            //  consistent when there are zero rows. But sparrow (or possibly Arrow) requires at least one key-value
+            //  pair in the dictionary, even if there are zero rows
+            auto offset_ptr = new int64_t[2];
```
vasil-pashov: Why are there two offsets?

nit: I'd use unique_ptr for this and string_ptr and then call unique_ptr::release() to pass them to sparrow.

cpp/arcticdb/column_store/column.hpp
```diff
@@ -728,59 +728,57 @@ class Column {
     // Only works if column is of numeric type and is monotonically increasing
     // Returns the index such that if val were inserted before that index, the order would be preserved
     // By default returns the lowest index satisfying this property. If from_right=true, returns the highest such index
+    // from (inclusive) and to (exclusive) can optionally be provided to search a subset of the rows in the column
     template<class T, std::enable_if_t<std::is_integral_v<T> || std::is_floating_point_v<T>, int> = 0>
```
vasil-pashov: nit: replace the enable_if with requires clause

cpp/arcticdb/arrow/arrow_utils.cpp
IvoDD: In short `len(offsets)` must always be `len(elements)+1` according to arrow spec.

I've updated to use `unique_ptr`s

In more detail:

String arrays in arrow are represented as an offset buffer and a string buffer. The string buffer is just a concatenation of all strings. So for example if you want to represent:
```
["a", "bc", "def"]
```
The string buffer will contain:
```
str_buffer = "abcdef"
```
Then the offset buffer just contains pointers to the beggining of each element. And to read the `i`-th element you just read everything between `i` and `i+1` pointers. This however requires storing one extra offset:
```
offset_buffer = [0, 1, 3, 6]
```

There is one extra layer of complexity of our string representation and that is the dict encoding explained in more detail [here](https://github.com/man-group/ArcticDB/pull/1831#discussion_r2140555459)

PR #2390, opened 2025-06-06
Fixes to run tests at Man

grusev:

1. A test now accepts several possible errors as different S3 storages might have different errors.
2 A new test is failing because on pandas 1.X which is at pegagus>next the x column will be silently converted to float64

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/integration/arcticdb/version_store/test_deletion.py
+++ b/python/tests/integration/arcticdb/version_store/test_deletion.py
@@ -331,6 +331,8 @@ def test_delete_versions_with_update(object_version_store, versions_to_delete, s
             expected = expected.combine_first(d)
             if not expected.empty:
                 expected.loc[d.index] = d
+        # pandas 1.0 patch, otherwise the col will be float64
+        expected['x'] = expected['x'].astype('int64')
         return expected
 
     # Write initial data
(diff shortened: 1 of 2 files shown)


poodlewars:
Shouldn't these tests run with Pandas 1 in our Github CI, using the compat requirements? https://github.com/man-group/ArcticDB/blob/fc9514f25712d8e86fbdbd2f7e37e64f3a10df40/build_tooling/requirements-compatibility-py38.txt#L4

python/tests/integration/arcticdb/version_store/test_deletion.py
```diff
@@ -331,6 +331,8 @@ def build_expected_df(dataframes):
             expected = expected.combine_first(d)
             if not expected.empty:
                 expected.loc[d.index] = d
+        # pandas 1.0 patch, otherwise the col will be float64
```
poodlewars, line 334: We've seen that this doesn't always happen with Pandas 1.0. It would be much better to understand why we need this change properly before merging this

grusev: At Man's internal we are testing with : pandas 1.0.5
At Github we do with 1.3.5

The problem is in between those pandas versions

We should merge this change, as there are more surprises probably waiting. Let's finish first the whole VAST project and see what is there still to investigate afterwards

poodlewars (approved):

PR #2392, opened 2025-06-06
9220057136: Allow block manager consolidation in Pandas < 2

alexowens90:
#### Reference Issues/PRs
[9220057136](https://man312219.monday.com/boards/7852509418/pulses/9220057136)

Pandas < 2 has a bug for which some operations do not work or produce results with the wrong dtype if the block manager has not been allowed to consolidate. This change makes it so that dataframe consolidation is only ever skipped in Pandas 2.X

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -738,6 +738,13 @@ class DataFrameNormalizer(_PandasNormalizer):
         super(DataFrameNormalizer, self).__init__(*args, **kwargs)
         self._skip_df_consolidation = IS_PANDAS_TWO and os.getenv("SKIP_DF_CONSOLIDATION") is not None
 
+    def set_skip_df_consolidation(self):
+        if IS_PANDAS_TWO:
+            self._skip_df_consolidation = True
+        else:
+            self._skip_df_consolidation = False
+
+
     def df_without_consolidation(self, columns, index, item, n_indexes, data):
         """
         This is a hack that allows us to monkey-patch the DataFrame Block Manager so it doesn't do any
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -736,7 +736,7 @@ class DataFrameNormalizer(_PandasNormalizer):
 
     def __init__(self, *args, **kwargs):
         super(DataFrameNormalizer, self).__init__(*args, **kwargs)
-        self._skip_df_consolidation = os.getenv("SKIP_DF_CONSOLIDATION") is not None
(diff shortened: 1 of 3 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -736,7 +736,14 @@ class DataFrameNormalizer(_PandasNormalizer):
 
     def __init__(self, *args, **kwargs):
         super(DataFrameNormalizer, self).__init__(*args, **kwargs)
-        self._skip_df_consolidation = os.getenv("SKIP_DF_CONSOLIDATION") is not None
+        self._skip_df_consolidation = IS_PANDAS_TWO and os.getenv("SKIP_DF_CONSOLIDATION") is not None
+
+    def set_skip_df_consolidation(self):
+        if IS_PANDAS_TWO:
```
G-D-Petrov, line 742: shouldn't this use the same check as above: `IS_PANDAS_TWO and os.getenv("SKIP_DF_CONSOLIDATION") is not None`

poodlewars (approved):

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -1060,3 +1062,19 @@ def test_required_field_inclusion(version_store_factory, dynamic_schema, segment
...
+    lib = version_store_factory()
+    assert lib._normalizer.df._skip_df_consolidation == (env_var_set and IS_PANDAS_TWO)
+
+
+@pytest.mark.parametrize("env_var_set", [True, False])
+def test_pandas_consolidation_v2(lmdb_library_factory, monkeypatch, env_var_set):
+    if env_var_set:
+        monkeypatch.setenv("SKIP_DF_CONSOLIDATION", "true")
+    lib = lmdb_library_factory()
+    assert lib._nvs._normalizer.df._skip_df_consolidation == IS_PANDAS_TWO
```
G-D-Petrov, line 1088: I think that there should also be tests that read out a df from ArcticDB and assert that the correct consolidation is used (e.g. `arcticdb.version_store._normalization.BlockManagerUnconsolidated` vs `pandas.core.internals.managers.BlockManager`)

python/arcticdb/version_store/_normalization.py
alexowens90: No, the env var has never been respected when constructing `Library` objects (which is where this is used)

python/tests/unit/arcticdb/version_store/test_normalization.py
alexowens90: Fair, added:
https://github.com/man-group/ArcticDB/pull/2392/commits/9b47bd5826b906226d3aa49f0ec4bd94b9f99878

PR #2395, opened 2025-06-09
Reimplement version map reload delete fix

G-D-Petrov:
#### Reference Issues/PRs
Monday ticket ref: 9383033019

- Reimplement the deletion fix that was reverted [here](https://github.com/man-group/ArcticDB/commit/01f6a6a7c6aed669267892d082afbe744fd90cdc)
- Adds a fix for the issue with `delete_snapshot` deleting data keys that shouldn't have been deleted
- Adds tests for the above
- Extends the CI to run with and without the version cache
- Fixes the tasks that are used in stress tests to pass the correct kwarg for prune previous

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/tests/integration/arcticdb/test_s3.py
```diff
@@ -222,3 +229,161 @@ def test_library_get_key_path(lib_name, storage_bucket, test_prefix):
...
+
+            lib.delete("s")
+            stats = qs.get_query_stats()
+            qs.reset_stats()
+            assert sum_all_operations(stats) == base_ops_count == (expected_ops + 2), visualize_stats_diff(
+                base_stats, stats
+            )
+
+
+def test_wrute_and_prune_previous_over_time(lib_name, storage_bucket, clear_query_stats):
```
alexowens90: typo

.github/workflows/build_steps.yml
```diff
@@ -265,17 +265,19 @@ jobs:
       matrix:
         type: ${{fromJSON(vars.TEST_DIRS_OVERRIDE || needs.compile.outputs.test_dirs)}}
         python_deps_id: ${{fromJson(inputs.python_deps_ids)}}
+        # Testing with 0(no version cache) and 2000000000(2s - default timeout) to see if the tests pass
+        version_cache_timeout: ["0", "2000000000"]
```
alexowens90: Looks from the checks that this doubling our number of tests? Given where all the version map caching logic lives, this seems a bit excessive, running all tests for one Python version/OS should build the same confidence?
I'm flagging because as I understand it, github throttles us a bit on our simultaneous runner count, so this could have a large impact on wall clock CI runtimes

IvoDD: What do you think about having the options named `No-Cache` and `With-Cache` or something similar?

Having the runner name `linux-3.11-2000000000` is a bit confusing.

Also what do you think about if `No-Cache` setting the env var to `0` but if `With-Cache` to not set the env variable at all? This way if we ever change the default we won't need to change it in the workflow as well.

python/tests/unit/arcticdb/version_store/test_deletion.py
```diff
@@ -177,6 +183,140 @@ def test_tombstones_deleted_data_keys_prune(lmdb_version_store_prune_previous, s
...
+    final_df = pd.concat([df, df3])
+    assert_frame_equal(lib.read(sym).data, final_df)
+
+    pre_delete_snapshot = lib.read(sym).data
+    delete_snapshot(lib, "snap")
+    post_delete_snapshot = lib.read(sym).data
+    assert_frame_equal(pre_delete_snapshot, post_delete_snapshot)
+    assert_frame_equal(pre_delete_snapshot, lib.read(sym, as_of=2).data)
+    assert_frame_equal(pre_delete_snapshot, final_df)
+    assert_frame_equal(df, lib.read(sym, as_of=1).data)
```
IvoDD: I'm confused. Shouldn't v1 be deleted by the `prune_previous` from the append?
I think it might be because we're using `prune_previous_versions` instead of `prune_previos_version`

.github/workflows/build_steps.yml
G-D-Petrov: Changed it to run only on Python311 for both Win/Linux by default and added the option to manually run it for all versions/OS

G-D-Petrov: Makes sense, implemented as suggested

python/tests/unit/arcticdb/version_store/test_deletion.py
G-D-Petrov: Yep, fixed

IvoDD (approved):

alexowens90 (approved):

PR #2400, opened 2025-06-11
Drop support for Python 3.7 and fix CI

IvoDD:
Drops support for end of life Python 3.7. And removes codepaths checking for `python < 3.8`

Also tweaks clang docker image to fix CI issue (where cmake was finding wrong python version).
Also had to install cmake3 because now by default the manylinux_2_28 images have cmake4.

#### Any other comments?
Link to a [successful build](https://github.com/man-group/ArcticDB/actions/runs/15613614393) using this branch's workflow and `arcticdb-dev-clang:fixed-path` docker image which I built locally.

And a link to [successful build with analysis tools](https://github.com/man-group/ArcticDB/actions/runs/15613127117/). Fails asv checks due to network error but benchmarking works fine

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -168,7 +168,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[8, 9, 10, 11, 12, 13]')}}
         arcticdb_version: ["oldest", "latest"]
         include:
           - python_deps_ids: [""]
@@ -253,7 +253,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[7, 8, 9, 10, 11, 12, 13]')}}
+        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[8, 9, 10, 11, 12, 13]')}}
     name: 3.${{matrix.python3}} Windows
     uses: ./.github/workflows/build_steps.yml
     secrets: inherit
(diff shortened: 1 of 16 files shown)


.github/workflows/build.yml
```diff
@@ -277,7 +277,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[8, 9, 10, 11, 12, 13]')}}
```
G-D-Petrov: I think that we shouldn't add version 12 and 13 as these tests run with older versions that we haven't published wheels for

.github/workflows/build.yml
```diff
@@ -299,7 +299,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[7, 8, 9, 10, 11]')}}
+        python3: ${{fromJson(vars.WINDOWS_PYTHON_VERSIONS || '[8, 9, 10, 11, 12, 13]')}}
```
G-D-Petrov: I think that we shouldn't add version 12 and 13 as these tests run with older versions that we haven't published wheels for

alexowens90 (approved):

.github/workflows/build.yml
```diff
@@ -145,7 +145,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python3: ${{fromJson(vars.LINUX_PYTHON_VERSIONS || '[6, 7, 8, 9, 10, 11]')}}
+        # This uses old arcticdb versions which won't work on python < 3.12
```
poodlewars, line 148: I don't understand this comment, don't you mean won't work on Python >= 3.12 given that that's the version you exclude below?

python/tests/util/mark.py
```diff
@@ -115,7 +115,7 @@
 FORK_SUPPORTED = pytest.mark.skipif(WINDOWS, reason="Fork not supported on Windows")
 
 ## MEMRAY supports linux and macos and python 3.8 and above
```
poodlewars, line 117: Drop the comments and strings like "python 3.8 and above"

PR #2403, opened 2025-06-12
Remove arcticdb package dependencies on tests and do checks during build

grusev:

Log: https://github.com/man-group/ArcticDB/actions/runs/15602928320/job/43946709069
A dependency on tests package from a file in tests package which is not packaged in wheel is causing unpredictable failures down the line

We have code here and there that will depend on that being able to execute from the wheel. And as in the wheel we do not have tests package there cannot be any imports from there. Perhaps we all know that including me but at the time I created original PR it was not on my mind, which proves we have to do checks for such st..id small things

What is in the PR:
- no more code in python/arcticdb  that depends on tests
- shell script to do the check at python/arcticdb
- customized workflow - just one the conda linux that has a step invoking this script

With that fix and this check all future problems will be prevented as they will become visible.

AWS S3 Log which this should fix: https://github.com/man-group/ArcticDB/actions/runs/15612826324
Changed linux conda build workflow: https://github.com/man-group/ArcticDB/actions/runs/15610614742/job/43970382577
```
Run build_tooling/checks.sh          
SUCCESS: No forbidden imports found.
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/build_tooling/checks.sh
+++ b/build_tooling/checks.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+pushd "$(pwd)" >/dev/null
+
+tooling_dir="$(dirname $BASH_SOURCE)"
+cd $tooling_dir/../python
+
+FORBIDDEN_CONTENT=("import\s+test" "from\s+test")
+SEARCH_DIR="arcticdb/"
+ERROR_FOUND=0
+
+for CONTENT in "${FORBIDDEN_CONTENT[@]}"; do
+    MATCHES=$(grep -rP "$CONTENT" "$SEARCH_DIR") 
+    if [ -n "$MATCHES" ]; then
+        echo "ERROR: Forbidden package '$CONTENT' is imported!"
+        echo "$MATCHES"  
+        ERROR_FOUND=1
+    fi
+done
+
+popd >/dev/null
+
+if [ "$ERROR_FOUND" -eq 1 ]; then
+    echo "ERRORS DETECTED!"
+    exit 1
+else
+    echo "SUCCESS: No forbidden imports found."
+fi
--- a/build_tooling/checks.sh
+++ b/build_tooling/checks.sh
@@ -5,12 +5,13 @@ pushd "$(pwd)" >/dev/null
 tooling_dir="$(dirname $BASH_SOURCE)"
 cd $tooling_dir/../python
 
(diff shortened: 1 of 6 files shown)


build_tooling/checks.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+pushd "$(pwd)" >/dev/null
+
+tooling_dir="$(dirname $BASH_SOURCE)"
+cd $tooling_dir/../python
+
+FORBIDDEN_CONTENT=("import\s+test" "from\s+test")
```
alexowens90: This isn't robust to multiple whitespace characters or newlines in import statements

build_tooling/checks.sh
```diff
@@ -0,0 +1,28 @@
...
+SEARCH_DIR="arcticdb/"
+ERROR_FOUND=0
+
+for CONTENT in "${FORBIDDEN_CONTENT[@]}"; do
+    MATCHES=$(grep -rP "$CONTENT" "$SEARCH_DIR")
```
alexowens90: Can use `-e` to match both patterns at once

build_tooling/checks.sh
grusev: Better version submited

build_tooling/checks.sh
grusev: Cannot do that actually. -e is not supported alongside -P.

build_tooling/checks.sh
grusev: Now the wrong output will be like:

```
ERROR: Forbidden package '\s*import(\s+\\*\n*)+test.*' is imported!
arcticdb/util/utils.py
arcticdb/util/marks.py
arcticdb/util/venv.py
ERROR: Forbidden package '\s*from(\s+\\*\n*)+tests.*' is imported!
arcticdb/util/utils.py
arcticdb/util/marks.py
arcticdb/util/venv.py
ERRORS DETECTED!
```

Tested with:

```
import \
  \
     tests
from \
  tests import sd
  import          tests.fkgjf
from tests import fdsfdsfds

  import \
   \
   tests

    from tests import asrt
```

G-D-Petrov (approved):

alexowens90 (approved):

PR #2404, opened 2025-06-12
Bugfix/9256801170/memory over allocation when reading shorts dfs

alexowens90:
#### Reference Issues/PRs
[9256801170](https://man312219.monday.com/boards/7852509418/pulses/9256801170)

When presizing a `ChunkedBuffer`, the behaviour was to round this up to the default block size (3968 bytes) if fewer bytes were requested, resulting in a major over-allocation if the buffer would only ever hold a few values.
Inspecting the code makes it clear that this method is only ever used with buffers that will never subsequently be expanded, and so there is no benefit to this overallocation, and we should always create a block that has exactly the correct buffer size.

The test output from the newly added tests (these will be backed out prior to merging):
```
	test_read_short_sym
		pre-fix
			RSS: 849.41 MB
			RSS: 1370.93 MB
			RSS: 2002.95 MB
			RSS: 2525.48 MB
			RSS: 3158.48 MB
			RSS: 3736.20 MB
			RSS: 4258.29 MB
			RSS: 4834.31 MB
			RSS: 5468.00 MB
			RSS: 6100.84 MB
		post-fix
			RSS: 457.21 MB
			RSS: 586.88 MB
			RSS: 826.82 MB
			RSS: 956.86 MB
			RSS: 1197.72 MB
			RSS: 1383.61 MB
			RSS: 1513.94 MB
			RSS: 1697.97 MB
			RSS: 1939.73 MB
			RSS: 2180.68 MB
	test_read_long_sym_date_range
		pre-fix
			RSS: 1105.40 MB
			RSS: 1675.86 MB
			RSS: 2246.67 MB
			RSS: 2821.30 MB
			RSS: 3391.05 MB
			RSS: 3967.81 MB
			RSS: 4542.89 MB
			RSS: 5109.78 MB
			RSS: 5685.29 MB
			RSS: 6254.56 MB
		post-fix
			RSS: 714.96 MB
			RSS: 895.02 MB
			RSS: 1076.36 MB
			RSS: 1258.63 MB
			RSS: 1435.95 MB
			RSS: 1618.65 MB
			RSS: 1801.58 MB
			RSS: 1984.29 MB
			RSS: 2166.22 MB
			RSS: 2344.72 MB
```

#### Any other comments?
Even with this change, we are still holding onto a lot of memory in addition to the raw buffers required to back the numpy arrays. A more thorough fix will come as part of [9106978040](https://man312219.monday.com/boards/7852509418/pulses/9106978040) once the initial Arrow reading PR is merged

### DIFF
--- a/cpp/arcticdb/column_store/chunked_buffer.hpp
+++ b/cpp/arcticdb/column_store/chunked_buffer.hpp
@@ -92,6 +92,9 @@ class ChunkedBufferImpl {
 
     ChunkedBufferImpl() = default;
 
+    explicit ChunkedBufferImpl(entity::AllocationType allocation_type) :
+            allocation_type_(allocation_type) {}
+
     explicit ChunkedBufferImpl(size_t size) {
         reserve(size);
     }
@@ -145,8 +148,14 @@ class ChunkedBufferImpl {
         *this = std::move(other);
     }
 
-    static auto presized(size_t size, entity::AllocationType allocation_type) {
-        ChunkedBufferImpl output(size, allocation_type);
+    static auto presized(size_t size) {
+        ChunkedBufferImpl output(entity::AllocationType::PRESIZED);
+        if (size > 0) {
+            if (size != DefaultBlockSize) {
+                output.handle_transition_to_irregular();
+            }
+            output.add_block(size, 0UL);
+        }
         output.ensure(size);
         return output;
     }
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -920,6 +920,7 @@ if(${TEST})
             codec/test/test_encode_field_collection.cpp
             codec/test/test_segment_header.cpp
             codec/test/test_encoded_field.cpp
+            column_store/test/test_chunked_buffer.cpp
             column_store/test/ingestion_stress_test.cpp
             column_store/test/test_column.cpp
             column_store/test/test_column_data_random_accessor.cpp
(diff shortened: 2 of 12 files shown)


cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -106,6 +106,9 @@ class ChunkedBufferImpl {
         }
     }
 
+    ChunkedBufferImpl(entity::AllocationType allocation_type) :
```
IvoDD: Maybe it's worth to have `explicit`?

python/tests/unit/arcticdb/test_memleak.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+import psutil
+
+from arcticdb import Arctic, LibraryOptions, QueryBuilder
+
+
+ac = Arctic("lmdb:///tmp/memleak_test")
+lib = ac.get_library("memleak_test", create_if_missing=True, library_options=LibraryOptions(columns_per_segment=1_000_000))
+
+def test_write_data():
```
IvoDD: This will get executed by pytest. I'm thinking this was not your intention given the `/tmp/memleak_test` above.

What do you think about reworking this file to have some proper memory leak unit tests instead?
Maybe putting it alongside @grusev 's stress tests [here](https://github.com/man-group/ArcticDB/blob/01f6a6a7c6aed669267892d082afbe744fd90cdc/python/tests/stress/arcticdb/version_store/test_mem_leaks.py#L376)

IvoDD (changes requested):

python/tests/unit/arcticdb/test_memleak.py
alexowens90: I'm gonna rip the test out before merging. The issue is that it isn't a memory leak, everything allocated is still referenced by a Python object, and is correctly freed when you delete the dataframe (or, more precisely, all of the backing numpy arrays).
When I get to the second part in the ticket referenced in the PR description, I'll add some ASV benchmarks to avoid regressing in this direction in future.

IvoDD: Ah I aggree an ASV peakmem benchmark should serve better.

PR #2405, opened 2025-06-13
Enhancement 9351235967: Support projecting a column of constant values

alexowens90:
#### Reference Issues/PRs
[9351235967](https://man312219.monday.com/boards/7852509418/pulses/9351235967)

Allow projections with a constant value of the form:
```
lazy_df = lib.read(sym, lazy=True)
lazy_df["new_col"] = 5
```

### DIFF
--- a/python/arcticdb/version_store/processing.py
+++ b/python/arcticdb/version_store/processing.py
@@ -1245,6 +1245,13 @@ def visit_expression(expr):
     expression_context = _ExpressionContext()
     input_columns = set()
     valueset_keys = dict()
-    _visit(expr)
-    expression_context.root_node_name = _ExpressionName(expr.get_name())
+    if isinstance(expr, ExpressionNode):
+        _visit(expr)
+        expression_context.root_node_name = _ExpressionName(expr.get_name())
+    elif isinstance(expr, (np.ndarray, list)):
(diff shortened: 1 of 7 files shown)


python/arcticdb/version_store/processing.py
```diff
@@ -1245,6 +1245,13 @@ def _handle_leaf(node):
...
-    _visit(expr)
-    expression_context.root_node_name = _ExpressionName(expr.get_name())
+    if isinstance(expr, ExpressionNode):
+        _visit(expr)
+        expression_context.root_node_name = _ExpressionName(expr.get_name())
+    elif isinstance(expr, (np.ndarray, list)):
+        raise ArcticNativeException("Query cannot create a new column from a list/set/array of values")
+    else:
+        key = to_string(expr)
+        expression_context.add_value(key, create_value(expr))
```
poodlewars, line 1255: Is there enough validation on the constant value? What if it is a type that we don't support, like a decimal or some user-defined class?

alexowens90: Yep, it's the [same codepath](https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/version_store/processing.py#L1163) as how we deal with all literal values being ingested into an AST

PR #2407, opened 2025-06-13
Add utilities to let the storage failure simulator simulate high latency conditions

poodlewars:
To be used in a microbenchmark in arcticdb-enterprise - https://github.com/man-group/arcticdb-enterprise/pull/232

### DIFF
--- a/cpp/arcticdb/storage/test/in_memory_store.hpp
+++ b/cpp/arcticdb/storage/test/in_memory_store.hpp
@@ -193,6 +193,7 @@ public:
     }
 
     bool key_exists_sync(const entity::VariantKey& key) override {
+        StorageFailureSimulator::instance()->go(FailureType::READ);
         std::lock_guard lock{mutex_};
         return util::variant_match(key,
            [&](const RefKey& key) {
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/storage/test/in_memory_store.hpp
```diff
@@ -263,7 +264,9 @@ class InMemoryStore : public Store {
                 return it->second->clone();
             });
 
-        return {VariantKey{key}, encode_dispatch(std::move(segment_in_memory), codec_, EncodingVersion::V1)};
+        Segment segment = encode_dispatch(std::move(segment_in_memory), codec_, EncodingVersion::V1);
+        (void) segment.calculate_size();
```
maxim-morozov, line 268: why do you need this call?

alexowens90 (approved):

cpp/arcticdb/storage/test/in_memory_store.hpp
poodlewars: It modifies the segment, downstream stuff expects the precalculated size to be there

PR #2411, opened 2025-06-16
Fix file descriptors for v1 StorageFixtures [Partially fixes the problem]

grusev:

(IMPORTANT:  it is important to note that this approach reduces the file descriptors usade with 40-50% it is not sure it eliminates all leakages, just does that from the test libraries caches, other leakages might still be present)

The main goal of this PR is to limit the number of file descriptors that are needed in order to run tests on Linux and MacOs as those OS-es have file descriptor limits. Windows does not have file descriptor limits.

The issue we have is with StorageFixture class in api.py - it caches all NativeVersionStore objects it creates when ```create_version_store_factory``` method is invoked. As our storage factories run now in session scope, the lifespan of this cache is the duration of tests. Thus for the duration of tests we have objects that keep their connections to storages. Additionally the storage space is nor wiped after the the test execution but only at the end. Thus if there is catatstrophic failure preventing reaching the end the storage will not be freed.

Approch taken is to shorten lifespan of the NativeVersionStores created for each test only for the duration of the test. Wherever it is possible the changes were implemented in the fixtures -  each fixture that creates NVS object is in charge of cleaning it now. For factory fixtures that is not possible and therefore tests were modified to have ```try ... finally``` clause.

There are also 2 helper utilities ```delete_nvs``` and ```delete_library``` implemented which help taking care of some exceptions that may occur during that process that would otherwise trigger failed or error state during cleanup

Also during the development it was noticed that LMDB tests and other types of tests rarely cleanup after themselves. That manifested in evergrowing disk space used by tests that leads to out of space errors for runners. Therefore this PR also tries to address that problem that exists with such tests and is not limited to v1 API fixtures only but also with v2 API fixtures

This PR also introduces a background process that will collect info fo file descriptor numbers during the execution of tests and at the end it will print the maximim number of file descriptors that we reached. This monitoring is needed to objectivly measure and understand needs of tests during test execution. In future this approach could be used to track much more stats about test execution. (NOTE: works only on local machines for now. Does not work on github unfortunately)

* Test Results (locally on WSL)

EXECUTION WITH FIX GCP/AWS_S3 TESTS FROM: python/tests/unit/arcticdb/version_store/test_parallel.py

All Tests finished after fix on a machine with limit 1024 file descriptors
 
During execution opened files reached 893 for GCP (for AWS S3 tests it was 725, or approx 23% more opened files on same tests for GCP bs AWS S3)

As the tests required arround 1550 opened files  for GCP when executed without this fix we have more than 45% reduction of file descriptors usage

NOTE: For monitoring following script was used:
``` 
while true; do
  lsof -c pytest | wc -l
  sleep 10
done
``` 

IMPORTANT: Because with the fix each NativeVersionStore was deleted immediately after usage all storage space was freed as soon as possible.  That is contrary to what was used to be before when storage was deleted at the end. And if file descriptors are exhausted before that no storage can be reclaimed

** Performance Analysis of code changes

Here we compare the execution of one and the same test suite with old and with new code in GitHub

- 3.8 Linux  - integration tests
-- old code - https://github.com/man-group/ArcticDB/actions/runs/15825356207/job/44604755311
```
= 3139 passed, 1356 skipped, 90 xfailed, 4 xpassed, 14 warnings in 1481.28s (0:24:41) =
```

-- NEW CODE -https://github.com/man-group/ArcticDB/actions/runs/15824842292/job/44602814883
```
= 3121 passed, 1357 skipped, 90 xfailed, 4 xpassed, 10 warnings in 1413.03s (0:23:33) =
```

- 3.11 Windows - integration tests

-- old code https://github.com/man-group/ArcticDB/actions/runs/15825356207/job/44605215726
```
= 2806 passed, 1655 skipped, 86 xfailed, 4 xpassed, 3740 warnings in 3413.78s (0:56:53) =
```

-- NEW CODE: https://github.com/man-group/ArcticDB/actions/runs/15824842292/job/44603214947
```
= 2788 passed, 1656 skipped, 86 xfailed, 4 xpassed, 3731 warnings in 3263.39s (0:54:23) =
```

- 3.13 Linux - integration tests
- old code - https://github.com/man-group/ArcticDB/actions/runs/15825356207/job/44604764324
```
= 3149 passed, 1346 skipped, 90 xfailed, 4 xpassed, 7647 warnings in 1337.87s (0:22:17) =
```

--NEW CODE - https://github.com/man-group/ArcticDB/actions/runs/15824842292/job/44602827969
```
= 3131 passed, 1347 skipped, 90 xfailed, 4 xpassed, 7642 warnings in 1374.96s (0:22:54) =
```

Analysis: there are no evidences that new fixtures are degrading the performance. The code at par with previous one and sligly faster which can be due to lower (18) number of tests or the code changes itself. 

*** Unit tests performance analysis

The only notable degradation is observed at unit tests level:
- In the past unit tests used to run for 5 minutes https://github.com/man-group/ArcticDB/actions/runs/15825356207/job/44604768352
- Currently they are running for 15 minutes: https://github.com/man-group/ArcticDB/actions/runs/15824842292/job/44602861927

The main difference is adding reclaiming of space for unit tests. The old code did not had it as majority tests were LMDB tests

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


build_tooling/monitor_opened_files.sh
```diff
@@ -0,0 +1,23 @@
...
+    timestamp=$(date '+%Y-%m-%d %H:%M:%S')
+
+    output="$timestamp"
+
+    for pid in $(pgrep -f 'pytest|multiprocessing'); do
+        count=$(lsof -p "$pid" 2>/dev/null | wc -l)
+        output+=" $count"
+    done
+
+    echo -e "$output" >> "$LOGFILE"
```
maxim-morozov, line 27: It worth to print a bit more information on top of time stamp and number. Especially if we are printing it into the centralised log file

build_tooling/monitor_opened_files.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+LOGFILE="${1:-opened_files.log}"
+
+PROCESS_NAME="pytest"
+
+# Clear or create the log file
+: > "$LOGFILE"
```
maxim-morozov, line 8: We are printing all runners to the same log file. I think we should provide different names to prevent runners to clash

build_tooling/parallel_test.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -e
 
+# Save firectory from which we started the script to go back after it finishes
```
maxim-morozov: Typo - Save **directory**

build_tooling/process_opened_files_log.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+LOGFILE="${1:-opened_files.log}"
+
+for i in {3..20}; do
```
maxim-morozov, line 8: Why those processes are 3-20?

build_tooling/process_opened_files_log.sh
```diff
@@ -0,0 +1,14 @@
...
+    if [[ -f "$LOGFILE" ]]; then
+        AWK='{ print $'$i' }'
+        max=$(awk "$AWK" "$LOGFILE" | sort -nr | head -n 1)
+        echo "Maximum open files during pytest process $i run: $max"
```
maxim-morozov, line 12: The number of the process doesn't provide clarity of what this process is. Would be good to have an idea what process is and also which process opened most file descriptors. It would be hard to debug based on the just max number.

python/arcticdb/storage_fixtures/api.py
```diff
@@ -97,7 +97,9 @@ def _factory_impl(
         suffix = 0
         while f"{name}{suffix or ''}" in libs_from_factory:
             suffix += 1
-        libs_from_factory[f"{name}{suffix or ''}"] = out
+        # No caching for NativeVersionStores at StorageFixture for non- Windows
+        # Preserve only names of created NativeVersionStores
+        libs_from_factory[f"{name}{suffix or ''}"] = out if _WINDOWS else None
```
maxim-morozov, line 102: Windows also has limit of file descriptors, they are just called handlers.

maxim-morozov: Also, did you have measurements what would be an impact on running tests without cache. How long does it take?

maxim-morozov: Why do we add None into the `libs_from_factory` instead of just skipping it?

python/arcticdb/storage_fixtures/api.py
```diff
@@ -125,8 +127,9 @@ def create_version_store_factory(self, default_lib_name: str, **defaults) -> Cal
 
     def slow_cleanup(self, failure_consequence=""):
         for lib in self.libs_from_factory.values():
-            with handle_cleanup_exception(self, lib, consequence=failure_consequence):
-                lib.version_store.clear()
+            if lib is not None:
```
maxim-morozov, line 130: If we were skipping adding Nones into the `libs_from_factory` we would not need to do this check

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -252,6 +253,20 @@ def create_test_cfg(self, lib_name: str) -> EnvironmentConfigsMap:
         return cfg, self.native_config
 
 
+def check_bucket(sff: Union['BaseS3StorageFixtureFactory', 'BaseGCPStorageFixtureFactory']):
```
maxim-morozov: can we give more meaningful names, rather `sff`. Took me some time to figure out it stands for StorageFixtureFactory

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -252,6 +253,20 @@ def create_test_cfg(self, lib_name: str) -> EnvironmentConfigsMap:
...
+    content = s3_tool.list_bucket(sff.default_prefix)
+
+    logger.warning(f"Total objects left: {len(content)}")
+    logger.warning(f"First 100: {content[0:100]}")
+    logger.warning(f"BUCKET: {sff.default_bucket}")
+    left_from = set()
+    for key in content:
+        library_name = key.split("/")[1] # get the name from object
+        left_from.add(library_name)
+    logger.warning(f"Left overs from libraries: {left_from}")
```
maxim-morozov, line 268: Worth to specify key as well, to know what was left

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -302,6 +317,8 @@ def cleanup_bucket(self, b: S3Bucket):
             # and if we try to delete the bucket, it will fail
             b.slow_cleanup(failure_consequence="The following delete bucket call will also fail. ")
 
+            check_bucket(self)
```
maxim-morozov, line 321: Why `check_bucket` is not a member function of StorageFixtureFactory?

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -717,6 +736,19 @@ def create_bucket(s3_client, bucket_name, max_retries=15):
             time.sleep(1)   
 
 
+def list_moto_storage(moto: 'MotoS3StorageFixtureFactory', bucket: Union['S3Bucket']):
```
maxim-morozov: Very similar to the `check_bucket` function. Can all this be consolidated to the member function of StorageFixtureFactory?

python/arcticdb/util/utils.py
```diff
@@ -93,6 +97,33 @@ def __init__(self, message: str):
         super().__init__(sanitized_message)
 
 
+def delete_nvs(nvs: NativeVersionStore, ac: Arctic = None):
+    if isinstance(nvs, NativeVersionStore):
+        get_logger().info(f"Removing data for NativeVersionStore library_path: {nvs.library().library_path}")
```
maxim-morozov: You are not assign the returned logger to anything and you are using the local logger underneath. It would probably crash

python/arcticdb/util/utils.py
```diff
@@ -93,6 +97,33 @@ def __init__(self, message: str):
...
+    
+
+def delete_library(ac: Arctic, lib_name: str):
+    try:
+        ac.delete_library(lib_name)
+    except arcticdb_ext.exceptions.StorageException as e:
+        if "LMDB" in repr(e): # skip LMDB errors
+            pass
+        else: 
+            logger.warning(f"Caught exception during deletion of library '{lib_name}' - {repr(e)}")
```
maxim-morozov, line 124: Same here. The logger is not defined

build_tooling/monitor_opened_files.sh
```diff
@@ -0,0 +1,37 @@
...
+        fi
+        output+=" $count"
+    done
+
+    echo -e "$output" >> "$LOGFILE"
+
+    # Processes dump to file
+    echo $timestamp >> "$PROCESSESFILE"
+    ps -ef >> "$PROCESSESFILE"
+    echo "======================================" >> "$PROCESSESFILE"
```
maxim-morozov: Can we remove this echo

python/tests/conftest.py
```diff
@@ -719,75 +733,100 @@ def in_memory_store_factory(mem_storage, lib_name) -> Callable[..., NativeVersio
 # endregion
 # region ================================ `NativeVersionStore` Fixtures =================================
 @pytest.fixture
-def real_s3_version_store(real_s3_store_factory):
-    return real_s3_store_factory()
+def real_s3_version_store(real_s3_store_factory) -> Generator[NativeVersionStore, Any, Any]:
+    nvs = real_s3_store_factory()
```
maxim-morozov: It's a same pattern across all of those fixtures, factory, yield, delete. Instead of repeating this pattern. You can wrap it in a class and call `yield from` - https://peps.python.org/pep-0380/

python/arcticdb/util/utils.py
```diff
@@ -93,6 +97,33 @@ def __init__(self, message: str):
...
 
+def delete_nvs(nvs: NativeVersionStore, ac: Arctic = None):
+    if isinstance(nvs, NativeVersionStore):
+        get_logger().info(f"Removing data for NativeVersionStore library_path: {nvs.library().library_path}")
+        try:
+            nvs.version_store.clear()
+            logger.info(f"SUCCESS in deletion")  
+        except Exception as e:
+            logger.warning(f"Exception caught during NativeVersionStore clear: {repr(e)}")
+    elif isinstance(nvs, Library):
```
maxim-morozov, line 109: Why do we need to check instance of this? I expect as with any python function, people should pass appropriate type of it will crash or throw exception.

python/tests/integration/arcticdb/test_arctic.py
```diff
@@ -1274,21 +1277,23 @@ def test_segment_slicing(arctic_client, lib_name):
...
+        rows = 12
+        columns = 3
+        data = {}
+        for col in range(columns):
+            data[f"col{col}"] = np.arange(100 * col, (100 * col) + rows)
+        lib.write(symbol, pd.DataFrame(data))
+        num_data_segments = len(lib._nvs.read_index(symbol))
+        assert num_data_segments == math.ceil(rows / rows_per_segment) * math.ceil(columns / columns_per_segment)
+    finally:
+        ac.delete_library(lib_name)
```
maxim-morozov, line 1296: It's a same pattern across all of those tests. We should think how to wrap it in order not to repeat the same code.

python/tests/integration/arcticdb/test_s3.py
```diff
@@ -147,20 +160,24 @@ def s3_storage_dots_in_path(request):
     ) as f:
         with f.create_fixture() as g:
             yield g
+        list_moto_storage(f, g)
```
maxim-morozov: How do you use g outside the with close? Which I think gonna be already destroyed.

python/tests/integration/arcticdb/test_s3.py
```diff
@@ -331,59 +356,65 @@ def test_delete_over_time(lib_name, storage_bucket, clear_query_stats):
...
-            qs.reset_stats()
-            assert sum_all_operations(stats) == base_ops_count == expected_ops, visualize_stats_diff(base_stats, stats)
+            # make sure that the write and prune makes a constant number of operations
+            for i in range(iters):
+                lib.write("s", data=create_df(), prune_previous=True)
+                stats = qs.get_query_stats()
+                qs.reset_stats()
+                assert sum_all_operations(stats) == base_ops_count == expected_ops, visualize_stats_diff(base_stats, stats)
+        finally:
+            delete_nvs(lib)
```
maxim-morozov, line 419: Need space in the end of the file

python/tests/integration/toolbox/test_storage_mover.py
```diff
@@ -86,10 +87,13 @@ def test_storage_mover_single_go(lmdb_version_store_v1, arctidb_native_local_lib
...
-    s = StorageMover(lmdb_version_store_v1._library, dst_lib._library)
-    s.go()
+    try:
+        s = StorageMover(lmdb_version_store_v1._library, dst_lib._library)
+        s.go()
 
-    compare_two_libs(lmdb_version_store_v1, dst_lib)
+        compare_two_libs(lmdb_version_store_v1, dst_lib)
+    finally:
+        delete_nvs(dst_lib)
```
maxim-morozov, line 96: Why are we deleting only the destination library and not the source?

python/tests/integration/toolbox/test_storage_mover.py
```diff
@@ -305,37 +322,40 @@ def _tmp_test_body(to_copy, existing, lib_with_gaps_and_reused_keys, version_sto
...
+        target_keys = lt.find_keys(KeyType.TABLE_DATA)
+        target_keys.sort()
+        if existing:
+            target_keys.pop(0)
+
+        for a, e in zip(target_keys, expected_target):
+            assert a.content_hash == e.content_hash
+            assert a.creation_ts >= source_keys[-1].creation_ts
+    finally:
+        delete_nvs(target)
```
maxim-morozov: New line at the end of the file is missing

python/tests/stress/arcticdb/version_store/test_stress_dynamic_bucketize.py
```diff
@@ -29,18 +30,21 @@ def test_dynamic_bucketize_append_variable_width(
...
+            df2 = get_wide_df(count, random.randrange(max_col_width) + 1, max_col_width)
+            df1 = pd.concat([df1, df2])
+            lib.append(symbol, df2)
+            count += 1
+        res = lib.read(symbol).data
+        df1 = df1.reindex(sorted(list(df1.columns)), axis=1)
+        res = res.reindex(sorted(list(res.columns)), axis=1)
+        assert_frame_equal(res, df1)
+    finally:
+    	delete_nvs(lib)
```
maxim-morozov: New line is missing at the end of the file.

build_tooling/monitor_opened_files.sh
grusev: This is a background process which prints the details for each pytest and its child process to a file. In that aspect it is just a single writer that takes stats on 10s basis. And we have no way how to identify each runners from its CMD column of process list. Therefore the line printed contains just a timestamp and for each test executor the opened file descriptors. The number of active child process varies in the time also

build_tooling/process_opened_files_log.sh
grusev: We have no idea how many processes will be spawned, usually they are at least 4 child process, up to how many will depend perhaps on our settings and the githib runner. But in any case if no process exists no data will be written, thus each line may contain different number of values printed. And as we do not have idea how many will actually be bigger number than initially anticipated is always good as with time this will grow perhaps.

build_tooling/process_opened_files_log.sh
grusev: that can eventually happen only from pytest type of logger. This indeed cannot have this information. It is only for us to have an initial idea at this moment do we have an issue or not. Later when we start to see issues we should thing of better way to do monitoring from within pytest istels. Currently I think the fast way is better as we just need to open the curtain

python/tests/integration/arcticdb/test_s3.py
grusev: that was only for debugging purposes. Will remove it

python/tests/integration/toolbox/test_storage_mover.py
grusev: yes - only this lib is created in the test. The source lib is created outside the test in the fixxture which will take care about it (see lib_with_gaps_and_reused_keys() fixture)

build_tooling/monitor_opened_files.sh
grusev: this was developed in pair with process_opened_files.sh specifically to try to monitor file descriptors only through time.

And they work perfectly on WSL but not on GitHub  where it seems it is impossible to obtain numbers different that 0 perhaps for some security reasons I do not understand yet.

So this approach can be used for now only locally. Thus perhaps no more investments here are needed....

In General the idea of this file is to add different stats that can be monitored over time - those stats would then be in diffferent files so the minimal info that is logged in a file is the timestamp and the resource unit at that time. This gives idea for how it changes over time

python/arcticdb/storage_fixtures/api.py
grusev: For Windows file descriptors there is a limit which is arround 16 mil per process. That is why we have not seen any problems on windows platforms in contrast to unix where the defaults are very low 1024. Thus we can keep the old caching working on windows 

For measurments see the description - generally this is slighly faster except fro unit tests

I assign None so that we keep the structure in place and not require rewriting it further the code except to avoid calling clear() on None. Thus changes to original behavior are minimal and can easily be switched on again as Windows example. Plus this dict now is caching the names of libraries we create now (provided that one day we want to use it for some reasons)

python/arcticdb/storage_fixtures/api.py
grusev: yep, but this way this big change is non-intrusive and easily backed out - the original code no matter how tiny stays with us still

python/arcticdb/storage_fixtures/s3.py
grusev: that is also avail for first 100 items on line 262:
```
    logger.warning(f"First 100: {content[0:100]}")
```
With the commented code we are extracting the name of the test which produced the keys. Thus this list should be shorter as there should be several keys per test (the name of symbol is produced as concatentation of test name and other things, thus we are interested exactly in that which will pinpint us toproblematic area)

Generally this now should be very rare - this is left only for REAL storages AWS and GCP

python/arcticdb/storage_fixtures/s3.py
grusev: StorageFixtureFactory is very abstract and it lacks variables used in function default_key, endpoint, default_bucket. Thus it does not make sense to be there.

Further BaseGCPStorageFixtureFactory and BaseS3StorageFixtureFactory class are different so it cannot be part of any one currently therefore it can only be free function at the moment

python/arcticdb/storage_fixtures/s3.py
grusev: see previous comment

grusev: This will do nothing actually in real tests. It will only print content if the test has left any, but since if any leftovers will trigger error from clean_bucket the test creator will have to fix this before code is merged. So this functionality will work only if there is an error and ideally not in GitHub. But if it triggers there we will have clue what is wronf

I have made it non-public part of MotoS3StorageFixtureFactory now:

```
    def _log_moto_storage_content(self, bucket: Union['S3Bucket']):
```    

NOTE: even if I make it part of BaseS3StorageFixtureFactory it will not be part of BaseGCPStorageFixtureFactory as the later do not have boto client

python/arcticdb/util/utils.py
grusev: this logger is intelligent one - it has quite lots of options to use it + it sanitizes the text. But generally uility functions should avoid usage of loggers, that thould be part of the test, but this is very impotant info that is an exception and should be logger centrally

grusev: get_logger() will always succeed but makes sense to have common logger for all helpers:
logger = get_logger()

python/arcticdb/util/utils.py
grusev: unfortunately yes.
We have 2 types of factories that return either NativeVersionStore or Library. So we need a function that can handle both.
Unfortunately the Library object cannot delete the library itself like the NativeVersionLibrary. Library deletion of v2 libraries can happen onl;y via Arctic class, but Library object does not have reference to it at this moment. So effectivly every usage of this type of factoy is leaking storage at the moment

python/tests/integration/arcticdb/test_arctic.py
grusev: yes but that is not quite easy excercise. I do think we can improve significantly fixtures but that is quite some effort to understand to what direction we want to go first.

If I was to design for instance the fixtures they would not return Arcic object but rather TestArctic object which is a thin wrapper over Arctic and would have refences to libs created so they can be deleted after its usage. Our current design is to return the objects directly so that is leaving the test creator with the obligation to do that

python/tests/conftest.py
grusev: Good one. I always told myself I will do it at the end and yet forgot it

build_tooling/monitor_opened_files.sh
grusev: since monitoring could not work on gihub this section was removed so now only one file will be created (cannot obtain opened files perhaps due to some restrictions)

python/arcticdb/util/utils.py
grusev: fixed

python/arcticdb/util/utils.py
```diff
@@ -93,6 +98,33 @@ def __init__(self, message: str):
         super().__init__(sanitized_message)
 
 
+def delete_nvs(nvs: NativeVersionStore, ac: Arctic = None):
+    if isinstance(nvs, NativeVersionStore):
+        logger.info(f"Removing data for NativeVersionStore library_path: {nvs.library().library_path}")
+        try:
+            nvs.version_store.clear()
```
phoebusm, line 105: This is quite similar to `slow_cleanup`. If `libs_from_factory` is the cache we are trying to clean up, why not making `slow_cleanup` being called whenever `S3Bucket` (function fixture object type) exits. That should make the fix much more simple, if `libs_from_factory` is really the culprit

phoebusm:
> As the tests required arround 1550 opened files for GCP when executed without this fix we have more than 45% reduction of file descriptors usage

So I guess it's only a partial fix?

maxim-morozov:
The new PR is in https://github.com/man-group/ArcticDB/pull/2431

PR #2414, opened 2025-06-17
Validate staged data symbol names

poodlewars:
We have quite strict validation we apply to newly written symbols at the time of the write. But this logic is accidentally not implemented for staged writes, where we should blow up when we compact and write the new symbol. See `name_validation.cpp` for our validation logic - symbols must be at most 254 chars long, in ASCII range [32, 126] and must not include `*` `<` or `>`.

If a symbol already exists with a name that fails this validation we still allow staged writes to it - to keep workflows that predate our stricter validation working.

See Monday ticket 8644487034 for more.


python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -125,6 +138,28 @@ def test_unhandled_chars_default(object_version_store, unhandled_char):
...
+    df = sample_dataframe()
+    with pytest.raises(UserInputException):
+        object_version_store.write(sym, df, parallel=True)
+    assert sym not in object_version_store.list_symbols_with_incomplete_data()
+
+
+@pytest.mark.parametrize("sym", [chr(32), chr(33), chr(125), chr(126), "fine", "l" * 254])
+@pytest.mark.storage
+def test_ok_chars_staged_data(object_version_store, sym):
+    """Test that by default, the problematic chars are raising an exception at staging time."""
```
G-D-Petrov: This comment is wrong

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -113,6 +112,20 @@ def test_s3_breaking_chars(object_version_store, breaking_char):
...
+def test_s3_breaking_chars_staged(object_version_store, breaking_char):
+    """Test that chars that are not supported are raising the appropriate exception and that we fail on write without
+    corrupting the db.
+    """
+    sym = f"prefix{breaking_char}postfix"
+    df = sample_dataframe()
+    with pytest.raises(UserInputException):
+        object_version_store.write(sym, df, incomplete=True)
+
+    assert sym not in object_version_store.list_symbols_with_incomplete_data()
```
G-D-Petrov: I would add a sanity assert that `object_version_store.list_symbols_with_incomplete_data()` is empty.
Same for `test_unhandled_chars_staged_data`

G-D-Petrov (approved):

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -404,9 +405,18 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
 void write_parallel_impl(
     const std::shared_ptr<Store>& store,
+    const std::shared_ptr<VersionMap>& version_map,
     const StreamId& stream_id,
     const std::shared_ptr<InputTensorFrame>& frame,
     const WriteIncompleteOptions& options) {
+    // Apply validation for new symbols, but don't interfere with pre-existing symbols that would fail our modern validation.
+    CheckOutcome check_outcome = verify_symbol_key(stream_id);
+    if (std::holds_alternative<Error>(check_outcome)) {
+        if (auto&& [key, deleted] = get_latest_version(store, version_map, stream_id); !key) {
```
alexowens90: I think this check is too restrictive, I think we just want to check if the vref key exists. Otherwise the workflow:
```
lib.stage(invalid_sym_name, data)
lib.finalize_staged_data(invalid_sym_name)
lib.delete(invalid_sym_name)
```
will break

alexowens90 (changes requested):

cpp/arcticdb/stream/incompletes.cpp
poodlewars: Hmm maybe `get_latest_version` doesn't do what I thought. Either way the testing here is rubbish. I'll have a look

poodlewars: I've pushed a new test that covers that workflow - I think what I have is correct. The `get_latest_version` returns a non-optional key even for deleted symbols. It's just checking the vref key under the hood, but I think it's cleanest to just use the `get_latest_version` abstraction.

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -407,6 +408,16 @@ void write_parallel_impl(
     const StreamId& stream_id,
     const std::shared_ptr<InputTensorFrame>& frame,
     const WriteIncompleteOptions& options) {
+    // Apply validation for new symbols, but don't interfere with pre-existing symbols that would fail our modern validation.
+    CheckOutcome check_outcome = verify_symbol_key(stream_id);
+    if (std::holds_alternative<Error>(check_outcome)) {
+        VersionMapEntry ref_entry;
+        read_symbol_ref(store, stream_id, ref_entry);
```
poodlewars: This does slightly more work in-memory than we strictly need but shouldn't be a big deal and keeps the code a bit cleaner.

alexowens90 (approved):

PR #2415, opened 2025-06-18
Validate snapshot names 

poodlewars:
We have quite strict validation we apply to newly written symbols at the time of the write. But this logic was accidentally not implemented for snapshots. See `name_validation.cpp` for our validation logic - symbols must be at most 254 chars long, in ASCII range [32, 126] and must not include * < or >. This PR introduces this validation for snapshot names.

Monday: 8667974441

Similar to PR 2414 for staged data.


cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -653,6 +653,9 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
        .def("indexes_sorted",
              &PythonVersionStore::indexes_sorted,
              py::call_guard<SingleThreadMutexHolder>(), "Returns the sorted indexes of a symbol")
+        .def("verify_snapshot",
+             &PythonVersionStore::verify_snapshot,
+             py::call_guard<SingleThreadMutexHolder>(), "Create a snapshot")
```
alexowens90: nit: copy/pasted string

PR #2417, opened 2025-06-19
Batch restore version with timestamp as_of

poodlewars:
Support `batch_restore_version` with timestamp `as_of` parameters.

Monday: 8560360589


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1597,29 +1578,41 @@ std::map<StreamId, VersionVectorType> get_multiple_sym_versions_from_query(
...
-        fut_vec.emplace_back(async::submit_io_task(AsyncRestoreVersionTask{store(), version_map(), *stream_id, version->second, maybe_prev}));
+    std::vector<StreamId> symbols_with_missing_versions;
+    for (const auto& [i, opt_key] : folly::enumerate(versions_to_restore)) {
+        if(!opt_key) {
+            symbols_with_missing_versions.push_back(stream_ids.at(i));
+        }
     }
-    auto output = folly::collect(fut_vec).get();
 
-    std::vector<folly::Future<folly::Unit>> symbol_write_futs;
```
poodlewars, line 1618: Rationale for why we do not need to touch the symbol list is in the message for commit c7d12b42ad44f66dfd3585db3138e693f4b55bd4

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1597,29 +1578,41 @@ std::map<StreamId, VersionVectorType> get_multiple_sym_versions_from_query(
...
+        }
+    }
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(duplicate_streams.empty(), "Duplicate symbols in restore_version request. Symbols submitted more than once [{}]",
+                                                          fmt::join(duplicate_streams, ","));
 
-    for(const auto& stream_id : folly::enumerate(stream_ids)) {
-        auto prev = previous->find(*stream_id);
-        auto maybe_prev = prev == std::end(*previous) ? std::nullopt : std::make_optional<AtomKey>(prev->second);
+    auto previous = batch_get_latest_version(store(), version_map(), stream_ids, true);
+    auto versions_to_restore = folly::collect(batch_get_versions_async(store(), version_map(), stream_ids, version_queries)).get();
```
alexowens90, line 1594: We could make this more performant by chaining to the individual futures returned here as we do in the other batch methods, but I don't think this API is widely used enough to worry about it for now

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: What if the version being restored is from a snapshot, but the symbol is otherwise deleted?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1597,29 +1578,41 @@ std::map<StreamId, VersionVectorType> get_multiple_sym_versions_from_query(
...
-    std::vector<folly::Future<folly::Unit>> symbol_write_futs;
-    for(const auto& item : output) {
-        symbol_write_futs.emplace_back(async::submit_io_task(WriteSymbolTask(store(), symbol_list_ptr(), item.first.key_.id(), item.first.key_.version_id())));
+    missing_data::check<ErrorCode::E_NO_SUCH_VERSION>(symbols_with_missing_versions.empty(), "Could not find the requested versions for some symbols during restore_version. "
+                                                                                                 "Symbols with missing versions [{}]",
+                                                          fmt::join(symbols_with_missing_versions, ","));
+
+    std::vector<folly::Future<std::pair<VersionedItem, TimeseriesDescriptor>>> fut_vec;
+    for(const std::optional<AtomKey>& key_opt : versions_to_restore) {
+        AtomKey key = *key_opt;
```
alexowens90: Why copy?

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: This was deliberate to try and make sure that the batch either succeeds or it doesn't (insofar as we can ever make a promise like that)

alexowens90 (approved):

G-D-Petrov (approved):

PR #2419, opened 2025-06-19
Staged data new API

ognyanstoimenov:
New stage API

* `stage` now returns `StageResult` containing the staged append_data (feature flagged)
* Pickling support for `StageResult` and `AtomKey`
* `finalize_staged_data` and `sort_and_finalize_staged_data` now have `stage_results` (currenlty named `_stage_results` while developing) parameter (which currently does nothing)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/arcticdb/version_store/library.py
```diff
@@ -811,7 +811,7 @@ def stage(
         sort_columns:
             Sort the data by specific columns prior to writing.
         """
-        self._nvs.stage(
+        return self._nvs.stage(
```
alexowens90, line 816: Method return type needs updating too

python/arcticdb/version_store/library.py
alexowens90, line 1: General point for discussion - should we update the docstrings with the new return type/arguments, marking them as unused? Or leave off? I'm leaning towards the former

alexowens90: Oh I see `stage` is still returning `None` unless the config int. Question still applies to new arguments to finalize/sort and finalize though

python/tests/unit/arcticdb/version_store/test_sort.py
```diff
@@ -5,9 +5,15 @@
...
 from arcticdb_ext.storage import KeyType
 from arcticdb_ext.version_store import SortedValue
+from arcticdb_ext import set_config_int
 
 from arcticdb.util.test import random_strings_of_length
 
+# Reuse all current testing with both the new and old API
+@pytest.fixture(params=[True, False], autouse=True)
+def setup_is_v2(request):
+    set_config_int("Stage.IsV2APIEnabled", 1 if request.param else 0)
```
alexowens90: This doesn't seem to be used?

python/arcticdb/version_store/_store.py
```diff
@@ -529,9 +530,14 @@ def stage(
             norm_failure_options_msg=norm_failure_options_msg,
         )
         if isinstance(item, NPDDataFrame):
-            self.version_store.write_parallel(symbol, item, norm_meta, validate_index, sort_on_index, sort_columns)
+            is_v2_stage_api_enabled = get_config_int("Stage.IsV2ApiEnabled") == 1
```
alexowens90: Can we not use V2 for yet another different feature 😆 
I think we should also make it clearer that this is a private/testing only config option and that nobody should be using it

python/tests/unit/arcticdb/version_store/test_stage.py
```diff
@@ -0,0 +1,71 @@
...
+        assert all(x is None for x in staged_results)
+    else:
+        assert all(len(staged_result._staged_segments) == 2 and staged_result._version == 0 for staged_result in staged_results)
+
+    assert_frame_equal(lib.read(sym).data, not_staged, check_freq=False)
+    lib.finalize_staged_data(sym, mode=finalize_mode)
+    expected_df = pd.concat([not_staged] + data_to_stage) if finalize_mode is StagedDataFinalizeMethod.APPEND else pd.concat(data_to_stage)
+    assert_frame_equal(lib.read(sym).data, expected_df, check_freq=False)
+
+def stage_result_worker(stage_result, result_queue):
```
alexowens90: Can just use `pickle` directly to test this
https://github.com/man-group/ArcticDB/blob/master/python/tests/unit/arcticdb/version_store/test_query_builder.py#L76

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -162,7 +162,18 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
...
+        return py::make_tuple(key.id(), key.version_id(), key.creation_ts(), key.content_hash(), key.start_index(), key.end_index(), key.type());
+    },[](py::tuple t) {
+        if (t.size() != 7)
+            throw std::runtime_error("Invalid AtomKey pickle object!");
+
+        AtomKey key(t[0].cast<StreamId>(), t[1].cast<VersionId>(), t[2].cast<timestamp>(),
+            t[3].cast<ContentHash>(), t[4].cast<IndexValue>(), t[5].cast<IndexValue>(),
+            t[6].cast<KeyType>());
+        return key;
+    }
```
alexowens90, line 175: This needs some thorough testing. Does this all "just work" even if there aren't Python bindings for the individual components of the atom key? What happens if one of the uint64 fields like the content hash is larger than what is representable by an int64?

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -162,7 +162,18 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
     .def(pybind11::self != pybind11::self)
     .def("__repr__", &AtomKey::view)
     .def(py::self < py::self)
-    ;
+    .def(py::pickle([] (const AtomKey& key) {
+        return py::make_tuple(key.id(), key.version_id(), key.creation_ts(), key.content_hash(), key.start_index(), key.end_index(), key.type());
+    },[](py::tuple t) {
+        if (t.size() != 7)
+            throw std::runtime_error("Invalid AtomKey pickle object!");
```
alexowens90: Can do this in one line with `util::check`

python/tests/unit/arcticdb/version_store/test_sort.py
ognyanstoimenov: the autouse=True part applies it to all the tests in the file without specifying it explicitly. I will remove this once we un-feature flag the new API

python/arcticdb/version_store/_store.py
ognyanstoimenov: Totally. 😃  I thought this was the agreed way to name new features but will be glad to change it to something else

cpp/arcticdb/version/python_bindings.cpp
ognyanstoimenov: yep, I will add more testing if we're happy with the API

python/arcticdb/version_store/library.py
poodlewars, line 1: What are we going to do about `write` with `staged=True`? I would think we should deprecate it and start logging warnings when it's used? (Python has some quite fancy stuff to help with that https://docs.python.org/3/library/warnings.html ).

python/arcticdb/version_store/_store.py
poodlewars, line 1: Need to decide what to do about `write` with `parallel=True` or `incomplete=True`. Probably deprecate?

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -327,6 +338,24 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
         .def_property_readonly("creation_ts", &DescriptorItem::creation_ts)
         .def_property_readonly("timeseries_descriptor", &DescriptorItem::timeseries_descriptor);
 
+    py::class_<StageResult>(version, "StageResult")
+        .def(py::init([]() { return StageResult({}); }))
+	.def_property_readonly("_staged_segments", [](const StageResult& self) { return self.staged_segments; })
+	.def_property_readonly("_version", [](const StageResult& self) { return self.version; })
```
poodlewars: Ah great glad you have the `_version` property 👍 

Don't think either property should be protected FWIW, this is just POD.

python/arcticdb/version_store/_store.py
```diff
@@ -2239,6 +2251,8 @@ def compact_incomplete(
               ``compact_incomplete``.
 
             To manually delete staged data, use the ``remove_incomplete`` function.
+        _stage_results : `Optional[List[StageResult]]`, default=None
```
poodlewars: I would leave this undocumented, on the `library.py` too

cpp/arcticdb/version/python_bindings.cpp
poodlewars: We need to think about forwards compat as the pickled structure changes. We can include a version number inside the pickled atom key structure to mark the serialization format.

In particular we might see serialized atom keys with different size()s over time so the check is too strict, >= 7 would be more suitable.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -334,6 +343,23 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
...
+    py::class_<StageResult>(version, "StageResult")
+        .def(py::init([]() { return StageResult({}); }))
+	.def_property_readonly("staged_segments", [](const StageResult& self) { return self.staged_segments; })
+	.def_property_readonly("version", [](const StageResult& self) { return self.version; })
+        .def(py::pickle(
+            [](const StageResult& s) {
+                return py::make_tuple(s.staged_segments, s.version);
+            },
+            [](py::tuple t) {
+                util::check(t.size() == 2, "Invalid StageResult pickle object!");
```
poodlewars: Similar comments about forwards compat of the pickled structure, we should put its version number as its first field and drive everything from that. It could end up having more fields over time so the `size()` check is too strict.

python/tests/unit/arcticdb/version_store/test_pickle_atomkey.py
poodlewars, line 1: Need tests for the different key types, there's no testing that they're being handled correctly at the moment.

Once this is released, we'll need to add compat tests to check that master can unpickle keys written by this binary.

poodlewars (changes requested):

python/arcticdb/version_store/_store.py
ognyanstoimenov: Alex had a comment about documenting this, but I agree it is better hidden for now

poodlewars (approved):

PR #2431, opened 2025-06-26
Fix for file descriptors for v1 StorageFixtures [Partially fixes the problem] version 2

grusev:

The main problem is stated with previous PR#2411

This PR explores better fix with much less code having the same outcome - reducing the file descriptors but perhaps not eliminating constant growth

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -586,26 +586,26 @@ def arctic_client_lmdb(request, encoding_version) -> Arctic:
 
 
 @pytest.fixture
-def arctic_library(arctic_client, lib_name) -> Library:
+def arctic_library(arctic_client, lib_name) -> Generator[Library, None, None]:
     yield arctic_client.create_library(lib_name)
     arctic_client.delete_library(lib_name)
 
 
 @pytest.fixture
-def arctic_library_dynamic(arctic_client, lib_name) -> Library:
+def arctic_library_dynamic(arctic_client, lib_name) -> Generator[Library, None, None]:
     lib_opts = LibraryOptions(dynamic_schema=True)
     yield arctic_client.create_library(lib_name, library_options=lib_opts)
     arctic_client.delete_library(lib_name)
 
 
 @pytest.fixture
-def arctic_library_v1(arctic_client_v1, lib_name) -> Library:
+def arctic_library_v1(arctic_client_v1, lib_name) -> Generator[Library, None, None]:
     yield arctic_client_v1.create_library(lib_name)
     arctic_client_v1.delete_library(lib_name)
 
 
 @pytest.fixture
-def arctic_library_lmdb(arctic_client_lmdb, lib_name) -> Library:
+def arctic_library_lmdb(arctic_client_lmdb, lib_name) -> Generator[Library, None, None]:
     yield arctic_client_lmdb.create_library(lib_name)
     arctic_client_lmdb.delete_library(lib_name)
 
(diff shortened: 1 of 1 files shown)


python/tests/conftest.py
```diff
@@ -646,9 +647,22 @@ def basic_arctic_library(basic_arctic_client, lib_name) -> Library:
...
+
+
 @pytest.fixture
-def version_store_factory(lib_name, lmdb_storage) -> Callable[..., NativeVersionStore]:
-    return lmdb_storage.create_version_store_factory(lib_name)
+def version_store_factory(lib_name, lmdb_storage) -> Generator[Callable[..., NativeVersionStore], None, None]:
+    # Do not delete LMDB library on windows
+    # Otherwise there will be no storage space left for unit tests
+    # very peculiar behavior for LMDB, not investigated yet
+    yield from _store_factory(lib_name, lmdb_storage, not WINDOWS)
```
phoebusm: Instead of wrapping the factory function, can we use the existing code path in https://github.com/man-group/ArcticDB/blob/5407104134ae0c9907a9fbdd1ac3a75907606efd/python/arcticdb/storage_fixtures/s3.py#L113

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -111,6 +113,10 @@ def __init__(
             # client_cert_dir is skipped on purpose; It will be tested manually in other tests
 
     def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            self.slow_cleanup()
```
phoebusm: The `cleanup_bucket` below also calls `slow_cleanup()` too.
I think a better way to do so is setting `self.factory.clean_bucket_on_fixture_exit` `True` by default in the constructor

grusev: you have a point there. By default clean_bucket_on_fixture_exit  is set to True. Still there might be cases when no cleanup is supposed to happen, thus the if is ok.
We do cleanup by using clear() and delete_library() methods. But do they really delete all things? Added new method check_bucket that will assure nothing is left behind. If something is left behind it will fail in the branch you work on and will display info what is left behind to track and see why.
It is not supposed to fail on master, although that it can when some bad thing has been merged.
As I have not seen such exhaustive checks anywhere adding this should be good thing. It will work on real storages only

phoebusm (approved):

python/tests/conftest.py
grusev: unfortunately turned out we cannot use it. The bucket is creation in another fixture at session scope. We want to dispose objects exactly here and not at the end of the session

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -172,6 +177,21 @@ def copy_underlying_objects_to(self, destination: "S3Bucket"):
...
+        content = s3_tool.list_bucket(self.bucket)
+
+        logger.warning(f"Total objects left: {len(content)}")
+        logger.warning(f"First 100: {content[0:100]}")
+        logger.warning(f"BUCKET: {self.bucket}")
+        left_from = set()
+        for key in content:
+            library_name = key.split("/")[1] # get the name from object
+            left_from.add(library_name)
+        logger.warning(f"Left overs from libraries: {left_from}")
```
G-D-Petrov, line 192: nit: should be Leftovers

python/tests/stress/arcticdb/version_store/test_stress_finalize_staged_data.py
```diff
@@ -90,7 +90,8 @@ def test_finalize_monotonic_unique_chunks(arctic_library_lmdb):
     df = cachedDF.generate_dataframe_timestamp_indexed(num_rows_initially, total_number_rows, cachedDF.TIME_UNIT)
 
     cnt = 0
-    for iter in [500, 1000, 1500, 2000]:
+    iterations = [500, 1000] if MACOS else [500, 1000, 1500, 2000]
```
G-D-Petrov, line 94: Add a comment on why this is needed

python/tests/unit/arcticdb/version_store/test_symbol_concatenation.py
```diff
@@ -627,6 +629,7 @@ def test_symbol_concat_non_existent_symbol(lmdb_library):
         concat(lib.read_batch([sym, "non-existent symbol"], lazy=True)).collect()
 
 
+@pytest.mark.skipif(MACOS_WHEEL_BUILD, reason="Fatal Python error: Segmentation fault")
```
G-D-Petrov: This applies to add of the skipifs, can you please open a ticket for investigating this problem and link it here?

G-D-Petrov (approved):

python/arcticdb/storage_fixtures/s3.py
grusev: Yes, there should not be any. This check is added mainly to help investigation of possible future occassions. I have used it to track omissions in my approach initially, and was helpuf thus decided to leave it here for others

PR #2435, opened 2025-07-01
Performance 9380719845: Reduce memory overhead when reading dataframes from ArcticDB

alexowens90:
#### Reference Issues/PRs
[9106978040](https://man312219.monday.com/boards/7852509418/pulses/9106978040)

Prior to this change:

1. The `SegmentInMemory` returned by calls to `read` (and similar) was wrapped in a Python object, which was then used as the base object for each individual `ndarray`. This meant that even if 99% of the arrays were freed, the entire `SegmentInMemory`  (i.e. all columns) would not be freed until the final array was freed.
2. The buffers underlying this `SegmentInMemory` were allocated as `presized`. In practice, this meant for small buffers that there was a lot of overhead taken up with metadata and page-alignment.

These issues are addressed here by:

1. Giving each `ndarray` its own base object (the new class `NumpyBufferHolder`) referencing only the memory needed for that individual array.
2. Using the new (from Arrow work) `detachable` allocation type for buffers that will eventually be used to back `ndarrays`. This removes the page alignment, and allows all of the metadata around our buffer class to be freed, leaving just the information necessary to decref any objects necessary when the `ndarray` is destroyed, and then free the memory.

This has also eliminated the need for ping-ponging off the Python layer, and massively simplified the memory management of `PandasOutputFrame` and `FrameData`, and completely removing `FrameDataWrapper`. The model for Pandas is now almost identical to that with Arrow.

Change tested by writing data to an S3-backed library using:
```
def write_data():
    lib.version_store.clear()
    num_rows = 1_000
    num_columns = 10_000
    df = pd.DataFrame(np.random.randint(0, 100, size=(num_rows, num_columns)), index=pd.date_range("2025-01-01", periods=num_rows))
    lib.write("sym", df)
```
and then reading it back using:
```
def read_data():
    date_range = (pd.Timestamp("2025-01-01"), pd.Timestamp("2025-01-04"))
    retrieved = []

    for i in range(10):
        small = lib.read("sym", query_builder=QueryBuilder().date_range(date_range)).data
        retrieved.append(small)
        print(f"RSS: {psutil.Process(os.getpid()).memory_info().rss / 1e6} MB")
```
Results:
```
5.8.2  5.9.0  Branch
  345    313     313
  395    321     314
  443    330     315
  492    339     316
  539    346     318
  588    355     318
  636    362     320
  685    371     321
  734    380     322
  785    388     322
```
The difference between 5.8.2 and 5.9.0 is down to https://github.com/man-group/ArcticDB/pull/2404


cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -14,16 +14,55 @@
 
 namespace arcticdb {
 
+sparrow::array empty_arrow_array_from_type(const TypeDescriptor& type, std::string_view name) {
```
IvoDD: Note to self: Changes in this file should be none after rebase.
Similar for other files like `chunked_buffer.hpp`, `query.hpp` and many others.

cpp/arcticdb/column_store/memory_segment_impl.cpp
```diff
@@ -96,8 +96,17 @@ void SegmentInMemoryImpl::create_columns(
     for (size_t i = old_size; i < size_t(descriptor_->field_count()); ++i) {
         auto type = descriptor_->fields(i).type();
         util::check(type.data_type() != DataType::UNKNOWN, "Can't create column in create_columns with unknown data type");
-        columns_.emplace_back(
-                std::make_shared<Column>(descriptor_->fields(i).type(), expected_column_size, presize, allow_sparse, output_format, mode));
+        if (presize == AllocationType::DETACHABLE && is_fixed_string_type(descriptor_->fields(i).type().data_type())) {
```
IvoDD: Maybe not for this PR but now that `AllocationType` is not only `DYNAMIC` and `PRESIZED` but also includes `DETACHABLE` we should rename `presize` to `allocation_type`.

cpp/arcticdb/processing/test/benchmark_ternary.cpp
```diff
@@ -1,3 +1,6 @@
+
+
+
```
IvoDD: Nit: Empty lines at beginning of file

cpp/arcticdb/python/numpy_buffer_holder.hpp
```diff
@@ -0,0 +1,58 @@
...
+        other.type_ = make_scalar_type(DataType::UNKNOWN);
+        other.ptr_ = nullptr;
+        other.row_count_ = 0;
+    }
+
+    ~NumpyBufferHolder() {
+        if (ptr_) {
+            if (is_dynamic_string_type(type_.data_type())) {
+                if (is_object_type(type_)) {
+                    if (is_array_type(type_) || !is_arrow_output_only_type(type_)) {
```
IvoDD: I find this sequence of nested `if`s confusing. 

Both of the below have been issues with the previous implementation inside `pandas_output_frame` but I think now that we're reworking it it is a good time to fix. 

- the `is_object_type` check is obsolete as it will always be true when `is_dynamic_string_type` is `True`.
- I think we should not allow `is_arrow_output_only_type=True` even in the case for `is_array_type=True`. So we can change the last if to only be `if (!is_arrow_output_only_type)`.

cpp/arcticdb/python/numpy_buffer_holder.hpp
```diff
@@ -0,0 +1,58 @@
...
+                        // pybind11 has taken the GIL for us so this is safe
+                        auto py_ptr = reinterpret_cast<PyObject**>(ptr_);
+                        for (size_t idx = 0; idx < row_count_; ++idx, ++py_ptr) {
```
IvoDD: Just to confirm: Can a user on python side modify the `NumpyBufferHolder`s we give them?

E.g. could a user truncate the columns themselves in a way so that numpy handles the deletion of the last several rows? Then we would still try to decref all `row_count_` values even though some of them were deleted by numpy.

cpp/arcticdb/column_store/column_utils.hpp
```diff
@@ -115,41 +107,20 @@ inline py::array array_at(const SegmentInMemory& frame, std::size_t col_pos, Out
...
-                if(!is_py_none(ptr))
-                    api.PyArray_SetBaseObject_(ptr, anchor.inc_ref().ptr());
-
-                it.next();
+            auto py_ptr = reinterpret_cast<PyObject**>(ptr);
+            for (size_t idx = 0; idx < frame.row_count(); ++idx, ++py_ptr) {
+                util::check(py_ptr != nullptr, "Can't set base object on null item");
+                if(!is_py_none(*py_ptr)) {
+                    api.PyArray_SetBaseObject_(*py_ptr, base_obj.inc_ref().ptr());
+                }
```
IvoDD, line 115: How will this work for sparse data? 
I think in this case `ptr` will only point to the densified column and thus iterating to `frame.row_count()` will reach out of allocated memory.

python/arcticdb/version_store/_normalization.py
```diff
@@ -1369,12 +1365,13 @@ def denormalize_user_metadata(udm, ext_obj=None):
 
 
 def denormalize_dataframe(ret):
-    read_result = ReadResult(*ret)
-    frame_data = FrameData.from_cpp(read_result.frame_data)
+    pandas_output_frame = ret[1]
+    norm = ret[2]
+    frame_data = FrameData(*pandas_output_frame.extract_numpy_arrays())
```
IvoDD, line 1370: Why not use the `ReadResult` constructor?
It will give you the constructed `frame_data` and `norm` in a more readable way.

python/arcticdb/version_store/_store.py
```diff
@@ -2066,9 +2067,9 @@ def _post_process_dataframe(self, read_result, read_query, implement_read_index=
             # post filter
             start_idx, end_idx = self._compute_filter_start_end_row(read_result, read_query)
             data = []
-            for c in read_result.frame_data.value.data:
+            for c in read_result.frame_data.data:
                 data.append(c[start_idx:end_idx])
-            read_result.frame_data = FrameData(data, read_result.frame_data.names, read_result.frame_data.index_columns)
+            read_result.frame_data = FrameData(data, read_result.frame_data.names, read_result.frame_data.index_columns, read_result.frame_data.row_count, read_result.frame_data.offset)
```
IvoDD: I don't think the `row_count` will be used after this point, but what do you think about setting the new `row_count=end_idx-start_idx`?

IvoDD (commented):
Most of the comments are minor. Only the one for sparse data inside the `array_at` construction is imporant.

Apart from those what do you think about:
- Adding asv peakmem tests similar to the tests you ran for the peakmem description?
- Now that all columns can be deallocated separately it would also be useful to test that if a user removes a subset of columns from returned dataframe we should also see memory decrease. Maybe we can test that as well, maybe again with an asv peakmem test?
- I'm slightly worried about any performance impact we might get from the switch from `PRESIZED` to `DETACHABLE`. Maybe not worth adding explicit C++ tests for this but paying careful attention to ASV results.

cpp/arcticdb/python/numpy_buffer_holder.hpp
alexowens90: Agreed, I copied it from the destructor of `PandasOutputFrame`, but it was more complex there to handle array columns having a different underlying buffer structure.

cpp/arcticdb/python/numpy_buffer_holder.hpp
alexowens90: I don't think so. The way our Python post-processing with `date_range/row_range` works is to use the [slice operator](https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/version_store/_store.py#L2082), which returns a view without modifying the underlying object.
Any other kind of truncation would necessitate a copy, which will `incref` all of the contained objects, and then it is correct that we still `decref` them all on destruction

cpp/arcticdb/column_store/column_utils.hpp
alexowens90: By this stage, all sparse data has been inflated into dense data

python/arcticdb/version_store/_normalization.py
alexowens90: Introduces a circular dependency

cpp/arcticdb/python/numpy_buffer_holder.hpp
```diff
@@ -0,0 +1,56 @@
...
+        other.row_count_ = 0;
+    }
+
+    ~NumpyBufferHolder() {
+        if (ptr_) {
+            if (is_dynamic_string_type(type_.data_type())) {
+                if (is_arrow_output_only_type(type_)) {
+                    log::version().error("Unexpected arrow output format seen in NumpyBufferHolder");
+                } else {
+                    // pybind11 has taken the GIL for us so this is safe
```
poodlewars, line 40: Worth a `PyGILState_Check()` assertion, at least for debug builds?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -849,7 +849,22 @@ struct ReduceColumnTask : async::BaseTask {
...
-                column.default_initialize_rows(0, frame_.row_count(), false);
+                if (is_fixed_string_type(field_type)) {
+                    // Special case where we have a fixed-width string column that is all null (e.g. dynamic schema
+                    // where this column was not present in any of the read row-slices)
+                    // All other column types are allocated in the output frame as detachable by default since we know
+                    // we will be handing them off to Python to free. This is not the case for fixed-width string
+                    // columns, since the buffer still contains string pool offsets at this point, and so will usually
+                    // be swapped out with the inflated strings buffer. However, if there are no strings to inflate, we
+                    // still need to make this buffer detachable and full of nulls of the correct length
+                    auto buffer_size = frame_.row_count() * (field_type == DataType::UTF_FIXED64 ? 4 : 1);
```
poodlewars, line 860: I presume the 1 case only applies to `ASCII_FIXED64`? Worth an assertion here?

PR #2437, opened 2025-07-02
Update_batch additional tests

grusev:
#### Any other comments?

Included are 3 integration tests covering update_batch with all supported types of dataframes and different type of updates over symbol. Covers both static and dynamic schema with checks types promotion. Updates on static schema are also tested against symbols spreading accross several segments.

All tests also cover real storages.

Several utilities have also been added to help integration scenarios testing with update and update_batch in future. 

A small set of ASV tests were also developed to keep track of time and mem

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_with_conda.yml
+++ b/.github/workflows/build_with_conda.yml
@@ -130,13 +130,10 @@ jobs:
 
 
   macos:
-    strategy:
-      matrix:
-        include:
-          - os: macos-14
-
-      fail-fast: false
-    runs-on: ${{ matrix.os }}
+    if: |
+      always() &&
+      !cancelled()
+    runs-on: macos-14
     env:
       ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
     steps:
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -555,6 +555,9 @@ set(arcticdb_srcs
 
 add_library(arcticdb_core_object OBJECT ${arcticdb_srcs})
 
+# Disable default api so both UTF-8 and UTF-32 can be supported
+target_compile_definitions(arcticdb_core_object PUBLIC PCRE2_CODE_UNIT_WIDTH=0)
+
 if(${ARCTICDB_USE_PCH})
     message(STATUS "Using precompiled headers for target arcticdb_core_object")
     set(BASE_PCH
@@ -794,6 +797,9 @@ target_include_directories(arcticdb_core_object
 
 add_library(arcticdb_core_static STATIC $<TARGET_OBJECTS:arcticdb_core_object>)
 
+# Disable default api so both UTF-8 and UTF-32 can be supported
+target_compile_definitions(arcticdb_core_static PUBLIC PCRE2_CODE_UNIT_WIDTH=0)
+
 target_include_directories(arcticdb_core_static PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
         $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
@@ -869,7 +875,6 @@ pybind11_add_module(arcticdb_ext MODULE
         python/arctic_version.cpp
         python/arctic_version.hpp
         )
-
 set (additional_link_flags "")
 if (HIDE_LINKED_SYMBOLS AND (NOT WIN32))
 
@@ -881,7 +886,6 @@ if (HIDE_LINKED_SYMBOLS AND (NOT WIN32))
 endif()
 
 target_link_libraries(arcticdb_ext
-        PUBLIC
         PRIVATE
         arcticdb_python
         ${additional_link_flags}
@@ -1130,6 +1134,7 @@ if(${TEST})
             version/test/rapidcheck_version_map.cpp)
 
     add_executable(arcticdb_rapidcheck_tests ${rapidcheck_srcs})
+
     install(TARGETS arcticdb_rapidcheck_tests RUNTIME
                 DESTINATION .
                 PERMISSIONS ${EXECUTABLE_PERMS}
@@ -1183,7 +1188,6 @@ if(${TEST})
     endif()
 
     gtest_discover_tests(arcticdb_rapidcheck_tests PROPERTIES DISCOVERY_TIMEOUT 60)
-
     if(${ARCTICDB_USE_PCH})
         message(STATUS "Using precompiled headers for target test_unit_arcticdb")
         target_precompile_headers(
--- a/cpp/arcticdb/util/regex_filter.hpp
+++ b/cpp/arcticdb/util/regex_filter.hpp
@@ -7,7 +7,10 @@
 
 #pragma once
 
-#define PCRE2_CODE_UNIT_WIDTH 0 //Disable default api so both UTF-8 and UTF-32 can be supported
+#if !defined(PCRE2_CODE_UNIT_WIDTH) || PCRE2_CODE_UNIT_WIDTH != 0
+#error "ArcticDB requires PCRE2_CODE_UNIT_WIDTH to be 0"
+#endif
+
 #include <pcre2.h>
 #include <boost/locale.hpp>
 #include <arcticdb/util/constructors.hpp>
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -662,7 +662,7 @@ def version_store_factory(lib_name, lmdb_storage) -> Generator[Callable[..., Nat
     # Otherwise there will be no storage space left for unit tests
     # very peculiar behavior for LMDB, not investigated yet
     # On MacOS ARM build this will sometimes hang test execution, so no clearing there either
-    yield from _store_factory(lib_name, lmdb_storage, not (WINDOWS or MACOS_WHEEL_BUILD))     
+    yield from _store_factory(lib_name, lmdb_storage, not (WINDOWS or MACOS))     
 
 
 @pytest.fixture
--- a/python/tests/integration/arcticdb/version_store/test_basic_version_store.py
+++ b/python/tests/integration/arcticdb/version_store/test_basic_version_store.py
@@ -47,7 +47,7 @@ from arcticdb.util.test import (
 from tests.util.date import DateRange
 from arcticdb.util.test import equals
 from arcticdb.version_store._store import resolve_defaults
-from tests.util.mark import MACOS_WHEEL_BUILD
+from tests.util.mark import MACOS, MACOS_WHEEL_BUILD
 
 
 @pytest.fixture()
@@ -101,7 +101,7 @@ def test_special_chars(object_version_store):
 
 @pytest.mark.parametrize("breaking_char", [chr(0), "\0", "*", "<", ">"])
 @pytest.mark.storage
-@pytest.mark.skipif(MACOS_WHEEL_BUILD, reason="Halts execution on MacOS ARM")
+@pytest.mark.skipif(MACOS, reason="Halts execution on MacOS ARM")
 def test_s3_breaking_chars(object_version_store, breaking_char):
     """Test that chars that are not supported are raising the appropriate exception and that we fail on write without
     corrupting the db.
@@ -116,7 +116,7 @@ def test_s3_breaking_chars(object_version_store, breaking_char):
 
 @pytest.mark.parametrize("breaking_char", [chr(0), "\0", "*", "<", ">"])
 @pytest.mark.storage
-@pytest.mark.skipif(MACOS_WHEEL_BUILD, reason="Halts execution on MacOS ARM")
+@pytest.mark.skipif(MACOS, reason="Halts execution on MacOS ARM")
 def test_s3_breaking_chars_staged(object_version_store, breaking_char):
     """Test that chars that are not supported are raising the appropriate exception and that we fail on write without
     corrupting the db.
@@ -131,7 +131,7 @@ def test_s3_breaking_chars_staged(object_version_store, breaking_char):
 
 @pytest.mark.parametrize("unhandled_char", [chr(0), chr(30), chr(127), chr(128)])
 @pytest.mark.storage
-@pytest.mark.skipif(MACOS_WHEEL_BUILD, reason="Halts execution on MacOS ARM")
+@pytest.mark.skipif(MACOS, reason="Halts execution on MacOS ARM")
 def test_unhandled_chars_default(object_version_store, unhandled_char):
     """Test that by default, the problematic chars are raising an exception"""
     sym = f"prefix{unhandled_char}postfix"
@@ -144,7 +144,7 @@ def test_unhandled_chars_default(object_version_store, unhandled_char):
 
 @pytest.mark.parametrize("sym", [chr(0), chr(30), chr(127), chr(128), "", "l" * 255])
 @pytest.mark.storage
-@pytest.mark.skipif(MACOS_WHEEL_BUILD, reason="Halts execution on MacOS ARM")
+@pytest.mark.skipif(MACOS, reason="Halts execution on MacOS ARM")
 def test_unhandled_chars_staged_data(object_version_store, sym):
     """Test that by default, the problematic chars are raising an exception at staging time."""
     df = sample_dataframe()
--- a/python/tests/unit/arcticdb/version_store/test_column_type_changes.py
+++ b/python/tests/unit/arcticdb/version_store/test_column_type_changes.py
@@ -17,7 +17,7 @@ from arcticdb_ext.storage import KeyType
 from arcticdb.util.test import assert_frame_equal
 from arcticdb_ext.types import DataType
 
-from tests.util.mark import MACOS_WHEEL_BUILD
+from tests.util.mark import MACOS, MACOS_WHEEL_BUILD
 
 
 @pytest.mark.parametrize("dynamic_schema", [True, False])
@@ -220,7 +220,7 @@ def test_type_promotion_int32_and_float32_up_to_float64(lmdb_version_store_dynam
     assert data.dtypes["a"] == np.float64
     assert expected_result.dtypes["a"] == np.float64
 
-
+@pytest.mark.xfail(MACOS, reason="bug??? https://github.com/man-group/ArcticDB/actions/runs/16517098026/job/46710177373?pr=2506")
 def test_type_promotion_int64_and_float64_up_to_float64(lmdb_version_store_dynamic_schema):
     """We unavoidably lose precision in this case, this test just shows what happens when we do."""
     # Given


python/arcticdb/util/test.py
```diff
@@ -420,8 +420,11 @@ def populate_db(version_store):
 
 def random_integers(size, dtype, min_value: int = None, max_value: int = None):
     # We do not generate integers outside the int64 range
-    platform_int_info = np.iinfo("int_")
     iinfo = np.iinfo(dtype)
+    if "uint" not in str(dtype):
+        platform_int_info = np.iinfo("int_")
+    else:
+        platform_int_info = iinfo
```
vasil-pashov: I don't why this is needed. Why is `iinfo = np.iinfo(dtype)` not enough?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+        freq = TimestampNumber.DEFAULT_FREQ if isinstance(start_time, Timestamp) else start_time.get_type()
+
+        cols=int(number_columns)
+        rows=int(number_rows)
+        upgradable_dtypes = [np.int8, np.int16, np.int32, np.uint16, np.uint32, np.float32]
+        gen = DFGenerator(size=rows, seed=self.seed) 
+        for i in range(cols):
+                dtype = upgradable_dtypes[i % len(upgradable_dtypes)]
+                desired_type = self._resolve(dtype)
+                if 'int' in str(desired_type):
```
vasil-pashov: numpy has `issubdtype` method. Use that instead of comparing strings,

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+            raise ValueError(f"Invalid update position type: {position_type}")
+        return update_df
+    
+
+def read_batch(lib: Library, symbol_names: List[str]) -> Dict[str, VersionedItem]:
+    read_results = lib.read_batch(symbol_names)
+    read_data = dict()
+    for result in read_results:
+        read_data[result.symbol] = result
+    return read_data
```
vasil-pashov: dict comprehension will make this in one line

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+        elif position_type == UpdatePositionType.RIGHT_AFTER:
+            update_df = self.generator.get_dataframe(number_cols, number_rows, end.inc(1))
+        elif position_type == UpdatePositionType.AFTER:
+            update_df = self.generator.get_dataframe(number_cols, number_rows, end.inc(2))
+        else:
+            raise ValueError(f"Invalid update position type: {position_type}")
+        return update_df
+    
+
+def read_batch(lib: Library, symbol_names: List[str]) -> Dict[str, VersionedItem]:
```
vasil-pashov: Type hint is wrong. The result of batch read is either VersionedItem or DataError. I'd also change the name so that it's not exactly as in the library API, I believe it's easier to read this way.

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+    The generator produces only dataframes with columns which types can be upgraded
+    with ArcticDB
+    """
+
+    def __init__(self, seed=None):
+        super().__init__()
+        self._type_conversion_dict: Dict[type, type] = self.no_upgrade_types()
+        self.seed = seed
+
+    def define_upgrade_types(self, mappings: Dict[type, type]):
```
vasil-pashov: I think that either both define_upgrade_types and no_upgrade_types should return or none of them should return.

I haven't seen type as a type hint. What does it do?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+    for a specified number of columns. But then via type mapping dictionary
+    the next generation could be forced instead of int32 to generate int64
+    for instance. 
+    The generator produces only dataframes with columns which types can be upgraded
+    with ArcticDB
+    """
+
+    def __init__(self, seed=None):
+        super().__init__()
+        self._type_conversion_dict: Dict[type, type] = self.no_upgrade_types()
```
vasil-pashov: The first line of self.no_upgrade_types() sets it to empty dict no need to assign it again after that

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+
+    def _resolve(self, base_type):
+        return self._type_conversion_dict.get(base_type, base_type)
+
+    def get_dataframe(self, number_columns:int, number_rows: int, 
+                      start_time: Union[Timestamp, TimestampNumber] = Timestamp("2033-12-11"), **kwargs) -> pd.DataFrame:
+        freq = TimestampNumber.DEFAULT_FREQ if isinstance(start_time, Timestamp) else start_time.get_type()
+
+        cols=int(number_columns)
+        rows=int(number_rows)
```
vasil-pashov: Why are those explicitly cast to int? Either the type hint is wrong, or there's a bug in the code passing non-ints to the function. In any case we should find the root cause.

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+        for i in range(cols):
+                dtype = upgradable_dtypes[i % len(upgradable_dtypes)]
+                desired_type = self._resolve(dtype)
+                if 'int' in str(desired_type):
+                    gen.add_int_col(f"col_{i}", desired_type)
+                    pass
+                elif 'float' in str(desired_type):
+                    gen.add_float_col(f"col_{i}", desired_type)
+                else:
+                    return f"Unsupported type {dtype}"
```
vasil-pashov: Why are we returning string here? I don't see anything handling that. It makes the typehint wrong. This should raise exception.

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+        """
+        self.generator: DataFrameGenerator = gen
+
+    def generate_sequence(self, original_dataframe: pd.DataFrame, number_cols:int,  number_rows:int) -> List[pd.DataFrame]:
+        """
+        Generates sequence of updates covering all updates type.
+        """
+        sequence: List[pd.DataFrame] = list()
+
+        for count, position_type in enumerate(UpdatePositionType):
```
vasil-pashov: Count is not used. Use comprehension

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+NUMBER_ROW_SEGMENTS = 10
+NUMBER_COL_SEGMENTS = 10
+DEFAULT_FREQ = 's'
+
+
+class UpdatePositionType(Enum):
+    '''Enum specifying relative position of an update start or end time relative to
+    original dataframe start and end time'''
+    BEFORE = 0 
+    RIGHT_BEFORE = 1
```
vasil-pashov: I think before and right before need some comment or different name. It looks to me that both of those will generate data that does not overlap with the dataframe on disk.

I don't see any options that have some data outside of the on disk and then overlaps with either the beginning or the end e.g.
on: disk [5, 10]
update: [3, 7]
update: [7, 13]
I also don't see anything that's larger than the dataframe on disk
update: [3, 13]

Is this intentional?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+        end = TimestampNumber.from_timestamp(original_dataframe.index[-1], DEFAULT_FREQ)
+        rows = original_dataframe.shape[0]
+        if position_type == UpdatePositionType.BEFORE:
+            update_df = self.generator.get_dataframe(number_cols, number_rows, start.dec(number_rows+1))
+        elif position_type == UpdatePositionType.RIGHT_BEFORE:
+            update_df = self.generator.get_dataframe(number_cols, number_rows, start.dec(number_rows))
+        elif position_type == UpdatePositionType.INSIDE_OVERLAP_BEGINNING:
+            update_df = self.generator.get_dataframe(number_cols, number_rows, start)
+        elif position_type == UpdatePositionType.INSIDE:
+            to_update = number_rows if (number_rows + 2) <= rows else rows - 2
```
vasil-pashov, line 198: Why do we check if number_rows + 2 <= rows? I think this needs a comment

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+    for update in update_batch:
+        start_index = original_dataframes[update.symbol].index[0] - pd.Timedelta(days=1)
+        previous_day = start_index - pd.Timedelta(days=1)
+        update.date_range = (previous_day, start_index)
+    update_result = lib.update_batch(update_batch, prune_previous_versions=True)
+    read_data = read_batch(lib, symbol_names)
+    for index, result in enumerate(update_result):
+        symbol = symbol_names[index]
+        if not hasattr(result, "version"):
+            # To catch rare problem on 3.8
```
vasil-pashov, line 421: Elaborate on this. Does this happen only on python 3.8? Is it reproducible or flaky? Is it a known issue. Have we done any investigation on that. It seems pretty odd to have an attribute randomly missing from an object. Is this related to batch_update or it can happen on any call?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+        symbol = symbol_names[index]
+        assert result.version == 6
+        assert_frame_equal(expected_results[symbol], read_data[symbol].data)
+
+
+def dataframe_simulate_arcticdb_update_dynamic(expected_df: pd.DataFrame, update_df: pd.DataFrame) -> pd.DataFrame:
+    """ Simulates partially dynamic update, when update df has new columns.
+    The scenario where there are type promotions of existing columns is not covered yet
+    """
+    def add_missing_ending_columns(to_df: pd.DataFrame, from_df: pd.DataFrame) -> pd.DataFrame:
```
vasil-pashov: Why is this a nested function. Also why is the return value ignored?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,501 @@
...
+    INSIDE_OVERLAP_END = 6
+    RIGHT_AFTER = 7
+    AFTER = 8
+
+class DataFrameGenerator(ABC):
+
+    @abstractmethod
+    def get_dataframe(self, number_columns:int, number_rows: int, 
+                      start_time: Union[Timestamp, TimestampNumber], **kwargs) -> pd.DataFrame:
+        pass
```
vasil-pashov: I don't see how we benefit from such abstraction. Python lists are heterogeneous, this doesn't add any function that can be reused by the derived classes, the check if get_dataframe exists is at runtime. It looks like Java translated to Python.

python/arcticdb/util/test.py
grusev: that is a common function, which was not working on uints. Thus in order to make it work for uints the else was added, the other part is preserved so it behaves as it used to and no unexpected problems arise. Thus I cannot comments exactly why np.iinfo("int_") was chosen.

python/tests/integration/arcticdb/test_update.py
grusev: now def read_batch_as_dict(lib: Library, symbol_names: List[str]) -> Dict[str, Union[VersionedItem, DataError]]:

python/tests/integration/arcticdb/test_update.py
grusev: Python cannot enforce type hints, thus hints are only recommendations not rule, so those 2 calls should be ok in this case where this rule is enforced no matter what is passed

python/tests/integration/arcticdb/test_update.py
grusev: Definately, I planned it for raise TypeError("Unsupported type {dtype}") but forgot to do it. Thanks!

python/tests/integration/arcticdb/test_update.py
grusev: Great catch! You are correct, those 3 indeed are missing.

class UpdatePositionType(Enum):
    '''Enum specifying relative position of an update start or end time relative to
    original dataframe start and end time'''
    BEFORE = 0  # End of the update is much before start of original DF
    RIGHT_BEFORE = 1  # End of the update is exactly before start of original DF
    BEFORE_OVERLAP_START = 2 # End of the update overlaps with the start of the original DF
    INSIDE_OVERLAP_BEGINNING = 3 # Start of update overlaps with the start of the original DF, no restrictions on end
    TOTAL_OVERLAP = 4 # Exact same start and end and number of rows as the original DF
    OVERSHADOW_ORIGINAL = 5 # Starts before start of original and ends after the original DF
    INSIDE = 6 # literally inside the original DF, update does not overlap neither start nor end
    INSIDE_OVERLAP_END = 7 # end of both update and original overlap, no restrictions for start
    AFTER_OVERLAP_END = 8 # end of original overlaps with start of of update
    RIGHT_AFTER = 9 # start of update is exactly next to original
    AFTER = 104 # start of the update is after at least 1 duration of end of original
    
Added:
    - BEFORE_OVERLAP_START 
    - OVERSHADOW_ORIGINAL 
    - AFTER_OVERLAP_END 
    
The most critical point of overlap is always overlap with one row at beging and at end and all that is covered now.

python/tests/integration/arcticdb/test_update.py
grusev: I think I documented that  now - the idea is that the generated DF must be inside physically in the original one, thus the returned generated row may have different than requested  from used number of rows

python/tests/integration/arcticdb/test_update.py
grusev: this is a bug escape I currently investigate - will log issue upon it once all details are relevealed but looks like we do not handle in 3.8 one case

ERROR:tests.integration.arcticdb.test_update:Expected Version UNKNOWN of symbol 'some heck of a symbol!.!0' unable to be retrieved: Error Category: 'UNKNOWN' Exception String: 'arcticdb::ArcticCategorizedException<(arcticdb::ErrorCategory)4>: E_DESCRIPTOR_MISMATCH No valid common type between TD<type=FLOAT64, dim=0> and TD<type=UTF_DYNAMIC64, dim=0> for column col_2' with version attribute <arcticdb_ext.version_store.DataError object at 0x7f8594b7cef0>

python/tests/integration/arcticdb/test_update.py
grusev: it is private to dataframe_simulate_arcticdb_update_dynamic function, not intended to be reused.
The plan is at certain point dataframe_simulate_arcticdb_update_dynamic to be full simulator of Arctic logic in updates
removed returned value and added doc. Basically the to_df is modified so no reason of returning value

python/tests/integration/arcticdb/test_update.py
grusev: This is a standard OOP pattern for interface, as python supports OOP then it should be ok also. Moreover this exact approach was is used already in ASV real tests. So I used it here also, What would be be then more pythonic way to express that certain class supports this specific method which is basis of update generation?

python/tests/integration/arcticdb/test_update.py
grusev: there is an issue logged now and also the part that was failing is now xfailed - Linux 3.8 specific but underlines some logic weakness

python/arcticdb/util/test.py
vasil-pashov: This is working in both numpy 1 and numpy 2
```
>>> import numpy as np
>>> np.iinfo("uint32")
iinfo(min=0, max=4294967295, dtype=uint32)
>>> np.iinfo("uint16")
iinfo(min=0, max=65535, dtype=uint16)
```

python/tests/integration/arcticdb/test_update.py
vasil-pashov: What's the point of having explicit contract via type hints if they are not respected. My point is that if something is not passing int to this function it's a bug and if we decide to allow anything to be passed the type hint useless

python/arcticdb/util/test.py
vasil-pashov: Also avoid string checks in dtypes. Prefer the built in functions of either numpy or pandas to check the properties of the dtype.

python/tests/integration/arcticdb/test_update.py
vasil-pashov: I think some examples will be useful, IMO RIGHT_BEFORE and BEFORE_OVERLAP_START are a bit ambiguous. Is this correct assuming df with index [5, 10]
* RIGHT_BEFORE -> [2, 4]
* BEFORE_OVERLAP_START -> [2, 5]

python/tests/integration/arcticdb/test_update.py
vasil-pashov: The pythonic way not to have abstract class at all and use duck typing. Python is multi-paradigm language but just because it supports multiple paradigms doesn't mean we have to use all of them.

python/tests/integration/arcticdb/test_update.py
grusev: not sure what is the benefit of doing that. This works for int and uint, if I do the change there will be at least one elif more

vasil-pashov: Because relying on strings is error prone and does not provide type safety and is less obvious what you're trying to do. You can use `np.issubdtype(dtype, np.integer)` to check if something is any int or uint type.

python/arcticdb/util/test.py
```diff
@@ -244,10 +244,13 @@ def assert_frame_equal_rebuild_index_first(expected: pd.DataFrame, actual: pd.Da
     First will rebuild index for dataframes to assure we
     have same index in both frames when row range index is used
     """
-    if PANDAS_VERSION < CHECK_FREQ_VERSION:
-        if expected.shape[0] == expected.shape[0] and expected.shape[0] == 0:
-            assert expected.columns.equals(expected.columns)    
-            return
+    if IS_PANDAS_ONE:
+        if (("object" in str(expected.dtype)) and ("float" in str(actual.dtype)) or
```
vasil-pashov: Use np.issubdtype/np.isdtype

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,564 @@
...
+        cols=int(number_columns)
+        rows=int(number_rows)
+        upgradable_dtypes = [np.int8, np.int16, np.int32, np.uint16, np.uint32, np.float32]
+        gen = DFGenerator(size=rows, seed=self.seed) 
+        for i in range(cols):
+                dtype = upgradable_dtypes[i % len(upgradable_dtypes)]
+                desired_type = self._resolve(dtype)
+                if 'int' in str(desired_type):
+                    gen.add_int_col(f"col_{i}", desired_type)
+                    pass
```
vasil-pashov: Why?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,564 @@
...
+        ], indirect=True)
+@pytest.mark.only_fixture_params(["lmdb", "real_s3", "real_gcp"])
+def test_update_batch_all_supported_datatypes_over_several_segments(custom_library):
+    '''
+    Test assures that update batch works with all supported datatypes, 
+    updates work over several segments of the library and executing several times
+    same updates does not alter the result
+    '''
+    lib: Library = custom_library
+    seed(SEED)
```
vasil-pashov: Is the use of different seed here and in the next test intentional and why?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,564 @@
...
+from tests.util.mark import LINUX
+
+
+logger = get_logger()
+
+SEED = 1001
+
+
+NUMBER_ROW_SEGMENTS = 10
+NUMBER_COL_SEGMENTS = 10
```
vasil-pashov:
```suggestion
ROWS_PER_SEGMENT = 10
COLS_PER_SEGMENTS = 10
```
Would be more appropriate I think. In Arctic's vocabulary a segment is comprised of both rows and columns. So row segment doesn't make sense.

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,564 @@
...
+    for index, result in enumerate(update_result):
+        symbol = symbol_names[index]
+        assert result.version == 4 # This will become 6 when uncommented above once bug is fixed
+        assert_frame_equal(expected_results[symbol], read_data[symbol].data)
+
+
+
+@pytest.mark.xfail(LINUX and (sys.version_info[:2] == (3, 8)), 
+                   reason = "update_batch return unexpected exception (9589648728)",
+                   strict=False)
```
vasil-pashov: Is this a flaky test? Or is it 100% reproducible on the given platforms? If it's reproducible it shouldn't be marked as strict=False

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,564 @@
...
+    df_0col = df[0:0]
+    lib.write_batch([WritePayload(symbol, df)])
+    update = UpdatePayload(symbol, df_0col)
+    update_result = lib.update_batch([update], prune_previous_versions=True)
+    assert update_result[0].version == 1
+
+
+@pytest.mark.xfail(LINUX and (sys.version_info[:2] == (3, 8)), 
+                   reason = "update_batch return unexpected exception (9589648728)",
+                   strict=False)
```
vasil-pashov: Same as https://github.com/man-group/ArcticDB/pull/2437/files#r2222559649

python/tests/integration/arcticdb/test_update.py
```diff
@@ -0,0 +1,564 @@
...
+    ug = UpdatesGenerator(g)
+    original_num_cols = 30
+    original_number_rows = 50
+    symbol_prefix = "different types of updates"
+
+    symbol_names = []
+    update_batch:List[UpdatePayload] = []
+    write_batch:List[UpdatePayload] = []
+    expected_results = dict()
+    dataframes_lenghts = [1, original_number_rows // 2, original_number_rows, original_number_rows * 1.2]
```
vasil-pashov, line 537: Is column reordering handled in any test here or in the existing unit tests. E.g.
```python
lib.write(pd.DataFrame({"a": [1], "b": [1]}))
lib.write(pd.DataFrame({"b": [1], "a": [1]}))
```
it would be good to test this property of dynamic schema if it's not already.

python/tests/integration/arcticdb/test_update.py
vasil-pashov: I don't think type works in type hint? type is a python function so it's preserved, you probably want to specify that this is a dtype but I don't think there's a typehint for that. You can check numpy's docs to be sure.

python/tests/integration/arcticdb/test_update.py
grusev: This is something worth covering, but not here at update_batch tests, I would rather cover it here: 

https://github.com/man-group/ArcticDB/pull/2470/

python/tests/integration/arcticdb/test_update.py
grusev: It is perfectly fine to be there and not be there. Since I have no sentiment I will remove them

python/tests/integration/arcticdb/test_update.py
grusev: yes.

grusev: very good suggestion

added:

    Example:
    If we have a dataframe with index [5, 10]. The following would be true:

        BEFORE -> [1, 3]
        RIGHT_BEFORE -> [2, 4]
        BEFORE_OVERLAP_START -> [2, 5]
        INSIDE_OVERLAP_BEGINNING -> [5, 7]
        TOTAL_OVERLAP -> [5, 10]
        INSIDE -> [6,8]
        etc .... ()

python/tests/integration/arcticdb/test_update.py
grusev: removed pass

python/tests/integration/arcticdb/test_update.py
grusev: not everyhting was clear for this xfail, the error was happening on 3.8 and on linux - hence the initial xfail.

Later it turned out that the reason is Pandas 1 behavior - see the issue loged - there are api which work differently than others in arcticdb

@pytest.mark.xfail(IS_PANDAS_ONE, 
                   reason = "update_batch return unexpected exception (9589648728)")

python/tests/integration/arcticdb/test_update.py
grusev: solved

python/tests/integration/arcticdb/test_update.py
grusev: not intentitionally, just ommission, now all tests use same seed.

python/tests/integration/arcticdb/test_update.py
grusev: done

vasil-pashov (approved):

PR #2441, opened 2025-07-03
Implement summming of bools as uint64 instead of bool

vasil-pashov:
#### Reference Issues/PRs

Prior this commit when sum aggregation was performed on bool column the result would be of bool type. The result was computed by applying OR on all values in the group. This not what pandas (and polars) do.

With this new implementation the sum of bools will equal to the number of True values in the group and the dtype will be uint64.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/entity/types.hpp
+++ b/cpp/arcticdb/entity/types.hpp
@@ -191,6 +191,7 @@ constexpr uint8_t combine_val_bits(ValueType v, SizeBits b = SizeBits::UNKNOWN_S
 
 } // namespace anonymous
 
+// When adding DataType here add it to the all_data_types function
 enum class DataType : uint8_t {
     UINT8 = detail::combine_val_bits(ValueType::UINT, SizeBits::S8),
     UINT16 = detail::combine_val_bits(ValueType::UINT, SizeBits::S16),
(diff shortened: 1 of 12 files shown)


python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -513,19 +538,25 @@ def test_sum_aggregation_type(lmdb_version_store_dynamic_schema_v1, first_dtype,
...
+            expected_df = pd.DataFrame({"to_sum": np.array([2], expected_type)}, index=[first_group])
+        else:
+            expected_df = pd.DataFrame({"to_sum": np.array([1, 1], expected_type)}, index=["1", "0"])
+        expected_df.index.name = "grouping_column"
+        assert_frame_equal(expected_df, data, check_dtype=True)
 
 @pytest.mark.parametrize("extremum", ["min", "max"])
-@pytest.mark.parametrize("dtype", [np.int32, np.float32])
-def test_extremum_aggregation_with_missing_aggregation_column(lmdb_version_store_dynamic_schema_v1, extremum, dtype):
+@pytest.mark.parametrize("dtype_value", [(np.int32, 0), (np.float32, np.nan), (bool, False)])
```
IvoDD: Nit: Can also parameterize on two arguments:
```
@pytest.mark.parametrize("dtype, default_value", [(np.int32, 0), (np.float32, np.nan), (bool, False)])
test_something(version_store, extremum, dtype, default_value)
```
I think just the name of the test in pytest logs will be clearer this way

cpp/arcticdb/entity/types.hpp
```diff
@@ -215,6 +216,31 @@ enum class DataType : uint8_t {
...
+        DataType::NANOSECONDS_UTC64,
+        DataType::ASCII_FIXED64,
+        DataType::ASCII_DYNAMIC64,
+        DataType::UTF_FIXED64,
+        DataType::UTF_DYNAMIC64,
+        DataType::EMPTYVAL,
+        DataType::BOOL_OBJECT8,
+        DataType::UTF_DYNAMIC32
+    };
+}
```
alexowens90: Since this is only used for testing, can we move it to some test utils file?

alexowens90 (approved):

PR #2444, opened 2025-07-07
Make MEAN groupby return timestamp when the input is timestamp

vasil-pashov:
Prior this PR the result of using the mean aggregation in a groupby statement with timestamp column would be a double column. This is not how Pandas and Polars handle these. The result should be a timestamp value. In case the mean is a floating point we take the integer part without rounding.

Add python tests for sum aggregation
Move aggregation hypothesis tests from unit tests folder to the aggregation folder
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/processing/unsorted_aggregation.cpp
alexowens90, line 1: Comment on line 414 is now out of date

PR #2448, opened 2025-07-07
Extend testing for the index names returned by get_info

poodlewars:
This is to fix https://man312219.monday.com/boards/7852509418/views/168855452/pulses/8667920777 which was actually working correctly - the test case was incorrect (it was not actually setting names on the index being written).

Also fix some obfuscation that shouldn't happen when an index and a column share the same name.

The testing exposes several rough edges with the index names exposed by `get_info` on `Series` - but it does not seem particularly worthwhile changing their behaviour now.

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -457,7 +457,8 @@ def _denormalize_columns_names(columns_names, norm_meta):
                 columns_names[idx] = ""
             elif col_data.is_int:
                 columns_names[idx] = int(col_data.original_name)
-            else:
+            elif col_data.original_name:
+                # Very old clients may not have written the original_name, so if it's not there just leave the name alone
                 columns_names[idx] = col_data.original_name
 
     return columns_names
@@ -483,7 +484,7 @@ def _denormalize_columns(item, norm_meta, idx_type, n_indexes):
     return columns, denormed_columns, data
 
 
-def _normalize_columns_names(columns_names, norm_meta, dynamic_schema=False):
+def _normalize_columns_names(columns_names, index_names, norm_meta, dynamic_schema=False):
     counter = Counter(columns_names)
     for idx in range(len(columns_names)):
         col = columns_names[idx]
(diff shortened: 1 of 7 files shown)


python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -948,10 +987,12 @@ def test_update_times(basic_store):
 
 
 @pytest.mark.storage
-def test_get_info_multi_index(basic_store):
+@pytest.mark.parametrize("index_names", [("blah", None), (None, None), (None, "blah"), ("blah1", "blah2")])
```
G-D-Petrov: Can we add cases with indexes with the same name as the column (`col1`)?
The test is not checking for the names of the indexes in the info, is this expected?

python/tests/unit/arcticdb/version_store/test_unicode.py
```diff
@@ -298,11 +299,13 @@ def test_get_info(lmdb_version_store, batch):
...
         assert list(df_1.columns) == res[0]["col_names"]["columns"]
         assert list(df_2.columns) == res[1]["col_names"]["columns"]
+        assert res[0]["col_names"]["index"] == [unicode_str]
+        assert res[1]["col_names"]["index"] == [unicode_str]
     else:
         for sym, df in [("sym_1", df_1), ("sym_2", df_2)]:
             res = lmdb_version_store.get_info(sym)
             assert list(df.columns) == res["col_names"]["columns"]
-            # assert res["col_names"]["index"] == [unicode_str]  # index names are not exposed by get_info, seems to be a bug 8667920777
+            assert res["col_names"]["index"] == [unicode_str]  # index names are not exposed by get_info, seems to be a bug 8667920777
```
G-D-Petrov: I guess we can remove the comment now?

python/arcticdb/version_store/_normalization.py
```diff
@@ -516,7 +516,7 @@ def _normalize_columns_names(columns_names, index_names, norm_meta, dynamic_sche
             if dynamic_schema and (counter[col] > 1):
                 raise ArcticNativeException("Same column names not allowed in dynamic_schema")
             new_name = col_str
-            if counter[col] > 1 or col in index_names:
```
poodlewars: I'm now not sure that this is a great idea. I should double check what happens when we have duplicated column names that also match the index names (especially with multi-index).

PR #2450, opened 2025-07-08
Small non-functional changes in dedup map

vasil-pashov:
Do not wrap the map in unique pointer
Better variable naming

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/de_dup_map.hpp
+++ b/cpp/arcticdb/version/de_dup_map.hpp
@@ -6,6 +6,11 @@
  */
 
 #pragma once
+#include <vector>
+#include <unordered_map>
+#include <optional>
+#include <algorithm>
+#include <arcticdb/entity/atom_key.hpp>
 
 namespace arcticdb {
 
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/version/de_dup_map.hpp
```diff
@@ -6,52 +6,48 @@
...
-    std::optional<AtomKey> get_key_if_present(const AtomKey &key) const {
-        if (!de_dup_map_) {
-            util::raise_rte("Invalid de dup map object");
-        }
-        auto de_dup_candidates = de_dup_map_->find(key.content_hash());
-
-        if (de_dup_candidates == de_dup_map_->end()) {
+    DeDupMap() = default;
+    ~DeDupMap() = default;
+    DeDupMap(const DeDupMap&) = delete;
```
poodlewars: Doesn't really matter, but there's an `ARCTICDB_NO_COPY` macro

cpp/arcticdb/version/de_dup_map.hpp
```diff
@@ -6,52 +6,48 @@
...
-            return std::nullopt;
-        } else {
-            return std::optional<AtomKey>(*key_iterator);
-        }
+        const auto key_iterator = std::ranges::find_if(de_dup_candidates->second,
+                                                 [&](const auto &k) {
+                                                     return k.start_index() == key.start_index() &&
+                                                            k.end_index() == key.end_index();
+                                                 });
+        return key_iterator != de_dup_candidates->second.end() ? std::optional{*key_iterator} : std::nullopt;
```
poodlewars: (nit) Invert the ternary operator to avoid a double negative

cpp/arcticdb/version/de_dup_map.hpp
```diff
@@ -6,52 +6,48 @@
...
-
-    std::optional<AtomKey> get_key_if_present(const AtomKey &key) const {
-        if (!de_dup_map_) {
-            util::raise_rte("Invalid de dup map object");
-        }
-        auto de_dup_candidates = de_dup_map_->find(key.content_hash());
-
-        if (de_dup_candidates == de_dup_map_->end()) {
+    DeDupMap() = default;
+    ~DeDupMap() = default;
```
poodlewars: These two are implicitly declared anyway right? Why declare them ourselves?

cpp/arcticdb/version/de_dup_map.hpp
```diff
@@ -6,52 +6,48 @@
...
         }
-        if (de_dup_map_->count(key.content_hash()) != 0)
-            de_dup_map_->at(key.content_hash()).push_back(key);
-        else
-            de_dup_map_->insert({key.content_hash(), std::vector<AtomKey>({key})});
     }
 
 private:
-    std::unique_ptr<std::unordered_map<ContentHash, std::vector<AtomKey>>> de_dup_map_;
+    std::unordered_map<ContentHash, std::vector<AtomKey>> de_dup_map_;
```
poodlewars, line 48: I suppose there is a real change here for when we move the `DeDupMap` (we now move the `unordered_map` rather than the `unique_ptr`). I would expect any change in performance to be tiny though.

cpp/arcticdb/version/de_dup_map.hpp
vasil-pashov: Once you delete the copy constructor the default constructor is not generated by the compiler, so we must declare it. You're right about the destructor though, there's no need to declare it.

cpp/arcticdb/version/de_dup_map.hpp
vasil-pashov: Although, it's not specified in the standard but most implementation keep a pointer to the beginning of the bucket list, so it's basically the same, there a few helper size_t values, but I don't think they matter.
My second point is that we wrap the whole thing in a shared pointer anyway, so the unique ptr is redundant.

cpp/arcticdb/version/de_dup_map.hpp
```diff
@@ -6,52 +6,46 @@
...
+                                                     return k.start_index() == key.start_index() &&
+                                                            k.end_index() == key.end_index();
+                                                 });
+        return key_iterator == de_dup_candidates->second.end() ? std::nullopt : std::optional{*key_iterator};
     }
 
     void insert_key(const AtomKey &key) {
-        if (!de_dup_map_) {
-            util::raise_rte("Invalid de dup map object");
+        if (de_dup_map_.contains(key.content_hash())) {
```
alexowens90: nit:
`if (auto it = de_dup_map_.find(key.content_hash()); it != de_dup_map_.end()) {`
to avoid the double lookup

alexowens90 (approved):

PR #2460, opened 2025-07-11
Fix ref key structure after a write_and_prune_previous call

G-D-Petrov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_map.hpp
+++ b/cpp/arcticdb/version/version_map.hpp
@@ -297,13 +297,13 @@ public:
 
         std::vector<AtomKey> keys_to_write;
         std::optional<AtomKey> tombstone_all_key;
+        keys_to_write.push_back(key);
         if (!result.empty()) {
             auto first_key_to_tombstone = previous_key ? previous_key : entry->get_first_index(false).first;
             tombstone_all_key = get_tombstone_all_key(first_key_to_tombstone.value(), store->current_timestamp());
             entry->try_set_tombstone_all(tombstone_all_key.value());
             keys_to_write.push_back(tombstone_all_key.value());
         }
-        keys_to_write.push_back(key);
 
         auto previous_index = do_write(store, key.version_id(), key.id(), std::span{keys_to_write}, entry);
         write_symbol_ref(store, *entry->keys_.cbegin(), previous_index, entry->head_.value());
--- a/python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
+++ b/python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
@@ -9,7 +9,7 @@ from arcticdb.util.test import (
     assert_frame_equal,
     config_context,
     config_context_string,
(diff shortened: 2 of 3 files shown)


python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
```diff
@@ -152,62 +153,90 @@ def test_delete_over_time(lib_name, s3_and_nfs_storage_bucket, clear_query_stats
 
 
 def test_write_and_prune_previous_over_time(lib_name, s3_and_nfs_storage_bucket, clear_query_stats):
-    expected_ops = 9
+    expected_ops = 17
```
G-D-Petrov, line 156: The test wasn't actually deleting in the past, that's we why we have more expected_ops now

python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
```diff
@@ -152,62 +153,90 @@ def test_delete_over_time(lib_name, s3_and_nfs_storage_bucket, clear_query_stats
...
+        assert stats["storage_operations"].keys() == {"S3_GetObject"}, pformat(stats)
+        get_obj_ops = stats["storage_operations"]["S3_GetObject"]
+        # We expect 3 get object operations:
+        # - 1 for the version ref key
+        # - 1 for the index key
+        # - 1 for the data key
+        assert len(get_obj_ops.keys()) == 3, pformat(stats)
+        assert get_obj_ops["VERSION_REF"]["count"] == 1, pformat(stats)
+        assert get_obj_ops["TABLE_INDEX"]["count"] == 1, pformat(stats)
+        assert get_obj_ops["TABLE_DATA"]["count"] == 1, pformat(stats)
```
poodlewars, line 204: It would be good to add a test that checks how many VERSION keys we read in cases where we do actually need to walk through the version chain too (if such a test doesn't already exist)

cpp/arcticdb/version/version_map.hpp
```diff
@@ -297,13 +297,13 @@ class VersionMapImpl {
 
         std::vector<AtomKey> keys_to_write;
         std::optional<AtomKey> tombstone_all_key;
+        keys_to_write.push_back(key);
```
poodlewars, line 300: Does `write_tombstones_internal` need a similar change? Worth checking

poodlewars (approved):

python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
G-D-Petrov: Added a test for this

cpp/arcticdb/version/version_map.hpp
G-D-Petrov: There the case is different as in that case we want to point to the tombstone key.
I've added a test to assert the structure though

PR #2425, opened 2025-06-23
Implement batch_delete operations

G-D-Petrov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #2463, opened 2025-07-11
Add batch_delete APIs and functionality

G-D-Petrov:
#### Reference Issues/PRs
Monday ticket ref: 7855100642

Original PR to master: https://github.com/man-group/ArcticDB/pull/2425

* Refactors `tombstone_version` into an async variant of it
* Implements batch delete functionality on top of `tombstone_versions_async`
* Adds tests for the new functionality

### DIFF
--- a/cpp/arcticdb/version/version_functions.hpp
+++ b/cpp/arcticdb/version/version_functions.hpp
@@ -181,11 +181,14 @@ inline version_store::TombstoneVersionResult populate_tombstone_result(
         // It is possible to have a tombstone key without a corresponding index_key
         // This scenario can happen in case of DR sync
         if (entry->is_tombstoned(version_id)) {
-            util::raise_rte("Version {} for symbol {} is already deleted", version_id, stream_id);
+            missing_data::raise<ErrorCode::E_NO_SUCH_VERSION>(  
+                "Version {} for symbol {} is already deleted", version_id, stream_id);
         } else {
-            if (!latest_key || latest_key->version_id() < version_id)
-                util::raise_rte("Can't delete version {} for symbol {} - it's higher than the latest version",
-                        stream_id, version_id);
+            if (!latest_key || latest_key->version_id() < version_id) {
+                missing_data::raise<ErrorCode::E_NO_SUCH_VERSION>(
+                    "Can't delete version {} for symbol {} - it's higher than the latest version",
+                    version_id, stream_id);
+            }
         }
 
 
@@ -203,7 +206,7 @@ inline version_store::TombstoneVersionResult populate_tombstone_result(
         }
     }
 
-    util::check(
+    storage::check<ErrorCode::E_KEY_NOT_FOUND>(
         res.keys_to_delete.size() == version_ids.size(),
         "Expected {} index keys to be marked for deletion, got {} keys: {}",
         version_ids.size(), 
--- a/cpp/arcticdb/version/version_functions.hpp
+++ b/cpp/arcticdb/version/version_functions.hpp
@@ -279,51 +279,6 @@ inline version_store::TombstoneVersionResult tombstone_versions(
     return tombstone_versions_async(store, version_map, stream_id, version_ids, creation_ts).get();
 }
 
-inline std::vector<version_store::TombstoneVersionResult> tombstone_versions_batch(
-    const std::shared_ptr<Store> &store,
-    const std::shared_ptr<VersionMap> &version_map,
-    const std::vector<std::pair<StreamId, std::unordered_set<VersionId>>> &stream_ids_and_version_ids,
-    const std::optional<timestamp>& creation_ts=std::nullopt) {
-    std::vector<folly::Future<version_store::TombstoneVersionResult>> futures;
-    for (const auto& [stream_id, version_ids] : stream_ids_and_version_ids) {
-        futures.push_back(tombstone_versions_async(store, version_map, stream_id, version_ids, creation_ts));
-    }
-
-    auto tries = folly::collectAll(futures).get();
-    std::vector<version_store::TombstoneVersionResult> results;
-    results.reserve(tries.size());
-    for (auto& try_result : tries) {
-        results.push_back(std::move(try_result.value()));
-    }
-    return results;
-}
-
-inline std::vector<version_store::TombstoneVersionResult> tombstone_versions_batch_from_vectors(
-    const std::shared_ptr<Store> &store,
-    const std::shared_ptr<VersionMap> &version_map,
-    const std::vector<StreamId> &stream_ids,
-    const std::vector<std::vector<VersionId>> &version_ids,
-    const std::optional<timestamp>& creation_ts=std::nullopt) {
-    std::vector<std::pair<StreamId, std::unordered_set<VersionId>>> stream_ids_and_version_ids;
-    stream_ids_and_version_ids.reserve(stream_ids.size());
-    
-    for (size_t i = 0; i < stream_ids.size(); ++i) {
-        // Skip if version_ids[i] is empty
-        if (version_ids[i].empty()) {
-            continue;
-        }
-        std::unordered_set<VersionId> version_set(version_ids[i].begin(), version_ids[i].end());
-        stream_ids_and_version_ids.emplace_back(stream_ids[i], std::move(version_set));
-    }
-    
-    // If no valid version IDs to process, return empty result
-    if (stream_ids_and_version_ids.empty()) {
-        return {};
-    }
-    
-    return tombstone_versions_batch(store, version_map, stream_ids_and_version_ids, creation_ts);
-}
-
 inline std::optional<AtomKey> get_index_key_from_time(
     timestamp from_time,
     const std::vector<AtomKey> &keys) {
--- a/cpp/arcticdb/version/version_functions.hpp
+++ b/cpp/arcticdb/version/version_functions.hpp
@@ -218,7 +218,7 @@ inline folly::Future<version_store::TombstoneVersionResult> finalize_tombstone_r
     version_store::TombstoneVersionResult res,
     const std::shared_ptr<VersionMap>& version_map,
     const std::shared_ptr<VersionMapEntry>& entry,
-    AtomKey tombstone_key) {
+    [[maybe_unused]] AtomKey tombstone_key) {
     ARCTICDB_DEBUG(log::version(), "Finalizing result for tombstone key {}", tombstone_key);
     // Update the result with final state
     if (version_map->validate())
--- a/cpp/arcticdb/version/version_functions.hpp
+++ b/cpp/arcticdb/version/version_functions.hpp
@@ -162,8 +162,7 @@ inline version_store::TombstoneVersionResult populate_tombstone_result(
     const std::shared_ptr<VersionMapEntry>& entry,
     const std::unordered_set<VersionId>& version_ids,
     const StreamId& stream_id,
-    const std::shared_ptr<Store>& store,
-    const std::optional<timestamp>& creation_ts=std::nullopt) {
+    const std::shared_ptr<Store>& store) {
     version_store::TombstoneVersionResult res(entry->empty(), stream_id);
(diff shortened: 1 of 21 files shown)


vasil-pashov (approved):

python/arcticdb/version_store/library.py
```diff
@@ -13,6 +13,7 @@
 import pytz
 from enum import Enum, auto
 from typing import Optional, Any, Tuple, Dict, Union, List, Iterable, NamedTuple
+from dataclasses import dataclass
```
alexowens90: Unused?

python/arcticdb/version_store/_store.py
```diff
@@ -2643,6 +2644,59 @@ def delete_versions(self, symbol: str, versions: List[int]):
...
+        ValueError
+            If version_ids is empty for any symbol.
+        """
+        # make sure that the versions are not empty
+        for symbol, version_ids in zip(symbols, versions):
+            if not version_ids:
+                raise ValueError(f"version_ids cannot be empty for symbol '{symbol}'")
+        return self.version_store.batch_delete(symbols, versions)
+
+    def batch_delete_symbols(self, symbols: List[str]) -> List[Optional[DataError]]:
```
alexowens90, line 2685: I'm not familiar enough with the bridge magic, will `DaatError` be importable inside Man automagically?

python/arcticdb/version_store/_store.py
```diff
@@ -2643,6 +2644,59 @@ def delete_versions(self, symbol: str, versions: List[int]):
...
+        Returns
+        -------
+        `List[DataError]`
+            List of DataError objects, one for each symbol that was not deleted due to an error.
+            If the symbol was already deleted, there will be no error, just a warning.
+
+        Raises
+        ------
+        ValueError
+            If version_ids is empty for any symbol.
```
alexowens90:
```suggestion
            If versions is empty for any symbol.
```

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -956,6 +956,60 @@ void PythonVersionStore::delete_versions(
     }
 }
 
+std::vector<std::optional<DataError>> PythonVersionStore::batch_delete(
+    const std::vector<StreamId>& stream_ids,
+    const std::vector<std::vector<VersionId>>& version_ids) {
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(stream_ids.size() == version_ids.size(), "when calling batch_delete_versions, stream_ids and version_ids must have the same size");
```
alexowens90, line 964: Might be worth a comment that this error message is correct as it can only be triggered by `batch_delete_versions` due to upstream checks in other code paths that get here

cpp/arcticdb/version/version_functions.hpp
```diff
@@ -226,6 +231,100 @@ inline version_store::TombstoneVersionResult tombstone_versions(
...
+    const std::shared_ptr<Store> &store,
+    const std::shared_ptr<VersionMap> &version_map,
+    const StreamId &stream_id,
+    const std::unordered_set<VersionId>& version_ids) {
+    ARCTICDB_DEBUG(log::version(), "Tombstoning versions {} for stream {}", version_ids, stream_id);
+
+    return async::submit_io_task(CheckReloadTask{store,
+                                    version_map,
+                                    stream_id,
+                                    LoadStrategy{LoadType::ALL, LoadObjective::UNDELETED_ONLY}})
```
alexowens90, line 297: (I know it was like this before, but just noticed it now)
Is this the right load type? If you call write 10 times, and then delete version 8, it will load the whole version chain.
Could it just load down to the smallest element of `version_ids`?

cpp/arcticdb/version/version_store_api.hpp
```diff
@@ -201,6 +201,10 @@ class PythonVersionStore : public LocalVersionedEngine {
         const StreamId& stream_id,
         const std::vector<VersionId>& version_ids);
 
+    std::vector<std::optional<DataError>> batch_delete(
+        const std::vector<StreamId>& stream_ids,
+        const std::vector<std::vector<VersionId>>& version_ids);
```
alexowens90, line 206: Not for this PR, but this (and equivalent argument in non-batch methods) should be `SignedVersionId`, so that users can do things like delete the latest version

cpp/arcticdb/version/version_functions.hpp
alexowens90: Oh I guess it's needed to know if there are any versions left alive

cpp/arcticdb/version/version_tasks.hpp
```diff
@@ -153,6 +153,38 @@ struct WriteVersionTask : async::BaseTask {
         return folly::Unit{};
     }
 };
+
+struct WriteTombstoneTask : async::BaseTask {
```
alexowens90: nit: `WriteTombstonesTask`

cpp/arcticdb/version/version_functions.hpp
```diff
@@ -226,6 +231,100 @@ inline version_store::TombstoneVersionResult tombstone_versions(
...
+        .thenValue([store, version_map, stream_id, version_ids](std::shared_ptr<VersionMapEntry>&& entry) {
+            return process_tombstone_versions(store, version_map, stream_id, version_ids, std::move(entry));
+        });
+}
+
+inline folly::Future<version_store::TombstoneVersionResult> tombstone_all_async(
```
alexowens90, line 303: Would be good if the non-batch version also used this function so we only have one codepath to maintain

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -956,6 +956,60 @@ void PythonVersionStore::delete_versions(
...
+    std::vector<IndexTypeKey> keys_to_delete;
+    std::vector<std::pair<StreamId, VersionId>> symbols_to_delete;
+    std::vector<std::string> failed_symbols;
+
+    for (const auto& result : results) {
+        util::variant_match(result,
+            [&](const version_store::TombstoneVersionResult& tombstone_result) {
+                return_results.emplace_back(std::nullopt);
+
+                if(tombstone_result.keys_to_delete.empty()) {
```
alexowens90, line 978: I think this is an error? At least the index key should always be deleted?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -956,6 +956,60 @@ void PythonVersionStore::delete_versions(
...
+                    log::version().warn("Nothing to delete for symbol '{}'", tombstone_result.symbol);
+                    return;
+                }
+
+                if (!tombstone_result.keys_to_delete.empty() && !cfg().write_options().delayed_deletes()) {
```
alexowens90: nit: first check is unnecessary, both because of above return statement, and because `insert` will handle `keys_to_delete` being empty

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -956,6 +956,60 @@ void PythonVersionStore::delete_versions(
...
+                    keys_to_delete.insert(keys_to_delete.end(), tombstone_result.keys_to_delete.begin(), tombstone_result.keys_to_delete.end());
+                }
+
+                if(tombstone_result.no_undeleted_left && cfg().symbol_list() && !tombstone_result.keys_to_delete.empty()) {
+                    symbols_to_delete.emplace_back(tombstone_result.symbol, tombstone_result.latest_version_);
+                }
+            },
+            [&](const DataError& data_error) {
+                return_results.emplace_back(std::make_optional(data_error));
```
alexowens90: Can move `data_error` in

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -956,6 +956,60 @@ void PythonVersionStore::delete_versions(
...
+    const std::vector<std::vector<VersionId>>& version_ids) {
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(stream_ids.size() == version_ids.size(), "when calling batch_delete_versions, stream_ids and version_ids must have the same size");
+    
+    auto results = batch_delete_internal(stream_ids, version_ids);
+    
+    std::vector<std::optional<DataError>> return_results;
+
+    std::vector<IndexTypeKey> keys_to_delete;
+    std::vector<std::pair<StreamId, VersionId>> symbols_to_delete;
+    std::vector<std::string> failed_symbols;
```
alexowens90: Unused

python/tests/integration/arcticdb/version_store/test_snapshot.py
```diff
@@ -28,6 +28,7 @@
 )
 from tests.util.storage_test import get_s3_storage_config
 from arcticdb_ext.storage import KeyType
+from arcticdb.version_store.library import DeleteRequest, WritePayload
```
alexowens90: Unused

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -497,6 +498,132 @@ def test_delete_batch(library_factory, sym):
...
+    lib.write("sym1", df2)
+    lib.write("sym2", df3)
+
+    # Test 1: Delete all versions of a symbol (string input)
+    result = lib.delete_batch(["sym1"])
+    assert len(result) == 1
+    assert result[0] is None
+
+    # Verify sym1 is deleted
+    assert not lib.has_symbol("sym1")
```
alexowens90, line 522: `has_symbol` uses the version chain directly, would be good to also assert that the symbol list cache is correct

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -497,6 +498,132 @@ def test_delete_batch(library_factory, sym):
...
+    # We should just have sym1
+    assert not [ver for ver in lib._nvs.list_versions() if ver["symbol"] == "sym2"]
+
+    lib.write("sym3", original_data)
+    lib.snapshot("sym3_snap")
+
+    # This version of sym3 is not in a snapshot
+    lib.write("sym3", v1_data)
+
+    # This should NOT delete sym1 and shoud return an error
```
alexowens90: Can assert this

python/arcticdb/version_store/_store.py
alexowens90, line 1: Need to remember to update `NativeVersionStore` in `man.core` so that we get kibana logs

cpp/arcticdb/version/version_functions.hpp
G-D-Petrov: Yes exactly

cpp/arcticdb/version/version_store_api.hpp
G-D-Petrov: Opened an issue for this: https://github.com/man-group/ArcticDB/issues/2471

cpp/arcticdb/version/version_store_api.cpp
G-D-Petrov: This is to cover the case where the symbol is already deleted.
In a previous PR for multi deletes, we agreed that this case should be a warning and not an error and this keeps the same behavior.

python/arcticdb/version_store/_store.py
G-D-Petrov: Looks like it, we are already using it in other functions in this file.
And I see that other teams inside Man are importing in in their code.

python/tests/integration/arcticdb/test_arctic_batch.py
G-D-Petrov: The comment was a bit wrong, I changed it.
The code was already asserting the expected behavior.

cpp/arcticdb/version/version_store_api.cpp
alexowens90: Yeh OK. I dislike the mismatch we have where `delete` is a no-op if the symbol is already deleted, but deleting a specific version raises if that version has been deleted. Hard to change now though

PR #2470, opened 2025-07-17
Increase coverage - append and update v1 API

grusev:

Small fix for append() documentation.

extended existing test to cover nd arrays with all types

added common methods to create series and np arrays of any arcticdb supported data type to help testing 

added several tests use append/update in different dynamically growing symbols

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/util/test.py
+++ b/python/arcticdb/util/test.py
@@ -158,19 +158,26 @@ def dataframe_dump_to_log(label_for_df, df: pd.DataFrame):
         print("-" * 80)
 
 
-def dataframe_simulate_arcticdb_update_static(existing_df: pd.DataFrame, update_df: pd.DataFrame) -> pd.DataFrame:
+def dataframe_simulate_arcticdb_update_static(existing_df: Union[pd.DataFrame, pd.Series], 
+                                              update_df: Union[pd.DataFrame, pd.Series]) -> Union[pd.DataFrame, pd.Series]:
     """
-    Does implement arctic logic of update() method functionality over pandas dataframes.
-    In other words the result, new data frame will have the content of 'existing_df' dataframe
+    Does implement arctic logic of update() method functionality over pandas dataframes/series.
+    In other words the result, new data frame will have the content of 'existing_df' dataframe/series
     updated with the content of "update_df" dataframe the same way that arctic is supposed to work.
     Useful for prediction of result content of arctic database after update operation
     NOTE: you have to pass indexed dataframe
     """
 
-    assert existing_df.dtypes.to_list() == update_df.dtypes.to_list(), (
-        "Dataframe must have identical columns types in same order"
-    )
-    assert existing_df.columns.to_list() == update_df.columns.to_list(), "Columns names also need to be in same order"
+    if isinstance(existing_df, pd.Series):
+        if len(update_df) < 1:
+            return existing_df # Nothing to update
+        assert existing_df.dtype == update_df.dtype, f"Series must have same type {existing_df.dtype} == {update_df.dtype}"
+        assert existing_df.name == update_df.name, "Series name must be same"
+    else:
+        assert existing_df.dtypes.to_list() == update_df.dtypes.to_list(), (
+            "Dataframe must have identical columns types in same order"
+        )
+        assert existing_df.columns.to_list() == update_df.columns.to_list(), "Columns names also need to be in same order"
 
     start2 = update_df.first_valid_index()
     end2 = update_df.last_valid_index()
--- a/python/arcticdb/util/utils.py
+++ b/python/arcticdb/util/utils.py
@@ -13,7 +13,7 @@ import re
 import string
 import time
 import sys
-from typing import Dict 
+from typing import Dict, Set 
 from typing import Literal, Any, List, Tuple, Union, get_args
 import numpy as np
 import pandas as pd
@@ -25,7 +25,7 @@ from arcticdb.version_store.library import Library
 # Types supported by arctic
 ArcticIntType = Union[np.uint8, np.uint16, np.uint32, np.uint64, np.int8, np.int16, np.int32, np.int64]
 ArcticFloatType = Union[np.float64, np.float32]
-ArcticTypes = Union[ArcticIntType, ArcticFloatType, str]
+ArcticTypes = Union[ArcticIntType, ArcticFloatType, bool, str]
 supported_int_types_list = list(get_args(ArcticIntType))
 supported_float_types_list = list(get_args(ArcticFloatType))
 supported_types_list = list(get_args(ArcticTypes))
@@ -85,6 +85,72 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
     loggers[name] = logger
     return logger
 
+
+def set_seed(seed=None):
+    """Sets seed to random libraries if not None"""
+    if seed is not None:
+       np.random.seed(seed)
+       random.seed(seed)
+
+
+def generate_random_numpy_array(size, _type, seed=8238):
+    """ Generates random numpy array of specified type
+    """
+    set_seed(seed)
+    arr = []
+    if 'int' in str(_type):
+        arr = random_integers(size, _type)
+    elif 'float' in str(_type):
+        arr = np.arange(size, dtype=_type)
+    elif 'bool' in str(_type):
+        arr = np.random.randn(size) > 0
+    elif 'str' in str(_type):
+        length = 10
+        arr = [random_string(length) for _ in range(size)]
+        arr = np.array(arr, dtype=f"U{size}")
+    else:
+        raise TypeError("Unsupported type {dtype}")        
+    return arr
+
+
+def generate_random_series(type: ArcticTypes, length: int, name: str, 
+                    start_time: pd.Timestamp=None, freq: str='s', seed=3247) -> pd.Series:
+    """Generates random series of specified type with or without index"""
+    set_seed(seed)
+    index = None
+    if start_time:
+        index = pd.date_range(start_time, periods=length, freq=freq)
+    return pd.Series(generate_random_numpy_array(length, type, seed), 
+                           index=index,
+                           name=name)
+
+
+def verify_dynamically_added_columns(updated_df: pd.DataFrame, row_index: Union[int, pd.Timestamp],
+                                     new_columns_to_verify: Set[str]):
+    """ Verifies the value of dynamically added columns to dataframes 
+    after append/update operation with dataframe that is having additional new rows
+
+    row_index is either location of the row in dataframe or its index (timestamp)
+    """
+    updated_df_columns = set(updated_df.columns.to_list())
+    assert updated_df_columns.issuperset(new_columns_to_verify)
+    for col in new_columns_to_verify:
+            dtype = updated_df[col].dtype
+            if isinstance(row_index, pd.Timestamp):
+                value = updated_df[col].loc[row_index]
+            else:
+                value = updated_df[col].iloc[row_index]
+            if 'int' in str(dtype):
+                assert 0 == value, f"column {col}:{dtype} -> 0 == {value}"
+            elif 'float' in str(dtype):
+                assert pd.isna(value), f"column {col}:{dtype} -> Nan == {value}"
+            elif 'bool' in str(dtype):
+                assert False == value, f"column {col}:{dtype} -> False == {value}"
+            elif 'object' in str(dtype):
+                assert value is None , f"column {col}:{dtype} -> None == {value}"
+            else:
+                raise TypeError(f"Unsupported dtype: {dtype}")
+            
 
 class GitHubSanitizingException(Exception):
     def __init__(self, message: str):
@@ -525,9 +591,7 @@ class DFGenerator:
         self.__types = {}
         self.__df = None
         self.__index = None
-        if seed is not None:
-            np.random.seed(seed)
-            random.seed(seed)
+        set_seed(seed)
 
     def generate_dataframe(self) -> pd.DataFrame:
         if self.__df is None:
--- a/python/arcticdb/util/utils.py
+++ b/python/arcticdb/util/utils.py
@@ -16,6 +16,7 @@ import sys
 from typing import Dict, Set 
 from typing import Literal, Any, List, Tuple, Union, get_args
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 
 from arcticdb.util.test import create_datetime_index, get_sample_dataframe, random_integers, random_string
@@ -25,7 +26,7 @@ from arcticdb.version_store.library import Library
 # Types supported by arctic
 ArcticIntType = Union[np.uint8, np.uint16, np.uint32, np.uint64, np.int8, np.int16, np.int32, np.int64]
 ArcticFloatType = Union[np.float64, np.float32]
-ArcticTypes = Union[ArcticIntType, ArcticFloatType, bool, str]
+ArcticTypes = Union[ArcticIntType, ArcticFloatType, bool, str, np.datetime64]
 supported_int_types_list = list(get_args(ArcticIntType))
 supported_float_types_list = list(get_args(ArcticFloatType))
 supported_types_list = list(get_args(ArcticTypes))
@@ -93,7 +94,20 @@ def set_seed(seed=None):
        random.seed(seed)
 
 
-def generate_random_numpy_array(size, _type, seed=8238):
+def generate_random_timestamp_array(size: int, 
+                               start: str = '2020-01-01', 
+                               end: str = '2030-01-01', 
+                               seed: int = 432432) -> npt.NDArray[np.datetime64]:
+    """ Generates an array of random timestamps"""
+    if seed: 
+        np.random.seed(seed)
+    start_ts = pd.Timestamp(start).value // 10**9
+    end_ts = pd.Timestamp(end).value // 10**9
+    random_seconds = np.random.randint(start_ts, end_ts, size=size)
+    return np.array(pd.to_datetime(random_seconds, unit='s'))
+
+
+def generate_random_numpy_array(size: int, _type, seed: int = 8238) -> npt.NDArray[Any]:
     """ Generates random numpy array of specified type
     """
     set_seed(seed)
@@ -108,6 +122,8 @@ def generate_random_numpy_array(size, _type, seed=8238):
         length = 10
         arr = [random_string(length) for _ in range(size)]
         arr = np.array(arr, dtype=f"U{size}")
+    elif 'datetime' in str(_type):
+        arr = generate_random_timestamp_array(size, seed=seed)
     else:
         raise TypeError("Unsupported type {dtype}")        
     return arr
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -720,7 +720,8 @@ class NativeVersionStore:
             Data to be appended.
         metadata : `Optional[Any]`, default=None
             Optional metadata to persist along with the new symbol version. Note that the metadata is
-            not combined in any way with the metadata stored in the previous version.
+            not combined in any way with the metadata stored in the previous version. Additionally metadata is not stored
+            also when incomplet=True
         prune_previous_version
             Removes previous (non-snapshotted) versions from the database.
         validate_index: bool, default=False
--- a/python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
+++ b/python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
@@ -1,0 +1,223 @@
+"""
+Copyright 2023 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+import re
+from typing import Set
+import pytest
+import pandas as pd
+import numpy as np
+
+from arcticdb.util.test import (
+    assert_frame_equal,
+    assert_series_equal,
+    dataframe_simulate_arcticdb_update_static,
+)
+from arcticdb.util.utils import ArcticTypes, generate_random_series, set_seed, supported_types_list, generate_random_numpy_array, verify_dynamically_added_columns
+from arcticdb.version_store._store import NativeVersionStore
+from datetime import datetime, timedelta
+
+from tests.util.mark import SLOW_TESTS_MARK
+
+
+def add_index(df: pd.DataFrame, start_time: pd.Timestamp):
+    df.index = pd.date_range(start_time, periods=df.shape[0], freq='s')
+
+
+def make_df_appendable(original_df: pd.DataFrame, append_df:pd.DataFrame):
+    """ Creates such index for `append_df` so that it can be appended to `original_df`"""
+    last_time = original_df.index[-1]
+    add_index(append_df, last_time + timedelta(seconds=1))
+
+
+def create_all_arcticdb_types_df(length: int, column_prefix: str = ""):
+    """ Creates a dataframe with columns of all supported arcticdb data types."""
+    arr = []
+    for dtype in supported_types_list:
+        name = f"{column_prefix}_{np.dtype(dtype).name}"
+        arr.append(generate_random_series(dtype, length, name, seed=None))
+    return pd.concat(arr, axis=1)
+
+
+def wrap_df_add_new_columns(df: pd.DataFrame, prefix: str = ""):
+    """ Adds columns at the beginning and at the end of dataframe 
+    Columns added are of all supported arcticdb types and have
+    specified prefix
+    """
+    length = df.shape[0]
+    df1 = create_all_arcticdb_types_df(length, column_prefix=f"{prefix}_pre_")
+    df1.index = df.index
+    df2 = create_all_arcticdb_types_df(length, column_prefix=f"{prefix}_post_")
+    df2.index = df.index
+    df = pd.concat([df1, df, df2], axis=1)
+    return df
+
+
+def get_metadata():
+    '''Returns weird and complex metadata'''
+    metadata = {
+        "experiment": {
+            "id": 42,
+            "params": {
+                "learning_rate": 0.01,
+                "batch_size": 32
+            }
+        },
+        "source": "integration_test",
+        "tags": ["v1.2", "regression"]
+    }
+    return [metadata, {}, [metadata, {}, [metadata, {}], 1], "", None]
+
+
+@SLOW_TESTS_MARK
+@pytest.mark.parametrize("dtype", supported_types_list)
+@pytest.mark.parametrize("schema", [True, False])
+def test_write_append_update_read_scenario_with_different_series_combinations(version_store_factory, dtype, schema):
+    """This test covers series with timestamp index of all supported arcticdb types.
+    Write, append and read combinations of different boundary length sizes of series
+    """
+    segment_row_size = 3
+    lib: NativeVersionStore = version_store_factory(dynamic_schema=schema, segment_row_size=segment_row_size)
+    set_seed(3484356)
+    max_length = segment_row_size * 29
+    symbol = f"symbol-{re.sub(r'[^A-Za-z0-9]', '_', str(dtype))}"
+    name = "some_name!"
+    timestamp = pd.Timestamp(4839275892348)
+    series_length = [ 1, 0, 2, max_length]
+    meta = get_metadata()
+
+    for length in series_length:
+        total_length = 0
+        series = generate_random_series(dtype, length, name, start_time=timestamp, seed=None)
+        total_length += length
+        lib.write(symbol, series)
+        result_series = series
+        assert_series_equal(lib.read(symbol).data, series, check_index_type=(len(series) > 0))
+        
+        for append_series_length in series_length:
+            append_series = generate_random_series(dtype, append_series_length, name, 
+                                                   start_time=timestamp + timedelta(seconds=total_length), seed=None)
+            lib.append(symbol, append_series, metadata=meta)
+            result_series = pd.concat([result_series, append_series])
+            ver = lib.read(symbol)
+            assert_series_equal(result_series, ver.data, check_index_type=(len(result_series) > 0))
+            assert meta == ver.metadata
+            # Note update is of same length but starts in previous period
+
+            update_series = generate_random_series(dtype, append_series_length, name, 
+                                                   start_time=timestamp + timedelta(seconds=total_length - 1), seed=None)
+            total_length += append_series_length
+            lib.update(symbol, update_series, metadata=meta)
+            result_series = dataframe_simulate_arcticdb_update_static(result_series, update_series)
+            ver = lib.read(symbol)
+            assert_series_equal(result_series, ver.data, check_index_type=(len(result_series) > 0))
+            assert meta == ver.metadata
+
+
+@pytest.mark.storage
+def test_append_update_dynamic_schema_add_columns_all_types(version_store_and_real_s3_basic_store_factory):
+    """
+    The test does series of append operations with new columns of all supported column types.
+    The resulting symbol will have additional columns each time added, with predefined default 
+    values for different data types
+    """
+
+    set_seed(32432)
+    counter = 1
+    lib: NativeVersionStore = version_store_and_real_s3_basic_store_factory(
+        dynamic_schema=True, dynamic_strings=True, segment_row_size=3)
+    symbol = "232_43213dfmkd_!"
+    col_name = 'MIDDLE'
+    start_time = pd.Timestamp(32513454)
+
+    def get_df() -> pd.DataFrame:
+        """ Creates new dataframe with one constant one column and one row where the value is 
+        auto incremented on each new dataframe created"""
+        nonlocal counter
+        df = pd.DataFrame({col_name: [counter]})
+        counter += 1
+        return df
+
+    initial_df = get_df()
+    add_index(initial_df, start_time)
+    lib.write(symbol, initial_df)
+    read_data = lib.read(symbol).data
+
+    # Will do append operation times to confirm growth is not a problem
+    for i in range(10):
+        num_versions_before = len(lib.list_versions(symbol))
+        middle_df = get_df()
+        add_index(middle_df, start_time)
+        append_df = wrap_df_add_new_columns(middle_df, i)
+        make_df_appendable(read_data, append_df) # timestamps of append_df to be after the initial
+
+        # Get new columns that will be added to appended rows
+        new_columns_to_appended_df = set(read_data.columns.to_list())
+        # Remove the common column
+        new_columns_to_appended_df.remove(col_name)
+        # Get new columns that are added with this append
+        new_columns_to_symbol = set(append_df.columns.to_list())
+        # Remove the common column
+        new_columns_to_symbol.remove(col_name)
+
+        # Do append with different mixes of parameters
+        if i % 2 == 0:
+            meta = get_metadata()
+            lib.append(symbol, append_df, prune_previous_version=True, validate_index=True, metadata=meta)
+            assert num_versions_before, lib.list_versions(symbol) # previous version is pruned
+        else:
+            meta = None
+            # NOTE: metadata is not stored then incomplete=True
+            lib.append(symbol, append_df, incomplete=True, prune_previous_version=False, validate_index=True, metadata=meta)
+            lib.compact_incomplete(symbol, True, False)
+            assert num_versions_before, lib.list_versions(symbol) + 1 # previous version is not pruned
+
+        # Verify  metadata and dynamically added columns
+        ver = lib.read(symbol)
+        read_data:pd.DataFrame = ver.data
+        assert meta == ver.metadata
+        verify_dynamically_added_columns(read_data, -2, new_columns_to_symbol)
+        verify_dynamically_added_columns(read_data, -1, new_columns_to_appended_df)
+
+    # Will only update last row with 1 row dataframe
+    update_timestamp = read_data.index[-1]
+    # Will take previous to the update row timestamp
+    previous_timestamp = read_data.index[-2]
+
+    # repeat update operation several times, to confirm growth is not a problem
+    for i in range(5):
+        num_versions_before = len(lib.list_versions(symbol))
+        middle_df = get_df()
+        add_index(middle_df, update_timestamp)
+        update_df = wrap_df_add_new_columns(middle_df, f"upd_{i}")
+
+        # Get new columns that will be added to updated rows
+        new_columns_to_update_df = set(read_data.columns.to_list())
+        # Remove the common column
+        new_columns_to_update_df.remove(col_name)
+        # Get new columns that are added with this update
+        new_columns_to_symbol = set(update_df.columns.to_list())
+        # Remove the common column
+        new_columns_to_symbol.remove(col_name)
+
+        if i % 2 == 0:
+            meta = get_metadata()
+            lib.update(symbol, update_df, prune_previous_version=False, metadata=meta)
+            assert num_versions_before, lib.list_versions(symbol) + 1 # previous version is not pruned
+        else:
+            meta = "just a message"
+            lib.update(symbol, update_df, prune_previous_version=True, metadata=meta)
+            assert num_versions_before, lib.list_versions(symbol) # previous version is pruned
+
+        # Verify  metadata and dynamically added columns
+        ver = lib.read(symbol)
+        read_data:pd.DataFrame = ver.data
+        assert meta == ver.metadata
+        verify_dynamically_added_columns(read_data, previous_timestamp, new_columns_to_symbol)
+        verify_dynamically_added_columns(read_data, update_timestamp, new_columns_to_update_df)
+
+
--- a/python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
+++ b/python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
@@ -13,14 +13,22 @@ import pandas as pd
 import numpy as np
 
 from arcticdb.util.test import (
-    assert_frame_equal,
     assert_series_equal,
     dataframe_simulate_arcticdb_update_static,
 )
-from arcticdb.util.utils import ArcticTypes, generate_random_series, set_seed, supported_types_list, generate_random_numpy_array, verify_dynamically_added_columns
-from arcticdb.version_store._store import NativeVersionStore
-from datetime import datetime, timedelta
+from arcticdb.util.utils import ArcticTypes, generate_random_series, set_seed, supported_types_list, verify_dynamically_added_columns
+from arcticdb.version_store._store import NativeVersionStore, VersionedItem
+from datetime import timedelta, timezone
 
+from arcticdb.exceptions import (
+    ArcticNativeException,
+)
+from arcticdb_ext.exceptions import (
+    InternalException,
+    NormalizationException,
+)
+
+from benchmarks.bi_benchmarks import assert_frame_equal
 from tests.util.mark import SLOW_TESTS_MARK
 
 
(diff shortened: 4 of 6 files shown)


python/arcticdb/util/test.py
```diff
@@ -158,19 +158,26 @@ def dataframe_dump_to_log(label_for_df, df: pd.DataFrame):
...
     updated with the content of "update_df" dataframe the same way that arctic is supposed to work.
     Useful for prediction of result content of arctic database after update operation
     NOTE: you have to pass indexed dataframe
     """
 
-    assert existing_df.dtypes.to_list() == update_df.dtypes.to_list(), (
-        "Dataframe must have identical columns types in same order"
-    )
-    assert existing_df.columns.to_list() == update_df.columns.to_list(), "Columns names also need to be in same order"
+    if isinstance(existing_df, pd.Series):
```
IvoDD: We should probably not allow combinations of `pd.Series` and `pd.DataFrame`. So we can make these ifs like:
```
if isinstance(existing_df, pd.Series) and isinstance(update_df, pd.Series):
     some code
elif isinstance(existing_df, pd.DataFrame) and isinstance(update_df, pd.DataFrame):
     some other code
else:
     raise("Expected existing_df and update_df to have the same type. Types: {type(existing_df)} and {type(update_df)}")

File: python/arcticdb/util/test.py [hunk 2]
```diff
@@ -236,6 +243,19 @@ def wrapper(*args, **kwargs):
...
+def assert_series_equal_pandas_1(expected: pd.Series, actual: pd.Series, **kwargs):
+    """For Pandas 1 type of empty series will be float64 when returned by arctic"""
+    if IS_PANDAS_ONE:
+        if (
+            (np.issubdtype(expected.dtype, np.object_) and np.issubdtype(actual.dtype, np.floating)) or
+            (np.issubdtype(expected.dtype, np.floating) and np.issubdtype(actual.dtype, np.object_))
+            ):
+            if (expected.size == 0) and (actual.size == 0):
+                assert expected.name == actual.name
+                return
```
IvoDD, line 224: I think we should still check that the index on both series is the the same. E.g. we should raise on such series:
```
series_1 = pd.Series([], index=pd.date_range(pd.Timestamp(0), periods=0))
series_2 = pd.Series([], index=[])
series_2.index.name = "index"
pd.testing.assert_series_equal(series_1, series_2)
```
What about  What do you think about something like:
```
if IS_PANDAS_ONE:
    if expected.size == 0:
         expected = expected.astype(np.float64)
    if actual.size == 0:
         actual= actual.astype(np.float64)
assert_series_equal(expected, actual, **kwargs)
```

python/arcticdb/util/test.py
```diff
@@ -236,6 +243,19 @@ def wrapper(*args, **kwargs):
 assert_series_equal = maybe_not_check_freq(pd.testing.assert_series_equal)
 
 
+def assert_series_equal_pandas_1(expected: pd.Series, actual: pd.Series, **kwargs):
+    """For Pandas 1 type of empty series will be float64 when returned by arctic"""
```
IvoDD, line 214: Is this behavior documented? I'm surprised we have a different output between pandas 1 and pandas 2 in this case.
If possible we should just fix it. 
If not what do you think about renaming this function to `assert_series_equal_with_empty_handling` or somthing similiar because it does indeed work for both pandas 1 and pandas 2 it just does some extra handling for pandas 1 and empty series.

python/arcticdb/util/utils.py
```diff
@@ -86,6 +87,109 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
...
+    Sample usage:
+        for package in list_installed_packages():
+            print(package)
+    """
+    try:
+        # Python 3.8+
+        from importlib.metadata import distributions
+        return [f"{dist.metadata['Name']}=={dist.version}" for dist in distributions()]
+    except ImportError:
+        # Previous pythons
```
IvoDD: We no longer run CI with python <3.8, so let's simplify the imports. See #2400

python/arcticdb/util/utils.py
```diff
@@ -86,6 +87,109 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
...
+            return [f"{dist.project_name}=={dist.version}" for dist in pkg_resources.working_set]
+        except ImportError:
+            raise RuntimeError("Neither importlib.metadata nor pkg_resources is available.")
+
+
+def set_seed(seed=None):
+    """Sets seed to random libraries if not None"""
+    if seed is not None:
+       np.random.seed(seed)
+       random.seed(seed)
```
IvoDD, line 62: There already exists a `set_seed` in `test_update.py`. Let's remove that one and keep this one as it makes more sense.

python/arcticdb/util/utils.py
```diff
@@ -86,6 +87,109 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
...
+
+def generate_random_timestamp_array(size: int, 
+                               start: str = '2020-01-01', 
+                               end: str = '2030-01-01', 
+                               seed: int = 432432) -> npt.NDArray[np.datetime64]:
+    """ Generates an array of random timestamps"""
+    if seed: 
+        np.random.seed(seed)
+    start_ts = pd.Timestamp(start).value // 10**9
+    end_ts = pd.Timestamp(end).value // 10**9
```
IvoDD, line 73: Why divide by `10**9` and then use `s` precision? We can just use `ns` precision

python/arcticdb/util/utils.py
```diff
@@ -86,6 +87,109 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
...
+
+def generate_random_numpy_array(size: int, _type, seed: int = 8238) -> npt.NDArray[Any]:
+    """ Generates random numpy array of specified type
+    """
+    set_seed(seed)
+    arr = []
+    if 'int' in str(_type):
+        arr = random_integers(size, _type)
+    elif 'float' in str(_type):
+        arr = np.arange(size, dtype=_type)
```
IvoDD: `arange` is not random. Maybe `np.random.random(size)`?

python/arcticdb/util/utils.py
```diff
@@ -86,6 +87,109 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
...
+       random.seed(seed)
+
+
+def generate_random_timestamp_array(size: int, 
+                               start: str = '2020-01-01', 
+                               end: str = '2030-01-01', 
+                               seed: int = 432432) -> npt.NDArray[np.datetime64]:
+    """ Generates an array of random timestamps"""
+    if seed: 
+        np.random.seed(seed)
```
IvoDD, line 71: Maybe we can `set_seed` instead to have our seed handling consistent?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -259,13 +263,29 @@ def test_upsert_with_delete(lmdb_version_store_big_map):
...
 
 def test_append_numpy_array(lmdb_version_store):
-    np1 = random_integers(10, np.uint32)
-    lmdb_version_store.write("test_append_numpy_array", np1)
-    np2 = random_integers(10, np.uint32)
-    lmdb_version_store.append("test_append_numpy_array", np2)
-    vit = lmdb_version_store.read("test_append_numpy_array")
-    expected = np.concatenate((np1, np2))
-    assert_array_equal(vit.data, expected)
+    '''Tests append will all supported by arctic data types'''
```
IvoDD:
```suggestion
    '''Tests append with all supported by arctic data types'''
```

python/arcticdb/version_store/_store.py
```diff
@@ -739,7 +739,8 @@ def append(
             Data to be appended.
         metadata : `Optional[Any]`, default=None
             Optional metadata to persist along with the new symbol version. Note that the metadata is
-            not combined in any way with the metadata stored in the previous version.
+            not combined in any way with the metadata stored in the previous version. Additionally metadata is not stored
+            also when incomplet=True
```
IvoDD:
```suggestion
            also when incomplete=True
```

python/arcticdb/util/utils.py
```diff
@@ -86,6 +87,109 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
...
+        arr = np.array(arr, dtype=f"U{size}")
+    elif 'datetime' in str(_type):
+        arr = generate_random_timestamp_array(size, seed=seed)
+    else:
+        raise TypeError("Unsupported type {dtype}")        
+    return arr
+
+
+def generate_random_series(type: ArcticTypes, length: int, name: str, 
+                    start_time: pd.Timestamp=None, freq: str='s', seed=3247) -> pd.Series:
```
IvoDD:
```suggestion
                    start_time: Optional[pd.Timestamp]=None, freq: str='s', seed=3247) -> pd.Series:
```

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -259,13 +263,29 @@ def test_upsert_with_delete(lmdb_version_store_big_map):
...
+    logger = get_logger()
+    for index, _type in enumerate(supported_types_list):
+        sym = f"test_append_numpy_array_{index}"
+        logger.info(f"Storing type: {_type} in symbol: {sym}")
+        #np1 = random_integers(10, t)
```
IvoDD:
```suggestion
```

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -259,13 +263,29 @@ def test_upsert_with_delete(lmdb_version_store_big_map):
...
+        np1 = generate_random_numpy_array(10, _type)
+        try: 
+            lmdb_version_store.write(sym, np1)
+        except arcticdb.exceptions.ArcticDbNotYetImplemented as e:
```
IvoDD, line 276: I think this is because of the fixed strings handling in Windows?

If so we can try using `dynamic_strings=True`. In fact `dynamic_strings=True` is the default almost always. It is quite unfortunate if the `lmdb_version_store` fixutre has `dynamic_strings=False` default.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+    df.index = pd.date_range(start_time, periods=df.shape[0], freq='s')
+
+
+def make_df_appendable(original_df: pd.DataFrame, append_df:pd.DataFrame):
+    """ Creates such index for `append_df` so that it can be appended to `original_df`"""
+    last_time = original_df.index[-1]
+    add_index(append_df, last_time + timedelta(seconds=1))
+
+
+def create_all_arcticdb_types_df(length: int, column_prefix: str = ""):
```
IvoDD, line 52: Is there a difference between this and `arcticdb.util.test.sample_dataframe` apart from the columnp prefixes? If not, we can add an optional argument to the `sample_dataframe` to reuse that.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+    Verifies:
+     - append/update operations over symbol containing timestamped series
+     - append/update operations with different types of series - empty, one element, many elements
+     - tests repeated over each supported arcticdb type - ints, floats, str, bool, datetime
+     - tests work as expected over static and dynamic schema with small segment row size 
+    """
+    if LINUX and (sys.version_info[:2] == (3, 8)) and dtype == np.float64:
+        """ https://github.com/man-group/ArcticDB/actions/runs/16363364782/job/46235614310?pr=2470
+        """
+        print ("Test fails due to issue (9589648728), Skipping")
```
IvoDD: Better to use `pytest.skip`. This way it will appear as a skipped test instead of successful test.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+        },
+        "source": "integration_test",
+        "tags": ["v1.2", "regression"]
+    }
+    return [metadata, {}, [metadata, {}, [metadata, {}], 1], "", None]
+
+
+@SLOW_TESTS_MARK
+@pytest.mark.parametrize("dtype", supported_types_list)
+@pytest.mark.parametrize("schema", [True, False])
```
IvoDD: `dynamic_schema` I think is better than `schema`. This way it's clear it is for the library option and not somehow configuring the schema of the series.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+
+@pytest.mark.storage
+def test_append_update_dynamic_schema_add_columns_all_types(version_store_and_real_s3_basic_store_factory):
+    """
+    The test does series of append operations with new columns of all supported column types.
+    The resulting symbol will have additional columns each time added, with predefined default 
+    values for different data types
+
+    Verifies:
+     - continuous appends/updates with new columns works as expected across dynamic and static schema
```
IvoDD: This test does not test static schema?

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+            lib.append(symbol, append_df, incomplete=True, prune_previous_version=False, validate_index=True, metadata=meta)
+            lib.compact_incomplete(symbol, True, False)
+            assert num_versions_before, lib.list_versions(symbol) + 1 # previous version is not pruned
+
+        # Verify  metadata and dynamically added columns
+        ver = lib.read(symbol)
+        read_data:pd.DataFrame = ver.data
+        assert meta == ver.metadata
+        verify_dynamically_added_columns(read_data, -2, new_columns_to_symbol)
+        verify_dynamically_added_columns(read_data, -1, new_columns_to_appended_df)
```
IvoDD: Why not use `pd.concat` and `assert_frame_equal` to confirm the entire dataframe is as expected instead of just a couple of rows?

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+            meta = "just a message"
+            lib.update(symbol, update_df, prune_previous_version=True, metadata=meta)
+            assert num_versions_before, lib.list_versions(symbol) # previous version is pruned
+
+        # Verify  metadata and dynamically added columns
+        ver = lib.read(symbol)
+        read_data:pd.DataFrame = ver.data
+        assert meta == ver.metadata
+        verify_dynamically_added_columns(read_data, previous_timestamp, new_columns_to_symbol)
+        verify_dynamically_added_columns(read_data, update_timestamp, new_columns_to_update_df)
```
IvoDD: Similarly we can perform the updates artificially via `dataframe_simulate_arcticdb_update`?
I know currently `dataframe_simulate_arcicdb_update_static` works for static dataframes but I think it should work well enought for dynamic dataframes as well. Might need some extra work to get this right.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+    )
+
+    # Test append to non-existent symbol
+    for sym in [symbol, None, ""]:
+        try:
+            lib.append(sym)
+            assert False, "Expected exception was not raised"
+        except Exception as e:
+            # Verify exception type and message content
+            assert isinstance(e, TypeError)
```
IvoDD: This exception handling logic is hard to follow. E.g. the except call would catch the assertion error and only the follow up assert would fail. Can we use `pytest.raises` which is exactly designed to test that some code snippet raises the appropriate exception?

python/arcticdb/util/utils.py
```diff
@@ -86,6 +87,109 @@ def get_logger(bencmhark_cls: Union[str, Any] = None):
...
+    set_seed(seed)
+    index = None
+    if start_time:
+        index = pd.date_range(start_time, periods=length, freq=freq)
+    return pd.Series(generate_random_numpy_array(length, type, seed), 
+                           index=index,
+                           name=name)
+
+
+def verify_dynamically_added_columns(updated_df: pd.DataFrame, row_index: Union[int, pd.Timestamp],
```
IvoDD, line 136: Let's discuss, I think there is a better way to run the assertions on the scenario tests.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+    def run_test(lib, initial_data, start, end, update_expected_at_index, length_of_result_df, scenario):
+        nonlocal update_data
+        lib.write(symbol, initial_data) # always reset symbol
+        
+        lib.update(symbol, update_data, date_range=(start, end))
+        
+        result =  lib.read(symbol).data
+        
+        assert result.iloc[update_expected_at_index]["value"] == 999, f"Failed for {scenario} Scenario"
+        assert len(result) == length_of_result_df
```
IvoDD, line 391: Similarly I think we can calculate the exact expected dataframe and run `assert_frame_equal`

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -0,0 +1,584 @@
...
+                scenario="open start")
+
+    # Both start and end are open - in this case only the update will be the result
+    result_df = run_test(lib, initial_data, 
+                start=None, end=None, 
+                update_expected_at_index=0, length_of_result_df=update_data.shape[0], 
+                scenario="both open")
+    assert_frame_equal(update_data, result_df)
+
+    lib.list_symbols_with_incomplete_data
```
IvoDD:
```suggestion
```

python/arcticdb/util/utils.py
grusev: that was intention, I forgot it however, thanks for noticing this`! Removed from test_update.py

python/arcticdb/util/utils.py
grusev: we have to delete this with 10**9 otherwise it will not be timestamp ... So this is a requirement for keeping the number as timestamp then the precision of 's' is more humanly meangful, nothing more or less. Yes it could be ns no problem but most of timestamps would perhaps be 's' or 'ms' based

I will leave this as is

python/tests/unit/arcticdb/version_store/test_append.py
grusev: will see what I can do

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
grusev: I'd rather not touch sample_dataframe at this point of time.
I would rather implement sample_dataframe using new one, but that is for future. Lets keep them separate for now.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
grusev: will remove static

python/arcticdb/util/test.py
grusev: Since Pandas 1 is very low prio, I would do this which works: (no more things needed here)

```
            if (expected.size == 0) and (actual.size == 0):
                assert expected.name == actual.name
                # Compare the indexes as indexes without the frequency which can be None sometimes for some types (str)
                assert pd.Index(expected.index).equals(pd.Index(actual.index)), f"Investigate why {expected.index} == {actual.index}"
                return
```

python/arcticdb/util/test.py
grusev: That is the behavior of arcticdb with Pandas 1.x ... Do not know if this is documented or not, and as Pandas 1 is near end of life I will leave this as is

python/arcticdb/util/utils.py
grusev: actually the else is for python 3.8 added as comment

python/arcticdb/util/utils.py
grusev: there was no good random function with almost whole range, so I created something that is better than both and closer to what perhaps we need:

```
def generate_random_float_array(size: int, dtype: 
                                np.floating = np.float32 ):
    """Pseudo random float algorithm supporting np.float* types"""
    def power_sequence(max_power):
        exponents = np.arange(max_power, -(max_power+1), -1)
        sequence = np.power(10.0, exponents)
        return sequence
    
    if dtype == np.float32:
        power_arr = power_sequence(37)
    elif dtype == np.float16:
        power_arr = power_sequence(4)
    elif np.issubdtype(dtype, np.floating):
        power_arr = power_sequence(100)
    else:
        raise TypeError(f"Type not supported {dtype}")

    random_part = np.random.random(size).astype(dtype)
    scaling_part = np.random.choice(power_arr, size=size).astype(np.float64)
    result = random_part * scaling_part
    return result.astype(dtype)
```

python/arcticdb/util/utils.py
grusev: set_seed sets both lib seeds, and this function uses only one of the libs, so I intentionally not used set_seed. so that it does not screw up any test the function will be used

python/arcticdb/version_store/_store.py
```diff
@@ -744,7 +744,8 @@ def append(
             Data to be appended.
         metadata : `Optional[Any]`, default=None
             Optional metadata to persist along with the new symbol version. Note that the metadata is
-            not combined in any way with the metadata stored in the previous version.
+            not combined in any way with the metadata stored in the previous version. Additionally metadata is not stored
+            also when incomplete=True
```
ognyanstoimenov: `Additionally, metadata is not stored when incomplete=True` reads better

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+        if dynamic_schema:
+            result_df = chunks[0]
+            for i in range(1, len(chunks), 1):
+                result_df = cls.simulate_arctic_append(result_df, chunks[i])
+            return result_df
+        else:
+            return pd.concat(chunks)
+
+
+class ArcticSymbolSimulatorWrapper(ArcticSymbolSimulator):
```
ognyanstoimenov: I don't get the need for this wrapper

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+            result[col] = [ARCTICDB_NA_VALUE_STRING] * len(result) # string is None
+        elif pd.api.types.is_datetime64_any_dtype(col_type):
+            result[col] = ARCTICDB_NA_VALUE_TIMESTAMP
+        else:
+            raise TypeError(f"Unsupported column type in '{col}': {col_type}")
+
+    return result
+
+
+class ArcticSymbolSimulator:
```
ognyanstoimenov, line 64: I am a bit confused by this, should every test that expects certain result use the simulator and expand it? If this is to go in, I think it would be good to communicate with the others how to use it. When is it useful, when to modify it. Otherwise we are at risk of duplicating a lot of code, or not using this one at all. Also it is a big class that doesn't seem to be used that much in the tests even currently?

I see there is a lot of complicated logic here and I see that this class even has tests, so it adds a layer of complexity to writing simple tests. 
So if one is to develop a new feature, they have to write the source, write the simulator for it (whatever this means), and then write tests for both the code and simulator?


Maybe this needs a broader discussion?

ognyanstoimenov (changes requested):

python/arcticdb/util/arctic_simulator.py
grusev: The Wrapper class is only for tests that cover ArcticSymbolSimulator class: https://github.com/man-group/ArcticDB/pull/2470/files#diff-d849db03718b5cdb695647b1f93d9f58dd46cecb7a825aec2ba6c20931f30bd1

The main idea is that when you do changes they need to be mirrored on arctic symbol also.

And to be able to track quickly differences the operations are better done at one line. This is easier to track errors by mistake during creating a test. For example:

```
def test_simulator_append_serries_supported_combos(lmdb_library_dynamic_schema):

    def test_append_serries(s1, s2):
        lib = lmdb_library_dynamic_schema
        asym = ArcticSymbolSimulatorWrapper().associate_arctic_lib(lib)
        asym.write(s1).arctic_lib().write("s", s1)
        asym.append(s2).arctic_lib().append("s", s2)
        asym.assert_equal_to_associated_lib("s")
        return asym.read()

    s1 = pd.Series([10, 20, 30], name="name")
    s2 = pd.Series([100, 200, 300], name="name")
    test_append_serries(s1, s2)

    s1 = pd.Series([10, 20, 30])
    s2 = pd.Series([100, 200, 300])
    test_append_serries(s1, s2)

    # Series and Dataframe - results in Dataframe
    s1 = pd.Series([10, 20, 30], name="name")
    s2 = pd.DataFrame(pd.Series([100, 200, 300], name="name"))
    df: pd.DataFrame = test_append_serries(s1, s2)
    assert 1 == df.shape[1]

    # Series and Dataframe - results in Dataframe
    # when column names different - we have 2 cols
    s1 = pd.Series([10, 20, 30], name="name")
    s2 = pd.DataFrame(pd.Series([100, 200, 300], name="name2"))
    df: pd.DataFrame = test_append_serries(s1, s2)
    assert 2 == df.shape[1]
```

python/arcticdb/util/arctic_simulator.py
grusev: Good questions. I actually forgot to update the description for the simulator why it was developed and how it can help. We had discuss it Ivo and we agreed that now it is the time to create it - we had already some pieces implemented for different PRs, thus it made sense to make the leap and do some refactoring to make things more clean

The idea for the Simulator is based on so called Test Oracles. https://en.wikipedia.org/wiki/Test_oracle

It indeed looks like it adds complexity but in practices it removes complexity for validating the results of a complex solution.

Oracles are used a lot in Software Engineering in many areas. One excellent example of test oracle was used at MS when they had to rewrite Excel completely - so how to verify the new excel behaves as it should? Easy - just do same things on the old Excel and on the new Excel and compare - if something deviates then analyze as it is potential bug.

In our case we use simple high level programming to simulate what arctic does in much more complex way appends and updates especially with dynamic schema. How to verify all those things work as expected over arbitrary complex input data? Series? Dataframes with hundreds of columns and rows.

You can cover for sure the small dataframes because you can code both input and output data in the tests. 

But how can bring confidence through empirical evidences that this works at scale? The approach with hardcoded input and output is out of question.

To solve this we have to use Generators and Real Life samples. And have tests that can validate the results.

So this is the primary use case - use in testing with large scale data in higher abrstraction tests. Thus those oracles are part of system, stress etc testing and rarely in unit testing

There is also another use case of Test Oracles - they are built along side the real product with simple and cost effective methods, so that once the solution is ready the test are also ready and waiting. You create the tests against the oracle or use oracle directly in the tests.

How this is relevant to Arctic as we have already a product that is built? The product will elvolve and with evolution the logic of specific operations will become unbearble to handle well. In that case this oracle can once again be used - a prototype of solution could be done with some small scale changes those validated and then when you start actual rework and you finish you have set of tests 

This is usage scenario 2.

Both are critical as they remove huge impediments and although that it looks complex it is easier to catch a problem in couple hundred of lines than in a couple of hundred thousand lines of code.

What you gain is time for validation and tests ready at much lower cost than doing them manually.

ognyanstoimenov: Okay, I still think this needs to be shown to the team so that we don't just add it and never use it.
Also what about this?
> So if one is to develop a new feature, they have to write the source, write the simulator for it (whatever this means), and then write tests for both the code and simulator?

python/arcticdb/util/arctic_simulator.py
ognyanstoimenov: Do we need tests for the simulator class? Also can't the wrapper methods just be part of the Simulator itself?

grusev: Yes, we do need tests for simulator class, this is the only way to show it is doing exactly same thing. There is no other way to validated behavior is same unless using tests. 

Also, note that the tests constructed this was are testing not only Simulator they are doing same for the product under test functionality.

As an example one of the tests is xfail due to bug in arcticdb caught exactly with during construction of validation tests. And this bug was serious. 

To summarize: the tests validate that _both_ arcticdb and simulator are behaving as it should, they are not only for simulator.

>> Also can't the wrapper methods just be part of the Simulator itself?

They were initially. In a previous round of discussions with Ivo we decided that the Simulator should offer clean interface very close to arctic library, thus they are moved into wrapper class for that reason, accept them as extensions.

We also agreed that they will not be used in arcticdb tests so that the tests are more simple. 

Therefore the intention for the Wrapper is to be used for now only for Simulator tests, which we found out are not only about it but also for arcticdb, but there therei usage should be ok as the intention is to demonstrate each Simulator op is exactly same as the arcticdb op

python/arcticdb/util/arctic_simulator.py
grusev: The team will be slowly engaged with it one by one when time allows as each one is having their own tasks and focus. Everyone will be at some time on board with it, no sooner then they need to and have time to.

As for the usage of this class it is vital for higher abstraction tests. There is no alternative,

For unit tests - the approach with input and expected outputs hardcoded is more beneficial and should be continued

To summarize, this approach is not displace previous but it complements it and adds new opportunities that did not existed before

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+        as_of = as_of if as_of is not None else len(self._versions) - 1
+        assert as_of < len(self._versions)
+        df = self._versions[as_of]
+        return df.copy(deep=True) if df is not None else None
+    
+    def assert_equal_to(self, other_df_or_series: Union[pd.DataFrame, pd.Series]):
+        self.assert_frames_equal(self.read(), other_df_or_series)
+
+    @classmethod
+    def assert_frames_equal(self, expected: Union[pd.DataFrame, pd.Series], actual: Union[pd.DataFrame, pd.Series]):
```
IvoDD: Let's make it more clear what this method does. E.g. name it `assert_frame_equal_rebuild_index_first`?
Also why is it a `classmethod`? Seems like `staticmethod` would be more suitable. Not just for this but for all of the `simulate_*` methods as well.

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+
+class ArcticSymbolSimulatorWrapper(ArcticSymbolSimulator):
+    """The wrapper class provides ability to do one line operations with simulator and arcticdb
+
+    Use like:
+
+        asym = ArcticSymbolSimulator(keep_versions=True).associate_arctic_lib(lib)
+        asym.write(initial_df).arctic_lib().write(symbol, initial_df)
+        asym.append(initial_df).arctic_lib().append(symbol, initial_df)
+        asym.assert_equal_to_associated_lib(symbol)
```
IvoDD: I think the usage of the wrapper can be made more intuitive if:
- Rename `ArcticSymbolSimulatorWrapper` to `ArcticSimulatorAndLibrary`
- Add methods to the class which perform the write/update/append to both the Simulator and the library:
  - Now we need to: `wrapper.arctic_lib().write(sym, df)` and `wrapper.write(sym, df)` to make sure they have the same state.
  - It would be much simpler if `wrapper.write(sym, df)` calls both `arctic_lib().write()` and `super().write()`

This would allow easy writing of proper scenario tests like e.g.:
```
operations = [("WRITE", df_1), ("APPEND", df_2), ("UPDATE", df_3)] # This can be dynamically generated by e.g. hypothesis
for operation, df in operations:
    if operation == "WRITE":
        wrapper.write(sym, df)
    # Similar for update and append
wrapper.assert_equal(sym)
```

python/tests/unit/simulator/test_symbol_symulator.py
```diff
@@ -0,0 +1,311 @@
+"""
```
IvoDD: Nit: Typo in file name s**i**mulator

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+    cols_df1 = set(df1.columns)
+    cols_df2 = set(df2.columns)
+
+    # Determine column groups
+    only_in_df1 = list(cols_df1 - cols_df2)
+    only_in_df2 = list(cols_df2 - cols_df1)
+    common_cols = list(cols_df1 & cols_df2)
+
+    # Check type consistency for common columns
+    if not dynamic_schema:
```
IvoDD: We also want to verify `len(only_in_df1) == 0 and len(only_in_df2) == 0` for static schema. I.e. it does not allow adding new columns.

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+    @classmethod
+    def assert_frames_equal(self, expected: Union[pd.DataFrame, pd.Series], actual: Union[pd.DataFrame, pd.Series]):
+        if isinstance(expected, pd.Series) and isinstance(actual, pd.Series):
+            assert_series_equal_pandas_1(expected, actual)
+        else:
+            actual_df_same_col_sequence = actual[expected.columns]
+            assert_frame_equal_rebuild_index_first(expected, actual_df_same_col_sequence)
+
+    @classmethod
+    def simulate_arctic_append(cls, df1: Union[pd.DataFrame, pd.Series],
```
IvoDD: I think this can be simplified by using `pandas.concat`. I might be missing some complexity but would be interesting to experiment with something like:
```
if not dynamic_schema:
     assert_schema_equals(df1, df2)
ignore_index = isinstance(df1.index, RangeIndex):
result = pd.concat([df1, df2], ignore_index=ignore_index)
# And some additional logic around empty dataframes
# And maybe additonal logic to reorder the columns in the same way arcticdb would order them (i.e. leave the columns of df1 first and only add new columns from df2 at the end) 
return result
```

Some of the checks performed in this function are not performed by arcticdb's append I think. E.g.:
- Does arcticdb check range index steps are the same?
- I think if `df1.empty()` we would not just return `df2`. We would have df2 but with the schema from df1
- I think arcticdb raises after Vasil's fixes if we try to append a Series to a DataFrame

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+        chunks = []
+        df1 = existing_df[existing_df.index < start2]
+        chunks.append(df1)
+        chunks.append(update_df)
+        df2 = existing_df[existing_df.index > end2]
+        chunks.append(df2)
+        if dynamic_schema:
+            result_df = chunks[0]
+            for i in range(1, len(chunks), 1):
+                result_df = cls.simulate_arctic_append(result_df, chunks[i])
```
IvoDD: I like reusing the `simulate_arctic_append`. However I think this would not work precisely if some of the chunks are empty. I.e. the column ordering might get messed up if:

```
chunk[0] = empty
chunk[1] = update_df
chunk[2] = non-empty
```
Then the `chunk[1]` columns would appear before the `chunk[2]` columns which is not how arcticdb update would work I think

python/arcticdb/util/arctic_simulator.py
```diff
@@ -0,0 +1,331 @@
...
+            for col in source_float32_cols:
+                if append_df_to_df1[col].dtype == np.float64:
+                    append_df_to_df1[col] = append_df_to_df1[col].astype(np.float32)        
+
+            return append_df_to_df1
+
+    @classmethod
+    def simulate_arctic_update(cls, existing_df: Union[pd.DataFrame, pd.Series], 
+                               update_df: Union[pd.DataFrame, pd.Series],
+                               dynamic_schema: bool = True) -> Union[pd.DataFrame, pd.Series]:
```
IvoDD, line 154: We should also support the `date_range` argument

grusev: For sure, but lets make this in another PR with concrete test at hand. This PR is quite big already

python/arcticdb/util/arctic_simulator.py
grusev: Will filter out empty chunks

python/arcticdb/util/arctic_simulator.py
grusev: will fix

python/arcticdb/util/arctic_simulator.py
grusev: Lets leave the design for this out for now. I have simplified tests for simulator and now they use simple methods and directly ArcticSymbolSimulator. 

Along with that I had to refactor the tests due to change of arcticdb which does not allow now mixinf Dataframes and Series

As a result at this point the tests are very close to other tests that we have

python/arcticdb/util/arctic_simulator.py
grusev: I think that could wait. We have a working simulator with well defined interface.

Thus any simplification and optimizations of its internal structure can happen at any point in future. SimulatorV2 can easily withou problem replace SimulatorV1 at any point of time we have good and optimized and more understandable solution

python/tests/integration/arcticdb/version_store/test_snapshot.py
```diff
@@ -839,7 +839,7 @@ def test_delete_snapshot_on_updated_and_appended_dataframe(basic_store_tiny_segm
     # this dataframe has both values before the date and within the dates of df_1
     df_2 = create_df_index_datetime(10, 1, 5)
     df_3 = create_df_index_datetime(10, 20, 30)
-    df_updated = dataframe_simulate_arcticdb_update_static(df_1, df_2)
+    df_updated = ArcticSymbolSimulator().simulate_arctic_update(df_1, df_2, dynamic_schema=False)
```
ognyanstoimenov: no need to instantiate it here as it is static

python/tests/integration/arcticdb/test_read_batch_more.py
```diff
@@ -61,9 +61,9 @@ def test_read_batch_2tables_7reads_different_slices(arctic_library):
     df1_1 = create_df_index_datetime(num_columns=7, start_hour=4, end_hour=6)
     df1_2 = create_df_index_datetime(num_columns=7, start_hour=6, end_hour=10)
     df1_3 = create_df_index_datetime(num_columns=7, start_hour=0, end_hour=10)
-    df1_till2 = dataframe_simulate_arcticdb_update_static(df1_0, df1_1)  # DF of state 0+1
+    df1_till2 = ArcticSymbolSimulator().simulate_arctic_update(df1_0, df1_1, dynamic_schema=False)  # DF of state 0+1
```
ognyanstoimenov: remove ()

ognyanstoimenov: this applies to couple of other places as well

ognyanstoimenov (commented):
I am fine with most of it as we discussed offlline. Left couple of minor comments. Still not super convinced that testing the simulator against arctic (in test_symbol_simulator.py) while also testing arctic against the simulator makes sence but maybe it is just me I am new to the test oracle approach. 

Also I think it may be worth keeping the whole simulator as static class (without keeping versions or any state) so that it is clearer what is happening in the tests, although that would need some refactoring to the existing tests

python/arcticdb/util/utils.py
```diff
@@ -8,23 +8,157 @@
...
+    else:
+        raise TypeError(f"Type not supported {dtype}")
+
+    random_part = np.random.random(size).astype(dtype)
+    scaling_part = np.random.choice(power_arr, size=size).astype(np.float64)
+    result = random_part * scaling_part
+    return result.astype(dtype)
+
+
+def generate_random_numpy_array(size: int, dtype, seed: int = 8238) -> npt.NDArray[Any]:
```
IvoDD: We should probably accept an optional seed. I.e. `seed: Optional[int]`

python/arcticdb/util/utils.py
```diff
@@ -8,23 +8,157 @@
...
+        arr = generate_random_timestamp_array(size, seed=seed)
+    else:
+        raise TypeError("Unsupported type {dtype}")        
+    return arr
+
+
+def generate_random_series(type: ArcticTypes, length: int, name: str, 
+                    start_time: Optional[pd.Timestamp]=None, freq: str='s', seed=3247) -> pd.Series:
+    """Generates random series of specified type with or without index"""
+    set_seed(seed)
```
IvoDD, line 127: This `set_seed` doesn't do much given that we call the `set_seed` in `generate_random_numpy_array`.

We should think more about seeding but that can be done in a separate PR. Let's maybe open a ticket for this to discuss later.
- I think setting seeds is best done in most top level function call. This allows multiple calls to the same function to give different results. E.g. if I'm generating a dataframe and use `generate_random_series` to generate two columns for it they will be the same if I don't specifically set `seed=None`.
- Implication is that by default `seed=None` for 90% of functions.

python/arcticdb/util/utils.py
IvoDD: Now that we have the symbol simulator can we remove `verify_dynamically_added_columns`?

python/tests/integration/arcticdb/test_update.py
```diff
@@ -531,8 +499,8 @@ def test_update_batch_different_updates_dynamic_schema(custom_library):
     start_time = TimestampNumber.from_timestamp(timestamp=Timestamp("1/2/2017"))
     g = BasicDataFrameGenerator()
     ug = UpdatesGenerator(g)
-    original_num_cols = 30
-    original_number_rows = 50
+    original_num_cols = 5
+    original_number_rows = 5
```
IvoDD: Why decrease num cols and num rows?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -260,13 +265,28 @@ def test_upsert_with_delete(lmdb_version_store_big_map):
...
-    np1 = random_integers(10, np.uint32)
-    lmdb_version_store.write("test_append_numpy_array", np1)
-    np2 = random_integers(10, np.uint32)
-    lmdb_version_store.append("test_append_numpy_array", np2)
-    vit = lmdb_version_store.read("test_append_numpy_array")
-    expected = np.concatenate((np1, np2))
-    assert_array_equal(vit.data, expected)
+    '''Tests append with all supported by arctic data types'''
+    logger = get_logger()
+    for index, _type in enumerate(supported_types_list):
```
IvoDD, line 270: Nit: Might be easier to read if we pytest parameterize over the `supported_types_list`. Doesn't matter much.

python/tests/unit/arcticdb/version_store/test_append.py
IvoDD: Looks like we missed this. I think if we `lmdb_version_store.write(sym, np1, dynamic_strings=True)` we can get rid of the try except.

python/tests/unit/simulator/test_symbol_simulator.py
```diff
@@ -0,0 +1,359 @@
...
+def test_simulator_append_serries_supported_errors(lmdb_library_dynamic_schema):
+
+    def test_append_serries_with_error(s1, s2):
+        lib = lmdb_library_dynamic_schema
+        asim = ArcticSymbolSimulator()
+        asim.write(s1)
+        lib.write("s", s1)
+        asim.append(s2)
+        with pytest.raises(SchemaException):
+            lib.append("s", s2)
```
IvoDD, line 354: Can be left for a follow up PR but it would be good to make simulator behave as identical to the actual arctic as possible.
E.g. in this case raise the same exceptions as the library. (unless we have a bug in arctic which needs to be fixed)

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
IvoDD: I'm ok with this in this case as this is only used in this file. 

But in general we should be thinking in the direction of having a more unified test data generation. Similar to how we now have too many fixutres and we have to spend a significant effort to simplify them.

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
IvoDD: We can use the `ArcticSymbolSimulator` now to assert the entire dataframe is as expected?

python/arcticdb/util/utils.py
grusev: I aztually do agree. I can say it in different way

- Each function should be able to set seed if asked,  As first line
- Then if it is calling other functions that accept seed, it must propagate sed=None

I have applied that here and will open a PR on this

python/tests/integration/arcticdb/test_update.py
grusev: Leftovers from debugging a problem experienced during refactoring. Thanks!

python/tests/unit/simulator/test_symbol_simulator.py
grusev: Yes this is for further discussion. Note that creation and maintenance of simulators could be very time consuming if we go for complete simulation like simulating also exceptions and other stuff. I would rather leave such out initially as they will be rarely used but if added will add more cost than benefit.

IvoDD (approved):

PR #2474, opened 2025-07-21
Bugfix 8083916814: Respect pickle_on_failure kwarg

alexowens90:
#### Reference Issues/PRs
[8083916814](https://man312219.monday.com/boards/7852509418/pulses/8083916814)

The `pickle_on_failure` argument to `write` and `batch_write` on the V1 API was not respected, and always defaulted to the lib config setting.

### DIFF
--- a/python/arcticdb/flattener.py
+++ b/python/arcticdb/flattener.py
@@ -117,27 +117,6 @@ class Flattener:
     def deserialize_primitives(obj):
         return _msgpack_compat.unpackb(obj, raw=False)
 
-    def will_obj_be_partially_pickled(self, obj):
-        to_write = dict()
-        self._create_meta_structure(obj, "dummy", to_write)
-        msgpack_normalizer = MsgPackNormalizer()
-        msgpack_normalizer.strict_mode = True  # To prevent msgpack from falling back to pickle silently
-        base_normalizer = CompositeNormalizer(msgpack_normalizer, use_norm_failure_handler_known_types=False)
-        for sym, obj_to_write in to_write.items():
-            try:
-                opt_custom = self.custom_normalizer.normalize(obj_to_write)
-                if opt_custom is not None:
-                    item, custom_norm_meta = opt_custom
-                    base_normalizer.normalize(item, pickle_on_failure=False)
-                else:
-                    base_normalizer.normalize(obj_to_write, pickle_on_failure=False)
-                # Note that we are fine with msgpack serialization, but not fall back to pickle for msgpack.
-            except Exception:
-                log.info("{} with key {} will be pickled".format(obj_to_write, sym))
-                return True
-
-        return False
-
     def _create_meta_structure(self, obj, sym, to_write, depth=0, original_symbol=None):
         if original_symbol is None:
             original_symbol = sym  # just used for error messages


python/tests/unit/arcticdb/version_store/test_pickle_on_failure.py
```diff
@@ -0,0 +1,85 @@
...
+        monkeypatch.setenv("USE_NORM_FAILURE_HANDLER_KNOWN_TYPES", "true")
+        should_succeed = True
+    if env_var_2:
+        monkeypatch.setenv("PICKLE_ON_FAILURE", "true")
+        should_succeed = True
+    if arg is not None:
+        should_succeed = arg
+    assert should_succeed == lib._resolve_pickle_on_failure(arg)
+
+    # Normalizable type with non-normalizable object column type
```
poodlewars, line 36: Tests for non-normalizable types with (non-) normalizable contents?

python/arcticdb/version_store/_normalization.py
```diff
@@ -1165,18 +1165,12 @@ def denormalize(self, item, norm_meta):
...
 
 class CompositeNormalizer(Normalizer):
-    def __init__(self, fallback_normalizer=None, use_norm_failure_handler_known_types=False):
+    def __init__(self, fallback_normalizer=None):
         self.df = DataFrameNormalizer()
         self.series = SeriesNormalizer()
         self.tf = TimeFrameNormalizer()
         self.np = NdArrayNormalizer()
 
-        if use_norm_failure_handler_known_types and fallback_normalizer is not None:
```
poodlewars, line 1177: IIUC the intent of this block was that if we see a DataFrame, Series, TimeFrame or NdArray that we cannot normalize natively then we use the `fallback_normalizer`. `get_normalizer_for_type` has this logic,

```
    def get_normalizer_for_type(self, item):
        # TODO: this should use customcompositenormalizer as well.
        if isinstance(item, DataFrame):
            if (
                item.empty
                and not isinstance(self.df, KnownTypeFallbackOnError)
                and self.fallback_normalizer is not None
            ):
                return KnownTypeFallbackOnError(self.df, self.fallback_normalizer).normalize
            return self.df.normalize

        if isinstance(item, Series):
            if (
                item.empty
                and not isinstance(self.series, KnownTypeFallbackOnError)
                and self.fallback_normalizer is not None
            ):
                return KnownTypeFallbackOnError(self.series, self.fallback_normalizer).normalize

            return self.series.normalize

        if isinstance(item, TimeFrame):
            return self.tf.normalize

        if isinstance(item, np.ndarray):
            return self.np.normalize

        if self.fallback_normalizer is not None:
            # Msgpack normalize if everything else fails.
            return self.fallback_normalizer.normalize

        return None
```

so I don't see how the fallback normalizer will now be applied to dataframes, series, timeframes or ndarrays?

`flattener.py` has

```
def will_obj_be_partially_pickled(self, obj):
  ...
  base_normalizer = CompositeNormalizer(msgpack_normalizer, use_norm_failure_handler_known_types=False)
  ...
```

hasn't the constructor change broken this line?

poodlewars: BTW there are some tests that create `CompositeNormalizer` objects in `man.arcticdb` that will need updating

poodlewars (approved):

phoebusm (approved):

PR #2479, opened 2025-07-22
maint: Replace Folly's ranges with the standard library's

jjerphan:
#### Reference Issues/PRs

Part of https://github.com/man-group/ArcticDB/issues/1412.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #2480, opened 2025-07-22
Workaround folly issue with clang 19

JohanMabille:
This workaround allows to build ArcticDB with clang 19 and revert the pinning done in #2476


JohanMabille:
So it looks like folly/Range is instantiated by other parts of folly. Closing this one in favor of https://github.com/man-group/ArcticDB/pull/2479 (which does not allow to remove the pin yet).

PR #2483, opened 2025-07-22
maint: Replace `folly/gen/Base.h`

jjerphan:
#### Reference Issues/PRs

Part of https://github.com/man-group/ArcticDB/issues/1412.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+template<typename Container, typename KeyFunc>
+auto group_by(Container&& container, KeyFunc&& key_func) {
+    // Use decltype to get the correct value type from the range
+    using ValueType = std::remove_reference_t<decltype(*std::begin(container))>;
+    using KeyType = std::invoke_result_t<KeyFunc, ValueType>;
+    
+    std::unordered_map<KeyType, std::vector<ValueType>> result;
+    
+    for (auto&& item : std::forward<Container>(container)) {
+        auto key = std::invoke(std::forward<KeyFunc>(key_func), item);
```
vasil-pashov, line 24: I don't think we have to forward key_func here. If key_func is rvalue ref forwarding will result in move on the first call to std::invoke and latter calls to invoke will end up being use after move.

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+// Standard library replacement for folly::gen::groupBy
+template<typename Container, typename KeyFunc>
+auto group_by(Container&& container, KeyFunc&& key_func) {
+    // Use decltype to get the correct value type from the range
+    using ValueType = std::remove_reference_t<decltype(*std::begin(container))>;
+    using KeyType = std::invoke_result_t<KeyFunc, ValueType>;
+    
+    std::unordered_map<KeyType, std::vector<ValueType>> result;
+    
+    for (auto&& item : std::forward<Container>(container)) {
```
vasil-pashov, line 23: Ranged for loops can be deceiving. Unfortunately this will result in always copying the contents of container. You'd have to use move iterators like so:
```c++
    std::ranges::subrange moving_range(std::make_move_iterator(container.begin()), std::make_move_iterator(container.end()));
    for(auto&& item : moving_range) {
        ...
        result[std::move(key)].push_back(std::forward<ValueType>(item));
    }
```

Check this out to verify: https://godbolt.org/z/eE8cso6zc

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+// Standard library replacement for folly::gen::zip with move semantics
+template<typename Container1, typename Container2>
+auto zip_move(Container1&& container1, Container2&& container2) {
+    using ValueType1 = std::remove_reference_t<decltype(*std::begin(container1))>;
+    using ValueType2 = std::remove_reference_t<decltype(*std::begin(container2))>;
+    
+    std::vector<std::pair<ValueType1, ValueType2>> result;
+    result.reserve(std::min(container1.size(), container2.size()));
+    
+    auto it1 = std::begin(container1);
```
vasil-pashov, line 40: nit: I'd use move iterators here as well. This way you don't have to call std::move on the items below. Both are fine options though.

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+        ++it1;
+        ++it2;
+    }
+    
+    return result;
+}
+
+// Helper for foreach on grouped data
+template<typename GroupedContainer, typename Func>
+void foreach_group(GroupedContainer&& grouped, Func&& func) {
```
vasil-pashov, line 56: Same as in group_by regarding moving func and using move iterators on the container

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+    for (auto&& [key, values] : std::forward<GroupedContainer>(grouped)) {
+        std::invoke(std::forward<Func>(func), key, std::forward<decltype(values)>(values));
+    }
+}
+
+// Standard library replacement for folly::gen::from with move semantics
+template<typename Container>
+auto from_move(Container&& container) {
```
vasil-pashov, line 64: I don't understand the use cases for this. But it also doesn't seem to be used anywhere. Can we remove it.

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+namespace arcticdb::util {
+
+// Standard library replacement for folly::gen::groupBy
+template<typename Container, typename KeyFunc>
+auto group_by(Container&& container, KeyFunc&& key_func) {
+    // Use decltype to get the correct value type from the range
+    using ValueType = std::remove_reference_t<decltype(*std::begin(container))>;
+    using KeyType = std::invoke_result_t<KeyFunc, ValueType>;
+    
+    std::unordered_map<KeyType, std::vector<ValueType>> result;
```
vasil-pashov, line 21: nit: an option to specify the map type as a template template parameter would be cool.

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+// Standard library replacement for folly::gen::map
+template<typename Container, typename Func>
+auto map(Container&& container, Func&& func) {
+    using InputType = std::remove_reference_t<decltype(*std::begin(container))>;
+    using OutputType = std::invoke_result_t<Func, InputType>;
+    
+    std::vector<OutputType> result;
+    result.reserve(container.size());
+    
+    for (auto&& item : std::forward<Container>(container)) {
```
vasil-pashov, line 77: Nit: use std::transform (or ranges::transform) when we switch to C++23 we could also combine ranges::transform and ranges::to.

https://github.com/man-group/ArcticDB/pull/2483/files#r2224897222

I also think that the output container should be the same as in the input instead of getting always a vector.

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+// Standard library replacement for folly::gen::filter
+template<typename Container, typename Func>
+auto filter(Container&& container, Func&& func) {
+    using ValueType = std::remove_reference_t<decltype(*std::begin(container))>;
+    
+    std::vector<ValueType> result;
+    result.reserve(container.size()); // Reserve for worst case
+    
+    for (auto&& item : std::forward<Container>(container)) {
```
vasil-pashov, line 92: Use std::ranges::copy_if with move iterators. I'd also pass std::back_inserter(result) to copy_if and don't bother with reserving the worst case and then calling shrink to fit.

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+    
+    result.shrink_to_fit();
+    return result;
+}
+
+// Standard library replacement for folly::gen::as<std::vector>
+template<typename Container>
+auto as_vector(Container&& container) {
```
vasil-pashov, line 104: This is not used but it also definitely doesn't the input into vector

cpp/arcticdb/util/std_ranges_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+    return std::forward<Container>(container);
+}
+
+// Generator-like interface for lazy evaluation
+template<typename T>
```
vasil-pashov, line 109: Are generators used anywhere in the code? If not can we remove them?

cpp/arcticdb/util/std_ranges_utils.hpp
vasil-pashov, line 1: It would also be good if we had some unit tests for these util functions

cpp/arcticdb/storage/azure/azure_storage.cpp
```diff
@@ -237,18 +235,22 @@ void do_remove_impl(
...
-                auto blob_name = object_path(b.bucketize(key_type_dir, *k), *k);
-                to_delete.emplace_back(std::move(blob_name));
-                if (to_delete.size() == delete_object_limit) {
-                    submit_batch(to_delete);
-                }
+    // Group keys by type using standard library
+    auto grouped_keys = arcticdb::util::group_by(variant_keys, fmt_db);
+    
+    // Process each group
+    arcticdb::util::foreach_group(grouped_keys, [&](auto&& key_type, auto&& keys) {
```
vasil-pashov, line 242: I'd use std::move(grouped_keys) just to be more explicit

cpp/arcticdb/util/std_ranges_utils.hpp
vasil-pashov: I'm still wondering if that's the design we want to go as:
```c++
auto container = SomeContainer();
group_by(container, some_func);
```
This will result in moving the contents of container when it creates the groups, which IMO is not obvious. Maybe a more obvious solution would be to keep this but rename it to something like group_by_helper and then have group_by with two overloads one for const ref and one for rvalue ref. This way it'll be obvious when moving is happening.

cpp/arcticdb/storage/mongo/mongo_storage.cpp
```diff
@@ -151,13 +151,16 @@ bool MongoStorage::do_fast_delete() {
...
     ARCTICDB_SAMPLE(MongoStorageRemove, 0)
     std::vector<VariantKey> keys_not_found;
 
-    (fg::from(variant_keys) | fg::move | fg::groupBy(fmt_db)).foreach([&](auto &&group) {
-        for (auto &k : group.values()) {
+    // Group keys by type using standard library
+    auto grouped_keys = arcticdb::util::group_by(variant_keys, fmt_db);
+    
+    // Process each group
+    arcticdb::util::foreach_group(grouped_keys, [&](ARCTICDB_UNUSED auto&& key_type, auto&& keys) {
```
vasil-pashov, line 162:
```suggestion
    arcticdb::util::foreach_group(grouped_keys, [&](auto&&, auto&& keys) {
```

cpp/arcticdb/stream/stream_reader.hpp
```diff
@@ -116,29 +116,32 @@ class StreamReader {
     }
 
     auto generate_rows() {
-        return folly::gen::from(key_gen_())
-            | generate_segments_from_keys(*store_, IDX_PREFETCH_WINDOW, opts_)
-            | generate_keys_from_segments(*store_, entity::KeyType::TABLE_DATA, entity::KeyType::TABLE_INDEX)
-            | generate_segments_from_keys(*store_, DATA_PREFETCH_WINDOW, opts_)
-            | generate_rows_from_data_segments();
+        auto keys = key_gen_();
```
vasil-pashov, line 119: It looks like the previous version was lazy evaluated. Can you verify that this new version is also lazy?

cpp/arcticdb/stream/stream_utils.hpp
```diff
@@ -202,56 +202,62 @@ class KeyRangeIterator : public IndexRangeFilter {
 
 inline auto generate_segments_from_keys(
     arcticdb::stream::StreamSource &read_store,
-    std::size_t prefetch_window,
+    ARCTICDB_UNUSED std::size_t prefetch_window,
```
vasil-pashov, line 205:
```suggestion
    std::size_t,
```

cpp/arcticdb/stream/stream_utils.hpp
```diff
@@ -202,56 +202,62 @@ class KeyRangeIterator : public IndexRangeFilter {
...
     const storage::ReadKeyOpts& opts) {
-    using namespace folly::gen;
-    return
-        map([&read_store](auto &&key) {
+
+    return [&read_store, opts](auto&& keys) {
+        std::vector<std::pair<entity::VariantKey, SegmentInMemory>> result;
+        result.reserve(keys.size());
+
+        for (auto&& key : keys) {
```
vasil-pashov, line 212: use move iterators

cpp/arcticdb/stream/stream_utils.hpp
```diff
@@ -202,56 +202,62 @@ class KeyRangeIterator : public IndexRangeFilter {
...
 inline auto generate_segments_from_keys(
     arcticdb::stream::StreamSource &read_store,
-    std::size_t prefetch_window,
+    ARCTICDB_UNUSED std::size_t prefetch_window,
     const storage::ReadKeyOpts& opts) {
-    using namespace folly::gen;
-    return
-        map([&read_store](auto &&key) {
+
+    return [&read_store, opts](auto&& keys) {
```
vasil-pashov, line 208: Returning lambda capturing a ref makes me feel uneasy. I know it was like this before but can you check that we're not trying to use this returned lambda after that reference becomes invalid.

cpp/arcticdb/stream/stream_utils.hpp
```diff
@@ -202,56 +202,62 @@ class KeyRangeIterator : public IndexRangeFilter {
...
-        return folly::gen::detail::GeneratorBuilder<entity::AtomKey>() + [&](auto &&yield) {
-            std::stack<std::pair<entity::VariantKey, SegmentInMemory>> key_segs;
-            key_segs.push(std::forward<decltype(key_seg)>(key_seg));
-            while(!key_segs.empty()) {
-                auto [key, seg] = std::move(key_segs.top());
-                key_segs.pop();
+    return [expected_key_type, expected_index_type, &read_store](auto&& key_segs) {
+        std::vector<entity::AtomKey> result;
+
+        for (auto&& key_seg : key_segs) {
```
vasil-pashov, line 236: move iterators

vasil-pashov (changes requested):

PR #2488, opened 2025-07-23
Arrow output normalization

IvoDD:
Makes sure arrow output is consistent with pandas normalization metadata.
Change includes:
- properly constructing timestamp type in C++ layer
- Introduction of ArrowTableNormalizer
    - Handles index column timezones
    - Handles index column renames
    - Constructs `pyarrow.Schema.pandas_metadata` which `pyarrow` can then use to reconstruct the pandas dataframe.
- Arrow normalization tests
- Simplifies testing logic for existing tests by allowing to reuse `assert_frame_equal`

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -52,6 +52,22 @@ inline sparrow::primitive_array<bool> create_primitive_array(
...
 
+template <typename T>
+sparrow::timestamp_without_timezone_nanoseconds_array create_timestamp_array(
+        T* data_ptr,
+        size_t data_size,
+        std::optional<sparrow::validity_bitmap>& validity_bitmap) {
+    // We default to using timestamps without timezones. If the normalization metadata contains a timezone it will be
+    // applied during normalization in python layer.
+    sparrow::u8_buffer<sparrow::zoned_time_without_timezone_nanoseconds> buffer(
+        reinterpret_cast<sparrow::zoned_time_without_timezone_nanoseconds*>(data_ptr), data_size);
```
alexowens90, line 63: Presumably this relies on `zoned_time_without_timezone_nanoseconds` also being 64 bit? Could add a static assert to this effect

python/arcticdb/version_store/_normalization.py
```diff
@@ -583,6 +584,204 @@ def denormalize(self, item, norm_meta):
...
+            else:
+                multi_index_meta = pandas_meta.multi_index
+                pandas_indexes = multi_index_meta.field_count+1
+                fake_field_pos = set(multi_index_meta.fake_field_pos)
+                for index_col_idx in range(pandas_indexes):
+                    index_level_num = index_col_idx + 1
+
+                    tz = multi_index_meta.timezone.get(index_col_idx, "")
+                    if tz != "":
+                        timezones[index_col_idx] = tz
```
alexowens90: I think the top-level index timezone (if present) is in [this](https://github.com/man-group/ArcticDB/blob/master/cpp/proto/arcticc/pb2/descriptors.proto#L140) protobuf field (although if you have tests that prove me wrong happy to be overruled)

python/arcticdb/version_store/_normalization.py
```diff
@@ -583,6 +584,204 @@ def denormalize(self, item, norm_meta):
...
+    pandas_indexes: Optional[int]
+        Num index columns to place in pandas_metadata
+    pandas_renames: Mapping[int, Union[int, str, None]]
+        Column renames which only apply to pandas_metadata. E.g. renaming a column to an int
+    """
+    renames : Mapping[int, str]
+    timezones: Mapping[int, str]
+    range_index: Optional[Dict[str, Any]]
+    pandas_indexes: Optional[int]
+    pandas_renames: Mapping[int, Union[int, str, None]]
```
alexowens90: `renames` and `pandas_renames` is a bit confusing, they're both about pandas really

python/arcticdb/version_store/_normalization.py
```diff
@@ -583,6 +584,204 @@ def denormalize(self, item, norm_meta):
...
+                        pandas_renames[i] = ""
+                    elif col_data.is_int:
+                        pandas_renames[i] = int(col)
+                    elif col_data.original_name != col:
+                        pandas_renames[i] = col_data.original_name
+
+
+        elif input_type == "series":
+            frame_meta = norm_meta.series
+            log.warn(f"No normalization was applied to series metadata when reading as arrow: {frame_meta}")
```
alexowens90: Series can have all the same index metadata as dataframes?

python/arcticdb/version_store/_normalization.py
```diff
@@ -583,6 +584,204 @@ def denormalize(self, item, norm_meta):
...
+        for i, col in enumerate(table.columns):
+            field = table.field(i)
+            if i in op.renames:
+                field = field.with_name(op.renames[i])
+            if i in op.timezones:
+                timezone = op.timezones[i]
+                # All arcticdb timestamps are stored as UTC for timezone aware timestamps.
+                col = pa.compute.assume_timezone(col, timezone="UTC")
+                col = col.cast(pa.timestamp("ns", timezone))
+                field = field.with_type(pa.timestamp("ns", timezone))
```
alexowens90, line 701: When profiling, we should see if these 3 lines are trivial or costly

python/arcticdb/version_store/_normalization.py
```diff
@@ -583,6 +584,204 @@ def denormalize(self, item, norm_meta):
...
+            new_columns.append(col)
+            new_fields.append(field)
```
alexowens90, line 703: Zero copy?

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
```diff
@@ -0,0 +1,94 @@
...
+    lib.write("arrow", df)
+    table = lib.read("arrow", _output_format=OutputFormat.ARROW).data
+    assert table.schema.field(0).name == "index1"
+    assert table.schema.field(0).type == pa.dictionary(pa.int32(), pa.large_string())
+    assert table.schema.field(1).name == "index2"
+    assert table.schema.field(1).type == pa.int64()
+    assert_frame_equal_with_arrow(table, df)
+
+
+def test_multi_index_with_tz(lmdb_version_store_v1):
```
alexowens90, line 135: See comment above, I think the primary index timezone isn't stored in the map, we should add a test either way

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
alexowens90, line 1: Additional tests I think would be useful:

- Range index (both default constructed, and with non-default start/step)
- Datetime index column with no name (and what happens if there is another column called `index`)`
- Multiindex with various combinations of names set/unset

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
```diff
@@ -0,0 +1,94 @@
...
+    assert table.schema.field(0).type == pa.float64()
+    assert table.schema.field(1).name == "y"
+    assert table.schema.field(1).type == pa.float64()
+    assert table.schema.field(2).name == "__col_x__2"
+    assert table.schema.field(2).type == pa.float64()
+    assert_frame_equal_with_arrow(table, df)
+
+def test_int_column_name(lmdb_version_store_v1):
+    lib = lmdb_version_store_v1
+    df = pd.DataFrame(np.arange(30, dtype=np.float64).reshape(10, 3), columns=[1, 2, "x"])
```
alexowens90, line 187: Is there a weird corner case where `columns=[1, "1"]`?

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
```diff
@@ -0,0 +1,94 @@
...
+    assert table.schema.field(1).type == pa.timestamp("ns", "America/Los_Angeles")
+    assert_frame_equal_with_arrow(table, df)
+
+
+def test_duplicate_column_name(lmdb_version_store_v1):
+    lib = lmdb_version_store_v1
+    df = pd.DataFrame(np.arange(30, dtype=np.float64).reshape(10, 3), columns=["x", "y", "x"])
+    lib.write("arrow", df)
+    table = lib.read("arrow", _output_format=OutputFormat.ARROW).data
+    assert table.schema.field(0).name == "__col_x__0"
```
alexowens90, line 177: I'm guessing this is our internal representation when there are duplicated column names? How does `pa.Table.from_pandas` handle this case?

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
```diff
@@ -0,0 +1,94 @@
...
+    assert table.schema.field(2).type == pa.float64()
+    assert_frame_equal_with_arrow(table, df)
+
+def test_int_column_name(lmdb_version_store_v1):
+    lib = lmdb_version_store_v1
+    df = pd.DataFrame(np.arange(30, dtype=np.float64).reshape(10, 3), columns=[1, 2, "x"])
+    lib.write("arrow", df)
+    table = lib.read("arrow", _output_format=OutputFormat.ARROW).data
+    expected_df = lib.read("arrow", _output_format=OutputFormat.PANDAS).data
+    assert table.schema.field(0).name == "1"
```
alexowens90, line 190: How does pa.Table.from_pandas handle this case?

python/arcticdb/version_store/_normalization.py
IvoDD: Copying a column like this is zero-copy because it just copies the pointer to the data instead of copy-ing the entire array.

Proof that copying a table like this is cheap and only when casting is a genuine memcopy required: 
image

python/arcticdb/version_store/_normalization.py
IvoDD: They require a memcopy (as per comment below), so indeed are not efficient. 

However it would require some plumbing to make the timezone from normalization metadata available at the interaction with sparrow. Even then will need to look at arrow spec whether timezoned data is stored as nanoseconds in UTC.

I'm thinking of leaving as-is and when we have thorough benchmarks to decide whether to invest time in saving this memcopy in case it takes a big chunk of time.

python/arcticdb/version_store/_normalization.py
IvoDD: Updated:
```
    Attributes
    ----------
    renames_for_table : Mapping[int, str]
        Column renames coming from normalization metadata to be applied to the pyarrow table. E.g. index names
    renames_for_pandas_metadata: Mapping[int, Union[int, str, None]]
        Column renames which can only be applied to pandas_metadata. E.g. renaming a column to an int

```

python/arcticdb/version_store/_normalization.py
IvoDD: Yeah was thinking to handle together with 9360502457 but made it so we just return a pyarrow.Table with the same index denormalization.

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
IvoDD: Added the tests. Also opened #2539 with a couple of issues with pandas normalization, which I haven't tested here as they can't be written.

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
IvoDD: Good question, it's a bug during normalization, see #2539

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
IvoDD: It raises:
```
>>> df = pd.DataFrame(np.arange(6).resize(3, 2), columns=["x", "x"])
>>> pa.Table.from_pandas(df)
ValueError: Duplicate column names found: ['x', 'x']
```
But we should handle this case gracefully as we can have data already written like so in storage.

IvoDD: It's also worth noting that we do have different column naming to pyarrow for unnamed indices. I don't think we should strictly mimic `pyarrow.Table.from_pandas` in such cases. As long as pandas normalization metadata allows us to round trip correctly it should be fine.

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
IvoDD: It converts to "1":
```
>>> df = pd.DataFrame(np.arange(6).resize(3, 2), columns=["x", 1])
>>> pa.Table.from_pandas(df).column_names
<stdin>:1: UserWarning: The DataFrame has column names of mixed type. They will be converted to strings and not roundtrip correctly.
['x', '1']
```

alexowens90 (approved):

cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -51,6 +51,23 @@ inline sparrow::primitive_array<bool> create_primitive_array(
     }
 }
 
+template <typename T>
+sparrow::timestamp_without_timezone_nanoseconds_array create_timestamp_array(
+        T* data_ptr,
+        size_t data_size,
+        std::optional<sparrow::validity_bitmap>& validity_bitmap) {
```
vasil-pashov: nit: passing by rvalue ref makes it obvious that this shouldn't be used after the call. This way it's prone to use after move

python/arcticdb/util/test.py
```diff
@@ -252,6 +252,41 @@ def assert_frame_equal_rebuild_index_first(expected: pd.DataFrame, actual: pd.Da
...
+    new_table = pa.Table.from_arrays(
+        new_columns,
+        schema=pa.schema(new_fields).with_metadata(metadata)
+    )
+    return new_table.to_pandas()
+
+def assert_frame_equal_with_arrow(left, right):
+    import pyarrow as pa
+    if isinstance(left, pa.Table):
+        left = convert_arrow_to_pandas_and_remove_categoricals(left)
```
vasil-pashov, line 284: nit: I think it makes sense to assert the right is pandas dataframe

cpp/arcticdb/arrow/array_from_block.hpp
IvoDD: Thanks, fixed this one and a few others (which also included copies).

python/arcticdb/util/test.py
IvoDD: The intention is to use this method across a lot of tests in the follow up PR.
So there will be value for it working for both:
`assert_equal(dataframe, table)` and `assert_equal(table, dataframe)`.
There is even some value in asserting two pyarrow tables are equal when converted to pandas (because the pyarrow.equals method doesn't compare the pandas_metadata)

PR #2489, opened 2025-07-23
Arrow read support (experimental release)

IvoDD:
Adds a frontend API for using Arrow. The API looks like below:
```
from arcticdb import Arctic, OutputFormat
ac = adb.Arctic(uri, output_format=OutputFormat.EXPERIMENTAL_ARROW) # Sets a runtime output format option so all read operations return arrow tables.
lib = ac["lib_name"]
lib.read(sym).data # This will return pyarrow.Table
lib.read(sym, output_format=OutputFormat.PANDAS) # We can also override the output format for any specific read operation
```

All read operations `read`, `read_batch`, `read_batch_and_join` and their `lazy` equivalents adhere to the `output_format` argument.

The changes in this PR are:
- Change the `_output_format` argument to `output_format`
- Separates the internal C++ `OutputFormat` from the python one. For python we use a `StrEnum` which allows users to pass both the enum and the string value
- Adds a `RuntimeOptions` class stored inside the `Arctic` and `NativeVersionStore` instances. `RuntimeOptions` contains only the `output_format` currently but will later on include things like `arrow_string_column_encoding` and other layout configurations.
- Allow passing `output_format` in all V2 APIs and make it work for `lazy=True` cases
- Run all query builder tests also with `output_format=ARROW` 
- Clean up `test_arrow` and `test_arrow_normalization` to adhere to new API
- Add tests for all user facing arrow APIs in `test_arrow_api.py` 
- Fixes an issue where `read_batch_and_join` didn't respect the input `ReadOptions`
- Introduces the `arcticdb.dependencies` for handling optional dependencies.

#### Any other comments?
Writing tests for the optional dependencies is difficult. So I ran a manual test:

In a venv with pyarrow:
```
>>> import arcticdb as adb
>>> ac = adb.Arctic("lmdb:///tmp/test-arrow")
>>> lib = ac["test"]
>>> lib.read("test", output_format="pandas").data
   x
0  5
1  6
2  7
>>> lib.read("test", output_format="experimental_arrow").data
pyarrow.Table
x: int64
----
x: [[5,6,7]]
```

And in a venv without pyarrow:
```
>>> import arcticdb as adb
>>> ac = adb.Arctic("lmdb:///tmp/test-arrow")
>>> lib = ac["test"]
>>> lib.read("test", output_format="pandas").data
   x
0  5
1  6
2  7
>>> lib.read("test", output_format="experimental_arrow").data
  File "/home/ivo/source/read_as_arrow/python/arcticdb/version_store/library.py", line 1887, in read
    return self._nvs.read(
           ^^^^^^^^^^^^^^^
  File "/home/ivo/source/read_as_arrow/python/arcticdb/version_store/_store.py", line 2063, in read
    version_query, read_options, read_query = self._get_queries(
                                              ^^^^^^^^^^^^^^^^^^
  File "/home/ivo/source/read_as_arrow/python/arcticdb/version_store/_store.py", line 1970, in _get_queries
    read_options = self._get_read_options(**kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ivo/source/read_as_arrow/python/arcticdb/version_store/_store.py", line 1958, in _get_read_options
    output_format_to_internal(
  File "/home/ivo/source/read_as_arrow/python/arcticdb/options.py", line 162, in output_format_to_internal
    raise ModuleNotFoundError("ArcticDB's pyarrow optional dependency missing but is required to use arrow output format.")
ModuleNotFoundError: ArcticDB's pyarrow optional dependency missing but is required to use arrow output format.
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1240,7 +1240,7 @@ MultiSymbolReadOutput LocalVersionedEngine::batch_read_and_join_internal(
     }
     auto clauses_ptr = std::make_shared<std::vector<std::shared_ptr<Clause>>>(std::move(clauses));
     return folly::collect(symbol_processing_result_futs).via(&async::io_executor())
-    .thenValueInline([this, &handler_data, clauses_ptr, component_manager, &read_options](std::vector<SymbolProcessingResult>&& symbol_processing_results) mutable {
+    .thenValueInline([this, &handler_data, clauses_ptr, component_manager, read_options](std::vector<SymbolProcessingResult>&& symbol_processing_results) mutable {
         auto [input_schemas, entity_ids, res_versioned_items, res_metadatas] = unpack_symbol_processing_results(std::move(symbol_processing_results));
         auto pipeline_context = setup_join_pipeline_context(std::move(input_schemas), *clauses_ptr);
         return schedule_remaining_iterations(std::move(entity_ids), clauses_ptr)
@@ -1248,10 +1248,10 @@ MultiSymbolReadOutput LocalVersionedEngine::batch_read_and_join_internal(
             auto proc = gather_entities<std::shared_ptr<SegmentInMemory>, std::shared_ptr<RowRange>, std::shared_ptr<ColRange>>(*component_manager, std::move(processed_entity_ids));
             return collect_segments(std::move(proc));
         })
-        .thenValueInline([store=store(), &handler_data, pipeline_context, &read_options](std::vector<SliceAndKey>&& slice_and_keys) mutable {
+        .thenValueInline([store=store(), &handler_data, pipeline_context, read_options](std::vector<SliceAndKey>&& slice_and_keys) mutable {
             return prepare_output_frame(std::move(slice_and_keys), pipeline_context, store, read_options, handler_data);
         })
-        .thenValueInline([&handler_data, pipeline_context, res_versioned_items, res_metadatas, &read_options](SegmentInMemory&& frame) mutable {
+        .thenValueInline([&handler_data, pipeline_context, res_versioned_items, res_metadatas, read_options](SegmentInMemory&& frame) mutable {
             // Needed to force our usual backfilling behaviour when columns have been outer-joined and some are not present in all input symbols
             ReadOptions read_options_with_dynamic_schema = read_options.clone();
             read_options_with_dynamic_schema.set_dynamic_schema(true);
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -500,7 +500,7 @@ class LazyDataFrameCollection(QueryBuilder):
         }
         check(
             len(output_format_set) in [0, 1],
-            f"LazyDataFrameCollection init requires all provided lazy dataframes to have the same output_format, but received: {[output_format for output_format in output_format_set]}",
+            f"LazyDataFrameCollection init requires all provided lazy dataframes to have the same output_format, but received: {output_format_set}",
             )
         super().__init__()
         self._lazy_dataframes = lazy_dataframes


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1240,22 +1240,22 @@ MultiSymbolReadOutput LocalVersionedEngine::batch_read_and_join_internal(
...
+            return prepare_output_frame(std::move(slice_and_keys), pipeline_context, store, read_options, handler_data);
         })
-        .thenValueInline([&handler_data, pipeline_context, res_versioned_items, res_metadatas](SegmentInMemory&& frame) mutable {
+        .thenValueInline([&handler_data, pipeline_context, res_versioned_items, res_metadatas, &read_options](SegmentInMemory&& frame) mutable {
             // Needed to force our usual backfilling behaviour when columns have been outer-joined and some are not present in all input symbols
-            ReadOptions read_options;
-            read_options.set_dynamic_schema(true);
-            return reduce_and_fix_columns(pipeline_context, frame, read_options, handler_data)
+            ReadOptions read_options_with_dynamic_schema = read_options;
+            read_options_with_dynamic_schema.set_dynamic_schema(true);
```
alexowens90, line 1257: This probably isn't doing what you want. `ReadOptions` uses the pimpl pattern with a shared pointer, so creating a copy and then modifying it also modifies the original

python/arcticdb/arctic.py
```diff
@@ -100,7 +107,8 @@ def __getitem__(self, name: str) -> Library:
             self._library_manager.get_library(lib_mgr_name, storage_override, native_storage_config=self._library_adapter.native_config()),
             repr(self._library_adapter),
             lib_cfg=self._library_manager.get_library_config(lib_mgr_name, storage_override),
-            native_cfg=self._library_adapter.native_config()
+            native_cfg=self._library_adapter.native_config(),
+            runtime_options=self._runtime_options
```
alexowens90: Should we make this overwritable in `get_library`?

python/arcticdb/version_store/_store.py
```diff
@@ -118,6 +122,14 @@ def resolve_defaults(param_name, proto_cfg, global_default, existing_value=None,
     if param_value is not None:
         return param_value
 
+    try:
+        if runtime_options is not None:
```
alexowens90, line 127: nit: Unneeded

python/arcticdb/version_store/library.py
```diff
@@ -1767,7 +1779,8 @@ def read(
         row_range: Optional[Tuple[int, int]] = None,
         columns: Optional[List[str]] = None,
         query_builder: Optional[QueryBuilder] = None,
-        lazy: bool = False,
+        output_format : Optional[OutputFormat] = None,
+        lazy: bool = False
```
alexowens90: Need to maintain the argument order for the (admittedly niche) case where someone is providing all of the arguments without `=`
e.g. `lib.read("sym", 0, (None, None), (None, None), [], None, False)`

python/arcticdb/version_store/library.py
```diff
@@ -1767,7 +1779,8 @@ def read(
         row_range: Optional[Tuple[int, int]] = None,
         columns: Optional[List[str]] = None,
         query_builder: Optional[QueryBuilder] = None,
-        lazy: bool = False,
+        output_format : Optional[OutputFormat] = None,
```
alexowens90: Missed `head` and `tail`

python/arcticdb/util/test.py
```diff
@@ -252,39 +252,33 @@ def assert_frame_equal_rebuild_index_first(expected: pd.DataFrame, actual: pd.Da
     assert_frame_equal(left=expected, right=actual)
 
 
-def convert_arrow_to_pandas_and_remove_categoricals(table):
+def stringify_dictionary_encoded_columns(table, string_type=None):
```
alexowens90: Until we support returning string types directly, I think we should make this function user-facing, so maybe move to a file called `arrow_utils.py` or something.

python/tests/conftest.py
```diff
@@ -1122,6 +1134,11 @@ def lmdb_version_store_tiny_segment(version_store_factory) -> NativeVersionStore
     return version_store_factory(column_group_size=2, segment_row_size=2, lmdb_config={"map_size": 2**30})
 
 
+@pytest.fixture
+def lmdb_version_store_tiny_segment_dynamic_strings(version_store_factory) -> NativeVersionStore:
+    return version_store_factory(column_group_size=2, segment_row_size=2, dynamic_strings=True, lmdb_config={"map_size": 2**30})
```
alexowens90, line 1139: nit: `column_group_size` doesn't do anything with dynamic schema

python/tests/unit/arcticdb/test_arrow_api.py
```diff
@@ -0,0 +1,95 @@
...
+import pandas as pd
+import pyarrow as pa
+import pytest
+
+from arcticdb import LazyDataFrame, DataError, concat
+from arcticdb.options import OutputFormat
+from arcticdb.util.test import assert_frame_equal_with_arrow, sample_dataframe
+
+
+all_output_format_args = [None, OutputFormat.PANDAS, "PANDAS", OutputFormat.EXPERIMENTAL_ARROW, "EXPERIMENTAL_ARROW"]
```
alexowens90: Does lowercase work as well?

python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -9,18 +9,31 @@
...
 import dateutil
+from typing import Union
 
 from arcticdb.version_store.processing import QueryBuilder
-from arcticdb.util.test import assert_frame_equal
+from arcticdb.util.test import assert_frame_equal, assert_frame_equal_with_arrow, stringify_dictionary_encoded_columns
+from arcticdb.options import OutputFormat
 import arcticdb.toolbox.query_stats as qs
 
 pytestmark = pytest.mark.pipeline
```
alexowens90, line 24: There are a bunch of other files, plus some individual tests (e.g. in `test_basic_version_store.py`) with this mark. We don't necessarily need to run all of those tests returning Arrow tables all the time, but would be good to convince ourselves once that they all pass

cpp/arcticdb/version/local_versioned_engine.cpp
IvoDD: Good catch. Added a `clone` method on `ReadOptions`

python/arcticdb/version_store/_store.py
IvoDD: I agree it should be unneeded for the most common case where we use `NativeVersionStore.resolve_runtime_defaults` because `self._runtime_defaults` is always set from the V2 API.

However the `NativeVersionStore` also has a static method `resolve_defaults` which can't know any runtime defaults as they are a member. I think it is bad design to have so many `resolve_defaults` variants but I'm keen to leave this refactor for another PR.

Also the V1 API inside man group will leave the `self._runtime_options=None`, so we need to be able to handle this case.

python/arcticdb/version_store/library.py
IvoDD: Good catch. 
We should have probably added a `*` right after `symbol` argument but it's too late for this now.

python/arcticdb/util/test.py
IvoDD: Moved to `arcticdb.util.arrow`

python/tests/conftest.py
IvoDD: Confusingly enough this is `dynamic_strings=True` not `dynamic_schema`. 
The original `lmdb_version_store_tiny_segment` fixture uses `dynamic_strings=False` which I didn't want to change as it will impact a lot of tests.

python/tests/unit/arcticdb/test_arrow_api.py
IvoDD: It didn't. Changed all output format types `Union[OutputFormat, str]` as this is required to support lowercase as well.

python/arcticdb/arctic.py
```diff
@@ -59,10 +59,18 @@ def __init__(self, uri: str, encoding_version: EncodingVersion = DEFAULT_ENCODIN
             URI specifying the backing store used to access, configure, and create Arctic libraries.
             For more details about the parameters, please refer to the [Arctic URI Documentation](./arctic_uri.md).
 
-        encoding_version: EncodingVersion, default DEFAULT_ENCODING_VERSION
+        encoding_version: EncodingVersion, default = DEFAULT_ENCODING_VERSION
```
alexowens90: Not your fault, but DEFAULT_ENCODING_VERSION doesn't get resolved, it just shows up in the docs:
https://docs.arcticdb.io/latest/api/arctic/#arcticdb.Arctic.__init__

python/arcticdb/arctic.py
```diff
@@ -59,10 +59,18 @@ def __init__(self, uri: str, encoding_version: EncodingVersion = DEFAULT_ENCODIN
...
             When creating new libraries with this Arctic instance, the default encoding version to use.
             Can be overridden by specifying the encoding version in the LibraryOptions argument to create_library.
 
+        output_format: Union[OutputFormat, str], default = OutputFormat.PANDAS
+            Controls the default output format of all operations returning a dataframe.
+            The default behavior (OutputFormat.PANDAS) is to return `pandas.DataFrame`s backed by numpy arrays.
```
alexowens90: Or `Series`

python/arcticdb/version_store/_store.py
alexowens90: Yeh fair enough

python/arcticdb/version_store/library.py
alexowens90: Yup, we will just suffer forever instead

setup.cfg
```diff
@@ -106,6 +106,9 @@ exclude=
     static
 
 [options.extras_require]
+arrow =
+    pyarrow
```
alexowens90, line 112: Conda?
If we add any pinning here then we'll need to add the same check in `dependencies.py` that `pyarrow.__version__` obeys our pins. I think for now we can just add a comment here to that effect. If we want to get fancy with it later we can [parse](https://setuptools.pypa.io/en/latest/setuptools.html#configuration-api) `setup.cfg` to avoid double definitions

python/tests/conftest.py
alexowens90: Yep, ignore me

setup.cfg
IvoDD: Conda doesn't really support optional dependencies. See [issue](https://github.com/conda/conda/issues/12098)

We already have pyarrow as a dependency for the tests, so there's nothing more to change for conda.

Also added a comment in the `setup.cfg`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1240,22 +1240,22 @@ MultiSymbolReadOutput LocalVersionedEngine::batch_read_and_join_internal(
     }
     auto clauses_ptr = std::make_shared<std::vector<std::shared_ptr<Clause>>>(std::move(clauses));
     return folly::collect(symbol_processing_result_futs).via(&async::io_executor())
-    .thenValueInline([this, &handler_data, clauses_ptr, component_manager](std::vector<SymbolProcessingResult>&& symbol_processing_results) mutable {
+    .thenValueInline([this, &handler_data, clauses_ptr, component_manager, &read_options](std::vector<SymbolProcessingResult>&& symbol_processing_results) mutable {
```
vasil-pashov: Capture `read_options` by value here and in other placed in this function. The reason is that any of the of the functions can throw and we're using `.get` on the future which just propagates the exception. But there might be other threads working on the data. The stack will get cleaned though and `read_options` will be destroyed and then the other treads will segfault. Read options holds only a shared_ptr so it's cheap to copy.

python/arcticdb/util/arrow.py
```diff
@@ -0,0 +1,18 @@
...
+
+    Useful for testing when comparing to the source dataframe where we want regular large_string columns instead of
+    categorical columns.
+    """
+    if string_type is None:
+        string_type = pa.large_string()
+    for i, name in enumerate(table.column_names):
+        if pa.types.is_dictionary(table.column(i).type):
+            table = table.set_column(i, name, table.column(name).cast(string_type))
+    return table
```
vasil-pashov, line 18: This equivalent to the removed code from `convert_arrow_to_pandas_and_remove_categoricals` right?

python/arcticdb/version_store/library.py
```diff
@@ -491,10 +494,19 @@ def __init__(
             len(lib_set) in [0, 1],
             f"LazyDataFrameCollection init requires all provided lazy dataframes to be referring to the same library, but received: {[lib for lib in lib_set]}",
         )
+        output_format_set = {
+            lazy_dataframe.read_request.output_format for lazy_dataframe in lazy_dataframes
+            if lazy_dataframe.read_request.output_format is not None
+        }
+        check(
+            len(output_format_set) in [0, 1],
+            f"LazyDataFrameCollection init requires all provided lazy dataframes to have the same output_format, but received: {[output_format for output_format in output_format_set]}",
```
vasil-pashov: Can't we use `output_format_set` instead of `[output_format for output_format in output_format_set]`

vasil-pashov (changes requested):

vasil-pashov (approved):

cpp/arcticdb/version/local_versioned_engine.cpp
IvoDD: Discussed offline: Should be fine because these functions are executed only once sequentially.

In any case reasoning about this is easier if we just copy and this is a shared_ptr so copy is cheap. Will do the copy.

python/arcticdb/util/arrow.py
IvoDD: Yes, we had ended up with two copies of the same function (one from Alex and one from me and Alex's one was better so it won)

PR #2490, opened 2025-07-23
Fix column selection for processing pipeline

vasil-pashov:
#### Reference Issues/PRs
Monday: 9492480789

The processing pipeline had a known bug that column selection does not work that was made more evident when modify_schema was added. More specifically synthetic columns were not added to the end result.

With this commit everything passing through the processing pipeline now removes filtered columns from the end stream descriptor. This also fixes bug that was introduced with an earlier commit and xfailed concat tests are now working.

#### Any other comments?
I could not evade using the pipeline_context. Things were too tangled to fix in a single PR. There is a ticket in Monday to refactor this portion of the code see: 9654333036

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -930,19 +930,36 @@ static StreamDescriptor generate_initial_output_schema_descriptor(const Pipeline
     return desc;
 }
 
-static OutputSchema create_initial_output_schema(const PipelineContext& pipeline_context) {
+static OutputSchema create_initial_output_schema(PipelineContext& pipeline_context) {
     internal::check<ErrorCode::E_ASSERTION_FAILURE>(pipeline_context.norm_meta_,
                                                     "Normalization metadata should not be missing during read_and_process");
     return OutputSchema{generate_initial_output_schema_descriptor(pipeline_context), *pipeline_context.norm_meta_};
 }
 
 static OutputSchema generate_output_schema(
-    const PipelineContext& pipeline_context,
-    const std::span<const std::shared_ptr<Clause>> clauses
+    PipelineContext& pipeline_context,
+    std::shared_ptr<ReadQuery> read_query
 ) {
     OutputSchema output_schema = create_initial_output_schema(pipeline_context);
-    for (const auto& clause: clauses) {
+    for (const auto& clause: read_query->clauses_) {
         output_schema = clause->modify_schema(std::move(output_schema));
+    }
+    if (read_query->columns) {
+        std::unordered_set<std::string_view> selected_columns(read_query->columns->begin(), read_query->columns->end());
+        FieldCollection fields_to_use;
+        if (!pipeline_context.filter_columns_) {
+            pipeline_context.filter_columns_ = std::make_shared<FieldCollection>();
+        }
+        for (const Field& field : output_schema.stream_descriptor().fields()) {
+            if (selected_columns.contains(field.name())) {
+                fields_to_use.add(field.ref());
+                if (!pipeline_context.is_in_filter_columns_set(field.name())) {
+                    pipeline_context.filter_columns_->add(field.ref());
+                }
+            }
+        }
+        pipeline_context.filter_columns_set_ = std::move(selected_columns);
+        output_schema.set_stream_descriptor(StreamDescriptor{output_schema.stream_descriptor().data_ptr(), std::make_shared<FieldCollection>(std::move(fields_to_use))});
     }
     return output_schema;
 }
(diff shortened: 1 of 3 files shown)


python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -1113,3 +1113,26 @@ def test_to_strings():
...
+def test_column_select_projected_column_and_filter_it(version_store_factory, dynamic_schema):
+    lib = version_store_factory(dynamic_schema=dynamic_schema)
+    sym = "sym_0"
+    lib.write(sym, pd.DataFrame({"a": [1, 2]}))
+    qb = QueryBuilder()
+    qb = qb.apply("new_column", qb["a"] + 2)
+    qb = qb[qb["new_column"] > 3]
+    result = lib.read(sym, columns=["new_column"], query_builder=qb).data
+    expected = pd.DataFrame({"new_column": [4]})
+    assert_frame_equal(expected, result)
```
alexowens90, line 1146: Can we have some tests with column slicing and using query stats? That way we'll prove that data key selection is working, and that we're not reading any data keys unnecessarily
Another test could be to create a projected column, filter on it, and select a different column that wasn't involved in any of the query steps

cpp/arcticdb/version/version_core.cpp
```diff
@@ -930,20 +930,37 @@ static StreamDescriptor generate_initial_output_schema_descriptor(const Pipeline
     return desc;
 }
 
-static OutputSchema create_initial_output_schema(const PipelineContext& pipeline_context) {
+static OutputSchema create_initial_output_schema(PipelineContext& pipeline_context) {
```
alexowens90, line 933: Why?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -930,20 +930,37 @@ static StreamDescriptor generate_initial_output_schema_descriptor(const Pipeline
...
 ) {
     OutputSchema output_schema = create_initial_output_schema(pipeline_context);
-    for (const auto& clause: clauses) {
+    for (const auto& clause: read_query->clauses_) {
         output_schema = clause->modify_schema(std::move(output_schema));
     }
+    if (read_query->columns) {
+        std::unordered_set<std::string_view> selected_columns(read_query->columns->begin(), read_query->columns->end());
+        FieldCollection fields_to_use;
+        if (!pipeline_context.filter_columns_) {
```
alexowens90, line 950: Are there any cases where `static_cast<bool>(pipeline_context.filter_columns_)` is `true` and the collection is non-empty?
If not, we should assert it is empty

vasil-pashov: Yes, when there are only synthetic columns. It's computed [here](https://github.com/man-group/ArcticDB/blob/7ff4b4a8c1e8b7d05201ef4da52d59c71fa6e63d/cpp/arcticdb/pipeline/read_pipeline.hpp#L122) when setup_pipeline_context is called. The function I linked includes only columns that are part of the TSD.

Unfortunately things are quite tangled at the moment. The pipeline context is used downstream in for example in `allocate_frame` and a few more places. I tried to reduce the used of `pipeline_context.filter_columns_` and `pipeline_context.filter_columns_set_` downstream but it's not trivial. Bunch of errors and segfault started appearing so I opened an additional task for that.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: See my other comment: https://github.com/man-group/ArcticDB/pull/2490/files#r2228294964

python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -1113,3 +1114,50 @@ def test_to_strings():
...
+@pytest.mark.parametrize("dynamic_schema", [True, False])
+def test_filter_synthetic_column_and_select_on_disk_column(s3_store_factory, dynamic_schema):
+    lib = s3_store_factory(dynamic_schema=dynamic_schema, column_group_size=2)
+    sym = "sym_0"
+    lib.write(sym, pd.DataFrame({"a": [1, 2], "b": [7, 8], "c": [5, 6]}))
+    qb = QueryBuilder()
+    qb = qb.apply("new_column", qb["a"] + 2)
+    qb = qb[qb["new_column"] > 3]
+    with qs.query_stats():
+        result = lib.read(sym, columns=["b"], query_builder=qb).data
```
alexowens90: Can we parametrize this test on selecting column `b` and `c`?

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_query_builder.py
vasil-pashov: Added

PR #2493, opened 2025-07-24
Add warning when writing empty dataframe

vasil-pashov:
#### Reference Issues/PRs
Monday: 9654696653

In version 6 writing empty dataframe can lead to unexpected behavior in the processing pipeline. Since there are strict type checks some workflows that worked in prior versions will now throw exceptions. The type of the columns on of empty dataframes depends on the version of Pandas that wrote the dataframe. In some cases it can be float64 while in other object. Since the processing pipeline now checks the types of the columns that take part in a computation even if the dataframe is wrong exceptions can be thrown when filtering or projection empty dataframe. For example if the column is of object dtype and the processing pipeline was expression like `qb = qb[qb["empty"] < 5]`.

This adds warnings in the python layer of the APIs that can write an empty dataframe. The warning is controlled by environment variable.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -26,6 +26,7 @@ from numpy import datetime64
 from pandas import Timestamp, to_datetime, Timedelta
 from typing import Any, Optional, Union, List, Sequence, Tuple, Dict, Set
 from contextlib import contextmanager
+import time
 
 from arcticc.pb2.descriptors_pb2 import IndexDescriptor, TypeDescriptor
 from arcticdb_ext.version_store import SortedValue, StageResult
@@ -669,6 +670,7 @@ class NativeVersionStore:
                 self.version_store.write_parallel(symbol, item, norm_meta, validate_index, False, None)
                 return None
             else:
+                _log_warning_on_writing_empty_dataframe(data, symbol)
                 vit = self.version_store.write_versioned_dataframe(
                     symbol, item, norm_meta, udm, prune_previous_version, sparsify_floats, validate_index
                 )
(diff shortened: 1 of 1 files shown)


python/arcticdb/version_store/_store.py
```diff
@@ -3292,3 +3295,11 @@ def resolve_dynamic_strings(kwargs):
         dynamic_strings = True
 
     return dynamic_strings
+
+def _log_warning_on_writing_empty_dataframe(dataframe, symbol):
+    if dataframe.empty and os.getenv("ARCTICDB_WARN_ON_WRITING_EMPTY_DATAFRAME", "1") == "1":
```
alexowens90: Is `dataframe.empty` equivalent to `len(dataframe) == 0`?

python/arcticdb/version_store/_store.py
```diff
@@ -3292,3 +3295,11 @@ def resolve_dynamic_strings(kwargs):
...
+        log.warning("Writing empty dataframe to ArcticDB for symbol \"{}\". The dtypes of empty columns depend on the"
+                    "Pandas version being used. This can lead to unexpected behavior in the processing pipeline. For"
+                    " example if the empty columns are of object dtype they cannot be part of numeric computations in"
+                    "the processing pipeline such as filtering (qb = qb[qb['empty_column'] < 5]) or projection"
+                    "(qb = qb.apply('new', qb['empty_column'] + 5)). Column types: {}.", symbol, list(dataframe.dtypes.items()))
```
alexowens90: Can we include the version of Pandas being used, and therefore what the default dtype for an empty column is? And specify that the `coerce_columns` argument can be used to overcome this?

python/arcticdb/version_store/_store.py
```diff
@@ -660,6 +660,7 @@ def write(
                 self.version_store.write_parallel(symbol, item, norm_meta, validate_index, False, None)
                 return None
             else:
+                _log_warning_on_writing_empty_dataframe(data, symbol)
```
alexowens90, line 673: If you call `append`, or `update` with `upsert=True`, can you get into the same position?

python/arcticdb/version_store/_store.py
vasil-pashov: Yeah, but we have some pretty obscure functionality allowing passing a string or a list to write and some of the tests use that and this crashes. Good catch! I'll use len(dataframe) == 0 as it's seems safer.

PR #2497, opened 2025-07-24
Storage lock increase wait time and add artificial slow writes

ognyanstoimenov:
In continuation of https://github.com/man-group/ArcticDB/pull/2359 which was reverted. This is the same but without the check for slow writes, as that was causing the lock to never be taken on inherently slow storages.

This is the diff with the reverted PR
https://github.com/man-group/ArcticDB/compare/symbol_list_slow_writes_tests_before_revert..symbol_list_slow_writes_tests_after_revert

Changelog is similar to #2359:
* Storage lock wait increased to 1000ms
* ~Locking gives up on slow writes~
* StorageFailureSimulator extended to simulate slow writes
* Stress tests (some are removed from #2359)
* Refactoring the storage lock share common code

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/util/configs_map.hpp
```diff
@@ -21,32 +21,32 @@ using namespace arcticdb::proto::config;
...
     }
 
 #define HANDLE_TYPE(LABEL, TYPE)     \
     void set_##LABEL(const std::string& label, TYPE val) { \
+        std::lock_guard<std::mutex> lock(mutex_); \
         map_of_##LABEL[boost::to_upper_copy<std::string>(label)] = val; \
     } \
 \
     TYPE get_##LABEL(const std::string& label, TYPE default_val) const { \
+        std::lock_guard<std::mutex> lock(mutex_); \
```
poodlewars: Why do we need this guard? We call these `get_` methods all the time, often in fairly performance sensitive blocks (which perhaps we shouldn't), so I'm a bit concerned about the performance impact

poodlewars (approved):

poodlewars (changes requested):
Perhaps this benchmark regression is real. Will be good to get to the bottom of it.

```
 | Change   | Before [91d18651]    | After [66842fe3]    |   Ratio | Benchmark (Parameter)                                                             |
|----------|----------------------|---------------------|---------|-----------------------------------------------------------------------------------|
| +        | 168±2ms              | 194±4ms             |    1.16 | basic_functions.BatchBasicFunctions.time_read_batch_with_date_ranges(25000, 500)  |
| +        | 188±2ms              | 218±5ms             |    1.16 | basic_functions.BatchBasicFunctions.time_read_batch_with_date_ranges(50000, 500)  |
| +        | 351±4ms              | 404±8ms             |    1.15 | basic_functions.BatchBasicFunctions.time_read_batch_with_date_ranges(25000, 1000) |
```

cpp/arcticdb/util/configs_map.hpp
IvoDD: I agree these locks shouldn't be needed. It could only be useful for tests which are setting and getting values from multiple threads. Such tests need better design.

Some general notes, not suggesting we do any of this:

The mutex lock without contention is [quite cheap](https://news.ycombinator.com/item?id=15158037#:~:text=C%2B%2B%20program%20(2006)-,An%20uncontended%20mutex%20is%20just%20as%20fast%20as%20a%20spinlock,with%20some%20context%20switch%20overhead). but in this use case there could be contention from multiple readers whichmight explain the failing ASV benchmark. There exist [single-writer multiple-reader](https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock) primitives which can resolve this. 

Still even just hashing the string every time and accessing the hashmap can be expensive but should again be on the order of tens of nanoseconds (especially if compiler is smart enough to hash the string compile time). 
I do not think we really use the configsmap in settings where 100nanoseconds matter.

python/tests/stress/arcticdb/version_store/test_stress_symbol_list_cache.py
```diff
@@ -0,0 +1,109 @@
...
+    set_config_int("SymbolList.MaxDelta", 1) # Trigger symbol list compaction on every list_symbols call
+    set_log_level(specific_log_levels = {"lock":"DEBUG"})
+    lib.list_symbols()
+
+@pytest.fixture(params=[
+    (0, 0, 0), # Probability of slowdown, min ms, max ms
+    (0.3, 10, 50),
+    (0.3, 100, 300),
+    (0.3, 300, 500),
+    (0.3, 700, 1200)
```
IvoDD: I think with the current setup the last fixture parameter will bring enough slowness to introduce a genine test failure.

I do think this test is useful for when we eventually do something about the symbol list, so we shouldn't remove it but probably `xfail` it.

Also could you (after xfail-ing) run the real storage test to see this test indeed fails and will prove itself useful for future symbol list development?

ognyanstoimenov:
The lock was indeed the problem for the asv degradation: [with lock](https://github.com/man-group/ArcticDB/actions/runs/16498009254/job/46648698640) vs [without lock](https://github.com/man-group/ArcticDB/actions/runs/16569596096/job/46874999625) 
so I've removed it 

My main concern is this relies on us being disciplined not to write anything to the configmap in multithreaded context and only read from it, while writing is done upfront

poodlewars (approved):

PR #2498, opened 2025-07-24
Azure tests added 

grusev:

Storage test execution against Azure blob storage for tests along with github workflow changes to support that in github with persistance test executions

Environment variables needed to execute azure storage tests locally:
```
export ARCTICDB_REAL_AZURE_CONTAINER=....
export ARCTICDB_REAL_AZURE_CONNECTION_STRING=....
export ARCTICDB_STORAGE_AZURE=1 
```

**Additionally**:
1.  there is support for ASV real storage test on Azure + log sanitization for AccountKey
NOTE: The ASV tests cannot work under this PR due to the change introducing common logging.py for all tests. This is run from a build with merged PR (in my fork) https://github.com/grusev/ArcticDB/actions/runs/16874994643/job/47797467959
2.  Installation tests now can run with Azure also

**Other framework enhancements**:
1.  Logging is now centralized in arcticdb\util\logger.py is the central logging utility. It contains sanitization code for secrets. Previously 2 loggers existed for ASV and functional tests. Now they are combined into one. Note that logger.py also exists in installation_tests. This is required due to the fact that this package although locally inside python folder is a package of its own and cannot use any of standard libs from arcticdb in order to not depend on any version, it is created to support ALL released versions.
2. As you will see in the logs below currently there is a MacOS related problem that affect significant number of tests. In order not to clutter code with many xfail statements there is a central code at conftest.py (again replicated twice for installation_tests also) that will XFAIL automatically all MacOs tests that experience that error. 

Imprtant note: this mechanism can be enhanced further to get rid potentially of all xfail statements and centralize xfailing in one file. That needs to be discussed first before accepted as it has both benefits and drawbacks. The case where we have massive failures like this one where benefits are much more than the drawbacks and therefore it perhaps should be the default option to go for in future.

**Workflow enhancements**: 
- 'Build with conda' - ability to debug, ability to enable debug logging, custom commands execution. Azure tests can now work by passing ARCTICDB params for the test, a kind of backdoor)
- "Installation tests" - same (Azure tests now can work fluently by selecting them)

Both enhancement make possible to run Azure tests with those workflows makes possible logging debug information when needed and connect to git executor on demand. With those enhancements we now get enhanced ability to run and debug issues on Linux, Windows, MacOS using either builds from the top of branch as well as arcticdb install packages from conda and pypi

**Build and Test**: https://github.com/man-group/ArcticDB/actions/runs/16614944930/job/47006507855
Windows is ok. Linux is having problems perhaps due to different image
Working tests on Linux and Windows: https://github.com/man-group/ArcticDB/actions/runs/16621045959/job/47026145204 (reduced to Python 3.12 tests only)
Full run : https://github.com/man-group/ArcticDB/actions/runs/16623323259/job/47033988304
One test failed due to error that needs to be investigated more closely, but the error is exatly same as the one with MacOS

**Installation Tests**: https://github.com/man-group/ArcticDB/actions/runs/16588682801/job/46919014211
The tests work on Azure with Linux. There is problem with MacOs. Some tests fail due to bug most probably (to be logged)

**Build with Conda** tets (running Azure tests): https://github.com/man-group/ArcticDB/actions/runs/16595281766/job/46940289302
The linux tests pass, while MacOS are full with errors of same type as previous:
```
___ ERROR at setup of test_batch_methods_with_negative_as_of[real_azure-0] ____
[gw2] darwin -- Python 3.13.5 /Users/runner/micromamba/envs/arcticdb/bin/python
arctic_client = Arctic(config=azure(endpoint=DefaultEndpointsProtocol=https;AccountName=arcticdbgithub;AccountKey=...;EndpointSuffix=core.windows.net, container=githubblob))
lib_name = 'test_batch_methods_with_negati.19784_8719781760_2025-07-29T12_11_50__56d7de65-510c-4134-9e1f-3eaf2ff19759'
    @pytest.fixture
    def arctic_library(arctic_client, lib_name) -> Generator[Library, None, None]:
>       yield arctic_client.create_library(lib_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/conftest.py:628: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
arcticdb/arctic.py:205: in create_library
    if self.has_library(name):
       ^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = Arctic(config=azure(endpoint=DefaultEndpointsProtocol=https;AccountName=arcticdbgithub;AccountKey=...;EndpointSuffix=core.windows.net, container=githubblob))
name = 'test_batch_methods_with_negati.19784_8719781760_2025-07-29T12_11_50__56d7de65-510c-4134-9e1f-3eaf2ff19759'
    def has_library(self, name: str) -> bool:
        """
        Query if the given library exists
    
        Parameters
        ----------
        name: str
            Name of the library to check the existence of.
    
        Returns
        -------
        True if the library exists, False otherwise.
        """
>       return self._library_manager.has_library(self._library_adapter.get_name_for_library_manager(name))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       arcticdb_ext.exceptions.InternalException: Azure::Storage::StorageException(404 The specified blob does not exist.
E       
E       Request ID: 5d961a76-e01e-005f-5e81-005af8000000)
arcticdb/arctic.py:254: InternalException
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,11 +70,10 @@ jobs:
     runs-on: ubuntu-latest
     env:
       INPUT_STORAGE: ${{ inputs.persistent_storage }}
-      CONN_STR: ${{ secrets.AZURE_CONNECTION_STRING }}
     outputs:
       storage: ${{ steps.set-storage.outputs.storage }}
-      DEBUG_LOGGING_ENABLED: ${{ steps.set-storage.outputs.DEBUG_LOGGING_ENABLED }}
-      DEBUG_ENABLED: ${{ steps.set-storage.outputs.DEBUG_ENABLED }}      
+      DEBUG_LOGGING_ENABLED: ${{ steps.parse-flags.outputs.DEBUG_LOGGING_ENABLED }}
+      DEBUG_ENABLED: ${{ steps.parse-flags.outputs.DEBUG_ENABLED }}      
     steps:
       - name: Determine Storage Type Based On Schedule
         id: set-storage
@@ -109,8 +108,6 @@ jobs:
 
           echo "Final storage type for workflow execution: $_storage"
           echo "storage=$_storage" >> $GITHUB_OUTPUT
-
-          connection_string=$CONN_STR
 
       - name: Parse flags
         id: parse-flags


.github/workflows/build_steps.yml
```diff
@@ -400,8 +400,19 @@ jobs:
           aws_access_key: "${{ secrets.AWS_S3_ACCESS_KEY }}"
           aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"
 
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.debug_enabled }}
```
G-D-Petrov: nit: the `inputs....` are only available in `workflow_dispatch` so you don't need this check - `github.event_name == 'workflow_dispatch'`

.github/workflows/build_with_conda.yml
```diff
@@ -107,11 +122,34 @@ jobs:
       - name: Check no arcticdb file depend on tests package
         shell: bash -l {0}
         run: |
-          build_tooling/checks.sh          
+          build_tooling/checks.sh       
+          
+      - name: Set persistent storage variables
```
G-D-Petrov, line 136: Do you have a run were this was tested and works?

.github/workflows/build_with_conda.yml
```diff
@@ -107,11 +122,34 @@ jobs:
...
+          aws_access_key: "${{ secrets.AWS_S3_ACCESS_KEY }}"
+          aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"
+          gcp_access_key: "${{ secrets.GCP_S3_ACCESS_KEY }}"
+          gcp_secret_key: "${{ secrets.GCP_S3_SECRET_KEY }}"
+          azure_container: "githubblob" # DEFAULT BUCKET FOR AZURE
+          azure_connection_string: "${{ secrets.AZURE_CONNECTION_STRING }}"
+          persistent_storage: "true"
+
+      - name: Set ArcticDB Debug Logging
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.run_enable_logging }}
```
G-D-Petrov: nit: the inputs.... are only available in workflow_dispatch so you don't need this check - github.event_name == 'workflow_dispatch'

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -74,28 +99,45 @@ def _safe_enter(self):
...
+                # Things to do against real Azure
+                uri = f"{self.factory.get_arctic_uri()};SharedAccessSignature={sas}"
+                self.client = ContainerClient.from_connection_string(uri, self.container, **self._get_policy())
+            else:
+                # Container is in a mock env - Azurite
+                self._set_uri_and_client_azurite("SharedAccessSignature=" + sas)
 
     def __exit__(self, exc_type, exc_value, traceback):
+        if self.is_real_azure():
+            if self.factory.clean_bucket_on_fixture_exit:
```
G-D-Petrov: why do we need the `is_real_azure` check here?
Wouldn't be better to just use `self.factory.clean_bucket_on_fixture_exit`

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -74,28 +99,45 @@ def _safe_enter(self):
...
+                self.factory.cleanup_container(self)
+                if len(self.libs_from_factory) > 0:
+                    get_logger().warning(f"Libraries not cleared remaining {self.libs_from_factory.keys()}")
+            
         if self.client:
```
G-D-Petrov, line 115: nit: can be rewritten as `if self.client and not self.is_real_azure():` and reduce the indentation/make it easier to read

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -399,6 +400,20 @@ def real_gcp_from_environment_variables(
         out.default_prefix = os.getenv("ARCTICDB_PERSISTENT_STORAGE_UNIQUE_PATH_PREFIX", "") + additional_suffix
     return out
 
+def real_azure_from_environment_variables(
```
G-D-Petrov: shouldn't this be in `python/arcticdb/storage_fixtures/azure.py`

python/installation_tests/client_utils.py
```diff
@@ -168,6 +174,30 @@ def get_real_gcp_uri(shared_path: bool = True):
     )
     return aws_uri
 
+def real_azure_credentials(shared_path: bool = True):
```
G-D-Petrov, line 181: can we import this from `python/arcticdb/storage_fixtures/...`

python/tests/conftest.py
```diff
@@ -662,7 +700,8 @@ def version_store_factory(lib_name, lmdb_storage) -> Generator[Callable[..., Nat
     # Otherwise there will be no storage space left for unit tests
     # very peculiar behavior for LMDB, not investigated yet
     # On MacOS ARM build this will sometimes hang test execution, so no clearing there either
-    yield from _store_factory(lib_name, lmdb_storage, not (WINDOWS or MACOS))     
+    check_local_storage_enabled()
```
G-D-Petrov: does this work? I think that we had a discussion in that past that this wouldn't work

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -179,6 +184,7 @@ def test_empty_snapshot_name_not_allowed(object_version_store):
 @pytest.mark.parametrize("unhandled_char", [chr(0), chr(30), chr(127), chr(128)])
 @pytest.mark.storage
 def test_unhandled_chars_update_upsert(object_version_store, unhandled_char):
+    xfail_azure_chars(object_version_store, unhandled_char)
```
G-D-Petrov, line 187: do we have an explanation on why this is failing/expected to fail?

.github/workflows/build_with_conda.yml
grusev: yes: https://github.com/man-group/ArcticDB/actions/runs/16595281766/job/46940289302

python/installation_tests/client_utils.py
grusev: no we cannot. Installation test must not import anything from normal project.
Generally if we have to structure the things well here we have to do something like the new logger.py file and symlink in installation_tests: https://github.com/man-group/ArcticDB/pull/2498/files

But that would be quite refactoring as I would need to do it for all storage types that read from  environment vars = GCP, S3, Azure ....

I will leave this work for another PR that is focused on that only

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
grusev: Yes we have where the function is defined marks.py see last line in this snippet:
```
def xfail_azure_chars(nvs, symbol_name):

    def contains_problem_chars(text: str) -> list:
        target_chars = [chr(c) for c in range(0, 32)] + [
            chr(126), chr(127), chr(140), chr(142), chr(143), chr(156)
        ]
        found = [(ord(char), repr(char)) for char in text if char in target_chars]
        return found

    storage_is_azure = False
    try:
        storage_is_azure = "azure" in nvs.get_backing_store()
    except:
        # for v2 API
        storage_is_azure = "azure" in nvs._nvs.get_backing_store()
    if (storage_is_azure) and (contains_problem_chars(symbol_name)):
        pytest.xfail("Char is not supported by azure and filtered by arcticdb (see 9669996138)")
```

python/tests/conftest.py
grusev: yes this works on all OS-es as per links in dexcription and my tests

python/arcticdb/storage_fixtures/s3.py
grusev: I can move it there

.github/workflows/build.yml
```diff
@@ -94,6 +109,22 @@ jobs:
...
+          connection_string=$CONN_STR
+
+          # Expected keys
+          keys=("DefaultEndpointsProtocol" "AccountName" "AccountKey" "EndpointSuffix")
+
+          # Check for presence of each key and extract value
+          for key in "${keys[@]}"; do
+              if [[ "$connection_string" =~ ${key}=([^;]+) ]]; then
+                  value="${BASH_REMATCH[1]}"
+                  echo "$key: $value"
```
phoebusm: Will that echo the secret?

.github/workflows/installation_tests.yml
```diff
@@ -227,24 +260,46 @@ jobs:
...
+          echo "Run custom pytest command: $command"
+          python -m pip  install pytest-repeat
+          echo "Run custom pytest command: $ARCTICDB_PYTEST_ARGS"
+          eval "$command"
+        else
+          echo $pwd
+          echo "Running tests against arcticdb version: ${{ env.arcticdb_version }}"
+          pytest -v --log-file="tests.log" -n auto python/installation_tests/test_installation.py
+          ## For testing
+          #pytest -v --log-file="tests.log" python/tests/integration/arcticdb/test_arctic.py::test_append_prune_previous_versions
```
phoebusm: I don't think we should keep testing lines in the PR

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -36,8 +42,17 @@ class AzureContainer(StorageFixture):
     client: Optional["ContainerClient"]
     _admin_client: Optional["ContainerClient"] = None
 
-    def _set_uri_and_client(self, auth: str):
-        from azure.storage.blob import ContainerClient, LinearRetry
+    def _get_policy(self) -> str:
```
phoebusm, line 47: Any reason why timeout is needed to be specified?

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -74,28 +99,45 @@ def _safe_enter(self):
...
+                    get_logger().warning(f"Libraries not cleared remaining {self.libs_from_factory.keys()}")
+            
         if self.client:
-            if self._admin_client:
-                self._admin_client.delete_container(timeout=3)
-                self._admin_client.close()
-            else:
-                self.client.delete_container(timeout=3)
+            if not self.is_real_azure():
+                if self._admin_client:
```
phoebusm: Why not always declaring `_admin_client` to reduce the complexity here?

python/arcticdb/storage_fixtures/azure.py
phoebusm: Agree with Georgi

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -166,3 +212,99 @@ def __exit__(self, exc_type, exc_value, traceback):
...
     def create_fixture(self) -> AzureContainer:
         return AzureContainer(self)
+
+    def cleanup_container(self, b: AzureContainer):
+        # Do nothing in this case
+        pass            
+
+def find_ca_certs():
+    # Common CA certificates locations
+    possible_paths = [
```
phoebusm: Can it be replaced with `ssl.get_default_verify_paths()`? Hardcoding default paths doesn't look good to me

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -166,3 +212,99 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    client_cert_file : str = None
+    protocol: str = None
+    clean_bucket_on_fixture_exit = True
+
+    def __init__(self, native_config: Optional[dict] = None):
+        self.native_config = native_config
+        if _LINUX:
+            ca_certs_file = find_ca_certs()
+            assert ca_certs_file, f"CA file: {ca_certs_file} not found!"
+            ca_certs_path = copy_ca_certs(ca_certs_file, "ca-certificates.crt")
```
phoebusm, line 268: Assuming this is the factory for real azure storage test, I wonder why ca cert path is needed to be specfically set?

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -166,3 +212,99 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        if prefix:
+            AzureStorageFixtureFactory.default_prefix = prefix
+        return self
+
+    def get_arctic_uri(self):
+        url = f"azure://Container={self.default_container};Path_prefix={self.default_prefix}"
+        if self.client_cert_file:
+           url += f";CA_cert_path={self.client_cert_file}"
+        if self.connection_string:
+            url += f"{url};{self.connection_string}"
```
phoebusm: `url` is duplicated

python/installation_tests/client_utils.py
G-D-Petrov: Let's leave it as is, my suggestion wouldn't actually work because the previous versions don't have this functionality

.github/workflows/build.yml
grusev: No it will not, but I will remove it anyway. I needed to troubleshoot that the string was correct. Git automatically filters out secrets so such things are not problematic, ok for troubleshooting but indeed must be out of code. Thanks!!!

python/arcticdb/storage_fixtures/azure.py
grusev: nope, actually I am keeping this line as it was, just refactored it slightly to make room for handling both Azure and Azurite (previously only Azurite was handled): 

```
# The retry_policy instance will be modified by the pipeline, so cannot be constant
        policy = {"connection_timeout": 1, "read_timeout": 2, "retry_policy": LinearRetry(retry_total=3, backoff=1), "connection_verify": f.client_cert_file}
```

python/arcticdb/storage_fixtures/azure.py
grusev: I did not knew about this merhod, so I cannot know if it will work - will try it. Thanks!

python/arcticdb/storage_fixtures/azure.py
grusev: Nope will not will append to url ... connection string was not part of URL and the connection string is not an URL
(will add this like comment so that it is not mistaken)
```
connection_string = 'DefaultEndpointsProtocol=https;AccountName=...;AccountKey=...;EndpointSuffix=core.windows.net'
```

```
url = f"azure://Container={self.default_container};Path_prefix={self.default_prefix}"
        if self.client_cert_file:
           url += f";CA_cert_path={self.client_cert_file}"
        if self.connection_string:
            url += f"{url};{self.connection_string}"
```

python/arcticdb/storage_fixtures/azure.py
grusev: nope the code does not set it this is local var only the code does this:

```
            ca_certs_path = copy_ca_certs(ca_certs_file, "ca-certificates.crt")
            AzureStorageFixtureFactory.client_cert_file = ca_certs_path
```

So actually only client_cert_file is set for the class

python/arcticdb/storage_fixtures/azure.py
grusev: by definition _admin_client is different than client. So the logic here is preserved for the Azurite and adapted to handle both Azurite and Azure. 
As I have not deep dived in Azurite I am not inclined to make changes.
I will add comments to code but rather not change it and trigger unnecessary side effects

The code used to be:
```
                if self._admin_client:
                    self._admin_client.delete_container(timeout=3)
                    self._admin_client.close()
                else:
                    self.client.delete_container(timeout=3)
```

now an if is added because we have Azure also not only Azurite. So this code assures all is working as it was for Azurite and new parts are not affecting that ...

Thus the change is clean and although might seem little oscure is doing correct thing

python/arcticdb/storage_fixtures/azure.py
grusev: not exacly ... 

```
self.client.close()
self.client = None
```

should be common for Azurite and Azure.

python/arcticdb/storage_fixtures/azure.py
grusev: The logic for cleaning azurite and Azure is different. In this PR I do not plan to refactor Azurite this keeping the things separate for now and adding comments. In future PR we can refactor this code if we know what we want to achieve, but this is not that PR (see other comments below for more details why this is left the way it is)

```
    def __exit__(self, exc_type, exc_value, traceback):
        if self.is_real_azure():
            # This code is for cleaning Azure, the code for Azurite is down
            if self.factory.clean_bucket_on_fixture_exit:
                self.factory.cleanup_container(self)
                if len(self.libs_from_factory) > 0:
                    get_logger().warning(f"Libraries not cleared remaining {self.libs_from_factory.keys()}")
            
        if self.client:
            if not self.is_real_azure():
                # This code is only for Azurite cleaning, it is faster than this for Azure
                if self._admin_client:
                    self._admin_client.delete_container(timeout=3)
                    self._admin_client.close()
                else:
                    self.client.delete_container(timeout=3)
            # This code is for both Azure and Azurite
            self.client.close()
            self.client = None
```

python/arcticdb/storage_fixtures/azure.py
phoebusm: I am not sure this is what you want:
```
In [1]: url = "abc"

In [2]: connection_string = "123"

In [3]: url += f"{url};{connection_string}"

In [4]: url
Out[4]: 'abcabc;123'
```

.github/workflows/build_with_conda.yml
```diff
@@ -241,15 +239,15 @@ jobs:
           gcp_secret_key: "${{ secrets.GCP_S3_SECRET_KEY }}"
           azure_container: "githubblob" # DEFAULT BUCKET FOR AZURE
           azure_connection_string: "${{ secrets.AZURE_CONNECTION_STRING }}"
-          persistent_storage: "true"
+          persistent_storage: "no"
```
phoebusm: It's a bit confusing that `set_persistent_storage_env_vars` is run but `persistent_storage` is set `no`

python/arcticdb/storage_fixtures/azure.py
G-D-Petrov: agreed, I didn't see that

python/arcticdb/storage_fixtures/azure.py
phoebusm: There are existing softlink for ca certs in CI: https://github.com/man-group/arcticdb-enterprise/blob/8625b07c7d6d3c9838329cfaf4fb91b3b0ee9633/.github/workflows/build.yml#L199
So would you please share what issue requires this new function (`find_ca_certs`)

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
phoebusm: Can we not feeding the problematic char if the storage is azure, instead of test against it but having extra logic to xfail it? The logic brings complexity to the tests

python/installation_tests/client_utils.py
```diff
@@ -178,6 +208,13 @@ def create_arctic_client(storage: StorageTypes, **extras) -> Arctic:
                 __ARCTIC_CLIENT_GPC = Arctic(get_real_gcp_uri(shared_path=False), **extras)
             return __ARCTIC_CLIENT_GPC
 
+    if CONDITION_AZURE_AVAILABLE:
+        if storage == StorageTypes.REAL_AZURE and is_storage_enabled(storage):
+            global __ARCTIC_CLIENT_AZURE
+            if __ARCTIC_CLIENT_AZURE is None:
+                __ARCTIC_CLIENT_AZURE = Arctic(get_real_azure_uri(shared_path=False), **extras)
```
phoebusm: Why the client has to be global but not pytest session fixture?

python/installation_tests/client_utils.py
```diff
@@ -10,22 +10,32 @@
...
+LINUX = sys.platform.lower().startswith("linux")
+logger = get_logger()
+
 
 CONDITION_GCP_AVAILABLE = (
     True if "dev" in arcticdb.__version__ else version.Version(arcticdb.__version__) >= version.Version("5.3.0")
 )
 
+CONDITION_AZURE_AVAILABLE = (
+    True if "dev" in arcticdb.__version__ else version.Version(arcticdb.__version__) >= version.Version("5.0.0")
```
phoebusm: Why >= 5.0.0? Azure should start to be available from v4.x at least

python/installation_tests/client_utils.py
```diff
@@ -10,22 +10,32 @@
...
 import tempfile
 
 import arcticdb
 from arcticdb.arctic import Arctic
 from datetime import datetime
 from packaging import version
 
+from logger import get_logger
+
+LINUX = sys.platform.lower().startswith("linux")
```
phoebusm: Why not import it from mark.py?

python/installation_tests/client_utils.py
```diff
@@ -168,6 +174,30 @@ def get_real_gcp_uri(shared_path: bool = True):
...
+
+def get_real_azure_uri(shared_path: bool = True):
+    (
+        container,
+        connection_str,
+        path_prefix,
+    ) = real_azure_credentials(shared_path)
+    ca_certs = ""
+    if LINUX:
+        ca_certs_file = "/etc/ssl/certs/ca-certificates.crt"
```
phoebusm, line 199: Why sometimes it's hardcoded but sometimes get it from a given list?
Again, why not getting from `ssl` lib?

python/installation_tests/conftest.py
```diff
@@ -61,9 +71,20 @@ def ac_library(request, ac_client, lib_name) -> Generator[Library, None, None]:
...
     if ac is None: pytest.skip()
+    logger.info(f"Create library : {lib_name}")
     ac.create_library(lib_name, **config)
     lib = ac.get_library(lib_name)
     yield lib
     ac.delete_library(lib_name)    
 
+#region Pytest special xfail handling
+
+def pytest_runtest_makereport(item, call):
```
phoebusm, line 91: I think I have seen the same functions more than once in the PR. Can we have one only?

python/tests/pytest_xfail.py
```diff
@@ -0,0 +1,46 @@
...
+"""
+
+import sys
+import traceback
+
+import pytest
+
+
+MACOS = sys.platform == "darwin"
+XFAILMESSAGE= "This is due issue 9692682845 - has_library may return error on Mac_OS"
```
phoebusm, line 16: Have this been raised to azure sdk repo? 
It'll be great if we can have [azure log](https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/README.md) to support whether it's really azure sdk issue

python/tests/util/storage_test.py
```diff
@@ -1,4 +1,5 @@
 from enum import Enum
+import random
```
phoebusm: I don't think it's in use

python/installation_tests/client_utils.py
```diff
@@ -168,6 +174,30 @@ def get_real_gcp_uri(shared_path: bool = True):
...
+        os.path.exists(ca_certs_file), f"CA file: {ca_certs_file} not found!"
+        ca_certs = f";CA_cert_path={ca_certs_file}"
+    azure_uri = f"azure://Container={container};Path_prefix={path_prefix};{connection_str}{ca_certs}"
```
phoebusm, line 202: I have found below function in your PR (`python/tests/util/storage_test.py`) and they look duplicated:
```
def get_real_azure_uri(shared_path: bool = True):
    return real_azure_from_environment_variables(
        shared_path=shared_path,
        ).get_arctic_uri()
```

python/tests/util/storage_test.py
```diff
@@ -132,10 +134,16 @@ def get_real_gcp_uri(shared_path: bool = True):
     )
     return aws_uri
 
+def get_real_azure_uri(shared_path: bool = True):
+    return real_azure_from_environment_variables(
```
phoebusm, line 137: I don't think the clean up has been done properly in this way, as the factory class is exited before the uri is returned by this function

python/tests/util/storage_test.py
```diff
@@ -145,8 +153,11 @@ def persistent_test_type() -> PersistentTestType:
     will raise error
     """
     if PERSISTENT_STORAGE_TESTS_ENABLED:
-        if os.getenv("ARCTICDB_PERSISTENT_STORAGE_TESTS_TYPE", "").lower() == "gcp":
+        storage_type = os.getenv("ARCTICDB_PERSISTENT_STORAGE_TESTS_TYPE", "").lower()
```
phoebusm, line 155: Maybe consolidate these into `mark.py`?

python/tests/hypothesis/arcticdb/test_resample.py
```diff
@@ -12,7 +12,9 @@
     compute_common_type_for_columns
 )
 from arcticdb.util._versions import IS_PANDAS_TWO
-from arcticdb.util.utils import get_logger
+from hypothesis.extra.pandas import column
```
phoebusm: Is this in use?

python/tests/conftest.py
```diff
@@ -149,18 +152,24 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize("encoding_version", [EncodingVersion.V1] if only_v1 else list(EncodingVersion))
 
 
+def check_local_storage_enabled():
```
phoebusm, line 155: Such way of skipping test brings extra complexity. Can we skip the test in the way similar in `def arctic_client`?

python/benchmarks/real_read_write.py
```diff
@@ -11,7 +11,8 @@
 import pandas as pd
 
 from arcticdb.options import LibraryOptions
-from arcticdb.util.environment_setup import DataFrameGenerator, TestLibraryManager, LibraryPopulationPolicy, LibraryType, Storage, get_console_logger, populate_library_if_missing
+from arcticdb.util.environment_setup import DataFrameGenerator, TestLibraryManager, LibraryPopulationPolicy, LibraryType, Storage, populate_library_if_missing
+from arcticdb.util.logger import get_logger
```
phoebusm: Can it be imported as `_get_logger` to avoid confusion below?

python/benchmarks/real_query_builder.py
```diff
@@ -8,7 +8,8 @@
 
 from logging import Logger
 import pandas as pd
-from arcticdb.util.environment_setup import DataFrameGenerator, TestLibraryManager, LibraryPopulationPolicy, LibraryType, Storage, get_console_logger, populate_library_if_missing
+from arcticdb.util.environment_setup import DataFrameGenerator, TestLibraryManager, LibraryPopulationPolicy, LibraryType, Storage, populate_library_if_missing
+from arcticdb.util.logger import get_logger
```
phoebusm: Same here

python/arcticdb/version_store/helper.py
```diff
@@ -430,8 +430,8 @@ def get_azure_proto(
             azure.prefix = f"{lib_name}{time.time() * 1e9:.0f}"
     else:
         azure.prefix = lib_name
-    azure.ca_cert_path = ca_cert_path
-    azure.ca_cert_dir = ca_cert_dir
+    azure.ca_cert_path = ca_cert_path if ca_cert_path else ""
```
phoebusm, line 433: I think we should fix the caller which passes `None` as the typehint hasn't be respected.
And the original implementation works

python/arcticdb/storage_fixtures/azure.py
phoebusm: This is a bit confusing. Would you please move the factory specific logic into `AzureStorageFixtureFactory` and `AzuriteStorageFixtureFactory` to keep the design more OO?

python/arcticdb/version_store/helper.py
grusev: the hint is [ca_cert_path: str = ""] this hint does not exclude None, so the caller is respecting the hint. 

The problem is the following call so indeed the caller here is the function that I fixed.
storage.config.Pack(azure, type_url_prefix="cxx.arctic.org")

Thus the fix is ok

python/tests/conftest.py
grusev: no, this is normal and supported way to skipping moreover this is the right way to skip fixture that does not have parameters. Only fixtures that have 2 or more parameters are handled that way.

python/tests/hypothesis/arcticdb/test_resample.py
grusev: good catch - removed

python/tests/util/storage_test.py
grusev: Not in this PR. There are other like this in this file and not only. 
There is a greater need of consolidation that will happen outside of this PR. The target will be to consolidate all env variable reads and most of factories that use them in one file without other dependencies of the project - will that be mark.py or other is yet to be seen but we have 3 distinct projects that require reading:
 - pytest unit and intgeration
 - installation tests - they are separate from others for a good reason
 - asv tests
 
ASV tests demand that all such things are outside of 'tests' package because they can work only with arcticdb build and installed module. Thus such file is better be located at arcticdb/util folder

Another task is created

python/tests/pytest_xfail.py
grusev: no I have not - that should be done when the referenced issue is decided to be tackled. No need to be done earlier. With this PR the goal is to find what works and what does not, What does not to fix if need adaptation to work, what not to raise issues, they will be prioritized later. For now although we have issues that prevent supporting Azure na MacOS the prio is low and perhaps will take time until we get to that moment to make it work properly

python/installation_tests/conftest.py
grusev: yes you are seeing it because it needs to be in conftest.py file. We have separate conftest.py files for standard tests and installation tests. If you look closer you will see that actual implementation of the function is done only once. In a separate file and symlinked to installa project. But those functions which are actually pytests hooks need to be created individually in each set of tests

python/installation_tests/client_utils.py
grusev: Very good catch ... but we cannot that easily. This works as installation tests use ubuntu. Other tests does not use ubuntu.

Installation tests cannot load anything for now from the normal project. They have to be separate. Loading anything from the common project is very dangerous at this point. 

What we can do on installation tests is to symlink a file from main project that has no other dependencies on it. That approach is already used for 2 files one of them is logger.py

This PR was going to become enormous if I had to fix all past sins we made. That is why there has to be an end. And the end is here.

This works for installation tests. It is not going to be used in near future in runs. This PR makes is possible to run tests in GH and assure all is fine.

I have commented bellow that another PR is needed for Env Vars and factories and such issue is already created. That issue will tackle this problem

python/installation_tests/client_utils.py
grusev: Will make it from 4x. No problem. For now GCP and Azure tests are not yet executed in installation tests

python/installation_tests/client_utils.py
grusev: ArcticDB approach to architecturing tests is not optimal.  Fixtures cannot be used in normal functions etc in normal code. Hence they are useless there. But we have times when we do need creating clients outside of tests. What should we do then? Create functions for that and have fixtures and functions? 

Answer is no.

So we have to start with functions. Have them and then create fixtures we need

As installation tests are separate from other tests (they cannot and should not import anything from there as this will trigger cascading import of all things, which will not work) here this concept is utilized.

Once we get to refactor the main project, this concept should be considered as start there also.

.github/workflows/build_with_conda.yml
grusev: persistent_storage values are "no", "GCP, "AZURE", ... it is enum not bool, it used to be bool once upon a time when only AWS S3 was the option, not anymore. But yml files have no checks so this was skipped during initial refactoring

python/arcticdb/storage_fixtures/azure.py
grusev: sorry have not realized that I am using += and f"{url};...." and those are one and the same things - my bad, one of those copy and paste sins I make. Interestingly enough this works so I easily dismissed your note without looking deeper what you might mean

python/arcticdb/storage_fixtures/azure.py
grusev: tests did not work when running in Github because ca certs could not be found on their usual location (see installation tests which run on normal ubntu against our tests which work inside a container for Build and Test) so I needed to create a function that will go through known places where it could be. Did not knew about this : ssl.get_default_verify_paths() now that I know I find it great - now we are sure that we have 99% chance to work on evry Linux. So the code for Azure does not rely anymore on having this cacert at specific location.

The correct link to container is this https://github.com/man-group/ArcticDB/blob/0b6f945f7c4650938014729e1ffa6cf4bb3fe56b/.github/workflows/build.yml#L253

python/arcticdb/storage_fixtures/azure.py
grusev: done
```
    def __exit__(self, exc_type, exc_value, traceback):
        if self.factory.clean_bucket_on_fixture_exit:
            self.factory.cleanup_container(self)
        if self.client:
            self.client.close()
            self.client = None
```

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
grusev: Let's leave this for 9669996138 resolution

For now this is difference between other storages and Azure. This is clear omissin to raise proper exception and this is the only way to xfail in this situation

python/installation_tests/client_utils.py
grusev: mark.py does not import any other libs I can symlink it

python/installation_tests/client_utils.py
grusev: hose functions need to be duplicated, installation tests cannot import anything from normal tests

python/tests/util/storage_test.py
grusev: This file is not a test file - it is a procedure file invoked from command line with different commands - populate, verify, clean. Thus we do not need fixtures just the URL where those need to run

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -288,8 +281,8 @@ def __str__(self):
         return f"[{type(self)}=Container:{self.default_container}], ConnectionString:{self.connection_string}"
 
     def initialize_from_connection_sting(self, constr: str, container: str, prefix: str = None) -> "AzureStorageFixtureFactory":
-        assert constr, "Azure connection string not available"
-        assert container, "Azure container not available"
+        get_logger().error(f"Azure connection string not available: {constr}")
+        get_logger().error(f"Azure container not available: {container}")
```
phoebusm: I wonder why error is logged here?

python/tests/pytest_xfail.py
phoebusm: Yea you have a point

python/arcticdb/storage_fixtures/azure.py
phoebusm: Yes I understand Azurite is a mess which may need those special retry policy to work. I am just suprised to see Azure also requires such. 
I wonder the same policy is needed for the azure C++ sdk.

.github/workflows/build_with_conda.yml
phoebusm: Sorry I still don't get it. If `persistent_storage` is hardcoded `no`, does `set_persistent_storage_env_vars` need to be run?

python/installation_tests/client_utils.py
phoebusm: https://github.com/man-group/ArcticDB/blob/bf7a719c8e44c1472d0a4a2d3bb1e582511dd146/python/arcticdb/adapters/azure_library_adapter.py#L66

Apology in advance as I may not fully understand `installation_tests`. In the above line, the ca certs should be automatically set. I really wonder it needs to be set again here.

python/installation_tests/client_utils.py
phoebusm: Can we have something like `XX_common.py` to consolidate the functions that can be shared?

phoebusm: May I ask why it can't be shared actually?

python/installation_tests/client_utils.py
phoebusm: If this has to be a global object, it feels wrong to me as `extras` can be different in everything call but the object being returned is always the first one created with the first `extras` being passed
Sorry for spotting this at this PR, not rather earlier.

python/arcticdb/storage_fixtures/azure.py
grusev: Because the connection string that is created is invalid. Initially I wanted to assert immediately but that breaks things in factories where Azure is created along with S3 and GCP. They will always return some connection string even if invalid. Thus I had to keep this returning wrong string so that code is is different for azure. Therefore the only way to notify the person who is trying to run the code is logging an error.

python/installation_tests/client_utils.py
grusev: installation tests can be executed against versions from 1.6 till now. Out common code if tigthly coupled with current version in development.

So the only possible way to have this work both ways is to separate things logically AND physically. Otherwise if you start importing something from tests folder usually you end up importing all things.

So installations tests cannot work with anything from tests folder. Instead they have to rely only on what is installed by arcticdb library from pypi or conda that is currently being tested

So when you consider anything from installaltion tests you should ignore all you knowledge and expectations from standard tests

python/installation_tests/client_utils.py
grusev: See my comment below.

.github/workflows/build_with_conda.yml
grusev: yes, with Github this is the section where all persistant storage things are tied up from secrets and user selections to actual environment variable for our tests

python/installation_tests/conftest.py
```diff
@@ -45,12 +45,16 @@ def ac_client(request) -> Generator[Arctic, None, None]:
...
-    ac = create_arctic_client(storage, **extras)            
+    ac:Arctic = create_arctic_client(storage, **extras)            
     arctic_uri = ac.get_uri() if ac else "Arctic is None (not created)"
     logger.info(f"Arctic uri : {arctic_uri}")
     if ac is None:
         pytest.skip("Storage not activated")
     yield ac
+    libs = ac.list_libraries()
+    for lname in libs:
+        logger.error(f"Library '{lname}' not deleted after test."
```
phoebusm, line 56: Shall we throw exception here? Error logged can be missed easily.

python/installation_tests/client_utils.py
```diff
@@ -203,26 +205,41 @@ def get_real_azure_uri(shared_path: bool = True):
 
 def create_arctic_client(storage: StorageTypes, **extras) -> Arctic:
 
+    sorted_extras = dict(sorted(extras.items()))
+
+    def create_arctic(dct: Dict[str, Arctic], uri: str, extras) -> Arctic:
+        key = f"{uri}{extras}"
+        if key not in dct:
+            dct[key] = Arctic(uri, **extras)
```
phoebusm, line 240: Per discussed, this could be a issue potentially as it keeps the client alive. Since only two clients are needed per storage, we won't hit it though.
Still it would be nice to add comment to it to explain the situation.

python/installation_tests/conftest.py
grusev: The error will be triggered during development time only. This is development error and not an installation test bug.
https://github.com/man-group/ArcticDB/pull/2498/commits/1d6f850f7d3108e6fe7920f9eb7a902fc75570f8

python/installation_tests/client_utils.py
grusev: Yes, as we discussed this design decision is different for a reason. That is explained now in the comment: https://github.com/man-group/ArcticDB/pull/2498/commits/1d6f850f7d3108e6fe7920f9eb7a902fc75570f8

.github/workflows/build_with_conda.yml
```diff
@@ -107,11 +122,34 @@ jobs:
...
+        if: ${{ env.real_tests_storage_type != 'no' }}
+        uses: ./.github/actions/set_persistent_storage_env_vars
+        with:
+          aws_access_key: "${{ secrets.AWS_S3_ACCESS_KEY }}"
+          aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"
+          gcp_access_key: "${{ secrets.GCP_S3_ACCESS_KEY }}"
+          gcp_secret_key: "${{ secrets.GCP_S3_SECRET_KEY }}"
+          azure_container: "githubblob" # DEFAULT BUCKET FOR AZURE
+          azure_connection_string: "${{ secrets.AZURE_CONNECTION_STRING }}"
+          persistent_storage: "true"
```
G-D-Petrov: As we discussed, this shouldn't be true but rather a specific option from the drop down

.github/workflows/build_with_conda.yml
G-D-Petrov: As we discussed, this shouldn't be hardcoded but we should have a drop down to specify it

G-D-Petrov (approved):

phoebusm (approved):

PR #2502, opened 2025-07-24
Upgrade folly

JohanMabille:
Upgrade to folly 2025.04.14 (available on both vcpkg and conda-forge), that should allow to build with clang 19.

### DIFF
--- a/cpp/third_party/vcpkg_overlays/folly/0008-win-dont-propagate-cpp-version.patch
+++ b/cpp/third_party/vcpkg_overlays/folly/0008-win-dont-propagate-cpp-version.patch
@@ -1,0 +1,18 @@
+diff --git a/CMake/FollyCompilerMSVC.cmake b/CMake/FollyCompilerMSVC.cmake
+index a7933d9e1..034a2b206 100644
+--- a/CMake/FollyCompilerMSVC.cmake
++++ b/CMake/FollyCompilerMSVC.cmake
+@@ -118,11 +118,11 @@ function(apply_folly_compile_options_to_target THETARGET)
+       /Zc:threadSafeInit # Enable thread-safe function-local statics initialization.
+       /Zc:throwingNew # Assume operator new throws on failure.
+ 
+-      /permissive- # Be mean, don't allow bad non-standard stuff (C++/CLI, __declspec, etc. are all left intact).
+-      /std:${MSVC_LANGUAGE_VERSION} # Build in the requested version of C++
+       /utf-8 # fmt needs unicode support, which requires compiling with /utf-8
+ 
+     PRIVATE
++      /permissive- # Be mean, don't allow bad non-standard stuff (C++/CLI, __declspec, etc. are all left intact).
++      /std:${MSVC_LANGUAGE_VERSION} # Build in the requested version of C++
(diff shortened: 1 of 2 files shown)


environment-dev.yml
```diff
@@ -63,8 +63,9 @@ dependencies:
   # Build dependencies for tests
   # Use libarrow==18.0.0 only on Linux and MacOS (resolution problem on Windows)
   # TODO: Remove this when the issue is resolved
-  - sel(unix): libarrow==18.0.0
-  - sel(win): libarrow
+  #- sel(unix): libarrow==18.0.0
+  #- sel(win): libarrow
```
jjerphan, line 67: This comment now has to be removed. It can be removed now or in another PR IMO.

environment-dev.yml
```diff
@@ -46,7 +46,7 @@ dependencies:
   - azure-identity-cpp
   - azure-storage-blobs-cpp
   - fmt
-  - folly==2023.09.25.00
+  - folly==2025.04.14.00
```
jjerphan, line 49: For readers' information, [`conda-forge/folly-feedstock` now has a `2025.04.14.00` branch](https://github.com/conda-forge/folly-feedstock/commits/2025.04.14.00/) dedicated to this version which is in sync with conda-forge.

cpp/third_party/vcpkg_overlays/folly/0004-Do-not-use-PRId64.patch
jjerphan, line 1: This has required and likely will be required for the Windows build of ArcticDB.

We could keep it.

cpp/third_party/vcpkg_overlays/folly/0003-Set-FOLLY_HAVE_SO_TIMESTAMPING-to-0.patch
jjerphan, line 1: This has required and likely will be required for the Windows build of ArcticDB.

We could keep it.

environment-dev.yml
JohanMabille: It should be removed in a dedicated PR, it is orthogonal to the upgrade of folly (even if upgrading folly is partially motivated by this pin).

cpp/third_party/vcpkg_overlays/folly/0004-Do-not-use-PRId64.patch
JohanMabille: I prefer to keep only what we need for sure. We can reintroduce it if the Windows build fails.

cpp/third_party/vcpkg_overlays/folly/0003-Set-FOLLY_HAVE_SO_TIMESTAMPING-to-0.patch
JohanMabille: Same remark here

environment-dev.yml
JohanMabille: Ah no I misread it, I thought it was the pinning of Clang, sorry for the confusion. But yeah, let's remove it in another PR, this one should be merged ASAP.

alexowens90 (approved):

PR #2505, opened 2025-07-24
Enhancement 9360445961: Support for reading dynamic schema with changing column types and missing columns with Arrow

alexowens90:
#### Reference Issues/PRs
[9360445961](https://man312219.monday.com/boards/7852509418/pulses/9360445961)

Support reading into Arrow format symbols with dynamic schema where column types are changing (this was already working, just added a test) and where columns are missing from some row slices.

### DIFF
--- a/cpp/arcticdb/arrow/array_from_block.hpp
+++ b/cpp/arcticdb/arrow/array_from_block.hpp
@@ -13,11 +13,10 @@
 
 namespace arcticdb {
 
-inline std::optional<sparrow::validity_bitmap> create_validity_bitmap(size_t offset, const Column& column) {
+inline std::optional<sparrow::validity_bitmap> create_validity_bitmap(size_t offset, const Column& column, size_t bitmap_size) {
     if(column.has_extra_buffer(offset, ExtraBufferType::BITMAP)) {
         auto &bitmap_buffer = column.get_extra_buffer(offset, ExtraBufferType::BITMAP);
-        const auto bitmap_size = bitmap_buffer.block(0)->bytes() / sizeof(uint32_t);
-        return sparrow::validity_bitmap{reinterpret_cast<uint8_t *>(bitmap_buffer.block(0)->release()), bitmap_size };
+        return sparrow::validity_bitmap{reinterpret_cast<uint8_t *>(bitmap_buffer.block(0)->release()), bitmap_size};
     } else {
         return std::nullopt;
     }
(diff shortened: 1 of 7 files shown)


cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -72,6 +71,20 @@ sparrow::dictionary_encoded_array<T> create_dict_array(
     }
 }
 
+// TODO: This is super hacky. Our string column return type is dictionary encoded, and this should be
+//  consistent when there are zero rows or the validity bitmap is all zeros. But sparrow (or possibly Arrow) requires at
+//  least one key-value pair in the dictionary, even if there are zero rows
+std::pair<sparrow::u8_buffer<int64_t>, sparrow::u8_buffer<char>> minimal_strings_dict() {
```
IvoDD: We always use these two buffers to construct a `sparrow::big_string_array` afterwards. What do you think about making the return type a `sparrow::big_string_array`?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -778,6 +792,22 @@ class NullValueReducer {
         return context_row.slice_and_key().slice_.row_range.first;
     }
 
+    void backfill_all_zero_validity_bitmaps(size_t offset_bytes_start, size_t offset_bytes_end) {
+        // Explanation: offset_bytes_start and offset_bytes_end should both be elements of block_offsets_ by
+        // construction. We must add an all zeros validity bitmap for each row-slice read from storage where this
+        // column was missing, in order to correctly populate the Arrow record-batches for the output
```
IvoDD, line 796: We can make the API make sure only elements of `block_offsets_` are selected by accepting indices to the vector. I.e.
```
void backfill_all_zero_validity_bitmaps(size_t block_start_index, size_t blocks_end_index) {
```

This will also save us the two expensive calls to `lower_bound` below.

IvoDD: Reading more carefully it looks like we can easily save only one of the `lower_bound` calls because we only know `block_end_index` upfront.

I think we can also get the `block_start_index` by using `PipelineContextRow::index_` from the previous iteration. I.e. We could keep a state `NullReducer::previous_block_index_ = row.index_`. That new state could also replace `pos_`.
Anyway I'm not super confident that the `PipelineContextRow::index_` corresponds to the block index always. Happy to skip this `block_start_index` logic if it complicates the `NullReducer` too much.

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -356,4 +339,122 @@ def test_arrow_layout(lmdb_version_store_tiny_segment):
...
+    # TODO: Remove to_pandas() when we support writing Arrow structures directly
+    lib.write(sym, write_table.to_pandas())
+    lib.append(sym, append_table.to_pandas())
+    expected = pa.concat_tables([write_table, append_table], promote_options="permissive")
+    received = lib.read(sym, _output_format=OutputFormat.ARROW).data
+    assert expected.equals(received)
+
+
+@pytest.mark.parametrize("rows_per_row_slice", [1, 7, 8, 9, 100_000])
+def test_arrow_dynamic_schema_missing_columns_numeric(lmdb_version_store_dynamic_schema_v1, rows_per_row_slice):
```
IvoDD: I think it would also be good to parametrize on `segment_row_count`. This way we can test that setting consecutive block bitmaps works as expected.

cpp/arcticdb/pipeline/read_frame.cpp
alexowens90: Yeh it's all a bit messy, and I'm not super confident there won't be bugs (what happens when `frame.offset() != 0`? When does this even happen?)
I doubt this will be a performance bottleneck in practice, it is `O(log(n))` in the number of data segments, almost everything else we do is `O(n)`

alexowens90: I've removed the easy one anyway

cpp/arcticdb/arrow/array_from_block.hpp
```diff
@@ -72,6 +71,20 @@ sparrow::dictionary_encoded_array<T> create_dict_array(
     }
 }
 
+// TODO: This is super hacky. Our string column return type is dictionary encoded, and this should be
```
poodlewars, line 74: This is a mess - have we raised with Sparrow maintainers?

PR #2507, opened 2025-07-25
Make append_batch and update_batch noop with empty dataframes when there's an existing version

vasil-pashov:
#### Reference Issues/PRs
Monday: 9663246666

Update and append are noop when there is an existing version and the input is an empty dataframe. They do not create new version, instead of that they return the latest live version's VersionItem.

`update_batch` and `append_batch` do not behave this way, they create new version keeping the same data. This is a bug and they should behave the same way as regular update and append.
#### Any other comments?
`append_batch` and `update_batch` should not create symbol list entries if the symbols exist. There's a new ticket for that Monday: 9682041273
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/integration/arcticdb/test_arctic_batch.py
+++ b/python/tests/integration/arcticdb/test_arctic_batch.py
@@ -717,6 +717,19 @@ def test_append_batch_missing_keys(arctic_library):
     assert read_dataframe.metadata == "great_metadata_s2"
     assert_frame_equal(read_dataframe.data, pd.concat([df2_write, df2_append]))
 
+def test_append_batch_empty_dataframe_does_not_increase_version(lmdb_version_store_v1):
+    lib = lmdb_version_store_v1
+    lib.batch_write(["sym1", "sym2"], [pd.DataFrame({"a": [1, 2, 3]}), pd.DataFrame({"b": [1, 2, 3, 4]})])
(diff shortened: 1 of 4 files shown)


python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -717,6 +718,34 @@ def test_append_batch_missing_keys(arctic_library):
...
+    assert append_result[0].version == 1
+    assert append_result[1].version == 0
+
+    sym_1_vit, sym_2_vit = lib.read("sym1"), lib.read("sym2")
+
+    assert sym_1_vit.version == 1
+    assert_frame_equal(sym_1_vit.data, pd.DataFrame({"a": [1, 2, 3, 5, 6, 7]}))
+    assert(len(lib_tool.find_keys_for_symbol(KeyType.VERSION, "sym1")) == 2)
+    assert(len(lib_tool.find_keys_for_symbol(KeyType.TABLE_INDEX, "sym1")) == 2)
+    assert(len(lib_tool.find_keys_for_symbol(KeyType.TABLE_DATA, "sym1")) == 2)
```
IvoDD, line 743: Maybe also assert the `KeyType.SYMBOL_LIST` remain unchanged. I think for regular updates and appends we don't write a new symbol list entry if there was a previous version.

IvoDD: Similar assertion for update would be useful.

PR #2515, opened 2025-07-28
Utility for writing segment_in_memory

deyan37:

This pull request implements a feature in the local versioned engine for writing a SegmentInMemory similarly to how an InputTensorFrame would be written using write_versioned_dataframe_internaly (this is what is used in the standard lib.write()). 

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/pipeline/slicing.cpp
```diff
@@ -29,6 +29,18 @@ SlicingPolicy get_slicing_policy(
     return FixedSlicer{options.column_group_size, options.segment_row_size};
 }
 
+SlicingPolicy get_slicing_policy(
+        const WriteOptions& options,
+        const SegmentInMemory& segment) {
+    if(options.bucketize_dynamic) {
+        const auto col_count = segment.num_columns();
```
IvoDD: This calculation is wrong. `num_columns` includes the index.
In any case we don't handle `HashedSlicer` correctly later on. Let's just raise that this is not yet implemented for Segments.

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -116,5 +120,28 @@ inline auto get_partial_key_gen(std::shared_ptr<InputTensorFrame> frame, TypedSt
     };
 }
 
+inline auto get_partial_key_gen(const SegmentInMemory& segment, TypedStreamVersion key) {
+    using PartialKey = stream::StreamSink::PartialKey;
+
+    return [&segment, key = std::move(key)](const SegmentInMemory& s) {
+        if (segment.descriptor().index().field_count() != 0) {
+            util::check(static_cast<bool>(segment.descriptor().index().type() == IndexDescriptor::Type::TIMESTAMP), "Got null index tensor in get_partial_key_gen");
+            auto& idx = segment.column(0);
```
IvoDD: The `segment.column(0)` is wrong. We should use the slice index column: `s.column(0)`.

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -116,5 +120,28 @@ inline auto get_partial_key_gen(std::shared_ptr<InputTensorFrame> frame, TypedSt
     };
 }
 
+inline auto get_partial_key_gen(const SegmentInMemory& segment, TypedStreamVersion key) {
```
IvoDD: As discussed maybe let's not use the source `SegmentInMemory`. We can only pass the `TypedStreamVersion`.

At this point this a quite different function with a different return type. Let's name it differently e.g. `get_partial_key_gen_for_segment_slices`

cpp/arcticdb/pipeline/slicing.hpp
IvoDD: This highlights we're not testing the case with an index column. Let's also write a C++ test which uses `write_sliced_segment` with a segment which has an index. 

I appreciate that the python segments we construct won't have indices but we want to make this function generally useful.

python/tests/integration/toolbox/test_library_tool.py
```diff
@@ -361,3 +361,23 @@ def read_type(key, column):
...
+    lib_tool.write_segment_in_memory(sym, segment)
+    dataframe = lib.read(sym).data
+
+    assert_frame_equal(dataframe, sample_df)
+
+    data_key_count = len(lib_tool.find_keys(KeyType.TABLE_DATA))
+    index_key_count = len(lib_tool.find_keys(KeyType.TABLE_INDEX))
+    version_key_count = len(lib_tool.find_keys(KeyType.VERSION))
+
+    assert data_key_count == len(sample_df) // 2
```
IvoDD: Let's also assert the data key ranges are as expected

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -182,21 +183,83 @@ folly::Future<std::vector<SliceAndKey>> slice_and_write(
...
-        const std::shared_ptr<InputTensorFrame>& frame,
-        const SlicingPolicy &slicing,
+        const std::shared_ptr<stream::StreamSink>& sink,
+        const std::shared_ptr<DeDupMap>& de_dup_map
+) {
+    ARCTICDB_SUBSAMPLE_DEFAULT(SliceSegment)
+    auto slices = arcticdb::util::variant_match(
+            slicing,
+            [&segment](FixedSlicer slicing) { return segment.split(slicing.row_per_slice()); },
+            [&segment](HashedSlicer slicing) { return segment.split(slicing.row_per_slice()); },
```
IvoDD: Our code won't work for a `HashedSlicer`. Let's instead raise an unimplemented error.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,35 +764,6 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
     return res.at(0);
 }
 
-VersionedItem LocalVersionedEngine::write_individual_segment(
-    const StreamId& stream_id,
-    SegmentInMemory&& segment,
```
IvoDD, line 762: This function takes the `SegmentInMemory&&` as an rvalue reference. Our new function only takes an lvalue reference, so it is still useful to keep this one as it allows to write the segment without copying.

IvoDD: Thinking more about it, I think it might be better to just update this function to use the slicing instead of making the `write_versioned_dataframe_internal` to take an variant type.

As it turned out the variant with SegmentInMemory has a lot of caveats (slicing by columns is not fully supported etc.) So better to just keep the API separate. Once we have fully featured column slicing and hash bucketizing for segments we can unify the APIs again. Let's discuss offline

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -182,21 +183,83 @@ folly::Future<std::vector<SliceAndKey>> slice_and_write(
...
+                tsd.set_stream_descriptor(segment.descriptor());
+                tsd.set_total_rows(segment.row_count());
+
+                // Create some basic normalization metadata that we need to be able to read the segment as a pandas
+                // dataframe later
+                arcticdb::proto::descriptors::NormalizationMetadata norm_meta;
+                norm_meta.mutable_df()->mutable_common()->mutable_index()->set_is_physically_stored(false);
+                norm_meta.mutable_df()->mutable_common()->mutable_index()->set_start(0);
+                norm_meta.mutable_df()->mutable_common()->mutable_index()->set_step(1);
+                tsd.set_normalization_metadata(std::move(norm_meta));
```
IvoDD: This normalization metadata construction would not allow using a timeseries index.

I think it's fine to leave it like this if we do this in the `write_individual_segment` method. As per other comment I think this will be the way to go.

We could also make the normalization metadata as an optional argument to `write_individual_segment` and populate it like so if missing.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -182,21 +183,83 @@ folly::Future<std::vector<SliceAndKey>> slice_and_write(
...
+            [&segment](FixedSlicer slicing) { return segment.split(slicing.row_per_slice()); },
+            [&segment](HashedSlicer slicing) { return segment.split(slicing.row_per_slice()); },
+            [segment]([[maybe_unused]] NoSlicing slicing) { return std::vector<SegmentInMemory>{segment}; }
+    );
+    if (slices.empty())
+        return folly::makeFuture(std::vector<SliceAndKey>{});
+
+    TypedStreamVersion tsv{std::move(key.id), key.version_id, KeyType::TABLE_DATA};
+    int64_t write_window = write_window_size();
+    return folly::collect(folly::window(std::move(slices), [key, sink, de_dup_map, &segment, tsv=std::move(tsv)](auto&& slice) {
```
IvoDD: I think we can avoud passing around the segment to the lambda and only rely on the slices.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -182,21 +183,83 @@ folly::Future<std::vector<SliceAndKey>> slice_and_write(
...
+                               auto frame_slice = FrameSlice{std::make_shared<entity::StreamDescriptor>(slice.descriptor()), {arcticdb::pipelines::get_index_field_count(slice), slice.descriptor().field_count()}, {slice.offset(), slice.offset() + slice.row_count()}};
```
IvoDD: This line is a bit hard to read, what do you think about splitting it to several variables like:
```
auto descriptor = ...
auto column_slice = ...
auto row_slice = ...
aitp frame_slice = FrameSlice(std::move(descriptor), std::move(row_slice), std::move(row_slice))
```

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -182,21 +183,83 @@ folly::Future<std::vector<SliceAndKey>> slice_and_write(
...
+                               auto ks = std::tuple<stream::StreamSink::PartialKey, SegmentInMemory, FrameSlice>{
+                                       get_partial_key_gen(segment, tsv)(slice), slice, frame_slice
```
IvoDD: This will copy construct the slice. What do you think about moving it?
Also we need to be careful to not have something like `something(x, std::move(x))` because argument calculation ordering is not guaranteed. To resolve this we can do the call to `get_partial_key_gen` above to a variable.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -182,21 +183,83 @@ folly::Future<std::vector<SliceAndKey>> slice_and_write(
...
+            [segment]([[maybe_unused]] NoSlicing slicing) { return std::vector<SegmentInMemory>{segment}; }
+    );
+    if (slices.empty())
+        return folly::makeFuture(std::vector<SliceAndKey>{});
+
+    TypedStreamVersion tsv{std::move(key.id), key.version_id, KeyType::TABLE_DATA};
+    int64_t write_window = write_window_size();
+    return folly::collect(folly::window(std::move(slices), [key, sink, de_dup_map, &segment, tsv=std::move(tsv)](auto&& slice) {
+                               auto frame_slice = FrameSlice{std::make_shared<entity::StreamDescriptor>(slice.descriptor()), {arcticdb::pipelines::get_index_field_count(slice), slice.descriptor().field_count()}, {slice.offset(), slice.offset() + slice.row_count()}};
+                               auto ks = std::tuple<stream::StreamSink::PartialKey, SegmentInMemory, FrameSlice>{
```
IvoDD: You can use `std::make_tuple`. See more info [here](https://stackoverflow.com/questions/34180636/what-is-the-reason-for-stdmake-tuple)

cpp/arcticdb/pipeline/slicing.hpp
IvoDD: Also this function is not really used like a generator. It would be better to just have:
```
PartialKey get_partial_key_for_segment_slice(const TypedStreamVersion& key, const SegmentInMemory& slice)
```

cpp/arcticdb/pipeline/write_frame.hpp
```diff
@@ -67,13 +75,13 @@ folly::Future<std::vector<SliceAndKey>> write_slices(
         bool sparsify_floats);
 
 folly::Future<entity::AtomKey> write_frame(
```
IvoDD: Relating to other comments, what do you think about leaving `write_frame` and all functions above it in the call stack unchanged?

and instead reviving `write_individual_segment` as `write_sliced_segment` which does what we want? Let's discuss offline.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -182,21 +183,83 @@ folly::Future<std::vector<SliceAndKey>> slice_and_write(
...
+        const SegmentInMemory& segment,
+        const SlicingPolicy& slicing,
         IndexPartialKey&& key,
-        const std::shared_ptr<InputTensorFrame>& frame,
-        const SlicingPolicy &slicing,
+        const std::shared_ptr<stream::StreamSink>& sink,
+        const std::shared_ptr<DeDupMap>& de_dup_map
+) {
+    ARCTICDB_SUBSAMPLE_DEFAULT(SliceSegment)
+    auto slices = arcticdb::util::variant_match(
```
IvoDD: Let's also leave a `TODO` to do the segment splitting in parallel. Currently we'll likely be CPU bound on the memory copies required by `split`, because we do them sequentially.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -72,46 +72,63 @@ static void modify_descriptor(const std::shared_ptr<pipelines::PipelineContext>&
...
+        const WriteOptions& options,
+        const std::shared_ptr<DeDupMap>& de_dup_map,
+        bool sparsify_floats,
+        bool validate_index,
+        bool no_slice
+) {
     ARCTICDB_SUBSAMPLE_DEFAULT(WaitForWriteCompletion)
-    ARCTICDB_DEBUG(log::version(), "write_dataframe_impl stream_id: {} , version_id: {}, {} rows", frame->desc.id(), version_id, frame->num_rows);
-    auto atom_key_fut = async_write_dataframe_impl(store, version_id, frame, options, de_dup_map, sparsify_floats, validate_index);
+    //ARCTICDB_DEBUG(log::version(), "write_dataframe_impl stream_id: {} , version_id: {}, {} rows", frame->desc.id(), version_id, frame->num_rows);
```
IvoDD: We don't want to disable the debug logging.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -72,46 +72,63 @@ static void modify_descriptor(const std::shared_ptr<pipelines::PipelineContext>&
...
-    bool validate_index
-    ) {
+        const std::shared_ptr<Store>& store,
+        VersionId version_id,
+        const std::variant<std::shared_ptr<pipelines::InputTensorFrame>, SegmentInMemory>& frame,
+        const WriteOptions& options,
+        const std::shared_ptr<DeDupMap>& de_dup_map,
+        bool sparsify_floats,
+        bool validate_index,
+        bool no_slice
```
IvoDD: As mentioned in other comments, let's probably leave the write_dataframe_imp unchanged.
I.e. we can add the `no_slice` argument to `write_individual_segment`.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -624,6 +624,17 @@ VersionedItem PythonVersionStore::write_versioned_dataframe(
...
 }
 
+VersionedItem PythonVersionStore::write_versioned_segment(
+        const StreamId& stream_id,
+        const SegmentInMemory& segment,
+        bool prune_previous_versions,
+        bool sparsify_floats,
+        bool validate_index) {
+    ARCTICDB_SAMPLE(WriteVersionedDataframe, 0)
+    auto versioned_item = write_versioned_dataframe_internal(stream_id, segment, prune_previous_versions, sparsify_floats, validate_index);
```
IvoDD: We can change this to use the new `write_sliced_segment` instead.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -22,6 +22,7 @@
 #include <vector>
 #include <array>
 #include <ranges>
+#include <arcticdb/pipeline/read_frame.hpp>
```
IvoDD: Nit: Remove unneeded import

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -116,5 +116,30 @@ inline auto get_partial_key_gen(std::shared_ptr<InputTensorFrame> frame, TypedSt
     };
 }
 
+inline stream::StreamSink::PartialKey get_partial_key(const IndexDescriptorImpl& index, TypedStreamVersion key, const SegmentInMemory& slice) {
```
IvoDD: We can also pass the `TypedStreamVersion` by `const&`.

The reason it might be useful to copy it is if it is called in a `folly::Future` and could loose the reference to the original object. However in this case we are guaranteeing the reference exists in the lamba wrapping the call to `get_partial_key`.

It will be worth checking the equivalent `get_partial_key_gen` whether it does some unneeded copies as well. If so, let's remove unneeded copies.

cpp/arcticdb/pipeline/write_frame.hpp
```diff
@@ -67,13 +67,12 @@ folly::Future<std::vector<SliceAndKey>> write_slices(
...
-    const std::shared_ptr<Store> &store,
-    const std::shared_ptr<DeDupMap>& de_dup_map = std::make_shared<DeDupMap>(),
-    bool allow_sparse = false
-);
+        IndexPartialKey&& key,
+        const std::shared_ptr<InputTensorFrame>& frame,
+        const SlicingPolicy &slicing,
+        const std::shared_ptr<Store>& store,
+        const std::shared_ptr<DeDupMap>& de_dup_map,
+        bool sparsify_floats = false);
```
IvoDD: Nit: Let's return the whitespace change to how it was before.
If we end up fixing the formatting let's do it in another PR and fix all of the wildly inconsistent formattings we have used in this file. This way it will be easier for reviewers to differentiate function vs non-functional changes

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -733,13 +733,13 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
-        de_dup_map,
-        allow_sparse,
-        validate_index);
+            store(),
+            version_id,
+            frame,
+            write_options,
+            de_dup_map,
+            allow_sparse,
+            validate_index);
```
IvoDD: Similar to previous comment let's separate formatting changes to untouched code.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
-}
-
-VersionedItem LocalVersionedEngine::write_individual_segment(
-    const StreamId& stream_id,
-    SegmentInMemory&& segment,
-    bool prune_previous_versions
-    ) {
+VersionedItem LocalVersionedEngine::write_segment(
+        const StreamId& stream_id,
+        const SegmentInMemory& segment,
```
IvoDD: Let's get it by rvalue reference. This way if using `Slicing::NoSlicing` we can write it without copying.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
+    }
+    auto partial_key = IndexPartialKey(stream_id, version_id);
 
-    auto frame_slice = FrameSlice{segment};
-    auto descriptor = make_timeseries_descriptor(segment.row_count(),segment.descriptor().clone(), {}, std::nullopt,std::nullopt, std::nullopt, false);
-    auto key = store_->write(pk, std::move(segment)).get();
-    std::vector sk{SliceAndKey{frame_slice, to_atom(key)}};
-    auto index_key_fut = index::write_index(index, std::move(descriptor), std::move(sk), IndexPartialKey{stream_id, version_id}, store_);
-    auto versioned_item = VersionedItem{std::move(index_key_fut).get()};
+    auto slices = (slicing == Slicing::NoSlicing) ? std::vector<SegmentInMemory>({segment}) : segment.split(get_write_options().segment_row_size);
```
IvoDD: I think a `switch` would be more readable and is better for maintenance if e.g. adding new variants to the enum.

IvoDD: Let's also move the segment in case it is a single slice.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
+    auto tsd = TimeseriesDescriptor();
+    tsd.set_stream_descriptor(segment.descriptor());
+    tsd.set_total_rows(segment.row_count());
+    arcticdb::proto::descriptors::NormalizationMetadata norm_meta;
+    norm_meta.mutable_df()->mutable_common()->mutable_index()->set_is_physically_stored(false);
+    norm_meta.mutable_df()->mutable_common()->mutable_index()->set_start(0);
+    norm_meta.mutable_df()->mutable_common()->mutable_index()->set_step(1);
+    tsd.set_normalization_metadata(std::move(norm_meta));
+
+    auto atom_key_fut = std::move(fut_slice_keys).thenValue([partial_key = std::move(partial_key), &sink, tsd, index = std::move(index)](auto&& slice_keys) {
```
IvoDD: We should be able to safely move the `tsd` to avoid a copy.

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -116,5 +116,30 @@ inline auto get_partial_key_gen(std::shared_ptr<InputTensorFrame> frame, TypedSt
...
+    using PartialKey = stream::StreamSink::PartialKey;
+
+    if (index.field_count() != 0) {
+        util::check(static_cast<bool>(index.type() == IndexDescriptor::Type::TIMESTAMP), "Got null index tensor in get_partial_key_gen");
```
IvoDD: Let's update the error message. E.g. `Got unexpected index type in get_partial_key`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
-    auto range = get_range_from_segment(index, segment);
 
-    stream::StreamSink::PartialKey pk{
-        KeyType::TABLE_DATA, version_id, stream_id, range.start_, range.end_
-    };
+    auto write_options = get_write_options();
+    auto de_dup_map = get_de_dup_map(stream_id, maybe_prev, write_options);
+
+    if(version_id == 0){
+        auto check_outcome = verify_symbol_key(stream_id);
```
IvoDD, line 777: Why do we need this check?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
-    auto index_key_fut = index::write_index(index, std::move(descriptor), std::move(sk), IndexPartialKey{stream_id, version_id}, store_);
-    auto versioned_item = VersionedItem{std::move(index_key_fut).get()};
+    auto slices = (slicing == Slicing::NoSlicing) ? std::vector<SegmentInMemory>({segment}) : segment.split(get_write_options().segment_row_size);
+
+    auto sink = store();
+
+    TypedStreamVersion tsv{partial_key.id, partial_key.version_id, KeyType::TABLE_DATA};
+    int64_t write_window = write_window_size();
+    auto fut_slice_keys = folly::collect(folly::window(std::move(slices), [&sink, de_dup_map, &segment, tsv=std::move(tsv)](auto&& slice) {
+            auto frame_slice = FrameSlice{std::make_shared<entity::StreamDescriptor>(slice.descriptor()), {arcticdb::pipelines::get_index_field_count(slice), slice.descriptor().field_count()}, {slice.offset(), slice.offset() + slice.row_count()}};
```
IvoDD: Let's either split this long line as into several variables. E.g.:
```
auto descriptor = ...
auto column_slice = ...
auto row_slice = ...
auto frame_slice = FrameSlice{std::move(descriptor), std::move(column_slice), std::move(row_slice)}
```
Or at least make it more readable by splitting into several lines. E.g.:
```
auto frame_slice = FrameSlice(
    std::make_shared<entity::StreamDescriptor>(slice.descriptor()), 
    {arcticdb::pipelines::get_index_field_count(slice), slice.descriptor().field_count()}, 
    {slice.offset(), slice.offset() + slice.row_count()}
);
```

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
+            auto pkey = get_partial_key(segment.descriptor().index(), tsv, slice);
+            auto ks = std::tuple<stream::StreamSink::PartialKey, SegmentInMemory, FrameSlice>{
+                    pkey, std::move(slice), frame_slice
```
IvoDD: We can move both `pkey` and `frame_slice` as they were just constructed above.
Also let's use `std::make_tuple`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
+    // Create a TimeSeriesDescriptor that is needed for writing the index key
+    auto index = stream::index_type_from_descriptor(segment.descriptor());
+    auto tsd = TimeseriesDescriptor();
+    tsd.set_stream_descriptor(segment.descriptor());
+    tsd.set_total_rows(segment.row_count());
+    arcticdb::proto::descriptors::NormalizationMetadata norm_meta;
+    norm_meta.mutable_df()->mutable_common()->mutable_index()->set_is_physically_stored(false);
+    norm_meta.mutable_df()->mutable_common()->mutable_index()->set_start(0);
+    norm_meta.mutable_df()->mutable_common()->mutable_index()->set_step(1);
+    tsd.set_normalization_metadata(std::move(norm_meta));
```
IvoDD: Maybe create this `TimeseriesDescriptor` with the rowrange metadata only if it is not already written in the `SegmentInMemory`.

cpp/arcticdb/version/test/test_sparse.cpp
```diff
@@ -419,7 +419,7 @@ TEST_F(SparseTestStore, SegmentWithExistingIndex) {
     bool written = false;
     DynamicAggregator aggregator(std::move(schema), [&](SegmentInMemory &&segment) {
         if(!written) {
-            test_store_->write_individual_segment(stream_id, std::move(segment), false);
+            test_store_->write_versioned_dataframe_internal(stream_id, std::move(segment), false, false, false);
```
IvoDD: Looks like we have forgotten to update the C++ tests.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -74,30 +74,31 @@ static void modify_descriptor(const std::shared_ptr<pipelines::PipelineContext>&
...
-    bool sparsify_floats,
-    bool validate_index
-    ) {
+        const std::shared_ptr<Store>& store,
+        VersionId version_id,
+        const std::shared_ptr<pipelines::InputTensorFrame>& frame,
+        const WriteOptions& options,
+        const std::shared_ptr<DeDupMap> &de_dup_map,
+        bool sparsify_floats,
+        bool validate_index
```
IvoDD: Nit: whitespaces changes

python/tests/integration/toolbox/test_library_tool.py
IvoDD: Looks like we have forgotten this.

python/tests/integration/toolbox/test_library_tool.py
```diff
@@ -363,6 +364,49 @@ def read_type(key, column):
...
+    assert index_key_count == 1
+    assert version_key_count == 1
+
+    
+def test_write_segment_in_memory_no_slicing(lmdb_version_store_tiny_segment):
```
IvoDD: What do you think about merging this test with the above and using a `@pytest.mark.parameterize`?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -748,44 +748,78 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
+    auto versioned_item = VersionedItem(std::move(atom_key_fut).get());
+
+    if(cfg().symbol_list())
+        symbol_list().add_symbol(store(), stream_id, versioned_item.key_.version_id());
 
     write_version_and_prune_previous(prune_previous_versions, versioned_item.key_, deleted ? std::nullopt : maybe_prev);
     return versioned_item;
 }
 
+std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_version(
```
IvoDD: This change seems unrelated? Maybe it a not well handled merge conflict.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,81 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
-    ) {
+VersionedItem LocalVersionedEngine::write_segment(
+        const StreamId& stream_id,
+        SegmentInMemory&& segment,
+        bool prune_previous_versions,
+        Slicing const& slicing
+) {
     ARCTICDB_SAMPLE(WriteVersionedDataFrame, 0)
-
+    assert(segment.descriptor().id() == stream_id);
```
IvoDD: Use `util::check` instead

cpp/arcticdb/version/local_versioned_engine.cpp
IvoDD: It's to avoid invalid symbol names. Similar to the regular write.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,81 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
+    case Slicing::RowSlicing:
+        slices = segment.split(get_write_options().segment_row_size);
+        break;
+    }
+
+    auto sink = store();
+
+    TypedStreamVersion tsv{partial_key.id, partial_key.version_id, KeyType::TABLE_DATA};
+    int64_t write_window = write_window_size();
+    auto fut_slice_keys = folly::collect(folly::window(std::move(slices), [&sink, de_dup_map, &index_desc = segment.descriptor().index(), tsv=std::move(tsv)](auto&& slice) {
```
IvoDD: Sorry if this is yet another iteration on the references. I think we should not capture any stack allocated variable by reference in this lambda.

I.e. let's remove the reference on `sink` and `index_desc`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,81 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
+            RowRange row_slice = {slice.offset(), slice.offset() + slice.row_count()};
+            auto frame_slice = FrameSlice{descriptor, column_slice, row_slice};
+            auto pkey = get_partial_key_for_segment_slice(index_desc, tsv, slice);
+            auto ks = std::make_tuple(
+                    std::move(pkey), std::move(slice), std::move(frame_slice)
+            );
+            return sink->async_write(std::move(ks), de_dup_map);
+        },write_window)).via(&async::io_executor());
+
+    // Create a TimeSeriesDescriptor that is needed for writing the index key
```
IvoDD: Nit: Move comment two lines below and change to something like:

`Create a TimeseriesDescriptor needed for the index key if segment doesn't already have one`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,81 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
+    auto index = stream::index_type_from_descriptor(segment.descriptor());
+
+    auto tsd = TimeseriesDescriptor();
```
IvoDD: Nit: it's annoying that C++ doesn't have a great way to conditionally initilize a variable and we have to default construct it first and then use the assignment operator.

What do you think about using a lambda function to assign the value of the tsd? E.g. similar to how we conditionally initilizae the output_frame [here](https://github.com/man-group/ArcticDB/blob/7f40f2647b64db91f1e56f625d1462cb1f053d22/cpp/arcticdb/entity/read_result.hpp#L89-L95)

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,81 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
+        arcticdb::proto::descriptors::NormalizationMetadata norm_meta;
+        norm_meta.mutable_df()->mutable_common()->mutable_index()->set_is_physically_stored(false);
+        norm_meta.mutable_df()->mutable_common()->mutable_index()->set_start(0);
+        norm_meta.mutable_df()->mutable_common()->mutable_index()->set_step(1);
+        tsd.set_normalization_metadata(std::move(norm_meta));
+    }
+    else {
+        tsd = segment.index_descriptor();
+    }
+    auto atom_key_fut = std::move(fut_slice_keys).thenValue([partial_key = std::move(partial_key), &sink, tsd = std::move(tsd), index = std::move(index)](auto&& slice_keys) {
```
IvoDD: Again let's not capture the `&sink` via reference. The `store` is a shared_ptr so it is not expensive to copy.
It would probably be better if we just did `sink = store()` instead of creating a local stack allocated variable.

cpp/arcticdb/version/test/test_sparse.cpp
```diff
@@ -479,7 +479,7 @@ TEST_F(SparseTestStore, SegmentAndFilterColumn) {
     bool written = false;
     DynamicAggregator aggregator(std::move(schema), [&](SegmentInMemory &&segment) {
         if(!written) {
-            test_store_->write_individual_segment(stream_id, std::move(segment), false);
+            test_store_->write_versioned_segment(stream_id, segment, false, arcticdb::version_store::Slicing::RowSlicing);
```
IvoDD: Let's preserve the old behavior and keep using `NoSlicing` for all these tests.

Also for C++ tests I think it makes more sense to use the `write_segment` variant which takes rvalue ref. The lvalue one is just for python tests.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -624,6 +624,16 @@ VersionedItem PythonVersionStore::write_versioned_dataframe(
     return versioned_item;
 }
 
+VersionedItem PythonVersionStore::write_versioned_segment(
+        const StreamId& stream_id,
+        SegmentInMemory& segment,
```
IvoDD: Maybe add a small comment explaining why we need to take an lvalue reference (because pybind requires it)

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -624,6 +624,16 @@ VersionedItem PythonVersionStore::write_versioned_dataframe(
     return versioned_item;
 }
 
+VersionedItem PythonVersionStore::write_versioned_segment(
```
IvoDD: Nit: Maybe prefix with `test_` so it's clear this is only used for tests.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,81 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
-VersionedItem LocalVersionedEngine::write_individual_segment(
-    const StreamId& stream_id,
-    SegmentInMemory&& segment,
-    bool prune_previous_versions
-    ) {
+VersionedItem LocalVersionedEngine::write_segment(
+        const StreamId& stream_id,
+        SegmentInMemory&& segment,
+        bool prune_previous_versions,
+        Slicing const& slicing
```
IvoDD: For trivially copiable types like `int`, `bool`, and trivial enums like `Slicing` it's easier to just pass by copy.
So, let's not use `const&` here and in other places for `Slicing`.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -624,6 +624,16 @@ VersionedItem PythonVersionStore::write_versioned_dataframe(
...
+        const StreamId& stream_id,
+        SegmentInMemory& segment,
+        bool prune_previous_versions,
+        Slicing const& slicing) {
+    ARCTICDB_SAMPLE(WriteVersionedDataframe, 0)
```
IvoDD: Let's remove the sampling here or at least change it to `WriteVersionedSegment`, so it's clear that it's a different operation.

python/tests/integration/toolbox/test_library_tool.py
```diff
@@ -363,6 +364,35 @@ def read_type(key, column):
...
+    sample_df = sample_dataframe()
+
+    segment = lib_tool.dataframe_to_segment_in_memory(sym, sample_df)
+    lib_tool.write_segment_in_memory(sym, segment, slicing)
+    dataframe = lib.read(sym).data
+
+    assert_frame_equal(dataframe, sample_df)
+
+    data_keys = lib_tool.find_keys(KeyType.TABLE_DATA)
+    data_key_count = len(data_keys)
```
IvoDD: Nit: `data_key_count` is unused

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -116,5 +116,30 @@ inline auto get_partial_key_gen(std::shared_ptr<InputTensorFrame> frame, TypedSt
...
 }
 
+inline stream::StreamSink::PartialKey get_partial_key_for_segment_slice(const IndexDescriptorImpl& index, const TypedStreamVersion& key, const SegmentInMemory& slice) {
+    using PartialKey = stream::StreamSink::PartialKey;
+
+    if (index.field_count() != 0) {
+        util::check(static_cast<bool>(index.type() == IndexDescriptor::Type::TIMESTAMP), "Got unexpected index type in get_partial_key_for_segment_slice");
+        auto& idx = slice.column(0);
+        assert(idx.scalar_at<timestamp>(0).has_value());
+        assert(idx.scalar_at<timestamp>(slice.row_count()-1).value());
```
alexowens90: `util::check`

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -116,5 +116,30 @@ inline auto get_partial_key_gen(std::shared_ptr<InputTensorFrame> frame, TypedSt
...
+                key.type, key.version_id, key.id, start, end_index_generator(end)
+        };
+    }
+    else {
+        return PartialKey{
+                key.type,
+                key.version_id,
+                key.id,
+                entity::safe_convert_to_numeric_index(slice.offset(), "Rows"),
+                entity::safe_convert_to_numeric_index(slice.offset() + slice.row_count() - 1, "Rows")
```
alexowens90: Is the `-1` here consistent with our other writing methods with non-timeseries data?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,83 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
+    }
+    auto partial_key = IndexPartialKey(stream_id, version_id);
+
+    // TODO: do the segment splitting in parallel
+    std::vector<SegmentInMemory> slices;
+    switch (slicing) {
+    case Slicing::NoSlicing:
+        slices = std::vector<SegmentInMemory>({segment});
+        break;
+    case Slicing::RowSlicing:
```
alexowens90, line 790: Do we need this for the primary use case of this method? The `split` will be quite slow and memory intensive

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -757,30 +757,83 @@ std::pair<VersionedItem, TimeseriesDescriptor> LocalVersionedEngine::restore_ver
...
+        slices = segment.split(get_write_options().segment_row_size);
```
alexowens90: `filter_down_stringpool` should be `true` in this call. Otherwise, every segment will have the same stringpool as the original segment, which will be wasteful of disk space in most cases

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -758,6 +762,9 @@ void register_bindings(py::module &version, py::exception<arcticdb::ArcticExcept
         .def("write_versioned_dataframe",
              &PythonVersionStore::write_versioned_dataframe,
              py::call_guard<SingleThreadMutexHolder>(), "Write the most recent version of this dataframe to the store")
+        .def("write_versioned_segment",
```
alexowens90: I'm not sure this should be public, I don't think we want users calling it?

python/tests/integration/toolbox/test_library_tool.py
```diff
@@ -363,6 +364,34 @@ def read_type(key, column):
...
+@pytest.mark.parametrize("slicing", [Slicing.NoSlicing, Slicing.RowSlicing])
+def test_write_segment_in_memory(lmdb_version_store_tiny_segment, slicing):
+    lib = lmdb_version_store_tiny_segment
+    lib_tool = lib.library_tool()
+    sym = "sym"
+    sample_df = sample_dataframe()
+
+    segment = lib_tool.dataframe_to_segment_in_memory(sym, sample_df)
+    lib_tool.write_segment_in_memory(sym, segment, slicing)
+    dataframe = lib.read(sym).data
```
alexowens90, line 377: Could also assert that the version number is correct in the returned `VersionedItem`, both when this is the first version, and if there is an existing complex version chain

cpp/arcticdb/version/local_versioned_engine.cpp
deyan37: The primary use case for now is writing reports from the reachability checker. In general, they won't have too many rows.    If there is a report with many rows, the advantage with splitting is that we write in parallel and we can read a row slice more efficiently. If the splitting is really slow we can parallelize it later in a follow-up PR @IvoDD

cpp/arcticdb/pipeline/slicing.hpp
deyan37: I removed the `-1` since the ranges were indeed exclusive, thanks

cpp/arcticdb/version/python_bindings.cpp
deyan37: Yes, I renamed it to `_test_write_versioned_segment`.

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Yeh OK

alexowens90 (approved):

PR #2522, opened 2025-07-29
Enhancement/9288960436/arrow support for reading sparse float columns

alexowens90:
#### Reference Issues/PRs
[9288960436](https://man312219.monday.com/boards/7852509418/pulses/9288960436)

Adds support for Arrow reading sparse float data. Currently limited to floats only because it is difficult to end-to-end test sparse data of other types. Will probably also work out of the box for other numeric types, strings will probably require more effort.

### DIFF
--- a/cpp/arcticdb/util/sparse_utils.cpp
+++ b/cpp/arcticdb/util/sparse_utils.cpp
@@ -27,4 +27,42 @@ util::BitSet scan_object_type_to_sparse(
     return bitset;
 }
 
+util::BitMagic truncate_sparse_map(
+        const util::BitMagic& input_sparse_map,
+        size_t start_row,
+        size_t end_row
+) {
+    // The output sparse map is the slice [start_row, end_row) of the input sparse map
+    // BitMagic doesn't have a method for this, so hand-roll it here
+    // Ctor parameter is the size
+    util::BitMagic output_sparse_map(end_row - start_row);
+    util::BitSetSizeType set_input_bit;
+    // find returns false if there are no set bits after (inclusive) start_row
+    if (input_sparse_map.find(start_row, set_input_bit)) {
+        util::BitSet::bulk_insert_iterator inserter(output_sparse_map);
+        while (set_input_bit < end_row) {
+            inserter = set_input_bit - start_row;
+            set_input_bit = input_sparse_map.get_next(set_input_bit);
+            // get_next returns 0 if there are no more set bits
+            if (set_input_bit == 0) {
+                break;
+            }
+        }
+        inserter.flush();
+    } // else no set bits after (inclusive) start_row. Ctor for output_sparse_map used initialises all bits to zero
+    return output_sparse_map;
+
+    // The above code is equivalent to the (easier to read) implementation:
+    //    for (auto en = input_sparse_map.first(); en != input_sparse_map.end(); ++en) {
+    //        if (*en < start_row) {
+    //            continue;
+    //        } else if (*en < end_row) {
+    //            inserter = *en - start_row;
+    //        } else {
+    //            break;
+    //        }
+    //    }
+    // but is more efficient in the case that the first set bit is deep into input_sparse_map
+}
+
 } //namespace arcticdb::util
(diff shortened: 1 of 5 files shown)


cpp/arcticdb/util/sparse_utils.cpp
```diff
@@ -27,4 +27,53 @@ util::BitSet scan_object_type_to_sparse(
...
+            // Shift start_row elements to the left
+            inserter = set_input_bit - start_row;
+        }
+    }
+    do {
+        set_input_bit = input_sparse_map.get_next(set_input_bit);
+        if (set_input_bit != 0 && set_input_bit < end_row) {
+            inserter = set_input_bit - start_row;
+        }
+    } while (set_input_bit != 0);
```
IvoDD: I know this code was like this before but I think we can simplify this logic if we do something like:
```
if (start_row == 0) {
    return input_sparse_map.resize(end_row)
}
set_input_bit = input_sparse_map.get_next(set_input_bit);
while (set_input_bit != 0 && set_input_bit < end_row) {
    inserter = set_input_bit - start_row;
    set_input_bit = input_sparse_map.get_next(set_input_bit);
}
```
To make the resize work inside this function we'd also need to take ownership of the `input_sparse_map`.
In any case the simplification comes from two places:
- We never really call this function with start_row = 0 because of the optimization in `handle_truncation`.
- I think it's easier to read if the condition `set_input_bit!=0 && set_input_bit < end_row` is in a single place.
- Also this way the while loop can end early when we reach `end_row`

alexowens90: We can't take ownership of the sparse map because of it's usage in `column.cpp`. I agree that this implementation is convoluted as hell though, I'll see if I can tidy it up

IvoDD (approved):

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -482,3 +483,120 @@ def test_arrow_dynamic_schema_missing_columns_hypothesis(lmdb_version_store_dyna
...
+    received = lib.read(sym, _output_format=OutputFormat.ARROW).data
+    assert table.equals(received)
+
+
+@pytest.mark.parametrize("type", [pa.float32(), pa.float64()])
+@pytest.mark.parametrize("dynamic_schema", [True, False])
+def test_arrow_sparse_floats_row_sliced(version_store_factory, type, dynamic_schema):
+    lib = version_store_factory(segment_row_size=2, dynamic_schema=dynamic_schema)
+    sym = "test_arrow_sparse_floats_row_sliced"
+    table_0 = pa.table({"col": pa.array([1, None], type)})
```
vasil-pashov, line 508: What will happen if the array contains `float("nan")`

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -482,3 +483,120 @@ def test_arrow_dynamic_schema_missing_columns_hypothesis(lmdb_version_store_dyna
...
+    df = pd.concat([df_0, df_1, df_2])
+    df.index = pd.RangeIndex(0, 15)
+    lib.write(sym, df, sparsify_floats=True)
+    row_range = (row_range_start, row_range_start + row_range_width)
+    expected = pa.concat_tables([table_0, table_1, table_2]).slice(offset=row_range[0], length=row_range[1] - row_range[0])
+    received = lib.read(sym, row_range=row_range, _output_format=OutputFormat.ARROW).data
+    assert expected.equals(received)
+
+
+@use_of_function_scoped_fixtures_in_hypothesis_checked
```
vasil-pashov, line 564: nit: other hypothesis tests are in their own file in the hypothesis folder because hypothesis is run in a different worker in the CI.

python/tests/unit/arcticdb/version_store/test_arrow.py
alexowens90: Nothing special, NaN is just a valid floating point value, `null` is a separate concept in Arrow

python/tests/unit/arcticdb/version_store/test_arrow.py
alexowens90: I want to keep all of the testing for Arrow in one place while it is still changing so quickly, we can rearrange once we're at a stable point

PR #2536, opened 2025-07-31
Fix handling of segments contained only of None values in sort merge

vasil-pashov:
#### Reference Issues/PRs
Monday: 9712540721

Sort and finalizes uses stream::Aggregator in combination with RowBuilder to combine the segments. The aggregator has row slicing policy controlled by a ConfigMap variable Merge.SegmentSize when the row builder accumulates more than Merge.SegmentSize rows (or the input is exhausted) it calls commit and writes the segment. The logic in do_merge skips `np.nan` and `None` values for string columns and does not add them to the corresponding column. It can happen so that a committed segment consist only of `np.nan` and `None` values. We don't handle this case properly in the encoding logic and throw an exception because the size of the column is 0 bytes.

#### Any other comments?
See Monday issues: 9712629170 and 9712849046
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-/cpp/ @alexowens90 @willdealtry @poodlewars
-/python/ @alexowens90 @willdealtry @poodlewars
+/cpp/ @alexowens90 @IvoDD @poodlewars
+/python/ @alexowens90 @IvoDD @poodlewars
--- a/.github/actions/set_persistent_storage_env_vars/action.yml
+++ b/.github/actions/set_persistent_storage_env_vars/action.yml
@@ -10,6 +10,8 @@ inputs:
   aws_secret_key:    {required: true, type: string, description: The value for the AWS Secret key}
   gcp_access_key:    {required: true, type: string, description: The value for the GCP Access key}      
   gcp_secret_key:    {required: true, type: string, description: The value for the GCP Secret key}
+  azure_container:    {required: true, type: string, description: The value for the Azure container}
+  azure_connection_string:    {required: true, type: string, description: The value for the Azure connection string}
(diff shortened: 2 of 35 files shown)


cpp/arcticdb/stream/merge.hpp
```diff
@@ -109,6 +110,21 @@ void do_merge(
                                 }
                             });
                         }
+                    } else {
+                        if constexpr(is_sequence_type(row_field_descriptor_tag.data_type())) {
```
vasil-pashov, line 114: This is the fix

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -947,4 +947,12 @@ def test_appending_to_empty_with_differing_columns_fails(self, lmdb_library, to_
...
         lib.write(symbol, to_append, staged=True)
         with pytest.raises(SchemaException, match="wrong_col"):
-            lib.sort_and_finalize_staged_data(symbol, mode=StagedDataFinalizeMethod.APPEND)
\ No newline at end of file
+            lib.sort_and_finalize_staged_data(symbol, mode=StagedDataFinalizeMethod.APPEND)
+
+def test_staged_segment_has_only_none(lmdb_library):
+    symbol = "symbol"
+    lib = lmdb_library
+    df = pd.DataFrame({"a": [None, None, None]}, index=pd.DatetimeIndex([pd.Timestamp(1), pd.Timestamp(2), pd.Timestamp(3)]))
```
poodlewars: Also a test with, say, [None, "str", None]?
And similar tests where we have multiple row and column slices?

IvoDD: +1, maybe also a test with multiple staged segments some of which are empty and some not

alexowens90 (approved):

IvoDD (approved):

PR #2544, opened 2025-08-04
Remove signaled threads on TaskScheduler destruction on Windows

ognyanstoimenov:
Hang on os._exit() in certain scenarios.

The problem only happens on windows because os._exit() kills all threads except the main one. Where on UNIX systems all threads are cleared and there is no hang. See https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-exitprocess

This assumes that by the time task scheduler's destructor is called, all spawned threads are no longer alive. They are either 
* Stopped gracefully by global_lifetimes.cpp::shutdown_globals
* Killed abruptly via e.g. `os._exit()` 

The latter was a potential problem until now because if threads are killed this way they aren't removed from `ThreadExecutor::taskList_` and when shutting down the `ThreadExecutor` it waited for all threads from that list to signal that they are done (via [this ](https://github.com/facebook/folly/blob/f4e89e3959fd10a34cd5284c2dbdedc0dab35cd0/folly/executors/ThreadPoolExecutor.cpp#L414) semaphore) . And since they are gone that wait is forever.

The hang only happened if ONLY CPU threads are spawned. Their clearing in the destructor is AFTER IO threads. If IOThreads exist they were killed first. No hang was happening there but again exception was thrown (which is again bad as exception in destructor is UB in c++). This PR reorders their clearing in the stop() method for consistency 

The fix removes orphaned threads from the `threadList_` (windows only) so there are no threads to wait for.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/async/task_scheduler.cpp
+++ b/cpp/arcticdb/async/task_scheduler.cpp
@@ -19,14 +19,12 @@ std::once_flag TaskScheduler::init_flag_;
 std::once_flag TaskScheduler::shutdown_flag_;
 bool TaskScheduler::forked_ = false;
 
-void TaskScheduler::destroy_instance() {
-    std::call_once(TaskScheduler::shutdown_flag_, &TaskScheduler::stop_threads);
-}
-
-void TaskScheduler::stop_threads() {
-    if(TaskScheduler::instance_) {
-        TaskScheduler::instance()->stop();
-    }
+void TaskScheduler::stop_active_threads() {
+    std::call_once(shutdown_flag_, [] {
+        if (instance_) {
+            instance()->stop();
+        }
+    });
 }
 
 void TaskScheduler::reattach_instance() {
@@ -52,7 +50,7 @@ void TaskScheduler::init(){
 }
 
 TaskSchedulerPtrWrapper::~TaskSchedulerPtrWrapper() {
-    ptr_->all_threads_dead();
+    ptr_->stop_orphaned_threads();
     delete ptr_;
 }
 
--- a/cpp/arcticdb/async/task_scheduler.cpp
+++ b/cpp/arcticdb/async/task_scheduler.cpp
@@ -20,14 +20,12 @@ std::once_flag TaskScheduler::shutdown_flag_;
 bool TaskScheduler::forked_ = false;
 
 void TaskScheduler::destroy_instance() {
-    std::call_once(TaskScheduler::shutdown_flag_, &TaskScheduler::stop_and_destroy);
+    std::call_once(TaskScheduler::shutdown_flag_, &TaskScheduler::stop_threads);
 }
 
-void TaskScheduler::stop_and_destroy() {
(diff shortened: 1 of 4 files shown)


poodlewars (approved):

cpp/arcticdb/async/task_scheduler.cpp
```diff
@@ -54,6 +50,7 @@ void TaskScheduler::init(){
 }
 
 TaskSchedulerPtrWrapper::~TaskSchedulerPtrWrapper() {
+    ptr_->stop_orphaned_threads();
```
alexowens90, line 53: There's still a narrow race condition here, where the threads become signalled after the call to `stop_orphaned_threads()` and before the `delete`, but I don't think we can do anything about that

python/tests/stress/arcticdb/version_store/test_deallocation.py
```diff
@@ -25,3 +28,33 @@ def test_many_version_store(basic_store_factory):
...
         vit = version_store.read(symbol)
         assert_frame_equal(vit.data, df2)
+
+
+def killed_worker(lib, io_threads, cpu_threads):
+    if io_threads:
+        lib.delete_snapshot("snap")
+    if cpu_threads:
+        lib.read("sym")
+    os._exit(0)
```
alexowens90, line 38: It's not obvious to me that this proves the fix works. There won't be any threads doing any work at the time of the `_exit` call, and folly is free to terminate idle threads. This process could spawn some Python threads to repeatedly call the 2 API methods, and then call `_exit` from the main thread, that way (most likely) some threads will be doing work at the time of the `_exit` call.
We don't need to leave them in, but adding some logs to prove that the expected code paths are being taken (just verified by eye) would also help convince that this is the right fix

python/tests/stress/arcticdb/version_store/test_deallocation.py
```diff
@@ -25,3 +28,33 @@ def test_many_version_store(basic_store_factory):
...
+
+@pytest.mark.parametrize("io_threads_cpu_threads_spawned_in_child", [
+    (True, True),
+    (True, False),
+    (False, True),
+    (False, False),
+])
```
alexowens90:
```suggestion
@pytest.mark.parametrize("io_threads_spawned_in_child", [True, False])
@pytest.mark.parametrize("cpu_threads_spawned_in_child", [True, False])
```

python/tests/stress/arcticdb/version_store/test_deallocation.py
ognyanstoimenov: I ran the same tests without the fix to prove they hang: https://github.com/man-group/ArcticDB/actions/runs/16834505248/job/47691535534

alexowens90 (approved):

PR #2546, opened 2025-08-04
Trim asv benchmarks so they take less time

G-D-Petrov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/benchmarks/version_chain.py
+++ b/python/benchmarks/version_chain.py
@@ -5,6 +5,7 @@ Use of this software is governed by the Business Source License 1.1 included in 
 
 As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
 """
+
 import datetime
 import os
 
@@ -15,6 +16,7 @@ import arcticdb as adb
 import datetime
 import math
 import sys
+import time
 
 from .common import *
 
@@ -22,7 +24,7 @@ from .common import *
 class IterateVersionChain:
     number = 10
     timeout = 6000
-    warmup_time = 0    
+    warmup_time = 0
     CONNECTION_STRING = "lmdb://version_chain?map_size=20GB"
     LIB_NAME = "lib"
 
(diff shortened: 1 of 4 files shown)


python/benchmarks/version_chain.py
```diff
@@ -36,21 +38,39 @@ def symbol(self, num_versions, deleted):
...
+                symbol = self.symbol(num_versions, deleted)
+                delete_points[symbol] = math.floor(deleted * num_versions)
 
+        start_time = time.time()
+        adb._ext.set_config_int("VersionMap.ReloadInterval", sys.maxsize)
+        # Batch operations by symbol to reduce overhead
         for num_versions in num_versions_list:
             for deleted in deleted_list:
                 symbol = self.symbol(num_versions, deleted)
+                delete_point = delete_points[symbol]
```
IvoDD, line 62: Wouldn't it be simpler if we just did here `delete_point = math.floor(deleted*num_versions)`? 
I think using the dict adds some minor extra complexity.

Also out of curiousity what is the impact of the `math.floor` calls? I'd assume they are orders of magnitude faster than the `lib.write`.

python/benchmarks/version_chain.py
```diff
@@ -15,14 +16,15 @@
...
+import time
 
 from .common import *
 
 
 class IterateVersionChain:
     number = 10
     timeout = 6000
-    warmup_time = 0    
+    warmup_time = 0
```
IvoDD, line 27: Just to confirm. `rounds` defaults to 2 as per [docs](https://asv.readthedocs.io/en/latest/benchmarks.html#timing-benchmarks). Will it still run `setup_cache` just once? If not maybe we should just use `rounds=1`

IvoDD (approved):

python/benchmarks/version_chain.py
```diff
@@ -36,21 +38,39 @@ def symbol(self, num_versions, deleted):
...
         self.ac.delete_library(IterateVersionChain.LIB_NAME)
         lib = self.ac.create_library(IterateVersionChain.LIB_NAME)
 
         small_df = generate_random_floats_dataframe(2, 2)
+        # Pre-calculate delete points to avoid repeated math.floor calls
+        delete_points = {}
+        for num_versions in num_versions_list:
+            for deleted in deleted_list:
+                symbol = self.symbol(num_versions, deleted)
+                delete_points[symbol] = math.floor(deleted * num_versions)
```
vasil-pashov, line 54: It's odd that  `deleted` and `num_versions` can be float. Do you know why is that?

python/benchmarks/version_chain.py
G-D-Petrov: I was just trying to optimize away the calculation as that tends to be a bit slow in Python.
The impact of precomputing the delete_point in advance is just a couple of seconds

python/benchmarks/version_chain.py
G-D-Petrov: `num_versions` is an int and `deleted` is a float that is trying express where we should have a deleted version in the version chain as a percentage. e.g. after 99% of versions have been written, delete a version

python/benchmarks/version_chain.py
G-D-Petrov: This is a good question, will check

PR #2468, opened 2025-07-15
Fix will_item_be_pickled not consider recursive normalization

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/9122332360
#### What does this implement or fix?
Fix will_item_be_pickled not consider recursive normalization if the option is ON.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #2548, opened 2025-08-04
Add docstring and warning for disparity between `will_item_be_pickled` and `is_symbol_pickle`

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/9122332360
The [PR](https://github.com/man-group/ArcticDB/pull/2468) that trying to align `will_item_be_pickled` and `is_symbol_pickled` has been scrapped as the change is likely to break users' logic.
Instead better docstrings are added for both. Better warning for `will_item_be_pickled` only as doing so for `is_symbol_pickle` requires substantatial effort
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -12,6 +12,7 @@ import datetime
 import os
 import sys
 from warnings import warn
+from collections import namedtuple
 
 import pandas as pd
 import numpy as np
@@ -60,6 +61,7 @@ from arcticdb_ext.version_store import DataError, KeyNotFoundInStageResultInfo
 from arcticdb_ext.version_store import sorted_value_name
 from arcticdb_ext.version_store import ArrowOutputFrame, InternalOutputFormat
 from arcticdb.options import RuntimeOptions, OutputFormat, output_format_to_internal
+from arcticdb_ext.log import LogLevel as _LogLevel
 from arcticdb.authorization.permissions import OpenMode
 from arcticdb.exceptions import ArcticDbNotYetImplemented, ArcticNativeException, MissingKeysInStageResultsError
 from arcticdb.flattener import Flattener
@@ -90,6 +92,7 @@ import arcticdb_ext as ae
 
 IS_WINDOWS = sys.platform == "win32"
 
+FlattenResult = namedtuple("FlattenResult", ["is_recursive_normalize_preferred", "metastruct", "to_write"])
 
 def resolve_defaults(param_name, proto_cfg, global_default, existing_value=None, uppercase=True, runtime_options=None, **kwargs):
     """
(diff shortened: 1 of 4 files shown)


python/arcticdb/version_store/_store.py
```diff
@@ -2951,9 +2991,25 @@ def will_item_be_pickled(self, item):
             )
         except Exception:
             # This will also log the exception inside composite normalizer's normalize
-            return True
+            result = True
 
-        return norm_meta.WhichOneof("input_type") == "msg_pack_frame"
+        result |= norm_meta.WhichOneof("input_type") == "msg_pack_frame"
```
alexowens90, line 3022: Is this correct if the item to be normalized is a POD dict like `{"hello": "there"}`? Then I think the norm meta will still say msgpack, but nothing will be pickled?

python/arcticdb/version_store/_store.py
```diff
@@ -2951,9 +2991,25 @@ def will_item_be_pickled(self, item):
...
+            proto_cfg = self._lib_cfg.lib_desc.version.write_options
+            resolved_recursive_normalizers = resolve_defaults(
+                "recursive_normalizers", proto_cfg, global_default=False, uppercase=False, **{"recursive_normalizers": recursive_normalizers}
+            )
+            if resolved_recursive_normalizers:
+                is_recursive_normalize_preferred, _, _ = self._try_flatten(item, "")
+                if is_recursive_normalize_preferred:
+                    log.warning("As the library setting recursive_normalizers is enabled, the item "
+                                "will be recursively normalized in `write`. "
+                                "However, for backward compatibility, this API will still return True.")
```
alexowens90: It isn't necessarily the library setting, it could be the env var or the argument to this method.
I would also combine with the other call to `log.warning`, there is no guarantee these logs will appear near each other in a busy logfile.
I would also add more detail as to why this method has historically returned true even if pickling is not involved (not date_range searchable, queryable, etc).
I think we also discussed being able to disable this log with an env var?

python/arcticdb/version_store/_store.py
phoebusm:
```
In [11]: dict_data = {"hello": "there"}

In [12]: lib._nvs.write("test", dict_data, recursive_normalizers=True)
Out[12]: VersionedItem(symbol='test', library='test', data=n/a, version=2, metadata=None, host='S3(endpoint=s3.eu-west-1.amazonaws.com, bucket=arcticdb-ci-test-bucket-02)', timestamp=1754503211644392787)

In [13]: lib._nvs.is_symbol_pickled("test")
Out[13]: True

In [14]: lib._nvs.will_item_be_pickled(dict_data, recursive_normalizers=True)
Out[14]: True
```
Yes it'll msgpack-normalized. Both APIs return true but because of different reasons. 
`is_symbol_pickled`: as the type of data is not native
`will_item_be_pickled`: As it considers msgpack-normalized data "pickled"

python/arcticdb/version_store/_store.py
phoebusm: The warning can be disabled by setting the log level to `INFO`.

alexowens90: Log level is a global setting though, we want to be able to just disable this message

python/tests/unit/arcticdb/version_store/test_normalization.py
```diff
@@ -541,6 +541,41 @@ def test_will_item_be_pickled(lmdb_version_store, sym):
...
 
 
+@pytest.mark.parametrize(
+    "data",
+    [
+        {"a": {"b": {"c": {"d": np.arange(24)}}}},
+        {"a": [1, 2, 3], "b": {"c": np.arange(24)}, "d": [TestCustomNormalizer()]} # A random item that will be pickled
+    ]
+)
+def test_will_item_be_pickled_recursive_normalizer(lmdb_version_store_v1, data, capfd):
```
alexowens90: We stopped using capfd as all the tests it was used for were [flaky](https://github.com/search?q=repo%3Aman-group%2FArcticDB+get_stderr&type=code). If all these tests passed for you locally we can live with the warning messages not being tested

python/arcticdb/version_store/_store.py
```diff
@@ -2951,9 +2992,33 @@ def will_item_be_pickled(self, item):
             )
         except Exception:
             # This will also log the exception inside composite normalizer's normalize
-            return True
+            result = True
 
-        return norm_meta.WhichOneof("input_type") == "msg_pack_frame"
+        result |= norm_meta.WhichOneof("input_type") == "msg_pack_frame"
+        log_warning_message = strtobool(os.getenv("VersionStore.WillItemBePickledWarningMsg", "1")) and log.is_active(_LogLevel.WARN)
```
alexowens90: Env var should follow our [naming convention](https://docs.arcticdb.io/latest/runtime_config/#environment-variables)

phoebusm: It should be a config instead of env var

G-D-Petrov (approved):

PR #2550, opened 2025-08-05
Support open ended row_range on QueryBuilder and read methods

poodlewars:
Changes `QueryBuilder.row_range` and `read` and `ReadRequest`'s `row_range` arguments to accept `None` in either position, in order to leave that part of the row range open ended. For example `(5, None)` would include everything from the 5th row onwards.

Monday: 7855088823

### DIFF
--- a/cpp/arcticdb/processing/clause.hpp
+++ b/cpp/arcticdb/processing/clause.hpp
@@ -681,10 +681,20 @@ struct RowRangeClause {
         clause_info_.input_structure_ = ProcessingStructure::ALL;
     }
 
-    explicit RowRangeClause(int64_t start, int64_t end):
-            row_range_type_(RowRangeType::RANGE),
-            user_provided_start_(start),
-            user_provided_end_(end) {
+    explicit RowRangeClause(std::optional<int64_t> start, std::optional<int64_t> end) {
+        util::check(start || end, "Expect at least one of start and end to be present");
+        if (start && end) {
+            row_range_type_ = RowRangeType::RANGE;
+            user_provided_start_ = *start;
+            user_provided_end_ = *end;
+        } else if (start) {
+            util::check(start != 0, "Did not expect end=nullopt and start==0");
+            row_range_type_ = RowRangeType::TAIL;
+            n_ = -1 * start.value();
+        } else if (end) {
+            row_range_type_ = RowRangeType::HEAD;
+            n_ = end.value();
+        }
         clause_info_.input_structure_ = ProcessingStructure::ALL;
     }
 
--- a/cpp/arcticdb/processing/clause.hpp
+++ b/cpp/arcticdb/processing/clause.hpp
@@ -682,12 +682,14 @@ struct RowRangeClause {
     }
 
     explicit RowRangeClause(std::optional<int64_t> start, std::optional<int64_t> end) {
+        // start and end both absent is a no-op, the Python layer just skips the clause in this case
         util::check(start || end, "Expect at least one of start and end to be present");
         if (start && end) {
             row_range_type_ = RowRangeType::RANGE;
             user_provided_start_ = *start;
             user_provided_end_ = *end;
         } else if (start) {
+            // start=0 and end absent is a no-op, the Python layer just skips the clause in this case
             util::check(start != 0, "Did not expect end=nullopt and start==0");
             row_range_type_ = RowRangeType::TAIL;
             n_ = -1 * start.value();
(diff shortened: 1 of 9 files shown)


python/tests/unit/arcticdb/version_store/test_row_range.py
```diff
@@ -89,3 +91,31 @@ def test_row_range_pickled_symbol(lmdb_version_store):
...
+
+
+@pytest.mark.parametrize("row_range,expected", (
+    ((-5, None), pd.DataFrame({"a": np.arange(95, 100)})),
+    ((5, None), pd.DataFrame({"a": np.arange(5, 100)})),
+    ((0, None), pd.DataFrame({"a": np.arange(0, 100)})),
+    ((None, -5), pd.DataFrame({"a": np.arange(95)})),
+    ((None, 5), pd.DataFrame({"a": np.arange(5)})),
+    ((None, 0), pd.DataFrame({"a": []}, dtype=np.int64)),
+    ((None, None), pd.DataFrame({"a": np.arange(100)})),
```
vasil-pashov, line 103: Looking at `RowRangeClause`
```
util::check(start || end, "Expect at least one of start and end to be present");
```
(None, None) is supposed to throw but the test passes. What am I missing?

python/arcticdb/version_store/processing.py
```diff
@@ -971,6 +973,15 @@ def row_range(self, row_range: Tuple[int, int]):
...
         end = row_range[1]
 
+        if start is not None and not isinstance(start, int):
+            raise UserInputException(f"The first element of row_range must be None or int but was: {start}")
+
+        if end is not None and not isinstance(end, int):
+            raise UserInputException(f"The second element of row_range must be None or int but was: {end}")
+
+        if end is None and not start:
+            return self  # no-op
```
phoebusm, line 983: I am a bit curious why it's a no-op rather than raising exception?

python/tests/unit/arcticdb/version_store/test_row_range.py
poodlewars: It's handled in the Python layer in `processing.py`

python/arcticdb/version_store/processing.py
poodlewars: Because it's a valid slicing operation?

cpp/arcticdb/processing/clause.hpp
```diff
@@ -681,10 +681,20 @@ struct RowRangeClause {
...
-            user_provided_start_(start),
-            user_provided_end_(end) {
+    explicit RowRangeClause(std::optional<int64_t> start, std::optional<int64_t> end) {
+        util::check(start || end, "Expect at least one of start and end to be present");
+        if (start && end) {
+            row_range_type_ = RowRangeType::RANGE;
+            user_provided_start_ = *start;
+            user_provided_end_ = *end;
+        } else if (start) {
+            util::check(start != 0, "Did not expect end=nullopt and start==0");
```
alexowens90, line 693: Might be worth a comment that this case is a no-op and so handled in the Python layer, I was confused why this check was needed for a valid slicing operation

python/tests/unit/arcticdb/version_store/test_row_range.py
```diff
@@ -89,3 +91,31 @@ def test_row_range_pickled_symbol(lmdb_version_store):
...
+    ((5, 3), pd.DataFrame({"a": []}, dtype=np.int64)),
+))
+@pytest.mark.parametrize("api", ("query_builder", "read", "read_batch"))
+def test_row_range_open_ended_negative_start(lmdb_version_store_v1, api, row_range, expected):
```
alexowens90: nit: Bad test name

python/tests/unit/arcticdb/version_store/test_row_range.py
```diff
@@ -89,3 +91,31 @@ def test_row_range_pickled_symbol(lmdb_version_store):
...
+    symbol = "test_row_range"
+    df = pd.DataFrame({"a": np.arange(100)})
```
alexowens90, line 109: Is `expected` always equal to `df[row_range[0]:row_range[1]]`? If so might be more readable

alexowens90 (approved):

python/arcticdb/version_store/processing.py
phoebusm: My bad! I have misread it

python/tests/unit/arcticdb/version_store/test_row_range.py
poodlewars: Almost, but the index messes things up a bit. I think it's quite nice having the result you expect in black and white in the parameters.

PR #2553, opened 2025-08-05
Fix release build on GCC 13.3

alexowens90:
GCC 13.3 release build has some stricter checks than compilers we use in our CI, see comments on individual changes for details
Still does not work out of the box due to a bug in the `stringop-overflow` check

### DIFF
--- a/cpp/arcticdb/processing/clause.cpp
+++ b/cpp/arcticdb/processing/clause.cpp
@@ -363,9 +363,16 @@ std::vector<EntityId> AggregationClause::process(std::vector<EntityId>&& entity_
     auto row_slices = split_by_row_slice(std::move(proc));
 
     // Sort procs following row range descending order, as we are going to iterate through them backwards
+    // front() is UB if vector is empty. Should be non-empty by construction, but exception > UB
+    internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+            ranges::all_of(row_slices, [](const auto& proc) {
+                return proc.row_ranges_.has_value() && !proc.row_ranges_->empty();
+            }),
+            "Unexpected empty row_ranges_ in AggregationClause::process"
+            );
     ranges::sort(row_slices,
                  [](const auto& left, const auto& right) {
-                     return left.row_ranges_->at(0)->start() > right.row_ranges_->at(0)->start();
+                    return left.row_ranges_->front()->start() > right.row_ranges_->front()->start();
                  });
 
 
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/processing/clause.cpp
```diff
@@ -363,9 +363,16 @@ std::vector<EntityId> AggregationClause::process(std::vector<EntityId>&& entity_
...
+    internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+            ranges::all_of(row_slices, [](const auto& proc) {
+                return proc.row_ranges_.has_value() && !proc.row_ranges_->empty();
+            }),
+            "Unexpected empty row_ranges_ in AggregationClause::process"
+            );
     ranges::sort(row_slices,
                  [](const auto& left, const auto& right) {
-                     return left.row_ranges_->at(0)->start() > right.row_ranges_->at(0)->start();
+                    return left.row_ranges_->front()->start() > right.row_ranges_->front()->start();
```
alexowens90, line 375: Compiler cannot tell that `row_ranges_` has a value and is non-empty by construction

cpp/arcticdb/util/allocator.cpp
```diff
@@ -319,7 +320,7 @@ namespace arcticdb {
             uintptr_t(ret));
 #endif
         auto ts = current_timestamp();
-        TracingPolicy::track_realloc(std::make_pair(uintptr_t(ptr.first), ptr.second), std::make_pair(uintptr_t(ret), ts), size);
+        TracingPolicy::track_realloc(old_addr, std::make_pair(uintptr_t(ret), ts), size);
```
alexowens90, line 323: `std::realloc` invalidates the `uint8*` in `ptr.first`

cpp/arcticdb/storage/coalesced/multi_segment_utils.hpp
```diff
@@ -27,7 +27,7 @@ uint64_t get_symbol_prefix(const StreamId& stream_id) {
             for(size_t p = begin, i = 0; p < end && i < string_id.size(); ++p, ++i) {
                 const auto c = string_id[i];
                 util::check(c < 127, "Out of bounds character {}", c);
-                target[p] = c;
+                target[i] = c;
```
alexowens90, line 30: This looks like it was just a bug. `StorageType` is `uint32_t` in the only use of this function, which means `begin == 4`, and `target` would always be out of bounds from `data`

poodlewars (approved):

vasil-pashov (approved):

PR #2560, opened 2025-08-08
Create common logger for all tests

grusev:

ASV Tests need logger
unit and functional tests need logger
installation tests need logger

All those types of tests need one logger that will do Sanitization and other things. Currenlty ASV has one and unit and functional tests have another and they cannot share it due to limitations imposed by ASV. This PR creates a common one that is needed by the Azure PR https://github.com/man-group/ArcticDB/pull/2498 which unites all tets to use this logger. 

However ASV tests cannot run successfully unless this logger is already present in master. So this PR is to fix this problem by merging it to master prior and then update the Azure PR to verify all is fine

The logger location should be at arcticdb\util so that all types of tests have access to it. ASV tests do not have access to tests package

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/util/logger.py
+++ b/python/arcticdb/util/logger.py
@@ -1,0 +1,77 @@
+"""
+Copyright 2025 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file LICENSE.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+import inspect
+import logging
+import os
+import re
+from typing import Any, Dict, Union
+
+
+class GitHubSanitizingHandler(logging.StreamHandler):
+    """
+    The handler sanitizes messages only when execution is in GitHub
+    """
+
+    def emit(self, record: logging.LogRecord):
+        # Sanitize the message here
+        record.msg = self.sanitize_message(record.msg)
+        super().emit(record)
+
+    @staticmethod
+    def sanitize_message(message: str) -> str:
+        if (os.getenv("GITHUB_ACTIONS") == "true") and isinstance(message, str):
+            # Use regex to find and replace sensitive access keys
+            sanitized_message = re.sub(r'(secret=)[^\s&]+', r'\1***', message)
+            sanitized_message = re.sub(r'(access=)[^\s&]+', r'\1***', sanitized_message)
+            sanitized_message = re.sub(r'AccountKey=([^;]+)', r'AccountKey=***', sanitized_message) 
+            return sanitized_message
+        return message
+
+
+loggers:Dict[str, logging.Logger] = {}
+
+
+def get_logger(bencmhark_cls: Union[str, Any] = None):
+    """
+    Creates logger instance with associated console handler.
+    The logger name can be either passed as string or class,
+    or if not automatically will assume the caller module name
+    """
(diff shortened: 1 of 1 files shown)


python/arcticdb/util/logger.py
```diff
@@ -0,0 +1,77 @@
...
+    The handler sanitizes messages only when execution is in GitHub
+    """
+
+    def emit(self, record: logging.LogRecord):
+        # Sanitize the message here
+        record.msg = self.sanitize_message(record.msg)
+        super().emit(record)
+
+    @staticmethod
+    def sanitize_message(message: str) -> str:
```
vasil-pashov, line 27: If we know that message is URL it would be safer to use some library for parsing URLs to obfuscate the parameters. If there are other things in the message then I guess this would be fine.

A more pythonic style would be to make it a free function.

python/arcticdb/util/logger.py
```diff
@@ -0,0 +1,77 @@
...
+    The logger name can be either passed as string or class,
+    or if not automatically will assume the caller module name
+    """
+    logLevel = logging.INFO
+    if bencmhark_cls:
+        if isinstance(bencmhark_cls, str):
+            value = bencmhark_cls
+        else:
+            value = type(bencmhark_cls).__name__
+        name = value
```
vasil-pashov, line 52:
```suggestion
        name = bencmhark_cls if isinstance(bencmhark_cls, str) else type(bencmhark_cls).__name__
```

python/arcticdb/util/logger.py
```diff
@@ -0,0 +1,77 @@
...
+loggers:Dict[str, logging.Logger] = {}
+
+
+def get_logger(bencmhark_cls: Union[str, Any] = None):
+    """
+    Creates logger instance with associated console handler.
+    The logger name can be either passed as string or class,
+    or if not automatically will assume the caller module name
+    """
+    logLevel = logging.INFO
```
vasil-pashov, line 46:
```suggestion
    log_level = logging.INFO
```

Is there a way to request different log level?

python/arcticdb/util/logger.py
```diff
@@ -0,0 +1,77 @@
...
+    console_handler = GitHubSanitizingHandler()
+    console_handler.setLevel(logLevel)
+    formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+    console_handler.setFormatter(formatter)
+    logger.addHandler(console_handler)
+    loggers[name] = logger
+    return logger
+
+
+class GitHubSanitizingException(Exception):
```
vasil-pashov, line 72: When will this be used?

python/arcticdb/util/logger.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+
+    def emit(self, record: logging.LogRecord):
+        # Sanitize the message here
+        record.msg = self.sanitize_message(record.msg)
+        super().emit(record)
+
+    @staticmethod
+    def sanitize_message(message: str) -> str:
+        if (os.getenv("GITHUB_ACTIONS") == "true") and isinstance(message, str):
```
vasil-pashov, line 28: What can message be if it's not a string? If we expect it to always be a string it's better to remove the check and let it crash when something else is passed, otherwise it'll silently pass with unexpected input

PR #2571, opened 2025-08-12
Publish pytest xml data

G-D-Petrov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -434,6 +434,18 @@ jobs:
       pytest_xdist_mode: ${{matrix.pytest_xdist_mode}}
       pytest_args: ${{inputs.pytest_args}}
 
+  publish_pytest_data:
+    needs: [common_config,build-python-wheels-linux, build-python-wheels-windows, build-python-wheels-macos]
+    if: |
+      always() &&
+      !failure() &&
+      !cancelled()
+    uses: ./.github/workflows/publish_pytest_data.yml
+    secrets: inherit
+    permissions: {contents: write}
+    with:
+      environment: ${{needs.common_config.outputs.publish_env}}
+
   can_merge:
     needs: [cpp-test-linux, cpp-test-windows, cpp-test-macos, build-python-wheels-linux, build-python-wheels-windows, build-python-wheels-macos, persistent_storage_verify_linux, persistent_storage_verify_windows]
     if: |
--- a/.github/workflows/publish_pytest_data.yml
+++ b/.github/workflows/publish_pytest_data.yml
@@ -1,0 +1,26 @@
+name: Publish pytest data to S3
+on: 
+  workflow_call:
+    inputs:
+      environment: { type: string, required: true }
+  workflow_dispatch:
+    inputs:
+      environment: { type: environment, required: true }
+      run_id: { type: number, required: true }
+run-name: Publish ${{github.ref}} to arcticdb-pytest-data
+jobs:
+  sync_pytest_xmls:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Gather pytest xmls from run ${{ github.run_id }}
+        id: gather-pytest-xmls
+        uses: actions/download-artifact@v4
+        with:
+            pattern: "pytest-*"
+            run-id: ${{ github.run_id }}
+            github-token: ${{github.token}}  
+
+      - name: Print pytest xmls
+        run: |
+          ls -la ${{runner.temp}}
+          cat ${{runner.temp}}/*test*


build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+        else:
+            all_runs = [run.stem.split("_")[1] for run in download_dir.glob("*.xml")]
+    else:
+        # Get a single workflow run
+        all_runs = [run_id]
+
+    # Process runs and download artifacts
+    print(f"\nProcessing {len(all_runs)} workflow runs...")
+
+    # Process runs in parallel
```
IvoDD: The csv extraction and writing is sequential? Maybe it's not that slow and it's fine but maybe at least tweak the comment?

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+    csv_path = unified_xml.with_suffix(".csv")
+
+    if not csv_path.exists():
+        # Create CSV from unified XML
+        csv_path = create_csv_from_unified_xml(unified_xml, csv_path)
+
+    return csv_path
+
+
+def get_all_workflow_runs_parallel(max_workers=5, max_pages=20):
```
IvoDD: Nit: Maybe remove the default args and just rely on the click CLI defaults?

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+        "-H",
+        "X-GitHub-Api-Version: 2022-11-28",
+        f"/repos/man-group/ArcticDB/actions/runs?per_page={per_page}&page={page}",
+    ]
+
+    output = subprocess.run(cmd, capture_output=True, text=True)
+
+    if output.returncode != 0:
+        print(f"Error on page {page}: {output.stderr}")
+        return []
```
IvoDD: I suspect this should not happen that frequently? Do we want to silently fail midway and only upload all csv up to the failure point? I suspect raising and exception might be better

alexowens90: I think this means the workflow will succeed even if the command failed? Could we pass `check=True` to the subprocess run so it's obvious when the scraping is failing?

build_tooling/process_pytest_artifacts.py
alexowens90: Why is `max_pages` an input? Can it not be queried from the API somehow?

build_tooling/process_pytest_artifacts.py
IvoDD: Similar question for `get_artifacts_for_run`

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+
+
+def save_csv_files_to_lib(csv_files):
+    lib = get_results_lib("pytest_results")
+
+    for csv_file in csv_files:
+        run_id = csv_file.stem
+        print(f"Saving {csv_file} for {run_id}")
+        df = pd.read_csv(csv_file)
+        lib.write(run_id, df)
```
alexowens90: This is pretty opaque for a symbol name, particularly one the workflow URL gets cleaned up. Could we include something more identifiable (branch name, commit hash, timestamp) as well?

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+
+        for future in as_completed(future_to_page):
+            page_num = future_to_page[future]
+            try:
+                runs = future.result()
+                if runs:
+                    all_runs.extend(runs)
+                else:
+                    print(f"Empty page {page_num}, reached the end")
+                    break
```
IvoDD: Is the order of iterating `as_completed(future_to_page)` guaranteed to be the same as the order they are put in?
I think they will be returned in the ~order they complete in which can be arbirary. Unclear docs [here](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed). Even if `as_completed` adhered to initial ordering I don't think we have guarantees on the dictionary iteration?

So if e.g. `fetch_page(max_pages)` finishes before the rest we would end early and not get all pages?

build_tooling/process_pytest_artifacts.py
alexowens90: This is running on every commit, right? That'll generate a lot of symbols, might be worth running `list_symbols` after the `write` so that the cache gets compacted regularly

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+            print("Downloading pytest XMLs...")
+            start_time = time.time()
+            download_pytest_xmls_in_parallel(all_runs, download_dir, max_workers=max_workers)
+            end_time = time.time()
+            print(f"Time taken: {end_time - start_time:.2f} seconds")
+        else:
+            all_runs = [run.stem.split("_")[1] for run in download_dir.glob("*.xml")]
+    else:
+        # Get a single workflow run
+        all_runs = [run_id]
```
IvoDD: Shouldn't we `download_pytest_xmls` in this case? I think we should probably separate getting the runs from getting the xmls like so:
```
if download_dir.exists():
    all_runs = extract_runs_from_xmls()
else:
    if run_id is None:
         all_runs = get_all_workflow_runs_parallel()
    else:
        all_runs = [run_id]
    download_pytest_xmls_in_parallel(all_runs)
```

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+        run_started_at = run.get("run_started_at", "Unknown")
+    else:
+        assert isinstance(run, str), f"Run must be a dict or a string, got {type(run)}"
+        run_id = run
+        run_name = "Unknown"
+        run_status = "Completed"
+        run_conclusion = "Unknown"
+        run_started_at = "Unknown"
+
+    if run_conclusion == "cancelled":
```
IvoDD: Should we also `run_conclusion.lower()` similar to `run_status.lower()` below?

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+def process_workflow_run(run: Union[dict, str], download_dir: Path) -> List[Path]:
+    """Process a single workflow run - get artifacts and download them"""
+    if isinstance(run, dict):
+        run_id = run["id"]
+        run_name = run.get("name", "Unknown")
+        run_status = run.get("status", "Unknown")
+        run_conclusion = run.get("conclusion", "Unknown")
+        run_started_at = run.get("run_started_at", "Unknown")
+    else:
+        assert isinstance(run, str), f"Run must be a dict or a string, got {type(run)}"
```
IvoDD: Myabe it would be worth explaining when `isinstance(run, str)` in a comment? 
Is it only when we already have made the downloads and are extracting the runs from xmls? 
If so, why do we need to handle this case in `process_workflow_run`? Won't the xmls already be there and we wouldn't need to extract the zips again?

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+
+    # Create directory for this artifact
+    artifact_dir = download_dir / f"run_{artifact['workflow_run']['id']}_{artifact_name}"
+
+    # check if the directory already exists
+    if artifact_dir.exists():
+        # print(f"Artifact {artifact_name} already exists, skipping...")
+        return None
+
+    artifact_dir.mkdir(parents=True, exist_ok=True)
```
IvoDD: Nit: Given the check above we don't need the `exists_ok=True`?

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+            "-H",
+            "X-GitHub-Api-Version: 2022-11-28",
+            f"/repos/man-group/ArcticDB/actions/artifacts/{artifact_id}/zip",
+        ]
+
+        output = subprocess.run(cmd, stdout=f, stderr=subprocess.PIPE)
+
+    if output.returncode != 0:
+        print(f"Error downloading {artifact_name}: {output.stderr.decode()}")
+        zip_path.unlink(missing_ok=True)  # Remove the file if download failed
```
IvoDD: Shouldn't we also remove the directory? Otherwise a follow up run would see `artifact_dir.exists()` and not try to download the zip at all.

build_tooling/process_pytest_artifacts.py
```diff
@@ -0,0 +1,556 @@
...
+        if py_match:
+            python_version = f"{py_match.group(1)}"
+
+        subversion = python_version[1:]
+
+        assert subversion in ["6", "7", "8", "9", "10", "11", "12", "13"], python_version
+
+        python_version = f"3.{subversion}"
+
+        return python_version, test_type
```
IvoDD: It might be useful to also extract the `DefaultCache` vs `NoCache`. E.g. it would be interesting to see if `NoCache` runs are super slow.

build_tooling/process_pytest_artifacts.py
G-D-Petrov: The single run case should be used when the script is run by the CI and then we are assuming that the xmls have been downloaded/extracted by [this action](https://github.com/man-group/ArcticDB/pull/2571/files#diff-8ea013d267079d62e7235b6be73aa91b74f8e2bd87fce185b5bee65b4456df68R30)

build_tooling/process_pytest_artifacts.py
G-D-Petrov: Yeah, added a step at the end of the script to remove the directory

build_tooling/process_pytest_artifacts.py
G-D-Petrov: This part has been changed and now takes in just a `Run` object

build_tooling/process_pytest_artifacts.py
G-D-Petrov: changed to remove the `break`

build_tooling/process_pytest_artifacts.py
G-D-Petrov: yeah, changed to throw

build_tooling/process_pytest_artifacts.py
G-D-Petrov: Added a way to query the API but kept the `max_pages` bit because the GH API seems to enable us to handle about 20 pages before throttling us

build_tooling/process_pytest_artifacts.py
```diff
@@ -53,39 +140,38 @@ def get_artifacts_for_run(artifact_download_url):
...
+    except subprocess.CalledProcessError as e:
+        print(f"Error running command: {' '.join(cmd)}")
+        print(f"Return code: {e.returncode}")
+        print(f"Error output: {e.stderr.decode('utf-8')}")
+        raise
 
     if output.returncode != 0:
-        print(f"Error getting artifacts for run {artifact_download_url}: {output.stderr}")
-        return []
+        raise Exception(f"Error getting artifacts for run {artifact_download_url}: {output.stderr}")
```
IvoDD: Nit: This if is obsolete with the `check=True` above.

build_tooling/process_pytest_artifacts.py
```diff
@@ -488,68 +558,88 @@ def get_results_lib(arcticdb_library, arcticdb_client_override=None):
...
-            all_runs = [run.stem.split("_")[1] for run in download_dir.glob("*.xml")]
+    if use_github_actions:
+        try:
+            run_obj = create_run_from_github_actions()
+            print(f"Created Run object from GitHub Actions environment: {run_obj}")
+            all_runs = {str(run_obj.run_id): run_obj}
+        except Exception as e:
+            print(f"Error creating Run object from GitHub Actions environment: {e}")
+            print("Falling back to manual parameter mode...")
+            use_github_actions = False
```
IvoDD: We'll only be using `--use-github-actions` inside github actions and would expect all env variables to be correctly populated. If for some reason they are not I think it's better to noisily fail rather than falling back to the slower `use_github_actions = False`. 

Also setting the `use_github_actions = False` here wouldn't cause is to go in the else branch but rather it would just fail later because `all_runs` is not assigned. So I gues re-raising the exception would make more sense.

IvoDD (approved):

PR #2574, opened 2025-08-13
Add missing type hints to Library and fix append return type

alexowens90:
Adds missing type hints for return types of public methods of `Library` class. Also fixes the return type of `append` (previously `Optional[VersionedItem]`, presumably copy-pasted from `NativeVersionStore`, but it is only `None` if data is being staged, which isn't possible in the V2 version of `append`)

### DIFF
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -780,7 +780,7 @@ class Library:
         self._nvs._normalizer.df.set_skip_df_consolidation()
         self._dev_tools = DevTools(nvs)
 
-    def __repr__(self):
+    def __repr__(self) ->str:
         return "Library(%s, path=%s, storage=%s)" % (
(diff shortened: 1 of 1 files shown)


python/arcticdb/version_store/library.py
```diff
@@ -1157,7 +1157,7 @@ def append(
         metadata: Any = None,
         prune_previous_versions: bool = False,
         validate_index: bool = True,
-    ) -> Optional[VersionedItem]:
+    ) -> VersionedItem:
```
G-D-Petrov, line 1160: the append in `NativeVersionStore` returns `Optional[VersionedItem]` but the `None `can be returned only if it is called with `incomplete=True`.
This method here in Library cannot be called with `incomplete=True` so that is why we are dropping the optional, right?

G-D-Petrov (approved):

PR #2576, opened 2025-08-13
Use std::unique_lock instead of OnExit with std::mutex

alexowens90:
We need something like `OnExit` to unlock our `StorageLock` with RAII, as it does not meet the C++ `Lockable` requirements, but there is no need to use it with a `std::mutex`

### DIFF
--- a/cpp/arcticdb/util/storage_lock.hpp
+++ b/cpp/arcticdb/util/storage_lock.hpp
@@ -38,31 +38,6 @@ SegmentInMemory lock_segment(const StreamId &name, uint64_t timestamp) {
 
 } // namespace
 
-struct OnExit {
-    folly::Func func_;
-    bool released_ = false;
-
-    ARCTICDB_NO_MOVE_OR_COPY(OnExit);
-
-    explicit OnExit(folly::Func&& func) :
-        func_(std::move(func)) {}
-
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -109,21 +84,12 @@ class StorageLock {
...
-        OnExit x{[that=this] () {
-            that->mutex_.unlock();
-        }};
-
-        const bool try_lock = try_acquire_lock(store);
-        if (try_lock) {
-            x.release();
-        }
-
-        return try_lock;
```
ognyanstoimenov, line 126: I see there is a slight change in logic introduced. Previously the `mutex_` wasn't unlocked if the storage lock was successfully taken with `try_acquire_lock`. Now it is unlocked regardless. Is that desired? If we are to mimic an std::mutex with `StorageLock::try_lock` maybe we shouldn't unlock the internal mutex

alexowens90: Ah you are right, this change is just busted

alexowens90:
Does not maintain previous behaviour, see https://github.com/man-group/ArcticDB/pull/2576/files#r2273321989

PR #2582, opened 2025-08-14
Threads already dead after revert

ognyanstoimenov:
This reapplies the reverted PR #2544 with two changes that previously caused flaky tests:
1. Increase test timeout from 5 to 10 seconds. Previously some successful tests took up to 4+ seconds so it is reasonable to increase the limit.
2. Increase the timeout of windows api call. Previously the function immediatly returned which caused some rare cases where the function was wrongly returning that the thread is not signaled and thus causing the tests to fail.

I've tested both before the change and after.
Before:
* Out of 100 repetitions of the test ~ 4 were failing

After
* Out of 200 repetitions - no failures

For easier review here is the diff with the reverted PR: https://github.com/man-group/ArcticDB/compare/threads_already_dead..threads_already_dead_after_revert

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -229,8 +240,8 @@ class TaskScheduler {
 
     void stop() {
         ARCTICDB_DEBUG(log::schedule(), "Stopping task scheduler");
-        cpu_exec_.stop();
         io_exec_.stop();
+        cpu_exec_.stop();
```
phoebusm, line 244: I wonder the implication of swapping the order

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -101,6 +101,18 @@ struct SchedulerWrapper : public SchedulerType {
     void ensure_active_threads() {
         SchedulerType::ensureActiveThreads();
     }
+
+    void stop_orphaned_threads() {
+#ifdef _WIN32
+        const auto to_remove_copy = SchedulerType::threadList_.get();
+        for (const auto& thread : to_remove_copy) {
+            const bool is_signaled = WaitForSingleObject(thread->handle.native_handle(), 5000) == WAIT_OBJECT_0;
```
alexowens90, line 109: I guess most of these calls return immediately, so not required, but looks like `WaitForMultipleObjects ` is designed for this case

cpp/arcticdb/async/task_scheduler.hpp
```diff
@@ -101,6 +101,18 @@ struct SchedulerWrapper : public SchedulerType {
...
+            if (is_signaled) {
+                SchedulerType::threadList_.remove(thread);
```
alexowens90, line 111: Is there any way for the thread to leave the threadlist since `to_remove_copy` was created? If so, what would happen here?

ognyanstoimenov: I don't think it is possible as in this case all other threads should already be dead.

PR #2583, opened 2025-08-14
Resilient start of simulated storages

grusev:

Problem shoed here: https://github.com/man-group/ArcticDB/actions/runs/16915115493/job/47928193692

On runners time start seems to be varying as found in a comment for moto - it was 240 secs. For azurite and mongo it was default 20. 

The fix now makes start timeout equal and adds one retry. This code will not harm execution on laptops, but will make startup of  tests for simulated and local storages on runners resilient.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/storage_fixtures/mongo.py
+++ b/python/arcticdb/storage_fixtures/mongo.py
@@ -123,9 +123,17 @@ class ManagedMongoDBServer(StorageFixtureFactory):
 
     def _safe_enter(self):
         cmd = [self._executable, "--port", str(self._port), "--dbpath", self._data_dir]
-        self._p = GracefulProcessUtils.start(cmd)
-        self.mongo_uri = f"mongodb://localhost:{self._port}"
-        wait_for_server_to_come_up(f"http://localhost:{self._port}", "mongod", self._p)
+        for i in range(2): # retry in case of connection problems
+                try:
+                    self._p = GracefulProcessUtils.start(cmd)
+                    self.mongo_uri = f"mongodb://localhost:{self._port}"
+                    # There is a problem with the performance of the socket module in the MacOS 15 GH runners - https://github.com/actions/runner-images/issues/12162
+                    # Due to this, we need to wait for the server to come up for a longer time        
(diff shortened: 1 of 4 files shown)


G-D-Petrov (approved):

python/arcticdb/storage_fixtures/mongo.py
```diff
@@ -123,9 +123,17 @@ def __init__(self, data_dir: Optional[str] = None, port=0, executable="mongod"):
 
     def _safe_enter(self):
         cmd = [self._executable, "--port", str(self._port), "--dbpath", self._data_dir]
-        self._p = GracefulProcessUtils.start(cmd)
-        self.mongo_uri = f"mongodb://localhost:{self._port}"
-        wait_for_server_to_come_up(f"http://localhost:{self._port}", "mongod", self._p)
+        for i in range(2): # retry in case of connection problems
```
G-D-Petrov: nit: Might be better to move the retry logic inside `wait_for_server_to_come_up`
This will improve the code reuse a bit.

grusev: As we do retry the start has to be also part of it. Thus wait_for_server_to_come_up is not best. A new method at GracefulProcessUtils will be created for that 
As s3 method for start is slighly different it will remain the old implementation

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -812,19 +812,24 @@ def _start_server(self):
...
-                self.key_file if self.http_protocol == "https" else None,
-                self.cert_file if self.http_protocol == "https" else None,
-            ),
-        )
-        self._p.start()
-        # There is a problem with the performance of the socket module in the MacOS 15 GH runners - https://github.com/actions/runner-images/issues/12162
-        # Due to this, we need to wait for the server to come up for a longer time
-        wait_for_server_to_come_up(self.endpoint, "moto", self._p, timeout=240)
-
+        for i in range(2): # retry in case of connection problems
```
G-D-Petrov: Looking back at this, there is already retry logic in `def _safe_enter` function.
So I don't think that the retry here is needed.

PR #2585, opened 2025-08-15
Fix workflow problems for Persistеnce test executions and ASV AWS S3 tests execution

grusev:

ASV tests problems with AWS S3: the code wrongly assumed aws bucket from azure env vars. This works locally as I did had Azure vars. But the ASW workflow does not supply those vars. This is fixed and run is initiated from Actions: https://github.com/man-group/ArcticDB/actions/runs/16987111363

Persistance tests are executed from a separate workflow. This workflow runc cleanup seed and verify tasks. The workflow does not install any arcticdb test libraries. Further it works with old arctic also, therefore it is much more closer to installation tests and must not include any new stuff from master but stuff that has been there in 3.0.0 only. This again was invisible requirement during development of Azure. Thus all worked fine
Run from Actions with AWS_S3: https://github.com/man-group/ArcticDB/actions/runs/16988249102

Both could only be revealed if executed against latest changed sources from Azure branc. As those workflows are not part pf the PR they were forgotten ....

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/arrow/arrow_handlers.cpp
+++ b/cpp/arcticdb/arrow/arrow_handlers.cpp
@@ -63,41 +63,41 @@ void ArrowStringHandler::convert_type(
     struct DictEntry {
         int32_t offset_buffer_pos_;
         int64_t string_buffer_pos_;
+        std::string_view strv;
     };
     std::vector<StringPool::offset_t> unique_offsets_in_order;
     ankerl::unordered_dense::map<StringPool::offset_t, DictEntry> unique_offsets;
+    // Trade some memory for more performance
+    // TODO: Use unique count column stat in V2 encoding
+    unique_offsets_in_order.reserve(source_column.row_count());
+    unique_offsets.reserve(source_column.row_count());
     int64_t bytes = 0;
+    int32_t unique_offset_count = 0;
     auto dest_ptr = reinterpret_cast<int32_t*>(dest_column.bytes_at(mapping.offset_bytes_, source_column.row_count() * sizeof(int32_t)));
 
     // First go through the source column once to compute the size of offset and string buffers.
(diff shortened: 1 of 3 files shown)


.github/workflows/persistent_storage.yml
```diff
@@ -102,13 +105,14 @@ jobs:
 
       - name: Install latest release
         run: |
-          pip install pytest arcticdb "protobuf<5"
+          pip install --force-reinstall pytest arcticdb "protobuf<5"
```
G-D-Petrov: Why do we need force-reinstall here? This will be a fresh runner everytime

grusev: will remove it - leftover of initial attempt to add test libs which did not worked :-(

python/tests/util/storage_test.py
```diff
@@ -96,6 +96,19 @@ def real_gcp_credentials(shared_path: bool = True):
...
 
+def real_azure_credentials(shared_path: bool = True):
+    if shared_path:
+        path_prefix = os.getenv("ARCTICDB_PERSISTENT_STORAGE_SHARED_PATH_PREFIX")
+    else:
+        path_prefix = os.getenv("ARCTICDB_PERSISTENT_STORAGE_UNIQUE_PATH_PREFIX", "")
+    constr=os.getenv("ARCTICDB_REAL_AZURE_CONNECTION_STRING"),
+    container=os.getenv("ARCTICDB_REAL_AZURE_CONTAINER"), 
+
+    clear = str(os.getenv("ARCTICDB_REAL_GCP_CLEAR")).lower() in ("true", "1")
```
IvoDD: AZURE, not GCP?

PR #2586, opened 2025-08-15
Arrow fix string cols with nones and nans

alexowens90:
Allow reading into Arrow string columns containing Nones and NaNs

### DIFF
--- a/cpp/arcticdb/arrow/arrow_handlers.cpp
+++ b/cpp/arcticdb/arrow/arrow_handlers.cpp
@@ -57,9 +57,6 @@ void ArrowStringHandler::convert_type(
     const std::shared_ptr<StringPool>& string_pool) const {
     using ArcticStringColumnTag = ScalarTagType<DataTypeTag<DataType::UTF_DYNAMIC64>>;
     auto input_data = source_column.data();
-    auto pos = input_data.cbegin<ArcticStringColumnTag>();
-    const auto end = input_data.cend<ArcticStringColumnTag>();
-
     struct DictEntry {
         int32_t offset_buffer_pos_;
         int64_t string_buffer_pos_;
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -75,29 +72,52 @@ void ArrowStringHandler::convert_type(
...
-    auto string_ptr = reinterpret_cast<char*>(string_buffer.data());
-    for (auto unique_offset: unique_offsets_in_order) {
-        const auto& entry = unique_offsets[unique_offset];
-        *offsets_ptr++ = entry.string_buffer_pos_;
-        memcpy(string_ptr, entry.strv.data(), entry.strv.size());
-        string_ptr += entry.strv.size();
+    inserter.flush();
+    // At this point bitset has ones where the source column contained None or NaN
+    // Inverting and shrinking to the source column size it then makes a sparse map for the input data
+    bitset.invert();
```
IvoDD, line 97: Is inverting an empty bitset cheaper than setting each bit? I imagine this was done as a speed improvement for the common case where no bits are set.

It would be less confusing if we were explicitly setting the bits we need.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -75,29 +72,52 @@ void ArrowStringHandler::convert_type(
...
+    // TODO: row_count() here won't be right when the original data was sparse, but we don't support sparse
+    // string columns yet anyway
+    bitset.resize(source_column.row_count());
```
IvoDD, line 100: I am confused how this works, if `resize` fills the end with zeros, shouldn't we `resize` before `invert`ing?

This way if we have the column: `["a", None, "b", "c"]` we would get the bitset `01` then we would invert it to `10` and then we would resize to 4 `1000` which seems wrong.

I'm obviously misunderstanding something as the tests pass

PR #2590, opened 2025-08-18
asv flakiness fixes

grusev:

ASV test for finalize staged data on LMDB could be flaky. The reason was that it is executed exactly once as the staged data can be finalizes once.

With LMDB there is also conviniency that db is staored as files locally. Thus a solution where the database is created once, then backed up and then after each test it is copied back is implemented. This makes possible to have 5 executions which are more than sufficient to get stable avg, as finalization takes approx 8 secs (ie it is realtively slow and likelyhood to have problems vs situation where test is in ms is extremely low over 40 secs time, compared to situation where measurement times is < 1s

-------
Removed alse persistent_query_builder ASV tests it was source for flakiness (due to calculation based on 2 samples only) and it has been completely superseeded by: https://github.com/man-group/ArcticDB/blob/master/python/benchmarks/real_query_builder.py which has much more things to offer and utilizes fully the persistent storage.

NOTE: persistent_query_builder used to run with PRs, while the new one is overnight. This is the only downside but if needed can be included to run along LMDB tests

-------
version_chain - the only thing here that can be done with test is to pring back warm-up time and increase with 30% the number of samples gathered for the calculation

In fact I find this test extremely suitable for Real tests where we use Persistent storage for one time setup of data. As it costs 35 minutes only to setup LMDB this test is perfect candidate to be converted to such test. Moreover lets examine this log:
https://github.com/man-group/ArcticDB/actions/runs/17034135989/job/48282781563?pr=2590

```
| Change   | Before [99c0afa0]    | After [2d4962e0]    |   Ratio | Benchmark (Parameter)                                                                     |
|----------|----------------------|---------------------|---------|-------------------------------------------------------------------------------------------|
| +        | 98.2±4ms             | 142±20ms            |    1.45 | persistent_query_builder.PersistentQueryBuilderFunctions.time_query_adv_query_2(1000000)  |
| +        | 5.33±0.09ms          | 6.84±0.08ms         |    1.28 | version_chain.IterateVersionChain.time_read_alternating(25000, 'default', 0.0)            |
| +        | 5.43±0.08ms          | 6.51±0.3ms          |    1.2  | version_chain.IterateVersionChain.time_read_alternating(25000, 'forever', 0.0)            |
| +        | 5.42±0.09ms          | 6.41±0.08ms         |    1.18 | version_chain.IterateVersionChain.time_read_from_epoch(25000, 'default', 0.0)             |
| -        | 701±60ms             | 575±10ms            |    0.82 | persistent_query_builder.PersistentQueryBuilderFunctions.time_query_adv_query_2(10000000) |
```

It is seen that measurment for version_chain that are flaky are approx 6 ms each. Thus even with increased 30% number of samples from 10 to 13 we are going to experience further flakiness as both LMDB and Arcticdb are on the same machine and where times are in ms you can not rely that LMDB (or other related with it process) will not consume more proc time. Thus separating the storage and application under test is the right thing to do.

RESULTS FROM TESTS:

Run from ASV workflow without this PR: https://github.com/man-group/ArcticDB/actions/runs/17064710600/job/48379213172?pr=2594
2h 15m 29s

Run from this modified workflow: https://github.com/man-group/ArcticDB/actions/runs/17071349517/job/48401001814?pr=2590
2h 18m 11s

Just 3 minutes overall increase from 135 to 138 mins which is 2%

Additional run from Actions:
https://github.com/man-group/ArcticDB/actions/runs/17096787292

================================
Locally measured times for tests:

```
2025-08-21 11:01:16,531 - benchmarks.finalize_staged_data - INFO - SETUP_CACHE TIME: 25.863695859909058
Execution time: 81 seconds
--------------------
2025-08-21 11:03:59,690 - benchmarks.list_functions - INFO - SETUP_CACHE TIME: 58.31131935119629 
Execution time: 75 seconds
--------------------
2025-08-21 11:07:53,018 - benchmarks.list_snapshots - INFO - SETUP_CACHE TIME: 97.5234956741333
Execution time: 271 seconds
--------------------
2025-08-21 11:12:26,384 - benchmarks.local_query_builder - INFO - SETUP_CACHE TIME: 7.684977769851685
Execution time: 202 seconds
--------------------
2025-08-21 11:16:31,375 - benchmarks.bi_benchmarks - INFO - SETUP_CACHE TIME: 16.876145601272583
Execution time: 152 seconds
--------------------
2025-08-21 11:19:43,333 - benchmarks.basic_functions - INFO - SETUP_CACHE TIME: 9.34851861000061
BasicFunctions
Execution time: 284 seconds
--------------------
2025-08-21 11:31:22,862 - benchmarks.basic_functions - INFO - SETUP_CACHE TIME: 80.07152652740479
BatchBasicFunctions
Execution time: 1227 seconds
--------------------
2025-08-21 12:03:15,853 - benchmarks.basic_functions - INFO - SETUP_CACHE TIME: 11.19944167137146
ModificationFunctions
Execution time: 481 seconds
--------------------
2025-08-21 12:15:20,464 - benchmarks.arrow - INFO - SETUP_CACHE TIME: 17.815497636795044                                                                                                              arrow.ArrowReadNumeric
2025-08-21 13:19:27,823 - benchmarks.arrow - INFO - SETUP_CACHE TIME: 9.17694902420044
arrow.ArrowReadStrings
Execution time: 4075 seconds
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/benchmarks/finalize_staged_data.py
```diff
@@ -23,61 +24,70 @@ class FinalizeStagedData:
...
-        print(">>> Library:", self.lib)
-        print(">>> Symbol:", self.symbol)
+        self.logger.info(f"LIBRARY: {self.lib}")
+        self.logger.info(f"Created Symbol: {self.symbol}")
         self.lib.finalize_staged_data(self.symbol, mode=StagedDataFinalizeMethod.WRITE)
 
     def teardown(self, cache: CachedDFGenerator, param: int):
-        self.ac.delete_library(self.lib_name)
+        # After the modification functions clean up the changes by replacing the modified ARCTIC_DIR with the original ARCTIC_DIR_ORIGINAL
+        # TODO: We can use dirs_exist_ok=True on copytree instead of removing first if we run with python version >=3.8
```
alexowens90: We do only run with Python version >= 3.8?

python/benchmarks/finalize_staged_data.py
```diff
@@ -23,61 +24,70 @@ class FinalizeStagedData:
...
+            df = cachedDF.generate_dataframe_timestamp_indexed(200, 0, cachedDF.TIME_UNIT)
+            list_of_chunks = [10000] * param
+
+            lib.write(symbol, data=df, prune_previous_versions=True)
+            self.logger.info(f"LIBRARY: {lib}")
+            self.logger.info(f"Created Symbol: {symbol}")
+            stage_chunks(lib, symbol, cachedDF, INITIAL_TIMESTAMP, list_of_chunks)
+        # We use the fact that we're running on LMDB to store a copy of the initial arctic directory.
+        # Then on each teardown we restore the initial state by overwriting the modified with the original.
+        copytree(FinalizeStagedData.ARCTIC_DIR, FinalizeStagedData.ARCTIC_DIR_ORIGINAL)
```
alexowens90, line 70: ASV doesn't seem to have a `teardown_cache` concept, so after the final run of the benchmark, a copy of the data will be left in `ARCTIC_DIR_ORIGINAL`. I don't think it'll be easy to fix this, so can we make `ARCTIC_DIR_ORIGINAL` a temporary directory so that at least on developer machines it'll get cleaned up on reboots? You can use the [tempfile](https://docs.python.org/3/library/tempfile.html) module for this

alexowens90 (changes requested):

python/benchmarks/finalize_staged_data.py
grusev: No need to change anything, the constrcted lmdb file is already in tmp folder, added log to show that:

```
            OUTPUT -------->
             2025-08-21 10:21:35,186 - benchmarks.finalize_staged_data - INFO - Arctic(config=LMDB(path=/tmp/tmp4_3x97xe/staged_data))
```

python/benchmarks/finalize_staged_data.py
grusev: this was copy and paste directly from another test. Will change it there also

python/benchmarks/finalize_staged_data.py
G-D-Petrov: I think that Alex meant that we should use something like [this](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) to make sure that the directory is cleaned automatically, not just that it is put in the tmp dir.
We should use that and perhaps also it is for the rest of the ASV benchmarks.

grusev: That is not a job for ASV, that is a job for the invoker. ASV runs tests and setup and different threads. Thus Deleting shared files from one thread before others have exited will lead to unpredictable results. Since there is no ASV method exposed it is unsafe to think we could reliably do that.

Thus this job is always on the caller, not on the test. 

Here is possible solution:

```
start=$(date +%s)
asv run --python=same -v --show-stderr --bench $1
if [ $? -ne 0 ]; then
    echo "⚠️ WARNING: ASV failed with exit code $?"
fi
end=$(date +%s)
rm -rf /tmp/* 
echo "Execution time: $((end - start)) seconds"
```

alexowens90: @G-D-Petrov is right, I'm not talking about the lmdb library, I'm talking about the copy taken. If we put this in a directory created by `tempfile` then if the benchmarks exit cleanly then it will get cleaned up for us, and in the worst case when it crashes out it will be in `/tmp` and the OS will clean it on reboot

grusev: No need to do requested changes. Upon execution of a test ASV creates temp folder. Then all operations done in the test are executed as this temp folder as current working directory. After the completion of asv this temp folder is cleaned:

I have modified slighly the script to show all that:

```
             OUTPUT -------->
             2025-08-26 14:13:54,128 - benchmarks.finalize_staged_data - INFO - Arctic(config=LMDB(path=/tmp/tmpqkmeks0g/staged_data))
             2025-08-26 14:13:54,867 - benchmarks.finalize_staged_data - INFO - LIBRARY: Library(Arctic(config=LMDB(path=/tmp/tmpqkmeks0g/staged_data)), path=Finalize_Staged_Data_LIB, storage=lmdb_storage)
             2025-08-26 14:13:54,867 - benchmarks.finalize_staged_data - INFO - Created Symbol: symbol100
             2025-08-26 14:13:57,613 - benchmarks.finalize_staged_data - INFO - Destination folder is: staged_data_original
             2025-08-26 14:13:57,613 - benchmarks.finalize_staged_data - INFO - Current directory: /tmp/tmpqkmeks0g
             2025-08-26 14:13:57,613 - benchmarks.finalize_staged_data - INFO - /tmp/tmpqkmeks0g/staged_data
             2025-08-26 14:13:57,613 - benchmarks.finalize_staged_data - INFO - /tmp/tmpqkmeks0g/staged_data_original
             2025-08-26 14:13:57,617 - benchmarks.finalize_staged_data - INFO - SETUP_CACHE TIME: 3.5004539489746094
			 
			 ................................................
			 
			 
			(310) grusev@DESKTOP-4L71HMK:~/source/asv_flaky_fix$ ls /tmp/tmpqkmeks0g
			ls: cannot access '/tmp/tmpqkmeks0g': No such file or directory
			(310) grusev@DESKTOP-4L71HMK:~/source/asv_flaky_fix$ ls /tmp
			ca-certificates.crt                                                             systemd-private-0ecda0c639b94c4c87d59932e8aeed65-systemd-resolved.service-0FYMYi
			snap-private-tmp                                                                systemd-private-0ecda0c639b94c4c87d59932e8aeed65-systemd-timesyncd.service-4S1xfg
			systemd-private-0ecda0c639b94c4c87d59932e8aeed65-systemd-logind.service-sFDjSD
			(310) grusev@DESKTOP-4L71HMK:~/source/asv_flaky_fix$
```

alexowens90: Ah OK understood. No need to leave these logs in then, but worth a comment explaining why the `ARCTIC_DIR_ORIGINAL` directory will not be orphaned by the benchmark

G-D-Petrov (approved):

PR #2594, opened 2025-08-19
Ability to turn on/off storages during local testing

grusev:

During local testing we need to be able to turn on and off storages, depending on our testing needs for particular features we work on. This PR makes this possible allowing each storage type to be enabled/disabled using environment variables. 

Some specifics:
 - By default if real storage is enabled, for example AWS S3 its respective simulated storage will be disabled as there is no need to do the test on simulated storage
 - There is ability to turn on/off all local storages but also control each one of them on their own

This allows creationg of profiles - enabling such profile sets certains storages for tests and others not.

Another thing that is also addressed is the ability to enable disable V2 encoding the same way, for which conftest is also modified

Overall mark.py is now cleaner with variables that enable the check if storage is to be used or not and skip marks that now use those variables.

-------

Local storage switch on of, by default it is on:
ARCTICDB_LOCAL_STORAGE_TESTS_ENABLED

Local storages, to be able to switch off individually:
ARCTICDB_STORAGE_LMDB
ARCTICDB_STORAGE_AZURITE
ARCTICDB_STORAGE_MONGO
ARCTICDB_STORAGE_MEM
ARCTICDB_STORAGE_NFS
ARCTICDB_STORAGE_SIM_S3
ARCTICDB_STORAGE_SIM_GCP

Setting any local storage to 0 will disable it if ARCTICDB_LOCAL_STORAGE_TESTS_ENABLED is 1, And even if ARCTICDB_LOCAL_STORAGE_TESTS_ENABLED is 0 setting a local storage to 1 will enable it.

We have also real storages like:
ARCTICDB_STORAGE_AWS_S3
ARCTICDB_STORAGE_GCP

Which are not grouped with local storages

Also enabling real storage will disable its simulated counterpart.

With encodings:
ARCTICDB_TEST_ENCODING_V1
ARCTICDB_TEST_ENCODING_V2

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


build_tooling/test_profiles/lmdb_only.sh
alexowens90, line 1: Maybe `lmdb_v1_only.sh`?

build_tooling/test_profiles/real_storage_and_local.sh
```diff
@@ -0,0 +1,16 @@
+# In  this case the simulated real storages will be skipped
```
alexowens90: This comment seems wrong?

python/tests/conftest.py
```diff
@@ -610,7 +618,7 @@ def arctic_client(request, encoding_version) -> Arctic:
 def arctic_client_v1(request) -> Arctic:
     filter_out_unwanted_mark(request, request.param)
     storage_fixture: StorageFixture = request.getfixturevalue(request.param + "_storage")
-    ac = storage_fixture.create_arctic(encoding_version=EncodingVersion.V1)
+    ac = storage_fixture.create_arctic(encoding_version=DEFAULT_ENCODING)
```
alexowens90: It was better before, otherwise if we change `DEFAULT_ENCODING` we will suddenly lose test coverage of V1

alexowens90 (changes requested):
How have you confirmed that our test coverage in the CI is not accidentally reduced by this change?

build_tooling/test_profiles/real_storage_and_local.sh
grusev: Better one 
```
# In  this case the simulated real storages will be skipped,
# because even they are enabled, the real storages are also enabled
# and then simulated will not be executed
```

python/tests/conftest.py
grusev: If we change DEFAULT_ENCODING then perhaps V2 will be default encoding hence triggering more need to do more coverage with default encoding. Thus perhaps those arctic_client_v1 etc fixtures would need to be renamed with default_encoding? and perhaps new fixtures for old new would need to be created.

Thus this change was exactly targeted to decide if we want a slight semantic change. Or we actually indeed would need to stick to old.

I do not have any preference here just pointing why this change could be ok

grusev:
> This is not cross platform. It won't work on windows. We can still use environment variables bu we should switch to using `.env` files instead of `.sh`.

Good point! done!

grusev:
> Fix so that V2 encoding is tested on CI on PRs

Fixed by default on GitHub V2 is on (unless switched off), and locally off (unless switched on)

```
# On local environment default is without V2 on github it is always on
TEST_ENCODING_V2 = getenv_strip("ARCTICDB_TEST_ENCODING_V2", "1" if RUNS_ON_GITHUB else "0") == "1"
```

vasil-pashov (approved):

PR #2600, opened 2025-08-20
Add a way to force write a non-incremental version

G-D-Petrov:
#### Reference Issues/PRs
Monday ref ticket: 9825177038
Needed for this PR in enterprise - https://github.com/man-group/arcticdb-enterprise/pull/253

Adds a way to force write a non-increasing version in the version chain.
This is needed for 2 reasons:
- adds a way to simulate/test operation (e.g. replication) on libraries that have had non-increasing versions in the past
- enables such versions to be written in very specific cases (e.g. replication of such previously broken libraries, see examples of intended use in [this PR](https://github.com/man-group/arcticdb-enterprise/pull/253))

This should be used only in testing and by background jobs, so there should be no way for an end user to enable this behavior.
That is way this was added as a parameter to be called by the function manually, instead of as a env variable or something similar.
There should be no change in behavior for end users of ArcticDB.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/version/version_map.hpp
```diff
@@ -557,9 +562,16 @@ class VersionMapImpl {
...
-                    "Trying to write TABLE_INDEX key with a non-increasing version. New version: {}, Last version: {}. This is most likely due to parallel writes to the same symbol, which is not supported.",
+                auto message = fmt::format("Trying to write TABLE_INDEX key with a non-increasing version. New version: {}, Last version: {} This is most likely due to parallel writes to the same symbol, which is not supported.",
                     key.version_id(), original_head ? original_head->version_id() : VariantId{""});
+                if (prevent_non_increasing_version_id) {
+                    storage::check<ErrorCode::E_NON_INCREASING_INDEX_VERSION>(!original_head.has_value() || key.version_id() > original_head->version_id(),
+                        "{}",
+                        message);
+                } else {
+                    // This should happen only in tests and background jobs
+                    log::version().warn("{}", message);
```
IvoDD: Maybe we shouldn't log every time but rather only when we are writing a non incresing index.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -557,9 +562,16 @@ class VersionMapImpl {
             if (key.type() == KeyType::TABLE_INDEX) {
                 util::check(!has_index_key, "There should be at most one index key in the list of keys when trying to write an entry to the store, keys: {}", fmt::format("{}", keys));
                 has_index_key = true;
-                storage::check<ErrorCode::E_NON_INCREASING_INDEX_VERSION>(!original_head.has_value() || key.version_id() > original_head->version_id(),
-                    "Trying to write TABLE_INDEX key with a non-increasing version. New version: {}, Last version: {}. This is most likely due to parallel writes to the same symbol, which is not supported.",
+                auto message = fmt::format("Trying to write TABLE_INDEX key with a non-increasing version. New version: {}, Last version: {} This is most likely due to parallel writes to the same symbol, which is not supported.",
```
IvoDD: I think the `storage::check` is smart enough to not spend time constructing the error message in case there is no error.
I'm slightly worried now that we're constructing it every time we might get a performance penalty for writing a version.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -234,11 +234,14 @@ class VersionMapImpl {
             entry->validate();
     }
 
-    void write_version(std::shared_ptr<Store> store, const AtomKey &key, const std::optional<AtomKey>& previous_key) {
+    void write_version(std::shared_ptr<Store> store,
+        const AtomKey &key,
+        const std::optional<AtomKey>& previous_key,
+        const bool prevent_non_increasing_version_id = true) {
```
IvoDD, line 243: Maybe an inline comment explaining that this should almost never be false?

cpp/arcticdb/version/version_map.hpp
G-D-Petrov: good point, changed it back and also changed the log to be more explicit

cpp/arcticdb/version/version_map.hpp
```diff
@@ -557,9 +565,19 @@ class VersionMapImpl {
...
+                bool is_version_increasing = !original_head.has_value() || key.version_id() > original_head->version_id();
+
+                if (!is_version_increasing) {
+                    if (prevent_non_increasing_version_id) {
+                        storage::raise<ErrorCode::E_NON_INCREASING_INDEX_VERSION>(
+                            "Trying to write TABLE_INDEX key with a non-increasing version. New version: {}, Last version: {} This is most likely due to parallel writes to the same symbol, which is not supported.",
+                            key.version_id(), original_head ? original_head->version_id() : VariantId{""});
+                    } else {
+                        // This should happen only in tests and background jobs
+                        log::version().warn("Force writing TABLE_INDEX key with a non-increasing version. New version: {}, Last version: {}",
```
alexowens90: Could add reading `as_of` version numbers and timestamps may no longer work as expected

PR #2603, opened 2025-08-21
delete_container() operation for Azurite should not be reason tests to fail

grusev:
…o fail

https://github.com/man-group/ArcticDB/actions/runs/17095023477/job/48477853156

the problem in the log is that during __exit__ of fixture delete_container on Azurite throws error.
This should not be a reason to fail execution

This is targeted fix only for Azurite and will not harm Azure

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


G-D-Petrov (approved):
It is a bit strange that we are running into these error, do we know what is causing them?

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -208,14 +208,24 @@ def create_fixture(self) -> AzureContainer:
         return AzureContainer(self)
 
     def cleanup_container(self, b: AzureContainer):
+
+        def test_delete_container_safely(client, timeout):
```
alexowens90:
```suggestion
        def delete_container_safely(client, timeout):
```

python/arcticdb/storage_fixtures/azure.py
```diff
@@ -208,14 +208,24 @@ def create_fixture(self) -> AzureContainer:
...
+            from azure.core.exceptions import ResourceNotFoundError
```
alexowens90: Put with the other imports

alexowens90 (changes requested):

alexowens90 (approved):

PR #2606, opened 2025-08-22
initial version

grusev:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -377,9 +377,13 @@ jobs:
         with:
           node-version: '16'
 
+      - name: Install Azurite
+        uses: ./.github/actions/retry_install-azurite
+        with:
+          max_retries: "3"
+
       - name: Install the wheel and dependencies
         run: |
-          npm install -g azurite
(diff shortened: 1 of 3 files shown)


.github/workflows/build_steps.yml
```diff
@@ -377,9 +377,13 @@ jobs:
...
 
+      - name: Install Azurite
+        uses: ./.github/actions/retry_install_azurite
+        with:
+          max_retries: "3"
+
       - name: Install the wheel and dependencies
         run: |
-          npm install -g azurite
           cmake -P cpp/CMake/CpuCount.cmake | sed 's/^-- //' | tee -a $GITHUB_ENV
```
G-D-Petrov, line 389: Presumably the same could happen with the `pip install ....` below, wouldn't it be better to retry the whole step with something like [this](https://github.com/marketplace/actions/retry-step)

grusev: Pip has already such mechanism and I had fixed it  at installation steps, only 3.8 version had different interface though
```
pip install requests --retries 10 --timeout 60
```

Will give it a try with this plugin seems very nice and better than my awkward solution

G-D-Petrov (approved):

PR #2608, opened 2025-08-22
Better error handling and retries on pip install fails for compat tests

IvoDD:
Raises exceptions in case Venv.init_venv fails
Also adds retries in case e.g. a network error causes the venv init to fail

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/util/venv.py
+++ b/python/arcticdb/util/venv.py
@@ -7,6 +7,7 @@ import tempfile
 import venv
 import pandas as pd
 import numpy as np
+from retrying import retry
 
 
 from typing import Dict, List, Optional, Union
@@ -47,10 +48,12 @@ def run_shell_command(
             stdin=subprocess.DEVNULL,
         )
     if result.returncode != 0:
-        logger.error(
+        error_message = (
             f"Command '{command_string}' failed with return code {result.returncode}\n"
             f"stdout:\n{result.stdout.decode('utf-8')}\nstderr:\n{result.stderr.decode('utf-8')}"
         )
+        logger.error(error_message)
+        raise ErrorInVenv(error_message)
     return result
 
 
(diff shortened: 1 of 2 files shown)


python/arcticdb/util/venv.py
```diff
@@ -74,6 +77,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.tear_down_venv()
 
+    # We have seen some flakyness in pip installing the required dependencies. We retry several times in case
+    # installing fails.
+    @retry(stop_max_attempt_number=5, wait_exponential_multiplier=1000, wait_exponential_max=10000)
```
grusev, line 82: I am not sure if that is enough - I do not think that I saw a fail setting up the environment with pip. Perhaps could be good idea to include here a check after pip installs dependencies that everything is ok like loading arcticdb and doing something here to trigger eventually error so that retry kicks in?

python/arcticdb/util/venv.py
```diff
@@ -91,7 +97,7 @@ def tear_down_venv(self):
 
     def execute_python_file(self, python_path: Union[str, os.PathLike]) -> subprocess.CompletedProcess:
         command = [get_os_specific_venv_python(), python_path]
-        return run_shell_command(command, self.path)
+        run_shell_command(command, self.path)
```
grusev, line 100: if we do not return a value perhaps the typint in signature needs to not do that also?

python/arcticdb/util/venv.py
IvoDD: The setup failures were not seen before because we were silently failing (and pytest would hide the logs of fixture creation because it succeeded). Now we raise an exception if the pip command fails and we will get a session level fixture failure log in case that happens.

grusev (approved):

PR #2609, opened 2025-08-25
Flexible execution and easier analysis with autogenerated Marks

grusev:

Currently our tests lack metadata to help quicly filtered out tests or do in depth analysis of existing tests:

- how many tests we have for each one of storage types we test against
- how many tests we have for each test type/level - unit,  integration
- how many test we have in different cross sections of marks

With this PR several things are introduced:

- dynamic assignments of marks based on physical structure of directories - from there we can obtain which tests are at what level
- dynamic assignment of marks based on fixture usage - each test is marked with specific mark if certain storage fixture is used - lmdb, s3, real_s3 etc

that allows quick queries over our tests like:

``` 
pytest -s --co -m "(lmdb and unit) or (lmdb and integration)"
```

to obtain better understanding where and what tests we have. Those marks will be possible to be used further in test execution selection

Overall this approach adds good benefit for adding with little effort important metadata to the tests. Further that metadata could be enhanced significantly when needed. Like taking info from external sources - like databases. xls. github to dynamically mark tests with certain properties, like flaky, quaranteen etc.

In other words there is no longer need marks to be added and maintained all the time by the team.

Effectivley marks now are combination of all those we add explicitly to the test and those assigned to a test by external resources.

Additionally this PR introduces small additions to Marks namagement:
 - ability to assign many marks to a test on a single line
 - ability to group marks and avoidance of misspelling of marks through Mark and Marks classes

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


build_tooling/list_pytests.sh
vasil-pashov, line 1: I guess it's ok for starters but it'll be better if this was a python script because that way we can run it on Windows

python/tests/unit/arcticdb/version_store/test_filtering_hypothesis.py
```diff
@@ -329,6 +329,7 @@ def test_filter_with_column_slicing(lmdb_version_store_tiny_segment, df):
     ),
     val=numeric_type_strategies(),
 )
+@reproduce_failure('6.72.4', b'AXicY2RmYGZmYGBQYGBnZGJgZISzmBiQACOCZIQKAQARdwBn')
```
vasil-pashov: Why is this part of this PR?

python/tests/conftest.py
```diff
@@ -1505,3 +1507,207 @@ def clear_query_stats():
...
+        evaluate_item(item, Marks.integration.name, Marks.integration.mark)
+        evaluate_item(item, Marks.stress.name, Marks.stress.mark)
+        evaluate_item(item, Marks.hypothesis.name, Marks.hypothesis.mark)
+        evaluate_item(item, Marks.nonreg.name, Marks.integration.mark)
+        evaluate_item(item, Marks.version_store.name, Marks.version_store.mark)
+        evaluate_item(item, Marks.toolbox.name, Marks.toolbox.mark)
+
+        # --- Auto‑mark by fixtures ---
+        fixtures = set(item.fixturenames)
+        ALL_FIXTUR_NAMES.update(fixtures)
```
vasil-pashov: typo
```suggestion
        ALL_FIXTURE_NAMES.update(fixtures)
```

build_tooling/list_pytests.sh
grusev: We primerily work on WSL on Windows, and since this is was more about proof of concept I decided to make it quick. If need araises will make it on python to work on win also, although on win we also have bash which is part of git package

python/tests/unit/arcticdb/version_store/test_filtering_hypothesis.py
grusev: Thanks! leftover will remove it

python/tests/conftest.py
```diff
@@ -1505,3 +1507,207 @@ def clear_query_stats():
...
+        if item.get_closest_marker(mark_name):
+            continue
+            
+        for pattern in patterns:
+            for value in source_values:
+                value_lower = value.lower()
+                if isinstance(pattern, str):
+                    if pattern.lower() in value_lower:
+                        item.add_marker(mark_name)
+                        break
```
IvoDD, line 1644: These breaks will exit only the inner for loop, so we could end up adding the same mark several times (once per pattern).

We can e.g. use `for pattern, value in itertools.product(patterns, source_values)`

python/tests/conftest.py
```diff
@@ -1505,3 +1507,207 @@ def clear_query_stats():
...
+        + ALL_FIXTURES + BASIC_STORE_FIXTURES + BASIC_ARCTIC_FIXTURES + VERSION_STORE_AND_REAL_FIXTURES + OBJECT_STORE_FIXTURES,
+    Marks.dynamic_schema.name: [re.compile(r".*(dynamic_schema|dynamic(?!string)).*", re.I)],
+    Marks.empty_types.name : ["empty_types", "lmdb_version_store_delayed_deletes_v1", "lmdb_version_store_delayed_deletes_v2"],
+    Marks.delayed_deletes.name : ["delayed_deletes"],
+    Marks.use_tombstones.name : ["tombstone", "basic_store_prune_previous", "basic_store_prune_previous"],
+    Marks.sync_passive.name : ["sync_passive"],
+    Marks.bucketize_dynamic.name : ["buckets"],
+    Marks.prune_previous.name : ["prune_previous", "lmdb_version_store_delayed_deletes_v1", "lmdb_version_store_tombstone_and_pruning", 
+                                 "basic_store_delayed_deletes_v1", "basic_store_delayed_deletes_v2"],
+    Marks.segment_size.name : ["segment", "lmdb_version_store_no_symbol_list"],
```
IvoDD: This approach for marks is only a best effort. It is maybe impossible with the current set of tests and fixtures to correctly assign marks to tests in this way.

E.g. see [this test](https://github.com/man-group/ArcticDB/blob/dd746f07084e2b6c3000234b4bc3f203be55584d/python/tests/unit/arcticdb/version_store/test_arrow.py#L176) which sets a segment size without using a specific fixture.

So marks like `segment_size`, `dynamic_schema`, `dynamic_strings` are quite easy to be incorrect. E.g. what if someone uses a parametrization `@pytest.mark.parameterize("dynamic", [True, False])`. How do we know whether it refers to dynamic strings or dynamic schema? What if the parametrization was e.g. `parameterize("schema", "dynamic", "static")`.

IvoDD: I'm happy with this best effort approach for now.

For longer term (as we have discussed offline) we should entirely replace most/all of our fixtrures with a universal library creation method and use pytest fixture only for the parametrization like e.g.

```
@contextlib.contextmanager
def create_version_store(
        storage_type: StorageType = StorageType.LMDB,
        *
        dynamic_schema: Optional[Schema] = None, # dynamic or static schema
        slicing: Optional[Tuple[int, int]] = None, # segment row and column slicing
        encoding_version: Optional[EncodingVersion] = EncodingVersion.V1, #v1 or v2 encoding
        output_format: Optional[OutputFormat] = OutputFormat.PANDAS, # Pandas or Arrow
) -> NativeVersionStore:
    """
    Creates a `NativeVersionStore` according to the configurations and cleans up after use
    """
    store_factory = get_store_factory(storage_type)
    lib = store_factory(dynamic_schema=dynamic_schema, segment_row_size=slicing[0], column_group_size=slicing[1], ...)
    try:
        yield lib
    finally:
        cleanup(lib)
```
With this approach fixtures can be used only for different parametrizations of the input e.g. for dynamic schema or for slicing e.g.
```
@pytest.fixture(params=[True, False])
def dynamic_or_static_schema(request) -> Schema:
    return request.param

@pytest.fixture()
def tiny_segment_slicing():
    return (2, 2)
```

Which could appear in tests like so (e.g. if we want a test which tests both dynamic and static schema over small segment slices):
```
def test_something(dynamic_or_static_schema, tiny_segment_slicing):
    with create_version_store(StorageType.LMDB, dynamic_schema=dynamic_and_static_schema, tiny_segment_slicing) as lib:
         # Use lib here for tests
```

With this approach if every test uses the `create_version_store` method and the method only accepts a new `Schema` type instead of e.g. a plain `bool` writing tests would force people to use the fixture and we can dynamically assign a mark for the test correctly.

python/tests/conftest.py
```diff
@@ -1505,3 +1507,207 @@ def clear_query_stats():
...
+BASIC_STORE_FIXTURES = [re.compile(r"^(basic_store.*|basic_version_.*) ", re.I)]
+OBJECT_STORE_FIXTURES = [re.compile(r"^(object_store.*|object_version_.*)", re.I)]
+LOCAL_OBJECT_STORE_FIXTURES = [re.compile(r"^(local_object_store.*|local_object_version.*)", re.I)]
+VERSION_STORE_AND_REAL_FIXTURES = [re.compile(r"^version_store_and_real*", re.I)]
+
+FIXTURES_TO_MARK = {
+    Marks.lmdb.name: [re.compile(r"^lmdb_.*", re.I)] 
+        + ALL_FIXTURES_AND_LMDB + VERSION_STORE_AND_REAL_FIXTURES + BASIC_STORE_FIXTURES,
+    Marks.mem.name: [re.compile(r"^(mem_.*|in_memory_.*)", re.I)] + ALL_FIXTURES + BASIC_STORE_FIXTURES,
+    Marks.s3.name: [re.compile(r"^s3_.*", re.I)]
```
IvoDD: It's hard to verify all of these regexes but for example we have fixtures like [mock_s3 variants](https://github.com/man-group/ArcticDB/blob/d21054d5a4b95be26301cadc9cde87fbc3fa859a/python/tests/conftest.py#L821-L823) which still test the s3 codepaths but don't use a moto s3 client.

python/tests/util/marking.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Allows combining all or several marks on one line and thus significantly reducing 
+    the the file content and allowing grouping of marks on smaller space
+
+    NOTE: The list should be ordered the same way as you intend to apply marks if 
+    on miltiple lines
+    
+    Usage:
+    ------
+        @pmark([Marks.abc, Marks.cde])
```
IvoDD: Nit: typo, should be `@mark`

python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
```diff
@@ -18,8 +18,10 @@
 from arcticdb_ext import set_config_int
 from arcticdb_ext.storage import KeyType
 from arcticc.pb2.descriptors_pb2 import TypeDescriptor
+from tests.conftest import Marks
 from tests.util.date import DateRange
 from tests.util.mark import MACOS_WHEEL_BUILD
+from tests.util.marking import marks
 
 
 @pytest.mark.storage
```
IvoDD, line 28: There is an enormous amount of marks spread throughout the code which do not use the new mechanism for marking.
E.g. this one.

Also I've recently dealth with pipeline marks which are used in many files untouched by this PR. E.g. [here](https://github.com/man-group/ArcticDB/blob/bf839b584b03e449fe8b3f1409f7bd84e3d0123c/python/tests/hypothesis/arcticdb/test_resample.py#L24) but also many other places.

I suggest creating a new PR to follow up on this one which does all of the mark replacements to the new format.

grusev: Yep a task will be created in monday for that. I did not wanted to make huge PR this time. As start we need to agrre on the principle, and when we agree we will do the necessary changes

python/tests/conftest.py
grusev: Basically regexp do a match like "if fixture name starts with ... ". mock_s3 has been added.:

```
Marks.s3.name: [re.compile(r"^(s3_.*|mock_s3.*)", re.I)] 
```

Having marks separated from the actual test makes it easy to initially plugin such mechanism but makes it hard to maintain in longer term unless a naming convention is not made.

Also, please note that in some of the test suites there are also fixtures. Those fixtures *are not yet* part of this marks model. I will add them once we agree this is good model and we should use it. they will be part of another task

python/tests/conftest.py
grusev: Lets start with mentioning our current situation - and it is we do not have _any_goog knowledge nor method to extract even approximated results. In that aspect this effort is groundbreaking as it offers not exact but a very good approximation. 

In order to have knowledge of what tests we have we have to build a metadata library arround tests. Pytest offers marks as the only option to build something like metadata library. Note that marks _can_ also have parameters which is not utilized here and also standard pytest examples and usage is not tuned to that also.

How to use marks effectivley will be another topic, and as you rightly suggested it is tied to making a better design for fixtures.

For now I will share the following recommendation for marks:
- lets start in small scale 
- to describe only the most important things. (One extremely important thing is test priority which we do not have anywhere for instance)
- description of every possible property is perhaps not needed

Note that whateve we do such mechanisms are bot exact. They are only an approximation of what we have. There is no way to make exact machanism or if there is it will be prohitivly expensive and disruptive.

Marks or any metadata is made usually at some moment in time and at that time it had its truth. In next months that truth might change.

python/tests/conftest.py
grusev: it is possible indeed, but:
 - would be very rare - patterns are made in such a way that they would be very distinctive. In other words a pattern needs to match exactly what we want, the next pattern of the same mark will have to be quite different also or it will not have value, Thus for a mark it is likely to have only one pattern that will match.
 - pytest does not care if you mark something with one and the same mark 2,3,4 or 4000 times. It will be just one mark for it,
 
 Total time for the whole algo is 4 sec, thus even if we do optimization of the code we are not going to gain advantage, plus this is done only once at the very begining and there is a way to turn this mechanism off.
 
 I have addressed this anyway

IvoDD: Yeah agree it doesn't have impact, was more of a confusing part one needs to think about when reading the code.

Probably still think using a single for loop over the catresian product would be easier to reed but doesn't matter too much. Now it's logically what I expect.

IvoDD (approved):

PR #2610, opened 2025-08-25
Parametrize all pipeline tests to run with arrow output format

IvoDD:
#### Reference Issues/PRs
Monday ref: 18168500630 
Adds xfail test for: 18197986461

Consists of two commits that can/should be reviewed seperately:

- First commit does a small change to align arrow normalization to pandas normalization.
Also adds an xfail test for a normalization issue after groupby.

- Second commit does the parametrization of all pipeline tests.
It does so by adding a special atribute `_test_convert_arrow_back_to_pandas` to `NativeVersionStore` used only for testing.
When set this reads as `EXPERIMENTAL_ARROW` but then uses `convert_arrow_to_pandas_for_tests` to convert back to pandas.
This allows running all pipeline tests without the need to change too many pandas specific checks e.g. `assert df.iloc[1]==something` or attaching a `convert_arrow_to_pandas_for_tests` after every single read.

#### Any other comments?

The required changes to parametrize pipeline tests was done with vim macros so should cover all tests. Tests which are skipped (e.g. because they test fixed strings) have deliberate comments explaining why they are not parameterized for arrow.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -77,6 +77,7 @@ from .util.mark import (
     PANDAS_2_COMPAT_TESTS_MARK,
     MACOS,
     ARCTICDB_USING_CONDA,
+    PYARROW_POST_PROCESSING,
 )
 from arcticdb.storage_fixtures.utils import safer_rmtree
 from packaging.version import Version
@@ -1074,7 +1075,10 @@ def lmdb_version_store_arrow(lmdb_version_store_v1) -> NativeVersionStore:
     return store
 
 
-@pytest.fixture(params=list(OutputFormat))
+# TODO: Revert to `params=list(OutputFormat)` once bug https://github.com/apache/arrow/issues/47234 is fixed
+@pytest.fixture(
+    params=[OutputFormat.PANDAS, pytest.param(OutputFormat.EXPERIMENTAL_ARROW, marks=PYARROW_POST_PROCESSING)]
+)
 def any_output_format(request) -> OutputFormat:
     return request.param
 
--- a/python/tests/util/mark.py
+++ b/python/tests/util/mark.py
@@ -229,6 +229,10 @@ SSL_TEST_SUPPORTED = sys.platform == "linux"
 
 FORK_SUPPORTED = pytest.mark.skipif(WINDOWS, reason="Fork not supported on Windows")
 
+PYARROW_POST_PROCESSING = pytest.mark.skipif(
+    WINDOWS, reason="pyarrow 21.0.0 doesn't correctly apply fill_null: https://github.com/apache/arrow/issues/47234"
+)
+
 ## MEMRAY supports linux and macos and python 3.8 and above
 MEMRAY_SUPPORTED = MACOS or LINUX
 MEMRAY_TESTS_MARK = pytest.mark.skipif(


python/arcticdb/version_store/_normalization.py
```diff
@@ -775,7 +775,7 @@ def denormalize(self, item, norm_meta):
             is_empty_table_with_datetime_index = len(item) == 0 and not index_meta.step
             if index_meta.is_physically_stored or is_empty_table_with_datetime_index:
                 pandas_indexes = 1
-                if index_meta.tz:
+                if index_meta.tz and len(item.columns) > 0 and pa.types.is_timestamp(item.columns[0].type):
```
vasil-pashov, line 778: Why was this required? Is it because of writing empty data frames?

IvoDD: It is so that we don't try to apply the timezone to a non timestamp column. 

This can happen if we end up doing a `group_by` on a dataframe which has an index with timezone.
It is still not the correct thing to do as highlighted by [this test](https://github.com/man-group/ArcticDB/pull/2610/files#diff-c15db9d4fe1011ff8e7a6ef29c1c12c8665be13a445a5395dad8506f81c89074R1332).

This change just mimics the behavior of pandas case [here](https://github.com/man-group/ArcticDB/blob/4663a6aac371a694997cac08c14ebbfc546c6ca9/python/arcticdb/version_store/_normalization.py#L450-L452) (i.e. we apply the timezone only if index ends up a timestamp)

vasil-pashov: I think it's worth leaving a comment

python/arcticdb/util/arrow.py
```diff
@@ -17,3 +17,22 @@ def stringify_dictionary_encoded_columns(table, string_type=None):
         if pa.types.is_dictionary(table.column(i).type):
             table = table.set_column(i, name, table.column(name).cast(string_type))
     return table
+
+
+def convert_arrow_to_pandas_for_tests(table):
```
alexowens90, line 22: Assume you moved this here so users can easily verify their tests also still pass?

python/arcticdb/version_store/_store.py
```diff
@@ -2509,6 +2523,8 @@ def _adapt_read_res(self, read_result: ReadResult) -> VersionedItem:
             else:
                 table = pa.Table.from_batches(record_batches)
             data = self._arrow_normalizer.denormalize(table, read_result.norm)
+            if self._test_convert_arrow_back_to_pandas:
+                data = convert_arrow_to_pandas_for_tests(data)
```
alexowens90, line 2527: Did you consider making these changes in the `ArrowTableNormalizer`? Not sure if there are any advantages but seems like a logical place for them

python/arcticdb/util/arrow.py
IvoDD: Main reason was that using the `convert_arrow_to_pandas_for_tests` in `NativeVersionStore` or `ArrowTableNormalizer` introduced circular dependencies with the testing utils.

I aggree it is still probably useful for updating user tests if e.g. codex decide to change default output format without rewriting a lot of tests.

python/arcticdb/version_store/_store.py
IvoDD: The `_test_convert_arrow_back_to_pandas` affects 2 different aspects:
- Conversion from arrow to pandas (which I agree can make sense to move to normalizer)
- Always enabling `dynamic_strings=True` on write [here](https://github.com/man-group/ArcticDB/pull/2610/files#diff-fd55a7274e8a49f65e2f62074974d20809bde05f639f4d55fbe7f8430f772358R773-R776), so that we don't have to special case all tests using `dynamic_strings=False` (most of which do so through the `lmdb_dynamic_schema_v1` fixture which uses fixed width strings by default)

So since this flag already needs to live within the `NativeVersionStore` I think it makes sense to isolate its effect to that scope, instead of part of changes happening in `ArrowTableNormalizer` and part in `NativeVersionStore`.

However I don't feel strongly, happy to also pass as extra kwarg to the denormalizer if you think it makes it more readable

alexowens90: Ah yeh, hadn't thought about the write path aspect

PR #2612, opened 2025-08-25
Create mocked tests for merging sorted timeseries

vasil-pashov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_merge.py
+++ b/python/tests/unit/arcticdb/version_store/test_merge.py
@@ -58,6 +58,132 @@ def raise_wrapper(exception, message=None):
 
 class TestMergeTimeseries:
 
+    @pytest.mark.parametrize("strategy", (
+        MergeStrategy(MergeAction.UPDATE, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.INSERT, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT),
+        MergeStrategy(MergeAction.UPDATE, MergeAction.INSERT),
+        MergeStrategy(MergeAction.INSERT, MergeAction.INSERT)
+    ))
+    def test_merge_matched_update_with_metadata(self, lmdb_library, strategy, monkeypatch):
+        lib = lmdb_library
+
+        target = pd.DataFrame({"a": [1, 2, 3], "b": [1.0, 2.0, 3.0]}, index=pd.date_range("2024-01-01", periods=3))
+        write_vit = lib.write("sym", target)
+
+        source = pd.DataFrame(
+            {"a": [1, 20, 3], "b": [1.0, 20.0, 3.0]},
+            index=pd.DatetimeIndex(["2024-01-01 10:00:00", "2024-01-02", "2024-01-04"])
+        )
+
+        metadata = {"meta": "data"}
+        monkeypatch.setattr(lib, "merge", lambda *args, **kwargs: VersionedItem(
+            symbol=write_vit.symbol,
+            library=write_vit.library,
+            data=None,
+            version=write_vit.version + 1,
+            metadata=metadata,
+            host=write_vit.host,
+            timestamp=write_vit.timestamp + 1,
+        ), raising=False)
+
+        merge_vit = lib.merge("sym", source, metadata=metadata, strategy=strategy)
+        assert merge_vit.version == 1
+        assert merge_vit.symbol == write_vit.symbol
+        assert merge_vit.timestamp > write_vit.timestamp
+        assert merge_vit.metadata == metadata
+        assert merge_vit.library == write_vit.library
+        assert merge_vit.host == write_vit.host
+        assert merge_vit.data is None
+
+        monkeypatch.setattr(lib, "read", lambda *args, **kwargs: VersionedItem(
+            symbol=merge_vit.symbol,
+            library=merge_vit.library,
+            data=None, # Not going to check the data, only the metadata
+            version=merge_vit.version,
+            metadata=merge_vit.metadata,
+            host=merge_vit.host,
+            timestamp=merge_vit.timestamp,
+        ))
+        read_vit = lib.read("sym")
+        assert_vit_equals_except_data(merge_vit, read_vit)
+
+        lt = lib._dev_tools.library_tool()
+
+        monkeypatch.setattr(lt, "find_keys_for_symbol", mock_find_keys_for_symbol({KeyType.TABLE_DATA: 2, KeyType.TABLE_INDEX: 2, KeyType.VERSION: 2}))
+
+        assert len(lt.find_keys_for_symbol(KeyType.TABLE_DATA, "sym")) == 2
+        assert len(lt.find_keys_for_symbol(KeyType.TABLE_INDEX, "sym")) == 2
+        assert len(lt.find_keys_for_symbol(KeyType.VERSION, "sym")) == 2
+
+    @pytest.mark.parametrize("metadata", ({"meta": "data"}, None))
+    @pytest.mark.parametrize("strategy", (
+        MergeStrategy(MergeAction.UPDATE, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.INSERT, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT),
+        MergeStrategy(MergeAction.UPDATE, MergeAction.INSERT),
+        MergeStrategy(MergeAction.INSERT, MergeAction.INSERT)
+    ))
+    def test_merge_does_not_write_new_version_with_empty_source(self, lmdb_library, metadata, strategy, monkeypatch):
+        lib = lmdb_library
+        target = pd.DataFrame({"a": [1, 2, 3], "b": [1.0, 2.0, 3.0]}, index=pd.date_range("2024-01-01", periods=3))
+        write_vit = lib.write("sym", target)
+        monkeypatch.setattr(lib.__class__, "merge", lambda *args, **kwargs: write_vit, raising=False)
+        merge_vit = lib.merge("sym", pd.DataFrame(), metadata=metadata, strategy=strategy)
+        assert_vit_equals_except_data(write_vit, merge_vit)
+        assert merge_vit.data is None and write_vit.data is None
+        lt = lib._dev_tools.library_tool()
+        assert len(lt.find_keys_for_symbol(KeyType.TABLE_DATA, "sym")) == 1
+        assert len(lt.find_keys_for_symbol(KeyType.TABLE_INDEX, "sym")) == 1
+        assert len(lt.find_keys_for_symbol(KeyType.VERSION, "sym")) == 1
+
+    @pytest.mark.parametrize("strategy", (
+        None,
+        MergeStrategy(MergeAction.UPDATE, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.INSERT, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT),
+        MergeStrategy(MergeAction.UPDATE, MergeAction.INSERT),
+        MergeStrategy(MergeAction.INSERT, MergeAction.INSERT)
+    ))
+    @pytest.mark.parametrize("source", [
+        pd.DataFrame(
+            {"a": np.array([1, 2, 3], dtype=np.int8),"b": [1.0, 2.0, 3.0]},
+            index=pd.DatetimeIndex([
+                pd.Timestamp("2024-01-01"),
+                pd.Timestamp("2024-01-01 10:00:00"),
+                pd.Timestamp("2024-01-02")
+            ])
+        ),
+        pd.DataFrame(
+            {"c": np.array([1, 2, 3], dtype=np.int8), "b": [1.0, 2.0, 3.0]},
+            index=pd.DatetimeIndex([
+                pd.Timestamp("2024-01-01"),
+                pd.Timestamp("2024-01-01 10:00:00"),
+                pd.Timestamp("2024-01-02")
+            ])
+        ),
+        pd.DataFrame(
+            {"b": [1.0, 2.0, 3.0]},
+            index=pd.DatetimeIndex([
+                pd.Timestamp("2024-01-01"),
+                pd.Timestamp("2024-01-01 10:00:00"),
+                pd.Timestamp("2024-01-02")
+            ])
+        )
+    ])
+    def test_static_schema_merge_throws_when_schemas_differ(self, lmdb_library, strategy, source, monkeypatch):
+        lib = lmdb_library
+        target = pd.DataFrame({"a": [1, 2, 3], "b": [1.0, 2.0, 3.0]}, index=pd.date_range("2024-01-01", periods=3))
+        lib.write("sym", target)
+
+        monkeypatch.setattr(lib.__class__, "merge", raise_wrapper(SchemaException), raising=False)
+
+        with pytest.raises(SchemaException):
+            lib.merge("sym", source, strategy=strategy)
+
+    # ================================================================================================
+    # ==================================== TEST UPDATE ON MATCH ======================================
+    # ================================================================================================
     def test_merge_update(self, lmdb_library, monkeypatch):
         lib = lmdb_library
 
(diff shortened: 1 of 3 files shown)


python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+    # ================================================================================================
+    # ==================================== TEST UPDATE ON MATCH ======================================
+    # ================================================================================================
+    def test_merge_update(self, lmdb_library, monkeypatch):
+        lib = lmdb_library
+
+        target = pd.DataFrame({"a": [1, 2, 3], "b": [1.0, 2.0, 3.0]}, index=pd.date_range("2024-01-01", periods=3))
+        write_vit = lib.write("sym", target)
+
+        source = pd.DataFrame(
```
maxim-morozov, line 233: Can we have completely different values between source and target, so it's easier to see the result

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+        monkeypatch.setattr(lt, "find_keys_for_symbol", mock_find_keys_for_symbol({KeyType.TABLE_DATA: 2, KeyType.TABLE_INDEX: 2, KeyType.VERSION: 2}))
+
+        assert len(lt.find_keys_for_symbol(KeyType.TABLE_DATA, "sym")) == 2
+        assert len(lt.find_keys_for_symbol(KeyType.TABLE_INDEX, "sym")) == 2
+        assert len(lt.find_keys_for_symbol(KeyType.VERSION, "sym")) == 2
+
+    def test_merge_update_writes_new_version_even_if_nothing_is_changed(self, lmdb_library, monkeypatch):
+        # In theory, it's possible to make so that it doesn't write a new version when nothing is matched, but the source
+        # is not empty. This has lots of edge cases and will burden the implementation for almost no gain. If nothing is
+        #  changed, we'll keep the same index and data keys and just write a new version key which is cheap.
```
maxim-morozov, line 298: I am not sure it makes sense to have new version if nothing changed. How difficult it is to keep the version as it is when nothing changed? Also what would be the case for empty dataframe?

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+        source = pd.DataFrame(
+            {"a": [1, 2, 30, 40], "b": [10.0, 20.0, 30.0, 40.0]},
+            index=pd.DatetimeIndex([
+                "2024-01-01", # Matches index and column
+                "2024-01-02 01:00:00", # Matches column, but not index
+                "2024-01-03", # Matches index, but not column
+                "2024-01-04" # Does not match either
+        ]))
+        monkeypatch.setattr(lib.__class__, "merge", lambda *args, **kwargs: None, raising=False)
+        lib.merge("sym", source, on=["a"], strategy=MergeStrategy(not_matched=MergeAction.DO_NOTHING))
```
maxim-morozov: Something we should discuss, but if on is specified, I think we should match only on columns in on, not the index

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+        lib.write("sym", target)
+
+        source = pd.DataFrame({
+            "a": [10, 20, 30, 40, 50],
+            "b": ["a", "b", "c", "d", "e"],
+            "c": ["A", "D", "A", "B", "C"],
+            "d": [10.1, 50.5, 30.3, 40.4, 70.7],
+            "e": [100, 500, 600, 400, 800]
+        }, index=pd.DatetimeIndex([
+            "2024-01-01",      # index + b + d + e match (should update)
```
maxim-morozov: I think we should find the way to specify that merge is on index as well.

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+            index=pd.DatetimeIndex([pd.Timestamp("2024-01-01"), pd.Timestamp("2024-01-01"), pd.Timestamp("2024-01-02")])
+        )
+        lib.write("sym", target)
+        monkeypatch.setattr(lib.__class__, "merge", lambda *args, **kwargs: None, raising=False)
+        source = pd.DataFrame(
+            {"a": [5], "b": [20.0]},
+            index=pd.DatetimeIndex([pd.Timestamp("2024-01-01")])
+        )
+        lib.merge("sym", source, strategy=MergeStrategy(not_matched=MergeAction.DO_NOTHING))
+        expected = pd.DataFrame(
```
maxim-morozov, line 489: I am not sure about this case. What happens if we have multiple rows at source? It looks like it will throw from the case below. But why do we need merge at all then? It's not very intuitive. Should we discuss it?

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+            symbol=write_vit.symbol,
+            library=write_vit.library,
+            data=None,
+            version=write_vit.version + 1,
+            metadata=write_vit.metadata,
+            host=write_vit.host,
+            timestamp=write_vit.timestamp + 1,
+        ), raising=False)
+        source = pd.DataFrame({"a": [10, 20, 30], "b": [10.0, 20.0, 30.0]}, index=pd.date_range("2024-01-01", periods=3))
+        merge_vit = lib.merge("sym", source, strategy=MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT))
```
maxim-morozov, line 743: Yeah, something for us to decide, but my preference is not to advance version if nothing changed

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+        source = pd.DataFrame(
+            {"a": [1, 2, 3], "b": [1.0, 2.0, 3.0]},
+            index=pd.DatetimeIndex([
+                pd.Timestamp("2024-01-01"), # Matches first row of target
+                pd.Timestamp("2024-01-01"), # Also matches first row of target
+                pd.Timestamp("2024-01-04")
+            ]))
+
+        monkeypatch.setattr(lib.__class__, "merge", lambda *args, **kwargs: None, raising=False)
+        lib.merge("sym", source, strategy=MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT))
```
maxim-morozov: What would be the output? Only the last row inserted?

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+    # timestamp   ticker   ask   bid ....
+    # 01-01-2024  AAPL     100   99
+    # 01-01-2024  MSFT     200   199
+    # 01-02-2024  AAPL     101   100
+
+    # Then some provider might have outage on 01-02-2024 and provide update. Then the user can call
+    # lib.merge(sym, source, strategy=MergeStrategy(MergeAction.INSERT, MergeAction.DO_NOTHING)) to insert the data.
+
+    # I think the example is flawed:
+    # 1) If a table contains data from multiple providers, then there would be multiple rows for the same stock at the
```
maxim-morozov: Yes, I don't think match insert makes sense. The case I was talking about was either Non-Match Insert or Match-Update.

Think about stocks. Each exchange has set of stocks and if you are trying to build ticks data from all exchanges in arctic, there will be no overlap. Basically of you build data from LSE and NYSE and LSE has an outage, you will not get tick data from LSE, but you will continue getting NYSE tick data. You probably will continue doing appends, but when LSE outage is addressed, LSE might issue all the missing ticks. You basically need a way to insert those ticks between all NYSE ticks already there.

In this case the call probably will look like:

```
lib.merge(sym, source, strategy=MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT))
```

Or another very common case is when LSE had an outage and sent wrong prices (ticks) for their traded stocks or some of the stocks were missing. In this case, it will be:

```
lib.merge(sym, source, strategy=MergeStrategy(MergeAction.UPDATE, MergeAction.INSERT))
```

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+from typing import NamedTuple
+
+class MergeAction(Enum):
+    UPDATE = 1
+    INSERT = 2,
+    DO_NOTHING = 3
+
+class MergeStrategy(NamedTuple):
+    matched: MergeAction = MergeAction.UPDATE
+    not_matched: MergeAction = MergeAction.INSERT
```
alexowens90: Could this be `not_matched_by_target`? We might never implement `not_matched_by_source`, but given the SQL inspiration we should make this clear

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+#     INSERT_NOT_MATCHED = 8
+#
+# this can be used like
+# lib.merge("sym", source, strategy=MergeAction.DO_NOTHING_MATCHED | MergeAction.INSERT_NOT_MATCHED)
+# lib.merge("sym", source, strategy=MergeAction.INSERT_MATCHED | MergeAction.INSERT_NOT_MATCHED)
+#
+# but it's not shorter than
+# lib.merge("sym", source, strategy=MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT))
+# lib.merge("sym", source, strategy=MergeStrategy(matched=MergeAction.DO_NOTHING, not_matched=MergeAction.INSERT))
+# lib.merge("sym", source, strategy=MergeStrategy(MergeAction.INSERT, MergeAction.INSERT))
```
alexowens90: I dislike the flags OR approach, it is more C than Pythonic

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+    # ================================= APPLIES TO ALL STRATEGIES ====================================
+    # ================================================================================================
+    @pytest.mark.parametrize("strategy", (
+        MergeStrategy(MergeAction.UPDATE, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.INSERT, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT),
+        MergeStrategy(MergeAction.UPDATE, MergeAction.INSERT),
+        MergeStrategy(MergeAction.INSERT, MergeAction.INSERT)
+    ))
+    def test_merge_matched_update_with_metadata(self, lmdb_library, strategy, monkeypatch):
```
alexowens90, line 71: The metadata matching the previous version after a modification operation is a departure from `append/update`. I think it would be nice to have as an option for all of them, but in the first implementation I think consistency would be better, so the metadata should need to be explicitly provided

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+
+    @pytest.mark.parametrize("metadata", ({"meta": "data"}, None))
+    @pytest.mark.parametrize("strategy", (
+        MergeStrategy(MergeAction.UPDATE, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.INSERT, MergeAction.DO_NOTHING),
+        MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT),
+        MergeStrategy(MergeAction.UPDATE, MergeAction.INSERT),
+        MergeStrategy(MergeAction.INSERT, MergeAction.INSERT)
+    ))
+    def test_merge_does_not_write_new_version_with_empty_source(self, lmdb_library, metadata, strategy, monkeypatch):
```
alexowens90, line 130: I guess there are also merge strategies that can be a no-op (e.g. only updating matching rows where no rows match) that could also not create new versions, but not necessary for first impl I would say

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+            symbol=write_vit.symbol,
+            library=write_vit.library,
+            data=None,
+            version=1,
+            metadata=write_vit.metadata,
+            host=write_vit.host,
+            timestamp=write_vit.timestamp + 1,
+        ), raising=False)
+
+        merge_vit = lib.merge("sym", source, strategy=MergeStrategy(not_matched=MergeAction.DO_NOTHING))
```
alexowens90: Some syntactic sugar of being able to specify enum values with case-insensitive strings usually goes down well to avoid an additional import

python/tests/unit/arcticdb/version_store/test_merge.py
alexowens90: This is more subtle than it might appear. Pointing the new version key at the old index key would be a new pattern that we do not use elsewhere, and could have some knock on effects in replication, which cares about version numbers of index/data keys.
I don't think we should worry about this too much on day 1 and just create a new version that happens to be identical to the previous version, and circle back round to optimise this later.

python/tests/unit/arcticdb/version_store/test_merge.py
alexowens90: I disagree, I think the default should always include the index if it is a timeseries, with users having to explicitly opt out.
Additionally, in this case (and in a lot of real data), the index does not have a name. We use the internal name "index", but users have no way of knowing this, and anyway could have another column called "index".

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+                pd.Timestamp("2024-01-01"), # Matches first row of target
+                pd.Timestamp("2024-01-01"), # Also matches first row of target
+                pd.Timestamp("2024-01-04")
+            ]))
+
+        monkeypatch.setattr(lib.__class__, "merge", lambda *args, **kwargs: None, raising=False)
+        lib.merge("sym", source, strategy=MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT))
+
+    # ================================================================================================
+    # ======================================== TEST UPSERT ===========================================
```
alexowens90: nit: This word is overloaded and doesn't mean here what it means in `update`

python/tests/unit/arcticdb/version_store/test_merge.py
alexowens90, line 1: (I've reviewed for API rather than behaviours)
Maybe I've missed it, but I can't see a way to specify NOT merging on a timeseries index?

python/tests/unit/arcticdb/version_store/test_merge.py
```diff
@@ -0,0 +1,1349 @@
...
+
+# class MergeAction(Flags):
+#     UPDATE = 0
+#     DO_NOTHING_MATCHED = 1
+#     DO_NOTHING_NOT_MATCHED = 2
+#     INSERT_MATCHED = 4
+#     INSERT_NOT_MATCHED = 8
+#
+# this can be used like
+# lib.merge("sym", source, strategy=MergeAction.DO_NOTHING_MATCHED | MergeAction.INSERT_NOT_MATCHED)
```
alexowens90: Are fixed on `merge` as the name for this? I think I preferred `merge_into` as being more explicit

python/tests/unit/arcticdb/version_store/test_merge.py
alexowens90, line 1: Another thing I might have missed, what is the behaviour of there is no existing data? (either new symbol or symbol where all versions are deleted)
I guess it will depend on the merge strategy?

vasil-pashov: Good catch I missed that. I'll add new examples.

vasil-pashov: Do you mean explicitly opting out of matching on the index. I thought we don't want this, but it's doable. I'll add mocks for that as well.

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: Sounds reasonable

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: My understanding was that the index should be always picked up when we have ordered datetime index. As Alex pointed above often the index name is blank which makes it more complicated.

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: The idea is the idea is that if the target has duplicate rows (either fully duplicated or duplicated only in the columns we match on) then both of them get updated when they are matched with a row from source. I think it's reasonable to assume that dataframes can have duplicates and that all duplicates get updated.

On the other hand if the source has two rows that both match the target it's not clear which value should we update with. E.g.

Target
idx a
0    1

source
idx a
0    2
0    3

it's not clear whether the target should be (idx=0, a=2) or (idx=0, a=3)

We can jump on a call I'm missing anything.

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: I think it will not be trivial but it's a bit hard to estimate it without having any code to play with. My suggestion is to implement at least one part merge or insert and then circle back to this to estimate how complicated it's going to be.

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: Yes, apparently I forgot to add that part. Will fix.

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: I think merge is more consistent with the other methods e.g. append is not append_to, but I don't have a strong opinion.

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: Yes, I agree that 
```python
lib.merge(sym, source, strategy=MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT))
lib.merge(sym, source, strategy=MergeStrategy(MergeAction.UPDATE, MergeAction.INSERT))
```
will suffice. I'll remove the cases for insert on match.

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: From slack discussion:

If we allow to exclude the index from the query it might create unordered timeseries e.g.
target
index[ns]   col
0           1
1           2
2           3

Source:
index[ns]   col
5           2

lib.merge(sym, source, on=["col"], match_index=False)
index[ns]   col
0           1
5           2
2           3

Decision is to start without this and implement if required by users

python/tests/unit/arcticdb/version_store/test_merge.py
vasil-pashov: Also not matching the index can produce unordered dataframe see: https://github.com/man-group/ArcticDB/pull/2612#discussion_r2376250357

maxim-morozov: I think it's misleading and not very transparent. We can do some thing like this:

``` 
lib.merge("sym", source, on=[ArcticDB.Index, "a"], strategy=MergeStrategy(not_matched=MergeAction.DO_NOTHING))
```

and by default Merge function will look something like:
```
def merge(symbol, source, on=[ArcticDB.Index], ...
```

Basically my point is that when you are using SQL for example, you don't expect that DB behind the scene will do additional columns comparisons on top what you specified in `where` or `on` clauses

python/tests/unit/arcticdb/version_store/test_merge.py
maxim-morozov: That makes sense. I do see more complex scenario where in your case above the matching row will be replaced by 2 rows from the source, but I don't think we need to worry about it now.

PR #2623, opened 2025-08-28
v1 API increase coverage - write operation

grusev:

The goal for this  PR is to improve coverage of v1 API write() method. Initial analysis showed that not all types of data was used in the existing tests and not every combination of parameters is examined. 

The result is a single scenario test that loops through all combination of parameters and executing different write operarions with all supported data types. Unlike existing tests this scenario loops through series of writes and reads on single symbol without wiping out after each one. Thus the test also examines that all write operation types can be performed on a single symbol.

As most of the actual usage of Arctic is actually with sparse data, and such thing was not present in the tests. A number of enhancements for utility functions have been made to be possible to produce sparse arrays and dataframes., which are used primerily in this test but are intended to be used in other future tests also.

2 interesting observations were made during the test, which could be considered issues or known limitations:
 - wrting dataframe with timestamp index which is not sorted is ok, but after reading it back the dataframe is sorted. I could not find documentation that covers this so I could not decide if issue or limitation this is
 - writing empty dataframe with no cols is returned with unexpected index timestamp instead of range 

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -204,6 +204,11 @@ jobs:
           python python/utils/asv_checks.py
         continue-on-error: false
 
+      - name: Disk usage
+        if: always()
+        run: du -m . | sort -n | tail -n 50 || true; df -h
+        continue-on-error: true
+
   # code_coverage:
   #   runs-on: "ubuntu-22.04"
   #   container:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -177,7 +177,7 @@ jobs:
                     - /:/mnt
             windows_matrix:
               - os: windows
-                distro: windows-latest
+                distro: windows-2022
                 cmake_preset_prefix: windows-cl
                 cibw_build_suffix: win_amd64
                 build_dir: C:/cpp_build


python/arcticdb/util/utils.py
```diff
@@ -622,7 +729,10 @@ def add_int_col(self, name: str, dtype: ArcticIntType = np.int64, min: int = Non
     
     def add_float_col(self, name: str, dtype: ArcticFloatType = np.float64, min: float = None, max: float = None,
```
G-D-Petrov: Shouldn't we do this for `add_int_col` as well?

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -560,3 +563,130 @@ def check_incomplete_staged(sym: str, remove_staged: bool = True) -> None:
...
+    pipeline for heterogeneous, sparse, and non-tabular payloads under different
+    configuration flags.
+    """
+
+    max_length: int = 100
+    max_cols: int = 200
+    sym: str = "__qwerty124"
+
+    nvs: NativeVersionStore = version_store_and_real_s3_basic_store_factory(
+        dynamic_schema=True, segment_row_size=(max_length // 2),
```
G-D-Petrov: nit: I'd put all of these as params to the test and test with both `dynamic_schema=True` and `dynamic_schema=False`

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -560,3 +563,130 @@ def check_incomplete_staged(sym: str, remove_staged: bool = True) -> None:
...
+            #   (such as date_range filtering and column selection) will not work on pickled data.            
+            arr_to_write = arr_str.tolist() if WINDOWS else arr_str 
+            nvs.write(sym, arr_to_write, pickle_on_failure=pickle_on_failure, validate_index=validate_index)
+            np.testing.assert_array_equal(arr_to_write, nvs.read(sym).data)
+
+            meta = get_metadata()
+            nvs.write(sym, arr_datetime, pickle_on_failure=pickle_on_failure, validate_index=validate_index,
+                    metadata=meta)
+            np.testing.assert_array_equal(arr_datetime, nvs.read(sym).data)
+            assert 2 <= len(nvs.list_versions()) # Assert it grows
```
G-D-Petrov: nit: this is not great check even if the versions don't grow after the first 2, we will never catch it

python/arcticdb/util/utils.py
grusev: For now only float, str, datetime have real sparse representation. We do not have true None/NA value for int types, we use 0 which is ambiguous. Thus in this case I preferred not to do anythings. 
If needed we can add this feature later for ints - np for me from test code (utility) perspective

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
grusev: The test alternates versions checks with checks when we use prune previous. Thus in this test we assure that primaririly the prune previous version works. Which is enough with making sure versions go back to zero and then they go back to at least 2 (ie they grow). Thus in this test we do not verify indeed that versions can grow to certain number. That for me is outside of the scope of the test. 

In otder to verify the version number correctly we should make just several tests but some of them are expensive - there should be 2 of them with max-version-number minus 1 then increasing version number to max and then increasing it again. Those test are expensive unless we have *persistant* storage with prepared data _or_ we can control max versions as aparameter and set it to low number like 20. That would be the proper test to verify versioning limits, therefore I did not made more than just the essential to verify the prune previous works correctly with versions (several times applied on one symbol)

PR #2632, opened 2025-09-03
Fix range intersection calculation in query.cpp

G-D-Petrov:
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/pipeline/query.cpp
+++ b/cpp/arcticdb/pipeline/query.cpp
@@ -102,7 +102,7 @@ std::unique_ptr<util::BitSet> build_bitset_for_index(
         const auto range_start = std::get<timestamp>(rg.start_);
         const auto range_end = std::get<timestamp>(rg.end_);
         for(auto i = 0u; i < container.size(); ++i) {
-            const auto intersects = range_intersects<RawType>(range_start, range_end, *start_idx_pos, *end_idx_pos - 1);
+            const auto intersects = range_intersects<RawType>(range_start, range_end, *start_idx_pos, *end_idx_pos);
             (*res)[i] = intersects;
             if(intersects)
                 ARCTICDB_DEBUG(log::version(), "range intersects at {}", i);


python/tests/conftest.py
```diff
@@ -1366,6 +1387,35 @@ def lmdb_version_store_static_and_dynamic(request) -> Generator[NativeVersionSto
...
+@pytest.fixture(
+    scope="function",
+    params=(
+        "lmdb_version_store_empty_types_v1",
+        "lmdb_version_store_empty_types_dynamic_schema_v1",
+    ),
+)
+def lmdb_version_store_static_and_dynamic_v1(request) -> Generator[NativeVersionStore, None, None]:
+    """
+    Designed to test V1 combinations between schema and encoding version for LMDB
```
IvoDD: Nit: I find this doc slightly confusing, not sure what `V1 combinations between schema and encoding version` means. What about something like:
```
Designed to test both static and dynamic schema. Uses only lmdb with encoding V1.
```

python/tests/conftest.py
```diff
@@ -1366,6 +1387,35 @@ def lmdb_version_store_static_and_dynamic(request) -> Generator[NativeVersionSto
...
+
+
+@pytest.fixture(
+    scope="function",
+    params=(
+        "lmdb_version_store_empty_types_v1",
+        "lmdb_version_store_empty_types_dynamic_schema_v1",
+    ),
+)
+def lmdb_library_static_and_dynamic_v1(request) -> Generator[NativeVersionStore, None, None]:
```
IvoDD: Nit: Maybe it would be simpler if it just used the `lmdb_version_store_static_and_dynamic_v1` from above?

python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
```diff
@@ -290,20 +290,31 @@ def get_num_data_keys_intersecting_row_range(index, start, end):
...
     count = 0
-    for _, row in index.reset_index().iterrows():
+    for i, (_, row) in enumerate(index.reset_index().iterrows()):
         # end is inclusive when doing date_range but end_index in the column is exclusive
-        if (start is None or start < row["end_index"]) and (end is None or end >= row["start_index"]):
+        if exclude_fully_included:
+            condition1 = start is None or start < row["end_index"]
+            condition2 = end is None or end >= row["start_index"]
+        else:
+            condition1 = start is None or start <= row["end_index"]
```
IvoDD, line 301: `condition1`'s acceptance of equality is the new change here right?

Why is it only for `not exclude_fully_included`? I.e. only for reads and not for updates?
In any case a comment explaining this would be useful.

python/tests/unit/arcticdb/version_store/test_read_index.py
```diff
@@ -204,90 +193,76 @@ def test_as_of_snapshot(self, lmdb_storage, lib_name, dynamic_schema, index):
...
         data = {"a": np.arange(5), "b": np.arange(8)}
-        ac = lmdb_storage.create_arctic()
-        lib = ac.create_library(lib_name, LibraryOptions(dynamic_schema=dynamic_schema))
+        lib = lmdb_library_static_and_dynamic_v1
         lib._nvs.write("sym_recursive", data, recursive_normalizers=True)
         with pytest.raises(UserInputException) as exception_info:
             lib.read("sym_recursive", columns=[])
-        assert "normalizers" in str(exception_info.value)
+        # TODO
+        # assert "normalizers" in str(exception_info.value)
```
IvoDD: Looks like a leftover TODO

cpp/arcticdb/pipeline/index_writer.hpp
```diff
@@ -94,11 +94,13 @@ class IndexWriter {
         );
 
         bool new_col_group = !current_col_.has_value() || *current_col_ < slice.col_range.first;
-        util::check_arg(!current_row_.has_value() || new_col_group
-                            ||
-                                (*current_col_ == slice.col_range.first && *current_row_ < slice.row_range.first),
-                        "expected increasing row group, last col range left value {}, arg {}",
-                        current_col_.value_or(-1), slice.col_range
+        bool has_val = !current_row_.has_value();
```
IvoDD: Nit: probably rename to `missing_val` or `not_has_val`

cpp/arcticdb/pipeline/query.cpp
```diff
@@ -57,10 +58,11 @@ std::unique_ptr<util::BitSet> build_bitset_for_index(
 
         // End index column is exclusive. We want to find the last position where `range_start` is < end_index at position.
         // This is equivalent to finding the first position where range_start + 1 >= end_index at position.
+        const auto adjusted_range_start = is_read_operation ? range_start : range_start + 1;
```
IvoDD, line 59: A comment explaining why even though end index is exclusive we search for `range_start` instead of `range_start+1` would be quite useful I think. E.g. something like
```
End index column should be exclusive when written from new versions of ArcticDB. 
However older versions of ArcticDB (pre 2022) have had bugs where end index is not always exclusive and might be inclusive. 
To ensure backwards compatibilty with those versions we load row slices where `range_start >= end_index` instead of `range_start + 1 >= end_index`.
```

IvoDD: Also I think it would be useful if we can understand more about the root cause of the old bug with end_index not being exclusive.

E.g. could this bug manifest in such a way that it breaks the assumtion of monotinic increases of end_index column?
E.g. if we have 3 row slices each with a single index value `5` could we end up with index looking like:
```
start_index |  end_index
------------+------------
 5          |  6
 5          |  5   # Only this row slice affected by bug and breaking monotonicity of end_index required by lower_bound
 5          |  6
```

cpp/arcticdb/pipeline/query.hpp
```diff
@@ -323,12 +325,12 @@ inline std::vector<FilterQuery<ContainerType>> build_update_query_filters(
                         },
                         [&](const IndexRange &index_range) {
                             util::check(std::holds_alternative<stream::TimeseriesIndex>(index), "Cannot partition by time when a rowcount-indexed frame was supplied");
-                            queries.emplace_back(create_index_filter<ContainerType>(IndexRange{index_range}, dynamic_schema, column_groups));
+                            queries.emplace_back(create_index_filter<ContainerType>(IndexRange{index_range}, dynamic_schema, column_groups, false));
```
IvoDD: Why do we need to use a different behavior for updates? Couldn't the same end_index exclusivity bug affect update where e.g.
```
Initial table has index values (two per row slice):
1
2
---
3
4
Hence index key looks like
start_index | end_index
------------+------------
1           |  2 # Affected by old bug
3           |  5 # Not affected by bug

Then we attempt to update with an inclusive date_range (2-3)
2
3

We will need to read both slices to remove/replace the value for index 2, hence will need the updated query
```

cpp/arcticdb/pipeline/test/test_query.cpp
```diff
@@ -10,60 +10,72 @@
...
     using namespace arcticdb::pipelines;
     TestContainer container;
     container.seg().set_range(2, 4);
     container.seg().set_range(5, 7);
     IndexRange rg(NumericIndex{ 6 }, NumericIndex{8});
-    auto bitset = build_bitset_for_index<TestContainer, TimeseriesIndex>(container, rg, true, false, std::unique_ptr<util::BitSet>{});
+    auto bitset = build_bitset_for_index<TestContainer, TimeseriesIndex>(container, rg, true, false, is_read_operation(), std::unique_ptr<util::BitSet>{});
     ASSERT_EQ((*bitset)[1], true);
     ASSERT_EQ(bitset->count(), 1);
 }
```
IvoDD, line 83: Maybe also add a C++ test showing the edge case condition?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -742,3 +750,70 @@ def test_append_series_with_different_row_range_index_name(lmdb_version_store_dy
...
+    while cur_date <= end_date:
+        df = create_random_data(at_date=cur_date)
+        create_or_append(lib, "sym", df)
+        cur_date = get_next_business_date(cur_date)
+
+    # remove date
+    lib.delete("sym", date_range=(datetime(2025, 9, 2), datetime(2025, 9, 3)))
+
+    data = lib.read("sym", date_range=(datetime(2025, 9, 2), datetime(2025, 9, 2))).data
+    print(data.head())
```
IvoDD: Nit: Remove debug print and data read

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -742,3 +750,70 @@ def test_append_series_with_different_row_range_index_name(lmdb_version_store_dy
     # See Monday 9797097831, it would be best to require that index names are always matching. This is the case for
     # datetime index because it's a physical column. It's a potentially breaking change.
     assert lib.read("sym").data.index.name == "index_name_2"
+
+
+from pandas._libs.tslibs.offsets import BDay
+
+from datetime import datetime
```
IvoDD: Nit: Maybe move imports to top?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -742,3 +750,70 @@ def test_append_series_with_different_row_range_index_name(lmdb_version_store_dy
...
+    """Returns next business date from datetime 'd' (uses pandas BDay)."""
+
+    return (d + BDay(1)).to_pydatetime()
+
+
+def create_or_append(arctic_lib: NativeVersionStore, symbol: str, data: pd.DataFrame) -> None:
+    if symbol not in arctic_lib.list_symbols():
+        arctic_lib.write(symbol, data)
+    else:
+        arctic_lib.append(symbol, data)
```
IvoDD: Why not use `lib.append(write_if_missing=True)`?

python/tests/conftest.py
G-D-Petrov: There was another already existing fixture so I reused that one

python/tests/unit/arcticdb/version_store/test_read_index.py
G-D-Petrov: Removed

cpp/arcticdb/pipeline/query.hpp
G-D-Petrov: Adding the same behavior to updates results in cases where the index is no longer monotonically increasing e.g. like in [this comment](https://github.com/man-group/ArcticDB/pull/2632#discussion_r2348631851).

alexowens90: We should at least have a separate ticket to solve this then, as if we're doing an update we are re-writing data anyway and therefore could fix it properly

alexowens90 (approved):

cpp/arcticdb/pipeline/query.cpp
IvoDD: I think adding the comment here with a reference to the ticket for update would be useful

cpp/arcticdb/pipeline/query.hpp
G-D-Petrov: Ticket added [here](https://github.com/man-group/ArcticDB/issues/2655)

PR #2633, opened 2025-09-04
[9898177828] Use `default_value` in `NullReducer` for `OutputFormat::ARROW`

IvoDD:
#### Reference Issues/PRs
Monday ref: 9898177828

When doing aggregation we explicitly default `sum=0` for slices with no underlying values.
For arrow this means to not set the validity bitmap in this case and to default initialize the values.

The change includes:
- Small refactor of `NullReducer` to extract common parts between `reduce` and `finalize` in `backfill_up_to_frame_offset`
- Modification of `Column::default_initialize` to work across several blocks
- Removes broken `memset` method from `ChunkedBuffer` and instead provides a new `util::initialize` method which can initialize a `ChunkedBuffer` across blocks

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/column_store/chunked_buffer.hpp
+++ b/cpp/arcticdb/column_store/chunked_buffer.hpp
@@ -311,7 +311,7 @@ class ChunkedBufferImpl {
 
     uint8_t* bytes_at(size_t pos_bytes, size_t required) {
         auto [block, pos, _] = block_and_offset(pos_bytes);
-        util::check(pos + required <= block->bytes(), "Block overflow, position {} is greater than block capacity {}", pos, block->bytes());
+        util::check(pos + required <= block->bytes(), "Block overflow, position {} is greater than block capacity {}", pos + required, block->bytes());
         return &(*block)[pos];
     }
 
@@ -366,21 +366,19 @@ class ChunkedBufferImpl {
         }
     }
 
-    void memset_buffer(size_t offset, size_t bytes, char value) {
+    std::vector<std::pair<uint8_t*, size_t>> byte_blocks_at(size_t offset, size_t bytes) {
+        check_bytes(offset, bytes);
+        std::vector<std::pair<uint8_t*, size_t>> result;
         auto [block, pos, block_index] = block_and_offset(offset);
         while(bytes > 0) {
-            const auto size_to_write = block->bytes() - pos;
-            memset(block->data() + pos, size_to_write, value);
+            block = blocks_[block_index];
+            const auto size_to_write = std::min(bytes, block->bytes() - pos);
+            result.push_back({block->data() + pos, size_to_write});
             bytes -= size_to_write;
-            if(bytes > 0) {
-                ++block_index;
-                if(block_index == blocks_.size())
-                    return;
-
-                block = blocks_[block_index];
-                pos = 0;
-            }
+            ++block_index;
+            pos = 0;
         }
+        return result;
     }
 
     template<typename T>
--- a/cpp/arcticdb/column_store/chunked_buffer.hpp
+++ b/cpp/arcticdb/column_store/chunked_buffer.hpp
@@ -366,15 +366,17 @@ class ChunkedBufferImpl {
         }
     }
 
-    std::vector<std::pair<uint8_t*, size_t>> byte_blocks_at(size_t offset, size_t bytes) {
-        check_bytes(offset, bytes);
+    // Returns a vector of continuous buffers, each designated by a pointer and size
+    // Similar to `bytes_at` but will work if the requested range spans multiple continuous blocks.
+    std::vector<std::pair<uint8_t*, size_t>> byte_blocks_at(size_t pos_bytes, size_t required_bytes) {
+        check_bytes(pos_bytes, required_bytes);
         std::vector<std::pair<uint8_t*, size_t>> result;
-        auto [block, pos, block_index] = block_and_offset(offset);
-        while(bytes > 0) {
+        auto [block, pos, block_index] = block_and_offset(pos_bytes);
+        while(required_bytes > 0) {
             block = blocks_[block_index];
-            const auto size_to_write = std::min(bytes, block->bytes() - pos);
+            const auto size_to_write = std::min(required_bytes, block->bytes() - pos);
             result.push_back({block->data() + pos, size_to_write});
-            bytes -= size_to_write;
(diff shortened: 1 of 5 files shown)


python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -667,3 +667,47 @@ def test_project_dynamic_schema_complex(lmdb_version_store_dynamic_schema_v1):
...
+
+    table = lib.read(sym, query_builder=q).data
+    print(table.to_pandas().sort_index())
+    import polars as pl
+    print(pl.from_arrow(table.column("mean_col")))
+    # sum_col is correctly filled with 0s instead of nulls
+    assert pc.count(table.column("sum_col"), mode="only_null") == 0
+    # TODO: Fix the TODOs in `CopyToBufferTask` to make num_nulls=5 as expected
+    # For this test it so happens that one present and one missing value end up in the same bucket.
+    # Copying then default initializes the missing values instead of setting the validity bitmap.
```
IvoDD, line 704: This TODO will be addressed in a follow up PR as it will be a slightly more involved change.

cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -366,21 +366,19 @@ class ChunkedBufferImpl {
         }
     }
 
-    void memset_buffer(size_t offset, size_t bytes, char value) {
+    std::vector<std::pair<uint8_t*, size_t>> byte_blocks_at(size_t offset, size_t bytes) {
```
alexowens90: This is quite an odd API, worth a comment explaining what it does and why it's useful

cpp/arcticdb/util/sparse_utils.hpp
```diff
@@ -86,39 +86,6 @@ void default_initialize(uint8_t* data, const size_t bytes) {
...
-        if (const std::shared_ptr<TypeHandler>& handler = arcticdb::TypeHandlerRegistry::instance()->get_handler(type_descriptor);handler) {
-            handler->default_initialize(buffer, offset, bytes, shared_data, handler_data);
-        } else {
-            internal::raise<ErrorCode::E_INVALID_ARGUMENT>(
-                "Default initialization for {} is not implemented.",
-                type_descriptor
-            );
-        }
-    }
-}
```
alexowens90, line 119: Was this just unused?

python/tests/unit/arcticdb/version_store/test_arrow.py
alexowens90: Yeh the right thing to do is have the aggregation clause produce sparse data, then when outputting as Arrow it will just be right

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -667,3 +667,47 @@ def test_project_dynamic_schema_complex(lmdb_version_store_dynamic_schema_v1):
...
+        "sum_col": "sum",
+        "min_col": "min",
+        "max_col": "max",
+        "count_col": "count",
+    })
+
+    table = lib.read(sym, query_builder=q).data
+    print(table.to_pandas().sort_index())
+    import polars as pl
+    print(pl.from_arrow(table.column("mean_col")))
```
alexowens90: nit: Remove prints

python/tests/unit/arcticdb/version_store/test_arrow.py
```diff
@@ -667,3 +667,47 @@ def test_project_dynamic_schema_complex(lmdb_version_store_dynamic_schema_v1):
...
+    # sum_col is correctly filled with 0s instead of nulls
+    assert pc.count(table.column("sum_col"), mode="only_null") == 0
+    # TODO: Fix the TODOs in `CopyToBufferTask` to make num_nulls=5 as expected
+    # For this test it so happens that one present and one missing value end up in the same bucket.
```
alexowens90, line 703: The number of buckets defaults to the number of cores, so this test probably won't reliably pass on all machines

cpp/arcticdb/util/sparse_utils.hpp
IvoDD: Yeah it was not used and I needed a similar function which was taking a `default_value` argument, so removed this `default_initialize` which was not used anywhere and added the new `initialize` method which takes an optional default value.

python/tests/unit/arcticdb/version_store/test_arrow.py
IvoDD: Ah good catch, commenting out the assertions and leaving the TODO.

alexowens90: In the next PR can set `Partition.NumBuckets` to get consistent behaviour across different machines

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -763,38 +763,41 @@ class NullValueReducer {
...
+    void backfill_all_zero_validity_bitmaps_up_to(size_t up_to_block_offset) {
         // Fills up all validity bitmaps with zeros from `column_block_idx_` until reaching `up_to_block_offset`.
-        // If `up_to_block_offset` is `std::nullopt` then fills up until the end of the column.
         const auto& block_offsets = column_.block_offsets();
-        util::check(!up_to_block_offset.has_value() || up_to_block_offset.value() <= block_offsets.back(), "up_to_block_offset outside of range");
-        for (; column_block_idx_ < block_offsets.size() - 1; ++column_block_idx_) {
-            if (up_to_block_offset.has_value() && block_offsets.at(column_block_idx_) >= up_to_block_offset.value()) {
+        util::check(up_to_block_offset <= block_offsets.back(), "up_to_block_offset {} outside of range {}", up_to_block_offset, block_offsets.back());
+        for (; column_block_idx_ < block_offsets.size() - 1 && block_offsets.at(column_block_idx_) < up_to_block_offset; ++column_block_idx_) {
+            if (block_offsets.at(column_block_idx_) >= up_to_block_offset) {
```
G-D-Petrov: Isn't this `if` redundant? Looks like it is checking the same condition as in the loop

IvoDD: Ah yes that was the intention of adding the new condition but somehow missed removing the if. Fixing now

PR #2646, opened 2025-09-15
Cleanup CI buckets/containers

grusev:

Release storage on regular basis. 

We leak storage due to different types of failures that cause symbols and libraries not to be cleaned after tests. This PR adds a step for cleaning the storages - AWS S3, GCP, Azure once a week. By default all data older than 28 days will be cleaned, thus presevrving info for the last moth.

The PR adds small boto and azure library for obtaining info and cleaning buckets/containers. This library can be further reused and enhanced if needed for more bucket management that we will need

Prior to this PR the leftovers were as follows:

```
AWS TOTAL SIZE : 891,874,609,188
GCP TOTAL SIZE : 3,363,272,750
AZURE TOTAL SIZE: 2,706,787,373
```

Cleanup buckets with no more than 1 month old data (initial run) : https://github.com/man-group/ArcticDB/actions/runs/17733685971/job/50390234882

Next run after the buckets are clean: https://github.com/man-group/ArcticDB/actions/runs/17756785784/job/50460916726

```
2025-09-16 06:28:50,748 - __main__ - INFO - Cleaning before: 2025-08-19 06:28:50.748631+00:00
2025-09-16 06:28:50,749 - __main__ - INFO - Cleaning-up GCP storage
2025-09-16 06:29:05,352 - __main__ - INFO - GCP TOTAL SIZE: 339770117
2025-09-16 06:29:19,172 - utils.bucket_management - INFO - Found 0 objects to delete before 2025-08-19 06:28:50.748631+00:00
2025-09-16 06:29:30,408 - __main__ - INFO - GCP TOTAL SIZE: 339770117
2025-09-16 06:29:30,408 - __main__ - INFO - Cleaning-up Azure storage
2025-09-16 06:29:35,005 - __main__ - INFO - AZURE TOTAL SIZE: 3591350
2025-09-16 06:29:38,719 - utils.bucket_management - INFO - Found 0 blobs to delete before 2025-08-19 06:28:50.748631+00:00
2025-09-16 06:29:42,433 - __main__ - INFO - AZURE TOTAL SIZE: 3591350
2025-09-16 06:29:42,433 - __main__ - INFO - Cleaning-up S3 storage
2025-09-16 06:29:43,227 - __main__ - INFO - AWS S3 TOTAL SIZE: 0
2025-09-16 06:29:43,418 - __main__ - INFO - AWS S3 TOTAL SIZE: 0
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


.github/workflows/delete_sts_roles.yml
```diff
@@ -19,16 +19,31 @@ jobs:
...
           aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"        
+          gcp_access_key: "${{ secrets.GCP_S3_ACCESS_KEY }}"
+          gcp_secret_key: "${{ secrets.GCP_S3_SECRET_KEY }}"
+          azure_container: "githubblob" # DEFAULT BUCKET FOR AZURE
+          azure_connection_string: "${{ secrets.AZURE_CONNECTION_STRING }}"
+
+      - name: Delete STS Roles
+        run: |
+          cd python
+          rm -rf arcticc
```
IvoDD: Why is removing arcticc needed here but not for bucket cleanup?

python/utils/cleanup_test_buckets.py
```diff
@@ -0,0 +1,51 @@
...
+cutoff = now - timedelta(days=28)
+
+logger.info(f"Cleaning before: {cutoff}")
+
+logger.info("Cleaning-up GCP storage")
+gcp = gcp_client()
+gcp_bucket = os.getenv("ARCTICDB_REAL_GCP_BUCKET")
+logger.info(f"GCP TOTAL SIZE: {get_gcp_bucket_size(gcp, gcp_bucket)}")
+delete_gcp_bucket(gcp, gcp_bucket, cutoff)
+logger.info(f"GCP TOTAL SIZE: {get_gcp_bucket_size(gcp, gcp_bucket)}")
```
IvoDD: Nit: Maybe change logs to containt words "BEFORE" and "AFTER" so it makes easier to e.g. grep results

python/utils/bucket_management.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Parameters:
+        client: boto3 S3-compatible client (e.g., for GCS via HMAC).
+        bucket_name: Name of the bucket.
+        handler : Function to apply to each qualifying object.
+        cutoff_date (Optional): Only include objects older than this date.
+                                    Defaults to current UTC time.
+    """
+    if cutoff_date is None:
+        cutoff_date = datetime.now(timezone.utc)
```
IvoDD, line 69: Nit: we might miss a few items due to clock skew. Might be slightly better to call the handler `if cutoff_date is None or LastModified < cutoff_date`

python/utils/bucket_management.py
```diff
@@ -0,0 +1,299 @@
...
+
+    list_bucket(client, bucket_name, collect_key, cutoff_date)
+    logger.info(f"Found {len(keys_to_delete)} objects to delete before {cutoff_date or datetime.now(timezone.utc)}")
+
+    def delete_key(key: str) -> None:
+        client.delete_object(Bucket=bucket_name, Key=key)
+        logger.info(f"Deleted: {key}")
+
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        executor.map(delete_key, keys_to_delete)
```
IvoDD, line 104: Does GCP not support batch deletions like we use for aws?

python/utils/bucket_management.py
```diff
@@ -0,0 +1,299 @@
...
+            logger.info(f"Deleted: {blob_name}")
+        except Exception as e:
+            logger.error(f"Failed to delete {blob_name}: {e}")
+
+
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        executor.map(delete_blob, blobs_to_delete)
+
+
+def get_s3_bucket_size(
```
IvoDD: There is a lot of duplication between `get_gcp_bucket_size` and `get_s3_bucket_size`. In fact I think they do the exact same thing.

I think I like the `get_s3_bucket_size` better. We can make the `get_gcp_bucket_size` just use the `get_s3_bucket_size`

python/utils/bucket_management.py
```diff
@@ -0,0 +1,299 @@
...
+        handler : Function to apply to each qualifying blob.
+        cutoff_date (Optional[datetime]): Only include blobs older than this date.
+                                          Defaults to current UTC time.
+    """
+    if cutoff_date is None:
+        cutoff_date = datetime.now(timezone.utc)
+
+    container_client = client.get_container_client(container_name)
+    for blob in container_client.list_blobs():
+        if blob.last_modified and blob.last_modified < cutoff_date:
```
IvoDD, line 145: Do you know when `blob.last_modified` is missing? Shouldn't we in this case still count it to the total count in case `cutoff_date=None`?

python/utils/bucket_management.py
```diff
@@ -0,0 +1,299 @@
...
+        nonlocal total_size
+        total_size += blob.size
+
+    list_azure_container(blob_service_client, container_name, size_accumulator, cutoff_date)
+    return total_size
+
+
+def delete_azure_container33(
+    blob_service_client: BlobServiceClient,
+    container_name: str,
```
IvoDD: This looks like a leftover function which is worse than `delete_azure_container`. Let's remove it.

python/utils/bucket_management.py
```diff
@@ -0,0 +1,299 @@
...
+    def collect_blob(blob: BlobProperties) -> None:
+        blobs_to_delete.append(blob.name)
+
+    list_azure_container(client, container_name, collect_blob, cutoff_date)
+
+    logger.info(f"Found {len(blobs_to_delete)} blobs to delete before {cutoff_date or datetime.now(timezone.utc)}")
+
+    def delete_blob(blob_name: str) -> None:
+        try:
+            container_client.delete_blob(blob_name)
```
IvoDD, line 190: It looks like there exists a [batch deletion](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.specialized.blobbatchclient.deleteblobs?view=azure-dotnet) operation for azure as well.

Should we use that for better performance and less IO price (not sure how price is calclated for azure)?

python/utils/bucket_management.py
grusev: yes - GCP does not support it

python/utils/bucket_management.py
grusev: cut off date cannot be None. it is assigned value just before that - current time . So all is counted even if cutoff date is none

also, last modified should be there, should not be missing, but if that happends we will not fail.

python/utils/bucket_management.py
grusev: I have thought of that but forgot to make it at last. Thanks for spoting it!

python/utils/bucket_management.py
grusev: The main goal of current ops are to delete all objects before given time. We do not aim to empty the bucket as operations might be still working. That is why we run clean buckets with cut_off date before 28 days.

.github/workflows/delete_sts_roles.yml
grusev: We have to remove them once from the file system. If left there, the code will not work as those packages doe not contain build protobufs. If removed protobufs will be taken from the installed lib

python/utils/bucket_management.py
grusev: No need to do optimizations now, the data is not large and it is getting deleted very fast already. I have left a comment in the code so that we utilize this at point of time when azure is used more.

.github/workflows/delete_sts_roles.yml
```diff
@@ -1,4 +1,4 @@
-name: Scheduled Deletion of STS Roles
+name: Scheduled Cleanup
```
G-D-Petrov: nit: rename the file as well to match the name herre

G-D-Petrov (approved):

PR #2650, opened 2025-09-16
Staged data with tokens demo notebook

ognyanstoimenov:
Add staged data notebook to the website. 

Preview of the notebook: https://github.com/man-group/ArcticDB/blob/1ae8d5a2c6a75ce6a33b853b81401d0ab40af303/docs/mkdocs/docs/notebooks/ArcticDB_staged_data_with_tokens.ipynb
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


docs/mkdocs/docs/notebooks/ArcticDB_staged_data_with_tokens.ipynb
jamesblackburn, line 1: A few comments aiming to make this more user-centric.

Add:
* An introduction to Staging data, and description of the "Why" use-cases e.g. parallel loading of data. Then describe the high-level flow: stage data in parallel or sequentially, collect the `tokens` / `cookies` and finalise the staged data.

Remove:
* Old API: If we want the new way to be standard then no point referencing the old way

Change: 
* New API -> API
* Document "stage result": i.e. every staged write returns a 'cookie/token' that is used for the finalisation stage.

Wy do we need this: -> Move to the introduction. Likely doesn't require code just description of how the API avoids pitfalls. Staging while finalising: similar just requires a comment rather than code. 

How do you clean-up unstaged data in future runs? Should there be a clear or equivalent method to cleanup before running a large backfill.

PR #2660, opened 2025-09-18
[9898173277] Resolve row range edge cases for arrow

IvoDD:
#### Reference Issues/PRs
Fixes monday ref 9898173277
Fixes monday ref 9943908001 when using `query_builder`'s `date_range`.

- Small refactor to use abstract common `is_slice_in_row_range` and `is_slice_in_index_range` to be used both in regular reads and in processing pipeline. This resolves the issue fixed in #2632 for query builder date ranges as well.
- If requested row/date range is empty, we no longer read any data keys (previously we would read one)
- Removes several unused methods from `query.hpp` and `index_range.hpp` (index_range one was quite broken)
- Adds arrow tests for:
  - open ended row/date range filters
  - empty row/date ranges
  - negative row ranges

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/pipeline/query.cpp
```diff
@@ -16,17 +16,31 @@ namespace arcticdb::pipelines {
...
-IndexValue end_index(const index::IndexSegmentReader& isr, std::size_t row) {
-    return index::index_value_from_segment(isr.seg(), row, index::Fields::end_index);
+RowRange slice_row_range_at(const std::vector<SliceAndKey>& sk, std::size_t row) { return sk[row].slice_.row_range; }
+
+bool is_slice_in_row_range(const RowRange& slice_row_range, const RowRange& row_filter) {
+    // If the row_filter is empty we return false (i.e. don't read the slice).
+    // This is required for cases like a range (3, 2) which falls completely within an index row (0, 10). We
+    // know that if the range is empty we don't need to read the data key for (0, 10) because it won't contain
+    // any elements within the empty range.
+    return slice_row_range.first < row_filter.second && slice_row_range.second > row_filter.first &&
```
G-D-Petrov, line 32: nit: Shouldn't we do the `!row_filter.empty()` check first?
Same in `is_slice_in_index_range`

G-D-Petrov (approved):

alexowens90 (approved):
Is our off-by-one nanosecond testing with `QueryBuilder` date ranges robust enough already that we can be confident of no regressions here?

cpp/arcticdb/pipeline/query.cpp
IvoDD: The reason I've put it in the end is mainly for very slight efficiency with respect to short circuit evaluation of `&&`s. 

I think `!row_filter.empty()` will very infrequently be false so it wouldn't short circuit often if first.
Also the calculation inside `row_filter.empty()` involves several branches so I'd like that computation to get short circuited out where possible.

In practice this probably doesn't matter much due to compiler optimizations and efficient branch prediction, but am planning to keep as is unless it is confusing

PR #2663, opened 2025-09-19
WIP Enhancement 8987170104: arrow write numeric dense data

alexowens90:
#### Reference Issues/PRs
[8987170104](https://man312219.monday.com/boards/7852509418/pulses/8987170104)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/arrow/arrow_utils.hpp
```diff
@@ -19,4 +19,9 @@ std::vector<sparrow::array> arrow_arrays_from_column(const Column& column, std::
 
 std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(SegmentInMemory& segment);
 
+std::pair<SegmentInMemory, std::optional<size_t>> arrow_data_to_segment(
```
IvoDD, line 34: This method does the reordering of columns and returns the position of the index column if specified.

I think this would be slightly clearer and easier to follow if this column reordering is done during normalization. Then everything that touches the normalization metadata for arrow would happen in a single place instead of half of the fields being set by normalization and half in `py_ndf_to_frame`.

Also it should still be zero-copy as it will just be moving around pointers.

Also this method could have a simpler interface:
`SegmentInMemory arrow_data_to_segment(const std::vector<sparrow::record_batch>& record_batches)`

python/arcticdb/version_store/library.py
```diff
@@ -41,6 +42,7 @@
 
 
 NORMALIZABLE_TYPES = (pd.DataFrame, pd.Series, np.ndarray)
+_EXPERIMENTAL_NORMALIZABLE_TYPES = (pa.Table,)
```
IvoDD: I think this toplevel `pa.Table` will fail if we don' have pyarrow installed. Instead I think we should expose a method like:

```python
def check_acceptable_input_type(self, data):
    if _PYARROW_AVAILABLE and isinstance(data, pa.Table) and self._nvs._allow_arrow_input:
        pass
    elif not isinstance(data, NORMALIZABLE_TYPES):
        raise ArcticUnsupportedDataTypeException(
            "data is of a type that cannot be normalized. Consider using "
            f"write_pickle instead. type(data)=[{type(data)}]"
        )


```

python/arcticdb/version_store/_store.py
```diff
@@ -481,6 +487,8 @@ def _try_normalize(
                     coerce_columns=coerce_columns,
```
IvoDD, line 486: Nit: Remove the `TODO: just for pandas dataframes for now` above

python/arcticdb/version_store/_store.py
```diff
@@ -1129,8 +1171,8 @@ def get_column_stats_info(
         version_query = self._get_version_query(as_of, **kwargs)
         return self.version_store.get_column_stats_info_version(symbol, version_query).to_map()
 
-    def _batch_read_keys(self, atom_keys):
-        for result in self.version_store.batch_read_keys(atom_keys):
+    def _batch_read_keys(self, atom_keys, read_query):
+        for result in self.version_store.batch_read_keys(atom_keys, read_query):
```
IvoDD: Nit: `read_options` instead of `read_query`

python/arcticdb/version_store/_store.py
```diff
@@ -2155,53 +2223,54 @@ def tail(
...
-                data,
-                read_result.frame_data.names,
-                read_result.frame_data.index_columns,
-                row_count,
-                read_result.frame_data.offset,
+    def _post_process_dataframe(
+        self, read_result, read_query, read_options, implement_read_index=False, head=None, tail=None
+    ):
+        # Range filters for arrow are processed inside C++ layer. So we skip the post-processing in this case.
+        if not isinstance(read_result.frame_data, ArrowOutputFrame):
```
IvoDD, line 2236: Note to other reviewers:
Only change here is to not return early on `ArrowOutputFrame`. This is to allow reading `custom normalized data` with arrow output format.

The rest of the diff is just formatting because of increased indent.

python/arcticdb/version_store/_normalization.py
```diff
@@ -718,16 +719,25 @@ def apply_pyarrow_operations(self, table, op: ArrowNormalizationOperations):
...
+        for pa_record_batch in pa_record_batches:
+            arcticdb_record_batch = RecordBatchData()
+            pa_record_batch._export_to_c(arcticdb_record_batch.array(), arcticdb_record_batch.schema())
+            arcticdb_record_batches.append(arcticdb_record_batch)
+        norm_metadata = NormalizationMetadata()
+        index_column = kwargs.get("index_column", None)
+        if index_column is None:
+            norm_metadata.arrow_table.has_index = False
+        else:
+            check(isinstance(index_column, str), "Arrow index column specifier must be a string")
```
IvoDD, line 734: Maybe also check `index_column in table.column_names`?

python/arcticdb/version_store/_normalization.py
```diff
@@ -718,16 +719,25 @@ def apply_pyarrow_operations(self, table, op: ArrowNormalizationOperations):
...
+            norm_metadata.arrow_table.has_index = True
+            norm_metadata.arrow_table.index_column_name = index_column
```
IvoDD: We should also add `norm_metadata.arrow_table.index_column_position`?

cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -224,12 +224,19 @@ message NormalizationMetadata {
...
+        uint64 index_column_position = 3;
+    }
+
     oneof input_type {
         PandasDataFrame df = 1;
         PandasDataFrame series = 2;
         NormalisedTimeSeries ts = 3;
         MsgPackFrame msg_pack_frame = 4;
         NdArray np = 5;
+        ArrowTable arrow_table = 6;
```
IvoDD: For appending mix of arrow and pandas:

`oneof` might not be the best approach. We might want to maintain both the `PandasDataFrame` and `ArrowTable` normalization metadata.

E.g. if we 
```
lib.write("sym", pd_df_with_multiindex)
lib.append("sym", table_with_some_index, index_column="index_col")
```
We probably want to preserve the multiindex metadata if we read as pandas but we also want to keep index column position if reading back as arrow.

As you said arrow and pandas interactions with modification operations will be leftover for a follow up PR but I think it is worth deciding on the `protobuf` structure in this PR. Protobuf changes are hard to reason about backwards compatibility etc. and it would be best if we limit it to just one change.

cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -224,12 +224,19 @@ message NormalizationMetadata {
         repeated uint64 shape = 1;
     }
 
+    message ArrowTable {
+        bool has_index = 1;
+        string index_column_name = 2;
+        uint64 index_column_position = 3;
```
IvoDD, line 230: Not for this PR but we should open a ticket to gracefully raise if normalization metadata changes on append:

```python
lib.write("sym", arrow_table_with_A_and_B, index_column="A")
lib.append("sym", arrow_table_with_A_and_B, index_column="B")
```
Should probably just raise. Without the extra guards we could end up with super broken data.

python/arcticdb/version_store/_normalization.py
```diff
@@ -1451,6 +1504,9 @@ def get_normalizer_for_type(self, item):
         if isinstance(item, np.ndarray):
             return self.np.normalize
 
+        if isinstance(item, pa.Table) and allow_arrow_input:
```
IvoDD: I think this will trigger failures if we try pickling in an environment where `pyarrow` is not installed. Should add `_PYARROW_ENABLED and ...`

python/arcticdb/version_store/_normalization.py
```diff
@@ -1627,6 +1687,14 @@ def _strip_tz(s, e):
             # of duplicating exception messages.
             raise SortingException("E_UNSORTED_DATA When calling update, the input data must be sorted.")
         data = data.loc[pd.to_datetime(start) : pd.to_datetime(end)]
+    elif isinstance(data, pa.Table):
```
IvoDD: This can also raise if `pyarrow` is not installed.
We should probably try and make sure `pyarrow` is not installed in one of our `compat` env tests to catch issues like these.

python/arcticdb/version_store/_normalization.py
```diff
@@ -1627,6 +1687,14 @@ def _strip_tz(s, e):
...
+        check(index_column is not None, "Cannot update with pyarrow Table without specifying index column")
+        col = data.column(index_column)
+        start, end = _strip_tz(start, end)
+        check(
+            start <= col[0].as_py().tz_localize(None) and end >= col[-1].as_py().tz_localize(None),
+            "update with date_range and pyarrow Tbale not yet supported with date_range overlapping the data",
```
IvoDD: Nit: typo `Tbale`

cpp/proto/arcticc/pb2/descriptors.proto
IvoDD: Also we should think about what should happen if we use dynamic schema. Maybe for dynamic schema we should only allow `index_column_position=0`. Otherwise:
```
to_write with index_column = "index":
A index B
a  1    3.0
b  2    4.0

to append with index_column = "index":
index B
3     5.0
4     6.0

With current code read would return:
index A   B
1     a   3.0
2     b    4.0
3     N/A 5.0
4     N/A 6.0
```
What position should the index appear in then? With

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -98,4 +102,139 @@ std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(Segmen
...
+                    record_batch->names()[idx]
+            );
+            auto arrow_array_buffers = sparrow::get_arrow_array_buffers(*arrow_array, *arrow_schema);
+            // arrow_array_buffers.at(0) seems to be the validity bitmap, and may be NULL if there are no null values
+            // need to handle it being non-NULL and all 1s though
+            const auto* data = arrow_array_buffers.at(1).data<uint8_t>();
+            if (is_bool_type(data_type)) {
+                // Arrow bool columns are packed bitsets
+                // This is the limiting factor when writing a lot of bool data as it is serial. This should be moved to
+                // WriteToSegmentTask
```
IvoDD, line 363: Just note: Agreed, to do this we'll probably need a new `DataType::PACKED_BOOL` which will be input only and never written, just passed to `WriteToSegmentTask`

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -98,4 +102,139 @@ std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(Segmen
...
+                        make_scalar_type(data_types.at(idx)),
+                        Sparsity::NOT_PERMITTED,
+                        std::move(chunked_buffers.at(idx))
+                )
+        );
+    }
+    for (size_t idx = 0; idx < column_names.size(); ++idx) {
+        if (!index_column_position.has_value() || idx != *index_column_position) {
+            // The Arrow data may be semantically sparse, but this buffer is still dense, hence Sparsity::NOT_PERMITTED
+            seg.add_column(
```
IvoDD, line 391: Nit: Tiny bit of code duplication with the index case. Would be removed if we do column reordering in python.

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+const StreamDescriptor& InputFrame::desc() const { return const_cast<InputFrame*>(this)->desc(); }
+
+void InputFrame::set_offset(ssize_t off) const { offset = off; }
+
+void InputFrame::set_sorted(SortedValue sorted) {
+    switch (sorted) {
+    case SortedValue::UNSORTED:
```
IvoDD: Why not replace this switch statement with just:
```c++
desc().set_sorted(sorted)
```

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -0,0 +1,170 @@
...
+            input_data,
+            [row](SegmentInMemory& seg) {
+                util::check(
+                        row < seg.row_count(),
+                        "Out of range row {} requsted in InputFrame::index_value_at with segment of length",
+                        row,
+                        seg.row_count()
+                );
+                const auto& index_column = seg.column(0);
+                return *index_column.scalar_at<timestamp>(row);
```
IvoDD, line 80: This is `O(log(number_of_record_batches))` I think. I don't think there is much we can do about it here but at least a comment flagging that this could be slower than `O(1)` for arrow would be helpful.

cpp/arcticdb/pipeline/slicing.hpp
```diff
@@ -87,15 +87,13 @@ inline auto end_index_generator(T end_index) { // works for both rawtype and raw
...
+inline auto get_partial_key_gen(std::shared_ptr<InputFrame> frame, const TypedStreamVersion& key) {
     using PartialKey = stream::StreamSink::PartialKey;
 
     return [frame = std::move(frame), &key](const FrameSlice& s) {
         if (frame->has_index()) {
-            util::check(static_cast<bool>(frame->index_tensor), "Got null index tensor in get_partial_key_gen");
-            auto& idx = frame->index_tensor.value();
-            auto start = *idx.ptr_cast<timestamp>(slice_begin_pos(s, *frame));
-            auto end = *idx.ptr_cast<timestamp>(slice_end_pos(s, *frame));
+            auto start = frame->index_value_at(slice_begin_pos(s, *frame));
```
IvoDD, line 98: This is called once per each row slice. This means that calculating these start and end indices becomes: `O(num_row_slices_to_write * log(num_record_batches))`.

With a refactor we could find the next start and end in `O(1)` but probably that's not on the hot path (the actual work in `WriteToSegmentTask` should be more), so not worth changing but at least a comment would be good I think.

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -0,0 +1,170 @@
...
+                            util::check(
+                                    static_cast<bool>(input_tensors.index_tensor),
+                                    "Got null index tensor in set_index_range"
+                            );
+                            util::check(input_tensors.index_tensor->nbytes() > 0, "Empty index tensor");
+                            auto& tensor = input_tensors.index_tensor.value();
+                            auto start_t = tensor.ptr_cast<RawType>(0);
+                            auto end_t = tensor.ptr_cast<RawType>(static_cast<size_t>(tensor.shape(0) - 1));
+                            index_range.start_ = IndexValue(static_cast<timestamp>(*start_t));
+                            index_range.end_ = IndexValue(static_cast<timestamp>(*end_t));
```
IvoDD: Can we simplify this `variant_match` if we use `index_value_at`?
E.g. something like:
```c++
index_range.start_ = index_value_at(0)
index_range.end_ = index_value_at(num_rows - 1);
```

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -44,21 +44,59 @@ WriteToSegmentTask::WriteToSegmentTask(
...
+    }
+    seg.set_row_data((slice_.rows().second - slice_.rows().first) - 1);
+    return seg;
+}
+
+Column WriteToSegmentTask::slice_column(const SegmentInMemory& frame, size_t col_idx, size_t offset) const {
+    const auto& source_column = frame.column(col_idx);
+    const auto first_byte = (slice_.rows().first - offset) * get_type_size(source_column.type().data_type());
+    const auto bytes = ((slice_.rows().second - offset) * get_type_size(source_column.type().data_type())) - first_byte;
+    auto byte_blocks_at = source_column.data().buffer().byte_blocks_at(first_byte, bytes);
```
IvoDD, line 82: Just flagging: This is `O(log(num_record_batches))`. 

We could make it `O(1)` if we reuse the offsets calculated from the previous call.

This wouldn't have a big impact if input arrow table is very fragmented because we would need to do memory copies in that case, so I don't think we need to change anything. Although a comment explaining this might be helpful.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -98,4 +102,139 @@ std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(Segmen
...
+            auto arrow_array_buffers = sparrow::get_arrow_array_buffers(*arrow_array, *arrow_schema);
+            // arrow_array_buffers.at(0) seems to be the validity bitmap, and may be NULL if there are no null values
+            // need to handle it being non-NULL and all 1s though
+            const auto* data = arrow_array_buffers.at(1).data<uint8_t>();
+            if (is_bool_type(data_type)) {
+                // Arrow bool columns are packed bitsets
+                // This is the limiting factor when writing a lot of bool data as it is serial. This should be moved to
+                // WriteToSegmentTask
+                if (record_batch == record_batches.cbegin()) {
+                    chunked_buffer.ensure(total_rows);
```
IvoDD: We'll need to make sure we correctly handle deallocating the mix of owning and non-owning buffers within the segment

IvoDD: Also we are currently using `AllocationType::DYNAMIC` which is the default. If we have a very wide `1` row long table we would overallocate `4095*num_columns` bytes.
It would be better to use presized for bool columns.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -217,15 +257,8 @@ folly::Future<entity::AtomKey> append_frame(
             frame->index,
             [&index_segment_reader, &frame, ignore_sort_order](const TimeseriesIndex&) {
                 util::check(frame->has_index(), "Cannot append timeseries without index");
-                util::check(static_cast<bool>(frame->index_tensor), "Got null index tensor in append_frame");
-                auto& frame_index = frame->index_tensor.value();
-                util::check(
-                        frame_index.data_type() == DataType::NANOSECONDS_UTC64,
```
IvoDD, line 223: Why drop the assertion? We could still theoretically get a tensor frame with a non timeseries index which would break subsequent logic.

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -216,91 +218,119 @@ NativeTensor obj_to_tensor(PyObject* ptr, bool empty_types) {
...
-            res->index = stream::RowCountIndex();
-            res->desc.set_index_type(IndexDescriptor::Type::ROWCOUNT);
-            res->desc.add_scalar_field(index_tensor.dt_, index_column_name);
-            res->field_tensors.push_back(std::move(index_tensor));
+            frame.index = stream::RowCountIndex();
+            desc.set_index_type(IndexDescriptor::Type::ROWCOUNT);
+            desc.add_scalar_field(index_tensor.dt_, index_column_name);
+            field_tensors.push_back(std::move(index_tensor));
         }
+        opt_index_tensor = std::move(index_tensor);
```
IvoDD: Shouldn't we set the `opt_index_tensor` only in the if case? We're using after move in the `else` case.

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -141,19 +141,19 @@ TimeseriesDescriptor pack_timeseries_descriptor(
...
+        const std::shared_ptr<pipelines::InputFrame>& frame, size_t existing_rows,
         std::optional<entity::AtomKey>&& prev_key, bool allow_sparse
 ) {
     using namespace arcticdb::stream;
 
     auto offset_in_frame = 0;
     auto slice_num_for_column = 0;
     const auto num_rows = frame->num_rows;
-    auto index_tensor = std::move(frame->index_tensor);
+    auto index_tensor = std::move(frame->opt_index_tensor());
```
IvoDD, line 156: Here we assume `InputFrame` is not arrow. Maybe let's rename the method to `incomplete_segment_from_tensor_frame` and add a check that `frame.has_tensors()`.

The `incomplete_segment_from_frame` is somewhat hackily used with some of library tool methods that need to operate on segments and this might start breaking in surprising ways if users start passing to it arrow data. Probably we can add a low priority ticket to fix the library tool methods with arrow.

cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -275,12 +275,15 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
     bool sparsify_floats{false};
 
     auto next_key = std::nullopt;
-    auto segment = incomplete_segment_from_frame(frame, 0, next_key, sparsify_floats);
+    // This clone is unnecessarily expensive. We should have a sort method that produces a new segment (current one is
+    // in-place)
+    auto segment = frame->has_tensors() ? incomplete_segment_from_frame(frame, 0, next_key, sparsify_floats)
+                                        : frame->segment().clone();
```
IvoDD, line 286: Just note: We're not setting the `next_key` in case of arrow. 
This doesn't matter as that is only used from the library tool.

cpp/arcticdb/util/bitset.cpp
```diff
@@ -0,0 +1,62 @@
...
+    auto initial_bits = shift == 0 ? 0 : std::min(8 - shift, num_bits);
+    if (initial_bits > 0) {
+        uint8_t byte = *packed_bits++;
+        for (size_t bit = shift; bit < 8; ++bit) {
+            *dest_ptr++ = (byte >> bit) & 1;
+        }
+    }
+    auto leftover_bits = (num_bits - initial_bits) % 8;
+    size_t num_bytes = (num_bits - initial_bits) / 8;
+    // Experimentally, this approach is ~33% faster than the naive bit iteration approach used for the initial/leftover
```
IvoDD, line 36: Wow, nice!

cpp/arcticdb/version/version_core.cpp
```diff
@@ -136,7 +136,7 @@ IndexDescriptorImpl check_index_match(const arcticdb::stream::Index& index, cons
 }
 } // namespace
 
-void sorted_data_check_append(const InputTensorFrame& frame, index::IndexSegmentReader& index_segment_reader) {
+void sorted_data_check_append(const InputFrame& frame, index::IndexSegmentReader& index_segment_reader) {
     if (!index_is_not_timeseries_or_is_sorted_ascending(frame)) {
         sorting::raise<ErrorCode::E_UNSORTED_DATA>(
                 "When calling append with validate_index enabled, input data must be sorted"
```
IvoDD, line 143: We're always assuming arrow index is sorted. Should we verify when `validate_index=True`? Not sure what is the behavior for pandas but we probably should mimic it.

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -0,0 +1,757 @@
...
+    "data",
+    [
+        [True],
+        [True, False],
+        [False, True, True],
+        [True, False, False, True],
+        [False, True, True, False, True],
+    ],
+)
+def test_write_bools_sliced(lmdb_version_store_tiny_segment, data):
```
IvoDD, line 167: It would be good to also test more than 8 bools to see the bitset unpacking in action.
It would also be good to test more than 8 bools spanning across several record batches

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -0,0 +1,757 @@
...
+    sym = "test_write_bools_sliced"
+    table = pa.table({"col": pa.array(data)})
+    lib.write(sym, table)
+    received = lib.read(sym).data
+    assert table.equals(received)
+
+
+@pytest.mark.parametrize("num_rows", [1, 2, 3, 4, 5])
+@pytest.mark.parametrize("num_cols", [1, 2, 3, 4, 5])
+def test_write_sliced_with_index(lmdb_version_store_tiny_segment, num_rows, num_cols):
```
IvoDD, line 180: It would be good to test slicing with more than one record batch. I.e. having a test which tests:
- slicing one record batch into many slices
- slicing several record batches into one slice
- have record batch starting mid slice covering an entire slice and finishing mid slice

Can probably be combined in one test with the bools with slicing where we e.g. have parameterization:
```python
@pytest.mark.parametrize("row_slice_size", [1, 2, 10, 100_000])
@pytest.mark.parametrize("type", [pa.int32, pa.float64, pa.bool])
def test_many_record_batches_many_slices():
     record_batch_sizes = [1, 2, 1, 15, 13, 2, 1, 10]
```

I know the hypothesis tests similar things although it doesn't have different sized record batches and in general I prefer having a tiny non hypothesis test for something like this which can always catch an issue in an edge case rather than only 50% time catching the issue.

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -0,0 +1,757 @@
...
+    assert expected.equals(received)
+
+
+def test_write_owned_and_non_owned_buffers(lmdb_version_store_tiny_segment):
+    lib = lmdb_version_store_tiny_segment
+    lib.set_output_format("experimental_arrow")
+    lib._set_allow_arrow_input()
+    sym = "test_write_owned_and_non_owned_buffers"
+    rb0 = pa.RecordBatch.from_arrays([pa.array([0, 1, 2], pa.int32())], names=["col"])
+    rb1 = pa.RecordBatch.from_arrays([pa.array([3, 4, 5], pa.int32())], names=["col"])
```
IvoDD, line 268: Was the intention to make one of the record batches contain a view rather than an array?

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -0,0 +1,757 @@
...
+    append_table = pa.table({"col": pa.array([2, 3], pa.int64())})
+    lib.append(sym, append_table)
+
+    received = lib.read(sym).data
+    expected = pa.concat_tables([write_table, append_table]) if existing_data else append_table
+    assert expected.equals(received)
+
+
+@pytest.mark.parametrize("existing_data", [True, False])
+def test_append_with_index(lmdb_version_store_arrow, existing_data):
```
IvoDD, line 392: I think it would be good to parameterize one test for `append`, `update`, `stage` on smaller row sizes and on multiple record batches.
Or add a similar hypothesis test for other operations than writes.

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -0,0 +1,757 @@
...
+    lib.batch_append(["sym0", "sym1", "sym2"], [table_0, df_1, table_2], index_column_vector=[None, None, "ts"])
+    received_0 = lib.read("sym0").data
+    assert pa.concat_tables([table_0, table_0]).equals(received_0)
+    received_1 = lib.read("sym1", output_format="pandas").data
+    expected_1 = pd.concat([df_1, df_1])
+    expected_1.index = pd.RangeIndex(len(expected_1))
+    assert_frame_equal(expected_1, received_1)
+    received_2 = lib.read("sym2").data
+    assert table_2.equals(received_2)
+
```
IvoDD, line 687: Maybe also one `batch_update` test

cpp/proto/arcticc/pb2/descriptors.proto
IvoDD: We will also need to think about writing pyarrow-backed pandas.

cpp/arcticdb/arrow/test/test_arrow_write.cpp
```diff
@@ -0,0 +1,242 @@
...
+        if constexpr (std::is_same_v<TypeParam, bool>) {
+            ASSERT_EQ(*col.scalar_at<TypeParam>(idx), idx % 3 == 0);
+        } else {
+            ASSERT_EQ(*col.scalar_at<TypeParam>(idx), idx);
+        }
+    }
+    const auto& buffer = col.data().buffer();
+    ASSERT_EQ(buffer.bytes(), total_rows * sizeof(TypeParam));
+    if constexpr (std::is_same_v<TypeParam, bool>) {
+        ASSERT_EQ(buffer.blocks().size(), 1);
```
IvoDD, line 169: Would be nice to assert the allocated size of the bool buffer here. To ensure we don't overallocate 4KB for a few lines

cpp/arcticdb/arrow/test/benchmark_arrow_writes.cpp
```diff
@@ -0,0 +1,79 @@
...
+        ->Args({10'000'000, 10, -1})
+        ->Args({10'000'000, 10, 0})
+        ->Args({10'000'000, 10, 5});
+
+BENCHMARK(BM_arrow_write_multiple_record_batches)
+        // Short and wide data
+        ->Args({10, 100'000, 10})
+        // Long and thin data
+        ->Args({10'000'000, 10, 100})
+        ->Args({10'000'000, 10, 10'000});
```
IvoDD: Maybe also the edge case of very few rows per record batch for completeness (e.g. 1 or 10)?

cpp/arcticdb/arrow/test/benchmark_arrow_writes.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <arcticdb/arrow/test/arrow_test_utils.hpp>
+#include <arcticdb/arrow/arrow_utils.hpp>
+#include <arcticdb/column_store/memory_segment.hpp>
+
+using namespace arcticdb;
+
+// run like: --benchmark_time_unit=ms --benchmark_filter=.* --benchmark_min_time=5x
+
+static void BM_arrow_write_single_record_batch(benchmark::State& state) {
```
vasil-pashov: nit: I think the names of the tests are a bit misleading. It's not actually writing anything it's converting record batch to SegmentInMemory.

cpp/arcticdb/arrow/test/test_arrow_write.cpp
```diff
@@ -0,0 +1,242 @@
...
+    } else {
+        ASSERT_EQ(buffer.blocks().size(), rows_per_batch.size());
```
vasil-pashov, line 172: Is it possible and does it make sense to flatten the record batch vector so that we don't allocate a block per batch?

cpp/arcticdb/arrow/test/test_arrow_write.cpp
```diff
@@ -0,0 +1,242 @@
...
+    std::vector<sparrow::record_batch> record_batches;
+    record_batches.emplace_back(std::move(record_batch));
+    auto seg = arrow_data_to_segment(record_batches).first;
+
+    auto num_columns = numeric_data_types.size();
+    ASSERT_EQ(seg.fields().size(), num_columns);
+    ASSERT_EQ(seg.num_columns(), num_columns);
+    ASSERT_EQ(seg.row_count(), num_rows);
+    for (auto [idx, data_type] : folly::enumerate(numeric_data_types)) {
+        ARCTICDB_UNUSED const auto column_index = seg.column_index(fmt::format("{}", data_type));
```
vasil-pashov: Why is this mark as unused?

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -98,4 +102,139 @@ std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(Segmen
...
+            uint64_t(0),
+            [](const uint64_t& accum, const sparrow::record_batch& record_batch) {
+                return accum + record_batch.nb_rows();
+            }
+    );
+    std::vector<ChunkedBuffer> chunked_buffers(record_batch->nb_columns());
+    uint64_t start_row{0};
+    for (; record_batch != record_batches.cend(); ++record_batch) {
+        for (size_t idx = 0; idx < record_batch->nb_columns(); ++idx) {
+            auto& chunked_buffer = chunked_buffers.at(idx);
```
vasil-pashov: nit: std::vector::at performs bounds checking both in debug and release builds, while operator[] performs bounds checking only in debug. I just wanted to make sure that this is intentional.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -98,4 +102,139 @@ std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(Segmen
...
+            const auto& data_type = data_types.at(idx);
+            const auto& array = record_batch->get_column(idx);
+            auto [arrow_array, arrow_schema] = sparrow::get_arrow_structures(array);
+            schema::check<ErrorCode::E_UNSUPPORTED_COLUMN_TYPE>(
+                    arrow_array->null_count == 0,
+                    "Column '{}' contains null values, which are not currently supported",
+                    record_batch->names()[idx]
+            );
+            auto arrow_array_buffers = sparrow::get_arrow_array_buffers(*arrow_array, *arrow_schema);
+            // arrow_array_buffers.at(0) seems to be the validity bitmap, and may be NULL if there are no null values
```
vasil-pashov, line 357: I'm a bit confused because we don't handle `arrow_array_buffers.at(0)` at all. Is this something that's going to be done in future?

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
vasil-pashov, line 1: I think it's also worth having tests for combining pandas and arrow writes and reading as padnas/arrow.

vasil-pashov (approved):

cpp/arcticdb/arrow/arrow_utils.hpp
alexowens90: Discussed in person - looking up columns by name is not O(1), and we need to do the O(n) iteration here anyway

python/arcticdb/version_store/_normalization.py
alexowens90: Not O(1)

python/arcticdb/version_store/_normalization.py
alexowens90: Same

cpp/proto/arcticc/pb2/descriptors.proto
alexowens90: Even people not using the feature get the new protobuf - so we need to get it right first time, as we will not be able to change it in the future

cpp/arcticdb/pipeline/write_frame.cpp
alexowens90: Assertion inside `index_value_at`

cpp/arcticdb/stream/incompletes.cpp
alexowens90: Can also add a check that the passed in frame has tensors instead of a segment

cpp/arcticdb/version/version_core.cpp
alexowens90: At least update the docstrings that this does nothing with Arrow Tables

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: Add a comment to clarify that this is about our internal buffers, not Arrow views

cpp/arcticdb/arrow/test/test_arrow_write.cpp
alexowens90: Discussed in person - this is to keep things zero-copy

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: Not expected to work right now, but should add a test to demonstrate the behaviour at least (both ways round)

cpp/arcticdb/arrow/test/benchmark_arrow_writes.cpp
alexowens90: Fair, changed to `BM_arrow_convert_single_record_batch_to_segment` and `BM_arrow_convert_multiple_record_batches_to_segment`

cpp/arcticdb/arrow/test/test_arrow_write.cpp
alexowens90: PBKAC, fixed

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: Fair, swapped for `[]`

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: Discussed in person, this is the validity bitmap which we do not yet support.
I've checked and the `ArrowArray` and `ArrowSchema` do not seem to specify what each buffer is for.

cpp/proto/arcticc/pb2/descriptors.proto
alexowens90: I've added a test for changing index names, that already correctly raises.
For dynamic schema, I think we should follow our usual convention that a column's position is fixed the first time it is written. I've updated the [index column positioning ticket](https://man312219.monday.com/boards/7852509418/pulses/18042073623) to take care of this

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: Discussed in person - we should be able to use the `ExtraBuffer` pattern in `Column` to pass the packed bitsets through to `WriteToSegmentTask` zero-copy, and then perform the expansion in parallel there. I've updated the [task](https://man312219.monday.com/boards/7852509418/pulses/18053251438) description with this plan as well

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: See other comment for why column reordering is happening here
https://github.com/man-group/ArcticDB/pull/2663#discussion_r2398797873

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: `test_write_owned_and_non_owned_buffers` covers the first point (and runs clean under valgrind)

cpp/arcticdb/stream/incompletes.cpp
alexowens90: Agree on rename and check.
Discussed in person - correct fix is to use `WriteToSegmentTask` everywhere instead of this hacky semi-duplicated implementation

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: Covered in `test_arrow_api.py` as V1 API does not have a `batch_update` method

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: I've updated the hypothesis test to call write, append, and update.

cpp/arcticdb/arrow/test/benchmark_arrow_writes.cpp
```diff
@@ -61,19 +61,21 @@ static void BM_arrow_write_multiple_record_batches(benchmark::State& state) {
...
 }
 
-BENCHMARK(BM_arrow_write_single_record_batch)
+BENCHMARK(BM_arrow_convert_single_record_batch_to_segment)
         // Short and wide data
         ->Args({10, 100'000, -1})
         ->Args({10, 100'000, 0})
         ->Args({10, 100'000, 50'000})
-        // Long and thin data
+        // Long and thin data``````````````¬¬¬``¬¬¬¬``````
```
IvoDD: Nit: Typo?

PR #2667, opened 2025-09-23
Add exception when trying to write an index or a column with a boolean name

G-D-Petrov:
#### Reference Issues/PRs
Monday ref: 9911125848

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_write.py
+++ b/python/tests/unit/arcticdb/version_store/test_write.py
@@ -191,3 +191,21 @@ def test_write_bool_named_index(lmdb_version_store, idx):
 
     assert lmdb_version_store.list_symbols() == []
     assert lmdb_version_store.has_symbol(symbol) is False
+
+
+@pytest.mark.parametrize(
+    "idx", [pd.date_range(pd.Timestamp("2020-01-01"), periods=3), pd.RangeIndex(start=0, stop=3, step=1)]
+)
+@pytest.mark.parametrize("idx_names", [["index", True], [True, "index"]])
+def test_write_bool_named_multi_index(lmdb_version_store, idx, idx_names):
+    symbol = "bad_write"
+
+    df = pd.DataFrame({"col": [1, 2, 3]}, index=pd.MultiIndex.from_arrays([idx, idx], names=idx_names))
+
+    lmdb_version_store.write(symbol, df)
+
+    # We do allow for the boolean index names in multiindex and they get normalized to strings
+    # so this just tests the current behaviour and that we can read back the data correctly
+    df.index.names = [str(n) for n in idx_names]
+
+    assert_frame_equal(lmdb_version_store.read(symbol).data, df)
--- a/python/arcticdb/flattener.py
+++ b/python/arcticdb/flattener.py
@@ -22,7 +22,7 @@ except ImportError:
 from arcticdb import _msgpack_compat
 from arcticdb.log import version as log
 from arcticdb.version_store._custom_normalizers import get_custom_normalizer
-from arcticdb.version_store._normalization import MsgPackNormalizer, CompositeNormalizer, _check_valid_name
+from arcticdb.version_store._normalization import MsgPackNormalizer, CompositeNormalizer
 
 
 class Flattener:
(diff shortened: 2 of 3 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -498,6 +492,14 @@ def _denormalize_columns(item, norm_meta, idx_type, n_indexes):
     return columns, denormed_columns, data
 
 
+def _check_valid_name(name):
+    # bools are a subclass of int, so we need to check for them explicitly
+    if isinstance(name, bool) or not isinstance(name, (str, int)):
```
vasil-pashov, line 497: I was today years old when I learned that this is possible `isinstance(name, (str, int))`

python/arcticdb/version_store/_normalization.py
```diff
@@ -498,6 +492,14 @@ def _denormalize_columns(item, norm_meta, idx_type, n_indexes):
     return columns, denormed_columns, data
 
 
+def _check_valid_name(name):
```
vasil-pashov, line 495: Can you check if this is called when recursive/custom normalizers are used?

python/tests/unit/arcticdb/version_store/test_write.py
```diff
@@ -160,3 +160,34 @@ def test_write_only_nan_column(self, lmdb_version_store, dtype):
         lib.write(sym, pd.DataFrame({"a": [np.nan]}, dtype=dtype))
         data = lib.read(sym).data
         assert_frame_equal(data, pd.DataFrame({"a": [np.nan]}, dtype=dtype))
+
```
vasil-pashov, line 163: Does it make sense to test append/update. I guess the interesting case is with dynamic schema

initial
idx col
0   1

append
idx True
1   2

python/arcticdb/version_store/_normalization.py
G-D-Petrov: Me too :)

python/tests/integration/arcticdb/test_update.py
```diff
@@ -555,3 +557,21 @@ def test_update_batch_different_updates_dynamic_schema(custom_library):
         ArcticSymbolSimulator.assert_frame_equal_rebuild_index_first(
             expected_results[result.symbol], read_data[result.symbol].data
         )
+
+
+@pytest.mark.parametrize(
+    "idx", [pd.date_range(pd.Timestamp("2020-01-01"), periods=3), pd.RangeIndex(start=0, stop=3, step=1)]
```
alexowens90, line 563: nit: can't update with a range index anyway

python/tests/unit/arcticdb/version_store/test_write.py
```diff
@@ -160,3 +160,34 @@ def test_write_only_nan_column(self, lmdb_version_store, dtype):
...
+        lmdb_version_store.write(symbol, df)
+
+    assert lmdb_version_store.list_symbols() == []
+    assert lmdb_version_store.has_symbol(symbol) is False
+
+
+@pytest.mark.parametrize(
+    "idx", [pd.date_range(pd.Timestamp("2020-01-01"), periods=3), pd.RangeIndex(start=0, stop=3, step=1)]
+)
+def test_write_bool_named_index(lmdb_version_store, idx):
```
alexowens90, line 182: How about a multiindex where the names are `["ts", True]`?

python/tests/integration/arcticdb/test_update.py
G-D-Petrov: Yeah but the normalization path is a bit different for the 2 index types so I prefer to keep this check.
If we do manage to normalize the index, it should raise a different exception, so the test won't pass in that case.

python/tests/unit/arcticdb/version_store/test_write.py
G-D-Petrov: Similarly to recursive normalizers, we do allow this case and just [stringify the names](https://github.com/man-group/ArcticDB/blob/9169d2b701903e838b2a67ec327b599106922d82/python/arcticdb/version_store/_common.py#L145).
I've added a test for the behavior

PR #2684, opened 2025-10-01
Merge update initial implementation

vasil-pashov:
#### Reference Issues/PRs
Monday: 

This provides initial implementation of merge functionality supporting only the update part of it. It supports only matching on an ordered DatetimeIndex and static schema.

The algorithm takes advantage that both the source and the target are ordered.
1. Iterate over all slices in the index key and produce a list of object describing which of the slices can contain rows in source. This is done by performing lower_bound (binary_search) in the source index, searching for start index value stored in the slice. If the returned value is between key_start_index and key_end_index then the data segment could be affected. The complexity is `O(index_row_count * log(source_row_count))`. The information is stored as a pair, the index of the affected slice in the index key and the first index value from the source that falls into that slice.
2. Only the potentially affected data keys are read.
3. For each data key (in parallel) iterate over all index values in source that are between the first and last index values of the data key and perform lower_bound (binary search) to check if the index value from source is in the segment. If it is perform update. Complexity O(source_row_count * log(segment_size))

Next steps:
The iteration in step 3 above is row wise. This will be slow for DataFrames containing UTF string values as reading UTF strings requires holding the GIL and in general row wise iterations are not cache friendly. The reason this initial implementation uses row wise iteration is that it's easier to implement. Column wise iterations would need to either perfrom `O(slice_column_count * source_row_count * log(segment_size))` or use a caching mechanism matching source row to segment row another difficulty will be related to having the on clause. With on clause we need to check the entire row (across all segments) to know if update should be performed. The long term plan is to add additional step before `update_segment_inplace` that will iterate over all slices and generate a list of of pairs (UPDATE/INSERT, row_in_target_segment, row_in_source).

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -134,6 +134,62 @@ IndexDescriptorImpl check_index_match(const arcticdb::stream::Index& index, cons
 
     return desc;
 }
+
+void check_can_perform_merge(
+        const InputTensorFrame& source, const index::IndexSegmentReader& target_index_segment_reader,
+        const MergeStrategy& strategy, std::span<const std::string> on, bool match_on_timeseries_index,
+        bool dynamic_schema
+) {
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            strategy.matched == MergeAction::UPDATE && strategy.not_matched_by_target == MergeAction::DO_NOTHING,
+            "The only merge strategy currently supported is matched=UPDATE, not_matched_by_target=DO_NOTHING"
+    );
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            match_on_timeseries_index,
+            "Currently the set of columns to match on for a merge command must always include the index via setting "
+            "match_on_timeseries_index=True"
+    );
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            source.sorted() == SortedValue::ASCENDING,
+            "Currently the source data for a merge command must be sorted in ascending order"
+    );
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            on.empty(), "Currently the 'on' parameter of the merge command is not supported and must be empty"
+    );
+    const IndexDescriptorImpl& target_index_desc = target_index_segment_reader.tsd().index();
+    check_index_match(source.index, target_index_desc);
+    if (target_index_desc.type() == IndexDescriptor::Type::TIMESTAMP) {
+        sorting::check<ErrorCode::E_UNSORTED_DATA>(
+                target_index_segment_reader.sorted() == SortedValue::ASCENDING,
+                "When merging into a timeseries-indexed frame, the target frame must be "
+                "sorted in ascending order"
+        );
+    }
+    fix_descriptor_mismatch_or_throw(MERGE, dynamic_schema, target_index_segment_reader, source, false);
+}
+
+struct SliceAffectedByMerge {
+    /// Index in [0;source_row_count) of the first index value that falls into the range of the key
+    ssize_t first_source_index_position;
+    /// The index of the slice in the index segment reader
+    size_t slice_index;
+};
+
+std::vector<SliceAffectedByMerge> slices_affected_by_merge(
+        const InputTensorFrame& source, std::span<const SliceAndKey> slices
+) {
+    std::vector<SliceAffectedByMerge> affected_slices;
+    std::span<const timestamp> source_index = source.index_tensor->as_span<timestamp>();
+    for (size_t slice_index = 0; slice_index < slices.size(); slice_index++) {
+        const timestamp start = slices[slice_index].key().start_time();
+        const timestamp end = slices[slice_index].key().end_time();
+        if (const auto lower = ranges::lower_bound(source_index, start); lower != source_index.end() && *lower < end) {
+            const ssize_t first_source_index_position = lower - source_index.begin();
+            affected_slices.emplace_back(first_source_index_position, slice_index);
+        }
+    }
+    return affected_slices;
+}
 } // namespace
 
 void sorted_data_check_append(const InputTensorFrame& frame, index::IndexSegmentReader& index_segment_reader) {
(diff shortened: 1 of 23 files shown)


python/arcticdb/version_store/library.py
```diff
@@ -3021,3 +3026,28 @@ def _normalize_staged_data_mode(mode: Union[StagedDataFinalizeMethod, str]) -> S
...
+    def merge(
+        self,
+        symbol: str,
+        dataframe: NormalizableType,
+        strategy: MergeStrategy,
+        prune_previous_versions: bool = False,
+        metadata: Optional[Any] = None,
+        upsert: bool = False,
+    ) -> VersionedItem:
+        udm, item, norm_meta = self._nvs._try_normalize(
```
jamesblackburn, line 3039: + docs :)

python/arcticdb/version_store/library.py
```diff
@@ -3021,3 +3026,28 @@ def _normalize_staged_data_mode(mode: Union[StagedDataFinalizeMethod, str]) -> S
             raise ArcticInvalidApiUsageException(
                 "mode must be one of StagedDataFinalizeMethod.WRITE, StagedDataFinalizeMethod.APPEND, 'write', 'append'"
             )
+
+    def merge(
+        self,
+        symbol: str,
+        dataframe: NormalizableType,
```
jamesblackburn, line 3033: Parameter `data` to match `update` method?

python/arcticdb/version_store/library.py
```diff
@@ -393,6 +393,11 @@ def __repr__(self):
         )
 
 
+class MergeStrategy(NamedTuple):
```
jamesblackburn, line 396: What are the expected semantics of UPDATE vs INSERT? Just thinking of how this will be used, isn't `upsert` always what's wanted i.e. add missing, or update what's there?

python/arcticdb/version_store/library.py
```diff
@@ -3021,3 +3026,28 @@ def _normalize_staged_data_mode(mode: Union[StagedDataFinalizeMethod, str]) -> S
             raise ArcticInvalidApiUsageException(
                 "mode must be one of StagedDataFinalizeMethod.WRITE, StagedDataFinalizeMethod.APPEND, 'write', 'append'"
             )
+
+    def merge(
```
jamesblackburn, line 3030: Should we have another method on this, vs making this a feature of `update`? 
Update already has an `upsert` parameter which many would think would have upsert semantics.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -134,6 +134,62 @@ IndexDescriptorImpl check_index_match(const arcticdb::stream::Index& index, cons
...
+
+struct SliceAffectedByMerge {
+    /// Index in [0;source_row_count) of the first index value that falls into the range of the key
+    ssize_t first_source_index_position;
+    /// The index of the slice in the index segment reader
+    size_t slice_index;
+};
+
+std::vector<SliceAffectedByMerge> slices_affected_by_merge(
+        const InputTensorFrame& source, std::span<const SliceAndKey> slices
```
alexowens90, line 179: This will need the `match_on_timeseries_index' parameter as well, as we will always need to read everything when not matching on the index

cpp/arcticdb/version/version_core.cpp
```diff
@@ -548,6 +604,189 @@ folly::Future<AtomKey> async_update_impl(
...
+                target.mutable_tsd().set_user_metadata(std::move(source->user_meta));
+                return std::make_shared<SerializedIndex>(std::vector(target.begin(), target.end()), target.tsd());
+            })
+            .thenValue([&](std::shared_ptr<SerializedIndex>&& index) {
+                const std::vector<SliceAffectedByMerge> affected_slices =
+                        slices_affected_by_merge(*source, index->slice_and_keys);
+                std::vector<folly::Future<folly::Unit>> merge_segments_fut;
+                merge_segments_fut.reserve(affected_slices.size());
+                for (const SliceAffectedByMerge& affected : affected_slices) {
+                    merge_segments_fut.emplace_back(
```
alexowens90, line 743: This needs to be windowed to guarantee we don't read everything into memory at once

cpp/arcticdb/version/version_core.cpp
```diff
@@ -548,6 +604,189 @@ folly::Future<AtomKey> async_update_impl(
...
+                                                *source,
+                                                affected_slice,
+                                                affected.first_source_index_position
+                                        );
+                                        return std::move(key_segment);
+                                    })
+                                    .thenValue([store,
+                                                &update_info](std::pair<VariantKey, SegmentInMemory>&& key_segment) {
+                                        const AtomKey& key = std::get<AtomKey>(key_segment.first);
+                                        return store->write(
```
alexowens90, line 759: What about slicing? We can be inserting lots of rows, so could end up with a massive segment on disk

cpp/arcticdb/version/version_core.cpp
```diff
@@ -548,6 +604,189 @@ folly::Future<AtomKey> async_update_impl(
...
+                                                key.type(),
+                                                update_info.next_version_id_,
+                                                key.id(),
+                                                key.start_index(),
+                                                key.end_index(),
+                                                std::move(key_segment.second)
+                                        );
+                                    })
+                                    .thenValueInline([index, affected](VariantKey&& key) {
+                                        index->slice_and_keys[affected.slice_index].set_key(
```
alexowens90, line 769: The rowcount will now be wrong in the `SliceAndKey`?
See above comment on slicing though, I don't think this design of modifying the existing index will work in that case, better to just build up a new index key, more like `update`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -548,6 +604,189 @@ folly::Future<AtomKey> async_update_impl(
...
+        size_t row_in_source
+) {
+    std::span<const timestamp> source_index = source.index_tensor->as_span<timestamp>();
+    const Column& target_index = target.column(0);
+    using IndexType = ScalarTagType<DataTypeTag<DataType::NANOSECONDS_UTC64>>;
+    auto target_index_search_start = target_index.begin<IndexType>();
+    const auto target_index_end = target_index.end<IndexType>();
+    while (row_in_source < source.num_rows) {
+        const timestamp source_index_value = source_index[row_in_source];
+        if (slice_to_update.key().end_time() <= source_index_value) {
```
alexowens90, line 623: Using atom key end times is error prone due to various bugs we've had in the past, use the segment in memory as the source of truth whenever it has already been read

cpp/arcticdb/version/version_core.cpp
```diff
@@ -548,6 +604,189 @@ folly::Future<AtomKey> async_update_impl(
...
+            break;
+        }
+
+        auto target_row_it = std::lower_bound(target_index_search_start, target_index_end, source_index_value);
+        while (target_row_it != target_index_end && *target_row_it == source_index_value) {
```
alexowens90, line 628: Let's discuss this algorithm on a call. There are pros and cons to whatever we choose to do for index.
We should also design this with `match_on_index_column=false` and `!on.empty()` now as well, as they could both cause large changes to how this looks.

python/arcticdb/version_store/library.py
vasil-pashov: I guess upsert will be the most commonly used. Update and insert are more or less orthogonal and can be implemented separately. Thus we can present an option to the user to do just one operation.

python/arcticdb/version_store/library.py
vasil-pashov: Yeah, I find the semantics of the upsert parameter to be a bit confusing I guess the word got overloaded too much. Technically it's possible to add some flags and have this functionality as part of the update. I think the two things are doing quite different things and the implementation will be vastly different so I'm not quite convinced it'll be for the better to mash them into one API call.

PR #2685, opened 2025-10-01
Build speedup

alexowens90:


cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -7,16 +7,182 @@
...
 #include <arcticdb/arrow/arrow_utils.hpp>
-#include <arcticdb/arrow/array_from_block.hpp>
 #include <arcticdb/column_store/memory_segment.hpp>
 
+#include <sparrow/layout/primitive_data_access.hpp>
 #include <sparrow/record_batch.hpp>
 
 namespace arcticdb {
 
+inline std::optional<sparrow::validity_bitmap> create_validity_bitmap(
```
IvoDD: Nit: Now that it is in `cpp` it doesn't need to be `inline`?

IvoDD: Same thing for a few others in this file

IvoDD (approved):

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: Yup, copy-paste fail

PR #2691, opened 2025-10-03
[9898189624] Fix arrow reading empty frames

IvoDD:
#### Reference Issues/PRs
Fixes monday ref: 9898189624

During symbol concat we can end up with a Segment with zero columns. Convert that to arrow gracefully.

Also uncovered that arrow normalization doesn't correctly construct `pandas_metadata` for empty dataframes. This is also fixed and tested in this PR.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/arrow/arrow_utils.cpp
+++ b/cpp/arcticdb/arrow/arrow_utils.cpp
@@ -229,6 +229,11 @@ std::vector<sparrow::array> arrow_arrays_from_column(const Column& column, std::
 std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(SegmentInMemory& segment) {
     const auto total_blocks = segment.num_blocks();
     const auto num_columns = segment.num_columns();
+    if (num_columns == 0) {
+        // We can't construct a record batch with no columns, so in this case we return an empty list of record batches,
+        // which needs special handling in python.
+        return {};
+    }
     const auto column_blocks = segment.column(0).num_blocks();
     util::check(total_blocks == column_blocks * num_columns, "Expected regular block size");
 
--- a/cpp/arcticdb/arrow/arrow_utils.cpp
+++ b/cpp/arcticdb/arrow/arrow_utils.cpp
@@ -232,7 +232,7 @@ std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(Segmen
     if (num_columns == 0) {
         // We can't construct a record batch with no columns, so in this case we return an empty list of record batches,
         // which needs special handling in python.
-        return {};
+        return std::make_shared<std::vector<sparrow::record_batch>>();
     }
     const auto column_blocks = segment.column(0).num_blocks();
     util::check(total_blocks == column_blocks * num_columns, "Expected regular block size");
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -63,6 +63,11 @@ std::vector<sparrow::array> arrow_arrays_from_column(const Column& column, std::
 std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(SegmentInMemory& segment) {
     const auto total_blocks = segment.num_blocks();
     const auto num_columns = segment.num_columns();
+    if (num_columns == 0) {
+        // We can't construct a record batch with no columns, so in this case we return an empty list of record batches,
+        // which needs special handling in python.
+        return {};
```
vasil-pashov: I think the comment is a bit misleading. This actually returns shared pointer wrapping null i.e. std::shared_ptr<std::vector...> {nullptr} and not an empty list.

python/arcticdb/version_store/_normalization.py
```diff
@@ -742,7 +742,9 @@ def denormalize(self, item, norm_meta):
         index_type = pandas_meta.WhichOneof("index_type")
         if index_type == "index":
             index_meta = pandas_meta.index
-            if index_meta.is_physically_stored:
+            # Empty tables don't have `is_physically_stored=True` but we still output them with an empty DateTimeIndex.
+            is_empty_table_with_datetime_index = len(item) == 0 and not index_meta.step
```
vasil-pashov, line 746: Wouldn't it be better to store is_physically_stored in this case rather than trying to infer it from other fields?

IvoDD: I agree but unfortunately we can't change it now. Old writers will still have `is_physically_stored=false` for empty dataframes.

Also even if we did change it, it would break old readers reading new data.

cpp/arcticdb/arrow/arrow_utils.cpp
IvoDD: Ah thanks. I actually intended to return an emtpy list. It doesn't matter in this case because it either way get's converted to an empty list [here](https://github.com/man-group/ArcticDB/blob/7acd347e3a035b635fd71c1bde2b54e0135ac8db/cpp/arcticdb/arrow/arrow_output_frame.cpp#L26-L29).

Anyway updated to return an empty list.

python/arcticdb/version_store/_normalization.py
vasil-pashov: Ah, true.

vasil-pashov (approved):

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -63,6 +63,11 @@ std::vector<sparrow::array> arrow_arrays_from_column(const Column& column, std::
 std::shared_ptr<std::vector<sparrow::record_batch>> segment_to_arrow_data(SegmentInMemory& segment) {
     const auto total_blocks = segment.num_blocks();
     const auto num_columns = segment.num_columns();
+    if (num_columns == 0) {
+        // We can't construct a record batch with no columns, so in this case we return an empty list of record batches,
+        // which needs special handling in python.
+        return std::make_shared<std::vector<sparrow::record_batch>>();
+    }
```
alexowens90, line 236: Snap
https://github.com/man-group/ArcticDB/pull/2663/files#diff-8b01d82fb580cb2019d961fd12a93cc3249e3b9c99b218ea64d523d331dc468bR233-R235

alexowens90 (approved):

PR #2699, opened 2025-10-09
Performance 18146257858: speedup reading as_of tombstoned versions alive in snapshots

alexowens90:
#### Reference Issues/PRs
[18146257858](https://man312219.monday.com/boards/7852509418/pulses/18146257858)

Speeds up reading as_of a specific version number where the version number has been deleted, but is alive in at least one snapshot (note that this is only supported in the V1 API, we made a deliberate choice for this to throw a NoSuchVersionException in the V2 API as the performance can be so bad).

The main performance improvement comes from parallelising and short-circuiting the reading of snapshot ref keys read from storage. The previous implementation read all of the snapshot ref keys, and then searched through for index keys with the correct symbol and version ids. The new implementation reads snapshot ref keys in parallel, and stops reading snapshot ref keys once it finds an index key with the correct symbol and version id.

There are also some minor improvements to only materialise the atom key in memory of the desired symbol/version pair, rather than all index keys from all snapshots with the correct symbol, although in practice this should be dwarfed by the IO improvements.

### DIFF
--- a/cpp/arcticdb/version/snapshot.cpp
+++ b/cpp/arcticdb/version/snapshot.cpp
@@ -128,7 +128,7 @@ void iterate_snapshots(const std::shared_ptr<Store>& store, folly::Function<void
 }
 
 std::optional<size_t> row_id_for_stream_in_snapshot_segment(
-        SegmentInMemory& seg, bool using_ref_key, const StreamId& stream_id
+        SegmentInMemory& seg, bool using_ref_key, const StreamId& stream_id, const std::optional<VersionId> version_id
 ) {
     if (using_ref_key) {
         // With ref keys we are sure the snapshot segment has the index atom keys sorted by stream_id.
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/version/snapshot.cpp
```diff
@@ -187,6 +187,82 @@ std::unordered_set<entity::AtomKey> get_index_keys_in_snapshots(
     return index_keys_in_snapshots;
 }
 
+std::optional<AtomKey> index_key_for_stream_and_version_in_snapshot_segment(
+        SegmentInMemory& seg, bool using_ref_key, const StreamId& stream_id, VersionId version_id
+) {
+    // Logic is very similar to row_id_for_stream_in_snapshot_segment
```
IvoDD: We can maybe decrease the code duplication if we make `row_id_for_stream_in_snapshot_segment` take an optional version_id and use it here .e.g.:

```
std::optional<size_t> row_id_for_stream_in_snapshot_segment(
        SegmentInMemory& seg, bool using_ref_key, const StreamId& stream_id, std::optional<VersionId> version_id
) {
    // Do the same logic as here but only return the row_id
}

std::optional<AtomKey> index_key_for_stream_and_version_in_snapshot_segment(
        SegmentInMemory& seg, bool using_ref_key, const StreamId& stream_id, VersionId version_id
) {
     auto row_id = row_id_for_stream_in_snapshot_segment(seg, using_ref_key, stream_id, {version_id});
     if (row_id.has_value) {
          return read_key_row(row_id.value())
     }
     return std::nullopt
}
```

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -187,6 +187,82 @@ std::unordered_set<entity::AtomKey> get_index_keys_in_snapshots(
...
+        const std::shared_ptr<Store>& store, const StreamId& stream_id, VersionId version_id
+) {
+    std::vector<VariantKey> snapshot_keys;
+    iterate_snapshots(store, [&snapshot_keys](auto&& snapshot_key) {
+        snapshot_keys.emplace_back(std::move(snapshot_key));
+    });
+    std::optional<AtomKey> res;
+    std::mutex res_mutex;
+    std::atomic<bool> found{false};
+    const auto window_size = async::TaskScheduler::instance()->io_thread_count();
```
IvoDD, line 217: Maybe better to use `cpu_thread_count` because `store->read` will use IO pool for reading and CPU pool for decoding. So if we have too low cpu count that will be the bottleneck and we'll be spending time waiting in queues for cpu pool

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -187,6 +187,82 @@ std::unordered_set<entity::AtomKey> get_index_keys_in_snapshots(
...
+                                auto snapshot_segment = std::move(key_seg.second);
+                                auto opt_res = index_key_for_stream_and_version_in_snapshot_segment(
+                                        snapshot_segment,
+                                        variant_key_type(snapshot_key) == KeyType::SNAPSHOT_REF,
+                                        stream_id,
+                                        version_id
+                                );
+                                if (opt_res.has_value()) {
+                                    found.store(true);
+                                    std::lock_guard<std::mutex> lock(res_mutex);
```
IvoDD: I think we can remove the mutex and a use lock-free `compare_exchange_strong` on the `found`:
```
if (found.compare_exchange_strong(false, true)) {
    res = std::move(opt_res);
}
```

I think this will be faster and we wouldn't end up potentially writing to the res entry multiple times.

cpp/arcticdb/version/snapshot.cpp
alexowens90: By default, decodes actually happen in the IO threadpool, to prevent context switching between threads in the 2 pools.

IvoDD: Ah was confused by encoding code which calls `submit_cpu_task` [here](https://github.com/man-group/ArcticDB/blob/7acd347e3a035b635fd71c1bde2b54e0135ac8db/cpp/arcticdb/async/async_store.hpp#L77). Decoding is different

PR #2707, opened 2025-10-15
Delete written data keys when storage quota is reached.

ognyanstoimenov:
#### Reference Issues/PRs

Monday ref 9470086469
Implements rolling back any written data keys during a write/append/update if storage quota has been hit.

This quota exception handling is only implemented for data keys, so any quota-related failure on other key types is not covered.
Future work not covered in this PR:
* Test against real VAST storage as it is currently the only supported one that has quotas
* Implement the rollback mechanism when working with recursively normalized data.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/pipeline/test/test_rollback.cpp
+++ b/cpp/arcticdb/pipeline/test/test_rollback.cpp
@@ -126,8 +126,11 @@ class RollbackOnQuotaExceeded : public ::testing::TestWithParam<TestScenario> {
         StorageFailureSimulator::instance()->configure({{FailureType::DELETE, scenario.delete_failures}});
 
         if (scenario.single_thread) {
-            ConfigsMap::instance()->set_int("VersionStore.NumCPUThreads", 1);
-            ConfigsMap::instance()->set_int("VersionStore.NumIOThreads", 1);
+            std::vector<std::pair<std::string, std::optional<int64_t>>> configs = {
+                    {{"VersionStore.NumCPUThreads", 1}, {"VersionStore.NumIOThreads", 1}}
+            };
+            scoped_config_ = std::make_unique<ScopedConfig>(std::move(configs));
+            async::TaskScheduler::instance()->reattach_instance();
         }
         proto::storage::VersionStoreConfig version_store_cfg;
         version_store_cfg.mutable_write_options()->set_column_group_size(100);
@@ -137,10 +140,15 @@ class RollbackOnQuotaExceeded : public ::testing::TestWithParam<TestScenario> {
         version_store_ = std::make_unique<version_store::PythonVersionStore>(std::move(version_store));
     }
 
-    void TearDown() override { StorageFailureSimulator::instance()->reset(); }
+    void TearDown() override {
+        StorageFailureSimulator::instance()->reset();
+        scoped_config_.reset();
+        async::TaskScheduler::instance()->reattach_instance();
+    }
 
     std::unique_ptr<version_store::PythonVersionStore> version_store_;
     StreamId stream_id_{"sym"};
+    std::unique_ptr<ScopedConfig> scoped_config_;
 };
 
 class RollbackOnQuotaExceededUpdateOrAppend : public RollbackOnQuotaExceeded {
@@ -152,8 +160,11 @@ class RollbackOnQuotaExceededUpdateOrAppend : public RollbackOnQuotaExceeded {
         version_store_cfg.mutable_write_options()->set_column_group_size(100);
         version_store_cfg.mutable_write_options()->set_segment_row_size(10);
         if (scenario.single_thread) {
-            ConfigsMap::instance()->set_int("VersionStore.NumCPUThreads", 1);
-            ConfigsMap::instance()->set_int("VersionStore.NumIOThreads", 1);
+            std::vector<std::pair<std::string, std::optional<int64_t>>> configs = {
+                    {{"VersionStore.NumCPUThreads", 1}, {"VersionStore.NumIOThreads", 1}}
+            };
+            scoped_config_ = std::make_unique<ScopedConfig>(std::move(configs));
+            async::TaskScheduler::instance()->reattach_instance();
         }
         auto [version_store, _] = python_version_store_in_memory(version_store_cfg);
         version_store_ = std::make_unique<version_store::PythonVersionStore>(std::move(version_store));
@@ -290,8 +301,8 @@ const auto TEST_DATA_UPDATE = {
                 .check_data_keys = false,
                 .expected_written_data_keys = 0,
                 .num_writes = 1,
-                // If either of the rewrites (before or after) throws non-quota exception while the other throws quota,
-                // it is undefined which exception is propagated
+                // If either of the rewrites (before or after) throws non-quota exception while the other throws
+                // quota, it is undefined which exception is propagated
                 .write_expected_outcome = {UNKNOWN_EXCEPTION},
                 .operation = Operation::UPDATE,
         },


cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -6,6 +6,8 @@
  * will be governed by the Apache License, version 2.0.
  */
 
+#include "async/tasks.hpp"
```
vasil-pashov:
```suggestion
#include <arcticdb/async/tasks.hpp>
```

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -199,26 +201,23 @@ folly::Future<std::vector<SliceAndKey>> write_slices(
...
+                return async::submit_cpu_task(WriteToSegmentTask(
+                                                      frame,
+                                                      slice.first,
+                                                      slicing,
+                                                      get_partial_key_gen(frame, key),
+                                                      slice.second,
+                                                      frame->index,
+                                                      sparsify_floats
+                                              ))
+                        .then([sink, de_dup_map](auto&& ks) { return sink->async_write(ks, de_dup_map); });
```
vasil-pashov:
```suggestion
                        .then([sink, de_dup_map](auto&& ks) { return sink->async_write(std::forwad<decltype(ks)>(ks), de_dup_map); });
```

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -325,23 +329,23 @@ static RowRange partial_rewrite_row_range(
...
                         const auto& key = existing.key();
                         const SegmentInMemory& segment = key_segment.second;
                         const RowRange affected_row_range =
                                 partial_rewrite_row_range(segment, index_range, affected_part);
                         const auto num_rows = int64_t(affected_row_range.end() - affected_row_range.start());
-                        if (num_rows <= 0)
-                            return std::nullopt;
-
+                        if (num_rows <= 0) {
+                            return folly::Try<SliceAndKey>{}; // Empty future
```
vasil-pashov: How does this affect the callers of `async_rewrite_partial_segment`? It seems like the output type is changed but I don't see any changes to the code using `async_rewrite_partial_segment`.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -389,4 +393,54 @@ std::vector<SliceAndKey> flatten_and_fix_rows(
...
+            std::make_move_iterator(std::begin(slices)),
+            std::make_move_iterator(std::end(slices)),
+            std::back_inserter(keys),
+            [](SliceAndKey&& key) { return std::move(key).key(); }
+    );
+    return sink.remove_keys(std::move(keys));
+};
+
+folly::SemiFuture<std::vector<SliceAndKey>> rollback_on_quota_exceeded(
+        std::vector<folly::Try<SliceAndKey>>&& vec, const std::shared_ptr<stream::StreamSink>& sink
```
vasil-pashov: nit: Can you use a bit more descriptive name, probably slices or maybe_slices

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -389,4 +393,54 @@ std::vector<SliceAndKey> flatten_and_fix_rows(
...
+                has_quota_limit_exceeded = true;
+            }
+
+        } else if (k.hasValue()) {
+            succeeded.emplace_back(std::move(k.value()));
+        }
+    }
+
+    if (has_quota_limit_exceeded) {
+        return remove_slice_and_keys(std::move(succeeded), *sink).thenValue([exception](auto&&) {
```
vasil-pashov: I don't think `exception` is used in the lambda

vasil-pashov: So this rollbacks only on quota limit. Do we care about other kind of exceptions should it be fixed in another PR?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -251,11 +241,49 @@ using IntersectingSegments = std::tuple<std::vector<SliceAndKey>, std::vector<Sl
...
-                   collect(maybe_intersect_after_fut).via(&async::io_executor()).thenValueInline(filter_existing_slices)
+                           .thenValueInline([store](auto&& maybe_intersect_after) {
+                               return rollback_on_quota_exceeded(
+                                       std::move(maybe_intersect_after), std::static_pointer_cast<StreamSink>(store)
+                               );
+                           })
     )
+            .via(&async::cpu_executor())
+            .thenValue([store](auto&& try_before_and_after) -> folly::Future<IntersectingSegments> {
+                auto [try_intersect_before, try_intersect_after] = try_before_and_after;
```
vasil-pashov: Hmm, I think this is going to copy before and after. TBH I'm not sure if you should do 
` auto [try_intersect_before, try_intersect_after] = std::forwad<decltype(try_before_and_after)>(try_before_and_after);` or ` auto& [try_intersect_before, try_intersect_after] = try_before_and_after;`

cpp/arcticdb/pipeline/write_frame.cpp
ognyanstoimenov: Not sure how to check but I think folly always moves the result out to the argument (in this case `ks`) so this also should be okay although way uglier:
```suggestion
                        .then([sink, de_dup_map](folly::Try<std::tuple<PartialKey, SegmentInMemory, pipelines::FrameSlice>>&& ks) { return sink->async_write(std::move(ks), de_dup_map); });
```

So I'll implement your suggestion

cpp/arcticdb/version/version_core.cpp
ognyanstoimenov: Good catch! I think ` auto& [try_intersect_before, try_intersect_after] = try_before_and_after;` should be good

cpp/arcticdb/pipeline/write_frame.cpp
ognyanstoimenov: Yeah only on quota limit for now. I think if we have any other exception on the data keys it is most probably something network related so it isn't worth trying to roll back as we'd likely still get an error

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -88,6 +88,10 @@ inline bool is_not_found_error(const Aws::S3::S3Errors& error) {
             );
         }
 
+        if (err.GetExceptionName().find("InsufficientCapacity") != std::string::npos) {
```
alexowens90: Do we know what API guarantees Vast give about the HTTP response when a quota is hit? This seems like it might be fragile if they change a string

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -389,4 +393,54 @@ std::vector<SliceAndKey> flatten_and_fix_rows(
...
+    succeeded.reserve(vec.size());
+    for (auto& k : vec) {
+        if (k.hasException()) {
+            if (!exception.has_value()) {
+                exception = k.exception();
+            }
+
+            if (!has_quota_limit_exceeded && k.exception().template is_compatible_with<QuotaExceededException>()) {
+                has_quota_limit_exceeded = true;
+            }
```
alexowens90, line 421:
```suggestion
has_quota_limit_exceeded = has_quota_limit_exceeded || k.exception().template is_compatible_with<QuotaExceededException>();
```

cpp/arcticdb/pipeline/test/test_rollback.cpp
```diff
@@ -0,0 +1,430 @@
...
+    auto append_frame =
+            get_test_frame<TimeseriesIndex>(stream_id, get_test_timeseries_fields(), 30, start_index, update_val);
+    store.append_internal(stream_id, append_frame.frame_, false, false, false);
+    start_index += 30; // To avoid sameappends
+    return append_frame;
+}
+} // namespace
+
+#include <optional>
+#include <type_traits>
```
alexowens90: Move to top of file

cpp/arcticdb/pipeline/test/test_rollback.cpp
```diff
@@ -0,0 +1,430 @@
...
+        },
+        TestScenario{
+                .name = "Write_triggers_rollback_but_then_delete_fails",
+                .write_failures = make_fault_sequence({NONE, QUOTA, NONE, NONE, NONE, NONE}),
+                .delete_failures = make_fault_sequence({OTHER}),
+                .expected_written_ref_keys = 1,
+                .expected_written_version_keys = 1,
+                .expected_written_index_keys = 1,
+                .check_data_keys = false,
+                .expected_written_data_keys = 0,
```
alexowens90: I don't get why the second write doesn't succeed?

cpp/arcticdb/pipeline/test/test_rollback.cpp
```diff
@@ -0,0 +1,430 @@
...
+        },
+        TestScenario{
+                .name = "Update_triggers_rollback_but_then_delete_fails",
+                .write_failures = make_fault_sequence({NONE, QUOTA, NONE, NONE, NONE}),
+                .delete_failures = make_fault_sequence({OTHER}),
+                .expected_written_ref_keys = 0,
+                .expected_written_version_keys = 0,
+                .expected_written_index_keys = 0,
+                .check_data_keys = false,
+                .expected_written_data_keys = 0,
```
alexowens90: Won't there be data keys left in storage if the delete fails?

cpp/arcticdb/pipeline/test/test_rollback.cpp
alexowens90, line 1: Have you run these tests with the number of cpu and io threads set to 1 to check for deadlocks?

cpp/arcticdb/pipeline/test/test_rollback.cpp
ognyanstoimenov: It is expected to succeed, see `.write_expected_outcome = {OTHER, NONE}`, but I had mistakenly assumed it would flake if I check the data keys. Will fix, thanks

cpp/arcticdb/pipeline/test/test_rollback.cpp
ognyanstoimenov: Good spot again. I had missed initializing the Delete failure here. Will fix

cpp/arcticdb/pipeline/test/test_rollback.cpp
ognyanstoimenov: Added

cpp/arcticdb/storage/s3/detail-inl.hpp
ognyanstoimenov: The only documentation I managed to find is this https://support.vastdata.com/s/document-item?bundleId=vast-cluster-5.0-release-notes&topicId=vast-cluster-5-0-0-sp24-release-notes/enhancements-in-5-0-0-sp24.html&_LANG=enus which says that since VAST 5.0.0 `ORION-167022: Updated quota routines to respond with HTTP status code 507 (Insufficient Storage) if the quota limit was reached when attempting to write to an S3 bucket. Prior to this change, HTTP status code 500 was returned (Internal Server Error).`

So I think I should change this to handle that http response code

cpp/arcticdb/pipeline/test/test_rollback.cpp
alexowens90: I think you need to reinitialise the task scheduler for the threadpool changes to take effect. And you should set it back to whatever it was before in the `TearDown` methods

alexowens90 (approved):

PR #2714, opened 2025-10-16
Fix staged data corner case when using finalize with `delete_staged_data_on_failure` and update docs

ognyanstoimenov:
#### Reference Issues/PRs

Monday ref 18096936588
* Fixes a rare corner case where a throwing `finalize_staged_data` with `delete_staged_data_on_failure=True` and provided `stage_results` can result in deleting of every staged data key instead of just the provided ones. This only happens in the case where the operation throws before pipeline context is propagated
* Fixes docs for `delete_staged_data_on_failure` with `stage_results` to be clearer
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/test/test_version_store.cpp
+++ b/cpp/arcticdb/version/test/test_version_store.cpp
@@ -1089,32 +1089,26 @@ TEST(DeleteIncompleteKeysOnExit, TestDeleteIncompleteKeysOnExit) {
     auto& frame1 = wrapper1.frame_;
     auto wrapper2 = get_test_simple_frame(stream_id, 15, 2);
     auto& frame2 = wrapper2.frame_;
-    version_store.append_incomplete_frame(stream_id, frame1, true);
-    version_store.append_incomplete_frame(stream_id, frame2, true);
-    const auto staged_key_frame1_and_2 = get_staged_keys();
-    ASSERT_EQ(staged_key_frame1_and_2.size(), 2);
+    auto staged_keys_1 =
+            version_store.write_parallel_frame(stream_id, frame1, true, false, std::nullopt).staged_segments;
+    auto staged_keys_2 =
+            version_store.write_parallel_frame(stream_id, frame2, true, false, std::nullopt).staged_segments;
+    ASSERT_EQ(staged_keys_1.size(), 1);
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/version/test/test_version_store.cpp
```diff
@@ -1069,4 +1069,64 @@ TEST(VersionStore, TestWriteAppendMapHead) {
...
+    auto get_staged_keys = [store]() {
+        std::unordered_set<AtomKey> res;
+        store->iterate_type(KeyType::APPEND_DATA, [&](VariantKey&& found_key) { res.emplace(to_atom(found_key)); });
+        return res;
+    };
+    auto wrapper1 = get_test_simple_frame(stream_id, 15, 2);
+    auto& frame1 = wrapper1.frame_;
+    auto wrapper2 = get_test_simple_frame(stream_id, 15, 2);
+    auto& frame2 = wrapper2.frame_;
+    version_store.append_incomplete_frame(stream_id, frame1, true);
```
alexowens90: I think the logic could be simplified slightly if using `write_parallel_frame` rather than `append_incomplete_frame`, as that returns the `StageResult`

cpp/arcticdb/version/test/test_version_store.cpp
```diff
@@ -1069,4 +1069,64 @@ TEST(VersionStore, TestWriteAppendMapHead) {
...
+    ASSERT_EQ(staged_key_frame3.size(), 1);
+
+    StageResult result{staged_key_frame3};
+    std::optional<std::vector<StageResult>> stage_results;
+    stage_results = std::make_optional(std::vector{std::move(result)});
+
+    CompactIncompleteParameters params;
+    params.delete_staged_data_on_failure_ = true;
+    params.stage_results = stage_results;
+    version_store::get_delete_keys_on_failure(pipeline_context, store, params);
```
alexowens90: I had to stare at this for a while before realising it works because this function returns the `DeleteIncompleteKeysOnExit` which is immediately destroyed as there's no assignment. Worth a comment

alexowens90 (approved):

PR #2016, opened 2024-11-25
Add pypi python 3.13 build to the pipeline

phoebusm:
#### Reference Issues/PRs

Depends on https://github.com/man-group/ArcticDB/pull/2050. Need to rebase after #2050 is merged

#### What does this implement or fix?
Add python3.13 build to pipeline and upgrade grpc for 3.13 wheel build

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #2721, opened 2025-10-17
Add Python 3.14 support to build pipeline

fersarr:
- Add Python 3.14 to all build matrices (Linux, Windows, macOS)
- Add support for both cp314 (regular) and cp314t (free-threading) versions
- Existing protobuf/gRPC compatibility logic already covers Python 3.14
- Following the same pattern as Python 3.13 support addition

copying https://github.com/man-group/ArcticDB/pull/2016

python 3.14 build test


poodlewars:
We've added support ourselves. The free-threading build needs more care so we're going to postpone that.

PR #2727, opened 2025-10-23
Bugfix 18252295723: Allow grouping on empty columns

alexowens90:
#### Reference Issues/PRs
[18252295723](https://man312219.monday.com/boards/7852509418/pulses/18252295723)

Grouping  on empty columns used to throw, including if the input to the clause was empty because a date range deselected all rows. this allows grouping on empty columns, which returns an empty dataframe with the correct schema.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_aggregation.py
+++ b/python/tests/unit/arcticdb/version_store/test_aggregation.py
@@ -261,14 +261,14 @@ def test_group_empty_dataframe(lmdb_version_store_v1, any_output_format):
     lib = lmdb_version_store_v1
     lib._set_output_format_for_pipeline_tests(any_output_format)
     symbol = "test_group_empty_dataframe"
-    df = DataFrame({"grouping_column": [], "to_mean": np.array([], dtype=np.float64)})
+    df = DataFrame({"grouping_column": [], "to_count": []})
     lib.write(symbol, df)
-    q = QueryBuilder().groupby("grouping_column").agg({"to_mean": "mean"})
+    q = QueryBuilder().groupby("grouping_column").agg({"to_count": "count"})
     received = lib.read(symbol, query_builder=q).data
     assert not len(received)
     assert received.index.name == "grouping_column"
     assert len(received.columns) == 1
-    assert "to_mean" in received.columns
+    assert "to_count" in received.columns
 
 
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/processing/clause.cpp
```diff
@@ -339,14 +339,6 @@ AggregationClause::AggregationClause(
 std::vector<std::vector<EntityId>> AggregationClause::structure_for_processing(
         std::vector<std::vector<EntityId>>&& entity_ids_vec
 ) {
-    schema::check<ErrorCode::E_COLUMN_DOESNT_EXIST>(
-            ranges::any_of(
-                    std::as_const(entity_ids_vec),
-                    [](const std::vector<EntityId>& entity_ids) { return !entity_ids.empty(); }
-            ),
-            "Grouping column {} does not exist or is empty",
```
IvoDD, line 347: I think we should still raise if a user tries to group on a non-existent column? Is this checked elsewhere?

python/tests/unit/arcticdb/version_store/test_aggregation.py
```diff
@@ -264,8 +264,11 @@ def test_group_empty_dataframe(lmdb_version_store_v1, any_output_format):
     df = DataFrame({"grouping_column": [], "to_mean": []})
     lib.write(symbol, df)
     q = QueryBuilder().groupby("grouping_column").agg({"to_mean": "mean"})
-    with pytest.raises(SchemaException):
-        lib.read(symbol, query_builder=q)
+    received = lib.read(symbol, query_builder=q).data
+    assert not len(received)
+    assert received.index.name == "grouping_column"
+    assert len(received.columns) == 1
+    assert "to_mean" in received.columns
```
IvoDD, line 272: Maybe also a test which tries grouping on a non-existent column if there is no such test yet?

cpp/arcticdb/processing/clause.cpp
alexowens90: Yeh `modify_schema` will throw if the column doesn't exist at all in the timeseries descriptor

python/tests/unit/arcticdb/version_store/test_aggregation.py
alexowens90: Covered in `test_group_column_not_present[_dynamic]`

PR #2729, opened 2025-10-23
Performance 18254756429: Improve hash grouping aggregation parallelism

alexowens90:
#### Reference Issues/PRs
[18254756429](https://man312219.monday.com/boards/7852509418/pulses/18254756429)

Poor quality hash implementations of integral types, including at least some implementations of `std::hash `are basically a static cast. e.g. `std::hash<int64_t>{}(100) == 100`. This is fast, but leads to poor distributions in our bucketing, where we mod the hash with the number of buckets. In particular, if performing a grouping hash on a timeseries where the time points are dates results in all of the rows being partitioned into bucket zero, which then results in no parallelism in the aggregation clause.

Swap to using a consistent hash function across all supported platforms with improved uniformity.

### DIFF
--- a/cpp/arcticdb/processing/test/test_clause.cpp
+++ b/cpp/arcticdb/processing/test/test_clause.cpp
@@ -65,6 +65,37 @@ TEST(Clause, PartitionEmptyColumn) {
     ASSERT_TRUE(processed.empty());
 }
 
+TEST(Clause, PartitionHashQuality) {
+    // Poor quality hash implementations of integral types, including at least some implementations of std::hash are
+    // basically a static cast. e.g. std::hash<int64_t>{}(100) == 100. This is fast, but leads to poor distributions in
+    // our bucketing, where we mod the hash with the number of buckets. In particular, if performing a grouping hash on
+    // a timeseries where the time points are dates results in all of the rows being partitioned into bucket zero, which
+    // then results in no parallelism in the aggregation clause.
+    using namespace arcticdb;
+    ScopedConfig scoped_config({{"Partition.NumBuckets", 10}});
+    auto component_manager = std::make_shared<ComponentManager>();
+
+    PartitionClause<arcticdb::grouping::HashingGroupers, arcticdb::grouping::ModuloBucketizer> partition{"grouping"};
+    partition.set_component_manager(component_manager);
+
+    int64_t num_rows{100};
+    SegmentInMemory seg;
+    auto col = std::make_shared<Column>(
+            make_scalar_type(DataType::INT64), 0, AllocationType::DYNAMIC, Sparsity::NOT_PERMITTED
+    );
+    for (int64_t idx = 0; idx < num_rows; ++idx) {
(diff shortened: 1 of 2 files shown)


IvoDD (approved):

cpp/arcticdb/processing/test/test_clause.cpp
```diff
@@ -65,6 +65,37 @@ TEST(Clause, PartitionEmptyColumn) {
...
+    partition.set_component_manager(component_manager);
+
+    int64_t num_rows{100};
+    SegmentInMemory seg;
+    auto col = std::make_shared<Column>(
+            make_scalar_type(DataType::INT64), 0, AllocationType::DYNAMIC, Sparsity::NOT_PERMITTED
+    );
+    for (auto idx = 0; idx < num_rows; ++idx) {
+        col->set_scalar<int64_t>(idx, idx * 1'000'000'000);
+    }
```
jjerphan: Memory allocation grows linearly here before the process is `SIGKILL`ed after OOM. 

Most of the time seems to be spent here:
Image

jjerphan: The hard-coded constant influences memory allocation; if we decrease it (as shown bellow), the test passes for instance.

```diff
diff --git i/cpp/arcticdb/processing/test/test_clause.cpp w/cpp/arcticdb/processing/test/test_clause.cpp
index 3bbe66a00..f9f5add0a 100644
--- i/cpp/arcticdb/processing/test/test_clause.cpp
+++ w/cpp/arcticdb/processing/test/test_clause.cpp
@@ -84,7 +84,7 @@ TEST(Clause, PartitionHashQuality) {
             make_scalar_type(DataType::INT64), 0, AllocationType::DYNAMIC, Sparsity::NOT_PERMITTED
     );
     for (auto idx = 0; idx < num_rows; ++idx) {
-        col->set_scalar<int64_t>(idx, idx * 1'000'000'000);
+        col->set_scalar<int64_t>(idx, idx * 100'000);
     }
     seg.add_column("grouping", col);
     seg.set_row_id(num_rows - 1);
```

jjerphan: I can reproduce with gcc 11.2 from conda-forge (also used for the PyPI builds).

alexowens90: Yeh I just came to the same conclusion. The loop isn't exiting when `idx` hits 100.
`auto` is resolving to `int`, which is equivalent to `int32_t` here. Changing the `auto` to `int64_t` fixes the issue. Something about the `idx * 1'000'000'000` overflowing the max representable value of `int32_t` is breaking something, which feels like a compiler bug, since both arguments to `set_scalar` are passed by value.

jjerphan: Yes, time to refresh [and old but gold quiz](https://www.acepace.net/integerQuiz/).

vasil-pashov (approved):

PR #2730, opened 2025-10-23
Register cleanup function reliably

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/18261646279

There is possibility which arcticdb python objects has been released only after the library import is cleaned up.
In that case, if any object logs while deconstructing, the null log instance will make the process seg fault.
Register the cleanup function properly so it will cleanup when the interpreter exit
#### Any other comments?
No test can be given as I can't repeat the issue in a standalone script.
Still pybind11 has warned this may happen: https://pybind11.readthedocs.io/en/stable/advanced/misc.html#module-destructors
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/python/python_module.cpp
+++ b/cpp/arcticdb/python/python_module.cpp
@@ -382,15 +382,8 @@ PYBIND11_MODULE(arcticdb_ext, m) {
     register_metrics(m.def_submodule("metrics"));
     register_type_handlers();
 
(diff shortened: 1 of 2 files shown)


phoebusm:
Ignore asv benchmark result as the change only affects behavior at exit

PR #2732, opened 2025-10-24
Perf/18232416238/improve list versions performance when there are lots of snapshots

alexowens90:
#### Reference Issues/PRs
[18232416238](https://man312219.monday.com/boards/7852509418/pulses/18232416238)

- Adds extensive testing for `list_versions`, covering both the existing behaviour, as well as what the behaviour should be when the bugs discovered during this process have been fixed ([18262322490](https://man312219.monday.com/boards/7852509418/pulses/18262322490), [18279584183](https://man312219.monday.com/boards/7852509418/pulses/18279584183), and [18286248854](https://man312219.monday.com/boards/7852509418/pulses/18286248854)). All of these tests pass on `master`, to ensure that the refactoring does not introduce any unintended behaviour changes.
- Parallelise reading of snapshot keys.
- If only one symbol is requested, then only extract index keys for that symbol from the read snapshot segments.
- Parallelised the version chain traversals over symbols.
- Eliminated need for sorting the entire result vector at the end.
- Combined `SymbolVersionToSnapshotMap` and `SymbolVersionTimestampMap` into a single map, as in some cases inserting into these maps in `get_snapshot_version_info` was the bottleneck.
- Replaced some copies with moves.

### DIFF
--- a/python/.asv/results/benchmarks.json
+++ b/python/.asv/results/benchmarks.json
@@ -1640,7 +1640,7 @@
         "warmup_time": 0
     },
     "list_versions.ListVersions.peakmem_list_versions": {
-        "code": "class ListVersions:\n    def peakmem_list_versions(\n        self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots\n    ):\n        self.lib.list_versions(symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots)\n\n    def setup(self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots):\n        self.ac = Arctic(\"lmdb://list_versions\")\n        self.lib = self.ac[self._lib_name(num_symbols, num_versions, num_snapshots)]\n\n    def setup_cache(self):\n        start = time.time()\n        self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")",
+        "code": "class ListVersions:\n    def peakmem_list_versions(\n        self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots\n    ):\n        if snapshot is None or num_snapshots > 0:\n            self.lib.list_versions(\n                symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots\n            )\n\n    def setup(self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots):\n        self.ac = Arctic(\"lmdb://list_versions\")\n        self.lib = self.ac[self._lib_name(num_symbols, num_versions, num_snapshots)]\n\n    def setup_cache(self):\n        start = time.time()\n        self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")",
         "name": "list_versions.ListVersions.peakmem_list_versions",
         "param_names": [
             "num_symbols",
@@ -1686,10 +1686,10 @@
         "timeout": 6000,
         "type": "peakmemory",
         "unit": "bytes",
-        "version": "38ba0ce0863d8967098e1ef5ef9662ace2664386f330833f483dc437dbcb2998"
+        "version": "7196504c08aa080e91f3cbb9cbd8228eea01a4b78ecb20f5509bbc67634f8508"
     },
     "list_versions.ListVersions.time_list_versions": {
-        "code": "class ListVersions:\n    def time_list_versions(\n        self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots\n    ):\n        self.lib.list_versions(symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots)\n\n    def setup(self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots):\n        self.ac = Arctic(\"lmdb://list_versions\")\n        self.lib = self.ac[self._lib_name(num_symbols, num_versions, num_snapshots)]\n\n    def setup_cache(self):\n        start = time.time()\n        self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")",
+        "code": "class ListVersions:\n    def time_list_versions(\n        self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots\n    ):\n        if snapshot is None or num_snapshots > 0:\n            self.lib.list_versions(\n                symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots\n            )\n\n    def setup(self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots):\n        self.ac = Arctic(\"lmdb://list_versions\")\n        self.lib = self.ac[self._lib_name(num_symbols, num_versions, num_snapshots)]\n\n    def setup_cache(self):\n        start = time.time()\n        self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")",
         "min_run_count": 2,
         "name": "list_versions.ListVersions.time_list_versions",
         "number": 5,
@@ -1740,7 +1740,7 @@
         "timeout": 6000,
         "type": "time",
         "unit": "seconds",
-        "version": "51288a831ce0073bb37435b01232c3bcfc905449d6bfa8c5d8b2331f14691118",
+        "version": "ee6666cf1919afefb16b4814886b3b80160b921cb6ecbba442f0d473627ec943",
         "warmup_time": 0
     },
     "local_query_builder.LocalQueryBuilderFunctions.peakmem_filtering_numeric": {
--- a/python/benchmarks/list_symbols.py
+++ b/python/benchmarks/list_symbols.py
@@ -40,7 +40,7 @@ class ListSymbols:
             self.ac.delete_library(lib_name)
             lib = self.ac.create_library(lib_name)
             for sym in range(syms):
-                lib.write(f"{sym}_sym", generate_benchmark_df(ListFunctions.rows))
+                lib.write(f"{sym}_sym", generate_benchmark_df(ListSymbols.rows))
 
     def teardown(self, num_symbols):
         pass
--- a/python/benchmarks/list_versions.py
+++ b/python/benchmarks/list_versions.py
@@ -86,9 +86,15 @@ class ListVersions:
     def time_list_versions(
         self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots
     ):
-        self.lib.list_versions(symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots)
+        if snapshot is None or num_snapshots > 0:
+            self.lib.list_versions(
+                symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots
+            )
 
     def peakmem_list_versions(
         self, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots
     ):
-        self.lib.list_versions(symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots)
+        if snapshot is None or num_snapshots > 0:
+            self.lib.list_versions(
+                symbol=symbol, snapshot=snapshot, latest_only=latest_only, skip_snapshots=skip_snapshots
+            )


cpp/arcticdb/version/snapshot.cpp
```diff
@@ -373,31 +373,45 @@ std::set<StreamId> list_streams_in_snapshot(const std::shared_ptr<Store>& store,
...
     });
-
-    return res;
+    const auto window_size = async::TaskScheduler::instance()->io_thread_count();
+    auto futures = folly::window(
+            std::move(snapshot_keys),
+            [store, &snapshot_map, &stream_id](const VariantKey& snapshot_key) {
+                return store->read(snapshot_key).thenValueInline([&snapshot_map, &stream_id](auto&& key_seg) {
+                    auto snapshot_key = std::move(key_seg.first);
+                    auto snapshot_segment = std::move(key_seg.second);
```
vasil-pashov: nit: No need to invoke move semantics if we're we're passing references to to the objects, const refs should have less instructions (though it doesn't really matter)

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
 VersionResultVector get_latest_versions_for_symbols(
         const std::shared_ptr<Store>& store, const std::shared_ptr<VersionMap>& version_map,
-        const std::set<StreamId>& stream_ids, SymbolVersionToSnapshotMap& snapshots_for_symbol
+        std::set<StreamId>&& stream_ids, SymbolVersionToSnapshotInfoMap&& snapshots_for_symbol
 ) {
+    // folly::window requires subscript operator
+    std::vector<StreamId> symbols;
+    symbols.insert(
+            symbols.end(), std::make_move_iterator(stream_ids.rbegin()), std::make_move_iterator(stream_ids.rend())
+    );
```
vasil-pashov: nit: We can shorten this a bit
```suggestion
    std::vector<StreamId> symbols( std::make_move_iterator(stream_ids.rbegin()), std::make_move_iterator(stream_ids.rend()))
```

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
 
 VersionResultVector get_all_versions_for_symbols(
         const std::shared_ptr<Store>& store, const std::shared_ptr<VersionMap>& version_map,
-        const std::set<StreamId>& stream_ids, SymbolVersionToSnapshotMap& snapshots_for_symbol,
-        const SymbolVersionTimestampMap& creation_ts_for_version_symbol
+        std::set<StreamId>&& stream_ids, SymbolVersionToSnapshotInfoMap&& snapshots_for_symbol
 ) {
+    // folly::window requires subscript operator
+    std::vector<StreamId> symbols;
+    symbols.insert(
```
vasil-pashov: Same as https://github.com/man-group/ArcticDB/pull/2732/files#r2474246246

vasil-pashov (approved):

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -373,31 +373,45 @@ std::set<StreamId> list_streams_in_snapshot(const std::shared_ptr<Store>& store,
...
     ARCTICDB_SAMPLE(GetVersionsFromSnapshot, 0)
-    SnapshotMap res;
-
-    iterate_snapshots(store, [&res, &store](const VariantKey& vk) {
-        SnapshotId snapshot_id{fmt::format("{}", variant_key_id(vk))};
-        res[snapshot_id] = get_versions_from_snapshot(store, vk);
+    SnapshotMap snapshot_map;
+    std::vector<VariantKey> snapshot_keys;
+    iterate_snapshots(store, [&snapshot_map, &snapshot_keys](auto&& snapshot_key) {
+        SnapshotId snapshot_id{fmt::format("{}", variant_key_id(snapshot_key))};
```
IvoDD: I know this is not from this PR but why do we need to convert `StreamId` to a string?
Both `StreamId` and `SnapshotId` are `VariantId`s.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
+            symbols.end(), std::make_move_iterator(stream_ids.rbegin()), std::make_move_iterator(stream_ids.rend())
+    );
+    const auto window_size = async::TaskScheduler::instance()->io_thread_count();
+    auto opt_version_result_futures = folly::window(
+            std::move(symbols),
+            [&store, &version_map, &snapshots_for_symbol](const StreamId& stream_id) {
+                return async::submit_io_task(CheckReloadTask{
```
IvoDD, line 199: I noticed the vesion map uses a `mutex` instead of a `shared_mutex` for its map [here](https://github.com/man-group/ArcticDB/blob/7acd347e3a035b635fd71c1bde2b54e0135ac8db/cpp/arcticdb/version/version_map.hpp#L116).

This can slow us down if we only read cached entries with many io threads.

Can change it in a separate PR

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
-                    s_id,
-                    opt_version_key->version_id(),
-                    opt_version_key->creation_ts(),
-                    snapshots_for_symbol[{s_id, opt_version_key->version_id()}],
-                    false
-            );
+    res.reserve(opt_version_results.size());
+    // opt_version_results are already in alphabetical order of symbols, so no sorting required
+    for (auto&& opt_version_result_or_error : opt_version_results) {
+        if (opt_version_result_or_error.hasValue() && opt_version_result_or_error.value().has_value()) {
```
IvoDD: Isn't this going to silently swallow any e.g. network exceptions we get while iterating the version chain?

Previously we would raise?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
+    );
+    const auto window_size = async::TaskScheduler::instance()->io_thread_count();
+    auto symbol_version_vector_futures = folly::window(
+            std::move(symbols),
+            [&store, &version_map, &snapshots_for_symbol](const StreamId& stream_id) {
+                return async::submit_io_task(CheckReloadTask{
+                                                     store,
+                                                     version_map,
+                                                     stream_id,
+                                                     LoadStrategy{LoadType::ALL, LoadObjective::UNDELETED_ONLY}
```
IvoDD: I know this was like this before but loading `UNDELETED_ONLY` would stop at a tombstone all. This way it would miss a potentially snapshotted version before that?

python/tests/unit/arcticdb/version_store/test_list_versions.py
```diff
@@ -0,0 +1,289 @@
...
+    for snap_idx in range(num_snapshots):
+        snap = f"snap{snap_idx}"
+        versions = {f"sym{sym_idx}": min(snap_idx, sym_idx) for sym_idx in range(num_symbols)}
+        snapshots[snap] = versions
+        lib.snapshot(snap, versions=versions)
+        for version in all_versions:
+            if version["symbol"] in versions.keys() and version["version"] == versions[version["symbol"]]:
+                version["snapshots"].append(snap)
+    for sym_idx in range(num_symbols):
+        lib.delete_versions(f"sym{sym_idx}", list(range(0, sym_idx + 1, 2)))
```
IvoDD, line 45: It would be nice to also test with some versions deleted with a tombstone all. I think they go through a slightly different code path because they are never loaded by the `LoadType::UNDELETED_ONLY`

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
+                                             })
+                        .via(&async::cpu_executor())
```
IvoDD, line 261: Why do we go through `cpu_executor` in this case? I assume it's because of the sort?
If we decide to use CPU executor maybe we should limit the `window_size` to `cpu_thread_count` and use `submit_cpu_task`?

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
+                                            sym_version.first,
+                                            sym_version.second,
+                                            snapshot_info.ts,
+                                            std::move(snapshot_info.snapshots),
+                                            true
+                                    );
+                                }
+                            }
+                            // All symbols will be the same so compare only on date field
+                            std::sort(res.begin(), res.end(), [](const VersionResult& v1, const VersionResult& v2) {
```
IvoDD, line 291: The for loop on `all_versions` is already ordered by time. We can change the second for loop on `snapshots_for_symbol` also ordered by time. This will allow merging the two sorted lists in `O(n)` without the need for an explicit sort call.

To do this we will need to change the data structure `snapshots_for_symbol` from `unordered_map<std::pair<StreamId, VersionId>, SnapshotInfo>` to an `unordered_map<StreamId, std::map<VersionId, SnapshotInfo>>`. This will have the following benefits:
- Can iterate only on the versions for the requested symbols (resolves the `O(n**2)` from [previous comment](https://github.com/man-group/ArcticDB/pull/2732#discussion_r2481395069))
- Can iterate through the versions for that symbol in `O(n)` in order.
It has a downside:
- Construction of the ordered map will be `O(n*logn)` on number of versions for the symbol.

Even if we don't use an ordered map for the inner map we'll still get the first benefit without the downside.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
+                if (symbol_version_vector.hasValue()) {
+                    return accum + symbol_version_vector.value().size();
+                } else {
+                    return accum;
+                }
             }
+    ));
+    // symbol_version_vectors are already in alphabetical order of symbols, so no sorting required
+    for (auto&& symbol_version_vector : symbol_version_vectors) {
+        if (symbol_version_vector.hasValue()) {
```
IvoDD: Again previously we would raise if we encounter networking issue during version chain iteration.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -157,83 +161,169 @@ VersionResultVector list_versions_for_snapshot(
...
+                                        snapshots_it == snapshots_for_symbol.end()
+                                                ? std::vector<SnapshotId>()
+                                                : std::move(snapshots_it->second.snapshots),
+                                        false
+                                );
+                            }
+                            for (auto& [sym_version, snapshot_info] : snapshots_for_symbol) {
+                                // For all symbol, version combinations in snapshots, check if they have been pruned,
+                                // and if so use the information from the snapshot indexes and set deleted to true.
+                                if (sym_version.first == stream_id && !unpruned_versions.contains(sym_version)) {
```
IvoDD, line 280: Isn't this `O(n**2)` on number of symbols requested? The `snapshots_for_symbol` will contain the info for all symbols and for each symbol we would iterate over the entire thing.

I might be wrong though, given the big performance improvement you showed.

Suggestion in [next comment](https://github.com/man-group/ArcticDB/pull/2732#discussion_r2481341555).

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -118,14 +118,18 @@ struct VersionComp {
     }
 };
 
-using SymbolVersionToSnapshotMap = std::unordered_map<std::pair<StreamId, VersionId>, std::vector<SnapshotId>>;
-using SymbolVersionTimestampMap = std::unordered_map<std::pair<StreamId, VersionId>, timestamp>;
+struct SnapshotInfo {
+    std::vector<SnapshotId> snapshots;
+    timestamp ts;
```
IvoDD: Nit: I was confused for a long time that this is time of snapshot and not time of creation. Maybe a rename to `index_creation_ts` would be good. I don't feel too strongly about this (I was being kind of dumb)

cpp/arcticdb/version/snapshot.cpp
alexowens90: Yeh it was pointless, removed

cpp/arcticdb/version/version_store_api.cpp
alexowens90: I agree that can be a `shared_mutex`, and also agree it should be a separate PR. Opened a [ticket](https://man312219.monday.com/boards/7852509418/pulses/18330241519)

cpp/arcticdb/version/version_store_api.cpp
alexowens90: Yeh you're right. We did need this pattern when iterating the snapshots because deleting there removes the ref key, but that isn't the case here

cpp/arcticdb/version/version_store_api.cpp
alexowens90: Yes because of the sort.
RE `window_size`, that would have a negative performance impact in the (probably common) IO bound case.
I'm pretty sure that using `submit_cpu_task` explicitly won't make any difference, the CPU task will still have to wait for a free thread in the pool.

cpp/arcticdb/version/version_store_api.cpp
alexowens90: Yeh that is the case. A lot of the big performance improvement when listing versions for individual symbols comes from ignoring index keys in snapshots that aren't for the symbol requested:
https://github.com/man-group/ArcticDB/blob/e838dc7490b8d01bd0b9ba63e5ee1ad25ec93211/cpp/arcticdb/version/snapshot.cpp#L382
which then means that this iteration is only over the symbol of interest in that case

cpp/arcticdb/version/version_store_api.cpp
alexowens90: I did consider more extreme refactors like this, but decided against for a couple of reasons. The lack of existing testing makes iterating costly, as every time you add a test you need to check that the behaviour is the same on master. The new tests are an improvement, but by no means exhaustive. I personally find the current logic and data structures quite hard to follow due to the 16 combinations of possible input parameters, and by the looks of it most options were bodged onto the side later. So I thought it was safer to parallelise the existing algorithm and get rid of a couple of obvious inefficiencies than to rewrite everything.

PR #2734, opened 2025-10-27
[9360419971] Configurable strings format for arrow

IvoDD:
#### Reference Issues/PRs
Monday ref: 9360419971

Introduce `ArrowOutputStringFormat` enum.

It allows configuring the output format for string columns when using arrow output_format.

#### Any other comments?
Notes for reviewers: The PR is split in several commits and I suggest reviewing commit by commit.
First two are only refactors and don't change any existing behavior:

[Remove DataTypeMode and TypeHandler::type_size](https://github.com/man-group/ArcticDB/pull/2734/commits/7a78ea4988cebc129339368f2775b5fb52b4a79f) 
`DataTypeMode` was used to signify whether we are constructing the
segment to be offloaded to python or to be used internally. However it
was not assigned correctly and was not really used in a meaningfull way.

Also `TypeHandler::type_size` is not needed as the type size can be
inferred from the column type.

[Refactor TypeHandler to get ReadOptions](https://github.com/man-group/ArcticDB/pull/2734/commits/84f495989667263cf0a6b853748482b492da6212) 
This commit is only a refactor. It is in preparation to pass
`ArrowOutputConfig` (to be part of `ReadOptions`) down to the `TypeHandler`.
It also does the preparation to use a `extra_bytes_per_block`
to allow storing the offset buffers within the column.

The main behavior changes are in the third commit:
[Introduce ArrowOutputStringFormat](https://github.com/man-group/ArcticDB/pull/2734/commits/365071f40d0fe506f3483f9b6ff0d758eeeb8563) 
- Allow specifying a string output format for each column to be either
  `CATEGORICAL`, `LARGE_STRING`, `SMALL_STRING`
- Specify `extra_bytes_per_block` to output columns when using
  `LARGE_STRING` or `SMALL_STRING`
- ArrowStringHandler logic for handling the new output types
- sparrow construction of the new string output formats
- Truncation handling for string columns
- Tests and benchmarks

#### Benchmark results for actual string processing
I've reordered the benchmark results in a table (showing the mean time in microseconds) for easier preview of results.
Benchmark was run with:
```
./benchmarks --benchmark_time_unit=us --benchmark_filter="BM_arrow_string_handler" --benchmark_min_time=2000x --benchmark_repetitions=5
```

| Becnhmark                                 | CATEGORICAL | LARGE_STRING | SMALL_STRING |
| ----------------------------------------- | ----------- | ------------ | ------------ |
| BM_arrow_string_handler/10000/1/0/X       | 62          | 88.5         | 73.9         |
| BM_arrow_string_handler/100000/1/0/X      | 552         | 817          | 705          |
| BM_arrow_string_handler/10000/10000/0/X   | 229         | 96.3         | 80.3         |
| BM_arrow_string_handler/100000/100000/0/X | 3406        | 863          | 785          |
| BM_arrow_string_handler/10000/1/5000/X    | 63.9        | 76.8         | 77.1         |
| BM_arrow_string_handler/100000/1/50000/X  | 598         | 740          | 709          |
| BM_arrow_string_handler/10000/1/10000/X   | 7.59        | 12.9         | 15.3         |
| BM_arrow_string_handler/100000/1/100000/X | 7.18        | 33           | 32           |

Results look as expected. Conversion is faster if we have many unique strings because we don't have to build up an unordered map. Conversion is slower (but not by much) if we have few unique strings (because we need to allocate larger string buffer) or we have very sparse data (because we need to populate offset buffer even for missing values)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/.asv/results/benchmarks.json
+++ b/python/.asv/results/benchmarks.json
@@ -128,6 +128,7 @@
             ],
             [
                 "<ArrowOutputStringFormat.CATEGORICAL: 'CATEGORICAL'>",
+                "<ArrowOutputStringFormat.DICTIONARY_ENCODED: 'DICTIONARY_ENCODED'>",
                 "<ArrowOutputStringFormat.LARGE_STRING: 'LARGE_STRING'>",
                 "<ArrowOutputStringFormat.SMALL_STRING: 'SMALL_STRING'>"
             ]
@@ -163,6 +164,7 @@
             ],
             [
                 "<ArrowOutputStringFormat.CATEGORICAL: 'CATEGORICAL'>",
+                "<ArrowOutputStringFormat.DICTIONARY_ENCODED: 'DICTIONARY_ENCODED'>",
                 "<ArrowOutputStringFormat.LARGE_STRING: 'LARGE_STRING'>",
                 "<ArrowOutputStringFormat.SMALL_STRING: 'SMALL_STRING'>"
             ]
@@ -200,6 +202,7 @@
             ],
             [
                 "<ArrowOutputStringFormat.CATEGORICAL: 'CATEGORICAL'>",
+                "<ArrowOutputStringFormat.DICTIONARY_ENCODED: 'DICTIONARY_ENCODED'>",
                 "<ArrowOutputStringFormat.LARGE_STRING: 'LARGE_STRING'>",
                 "<ArrowOutputStringFormat.SMALL_STRING: 'SMALL_STRING'>"
             ]
@@ -241,6 +244,7 @@
             ],
             [
                 "<ArrowOutputStringFormat.CATEGORICAL: 'CATEGORICAL'>",
+                "<ArrowOutputStringFormat.DICTIONARY_ENCODED: 'DICTIONARY_ENCODED'>",
                 "<ArrowOutputStringFormat.LARGE_STRING: 'LARGE_STRING'>",
                 "<ArrowOutputStringFormat.SMALL_STRING: 'SMALL_STRING'>"
             ]


python/arcticdb/version_store/library.py
```diff
@@ -1893,6 +1897,8 @@ def read(
         query_builder: Optional[QueryBuilder] = None,
         lazy: bool = False,
         output_format: Optional[Union[OutputFormat, str]] = None,
+        arrow_string_format_default: Optional[ArrowOutputStringFormat] = None,
+        arrow_string_format_per_column: Optional[Dict[str, ArrowOutputStringFormat]] = None,
```
IvoDD: This PR only introduces the options on V2 API only for non batch methods: `read`, `head`, `tail`, `collect`.

I'll add the options to batch methods in a separate PR as this PR is long enough and I want to think about deduplicating logic to allow both specifying string format for individual `ReadRequests` and providing a general default for the entire batch. We do something similar for `output_format` already. Plus I want to add some `read_batch` tests

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -926,38 +931,30 @@ struct ReduceColumnTask : async::BaseTask {
         }();
 
         if (dynamic_schema && column_data == slice_map_->columns_.end()) {
-            if (const std::shared_ptr<TypeHandler>& handler =
-                        get_type_handler(read_options_.output_format(), column.type());
-                handler) {
-                handler->default_initialize(
-                        column.buffer(), 0, frame_.row_count() * handler->type_size(), shared_data_, handler_data_
-                );
```
IvoDD, line 934: This a slight behavior change to fix a bug.

If we're using a type handler to do the `default_initilize` for arrow we should also set the `validity_bitmap`. 
This happens in `NullValueReducer.finalize` and was only happening in the `else` branch.

This change entirely removes the `if` branch because `NullValueReducer` is already type handler aware and does the correct thing.

cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -556,6 +565,8 @@ class ChunkedBufferImpl {
 
     MemBlock* create_detachable_block(size_t capacity, size_t offset) const {
         auto [ptr, ts] = Allocator::aligned_alloc(sizeof(MemBlock));
+        // TODO: Comment on extra bytes
+        capacity += extra_bytes_per_block_.value_or(0);
```
vasil-pashov: If we're using `value_or(0)` can this be just a size_t that's 0 by default?

cpp/arcticdb/pipeline/read_frame.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2023 Man Group Operations Limited
+/* Copyright 2023 Man Group Operations Limiteread_framed
```
vasil-pashov: Is this intentional?

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -15,6 +15,16 @@
 #include <arcticdb/column_store/column_algorithms.hpp>
 
 namespace arcticdb {
+ArrowOutputStringFormat ArrowStringHandler::output_string_format(
+        const std::string_view& column_name, const ReadOptions& read_options
```
vasil-pashov: It's usually [better to pass string_view ](https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/)and other small objects by value.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -48,9 +58,85 @@ void ArrowStringHandler::handle_type(
...
+        const std::shared_ptr<StringPool>& string_pool
+) {
+    using ArcticStringColumnTag = ScalarTagType<DataTypeTag<DataType::UTF_DYNAMIC64>>;
+
+    int64_t bytes = 0u;
+    auto last_idx = 0u;
+
+    // `mapping.dest_bytes` does not count the extra offset value we need to store in the end.
+    // Thus, we request one extra value to assert that the buffer has that extra value allocated.
+    auto dest_ptr = reinterpret_cast<OffsetType*>(
```
vasil-pashov, line 73: nit: wrapping raw ptr's in span is a bit safer because out of bounds access will trigger an assertion in debug

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -48,9 +58,85 @@ void ArrowStringHandler::handle_type(
...
+                inserter = en.idx();
+            }
+        } else if (populate_inverted_bitset) {
+            inserter = en.idx();
+        }
+    });
+    while (last_idx <= mapping.num_rows_) {
+        dest_ptr[last_idx++] = static_cast<OffsetType>(bytes);
+    }
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
```
vasil-pashov, line 103: Both small and large string call this function but with a different type parameter. So in theory this can be thrown even if the input is a large string. In that case the error message can be a bit confusing.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -130,14 +216,56 @@ void ArrowStringHandler::
...
+    if (!buffer.extra_bytes_per_block().has_value()) {
+        // For categorical string columns the extra_bytes_per_block won't be set.
+        // Categorical string columns can contain any values as long as they are marked as 0.
+        return;
+    }
+    // Large or small string columns must be default initialized because monotonicity is required even for null values.
+    // `default_initialize` is called only on entire row slices so it is ok to memset buffers with 0s
+    auto dest_ptrs_and_sizes = buffer.byte_blocks_at(offset, byte_size);
+    for (auto& [dest_ptr, bytes] : dest_ptrs_and_sizes) {
+        auto bytes_with_extra = bytes + buffer.extra_bytes_per_block().value();
```
vasil-pashov: This will copy the whole vector. Use const ref

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -95,18 +95,61 @@ sparrow::dictionary_encoded_array<T> create_dict_array(
...
+    return sparrow::big_string_array(std::vector<std::string>{"a"});
+}
+
+template<typename OffsetType>
+sparrow::array minimal_strings_array() {
+    return sparrow::array(sparrow::string_array_impl<OffsetType>(std::vector<std::string>()));
+}
+
+template<typename TagType>
+sparrow::array string_array_from_block(
```
vasil-pashov, line 114: nit: Adding `requires is_sequence_type` improves readability IMO

cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -529,7 +550,8 @@ class ChunkedBufferImpl {
         // (i.e. bytes == block->bytes())
         util::check(bytes <= block->bytes(), "Can't truncate {} bytes from a {} byte block", bytes, block->bytes());
         auto remaining_bytes = block->bytes() - bytes;
-        auto new_block = create_block(remaining_bytes, block->offset_);
+        auto remaining_bytes_without_extra = remaining_bytes - extra_bytes_per_block_.value_or(0);
```
vasil-pashov: Can this become negative?

cpp/arcticdb/arrow/arrow_handlers.cpp
IvoDD: Isn't it going to get return value optimized?

cpp/arcticdb/arrow/arrow_handlers.cpp
IvoDD: Good point, I'll address the string view bit in this PR. 

We do however pass around many `std::shared_ptr`s as `const ref` in many many places around the code. I'll open a ticket to fix that in a big refactor.

cpp/arcticdb/column_store/chunked_buffer.hpp
IvoDD: Yeah we always use `value_or`. My intention was to make it clearer in the places where we check for extra bytes e.g. when constructing the sparrow to do `buffer.extra_bytes().has_value()` but `buffer.extra_bytes() > 0` is probably just as clear, so will fix that.

cpp/arcticdb/arrow/arrow_handlers.cpp
IvoDD: We use raw pointers in a lot of places for `reinterpret_cast<SomeType*>(bytes_at)`. I prefer tackling them together. Will open a ticket to change the `ChunkedBuffer::ptr_cast` to return a span

cpp/arcticdb/column_store/chunked_buffer.hpp
IvoDD: It should not because the `bytes` argument should always be less than `block->bytes() - extra_bytes_per_block`. Will update the assertion to check for this.

cpp/arcticdb/pipeline/read_frame.hpp
IvoDD: ah thanks fixed

cpp/arcticdb/arrow/arrow_handlers.cpp
IvoDD: This can theoretically only ever break for a `SMALL_STRING`. Otherwise we would most certainly OOM if we have more than `max(int64)` bytes in memory.
Also we use an `OffsetType = int64_t` for `LARGE_STRING` so it should not be possible `bytes > max(int64)` because bytes is also `int64`.
Adding a comment to explain this

cpp/arcticdb/arrow/arrow_handlers.cpp
vasil-pashov: Ah yes, I thought it's a member of the class but it's actually created in the method and returned. RVO will definitely kick in.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -241,6 +241,11 @@ void register_bindings(py::module& version, py::exception<arcticdb::ArcticExcept
             .value("PANDAS", OutputFormat::PANDAS)
             .value("ARROW", OutputFormat::ARROW);
 
+    py::enum_<ArrowOutputStringFormat>(version, "InternalArrowOutputStringFormat")
+            .value("CATEGORICAL", ArrowOutputStringFormat::CATEGORICAL)
```
alexowens90, line 245: I know Polars use categorical, whereas Arrow refers to dict-encoded. Could maybe allow both? Can just be an alias defined here, doesn't need to permeate into the C++ layer

cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -66,6 +68,18 @@ struct ReadOptions {
...
+        data_->arrow_output_config_.default_string_format_ = arrow_output_string_format;
+    }
+
+    void set_arrow_output_per_column_string_format(
+            std::unordered_map<std::string, ArrowOutputStringFormat>& per_column_string_format
+    ) {
+        data_->arrow_output_config_.per_column_string_format_ = per_column_string_format;
+    }
+
+    [[nodiscard]] const ArrowOutputConfig& arrow_output_config() const { return data_->arrow_output_config_; }
```
alexowens90: For all 3 of these new methods could `util::check(output_format_ == OutputFormat::ARROW)`

cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -224,7 +239,9 @@ class ChunkedBufferImpl {
                 add_block(std::max(DefaultBlockSize, extra_size), last_off);
             }
             res = last_block().end();
-            last_block().bytes_ = extra_size;
+            if (allocation_type_ != entity::AllocationType::DETACHABLE) {
+                last_block().bytes_ = extra_size;
```
alexowens90, line 243: Where is this set in the detachable case?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -129,26 +139,23 @@ SegmentInMemory allocate_chunked_frame(const std::shared_ptr<PipelineContext>& c
...
     auto [offset, row_count] = offset_and_row_count(context);
-    SegmentInMemory output{
-            get_filtered_descriptor(context, output_format),
-            row_count,
-            AllocationType::DETACHABLE,
-            Sparsity::NOT_PERMITTED,
-            output_format,
-            DataTypeMode::EXTERNAL
-    };
+    auto [desc, _] = get_filtered_descriptor_and_extra_bytes_per_column(context, read_options);
```
alexowens90:
```suggestion
    auto desc = get_filtered_descriptor_and_extra_bytes_per_column(context, read_options).first;
```

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -15,11 +15,21 @@
 #include <arcticdb/column_store/column_algorithms.hpp>
 
 namespace arcticdb {
+ArrowOutputStringFormat ArrowStringHandler::output_string_format(
+        std::string_view column_name, const ReadOptions& read_options
+) const {
+    const auto& arrow_config = read_options.arrow_output_config();
+    auto it = arrow_config.per_column_string_format_.find(std::string{column_name});
```
alexowens90: nit: Put inside the `if`

python/arcticdb/util/arrow.py
```diff
@@ -1,21 +1,21 @@
 from arcticdb.dependencies import pyarrow as pa
 
 
-def stringify_dictionary_encoded_columns(table, string_type=None):
```
alexowens90, line 4: We told people to use this during the beta, so we could leave it in as a thin wrapper around `cast_string_columns` to not break any existing code

python/arcticdb/version_store/library.py
```diff
@@ -292,6 +292,8 @@ class ReadRequest(NamedTuple):
     columns: Optional[List[str]] = None
     query_builder: Optional[QueryBuilder] = None
     output_format: Optional[Union[OutputFormat, str]] = None
+    arrow_string_format_default: Optional[ArrowOutputStringFormat] = None
+    arrow_string_format_per_column: Optional[Dict[str, ArrowOutputStringFormat]] = None
```
alexowens90: Could add to repr

python/arcticdb/version_store/library.py
alexowens90: Works for me, please add a ticket to the epic

python/arcticdb/version_store/library.py
```diff
@@ -1949,6 +1955,14 @@ def read(
             For more information see documentation of `Arctic.__init__`.
             If `None` uses the default output format from the `Library` instance.
 
+        arrow_string_format_default: Optional[ArrowOutputStringFormat], default=None
+            If using `output_format=EXPERIMENTAL_ARROW` it sets the output format of string columns for arrow.
+            See documentation of `ArrowOutputStringFormat` for more information on the different options.
+            If `None` uses the default arrow_string_format from the `Library` instance.
+
+        arrow_string_format_per_column: Optional[Dict[str, ArrowOutputStringFormat]], default=None
```
alexowens90: Do these overrides work with data written with a Pandas multiindex?

python/arcticdb/options.py
```diff
@@ -167,17 +167,41 @@ def output_format_to_internal(output_format: Union[OutputFormat, str]) -> Intern
         raise ValueError(f"Unknown OutputFormat: {output_format}")
 
 
+class ArrowOutputStringFormat(str, Enum):
```
alexowens90, line 171: Documentation of this class is referenced elsewhere so it needs some

cpp/arcticdb/arrow/test/test_arrow_read.cpp
```diff
@@ -136,42 +159,81 @@ TEST(ArrowRead, ColumnString) {
...
+                auto str_size = offset_buffer.cast<int64_t>(id + 1) - offset_begin;
+                auto str_in_column = std::string_view(
+                        reinterpret_cast<char*>(string_buffer.bytes_at(offset_begin, str_size)), str_size
+                );
+                EXPECT_EQ(str_in_column, column_values[global_row]);
+            }
+        } else {
+            // We should be using 64-bit offsets for LARGE_STRING and 32-bit offsets for SMALL_STRING
+            details::visit_scalar(column.type(), [&](auto&& impl) {
+                using TagType = std::decay_t<decltype(impl)>;
```
alexowens90, line 201: nit: Wrap in `if constexpr(is_sequnce_type(...))` to avoid generating dead code for other types

python/tests/unit/arcticdb/version_store/test_arrow_read.py
alexowens90, line 1: I might just be missing it, but all the tests seem to either use the library level default or the column overrides for all string columns, but not the read-time override, or combinations thereof?

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -64,8 +65,10 @@ def test_basic_write_strings(lmdb_version_store_arrow, type):
     sym = "test_basic_write_strings"
     table = pa.table({"col": pa.array(["hello", "bonjour", "gutentag", "nihao", "konnichiwa"], type)})
     lib.write(sym, table)
-    received = lib.read(sym).data
-    received = stringify_dictionary_encoded_columns(received, type)
+    arrow_string_format = (
+        ArrowOutputStringFormat.SMALL_STRING if type == pa.string() else ArrowOutputStringFormat.LARGE_STRING
+    )
```
alexowens90: This is making me wonder if we should support `pa.string()` and `pa.large_string()` as values passed to `arrow_string_format_default` (and the per column overrides equivalent). I don't know what we'd do for dict encoded though

python/tests/unit/arcticdb/version_store/test_arrow_read.py
alexowens90: OK I see this this covered in `test_arrow_api.py`, ignore this

alexowens90 (approved):

cpp/arcticdb/version/python_bindings.cpp
IvoDD: Done in python layer in `options.py`

cpp/arcticdb/column_store/chunked_buffer.hpp
IvoDD: When creating detachable blocks we use the `MemBlock` constructor which assigns both bytes and capacity [here](https://github.com/man-group/ArcticDB/blob/08b5e462e8c924b41866bd4e326f3bae3a4e4ab1/cpp/arcticdb/column_store/chunked_buffer.hpp#L602)

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
IvoDD: I guess it makes sense to allow using the pyarrow types. For categorical we could allow passing the precise type `pa.dictionary(pa.int32(), pa.large_string())` and raise otherwise.
Will add it.

PR #2736, opened 2025-10-28
Support recursive normalizer in v2 API write

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/18298965201
1. Makes `def write` and `def write_pickle` in v2 API support recursive normalizer
- Recursive normalizer enables ArcricDB to write and read nest data structures (`dict`, `list`, `tuple`) of dataframes and arrays without having to pickling the entire structure
- The data stored is not filterable, as is pickled data
- Please refer to https://docs.arcticdb.io/latest/notebooks/arcticdb_demo_recursive_normalizers for more details
3. Minor changes on formatter to support formatting individual files

Per discussed, in v2 API, recursive normalizer setting in LibraryOption will be respected. Default library option of recursive normalizer will be `False`. Existing libraries are unaffected.
#### Any other comments?
`def batch_write` in v1 API doesn't support recursive normalizer. Ticket: https://man312219.monday.com/boards/7852509418/pulses/7855436309
Therefore the support of recursive normalizer in corresponding v2 API will not be covered in this PR.

##### Pickling
`pickling` is a bit of a mess in V1 API. 
For `arrow` and `pandas` data, if the normalization fails, it can fallback to msgpack and pickling, depending on whether `pickle_on_failure` is `True`. 
However, for other kinds of data, it almost certainly fallback to msgpack and pickling. The only option to prevent pickling is a library config `strict_mode`. But I don't see there is any API to enable this option, in V1/V2 nor internally.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -1030,13 +1030,15 @@ class Library:
             the default library configuration is True, otherwise it is False.
             The library configuration can be modified via Arctic.modify_library_option(). Please refer to
             https://docs.arcticdb.io/latest/api/arctic/#arcticdb.Arctic.modify_library_option for more details.
-            The data structure can be nested or a mix of lists and dictionaries.
+            The data structure can be nested or a mix of lists, dictionaries and tuples.
             Example:
                 data = {"a": np.arange(5), "b": pd.DataFrame({"col": [1, 2, 3]})}
                 lib.write(symbol, data, recursive_normalizers=False) # ArcticUnsupportedDataTypeException will be thrown
                 lib.write(symbol, data, recursive_normalizers=True) # The data will be successfully written
                 ac.modify_library_option(lib, ModifiableLibraryOption.RECURSIVE_NORMALIZERS, True)
                 lib.write(symbol, data) # The data will be successfully written
+            Please refer to https://docs.arcticdb.io/latest/notebooks/arcticdb_demo_recursive_normalizers for more details
+            of this feature.
 
         Returns
         -------
@@ -1072,7 +1074,7 @@ class Library:
             if is_recursive_normalizers_enabled:
                 if staged:
                     raise ArcticUnsupportedDataTypeException(
-                        "Staged data must be of a type that can be natively normalized"
+                        "Staged data cannot be natively normalized. The recursive normalizer is enabled but is not allowed to work on staged data."
                     )
             else:
                 raise ArcticUnsupportedDataTypeException(
@@ -1126,6 +1128,7 @@ class Library:
             See documentation on `write`.
         recursive_normalizers: bool, default None
             See documentation on `write`.
+            If enabled, attempts to recursively normalize data before falling back to pickling.
             If the leaf nodes cannot be natively normalized, they will be pickled,
             resulting in the overall data being recursively normalized and partially pickled.
             Example:


python/arcticdb/options.py
```diff
@@ -125,12 +127,19 @@ def __init__(
         encoding_version: Optional[EncodingVersion], default None
             The encoding version to use when writing data to storage.
             v2 is faster, but still experimental, so use with caution.
+
+        recursive_normalizers: bool, default False
+            Whether to recursively normalize nested data structures when writing sequence-like or dict-like data.
+            The data structure can be nested or a mix of lists and dictionaries.
+            Note: If the leaf nodes cannot be natively normalized and must be written using write_pickle, those leaf nodes
+            will be pickled, resulting in the overall data being only partially normalized and partially pickled.
```
alexowens90, line 136: I think the docstrings here, and for `write` and `write_pickle` would benefit from some examples

python/arcticdb/version_store/library.py
```diff
@@ -958,6 +944,7 @@ def write(
         staged=False,
         validate_index=True,
         index_column: Optional[str] = None,
+        recursive_normalizers: bool = None,
```
alexowens90, line 978: Does `write` throw if a leaf node will be pickled?

alexowens90, line 978: What happens if you pass recursively normalized data and `staged=True`? I think it should throw

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
```diff
@@ -57,6 +59,44 @@ def assert_vit_equals_except_data(left, right):
...
+    if lib_option is None:
+        lib = arctic_client_s3.create_library(lib_name)
+    else:
+        lib = arctic_client_s3.create_library(lib_name, LibraryOptions(recursive_normalizers=lib_option))
+    data = {"a": np.arange(5), "b": pd.DataFrame({"col": [1, 2, 3]})}
+    if (lib_option is True and recursive_normalizers is not False) or recursive_normalizers is True:
+        with qs.query_stats():
+            lib.write(sym, data, recursive_normalizers=recursive_normalizers)
+        stats = qs.get_query_stats()
+        assert "MULTI_KEY" in stats["storage_operations"]["S3_PutObject"].keys()
```
alexowens90: Prefer using the library tool for this kind of key existence check post-fact

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
```diff
@@ -57,6 +59,44 @@ def assert_vit_equals_except_data(left, right):
...
+@pytest.mark.parametrize("recursive_normalizers", (True, False, None))
+def test_v2_api_pickle(arctic_library_s3, sym, recursive_normalizers, clear_query_stats):
+    lib = arctic_library_s3
+    data = (
+        {
+            "a": [1, 2, 3],
+            "b": {"c": np.arange(24)},
+            "d": [AlmostAListNormalizer()],  # A random item that will be pickled
+        },
+    )
```
alexowens90: What about if you call `write` with this data and recursive normalizers both enabled and disabled?

alexowens90 (changes requested):

python/arcticdb/version_store/library.py
phoebusm: I didn't realize V1 enable pickling by default, regardless of recursive normalized enabled or not. Will fix it

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
phoebusm: Should always throw exception

python/arcticdb/version_store/library.py
phoebusm: Added test

python/arcticdb/version_store/library.py
```diff
@@ -1086,6 +1129,8 @@ def write_pickle(
             prune_previous_version=prune_previous_versions,
             pickle_on_failure=True,
             parallel=staged,
```
phoebusm, line 1165: V2 API allows staging non-natively normalized data in `write_pickle` by passing `staged=True`. Maybe this combination should be blocked in future major release.

alexowens90: Agreed, I think there is a ticket to remove the `staged` argument from `write_pickle` as it doesn't make sense. Technically an API break though so needs to wait for a better reason to do `7.0.0`

python/arcticdb/version_store/library.py
```diff
@@ -1006,6 +993,19 @@ def write(
         index_column: Optional[str], default=None
             Optional specification of timeseries index column if data is an Arrow table. Ignored if data is not an Arrow
             table.
+        recursive_normalizers: bool, default None
```
vasil-pashov, line 1027: I think we should be a bit more descriptive in the docstring. Even though it's working in V1 almost no one from the outside world knows about it. IMO we should treat it as a new feature. We should also improve the description of the PR as it will go in the release notes.

python/arcticdb/version_store/library.py
```diff
@@ -1034,11 +1034,20 @@ def write(
...
+        )
         if not self._allowed_input_type(data):
-            raise ArcticUnsupportedDataTypeException(
-                "data is of a type that cannot be normalized. Consider using "
-                f"write_pickle instead. type(data)=[{type(data)}]"
-            )
+            if is_recursive_normalizers_enabled:
+                if staged:
+                    raise ArcticUnsupportedDataTypeException(
+                        "Staged data must be of a type that can be natively normalized"
```
vasil-pashov: I'd be a bit more explicit here and tell the user that they're trying to use recursive normalizers for staged data and that it's not allowed. It will be more useful for the user and for us when a support request comes in.

python/arcticdb/version_store/library.py
```diff
@@ -1076,6 +1093,18 @@ def write_pickle(
             See documentation on `write`.
         staged
             See documentation on `write`.
+        recursive_normalizers: bool, default None
+            See documentation on `write`.
+            If the leaf nodes cannot be natively normalized, they will be pickled,
```
vasil-pashov, line 1132: Interesting did I miss a discussion on this. I'd expect write pickle to pickle everything all the time regardless of normalizers.

python/arcticdb/version_store/library.py
phoebusm: I have given a bit more details in the PR desc.
And I have also created a notebook for the feature. In docstring, users will be referred to it for more details

python/arcticdb/version_store/library.py
phoebusm: Yes good spot. If both are enabled, recursive normalizer has priority over pickling. I have amended to docstring to explain the priority.
It follows V1 API on this.

PR #2738, opened 2025-10-30
Add rollback support for recursively normalized data 

ognyanstoimenov:
#### Reference Issues/PRs

This is part of ongoing work to support rollback -- deleting newly written orphaned data keys in an event of insufficient storage (or quota exceeded)  

Monday ref https://man312219.monday.com/boards/7852509418/pulses/18096835093

https://github.com/man-group/ArcticDB/pull/2707 adds rollback support for `write`, `append`, `update`

This PR extends that to support write with `recursive_normalizers=True`

Details:
When writing with recursive normalizers we write a multi_key to storage which contains indexs key of each of the elements.
Previously these seperate writes were happening independently one from another with each writing its own data and index keys.
Now the data keys of each are written up front, then a rollback happens if necessary, deleting all the data keys from all writes if a quota exception occurs.. Then only if all is okay, the index keys of each is written.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/scheduled_cleanup.yml
+++ b/.github/workflows/scheduled_cleanup.yml
@@ -39,12 +39,12 @@ jobs:
           cd python
           # remove the empty protobuf libs so that protobufs are loaded from installed lib
           rm -rf arcticc
-          PYTHONPATH=. python -m utils.s3_roles_delete
+          PYTHONPATH=. python utils/s3_roles_delete.py
 
       - name: Cleanup buckets
         run: |
           cd python
-          PYTHONPATH=. python -m utils.cleanup_test_buckets 
+          PYTHONPATH=. python utils/cleanup_test_buckets.py 
 
 
 
--- a/cpp/arcticdb/column_store/column.cpp
+++ b/cpp/arcticdb/column_store/column.cpp
@@ -880,33 +880,30 @@ std::vector<std::shared_ptr<Column>> Column::split(const std::shared_ptr<Column>
 }
 
 void Column::truncate_first_block(size_t start_row) {
-    if (!is_sparse()) {
-        auto bytes = start_row * data_type_size(type_, OutputFormat::NATIVE, DataTypeMode::INTERNAL);
-        data_.buffer().truncate_first_block(bytes);
-    }
+    util::check(!is_sparse(), "Truncation should only happen for dense columns.");
+    auto bytes = start_row * data_type_size(type_, OutputFormat::NATIVE, DataTypeMode::INTERNAL);
+    data_.buffer().truncate_first_block(bytes);
 }
 
 void Column::truncate_last_block(size_t end_row) {
-    if (!is_sparse()) {
-        const auto column_row_count = row_count();
-        util::check(
-                column_row_count >= static_cast<int64_t>(end_row),
-                "Cannot truncate column of length {} to row {}",
-                column_row_count,
-                end_row
-        );
-        auto bytes = (column_row_count - end_row) * data_type_size(type_, OutputFormat::NATIVE, DataTypeMode::INTERNAL);
-        data_.buffer().truncate_last_block(bytes);
-    }
+    util::check(!is_sparse(), "Truncation should only happen for dense columns.");
+    const auto column_row_count = row_count();
+    util::check(
+            column_row_count >= static_cast<int64_t>(end_row),
+            "Cannot truncate column of length {} to row {}",
+            column_row_count,
+            end_row
+    );
+    auto bytes = (column_row_count - end_row) * data_type_size(type_, OutputFormat::NATIVE, DataTypeMode::INTERNAL);
+    data_.buffer().truncate_last_block(bytes);
 }
 
 void Column::truncate_single_block(size_t start_row, size_t end_row) {
-    if (!is_sparse()) { // TODO: We need to also do truncation on sparse data
-        const auto type_size = data_type_size(type_, OutputFormat::NATIVE, DataTypeMode::INTERNAL);
-        auto start_offset = type_size * start_row;
-        auto end_offset = type_size * end_row;
-        data_.buffer().truncate_single_block(start_offset, end_offset);
-    }
+    util::check(!is_sparse(), "Truncation should only happen for dense columns.");
+    const auto type_size = data_type_size(type_, OutputFormat::NATIVE, DataTypeMode::INTERNAL);
+    auto start_offset = type_size * start_row;
+    auto end_offset = type_size * end_row;
+    data_.buffer().truncate_single_block(start_offset, end_offset);
 }
 
 /// Bytes from the underlying chunked buffer to include when truncating. Inclusive of start_byte, exclusive of end_byte
--- a/cpp/arcticdb/pipeline/index_writer.hpp
+++ b/cpp/arcticdb/pipeline/index_writer.hpp
@@ -98,8 +98,8 @@ class IndexWriter {
 
         bool new_col_group = !current_col_.has_value() || *current_col_ < slice.col_range.first;
         bool missing_row_val = !current_row_.has_value();
-        bool is_valid_col = *current_col_ == slice.col_range.first;
-        bool is_valid_row = *current_row_ < slice.row_range.first;
+        bool is_valid_col = new_col_group || *current_col_ == slice.col_range.first;
+        bool is_valid_row = missing_row_val || *current_row_ < slice.row_range.first;
         bool is_valid = (is_valid_col && is_valid_row);
         util::check_arg(
                 missing_row_val || new_col_group || is_valid,
--- a/cpp/arcticdb/pipeline/read_frame.cpp
+++ b/cpp/arcticdb/pipeline/read_frame.cpp
@@ -307,8 +307,7 @@ void decode_or_expand(
                 util::expand_dense_buffer_using_bitmap<RawType>(bv.value(), sparse.data(), dest);
             });
 
-            // TODO We must handle sparse columns in ArrowStringHandler and deduplicate logic between the two.
-            // Consider registering a sparse handler on the TypeHandlerRegistry.
+            // For arrow we must apply the truncation to the bitmap and set it as an extra buffer.
             if (output_format == OutputFormat::ARROW) {
                 bv->resize(dest_bytes / dest_type_desc.get_type_bytes());
                 handle_truncation(*bv, mapping.truncate_);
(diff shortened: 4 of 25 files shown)


cpp/arcticdb/pipeline/test/test_rollback.cpp
```diff
@@ -24,6 +24,36 @@ TestTensorFrame write_version_frame_with_three_segments(
     return frame;
 }
 
+void write_version_frame_with_three_segments_rec_norm(
+        version_store::PythonVersionStore& store, const StreamId& stream_id
+) {
+    std::vector<StreamId> rec_norm_ids;
+    std::vector<TestTensorFrame> frames;
```
vasil-pashov: What't the purpose of this? We only push elements to it but nothing else. Also is it possible to come up with better naming?

cpp/arcticdb/pipeline/write_frame.hpp
```diff
@@ -98,10 +98,54 @@ std::vector<SliceAndKey> flatten_and_fix_rows(
 
 std::vector<std::pair<FrameSlice, size_t>> get_slice_and_rowcount(const std::vector<FrameSlice>& slices);
 
-folly::SemiFuture<std::vector<SliceAndKey>> rollback_on_quota_exceeded(
+template<typename T, typename F>
```
vasil-pashov: Does this have to be a template? It looks like we can use concrete types. If it has to be a template is it possible to make it a bit more specific by adding some `requires` clauses?

cpp/arcticdb/pipeline/write_frame.hpp
```diff
@@ -98,10 +98,54 @@ std::vector<SliceAndKey> flatten_and_fix_rows(
...
+                                       try_slice.exception().template is_compatible_with<QuotaExceededException>();
+        } else if (try_slice.hasValue()) {
+            succeeded.emplace_back(std::move(try_slice.value()));
+        }
+    }
+
+    if (has_quota_limit_exceeded) {
+        return std::forward<F>(remove_future)(std::move(succeeded))
+                .via(&async::cpu_executor())
+                .thenValue([](auto&&...) {
```
vasil-pashov: I don't think we can get more than one parameter in a thenValue callback.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -458,39 +458,29 @@ folly::Future<std::vector<StreamSink::RemoveKeyResultType>> remove_slice_and_key
...
+    };
 
-    return succeeded;
+    return rollback_on_quota_exceeded(std::move(try_slices), remove_slice_and_keys_func);
 }
 
+folly::SemiFuture<SliceAndKeyBatches> rollback_batches_on_quota_exceeded(
+        SliceAndKeyTryBatches&& try_slices, const std::shared_ptr<stream::StreamSink>& sink
+) {
+    auto remove_keys = [sink](std::vector<std::vector<SliceAndKey>>&& slice_keys_per_batch) {
```
vasil-pashov: nit: I'd prefer this to be a free function instead of lambda, if possible.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -458,39 +458,29 @@ folly::Future<std::vector<StreamSink::RemoveKeyResultType>> remove_slice_and_key
...
-
-    if (exception.has_value()) {
-        exception->throw_exception();
-    }
+    auto remove_slice_and_keys_func = [sink](std::vector<SliceAndKey>&& slice_keys) {
+        return remove_slice_and_keys(std::move(slice_keys), *sink);
+    };
 
-    return succeeded;
+    return rollback_on_quota_exceeded(std::move(try_slices), remove_slice_and_keys_func);
```
vasil-pashov: nit: Either move the lambda or place it here instead of a variable to avoid copying the underlying struct

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1488,26 +1488,56 @@ folly::Future<VersionedItem> LocalVersionedEngine::write_index_key_to_version_ma
...
+        auto partial_key = IndexPartialKey{frame->desc().id(), version_id};
+        auto fut_slice_keys =
+                slice_and_write(frame, slicing_policy, std::move(partial_key), store(), de_dup_maps[idx], false);
+        batch_futures.emplace_back(std::move(fut_slice_keys));
     }
-    return results_fut;
+
+    return folly::collectAll(batch_futures)
+            .via(&async::cpu_executor())
+            .thenValue([store = store()](SliceAndKeyTryBatches&& batches) mutable {
```
vasil-pashov: Why is this mutable?

vasil-pashov (approved):

cpp/arcticdb/pipeline/write_frame.hpp
ognyanstoimenov: Ah yeah, you are right

cpp/arcticdb/pipeline/write_frame.hpp
ognyanstoimenov: I want to use it for both `vector<SlicaAndKey>` and `vector<vector<SliceAndKey>>` and actually any `vector<T>` should work. Maybe I've made it too generic with that `F` I'll try to make it more specific with `require`

cpp/arcticdb/version/local_versioned_engine.cpp
ognyanstoimenov: No idea :D I think I wanted to move the `store` ptr but that is unneeded. Will remove

cpp/arcticdb/pipeline/test/test_rollback.cpp
ognyanstoimenov: This is a weird one. I want the `TestTensorFrame`s to stay alive when I call
`store.batch_write_internal(..,. std::move(TestTensorFrame.frames_),...)` because some internals get destroyed with it and frame_ doesn't behave as expected. I will add a comment to explain that and change the naming

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1488,26 +1488,56 @@ folly::Future<VersionedItem> LocalVersionedEngine::write_index_key_to_version_ma
...
+            if (std::holds_alternative<Error>(check_outcome)) {
+                std::get<Error>(check_outcome).throw_error();
+            }
+        }
+
+        auto write_options = get_write_options();
+        frame->set_bucketize_dynamic(write_options.bucketize_dynamic);
+
+        auto slicing_policy = get_slicing_policy(write_options, *frame);
+        auto partial_key = IndexPartialKey{frame->desc().id(), version_id};
```
alexowens90: From line 1501 to here looks duplicated from `async_write_dataframe_impl`? If so, can we factor it out into a separate function?

cpp/arcticdb/pipeline/write_frame.hpp
alexowens90: I would go further, and just support the `vector<vector<SliceAndKey>>` version, and in the non-recursive case the outer vector is just of length one

cpp/arcticdb/pipeline/write_frame.hpp
```diff
@@ -98,10 +98,54 @@ std::vector<SliceAndKey> flatten_and_fix_rows(
...
+
+    return succeeded;
+}
+
+folly::SemiFuture<std::vector<SliceAndKey>> rollback_slices_on_quota_exceeded(
         std::vector<folly::Try<SliceAndKey>>&& try_slices, const std::shared_ptr<stream::StreamSink>& sink
 );
 
+using SliceAndKeyBatches = std::vector<std::vector<SliceAndKey>>;
+using SliceAndKeyTryBatches = std::vector<folly::Try<std::vector<SliceAndKey>>>;
```
alexowens90: This is personal preference, but I find these kinds of aliases unhelpful. `std::vector<folly::Try<std::vector<SliceAndKey>>>` is self-contained and understandable by itself, anywhere I see `SliceAndKeyTryBatches` I have to come back here to know exactly what it is referring to

cpp/arcticdb/pipeline/test/test_rollback.cpp
alexowens90: FWIW I'm also confused by this

cpp/arcticdb/pipeline/test/test_rollback.cpp
```diff
@@ -245,6 +275,77 @@ const auto TEST_DATA_WRITE = {
...
+                .num_writes = 1,
+                .write_expected_outcome = {QUOTA},
+                .operation = Operation::WRITE_REC_NORM
+        },
+        TestScenario{
+                .name = "All_succeed",
+                .write_failures = make_fault_sequence({NONE}),
+                .expected_written_ref_keys = 0, // The rec norm test uses batch_write which only writes data and index
+                .expected_written_version_keys = 0,
+                .expected_written_index_keys = 2,
```
alexowens90, line 315: Could also add an `expected_written_multiindex_keys` field

ognyanstoimenov: I had that before but I removed it as it is always zero. I have a comment above as to why:
`// The rec norm test uses batch_write which only writes data and index (indices)`
Couldn't find an easy way to run the test one layer up where the multiindex is written

cpp/arcticdb/pipeline/write_frame.hpp
ognyanstoimenov: I don't have a strong preference for this. The only reason I did it is to use it in lambda parameters instead of having to type out the whole thing. Happy to remove it though

cpp/arcticdb/pipeline/write_frame.hpp
ognyanstoimenov: > just support the vector<vector<SliceAndKey>> version, and in the non-recursive case the outer vector is just of length one

That would not look as pretty as I thought because of having to support both `vector<Try<vector<SliceAndKey>>` and `vector<Try<SliceAndKey>>`. Also would require changing the `write_slice` method to return ``
folly::SemiFuture<std::vector<folly::Try<vector<SliceAndKey>>>>`` insted of `
folly::SemiFuture<std::vector<folly::Try<SliceAndKey>>>` which would be even more confusing and counterintuitive.

I've however changed the template function to ditch the Function template parameter and only accept SliceAndKey or vector of SliceAndKey

cpp/arcticdb/pipeline/test/test_rollback.cpp
ognyanstoimenov: https://chat-man.slack.com/archives/C02NEG3V38X/p1760440782690259
I haven't debugged this further to find out the exact cause so I was being cautious.

I tested that removing the unused `std::vector<TestTensorFrame> frames;`  doesn't break things so I've removed it from here

cpp/arcticdb/pipeline/test/test_rollback.cpp
alexowens90: Ah OK fair enough

cpp/arcticdb/pipeline/test/test_rollback.cpp
ognyanstoimenov: ASAN fails with:
`2025-11-11T00:39:59.1906750Z ==4174==ERROR: AddressSanitizer: heap-use-after-free on address 0x521007364180 at pc 0x000009fb8040 bp 0x7fea0e6618b0 sp 0x7fea0e6618a8`
If I remove this though so I am bringing it back in https://github.com/man-group/ArcticDB/pull/2753

PR #2742, opened 2025-10-30
Improve recursive normalized data read performance

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/18298965201
##### `read`
`batch_read_keys` read index keys of individual leaf nodes one by one during submission of read tasks. This PR has made this step runs in parallel in C++ layer.
It has shown read performance improvment, espceially on slow network or data with more leaf nodes:
| Read                  | Time(s)                                                    |                  |                                    |                  |
|-----------------------|------------------------------------------------------------|-----------------:|------------------------------------|------------------|
|                       | **Remote AWS**                                                    |                  | **Local S3 Storage (moto)**        |                  |
|                       | **Before**                                                 | **After**        | **Before**                         | **After**        |
| 200 Large Dataframe   | 98.4112                                                    | 50.547           | 27.7294                            | 25.2147          |
| 2000 Small Dataframe  | 159.712                                                    | 9.73144          | 33.0835                            | 10.7383          |

##### `batch_read`
It has been changed to unify to code path with `read`, Now node keys are read in the same chain of root keys.
The performance has not bettered or worsened, as expected.
| Read                  | Time(s)            |          |                                |          |
|-----------------------|--------------------|----------|--------------------------------|----------|
|                       | **Remote AWS**     |          | **Local S3 Storage (moto)**    |          |
|                       | **Before**         | **After**| **Before**                     | **After**|
| 2000 Symbols × 200 Dataframe | 7.379              | 7.161    | 7.224                          | 7.252    |

#### Any other comments?
ASV benchmark fails because of unreliable arrow and peakmem tests. They can be ignored.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1327,7 +1327,7 @@ std::vector<std::variant<ReadVersionWithNodesOutput, DataError>> LocalVersionedE
                         .thenValue([this,
                                     read_query =
                                             read_queries.empty() ? std::make_shared<ReadQuery>() : read_queries[idx],
-                                    read_options,
+                                    read_options = batch_read_options.at(idx),
                                     &handler_data](ReadVersionOutput&& result) {
                             auto& keys = result.frame_and_descriptor_.keys_;
                             if (keys.empty()) {


python/benchmarks/real_nested_dict_read.py
```diff
@@ -0,0 +1,83 @@
...
+    def setup_cache(self):
+        manager = self.get_library_manager()
+        lib = manager.get_library(LibraryType.PERSISTENT)
+
+        num_entries = self.params[0][0]
+        num_symbols = self.params[1][0]
+
+        for symbol_idx in range(num_symbols):
+            symbol_name = self.get_symbol_name(symbol_idx)
+            if symbol_name not in lib.list_symbols():
```
alexowens90: Why is this needed?

python/benchmarks/real_nested_dict_read.py
```diff
@@ -0,0 +1,83 @@
...
+                data = {
+                    f"{i}": {"a": pd.DataFrame({"a": [1, 2, 3]}), "b": pd.DataFrame({"b": [1, 2, 3, 4, 5, 6]})}
+                    for i in range(num_entries)
+                }
+                lib._nvs.write(symbol_name, data, recursive_normalizers=True)
+
+        manager.log_info()
+
+    def get_symbol_name(self, symbol_idx=0):
+        return f"nested_dict_{self.params[0]}_sym{symbol_idx}"
```
alexowens90: `self.params[0]` is `[1000]`, this doesn't make much sense?

python/benchmarks/recursive_normalizer.py
alexowens90, line 1: Are there already lmdb benchmarks for reading recursively normalized data?
What about for writing?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1279,6 +1279,32 @@ std::vector<std::variant<ReadVersionOutput, DataError>> LocalVersionedEngine::ba
...
                             return read_frame_for_version(store, version_info, read_query, read_options, handler_data);
                         })
+                        .thenValue([this,
+                                    read_query =
+                                            read_queries.empty() ? std::make_shared<ReadQuery>() : read_queries[idx],
+                                    read_options,
+                                    &handler_data](ReadVersionOutput&& result) {
+                            auto& keys = result.frame_and_descriptor_.keys_;
+                            if (keys.empty()) {
+                                return folly::makeFuture<ReadVersionWithNodesOutput>(
```
alexowens90: nit: I don't think you need the explicit template argument, `folly::makeFuture(ReadVersionWithNodesOutput{std::move(result), {}})` should work

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -408,14 +408,29 @@ std::variant<VersionedItem, StreamId> get_version_identifier(
...
+    } else {
+        std::vector<folly::Future<ReadVersionOutput>> node_futures;
+        node_futures.reserve(keys.size());
+        for (const auto& key : keys) {
+            node_futures.emplace_back(
+                    read_frame_for_version_async_get_context(store(), key, read_query, read_options, handler_data)
+            );
+        }
+
+        return {std::move(root_result), folly::collect(node_futures).get()};
```
alexowens90: This can segfault if the data is corrupted. If one of the calls to `read_frame_for_version_async_get_context` throws, then this function will return, and references like `read_options` will become invalid. The 3 options are:

1. Make `read_frame_for_version_async_get_context` take ownership of it's arguments (e.g. move in `key`, copy in `read_options`, I'm not sure how we treat `handler_data` elsewhere.
2. Make shared pointers for these arguments
3. Use `folly::collectAll` and then reraise if any of the tasks threw. We don't generally use this pattern as it is a slower failure mode, although could be acceptable in this case

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1279,6 +1279,32 @@ std::vector<std::variant<ReadVersionOutput, DataError>> LocalVersionedEngine::ba
...
+                                );
+                            } else {
+                                std::vector<folly::Future<ReadVersionOutput>> node_futures;
+                                node_futures.reserve(keys.size());
+                                for (const auto& key : keys) {
+                                    node_futures.emplace_back(read_frame_for_version_async_get_context(
+                                            store(), key, read_query, read_options, handler_data
+                                    ));
+                                }
+                                return folly::collect(std::move(node_futures))
```
alexowens90: See https://github.com/man-group/ArcticDB/pull/2742/files#r2526897908

cpp/arcticdb/version/version_tasks.cpp
```diff
@@ -0,0 +1,12 @@
+#include <arcticdb/version/version_tasks.hpp>
+#include <arcticdb/version/version_core.hpp>
+
+namespace arcticdb {
+std::tuple<std::shared_ptr<pipelines::PipelineContext>, VersionedItem> GetContextAndVersionedItemTask::operator()(
```
alexowens90: I would call this `SetupPipelineContextTask`, and remove the call to `generate_result_versioned_item`, it doesn't involve any IO and is not CPU intensive, so does not need to be scheduled

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2665,46 +2666,68 @@ VersionedItem generate_result_versioned_item(const std::variant<VersionedItem, S
     return versioned_item;
 }
 
+auto get_read_and_process_fn(
```
alexowens90: This returning a lambda is needlessly confusing. There should just be a function that takes the same parameters as this, plus a pipeline context, and returns a future on a `ReadVersionOutput`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2665,46 +2666,68 @@ VersionedItem generate_result_versioned_item(const std::variant<VersionedItem, S
...
-                                            {frame,
-                                             timeseries_descriptor_from_pipeline_context(
-                                                     pipeline_context, {}, pipeline_context->bucketize_dynamic_
-                                             ),
-                                             {}}
-                                    };
-                                }
-                        );
-            });
+folly::Future<ReadVersionOutput> read_frame_for_version_async_get_context(
```
alexowens90: I don't understand why this is needed? What are the consequences if you replace all calls to this with calls to `read_frame_for_version`?

cpp/arcticdb/version/version_core.hpp
```diff
@@ -47,6 +47,11 @@ struct ReadVersionOutput {
     FrameAndDescriptor frame_and_descriptor_;
 };
 
+struct ReadVersionWithNodesOutput {
```
alexowens90: This probably belongs in `read_result.hpp`

cpp/arcticdb/entity/read_result.hpp
```diff
@@ -17,13 +17,29 @@
...
 
 #include <vector>
 
 namespace arcticdb {
 
 using OutputFrame = std::variant<pipelines::PandasOutputFrame, ArrowOutputFrame>;
 
+struct ARCTICDB_VISIBILITY_HIDDEN NodeReadResult {
+    NodeReadResult(
+            const std::string& symbol, OutputFrame&& frame_data,
```
alexowens90:
```suggestion
            const StreamId& symbol, OutputFrame&& frame_data,
```

cpp/arcticdb/entity/read_result.hpp
```diff
@@ -17,13 +17,29 @@
...
+            const arcticdb::proto::descriptors::NormalizationMetadata& norm_meta
```
alexowens90: Can this be passed by rvalue reference as well?

alexowens90 (changes requested):

python/benchmarks/real_nested_dict_read.py
phoebusm: The library and data are used is persistent. So if the symbol already exists in the library, there is no need to write. 
But I have amended it anyway to switch the backend to LMDB

python/benchmarks/recursive_normalizer.py
phoebusm: Yea why not LMDB. 
I didn't touch `write` but yea it's worth adding the write asv test

cpp/arcticdb/version/version_core.cpp
phoebusm: Per discussed, I'll fold that into the caller as all calls for `read_frame_for_version` will be made async anyway

cpp/arcticdb/version/version_core.cpp
phoebusm: Per discussed, I'll fold that into the caller as all calls for read_frame_for_version will be made async anyway

cpp/arcticdb/pipeline/pipeline_utils.hpp
```diff
@@ -47,6 +47,80 @@ inline void apply_type_handlers(SegmentInMemory seg, std::any& handler_data, Out
     }
 }
 
+inline ReadResult create_python_read_result(
```
phoebusm, line 50: Just moving existing function to a new place to resolve some circular dependency issue

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2666,68 +2666,56 @@ VersionedItem generate_result_versioned_item(const std::variant<VersionedItem, S
     return versioned_item;
 }
 
-auto get_read_and_process_fn(
```
phoebusm: https://github.com/man-group/ArcticDB/pull/2742/files/1c94f45e6d16af1226a945c2040fd27da1b107cb..407257ad91088f24fef13bce2062bcc195131d98#r2541952458

python/benchmarks/recursive_normalizer.py
alexowens90: OK cool, can remove the `real_` from this filename now then

python/benchmarks/real_nested_dict_read.py
alexowens90: I think the point was slightly missed. If we added another `num_dict_entries` parameter (e.g. so it's `[1000, 2000]`, then we will want separate symbol names for the different number of dict entries

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -424,12 +424,18 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
...
-
-        return {std::move(root_result), folly::collect(node_futures).get()};
+        auto node_trys = folly::collectAll(node_futures).get();
+        std::vector<ReadVersionOutput> node_results;
+        node_results.reserve(node_trys.size());
+        std::transform(
+                std::make_move_iterator(node_trys.begin()),
+                std::make_move_iterator(node_trys.end()),
+                std::back_inserter(node_results),
+                [](auto&& try_result) { return std::move(try_result).value(); }
```
alexowens90, line 436: Does `value()` reraise the original exception if any of the collected futures failed?

python/benchmarks/real_nested_dict_read.py
phoebusm: I have rewritten this bit to make it less ambiguous, as the data is not persisted anyway

cpp/arcticdb/version/local_versioned_engine.cpp
phoebusm: Yes confirmed `folly` will rethrow the same exceptions:
 `arcticdb_ext.exceptions.KeyNotFoundException: Not found: [d:pho__1__0:0:0xe25334ac5af8535d@1763572020610151946[0,5]]`

cpp/arcticdb/pipeline/pipeline_utils.hpp
vasil-pashov, line 50: Can this be moved to a .cpp file

cpp/arcticdb/python/adapt_read_dataframe.hpp
```diff
@@ -34,7 +34,17 @@ inline py::tuple adapt_read_df(ReadResult&& ret, std::any* const handler_data) {
...
     auto multi_key_meta = python_util::pb_to_python(ret.multi_key_meta);
-    return py::make_tuple(ret.item, std::move(ret.frame_data), pynorm, pyuser_meta, multi_key_meta, ret.multi_keys);
+    py::list node_results;
+    for (auto& node_result : ret.node_results) {
+        node_results.append(py::make_tuple(
+                node_result.symbol_,
+                std::move(node_result.frame_data_),
+                python_util::pb_to_python(node_result.norm_meta_)
+        ));
+    }
```
vasil-pashov: This seems duplicated in python_utils.hpp can it be extracted in a function?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -408,60 +408,35 @@ std::variant<VersionedItem, StreamId> get_version_identifier(
...
+    auto& keys = root_result.frame_and_descriptor_.keys_;
+    if (keys.empty()) {
+        return {std::move(root_result), {}};
+    } else {
+        std::vector<folly::Future<ReadVersionOutput>> node_futures;
+        node_futures.reserve(keys.size());
+        for (const auto& key : keys) {
+            node_futures.emplace_back(read_frame_for_version(store(), key, read_query, read_options, handler_data));
         }
+        auto node_trys = folly::collectAll(node_futures).get();
```
vasil-pashov, line 429: Since we're not wrapping exceptions as in batch data can't we use folly::collect to fail faster? In that case we must be very careful not to have segfaults on failures and make all parameters for the future shared_ptrs.

cpp/arcticdb/pipeline/pipeline_utils.hpp
phoebusm: It has been marked `inline` explicitly so I prefer not to touch that

cpp/arcticdb/version/local_versioned_engine.cpp
phoebusm: Me and Alex have discussion regarding this and we think, as fail faster is not that important here, this implemetation is prefered

PR #2747, opened 2025-11-06
Enhancement 9288963244: Support for reading fixed-width string columns into Arrow tables

alexowens90:
#### Reference Issues/PRs
[9288963244](https://man312219.monday.com/boards/7852509418/pulses/9288963244)

Allows reading of `UTF_FIXED64` string columns into Arrow Tables.
Note that these have only ever been writable inside Man.

#### Benchmarking
Existing Arrow string reading benchmarks in `benchmark_arrow_reads.cpp` were run before and after this change, and confirmed to be the same.

Additional benchmarks for the fixed-width string case have been added, and run with `./benchmarks --benchmark_time_unit=us --benchmark_filter="BM_arrow_string_handler" --benchmark_min_time=2000x --benchmark_repetitions=5`. The means comparing fixed-width to dynamic strings are presented below in an easier to read format than the raw results. They show that fixed-width strings are up to x5 slower than dynamic strings, which is somewhat expected as we have to do the UTF-32 to UTF-8 conversion.

| Num rows | Num unique strings | Return type | Dynamic strings | Fixed-width strings | Ratio |
|-|-|-|-|-|-|
| 10,000 | 1 | CATEGORICAL | 54 | 58 | 1.1  |
| 100,000 | 1 | CATEGORICAL | 380 | 409 | 1.1  |
| 10,000 | 10,000 | CATEGORICAL | 352 | 747 | 2.1 |
| 100,000 | 100,000 | CATEGORICAL | 5516 | 11087 | 2.0 |
| 10,000 | 1 | LARGE_STRING | 136 | 579 | 4.3 |
| 100,000 | 1 | LARGE_STRING | 1473 | 5609 | 3.8 |
| 10,000 | 10,000 | LARGE_STRING  | 144 | 673 | 4.7 |
| 100,000 | 100,000 | LARGE_STRING  | 1680 | 7441 | 4.4 |
| 10,000 | 1 | STRING | 160 | 581 | 3.6 |
| 100,000 | 1 | STRING | 1541 | 5548 | 3.6 |
| 10,000 | 10,000 | STRING  | 182 | 666 | 3.7 |
| 100,000 | 100,000 | STRING  | 1768 | 7193 | 4.1 |

### DIFF
--- a/cpp/arcticdb/arrow/arrow_handlers.cpp
+++ b/cpp/arcticdb/arrow/arrow_handlers.cpp
@@ -92,13 +92,13 @@ void encode_variable_length(
                                 source_column,
                                 [&] ARCTICDB_LAMBDA_INLINE(const auto& en) {
                                     if (is_a_string(en.value())) {
-                                        auto strv = string_pool->get_const_view(en.value());
                                         while (last_idx <= en.idx()) {
                                             // According to arrow spec offsets must be monotonic even for null values.
                                             // Hence, we fill missing values between `last_idx` and `en.idx` so that
                                             // they contain 0 rows in string buffer
                                             dest_ptr[last_idx++] = static_cast<OffsetType>(bytes);
                                         }
+                                        auto strv = string_pool->get_const_view(en.value());
                                         bytes += strv.size();
                                         strings.emplace_back(strv);
                                         if (!populate_inverted_bitset) {
@@ -117,7 +117,12 @@ void encode_variable_length(
                         for_each_enumerated<typename source_type_info::TDT>(
                                 source_column,
                                 [&] ARCTICDB_LAMBDA_INLINE(const auto& en) {
-                                    dest_ptr[last_idx++] = static_cast<OffsetType>(bytes);
+                                    while (last_idx <= en.idx()) {
+                                        // According to arrow spec offsets must be monotonic even for null values.
+                                        // Hence, we fill missing values between `last_idx` and `en.idx` so that
+                                        // they contain 0 rows in string buffer
+                                        dest_ptr[last_idx++] = static_cast<OffsetType>(bytes);
+                                    }
                                     // Fixed-width string columns don't support None/NaN values, so is_a_string check
                                     // not required
                                     auto str = util::utf32_to_u8(string_pool->get_const_view(en.value()));


cpp/arcticdb/arrow/test/benchmark_arrow_reads.cpp
```diff
@@ -23,24 +23,36 @@ static void BM_arrow_string_handler(benchmark::State& state) {
...
 
     auto handler = ArrowStringHandler();
 
     const double sparsity_ratio = num_sparse / num_rows;
-    auto source_type_desc = TypeDescriptor{DataType::UTF_DYNAMIC64, Dimension::Dim0};
+    const auto data_type = fixed_width_strings ? DataType::UTF_FIXED64 : DataType::UTF_DYNAMIC64;
+    auto source_type_desc = TypeDescriptor{data_type, Dimension::Dim0};
     auto [dest_type_desc, extra_bytes_per_block] =
-            handler.output_type_and_extra_bytes({DataType::UTF_DYNAMIC64, Dimension::Dim0}, "col", read_options);
+            handler.output_type_and_extra_bytes({data_type, Dimension::Dim0}, "col", read_options);
```
IvoDD: Nit: Use the `source_type_desc` instead

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -33,12 +34,18 @@ void ArrowStringHandler::handle_type(
 ) {
     ARCTICDB_SAMPLE(ArrowHandleString, 0)
     util::check(field.has_ndarray(), "String handler expected array");
+    // ASCII would probably "just work", but it is impossible to test from the Python layer these days
+    // Not sure how much data has been written with these types either, so prefer to see if anybody complains about it
+    // not working before putting effort into making it work and tested
```
IvoDD, line 39: Out of curiosity: Has it ever been possible to write fixed with strings which are ascii encoded and not utf-32?
If yes, maybe a very low priority ticket for testing might be useful

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -73,28 +78,63 @@ void encode_variable_length(
...
+                        );
+                        return strings;
+                    } else { // fixed-width string type
+                        std::vector<std::string> strings;
+                        // Experimented with storing a hash map from stringpool offset to UTF-8 strings. Speeds up the
+                        // "1 unique string" case by ~40%, but makes the "all strings unique" case ~x3 slower
+                        for_each_enumerated<typename source_type_info::TDT>(
+                                source_column,
+                                [&] ARCTICDB_LAMBDA_INLINE(const auto& en) {
+                                    dest_ptr[last_idx++] = static_cast<OffsetType>(bytes);
```
IvoDD: I think we still need a while loop as in other case to populate the values between sparse (due to sparse map) data.
E.g. we can still do processing over fixed width column and end up with a sparse fixed width string column.

IvoDD: Might be a good idea to parametrize the `test_sparse_strings_from_processing_pipeline` with fixed width strings

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -162,24 +211,86 @@ void encode_dictionary(
...
+                        unique_offsets_view.reserve(source_column.row_count());
+                        for_each_enumerated<typename source_type_info::TDT>(
+                                source_column,
+                                [&] ARCTICDB_LAMBDA_INLINE(const auto& en) {
+                                    if (is_a_string(en.value())) {
+                                        if (auto it = unique_offsets_view.find(en.value());
+                                            it == unique_offsets_view.end()) {
+                                            auto entry = unique_offsets_view
+                                                                 .emplace(
+                                                                         en.value(),
```
IvoDD, line 237: Not seen in the benchmark results so def not significant but isn't:
```
if (map.find(something) == map.end()){
    map.emplace(something)
}
```
slightly less efficient than `map.try_emplace(something)` because `something` will need to get hashed twice and we'll need to iterate over the hash bucket twice (not sure how ankerl hashmap is implemented)?

Appreciate we can't do this for the fixed string case because we don't want to do the expensive `utf32_to_u8` more than necassary.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -73,28 +78,63 @@ void encode_variable_length(
...
+                                    if (is_a_string(en.value())) {
+                                        auto strv = string_pool->get_const_view(en.value());
+                                        while (last_idx <= en.idx()) {
+                                            // According to arrow spec offsets must be monotonic even for null values.
+                                            // Hence, we fill missing values between `last_idx` and `en.idx` so that
+                                            // they contain 0 rows in string buffer
+                                            dest_ptr[last_idx++] = static_cast<OffsetType>(bytes);
+                                        }
+                                        bytes += strv.size();
+                                        strings.emplace_back(std::move(strv));
```
vasil-pashov: nit: string_view is a pointer and a size and it doesn't really have move semantics. This will result in calling the copy constructor which is ok. We can just remove the move as it does nothing in this case.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -120,11 +160,13 @@ void encode_variable_length(
...
         );
-        auto string_ptr = reinterpret_cast<char*>(string_buffer.data());
-        for (auto strv : strings) {
-            memcpy(string_ptr, strv.data(), strv.size());
-            string_ptr += strv.size();
-        }
+        util::variant_match(strings, [&string_buffer](const auto& strings_or_views) {
+            auto string_ptr = reinterpret_cast<char*>(string_buffer.data());
+            for (auto string_or_view : strings_or_views) {
+                memcpy(string_ptr, string_or_view.data(), string_or_view.size());
```
vasil-pashov, line 171: Is it possible to read the strings or view directly into the string buffer instead of creating vecotrs and then copying each element in the buffer?

cpp/arcticdb/arrow/arrow_handlers.cpp
alexowens90: I honestly don't know. It would be very hard to test from Python now, but doable from C++. Added a [ticket](https://man312219.monday.com/boards/7852509418/pulses/18372170929) to remind us anyway

cpp/arcticdb/arrow/arrow_handlers.cpp
alexowens90: I left in the `try_emplace` implementation initially, and found that fixed-width strings were then around twice as fast as dynamic, which is why I changed it. While the key needs to be hashed twice, these are integers, and so will be fast, whereas the `DictEntry` construction involves a string pool lookup, and this will happen in `try_emplace` regardless of whether the key exists in the map already or not.

cpp/arcticdb/util/string_utils.cpp
```diff
@@ -63,4 +65,12 @@ std::string safe_decode(const std::string& value) {
     }
     return unescaped.str();
 }
+
+std::string utf32_to_u8(std::string_view strv) {
+    std::u32string_view strv32{reinterpret_cast<const char32_t*>(strv.data()), strv.size() / 4};
+    // Strip trailing null characters
+    strv32 = strv32.substr(0, strv32.find_first_of(char32_t(0)));
+    return boost::locale::conv::utf_to_utf<char>(strv32.data(), strv32.data() + strv32.size());
```
vasil-pashov, line 73: nit: looks like we can do this with [STL](https://en.cppreference.com/w/cpp/locale/codecvt_utf8.html) but I don't insist.

vasil-pashov (approved):

cpp/arcticdb/arrow/arrow_handlers.cpp
alexowens90: We don't support writing `None/NaN` in fixed-width string columns, so that test can't be parametrized. I can't think of any other way to generate sparse data for fixed-width string columns?

cpp/arcticdb/arrow/arrow_handlers.cpp
alexowens90: Not in a single pass, as we don't know the total length of all of the strings in the column until we get to the end.
There are 2 possible approaches I can see:

1. Do 2 passes, with the first just being to figure out the buffer size
2. Use our `ChunkedBuffer` with a large-ish preallocation of the first block and a detachable allocation mode. Then if all the strings fitted in the first block, we can just hand-off the pointer to Arrow, and otherwise it is till fewer memcpys as it is blocks rather than individual strings.

We have a ticket for aggressive performance optimisations that this could be looked at as part of.

cpp/arcticdb/util/string_utils.cpp
alexowens90: It was deprecated in C++17

PR #2759, opened 2025-11-12
Enhancement 9722907222+18364157684: Denormalize all supported Pandas index and column names into Arrow

alexowens90:
#### Reference Issues/PRs
[9722907222](https://man312219.monday.com/boards/7852509418/pulses/9722907222)
[18364157684](https://man312219.monday.com/boards/7852509418/pulses/18364157684)

Pandas is very permissive in terms of it's column naming, allowing arbitrary objects as column names, as well as duplicates, and no names at all. This also applies to index columns. Of this large universe of possible column names, we allow:

- Strings (including the empty string "")
- Integers
- `None` (== no name)
- Duplicates

Arrow is much more restrictive, requiring all column names to be unique string values. As a large amount of data has already been written with these quirky column names, we want to be able to read all of it back in as Arrow format, with some kind of sensible naming scheme.

#### Prior Art
`pyarrow.Table.from_pandas` will use a stringified version of integer and `None` column names. Unnamed index columns are named `__index_level_i__`, where `i` is the level of the index, and placed as the last columns in the table, and seems to have a bug where if there is already a column with this name, then there will be duplicate column names. Duplicate column names will raise a `ValueError`.

`polars.from_pandas` will also stringify integer and `None` column names. Polars does not include index columns from Pandas by default, `include_index=True` must be passed, in which case the columns will be placed as the first columns in the resulting dataframe. Unnamed index columns will be called `"None"`. If a multiindex is provided with multiple unnamed columns then a `ValueError` is raised. Duplicate column names will also raise a `ValueError`.

#### Design
We must be more permissive than `pyarrow` or `polars` due to the data already written with these column names. The name resolution process is as follows:

- Existing names are always respected as far as possible
- If the data has a single unnamed timeseries index, this will be named `__index__` and placed as the first column in the resulting dataframe.
- If the data has a multiindex with one or more unnamed levels, these will be named `__index_level_i__` where `i` is the level of the index without a name, and placed as the first columns in the resulting dataframe.
- Integer and `None` column names will be stringified.
- Duplicate column names will be handled by prepending and appending an underscore character until a unique name is achieved. e.g. column names `["hello", "hello", "hello"]`, would become `["hello", "_hello_", "__hello__"]`
    - This also applies to integer and `None` column names if they are repeated.
    - This also applies to unnamed index columns. e.g. if there is already a column called `__index__` (with 2 underscores), then the index column will become `___index___` (with 3 underscores).

In this way we can read back all data we have previously written, and it will generally be possible to infer what the original column name was.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
+++ b/python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
@@ -84,10 +84,10 @@ def test_named_timeseries_index_no_clash(lmdb_version_store_arrow, index_name):
 
 @pytest.mark.parametrize("index_name", ["index", "__index__", "ts"])
 def test_named_timeseries_index_clash(lmdb_version_store_arrow, index_name):
-    columns = [index_name, index_name]
+    columns = [index_name, index_name, f"_{index_name}_"]
     df = pd.DataFrame(np.zeros((1, len(columns))), columns=columns, index=[pd.Timestamp(0)])
     df.index.name = index_name
-    expected_columns = [index_name, f"_{columns[0]}_", f"__{columns[1]}__"]
+    expected_columns = [index_name, f"_{columns[0]}_", f"__{columns[1]}__", f"__{columns[2]}__"]
     generic_arrow_norm_test(lmdb_version_store_arrow, "test_named_timeseries_index_clash", df, expected_columns)
 
 


python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
```diff
@@ -1,261 +1,305 @@
...
+    expected_columns = [index_name] + columns
+    generic_arrow_norm_test(lmdb_version_store_arrow, "test_named_timeseries_index_no_clash", df, expected_columns)
+
+
+@pytest.mark.parametrize("index_name", ["index", "__index__", "ts"])
+def test_named_timeseries_index_clash(lmdb_version_store_arrow, index_name):
+    columns = [index_name, index_name]
+    df = pd.DataFrame(np.zeros((1, len(columns))), columns=columns, index=[pd.Timestamp(0)])
+    df.index.name = index_name
+    expected_columns = [index_name, f"_{columns[0]}_", f"__{columns[1]}__"]
```
IvoDD: I have mixed feelings about this behavior. 

On one hand if we only have a dataframe with unnamed index and a single column named `__index__`. It would be quite weird for the user to try and read his column `lib.read(sym).data.column('__index__')` and get the index column instead?

On the other hand if we add the underscored with preference in front of the unnamed index column it will be less easy to identify what `"____index____", "__index__", "___index___"` stands for.

Anyway probably doesn't matter that much as who names their column `__index__`? I guess I'm leaning a bit more in direction of changing it as user could see their request break even if they are not using pandas weird features and just have an odd column name.

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
```diff
@@ -1,261 +1,305 @@
...
+    columns = ["col"]
+    df = pd.DataFrame(np.zeros((1, len(columns))), columns=columns, index=[pd.Timestamp(0)])
+    df.index.name = index_name
+    expected_columns = [index_name] + columns
+    generic_arrow_norm_test(lmdb_version_store_arrow, "test_named_timeseries_index_no_clash", df, expected_columns)
+
+
+@pytest.mark.parametrize("index_name", ["index", "__index__", "ts"])
+def test_named_timeseries_index_clash(lmdb_version_store_arrow, index_name):
+    columns = [index_name, index_name]
```
IvoDD: Should we also add an edge case test where columns are:
```
# Unnamed index
col_names = ["__index__", "___index___"]
```
In this case the result should probably end up the same but we'll end up adding just a single underscore to the columns (even though they are different but the index column would trigger a rename).

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
```diff
@@ -1,261 +1,305 @@
...
+)
+def test_unnamed_multiindex(lmdb_version_store_arrow, columns):
+    df = pd.DataFrame(
+        np.zeros((1, len(columns))), columns=columns, index=pd.MultiIndex.from_product([[pd.Timestamp(0)], ["id"]])
+    )
+    index_column_names = ["__index_level_0__", "__index_level_1__"]
+    if index_column_names[0] in columns:
+        index_column_names[0] = f"_{index_column_names[0]}_"
+    if index_column_names[1] in columns:
+        index_column_names[1] = f"_{index_column_names[1]}_"
```
IvoDD, line 124: Somewhat oddly this is different to the way we handle a single index. In that case we don't add any underscores to the index but add underscores to the columns.

For multiindex it looks like we add a single underscore to the index if index name matches and add two underscores to one of the columns.

In combination with [this comment](https://github.com/man-group/ArcticDB/pull/2759/files#r2522137999) I can suggest the following algorithm to make this consistent:
```
1. Add underscores to all named columns and indices in such a way they are all distinct (similar to how we do now)
2. Set in stone the named columns and indices and not change them further
3. Decide the name of the unnamed indices in the end and add as many underscores as needed to make them unique
```
This has the benefit of not changing the column names if they are not duplicated. So users can query them as expected. Anyway I don't feel strongly about this

python/arcticdb/version_store/_normalization.py
```diff
@@ -674,9 +674,9 @@ def construct_pandas_metadata(self, fields, op: ArrowNormalizationOperations) ->
         column_index = {
             "name": None,
             "field_name": None,
-            "pandas_type": "unicode",
+            "pandas_type": "mixed",
```
IvoDD, line 677: Isn't it incorrect to always set to mixed? E.g. in 90% of cases where all columns are regular strings shouldn't this be unicode? Wouldn't this cause the below basic example to fail:
```
df = pd.DataFrame({"col": [1, 2, 3]})
lib.write(sym, df)
table = lib.read(sym).data
assert_frame_equal(table.to_pandas(), df) # Raises because table.to_pandas() has a mixed column type but df has a unicode column type?
```

Probably I'm misunderstanding something since CI passes

python/arcticdb/version_store/_normalization.py
```diff
@@ -811,30 +862,45 @@ def denormalize(self, item, norm_meta):
 
                 if index_col_idx in fake_field_pos:
                     renames_for_pandas_metadata[index_col_idx] = None
+                    new_name = f"__index_level_{index_col_idx}__"
+                    while new_name in original_column_names:
```
IvoDD, line 866: This is the bit I find surprising from the comments in the test section.
I think we should do the underscore additions in a consistent way with the single index case which is looking at the `taken_col_names` (which is empty at that point) vs here where we look at the `original_column_names`.

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
alexowens90: That isn't the behaviour, if you look at `test_unnamed_timeseries_index`, if the index is unnamed, and there is already a column called `__index__`, then the index column is named `___index___` (i.e. with 3 underscores).

python/arcticdb/version_store/_normalization.py
alexowens90: The single index case also looks at `original_column_names`?

python/arcticdb/version_store/_normalization.py
alexowens90: The test you show here passes with the new implementation, I guess `assert_frame_equal` is tolerant to this difference.
I agree that ideally this would be unicode if all strings, and mixed otherwise, but I don't think we should invest too much effort into it.

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
alexowens90: I know this implementation is still not ideal, but between supported non-string names like integers and None, duplicated column names, our `__fkidx` multiindex naming, and all of these things possibly being duplicated, it is surprisingly hard to cover off all of the possible corner cases in a consistent manner.
All I really care about is that if you wrote data with these crazy schemas, that you can read it back, anything else is a bonus.

python/tests/unit/arcticdb/version_store/test_arrow_normalization.py
IvoDD: Ah ok this covers my main concern then

python/arcticdb/version_store/_normalization.py
IvoDD: Ah sorry was confused, nvm

python/arcticdb/version_store/_normalization.py
```diff
@@ -770,25 +810,37 @@ def denormalize(self, item, norm_meta):
...
                     # We apply timezone metadata only when the first column is a timestamp column.
                     # This matches the behavior and is required to handle `groupby`s which can change index type.
                     # TODO: This is still not correct if grouping by a timestamp column. Monday ref: 18197986461
                     timezones[0] = index_meta.tz
-                if index_meta.name:
-                    renames_for_table[0] = index_meta.name
                 if index_meta.fake_name:
                     renames_for_pandas_metadata[0] = None
+                    new_name = "__index__"
+                    while new_name in original_column_names:
```
vasil-pashov, line 837: I think it would be easier to follow if `taken_column_names` is initialized to be the same as `original_column_names` and then only check if `new_name in taken_column_names` instead of having some while loops check `original_column_names` as here and some check `taken_column_names` as in line 875

vasil-pashov (approved):

python/arcticdb/version_store/_normalization.py
alexowens90: The problem is that `taken_column_names` is not always the same as `original_column_names`, for instance if there are duplicated names, `None` column names, or integer column names

PR #2763, opened 2025-11-18
[18344690269] Add batch APIs for configurable strings

IvoDD:
#### Reference Issues/PRs
Monday ref: 18344690269

- Creates a new `BatchReadOptions` which separates batch level options and per symbol level options.
- Exposes APIs both on V1 and V2 for `read_batch`, `read_batch_and_join` and tests for them

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/pipeline/read_options.hpp
+++ b/cpp/arcticdb/pipeline/read_options.hpp
@@ -100,12 +100,18 @@ using ReadOptionsPerSymbol = std::variant<ReadOptions, std::vector<ReadOptions>>
 
 struct BatchReadOptionsData {
     ReadOptionsPerSymbol read_options_per_symbol_;
-    std::optional<bool> batch_throw_on_error_;
+    bool batch_throw_on_error_;
     OutputFormat output_format_ = OutputFormat::PANDAS;
+
+    BatchReadOptionsData(bool batch_throw_on_error) : batch_throw_on_error_(batch_throw_on_error) {}
 };
 
 struct BatchReadOptions {
-    std::shared_ptr<BatchReadOptionsData> data_ = std::make_shared<BatchReadOptionsData>();
+    std::shared_ptr<BatchReadOptionsData> data_;
+
+    BatchReadOptions(bool batch_throw_on_error) {
+        data_ = std::make_shared<BatchReadOptionsData>(batch_throw_on_error);
+    }
 
     void set_read_options(const ReadOptions& read_options) { data_->read_options_per_symbol_ = read_options; }
 
@@ -123,7 +129,7 @@ struct BatchReadOptions {
 
     void set_batch_throw_on_error(bool batch_throw_on_error) { data_->batch_throw_on_error_ = batch_throw_on_error; }
 
-    [[nodiscard]] const std::optional<bool>& batch_throw_on_error() const { return data_->batch_throw_on_error_; }
+    [[nodiscard]] bool batch_throw_on_error() const { return data_->batch_throw_on_error_; }
 
     void set_output_format(OutputFormat output_format) { data_->output_format_ = output_format; }
 
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -539,12 +539,6 @@ std::vector<std::variant<DescriptorItem, DataError>> LocalVersionedEngine::batch
         const std::vector<StreamId>& stream_ids, const std::vector<VersionQuery>& version_queries,
         const BatchReadOptions& batch_read_options
 ) {
-
-    internal::check<ErrorCode::E_ASSERTION_FAILURE>(
-            batch_read_options.batch_throw_on_error().has_value(),
-            "ReadOptions::batch_throw_on_error_ should always be set here"
-    );
-
     auto opt_index_key_futs = batch_get_versions_async(store(), version_map(), stream_ids, version_queries);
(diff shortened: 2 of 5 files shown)


python/tests/unit/arcticdb/version_store/test_arrow_read.py
```diff
@@ -1125,3 +1125,32 @@ def test_sparse_strings_from_processing_pipeline(lmdb_version_store_dynamic_sche
...
+    sym_1, sym_2 = "sym_1", "sym_2"
+    df_1 = pd.DataFrame({"col_1": ["a", "a", "bb"], "col_2": ["x", "y", "z"]})
+    df_2 = pd.DataFrame({"col_1": ["a", "aa", "aaa"], "col_2": ["a", "a", "a"]})
+    lib.batch_write([sym_1, sym_2], [df_1, df_2])
+
+    arrow_string_format_default = ArrowOutputStringFormat.SMALL_STRING
+    per_symbol_arrow_string_format_default = [None, ArrowOutputStringFormat.LARGE_STRING]
+    per_symbol_arrow_string_format_per_column = [
+        {"col_1": ArrowOutputStringFormat.CATEGORICAL},
+        {"col_2": ArrowOutputStringFormat.CATEGORICAL},
```
alexowens90, line 1142: Not specific to batch methods, but I assume it is just silently ignored if you provide a column here that doesn't exist, as it boils down to a map lookup?

python/tests/unit/arcticdb/test_arrow_api.py
```diff
@@ -324,3 +325,78 @@ def test_read_arctic_strings(
...
+    read_requests = [
+        ReadRequest(
+            symbol=sym_1, arrow_string_format_default=default_1, arrow_string_format_per_column={"col_2": per_column_1}
+        ),
+        ReadRequest(
+            symbol=sym_2,
+            arrow_string_format_default=default_2,
+            arrow_string_format_per_column={
+                "col_2": ArrowOutputStringFormat.CATEGORICAL,
+                "col_3": ArrowOutputStringFormat.LARGE_STRING,
```
alexowens90: It's a bit weird to specify this in individual `ReadRequest`s. `read_batch_and_join` is kind of a hybrid of `read` and `read_batch`, but when it comes to the output format, it should look more like `read`

python/tests/unit/arcticdb/version_store/test_arrow_read.py
IvoDD: Yes it is silently ignored if you provide a non existent column. 

This is useful for the join API where you can use the same map for each symbol (even though not every symbol will have every column). I added an example of this in the batch_read_and_join test.

python/tests/unit/arcticdb/test_arrow_api.py
IvoDD: Changed as discussed offline to instead happen via:
```
lazy_frames = lib.read_batch(syms, arrow_string_format_default=something arrow_string_format_per_column=something)
concat(lazy_frames).collect() # Here we will use the string configurations for the entire batch to be applied to the joined frame.
```

alexowens90 (approved):

cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -99,4 +95,39 @@ struct ReadOptions {
...
+        data_->read_options_per_symbol_ = read_options_per_symbol;
+    }
+
+    [[nodiscard]] ReadOptions at(size_t idx) const {
+        return util::variant_match(
+                data_->read_options_per_symbol_,
+                [&](const std::vector<ReadOptions>& read_options) { return read_options.at(idx); },
+                [&](ReadOptions read_options) { return read_options; }
+        );
+    }
```
vasil-pashov, line 128: Is `ReadOptions` cheap to copy because this will copy them? I'd change the signature to const ReadOptions&. In that case we should capture read_options by const ref as well in the visitor

nit: vector.at will always perform bounds check. Is that intentional?

cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -99,4 +95,39 @@ struct ReadOptions {
 
     [[nodiscard]] ReadOptions clone() const { return ReadOptions(std::make_shared<ReadOptionsData>(*data_)); }
 };
+
+using ReadOptionsPerSymbol = std::variant<ReadOptions, std::vector<ReadOptions>>;
+
+struct BatchReadOptionsData {
+    ReadOptionsPerSymbol read_options_per_symbol_;
+    std::optional<bool> batch_throw_on_error_;
```
vasil-pashov: I can see that this was moved from `ReadOptionsData` but optional<bool> doesn't make much sense to me in general. No need to change it though I just wanted to complain about the use of it in general.

cpp/arcticdb/pipeline/read_options.hpp
IvoDD: yes, `ReadOptions` is a `shared_ptr<ReadOptionsData>`, so better to pass by value.

Yeah I don't have a strong opinion on the `at`. It is not on the hot path and thought I could leave the bounds checking to more easily debug while developing if construct the `read_options` incorectly, So I'm thinking to leave as is.

cpp/arcticdb/pipeline/read_options.hpp
IvoDD: yeah I agree it doesn't make sense to be an optional. Will update.

PR #2766, opened 2025-11-18
Performance 10588043604: Decode only required columns in processing pipeline with dynamic schema

alexowens90:
#### Reference Issues/PRs
[10588043604](https://man312219.monday.com/boards/7852509418/pulses/10588043604)

`PipelineContext::overall_column_bitset_` is used to construct a hash set of columns that are to be decoded in the processing pipeline path.
Prior to this change, this bitset was only populated for static schema and (unsupported) bucketize dynamic schema data. This resulted in all columns being decoded with dynamic schema, which becomes a performance bottleneck with very wide dataframes where only a few columns are required for the processing and output.

Test coverage already fairly comprehensive in `test_basic_version_store.py::test_dynamic_schema_read_columns`, and `read_index` behaviour in `test_read_index.py`.

### DIFF
--- a/cpp/arcticdb/async/tasks.cpp
+++ b/cpp/arcticdb/async/tasks.cpp
@@ -49,6 +49,7 @@ pipelines::SegmentAndSlice DecodeSliceTask::decode_into_slice(storage::KeySegmen
     ARCTICDB_TRACE(log::codec(), "Creating segment");
     SegmentInMemory segment_in_memory(std::move(descriptor));
     decode_into_memory_segment(seg, hdr, segment_in_memory, desc);
+    segment_in_memory.set_row_data(std::max(segment_in_memory.row_count() - 1, ranges_and_key_.row_range().diff() - 1));
     return pipelines::SegmentAndSlice(std::move(ranges_and_key_), std::move(segment_in_memory));
 }
 } // namespace arcticdb::async
--- a/cpp/arcticdb/pipeline/read_pipeline.hpp
+++ b/cpp/arcticdb/pipeline/read_pipeline.hpp
@@ -156,9 +156,7 @@ std::vector<FilterQuery<ContainerType>> get_column_bitset_and_query_functions(
         bool column_groups
 ) {
     using namespace arcticdb::pipelines::index;
-    if (!dynamic_schema || column_groups) {
-        get_column_bitset_in_context(query, pipeline_context);
-    }
+    get_column_bitset_in_context(query, pipeline_context);
     return build_read_query_filters<ContainerType>(pipeline_context, query.row_filter, dynamic_schema, column_groups);
 }
 
--- a/cpp/arcticdb/pipeline/query.hpp
+++ b/cpp/arcticdb/pipeline/query.hpp
@@ -264,10 +264,9 @@ void build_col_read_query_filters(
         };
         queries.push_back(std::move(query));
     } else if (pipeline_context->overall_column_bitset_) {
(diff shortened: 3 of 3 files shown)


cpp/arcticdb/async/tasks.cpp
```diff
@@ -49,6 +49,7 @@ pipelines::SegmentAndSlice DecodeSliceTask::decode_into_slice(storage::KeySegmen
     ARCTICDB_TRACE(log::codec(), "Creating segment");
     SegmentInMemory segment_in_memory(std::move(descriptor));
     decode_into_memory_segment(seg, hdr, segment_in_memory, desc);
+    segment_in_memory.set_row_data(std::max(segment_in_memory.row_count() - 1, ranges_and_key_.row_range().diff() - 1));
```
IvoDD, line 52: When can the `row_range.diff()` be different than the decoded segment size?

Also out of curiosity was this change needed to make this PR work or it's just a general improvement? I agree it is more correct to e.g. populate the sparse map if a column is missing from the segment.

cpp/arcticdb/pipeline/read_pipeline.hpp
```diff
@@ -156,9 +156,7 @@ std::vector<FilterQuery<ContainerType>> get_column_bitset_and_query_functions(
         bool column_groups
 ) {
     using namespace arcticdb::pipelines::index;
-    if (!dynamic_schema || column_groups) {
-        get_column_bitset_in_context(query, pipeline_context);
-    }
+    get_column_bitset_in_context(query, pipeline_context);
```
IvoDD, line 159: Definetely not for this PR:

I think this will have a side effect of making something like `read(columns=all_cols_but_one)` slower for very short and wide dataframes.

It looks like there is a bunch of avoidable column name string munging operations (e.g. a bunch of copying [here](https://github.com/man-group/ArcticDB/blob/85f14e3afcae2e7760fb883a610021616e1e535c/cpp/arcticdb/async/tasks.cpp#L23-L33))

IvoDD (approved):

cpp/arcticdb/async/tasks.cpp
alexowens90: The decoded segment size is based on the longest column it sees while decoding. This change was needed to make the PR work in the case where:

- `columns=[]` (i.e. just the index column[s] on the V2 API)
- The data has a range index
- Some processing (e.g. head or tail) are applied

Covered in [these tests](https://github.com/man-group/ArcticDB/blob/master/python/tests/unit/arcticdb/version_store/test_read_index.py#L60-L74)

cpp/arcticdb/pipeline/read_pipeline.hpp
alexowens90: Possibly yeh. This is all a bit of a mess. We have [4](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/pipeline/pipeline_context.hpp#L119-L129) different fields in the pipeline context right now controlling slightly different aspects of this. The (hugely unnecessary) complexity of the code in `query.hpp` make it hard to change confidently, even though all it is doing is picking out row and column slices from the index to read.
I think there is a multi-stage refactor needed. One is to eliminate the pipeline context from everything after the data keys are read, and another to cleanup everything to do with column selection and index key filtering. I think the impetus will come from either:

1. The current index key filtering is too slow for huge data
2. We fully implement bucketize dynamic, when a lot of this code will need to be touched anyway

PR #2769, opened 2025-11-20
[18362773160] Introduce `EXPERIMENTAL_POLARS` output format

IvoDD:
#### Reference Issues/PRs
Monday ref: 18362773160

Polars output format is just a thin wrapper around the arrow output format. We create the polars dataframe zero-copy from the pyarrow table.

Also improves some docs.

Adds just a few extra tests for polars because:
- Extensive arrow testing covers most arrow related logic
- Parametrizing many tests to work with polars is difficult because `polars.DataFrame` does not have any concept of pandas metadata.

#### Any other comments?
I decided to go through `pyarrow` even though we could avoid the pyarrow dependency by using a [PyCapsule](https://docs.pola.rs/user-guide/misc/arrow/#using-polars-directly), because:
- This would require rewriting our arrow denormalization (which currently relies on pyarrow APIs)
- This would require extra testing coverage of polars output format. And it is harder to parametrize our existing tests because polars doesn't have a concept a pandas metadata.

Also needed to clean up some space for conda build. This is done in the conda workflow. A successful run with workflow from this branch can be seen [here](https://github.com/man-group/ArcticDB/actions/runs/19636884795/job/56229690182).

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_with_conda.yml
+++ b/.github/workflows/build_with_conda.yml
@@ -65,6 +65,13 @@ jobs:
       mongodb:
         image: mongo:4.4
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@v1.3.1
+        with:
+          tool-cache: false
+          large-packages: false # Time-consuming but doesn't save that much space (4GB)
+          docker-images: false  # We're using docker images we don't want to clear
+
       - uses: actions/checkout@v3.3.0
         # DONT use recursive submodules checkout to simulate conda feedstock build
         # with:


python/arcticdb/options.py
```diff
@@ -151,14 +151,34 @@ def __repr__(self):
...
     EXPERIMENTAL_ARROW = "EXPERIMENTAL_ARROW"
+    EXPERIMENTAL_POLARS = "EXPERIMENTAL_POLARS"
 
 
 def output_format_to_internal(output_format: Union[OutputFormat, str]) -> InternalOutputFormat:
     if output_format.lower() == OutputFormat.PANDAS.lower():
         return InternalOutputFormat.PANDAS
-    elif output_format.lower() == OutputFormat.EXPERIMENTAL_ARROW.lower():
+    elif output_format.lower() in [OutputFormat.EXPERIMENTAL_ARROW.lower(), OutputFormat.EXPERIMENTAL_POLARS.lower()]:
         if not _PYARROW_AVAILABLE:
```
alexowens90, line 181: I don't think any dependency resolver would let us get into a position where pyarrow is available and polars isn't, but we may as well check just in case

python/arcticdb/version_store/_store.py
```diff
@@ -1196,10 +1197,10 @@ def get_column_stats_info(
         version_query = self._get_version_query(as_of, **kwargs)
         return self.version_store.get_column_stats_info_version(symbol, version_query).to_map()
 
-    def _batch_read_keys(self, atom_keys, read_options):
+    def _batch_read_keys(self, atom_keys, read_options, output_format):
```
alexowens90: @phoebusm has removed this function entirely as part if the recursive normalizers performance improvement PR, so whoever is merging second will get a merge conflict

python/arcticdb/version_store/_store.py
```diff
@@ -1331,7 +1331,7 @@ def _batch_read_to_versioned_items(
         # Take a copy as _get_read_queries can modify the input argument, which makes reusing the input counter-intuitive
         query_builder = copy.deepcopy(query_builder)
         read_queries = self._get_read_queries(len(symbols), date_ranges, row_ranges, columns, query_builder)
-        batch_read_options = self._get_batch_read_options(
+        batch_read_options, output_format = self._get_batch_read_options(
```
alexowens90, line 1362: Couldn't we expose the `output_format` as a read-only property of the `batch_read_options` C++ object?

IvoDD: It is exposed but both the `ReadOptions` and `BatchReadOptions` have only the internal c++ `InternalOutputFormat` which in both cases is just `ARROW`. 

Here I need to differenciate between the python level `OutputFormat` which is different for `PYARROW` and `POLARS`.
I think I'd like to keep it this way because C++ layer doesn't need to know whether it's pyarrow or polars. Both are only python suger on top of the arrow c structures.

PR #2776, opened 2025-11-21
Arrow read support (read directly as `pyarrow.Table` or `polars.DataFrame`)

IvoDD:
## New Feature: PyArrow and Polars Output Formats

ArcticDB now supports returning data in Apache Arrow-based formats.

### Overview

In addition to the existing `PANDAS` output format (which remains the default), you can now specify `PYARROW` or `POLARS` output formats for read operations. Both formats use Apache Arrow's columnar memory layout, enabling:

- **Better performance**: Especially for dataframes with many string columns
- **Integration with Modern systems**: Native zero-copy integration with Arrow-based data processing tools like `polars` and `duckdb`

### Usage

*For detailed usage see the [arrow notebook](https://github.com/man-group/ArcticDB/blob/official-arrow-read-support/docs/mkdocs/docs/notebooks/ArcticDB_demo_read_as_arrow.ipynb)*

Output format can be configured at three levels:

**1. Arctic instance level (default for all libraries):**
```python
import arcticdb as adb
from arcticdb import OutputFormat

# Set default for all libraries from this Arctic instance
ac = adb.Arctic(uri, output_format=OutputFormat.PYARROW)
```

**2. Library level (default for all reads from this library):**
```python
# Set at library creation
lib = ac.create_library("my_lib", output_format=OutputFormat.POLARS)

# Or when getting an existing library
lib = ac.get_library("my_lib", output_format=OutputFormat.PYARROW)
```

**3. Per read operation (most granular control):**
```python
# Override on a per-read basis
result = lib.read("symbol", output_format=OutputFormat.PYARROW)
table = result.data  # Returns pyarrow.Table

result = lib.read("symbol", output_format=OutputFormat.POLARS)
df = result.data  # Returns polars.DataFrame
```

### Output Format Options

- **`OutputFormat.PANDAS`** (default): Returns `pandas.DataFrame` backed by numpy arrays
- **`OutputFormat.PYARROW`**: Returns `pyarrow.Table` objects
- **`OutputFormat.POLARS`**: Returns `polars.DataFrame` objects

### String Format Customization

When using Arrow-based output formats, you can customize how string columns are encoded using `ArrowOutputStringFormat`:

**`LARGE_STRING`** (default):
- 64-bit variable-size encoding
- Best for general-purpose use and large string data
- PyArrow: `pa.large_string()`, Polars: `pl.String`

**`SMALL_STRING`**:
- 32-bit variable-size encoding
- More memory efficient for smaller string data
- PyArrow: `pa.string()`, Polars: Not supported

**`CATEGORICAL`** / **`DICTIONARY_ENCODED`**:
- Dictionary-encoded with int32 indices
- Best for low cardinality columns (few unique values repeated many times)
- Deduplicates strings for significant memory savings
- PyArrow: `pa.dictionary(pa.int32(), pa.large_string())`, Polars: `pl.Categorical`

### Example: Using String Formats

```python
from arcticdb import ArrowOutputStringFormat

# Set default string format at Arctic level
ac = adb.Arctic(
    uri,
    output_format=OutputFormat.PYARROW,
    arrow_string_format_default=ArrowOutputStringFormat.CATEGORICAL
)

# Or per-column during read
result = lib.read(
    "symbol",
    output_format=OutputFormat.PYARROW,
    arrow_string_format_per_column={
        "country": ArrowOutputStringFormat.CATEGORICAL,  # Low cardinality
        "description": ArrowOutputStringFormat.LARGE_STRING  # High cardinality
    }
)
```

### Supported Operations

All read operations support the new output formats:
- `read()`
- `read_batch()`
- `read_batch_and_join()`
- `head()`
- `tail()`
- LazyDataFrame operations with `lazy=True`

### Notes

- The default output format remains `PANDAS` for backward compatibility
- Both `PYARROW` and `POLARS` formats share the same underlying Arrow memory layout
- For more details on Arrow's physical layouts, see: https://arrow.apache.org/docs/format/Columnar.html

### DIFF
--- a/docs/mkdocs/docs/notebooks/ArcticDB_demo_read_as_arrow.ipynb
+++ b/docs/mkdocs/docs/notebooks/ArcticDB_demo_read_as_arrow.ipynb
@@ -1056,7 +1056,10 @@
    "source": [
     "## Record Batch Structure\n",
     "\n",
-    "Large writes, appends, and updates result in Arrow based output formats composed of multiple record batches. Each record batch corresponds to a row-slice in ArcticDB's storage.\n",
+    "Arrow tables returned from ArcticDB are split in record batches. Each record batch corresponds to a row-slice in ArcticDB's storage.\n",
+    "\n",
+    "ArcticDB will slice data by rows if it was written in multiple operations (e.g. with append in example below) or if a large (>100k rows) dataframe is written in a single operation.\n",
+    "\n",
     "This applies to both Pyarrow and Polars. With pyarrow we can see the record batch separation directly."
    ]
   },
@@ -1435,7 +1438,9 @@
    "source": [
     "## Index Column Naming: `__index__`\n",
     "\n",
-    "When Pandas DataFrames have unnamed indexes, PyArrow/Polars output formats use the special column name `__index__`:"
+    "When Pandas DataFrames have unnamed indexes, PyArrow/Polars output formats use the special column name `__index__`:\n",
+    "\n",
+    "*Note that if there are duplicate column names (e.g. column `__index__` already exists) all duplicates are resolved by adding extra underscores to beginning and end of column name*"
    ]
   },
   {
--- a/python/tests/unit/arcticdb/version_store/test_arrow_read.py
+++ b/python/tests/unit/arcticdb/version_store/test_arrow_read.py
@@ -94,6 +94,20 @@ def test_read_empty(lmdb_version_store_arrow):
     assert_frame_equal_with_arrow(table, expected)
 
 
+def test_read_empty_column_filter(lmdb_version_store_arrow):
+    lib = lmdb_version_store_arrow
+    sym = "sym"
+    df = pd.DataFrame({"col": [1, 2, 3]})
+    lib.write(sym, df)
+    # Table has no rows (because it has no columns)
+    table = lib.read(sym, columns=[], implement_read_index=True).data
+    assert table.column_names == []
+    assert table.shape == (0, 0)
+    # However in pandas we can have 3 rows even when no columns because pandas includes the row range index
+    pandas_df = lib.read(sym, columns=[], implement_read_index=True, output_format=OutputFormat.PANDAS).data
+    assert pandas_df.shape == (3, 0)
+
+
 @pytest.mark.skipif(IS_PANDAS_ONE, reason="Different empty frame handling in pandas 1.x")
 def test_read_empty_with_columns(lmdb_version_store_arrow):
     lib = lmdb_version_store_arrow


python/arcticdb/version_store/_store.py
```diff
@@ -2728,8 +2728,8 @@ def _adapt_frame_data(self, frame_data, norm, output_format):
                 )
             if self._test_convert_arrow_back_to_pandas:
                 data = convert_arrow_to_pandas_for_tests(data)
-            if output_format.lower() == OutputFormat.EXPERIMENTAL_POLARS.lower():
-                data = pl.from_arrow(data)
+            if output_format.lower() == OutputFormat.POLARS.lower():
+                data = pl.from_arrow(data, rechunk=False)
```
IvoDD, line 2729: This `rechunk=False` is the only behavior change apart from the enum renames.

It is done to get better performance on read (rechunking requires memory copies and allocations).
In most cases the chunks will be of reasonable sizes (100k rows), so the rechunking would not provide much benefit on polars operations.

Users can always call `rechunk` on their dataframes if needed and an example is given in the notebook.

python/arcticdb/version_store/_store.py
```diff
@@ -1284,21 +1284,17 @@ def batch_read(
...
+            See `ArrowOutputStringFormat` documentation for details on available string formats.
+        arrow_string_format_per_column: Optional[Dict[str, Union[ArrowOutputStringFormat, "pa.DataType"]]], default=None
+            Per-column overrides for `arrow_string_format_default`. Keys are column names.
+            Applied to all symbols without a `per_symbol_arrow_string_format_per_column` set.
+        per_symbol_arrow_string_format_default: Optional[List[Optional[Union[ArrowOutputStringFormat, "pa.DataType"]]]], default=None
+            Per-symbol override for `arrow_string_format_default`. Overrides the batch-level default.
+            i-th entry corresponds to i-th element of `symbols`.
+        per_symbol_arrow_string_format_per_column: Optional[List[Optional[Dict[str, Union[ArrowOutputStringFormat, "pa.DataType"]]]]], default=None
+            Per-symbol, per-column overrides. Takes precedence over all other string format settings.
+            i-th entry corresponds to i-th element of `symbols`.
```
alexowens90, line 1297: It's a bit odd that these are full arguments to `batch_read`, but undocumented kwargs to `read/head/tail`

python/arcticdb/arctic.py
```diff
@@ -268,14 +267,16 @@ def create_library(
...
+            See `OutputFormat` documentation for details on available formats.
 
         arrow_string_format_default: Optional[Union[ArrowOutputStringFormat, "pa.DataType"]], default=None
-            If using `output_format=EXPERIMENTAL_ARROW` it sets the output format of string columns for arrow.
-            See documentation of `ArrowOutputStringFormat` for more information on the different options.
-            If `None` uses the default arrow_string_format from the `Library` instance.
+            Default string column format when using `PYARROW` or `POLARS` output formats on this library.
+            If `None`, uses the `arrow_string_format_default` from the `Arctic` instance.
+            Can be overridden per read operation.
+            See `ArrowOutputStringFormat` documentation for details on available string formats.
```
alexowens90, line 279: I think we should have a note that these are not persisted lib config settings, they just apply to the returned library object

python/arcticdb/options.py
```diff
@@ -195,24 +214,30 @@ def output_format_to_internal(output_format: Union[OutputFormat, str]) -> Intern
...
+        PyArrow: `pa.large_string()`, Polars: `pl.String`
+        Supports up to 2⁶³-1 bytes total string length per Arrow array.
+        Best for general-purpose use and when working with large string data.
 
     SMALL_STRING:
-    Produces string columns with type `pa.string()`. Total length of strings must fit in a 32-bit integer.
-    Does not deduplicate strings, so has better performance for columns with many unique strings.
-    Slightly faster than `LARGE_STRING` but does not work with very long strings.
+        Uses 32-bit variable-size encoding.
+        PyArrow: `pa.string()`, Polars: `pl.String`
```
alexowens90: Does this mean Polars converts everything to large string?

docs/mkdocs/docs/notebooks/ArcticDB_demo_read_as_arrow.ipynb
alexowens90, line 1: - "batche" typo
- Warning about lmdb already being open in the process should be removed if possible from the final version
- We should use [set_sorted](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.set_sorted.html) on timeseries index columns to speed up subsequent query performance [when we know this to be the case](https://github.com/man-group/ArcticDB/blob/master/cpp/proto/arcticc/pb2/descriptors.proto#L74) (probably a separate PR)
- `tz_data` symbol use an unnamed index before that behaviour has been introduced, I would just give it a name to avoid confusion when reading the notebook from start to finish
- In the filtering and column selection section, worth mentioning that timeseries indexes are always fetched, even if not specified in the `columns` list
- "PyArrow is useful when you need direct access to Arrow's columnar format" I didn't understand what this means exactly? Does it expose more info on chunks than Polars?
- In the duckdb example, is there an alternative to `to_df()` that returns Polars or PyArrow, to show Pandas isn't necessary in that workflow?
- `print(f"Number of record batches: {arrow_table.num_rows}")` seems wrong
- "When Pandas DataFrames have unnamed indexes, Arrow/Polars use the special column name __index__" Polars uses "None"
- "Converting to Polars/PyArrow and back preserves Pandas metadata" - I thought Polars dropped the Pandas metadata?
- Might be worth a cell on unnamed multiindexes

python/arcticdb/version_store/_store.py
IvoDD: Yeah I separated them from the `**kwargs` (in a previous PR) because they have a different behavior than the corresponding arguments for the other methods.

I.e. most kwargs are just passed around to `get_read_options` but here I want to explicitly override `arrow_string_format_default` with the `per_symbol` variant.

In general I'm very unhappy with the state of docs on the v1 API. I think we should expand all `**kwargs` to properly named arguments similar to the v2 api but this would be a significant effort.

If you prefer I can move these to undocumented `**kwargs` for consistency?

python/arcticdb/options.py
IvoDD: Ah interesting it indeed does:
```
>>> table
pyarrow.Table
col: string
----
col: [["str_0","str_1","str_2","str_3","str_4",...,"str_99995","str_99996","str_99997","str_99998","str_99999"]]
>>> pl_frame = pl.from_arrow(table)
>>> pl_frame.to_arrow()
pyarrow.Table
col: large_string
----
col: [["str_0","str_1","str_2","str_3","str_4",...,"str_99995","str_99996","str_99997","str_99998","str_99999"]]
```

I'll just disallow `SMALL_STRING` string output format for polars and raise an appropriate error message.

python/arcticdb/version_store/_store.py
alexowens90: I think it's OK, as you say, the V1 API docs aren't consistent, and kwargs make it error prone

PR #2778, opened 2025-11-25
No pickle V2 meta structure normalization for recursive normalizer

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/9686142250
Introduce new normalization (V2) for recursive normalizer metastructure as the existing one requires pickle.
Please note that data with V2 normalized metastructure is not readable by existing releases.
Plan of introduction:
| Version | Change |
|---------|--------|
| >= v6.7.0 | Deprecation warning when writing V1 meta structure; V2 meta structure can be enabled optionally |
| >= v7.0.0 | V2 meta structure will be enabled by default |
#### Any other comments?
I have noticed that currently, if custom normalizer returns one single object as the output of normalization, and it's used as part of the recursive normalized data, the object will not be denormalized by the custom normalizer during read. I think it behaves like this since day 1 so I'll leave it as it is.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/docs/mkdocs/docs/runtime_config.md
+++ b/docs/mkdocs/docs/runtime_config.md
@@ -104,7 +104,7 @@ Values:
 * 0: Disable
 * 1: Enable (Default)
 
-### VersionStore.RecursiveNormalizerMetastructureV2
+### VersionStore.RecursiveNormalizerMetastructure
 
 Controls whether the recursive normalizer will use meta structure V2
 
@@ -122,7 +122,7 @@ Controls whether the recursive normalizer will use meta structure V2
 | >= v6.7.0 | Deprecation warning when writing V1 meta structure; V2 meta structure can be enabled optionally |
 | >= v7.0.0 | V2 meta structure will be enabled by default |
 
-Values: * 0: Disable (Default) * 1: Enable
+Values: * 1: V1 (Default) * 2: V2  
 
 ### VersionStore.VersionStore.RecursiveNormalizerMetastructureV1DeprecationWarning
 
--- a/python/arcticdb/flattener.py
+++ b/python/arcticdb/flattener.py
@@ -35,7 +35,7 @@ class Flattener:
 
     def __init__(self):
         self.custom_normalizer = get_custom_normalizer(False)
-        self.meta_structure_v2 = get_config_int("VersionStore.RecursiveNormalizerMetastructureV2") == 1
+        self.meta_structure_v2 = get_config_int("VersionStore.RecursiveNormalizerMetastructure") == 2
         self.meta_structure_v1_deprecation_warning = (
             get_config_int("VersionStore.RecursiveNormalizerMetastructureV1DeprecationWarning") != 0
         )
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -51,6 +51,7 @@ from arcticdb.log import version as log
 from arcticdb_ext.log import LogLevel
 from arcticdb.version_store._common import _column_name_to_strings, TimeFrame
 from arcticdb.options import OutputFormat
+from arcticdb_ext import get_config_int
 
 PICKLE_PROTOCOL = 4
 
@@ -1718,13 +1719,14 @@ def normalize_metadata(metadata: Any) -> UserDefinedMetadata:
 
 
 def normalize_recursive_metastruct(metastruct: Dict[Any, Any]) -> UserDefinedMetadata:
-    from arcticdb.flattener import Flattener
-
-    meta_structure_v2 = Flattener().meta_structure_v2
-    if meta_structure_v2 is not True and Flattener().meta_structure_v1_deprecation_warning:
+    meta_structure_v2 = get_config_int("VersionStore.RecursiveNormalizerMetastructure") == 2
+    meta_structure_v1_deprecation_warning = (
+        get_config_int("VersionStore.RecursiveNormalizerMetastructureV1DeprecationWarning") != 0
+    )
+    if meta_structure_v2 is not True and meta_structure_v1_deprecation_warning:
         log.warn(
             "Recursive normalization metastruct in use (V1) is going to be deprecated after ArcticDB v7.0.0 release. "
-            "Please refer to https://docs.arcticdb.io/latest/runtime_config/#versionstorerecursivenormalizermetastructurev2 for more details."
+            "Please refer to https://docs.arcticdb.io/latest/runtime_config/#versionstorerecursivenormalizermetastructure for more details."
             "V2 recursive normalization metastruct can be read by >= v6.7.0 releases. Please consider switching once all readers are up-to-date",
         )
 
--- a/python/arcticdb/version_store/library.py
+++ b/python/arcticdb/version_store/library.py
@@ -1039,7 +1039,7 @@ class Library:
                 lib.write(symbol, data) # The data will be successfully written
             Please refer to https://docs.arcticdb.io/latest/notebooks/arcticdb_demo_recursive_normalizers for more details
             of this feature.
-            Please check https://docs.arcticdb.io/latest/runtime_config/#versionstorerecursivenormalizermetastructurev2 for the plan of
+            Please check https://docs.arcticdb.io/latest/runtime_config/#versionstorerecursivenormalizermetastructure for the plan of
             introducing meta structure v2. V2 has removed the dependency on pickle for normalizing meta structure. Please consider switching to V2
             as V1 will be deprecated in future v7.0.0 release. However note that reading V2 meta structure requires ArcticDB version >= 6.7.0,
             or KeyError will be raised.
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -1612,11 +1612,11 @@ def clear_query_stats():
     query_stats.reset_stats()
 
 
-@pytest.fixture(params=[1, 0])
+@pytest.fixture(params=[2, 1])
 def all_recursive_metastructure_versions(request):
-    set_config_int("VersionStore.RecursiveNormalizerMetastructureV2", request.param)
+    set_config_int("VersionStore.RecursiveNormalizerMetastructure", request.param)
     yield request.param
-    unset_config_int("VersionStore.RecursiveNormalizerMetastructureV2")
+    unset_config_int("VersionStore.RecursiveNormalizerMetastructure")
 
 
 # region Pytest special xfail handling
--- a/python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
+++ b/python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
@@ -1052,14 +1052,10 @@ expected = {
 }
 equals(expected, data)
 """
-                if all_recursive_metastructure_versions == 0  # V1
+                if all_recursive_metastructure_versions == 1  # V1
                 else """
-key_error_raised = False
-try:
-    data = lib.read('sym').data
-except KeyError:
-    key_error_raised = True
-assert key_error_raised
+with pytest.raises(KeyError):
+    lib.read("sym")
                 """
             )
             compat.old_lib.execute(
@@ -1068,6 +1064,7 @@ assert key_error_raised
 from arcticdb.version_store._custom_normalizers import CustomNormalizer, register_normalizer
 from arcticc.pb2.descriptors_pb2 import NormalizationMetadata
 import numpy as np
+import pytest
 
 {self._COMPAT_TEST_CLASS_DEFINITIONS}
 
--- a/docs/mkdocs/docs/notebooks/ArcticDB_demo_recursive_normalizers.ipynb
+++ b/docs/mkdocs/docs/notebooks/ArcticDB_demo_recursive_normalizers.ipynb
@@ -37,7 +37,9 @@
     "### Meta structure V2\n",
     "\n",
     "Meta structure V2 for recursive normalizer is introduced in v6.7.0. It has removed the dependency on pickle for normalizing the meta structure.\n",
-    "Please consider enabling V2 as V1 will be deprecated in future v7.0.0 release. To enabld it, please set environment variable **ARCTICDB_VersionStore_RecursiveNormalizerMetastructureV2_int** to **1**"
+    "Please consider enabling V2 as V1 will be deprecated in future v7.0.0 release. To enabld it, you need to set environment variable **ARCTICDB_VersionStore_RecursiveNormalizerMetastructure_int** to **2**.\n",
+    "\n",
+    "Reader of the data is needed to be >= v6.7.0 or **KeyError** will be raised while reading V2 meta structure"
    ]
   },
   {
--- a/python/tests/compat/requirements-1.6.2.txt
+++ b/python/tests/compat/requirements-1.6.2.txt
@@ -7,3 +7,4 @@ msgpack >=0.5.0
 pyyaml
 packaging
 pyarrow
+pytest
--- a/python/tests/compat/requirements-1.6.2.yml
+++ b/python/tests/compat/requirements-1.6.2.yml
@@ -11,3 +11,4 @@ dependencies:
   - pyyaml
   - packaging
   - pyarrow
+  - pytest
--- a/python/tests/compat/requirements-4.5.1.txt
+++ b/python/tests/compat/requirements-4.5.1.txt
@@ -2,3 +2,4 @@ arcticdb==4.5.1
 numpy<2
 pyarrow
 protobuf <6
+pytest
--- a/python/tests/compat/requirements-4.5.1.yml
+++ b/python/tests/compat/requirements-4.5.1.yml
@@ -6,3 +6,4 @@ dependencies:
   - numpy<2
   - pyarrow
   - protobuf<=5.27.5
+  - pytest
--- a/python/tests/compat/requirements-5.0.0.txt
+++ b/python/tests/compat/requirements-5.0.0.txt
@@ -2,3 +2,4 @@ arcticdb==5.0.0
 numpy<2
 pyarrow
 protobuf<6
+pytest
--- a/python/tests/compat/requirements-5.0.0.yml
+++ b/python/tests/compat/requirements-5.0.0.yml
@@ -6,3 +6,4 @@ dependencies:
   - numpy<2
   - pyarrow
   - protobuf<6
+  - pytest
--- a/python/tests/compat/requirements-5.1.2.txt
+++ b/python/tests/compat/requirements-5.1.2.txt
@@ -2,3 +2,4 @@ arcticdb==5.1.2
 numpy<2
 pyarrow
 protobuf<6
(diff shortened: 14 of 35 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -1719,8 +1719,17 @@ def normalize_metadata(metadata: Any) -> UserDefinedMetadata:
...
 
 def normalize_recursive_metastruct(metastruct: Dict[Any, Any]) -> UserDefinedMetadata:
+    from arcticdb.flattener import Flattener
+
+    meta_structure_v2 = Flattener().meta_structure_v2
+    if meta_structure_v2 is not True:
+        log.warn(
+            "Recursive metastruct normalization in use (V1) is going to be deprecated after ArcticDB v7.0.0 release. "
+            "Please refer to https://docs.arcticdb.io/latest/runtime_config/#versionstorerecursivenormalizermetastructurev2 for more details."
+        )
```
vasil-pashov: Usually we add an environment variable to control such warnings and a ticket to remove the environment variable when some condition is met (in this case v7 is released)

python/arcticdb/version_store/library.py
```diff
@@ -1039,6 +1039,8 @@ def write(
                 lib.write(symbol, data) # The data will be successfully written
             Please refer to https://docs.arcticdb.io/latest/notebooks/arcticdb_demo_recursive_normalizers for more details
             of this feature.
+            Please check https://docs.arcticdb.io/latest/runtime_config/#versionstorerecursivenormalizermetastructurev2 for the plan of
```
vasil-pashov: I'd add a sentence describing the main benefit of V2 metastruct  here (and remove it when v7 is released)

python/arcticdb/flattener.py
```diff
@@ -85,26 +88,31 @@ def is_sequence_like(item):
     def is_dict_like(item):
         return isinstance(item, collections.abc.MutableMapping)
 
+    def _encode_metastruct_item_type_v2(self, obj_type):
+        if obj_type == None or obj_type == type(None):
```
vasil-pashov:
```suggestion
        if obj_type is None or obj_type == type(None):
```

python/tests/conftest.py
```diff
@@ -1614,6 +1621,13 @@ def clear_query_stats():
     query_stats.reset_stats()
 
 
+@pytest.fixture(params=[1, 0])
+def recursive_normalizer_meta_structure_v2(request):
```
vasil-pashov: nit: I think the name is confusing I initially thought that this enables only the V2 struct. I think `recursive_normalizer_meta_structure` is better.

python/arcticdb/flattener.py
```diff
@@ -152,26 +160,42 @@ def _create_meta_structure(self, obj, sym, to_write, depth=0, original_symbol=No
 
         # Commit 450170d94 shows a non-recursive implementation of this function, but since `msgpack.packb` of the
         # result is itself recursive, there is little point to rewriting this function.
-        item_type, iterables, normalization_info = self.derive_iterables(obj)
+        item_type, iterables, normalizer = self.derive_iterables(obj)  # if obj
```
vasil-pashov: What do you mean by `# if obj`

python/arcticdb/flattener.py
```diff
@@ -184,23 +208,23 @@ def _create_meta_structure(self, obj, sym, to_write, depth=0, original_symbol=No
...
+                key_till_now = "{}{}{}".format(sym, self.SEPARATOR, str_k)
+                meta_struct["sub_keys"].append(
+                    self._create_meta_structure(
+                        v, key_till_now, to_write, depth=depth + 1, original_symbol=original_symbol
+                    )
                 )
-            key_till_now = "{}{}{}".format(sym, self.SEPARATOR, str_k)
-            meta_struct["sub_keys"].append(
-                self._create_meta_structure(v, key_till_now, to_write, depth=depth + 1, original_symbol=original_symbol)
-            )
```
vasil-pashov, line 203: Is it correct the changes here are only indentation due to the removing the return from ` if serialized_as_primitive` and making it if/else

python/arcticdb/flattener.py
```diff
@@ -241,14 +273,55 @@ def _create_original_obj_from_metastruct_new_v1(self, meta_struct, key_map):
...
-        if normalizer_used:
-            return normalizer_used.denormalize(base_struct, meta)
+        normalizer_class = self._get_class_type_from_flattened(normalizer_format, normalizer_module, normalizer_name)
+        if normalizer_class is not None:
+            normalizer = normalizer_class()
+            return normalizer.denormalize(base_struct, None)
 
         return base_struct
 
     def create_original_obj_from_metastruct_new(self, meta_struct, key_map):
```
vasil-pashov: nit: I'd drop the `_new` suffix

python/tests/conftest.py
```diff
@@ -1098,6 +1098,13 @@ def lmdb_version_store_v2(version_store_factory, lib_name) -> NativeVersionStore
     return version_store_factory(dynamic_strings=True, encoding_version=int(EncodingVersion.V2), name=library_name)
 
 
+@pytest.fixture(scope="function")
+def lmdb_version_store_custome_array_norm(version_store_factory):
```
vasil-pashov: typo
```suggestion
def lmdb_version_store_custom_array_norm(version_store_factory):
```

vasil-pashov: My preference is to avoid having this much specialized fixtures. It's a bit easier to read the test code if you can see explicitly that a normalizer is registered. I'll leave it to you and the other reviewer to take the final decision.

docs/mkdocs/docs/runtime_config.md
```diff
@@ -104,6 +104,26 @@ Values:
...
+### VersionStore.RecursiveNormalizerMetastructureV2
+
+Controls whether the recursive normalizer will use meta structure V2
+
+**Read Compatibility:**
+
+| Meta Structure Version | Read Support |
+|------------------------|--------------|
+| V1 | All existing and future ArcticDB releases |
+| V2 | ArcticDB v6.6.0 and later |
```
alexowens90: Looks like this will be at least 6.7.0 now

docs/mkdocs/docs/runtime_config.md
```diff
@@ -104,6 +104,26 @@ Values:
...
+
+**V1 meta structure phase-out plan:**
+
+| Version | Change |
+|---------|--------|
+| >= v6.6.0 | Deprecation warning when writing V1 meta structure; V2 meta structure can be enabled optionally |
+| >= v7.0.0 | V2 meta structure will be enabled by default |
```
alexowens90, line 123: We might need to give Man longer to adapt if we do 7.0.0 soon for some other reason

python/arcticdb/version_store/library.py
alexowens90: I think we had discussed the default when using the `Library` API would be V2? That way the only supported version as far as external customers are concerned is V2.
It becomes messier with Man group users of the `Library` API though. Would need to have a Man-specific toggle to allow writing the V1 metastruct in case they have any clients reading the data with older versions.
I think on balance the current approach is the best. We'll need to trail a lot of communication internally when the default is going to change.

python/arcticdb/version_store/_normalization.py
alexowens90: Should also check that it only logs once per call to `write`

python/arcticdb/version_store/_normalization.py
```diff
@@ -1719,8 +1719,17 @@ def normalize_metadata(metadata: Any) -> UserDefinedMetadata:
 
 
 def normalize_recursive_metastruct(metastruct: Dict[Any, Any]) -> UserDefinedMetadata:
+    from arcticdb.flattener import Flattener
```
alexowens90: nit: Why isn't this with the other imports?

python/arcticdb/version_store/_normalization.py
```diff
@@ -1719,8 +1719,17 @@ def normalize_metadata(metadata: Any) -> UserDefinedMetadata:
...
+
+    meta_structure_v2 = Flattener().meta_structure_v2
+    if meta_structure_v2 is not True:
+        log.warn(
+            "Recursive metastruct normalization in use (V1) is going to be deprecated after ArcticDB v7.0.0 release. "
+            "Please refer to https://docs.arcticdb.io/latest/runtime_config/#versionstorerecursivenormalizermetastructurev2 for more details."
```
alexowens90: This log should also specify the minimum version of arcticdb needed to read data written with the V2 metastruct

python/arcticdb/flattener.py
```diff
@@ -85,26 +88,31 @@ def is_sequence_like(item):
     def is_dict_like(item):
         return isinstance(item, collections.abc.MutableMapping)
 
+    def _encode_metastruct_item_type_v2(self, obj_type):
+        if obj_type == None or obj_type == type(None):
+            return None, None, None
+        else:
+            return "class", obj_type.__module__, obj_type.__name__
```
alexowens90: I don't understand the point of this hard-coded "class" string?

python/arcticdb/flattener.py
```diff
@@ -213,23 +237,31 @@ def create_meta_structure(self, obj, sym):
...
-        if meta_struct["leaf"]:
-            if meta_struct["data"]:
-                return self.deserialize_primitives(meta_struct["data"])
-            else:
-                return key_map[self.compact_v1(meta_struct["symbol"])]
+    def _get_class_type_from_flattened(self, type_format, type_module, type_name):
+        if type_format is None:
+            return None
+        if type_format != "class":
+            raise RuntimeError(f"Unexpected serialized key type format {type_format} in meta structure")
```
alexowens90: nit: Use `check`

python/arcticdb/flattener.py
```diff
@@ -241,14 +273,55 @@ def _create_original_obj_from_metastruct_new_v1(self, meta_struct, key_map):
...
     def create_original_obj_from_metastruct_new(self, meta_struct, key_map):
-        if meta_struct["__version__"] and meta_struct["__version__"] == 1:
-            return self._create_original_obj_from_metastruct_new_v1(meta_struct, key_map)
+        meta_struct_ver = meta_struct["__version__"]
+        if meta_struct_ver == 1:
+            return self._create_original_obj_from_metastruct_v1(meta_struct, key_map)
+        elif meta_struct_ver == 2:
+            return self._create_original_obj_from_metastruct_v2(meta_struct, key_map)
         else:
             log.error("Could not identify version of nested normalizer used, Got:", meta_struct)
```
alexowens90: This looks like an old client trying to read V2 metastruct data would just log an error and return `None` instead of raising?

python/arcticdb/flattener.py
```diff
@@ -85,26 +88,31 @@ def is_sequence_like(item):
     def is_dict_like(item):
         return isinstance(item, collections.abc.MutableMapping)
 
+    def _encode_metastruct_item_type_v2(self, obj_type):
+        if obj_type == None or obj_type == type(None):
+            return None, None, None
```
alexowens90: AFAICT this special case isn't needed?
```
>>> type(None).__module__
'builtins'
>>> type(None).__name__
'NoneType'
```

python/tests/conftest.py
alexowens90: It will also be confusing in test outputs, as 1 maps onto V2 and 0 maps onto V1. I would call it something like `all_recursive_metastructure_versions`, have `params=[1, 2]`, and call `set_config_int` with `request.param - 1`

python/tests/conftest.py
alexowens90: Yeh, it's only used once so agree this is over-specialised

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
```diff
@@ -970,8 +1011,113 @@ def test_write_new_read_old(self, old_venv_and_arctic_uri, lib_name):
                 dfs=dfs,
             )
 
+    @pytest.mark.skipif(MACOS_WHEEL_BUILD, reason="We don't have previous versions of arcticdb pypi released for MacOS")
+    def test_write_new_read_old_custom_normalizer(self, old_venv_and_arctic_uri, lib_name, root_custom_array):
```
alexowens90: This should be parametrized by `recursive_normalizer_meta_structure_v2` so we have a test for what happens if an old client tries to read a metastruct it doesn't understand

docs/mkdocs/docs/runtime_config.md
phoebusm: I think we can update the plan here anytime before flipping the switch

python/arcticdb/version_store/_normalization.py
phoebusm: Circular imports

python/arcticdb/flattener.py
phoebusm: Yes that's correct

python/arcticdb/flattener.py
phoebusm: Leftover comment. Will remove

python/arcticdb/version_store/_normalization.py
phoebusm: > Should also check that it only logs once per call to write

Only one metastructure normalization is done per write. And `batch_write` doesn't support recursive normalization atm.

python/tests/conftest.py
phoebusm: It's neater to make it a fixture as `register_normalizer` is needed to be run before `NativeVersionStore` object is created. I have put it back to the test file since it's only been used there

python/arcticdb/flattener.py
phoebusm: It's a naming for identifying whether the node being normalized is `None` or an object. 
As `getattr(sys.modules["builtins"], "NoneType")` doesn't work, I prefer having identification for the object type, rather than matching `type_module == "builtins" and type_name == "NoneType"`
Anyway I don't have strong opinion towards it

python/arcticdb/flattener.py
phoebusm: I have updated the code to remove this special case

python/arcticdb/version_store/_normalization.py
```diff
@@ -1718,8 +1719,18 @@ def normalize_metadata(metadata: Any) -> UserDefinedMetadata:
 
 
 def normalize_recursive_metastruct(metastruct: Dict[Any, Any]) -> UserDefinedMetadata:
+    from arcticdb.flattener import Flattener
+
+    meta_structure_v2 = Flattener().meta_structure_v2
+    if meta_structure_v2 is not True and Flattener().meta_structure_v1_deprecation_warning:
```
alexowens90: Constructing `Flattener` objects just to retrieve properties of them is a bit of an antipattern. e.g. if the `Flattener.__init__` method became non-trivial, you wouldn't expect it to impact here. I would just call `get_config_int` here directly (which also removes the circular import problem)

python/tests/conftest.py
```diff
@@ -1612,6 +1614,13 @@ def clear_query_stats():
     query_stats.reset_stats()
 
 
+@pytest.fixture(params=[1, 0])
```
alexowens90: https://github.com/man-group/ArcticDB/pull/2778#discussion_r2619575267

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
```diff
@@ -915,9 +924,50 @@ def test_data_layout(lmdb_version_store_v1):
...
+            return None
+        return [item._x, item._y, item._z], NormalizationMetadata.CustomNormalizerMeta()
+
+    def denormalize(self, item, norm_meta):
+        return CustomArray(*item)
+    """
+
     @pytest.mark.skipif(MACOS_WHEEL_BUILD, reason="We don't have previous versions of arcticdb pypi released for MacOS")
-    def test_compat_write_old_read_new(self, old_venv_and_arctic_uri, lib_name):
+    def test_compat_write_old_read_new(self, old_venv_and_arctic_uri, lib_name, all_recursive_metastructure_versions):
```
alexowens90, line 970: Old clients don't understand the env var, so this parametrization doesn't do anything?

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
```diff
@@ -970,8 +1020,129 @@ def test_write_new_read_old(self, old_venv_and_arctic_uri, lib_name):
...
+equals(expected, data)
+"""
+                if all_recursive_metastructure_versions == 0  # V1
+                else """
+key_error_raised = False
+try:
+    data = lib.read('sym').data
+except KeyError:
+    key_error_raised = True
+assert key_error_raised
```
alexowens90:
```suggestion
with pytest.raises(KeyError):
    lib.read("sym")
```

python/arcticdb/flattener.py
```diff
@@ -32,6 +35,10 @@ class Flattener:
 
     def __init__(self):
         self.custom_normalizer = get_custom_normalizer(False)
+        self.meta_structure_v2 = get_config_int("VersionStore.RecursiveNormalizerMetastructureV2") == 1
```
alexowens90: I think it would be less confusing like this:
```suggestion
        self.meta_structure_v2 = get_config_int("VersionStore.RecursiveNormalizerMetastructure") == 2
```
Otherwise if we add a V3, how will it work?

alexowens90 (changes requested):

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
phoebusm: Just to make sure flipping the env var won't limit only V2 metastructure can be read

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
phoebusm: It's a seperate script but wow I didn't know pytest could work outside the framework. Amended

python/arcticdb/flattener.py
phoebusm: Sorry have missed earlier commetns. Amended

python/tests/conftest.py
phoebusm: Sorry have missed earlier commetns. Amended

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
alexowens90: Ah OK

python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py
alexowens90: I also did not know this...

vasil-pashov (approved):

PR #2789, opened 2025-12-02
Enhancement 10681189884: Upgrade sparrow to 1.4.0

alexowens90:
#### Reference Issues/PRs
[10681189884](https://man312219.monday.com/boards/7852509418/pulses/10681189884)

Updates sparrow to 1.4.0. This requires a vcpkg baseline update, which does not work out of the box on linux. Binary search of vcpkg commits identifies [this](https://github.com/microsoft/vcpkg/pull/44653) as the culprit. We do not seem to be the only project affected ([ticket](https://github.com/microsoft/vcpkg/issues/48496)).
This only breaks the build of the arrow dependency, which is only used for testing decimals, which are not yet supported, so for now I've just removed the dependency and corresponding tests.

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -936,7 +936,7 @@ endif()
 if(${TEST})
     unset(Python_USE_STATIC_LIBS)
     find_package(Python 3 COMPONENTS Interpreter Development REQUIRED)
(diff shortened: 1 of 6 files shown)


IvoDD (approved):
LGTM, can you add a low priority ticket to readd the arrow tests and link to this PR

PR #2792, opened 2025-12-03
[18343808662] MemBlock refactor

IvoDD:
#### Reference Issues/PRs
Monday ref: 18343808662

This is only a refactor to prepare for parallel bool unpacking and should not change existing behavior.

It splits up the `MemBlock` into two separate classes inheriting a common interface `IMemBlock`:
- `DynamicMemBlock` is responsible for the `AllocationType::DYNAMIC` and has a capacity which can gradually be filled up with resizes. This is the type of memory blocks we use for in memory operations with `ChunkedBuffer`s e.g. when tick streaming or when performing appends etc.
- `ExternalMemBlock` is responsible for the `AllocationType::DETACHABLE` and for external pointers to views. It just holds a pointer to external memory which can be owning or not. This is what we use when either when returning Segments to python or when receiving views from python

This change also adds the scaffolding needed for the packed bits buffer via `ExternalPackedMemBlock`. It will be used in a follow up PR to parallelise bool unpacking.

#### Any other comments?
Also introduced a few micro benchmarks from which the results are:

| Benchmark                                             | Time before | Time after |
| ----------------------------------------------------- | ----------- | ---------- |
| BM_chunked_buffer_random_access/100000/203/1/0        | 16.7 ns     | 15.5 ns    |
| BM_chunked_buffer_random_access/100000/203/0/0        | 40.5 ns     | 34.9 ns    |
| BM_chunked_buffer_random_access/100000/203/0/2        | 112 ns      | 117 ns     |
| BM_chunked_buffer_allocate_with_ensure/100000/203/1/0 | 7.16 ms     | 7.84 ms    |
| BM_chunked_buffer_allocate_with_ensure/100000/203/0/0 | 7.61 ms     | 8.13 ms    |
| BM_chunked_buffer_allocate_with_ensure/100000/203/0/2 | 24.0 ms     | 17.9 ms    |

The benchmarks are run with:
```
./benchmarks --benchmark_filter=BM_chunked_buffer_allocate.* --benchmark_time_unit=ms --benchmark_min_time=3s --benchmark_repetitions=5
./benchmarks --benchmark_filter=BM_chunked_buffer_random_access.* --benchmark_time_unit=ns --benchmark_min_time=1s --benchmark_repetitions=5
```

I've not included the larger chunk size because that is identical for before and after. The only statistically significant difference is the allocation on a `DETACHABLE` segment improvement. That is because for `ExternalMemBlock`s we no longer need to allocate the intrinsic data that was previously always included in the `MemBlock`.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/column_store/block.cpp
+++ b/cpp/arcticdb/column_store/block.cpp
@@ -41,6 +41,14 @@ uint8_t* IMemBlock::ptr(size_t pos) { return data() + pos; }
 
 uint8_t* IMemBlock::end() const { return const_cast<uint8_t*>(ptr(physical_bytes())); }
 
+void IMemBlock::resize(size_t bytes) {
+    util::raise_rte("Can't resize a memory block which does not support `resize`. Bytes: {}", bytes);
+}
+
+uint8_t* IMemBlock::release() { util::raise_rte("Can't release a memory block which does not support `release`"); }
+
+void IMemBlock::abandon() { util::raise_rte("Can't abandon a memory block which does not support `abandon`"); }
+
 // DynamicMemBlock implementation
 DynamicMemBlock::DynamicMemBlock(size_t capacity, size_t offset, entity::timestamp ts) :
     bytes_(0),
@@ -74,10 +82,6 @@ void DynamicMemBlock::resize(size_t size) {
     );
     bytes_ = size;
 }
-
-uint8_t* DynamicMemBlock::release() { util::raise_rte("Can't release a dynamic block"); }
-
-void DynamicMemBlock::abandon() { util::raise_rte("Can't abandon a dynamic block"); }
 
 // ExternalMemBlock implementation
 ExternalMemBlock::ExternalMemBlock(
@@ -129,8 +133,6 @@ entity::timestamp ExternalMemBlock::timestamp() const { return timestamp_; }
 const uint8_t* ExternalMemBlock::data() const { return external_data_; }
 
 uint8_t* ExternalMemBlock::data() { return external_data_; }
-
-void ExternalMemBlock::resize(size_t bytes) { util::raise_rte("Can't resize a non dynamic block. Bytes: {}", bytes); }
 
 // ExternalPackedMemBlock implementation
 ExternalPackedMemBlock::ExternalPackedMemBlock(
--- a/cpp/arcticdb/column_store/block.hpp
+++ b/cpp/arcticdb/column_store/block.hpp
@@ -33,13 +33,13 @@ class IMemBlock {
 
     virtual ~IMemBlock() = default;
 
-    // The below methods make sense only for specific memory block types. Decided to put them in the interface for
-    // performance reasons (i.e. calling them would be a single vtable lookup vs using `get_type` + `dynamic_cast`).
+    // The below methods make sense only for specific memory block types.
+    // Default implementations raise errors - override in subclasses that support these operations.
     // Dynamic block specific methods
-    virtual void resize(size_t bytes) = 0;
+    virtual void resize(size_t bytes);
     // External block specific methods
-    [[nodiscard]] virtual uint8_t* release() = 0;
-    virtual void abandon() = 0;
+    [[nodiscard]] virtual uint8_t* release();
+    virtual void abandon();
 
     // Implemented methods
     [[nodiscard]] bool empty() const;
@@ -79,8 +79,6 @@ class DynamicMemBlock : public IMemBlock {
     [[nodiscard]] const uint8_t* data() const final;
     [[nodiscard]] uint8_t* data() final;
     void resize(size_t size) final;
-    [[nodiscard]] uint8_t* release() final;
-    void abandon() final;
 
     size_t bytes_ = 0UL;
     size_t capacity_ = 0UL;
@@ -119,7 +117,6 @@ class ExternalMemBlock : public IMemBlock {
     [[nodiscard]] entity::timestamp timestamp() const override;
     [[nodiscard]] const uint8_t* data() const override;
     [[nodiscard]] uint8_t* data() override;
-    void resize(size_t bytes) override;
 
   private:
     size_t bytes_ = 0UL;


cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -70,7 +70,7 @@ class ChunkedBufferImpl {
         void next() {
             pos_ += type_size_;
 
-            if (pos_ >= block_->bytes()) {
+            if (pos_ >= block_->logical_bytes()) {
```
IvoDD: Highlighting a tiny behavior change. The `Column::Iterator` was previously broken for blocks with `extra_bytes>0`

cpp/arcticdb/column_store/chunked_buffer.cpp
```diff
@@ -120,7 +120,7 @@ ChunkedBufferImpl<BlockSize> truncate(const ChunkedBufferImpl<BlockSize>& input,
     for (auto idx = start_idx; idx < end_idx; idx++) {
         auto input_block = input_blocks.at(idx);
         auto source_pos = idx == start_idx ? start_block_and_offset.offset_ : 0u;
-        auto source_bytes = std::min(remaining_bytes, input_block->bytes() - source_pos);
+        auto source_bytes = std::min(remaining_bytes, input_block->logical_bytes() - source_pos);
```
alexowens90: I don't think this would be correct for an `ExternalMemBlock`? I think this will only be called at the moment with `DYNAMIC` mem blocks, so maybe use `physical_bytes()` here and add a `util::check` on entry to the function that the input mem blocks are dynamic?

cpp/arcticdb/column_store/block.cpp
```diff
@@ -0,0 +1,176 @@
...
+ExternalPackedMemBlock::ExternalPackedMemBlock(
+        uint8_t* data, size_t size, size_t shift, size_t offset, entity::timestamp ts, bool owning
+) :
+    ExternalMemBlock(data, (size + shift - 1) / 8 + 1, offset, ts, owning),
+    logical_size_(size),
+    shift_(shift) {}
+
+MemBlockType ExternalPackedMemBlock::get_type() const { return MemBlockType::EXTERNAL_PACKED; }
+
+size_t ExternalPackedMemBlock::logical_bytes() const { return logical_size_; }
```
alexowens90: This is a bit confusing, as it is actually the number of logical bits in this case

cpp/arcticdb/column_store/block.hpp
```diff
@@ -10,12 +10,51 @@
...
+    [[nodiscard]] virtual size_t offset() const = 0;
+    [[nodiscard]] virtual entity::timestamp timestamp() const = 0;
+    [[nodiscard]] virtual const uint8_t* data() const = 0;
+    [[nodiscard]] virtual uint8_t* data() = 0;
+    // Dynamic block specific methods
+    virtual void resize(size_t bytes) = 0;
+    virtual void check_magic() const = 0;
+    // External block specific methods
+    [[nodiscard]] virtual uint8_t* release() = 0;
+    virtual void abandon() = 0;
```
alexowens90: It's a bit of a code smell to have interface methods for specific subclasses. The alternative is to have callers of these methods check the return value of `get_type` and do a `reinterpret_cast` as appropriate though, which also feels clunky

cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -131,9 +131,9 @@ class ChunkedBufferImpl {
         output.extra_bytes_per_block_ = extra_bytes_per_block_;
 
         for (auto block : blocks_) {
-            output.add_block(block->capacity_, block->offset_);
-            (*output.blocks_.rbegin())->copy_from(block->data(), block->bytes(), 0);
-            (*output.blocks_.rbegin())->resize(block->bytes());
+            output.add_block(block->capacity(), block->offset());
+            (*output.blocks_.rbegin())->resize(block->physical_bytes());
```
alexowens90: I guess this means `clone()` is never called with anything other than dynamic memory at the moment?
We could make it work for external memory, although the semantics of an owning external buffer mean that the clone wouldn't own it, and so wouldn't be identical to the source. I think just raising if someone tries to use `clone()` with external memory is fine for now, and if we need it in the future we can add a `view()` method or similar.

cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -131,9 +131,9 @@ class ChunkedBufferImpl {
...
+            (*output.blocks_.rbegin())->copy_from(block->data(), block->physical_bytes(), 0);
```
alexowens90:
```suggestion
            output.blocks_.back().copy_from(block->data(), block->physical_bytes(), 0);
```

cpp/arcticdb/column_store/chunked_buffer.cpp
IvoDD: yeah it would drop any `extra_bytes` if there are any and yeah truncation to a presized buffer doesn't really make sense if we have extra bytes. Will add an assertion that `type==DYNAMIC || type==EXTERNAL and extra_bytes==0`.

cpp/arcticdb/column_store/block.cpp
IvoDD: agreed, will rename to `logical_size`

cpp/arcticdb/column_store/block.hpp
IvoDD: yeah only 90% of the interface is common and the choices either to have them as methods on the interface or to do `dynamic_cast` after checking `get_type`. The other option would be to not use inheritance but rather `std::variant` which would be clunkier.

Main reason I chose having them in the interface was that presumably it will be faster to just read the vtable once instead of doing `get_type` which also reads vtable and `dynamic_cast` which needs to read the RTTI. 
Compiler might optimize away some of this though.

Will add a comment explaining why I think using this instead of `dynamic_cast` is presumably faster. Happy to be overruled though.

Note `reinterpret_cast` would probably be [undefined behavior](https://stackoverflow.com/questions/78593829/can-reinterpret-cast-be-considered-as-an-unsafe-workaround-when-dynamic-cast-is)

cpp/arcticdb/column_store/chunked_buffer.hpp
IvoDD: It would technically work (all blocks will be copied correctly) but yeah the clone would only have dynamic blocks.

This is actually used with non dynamic blocks in a couple of places:
In library tool [here](https://github.com/man-group/ArcticDB/blob/2a836a56615c056573b1a1b37a18a109fa236a26/cpp/arcticdb/toolbox/library_tool.cpp#L49)
which for some reason receives external memory blocks if using v2 encoding.

I've added an assertion that the block is either DYNAMIC or EXTERNAL with no extra bytes.

cpp/arcticdb/arrow/test/test_arrow_write.cpp
```diff
@@ -173,11 +173,12 @@ TYPED_TEST(ArrowDataToSegmentNumeric, MultipleRecordBatches) {
         ASSERT_EQ(buffer.block_offsets().size(), rows_per_batch.size() + 1);
         size_t bytes{0};
         for (size_t idx = 0; idx < rows_per_batch.size(); ++idx) {
-            ASSERT_TRUE(buffer.blocks()[idx]->is_external());
-            ASSERT_EQ(buffer.blocks()[idx]->bytes(), rows_per_batch[idx] * sizeof(TypeParam));
-            ASSERT_EQ(buffer.blocks()[idx]->offset_, bytes);
+            ASSERT_EQ(buffer.blocks()[idx]->get_type(), MemBlockType::EXTERNAL_WITH_EXTRA_BYTES);
+            ASSERT_EQ(buffer.blocks()[idx]->physical_bytes(), rows_per_batch[idx] * sizeof(TypeParam));
+            ASSERT_EQ(buffer.blocks()[idx]->logical_size(), rows_per_batch[idx] * sizeof(TypeParam));
```
vasil-pashov, line 178: nit: These two names look odd side by side. Does it make sense to either rename physical_bytes to `physical_size` or `logical_size` to `logical_bytes`

cpp/arcticdb/column_store/block.hpp
vasil-pashov: My two cents:
1. Both are ugly
2. We have something that is similar to the current proposal e.g. not all clauses support all clause methods and they throw exceptions. Thus I think this is ok
3. Reinterpret cast won't be UB https://en.cppreference.com/w/cpp/language/reinterpret_cast.html 5) handles this case.
4. Checking the vtable is not that slow especially if it's not in a hot loop
5. Adding `final` to the derived classes would help the compiler optimize vtables if possible

cpp/arcticdb/column_store/block.hpp
```diff
@@ -10,12 +10,54 @@
...
+    // Implemented methods
+    [[nodiscard]] bool empty() const;
+    [[nodiscard]] size_t free_space() const;
+    void add_bytes(size_t bytes);
+    void copy_to(uint8_t* target) const;
+    void copy_from(const uint8_t* src, size_t bytes, size_t pos);
+    [[nodiscard]] virtual uint8_t& operator[](size_t pos);
+    [[nodiscard]] const uint8_t* ptr(size_t pos) const;
+    [[nodiscard]] uint8_t* ptr(size_t pos);
+    [[nodiscard]] uint8_t* end() const;
```
vasil-pashov: `ExternalPackedMemBlock`'s override of operator[] throws. On the other hand you can call ptr() on it as it's part of the base class. But they are mostly doing the same thing. I'd remove the `virtual operator[]`.

cpp/arcticdb/column_store/block.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+ExternalMemBlock::ExternalMemBlock(
+        uint8_t* data, size_t size, size_t offset, entity::timestamp ts, bool owning, size_t extra_bytes
+) :
+    bytes_(size),
+    offset_(offset),
+    timestamp_(ts),
+    external_data_(data),
+    owns_external_data_(owning),
+    extra_bytes_(extra_bytes) {}
```
vasil-pashov: We can remove this. The other constructor's doing the same thing.

cpp/arcticdb/column_store/block.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+uint8_t* ExternalMemBlock::release() {
+    auto* tmp = external_data_;
+    external_data_ = nullptr;
+    owns_external_data_ = false;
+    return tmp;
+}
+
+void ExternalMemBlock::abandon() {
+    free_detachable_memory(external_data_, physical_bytes());
```
vasil-pashov: Shouldn't we check if this is owning the external data?

cpp/arcticdb/column_store/block.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+entity::timestamp ExternalMemBlock::timestamp() const { return timestamp_; }
+
+const uint8_t* ExternalMemBlock::data() const { return external_data_; }
+
+uint8_t* ExternalMemBlock::data() { return external_data_; }
+
+void ExternalMemBlock::resize(size_t bytes) { util::raise_rte("Can't resize a non dynamic block. Bytes: {}", bytes); }
+
+void ExternalMemBlock::check_magic() const {}
```
vasil-pashov: I know it's not releated to your changes but I think we should either get rid of the magic or start using it. AFAIK it was added for the V2 encoding which used native structures instead of protobuf. The magic was placed at the end of the buffer containing some object. This way you could
1. Inspect the raw byte buffer and get a sense of what's in it
2. Verify that something was loaded correctly

I don't think the mem blocks need it

cpp/arcticdb/column_store/block.hpp
```diff
@@ -24,122 +66,94 @@ struct MemBlock {
...
+    ExternalPackedMemBlock(uint8_t* data, size_t size, size_t shift, size_t offset, entity::timestamp ts, bool owning);
+
+    MemBlockType get_type() const override;
+    [[nodiscard]] size_t logical_size() const override;
+    [[nodiscard]] uint8_t& operator[](size_t pos) override;
+    [[nodiscard]] size_t shift() const;
+
+  private:
+    size_t logical_size_ = 0UL;
+    size_t shift_ = 0UL;
```
vasil-pashov, line 151: Discussed offline, but a comment about what shift is will be helpful

cpp/arcticdb/column_store/chunked_buffer.hpp
```diff
@@ -38,10 +38,10 @@ namespace arcticdb {
 template<size_t DefaultBlockSize>
 class ChunkedBufferImpl {
 
-    using BlockType = MemBlock;
+    using BlockType = IMemBlock;
 
-    static_assert(sizeof(BlockType) == BlockType::Align + BlockType::MinSize);
-    static_assert(DefaultBlockSize >= BlockType::MinSize);
+    static_assert(sizeof(DynamicMemBlock) == DynamicMemBlock::Align + DynamicMemBlock::MinSize);
```
vasil-pashov: It makes more sense for this to be near the DynamicMemBlock implementation rather than here

cpp/arcticdb/arrow/test/test_arrow_write.cpp
IvoDD: I have explicitly named `logical_size` to highlight that it does not refer to number of bytes. For the `ExternalPackedBuffer` buffer it refers to number of bits. 

On the other hand `physical_bytes` does refer to the actual size of memory occupied in bytes. Thus I think the distinction is useful.

cpp/arcticdb/column_store/block.hpp
IvoDD: Ah thanks for clarifying. Will add `final` to all methods that it makes sense for.

cpp/arcticdb/column_store/block.hpp
IvoDD: Ah thanks, the idea was `ptr` to use the `operator[]` but I missed that. Will remove `operator[]` as it's not used anywhere meaningfully.

cpp/arcticdb/column_store/block.cpp
IvoDD: yes, good catch. Previously we were only doing an assert [here](https://github.com/man-group/ArcticDB/pull/2792/files#diff-6a95db6224aa520eafb327c355ac4b47be0d36e1a3baf8b2200448d1cc88a0b0L114), but now we use it in `abandon_block`, so using an if.

cpp/arcticdb/column_store/block.cpp
IvoDD: agreed, removing entirely

cpp/arcticdb/column_store/chunked_buffer.cpp
IvoDD: The `type==EXTERNAL and extra_bytes==0` is needed for arrow writes.

cpp/arcticdb/column_store/block.hpp
alexowens90: `folly::Poly` doesn't have a concept of a base class implementation, which is why we have to explicitly throw from the derived classes for clauses. We could make the base class methods non-virtual and throw an exception, and only override in derived classes where the methods make sense.

vasil-pashov (approved):

alexowens90 (approved):
Subject to valgrind runs and list_versions slowdown

PR #2795, opened 2025-12-04
[18053251438] Parallelise bool unpacking for arrow write

IvoDD:
#### Reference Issues/PRs
Monday ref: 18053251438

After the MemBlock refactor we can now store packed bits inside the ChunkedBuffer in `arrow_data_to_segment`.

We later unpack the bools inside `WriteToSegmentTask` which runs in parallel for all segment slices.

#### Any other comments?
Timings to write a 10million rows x 10 bool columns

| Number of threads | Before | After |
|---------------|-------|---------------|
| 1 | 214.94 ms | 213.27 ms |
| 2 | 214.14 ms | 208.27 ms |
| 4 | 215.44 ms | 199.38 ms |
| 8 | 214.11 ms | 198.96 ms |

Note that there was still big variance of +- 20ms on my machine but the above is averaged over 200 runs

Dummy benchmar script because of unreliable asv benchmarks:
```
ac = adb.Arctic("lmdb:///tmp/test")
ac.delete_library("benchmark")
lib = ac.create_library("benchmark", output_format="PYARROW")
lib._nvs._set_allow_arrow_input()

num_rows = 10_000_000
num_cols = 10
table = pa.table({f"bool_{i}": pa.array([j % (i + 1) == 0 for j in range(num_rows)]) for i in range(num_cols)})

for num_threads in [1, 2, 4, 8]:
    with adb.util.test.config_context("VersionStore.NumCPUThreads", num_threads):
        adb_async.reinit_task_scheduler()
        time = timeit.timeit(lambda: lib.write("sym", table), number=50) / 50
        print(f"Time for {num_threads} is {time*1000:.2f} ms")
```

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/pipeline/write_frame.cpp
+++ b/cpp/arcticdb/pipeline/write_frame.cpp
@@ -84,8 +84,8 @@ Column WriteToSegmentTask::slice_column(
 ) const {
     const auto& source_column = frame.column(col_idx);
     const auto type_size = get_type_size(source_column.type().data_type());
-    const auto first_byte = (slice_.rows().first - offset) * type_size;
-    const auto bytes = ((slice_.rows().second - offset) * type_size) - first_byte;
+    const auto first_pos = (slice_.rows().first - offset) * type_size;
+    const auto bytes = ((slice_.rows().second - offset) * type_size) - first_pos;
 
     if (is_bool_type(source_column.type().data_type())) {
         // Bool columns from arrow come as packed bitsets and are stored in `MemBlockType::EXTERNAL_PACKED` memory.
@@ -101,17 +101,17 @@ Column WriteToSegmentTask::slice_column(
         auto dest_ptr = res.data().buffer().data();
 
         const auto& buffer = source_column.data().buffer();
-        auto [_, offset_in_block, block_index] = buffer.block_and_offset(first_byte);
+        auto [_, offset_in_block, block_index] = buffer.block_and_offset(first_pos);
         auto pos_in_res = 0u;
         while (pos_in_res < bytes) {
             const auto block = buffer.blocks()[block_index++];
             util::check(
                     block->get_type() == MemBlockType::EXTERNAL_PACKED,
                     "Expected to see a packed external block but got: {}",
-                    static_cast<int8_t>(block->get_type())
+                    block->get_type()
             );
-            const auto packed_block = dynamic_cast<ExternalPackedMemBlock*>(block);
-            auto num_bits = std::min(packed_block->logical_bytes() - offset_in_block, bytes - pos_in_res);
+            const auto packed_block = static_cast<ExternalPackedMemBlock*>(block);
+            auto num_bits = std::min(packed_block->logical_size() - offset_in_block, bytes - pos_in_res);
             auto offset_in_bits = packed_block->shift() + offset_in_block;
             packed_bits_to_buffer(packed_block->data(), num_bits, offset_in_bits, dest_ptr + pos_in_res);
             offset_in_block = 0;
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -86,6 +86,39 @@ Column WriteToSegmentTask::slice_column(
...
+                make_scalar_type(DataType::BOOL8),
+                slice_.rows().diff(),
+                AllocationType::PRESIZED,
+                Sparsity::NOT_PERMITTED
+        );
+        res.set_row_data(slice_.rows().diff() - 1);
+        auto dest_ptr = res.data().buffer().data();
+
+        const auto& buffer = source_column.data().buffer();
+        auto [_, offset_in_block, block_index] = buffer.block_and_offset(first_byte);
```
alexowens90: Is this correct? If (for example) we're slicing into 256 row chunks, and we're on the second row slice, then `first_byte == 256`. But we want to look at the 32nd byte of the input packed bitset?

cpp/arcticdb/column_store/block.cpp
vasil-pashov: Does size mean number of bits? I was left with the impression that all block constructors take in the number of bytes and this confused me.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -86,6 +86,39 @@ Column WriteToSegmentTask::slice_column(
...
+        auto pos_in_res = 0u;
+        while (pos_in_res < bytes) {
+            const auto block = buffer.blocks()[block_index++];
+            util::check(
+                    block->get_type() == MemBlockType::EXTERNAL_PACKED,
+                    "Expected to see a packed external block but got: {}",
+                    static_cast<int8_t>(block->get_type())
```
vasil-pashov: This won't mean much to people reading the error. Can you add fmt override for the enum values

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -86,6 +86,39 @@ Column WriteToSegmentTask::slice_column(
...
+            );
+            const auto packed_block = dynamic_cast<ExternalPackedMemBlock*>(block);
```
vasil-pashov: We've already established that it's ExternalPackedMemBlock, static_casting will be safe if performance is crucial. If you insist on using dynamic_cast you can then then check if the result is !nullptr to avoid the virtual call to `block->get_type()` on the happy path. If it's not you'd still want to call `block->get_type()` but then we're throwing so the virtual call will be dwarfed by the cost of the exception

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -86,6 +86,39 @@ Column WriteToSegmentTask::slice_column(
...
+            auto num_bits = std::min(packed_block->logical_bytes() - offset_in_block, bytes - pos_in_res);
```
vasil-pashov: shouldn't this be multiplied by 8 to be number of bits?

cpp/arcticdb/pipeline/write_frame.cpp
IvoDD: Sorry this was confusing. This is not the `first_byte` but rather the logical `first_pos`. Renamed it. Also `block_and_offset` is `logical_size` aware so this works fine.

cpp/arcticdb/column_store/block.cpp
IvoDD: On ExternalMemBlocks the size is the `logical_size`. Renamed in previous PR to reflect that.

cpp/arcticdb/pipeline/write_frame.cpp
IvoDD: Renamed in previous PR to `logical_size` to reflect that this is indeed the number of bits.

cpp/arcticdb/pipeline/write_frame.cpp
IvoDD: Switched to static cast

PR #2801, opened 2025-12-10
conda-build: Use parallel jobs for test suites in workflows

jjerphan:

Split the CI jobs for conda builds into several ones to shorten iterations by around three times.

The CI took around 1h30 before to run and now takes around 30 minutes.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/build_with_conda.yml
+++ b/.github/workflows/build_with_conda.yml
@@ -940,3 +940,14 @@ jobs:
           COMMANDLINE: ${{ inputs.run_commandline }}
           ARCTICDB_PYTEST_ARGS: ${{ inputs.run_custom_pytest_command }}
           NODE_OPTIONS: --openssl-legacy-provider
+
+  can_merge_conda:
+    name: All conda jobs completed
+    needs: [cpp_tests_linux, cpp_tests_macos, cpp_tests_windows, python_tests_linux, python_tests_macos, python_tests_windows]
+    if: |
+      always() &&
+      !failure() &&
+      !cancelled()
+    runs-on: ubuntu-22.04
+    steps:
+      - run: echo All conda jobs completed!


.github/workflows/build_with_conda.yml
```diff
@@ -63,29 +64,128 @@ on:
...
+      - name: Download build artifacts from previous runs
+        uses: actions/download-artifact@v4
+        continue-on-error: true
+        if: steps.find-run.outputs.run_id != ''
+        with:
+          name: build-linux
+          path: .
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          repository: ${{ github.repository }}
+          run-id: ${{ steps.find-run.outputs.run_id }}
```
vasil-pashov: Can this download the wrong artifacts? For example if there are two PRs from different branches?

jjerphan: Yes, this might be possible.

I guess we can fetch the build artifacts from the same PR / branch for safety.

vasil-pashov: Is it then possible that it can pick up the a previous buggy version e.g.

commit 1
commit 2 (with a bug but compiles)
commit 3

Is there a chance this picks up commit 2 or 1

jjerphan: I mostly want to make sure the `sccache` is properly used, but this might be an overkill.

Must this:

```yaml
          token: ${{ secrets.GITHUB_TOKEN }}
```

also be specified like for this other worklow?

https://github.com/man-group/ArcticDB/blob/1492e1ee7625cc20bcd438bec55aaa12d5c59cb5/.github/workflows/build_steps.yml#L95-L99

Not all workflows use this `token` for the step for `sccache` configuration.

vasil-pashov: According to the [docs](https://github.com/Mozilla-Actions/sccache-action?tab=readme-ov-file#using-on-github-enterprise-server-ghes) `token` is needed only for actions running on GitHub Enterprise Server. We're using the free runners so this shouldn't be needed. You can print the stats of sccache to make sure it was copied/used

jjerphan: I used a simpler approach with 358c29f.

vasil-pashov (approved):

PR #2802, opened 2025-12-12
Enhancement 10753358032: Log string pool decode timings in release builds

alexowens90:
#### Reference Issues/PRs
[10753358032](https://man312219.monday.com/boards/7852509418/pulses/10753358032)

Log how long it took to decode the string pool if the codec log level is at least debug

### DIFF
--- a/cpp/arcticdb/codec/codec.cpp
+++ b/cpp/arcticdb/codec/codec.cpp
@@ -366,6 +366,8 @@ void decode_string_pool(
 ) {
     if (hdr.has_string_pool_field()) {
         ARCTICDB_TRACE(log::codec(), "Decoding string pool");
+        interval timer;
+        timer.start();
         util::check(data != end, "Reached end of input block with string pool fields to decode");
         std::optional<util::BitMagic> bv;
         data += decode_ndarray(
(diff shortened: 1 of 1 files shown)


cpp/arcticdb/codec/codec.cpp
```diff
@@ -376,8 +378,8 @@ void decode_string_pool(
                 bv,
                 hdr.encoding_version()
         );
-
-        ARCTICDB_TRACE(log::codec(), "Decoded string pool to position {}", data - begin);
+        timer.end();
+        log::codec().debug("Decoded string pool to position {} in {}s", data - begin, timer.get_results_total());
```
IvoDD, line 382: I'm not sure how valuable will the `data-begin` be for users. It would probably be more meaningful if we include the `stream_id` for batch read operations and maybe the key being written if available.

I think the `SegmentInMemory& res` might have the `stream_id` available although I'm not sure it's populated at this stage. As for the key it would probably require passing it around from quite far so I wouldn't bother.

alexowens90: I left `data-begin` in to not lose information compared to the previous log. Yeh the key info would be most useful, although it's quite far away from here so I'd rather not involve it.

Issue #2799, opened 2025-12-08
[Bug] SSL connect error (curlCode: 35) on macOS (Apple Silicon) with Python 3.13 and Cloudflare R2


wy-z:
### Describe the bug

First of all, thank you for the great work on open-sourcing ArcticDB. It's an excellent tool.

### Description

I am encountering an SSL connection error when trying to connect to Cloudflare R2 using ArcticDB on macOS. The same configuration works perfectly on Linux, suggesting the issue is specific to the macOS environment or the specific wheel build.

### Environment

  * **OS:** macOS 15.x (Apple Silicon / arm64)
  * **Python Version:** 3.13
  * **ArcticDB Version/Wheel:** `arcticdb-6.5.0+man0-cp313-cp313-macosx_15_0_arm64.whl`
  * **Storage Backend:** Cloudflare R2 (S3 compatible)

### Reproduction Steps

I installed the specific wheel mentioned above. Here is the code snippet used to reproduce the issue:

```python
import arcticdb

# Connecting to Cloudflare R2
ac = arcticdb.Arctic("s3s://xxx.r2.cloudflarestorage.com:xxx?use_virtual_addressing=false&region=auto&access=xxx&secret=xxx")

# Triggering the network call
ac.list_libraries()
```

### Logs / Traceback

Upon execution, I receive the following `curlCode: 35` (SSL connect error):

```text
20251208 23:20:46.932768 23898244 W arcticdb | Failed to iterate key type with key 'C' : curlCode: 35, SSL connect error
20251208 23:20:46.932890 23898244 E arcticdb | Network error: S3Error:99, HttpResponseCode:-1, : curlCode: 35, SSL connect error for object '_arctic_cfg/cref/' This could be due to a connectivity issue or exhausted file descriptors. Having more than one open Arctic instance will use multiple file descriptors, you should reuse Arctic instances. If you need many file descriptors, consider increasing `ulimit -n`.
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    ac.list_libraries()
    ~~~~~~~~~~~~~~~~~^^
  File "/Users/weiyang/Ones/trade/q-trade/.venv/lib/python3.13/site-packages/arcticdb/arctic.py", line 358, in list_libraries
    raw_libs = self._library_manager.list_libraries()
[...]

PR #2816, opened 2025-12-18
Dropping support for SecureTransport on Mac using only OpenSSL

maxim-morozov:
Following issue #2799, we are disabling SecureTransport from Curl on macOS. Apple has deprecated SecureTransport, which does not support [TLS 1.3 (only up to TLS 1.2)](https://developer.apple.com/documentation/security/secure-transport) and is no longer being actively developed. Additional context can be found here: [link](https://daniel.haxx.se/blog/2025/01/14/secure-transport-support-in-curl-is-on-its-way-out/). Currently on Conda, we are already defaulting to OpenSSL, but recently released PyPI builds use SecureTransport due to static linking

#### Reference Issues/PRs
10725858850 

Curl VCPKG overlay that disables the SecureTransport protocol.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -204,6 +204,17 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/proto/arcticc/pb2/proto/)
 # for IDE resolution otherwise it is lost
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
+# Add `third_party` include directories (only used for vcpkg builds) so that headers can
+# be included without a compile-time check on `ARCTICDB_USING_CONDA` in the source files.
+if(NOT ${ARCTICDB_USING_CONDA})
+    # `semimap.h` is at `third_party/semimap/semimap.h`, so we need `third_party` in the path
+    # to allow <semimap/semimap.h> to resolve correctly
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/third_party)
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/third_party/lmdbcxx)
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/third_party/recycle/src)
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/third_party/Remotery/lib)
+endif()
+
 add_subdirectory(arcticdb)
 
 # arcticdb_core_static is used by other projects so we expose the headers via `arcticdb/.../abc.hpp`.
--- a/cpp/arcticdb/entity/performance_tracing.hpp
+++ b/cpp/arcticdb/entity/performance_tracing.hpp
@@ -32,11 +32,7 @@
 
 #ifdef USE_REMOTERY
 
-#ifdef ARCTICDB_USING_CONDA
-#include <remotery/Remotery.h>
-#else
 #include <Remotery.h>
-#endif
 
 class RemoteryInstance {
   public:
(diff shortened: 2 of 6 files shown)


cpp/third_party/vcpkg_overlays/curl/vcpkg.json
poodlewars, line 1: Could this just be a symlink to `cpp/vcpkg/ports/curl/vcpkg.json` rather than us having to maintain a copy?

Also, I seem to have cUrl 8.17.0 in my working copy. I assume the downgrade is not intentional?

cpp/third_party/vcpkg_overlays/curl/vcpkg-cmake-wrapper.cmake
poodlewars, line 1: Is there any difference here compared to the `curl` port? I wonder if this could be a symlink too.

cpp/third_party/vcpkg_overlays/curl/portfile.cmake
```diff
@@ -0,0 +1,163 @@
...
+)
+
+if("ssl" IN_LIST FEATURES AND
+    NOT "http3" IN_LIST FEATURES AND
+    # (windows & !uwp) | mingw to match curl[ssl]'s "platform"
+    ((VCPKG_TARGET_IS_WINDOWS AND NOT VCPKG_TARGET_IS_UWP) OR VCPKG_TARGET_IS_MINGW))
+    list(APPEND FEATURE_OPTIONS -DCURL_USE_SCHANNEL=ON)
+endif()
+
+if(VCPKG_TARGET_IS_OSX OR VCPKG_TARGET_IS_IOS)
```
poodlewars: Are these three lines the only customization in this port?

cpp/third_party/vcpkg_overlays/curl/portfile.cmake
```diff
@@ -0,0 +1,163 @@
...
+    list(APPEND FEATURE_OPTIONS -DCURL_USE_SECTRANSP=OFF)
```
poodlewars: I'm struggling to see what the effect of this option is on the curl build, where does the curl build read this?

cpp/third_party/vcpkg_overlays/curl/vcpkg.json
maxim-morozov: Yes, symlink is a good idea and yes, it should 8.17. Thanks

cpp/third_party/vcpkg_overlays/curl/vcpkg-cmake-wrapper.cmake
maxim-morozov: Yes, symlink is a good idea

cpp/third_party/vcpkg_overlays/curl/portfile.cmake
maxim-morozov: Yes, this is the only change.

cpp/third_party/vcpkg_overlays/curl/portfile.cmake
maxim-morozov: The issue is slightly settled and convoluted. 

Because of our VCPKG restriction, we are actually not linking curl 8.17, but using curl 8.4 (https://github.com/man-group/ArcticDB/blob/master/cpp/vcpkg.json#L108):

```
Installing 97/135 curl[core,non-http,sectransp,ssl]:arm64-osx@8.4.0...
Building curl[core,non-http,sectransp,ssl]:arm64-osx@8.4.0...
```

Which is installing sectransp as well, which is visible when running `cmake --trace-expand`. 

In 8.4, the vcpkg definition actually includes sectransp as part of the features - https://github.com/microsoft/vcpkg/blob/b40de44891dc1cab11d4722094ae44807a837b98/ports/curl/vcpkg.json#L103

Eventually it is running the map between features supported is here:
https://github.com/microsoft/vcpkg/blob/b40de44891dc1cab11d4722094ae44807a837b98/ports/curl/portfile.cmake#L33

PR #2821, opened 2025-12-19
Consolidate bindings for Enterprise ABI compatibility

phoebusm:
#### Reference Issues/PRs

Part of https://man312219.monday.com/boards/7852509418/pulses/18378732140 and 
https://man312219.monday.com/boards/7852509418/pulses/8371158618
1. Consolidate some binding functions so they can be reused in different repos
2. Fix sts fixtures missing configuration if the native version store has valid `native_config`
( Now `add_s3_library_to_env` set both native config and class variable for config `aws_auth` and `aws_profile`. They should only be set in native config. It can't be changed until `7.0.0`)

#### Any other comments?
Object error simulation is added to S3ClientTestWrapper. So the class should now have all error simulation feature of MockS3Client. Therefore we should remove and replace MockS3Client. Ticket: https://man312219.monday.com/boards/7852509418/pulses/11041458610
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/util/venv.py
+++ b/python/arcticdb/util/venv.py
@@ -35,20 +35,22 @@ def run_shell_command(
     result = None
     command_string = " ".join(command)
     env = os.environ.copy()
-    env["PYTHONPATH"] = cwd
+    env["PYTHONPATH"] = str(cwd)  # Windows only allow str object here
+    # cwd parameter only affects result of os.getcwd()
+    # env parameter puts cwd at index 1 in os.path; index 0 is always the path of the script
     if is_running_on_windows():
         # shell=True is required for running the correct python executable on Windows
-        result = subprocess.run(command, cwd=cwd, capture_output=True, shell=True)
+        result = subprocess.run(command, cwd=cwd, capture_output=True, shell=True, env=env)
     else:
         # On linux we need shell=True for conda feedstock runners (because otherwise they fail to expand path variables)
         # But to correctly work with shell=True we need a single command string.
         result = subprocess.run(
             command_string,
-            cwd=cwd,  # this only affects os.getcwd()
+            cwd=cwd,
             capture_output=True,
             shell=True,
             stdin=subprocess.DEVNULL,
-            env=env,  # this put cwd at index 1 in os.path; index 0 is always the path of the script
+            env=env,
         )
     if result.returncode != 0:
         error_message = (


cpp/arcticdb/entity/python_bindings_common.hpp
```diff
@@ -0,0 +1,17 @@
...
+ * will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <pybind11/pybind11.h>
+
+namespace arcticdb::entity::apy {
+
+void register_common_entity_bindings(pybind11::module& m, bool local_bindings);
```
poodlewars: What does "common" mean here?

poodlewars: We should avoid `bool` flags on APIs, it's hard to guess what the false in `entity::apy::register_common_entity_bindings(version, false);` means. Instead there should be separate functions for the `true` and `false` cases, which might push down to a private function with the `bool` in its signature.

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -525,6 +530,9 @@ def real_s3_sts_from_environment_variables(
         logger.error(f"Error creating access key: {e}")
         raise e
 
+    out.native_config = NativeVariantStorage(
```
poodlewars, line 533: I don't understand this change, although I've seen your comment,

```
Fix sts fixtures missing configuration if the native version store is reused
( Now add_s3_library_to_env set both native config and class variable for config aws_auth and aws_profile. They should only be set in native config. It can't be changed until 7.0.0)
```

What does `if the native version store is reused` mean? A unit test to go with this change would make this easier to review, and help us avoid breaking it again.

python/arcticdb/storage_fixtures/s3.py
```diff
@@ -525,6 +530,9 @@ def real_s3_sts_from_environment_variables(
...
+        NativeS3Settings(AWSAuthMethod.STS_PROFILE_CREDENTIALS_PROVIDER, profile_name, False)
```
poodlewars: Can we use kwargs for the last argument whenever we create this object? This also has the `True` / `False` readability problem.

cpp/arcticdb/storage/python_bindings_common.hpp
```diff
@@ -0,0 +1,38 @@
...
+    SECRET = 8,
+    ACCESS = 9,
+    BUCKET = 10,
+};
+
+s3::GCPXMLSettings gcp_settings(const pybind11::tuple& t);
+s3::S3Settings s3_settings(const pybind11::tuple& t);
+pybind11::tuple to_tuple(const s3::GCPXMLSettings& settings);
+pybind11::tuple to_tuple(const s3::S3Settings& settings);
+void register_common_bindings(pybind11::module& m, bool local_bindings);
```
IvoDD: Nit: Maybe rename to `register_common_storage_bindings`?

cpp/arcticdb/entity/python_bindings_common.cpp
```diff
@@ -0,0 +1,76 @@
...
+                                key.content_hash(),
+                                key.start_index(),
+                                key.end_index(),
+                                key.type()
+                        );
+                    },
+                    [](py::tuple t) {
+                        util::check(t.size() >= 7, "Invalid AtomKey pickle object!");
+
+                        [[maybe_unused]] const int serialization_version = t[0].cast<int>();
```
vasil-pashov: Is this needed? Why don't we remove it?

cpp/arcticdb/entity/python_bindings_common.cpp
```diff
@@ -0,0 +1,76 @@
...
+                                key.version_id(),
+                                key.creation_ts(),
+                                key.content_hash(),
+                                key.start_index(),
+                                key.end_index(),
+                                key.type()
+                        );
+                    },
+                    [](py::tuple t) {
+                        util::check(t.size() >= 7, "Invalid AtomKey pickle object!");
```
vasil-pashov, line 54: Why not check for `t.size() == 7`

cpp/arcticdb/entity/python_bindings_common.hpp
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software
+ * will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <pybind11/pybind11.h>
```
vasil-pashov: You don't need this header in the .hpp file. You can forward declare pybind11::module and place the header only in the .cpp

cpp/arcticdb/python/python_bindings_common.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include <pybind11/pybind11.h>
```
vasil-pashov: same as https://github.com/man-group/ArcticDB/pull/2821/changes#r2664451476

cpp/arcticdb/entity/python_bindings_common.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <pybind11/pybind11.h>
+#include <pybind11/operators.h>
+#include <pybind11/stl.h>
+
+namespace py = pybind11;
+
+namespace arcticdb::entity::apy {
+
+void register_common_entity_bindings(py::module& m, BindingScope scope) {
+    bool local_bindings = (scope == BindingScope::LOCAL);
```
vasil-pashov, line 23: `register_common_entity_bindings` is only called with BindingScope::GLOBAL why does it have an option to pass LOCAL?

I've probably missed some part of the discussion. So I wanted to double check the intent.

We will distinguish ArcticDB and ArcticDB Enterprise AtomKey and VersionItem? ArcticDB Enterprise users won't be able to pass in AtomKey return from ArcticDB.

This will also mean the the library tool currently won't be able to operate with AtomKeys from Enterprise.

cpp/arcticdb/storage/python_bindings_common.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <pybind11/pybind11.h>
+#include <arcticdb/util/error_code.hpp>
+#include <arcticdb/storage/s3/s3_settings.hpp>
+#include <arcticdb/storage/common.hpp>
+#include <arcticdb/python/python_bindings_common.hpp>
```
vasil-pashov: We can trim some of the includes using fwd declarations.

cpp/arcticdb/entity/python_bindings_common.hpp
phoebusm: That'll work as well. Is it for the sake of compilation speed?

cpp/arcticdb/entity/python_bindings_common.cpp
phoebusm: I think it's for some sort of future proof, if new setting is added to the class.

cpp/arcticdb/entity/python_bindings_common.cpp
phoebusm: > register_common_entity_bindings is only called with BindingScope::GLOBAL why does it have an option to pass LOCAL?

It used by downstream library, namely Enterprise.

phoebusm: > We will distinguish ArcticDB and ArcticDB Enterprise AtomKey and VersionItem? ArcticDB Enterprise users won't be able to pass in AtomKey return from ArcticDB.

It's never supposed to work out of the box as they are defined in different binaries. 
The reason why they are needed to be bound in enterprise, as they are the return type of some functions. If core and enterprise are ABI-compatible, not binding those class will give us pybind exception.
But it will be very useful to provide some helper functions to convert those two types between two binaries

python/tests/compat/arcticdb/test_core_api.py
poodlewars, line 1: These unit tests don't cover the `_core_api.py` functions?

python/arcticdb/_core_api.py
```diff
@@ -0,0 +1,20 @@
+"""
+Copyright 2023 Man Group Operations Ltd.
+NO WARRANTY, EXPRESSED OR IMPLIED.
+
+This module implements a backwards compatible version of msgpack functions.
```
poodlewars: Does it?

python/arcticdb/_core_api.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+
+# Treat everything here as public APIs!!!! Any change made here may break downstream repos!!!!
```
poodlewars: Worth mentioning that we should apply semantic versioning here

cpp/arcticdb/entity/python_bindings_common.cpp
poodlewars: It's for the staged data tokens API, which makes the staged data tokens (which wrap atom keys) pickleable. This helps to use the staged data API across processes which may have different ArcticDB versions.

cpp/arcticdb/python/python_bindings_common.cpp
poodlewars, line 1: Why do we need to make the metrics config pickleable? This relates to this comment on the enterprise PR https://github.com/man-group/arcticdb-enterprise/pull/281#discussion_r2650734598 . All we need is to add accessors to the fields in hte metrics config? It's confusing to use the pickling APIs for this purpose.

python/arcticdb/_core_api.py
phoebusm: Oops ctrl+c error

cpp/arcticdb/entity/python_bindings_common.hpp
phoebusm: Shared by multiple repos

python/arcticdb/storage_fixtures/s3.py
phoebusm: Use kwargs now but didn't update `use_internal_client_wrapper_for_testing` parameter type as it has been here for a while already and it's semi-public.

python/tests/compat/arcticdb/test_core_api.py
phoebusm: I think it should have (The only function in _core_api.py is `env_config_from_lib_config`)

python/arcticdb/storage_fixtures/s3.py
phoebusm: Callers of this functions should have loaded `aws_auth` and `aws_profile` from factory's `native_config`. But by mistake, only `aws_auth` and `aws_profile` settings in factory are set. So I have made the above changes and remove factory's ones in `7.0.0`

cpp/arcticdb/storage/s3/s3_client_wrapper.cpp
```diff
@@ -8,6 +8,7 @@
```
phoebusm, line 8: Please refer to `cpp/arcticdb/storage/mock/s3_mock_client.hpp` for the change in this file. This is for making s3 wrapper client supports error simulation

python/tests/integration/arcticdb/test_s3.py
```diff
@@ -32,8 +32,9 @@
 )
 
 
-def test_s3_storage_failures(mock_s3_store_with_error_simulation):
-    lib = mock_s3_store_with_error_simulation
+@pytest.mark.parametrize("version_store", ["wrapped_s3_version_store", "mock_s3_store_with_error_simulation"])
+def test_s3_storage_failures(request, version_store):
```
poodlewars, line 36: I assume this relates to this discussion https://github.com/man-group/arcticdb-enterprise/pull/281#discussion_r2650778098 . Did the other ideas in that discussion from me and Ivo not work?

The `S3ClientTestWrapper` is now very similar to the `MockS3Client`. What they do boils down to:

- Test for any simulated errors
- Push down to another client (arbitrary for the first, in-memory for the second).

Could we not now implement the `MockS3Client` as an `S3ClientTestWrapper` around an "in-memory" S3 client? (assuming that we actually want to modify the `S3ClientTestWrapper` at all and that mine and Ivo's alternatives did not work).

Since this is testing infra I would be OK with iterating on this after merge, but I do think it needs cleaning up.

python/arcticdb/_core_api.py
```diff
@@ -0,0 +1,15 @@
+"""
+Copyright 2023 Man Group Operations Ltd.
+NO WARRANTY, EXPRESSED OR IMPLIED.
+
+Non-public APIs depended by downstream repos
+Treat everything here as public APIs!!!!
```
poodlewars: This should say, "treat everything here as a public API, governed by our semantic versioning".

python/arcticdb/storage_fixtures/s3.py
poodlewars: Could you point me at the unit test for this change?

cpp/arcticdb/storage/s3/s3_client_wrapper.cpp
```diff
@@ -17,17 +18,24 @@ using namespace object_store_utils;
...
 
-std::optional<Aws::S3::S3Error> S3ClientTestWrapper::has_failure_trigger(const std::string& bucket_name) const {
-    bool static_failures_enabled = ConfigsMap::instance()->get_int("S3ClientTestWrapper.EnableFailures", 0) == 1;
+std::optional<Aws::S3::S3Error> S3ClientTestWrapper::has_failure_trigger(
+        const std::string& s3_object_name, const std::string& bucket_name, StorageOperation operation
+) const {
+    if (auto error = has_bucket_failure_trigger(bucket_name)) {
+        return error;
+    }
+    return has_object_failure_trigger(s3_object_name, operation);
```
IvoDD: I think the way this is currently setup will be problematic for a couple of reasons.

1. `has_object_failure_trigger` will be called on real s3 clients because we don't check for env variables etc. in it.
2. It may in some cases slow down s3 operation. `has_object_failure_trigger` does some more expensive string munging operations especially if object name is very long.

I think we should at least:
- Make the `has_object_failure_trigger` get called only if some (maybe new) config map value is set etc.

Maybe even simpler, can we just not create the MockS3Storage as we were doing before? I guess if the problem is passing down [this config](https://github.com/man-group/ArcticDB/blob/9f6f24f2bdbbc73279816fc6edd7330c7b2a13e9/cpp/arcticdb/storage/s3/nfs_backed_storage.cpp#L131) to enterprise, can we just make that take `config || some_value_from_config_map`?

python/tests/integration/arcticdb/test_s3.py
phoebusm: Replied in the comments

phoebusm: Will create a new ticket to consolidate mock s3 and wrap s3 client

cpp/arcticdb/storage/s3/s3_client_wrapper.cpp
IvoDD, line 27: Why do we need these changes to the `S3ClientTestWrapper`? It feels like it's currently duplicating some logic from `MockS3Client`? 

I'd be fine with moving it here and e.g. making this wrapper be able to wrap around a `MockS3Client` as well. But having this logic in both places feels obsolete.

python/arcticdb/storage_fixtures/s3.py
phoebusm: It's test in enterprise repo but I have updated a test locally for it too.
I have realized this bug has made library made by `real_s3_sts_version_store` fixture falls back to name+key authentication. 
Still most of the test for sts auth uses uri for initialization, which is not affected by this issue

cpp/arcticdb/storage/s3/s3_client_wrapper.cpp
phoebusm: Replied in https://github.com/man-group/arcticdb-enterprise/pull/281#discussion_r2650778098

PR #2824, opened 2025-12-23
Remove AsvBase from recursive_normalizer.py

poodlewars:
I'm working on ripping out `AsvBase` from the benchmark suites as it is complicated and hard to work with. This is a simple start where the recursive_normalizer benchmark never really need it at all.

Test run: https://github.com/man-group/ArcticDB/actions/runs/20467252929

Second test run: https://github.com/man-group/ArcticDB/actions/runs/20578815266

### DIFF
--- a/python/benchmarks/recursive_normalizer.py
+++ b/python/benchmarks/recursive_normalizer.py
@@ -7,13 +7,11 @@ As of the Change Date specified in that file, in accordance with the Business So
 """
 
 import pandas as pd
-from functools import lru_cache
 
 from arcticdb.util.logger import get_logger
 from arcticdb import Arctic
 
 
-@lru_cache(maxsize=None)
 def get_data(num_dict_entry):
     return {
         f"{i}": {"a": pd.DataFrame({"a": [1, 2, 3]}), "b": pd.DataFrame({"b": [1, 2, 3, 4, 5, 6]})}
@@ -25,9 +23,10 @@ def get_symbol_name(dict_entry, symbol_idx):
     return f"nested_dict_sym_{dict_entry}_{symbol_idx}"
 
 
-class LMDBRecursiveNormalizer:
+class LMDBRecursiveNormalizerRead:
     param_names = ["num_dict_entries", "num_symbols"]
     params = [[1000], [5]]
+    ARCTIC_URI = "lmdb://recursive_normalizer_read"
 
     def __init__(self):
         self.logger = get_logger()
@@ -35,7 +34,7 @@ class LMDBRecursiveNormalizer:
         self.lib = None
 
     def setup_cache(self):
-        ac = Arctic("lmdb://recursive_normalizer")
+        ac = Arctic(LMDBRecursiveNormalizerRead.ARCTIC_URI)
         if "lib" in ac:
             ac.delete_library("lib")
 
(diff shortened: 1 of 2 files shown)


python/benchmarks/recursive_normalizer.py
```diff
@@ -26,13 +26,6 @@ def get_symbol_name(dict_entry, symbol_idx):
 
 
 class LMDBRecursiveNormalizer:
-    rounds = 1
-    number = 3
-    repeat = 1
-    min_run_count = 1
-
-    timeout = 1200
```
IvoDD: I think we should have an increased timeout on all benchmarks (or have a clearer way to report when a benchmark times out). There is probably a better way than adding it on every benchmark class.

E.g. [this regression](https://chat-man.slack.com/archives/C02NEG3V38X/p1753972959456059?thread_ts=1753953094.535089&cid=C02NEG3V38X) was not caught before release because it was hard to identify whether asv run was timing out or it has failed with another issue.

python/benchmarks/recursive_normalizer.py
```diff
@@ -26,13 +26,6 @@ def get_symbol_name(dict_entry, symbol_idx):
 
 
 class LMDBRecursiveNormalizer:
-    rounds = 1
```
IvoDD: Not a problem, but noting that this has increased the benchmark runtime from [20s](https://github.com/man-group/ArcticDB/actions/runs/20486726620/job/58870671100#step:13:323) to [60s](https://github.com/man-group/ArcticDB/actions/runs/20467252929/job/58814063825#step:13:276)

IvoDD (approved):

python/benchmarks/real_finalize_staged_data.py
```diff
@@ -59,14 +58,14 @@ def setup(self, cache, num_chunks: int):
 
         self.lib = self.get_library_manager().get_library(LibraryType.MODIFIABLE)
 
-        INITIAL_TIMESTAMP: TimestampNumber = TimestampNumber(0, self.df_cache.TIME_UNIT)  # Synchronize index frequency
+        initial_timestamp: TimestampNumber = TimestampNumber(0, self.df_cache.TIME_UNIT)  # Synchronize index frequency
```
vasil-pashov, line 61: nit: The comment doesn't make sense to me.

python/benchmarks/recursive_normalizer.py
```diff
@@ -6,88 +6,75 @@
...
     def time_write_nested_dict(self, num_dict_entry, num_symbols):
-        self.read_lib._nvs.write(
+        self.lib._nvs.write(
             f"nested_dict_time_write_nested_dict_{num_dict_entry}",
-            self.get_data(num_dict_entry),
+            get_data(num_dict_entry),
             recursive_normalizers=True,
         )
 
     def peakmem_write_nested_dict(self, num_dict_entry, num_symbols):
```
vasil-pashov: Does the @lru_cache affect the RSS? Is the size of the cache part of the benchmark? If the memory used by the write is dwarfed by the cache size we might miss a bug where the amount of memory for the write is increased but it's less than 10% of the total amount.

vasil-pashov (approved):

PR #2825, opened 2025-12-24
3. Remove ASVBase from the finalize staged data benchmarks

poodlewars:
Start to replace the current real storage testing framework.

There are many layers of abstraction in the current code, such as the `TestLibraryManager` and `AsvBase`. There is also a notion of "persistent" storage - long running data stored on S3 and only manually updated, for read benchmarks to work against. This is complicated to work with and fragile - for example it is easy to forget to update "persistent" benchmarking libraries when we update benchmark parameters, so they silently stay stuck on the wrong parameters.

The datasets we write for benchmarking are actually fairly small but they are not being written efficiently at the moment (for example, they are not using our batch APIs).

This PR shows a new design, using `finalize_staged_data` as a starting point.

- Do not have separate `finalize_staged_data` and `real_finalize_staged_data` modules. Parameterize a single one by storage instead. This should be safer, since the exact benchmark code will at least run against LMDB on each PR.
- Control which storages to test against by environment variables (`ARCTICDB_STORAGE_AWS_S3=0`, `ARCTICDB_STORAGE_LMDB=1`) 
- Make sure that we only disable storages when the benchmark executes, rather than at discovery time. This makes sure that `benchmarks.json` includes all the storages so that adhoc real storage runs can execute.

This PR doesn't clean up data written to S3 yet, that will be easier to implement once all the real storage benchmarks are ported over. My plan for that is for the CI to create a temporary bucket, and drop it at the end of the benchmarking run.

Manual run against real storage: https://github.com/man-group/ArcticDB/actions/runs/20850427753/job/59904037259

Manual run against LMDB: https://github.com/man-group/ArcticDB/actions/runs/20850444779


python/benchmarks/finalize_staged_data.py
poodlewars, line 1: We remove `real_finalize_staged_data.py` and instead parameterize this over two `storages`. We also rip out the `AsvBase` base here.

python/benchmarks/finalize_staged_data.py
```diff
@@ -1,127 +1,78 @@
...
-        self.logger.info(f"LIBRARY: {self.lib}")
-        self.logger.info(f"Created Symbol: {self.symbol}")
-        self.lib.finalize_staged_data(self.symbol, mode=StagedDataFinalizeMethod.WRITE)
-
-    def teardown(self, param: int):
-        rmtree(FinalizeStagedData.ARCTIC_DIR)
-        copytree(FinalizeStagedData.ARCTIC_DIR_ORIGINAL, FinalizeStagedData.ARCTIC_DIR, dirs_exist_ok=True)
-        del self.ac
+        self.logger.info(f"SETUP_CACHE TIME: {time.time() - start}")
+        return lib_for_storage
```
IvoDD, line 51: How are the `lib_for_storage` library objects transitioned between the `setup_cache` and `setup_and_benchmark` processes? I assume they are pickled. Does pickling and unpickling a `Library` object work for all storages?

poodlewars: Yeah they're pickled. We have quite careful support for pickling in the `NativeVersionStore`, as people rely on it for multi-processing. It certainly works for LMDB and Amazon, which are the only ones we use at the moment, and if we find storages where this does not work that would be a bug worth fixing in its own right.

python/benchmarks/environment_setup.py
```diff
@@ -61,6 +67,72 @@ class LibraryType(Enum):
...
+    if storage == Storage.LMDB:
+        ac = Arctic(f"lmdb://tmp/{lib_name}")
+    elif storage == Storage.AMAZON:
+        factory = real_s3_from_environment_variables(shared_path=True, validate=True, log=True)
+        ac = factory.create_fixture().create_arctic()
+    else:
+        raise RuntimeError(f"storage {storage} not implemented for benchmark {__file__}")
+
+    if lib_name in ac:
+        ac.delete_library(lib_name)
```
IvoDD: Nit: `delete_library` is a no-op if library doesn't exist, so we can just drop the if.

python/benchmarks/finalize_staged_data.py
```diff
@@ -1,127 +1,81 @@
...
-    Check and benchmark performance of finalize_staged_data().
-    Due to specifics of this procedure we tune asv to make single measurement
-    which would be over a relatively big staged data.
-    """
 
+class FinalizeStagedData:
     number = 1
     rounds = 1
-    repeat = 5
+    repeat = 1
```
IvoDD: I'm a bit weary of designing benchmarks in a way that relies on a single benchmark execution. I.e. doing all of the setup in `setup_cache` which will never be repeated.

It is possible in the future we will want to be able to repeat the benchmark multiple times to decrease the variance. 

I'm ok with leaving `repeat=1` if benchmark is not flaky for now but I think we should `_prepopulate_library` inside `setup` to allow increasing the repeats in the future.

Also would you mind adding comments explaining these like:
```
number = 1 # Not safe to increase, each benchmark run relies on setup
repeat = 1 # Safe to increase, `setup` does required staging to correctly repeat benchmark multiple times
```

python/benchmarks/finalize_staged_data.py
```diff
@@ -1,127 +1,81 @@
...
+        return lib_for_storage
+
+    def _prepopulate_library(self, lib, df_generator):
+        for param in self.num_chunks:
+            initial_timestamp = TimestampNumber(0, df_generator.TIME_UNIT)
+
+            list_of_chunks = [10_000] * param
+
+            for suffix in ("-time", "-mem"):
+                symbol = _symbol_name(param) + suffix
```
IvoDD: Why not add a required `suffix` argument to `_symbol_name` since we just add it every time?

poodlewars: We don't add it every time, eg

```py
        self.symbol = _symbol_name(num_chunks)
```

IvoDD: In that case we still then add the suffix in the benchmarks themselves e.g. `self.symbol + "-time"` in several places.

I don't have a strong preference but it feels confusing to have `self.symbol` not be an actual symbol in the library but rather a symbol prefix. What do you think about either:
- renaming `self.symbol` to `self.symbol_prefix`
- adding `suffix` argument to `_symbol_name` and constructing the symbol in the benchmarks itself

IvoDD (approved):

.github/workflows/benchmark_commits.yml
```diff
@@ -100,10 +99,15 @@ jobs:
         run: | 
             if [ "${{ inputs.suite_to_run }}" == "REAL" ]; then
               # ASV tests with real storage starts with 'real_' in the test name
-              SUITE='^(real_).*'
+              # aseaton: We are currently refactoring away from this pattern to control the storages by environment variables
+              # instead. finalize_staged_data is the first one in the new pattern.
+              SUITE='(^real_.*|^finalize_staged_data.*)'
+              echo "ARCTICDB_STORAGE_AWS_S3=1" >> $GITHUB_ENV
```
poodlewars: TODO re-test that this works in the CI

python/benchmarks/finalize_staged_data.py
```diff
@@ -1,127 +1,85 @@
...
+        self.lib = lib_for_storage[storage]
+        if self.lib is None:
+            raise SkipNotImplemented
 
-from asv_runner.benchmarks.mark import SkipNotImplemented
-
+        assert len(self.lib.list_symbols()) == 0  # check we are in a clean state
+        initial_timestamp = TimestampNumber(0, self.df_generator.TIME_UNIT)
 
-class FinalizeStagedDataWiderDataframeX3(FinalizeStagedData):
```
poodlewars, line 94: This hasn't run for ages. I'm just going to remove it.

python/arcticdb/util/test_utils.py
vasil-pashov, line 1: It looks like this is just copy pasted from `python/arcticdb/util/utils.py` are there any actual changes besides moving to a new file?

python/benchmarks/finalize_staged_data.py
vasil-pashov: I agree, depending on a single run for the time tests doesn't look stable. The peakmem should be mostly fine, though

AS: I've adjusted this back up to 3

python/arcticdb/util/test_utils.py
poodlewars: No, I just wanted to rename it to make it clear it's for testing, since it ends up in the wheel that we release.

python/benchmarks/finalize_staged_data.py
poodlewars: I've adjusted this back up to 3

PR #2827, opened 2025-12-27
Fix normalization of np.str_ and np.bytes_ types

### LINKED ISSUE #2800: Support for `np.str_` types
### Describe the bug

ArcticDB doesn't support `np.str_` types inside dataframes or series.

### Steps/Code to Reproduce

```python
df = pd.DataFrame({"col": [np.str_("asdf")]})
lib.write("sym", df)
```

Produces a:
```
ArcticDbNotYetImplemented: Not supported: normalizing, symbol: sym, Reason: Failed to normalize column 'col' with dtype 'object'. Found first non-null value of type '<class 'numpy.str_'>', but only strings, unicode, and Timestamps are supported. Do you have mixed dtypes in your column?, Setting the pickle_on_failure parameter to True will allow the object to be written. However, many operations (such as date_range filtering and column selection) will not work on pickled data.
```

### Expected Results

Successfully writes the `np.str_` type

Python 3.11

### Backend storage used

_No response_

### Additional Context

_No response_

yashwantbezawada:
#### Reference Issues/PRs

Fixes #2800

When you try to write a DataFrame with `np.str_` values, you get this error:

```
ArcticDbNotYetImplemented: Failed to normalize column 'col' with dtype 'object'. 
Found first non-null value of type '<class 'numpy.str_'>', but only strings, unicode, 
and Timestamps are supported.
```

The issue is in `_accept_array_string()` - it checks `type(v) in (str, bytes)` which does an exact type match. But `np.str_` is its own class that inherits from `str`, so the check fails even though it is basically a string.

Switched to `isinstance(v, (str, bytes))` which handles subclasses properly. Same thing for `coerce_string_column_to_fixed_length_array()` - changed `to_type == str` to `issubclass(to_type, str)`.

#### Any other comments?

Quick sanity check that this makes sense:
```python
>>> isinstance(np.str_("hello"), str)
True
>>> isinstance(np.bytes_(b"hello"), bytes)
True
```

The C++ side already does the right thing - `PyUnicode_Check` and `PyBytes_Check` both check the type hierarchy, so `np.str_` values work fine once they get past the Python normalization.

Added some tests to cover this.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -157,8 +157,9 @@ _SUPPORTED_NATIVE_RETURN_TYPES = Union[FrameData]
 def _accept_array_string(v):
     # TODO remove this once arctic keeps the string type under the hood
     # and does not transform string into bytes
-    # Use isinstance to also support numpy string types (np.str_, np.bytes_)
-    return isinstance(v, (str, bytes))
+    # Use strict type equality to support numpy string types (np.str_, np.bytes_)
+    # but not arbitrary subclasses (see issue #704)
+    return type(v) in (str, bytes, np.str_, np.bytes_)
 
 
 def _is_nan(element):
@@ -188,8 +189,9 @@ def get_sample_from_non_empty_arr(arr, arr_name):
 
 def coerce_string_column_to_fixed_length_array(arr, to_type, string_max_len):
     # in python3 all text will be treated as unicode
-    # Use issubclass to also support numpy string types (np.str_)
-    if issubclass(to_type, str):
+    # Use strict type equality to support numpy string types (np.str_)
+    # but not arbitrary subclasses (see issue #704)
+    if to_type in (str, np.str_):
         if sys.platform == "win32":
             # See https://sourceforge.net/p/numpy/mailman/numpy-discussion/thread/1139250278.7538.52.camel%40localhost.localdomain/#msg11998404
             # Different wchar size on Windows is not compatible with our current internal representation of Numpy strings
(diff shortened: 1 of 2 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -157,7 +157,8 @@ def from_npd_df(df):
 def _accept_array_string(v):
     # TODO remove this once arctic keeps the string type under the hood
     # and does not transform string into bytes
-    return type(v) in (str, bytes)
+    # Use isinstance to also support numpy string types (np.str_, np.bytes_)
+    return isinstance(v, (str, bytes))
```
IvoDD: This will accept any type that inherits `str`, which I think is also a good change.

What do you think about putting that behavior in another test like:
```
    class CustomStrType(str, Enum):
        A = "A"
        B = "B"

    # Test custom string types in DataFrame
    df = pd.DataFrame({"col": [CustomStrType.A, CustomStrType.B]})
    lib.write(sym, df)
    result = lib.read(sym).data
    # Custom string values should be read back as regular strings
    assert result["col"][0] == "A"
    assert result["col"][1] == "B"
```

IvoDD (commented):
Thank you for contributing! Looks good! Just one suggestion for an extra test

python/arcticdb/version_store/_normalization.py
IvoDD: Actually looking back at #704 as pointed out by @vasil-pashov supporting all types inheriting strings has been problematic before.

Let's instead use a strict type equality like:
```
type(v) in (str, bytes, np.str_, np.bytes_)
```

IvoDD (changes requested):

python/arcticdb/version_store/_normalization.py
yashwantbezawada: Done! Changed to strict type equality as suggested:

```python
type(v) in (str, bytes, np.str_, np.bytes_)
```

Also updated `coerce_string_column_to_fixed_length_array()` similarly:

```python
to_type in (str, np.str_)
```

Added comments referencing issue #704 to explain the rationale for future readers.

IvoDD (approved):

PR #2833, opened 2025-12-31
Speed up version_chain benchmarks setup_cache

poodlewars:
The setup_cache creates two libraries, one with the tail of the version chain deleted. We used to write both libraries with Arctic APIs.

I change it here to write most of the versions of one library with Arctic APIs, up to the point where we want to delete the symbol, then copy that state with `shutil.copytree`. This roughly halves the amount of time spent in `setup_cache`, from ~17 minutes down to ~8 on my machine.

I also change the benchmarks to control their number of runs with `sample_time` rather than `number`. The high number was introduced for some benchmarks that run very quickly and so are noisy, but doing that across all the benchmarks isn't suitable as others are slow and stable.


python/benchmarks/version_chain.py
```diff
@@ -19,64 +20,95 @@
...
-                    lib.write(symbol, small_df)
-                    # Only check for deletion once per iteration
-                    if i == delete_point:
-                        lib.delete(symbol)
+        del lib
+        del ac
 
-        adb._ext.unset_config_int("VersionMap.ReloadInterval")
+        shutil.rmtree("version_chain_deleted", ignore_errors=True)
+        shutil.copytree("version_chain", "version_chain_tail_deleted")
```
IvoDD: We're removing `version_chain_deleted` but copying to `version_chain_tail_deleted`.

Maybe add constants for directory names so it's harder to make such error:
```
DIR_UNDELETED = "version_chain"
DIR_TAIL_DELETED = "version_chain_tail_deleted"
CONNECTION_STRING_UNDELETED = f"lmdb://{DIR_UNDELETED}"
CONNECTION_STRING_TAIL_DELETED = f"lmdb://{DIR_TAIL_DELETED}"
```

python/benchmarks/version_chain.py
```diff
@@ -19,64 +20,95 @@
...
+            deletion_point = delete_points[num_versions]
+            for i in range(deletion_point, num_versions):
+                lib.write(symbol, small_df)
+            # reasonableness checks
+            assert lib.read(symbol).version == num_versions - 1
+            # Only versions that have not been deleted are returned by list_versions
+            assert len(lib.list_versions(symbol)) == num_versions - deletion_point
+
+        adb._ext.unset_config_int("VersionMap.ReloadInterval")
+        print("IterateVersionChain: Setup cache took (s) :", time.time() - start_time, file=sys.stderr)
```
IvoDD: Nit: We're printing the time two times. Here and in `setup_cache`. Maybe just keep one?

python/benchmarks/version_chain.py
```diff
@@ -108,30 +140,32 @@ def setup(self, num_versions, caching, deleted):
             # Leave the default reload interval
             pass
 
-        self.ac = Arctic(IterateVersionChain.CONNECTION_STRING)
-        self.lib = self.ac[IterateVersionChain.LIB_NAME]
+        if deleted:
+            ac = Arctic(IterateVersionChain.CONNECTION_STRING_UNDELETED)
+        else:
+            ac = Arctic(IterateVersionChain.CONNECTION_STRING_TAIL_DELETED)
```
IvoDD, line 146: Maybe if and else should be the other way round?

python/benchmarks/version_chain.py
```diff
@@ -108,7 +111,6 @@ def _setup_cache(self):
             assert len(lib.list_versions(symbol)) == num_versions - deletion_point
 
         adb._ext.unset_config_int("VersionMap.ReloadInterval")
-        print("IterateVersionChain: Setup cache took (s) :", time.time() - start_time, file=sys.stderr)
```
IvoDD, line 111: Nit: Also remove the corresponding start time

IvoDD (approved):

PR #2834, opened 2025-12-31
2. ASV performance improvements

poodlewars:
Note that this includes a non-test change I came across while working on this, just to fix the logging when we write empty dataframes, which was malformed. This is e7ecd73.

The other changes are to speed up or otherwise improve our ASV execution.

- Adjust batch parameters to get a sensible execution time
- Remove time_read_batch_pure
- Test batch functions with 1M and 10M rows, not 1M and 1.5M rows which is not interesting
- Several changes to decompose benchmarks in to smaller classes. We have a pattern in several places where a parameterized benchmark class includes some functions that do not actually vary based on the parameters, so they run once for each parameterization for no reason.
- Improve the setup_cache time in comparison_benchmarks.py from about 60s to about 3s
- Fix some places where we write empty dataframes, which causes a lot of log warnings

This has reduced the benchmarking time to roughly 1h40 down from about 2h20m.

Please review commit-by-commit.

### DIFF
--- a/python/benchmarks/basic_functions.py
+++ b/python/benchmarks/basic_functions.py
@@ -7,40 +7,31 @@ As of the Change Date specified in that file, in accordance with the Business So
 """
 
 import time
-from typing import List
+
 from arcticdb import Arctic, QueryBuilder
 from arcticdb.version_store.library import UpdatePayload, WritePayload, ReadRequest
-from arcticdb.util.test import config_context
-from arcticdb_ext.version_store import DataError
 from asv_runner.benchmarks.mark import skip_benchmark
 
 import pandas as pd
 
 from benchmarks.common import *
 
-# Common parameters between BasicFunctions and ModificationFunctions
-WIDE_DF_ROWS = 5_000
-WIDE_DF_COLS = 30_000
 # We use larger dataframes for non-batch methods
-PARAMS = [1_000_000, 1_500_000]
+PARAMS = [1_000_000, 10_000_000]
 PARAM_NAMES = ["rows"]
-BATCH_PARAMS = ([25_000, 50_000], [500, 1000])
+BATCH_PARAMS = ([25_000, 50_000], [100, 200])
 BATCH_PARAM_NAMES = ["rows", "num_symbols"]
 DATE_RANGE = (pd.Timestamp("2022-12-31"), pd.Timestamp("2023-01-01"))
 
 
 class BasicFunctions:
-    number = 5
-    warmup_time = 0
-    timeout = 6000
+    sample_time = 0.1
     rounds = 1
-    CONNECTION_STRING = "lmdb://basic_functions?map_size=20GB"
-    ULTRA_SHORT_WIDE_DF_ROWS = 1
-    WIDE_DF_ROWS = WIDE_DF_ROWS
-    WIDE_DF_COLS = WIDE_DF_COLS
-    DATE_RANGE = DATE_RANGE
+
     params = PARAMS
     param_names = PARAM_NAMES
+
+    CONNECTION_STRING = "lmdb://basic_functions"
 
     def __init__(self):
         self.logger = get_logger()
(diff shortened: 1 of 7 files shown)


python/benchmarks/basic_functions.py
```diff
@@ -7,41 +7,33 @@
...
-    ULTRA_SHORT_WIDE_DF_ROWS = 1
-    WIDE_DF_ROWS = WIDE_DF_ROWS
-    WIDE_DF_COLS = WIDE_DF_COLS
-    DATE_RANGE = DATE_RANGE
+
     params = PARAMS
     param_names = PARAM_NAMES
 
+    CONNECTION_STRING = "lmdb://basic_functions"
+    DATE_RANGE = DATE_RANGE
```
vasil-pashov: Why do we need this?

python/benchmarks/basic_functions.py
```diff
@@ -164,13 +117,86 @@ def peakmem_write_staged(self, rows):
         self.fresh_lib._nvs.compact_incomplete(f"sym", False, False)
 
 
+class ShortWideWrite:
```
vasil-pashov, line 119: Is there a reason to keep the `ShortWideWrite` and `UltraShortWideWrite` separated by have parametrized `ShortWideRead`

python/benchmarks/basic_functions.py
```diff
@@ -164,13 +117,86 @@ def peakmem_write_staged(self, rows):
...
+
+
 class BatchBasicFunctions:
     rounds = 1
-    number = 1
-    repeat = 3
-    warmup_time = 0
-    timeout = 6000
-    CONNECTION_STRING = "lmdb://batch_basic_functions?map_size=20GB"
+    sample_time = 0.1
```
vasil-pashov, line 175: What was the process of coming up with this?

python/benchmarks/comparison_benchmarks.py
```diff
@@ -76,34 +66,34 @@ def delete_if_exists(self, path):
             os.remove(self.path)
 
     def create_dict(self, size):
+        ten_char_strings = [random_string(10) for _ in range(1000)]
```
vasil-pashov: I doubt it will matter much but maybe it'll be a good idea to add some seeds for all the random stuff below

python/benchmarks/modification_functions.py
```diff
@@ -0,0 +1,243 @@
...
+        self.lib.update("sym", self.df_update_half)
+
+    def time_update_upsert(self, lad: LargeAppendDataModify, rows):
+        self.lib.update("sym", self.df_update_upsert, upsert=True)
+
+    def time_append_single(self, lad: LargeAppendDataModify, rows):
+        self.lib.append("sym", self.df_append_single)
+
+    def time_append_large(self, lad: LargeAppendDataModify, rows):
+        large: pd.DataFrame = lad.df_append_large[rows].pop(0)
```
vasil-pashov, line 118: `lad` seems to be used only here. Does it make sense to make a separate benchmark class for it?

python/benchmarks/modification_functions.py
```diff
@@ -149,31 +111,133 @@ def time_update_half(self, lad: LargeAppendDataModify, rows):
...
 
-    def time_delete_multiple_versions(self, lad: LargeAppendDataModify, rows):
-        self.lib.delete("sym_delete_multiple", list(range(99)))
 
-    def time_delete_short_wide(self, lad: LargeAppendDataModify, rows):
+class ShortWideModificationFunctions:
+    rounds = 1
+    number = 1  # We do a single run between setup and teardown because we e.g. can't delete a symbol twice
+    repeat = 2
+    warmup_time = 0
```
phoebusm, line 129: I wonder why it's included in some tests and some doesn't

python/benchmarks/version_chain.py
```diff
@@ -21,7 +21,7 @@
 
 
 class IterateVersionChain:
-    timeout = 6000
+    timeout = 1000
```
phoebusm, line 24: I see this has been tweaked a few times. Given that its backend is LMDB so the storage less likely to fail, shall this be relaxed to prioritize outputting result at the end?
No strong opinion on this though

python/benchmarks/basic_functions.py
poodlewars: Thanks, the write ones can be folded together. They should be kept separate from the read one, so that their setup `df` does not affect the peakmem of the read benchmarks.

python/benchmarks/basic_functions.py
poodlewars: Nothing very scientific. I wanted to make sure that we re-run the faster and more volatile benchmark functions many times, but don't waste time re-running the slower and more stable benchmark functions a fixed number of times.

python/benchmarks/comparison_benchmarks.py
poodlewars: Yeah, I'll make them always be ASCII strings too.

python/benchmarks/modification_functions.py
poodlewars: Does the comment on line 127 help?

python/benchmarks/version_chain.py
poodlewars: The idea is that if it is running for an extremely long time, I'd rather it fails and we fix it. Otherwise there are limited ways to stop the benchmarking time growing and growing.

python/benchmarks/modification_functions.py
poodlewars: It would do yes, but since there aren't any peakmem tests here it doesn't do any harm as it is. The `lad` needs a lot of the same setup as the other benchmarks (so that there is something to append to) so it keeps the code a bit simpler to leave this as is.

PR #2841, opened 2026-01-07
Bugfix/10722938876 fix listing operations on s3 express

alexowens90:
#### Reference Issues/PRs
[10722938876](https://man312219.monday.com/boards/7852509418/views/168855452/pulses/10722938876)

S3 express (aka directory) buckets do not support `ListObjectsV2` requests with prefixes that do not end in a `/` character. At least in symbol list compaction (and possibly elsewhere), we rely on listing operations that do not conform to this API.
The main change is to trap when we get the AWS error corresponding to this request, and to retry the listing operation with all characters after the final `/` character stripped, and perform any further filtering of the returned list on the client side.
Note that this is already how the NFS-layout storage works due to the directory sharding, this solution is modelled after that one.

If it is identified that we are working with a directory bucket, then this information is stored in the `S3Storage` class, so that subsequent listing operations can skip the request that is doomed to fail, and go straight to the fallback mechanism.

### DIFF
--- a/cpp/arcticdb/storage/mock/s3_mock_client.cpp
+++ b/cpp/arcticdb/storage/mock/s3_mock_client.cpp
@@ -209,8 +209,7 @@ S3Result<ListObjectsOutput> MockS3Client::list_objects(
         auto res = Aws::S3::S3Error(Aws::Client::AWSError<Aws::S3::S3Errors>(
                 failure.first, "Simulated error", "Simulated error message for list_objects", failure.second
         ));
-        // We don't want to make the real implementation of list_objects non-const, hence the const_cast
-        const_cast<std::deque<std::pair<Aws::S3::S3Errors, bool>>&>(list_objects_failures_).pop_front();
+        list_objects_failures_.pop_front();
         return {res};
     }
     std::scoped_lock<std::mutex> lock(mutex_);
--- a/cpp/arcticdb/storage/mock/s3_mock_client.hpp
+++ b/cpp/arcticdb/storage/mock/s3_mock_client.hpp
@@ -85,7 +85,7 @@ class MockS3Client : public S3ClientInterface {
     // Failing listing operations based on symbol names isn't sufficient for testing the express bucket fallback
     // mechanism. Allow setting a list of failure modes for list_objects calls. This list will be popped from until
     // empty, at which point the listing operation will succeed.
-    std::deque<std::pair<Aws::S3::S3Errors, bool>> list_objects_failures_;
+    mutable std::deque<std::pair<Aws::S3::S3Errors, bool>> list_objects_failures_;
 };
 
 } // namespace arcticdb::storage::s3


cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -498,20 +490,65 @@ PathInfo calculate_path_info(
...
+void adjust_for_directory_bucket_limitations(KeyType key_type, PathInfo& path_info, Visitor<Predicate>& visitor) {
+    log::storage().debug(
+            "Storage does not support prefix matches not ending in delimiter '/'. Falling back to "
+            "iterating all keys of type {} and prefix matching in memory.",
+            key_type
+    );
+    // It is possible that should_retry_with_directory_bucket_limitations returned true for some other reason when we
+    // are not in a directory bucket. In that case, the next call to do_iterate_type_impl/
+    // do_visit_object_sizes_for_type_impl should fail in the same way. Then we will raise, and the directory_bucket_
+    // field of the calling S3Storage object will not be modified.
```
poodlewars: Note to self: check where this field is maintained

cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
```diff
@@ -217,34 +217,28 @@ void NfsBackedStorage::do_remove(std::span<VariantKey> variant_keys, RemoveOpts)
...
+    // We need this prefix filtering here instead of a regex like in s3/azure
     // because we are doing sharding through subdirectories
     // and the prefix might be partial(e.g. "sym_" instead of "sym_123")
     // so it cannot be hashed to the correct shard
-    const IterateTypePredicate func = [&v = visitor, prefix = prefix](VariantKey&& k) {
+    const IterateTypePredicate visitor_with_prefix_filtering = [&visitor, &prefix](VariantKey&& k) {
         auto key = unencode_object_id(k);
-        if (prefix.empty() || variant_key_view(key).find(prefix) != std::string::npos) {
-            return v(std::move(key));
+        if (prefix.empty()) {
```
poodlewars: Can re-use the `prefix_matching_visitor` logic

cpp/arcticdb/storage/test/test_s3_storage.cpp
```diff
@@ -414,3 +413,40 @@ TEST_F(S3StorageFixture, test_matching_key_type_prefix_list) {
 
     ASSERT_EQ(list_in_store(store, entity::KeyType::LOG), log_symbols);
 }
+
+// The first call to ListObjectsV2 with a directory bucket and a prefix not ending in the '/' delimiter returns this
+// error. We then retry effectively without the prefix, which should succeed
+TEST_F(S3StorageFixture, test_list_directory_bucket_success) {
+    auto* mock_s3_client = dynamic_cast<MockS3Client*>(&store.client());
+    mock_s3_client->add_list_objects_failure(Aws::S3::S3Errors::INVALID_REQUEST, false);
```
poodlewars: The bool on the API is hard to read, could have separate APIs for retryable and non-retryable errors?

cpp/arcticdb/storage/s3/detail-inl.hpp
poodlewars, line 555: I suppose there is also the case where (with a non-directory bucket):
- The first do_visit_object_sizes_for_type_impl fails (some intermittent problem on the storage)
- The second succeeds
- We mark the storage as a directory_bucket_ incorrectly

But I suppose that would be a bug in the S3 backend.

poodlewars (approved):
I had a look through the AWS docs trying to see a better way than this "fail, adapt, retry". I noticed that on AWS, S3 directory buckets have to end `--x-s3`, and general buckets cannot. If we also had a way to test whether a bucket is in AWS or not, we could make the decision explicitly, rather than using this vague failure.

HeadBucket also includes some responses that are only present for directory buckets.

These ways have their own downsides, but perhaps worth considering.

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -532,14 +569,18 @@ bool do_iterate_type_impl(
...
-                if (visitor(std::move(k))) {
+                if (visitor.primary_visitor_(std::move(k))) {
                     return true;
                 }
                 ARCTICDB_SUBSAMPLE(S3StorageCursorNext, 0)
             }
             continuation_token = output.next_continuation_token;
         } else {
             const auto& error = list_objects_result.get_error();
+            if (should_retry_with_directory_bucket_limitations(error.GetErrorType(), path_info, visitor)) {
```
maxim-morozov, line 600: I wonder if we can use AWS SDK to get information about the bucket. Something like using AccessAnalyzer here:  https://github.com/aws/aws-sdk-cpp/blob/main/generated/src/aws-cpp-sdk-accessanalyzer/include/aws/accessanalyzer/model/ResourceType.h#L35

cpp/arcticdb/storage/s3/detail-inl.hpp
alexowens90: Discussed in person, not an issue that is realistically likely to occur

cpp/arcticdb/storage/s3/detail-inl.hpp
alexowens90: AFAICT this is not set anywhere within the S3 client or as part of responses

cpp/arcticdb/storage/mock/s3_mock_client.cpp
```diff
@@ -196,6 +204,15 @@ S3Result<ListObjectsOutput> MockS3Client::list_objects(
         const std::string& name_prefix, const std::string& bucket_name,
         const std::optional<std::string>& continuation_token
 ) const {
+    if (!list_objects_failures_.empty()) {
+        auto failure = list_objects_failures_.front();
+        auto res = Aws::S3::S3Error(Aws::Client::AWSError<Aws::S3::S3Errors>(
+                failure.first, "Simulated error", "Simulated error message for list_objects", failure.second
+        ));
+        // We don't want to make the real implementation of list_objects non-const, hence the const_cast
+        const_cast<std::deque<std::pair<Aws::S3::S3Errors, bool>>&>(list_objects_failures_).pop_front();
```
vasil-pashov: IMO making `list_objects_failures_` mutable is a bit cleaner than const_casting.

cpp/arcticdb/storage/s3/s3_storage.cpp
```diff
@@ -120,30 +120,47 @@ void GCPXMLStorage::do_remove(VariantKey&& variant_key, RemoveOpts) {
     detail::do_remove_no_batching_impl(keys, root_folder_, bucket_name_, client(), FlatBucketizer{});
 }
 
+IterateTypePredicate prefix_matching_visitor(const IterateTypePredicate& visitor, const std::string& prefix) {
+    return [&](VariantKey&& key) { return detail::visit_if_prefix_matches(visitor, prefix, std::move(key)); };
```
vasil-pashov, line 124: nit: I'm not a fan of such wrappers as it obfuscates the fact that the returned lambda is capturing by refence.

vasil-pashov (approved):

PR #2842, opened 2026-01-08
1. ASV: Use master benchmark results from the database rather than re-running on each PR

poodlewars:
Each PR currently runs ASV twice - once using a build from master and once using a build from the HEAD commit. `asv continuous` then compares the results.

This PR instead looks up the "most recent" master commit that exists in the ASV database and compares against that. The step fails if we cannot find results for a master commit that's at least 2 days old. We run benchmarking on master in a nightly job.

Advantages: This should ~halve our ASV execution time. We often have issues where after changes to benchmarking code, the benchmarks cannot actually execute against master (eg we have refactored some utilities the benchmarks use). So we currently have to merge PRs with their benchmarking failing.

Disadvantages: Could cause issues if our master benchmarking repeatedly fails. Complexity in the workflow. New or modified benchmarks in the PR won't run against master.

Manual run to test the non-PR workflow: https://github.com/man-group/ArcticDB/actions/runs/20817761678

Test workflow run with a dummy performance regression: https://github.com/man-group/ArcticDB/actions/runs/20820771059/job/59808590534?pr=2842


build_tooling/transform_asv_results.py
```diff
@@ -111,18 +113,67 @@ def save_asv_results(lib, json_path):
...
+    print(f"origin/master last commit was at {master_date_pd}")
+
+    n = 0
+    stored_commit_ts = master_date_pd
+
+    # Scan back over master commits origin/master~N until either:
+    # - We find one in the ASV database
+    # - We do not find any results in the ASV database for a commit at least 2 days old
+    # We run benchmarks on master nightly, so 2 days should be plenty for the master results to become available
+    while master_date_pd - stored_commit_ts < pd.Timedelta(days=2):
```
IvoDD: This would not find anything if master has no pushes for the last 2 days. So I think every Monday we would end up rebuilding master until some PR is merged.

I think just making this a do while should be good enough for this case. I.e. if latest master commit is older than 2 days it ought to be in ASV database

IvoDD: Thinking a bit more we can also have the odd case where we have master with no pushes for the last 2 days with recent ASV results, but then when we push to master we would suddenly disregard all follow up builds.

Do we have information on when the ASV database was populated? I think we'd have clearer semantics if we use time of last ASV benchmark run.

IvoDD (approved):

PR #2844, opened 2026-01-09
4. Rewrite list operations benchmarks

poodlewars:
Please see PR #2825 .

This builds on that PR in the same way, aiming to simplify our benchmarking against real storages.

This PR rewrites `real_list_operations.py` by folding it in to existing benchmarks that run over LMDB.

I don't try to precisely replace the old `real_list_operations.py` benchmark - I'm just trying to make sure we get worthwhile coverage of the APIs in a benchmark that can run across arbitrary storages, without the complexity of the old framework.

Manual real storages run: https://github.com/man-group/ArcticDB/actions/runs/21010016553


python/benchmarks/list_snapshots.py
```diff
@@ -6,117 +6,80 @@
...
 
-    # For each test param n1xn2 defined:
-    #   n1 - the number of symbols in the library
-    #   n2 - Number of snapshots per each symbol
-    #        (We create also snapshot for each version of the symbol)
-    params = ["20x10", "40x20"]
-    param_names = ["symbols_x_snaps_per_sym"]
 
-    ARCTIC_URL = "lmdb://list_functions"
+class Snapshots:
```
poodlewars, line 30: This is not an exact replacement for the old `list_snapshots` benchmark. I think it's clearer than the old one.

python/benchmarks/list_symbols.py
```diff
@@ -6,54 +6,55 @@
...
+    storages = [Storage.LMDB, Storage.AMAZON]
+    num_symbols = [100, 1000]
 
-    rows = 50
+    params = [num_symbols, storages]
+    param_names = ["num_symbols", "storage"]
 
     def __init__(self):
         self.logger = get_logger()
+        self.test_counter = (
```
poodlewars, line 34: This was a good idea from the real storage benchmark that I've ported over here.

python/benchmarks/environment_setup.py
```diff
@@ -80,7 +80,7 @@ def is_storage_enabled(storage: Storage) -> bool:
         raise RuntimeError(f"Only LMDB and AMAZON storages are supported, received {storage}")
 
 
-def create_library(storage: Storage) -> Library:
+def create_libraries(storage: Storage, library_names: List[str]) -> List[Optional[Library]]:
```
vasil-pashov: I'm a bit confused about `library_names`. I'd expect that this defines the names for the libraries this function creates. However it seems that on line 123 we're creating libraries with random names. Can we get a comment on the purpose of this?

python/benchmarks/environment_setup.py
```diff
@@ -116,19 +116,42 @@ def create_library(storage: Storage) -> Library:
...
+def create_library(storage: Storage) -> Optional[Library]:
+    """
+    Create a library for a given storage test, for use in ASV benchmarking runs.
+
+    See create_libraries docstring for notes on running locally.
+    """
+    if not is_storage_enabled(storage):
+        return None
+
+    lib_name = "".join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
```
vasil-pashov: nit: we have `random_ascii_string` in test.py

python/benchmarks/environment_setup.py
```diff
@@ -116,19 +117,42 @@ def create_library(storage: Storage) -> Library:
...
+
+
+def create_library(storage: Storage) -> Optional[Library]:
+    """
+    Create a library for a given storage test, for use in ASV benchmarking runs.
+
+    See create_libraries docstring for notes on running locally.
+    """
+    if not is_storage_enabled(storage):
+        return None
```
IvoDD: Nit: I think we can skip this `if`. `create_libraries` would return `[None]` in this case

python/benchmarks/list_snapshots.py
```diff
@@ -6,117 +6,85 @@
...
+                print(f"lib_name={lib_name}, lib={lib}", file=sys.stderr)
+                if lib is None:
+                    continue
+                writes = [WritePayload(f"sym_{i}", simple_df) for i in range(n_syms)]
+
+                lib.write_batch(writes)
+
+                inputs = [(lib, i, md_size) for i in range(n_snaps)]
+                with multiprocessing.Pool(10) as p:
+                    p.starmap(create_snapshot, inputs)
```
IvoDD: I vaguely remember some issues with multiprocessing (if using fork) [here](https://chat-man.slack.com/archives/C02NEG3V38X/p1737645992945079?thread_ts=1737449305.537809&cid=C02NEG3V38X).

If we decide to use multiprocessing for this we should ensure we're not using `start_method=="fork"`. I think I'd prefer if we simplify this to just a for loop. 

The PR run which uses only lmdb wouldn't really be able to benefit from this either way because lmdb does some locks on writes IIRC.

python/benchmarks/list_snapshots.py
```diff
@@ -6,117 +6,85 @@
...
+        return libs_for_storage
+
+    def setup(self, libs_for_storage, storage, num_symbols, num_snapshots, metadata_entries, load_metadata):
+        self.lib = libs_for_storage[storage][get_lib_name(num_symbols, num_snapshots, metadata_entries)]
+        if self.lib is None:
+            raise SkipNotImplemented
+
+    def time_list_snapshots(
+        self, libs_for_storage, storage, num_symbols, num_snapshots, metadata_entries, load_metadata
+    ):
```
IvoDD, line 83: Any particular reason you have not added a peakmem benchmark as well?

python/benchmarks/list_symbols.py
```diff
@@ -6,54 +6,61 @@
...
-    def setup(self, num_symbols):
-        self.ac = Arctic("lmdb://list_symbols")
-        self.lib = self.ac[f"{num_symbols}_num_symbols"]
+        simple_df = pd.DataFrame({"a": [1]})
+        write_payloads = [WritePayload(f"{i}", simple_df) for i in range(num_symbols)]
+        self.lib.write_batch(write_payloads)
 
-    def time_list_symbols(self, num_symbols):
+    def time_list_symbols(self, *args):
+        assert self.test_counter == 1
```
IvoDD: Shouldn't this increment the counter so that we can avoid this method being called multiple times withing setup teardown?

python/benchmarks/list_versions.py
```diff
@@ -75,45 +75,91 @@ def _sym_name(self, sym_idx):
...
+        latest_only,
+        skip_snapshots,
+    ):
         self.check_if_skipped_parameter(
-            num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots
+            storage, num_symbols, num_versions, num_snapshots, symbol, snapshot, latest_only, skip_snapshots
         )
-        self.ac = Arctic("lmdb://list_versions")
-        self.lib = self.ac[self._lib_name(num_symbols, num_versions, num_snapshots)]
+        self.lib = lib_for_storage[storage][self._lib_name(num_symbols, num_versions, num_snapshots)]
```
IvoDD, line 131: Won't this fail if a storage is disabled? `lib_for_storage[storage]` would be `None` and calling `None[something]` would presumably break? 

I think this is not a problem for `list_snapshots.py` because there in `setup_cache` we are first creating the libraries and then skipping as opposed to [here](https://github.com/man-group/ArcticDB/pull/2844/changes#diff-3795b046877a4bd096fadfe4c5802d53c4215ed8070112977666a839aa652ec9R92).

How is this passing in CI now? Maybe I'm misunderstanding something

python/benchmarks/list_snapshots.py
poodlewars: I'll change to spawn. This was important because creating the snapshots in serial dominates the benchmarking time on S3.

python/benchmarks/list_versions.py
poodlewars: The `check_if_skipped_parameter` tests whether the storage is enabled. We could instead have this look more like the other tests, but since this file already has a function responsible for skipping stuff, it seemed best to keep that pattern here.

IvoDD: Ah sorry missed it, makes sense

python/benchmarks/list_symbols.py
```diff
@@ -54,13 +54,18 @@ def setup(self, lib_for_storage, num_symbols, storage):
         self.lib.write_batch(write_payloads)
 
     def time_list_symbols(self, *args):
-        assert self.test_counter == 1
+        self._check_test_counter()
+        self.test_counter += 1
```
IvoDD: Nit: We increment the counter twice.

PR #2845, opened 2026-01-12
Remove python 3.8 build

phoebusm:
* Remove python 3.8 build
* Upgrade pybind11 to 3.0.1 as legacy python version support is no longer needed
#### Any other comments?
It alsos frees up disk space at build wheel stage to fix the CI
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -130,7 +130,10 @@ jobs:
     name: Executes asv tests checks
     timeout-minutes: 120
     runs-on: ubuntu-latest
-    container: ghcr.io/man-group/arcticdb-dev:${{ inputs.dev_image_tag || 'latest' }}
+    container:     
+      image: ghcr.io/man-group/arcticdb-dev:${{ inputs.dev_image_tag || 'latest' }}
+      volumes:
+        - /:/mnt
     env:
         # 0 - uses S3 Cache, 1 - uses GHA cache
         # this way the external PRs can use the GHA cache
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -229,8 +229,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # This uses old arcticdb versions which won't work on python < 3.12
-        python3: ${{fromJson(inputs.run_full_matrix_of_persistent_tests && '[8, 9, 10, 11]' || '[11]')}}
+        # This uses old arcticdb versions which won't work on python >= 3.12
```
phoebusm, line 232: I think that's what should be as the version to be test (`v3.0.0`) has no 3.12 build

.github/workflows/build_steps.yml
```diff
@@ -142,6 +142,14 @@ jobs:
           nohup rm -rf /mnt/usr/local/lib/android &
         continue-on-error: true
 
+      - name: Remove GitHub default packages (Base Runner)
```
poodlewars, line 145: Why do we have both this and the step above it? Any idea why we only run the step above it for `cpp-tests`? And why we only use your step on `build-python-wheels`?

Looks like we also need a similar change for the analysis workflow.

I'll approve anyway since this is obviously better than the CI being broken and can iterate on it later.

cpp/third_party/pybind11
poodlewars, line 1: Our current version is 2.13.6. This takes us to 3.0.1.

Changelog https://pybind11.readthedocs.io/en/stable/changelog.html 

Their upgrade guide doesn't recommend any changes beyond this pin update, so I think we should be OK. https://pybind11.readthedocs.io/en/stable/upgrade.html

.github/workflows/build_steps.yml
phoebusm: The above one is for jobs running in a container, namely c++ tests

phoebusm: Pushed commit for fixing analysis workflow. CI run: https://github.com/man-group/ArcticDB/actions/runs/20959891968/job/60233913135
The publish step in the run fails as expected, as the commit is not on master branch yet

.github/workflows/analysis_workflow.yml
```diff
@@ -130,7 +130,10 @@ jobs:
     name: Executes asv tests checks
     timeout-minutes: 120
     runs-on: ubuntu-latest
-    container: ghcr.io/man-group/arcticdb-dev:${{ inputs.dev_image_tag || 'latest' }}
+    container:     
+      image: ghcr.io/man-group/arcticdb-dev:${{ inputs.dev_image_tag || 'latest' }}
+      volumes:
+        - /:/mnt
```
vasil-pashov, line 136: Why are we mounting this?

phoebusm: It's a hack so any workflow with container can have the base files deleted in the container

PR #2846, opened 2026-01-13
5. ASV Rewrite remaining real storage benchmarks and rip out framework for them

poodlewars:
This PR rewrites the real storage ASV benchmarks in to a new style where we parameterize existing benchmarks over storages. See PR #2825 for a description. This abandons the idea of "persistent storages" for benchmarking which I think is unnecessary - with proper use of Arctic APIs we can set up benchmarks on real storages in a reasonable time.

I've just removed the `comparison_benchmarks` since they rely on manual inspection to compare results across Arctic / Pandas and Parquet. I doubt we are getting any value from them.

Since this removes the last of the old "real storages" benchmarks this PR also rips out the old, complex, framework built on `AsvBase`. This is why we are able to delete so much code in `environment_setup.py`.

It also makes some CI changes that I've commented on inline.

After this is merged I will rewrite the docs https://github.com/man-group/ArcticDB/wiki/ASV-Benchmarks:-Real-storage-tests to briefly explain the new situation (which is a lot simpler than the old one).

DONE - ASV run against real storages https://github.com/man-group/ArcticDB/actions/runs/21211440282

### DIFF
--- a/python/benchmarks/environment_setup.py
+++ b/python/benchmarks/environment_setup.py
@@ -5,7 +5,6 @@ As of the Change Date specified in that file, in accordance with the Business So
 """
 
 import random
-import string
 from enum import Enum
 import os
 from typing import List, Optional, Iterable, Dict
--- a/python/.asv/results/benchmarks.json
+++ b/python/.asv/results/benchmarks.json
@@ -256,245 +256,378 @@
         "warmup_time": -1
     },
     "basic_functions.BasicFunctions.peakmem_read": {
-        "code": "class BasicFunctions:\n    def peakmem_read(self, rows):\n        self.lib.read(f\"sym\").data\n\n    def setup(self, rows):\n        self.ac = Arctic(BasicFunctions.CONNECTION_STRING)\n    \n        self.df = generate_pseudo_random_dataframe(rows)\n    \n        self.lib = self.ac[get_prewritten_lib_name(rows)]\n        self.fresh_lib = self.get_fresh_lib()\n\n    def setup_cache(self):\n        start = time.time()\n        self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")",
+        "code": "class BasicFunctions:\n    def peakmem_read(self, *args):\n        self.lib.read(f\"sym\").data\n\n    def setup(self, libs_for_storage, rows, use_query_stats, storage):\n        self.lib = libs_for_storage[storage][_lib_name(rows)]\n        if self.lib is None:\n            raise SkipNotImplemented\n    \n        if use_query_stats:\n            if storage != Storage.AMAZON:\n                raise SkipNotImplemented(\"Query stats only supported against S3\")\n            qs.enable()\n    \n        self.df = generate_pseudo_random_dataframe(rows)\n\n    def setup_cache(self):\n        start = time.time()\n        libs_for_storage = self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")\n        return libs_for_storage",
         "name": "basic_functions.BasicFunctions.peakmem_read",
         "param_names": [
-            "rows"
+            "num_rows",
+            "use_query_stats",
+            "storage"
         ],
         "params": [
             [
                 "1000000",
                 "10000000"
+            ],
+            [
+                "True",
+                "False"
+            ],
+            [
+                "<Storage.LMDB: 2>",
+                "<Storage.AMAZON: 1>"
             ]
         ],
-        "setup_cache_key": "basic_functions:41",
+        "setup_cache_key": "basic_functions:50",
+        "timeout": 600,
         "type": "peakmemory",
         "unit": "bytes",
-        "version": "4ed305e5b21d4c3df22cc239ad54e2ab9b8acf19c2a689e264e9145eca1a2094"
+        "version": "7e0663f5f53179157fae210d779bcfcedfc83dbfd58837a07832da84d46e9ef5"
     },
     "basic_functions.BasicFunctions.peakmem_read_with_columns": {
-        "code": "class BasicFunctions:\n    def peakmem_read_with_columns(self, rows):\n        COLS = [\"value\"]\n        self.lib.read(f\"sym\", columns=COLS).data\n\n    def setup(self, rows):\n        self.ac = Arctic(BasicFunctions.CONNECTION_STRING)\n    \n        self.df = generate_pseudo_random_dataframe(rows)\n    \n        self.lib = self.ac[get_prewritten_lib_name(rows)]\n        self.fresh_lib = self.get_fresh_lib()\n\n    def setup_cache(self):\n        start = time.time()\n        self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")",
+        "code": "class BasicFunctions:\n    def peakmem_read_with_columns(self, *args):\n        COLS = [\"value\"]\n        self.lib.read(f\"sym\", columns=COLS).data\n\n    def setup(self, libs_for_storage, rows, use_query_stats, storage):\n        self.lib = libs_for_storage[storage][_lib_name(rows)]\n        if self.lib is None:\n            raise SkipNotImplemented\n    \n        if use_query_stats:\n            if storage != Storage.AMAZON:\n                raise SkipNotImplemented(\"Query stats only supported against S3\")\n            qs.enable()\n    \n        self.df = generate_pseudo_random_dataframe(rows)\n\n    def setup_cache(self):\n        start = time.time()\n        libs_for_storage = self._setup_cache()\n        self.logger.info(f\"SETUP_CACHE TIME: {time.time() - start}\")\n        return libs_for_storage",
         "name": "basic_functions.BasicFunctions.peakmem_read_with_columns",
         "param_names": [
-            "rows"
+            "num_rows",
+            "use_query_stats",
+            "storage"
         ],
         "params": [
             [
                 "1000000",
                 "10000000"
+            ],
+            [
+                "True",
+                "False"
+            ],
+            [
+                "<Storage.LMDB: 2>",
+                "<Storage.AMAZON: 1>"
             ]
         ],
-        "setup_cache_key": "basic_functions:41",
+        "setup_cache_key": "basic_functions:50",
+        "timeout": 600,
         "type": "peakmemory",
         "unit": "bytes",
-        "version": "0655902640cee72832e909202e4187147f1378b546bd5b3518ab19f4004d5b32"
+        "version": "ccfd86bc53c7545e3e514a4987ec117bbd9a875392f85c425723498ba396f13d"
     },
     "basic_functions.BasicFunctions.peakmem_read_with_date_ranges": {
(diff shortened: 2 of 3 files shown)


.github/actions/set_persistent_storage_env_vars/action.yml
```diff
@@ -26,7 +26,6 @@ runs:
         if [[ -n "${{ inputs.persistent_storage }}" && "${{inputs.persistent_storage}}" != "no" ]]; then 
           echo "ARCTICDB_PERSISTENT_STORAGE_TESTS=1" >> $GITHUB_ENV
         fi
-        echo "ARCTICDB_PERSISTENT_STORAGE_UNIQUE_ID=${{ github.ref_name }}_${{ github.run_id }}" >> $GITHUB_ENV
```
poodlewars, line 29: This was just unused.

.github/workflows/analysis_workflow.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       commit: ${{ matrix.commits }}
       benchmark_all_tags: ${{ inputs.benchmark_all_tags || false }}
       dev_image_tag: ${{ inputs.dev_image_tag || 'latest' }}
-      storage: ${{  github.event_name == 'schedule' && 'ALL' || inputs.storage || 'LMDB'}}
+      storage: ${{  github.ref_name == 'master' && 'ALL' || inputs.storage || 'LMDB'}}
```
poodlewars, line 75: This is to make sure that the database of results on master always includes results across all storages, since now we run it on push to master as well as on this schedule.

.github/workflows/benchmark_commits.yml
```diff
@@ -100,22 +108,14 @@ jobs:
       - name: Configure what suite or tests to execute
         shell: bash -el {0}
         run: | 
+            SUITE='^.*'
             if [ "${{ inputs.storage }}" == "REAL" ]; then
-              # ASV tests with real storage starts with 'real_' in the test name
-              # aseaton: We are currently refactoring away from this pattern to control the storages by environment variables
```
poodlewars, line 105: This PR gets rid of the last examples of the old pattern, so we no longer need this logic.

.github/workflows/benchmark_commits.yml
```diff
@@ -55,24 +60,27 @@ jobs:
...
 
-      - name: Set persistent storage variables
-        uses: ./.github/actions/set_persistent_storage_env_vars
-        with:
-          bucket: "arcticdb-ci-benchmark-results"
-          aws_access_key: "${{ secrets.AWS_S3_ACCESS_KEY }}"
-          aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"
-          strategy_branch: "${{ inputs.commit }}"
-          shared_storage_prefix: "_github_runner_"
+      - name: Create test bucket
```
poodlewars, line 74: We used to also use the ASV bucket `arcticdb-ci-benchmark-results` as a scratch area for our benchmarking runs. Instead I now create a clean bucket for each run, for the ASV benchmarks to write Arctic data to. This keeps thing cleaner because we can just drop the entire bucket at the end, we don't need to worry about any cleanup in our own benchmarks.

build_tooling/test_profiles/lmdb_v1_only.env
poodlewars, line 1: These `test_profiles` were lying around unreferenced. I really doubt that anyone is using them.

.github/workflows/benchmark_commits.yml
```diff
@@ -161,15 +163,32 @@ jobs:
...
         shell: bash -el {0}
         run: |
+          echo "Using S3 bucket: ${ARCTICDB_REAL_S3_BUCKET}"
           ./tooling_venv/bin/python build_tooling/transform_asv_results.py --mode save ${{ github.ref != 'refs/heads/master' && format('--arcticdb_library {0}_asv_results', github.ref_name) || ''}}
 
           ./tooling_venv/bin/python build_tooling/transform_asv_results.py --mode analyze \
           ${{ github.ref != 'refs/heads/master' && format('--arcticdb_library {0}_asv_results', github.ref_name) || ''}} \
           --hash $(git rev-parse --short=8 ${{inputs.commit}})
+        env:
+          ARCTICDB_REAL_S3_BUCKET: ${{ env.ASV_RESULTS_BUCKET }}
```
IvoDD, line 191: Nit: Maybe it would be a bit easier if the `transform_asv_results.py` starts using `ASV_RESULTS_BUCKET`, so we can't easily forget the env override.

python/benchmarks/basic_functions.py
```diff
@@ -8,264 +8,335 @@
...
 
-        self.lib = self.ac[get_prewritten_lib_name(rows)]
-        self.fresh_lib = self.get_fresh_lib()
-
-    def get_fresh_lib(self):
-        self.ac.delete_library("fresh_lib")
-        return self.ac.create_library("fresh_lib")
+    def teardown(self, *args):
+        qs.disable()
+        qs.reset_stats()
```
IvoDD, line 90: Just to confirm, are these no-op for lmdb?

python/benchmarks/basic_functions.py
```diff
@@ -8,264 +8,335 @@
...
+    def setup(self, libs_for_storage, rows, num_symbols, storage):
+        self.lib = libs_for_storage[storage][_lib_name(rows)]
+        if self.lib is None:
+            raise SkipNotImplemented
+
+        self.df = generate_pseudo_random_dataframe(rows)
+        self.update_df = generate_pseudo_random_dataframe(rows // 2)
+
+    def time_update_batch(self, libs_for_storage, rows, num_symbols, storage):
         payloads = [UpdatePayload(f"{sym}_sym", self.update_df) for sym in range(num_symbols)]
```
IvoDD, line 306: I know this was like this before but `update_batch` performance can vary between first and follow up reruns.

I.e. first run it would need to read a single `rows` size segment and write two `rows//2` segments. On follow up runs it would only need to read one `rows//2` segment and write one `rows//2` segment. I.e. I wouldn't be surprised if there is an upwards of 2x difference between the two.

What do you think about moving these to `modification_functions.py`?

python/benchmarks/environment_setup.py
```diff
@@ -131,795 +91,29 @@ def create_libraries(storage: Storage, library_names: List[str]) -> List[Optiona
...
+def create_library(storage: Storage, library_options: LibraryOptions = LibraryOptions()) -> Optional[Library]:
     """
     Create a library for a given storage test, for use in ASV benchmarking runs.
 
     Returns None if the storage is not enabled.
 
     See create_libraries docstring for notes on running locally.
     """
+    if not is_storage_enabled(storage):
+        return None
```
IvoDD, line 108: Nit: Unneeded if

python/benchmarks/local_query_builder.py
IvoDD, line 1: Not needed to change but it would have been slightly easier if this was `git mv`-ed

python/benchmarks/modification_functions.py
```diff
@@ -25,174 +28,121 @@ def get_time_at_fraction_of_df(fraction, rows):
...
+    def setup(self, libs_for_storage, rows_and_cols, storage):
+        self.lib = libs_for_storage[storage]
+        if self.lib is None:
+            raise SkipNotImplemented
+
+        rows, cols = rows_and_cols
+
+        self.sym = _sym_name(rows, cols)
+        assert self.lib.has_symbol(self.sym)
+        self.lib._nvs.restore_version(self.sym, 0)
```
IvoDD, line 83: Nice idea!

python/benchmarks/modification_functions.py
```diff
@@ -25,174 +28,121 @@ def get_time_at_fraction_of_df(fraction, rows):
...
+        assert self.lib.get_description(self.sym).row_count == rows
+
+        self.df_update_single = generate_random_floats_dataframe_with_index(
+            1, cols, end_timestamp=get_time_at_fraction_of_df(0.5, rows)
+        )
+        self.df_update_half = generate_random_floats_dataframe_with_index(
+            rows, cols, end_timestamp=get_time_at_fraction_of_df(0.75, rows)
```
IvoDD, line 90: This previously was `rows//2` so it splits the dataframe in 3 chunks

python/benchmarks/modification_functions.py
```diff
@@ -25,174 +28,121 @@ def get_time_at_fraction_of_df(fraction, rows):
...
 
-    def time_update_upsert(self, lad: LargeAppendDataModify, rows):
-        self.lib.update("sym", self.df_update_upsert, upsert=True)
+    def time_update_half(self, *args):
+        self.lib.update(self.sym, self.df_update_half)
 
-    def time_append_single(self, lad: LargeAppendDataModify, rows):
-        self.lib.append("sym", self.df_append_single)
+    def time_update_upsert(self, *args):
+        self.lib.update(self.sym, self.df_update_upsert, upsert=True)
```
IvoDD, line 109: Nit: I don't like the `upsert=True` here. In arcticdb it means `write if symbol is deleted` where most other databases treat `upsert` as `insert rows for unmatched index values`, which update already does.

I know this was like this before, but what do you think about removing the `upsert` kwarg and maybe rename the benchmark to `time_update_outside_date_range` or similar?

IvoDD (approved):

.github/workflows/benchmark_commits.yml
```diff
@@ -100,22 +108,14 @@ jobs:
       - name: Configure what suite or tests to execute
         shell: bash -el {0}
         run: | 
+            SUITE='^.*'
```
vasil-pashov, line 111: Not for now, but I think we should get rid of this SUITE env as.

python/benchmarks/basic_functions.py
```diff
@@ -8,264 +8,335 @@
...
+
+    def setup(self, libs_for_storage, rows, use_query_stats, storage):
+        self.lib = libs_for_storage[storage][_lib_name(rows)]
+        if self.lib is None:
+            raise SkipNotImplemented
+
+        if use_query_stats:
+            if storage != Storage.AMAZON:
+                raise SkipNotImplemented("Query stats only supported against S3")
+            qs.enable()
```
vasil-pashov, line 84: I don't see the query stats being used. Is this added so that we track how they affect performance? If so maybe it would be better if they get their own benchmark, or maybe we can at least have a comment about the intent

python/benchmarks/basic_functions.py
```diff
@@ -8,264 +8,335 @@
...
-            if "prewritten" in lib:
-                continue
-            self.ac.delete_library(lib)
-
-        del self.ac
-
-    def setup(self, rows):
-        self.ac = Arctic(BasicFunctions.CONNECTION_STRING)
+        libs_for_storage = dict()
+        library_names = [_lib_name(rows) for rows in BasicFunctions.num_rows]
```
vasil-pashov, line 58: Is there a particular reason to create different libraries rather than having multiple symbols in the same library

poodlewars: Either way would work. I think grouping by library makes it easier to say "run over all symbols with N rows", which would otherwise be done by pattern matching on symbol names in a single library.

.github/workflows/benchmark_commits.yml
poodlewars: This is quite a chore actually, because the scripts uses `real_s3_from_environment_variables`. I think it's not worth it, so I'll leave this.

PR #2847, opened 2026-01-13
Util 10956701718: update with exponential defrag

alexowens90:
#### Reference Issues/PRs
[10956701718](https://man312219.monday.com/boards/7852509418/views/188354541/pulses/10956701718)

Adds a (hidden) utility method to defragment while updating. Limitations:

- Uses our existing Python API, and so performs many unnecessary IOs compared to an equivalent C++ implementation
- Requires data to be a timeseries
- Expects symbols to be append-only
- Expects the process calling this function to be the only one modifying these symbols
- Must use the same `factor` argument (see below for what this does) every time to avoid unexpected results
- Expects repeated calls with the same index values to feature the same column values

The defragmentation is exponential based on a provided `factor`. E.g. if the library is configured with 1k rows per segment, and the factor is set to 10, then the defrag levels will be `[1000, 100, 10]`. When appending new data, the existing data will be checked for row-slices that are more fragmented than the top level, and if combining some row slices will be over any of these levels. If so, it will tail the existing data to get these fragmented slices, combine them with the new data, and then update.

#### Any other comments?
Not suitable for general usage, hence the leading underscore on the entrypoint function name

### DIFF
--- a/python/arcticdb/util/append_and_defrag.py
+++ b/python/arcticdb/util/append_and_defrag.py
@@ -110,21 +110,20 @@ def _append_and_defrag_idempotent(
         df = pd.DataFrame({"col": np.arange(rows_per_df)}, index=rows_per_df * [ts])
         _append_and_defrag_idempotent(lib, [(sym, df)], factor)
     The row-slice structure after each iteration will be as follows:
-    0: 0-8 as we had 4 rows, and were appending 4 more. This triggers a defrag as 4 + 4 = 8 is >= the smallest defrag
-           threshold of 4
-    1:  0-8, 8-12 no compaction happens this iteration
+    0:  0-4, 4-8 no compaction happens this iteration
+    1:  0-4, 4-8, 8-12 no compaction happens this iteration
     2:  0-16 defrag triggered as we have hit the next defrag threshold of 16
     3:  0-16, 16-20 no compaction happens this iteration
-    4:  0-16, 16-24 same reasoning as iteration 0
-    5:  0-16, 16-24, 24-28 same reasoning as iteration 1
+    4:  0-16, 16-20, 20-24 no compaction happens this iteration
+    5:  0-16, 16-20, 20-24, 24-28 no compaction happens this iteration
     6:  0-16, 16-32 same reasoning as iteration 2
-    7:  0-16, 16-32, 32-36 same reasoning as iteration 3
-    8:  0-16, 16-32, 32-40 same reasoning as iteration 4
-    9:  0-16, 16-32, 32-40, 40-44 same reasoning as iteration 5
-    10: 0-16, 16-32, 32-48 same reasoning as iteration 6
-    11: 0-16, 16-32, 32-48, 48-52 same reasoning as iteration 3
-    12: 0-16, 16-32, 32-48, 48-56 same reasoning as iteration 4
-    13: 0-16, 16-32, 32-48, 48-56, 56-60 same reasoning as iteration 5
+    7:  0-16, 16-32, 32-36 no compaction happens this iteration
+    8:  0-16, 16-32, 32-36, 36-40 no compaction happens this iteration
+    9:  0-16, 16-32, 32-36, 36-40, 40-44 no compaction happens this iteration
+    10: 0-16, 16-32, 32-48 same reasoning as iteration 2
+    11: 0-16, 16-32, 32-48, 48-52 no compaction happens this iteration
+    12: 0-16, 16-32, 32-48, 48-52, 52-56 no compaction happens this iteration
+    13: 0-16, 16-32, 32-48, 48-52, 52-56, 56-60 no compaction happens this iteration
     14: 0-64 defrag triggered as we have hit the next defrag threshold of 64
     """
 
@@ -147,14 +146,10 @@ def _append_and_defrag_idempotent(
                     self.append_df = None
                 else:
                     # Drop columns we don't need as we're going to do a filter
-                    index.drop(
-                        columns=index.columns.difference(["start_row", "end_row", "start_col", "end_col"]), inplace=True
-                    )
+                    index = index[["start_row", "end_row", "start_col"]]
                     # Drop everything except the first column slice
                     first_start_col = index["start_col"][0]
                     index = index[index["start_col"] == first_start_col]
-                    # Lazy way to get unique values when there is column slicing (although main use case has dynamic schema enabled anyway)
-                    # Index keys will not be very long so not worried about efficiency here
                     self.start_indexes = index.index.to_list()
                     self.start_rows = index["start_row"].to_list()
                     self.end_rows = index["end_row"].to_list()
--- a/python/tests/unit/arcticdb/test_append_and_defrag.py
+++ b/python/tests/unit/arcticdb/test_append_and_defrag.py
@@ -364,11 +364,11 @@ def test_io_count_many_iterations(s3_library_factory):
     qs.reset_stats()
     # Iterations:
     # 0:    read 0 data keys, write 1 data key 0-4
-    # 1:    read 1 data keys, write 1 data key 0-8
-    # 2:    read 0 data keys, write 1 data key 0-8, 8-12
-    # 3:    read 2 data keys, write 1 data key 0-16
-    # 4-14: repeat this pattern, so total read 10 data keys, written 15 data keys 0-16, 16-32, 32-48, 48-56, 56-60
-    # 15:   read 5 data keys, write 1 data key 0-64
+    # 1:    read 0 data keys, write 1 data key 0-4, 4-8
+    # 2:    read 0 data keys, write 1 data key 0-4, 4-8, 8-12
+    # 3:    read 3 data keys, write 1 data key 0-16
+    # 4-14: repeat this pattern, so total read 9 data keys, written 15 data keys 0-16, 16-32, 32-48, 48-52, 52-56, 56-60
+    # 15:   read 6 data keys, write 1 data key 0-64
     # Cumulative total read 15 and written 16 data keys
     # 15-31: repeat this pattern for cumulative total read 30 and written 32 data keys
     assert_data_key_ios(stats, 30, 32)


python/arcticdb/util/update_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+            # end_rows[-1] - start_rows[idx] == sum(row_counts[idx:])
+            if (end_rows[-1] - start_rows[idx]) + new_df_row_count >= levels[0]:
+                return start_indexes[idx]
+            else:
+                return _generate_date_to_read_from(
+                    start_indexes[idx:], start_rows[idx:], end_rows[idx:], new_df_row_count, levels[1:], threshold
+                )
+
+
+def _update_and_defrag(
```
IvoDD: I think I'd name it `_append_and_defrag_if_valid` or similar. This will be a no-op if used with data in the middle of the segment. `update` kind of implies it will work fully with updates.

It would also be nice to include a docstring with a description, the limitations from the PR description. 

Also an example here (or in PR description) would be very useful for other reviewers/users.

python/arcticdb/util/append_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+            self.existing_data_tail = None
+            self.start_indexes = []
+            self.start_rows = []
+            self.end_rows = []
+            try:
+                index = lib._nvs.read_index(symbol)
+                # To ensure idempotency, we must not append the same data twice
+                start_index_of_new_df = append_df.index[0]
+                end_index_of_existing_df = index["end_index"][-1] - pd.Timedelta(1, unit="ns")
+                if start_index_of_new_df <= end_index_of_existing_df:
```
IvoDD, line 145: Just to confirm: 
- Users will not use timezones? I don't think `read_index` is doing index timezone normalization.
- Users will not want to append duplicate timestamps in separate operations

python/arcticdb/util/update_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+                index = lib._nvs.read_index(symbol)
+                # To ensure idempotency, we must not append the same data twice
+                start_index_of_new_df = append_df.index[0]
+                end_index_of_existing_df = index["end_index"][-1] - pd.Timedelta(1, unit="ns")
+                if start_index_of_new_df <= end_index_of_existing_df:
+                    self.append_df = None
+                else:
+                    # Lazy way to get unique values when there is column slicing (although main use case has dynamic schema enabled anyway)
+                    # Index keys will not be very long so not worried about efficiency here
+                    self.start_indexes = index.index.sort_values().drop_duplicates().to_list()
```
IvoDD: What if entire segment chunks contain the same datetime? E.g. index looks like:

start_index | end_index | start_row | end_row
----------- | ----------- | ----------- | ----------
0 | 1 | 0 | 100
0 | 1 | 100 | 200
0 | 1 | 200 | 300
1 | 2 | 300 | 400

IvoDD: In general not a big fan of `sort_values` + `drop_duplicates` as it is not performant. Can we not just filter on `start_col == 0`?

python/arcticdb/util/append_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+                del self.existing_data_tail
+                return res
+            elif self.append_df is not None:
+                return self.append_df
+            elif self.existing_data_tail is not None:
+                raise ArcticNativeException("Operation already applied so tail should not have been read")
+            else:
+                return None
+
+    check(0 < threshold <= 1, f"threshold must be in (0, 1], received {threshold}")
```
IvoDD, line 178: Nit: In the same spirit we can check `factor > 0`

python/arcticdb/util/append_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+    target_rows_per_slice = lib._nvs.lib_cfg().lib_desc.version.write_options.segment_row_size
+    if target_rows_per_slice == 0:
+        target_rows_per_slice = 100_000
+    levels = _generate_levels(target_rows_per_slice, factor)
+    symbol_infos = {item[0]: SymbolInfo(item[0], item[1], lib) for item in items}
```
IvoDD, line 183: We might need to read the index keys with multiprocessing? I imagine this can be a pretty big bottleneck

python/arcticdb/util/append_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+    read_requests = []
+    for symbol, symbol_info in symbol_infos.items():
+        ts_to_read_from = _generate_date_to_read_from(
+            symbol_info.start_indexes,
+            symbol_info.start_rows,
+            symbol_info.end_rows,
+            symbol_info.rows_to_append(),
```
IvoDD, line 190: Nit: We will needlessly iterate over index key for all noops where `rows_to_append()==0` because `append_df==None`

python/arcticdb/util/append_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        if ts_to_read_from is not None:
+            read_requests.append(ReadRequest(symbol, date_range=(ts_to_read_from, None)))
+    vits = lib.read_batch(read_requests)
+    for vit in vits:
+        check(isinstance(vit, VersionedItem), f"Unexpected DataError when tailing existing data: {vit}")
+        symbol_infos[vit.symbol].existing_data_tail = vit.data
+    update_payloads = []
+    for symbol, symbol_info in symbol_infos.items():
+        update_df = symbol_info.update_df()
```
IvoDD, line 202: If we decide to multiprocess reading of the index keys, we should also multiprocess `concat`-ing the dataframes here. It should be the next thing to be a bottleneck if done sequentially rather than in parallel.

python/tests/unit/arcticdb/test_append_and_defrag.py
```diff
@@ -0,0 +1,393 @@
...
+    _update_and_defrag(lib, [(sym, df)], factor, 1)
+    assert len(lib._nvs.read_index(sym)) == 2
+    assert_frame_equal(expected_result, lib.read(sym).data)
+    # We then reslice again into 0-64 and 64-72
+    ts += pd.Timedelta(1, unit="days")
+    df = pd.DataFrame({"col": np.arange(rows_per_df)}, index=rows_per_df * [ts])
+    expected_result = pd.concat([expected_result, df])
+    _update_and_defrag(lib, [(sym, df)], factor, 1)
+    assert len(lib._nvs.read_index(sym)) == 2
+    assert_frame_equal(expected_result, lib.read(sym).data)
```
IvoDD, line 126: Nit: I would find these tests a bit easier to read/verify if the important numbers were more densely put together. I.e. something like:
```
runs_of_defrag = [
    {"rows_to_add": 3, "expected_data_keys": 1}, # After adding 3 rows we expect a single slice 0-3
    {"rows_to_add": 3, "expected_data_keys": 1}, # After adding 3 more rows we expect to combine the slices to 0-6
    {"rows_to_add": 3, "expected_data_keys": 2}, # After adding 3 rows we expect a two slices 0-6 and 6-9
    # Etc ...
    {"rows_to_add": 3, "expected_data_keys": 5, "repeat": 15} # After doing 15 more appends we expect slices ....
]

for run in runs_of_defrag:
    repeat = run["repeat"] if "repeat" in run else 1
    for i in range(repeat):
        ts += pd.Timedelta(1, unit="days")
        df = pd.DataFrame({"col": np.arange(run["rows_to_add")}, index=run["rows_to_add"] * [ts])
        expected_result = pd.concat([expected_result, df])
        _update_and_defrag(lib, [(sym, df)], factor, 1)
        assert_frame_equal(expected_result, lib.read(sym).data)
    assert len(lib.read_index(sym), run["expected_data_keys"])
```

IvoDD: At this point after this is already written it probably doesn't make sense to change this.

python/tests/unit/arcticdb/test_append_and_defrag.py
```diff
@@ -0,0 +1,393 @@
...
+        expected_0 = pd.concat([expected_0, df_0])
+        expected_1 = pd.concat([expected_1, df_1])
+        _update_and_defrag(lib, [(sym_0, df_0), (sym_1, df_1)], factor, threshold)
+        if rng.random() <= repeat_probability:
+            _update_and_defrag(lib, [(sym_0, df_0), (sym_1, df_1)], factor, threshold)
+        assert_frame_equal(expected_0, lib.read(sym_0).data)
+        assert_frame_equal(expected_1, lib.read(sym_1).data)
+        ts += pd.Timedelta(1, unit="days")
+        # It is impossible to assert the exact number of row slices we will have at this point, but we can at least
+        # check that the first n blocks are all compacted to approximately rows_per_segment
```
IvoDD, line 409: It would be nice to put some rough assertions on the number of IOs we end up doing over the entire 1000 updates, to verify the defrag properties are as we would expect.

I.e. I would expect the total number of data key writes to be just very slightly over 2*1000 (we will have more than 1 data key write only when row slicing is required)

I would also expect the number of data key reads to be roughly:
for sym_1 we on average write 2 rows. So we would need to read around 8/2-1 chunks of size <8 to produce a new 8 chunk size once in 8/2 runs.
Similar for chunks between 8 and 40. I would expect to read around 4 chunks once every 40/2 runs. Etc.

So the expected number of reads per run for sym_1 would be `3 * 1/4 + 4 * 1/20 + 4 * 1/100 + 4 * 1/500`
Similarly for sym_2 I would expect roughly `1 * 1/2 + 4 * 1/8 + 4 * 1/40 + 4 * 1/200`.
This all sums up to an expectation of `2.118` reads for both symbols
So I guess asserts like:
```
assert 2000 <= num_data_key_write_ios <= 2010
assert 2118*0.7 <= num_data_key_read_ios <= 2118*1.3
```

python/arcticdb/util/append_and_defrag.py
alexowens90: Need to check

1. If timestamps in index segments read like this are UTC or tz naive
2. If Pandas allows comparing tz naive to tz aware Timestamp objects

python/arcticdb/util/append_and_defrag.py
alexowens90: >1

python/arcticdb/util/update_and_defrag.py
```diff
@@ -0,0 +1,128 @@
...
+    new_df_row_count: int,
+    levels: List[int],
+    threshold: float,
+) -> Optional[pd.Timestamp]:
+    check(len(start_rows) == len(end_rows), f"len(start_rows) != len(end_rows): {len(start_rows)} != {len(end_rows)}")
+    check(
+        len(start_rows) == len(start_indexes),
+        f"len(start_rows) != len(start_indexes): {len(start_rows)} != {len(start_indexes)}",
+    )
+    if not len(levels) or not len(start_rows):
```
alexowens90:
```suggestion
    if not len(levels) or not len(start_rows) or new_df_row_count == 0:
```

python/arcticdb/util/append_and_defrag.py
alexowens90: Checked, `read_index` returns timezone-naive timestamps, but they represent UTC. So updated to first localize to UTC, then to the input dataframe's timezone so that comparisons are valid.
Discussed in person, the use case does not involve appending duplicate timestamps in separate operations.

python/arcticdb/util/update_and_defrag.py
alexowens90: "What if entire segment chunks contain the same datetime?"
This is safe, as we will find the first row-slice that is sufficiently fragmented to include in a defrag, which will pick up the correct `start_index` value

python/arcticdb/util/append_and_defrag.py
alexowens90: Discussed offline, will not do this for the first implementation, will add later if needed

python/arcticdb/util/append_and_defrag.py
alexowens90: See https://github.com/man-group/ArcticDB/pull/2847#discussion_r2690339129

python/arcticdb/util/append_and_defrag.py
alexowens90: Added check from here https://github.com/man-group/ArcticDB/pull/2847#discussion_r2689703699

python/tests/unit/arcticdb/test_append_and_defrag.py
alexowens90: Added another test for IO counts with more calls to the function

python/arcticdb/util/append_and_defrag.py
```diff
@@ -68,15 +139,25 @@ def __init__(self, symbol: str, append_df: pd.DataFrame, lib: Library):
...
+                end_index_of_existing_df = index["end_index"][-1].tz_localize("UTC").tz_localize(
+                    start_index_of_new_df.tz
+                ) - pd.Timedelta(1, unit="ns")
                 if start_index_of_new_df <= end_index_of_existing_df:
                     self.append_df = None
                 else:
+                    # Drop columns we don't need as we're going to do a filter
+                    index.drop(
+                        columns=index.columns.difference(["start_row", "end_row", "start_col", "end_col"]), inplace=True
+                    )
```
IvoDD: Nit: Do we need `end_col`?
Also I think it's more readable to `index = index[["start_row", "end_row", "start_col"]]`

python/arcticdb/util/append_and_defrag.py
```diff
@@ -68,15 +139,25 @@ def __init__(self, symbol: str, append_df: pd.DataFrame, lib: Library):
...
+                    # Drop everything except the first column slice
+                    first_start_col = index["start_col"][0]
+                    index = index[index["start_col"] == first_start_col]
                     # Lazy way to get unique values when there is column slicing (although main use case has dynamic schema enabled anyway)
```
IvoDD: Nit: I think we should remove this stale comment, we're no longer getting the "unique" values

python/arcticdb/util/append_and_defrag.py
```diff
@@ -51,12 +51,83 @@ def _generate_date_to_read_from(
...
+    _append_and_defrag_idempotent(lib, [(sym, df)], factor)
+    for _ in range(15):
+        ts += pd.Timedelta(1, unit="days")
+        df = pd.DataFrame({"col": np.arange(rows_per_df)}, index=rows_per_df * [ts])
+        _append_and_defrag_idempotent(lib, [(sym, df)], factor)
+    The row-slice structure after each iteration will be as follows:
+    0: 0-8 as we had 4 rows, and were appending 4 more. This triggers a defrag as 4 + 4 = 8 is >= the smallest defrag
+           threshold of 4
+    1:  0-8, 8-12 no compaction happens this iteration
+    2:  0-16 defrag triggered as we have hit the next defrag threshold of 16
```
IvoDD, line 115: Are we going to do the defrag on step 0 here? `0-4` and `4-8` seems like a perfectly fine with `levels = [64, 16, 4]` ?

I think this entire exaple assumes `levels = [64, 16, 8, 4]`?

Also I find comments like "same reasoning as iteration 4" confusing when iteration 4 says "same reasoning as iteration 0". Maybe just always reference the leaf reasoning directly?

PR #2894, opened 2026-02-03
Improve version map cache validation for as_of queries by version and timestamp

maxim-morozov:
## Summary
- Improve cache validation for `as_of` queries to correctly detect when a reload is needed
- Ensure queries requesting versions beyond what's cached trigger a reload

## Problem
The version map cache could return stale results or incorrectly assume data was cached in two scenarios:

  1. **Version-based queries (DOWNTO):** When requesting a specific version ID that exists in storage but wasn't loaded yet, the cache could incorrectly report a hit.

  **Example:** Client B loads all versions (latest is v2 at ts=200). Client A writes v3 at ts=300. Client B queries `as_of=200`. The old code would incorrectly return from cache, missing v3.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/test/test_version_map.cpp
+++ b/cpp/arcticdb/version/test/test_version_map.cpp
@@ -1466,7 +1466,10 @@ TEST(VersionMap, HasCachedEntryDowntoReloadsWhenVersionNotLoaded) {
     // from has_cached_entry, causing check_reload to actually reload from storage).
     // Note: DOWNTO 5 will only load from latest (v5) down to v5, so just v5.
     entry = version_map_b->check_reload(
-            store, id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(5)}, __FUNCTION__
+            store,
+            id,
+            LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(5)},
+            __FUNCTION__
     );
     // The entry should now include version 5
     auto opt_v5 = entry->get_first_index(true);
@@ -1609,11 +1612,14 @@ TEST(VersionMap, FromTimeCacheInvalidationOnNewVersion) {
     // Client B reloads with FROM_TIME=250, which loads from newest (300) back to earliest <= 250 (200)
     // This loads v2 (300) and v1 (200), but not v0 (100)
     entry_b = version_map_b->check_reload(
-            store, id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(250)}, __FUNCTION__
+            store,
+            id,
+            LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(250)},
+            __FUNCTION__
     );
-    ASSERT_EQ(entry_b->get_indexes(true).size(), 2);  // v2 and v1
-    EXPECT_EQ(entry_b->load_progress_.latest_loaded_timestamp_, 300);  // newest loaded is v2
-    EXPECT_EQ(entry_b->load_progress_.earliest_loaded_timestamp_, 200);  // earliest loaded is v1
+    ASSERT_EQ(entry_b->get_indexes(true).size(), 2);                    // v2 and v1
+    EXPECT_EQ(entry_b->load_progress_.latest_loaded_timestamp_, 300);   // newest loaded is v2
+    EXPECT_EQ(entry_b->load_progress_.earliest_loaded_timestamp_, 200); // earliest loaded is v1
 
     // Now Client B's cache should be valid for timestamp 250 (100 <= 250 <= 300)
     // Wait, earliest is now 200, so 200 <= 250 <= 300 is still valid
--- a/cpp/arcticdb/version/test/test_version_map.cpp
+++ b/cpp/arcticdb/version/test/test_version_map.cpp
@@ -1392,6 +1392,100 @@ TEST(VersionMap, TombstoneAllFromEntry) {
     ASSERT_EQ(version_id, 2);
 }
 
+// Test that has_cached_entry correctly returns false for DOWNTO requests
+// when the requested version hasn't been loaded, even if is_earliest_version_loaded is true.
+// This tests the fix for the case where:
+// 1. Client A loads all versions and caches them (is_earliest_version_loaded = true)
+// 2. Client B writes new versions to storage
+// 3. Client A's cache shouldn't claim to have versions it hasn't loaded
+TEST(VersionMap, HasCachedEntryDowntoReloadsWhenVersionNotLoaded) {
+    ScopedConfig sc("VersionMap.ReloadInterval", std::numeric_limits<int64_t>::max());
+    auto store = std::make_shared<InMemoryStore>();
+    StreamId id{"test"};
+
+    // Client A writes versions v0, v1, v2
+    auto version_map_a = std::make_shared<VersionMap>();
+    write_versions(store, version_map_a, id, 3);
+
+    // Client B creates a new version map and loads all versions
+    auto version_map_b = std::make_shared<VersionMap>();
+    auto entry = version_map_b->check_reload(
+            store, id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+
+    // Verify that all versions are loaded and is_earliest_version_loaded is true
+    ASSERT_TRUE(entry->load_progress_.is_earliest_version_loaded);
+    ASSERT_EQ(entry->load_progress_.oldest_loaded_index_version_, VersionId{0});
+    ASSERT_EQ(entry->get_indexes(true).size(), 3); // v0, v1, v2
+
+    // Verify has_cached_entry returns true for all existing versions
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(0)}
+    ));
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(1)}
+    ));
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(2)}
+    ));
+
+    // Client A writes new versions v3, v4, v5 that Client B doesn't know about
+    for (int i = 3; i < 6; ++i) {
+        auto key = atom_key_with_version(id, i, i);
+        version_map_a->write_version(store, key, std::nullopt);
+    }
+
+    // Client B's cache still has is_earliest_version_loaded = true, but it doesn't have v3, v4, v5
+    // has_cached_entry should return false for DOWNTO requests for these new versions
+    // because even though is_earliest_version_loaded is true, the specific versions aren't in the cache
+    EXPECT_FALSE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(3)}
+    ));
+    EXPECT_FALSE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(4)}
+    ));
+    EXPECT_FALSE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(5)}
+    ));
+
+    // DOWNTO with negative indices should also correctly reload
+    // At this point, Client B thinks latest is v2, so -1 = v2, -2 = v1, -3 = v0
+    // But in storage, latest is actually v5
+    // The cache can still serve requests that resolve to loaded versions
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(-1)}
+    )); // -1 resolves to v2 in cache, which is loaded
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(-2)}
+    )); // -2 resolves to v1 in cache, which is loaded
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(-3)}
+    )); // -3 resolves to v0 in cache, which is loaded
+
+    // Trigger a reload by requesting version 5 via DOWNTO (which should return false
+    // from has_cached_entry, causing check_reload to actually reload from storage).
+    // Note: DOWNTO 5 will only load from latest (v5) down to v5, so just v5.
+    entry = version_map_b->check_reload(
+            store, id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(5)}, __FUNCTION__
+    );
+    // The entry should now include version 5
+    auto opt_v5 = entry->get_first_index(true);
+    ASSERT_TRUE(opt_v5.first.has_value());
+    ASSERT_EQ(opt_v5.first->version_id(), 5);
+
+    // has_cached_entry should now return true for version 5
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(5)}
+    ));
+
+    // Now request ALL to load all versions. Since is_earliest_version_loaded is now false
+    // (we only loaded down to v5, not v0), has_cached_entry should return false and trigger a full reload.
+    entry = version_map_b->check_reload(
+            store, id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+    ASSERT_EQ(entry->get_indexes(true).size(), 6); // v0-v5
+}
+
 #define GTEST_COUT std::cerr << "[          ] [ INFO ]"
 
 TEST_F(VersionMapStore, StressTestWrite) {
--- a/cpp/arcticdb/version/test/test_version_map.cpp
+++ b/cpp/arcticdb/version/test/test_version_map.cpp
@@ -1486,6 +1486,142 @@ TEST(VersionMap, HasCachedEntryDowntoReloadsWhenVersionNotLoaded) {
     ASSERT_EQ(entry->get_indexes(true).size(), 6); // v0-v5
 }
 
+TEST(VersionMap, LatestLoadedTimestampTracking) {
+    // Test that latest_loaded_timestamp_ is correctly populated and used for FROM_TIME cache validation
+    ScopedConfig sc("VersionMap.ReloadInterval", std::numeric_limits<int64_t>::max());
+    auto store = std::make_shared<InMemoryStore>();
+    auto version_map = std::make_shared<VersionMap>();
+    StreamId id{"test"};
+
+    auto entry = version_map->check_reload(
+            store, id, LoadStrategy{LoadType::NOT_LOADED, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+
+    // Write versions with specific timestamps: v0 at ts=100, v1 at ts=200, v2 at ts=300
+    auto key0 = atom_key_with_version(id, 0, 100);
+    version_map->do_write(store, key0, entry);
+    write_symbol_ref(store, key0, std::nullopt, entry->head_.value());
+
+    auto key1 = atom_key_with_version(id, 1, 200);
+    version_map->do_write(store, key1, entry);
+    write_symbol_ref(store, key1, std::nullopt, entry->head_.value());
+
+    auto key2 = atom_key_with_version(id, 2, 300);
+    version_map->do_write(store, key2, entry);
+    write_symbol_ref(store, key2, std::nullopt, entry->head_.value());
+
+    // Create a fresh version_map to test caching behavior
+    version_map = std::make_shared<VersionMap>();
+
+    // Load only the latest version
+    entry = version_map->check_reload(
+            store, id, LoadStrategy{LoadType::LATEST, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+    ASSERT_EQ(entry->get_indexes(true).size(), 1);
+    // After loading LATEST, latest_loaded_timestamp_ should be 300 (newest), earliest should also be 300
+    EXPECT_EQ(entry->load_progress_.latest_loaded_timestamp_, 300);
+    EXPECT_EQ(entry->load_progress_.earliest_loaded_timestamp_, 300);
+
+    // FROM_TIME with timestamp 250 should be cached (200 <= 300 newest, but need to load more for earliest)
+    // Since earliest is 300 and we need earliest <= 200, cache should NOT be valid
+    EXPECT_FALSE(version_map->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(200)}
+    ));
+
+    // FROM_TIME with timestamp 350 should NOT be cached because 350 > 300 (newest loaded)
+    // This is the key test - we might have missed a version written at ts=320
+    EXPECT_FALSE(version_map->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(350)}
+    ));
+
+    // Load all versions
+    entry = version_map->check_reload(
+            store, id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+    ASSERT_EQ(entry->get_indexes(true).size(), 3);
+    // After loading ALL, latest_loaded_timestamp_ should be 300 (newest), earliest should be 100 (oldest)
+    EXPECT_EQ(entry->load_progress_.latest_loaded_timestamp_, 300);
+    EXPECT_EQ(entry->load_progress_.earliest_loaded_timestamp_, 100);
+
+    // Now FROM_TIME with timestamp 200 should be cached (100 <= 200 <= 300)
+    EXPECT_TRUE(version_map->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(200)}
+    ));
+
+    // FROM_TIME with timestamp 350 should still NOT be cached (350 > 300)
+    EXPECT_FALSE(version_map->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(350)}
+    ));
+
+    // FROM_TIME with timestamp 50 is STILL cached because we've loaded everything.
+    // Even though 50 < 100 (earliest), we can answer "no version at that timestamp" from cache.
+    EXPECT_TRUE(version_map->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(50)}
+    ));
+}
+
+TEST(VersionMap, FromTimeCacheInvalidationOnNewVersion) {
+    // Test that FROM_TIME cache correctly detects when new versions have been written
+    ScopedConfig sc("VersionMap.ReloadInterval", std::numeric_limits<int64_t>::max());
+    auto store = std::make_shared<InMemoryStore>();
+    StreamId id{"test"};
+
+    // Client A writes initial versions
+    auto version_map_a = std::make_shared<VersionMap>();
+    auto entry_a = version_map_a->check_reload(
+            store, id, LoadStrategy{LoadType::NOT_LOADED, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+
+    auto key0 = atom_key_with_version(id, 0, 100);
+    version_map_a->do_write(store, key0, entry_a);
+    write_symbol_ref(store, key0, std::nullopt, entry_a->head_.value());
+
+    auto key1 = atom_key_with_version(id, 1, 200);
+    version_map_a->do_write(store, key1, entry_a);
+    write_symbol_ref(store, key1, std::nullopt, entry_a->head_.value());
+
+    // Client B loads all versions
+    auto version_map_b = std::make_shared<VersionMap>();
+    auto entry_b = version_map_b->check_reload(
+            store, id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+    ASSERT_EQ(entry_b->get_indexes(true).size(), 2);
+    EXPECT_EQ(entry_b->load_progress_.latest_loaded_timestamp_, 200);
+    EXPECT_EQ(entry_b->load_progress_.earliest_loaded_timestamp_, 100);
+
+    // Client B's cache should be valid for timestamps between 100 and 200
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(150)}
+    ));
+
+    // Client A writes a new version at ts=300
+    auto key2 = atom_key_with_version(id, 2, 300);
+    version_map_a->do_write(store, key2, entry_a);
+    write_symbol_ref(store, key2, std::nullopt, entry_a->head_.value());
+
+    // Client B's cache should NOT be valid for timestamp 250 anymore
+    // because 250 > 200 (Client B's newest loaded timestamp)
+    // A version might have been written between 200 and 250
+    EXPECT_FALSE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(250)}
+    ));
+
+    // Client B reloads with FROM_TIME=250, which loads from newest (300) back to earliest <= 250 (200)
+    // This loads v2 (300) and v1 (200), but not v0 (100)
+    entry_b = version_map_b->check_reload(
+            store, id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(250)}, __FUNCTION__
+    );
+    ASSERT_EQ(entry_b->get_indexes(true).size(), 2);  // v2 and v1
+    EXPECT_EQ(entry_b->load_progress_.latest_loaded_timestamp_, 300);  // newest loaded is v2
+    EXPECT_EQ(entry_b->load_progress_.earliest_loaded_timestamp_, 200);  // earliest loaded is v1
+
+    // Now Client B's cache should be valid for timestamp 250 (100 <= 250 <= 300)
+    // Wait, earliest is now 200, so 200 <= 250 <= 300 is still valid
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(250)}
+    ));
+}
+
 #define GTEST_COUT std::cerr << "[          ] [ INFO ]"
 
 TEST_F(VersionMapStore, StressTestWrite) {
--- a/cpp/arcticdb/version/test/test_version_map.cpp
+++ b/cpp/arcticdb/version/test/test_version_map.cpp
@@ -489,7 +489,7 @@ TEST(VersionMap, FixRefKey) {
     ASSERT_TRUE(version_map->check_ref_key(store, id));
 
     auto key4 = key3;
-    EXPECT_THROW(
+    ASSERT_THROW(
             {
                 // We should raise if we try to write a non-increasing index key
                 version_map->write_version(store, key4, key3);
@@ -528,11 +528,11 @@ TEST(VersionMap, FixRefKeyTombstones) {
     auto key1 = atom_key_with_version(id, 0, 1696590624524585339);
     version_map->write_version(store, key1, std::nullopt);
     auto key2 = atom_key_with_version(id, 0, 1696590624387628801);
-    EXPECT_THROW({ version_map->write_version(store, key2, key1); }, NonIncreasingIndexVersionException);
+    ASSERT_THROW({ version_map->write_version(store, key2, key1); }, NonIncreasingIndexVersionException);
     auto key3 = atom_key_with_version(id, 0, 1696590624532320286);
-    EXPECT_THROW({ version_map->write_version(store, key3, key2); }, NonIncreasingIndexVersionException);
+    ASSERT_THROW({ version_map->write_version(store, key3, key2); }, NonIncreasingIndexVersionException);
     auto key4 = atom_key_with_version(id, 0, 1696590624554476875);
-    EXPECT_THROW({ version_map->write_version(store, key4, key3); }, NonIncreasingIndexVersionException);
+    ASSERT_THROW({ version_map->write_version(store, key4, key3); }, NonIncreasingIndexVersionException);
     auto key5 = atom_key_with_version(id, 1, 1696590624590123209);
     version_map->write_version(store, key5, key4);
     auto entry = version_map->check_reload(
@@ -629,7 +629,7 @@ TEST(VersionMap, RecoverDeleted) {
 
     deleted = version_map->find_deleted_version_keys(store, id);
     ASSERT_EQ(deleted.size(), 3);
-    EXPECT_THROW({ get_all_versions(store, version_map, id); }, std::runtime_error);
+    ASSERT_THROW({ get_all_versions(store, version_map, id); }, std::runtime_error);
     version_map->recover_deleted(store, id);
 
     std::vector<AtomKey> expected{key3, key2, key1};
@@ -753,7 +753,7 @@ TEST(VersionMap, FollowingVersionChain) {
         auto follow_result = std::make_shared<VersionMapEntry>();
 
         version_map->follow_version_chain(store, ref_entry, follow_result, load_strategy);
-        EXPECT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{should_load_to});
+        ASSERT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{should_load_to});
     };
 
     check_strategy_loads_to(
@@ -805,8 +805,8 @@ TEST(VersionMap, FollowingVersionChainWithCaching) {
     auto check_loads_versions = [&](LoadStrategy load_strategy, uint32_t should_load_any, uint32_t should_load_undeleted
                                 ) {
         auto loaded = version_map->check_reload(store, id, load_strategy, __FUNCTION__);
-        EXPECT_EQ(loaded->get_indexes(true).size(), should_load_any);
-        EXPECT_EQ(loaded->get_indexes(false).size(), should_load_undeleted);
+        ASSERT_EQ(loaded->get_indexes(true).size(), should_load_any);
+        ASSERT_EQ(loaded->get_indexes(false).size(), should_load_undeleted);
     };
 
     check_loads_versions(
@@ -834,7 +834,7 @@ TEST(VersionMap, FollowingVersionChainWithCaching) {
 
     // LATEST should still be cached, but the cached entry now needs to have no undeleted keys
     check_loads_versions(LoadStrategy{LoadType::LATEST, LoadObjective::INCLUDE_DELETED}, 2, 0);
-    EXPECT_FALSE(version_map->has_cached_entry(
+    ASSERT_FALSE(version_map->has_cached_entry(
             id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(-1)}
     ));
     // FROM_TIME UNDELETED_ONLY should no longer be cached even though we used the same request before because the
@@ -844,14 +844,14 @@ TEST(VersionMap, FollowingVersionChainWithCaching) {
     );
     // We have the full version chain loaded, so has_cached_entry should always return true (even when requesting
     // timestamp before earliest version)
-    EXPECT_TRUE(version_map->has_cached_entry(
+    ASSERT_TRUE(version_map->has_cached_entry(
             id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, static_cast<timestamp>(-1)}
     ));
-    EXPECT_TRUE(version_map->has_cached_entry(
+    ASSERT_TRUE(version_map->has_cached_entry(
             id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(-1)}
     ));
-    EXPECT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}));
-    EXPECT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::ALL, LoadObjective::UNDELETED_ONLY}));
+    ASSERT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}));
+    ASSERT_TRUE(version_map->has_cached_entry(id, LoadStrategy{LoadType::ALL, LoadObjective::UNDELETED_ONLY}));
 
     // We add a new undeleted key
     auto key4 = atom_key_with_version(id, 3, 5);
@@ -890,7 +890,7 @@ TEST(VersionMap, FollowingVersionChainEndEarlyOnTombstoneAll) {
         version_map->follow_version_chain(store, ref_entry, follow_result, load_strategy);
         // When loading with any of the specified load strategies with include_deleted=false we should end following the
         // version chain early at version 1 because that's when we encounter the TOMBSTONE_ALL.
-        EXPECT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{1});
+        ASSERT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{1});
     }
 
     for (auto load_strategy :
@@ -901,7 +901,7 @@ TEST(VersionMap, FollowingVersionChainEndEarlyOnTombstoneAll) {
         version_map->follow_version_chain(store, ref_entry, follow_result, load_strategy);
         // When loading with any of the specified load strategies with include_deleted=true we should continue to the
         // beginning at version 0 even though it was deleted.
-        EXPECT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{0});
+        ASSERT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{0});
     }
 }
 
@@ -932,7 +932,7 @@ TEST(VersionMap, FollowingVersionChainWithWriteAndPrunePrevious) {
           LoadStrategy{LoadType::LATEST, LoadObjective::INCLUDE_DELETED}}) {
         follow_result->clear();
         version_map->follow_version_chain(store, ref_entry, follow_result, load_strategy);
-        EXPECT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{3});
+        ASSERT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{3});
     }
 
     for (auto load_strategy : {
@@ -944,7 +944,7 @@ TEST(VersionMap, FollowingVersionChainWithWriteAndPrunePrevious) {
         version_map->follow_version_chain(store, ref_entry, follow_result, load_strategy);
         // When loading with any of the specified load strategies with include_deleted=false we should end following the
         // version chain early at version 2 because that's when we encounter the TOMBSTONE_ALL.
-        EXPECT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{2});
+        ASSERT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{2});
     }
 
     for (auto load_strategy :
@@ -955,7 +955,7 @@ TEST(VersionMap, FollowingVersionChainWithWriteAndPrunePrevious) {
         version_map->follow_version_chain(store, ref_entry, follow_result, load_strategy);
         // When loading with any of the specified load strategies with include_deleted=true we should continue to the
         // beginning at version 0 even though it was deleted.
-        EXPECT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{0});
+        ASSERT_EQ(follow_result->load_progress_.oldest_loaded_index_version_, VersionId{0});
     }
 }
 
@@ -981,7 +981,7 @@ TEST(VersionMap, HasCachedEntry) {
         // Load to_load inside the clean version map cache
         clean_version_map->check_reload(store, id, to_load, __FUNCTION__);
         // Check whether to_check_if_cached is being cached by to_load
-        EXPECT_EQ(clean_version_map->has_cached_entry(id, to_check_if_cached), expected_outcome);
+        ASSERT_EQ(clean_version_map->has_cached_entry(id, to_check_if_cached), expected_outcome);
     };
 
     auto check_all_caching = [&](const std::vector<LoadStrategy>& to_load,
@@ -1394,20 +1394,16 @@ TEST(VersionMap, TombstoneAllFromEntry) {
 
 // Test that has_cached_entry correctly returns false for DOWNTO requests
 // when the requested version hasn't been loaded, even if is_earliest_version_loaded is true.
-// This tests the fix for the case where:
-// 1. Client A loads all versions and caches them (is_earliest_version_loaded = true)
-// 2. Client B writes new versions to storage
-// 3. Client A's cache shouldn't claim to have versions it hasn't loaded
 TEST(VersionMap, HasCachedEntryDowntoReloadsWhenVersionNotLoaded) {
     ScopedConfig sc("VersionMap.ReloadInterval", std::numeric_limits<int64_t>::max());
     auto store = std::make_shared<InMemoryStore>();
     StreamId id{"test"};
 
-    // Client A writes versions v0, v1, v2
+    // Client a writes versions v0, v1
     auto version_map_a = std::make_shared<VersionMap>();
-    write_versions(store, version_map_a, id, 3);
+    write_versions(store, version_map_a, id, 2);
 
-    // Client B creates a new version map and loads all versions
+    // Client b creates a new version map and loads all versions
     auto version_map_b = std::make_shared<VersionMap>();
     auto entry = version_map_b->check_reload(
             store, id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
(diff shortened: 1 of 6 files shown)


poodlewars (commented):
I think the idea of this looks fine but of course worrying that the C++ tests are failing and crashing.

I still think there are some odd consistency questions where we could see stuff like,

```
lib.read().version -> 7 (uses cache)
lib.read(as_of=-1) -> 6 (uses cache)
lib.read(as_of=8).version -> 8 (bypasses cache)
```

The existing design also has performance limitations where whenever there is a cache miss we call `entry->clear()` and invalidate the entire cache.

IMO the right design for this (which is a bit of a rewrite) is to always go to storage to check the VERSION_REF key and follow the version chain, using a cache of VERSION keys in-memory to skip unnecessary VERSION IOs. I think lots of the complexity comes from the fact that the cache is implicitly storing state from the mutable VERSION_REF key - it could be a lot more obviously correct if it only stores immutable state from the VERSION keys.

This design would also let us get rid of the complicated state machine tracking "does the cache represent a load down to a certain time, does it include deleted versions" etc. We'd just walk the version chain each time, checking the in-memory cache for each version key before we go to storage for it.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -855,12 +863,20 @@ class VersionMapImpl {
...
         if (requested_version_id >= 0) {
-            if (entry.load_progress_.oldest_loaded_index_version_ <= static_cast<VersionId>(requested_version_id)) {
+            // For positive version IDs, we need to check two things:
+            // 1. We have loaded far enough back: oldest_loaded_index_version_ <= requested_version_id
+            // 2. The requested version is within our known range: requested_version_id <= latest_known_version_id
+            // Without check 2, we'd incorrectly claim to have versions that were written after our cache was populated
+            auto opt_latest = entry.get_first_index(true).first;
+            if (opt_latest.has_value() &&
+                static_cast<VersionId>(requested_version_id) <= opt_latest->version_id() &&
+                entry.load_progress_.oldest_loaded_index_version_ <= static_cast<VersionId>(requested_version_id)) {
```
IvoDD: Nit: Move the `static_cast<VersionId>(requested_version_id)` to a separate varaible.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -738,11 +738,22 @@ class VersionMapImpl {
...
+                true;
         const bool has_loaded_earliest_undeleted =
                 entry->tombstone_all_.has_value() &&
                 entry->load_progress_.oldest_loaded_index_version_ <= entry->tombstone_all_->version_id();
-        if (has_loaded_everything ||
-            (!requested_load_strategy.should_include_deleted() && has_loaded_earliest_undeleted)) {
+
+        if ((has_loaded_everything && has_loaded_requested_version_id && has_loaded_requested_timestamp) ||
+            (!requested_load_strategy.should_include_deleted() && has_loaded_earliest_undeleted &&
+             has_loaded_requested_timestamp)) {
```
IvoDD: This logic is a bit hard to reason about. E.g. now we're ending up with the following wrong cache skip:

1. We load entire version chain
2. We request a time before earliest version
   a. Previously we would imediately return `true` because `has_loaded_everything=true`. Later code would decide that such an early timestamp has no version and return NoSuchVersion correctly without going to storage.
   b. Now `loaded_as_far_as_timestamp=false` and we end up returning `false` even though we can know that nothing has changed before the earliest version.

I suppose this is one reason for tests to be failing.

I think a much cleaner and easier to follow logic would be:
```
const bool has_requested_version_or_timestamp_after_latest = something_like_what_you_did_for_loaded_as_far_as_version_id()

if (has_requested_version_or_timestamp_after_latest) return false;

// keep the rest of the code unchanged
if (has_loaded_everything_deleted || has_loaded_everything_undeletered) return true;

switch(requested_load_type) {
    // do per load type checks as before
}
```
This makes it clear that we always disregard the cache if we know a newer version was requested than latest available in cache.

IvoDD (commented):
+1 to Alex's point on the tradoffs. I think it is especially weird for:
```
lib.read().version
lib.read(as_of=now()).version
```
to be different

+1 to Alex's point on the correct final design (although note it will always end up reading vref even when we know we're the only reader/writer, but we can have a separate flag on such behavior)

Have not reviewed tests, since not passing

cpp/arcticdb/version/version_map.hpp
maxim-morozov: Yeah, it needs a bit refactoring. This is just draft to demo the approach

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -1392,6 +1392,242 @@ TEST(VersionMap, TombstoneAllFromEntry) {
     ASSERT_EQ(version_id, 2);
 }
 
+// Test that has_cached_entry correctly returns false for DOWNTO requests
+// when the requested version hasn't been loaded, even if is_earliest_version_loaded is true.
+// This tests the fix for the case where:
+// 1. Client A loads all versions and caches them (is_earliest_version_loaded = true)
+// 2. Client B writes new versions to storage
+// 3. Client A's cache shouldn't claim to have versions it hasn't loaded
```
poodlewars: The labels A and B don't match the test below

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -1392,6 +1392,242 @@ TEST(VersionMap, TombstoneAllFromEntry) {
...
+    ASSERT_TRUE(opt_v5.first.has_value());
+    ASSERT_EQ(opt_v5.first->version_id(), 5);
+
+    // has_cached_entry should now return true for version 5
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::DOWNTO, LoadObjective::INCLUDE_DELETED, static_cast<SignedVersionId>(5)}
+    ));
+
+    // Now request ALL to load all versions. Since is_earliest_version_loaded is now false
+    // (we only loaded down to v5, not v0), has_cached_entry should return false and trigger a full reload.
```
poodlewars: What happens if load v3 rather than v5? I assume we'll still set `is_earliest_version_loaded` to False, which is fine for now. Still worth having a test.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -1392,6 +1392,242 @@ TEST(VersionMap, TombstoneAllFromEntry) {
...
+// This tests the fix for the case where:
+// 1. Client A loads all versions and caches them (is_earliest_version_loaded = true)
+// 2. Client B writes new versions to storage
+// 3. Client A's cache shouldn't claim to have versions it hasn't loaded
+TEST(VersionMap, HasCachedEntryDowntoReloadsWhenVersionNotLoaded) {
+    ScopedConfig sc("VersionMap.ReloadInterval", std::numeric_limits<int64_t>::max());
+    auto store = std::make_shared<InMemoryStore>();
+    StreamId id{"test"};
+
+    // Client A writes versions v0, v1, v2
```
poodlewars: Can simplify the test by writing v0 and v1 first, then v2 and v3 later? So 4 versions rather than 6. Similar for the timestamp test too.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -1392,6 +1392,242 @@ TEST(VersionMap, TombstoneAllFromEntry) {
...
+    EXPECT_TRUE(version_map_b->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(150)}
+    ));
+
+    // Client A writes a new version at ts=300
+    auto key2 = atom_key_with_version(id, 2, 300);
+    version_map_a->do_write(store, key2, entry_a);
+    write_symbol_ref(store, key2, std::nullopt, entry_a->head_.value());
+
+    // Client B's cache should NOT be valid for timestamp 250 anymore
```
poodlewars: The "anymore" is confusing - it never had a cached entry for T=250

cpp/arcticdb/version/test/test_version_map.cpp
poodlewars, line 1: - Verify `latest_loaded_timestamp` is set correctly in the exhaustive `HasCachedEntry` test

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
...
+
+        # Client B should be able to read the new versions by triggering a reload
+        # Test reading by explicit version number
+        for i in range(3, 6):
+            result = lib_b.read(symbol, as_of=i)
+            assert result.version == i
+            assert_equal(result.data, dataframes[i])
+
+        # Test reading by negative index (relative to latest)
+        # -1 should be v5, -2 should be v4, etc.
```
poodlewars, line 3458: Verify what the negative index reads return before the reads by explicit version number above?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
...
+        for i in range(3, 6):
+            lib_a.write(symbol, dataframes[i])
+
+        # Verify Client A can read the new versions
+        for i in range(6):
+            result = lib_a.read(symbol, as_of=i)
+            assert result.version == i
+
+        # Client B should be able to read the new versions by triggering a reload
+        # Test reading by explicit version number
```
poodlewars: Worth testing different orderings, eg what happens if we read 4 then 3 then 5? What is the output of `list_versions` at each step?

Might as well check all 6 permuations.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
...
+        for i in range(2, 4):
+            with distinct_timestamps(lib_a) as ts:
+                lib_a.write(symbol, dataframes[i])
+            timestamps.append(ts)
+
+        # Client B should be able to read using a timestamp after the new versions
+        # This should trigger a reload since the timestamp is newer than Client B's cached newest
+        result = lib_b.read(symbol, as_of=timestamps[3].after)
+        assert result.version == 3
+        assert_equal(result.data, dataframes[3])
```
poodlewars: Check `list_versions` output here?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
...
+
+        # Client B should also be able to read the intermediate version
+        result = lib_b.read(symbol, as_of=timestamps[2].after)
+        assert result.version == 2
+        assert_equal(result.data, dataframes[2])
```
poodlewars: Might as well check it can also read the earliest two timestamps?

cpp/arcticdb/version/version_map.hpp
```diff
@@ -758,16 +774,6 @@ class VersionMapImpl {
             auto opt_latest = entry->get_first_index(requested_load_strategy.should_include_deleted()).first;
             return opt_latest.has_value();
         }
-        case LoadType::DOWNTO:
-            // We check whether the oldest loaded version is before or at the requested one
-            return loaded_as_far_as_version_id(*entry, requested_load_strategy.load_until_version_.value());
-        case LoadType::FROM_TIME: {
```
poodlewars, line 764: Worth an assertion that we don't fall in to these cases in the switch

cpp/arcticdb/version/version_map.hpp
```diff
@@ -884,6 +896,44 @@ class VersionMapImpl {
...
 
+    /**
+     * Whether entry contains as much of the version map as specified by load_param. Checks whether
+     * the loaded timestamps cover the requested timestamp range.
+     */
+    bool loaded_as_far_as_timestamp(
+            const VersionMapEntry& entry, timestamp requested_timestamp, bool include_deleted_versions,
+            bool has_loaded_earliest
+    ) const {
+        if (requested_timestamp < 0) {
```
poodlewars: I wasn't aware that we support negative timestamps, how does this behave? Are there any tests?

poodlewars (changes requested):

cpp/arcticdb/version/version_map.hpp
alexowens90: Why wouldn't we? I think Windows has some issues with pre-epoch timestamps, but I know Vasil added some tests for them (e.g. [here](https://github.com/man-group/ArcticDB/blob/258565ae8ec9e0bbe9fa84031c39efb51149f3d5/python/tests/unit/arcticdb/version_store/test_sort_merge.py#L404))

cpp/arcticdb/version/version_map.hpp
```diff
@@ -884,6 +896,44 @@ class VersionMapImpl {
...
+            // In case of the negative timestamp, we check if everything was already loaded
+            return entry.load_progress_.is_earliest_version_loaded;
```
alexowens90: This is wrong, a negative timestamp just represents a pre-epoch time.

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -1392,6 +1392,242 @@ TEST(VersionMap, TombstoneAllFromEntry) {
...
+            store, id, LoadStrategy{LoadType::ALL, LoadObjective::INCLUDE_DELETED}, __FUNCTION__
+    );
+
+    // Verify that all versions are loaded and is_earliest_version_loaded is true
+    ASSERT_TRUE(entry->load_progress_.is_earliest_version_loaded);
+    ASSERT_EQ(entry->load_progress_.oldest_loaded_index_version_, VersionId{0});
+    ASSERT_EQ(entry->get_indexes(true).size(), 3); // v0, v1, v2
+
+    // Verify has_cached_entry returns true for all existing versions
+    EXPECT_TRUE(version_map_b->has_cached_entry(
```
alexowens90: Is there a reason to use `EXPECT_*` instead of `ASSERT_*` everywhere?

cpp/arcticdb/version/test/test_version_map.cpp
```diff
@@ -1392,6 +1392,242 @@ TEST(VersionMap, TombstoneAllFromEntry) {
...
+    EXPECT_EQ(entry->load_progress_.earliest_loaded_timestamp_, 300);
+
+    // FROM_TIME with timestamp 250 should be cached (200 <= 300 newest, but need to load more for earliest)
+    // Since earliest is 300 and we need earliest <= 200, cache should NOT be valid
+    EXPECT_FALSE(version_map->has_cached_entry(
+            id, LoadStrategy{LoadType::FROM_TIME, LoadObjective::INCLUDE_DELETED, static_cast<timestamp>(200)}
+    ));
+
+    // FROM_TIME with timestamp 350 should NOT be cached because 350 > 300 (newest loaded)
+    // This is the key test - we might have missed a version written at ts=320
```
alexowens90: FWIW I don't think this bit was worth the effort. With our default cache timeout of 2s, there is a vanishingly small chance this code path will ever be triggered, and even if it is, with clock skew, the user wouldn't even know if they got the "right" result.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
...
+    Test that reading a specific version correctly reloads the version chain when new versions
+    have been written by another client.
+
+    This tests the fix for has_cached_entry with DOWNTO LoadType where the cache might incorrectly
+    claim to have all versions loaded (is_earliest_version_loaded=true) but the specifically
+    requested version hasn't actually been loaded because it was written after the cache was populated.
+    """
+    # Set a very long reload interval to ensure the fix is being tested,
+    # not just the reload interval expiring
+    with config_context("VersionMap.ReloadInterval", sys.maxsize):
```
alexowens90, line 3414: Weird use of `sys.maxsize`, could just use a billion or something

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
...
+        symbol = "test_reload_version"
+        dataframes = [sample_dataframe() for _ in range(6)]
+
+        # Client A writes initial versions v0, v1, v2
+        for i in range(3):
+            lib_a.write(symbol, dataframes[i])
+
+        # Client B reads all versions to populate its cache
+        # This sets is_earliest_version_loaded=true in its version map cache
+        all_versions_b = lib_b.list_versions(symbol)
```
alexowens90: Do we need this step? The reading `as_of` below is more how real users would interact

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
...
+            lib_a.write(symbol, dataframes[i])
+
+        # Verify Client A can read the new versions
+        for i in range(6):
+            result = lib_a.read(symbol, as_of=i)
+            assert result.version == i
+
+        # Client B should be able to read the new versions by triggering a reload
+        # Test reading by explicit version number
+        for i in range(3, 6):
```
alexowens90: This is slightly odd, as after reading version 3, it will no longer need to go back to storage for more version chain operations

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -3397,3 +3397,134 @@ def find_expected_version(first_to_check):
         lib.delete(symbol)
         for i in range(num_of_versions):
             assert_correct_dataframe(i, set(range(num_of_versions)))
+
+
+@pytest.mark.storage
+def test_read_specific_version_reloads_when_new_versions_written(basic_store_factory):
```
alexowens90, line 3403: If we wanted to get fancy with it, we could use query stats to confirm key read counts, although I'm sufficiently convinced already that everything works.

cpp/arcticdb/version/version_map.hpp
vasil-pashov: Yes, I also have added tests in the resampling. As @alexowens90 pointed out below they represent pre-epoch dates. The problem we had was that we were using [gmtime](https://en.cppreference.com/w/cpp/chrono/c/gmtime.html) which does not work with pre-epoch dates on Windows. I replaced that with boost's implementation.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -737,12 +737,28 @@ class VersionMapImpl {
             return false;
         }
 
-        const bool has_loaded_everything = entry->load_progress_.is_earliest_version_loaded;
+        const bool has_loaded_earliest_version_all = entry->load_progress_.is_earliest_version_loaded;
```
IvoDD: Nit: I'd rename to `has_loaded_earliest_inluding_deleted`

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -209,6 +209,7 @@ struct LoadProgress {
     VersionId oldest_loaded_undeleted_index_version_ = std::numeric_limits<VersionId>::max();
     timestamp earliest_loaded_timestamp_ = std::numeric_limits<timestamp>::max();
     timestamp earliest_loaded_undeleted_timestamp_ = std::numeric_limits<timestamp>::max();
+    timestamp latest_loaded_timestamp_ = std::numeric_limits<timestamp>::min();
```
IvoDD: It looks like we are not keeping this consistent after write.

When doing `write_version` or `write_version_and_prune_previous` we are adding a new version to the cache but we do not update the `latest_loaded_timestamp`

This can cause issues like:
```
1. Client A reads latest version of sym X written at time 10. Cache contains latest_loaded_timestamp = 10
2. Client A writes a new version at time 20. Cached latest_loaded_timestamp does not get updated
3. Client A reads as_of=timestamp 15. Cache incorrectly reloads
```

We should fix and add tests for this case.

IvoDD (changes requested):

cpp/arcticdb/version/version_map_entry.hpp
IvoDD: This is problematic for users who know they are the sole writer and reader and set the version cache timeout to infinity.

If we decide to fix this properly we'll need to change [this code](https://github.com/man-group/ArcticDB/blob/258565ae8ec9e0bbe9fa84031c39efb51149f3d5/cpp/arcticdb/version/version_map.hpp#L561-L565)

cpp/arcticdb/version/test/test_version_map.cpp
maxim-morozov: EXPECT continues the test, so nice to have all failures, but looks like we use ASSERT everywhere, so I will change it

cpp/arcticdb/version/test/test_version_map.cpp
maxim-morozov: Yes, good point. I will add a test for this.

cpp/arcticdb/version/test/test_version_map.cpp
maxim-morozov: Yeah, I can verify this, but in this test as we are always loading from the head of the chain, this will be always the head of the chain

cpp/arcticdb/version/test/test_version_map.cpp
maxim-morozov: Yeah, I agree with you. I will revert and handle only DOWN TO version cases. With the timestamp can be quite unreliable.

cpp/arcticdb/version/version_map_entry.hpp
maxim-morozov: It's a good catch, however, I reverted the change for handling of as_of timestamp because of timeskew it might be not very reliable. I removed the `latest_loaded_timestamp` in this PR.

cpp/arcticdb/version/version_map.hpp
maxim-morozov: I don't think it makes sense. The if statement above is very simple, don't see how this will fall through in the switch statement.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
maxim-morozov: List versions actually goes to storage always as it specifies the LoadStrategy::ALL, so this would not actually do any difference to test it.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
maxim-morozov: Yeah, I just changed to load as of 3 and it should load all versions up to version 3.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
maxim-morozov: The  list_versions will go and reload the whole chain, so it would not give any indication.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
maxim-morozov: Removed this test as decided not to implement this logic for timestamp as Alex suggested.

alexowens90 (approved):

cpp/arcticdb/version/version_map.hpp
```diff
@@ -854,18 +858,24 @@ class VersionMapImpl {
      * @return true if and only if entry already contains data at least as far back as load_param requests
      */
     bool loaded_as_far_as_version_id(const VersionMapEntry& entry, SignedVersionId requested_version_id) const {
```
G-D-Petrov: nit: not directly related to your change but I think that this can be refactored to:
``` c++
bool loaded_as_far_as_version_id(const VersionMapEntry& entry, SignedVersionId requested_version_id) const {
    auto opt_latest = entry.get_first_index(true).first;
    if (!opt_latest.has_value())
        return false;

    auto latest_version = opt_latest->version_id();
    std::optional<VersionId> resolved_version;

    if (requested_version_id >= 0) {
        // For positive version IDs, we need to check two things:
        // 1. We have loaded far enough back: oldest_loaded_index_version_ <= requested_version_id
        // 2. The requested version is within our known range: requested_version_id <= latest_known_version_id
        // Without check 2, we'd incorrectly claim to have versions that were written after our cache was populated
        // Check 2 is done here; check 1 is done below in the common path.
        if (static_cast<VersionId>(requested_version_id) <= latest_version)
            resolved_version = static_cast<VersionId>(requested_version_id);
    } else {
        resolved_version = get_version_id_negative_index(latest_version, requested_version_id);
    }

    if (resolved_version.has_value() &&
        entry.load_progress_.oldest_loaded_index_version_ <= *resolved_version) {
        ARCTICDB_DEBUG(
            log::version(),
            "Loaded as far as required value {}, have {} to {}",
            requested_version_id,
            entry.load_progress_.oldest_loaded_index_version_,
            latest_version
        );
        return true;
    }
    return false;
}
```


and maybe we should add a log for missing latest case e.g.:
``` c++
if (!opt_latest.has_value()) {
     ARCTICDB_DEBUG(
            log::version(),
            "Tried to load version...."
     }

    return false;
}
```

IvoDD: +1 Could be even a bit simpler (we can just always check the <= latest_version) because of the way we calculate the negative version:
```c++
bool loaded_as_far_as_version_id(const VersionMapEntry& entry, SignedVersionId requested_version_id) const {
    auto opt_latest = entry.get_first_index(true).first;
    if (!opt_latest.has_value())
        return false;

    auto latest_version = opt_latest->version_id();
    VersionId resolved_version;

    if (requested_version_id >= 0) {
        
        resolved_version = static_cast<VersionId>(requested_version_id);
    } else {
        resolved_version = get_version_id_negative_index(latest_version, requested_version_id);
    }
    // For the resolved version ID, we need to check two things:
    // 1. We have loaded far enough back: oldest_loaded_index_version_ <= resolved_version
    // 2. The requested version is within our known range: resolved_version <= latest_version
    // Without check 2, we'd incorrectly claim to have versions that were written after our cache was populated
    // Check 2 will always be true for negative requested version ids, because it is calculated by subtracting from latest_version.
    if (entry.load_progress_.oldest_loaded_index_version_ <= resolved_version &&
         resolved_version <= latest_version) {
        ARCTICDB_DEBUG(
            log::version(),
            "Loaded as far as required value {}, have {} to {}",
            requested_version_id,
            entry.load_progress_.oldest_loaded_index_version_,
            latest_version
        );
        return true;
    }
    return false;
}
```

IvoDD (approved):

cpp/arcticdb/version/test/test_version_map.cpp
poodlewars: Can you show me where this has been addressed?

poodlewars (approved):

poodlewars (changes requested):
Marking as request changes again until we understand the ASV change

IvoDD (approved):

cpp/arcticdb/version/test/test_version_map.cpp
Maxim-Morozov-mangroup: Time stamp logic was reverted to the previous logic, no need to address it as `latest_loaded_timestamp` removed.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+    """Get the count of VERSION key reads from query stats."""
+    return stats.get("storage_operations", {}).get("Memory_GetObject", {}).get("VERSION", {}).get("count", 0)
+
+
+def get_version_ref_reads(stats):
+    """Get the count of VERSION_REF key reads from query stats."""
+    return stats.get("storage_operations", {}).get("Memory_GetObject", {}).get("VERSION_REF", {}).get("count", 0)
+
+
+def test_read_latest_one_version(in_memory_store_factory, clear_query_stats):
```
poodlewars: From here down to `test_list_versions` should just be a single parametrized test

cpp/arcticdb/version/version_map.hpp
poodlewars: Disagree. I don't see the downside of a sanity check.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -845,42 +841,52 @@ class VersionMapImpl {
...
-                            entry.load_progress_.oldest_loaded_index_version_,
-                            opt_latest->version_id()
-                    );
-                    return true;
-                }
-            }
+            resolved_version_id = static_cast<VersionId>(requested_version_id);
         }
+
+        if (resolved_version_id <= opt_latest->version_id() &&
```
poodlewars, line 878: Comment to explain why we have `resolved_version_id <= opt_latest->version_id()`

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+            lib.write(sym, make_df())
+        qs.enable()
+        qs.reset_stats()
+        lib.read(sym)
+        stats = qs.get_query_stats()
+        # LATEST only needs VERSION_REF
+        assert get_version_reads(stats) == 0
+        assert get_version_ref_reads(stats) == 1
+
+
```
poodlewars, line 81: I tried to think through all the cases to cover for reads with the cache enabled.

https://github.com/man-group/ArcticDB/pull/2894#discussion_r3000213518 makes me think I might misunderstand how FROM_TIME reads are supposed to work.

Left some thoughts on what's covered and what might be gaps below. I'm sure some of the gaps are false positives and you'll be able to point me at where they're covered.

Latest
- Read latest version, with it both:
  - cached from LATEST `test_read_latest_then_read_latest`
  - cached from DOWNTO `test_read_oldest_then_read_latest`
  - cached from FROM_TIME - **cannot find this test**
  - and uncached
- Read latest version where the head N versions in the chain are deleted
- Read latest where it does not actually exist
Downto or From Time
- Read a cached version
  - DOWNTO test_read_oldest_then_read_any
  - FROM_TIME - **cannot find this**
- Read an uncached version
  - DOWNTO test_read_latest_then_read_specific
  - FROM_TIME **cannot find**
- Read a version deleted by a tombstone_all
  - Covered
- Read a version deleted by a tombstone
  - Covered
- Read a version with some unrelated tombstones and tombstone alls in the chain
  - Covered
- Read something newer than the latest in the cache
  - test_client_b_reloads_for_unknown_version covers DOWNTO but not the FROM_TIME behaviour

All
- Just verify that a full cache is not used for an interesting chain including tombstones and tombstone alls
  - Covered

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+def test_read_latest_then_read_latest(in_memory_store_factory, clear_query_stats):
```
poodlewars: Again many of the tests below this one can easily be parametrized tests

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+        qs.enable()
+        qs.reset_stats()
+        lib.read(sym, as_of=0)
+        stats = qs.get_query_stats()
+        # Cache miss: must load all 5 VERSION keys + re-read VERSION_REF
+        assert get_version_reads(stats) == 5
+        assert get_version_ref_reads(stats) == 1
+
+
+def test_list_versions_then_read(in_memory_store_factory, clear_query_stats):
```
poodlewars: Can broaden this test - reading any of the versions should be a cache hit

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+    lib = in_memory_store_factory()
+    sym = "test_from_time_then_downto_with_tombstone_all"
+    with config_context("VersionMap.ReloadInterval", MAX_RELOAD_INTERVAL):
+        lib.write(sym, make_df())
+        lib.write(sym, make_df())
+        lib.delete(sym)
+        lib.write(sym, make_df())
+        lib.write(sym, make_df())
+        # FROM_TIME(now) loads all version keys including tombstone_all
+        lib.read(sym, as_of=pd.Timestamp.now())
```
poodlewars: I'm confused, this looks like a bug. Why do we need to load all version keys for an `as_of=NOW`?

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+    with config_context("VersionMap.ReloadInterval", MAX_RELOAD_INTERVAL):
+        lib.write(sym, make_df())
+        lib.write(sym, make_df())
+        lib.delete(sym)
+        lib.write(sym, make_df())
+        lib.write(sym, make_df())
+        lib.read(sym, as_of=pd.Timestamp.now())
+        qs.enable()
+        qs.reset_stats()
+        # DOWNTO(0) - version 0 was deleted, but cache should know this
```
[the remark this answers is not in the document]
poodlewars: Agree

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+        stats = qs.get_query_stats()
+        assert get_version_reads(stats) == 0
+        assert get_version_ref_reads(stats) == 1
+
+
+def test_cache_expires_after_ttl(in_memory_store_factory, clear_query_stats):
+    """After TTL expires, the cache should be invalidated and VERSION_REF re-read."""
+    lib = in_memory_store_factory()
+    sym = "test_cache_expires_after_ttl"
+    with config_context("VersionMap.ReloadInterval", 100):  # 100ms
```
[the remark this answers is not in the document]
poodlewars: Agree

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+        time.sleep(0.2)  # Wait for TTL to expire
+        qs.enable()
+        qs.reset_stats()
+        lib.read(sym)
+        stats = qs.get_query_stats()
+        assert get_version_reads(stats) == 0
+        assert get_version_ref_reads(stats) == 1
+
+
+def test_cache_valid_within_ttl(in_memory_store_factory, clear_query_stats):
```
poodlewars: This is already covered by other tests

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+    lib_a = in_memory_store_factory()
+    lib_b = in_memory_store_factory(reuse_name=True)
+    sym = "test_client_b_reloads_for_unknown_version"
+    with config_context("VersionMap.ReloadInterval", MAX_RELOAD_INTERVAL):
+        lib_a.write(sym, pd.DataFrame({"col": [0]}))
+        lib_b.read(sym)  # B caches v0
+        lib_a.write(sym, pd.DataFrame({"col": [1]}))
+        qs.enable()
+        qs.reset_stats()
+        result = lib_b.read(sym, as_of=1)
```
poodlewars: Also test what happens if B reads with `as_of` a timestamp

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+        lib.append(sym, make_ts_df(3))
+        qs.enable()
+        qs.reset_stats()
+        lib.read(sym)
+        stats = qs.get_query_stats()
+        assert get_version_reads(stats) == 0
+        assert get_version_ref_reads(stats) == 0
+
+
+def test_append_historical_read_one_prior_version(in_memory_store_factory, clear_query_stats):
```
poodlewars: Is this comment right? Isn't this happening because we put the latest but one version in the VERSION_REF key too? Are we even using the cache here?

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,1050 @@
...
+        lib.snapshot("snap1")
+        qs.enable()
+        qs.reset_stats()
+        lib.read(sym)
+        stats = qs.get_query_stats()
+        assert get_version_reads(stats) == 0
+        assert get_version_ref_reads(stats) == 0
+
+
+def test_head_uses_cache(in_memory_store_factory, clear_query_stats):
```
poodlewars: parameterize these `_uses_cache` functions

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: This is a good catch. It's not a bug, but the comment is wrong. The reload of all is triggered by `lib.delete(sym)`. It calls `delete_all_versions` and eventually calls `tombstone_from_key_or_all` and this one trigger `check_reload`  with `LoadType::ALL`

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: We do hit the cache, but the cache is miss, because writes do not update `oldest_loaded_index_version_`. Basically the workflow is we first `has_cached_entry` check the cache entry, but this returns `false` and then we are triggering reload. But the reload stops short at version_ref because it has penultimate version as well. 
I agree the comment is misleading. I will update it.

cpp/arcticdb/version/version_map.hpp
maxim-morozov: What do you mean? Comment to explain why we are checking the version we are looking for is smaller or equal the latest version?

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    pytest.param(setup_tombstone_even_plus_latest, id="tombstone_even_plus_latest"),
+]
+
+# Expected (VERSION_reads, VERSION_REF_reads) for each (chain, cache, as_of) combination.
+# Measured empirically with a fresh library per combination. Snapshot reads always bypass
+# the version map cache (0, 0). Chains with delete/delete_version do a full chain load
+# during setup, so cache_on results are often (0, 0) even for historical reads.
+EXPECTED_COUNTS = {
+    ("simple_6", "no_cache", "latest"): (0, 1),
+    ("simple_6", "no_cache", "version_latest"): (0, 1),
```
poodlewars: Can add some cases for negative `as_of` like `version_penultimate`

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    pytest.param(setup_tombstone_all_then_writes, id="tombstone_all_then_writes"),
+    pytest.param(setup_tombstone_latest, id="tombstone_latest"),
+    pytest.param(setup_tombstone_all_every_version, id="tombstone_all_every_version"),
+    pytest.param(setup_tombstone_even_plus_latest, id="tombstone_even_plus_latest"),
+]
+
+# Expected (VERSION_reads, VERSION_REF_reads) for each (chain, cache, as_of) combination.
+# Measured empirically with a fresh library per combination. Snapshot reads always bypass
+# the version map cache (0, 0). Chains with delete/delete_version do a full chain load
+# during setup, so cache_on results are often (0, 0) even for historical reads.
```
poodlewars: You could write tests without this behaviour by creating two clients, like you did in the client A and client B tests

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    ("multi_tombstone_all", "no_cache", "list_versions"): (5, 1),
+    ("multi_tombstone_all", "cache_on", "latest"): (0, 0),
+    ("multi_tombstone_all", "cache_on", "version_latest"): (0, 0),
+    ("multi_tombstone_all", "cache_on", "oldest_undeleted_version"): (0, 0),
+    ("multi_tombstone_all", "cache_on", "version_mid"): (0, 0),
+    ("multi_tombstone_all", "cache_on", "timestamp_mid"): (0, 0),
+    ("multi_tombstone_all", "cache_on", "snapshot"): (0, 0),
+    ("multi_tombstone_all", "cache_on", "list_versions"): (0, 0),
+    ("multi_individual_tombstone", "no_cache", "latest"): (3, 1),
+    ("multi_individual_tombstone", "no_cache", "version_latest"): (3, 1),
```
poodlewars: Why does this, and the one above it, need 3 VERSION reads?

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    pytest.param("downto_oldest", "from_time", id="downto_oldest_then_from_time"),
+    pytest.param("snapshot", "downto_oldest", id="snapshot_then_downto_oldest"),
+    pytest.param("snapshot", "from_time", id="snapshot_then_from_time"),
+    pytest.param("from_time", "snapshot", id="from_time_then_snapshot"),
+    pytest.param("downto_mid", "snapshot", id="downto_mid_then_snapshot"),
+    pytest.param("snapshot", "snapshot", id="snapshot_then_snapshot"),
+]
+
+# Expected (VERSION_reads, VERSION_REF_reads) for the second read in each
+# (chain, sequence) combination. Measured empirically with a fresh library
```
poodlewars: What does `Measured empirically with a fresh library per combination.` mean? Remove this comment?

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+# per combination.
+EXPECTED_CROSS_QUERY_COUNTS = {
+    ("simple_6", "latest_then_latest"): (0, 0),
+    ("simple_6", "downto_oldest_then_latest"): (0, 0),
+    ("simple_6", "from_time_then_downto_oldest"): (6, 1),
+    ("simple_6", "from_time_then_downto_mid"): (0, 0),
+    ("simple_6", "downto_mid_then_from_time"): (0, 0),
```
poodlewars: There don't seem to be tests where we read to a version then read with a `from_time` earlier than that version, which should incur one read.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    # live: 1,3. delete_version() does LoadType::ALL during setup, so cache_on
+    # has full chain loaded. no_cache walks further for older versions.
+    ("tombstone_even_plus_latest", "no_cache", 0): (10, 1),
+    ("tombstone_even_plus_latest", "no_cache", 2): (8, 1),
+    ("tombstone_even_plus_latest", "no_cache", 4): (6, 1),
+    ("tombstone_even_plus_latest", "no_cache", 5): (5, 1),
+    ("tombstone_even_plus_latest", "cache_on", 0): (0, 0),
+    ("tombstone_even_plus_latest", "cache_on", 2): (0, 0),
+    ("tombstone_even_plus_latest", "cache_on", 4): (0, 0),
+    ("tombstone_even_plus_latest", "cache_on", 5): (0, 0),
```
poodlewars: None of the `cache_on` cases are that interesting, in the sense that none of them are doing any IO

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
poodlewars: Please could you reply to this explaining which of the test cases address which of my cases?

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: They are different cases, Client A and Client B has 2 version maps and testing reload. Where this one testing the  hit miss of the same version map

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: Because the TOMBSTONE of version 3 and version 1 is at the start of the chain. The comment in `setup_multi_individual_tombstone` is incorrect.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: It means it creates a new library every time.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: We still need to validate it to make sure that we don't introduce any bugs in this case. I have run it with the previous arcticdb version to validate that it's consistent pre and post change

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: It will not be one read. As soon as we don't find `from_time` we invalidate the cache and reload. I will add the test, but it's just invalidation of the cache with the current logic.

poodlewars: Yes this one is a misunderstanding, no need to add the test

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    lib.delete(sym)
+    lib.write(sym, make_df())
+    lib.delete(sym)
+    vinfos = [lib.write(sym, make_df()) for _ in range(3)]
+    lib.snapshot(sym + "_snap", versions={sym: 3})
+    return ChainInfo(4, 2, 3, sym + "_snap", _ts_from_version_info(vinfos[1]), 5)
+
+
+def setup_multi_individual_tombstone(lib, sym):
+    """v5 -> TOMBSTONE(v3) -> TOMBSTONE(v1) -> v4 -> v3 -> v2 -> v1 -> v0"""
```
IvoDD: The docstring is incorrect. Should be:

```
"""TOMBSTONE(v3) -> TOMBSTONE(v1) -> v5 -> v4 -> v3 -> v2 -> v1 -> v0"""
```

or if you meant to make v5 after the tombstones you should update code below.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+        lib.write(sym, make_ts_df(0, 3)),
+        lib.update(sym, make_ts_df(1, 1)),
+        lib.update(sym, make_ts_df(2, 1)),
+    ]
+    lib.snapshot(sym + "_snap", versions={sym: 1})
+    return ChainInfo(2, 0, 1, sym + "_snap", _ts_from_version_info(vinfos[1]), 3)
+
+
+def setup_prune_chain(lib, sym):
+    """v2 only (v0, v1 pruned)"""
```
IvoDD: Might be more useful to docstring the version chain not the `v2 only (v0, v1 pruned)`:
```
TOMBSTONE_ALL(1) -> v2 -> v1 -> v0
```

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    vinfos = [lib.write(sym, make_df()) for _ in range(3)]
+    lib.prune_previous_versions(sym)
+    lib.snapshot(sym + "_snap", versions={sym: 2})
+    return ChainInfo(2, 2, 2, sym + "_snap", _ts_from_version_info(vinfos[2]), 3)
+
+
+def setup_tombstone_all_then_writes(lib, sym):
+    """v6 -> v5 -> v4 -> v3 -> v2 ->TOMBSTONE_ALL -> v1 (delete all first, then 6 writes)"""
```
IvoDD: The docstring is incorrect. Should be:

```
v6 -> v5 -> v4 -> v3 -> v2 -> v1 ->TOMBSTONE_ALL -> v0
```

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    pytest.param(setup_simple_6, id="simple_6"),
+    pytest.param(setup_multi_tombstone_all, id="multi_tombstone_all"),
+    pytest.param(setup_multi_individual_tombstone, id="multi_individual_tombstone"),
+    pytest.param(setup_append_chain, id="append_chain"),
+    pytest.param(setup_update_chain, id="update_chain"),
+    pytest.param(setup_prune_chain, id="prune_chain"),
+    pytest.param(setup_tombstone_all_then_writes, id="tombstone_all_then_writes"),
+    pytest.param(setup_tombstone_latest, id="tombstone_latest"),
+    pytest.param(setup_tombstone_all_every_version, id="tombstone_all_every_version"),
+    pytest.param(setup_tombstone_even_plus_latest, id="tombstone_even_plus_latest"),
```
IvoDD: We are missing several interesting version chains:
- Version chain written with `write(prune_privious=True)`. This will put an index key and a tombstone_all in the same version key
- Version chain written with `delete_versions` (notice plural). This will put multiple tombtones in the same version key
- edge case: everything is deleted by a combination of tombstone and tombstone all
   -  e.g. `TOMBSTONE(5) -> TOMBSTONE(4) -> v5 -> v4 -> TOMBSTONE_ALL(3) -> v3 -> v2 -> v1 -> v0`. We should read only up to v3 before we raise `NoSuchVersionChain`
- snapshot and then delete (all of the above are snapshotting only live versions). This is not necassary for this PR though as snapshotting doesn't really affect version chain traversal.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    return stats.get("storage_operations", {}).get("Memory_GetObject", {}).get("VERSION_REF", {}).get("count", 0)
+
+
+# =============================================================================
+# Chain setup functions
+# =============================================================================
+# Each returns a ChainInfo with metadata about the created version chain.
+
+ChainInfo = namedtuple(
+    "ChainInfo", ["latest_version", "oldest_undeleted_version", "mid_version", "snapshot", "mid_ts", "total_versions"]
```
IvoDD: I think `total_versions` is misleading and is unused and we should drop it. In all setup functions we make `total_versions = total_versions_with_index_keys` which is not exactly the number `lib_tool.find_keys(KeyType.VERSION)` would return.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
IvoDD: Testing `write_everything` + `cache_on` + `get_some_versions` is kind of weird.

`setup` always contains the entire version chain when it writes everything. The only reason we don't always get `(0, 0)` is because we are a bit suboptimal and when writing v0 we don't set [is_earliest_loaded](https://github.com/man-group/ArcticDB/blob/c6748f3c2a2e0193a8c093dce1c58c0a9bff1d79/cpp/arcticdb/version/version_map_entry.hpp#L212)

I agree it is useful to test cases where we modify an already cached entry but for measuring operations it is more useful if we modify a not entirely loaded version chain:
```
1. Load version chain up to point X
2. Perform modification operations in cache (e.g. write/delete)
3. Verify 0 operations before X and some number of operations after X
```

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    sym = f"sym_{uuid.uuid4().hex[:8]}"
+    cache_id = "no_cache" if reload_interval == 0 else "cache_on"
+    chain_id = chain_setup.__name__.replace("setup_", "")
+
+    with config_context("VersionMap.ReloadInterval", reload_interval):
+        chain_setup(lib, sym) if callable(chain_setup) else chain_setup.values[0](lib, sym)
+
+        qs.enable()
+        qs.reset_stats()
+        lib.list_versions(sym)
```
IvoDD: Can't we fold this test to the above by:

```
READ_CHAIN_OPERATIONS = ["latest", "version_latest", "oldest_undeleted_version", "version_mid", "timestamp_mid", "snapshot", 
"list_all" # NEW
]


def perform_read_operation(lib, sym, operation, chain_info):
    if operation == "list_all":
        return lib.list_versions(sym)
    elif operation == "latest":
        as_of = None
    elif operation == "version_latest":
        as_of = chain_info.latest_version
    elif operation == "oldest_undeleted_version":
        as_of = chain_info.oldest_undeleted_version
    elif operation == "version_mid":
        as_of = chain_info.mid_version
    elif operation == "timestamp_mid":
        as_of = chain_info.mid_ts
    elif operation == "snapshot":
        as_of = chain_info.snapshot
    return lib.read(sym, as_of=as_of)
```

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+        mutation_fn(lib, sym)
+        qs.enable()
+        qs.reset_stats()
+        lib.read(sym)
+        stats = qs.get_query_stats()
+        assert get_version_reads(stats) == 0
+        if reload_interval == 0:
+            assert get_version_ref_reads(stats) == 1
+        else:
+            assert get_version_ref_reads(stats) == 0
```
IvoDD: Also assert the `get_version_reads` not just the ref?

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+        assert get_version_ref_reads(stats) == 1
+
+
+def test_cache_expires_after_ttl(in_memory_store_factory, clear_query_stats):
+    lib = in_memory_store_factory()
+    sym = f"sym_{uuid.uuid4().hex[:8]}"
+    with config_context("VersionMap.ReloadInterval", 100_000_000):  # 100ms in nanoseconds
+        lib.write(sym, make_df())
+        lib.read(sym)
+        time.sleep(0.2)
```
IvoDD: This test doesn't actually require a sleep because of the [tolerance](https://github.com/man-group/ArcticDB/blob/c6748f3c2a2e0193a8c093dce1c58c0a9bff1d79/cpp/arcticdb/version/version_map.hpp#L110).

In any case all timing related tests are better positioned in C++ where we can modify the clock and should already be covered.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    pytest.param("downto_oldest", "latest", id="downto_oldest_then_latest"),
+    pytest.param("from_time", "downto_oldest", id="from_time_then_downto_oldest"),
+    pytest.param("from_time", "downto_mid", id="from_time_then_downto_mid"),
+    pytest.param("downto_mid", "from_time", id="downto_mid_then_from_time"),
+    pytest.param("downto_oldest", "from_time", id="downto_oldest_then_from_time"),
+    pytest.param("snapshot", "downto_oldest", id="snapshot_then_downto_oldest"),
+    pytest.param("snapshot", "from_time", id="snapshot_then_from_time"),
+    pytest.param("from_time", "snapshot", id="from_time_then_snapshot"),
+    pytest.param("downto_mid", "snapshot", id="downto_mid_then_snapshot"),
+    pytest.param("snapshot", "snapshot", id="snapshot_then_snapshot"),
```
IvoDD: This is not a full cartesian product. We don't have `latest_then_downto_mid` for example. We don't have `downto_mid_then_from_time` and many others.

We should make it a full cartesian product. And we should change the `EXPECTED_CROSS_QUERY_COUNTS` to not just list all expected versions because it is very verbose. We can instead:

```
def expected_version(label, chain_info):
    if label == "latest":
        return chain_info.latest_version
    if label == "from_time":
        return chain_info.mid_version
    ....

def should_be_cached(first_label, second_label, chain_info):
    expected_first = expected_version(first_label, chain_info)
    expected_second = expected_version(second_label, chain_info)
    return expected_first <= expected_second # Any operation which loads lower version should cache all operations which would load a higher version


# The expected operations for a single as_of if the previous entry is not cached
EXPECTED_READS_IF_UNCACHED = {
   ("simple_6", "latest"): (1, 1),
   ("simple_6", "down_to_mid") : (3, 1),
   ....
}

@pytest.parametrize(first, AS_OFS)
@pytest.parametrize(secod, AS_OFS)
def test(first, second):
    setup...
    lib.read(as_of=first)
    qs.enable()
    lib.read(as_of=second)
    if should_be_cached(first, second):
         assert stats == (0, 0)
     else:
         assert stats == EXPECTED_READS_IF_UNCACHED[chain, second]
```

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+    first_label,
+    second_label,
+):
+    """Test that reads with different as_of types interact correctly with the cache."""
+    lib = in_memory_store_factory()
+    sym = f"sym_{uuid.uuid4().hex[:8]}"
+    chain_id = chain_setup.__name__.replace("setup_", "")
+
+    with config_context("VersionMap.ReloadInterval", MAX_RELOAD_INTERVAL):
+        chain_info = chain_setup(lib, sym)
```
IvoDD: We should use separate libraries for `chain_setup` and `read` + `measurement`. We're not testing what we expect because the `chain_setup` has already populated the entire cache.

E.g. `latest_then_downto_mid` should not be `(0, 0)` but with the current setup it is.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,843 @@
...
+def setup_tombstone_latest(lib, sym):
+    """TOMBSTONE(v5) -> v5 -> v4 -> v3 -> v2 -> v1 -> v0 (delete latest version only)"""
+    vinfos = [lib.write(sym, make_df()) for _ in range(6)]
+    lib.delete_version(sym, 5)
+    lib.snapshot(sym + "_snap", versions={sym: 3})
+    return ChainInfo(4, 0, 3, sym + "_snap", _ts_from_version_info(vinfos[3]), 6)
+
+
+def setup_tombstone_all_every_version(lib, sym):
+    """v11->...->TOMBSTONE_ALL->v5->TOMBSTONE_ALL->...->v0 (delete after each of 6 writes, then 6 more)"""
```
IvoDD: The docstring is also incorrect. Topmost should be `TOMBSTONE_ALL`, not `v11`

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
IvoDD: Also suffers from preloading the cache with `setup_chain`

IvoDD (changes requested):

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,411 @@
...
+        elif as_of_type in ("oldest_undeleted_version", "version_neg_oldest"):
+            v = undeleted_asc[0]
+            return penultimate if v >= penultimate else v
+    return None
+
+
+def setup_simple_6(lib, sym):
+    """v5 -> v4 -> v3 -> v2 -> v1 -> v0"""
+    w = [lib.write(sym, make_df()) for _ in range(6)]
+    return [w[5], w[4], w[3], w[2], w[1], w[0]]
```
poodlewars: This is just `w[::-1]`? Same applies in many of the functions below

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,411 @@
...
+    for entry in chain:
+        if isinstance(entry, tuple):
+            tombstoned_ids.add(entry[1])
+        elif entry == TOMBSTONE_ALL:
+            tombstone_all_seen = True
+        else:
+            yield entry, entry.version in tombstoned_ids or tombstone_all_seen
+
+
+def _is_tombstone_head(ci):
```
poodlewars: Use a more meaningful name than `ci`

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
poodlewars: This has not been addressed

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,411 @@
...
+    return pd.Timestamp(version_info.timestamp, unit="ns", tz="UTC")
+
+
+def _iter_versions(chain):
+    """Yield (vinfo, is_deleted) newest-to-oldest"""
+    tombstoned_ids = set()
+    tombstone_all_seen = False
+    for entry in chain:
+        if isinstance(entry, tuple):
+            tombstoned_ids.add(entry[1])
```
poodlewars: assert `entry[0] == "tombstone"` or pull out a `class Tombstone: def __init__(self, v): self.version = v` and do an `isinstance(entry, Tombstone)` check here (and also a `class TombstoneAll: pass`).

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+def setup_tombstone_all_every_version(lib, sym):
+    """v11->...->v6 -> TOMBSTONE_ALL->v5->...->TOMBSTONE_ALL->v0"""
+    initial = []
+    for _ in range(6):
+        initial.append(lib.write(sym, make_df()))
+        lib.delete(sym)
+    rest = [lib.write(sym, make_df()) for _ in range(6)]  # v6..v11
+    chain = list(reversed(rest))
```
poodlewars: `chain = rest[::-1] + initial[::-1]`

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,411 @@
...
+        elif as_of_type in ("oldest_undeleted_version", "version_neg_oldest"):
+            return undeleted_asc[0]
+    elif len(undeleted_asc) > 1:  # we have penultimate index also cached
+        penultimate = undeleted_asc[-2]
+        if as_of_type in ("latest", "version_latest", "version_neg_1"):
+            return penultimate
+        elif as_of_type in ("version_mid", "version_neg_mid", "timestamp_mid"):
+            v = _mid_version(ci)
+            return penultimate if v >= penultimate else v
+        elif as_of_type in ("oldest_undeleted_version", "version_neg_oldest"):
```
poodlewars: Raise for unexpected `as_of_type`

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,411 @@
...
+
+def make_df():
+    return pd.DataFrame({"col": [1, 2, 3]})
+
+
+def make_ts_df(start_day=0, periods=3):
+    return pd.DataFrame({"col": range(periods)}, index=pd.date_range(f"2020-01-{start_day + 1:02d}", periods=periods))
+
+
+def get_version_reads(stats):
```
poodlewars, line 26: Use `query_stats_operation_count` https://chat-man.slack.com/archives/C02NEG3V38X/p1775745564700529

poodlewars (approved):

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,435 @@
...
+    version_id: int
+    type: TombstoneType
+
+
+def _ts_from_version_info(version_info):
+    return pd.Timestamp(version_info.timestamp, unit="ns", tz="UTC")
+
+
+def _iter_versions(chain):
+    """Yield (vinfo, is_deleted) newest-to-oldest"""
```
IvoDD, line 49: The `chain` data structure in python is insufficient for storing multiple entries in a single version key. This can happen for:
- version chain compaction (the last version key can hold multiple indices and tombstones)
- `batch_delete_versions` (a single version key can hold multipple tombstones)
- `write(prune_previous_versions=True)` (a single version key can hold both an index key and a tombstone all key)

Hence the above cases are missing for all version chain setups in this file.

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,435 @@
...
+    ("simple_6", setup_simple_6),
+    ("multi_tombstone_all", setup_multi_tombstone_all),
+    ("multi_individual_tombstone", setup_multi_individual_tombstone),
+    ("append_chain", setup_append_chain),
+    ("update_chain", setup_update_chain),
+    ("prune_chain", setup_prune_chain),
+    ("tombstone_all_then_writes", setup_tombstone_all_then_writes),
+    ("tombstone_latest", setup_tombstone_latest),
+    ("tombstone_all_every_version", setup_tombstone_all_every_version),
+    ("tombstone_even_plus_latest", setup_tombstone_even_plus_latest),
```
IvoDD, line 216: We are also missing versions where:
- All indices are deleted

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,435 @@
...
+
+AS_OF_TYPES = [
+    "latest",
+    "version_latest",
+    "oldest_undeleted_version",
+    "version_mid",
+    "timestamp_mid",
+    "version_neg_1",
+    "version_neg_mid",
+    "version_neg_oldest",
```
IvoDD, line 227: We are missing `as_of` requests that try to read an version deleted by a tombstone all.

There are interesting things to test there (e.g. if requesting v0 but there is a tombstone all at v5 we will stop iterating early at v5)

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
```diff
@@ -0,0 +1,435 @@
...
+        elif as_of_type == "version_neg_1":
+            return _total_versions(chain_info) - 1
+        elif as_of_type in ("version_mid", "version_neg_mid", "timestamp_mid"):
+            return _mid_version(chain_info)
+        elif as_of_type in ("oldest_undeleted_version", "version_neg_oldest"):
+            return undeleted_asc[0]
+        else:
+            raise Exception("Unexpected as_of_type")
+    elif len(undeleted_asc) > 1:  # we have penultimate index also cached
+        penultimate = undeleted_asc[-2]
```
IvoDD, line 115: Nit: Can we simplify the else branch by just doing `min(rloaded_version, penultimate)`?
E.g.:
```python
        if as_of_type in ("latest", "version_latest"):
            loaded = undeleted_asc[-1]
        elif as_of_type == "version_neg_1":
            loaded =_total_versions(chain_info) - 1
        elif as_of_type in ("version_mid", "version_neg_mid", "timestamp_mid"):
            loaded =_mid_version(chain_info)
        elif as_of_type in ("oldest_undeleted_version", "version_neg_oldest"):
            loaded = undeleted_asc[0]
        else:
            raise Exception("Unexpected as_of_type")
        
        if len(undeleted_asc) > 1:  # we have penultimate index also cached
            penultimate = undeleted_asc[-2]
            return min(loaded, penultimate)
        else:
            return loaded
```

IvoDD (approved):
Approved as the coverage is good enough even if not 100%

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: Fair point. I will add the case of having `oldest`, which will cover the cases where we have tombstone and tombstone_all

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: There is not much to test there except Exception. There are chains like `tombstone_all_every_version` which pretty much showing the same

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: Good idea

python/tests/integration/arcticdb/version_store/test_version_map_cache_storage_ops.py
maxim-morozov: The idea was not to represent the chain in Python, but rather give some representation to know what versions are active what not. Agree that we can cover more and more chains. We can extend the test and change it further as we need it. I think at the moment it gives sufficient coverage to make sure that the invalidation of the cache works correctly with newer `as_of` parameter.

PR #2934, opened 2026-02-25
Fix #2838: add modulo operator support to QueryBuilder expressions

### LINKED ISSUE #2838: Filtering datetime index by minute in QueryBuilder
I have a DataFrame of prices with a DatetimeIndex. I’m trying to filter rows in QueryBuilder based on a specific minute value from the index.

Example:
```python
q = adb.QueryBuilder()

q = q[q["index"].dt.minute == 10]
# AttributeError: 'ExpressionNode' object has no attribute 'dt'

q = q[q["index"].minute == 10]
# AttributeError: 'ExpressionNode' object has no attribute 'minute'
```

It seems that datetime accessors like .dt or .minute are not supported on ExpressionNode.

Question:
What is the correct way to filter a datetime index by time components (e.g., minute, hour, time, or exact timestamp) when using QueryBuilder? Is this functionality supported, or is there an alternative recommended approach?

academy-codex:
#### Reference Issues/PRs
Fixes #2838.

This PR adds modulo (`%`) operator support to ArcticDB query expressions, which enables minute-style filtering on datetime indexes via modulo arithmetic (for example, filtering rows where `index % 1h` falls in a target minute window).

Implemented end-to-end:
- Python `ExpressionNode.__mod__` and `ExpressionNode.__rmod__`
- C++ `OperationType::MOD` and binary dispatch wiring
- C++ expression type inference support for MOD
- pybind exposure of `OperationType.MOD`
- C++ formatter for `ModOperator` (`%`) used in expression/log stringification
- explicit MOD binary operator template instantiation and build wiring

Sample usage (replacement for unsupported `.dt.minute` on `ExpressionNode`):

```python
import pandas as pd
q = adb.QueryBuilder()

# Equivalent to: index.minute == 10
minute_in_hour = q["index"] % pd.Timedelta(hours=1)
q = q[(minute_in_hour >= pd.Timedelta(minutes=10)) &
      (minute_in_hour <  pd.Timedelta(minutes=11))]
```

Tests added:
- Python unit: modulo projection in QueryBuilder
- Python unit: datetime index minute filtering using modulo + `Timedelta`
- C++ unit: binary modulo dispatch behavior

Validation run in this branch:
- `cmake --build cpp/out/macos-release-build --target install_arcticdb_ext`
- `python -m pytest -q python/tests/unit/arcticdb/version_store/test_query_builder.py -k "projection_modulo or datetime_index_by_minute_with_modulo"`

#### Any other comments?
This keeps the user-facing approach aligned with existing arithmetic expression support and avoids introducing a separate datetime `.minute` accessor in QueryBuilder.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Is this contribution tested against all ArcticDB's features?
 - [x] Do all exceptions introduced raise appropriate error messages?
</details>


cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -226,6 +227,14 @@ std::variant<BitSetTag, DataType> ExpressionNode::compute(
                             res = data_type_from_raw_type<TargetType>();
                             break;
                         }
+                        case OperationType::MOD: {
+                            using TargetType = typename binary_operation_promoted_type<
```
poodlewars, line 231: I don't understand this type promotion logic, isn't the type of `a % b` always going to be `type(a)`?

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -51,6 +52,8 @@ enum class OperationType : uint8_t {
     AND,
     OR,
     XOR,
+    // Operator (kept here to preserve existing enum values)
```
poodlewars: Why is it important to preserve the existing enum values (but still change `TERNARY`)? Why can't this live with the other operators?

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -356,6 +360,17 @@ struct DivideOperator {
     }
 };
 
+struct ModOperator {
+    template<typename T, typename U, typename V = typename binary_operation_promoted_type<T, U, ModOperator>::type>
+    V apply(T t, U u) {
+        if constexpr (std::is_floating_point_v<V>) {
+            return std::fmod(static_cast<V>(t), static_cast<V>(u));
```
poodlewars: `fmod` won't match Pandas will it?

```
In [5]: df = pd.DataFrame({"a": [1.0,2.0,3.0, 4.0, 5.0, -1.0, -2.0, -3.0]})

In [6]: df["b"] = df["a"] % 2.0

In [7]: df
Out[7]: 
     a    b
0  1.0  1.0
1  2.0  0.0
2  3.0  1.0
3  4.0  0.0
4  5.0  1.0
5 -1.0  1.0
6 -2.0  0.0
7 -3.0  1.0
```

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -356,6 +360,17 @@ struct DivideOperator {
...
+        } else {
+            return static_cast<V>(t) % static_cast<V>(u);
```
poodlewars: What happens if `u` is 0?
What happens if the user does this with a non-numeric type, or a bool?

poodlewars (changes requested):
Thanks for the PR! Great to have some community contributions. Left some questions below.

cpp/arcticdb/processing/expression_node.cpp
alexowens90: Agree with @poodlewars that the type of `a % b` will always be the type of `a`.
I would suggest fixing `binary_operation_promoted_type` to handle the `ModOperator` case, which will then also automatically fix the output type of `ModOperator::apply` as well, and keep all of the logic in one place.

cpp/arcticdb/processing/test/test_operation_dispatch.cpp
```diff
@@ -75,6 +75,20 @@ TEST(OperationDispatch, binary_operator) {
     EXPECT_THROW(visit_binary_operator(value, empty_column, PlusOperator{}), SchemaException);
 }
 
+TEST(OperationDispatch, binary_operator_modulo) {
```
alexowens90: This can be folded into the test above.
The variants `val % col` and `col % col` should also be tested

cpp/arcticdb/processing/operation_types.hpp
alexowens90: Our approach with dividing by zero is to not try and handle it with an if in a hot loop, we just let the exception bubble up, so I think it is consistent to do the same here.
Non-numeric types are handled [higher up the stack](https://github.com/man-group/ArcticDB/blob/c2c75f4ac8bb7abb0ea581a584fdedb91ba568ec/cpp/arcticdb/processing/operation_dispatch_binary.hpp#L460).
As mentioned in other comments, `V == T`. This means that static_cast<V>(u) can be wrong if `sizeof(U) > sizeof(T)`, or if the sizes are equal but one is signed and the other unsigned.

python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -241,6 +241,41 @@ def can_read_back(write_with_time, filter_with_time):
     assert not can_read_back(us_time, notz_summer_time)
 
 
+def test_querybuilder_projection_modulo(lmdb_version_store_tiny_segment, any_output_format):
```
alexowens90: `test_projection.py` is a more suitable location for this test

python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -241,6 +241,41 @@ def can_read_back(write_with_time, filter_with_time):
...
+    q = q.apply("mod_col", q["col"] % 3)
+
+    expected = df.copy()
+    expected["mod_col"] = expected["col"] % 3
+    received = lib.read(symbol, query_builder=q).data
+
+    assert_frame_equal(expected, received)
+
+
+def test_querybuilder_filter_datetime_index_by_minute_with_modulo(lmdb_version_store_tiny_segment, any_output_format):
```
alexowens90: `test_filtering.py` is a more suitable location for this test

python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -241,6 +241,41 @@ def can_read_back(write_with_time, filter_with_time):
...
+    lib.write(symbol, df)
+
+    q = QueryBuilder()
+    q = q.apply("mod_col", q["col"] % 3)
+
+    expected = df.copy()
+    expected["mod_col"] = expected["col"] % 3
+    received = lib.read(symbol, query_builder=q).data
+
+    assert_frame_equal(expected, received)
```
alexowens90: Does this work if `received` is a `pyarrow.Table` or `polars.DataFrame`?

python/tests/unit/arcticdb/version_store/test_query_builder.py
alexowens90: This requires a lot more test coverage. I would suggest at least:

- Modding with and against negative integers, and negative and positive floats
- Modding against 0, NaN, and +-infinity
- Modding one column against another column, and a value against a column
- A lot of mixed-types testing, focusing on values that are not representable by the other type. e.g. `uint8(10) % int64(300) == 10`, and for `int64(300) % uint8(10) == 0`

cpp/arcticdb/processing/test/test_operation_dispatch.cpp
academy-codex: Done, I folded modulo checks into OperationDispatch.binary_operator and added explicit coverage for both col % col and value % col.

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Done, I moved projection modulo tests from [test_query_builder.py](app://-/index.html#) to [test_projection.py](app://-/index.html#).

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Done, I moved the DatetimeIndex minute modulo filter test to [test_filtering.py](app://-/index.html#).

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Agreed and updated. The modulo projection assertions now normalize read outputs with regularize_dataframe, so the checks are robust across Pandas/PyArrow/Polars output modes.

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Done, I substantially expanded modulo coverage in [test_projection.py](app://-/index.html#): negative ints/floats, 0/NaN/+/-inf cases, col%col and value%col, plus mixed-type edge cases including non-representable combinations (e.g. uint8 with int64/float64). Also aligned integer modulo sign semantics to Python/Pandas for negative divisors.

alexowens90:
@academy-codex the comments have not all been addressed, please go through them again and make all of the requested changes

cpp/arcticdb/processing/expression_node.cpp
academy-codex: Addressed in commit 3dc8cb22. I added binary_operation_promoted_type<LHS, RHS, ModOperator> specialization in operation_types.hpp so modulo promotion is centralized; ModOperator::apply now follows that target type. I also extended mixed signed/unsigned 64-bit modulo coverage in test_projection.py.

cpp/arcticdb/processing/operation_types.hpp
academy-codex: Addressed: MOD now lives with the arithmetic operators (ADD/SUB/MUL/DIV) in OperationType in operation_types.hpp.

cpp/arcticdb/processing/operation_types.hpp
academy-codex: Addressed: floating modulo now follows Python/Pandas sign semantics (result sign follows divisor) instead of raw std::fmod output. Added negative-float modulo coverage in both C++ and Python tests.

cpp/arcticdb/processing/operation_types.hpp
academy-codex: Addressed: integral modulo by zero is validated and raises user-input error; floating modulo by zero returns NaN. Non-numeric/bool remain blocked by existing expression type checks upstream. The 3dc8cb22 promotion change also avoids incorrect mixed signed/unsigned casting behavior.

cpp/arcticdb/processing/test/test_operation_dispatch.cpp
academy-codex: Confirmed and retained: modulo checks are folded into OperationDispatch.binary_operator and include both col % col and val % col variants.

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Confirmed and retained: projection modulo tests were moved to python/tests/unit/arcticdb/version_store/test_projection.py.

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Confirmed and retained: DatetimeIndex minute modulo filtering test was moved to python/tests/unit/arcticdb/version_store/test_filtering.py.

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Addressed: projection assertions normalize outputs via regularize_dataframe before comparison, so Pandas/PyArrow/Polars output modes are handled.

python/tests/unit/arcticdb/version_store/test_query_builder.py
academy-codex: Addressed with expanded coverage in test_projection.py: negative integer/float cases, special RHS values (0/NaN/+-inf handling where applicable), col%col and val%col variants, and mixed-type/non-representable combinations. Commit 3dc8cb22 additionally adds directional signed/unsigned 64-bit mixed cases.

PR #2943, opened 2026-02-27
Track C++ Google Benchmark microbenchmarks via ASV

vasil-pashov:
# Track C++ Google Benchmark microbenchmarks via ASV

## Registering benchmarks
C++ microbenchmark results were not visible in the ASV dashboard and could not block PRs on regressions. This wires the existing Google Benchmark suite into ASV so C++ performance regressions are caught with the same tooling and blocking semantics as Python benchmarks.
                                                                                                                                                                                                                                                                                                                                                                                              
`python/benchmarks/cpp_microbenchmarks.py` is a new ASV module bridging Google Benchmark and ASV. At import time it invokes the benchmark binary with `--benchmark_list_tests=true` and dynamically registers one ASV `track_time_ms` class per top-level Google Benchmark function. Parameter variants (e.g. `BM_foo/10000/1`) become ASV params. Each track_time_ms call shells out to the binary with `--benchmark_filter` and `--benchmark_format=json`.

## Handling flaky benchmarks

### Discovering flaky benchmarks

`build_tooling/check_cpp_bench_stability.py` is a script that can run a subset of google benchmarks multiple times. It can be used to discover flaky benchmarks. The script records statistics about the runs. If `max_time / min_time` is more than than ASV's threshold the benchmark is considered as potentially flaky.

### Handling flaky benchmarks

There we a couple of flaky benchmarks. Google benchmark supports several options similar to ASV that control benchmark execution and we can use to stabilize run time.

* Iterations - control the inner (benchmark loop). This tells GB how many time to run the body of the inner loop. Internally GB times each run of the body and gives us back the average.
* Repetitions - This tells GB how many time to run the benchmark function itself. It then returns statistics about each run, these statistics can be aggregated. We use the median. Example if we set `iterations` to 10 an then `repetitions` to 3 it will run the the whole body of the test 3 times and each run will run the inner benchmark loop 10 times. The timings we'll get will be for the time it took to run the whole test i.e. 10 iteration of the inner loop.
* MinTime - controls the inner loop (cannot be used alongside `iterations`). How many seconds to run the inner loop for.

Timing: when a benchmark uses ->Repetitions(N)->ReportAggregatesOnly(true), Google Benchmark emits only aggregate rows; the bridge picks the median row. This matches ASV's own convention for timing benchmarks. When no repetitions are configured there is one iteration row and its value is used directly.                                                                             
   
## Workflow changes                                                                                                                                                                                                                                                                                                                                                                            

Both `benchmark_commits.yml` and `analysis_workflow.yml` gain a new "Build C++ benchmarks" step that compiles the benchmarks CMake target. The binary path defaults to `cpp/out/linux-release-build/arcticdb/benchmarks` and can be overridden via `ARCTICDB_CPP_BENCHMARKS_BINARY`.                                                                                                                

Results land in the same ArcticDB library (asv_results on S3 bucket arcticdb-ci-benchmark-results, symbol = short commit hash) as the Python benchmarks. On non-master branches the library is <branch>_asv_results.                                                                                                                                                                        

The `analysis_workflow.yml` step is needed so cpp_microbenchmarks.py can register its classes at import time during the asv_checks.py hash verification — without it the check would fail on every PR.

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -218,6 +218,18 @@ jobs:
         run: |
           python -m pip install -ve .[Testing]
           
+      - name: Build C++ benchmarks (needed for ASV discovery)
+        # cpp_microbenchmarks.py invokes the binary with
+        # --benchmark_list_tests=true at import time to register ASV classes.
+        # asv_checks.py needs those classes so the regenerated benchmarks.json
+        # matches the committed one.
+        shell: bash -l {0}
+        run: |
+          cmake --preset linux-release -DTEST=ON cpp
+          cmake --build cpp/out/linux-release-build --target benchmarks
+        env:
+          CMAKE_BUILD_PARALLEL_LEVEL: ${{vars.CMAKE_BUILD_PARALLEL_LEVEL}}
+
       - name: Run ASV Tests Check script
         run: |
           python python/utils/asv_checks.py
--- a/python/benchmarks/cpp_microbenchmarks.py
+++ b/python/benchmarks/cpp_microbenchmarks.py
@@ -93,7 +93,14 @@ def escape_for_benchmark_filter(name: str) -> str:
 
 
 def run_one(binary: Path, full_name: str) -> float:
-    """Run a single benchmark by exact-match filter, return its real_time in milliseconds."""
+    """Run a single benchmark by exact-match filter, return its real_time in milliseconds.
+
+    When the C++ side sets ``->Repetitions(N)->ReportAggregatesOnly(true)``, the JSON
+    contains only aggregate rows (mean/median/stddev/cv) and we use the median —
+    robust against a single outlier repetition (typically the first, which pays
+    cold-cache costs). When no repetitions are configured, there is exactly one
+    iteration row and we use that value directly.
+    """
     result = subprocess.run(
         [
             str(binary),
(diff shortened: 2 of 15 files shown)


.github/workflows/analysis_workflow.yml
```diff
@@ -96,7 +120,7 @@ jobs:
         run: |
           apt update
           apt install -y git
-          python -m pip install arcticdb[Testing] "protobuf<6"
+          python -m pip install arcticdb[Testing] asv "protobuf<6"
```
poodlewars: `asv` is already in the Testing deps?

.github/workflows/benchmark_cpp_microbenchmarks.yml
```diff
@@ -0,0 +1,151 @@
...
+        run: |
+          . build_tooling/vcpkg_caching.sh # Linux follower needs another call in CIBW
+          echo -e "VCPKG_BINARY_SOURCES=$VCPKG_BINARY_SOURCES
+          VCPKG_ROOT=$PLATFORM_VCPKG_ROOT" | tee -a $GITHUB_ENV
+          cmake -P cpp/CMake/CpuCount.cmake | sed 's/^-- //' | tee -a $GITHUB_ENV
+          yum install -y awscli
+        env:
+          CMAKE_BUILD_PARALLEL_LEVEL: ${{vars.CMAKE_BUILD_PARALLEL_LEVEL}}
+
+      - name: Build C++ benchmarks binary
```
poodlewars: This was already done in the other workflow file?

.github/workflows/benchmark_cpp_microbenchmarks.yml
poodlewars, line 1: There's a lot of duplication with the analysis workflow. Can't we just fold the new steps in to that one?

.github/workflows/benchmark_cpp_microbenchmarks.yml
```diff
@@ -0,0 +1,151 @@
...
+        run: |
+          git config --global --add safe.directory .
+          python -m asv run --show-stderr --durations all --bench "^cpp_microbenchmarks\." ${{ inputs.commit }}^!
+
+      - name: Benchmark against master
+        if: github.event_name == 'pull_request'
+        shell: bash -l {0}
+        run: |
+          # Look up the most recent master commit stored in our ASV database and compare against it.
+          ARCTICDB_REAL_S3_BUCKET=${ASV_RESULTS_BUCKET} ./tooling_venv/bin/python build_tooling/transform_asv_results.py --mode extract-recent --arcticdb_library cpp_microbenchmarks_asv_results || exit 1
```
poodlewars: Why do we need a totally separate library to store these? Are they presented properly in the UI?

python/benchmarks/cpp_microbenchmarks.py
```diff
@@ -0,0 +1,206 @@
...
+import subprocess
+import tempfile
+from collections import defaultdict
+from pathlib import Path
+
+# Resolve relative to the repository root (three levels up from this file:
+# python/benchmarks/cpp_microbenchmarks.py → python/benchmarks → python → repo root).
+REPO_ROOT = Path(__file__).resolve().parents[2]
+
+CPP_BENCHMARKS_BINARY = Path(
```
poodlewars, line 43: We should make sure we blow up with a clear error message explaining that you need to build the benchmarks binary, if this is pointing at a binary that doesn't exist. Otherwise it will be hard for people to update the ASV `.json` file.

poodlewars (changes requested):

PR #2970, opened 2026-03-13
Unpin some dependencies

alexowens90:
- openssl
- curl


alexowens90:
Quick test of whether CI would pass with unpinned openssl and curl

PR #2974, opened 2026-03-17
Extend existing failure notifications for tracking flaky CI issues

G-D-Petrov:
Replaces the inline bash in `failure_notification.yaml` with standalone Python
scripts that parse CI failures, track them as GitHub issues, and send Slack
notifications. The new implementation adds three capabilities that the bash
version lacked:

- **Timeout detection** — test-runner steps that fail without producing test
  names are recognised as timeouts and tracked under a dedicated `CI timeout`
  issue instead of the generic "unparseable" bucket.
- **PR comment trigger** — developers can comment `/analyse-failures` on a PR
  to analyse its most recent failed CI run and get results posted back as a
  comment.
- **Correct JSON escaping** — Slack payloads are built with `json.dumps`
  instead of bash string interpolation, eliminating broken messages from
  special characters in test names.

## How failures are classified

```
                  ┌─────────────────────────┐
                  │ Are there parsed test    │
                  │ names (FAILED/ERROR)?    │
                  └────────┬────────────────┘
                     yes   │          no
                           ▼          │
                  ┌──────────────┐    ▼
                  │ test_failure │  ┌──────────────────────────┐
                  │              │  │ Are there infra steps    │
                  │ Track each   │  │ (non-test-runner)?       │
                  │ test as a    │  └────────┬─────────────────┘
                  │ flaky-test   │     yes   │         no
                  │ issue        │           ▼         │
                  └──────────────┘  ┌──────────────┐   ▼
                                    │ infra_failure │ ┌────────────────────────┐
                                    │              │ │ Were test-runner steps  │
                                    │ Track each   │ │ stripped out?           │
                                    │ step as a    │ └────────┬───────────────┘
                                    │ flaky-step   │    yes   │        no
                                    │ issue        │          ▼        │
                                    └──────────────┘ ┌─────────┐       ▼
                                                     │ timeout │  ┌─────────┐
                                                     │         │  │ unknown │
                                                     │ Single  │  │         │
                                                     │ CI      │  │ Unparse-│
                                                     │ timeout │  │ able    │
                                                     │ issue   │  │ issue   │
                                                     └─────────┘  └─────────┘
```

`--conclusion timed_out` from GitHub forces the `timeout` path regardless.

### Example 1: Test failure — [run 24664309055](https://github.com/man-group/ArcticDB/actions/runs/24664309055)

**Context:** `Building master on push`, conclusion `failure`.
One failed job: `3.9 Linux / integration-DefaultCache`, failed step: `Run test`.

Logs contain:
```
FAILED tests/integration/toolbox/test_library_tool.py::test_overwrite_append_data
```

**Pipeline output:**
- `failure_kind.txt` → `test_failure`
- `failing_tests.txt` → `tests/integration/toolbox/test_library_tool.py::test_overwrite_append_data`
- `failed_steps.txt` → empty (`Run test` filtered as test-runner step)
- **Issue created:** `Flaky test: tests/integration/toolbox/test_library_tool.py::test_overwrite_append_data` with `flaky-test` label
- **Slack:** `:rotating_light: *New* — \`tests/integration/toolbox/test_library_tool.py::test_overwrite_append_data\``

### Example 2: Infrastructure failure — [run 24807645574](https://github.com/man-group/ArcticDB/actions/runs/24807645574)

**Context:** `Building master on schedule`, conclusion `failure`.
One failed job: `3.11 Windows / compile (...)`, failed step: `Install Required MSVC`.

No test logs (compile job, not a test job).

**Pipeline output:**
- `failure_kind.txt` → `infra_failure`
- `failing_tests.txt` → empty
- `failed_steps.txt` → `Install Required MSVC`
- **Issue created:** `Flaky step: Install Required MSVC` with `flaky-step` label
- **Slack:** `:rotating_light: *New* — \`Install Required MSVC\``

### Example 3: Timeout — [run 24841138872](https://github.com/man-group/ArcticDB/actions/runs/24841138872)

**Context:** `Building master on workflow_dispatch`, conclusion `failure`.
Two failed jobs:
- `3.11 Windows / unit-DefaultCache` — step `Run test` timed out after 120 min
- `3.10 Windows / stress-DefaultCache` — step `Run test` timed out

Logs show tests running (at 95% progress) but no `FAILED`/`ERROR` lines — the
runner was killed before completion.

**Pipeline output:**
- `failure_kind.txt` → `timeout` (only test-runner steps, no parsed test names)
- `failing_tests.txt` → empty
- `failed_steps.txt` → empty (`Run test` filtered)
- **Issue created:** `CI timeout` with `ci-failure` label
- **Slack:** `:hourglass: *Timeout*`


.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
     types: [completed]
 
 jobs:
-  on-failure:
+  track-failures:
     runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
IvoDD: I think we should make the `track-failures` run on PR branches only after a human presses a button. Otherwise we will fill up github with false positives (e.g. I pushed a small change to my PR but broke all my tests because of dumb bug)

.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
...
-          status: ${{ github.event.workflow_run.conclusion }}
-          notification_title: " ${{github.event.workflow_run.name}} - ${{github.event.workflow_run.conclusion}} on ${{github.event.workflow_run.head_branch}} - <${{github.server_url}}/${{github.repository}}/actions/runs/${{github.event.workflow_run.id}}|View Failure>"
-          message_format: ":fire: *${{github.event.workflow_run.name}}* ${{github.event.workflow_run.conclusion}} in <${{github.server_url}}/${{github.repository}}/actions/runs/${{github.event.workflow_run.id}}|${{github.repository}}>"
-          footer: "Linked Repo <${{github.server_url}}/${{github.repository}}|${{github.repository}}> | <${{github.server_url}}/${{github.repository}}/actions/runs/${{github.event.workflow_run.id}}|View Failure>"
+      - name: Ensure labels exist
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+        run: |
+          gh label create "flaky-test" --repo "$REPO" --color "FBCA04" --description "Flaky test auto-tracked by CI" 2>/dev/null || true
```
IvoDD: Should we use an `autogenerated` label for all of these? (So that we can filter and remove all in case of noise or bugs)

Or prefix all these labels with autogenerated e.g. `autogenerated-flaky-test`

.github/workflows/failure_notification.yaml
IvoDD: Nit: Maybe rename the workflow

.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+
+          echo "Fetching failed jobs for run $RUN_ID..."
+
+          failed_jobs_json=$(gh api "repos/$REPO/actions/runs/$RUN_ID/jobs?filter=latest&per_page=100" \
```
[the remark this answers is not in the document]
IvoDD: I agree with claude's comment

.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
...
+          fi
+
+          echo "Failed jobs: $failed_jobs_json"
+
+          # --- Collect failed steps from each job (via API) ---
+          failed_steps=""
+          for job_id in $(echo "$failed_jobs_json" | jq -r '.[].id'); do
+            job_name=$(echo "$failed_jobs_json" | jq -r ".[] | select(.id == $job_id) | .name")
+            step_names=$(gh api "repos/$REPO/actions/jobs/$job_id" \
+              --jq '.steps[] | select(.conclusion == "failure") | .name')
```
IvoDD: I don't think large bash scripts should live standalone in github workflows. 
What do you think about extracting that to a separate file and calling into it with interface like:
```
track_failures --run-id something
```
and here we can just parse the stdout.

This will also make manual testing easier. E.g. you can run the track_failures script locally against a few example failures and see the output is correct. I think including such evidence in PR description would also be helpful.

.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
...
+
+      - name: Create or update issues and build summary
+        id: issues
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.event.workflow_run.id }}
+          REPO: ${{ github.repository }}
+          COMMIT_SHA: ${{ github.event.workflow_run.head_sha }}
+          FAILING_TESTS: ${{ steps.parse.outputs.tests }}
```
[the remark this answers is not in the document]
IvoDD: This is probably fine because of the somewhat strict regex pattern matching, but still worth considering this claude's suggestion. I guess if we extract the above to bash script we can pass `--output-dir` to it where it will put all the text files needed for next steps.

.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
...
+              body="$body"$'\n\n'"This test has been detected as failing in CI."
+              body="$body"$'\n\n'"### Failure history"
+              body="$body"$'\n\n'"| Date | Run | Commit |"
+              body="$body"$'\n'"|------|-----|--------|"
+              body="$body"$'\n'"| $run_date | [Run #$RUN_ID]($RUN_URL) | \`${COMMIT_SHA:0:10}\` |"
+              body="$body"$'\n\n'"---"
+              body="$body"$'\n'"*This issue was automatically created by the flaky test tracker workflow.*"
+              body="$body"$'\n'"*Add further failure occurrences as comments below.*"
+
+              track_item "$test_name" "Flaky test" "flaky-test" "$body"
```
IvoDD: I think opening a separate issue for each test can be problematic.

Most probably if >10 tests fail together their failures are correlated. This can happen in cases:
- Someone accidentally breaks 1000 tests on their PR and fat fingers the button to track failure
- Some fixture or shared setup breaks and results in dozens of failed tests

I guess we can try to produce a single item if >10 failures.

.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
...
+              body="$body"$'\n\n'"This CI step has failed. This is typically caused by transient infrastructure issues (network timeouts, package mirrors, etc.)."
+              body="$body"$'\n\n'"### Failure history"
+              body="$body"$'\n\n'"| Date | Run | Commit |"
+              body="$body"$'\n'"|------|-----|--------|"
+              body="$body"$'\n'"| $run_date | [Run #$RUN_ID]($RUN_URL) | \`${COMMIT_SHA:0:10}\` |"
+              body="$body"$'\n\n'"---"
+              body="$body"$'\n'"*This issue was automatically created by the flaky test tracker workflow.*"
+              body="$body"$'\n'"*Add further failure occurrences as comments below.*"
+
+              track_item "$step_entry" "Flaky step" "flaky-step" "$body"
```
IvoDD: Some general gh action outage can cause a lot of steps to fail together. Grouping them if many makes sense.

.github/workflows/failure_notification.yaml
```diff
@@ -5,20 +5,315 @@ on:
...
+          } >> "$GITHUB_OUTPUT"
+
+  notify-slack:
+    runs-on: ubuntu-latest
+    needs: [track-failures]
+    if: >-
+      (github.event.workflow_run.head_branch == 'master' ||
+       github.event.workflow_run.head_branch == '') &&
+      always() &&
+      needs.track-failures.result != 'skipped'
```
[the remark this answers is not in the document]
IvoDD: I think we should indeed notify in case of track-failures failure (otherwise we'll be losing coverage of failures silently), but we should probably have a friendly error message in that case.

.github/workflows/failure_notification.yaml
```diff
@@ -1,24 +1,265 @@
...
+
+permissions:
+  contents: read
+  issues: write
+  actions: read
+  pull-requests: write
 
 jobs:
-  on-failure:
+  resolve-pr-run:
```
IvoDD, line 22: I'm ok with triggering via PR comments but I think it has a couple downsides:
- Pollutes the PR description page with things not related to the PR (e.g. flaky master tests)
- Can't easily trigger for a failure before the last

I guess I was thinking of adding a button to the job summary of a failed "Build and Test" or "Build with conda" or other like:
```
    failure-analysis-link:                                                                     
      needs: [cpp-test-linux, cpp-test-windows, cpp-test-macos, build-python-wheels-linux,     
              build-python-wheels-windows, build-python-wheels-macos,                          
              persistent_storage_verify_linux, persistent_storage_verify_windows]              
      if: always() && failure()                                                                
      runs-on: ubuntu-latest                                                                   
      steps:                                                                                   
        - name: Write failure analysis link to summary                                         
          run: |
            
  url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/workflows/failure_notification.yaml"  
            cat >> "$GITHUB_STEP_SUMMARY" <<'EOF'
            ## Failure Analysis                                                                
                             
            EOF
            echo "[**Analyse failures →**]($url)" >> "$GITHUB_STEP_SUMMARY"                    
            echo "" >> "$GITHUB_STEP_SUMMARY"
            echo "Run ID: \`${GITHUB_RUN_ID}\`" >> "$GITHUB_STEP_SUMMARY"
```

I don't feel strongly about this, but using a link like above would allow us to delete `reselov_pr_run.py` and simplify logic here

I guess using both activation mechanisms is also an option

.github/workflows/failure_notification.yaml
```diff
@@ -1,24 +1,265 @@
...
+          {
+            echo "slack_summary<<SLACK_EOF"
+            cat /tmp/ci_failures/slack_summary.txt
+            echo "SLACK_EOF"
+          } >> "$GITHUB_OUTPUT"
+
+  notify-slack:
+    runs-on: ubuntu-latest
+    needs: [track-failures]
+    # Only send Slack notifications for automatic master triggers, not manual dispatch
```
IvoDD, line 177: Could be argued that manual dispatch calls on PRs are still flakyness present on master.

Still probably better to keep the notifications just for master to avoid excessive noise.

.github/scripts/parse_ci_failures.py
```diff
@@ -0,0 +1,193 @@
...
+        "api", f"repos/{repo}/actions/jobs/{job_id}",
+        "--jq", '.steps[] | select(.conclusion == "failure") | .name',
+    )
+    return [s for s in raw.splitlines() if s.strip()]
+
+
+def download_failed_logs(repo: str, run_id: str, output_path: str) -> bool:
+    """Download logs for the failed run. Returns True on success."""
+    result = subprocess.run(
+        ["gh", "run", "view", run_id, "--repo", repo, "--log-failed"],
```
IvoDD: Nit: Can use `run_gh` method

.github/scripts/parse_ci_failures.py
```diff
@@ -0,0 +1,193 @@
...
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    with open(output_path, "w") as f:
+        f.write(result.stdout)
+    if result.returncode != 0:
```
IvoDD: Maybe we can add some error handling in the general `run_gh` instead of just this one `gh run view` command

.github/scripts/tests/test_parse_ci_failures.py
IvoDD, line 1: Are these tests ever ran on the CI? 
I think it makes sense to enforce they pass before merging a PR.

.github/scripts/send_slack_notification.py
```diff
@@ -0,0 +1,130 @@
...
+
+def send_to_slack(webhook_url: str, payload: dict) -> None:
+    """POST a JSON payload to the Slack webhook."""
+    data = json.dumps(payload).encode("utf-8")
+    req = urllib.request.Request(
+        webhook_url,
+        data=data,
+        headers={"Content-Type": "application/json"},
+        method="POST",
+    )
```
IvoDD, line 82: How does authentication against the slack webhook work? I see we in gh workflow we have the url itself as a secret. 

Any chance urllib exception will leak the url in logs or something similar? It seems like in exception handling we at least don't log the url which is good.

.github/workflows/failure_notification.yaml
```diff
@@ -1,24 +1,265 @@
...
+      (github.event.workflow_run.head_branch == 'master' ||
+       github.event.workflow_run.head_branch == '') &&
+      always() &&
+      needs.track-failures.result != 'skipped' &&
+      needs.track-failures.result != 'cancelled'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Send Slack notification
```
IvoDD, line 189: We were previously using an existing action `ravsamhq/notify-slack-action@be814b201e233b2dc673608aa46e5447c8ab13f2`. 
Any reason to write our own slack notification code instead of building the message and using the existing action?

.github/scripts/track_ci_issues.py
```diff
@@ -0,0 +1,365 @@
...
+"""
+import argparse
+import json
+import subprocess
+from datetime import datetime, timezone
+
+GROUPING_THRESHOLD = 10
+
+
+def run_gh(*args: str, check: bool = True) -> str:
```
IvoDD: This method is defined in more than one place, maybe we can add a common method in `utils.py`?

.github/scripts/track_ci_issues.py
```diff
@@ -0,0 +1,365 @@
...
+
+class IssueTracker:
+    """Tracks CI failures as GitHub issues and accumulates a Slack summary."""
+
+    def __init__(self, repo: str, run_id: str, run_url: str, commit_sha: str):
+        self.repo = repo
+        self.run_id = run_id
+        self.run_url = run_url
+        self.short_sha = commit_sha[:10]
+        self.run_date = datetime.now(timezone.utc).strftime("%Y-%m-%d %H:%M UTC")
```
IvoDD: The run might be e.g. a week old, so the time of report might be misleading. Would be useful to use the time of the failure so we can cross reference against e.g. github outages.

.github/scripts/tests/test_parse_ci_failures.py
G-D-Petrov: added a flow to test them

.github/workflows/failure_notification.yaml
G-D-Petrov: The ravsamhq/notify-slack-action doesn't support what we need. It uses Slack's legacy attachments format with a fixed schema — no Block Kit support and no way to pass an arbitrary payload. We'd lose the structured summary (per-test     
  lines, grouped failure messages, timeout indicators) and the tracker-failure warning.

.github/scripts/send_slack_notification.py
G-D-Petrov: Added extra handling for the URLError.

But in any case, since SLACK_WEBHOOK_URL is a repository secret, GitHub Actions automatically masks its value in all log output. Any occurrence of the secret's value gets replaced with ***. So even if it did leak through a traceback, it would be masked.

.github/workflows/failure_notification.yaml
G-D-Petrov: I think that I prefer the implemented approach a bit more because:
1. It doesn't require changes to "Build and Test", "Build with conda", etc
2. Makes flaky tests a bit more clear in the PR itself, thus easier to see for reviewers

.github/scripts/tests/test_utils.py
```diff
@@ -0,0 +1,58 @@
...
+            args=["gh", "api", "repos/owner/repo"],
+            returncode=1,
+            stdout="",
+            stderr="Bad credentials",
+        )
+        with pytest.raises(subprocess.CalledProcessError):
+            run_gh("api", "repos/owner/repo")
+        captured = capsys.readouterr()
+        assert "gh command failed" in captured.err
+        assert "Bad credentials" in captured.err
```
IvoDD, line 58: I feel like these tests are a bit too much for a simple utility to run a github command.

IvoDD (approved):

.github/scripts/tests/test_utils.py
ognyanstoimenov: +1

ognyanstoimenov (approved):
Once this is up and running and it's verified to be working as expected we should probably get rid of the old flaky test tracking. I.e. revert https://github.com/man-group/ArcticDB/pull/2862 and https://github.com/man-group/ArcticDB/pull/2571

Issue #2611, opened 2025-08-25
Metadata got deleted while using lib.defragment_symbol_data


himsheda:
### Describe the bug

I had a library with defragmanted symbols. each symbol had data and metadata. but after defragment_symbol_data api call its returning a versioned item with metadata = None and if you read the metadata from db then its also None.

its a serious bug.

I am using arcticdb 4.4.3

### Steps/Code to Reproduce

```python
store["library_name"].defragment_symbol_data("symbol")
```
Output:
```python
VersionedItem(symbol='symbol', library='library_name', data=n/a, version=232, metadata=None, host='S3(endpoint=uri, bucket=bucket-name)', timestamp=1756122299472736959)
```

### Expected Results
```python
VersionedItem(symbol='symbol', library='library_name', data=n/a, version=232, metadata={<metadata>}, host='S3(endpoint=uri, bucket=bucket-name)', timestamp=1756122299472736959)
```

### OS, Python Version and ArcticDB Version

Python: 3.8.12 (default, Oct 12 2021, 13:49:34) 
[GCC 7.5.0]
OS: Linux-5.15.0-140-generic-x86_64-with-glibc2.17
ArcticDB: 4.4.3

### Backend storage used

Minio S3

### Additional Context

Try on the symbols won which `lib.is_symbol_fragmented(symbol)` gives true

WatchTree-19:
Still reproduces on **6.23.0**, a year after @vasil-pashov confirmed it on 6.1.1. I had a look at the root cause, which appears to be a single argument.

### Reproduction

```python
from arcticdb.config import set_config_int
set_config_int("SymbolDataCompact.SegmentCount", 1)
lib = adb.Arctic("lmdb:///tmp/x").get_library("t", create_if_missing=True)

base = pd.Timestamp("2024-01-01")
lib.write("s", pd.DataFrame({"a": [0]}, index=[base]), metadata={"tag": "keep-me", "n": 0})
for k in range(1, 40):                                    # 1-row appends -> many tiny segments
    lib.append("s", pd.DataFrame({"a": [k]}, index=[base + pd.Timedelta(minutes=k)]),
               metadata={"tag": "keep-me", "n": k})

before = lib.read("s").metadata
lib.defragment_symbol_data("s")
after = lib.read("s").metadata
```

```
fragmented?        True
metadata before :  {'tag': 'keep-me', 'n': 39}
metadata after  :  None
rows before/after: 40 / 40
```

[...]

PR #2991, opened 2026-03-28
Fix defragment_symbol_data dropping user metadata (#9889436827)

himsheda:
defragment_symbol_data_impl passed std::nullopt to collate_and_write for user_meta, silently discarding the metadata that had been correctly read into pipeline_context->user_meta_ from the previous version. Fix by passing the actual user_meta_ptr from the pipeline context.

Also fix the Python wrapper which hardcoded metadata=None in the returned VersionedItem regardless of what was stored; it now reads the metadata back from the newly-written version using the existing read_metadata pattern.

Add regression test test_defragment_preserves_metadata.

#### Reference Issues/PRs
https://github.com/man-group/ArcticDB/issues/2611

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -4230,15 +4230,10 @@ class NativeVersionStore:
         )
 
         result = self.version_store.defragment_symbol_data(symbol, segment_size, prune_previous_version)
-        return VersionedItem(
-            symbol=result.symbol,
-            library=self._library.library_path,
-            version=result.version,
-            metadata=None,
-            data=None,
-            host=self.env,
-            timestamp=result.timestamp,
-        )
+        version_query = self._get_version_query(result.version)
+        _, udm = self.version_store.read_metadata(symbol, version_query)
+        meta = denormalize_user_metadata(udm, self._normalizer) if udm else None
+        return self._convert_thin_cxx_item_to_python(result, meta)
 
     def library(self):
         return self._library
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -4230,10 +4230,15 @@ class NativeVersionStore:
         )
 
         result = self.version_store.defragment_symbol_data(symbol, segment_size, prune_previous_version)
-        version_query = self._get_version_query(result.version)
-        _, udm = self.version_store.read_metadata(symbol, version_query)
-        meta = denormalize_user_metadata(udm, self._normalizer) if udm else None
-        return self._convert_thin_cxx_item_to_python(result, meta)
+        return VersionedItem(
+            symbol=result.symbol,
+            library=self._library.library_path,
+            version=result.version,
+            metadata=None,
+            data=None,
+            host=self.env,
+            timestamp=result.timestamp,
+        )
 
     def library(self):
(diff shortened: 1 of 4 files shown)


python/arcticdb/version_store/_store.py
```diff
@@ -4022,15 +4022,10 @@ def defragment_symbol_data(
...
-            symbol=result.symbol,
-            library=self._library.library_path,
-            version=result.version,
-            metadata=None,
-            data=None,
-            host=self.env,
-            timestamp=result.timestamp,
-        )
+        version_query = self._get_version_query(result.version)
+        _, udm = self.version_store.read_metadata(symbol, version_query)
```
phoebusm: The metadata should already available in the C++ layer at this point. Consider returning it directly to avoid extra API call and storage read.

phoebusm (approved):

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -697,6 +697,22 @@ def test_defragment_no_work_to_do(sym, lmdb_version_store):
         lmdb_version_store.defragment_symbol_data(sym)
 
 
+def test_defragment_preserves_metadata(sym, lmdb_version_store):
+    meta = {"key": "value", "number": 42}
+    df1 = pd.DataFrame({"a": [1, 2]}, index=pd.date_range("2020-01-01", periods=2))
+    df2 = pd.DataFrame({"a": [3, 4]}, index=pd.date_range("2020-01-03", periods=2))
+
+    lmdb_version_store.write(sym, df1, metadata=meta)
+    lmdb_version_store.append(sym, df2)
```
phoebusm: I think once `append()`, the metadata won't be maintained

phoebusm (changes requested):
https://github.com/man-group/ArcticDB/pull/2991#discussion_r3009123928
Returning the metadata directly from C++ maybe too much scope for this PR. It would be sufficient to just update the docstring in `Library` to document that the returned metadata will be `None`

python/tests/unit/arcticdb/version_store/test_append.py
himsheda: Thanks. Its fixed

poodlewars:
> > [#2991 (comment)](https://github.com/man-group/ArcticDB/pull/2991#discussion_r3009123928) Returning the metadata directly from C++ maybe too much scope for this PR. It would be sufficient to just update the docstring in `Library` to document that the returned metadata will be `None`
> 
> Handled. @phoebusm Please have a look

Why is there still the `version_core.cpp` change?

himsheda:
> > > [#2991 (comment)](https://github.com/man-group/ArcticDB/pull/2991#discussion_r3009123928) Returning the metadata directly from C++ maybe too much scope for this PR. It would be sufficient to just update the docstring in `Library` to document that the returned metadata will be `None`
> > 
> > 
> > Handled. @phoebusm Please have a look
> 
> Why is there still the `version_core.cpp` change?

Its because this is the bug this PR is solving. Current code not only returns none as metadata but also stores it. Causing meradata corruption.

PR #3000, opened 2026-03-31
Refactor conda build

markovskipetar:
# Conda CI Workflows

Builds and tests ArcticDB using conda-forge dependencies across Linux x64, Linux ARM64, macOS ARM64, and Windows x64.

## Workflow Files

| File | Role |
|------|------|
| `.github/workflows/build_with_conda.yml` | Parent dispatcher — fans out to four platform jobs, defines `can_merge_conda` gate |
| `.github/workflows/compile_and_test_with_conda.yml` | Unified reusable workflow for all platforms (unix + windows) |

## Job Structure

Each platform runs three jobs: **compile** → **cpp_tests** + **python_tests** (parallel).

- **compile**: micromamba env setup, sccache, `pip install -e .` with `ARCTICDB_USING_CONDA=1`, uploads build artifacts
- **cpp_tests**: builds and runs `test_unit_arcticdb` + `arcticdb_rapidcheck_tests` via ctest (skipped on Windows)
- **python_tests**: matrix of 6 test types (`unit`, `integration`, `hypothesis`, `stress`, `compat`, `enduser`) with MongoDB + Azurite


.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,318 @@
...
+        uses: ./.github/actions/set_persistent_storage_env_vars
+        with:
+          aws_access_key: "${{ secrets.AWS_S3_ACCESS_KEY }}"
+          aws_secret_key: "${{ secrets.AWS_S3_SECRET_KEY }}"
+          gcp_access_key: "${{ secrets.GCP_S3_ACCESS_KEY }}"
+          gcp_secret_key: "${{ secrets.GCP_S3_SECRET_KEY }}"
+          azure_container: "githubblob"
+          azure_connection_string: "${{ secrets.AZURE_CONNECTION_STRING }}"
+          persistent_storage: ${{inputs.persistent_storage}}
+
```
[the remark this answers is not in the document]
G-D-Petrov: This is okay, we now want to always test the arm related flow

G-D-Petrov (approved):

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,318 @@
...
+      run_commandline:           {required: false, type: string,  default: '',    description: Custom commandline to run before tests}
+      run_custom_pytest_command: {required: false, type: string,  default: '',    description: Custom pytest command or additional arguments}
+
+jobs:
+  # ==================== JOB 1: COMPILE ====================
+  compile:
+    name: Compile (${{inputs.platform}})
+    if: |
+      always() &&
+      !cancelled()
```
IvoDD: Why are there different files for `_unix` and `_windows`? 

My main problem with this is the big amount of duplication between the two, so in the future when we change something we might forget to do it for the other workflow as well.

E.g. see this `if: always() && !cancelled`. It is missing on the windows build. Is there a reason for this?

I appreciate for some things we reqire completely different step (e.g. for Install Azurite). We can have `Install Azurite (Unix)` step and `Install Azurite (Windows)` step and skip them depending on the OS. E.g. see [how we do this in build_steps.yml](https://github.com/man-group/ArcticDB/blob/1679d9a96edfcb496312fd4020f98c58dffa61f9/.github/workflows/build_steps.yml#L529-L545) for reference.

However looking at the diff between the two the differences are not big:
E.g. windows doesn't build C++ tests. We can just pass `run_cpp_tests=false` or use `if: matrix.os != "windows"`
There are many others but we can go together through the diff and decide what can be easily unified.

.github/workflows/compile_and_test_with_conda_win.yml
```diff
@@ -0,0 +1,237 @@
...
+
+  # ==================== JOB 2: PYTHON TESTS ====================
+  # Note: C++ tests (rapidcheck) are currently disabled on Windows due to a linking issue.
+  python_tests:
+    name: Python Tests (${{inputs.platform}}) - ${{matrix.type}}
+    needs: [compile]
+    if: |
+      always() &&
+      !cancelled() &&
+      !failure()
```
IvoDD: This is another place where we differ between unix and windows? Any particular reason?

Also I think `always() && !cancelled() && !failures()` is equivalent to the default `success()`? Might be better to leave it to the default in this case.

.github/workflows/compile_and_test_with_conda_win.yml
```diff
@@ -0,0 +1,237 @@
...
+        with:
+          node-version: '24'
+
+      - name: Install Azurite
+        shell: bash -elo pipefail {0}
+        run: |
+          npm install -g azurite
+          npm_config_prefix=$(npm config get prefix)
+          npm_bin_dir="$npm_config_prefix"
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
```
IvoDD: Do we need this if? This code should only be executed on windows.

.github/workflows/compile_and_test_with_conda_win.yml
```diff
@@ -0,0 +1,237 @@
...
+        with:
+          environment-file: environment-dev.yml
+          environment-name: arcticdb
+          init-shell: bash cmd.exe
+          cache-environment: true
+          cache-environment-key: conda-env-${{inputs.platform}}
+          post-cleanup: 'none'
+
+      - name: Install ArcticDB from artifacts
+        shell: cmd /C call {0}
```
IvoDD: Out of curiosity why have we decided to use a non `bash` shell? I think this makes integrating the windonws and unix variants harder.

.github/workflows/compile_and_test_with_conda_win.yml
```diff
@@ -0,0 +1,237 @@
...
+          export ARCTICDB_RAND_SEED=$RANDOM
+          export ARCTICDB_WARN_ON_WRITING_EMPTY_DATAFRAME=0
+          if [[ "$(echo "$ARCTICDB_PYTEST_ARGS" | xargs)" == *pytest* ]]; then
+            command="python -m $ARCTICDB_PYTEST_ARGS"
+            echo "Run custom pytest command: $command"
+            eval "$command"
+          else
+            cd python
+            python -m pytest --timeout=3600 -v --tb=line -n logical --dist worksteal \
+              -m "not lmdb" \
```
IvoDD: Why do we skip lmdb tests on windows?

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,350 @@
...
```
IvoDD: Can we remove the `if:` clause? The `compile` job no longer `needs` anything.

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,350 @@
...
+          environment-file: environment-dev.yml
+          environment-name: arcticdb
+          init-shell: bash
+          cache-environment: true
+          cache-environment-key: conda-env-${{inputs.platform}}
+          post-cleanup: 'none'
+
+      - name: Setup MSVC
+        if: runner.os == 'Windows'
+        uses: ilammy/msvc-dev-cmd@v1
```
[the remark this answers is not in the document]
IvoDD: Or just using a precise `@v1.13.0` is better and will be consistent with our other actions.

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,350 @@
...
+        if: runner.os == 'Windows'
+        shell: bash -l {0}
+        run: echo "LIB=$CONDA_PREFIX/Library/lib;$LIB" >> $GITHUB_ENV
+
+      - name: Build ArcticDB with conda
+        run: |
+          python -m pip install --no-build-isolation --no-deps --retries 3 --timeout 400 -v -e .
+        env:
+          ARCTICDB_USING_CONDA: 1
+          ARCTICDB_BUILD_CPP_TESTS: 1
```
IvoDD: Do we need the `ARCTICDB_BUILD_CPP_TESTS: 1`? We have a separate step to build the windows C++ test [here](https://github.com/man-group/ArcticDB/pull/3000/changes#diff-5d04a595a40d7bb17cfc5b1bcbdfeab23c2d9ae3e9b6b88827fa28f4a923006aR166)

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,350 @@
...
+  # ==================== JOB 2: C++ TESTS ====================
+  # C++ tests are disabled on Windows due to a linking issue with rapidcheck.
+  cpp_tests:
+    name: C++ Tests (${{inputs.platform}})
+    needs: [compile]
+    if: |
+      always() &&
+      !cancelled() &&
+      inputs.run_cpp_tests &&
+      !contains(inputs.platform, 'win')
```
IvoDD: Better to have a single place where we disable the cpp tests on windows. What do you think about removing this ` !contains(inputs.platform, 'win')` and keeping the `run_cpp_tests: false` in the `build_with_conda.yml` with the comment.

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,344 @@
...
+          cd cpp/out/${{inputs.preset}}-build/
+          ctest --output-on-failure
+        env:
+          ARCTICDB_USING_CONDA: 1
+
+  # ==================== JOB 3: PYTHON TESTS ====================
+  python_tests:
+    name: Python Tests (${{inputs.platform}}) - ${{matrix.type}}
+    needs: [compile]
+    if: |
```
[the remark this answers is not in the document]
IvoDD: We can probably remove the `if` altogether to use the default `success` behavior which will be good enough in our case

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,344 @@
...
+
+      - name: Install npm
+        uses: actions/setup-node@v6.1.0
+        with:
+          node-version: '24'
+
+      - name: Install Azurite
+        shell: bash -elo pipefail {0}
+        run: |
+          npm install -g azurite
```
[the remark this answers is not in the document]
IvoDD: The [simpler Azurite install](https://github.com/man-group/ArcticDB/blob/3743b5310c8a8fdf2b183db610f4aeb1ecea3fb0/.github/workflows/build_steps.yml#L556-L568) in the vcpkg build pipeline should be good enough here as well. What do you think about using the same here?

We could even extract it to it's own action similar to the [mongodb one](https://github.com/man-group/ArcticDB/blob/master/.github/actions/install_mongodb/action.yml)

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,350 @@
...
+      - name: Test with pytest
+        run: |
+          openssl version -d || true
+          ulimit -a || true
+          echo "Run commandline: $COMMANDLINE"
+          eval "$COMMANDLINE"
+          export ARCTICDB_RAND_SEED=$RANDOM
+          export ARCTICDB_WARN_ON_WRITING_EMPTY_DATAFRAME=0
+          if [[ "$(echo "$ARCTICDB_PYTEST_ARGS" | xargs)" == pytest* ]]; then
+            python -m pip install pytest-repeat setuptools wheel
```
IvoDD: The previous step has already installed the `pytest-repeat` dependency and `setuptools wheel` should be installed earlier during the build.

We can also extract the protobuf build before the if and else like:
```
          export ARCTICDB_WARN_ON_WRITING_EMPTY_DATAFRAME=0
          python setup.py protoc --build-lib python # Moved before the if/else
          if [[ "$(echo "$ARCTICDB_PYTEST_ARGS" | xargs)" == pytest* ]]; then
            echo "Run custom pytest command: $ARCTICDB_PYTEST_ARGS"
            eval "$ARCTICDB_PYTEST_ARGS"
          else
            cd python
            BASETEMP_ARG=""
            # Whatever we previously had
```

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -0,0 +1,349 @@
...
+            python setup.py protoc --build-lib python
+            echo "Run custom pytest command: $ARCTICDB_PYTEST_ARGS"
+            eval "$ARCTICDB_PYTEST_ARGS"
+          else
+            python setup.py protoc --build-lib python
+            cd python
+            BASETEMP_ARG=""
+            if [[ "$RUNNER_OS" == "Windows" ]]; then
```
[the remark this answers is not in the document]
IvoDD: It can now be windows. All of linux, macos and windows flow through this file

.github/workflows/compile_and_test_with_conda.yml
```diff
@@ -323,13 +310,11 @@ jobs:
           eval "$COMMANDLINE"
           export ARCTICDB_RAND_SEED=$RANDOM
           export ARCTICDB_WARN_ON_WRITING_EMPTY_DATAFRAME=0
+          python setup.py protoc --build-lib python # Moved before the if/else
```
IvoDD: Nit: The comment `Moved before the if/else` was meant to be for the example. I don't think this inline comment makes sense standalone.

IvoDD (approved):

.github/workflows/build_with_conda.yml
```diff
@@ -14,8 +14,6 @@ on:
     branches:
       - master
 
-  pull_request:
```
IvoDD, line 21: This appears to be removed by accident

IvoDD (changes requested):

IvoDD (approved):

PR #3017, opened 2026-04-09
Support bools with column stats

poodlewars:
#### Reference Issues/PRs
Monday: 11292569665

Use column stats for comparisons against bool columns, like:

```
q["int_col"] > 4) & q["bool_col"]
```

This may be useful if a block is full of True or False values only, although it is unclear how often it will help in practice. You can imagine a flag that might be set to False on old data and True on newer data to see that this might actually be quite useful.

This also supports more verbose forms like:

```
False == q["bool_col"]
```

### DIFF
--- a/cpp/arcticdb/pipeline/column_stats_dispatch.cpp
+++ b/cpp/arcticdb/pipeline/column_stats_dispatch.cpp
@@ -148,6 +148,32 @@ std::vector<StatsComparison> visit_binary_membership_stats(
     );
 }
 
+bool value_is_nan(const Value& val) {
+    if (is_floating_point_type(val.data_type())) {
+        return details::visit_type(val.data_type(), [&val]<typename TagType>(TagType) -> bool {
+            using RawType = TagType::raw_type;
+            if constexpr (std::is_floating_point_v<RawType>) {
+                return std::isnan(val.get<RawType>());
+            }
+            return false;
+        });
+    }
+    return false;
+}
+
+StatsComparison to_stats_comparison_for_boolean(const ColumnStatsValues& csv) {
+    return unary_boolean_stats(csv, OperationType::IDENTITY);
+}
+
+StatsComparison to_stats_comparison_for_boolean(const Value& val) {
+    util::check(
+            is_bool_type(val.data_type()),
+            "Expected bool type in to_stats_comparison_for_boolean, got {}",
+            get_user_friendly_type_string(val.descriptor())
+    );
+    return val.get<bool>() ? StatsComparison::ALL_MATCH : StatsComparison::NONE_MATCH;
+}
+
 StatsComparison binary_boolean_stats(StatsComparison left, StatsComparison right, OperationType operation) {
     switch (operation) {
     case OperationType::AND:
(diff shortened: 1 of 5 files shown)


cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -378,6 +378,8 @@ struct LessThanOperator {
     bool operator()(T t, U u) const {
         return t < u;
     }
+    // MSVC warns about using < on bool (C4804). Cast to int to suppress.
+    bool operator()(bool t, bool u) const { return static_cast<int>(t) < static_cast<int>(u); }
```
alexowens90: I assume `false < true` is the test that should result in returning `true`? In which case:
```suggestion
    bool operator()(bool t, bool u) const { return !t && u; }
```
and similar for the other methods

cpp/arcticdb/pipeline/column_stats_dispatch.cpp
```diff
@@ -148,6 +148,32 @@ std::vector<StatsComparison> visit_binary_membership_stats(
...
+        return details::visit_type(val.data_type(), [&val]<typename TagType>(TagType) -> bool {
+            using RawType = TagType::raw_type;
+            if constexpr (std::is_floating_point_v<RawType>) {
+                return std::isnan(val.get<RawType>());
+            }
+            return false;
+        });
+    }
+    return false;
+}
```
alexowens90: I assume some kind of rebase fail from the nan-handling branch?

cpp/arcticdb/pipeline/column_stats_dispatch.cpp
```diff
@@ -173,36 +199,109 @@ StatsComparison binary_boolean_stats(StatsComparison left, StatsComparison right
...
-                        for (size_t i = 0; i < l.size(); ++i) {
-                            result.push_back(binary_boolean_stats(l.at(i), r.at(i), operation));
-                        }
-                        return result;
                     },
-                    [&](const auto&, const auto&) -> std::vector<StatsComparison> {
-                        size_t sz = std::max(stats_variant_size(left), stats_variant_size(right));
-                        log::version().warn("Unsupported case in visit_binary_boolean_stats");
-                        return std::vector(sz, StatsComparison::UNKNOWN);
+                    // StatsComparison x bool Value
```
alexowens90: We could short-circuit these as we do some others:
https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/version_store/processing.py#L153-L175
https://github.com/man-group/ArcticDB/blob/master/python/tests/unit/arcticdb/version_store/test_filtering.py#L1215
e.g. `q["col"] == True` is the same as `q["col"]`

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -1386,3 +1388,190 @@ def test_column_stats_empty_stats(in_memory_version_store, column_stats_filterin
...
+
+
+@pytest.mark.parametrize(
+    "query_expr",
+    [
+        lambda q: q["b"] & True,
+        lambda q: q["b"] | False,
+        lambda q: True & q["b"],
+        lambda q: False | q["b"],
+    ],
```
alexowens90, line 1474: These get short-circuited out on the Python layer, so don't really test anything new here

alexowens90 (approved):

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -1386,3 +1388,165 @@ def test_column_stats_empty_stats(in_memory_version_store, column_stats_filterin
...
+    assert_frame_equal(expected, result)
+    assert table_data_reads == expected_reads, f"Expected {expected_reads} TABLE_DATA reads, got {table_data_reads}"
+
+
+@pytest.mark.parametrize(
+    "query_expr,expected_reads",
+    [
+        pytest.param(lambda q: (q["int_col"] > 4) & q["bool_col"], 1, id="comparison_and_bool_col"),
+        pytest.param(lambda q: q["bool_col"] & (q["int_col"] > 4), 1, id="bool_col_and_comparison"),
+        pytest.param(lambda q: (q["int_col"] > 4) & q["bool_col"], 1, id="bool_col_and_comparison_flipped"),
```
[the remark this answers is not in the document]
vasil-pashov: Not sure about the intention but claude is right that these are the same

vasil-pashov (approved):

cpp/arcticdb/pipeline/column_stats_dispatch.cpp
poodlewars: Yep. Removed.

python/tests/unit/arcticdb/test_column_stats_optimisation.py
poodlewars: Cool - I'll leave as-is just so we have some regression testing.

PR #3018, opened 2026-04-09
ASV benchmarks for column stats

poodlewars:
Add benchmarks that run with and without column stats, on both ordered data (that should benefit a lot) and unordered data (that should not benefit). The unordered data helps us to measure the overhead of evaluating column stats when it cannot provide any benefit.

Since we're still waiting to merge #2958 the runs with and without column stats should be identical to start with. This will also help give us some nice evidence when we rebase and merge #2958.

Evidence of stability from 3 runs:

```
  | Case                              | Run 1   | Run 2   | Run 3   |   CV |
  |-----------------------------------|---------|---------|---------|------|
  | uint64 1M stats=T ordered=T      | 5.97ms  | 5.37ms  | 5.37ms  | 6.2% |
  | uint64 1M stats=F random=F       | 3.67ms  | 3.71ms  | 3.97ms  | 4.3% |
  | uint64 10M stats=T ordered=T     | 13.2ms  | 11.9ms  | 12.4ms  | 5.2% |
  | uint64 10M stats=F random=F      | 16.8ms  | 16.0ms  | 17.2ms  | 3.7% |
  | combined_or 1M stats=T ordered=T | 7.93ms  | 7.38ms  | 8.35ms  | 6.2% |
  | combined_or 10M stats=T ordered=T| 23.5ms  | 22.7ms  | 24.0ms  | 2.8% |
  | combined_or 10M stats=F random=F | 22.7ms  | 22.3ms  | 24.0ms  | 3.9% |
```

### DIFF
--- a/python/benchmarks/column_stats.py
+++ b/python/benchmarks/column_stats.py
@@ -1,0 +1,318 @@
+"""
+Copyright 2026 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+import time
+
+import numpy as np
+import pandas as pd
+from arcticdb import QueryBuilder
+from arcticdb_ext import set_config_int, unset_config_int
+from asv_runner.benchmarks.mark import SkipNotImplemented
+
+from benchmarks.common import *
+from arcticdb.util.logger import get_logger
+from benchmarks.environment_setup import (
+    Storage,
+    create_libraries_across_storages,
+    is_storage_enabled,
+)
+
+STORAGES = [Storage.LMDB, Storage.AMAZON]
+
+COLUMN_STATS = {
+    "uint64_col": {"MINMAX"},
+    "float_col": {"MINMAX"},
+    "bool_col": {"MINMAX"},
+    "datetime_col": {"MINMAX"},
+}
+
+
+def _symbol_name(rows, ordered):
+    suffix = "ordered" if ordered else "random"
+    return f"col_stats_{rows}_{suffix}"
+
+
+def _generate_column_stats_dataframe(n):
+    """Generate a DataFrame with ordered data suitable for column stats benchmarking.
+
+    Data is monotonically ordered so that each storage segment has a tight min/max range,
+    allowing column stats to effectively prune segments during filtered reads.
+    """
+    timestamps = pd.date_range(end="1/1/2023", periods=n, freq="s")
+    # Monotonically increasing datetime column spanning a separate range from the index
+    datetime_start = pd.Timestamp("2000-01-01")
+    datetime_col = pd.date_range(start=datetime_start, periods=n, freq="s")
+    df = pd.DataFrame(
+        {
+            "uint64_col": np.arange(n, dtype=np.uint64),
+            "float_col": np.linspace(0.0, 1000.0, n),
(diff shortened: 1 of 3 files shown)


python/benchmarks/column_stats.py
```diff
@@ -0,0 +1,318 @@
...
+    def time_isnotin_uint64(self, *args):
+        if not self.data_ordered:
+            # Not that interesting - we already run isin in the unordered case
+            raise SkipNotImplemented
+        q = QueryBuilder()
+        q = q[q["uint64_col"].isnotin(self.isin_values)]
+        self.lib.read(self.symbol, columns=["uint64_col"], query_builder=q)
+
+
+class ColumnStatsCreate:
```
alexowens90: Is this and `ColumnStatsManagement` worth benchmarking at this point? I think the current implementation will get ditched, and most stats will be generated at modification time, not through these explicit calls

python/benchmarks/column_stats.py
```diff
@@ -0,0 +1,318 @@
...
+        self.symbol = None
+
+    def setup_cache(self):
+        start = time.time()
+        lib_for_storage = create_libraries_across_storages(self.storages)
+        for rows in ColumnStatsQueryPerformance.num_rows:
+            for ordered in ColumnStatsQueryPerformance.data_ordered:
+                df = _generate_column_stats_dataframe(rows) if ordered else _generate_random_dataframe(rows)
+                sym = _symbol_name(rows, ordered)
+                for storage in ColumnStatsQueryPerformance.storages:
```
alexowens90: nit: This loop could go on the outside to avoid generating dfs that get binned

python/benchmarks/column_stats.py
```diff
@@ -0,0 +1,318 @@
...
+        self.lib.read(self.symbol, columns=["uint64_col"], query_builder=q)
+
+    def time_filter_float(self, *args):
+        q = QueryBuilder()
+        q = q[q["float_col"] > 900.0]
+        self.lib.read(self.symbol, columns=["float_col"], query_builder=q)
+
+    def time_filter_bool(self, *args):
+        q = QueryBuilder()
+        q = q[q["bool_col"] == True]
```
alexowens90: nit: You don't need the `== True`

alexowens90 (approved):

python/benchmarks/column_stats.py
poodlewars: I think the APIs will survive - we will still need a way to create stats over old data that isn't being modified any more.

python/benchmarks/column_stats.py
```diff
@@ -0,0 +1,575 @@
...
+    df.index = timestamps
+    df.index.name = "ts"
+    return df
+
+
+class ColumnStatsQueryPerformance:
+    """Benchmark query filtering performance with and without column stats enabled.
+
+    Runs against two data distributions:
+    - ordered: sorted so segment-level min/max stats allow pruning
```
vasil-pashov, line 108: I think this comment is slightly misleading. If I understood the implementation correctly, having ordered columns is not a prerequisite for taking advantage of column stats. It will indeed trigger pruning but it's a stronger requirement. The actual requirement is that we don't have completely random distribution allowing for each value to appear in each segment. I don't imply we change the benchmark implementation, we can rather change the comment to reflect that having ordered column is the easiest way to make sure pruning will happen.

vasil-pashov (approved):

python/benchmarks/column_stats.py
poodlewars: I think the comment is OK tbh - it isn't claiming that ordering is the only property that allows pruning.

PR #3021, opened 2026-04-09
Implement merge update for DataFrames using Multiindex

vasil-pashov:
#### Reference Issues/PRs
Monday:  https://man312219.monday.com/boards/7852509418/views/168855452

This is separated in two commits that can be reviewed separately.

**The actual implementation**

https://github.com/man-group/ArcticDB/commit/225918744b9223efd85ec5fc5bf80bc9eb080445 Multiindexes do not require any significant changes to the C++ logic. If the primary index is datetime they will act as if the a dataframe with DatetimeIndex is passed and follow the same rues for merge updates. In all other cases merge updates will act as if a dataframe with a RowrangeIndex is passed and it'll follow those rules. The change in C++ code is to handle all cases when columns with duplicate names appear. This is ambiguity and the decision is to throw. The commit also adds unit tests.

There are two ways we mangle columns.
1. The secondary indexes of panda Multiindex get an `__idx__ `prefix
2. If a data column is repeated more than once it becomes ` __col_<name>_<position_in_the_descriptor>`

There are a few possible combinations:

1. When datetime index is used never allow `on` to contain column named as the datetime index
2. If RowRange is used it's allowed to have a column in the dataframe named as the index because the index is not stored in the descriptor.
3. The primary index is never mangled. We must check if there's a mangled secondary index whose demangled name is the same as the primary, or if there's a data column named the same way as the primary index
4. The secondary indexes are mangled. We must check if there's a data column named the same way as the demangled name. We must also check if there's a data column named as the demangled secondary index but the data column is repeated.

**Make tests use explicit expected values**
https://github.com/man-group/ArcticDB/pull/3021/changes/11ae847f0cb2438e165b0d099389abbc638b0fc8
Working on this I had to make the oracle function used in hypothesis handle multiindex cases. Unlike most of the functions from the processing pipeline it doesn't have a direct alternative in pandas. The code became complicated on its own. So I find it both easier to read and safer if the unit tests have explicit expected results. This commit makes the generic test function takes the expected result. It test both the oracle and the C++ implementation.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -1768,3 +1756,253 @@ def test_throws_when_multiple_source_rows_match_same_target_row_via_nan(self, lm
         lib.write("sym", target)
         with pytest.raises(UserInputException, match="Multiple source rows match the same target row"):
             lib.merge_experimental("sym", source, strategy=self.strategy, on=["a"])
+
+    def test_on_column_named_same_as_index(self, lmdb_library):
+        """Row-range index name is protobuf-only and does not collide with data columns.
+        on=["col"] should match the data column and work without error."""
```
IvoDD, line 1762: All of these behaviors should be documented in the docstring of `merge_experimental`

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -1768,3 +1756,253 @@ def test_throws_when_multiple_source_rows_match_same_target_row_via_nan(self, lm
...
+        [
+            (["col", "other"], ["col"]),  # first level name matches data col
+            (["cat", "col"], ["col"]),  # secondary level matches data col
+        ],
+        ids=["first_level-dup_data", "secondary_level-dup_data"],
+    )
+    def test_duplicate_col_in_on_level_matches_data_raises(self, lmdb_library, is_datetime, index_names, data_cols):
+        # For rowrange, the first MI level becomes the row-range index and C++ does not
+        # check its name against data columns. This is a known bug.
+        level_values = [["A", "B", "C"], [1, 2, 3], [10, 20, 30]]
```
IvoDD, line 1838: Why is `len(level_values) != len(index_names)`?

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -1768,3 +1756,253 @@ def test_throws_when_multiple_source_rows_match_same_target_row_via_nan(self, lm
...
+        on = []
+        if is_datetime:
+            on.append("date")
```
IvoDD, line 1841: In this case `on == ["date"]`, why should that raise in the end? I think I'm just quite confused by this test

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -1768,3 +1756,253 @@ def test_throws_when_multiple_source_rows_match_same_target_row_via_nan(self, lm
...
+    def test_on_index_column(self, lmdb_library, is_datetime):
+        """on contains a column that is part of the multiindex (non-first level)."""
+        target, source = self._make_mi_frames(is_datetime)
+        generic_merge_test(lmdb_library, "sym", target, source, self.strategy, on=["a"])
+
+    @pytest.mark.parametrize("is_datetime", [True, False], ids=["datetime", "rowrange"])
+    def test_on_data_column(self, lmdb_library, is_datetime):
+        """on contains a column that is a data column, not in the multiindex."""
+        target, source = self._make_mi_frames(is_datetime)
+        generic_merge_test(lmdb_library, "sym", target, source, self.strategy, on=["c"])
```
IvoDD, line 1981: `"c"` would not match anything. Theoretically a bug with `c` column munging could slip through if it also ends up updating nothing. What do you think about making `c` have a match with target and only `d` being completely different so the update is not a no-op?

docs/mkdocs/docs/notebooks/ArcticDB_merge.ipynb
```diff
@@ -1936,18 +1936,59 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "7m0t30gjnq",
+   "source": "## MultiIndex Support\n\nMerge also works with `pd.MultiIndex` DataFrames. The behavior depends on the **first level** of the MultiIndex:\n\n- **Datetime first level**: The first level is used for timestamp-based matching (just like a regular `DatetimeIndex`). The remaining index levels are treated as regular columns — they can be updated by the merge or included in `on` for additional matching.\n- **Non-datetime first level**: The DataFrame is treated like a row-range-indexed DataFrame. All index levels become regular columns internally. The `on` parameter is **required** and matching is done purely on column values.\n\n### Example: MultiIndex with Datetime First Level\n\nWhen the first level of the MultiIndex is a datetime, merge matches on the datetime values. All other index levels and data columns are eligible for update.",
```
IvoDD, line 1942: I think `All index levels become regular columns internally` is confusing for users. I think a more simple explanation:
```
If first index is a datetime column it is automatically included in `on`. No other index levels are included in `on` by default. An empty `on` clause is not valid, so if first index is not datetime `on` must be explicitly specified.
```

cpp/arcticdb/processing/clause.cpp
```diff
@@ -348,6 +347,98 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
     return target_values;
 }
 
+size_t field_index_for_matching_on_column(std::string_view name, const StreamDescriptor& descriptor) {
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            descriptor.index().type() != IndexDescriptor::Type::TIMESTAMP || descriptor.field(0).name() != name,
+            "Column \"{}\" specified in the 'on' parameter cannot have the same name as the timestamp index column. "
```
IvoDD, line 353: I think this is a confusing error message for a user who does not have a duplicate column but just misunderstands that `on` by default includes the timestamp index.

E.g. user tries to `lib.merge(source, on=["ts"])` for a table with columns:
```
index = "ts"
columns = ["a", "b", "c"]
```

cpp/arcticdb/processing/clause.cpp
```diff
@@ -348,6 +347,98 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+            "Descriptor: {}",
+            name,
+            descriptor
+    );
+    const std::string repetition_prefix = fmt::format("__col_{}__", name);
+    const std::string mangled_name = stream::mangled_name(name);
+    auto field_it = std::ranges::find_if(descriptor.fields(), [&](const Field& field) {
+        return field.name() == name || field.name() == mangled_name || field.name().starts_with(repetition_prefix);
```
IvoDD, line 361: How would users pass `on` for an unnamed multiindex level? Is it just not possible

cpp/arcticdb/processing/clause.cpp
```diff
@@ -348,6 +347,98 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+    });
+    if (field_it == descriptor.fields().end()) {
+        user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>(
+                "Error while performing merge update with matching on columns. The \"on\" parameter contains a column "
+                "named \"{}\", but no such column exists in the following descriptor: {}",
+                name,
+                descriptor
+        );
+    }
+    if (field_it->name() == mangled_name) {
```
IvoDD, line 371: Wouldn't this be a bit simpler if:
```
    auto field_it = std::ranges::find_if(descriptor.fields(), [&](const Field& field) {
        return field.name() == name || field.name() == mangled_name || field.name().starts_with(repetition_prefix);
    });
   if (field_it == descriptor.fields().end()) { raise }
   auto repeated_field_it = field_it;
   ++repeated_field_it;
   if (repeated_field_it != descriptor.fields().end()) {
       repeated_field_it = std::find_if(repeated_field_it, descriptor.fields().end(), [&](const Field& field) {
           return field.name() == name || field.name() == mangled_name || field.name().starts_with(repetition_prefix);
       });
   }
   if (repeated_field_it != end()) {raise duplicate names}
   return std::distance(begin(), field_it);
```

It also doesn't rely on the ordering we currently expect to see (mangled_name, then regular names or repeated names)

cpp/arcticdb/processing/clause.cpp
```diff
@@ -348,6 +347,98 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+        // Pandas multiindex and the column appears as a secondary index column of the multiindex
+        const auto name_or_repeated_it = std::find_if(field_it, descriptor.fields().end(), [&](const Field& field) {
+            return field.name().starts_with(repetition_prefix) || field.name() == name;
```
IvoDD, line 374: How do names get mangled if we have repeated multiindex column names? we are not searching for a second repetition of the mangled name here.

python/arcticdb/util/test.py
```diff
@@ -1442,6 +1451,52 @@ def _apply_updates(
...
 
 
+def _merge_update_multiindex(target: pd.DataFrame, source: pd.DataFrame, on: List[str]) -> pd.DataFrame:
+    """Flatten MultiIndex, delegate to merge_update, reconstruct MultiIndex."""
+    index_names = list(target.index.names)
+    data_columns = list(target.columns)
+    overlap = [col for col in data_columns if col and col in index_names]
+    if len(overlap) > 0:
+        raise ValueError(
+            f"Cannot perform merge_update on a DataFrame with MultiIndex columns that overlap with index. Overlap: {overlap}"
```
IvoDD, line 1461: This is slightly different to the C++ behavior. This will raise if source has duplicate columns. C++ will raise only if the duplicate column is in the `on`

python/arcticdb/util/test.py
```diff
@@ -1468,6 +1523,9 @@ def merge_update(target: pd.DataFrame, source: pd.DataFrame, on: Optional[List[s
     """
     on = _validate_merge_update_inputs(target, source, on)
 
+    if isinstance(target.index, pd.MultiIndex):
+        return _merge_update_multiindex(target, source, on)
```
IvoDD, line 1527: Do we need the entirely separate method for multiindex? Would it just work if we:
1. Reset_index
2. Perform merge as before
3. Reconstruct index

cpp/arcticdb/processing/clause.cpp
```diff
@@ -348,6 +347,98 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
 }
 
+size_t field_index_for_matching_on_column(std::string_view name, const StreamDescriptor& descriptor) {
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            descriptor.index().type() != IndexDescriptor::Type::TIMESTAMP || descriptor.field(0).name() != name,
+            "Column \"{}\" specified in the 'on' parameter cannot have the same name as the timestamp index column. "
+            "Descriptor: {}",
+            name,
+            descriptor
+    );
```
alexowens90, line 357: Won't this throw if the user has an unnamed timeseries, plus an actual column called "index" that they are trying to merge on?

cpp/arcticdb/processing/clause.cpp
alexowens90: I don't think it should be

cpp/arcticdb/processing/clause.cpp
```diff
@@ -348,6 +347,98 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+    const std::string repetition_prefix = fmt::format("__col_{}__", name);
+    const std::string mangled_name = stream::mangled_name(name);
+    auto field_it = std::ranges::find_if(descriptor.fields(), [&](const Field& field) {
```
alexowens90, line 360: There is also this mangled pattern:
https://github.com/man-group/ArcticDB/blob/master/python/arcticdb/version_store/_normalization.py#L963
Although seems to be for an index column named `None`, so might not matter

cpp/arcticdb/processing/clause.cpp
```diff
@@ -348,6 +347,98 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
     return target_values;
 }
 
+size_t field_index_for_matching_on_column(std::string_view name, const StreamDescriptor& descriptor) {
```
alexowens90, line 350: I think this function would be easier to reason about if it looked something like:
```
[name_it, multiindex_mangled_it, repetition_it] = linear_search(fields)
if (auto error_msg = construct_error_message(name_it, multiindex_mangled_it, repetition_it, fields.end()); error_msg.has_value()) {
    user_input::raise<...>(*error_msg)
}
```
I think we also only really need 2 error messages:

1. Specified on column doesn't exist
2. Specified on column is duplicated
The details of where those columns are isn't really important

python/tests/unit/arcticdb/version_store/test_merge_update.py
vasil-pashov: Interesting that pandas didn't mind that at all. It cut the unused elements.

python/tests/unit/arcticdb/version_store/test_merge_update.py
vasil-pashov: Reworked

python/tests/unit/arcticdb/version_store/test_merge_update.py
vasil-pashov: makes sense

cpp/arcticdb/processing/clause.cpp
vasil-pashov: I don't think we should allow matching on unnamed columns

cpp/arcticdb/processing/clause.cpp
vasil-pashov: It throws on write

python/arcticdb/util/test.py
vasil-pashov: It's basically what the code does. The main gotcha is that pandas can and will change types when reindexing so the code has to
1. Store the proper dtypes
2. After reset index check and fix dtypes if necessary
3. Perform merge as before
4. Reconstruct the index via reset index
5. Fix dtypes if necessary

python/arcticdb/util/test.py
vasil-pashov: Fixed

cpp/arcticdb/processing/clause.cpp
vasil-pashov: Yeah, I don't think we should be able to match on columns named None. I'll add a test

cpp/arcticdb/version/version_core.cpp
```diff
@@ -317,6 +317,22 @@ void compact_row_slices(std::span<SliceAndKey> slices) {
         previous_col_slice_end = slice.slice().col_range.end();
     }
 }
+
+bool is_fake_index_name(const arcticc::pb2::descriptors_pb2::NormalizationMetadata& normalization_metadata) {
+    if (normalization_metadata.has_df() && normalization_metadata.df().has_common()) {
```
alexowens90: I think `has_common` will always be true?
Also need to check for Series

alexowens90 (approved):

python/arcticdb/util/test.py
IvoDD: Ah I think I missed the mutual recursion between`merge_update` and `_merge_update_multiindex`.

Anyway makes sense to leave 1, 2, 4 and 5 in a separate function

cpp/arcticdb/processing/clause.cpp
```diff
@@ -2676,4 +2582,44 @@ bool MergeUpdateClause::is_update_only() const {
...
+                    (descriptor.field(0).name() != name || (fake_index_name_ && descriptor.field(0).name() == name)),
+            "The \"on\" parameter must not contain the datetime index column \"{}\". Note that for date time indexed "
+            "column the index column is always used for matching. Descriptor: {}",
+            name,
+            descriptor
+    );
+    const boost::regex repetition_regex{fmt::format("^__col_{}__\\d+$", name)};
+    const std::string mutliindex_mangled_name = stream::mangled_name(name);
+    std::ranges::subrange data_fields{
+            std::next(descriptor.fields().begin(), descriptor.index().field_count()), descriptor.fields().end()
```
IvoDD, line 2623: Nit: Maybe worth noting that in case of multiindex we still store a single index column in `descriptor.index()`

IvoDD (approved):

PR #3023, opened 2026-04-10
Support preloaded index query with column stats

poodlewars:
#### Reference Issues/PRs
Monday 11526152128

Get rid of the `PreloadedIndexQuery` in `VersionIdentifier`, just use `IndexInformation` everywhere.

Kept the `PreloadedIndexQuery` around as its fields are const. The segments in the `IndexInformation` get moved from and cannot be marked const. I think it also helps to create the `IndexInformation` by cloning out of the (single) `PreloadedIndexQuery`, rather than creating a new `PreloadedIndexQuery` when its segments need cloning.


cpp/arcticdb/pipeline/query.hpp
```diff
@@ -40,13 +40,17 @@ struct SpecificVersionQuery {
...
         index_key_(std::move(index_key)),
-        index_seg_(std::move(index_seg)) {}
+        index_seg_(std::move(index_seg)),
+        column_stats_seg_(std::move(column_stats_seg)) {}
 
     // Key is just needed for constructing the VersionedItem to return
     const AtomKey index_key_;
     const SegmentInMemory index_seg_;
+    const std::optional<SegmentInMemory> column_stats_seg_;
 };
```
alexowens90, line 55: Worth adding a comment here and to `IndexInformation` similar to this PR description about why both are useful despite their similarity

alexowens90 (approved):

alexowens90 (changes requested):
Do not read column stats key unless configs map setting is truthy

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -436,7 +436,14 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
     const auto identifier = util::variant_match(
             version_query.content_,
             [&](const std::shared_ptr<PreloadedIndexQuery>& preloaded_index_query) -> VersionIdentifier {
-                return {preloaded_index_query};
+                // Clone because Python holds the PreloadedIndexQuery across multiple collect() calls.
```
vasil-pashov, line 440: For my own understanding. What goes wrong if we don't clone?

python/tests/unit/arcticdb/version_store/test_collect_schema.py
```diff
@@ -249,6 +269,8 @@ def test_collect_schema_and_collect_multiple_times(mem_library):
         assert query_stats_operation_count(stats, "Memory_GetObject", "VERSION_REF") == 1
         assert query_stats_operation_count(stats, "Memory_GetObject", "TABLE_INDEX") == 1
         assert query_stats_operation_count(stats, "Memory_GetObject", "TABLE_DATA") == 0
+        if create_column_stats:
```
vasil-pashov, line 272: I think it makes sense to also test
```
query_stats_operation_count(stats, "Memory_GetObject", "COLUMN_STATS") == 0
```
when create_column_stats is False

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -508,46 +515,67 @@ VersionedItem LocalVersionedEngine::read_modify_write_internal(
...
+                                }
+                            }
+                    );
+                }
+                DescriptorItem descriptor_item{
+                        std::move(key), start_index, end_index, std::move(timeseries_descriptor)
+                };
+                if (include_index_segment) {
+                    descriptor_item.index_segment_ = std::move(seg);
+                    if (!column_stats_try.hasException()) {
```
vasil-pashov, line 575: Should we log a warning if the column stats feature holds an exception?

vasil-pashov (approved):
LGTM given @alexowens90 comment https://github.com/man-group/ArcticDB/pull/3023#pullrequestreview-4155311227 is resolved

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: I'm not sure to be honest - just preserving the comment that was here before.

python/tests/unit/arcticdb/version_store/test_collect_schema.py
poodlewars: Yep, changing this exposes the bug that Alex spotted.

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: I want to make sure that we're quiet if the key is missing, and don't want to add the noise of verifying that the exception is a `KeyNotFoundException`. This did make me realise we need these,

```
```

around the column stats reads.

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: So you can call `collect()` from Python repeatedly with the same lazy dataframe

alexowens90 (approved):

PR #3024, opened 2026-04-10
Fix v2 encoding bugs exposed by column stats

poodlewars:
#### Reference Issues/PRs
11661249607

Add v2 encoding tests for column stats, and fix two bugs that it exposed. See comments in the diff for explanations of the two bugs.

#### Any other comments?

Only use v2 encoding for the creation tests and a simple optimisation test. Don't see a reason to run every optimisation test twice.


cpp/arcticdb/codec/codec.cpp
```diff
@@ -355,7 +355,7 @@ std::pair<std::optional<google::protobuf::Any>, StreamDescriptor> decode_metadat
 
     auto maybe_any = decode_metadata(hdr, data, begin);
     if (EncodingVersion(hdr.encoding_version()) == EncodingVersion::V2)
-        util::check_magic<DescriptorFieldsMagic>(data);
+        skip_descriptor(data, hdr);
```
poodlewars, line 358: decode_metadata_and_descriptor_fields in codec.cpp decodes metadata from a v2-encoded segment and then validates the next magic number. The v2 wire layout is:
```
  MetadataMagic | metadata | SegmentDescriptorMagic | SegmentDescriptor | identifier | DescriptorFieldsMagic | ...
```

  After decoding metadata, the function checked for DescriptorFieldsMagic immediately. But data actually points at SegmentDescriptorMagic at that point — the segment descriptor and identifier sit between them. So the check read 'S','D','s','c' (SegmentDescriptorMagic) where it expected 'D','e','s','c' (DescriptorFieldsMagic).

  The fix replaces the bare check_magic<DescriptorFieldsMagic>(data) with skip_descriptor(data, hdr), a function that already exists at line 262 of the same file. skip_descriptor advances data past the segment descriptor magic, the descriptor struct, the identifier, and the descriptor fields magic — which is exactly the gap that was being skipped over.

cpp/arcticdb/entity/field_collection.cpp
```diff
@@ -42,6 +42,7 @@ void FieldCollection::regenerate_offsets() {
         ++shape_ptr;
         ++offset_ptr;
     }
+    offsets_.commit();
```
poodlewars, line 45: When we call `create_column_stats` we attempt to merge together with any stats that already exist. When we load the existing segment we call `regenerate_offsets()`. Without this `commit()` we leave `pos_` at 0 which then breaks the `add_field` calls we make when we merge in the new statistics.

cpp/arcticdb/codec/codec.cpp
poodlewars, line 358: This only affects, `test_column_stats_header_metadata` and the library tool:

```
    # Verify descriptor field names match the offsets
    lib_tool = lib.library_tool()
    keys = lib_tool.find_keys_for_symbol(KeyType.COLUMN_STATS, sym)
    fields = lib_tool.read_descriptor(keys[0]).fields()
    for _, entry in header_all_entries(header):
        field_name = fields[entry.stats_seg_offset].name
        if entry.type == ColumnStatsType.MIN_V1:
            assert field_name == "v1_MIN(col_1)"
        else:
            assert field_name == "v1_MAX(col_1)"
```

and this test fails with: `E_ASSERTION_FAILURE Magic number failure, expected 1668506948(Desc) got 1668498515`. This means we're on `SegmentDescriptorMagic` but expect to be on `DescriptorFieldsMagic`.

The layout is:

```memory_layout.hpp
...
    SegmentDescriptorMagic segment_descriptor_magic_;
    SegmentDescriptor segment_descriptor_;
    SegmentIdentifierHeader identifier_header_;
    Optional<OpaqueField> identifier_data_;
    DescriptorFieldsMagic descriptor_magic_;
    CompressedField<FieldList> descriptor_fields_;
...
```

so we just need to skip the three fields in between, which is exactly what `skip_descriptor` does.

I've added `test_read_descriptor_table_index_key` for this too, which before this change failed with v2 for the same reason.

python/tests/conftest.py
```diff
@@ -2121,3 +2121,35 @@ def evaluate_item(item, part_string: str, mark_to_add: Mark):
 
 
 # endregion
+
+# region ======================== V1/V2 Encoding Fixture Variants ============================
+
+
+@pytest.fixture
+def lmdb_version_store_tiny_segment_v1_v2(version_store_factory, lib_name, encoding_version):
```
alexowens90: Instead of adding more fixtures like this, can we just use `version_store_factory` directly in the tests?

python/tests/integration/toolbox/test_library_tool.py
```diff
@@ -373,6 +373,36 @@ def read_type(key, column):
     assert_frame_equal(lib.read(sym).data, get_df(18, 0, np.int64))
 
 
+def test_read_descriptor_table_index_key(lmdb_version_store):
```
alexowens90, line 376: Worth a comment about what the bug was and why this test covers it. Stand-alone it looks like it must obviously pass for any of our code to work, and we'd need git archaeology to figure out why it's useful, which is fragile.

alexowens90 (approved):

python/tests/conftest.py
phoebusm: @alexowens90 I assume you're worrying about bloating the already large `conftest.py`? If so, since the fixture is only used in `test_column_stats_creation.py` only, maybe just placing it directly in that file instead. That keeps `conftest.py` clean while still eliminating the ~6-line repetition across 22 tests.

phoebusm (approved):

python/tests/conftest.py
poodlewars: I prefer having them exploded in the tests themselves. Easier to read than tracing through fixtures to see what's going on.

alexowens90: It's less about bloating the file, and more the proliferation of a new fixture for every new storage/encoding/lib config option combination. There are 32(!) instances of `def lmdb_version_store` in `conftest.py`, all of which could be replaced by `version_store_factory` (plus who knows how many for other storages, V2 API, defined in other files)

PR #3028, opened 2026-04-14
prune_previous: avoid pruning immediate versions to better support concurrency

jamesblackburn:
### Problem

  Concurrent writers appending to the same base version could lose data: writer 1's
  `prune_previous` physically deleted the base version's data segments before writer 2
  had finished referencing them.

  ### Fix — two-phase mark/sweep

  - **Mark** (`version_map::write_and_prune_previous`): write the new head plus an
    individual `TOMBSTONE` for every live previous version in one journal entry. Pruned
    versions become present-but-invisible *above* the `TOMBSTONE_ALL` line, so their data
    stays in storage and remains findable. Returns the loaded entry for reuse.
  - **Sweep** (`LocalVersionedEngine::delete_unreferenced_pruned_indexes`): reuse that entry
    to collect the above-line tombstoned set, retain the **anchor** (just-superseded previous
    head) plus anything inside the `PrunePreviousProtectionSecs` window, physically delete the
    rest (with snapshot/dedup/append-inheritance protection), then advance the `TOMBSTONE_ALL`
    line past the removed block.

  The bounded `UNDELETED_ONLY` load stops at the line, so prune/`delete_all` cost is
  proportional to retained data, not total history — a constant number of storage ops per
  call. Under `delayed_deletes` the physical delete is skipped but the line still advances so
  the chain stays bounded.

  ### Also

  - `delete_all_versions` reclaims prune-retained anchors via `finalize_tombstone_all_result`
    (filtered to individually-tombstoned keys), preventing a leak when a whole symbol is deleted.
  - `get_version_at_time` now finds snapshot-protected deleted versions when a live version also
    exists — prune leaves these behind more often now.

  ### Behavioural change ⚠️ 

  `prune_previous` no longer physically removes the immediately-previous version straight away.
  It is tombstoned (invisible to `list_versions`) but its data is retained as an anchor —
  reclaimed on a later prune once aged past `VersionStore.PrunePreviousProtectionSecs`
  (default 600s; set to 0 to disable the window, anchor still applies). Steady-state on-disk
  footprint is **anchor + latest** rather than just latest.

  ### API

  No signature changes to V1/V2 public APIs. Docstrings on all `prune_previous_versions` entry
  points updated to describe deferred deletion. New config: `VersionStore.PrunePreviousProtectionSecs`.

  ### Tests

  New `test_prune_previous.py` (window retention, anchor retention, burst reclaim,
  append/dedup dangling-reference safety, delete-all reclaim + snapshot safety); updated C++
  VersionMap/rapidcheck expectations; storage-op-count test now asserts *constant* per-prune
  ops instead of a hardcoded number.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1587,7 +1587,9 @@ void LocalVersionedEngine::write_version_and_prune_previous(
         bool prune_previous_versions, const AtomKey& new_version, const std::optional<IndexTypeKey>& previous_key
 ) {
     if (prune_previous_versions) {
-        auto pruned_indexes = version_map()->write_and_prune_previous(store(), new_version, previous_key);
+        auto pruned_indexes = version_map()->write_and_prune_previous(
+                store(), new_version, previous_key, cfg().write_options().delayed_deletes()
+        );
         delete_unreferenced_pruned_indexes(std::move(pruned_indexes), new_version).get();
     } else {
         version_map()->write_version(store(), new_version, previous_key);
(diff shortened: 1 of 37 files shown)


jamesblackburn:
Issue with as_of reads when a version is only held by snapshots: https://github.com/man-group/ArcticDB/pull/3034

PR #3029, opened 2026-04-14
Enhancement 11460943163: defrag dynamic schema

alexowens90:
#### Reference Issues/PRs
[11460943163](https://man312219.monday.com/boards/7852509418/pulses/11460943163)

Adds support for dynamic schema in `compact_data_experimental`

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -216,6 +216,7 @@ set(arcticdb_srcs
         column_store/column_data_random_accessor.hpp
         column_store/column.hpp
         column_store/column_algorithms.hpp
+        column_store/column_reslicer.hpp
         column_store/column_utils.hpp
         column_store/key_segment.hpp
         column_store/memory_segment.hpp
@@ -458,6 +459,7 @@ set(arcticdb_srcs
         column_store/chunked_buffer.cpp
         column_store/column.cpp
         column_store/column_data.cpp
+        column_store/column_reslicer.cpp
         column_store/column_utils.cpp
         column_store/column_data_random_accessor.cpp
         column_store/key_segment.cpp
@@ -1039,9 +1041,9 @@ if(${TEST})
             column_store/test/ingestion_stress_test.cpp
             column_store/test/test_column.cpp
             column_store/test/test_column_data_random_accessor.cpp
+            column_store/test/test_column_reslicer.cpp
             column_store/test/test_index_filtering.cpp
             column_store/test/test_memory_segment.cpp
-            column_store/test/test_segment_reslicer.cpp
             column_store/test/test_statistics.cpp
             entity/test/test_atom_key.cpp
             entity/test/test_key_serialization.cpp
--- a/cpp/arcticdb/column_store/column_reslicer.cpp
+++ b/cpp/arcticdb/column_store/column_reslicer.cpp
@@ -288,8 +288,8 @@ std::vector<std::optional<Column>> ColumnReslicer::initialise_output_columns() c
             auto num_rows = reslicing_info_.rows_in_slice(idx);
             auto set_bits = bitset.count_range(output_rows, output_rows + num_rows - 1, bit_index);
             output_values += set_bits;
-            if (set_bits == 0) {               // No values in the range of this output column
-                output_columns.emplace_back(); // std::nullopt
+            if (set_bits == 0) { // No values in the range of this output column
+                output_columns.emplace_back(std::nullopt);
             } else if (set_bits == num_rows) {
                 output_columns.emplace_back(
                         std::make_optional<Column>(type, num_rows, AllocationType::PRESIZED, Sparsity::NOT_PERMITTED)
--- a/cpp/arcticdb/column_store/column_reslicer.cpp
+++ b/cpp/arcticdb/column_store/column_reslicer.cpp
@@ -8,6 +8,8 @@
 
 #include <arcticdb/util/string_utils.hpp>
 
+#include <folly/container/Enumerate.h>
+
 #include <arcticdb/column_store/string_pool.hpp>
 #include <arcticdb/column_store/column_reslicer.hpp>
 
@@ -66,6 +68,11 @@ std::vector<std::optional<Column>> ColumnReslicer::reslice_columns(std::vector<S
             return reslice_by_iteration(string_pools);
         }
     }();
+    for (auto&& [idx, col] : folly::enumerate(res)) {
+        if (col.has_value()) {
+            col->set_row_data(reslicing_info_.rows_in_slice(idx) - 1);
+        }
+    }
     // This was a destructive process, reset the class so that it can be safely reused
     cols_or_row_counts_.clear();
     type_.reset();
--- a/cpp/arcticdb/column_store/column_reslicer.cpp
+++ b/cpp/arcticdb/column_store/column_reslicer.cpp
@@ -1,0 +1,316 @@
+/* Copyright 2026 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software
+ * will be governed by the Apache License, version 2.0.
+ */
+
+#include <arcticdb/util/string_utils.hpp>
+
+#include <arcticdb/column_store/string_pool.hpp>
+#include <arcticdb/column_store/column_reslicer.hpp>
+
+namespace arcticdb {
+
+ColumnReslicer::ColumnReslicer(const ReslicingInfo& reslicing_info) : reslicing_info_(reslicing_info) {}
+
+void ColumnReslicer::push_back(std::shared_ptr<Column> column, std::shared_ptr<StringPool> string_pool) {
+    if (column->is_sparse()) {
+        sparse_ = true;
+        // Will be implemented in a future PR
+        schema::raise<ErrorCode::E_UNSUPPORTED_COLUMN_TYPE>("compact_data not yet supported with sparse data");
+    }
+    if (type_.has_value()) {
+        if (type_->data_type() == DataType::UTF_DYNAMIC64) {
+            schema::check<ErrorCode::E_DESCRIPTOR_MISMATCH>(
+                    is_sequence_type(column->type().data_type()),
+                    "Cannot compact data: no common type between {} and {}",
+                    type_->data_type(),
+                    column->type().data_type()
+            );
+        } else {
+            if (column->type() != *type_) {
+                numeric_types_all_same_ = false;
+            }
+            // IntToFloatConversion::STRICT matches the behaviour when appending/updating data with dynamic schema
+            auto opt_common_type = has_valid_common_type(*type_, column->type(), IntToFloatConversion::STRICT);
+            // This should not be possible by construction, as has_valid_common_type is also used when
+            // appending/updating with dynamic schema
+            schema::check<ErrorCode::E_DESCRIPTOR_MISMATCH>(
+                    opt_common_type.has_value(),
+                    "Cannot compact data: no common type between {} and {}",
+                    type_->data_type(),
+                    column->type().data_type()
+            );
+            type_ = opt_common_type;
+        }
+    } else {
+        type_ = is_sequence_type(column->type().data_type()) ? make_scalar_type(DataType::UTF_DYNAMIC64)
+                                                             : column->type();
+    }
+    cols_or_row_counts_.emplace_back(ColumnWithStrings{std::move(column), std::move(string_pool)});
+}
+
+void ColumnReslicer::push_back(size_t row_count) {
+    cols_or_row_counts_.emplace_back(row_count);
+    sparse_ = true;
+}
+
+std::vector<std::optional<Column>> ColumnReslicer::reslice_columns(std::vector<StringPool>& string_pools) {
+    util::check(type_.has_value(), "ColumnReslicer::reslice_columns called without any calls to push_back");
+    auto res = [this, &string_pools]() {
+        if (!is_sequence_type(type_->data_type()) && numeric_types_all_same_) {
+            return reslice_by_memcpy();
+        } else {
+            return reslice_by_iteration(string_pools);
+        }
+    }();
+    // This was a destructive process, reset the class so that it can be safely reused
+    cols_or_row_counts_.clear();
+    type_.reset();
+    sparse_ = false;
+    numeric_types_all_same_ = true;
+    return res;
+}
+
+std::vector<std::optional<Column>> ColumnReslicer::reslice_by_memcpy() {
+    // In theory, this could be done zero copy by adding external blocks to the output columns, referencing data in the
+    // input columns. There are 2 arguments against this:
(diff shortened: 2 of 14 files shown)


python/tests/unit/arcticdb/version_store/test_compact_data.py
```diff
@@ -457,28 +457,146 @@ def test_compact_sparse_data(lmdb_version_store_dynamic_schema_v1):
...
-def test_compact_data_dynamic_schema_changing_types(lmdb_version_store_dynamic_schema_v1):
-    lib = lmdb_version_store_dynamic_schema_v1
+@pytest.mark.parametrize(
+    "first_type", ["uint8", "uint16", "uint32", "uint64", "int8", "int16", "int32", "int64", "float32", "float64"]
+)
+@pytest.mark.parametrize(
+    "second_type", ["uint8", "uint16", "uint32", "uint64", "int8", "int16", "int32", "int64", "float32", "float64"]
+)
+def test_compact_data_dynamic_schema_changing_types(
+    in_memory_store_factory, first_type, second_type, any_output_format
```
IvoDD, line 467: Nit: `any_output_format` is unused

python/tests/unit/arcticdb/version_store/test_compact_data.py
```diff
@@ -457,28 +457,146 @@ def test_compact_sparse_data(lmdb_version_store_dynamic_schema_v1):
...
+def test_compact_data_output_column_missing_from_slice_constant_types(in_memory_store_factory):
+    lib = in_memory_store_factory(dynamic_schema=True, segment_row_size=10)
+    sym = "test_compact_data_output_column_missing_from_slice_constant_types"
+    # This configuration tests the right thing because:
+    # - we start with row-slices of 5, 10, and 5 rows respectively
+    # - with segment_row_size=10, the acceptable range of row-slice length is [6, 13]
+    # - this means that structure_for_processing will put all segments together for processing
+    # - the column reslicer will then split these 20 rows into 2 segments of 10
+    # - Therefore, present_in_first will be missing from the second slice, and present_in_last will be missing from the
+    #   first slice in the resulting segments written to disk
```
IvoDD, line 543: `generic_compact_data_test` would not differentiate between second slice missing `present_in_first` and second slice having a fully sparse `present_in_first`.

What do you think about also asserting that the table data keys have the correct `SegmentDescriptor`s with the library tool?

python/tests/unit/arcticdb/version_store/test_compact_data.py
```diff
@@ -457,28 +457,146 @@ def test_compact_sparse_data(lmdb_version_store_dynamic_schema_v1):
...
-    lib.write(sym, write_df)
-    lib.append(sym, append_df)
-    with pytest.raises(SchemaException) as e:
-        lib.compact_data_experimental(sym)
-    assert "dynamic" in str(e.value)
+    df0 = pd.DataFrame({"col": np.arange(1, dtype=np.dtype(first_type))})
+    df1 = pd.DataFrame({"col": np.arange(1, dtype=np.dtype(second_type))})
+    lib.write(sym, df0)
+    lib.append(sym, df1)
+    generic_compact_data_test(lib, sym)
```
IvoDD, line 479: Would be interesting to have a test where we also do slicing alongside type promotion.

E.g. if we have:
```
5 rows int64
10 rows int8
5 rows int8
```
with `rows_per_segment = 10`
All `20` rows will be put in a single segment
and after slicing we should end up with:
```
10 rows int64
10 rows int64
```
Worth noting that the second slice could theoretically be int8 and use memcpy but the current implementation will use int64. I think converting all to the common type always is better because we won't incur the `static_cast` cost on follow up reads

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -0,0 +1,316 @@
...
+            auto opt_common_type = has_valid_common_type(*type_, column->type(), IntToFloatConversion::STRICT);
+            // This should not be possible by construction, as has_valid_common_type is also used when
+            // appending/updating with dynamic schema
+            schema::check<ErrorCode::E_DESCRIPTOR_MISMATCH>(
+                    opt_common_type.has_value(),
+                    "Cannot compact data: no common type between {} and {}",
+                    type_->data_type(),
+                    column->type().data_type()
+            );
+            type_ = opt_common_type;
```
IvoDD, line 46: What do you think about converting to the type in the timeseries descriptor instead of the common type?

I get that this is valuable for this to be generally usable outside of concat but we can make the constructor take a `std::optional<TypeDescriptor> common_type` to hint what the common type should be.

Main benefit is that it avoids static_cast on follow up reads. E.g. suppose we have the following on disc structure:
```
5 rows int8
5 rows int16
5 rows int32
5 rows int64
```

And we want to combine it with `rows_per_segment=10`
Currently we will end up with:
```
10 rows int16
10 rows int64
```
which on follow up reads will require static casting the int16 to int64.

If we instead always use the timeseries descriptor we would end up with:
```
10 rows int64
10 rows int64
```
which will be marginally faster on follow up reads.

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -0,0 +1,316 @@
...
+                        input_values += col.row_count();
+                        if (col.is_sparse()) {
+                            const auto& sparse_map = col.sparse_map();
+                            auto end_bit = sparse_map.end();
+                            for (auto set_bit = sparse_map.first(); set_bit < end_bit; ++set_bit) {
+                                inserter = *set_bit + idx;
+                            }
+                        } else {
+                            for (auto local_idx = 0; local_idx < col.last_row() + 1; ++local_idx) {
+                                inserter = local_idx + idx;
```
IvoDD, line 261: Potential performance improvement: We can replace this for with a single `bitset.set_range`. Unclear whether that will actually be faster as I'm not too familiar with `bitmagic`

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -0,0 +1,316 @@
...
+            } else if (set_bits == num_rows) {
+                output_columns.emplace_back(
+                        std::make_optional<Column>(type, num_rows, AllocationType::PRESIZED, Sparsity::NOT_PERMITTED)
+                );
+            } else { // Output column is sparse
+                output_columns.emplace_back(
+                        std::make_optional<Column>(type, set_bits, AllocationType::PRESIZED, Sparsity::PERMITTED)
+                );
+                output_columns.back()->set_sparse_map(
+                        util::truncate_sparse_map(bitset, output_rows, output_rows + num_rows)
```
IvoDD, line 295: Potential performance improvement: We can directly construct the sliced sparse maps instead of constructing one big one and truncating it several times.
This will be more complex (as it will bundle these two for loops together)
And will need benchmarking against alternative.

So definetely not for this PR

cpp/arcticdb/column_store/segment_reslicer.cpp
```diff
@@ -6,257 +6,88 @@
...
                 res.at(idx).add_column(col_name, std::make_shared<Column>(std::move(*sliced_cols.at(idx))));
             }
         }
     }
-    for (size_t idx = 0; idx < slicing_info.num_segments; ++idx) {
-        auto rows_to_consume = std::min(slicing_info.rows_in_slice(idx), total_rows);
-        res.at(idx).set_row_data(rows_to_consume - 1);
-        total_rows -= rows_to_consume;
+    for (size_t idx = 0; idx < reslicing_info.num_segments; ++idx) {
+        res.at(idx).set_row_data(reslicing_info.rows_in_slice(idx) - 1);
```
IvoDD, line 90: We are currently relying on this `set_row_data` to call `set_row_data` on each column in the output. We should probably also call `set_row_data` inside the `ColumnReslicer` so it can be safely used standalone?

cpp/arcticdb/column_store/test/test_column_reslicer.cpp
```diff
@@ -75,10 +71,231 @@ TYPED_TEST(SegmentReslicerDenseNumericStaticSchemaFixture, CombineIntoOne) {
     ASSERT_EQ(col.buffer().blocks().front()->capacity(), total_rows * sizeof(RawType));
 }
 
-TYPED_TEST(SegmentReslicerDenseNumericStaticSchemaFixture, SplitInTwo) {
+TYPED_TEST(ColumnReslicerDenseNumericSameTypeFixture, CombineIntoOneDynamicMissingFirstSlice) {
```
vasil-pashov, line 72: Having  two consecutive missing slices is another case worth covering

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -0,0 +1,316 @@
...
+ColumnReslicer::ColumnReslicer(const ReslicingInfo& reslicing_info) : reslicing_info_(reslicing_info) {}
+
+void ColumnReslicer::push_back(std::shared_ptr<Column> column, std::shared_ptr<StringPool> string_pool) {
+    if (column->is_sparse()) {
+        sparse_ = true;
+        // Will be implemented in a future PR
+        schema::raise<ErrorCode::E_UNSUPPORTED_COLUMN_TYPE>("compact_data not yet supported with sparse data");
+    }
+    if (type_.has_value()) {
+        if (type_->data_type() == DataType::UTF_DYNAMIC64) {
```
vasil-pashov, line 27: Do we need to handle other string types?

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -0,0 +1,316 @@
...
+}
+
+void ColumnReslicer::push_back(size_t row_count) {
+    cols_or_row_counts_.emplace_back(row_count);
+    sparse_ = true;
+}
+
+std::vector<std::optional<Column>> ColumnReslicer::reslice_columns(std::vector<StringPool>& string_pools) {
+    util::check(type_.has_value(), "ColumnReslicer::reslice_columns called without any calls to push_back");
+    auto res = [this, &string_pools]() {
```
vasil-pashov, line 64: Returning lambdas that capture references makes me feel uneasy. It's too easy to get it wrong and segfault.

cpp/arcticdb/column_store/column_reslicer.cpp
alexowens90: I did think about this. My reasoning about the current choice was that a column of `int8` will compress better than a column of `int64` where all the values are in `[-128, 127)`, although the difference is likely to be marginal.
Needs some measurements of both to be sure. I'll add a note to the "profiling and optimisation" ticket to investigate.

cpp/arcticdb/column_store/column_reslicer.cpp
alexowens90: I'll add a note to the profiling and optimisation ticket for this as well

cpp/arcticdb/column_store/column_reslicer.cpp
alexowens90: Yeh, I doubt the performance benefit would outweigh the readability benefit, but I'll check

cpp/arcticdb/column_store/segment_reslicer.cpp
alexowens90: Ah good catch, missed that when I factored stuff out into the `ColumnReslicer` class

cpp/arcticdb/CMakeLists.txt
```diff
@@ -1014,111 +1014,111 @@ if(${TEST})
...
-            async/test/test_async.cpp
-            codec/test/test_codec.cpp
-            codec/test/test_encode_field_collection.cpp
-            codec/test/test_segment_header.cpp
-            codec/test/test_encoded_field.cpp
-            column_store/test/test_chunked_buffer.cpp
-            column_store/test/ingestion_stress_test.cpp
-            column_store/test/test_column.cpp
-            column_store/test/test_column_data_random_accessor.cpp
+#            arrow/test/arrow_test_utils.hpp
```
IvoDD: Forgotten debugging leftover

alexowens90: 🤦

python/tests/unit/arcticdb/version_store/test_compact_data.py
```diff
@@ -552,34 +708,62 @@ def test_compact_data_hypothesis_general(
...
+        "strings_2": [str],
+    }
+    all_col_names = list(cols.keys())
+    string_values = random_strings_of_length(10, 5, True)
+    # Append random numbers of rows between 1 and 2 * rows_per_segment and then compact with an explicit argument
+    remaining_rows = num_rows
+    while remaining_rows > 0:
+        rows_to_take = rng.integers(1, 2 * rows_per_segment)
+        # Pick a subset of columns
+        num_columns = rng.integers(1, len(cols) + 1)
```
vasil-pashov: I think it's better to put as much of the random stuff in hypothesis generators or in the given clause so that it can converge in case of an error. It cannot track the randomness of the random generator and can timeout trying to converge to a minimal repro

For example the string values don't seem to be crucial for the correctness but `num_columns` `col_names`, `col_type` seem to be better suited to be in the given clause

IvoDD (approved):

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -0,0 +1,323 @@
...
+                bitset.count()
+        );
+        uint64_t output_values{0};
+        uint64_t output_rows{0};
+        for (idx = 0; idx < reslicing_info_.num_segments; ++idx) {
+            auto num_rows = reslicing_info_.rows_in_slice(idx);
+            auto set_bits = bitset.count_range(output_rows, output_rows + num_rows - 1, bit_index);
+            output_values += set_bits;
+            if (set_bits == 0) {               // No values in the range of this output column
+                output_columns.emplace_back(); // std::nullopt
```
vasil-pashov: nit: `output_columns.emplace_back(std::nullopt);` seems more natural than empty emplace back and a comment

vasil-pashov (approved):

cpp/arcticdb/column_store/column_reslicer.cpp
alexowens90: Nope, we always promote to UTF-8 for string columns:
https://github.com/man-group/ArcticDB/blob/491ffe0d67a08aa581d3f8693963ba69a30d15ce/cpp/arcticdb/column_store/column_reslicer.cpp#L51-L52

cpp/arcticdb/column_store/column_reslicer.cpp
alexowens90: The lambda isn't returned, it's called immediately

python/tests/unit/arcticdb/version_store/test_compact_data.py
alexowens90: In general I agree, but in this case, the number of columns and which ones are chosen (type is deterministic based on the name) is different for each row slice. The compromise is that the random number generator is seeded at the start of each test case, so results are deterministic for the input `num_rows` and `rows_per_segment` at least

cpp/arcticdb/column_store/column_reslicer.cpp
vasil-pashov: Ah true I can't read

PR #3030, opened 2026-04-14
Developer build performance improvements: Enable ccache for Linux builds + mold (if available)

jamesblackburn:
## Developer build performance improvements                                                                                
                                                                                                                                                    
  Three complementary build-speed optimisations for the `linux-debug` and `linux-release` presets.                                                  
  Conda and CI wheel builds are unaffected.                                                                                                         
                                                                                                                                                    
  ### Build times: before vs after                                                                                                                  
                                                                                                                                                    
  | Scenario | Before | After |                                                                                                                     
  |---|---|---|                                                        
  | Full clean build (cold ccache) | ~20 min | ~20 min |
  | Full clean build (ccache warm — e.g. fresh worktree, branch switch) | ~20 min | ~4–6 min |
  | Incremental: touch one `.cpp`, rebuild `arcticdb_ext` | ~56s | 6.5s |
  | No-op Build: | ~10s | ~1s |
                                              
  Incremental breakdown:                                                                                                                            
   
  | Stage | Before | After | Saving |                                                                                                               
  |---|---|---|---|                                                    
  | Compile (ccache warm hit) | minutes | ~1s | — |                                                                                                 
  | `ar` + `ranlib` (1.95 GB static lib) | ~23s | ~3.5s | −20s |       
  | Final link (`arcticdb_ext.so`, 965 MB) `ld` -> `mold` | ~24s | ~1s | −23s |
  | **Total** | **~56s** | **~10s** | **5.6×** |
                                              
  `make build-debug` end-to-end (including cmake reconfigure): **~11s**.                                                                            
   
  ---                                                                                                                                               
                                                                       
  ### What changed and why                                                                                                                          
                                                                       
  **ccache** (`1b2a92865`, `2159e93b1`)

  Enabled for ArcticDB source builds via a CMake launcher script. `hard_link` mode is set 
  so cached objects are hard-linked rather than copied, saving ~40% of cache I/O. On a cold 
  cache the first build takes the full ~20 min; after that, any file that hasn't changed since the 
  last build on any worktree compiles in ~1s regardless of how template-heavy it is.               
                                                                                                                                                    
  **mold linker** (`1f7d5b901`)                                                                                                                     
                                          
  Auto-detected on Linux and wired up via `-fuse-ld=mold` when sanitizers are not active. mold                                                      
  uses async I/O (`io_uring`) and parallelises symbol resolution. Measured on the 965 MB debug                                                      
  `.so`: link step drops from ~24s (GNU ld) to ~1s (mold, page-cache warm). Skipped automatically
  for sanitizer builds, which require `lld`.                                                                                                        
                                                                       
  **Thin archives** (`f9d00d9a7`)
                                                                                                                                                    
  `ARCTICDB_THIN_ARCHIVES=ON` is set by default in the `linux-debug` preset (inherited by                                                           
  `linux-release`). This overrides `CMAKE_CXX_ARCHIVE_CREATE` to use `ar qcT` — which stores                                                        
  paths to `.o` files rather than copying their data — and sets `CMAKE_CXX_ARCHIVE_FINISH` to a                                                     
  no-op, skipping `ranlib`. The symbol index is unnecessary here as the linker reads all members                                                    
  regardless — `arcticdb_core_static` contributes nearly its entire contents to the final `.so`.                                                    
  Saves ~20s per incremental build. Left `OFF` for conda presets where `arcticdb_core_static.a`                                                     
  is consumed by external C++ projects and must remain self-contained.       
                                                                       
  > **Note:** CMake 3.31's `THIN_ARCHIVE` target property was evaluated but does not update                                                         
  > `link.txt` under the Unix Makefiles generator. Setting `CMAKE_CXX_ARCHIVE_CREATE/FINISH`
  > directly is the reliable mechanism.


cpp/CMakeLists.txt
```diff
@@ -10,6 +10,17 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMake" ${CMAKE_MODULE_PATH})
+
+# Use ccache if available and no compiler launcher is already configured (e.g. sccache via presets)
```
poodlewars, line 14: I think this sits better in a `CMakeUserPresets.json`?

cpp/arcticdb/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ option(TEST "Build all tests." ON)
 option(BUILD_WITH_REMOTERY "Build ArcticDB with Remotery (used for performance analysis)" OFF)
 option(SSL_LINK "Link against SSL libraries." ON)
 option(HIDE_LINKED_SYMBOLS "Passed `exclude-libs` through to linker, ensuring linked library symbols are not exported. Shrinks binary size." ON)
+option(ARCTICDB_THIN_ARCHIVES "Use thin archives for intermediate static libraries. Faster incremental builds but archives are not self-contained — do not use for installed/distributed builds." OFF)
```
poodlewars: Is it safe to use this for our release build then? Which seems to be the current behaviour in this PR?

poodlewars (approved):

cpp/CMakeLists.txt
jamesblackburn: But that's not checked in? The issue this is trying to resolve for local development is when creating new git worktrees, have the builds run quickly out-the-box. Suspect this isn't possible unless we commit the CMakeUserPresets to the repo?

cpp/arcticdb/CMakeLists.txt
jamesblackburn: I did worry about this a bit - and `claude` claimed this. Triple-checking...

```
  Thin archives are safe for linux-release in CI for these reasons:                                                                                 
                                          
  1. .a files are never uploaded as artifacts — do_not_archive: ["*.so", "*.[ao]", "vcpkg_installed"] for Linux explicitly excludes them            
  2. No cmake --install step — there's no install step in any workflow that would copy arcticdb_core_static.a anywhere                              
  3. .o files always present during the build — thin archives just reference paths; since the build runs start-to-finish on one machine, the .o
  files are always there when the linker reads the thin archive                                                                                     
  4. Only arcticdb_ext.so is distributed — the wheel artifact contains only .so, not .a; the thin archive is discarded after linking
                                                                                                                                                    
  The linux-release preset is a developer/CI convenience preset for building ArcticDB itself. External C++ consumers building to package a portable 
  .a wouldn't use ArcticDB's presets, and they'd get ARCTICDB_THIN_ARCHIVES=OFF (the default) unless they explicitly opt in.                        
```

jamesblackburn: What each CI path uses                      
                                          
  ```
┌───────────────────────────────────┬───────────────────────────────────────────────────────────────────────┬────────────────────────────────┐
  │              CI Job               │                                Preset                                 │     ARCTICDB_THIN_ARCHIVES     │    
  ├───────────────────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────────────┤
  │ cpp-tests (Linux)                 │ linux-release                                                         │ ON (via preset inheritance)    │    
  ├───────────────────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────────────┤
  │ build-python-wheels (Linux)       │ linux-release (passed via CIBW_ENVIRONMENT_PASS_LINUX →               │ ON                             │
  │                                   │ ARCTIC_CMAKE_PRESET)                                                  │                                │
  ├───────────────────────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────────────┤
  │ conda builds                      │ linux-conda-release                                                   │ OFF (conda presets never set   │    
  │ (build_with_conda.yml)            │                                                                       │ it)                            │    
  └───────────────────────────────────┴───────────────────────────────────────────────────────────────────────┴────────────────────────────────┘    
```                                                                                                                                               
  So both the cpp-tests job and the wheel-building job run with thin archives on Linux release.
                                                                                                                                                    
  What makes it to a release                                                 
                                                                                                                                                    
  - PyPI wheels: wheel-cp* artifacts — these contain only arcticdb_ext.so. The .a is an intermediate that's discarded.
  - setup.py line 264 shows the wheel build target is install_arcticdb_ext, which copies the .so to the wheel. arcticdb_core_static.a is just a link-time intermediate.                                                    
  - Conda packages: Built entirely via build_with_conda.yml using linux-conda-release — unaffected by thin archives.                                
  - arcticdb_core_static.a never leaves any build machine in any path. It's excluded from artifact uploads (*.[ao]) and there is no cmake --install step in CI.

File: cpp/CMakeLists.txt (hunk 1 above)
Comment (IvoDD) [reply cpp/CMakeLists.txt:14]: My git worktree script symlinks untracked files like `CMakeUserPresets.json`. I guess we can also add a command for this in the `Makefile` if people think it's helpful. Anyway I'm happy for this to just be auto detected and not part of a preset

File: cpp/arcticdb/CMakeLists.txt (hunk 2 above)
Comment (IvoDD) [reply cpp/arcticdb/CMakeLists.txt]: Maybe we should update the description then from "do not use for installed/distributed builds" to something like "do not use if distributing link-time intermediate results like \`arcticdb_core_static.a\`"

**Review by IvoDD — APPROVED** [review]

File: cpp/arcticdb/CMakeLists.txt (hunk 2 above)
Comment (jamesblackburn) [reply cpp/arcticdb/CMakeLists.txt]: Thanks updated.

File: cpp/CMakeLists.txt (hunk 1 above)
Comment (vasil-pashov) [reply cpp/CMakeLists.txt:14]: I agree this sits better in a preset. `CMakeUserPresets.json` is git ignored so that everyone can have personal preset without having to open a PR.

Also why is this needed `env;CCACHE_HARDLINK=1`? Usually the path the exe is enough.

This can be added as a preset in CMakePresets (the global preset file) and other preset can inherit it. Similar to how vcpkg is done.

I'm on the fence if we should force the cache. I'd rather not to keep the CMake clean. In case you insist on forcing it it wold be better to add a cmake option so that it can be switched on/off. Even better if it's a flag that can set either sccache or ccache

File: cpp/CMakeLists.txt [hunk 3]
```diff
@@ -77,6 +88,16 @@ if(ARCTICDB_SANITIZER_FLAGS)
     message(STATUS "Building ArcticDB with sanitizers. Compiler flags: ${ARCTICDB_SANITIZER_FLAGS}")
 endif()
 
+# Use mold linker if available on Linux (faster linking via async I/O; ~30% speedup over GNU ld)
+# Skipped when sanitizers are active, as Clang sanitizer builds require lld.
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND NOT ARCTICDB_SANITIZER_FLAGS)
+    find_program(MOLD_PROGRAM mold)
+    if(MOLD_PROGRAM)
+        add_link_options(-fuse-ld=mold)
```
vasil-pashov, line 97: The correct way to set a custom linker is through CMAKE_LINKER_TYPE. But this again sits better in a preset. For example I have highly customized preset with a different linker, compiler, build system to my own liking and this will force a different linker on me

cpp/CMakePresets.json
```diff
@@ -115,13 +115,17 @@
       "hidden": true,
       "generator": "Unix Makefiles",
       "cacheVariables": {
-        "CMAKE_MAKE_PROGRAM": "make"
+        "CMAKE_MAKE_PROGRAM": "make",
+        "VCPKG_OVERLAY_TRIPLETS": "custom-triplets"
```
vasil-pashov: I think this should be paired with `VCPKG_TARGET_TRIPLET`. Is it doing anything now?

cpp/custom-triplets/x64-linux.cmake
```diff
@@ -0,0 +1,6 @@
+set(VCPKG_TARGET_ARCHITECTURE x64)
+set(VCPKG_CRT_LINKAGE dynamic)
+set(VCPKG_LIBRARY_LINKAGE static)
+set(VCPKG_CMAKE_SYSTEM_NAME Linux)
+
+set(VCPKG_CHAINLOAD_TOOLCHAIN_FILE "${CMAKE_CURRENT_LIST_DIR}/../CMake/ccache-toolchain.cmake")
```
vasil-pashov: I don't think we need to use the compiler cache for packages. For several reasons
1. We build them rarely
2. VCPKG has binary cache. Once a dependency is built it's cached so it's no longer rebuild
3. We have ~130 deps. The cache will be filled and since dependencies are not build often they'll get evicted anyway.

vasil-pashov (commented):
It looks like the provided speedup is reported with both mold and ccache enabled. When I last looked at link times mold did not lead to a greater performance (we have 180 TUs which is not much). I wonder if this is still the case. It would be interesting to see results with only ccache and mold disabled and ccache disabled and mold enabled

cpp/custom-triplets/x64-linux.cmake
jamesblackburn: Ok sounds reasonable

jamesblackburn:
> It looks like the provided speedup is reported with both mold and ccache enabled. When I last looked at link times mold did not lead to a greater performance (we have 180 TUs which is not much). I wonder if this is still the case. It would be interesting to see results with only ccache and mold disabled and ccache disabled and mold enabled

I think this is covered in the table at the top. 
* `ccache` brings the build time down by a factor of 3 for clean
* `mold` takes .so building down to ~1s - note it needs to be a new (2.x version of mold) which I've put in our internal environment
* Significant time of final link is building the .a archive (20s+), changing this to thin brings this down significantly

Does this cover the query?
<img width="526" height="188" alt="image" src="https://github.com/user-attachments/assets/67d8df28-7124-4f62-af57-848fa128864b" />


```
8.2 seconds end-to-end including cmake reconfigure. Breakdown from the output:                                                            
                                                                                                                                            
  - CMake configure: 5.1s (4.7s configuring + 0.4s generating)                                                                              
  - Compile version_store.cpp (ccache hit): [ 93%] Built target arcticdb_core_object — essentially instant                                  
  - ar qcT (thin archive): instant                                                                                                          
  - Link arcticdb_ext.so (mold): instant                                                                                                    
  - Symlink: instant                                                                                                                        
```

cpp/CMakeLists.txt
jamesblackburn: Ok have update this, can you have another look @vasil-pashov ?

I think we still need to detect if `mold` is available, but agree it shouldn't blat your custom settings.

Note that `claude` notes this depends on a new-ish version of cmake (>=3.29) (higher than ArcticDB defines), but clearly we're running on a newer version

Makefile
```diff
@@ -36,7 +40,7 @@ _VENV_PIP := $(VENV_DIR)/$(VENV_NAME)/bin/pip
 help: ## Show this help
 	@echo "ArcticDB Development Targets"
 	@echo ""
-	@grep -E '^[a-zA-Z_-]+:.*##' $(MAKEFILE_LIST) | \
+	@grep -hE '^[a-zA-Z_-]+:.*##' $(MAKEFILE_LIST) | \
```
jamesblackburn, line 43: This fixes an issue where the current `make` no argument doesn't show the actual build targets...

```
ArcticDB Development Targets

  Makefile             Show this help
  Makefile             Generate protobuf stubs
  Makefile             Create a dev venv (NAME=<name> required, CLEAN=1 to replace existing)
  Makefile             Print venv activate path (NAME=<name> required). Use: source $$(make activate NAME=x)
  Makefile             Full setup from scratch (CLEAN=1 to replace existing venv)
  Makefile             Run formatters (in-place)
  Makefile             Check formatting (no changes)
  Makefile             CMake configure (release) — always reconfigures
  Makefile             CMake configure (debug) — always reconfigures
  Makefile             Build arcticdb_ext (release) and symlink
  Makefile             Build arcticdb_ext (debug) and symlink
  Makefile             Build and run C++ unit tests (release, FILTER= for gtest_filter)
  Makefile             Build and run C++ unit tests (debug, FILTER= for gtest_filter)
  Makefile             Symlink release arcticdb_ext into python/
  Makefile             Symlink debug arcticdb_ext into python/
  Makefile             Run Python tests (TYPE=unit|integration|..., FILE= path, ARGS= extra pytest args)
  Makefile             Release build + symlink + run Python tests
  Makefile             Debug build + symlink + run Python tests
  Makefile             Build a pip wheel
  Makefile             Build and run C++ benchmarks (release, FILTER= for benchmark_filter)
  Makefile             Install arcticdb in editable mode
  Makefile             Run ASV Python benchmarks (BENCH= for --bench filter)

Variables (set in Makefile.local or on command line):
  RELEASE_PRESET   linux-release
  DEBUG_PRESET     linux-debug
  PROXY_CMD        withproxy
  CMAKE_JOBS       16
  PROTOC_VERS      456
  VENV_DIR         ~/venvs
  TMPDIR_OVERRIDE  /users/is/jblackburn/.tmp
```

jamesblackburn:
Latest build performance:

```
  ┌──────────────────────────────────┬───────┬──────────────────────────────────────────────────────────────┐
  │             Scenario             │ Time  │                       What's happening                       │
  ├──────────────────────────────────┼───────┼──────────────────────────────────────────────────────────────┤
  │ Real compile (ccache miss)       │ 55.5s │ Full debug compile of op_log.cpp + ar + mold link            │
  ├──────────────────────────────────┼───────┼──────────────────────────────────────────────────────────────┤
  │ ccache hit (touch, same content) │ 6.5s  │ ~1s ccache restore + ar (~3.5s) + mold link (~1s) + overhead │
  ├──────────────────────────────────┼───────┼──────────────────────────────────────────────────────────────┤
  │ No changes                       │ 1.7s  │ cmake --build scanning the build graph, nothing to do        │
  └──────────────────────────────────┴───────┴──────────────────────────────────────────────────────────────┘

  The ar step (~3.5s) dominates the ccache-hit case — even a single object file being "recompiled" forces a full thin-archive rebuild and
  relink. That's inherent to the static lib structure, not something we can fix here.
```

Makefile
```diff
@@ -82,31 +86,55 @@ lint-check: ## Check formatting (no changes)
...
-configure: ## CMake configure (release)
+# Files whose changes should trigger a cmake reconfigure.
+# Makefile.local is optional; $(wildcard) at parse time is fine since the file
+# is either always present or always absent for a given developer's machine.
+_CMAKE_INPUTS := cpp/CMakeLists.txt cpp/CMakePresets.json cpp/vcpkg.json \
+                 $(wildcard Makefile.local)
+
+# .configure-stamp is our own sentinel file — cmake never touches it, so its
+# mtime is fully under our control and won't interfere with cmake's internal
+# check-build-system mechanism (which uses .configure-stamp and cmake.check_cache).
```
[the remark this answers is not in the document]
jamesblackburn: Fixed

PR #3032, opened 2026-04-14
Arrow NaT handling

ognyanstoimenov:
## Handle NaT values in Arrow output format for timestamp columns
11525537906
* Add ArrowTimestampHandler that converts pandas NaT sentinel values (int64_t min) to Arrow nulls via a validity bitmap when reading
   NANOSECONDS_UTC64 columns in Arrow output format
* Sparse path iterates only materialized values via for_each_enumerated_flattened
* Dense path uses memcpy + separate NaT scan (tried to do it in a single pass and reuse for_each_enumerated_flattened from the sparse path but it was ~3x slower that way)
  * Inverting the bitset at the end as NaT values are expected to be fewer than non-NaT. Benchmarked x2 faster in the 0% NaTs
* Bitmap indices are built relative to the truncation window, avoiding a second pass through handle_truncation on the bitmap (which e.g the boolhandler does)

* c22af40763ff438521ac8ea233857e5ab86d558f also adds support for convert_type to be called on Columns with more than 1 non-contiguous block

Testing
* Existing test_datetime_col_with_nats un-xfailed and parametrized with row_range to cover truncation across segment 
* New test_datetime_col_with_nats_sparse covers dynamic_schema with row_range truncation
* C++ benchmarks added to benchmark_arrow_reads.cpp covering dense/sparse/truncated variants at multiple NaT ratios

<details>
<summary>Some local benchmarks I tried for different approaches </summary>

#### ArrowTimestampHandler benchmark results

Machine: 48 cores, 4 GHz, 36 MB L3. Debug build. `--benchmark_min_time=5x`.

## memcpy + separate scan vs single-pass for_each_enumerated (dense path)

10M rows, CPU time in ms, 3 runs averaged.

| NaT % | memcpy + scan | single-pass | ratio |
|-------|---------------|-------------|-------|
| 0%    | 185           | 585         | 3.2x  |
| 25%   | 369           | 762         | 2.1x  |
| 50%   | 296           | 657         | 2.2x  |
| 100%  | 85            | 435         | 5.1x  |

## Relative bitmap indices vs handle_truncation(validity bitmap) (truncated reads)

Truncation window: 25%-75% of the slice (50% of rows processed).

| Case | handle_truncation | relative indices | speedup |
|------|-------------------|------------------|---------|
| 1M dense 25% NaT   | 32.4 ms | 14.4 ms | 2.3x |
| 10M dense 25% NaT  | 371 ms  | 156 ms  | 2.4x |
| 1M half-sparse 25% NaT | 74.4 ms | 61.4 ms | 1.2x |

Building the validity bitmap with relative indices and skipping `handle_truncation` on the bitmap avoids re-iterating set bits in `truncate_sparse_map`. Dense path benefits most because the NaT scan loop also narrows to the truncated range.

 ## Inverted bitset optimisation (dense path)

  10M rows, CPU time in ms, 3 runs, debug build.

  | NaT % | no invert (ms) | invert (ms) | speedup |
  |-------|----------------|-------------|---------|
  | 0%    | 181            | 87          | 2.1x    |
  | 25%   | 356            | 314         | 1.1x   |
  | 50%   | 267            | 266         | 1.0x   |
  | 100%  | 82             | 179         | 0.5x   |

</details>


cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -522,4 +523,86 @@ void ArrowBoolHandler::default_initialize(ChunkedBuffer&, size_t, size_t, const
...
+                    if (en.value() != NaT) {
+                        inserter = en.idx() - first_idx;
+                    }
+                },
+                first_idx,
+                end_idx
+        );
+    } else {
+        const auto* src = reinterpret_cast<const timestamp*>(source_column.data().buffer().data());
+        memcpy(dest, src, m.num_rows_ * sizeof(timestamp));
```
IvoDD: A comment explaining that the memcpy is 3x faster than folding into the `for_each_enumerated` would be useful.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -522,4 +523,86 @@ void ArrowBoolHandler::default_initialize(ChunkedBuffer&, size_t, size_t, const
...
+
+        for (size_t i = first_idx; i < end_idx; ++i) {
+            if (src[i] != NaT) {
+                inserter = i - first_idx;
```
IvoDD, line 597: We usually expect a small percentage of values to be `pd.NaT`s. For such cases it should be faster to populate the inverted bitset and then invert it. Can you try that approach and see what the benchmarks say? I.e. I'd expect the dense 0% NaT to become faster.

IvoDD, line 597: Also a nice idea to directly construct the truncated validity bitmap. It is different to how we handle this in the other type handlers, we should consider doing the same for all type handlers in a separate PR.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -522,4 +523,86 @@ void ArrowBoolHandler::default_initialize(ChunkedBuffer&, size_t, size_t, const
...
+    util::check(field.has_ndarray(), "Timestamp handler expected array");
+    const auto& ndarray = field.ndarray();
+    const auto bytes = encoding_sizes::data_uncompressed_size(ndarray);
+
+    Column decoded_data{
+            m.source_type_desc_,
+            bytes / get_type_size(m.source_type_desc_.data_type()),
+            AllocationType::DYNAMIC,
+            Sparsity::PERMITTED
+    };
```
IvoDD: I think the main extra cost from this approach is that we now allocate a new tmp column and memcopy from it.

I know I suggested doing it like this initally but we should be able to speed up the dense case by decoding directly onto the dest_column. We can do this in a similar way to [here](https://github.com/man-group/ArcticDB/blob/609d990b6855eb475f4d37546fc177265cdb24dc/cpp/arcticdb/python/python_handlers.cpp#L182-L186) to look like e.g.:

```
def handle_type:
    if sparse:
        decoded_data = Column(size, DYNAMIC, PERMITTED) # Same as what we do now
    else:
        decoded_data = Column(0, NOT_PERMITTED)
        decoded_data.add_external_block(dest_column.bytes_at(...))
    
    data += decode_field(....)
    convert_type(....)


def convert_type:
    if sparse:
         same as before
     else:
         if (dest_ptr != src_ptr) {
                memcpy(....) # Only do the memcpy if the src and destination are in different places
          }

         same nat handling as above
```

python/tests/unit/arcticdb/version_store/test_arrow_read.py
```diff
@@ -204,10 +204,29 @@ def test_datetime_col_with_nats(lmdb_version_store_tiny_segment):
...
-    expected = lib.read("arrow", output_format=OutputFormat.PANDAS).data
-    assert table.column("x").null_count == 4
+    table = lib.read("arrow", row_range=row_range).data
+    expected = lib.read("arrow", row_range=row_range, output_format=OutputFormat.PANDAS).data
+    assert_frame_equal_with_arrow(table, expected)
+    assert table.column("x").is_null().to_pylist() == expected["x"].isna().tolist()
+
+
+@pytest.mark.parametrize("row_range", [None, (1, 5), (2, 7), (3, 6)])
+def test_datetime_col_with_nats_sparse(lmdb_version_store_tiny_segment_dynamic, row_range):
```
IvoDD, line 213: Currently there is no test which tests combining `pd.NaT` with sparse data

This is actually kind of hard to test because:
- Arrow sparse writes can't include `pd.NaT`
- Regular pandas writes can't really write sparse data

A mix of pd.NaT and sparse data will be able to be produced in the near future with `compact_data_experimental`.

I suggest writing a hacky test which uses arrow writes but masks the `pd.NaT`s with the corresponding arrow value in nanoseconds (I believe min int64_t)

cpp/arcticdb/arrow/arrow_handlers.cpp
ognyanstoimenov: Very cool hack with the `add_external_block`. Implemented.

cpp/arcticdb/arrow/arrow_handlers.cpp
ognyanstoimenov: Good idea. It is actually faster. Implemented it for the non-sparse case and added the benchmarks in the PR description

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -522,4 +523,100 @@ void ArrowBoolHandler::default_initialize(ChunkedBuffer&, size_t, size_t, const
...
+    const auto& ndarray = field.ndarray();
+    const auto bytes = encoding_sizes::data_uncompressed_size(ndarray);
+
+    auto decoded_data = [&m, bytes, &dest_column]() -> Column {
+        const bool num_non_empty_rows = bytes / get_type_size(m.source_type_desc_.data_type());
+        const bool is_sparse = num_non_empty_rows < m.num_rows_;
+        if (is_sparse) {
+            return Column(m.source_type_desc_, num_non_empty_rows, AllocationType::DYNAMIC, Sparsity::PERMITTED);
+        } else {
+            Column column(m.source_type_desc_, Sparsity::NOT_PERMITTED);
```
ognyanstoimenov, line 541: Is Sparsity::NOT_PERMITTED correct here for arrow?

python/tests/unit/arcticdb/version_store/test_arrow_read.py
```diff
@@ -229,6 +228,34 @@ def test_datetime_col_with_nats_sparse(lmdb_version_store_tiny_segment_dynamic,
...
+            1735689600000000000,  # 2025-01-01
+            nat_sentinel,
+            0,
+            nat_sentinel,
+            1735776000000000000,  # 2025-01-02
+            0,
+        ],
+        dtype=np.int64,
+    )
+    mask = np.array([False, False, True, False, False, True])
```
IvoDD: Nit: Can't we construct the timestamps by:
```
timestamps = pa.array([timestamp, nat_sentinel, None, nat_sentinel, timestamp, None], type=pa.int64)
timestamps = timestamps.cast(pa.timestamp('ns'))
```

This would avoud the mask construction

python/tests/unit/arcticdb/version_store/test_arrow_read.py
```diff
@@ -184,9 +184,8 @@ def test_strings_with_nones_and_nans(lmdb_version_store_tiny_segment, row_range,
 
 
 @pytest.mark.parametrize("row_range", [None, (2, 3), (2, 5), (2, 6), (3, 5)])
-def test_datetime_col_with_nats(lmdb_version_store_tiny_segment, row_range):
-    lib = lmdb_version_store_tiny_segment
-    lib.set_output_format(OutputFormat.PYARROW)
+def test_datetime_col_with_nats(lmdb_version_store_arrow, row_range):
```
IvoDD: This change removed the `tiny_segment`. Was that intentional?

Previously the test had the nice property of having a mix of all [value, value], [value, nat], [nat, value], [nat, nat] segments.

cpp/arcticdb/arrow/arrow_handlers.cpp
IvoDD: Yes! In this case we know the resulting column does not have a sparse map (we're in the else in the if).

Side note, surprisingly the `dest_column` for arrow is also constructed as `Sparsity::NOT_PERMITTED` because it is an output buffer for arrow. Arrow buffers are always dense on their own. An arrow array becomes sparse only when we combine a dense value buffer with a dense bitmap. The bitmap just invalidates certain rows of the values buffer but it doesn't compact them. Happy to chat offline if that's confusing.

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -532,12 +532,18 @@ void ArrowTimestampHandler::handle_type(
...
-            Sparsity::PERMITTED
-    };
+    auto decoded_data = [&m, bytes, &dest_column]() -> Column {
+        const auto num_non_empty_rows = bytes / get_type_size(m.source_type_desc_.data_type());
+        const bool is_sparse = num_non_empty_rows < m.num_rows_;
+        if (is_sparse) {
+            return Column(m.source_type_desc_, num_non_empty_rows, AllocationType::DYNAMIC, Sparsity::PERMITTED);
+        } else {
+            Column column(m.source_type_desc_, Sparsity::NOT_PERMITTED);
+            // Point to dest_column's buffer without owning it.
```
IvoDD: I'd rephrase the comment:
```
When the column is not sparse we can decode directly onto the destination buffer. So, we make decoded_data hold a non-owning pointer to the destination buffer.
```

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -547,43 +553,53 @@ void ArrowTimestampHandler::handle_type(
...
                 end_idx
         );
+        inserter.flush();
     } else {
-        const auto* src = reinterpret_cast<const timestamp*>(source_column.data().buffer().data());
-        memcpy(dest, src, m.num_rows_ * sizeof(timestamp));
+        auto column_data = source_column.data();
+        auto* write_ptr = dest;
+        // 3x faster than folding this into the is_sparse case above and doing for_each_enumerated over source
+        while (auto block = column_data.next<TimestampTag>()) {
```
IvoDD: Why do we use a `while`? Isn't the source column guaranteed to have a single contiguous buffer?

Otherwise the `column.buffer().add_external_block(dest_column.bytes_at(m.offset_bytes_, bytes), bytes);` would fail

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -547,43 +553,53 @@ void ArrowTimestampHandler::handle_type(
...
                 first_idx,
                 end_idx
         );
+        inserter.flush();
     } else {
-        const auto* src = reinterpret_cast<const timestamp*>(source_column.data().buffer().data());
-        memcpy(dest, src, m.num_rows_ * sizeof(timestamp));
+        auto column_data = source_column.data();
+        auto* write_ptr = dest;
+        // 3x faster than folding this into the is_sparse case above and doing for_each_enumerated over source
```
IvoDD: Maybe rephrase:
```
// 3x faster for dense to `memcpy` + scan each element than doing the copy in the for_each_enumerated.
// If going through the `handle_type` codepath we can skip the memcpy as we have decoded directly onto the destination buffer.
```

cpp/arcticdb/arrow/test/benchmark_arrow_reads.cpp
```diff
@@ -157,12 +157,14 @@ BENCHMARK(BM_arrow_string_handler)
 
 // ── Timestamp NaT handler benchmark ─────────────────────────────────────────
 
-// Args: {num_rows, nat_percentage, num_sparse, truncate_25pct}
+// Args: {num_rows, nat_percentage, num_sparse, truncate_25pct, num_source_chunks}
+// num_source_chunks: 0 or 1 = single block, >1 = multi-block source
```
IvoDD, line 161: I think `num_source_chunks` is obsolete. We always call `handle_type` or `convert_type` with a contiguous buffer. Type handlers don't support multi blocked columns in general, so we can remove this. complication.

cpp/arcticdb/arrow/test/test_arrow_read.cpp
```diff
@@ -365,4 +366,98 @@ TEST(ArrowRead, ConvertSegmentMultipleStringColumns) {
...
     }
-}
\ No newline at end of file
+}
+
+struct TimestampConvertParam {
+    std::string name;
+    bool sparse;
+    bool has_nats;
+    size_t source_chunk_size; // 0 = single block, >0 = multi-block with this chunk size
```
IvoDD, line 375: Again this is not needed. Source column is always contiguous. Only destination column might be fragmented for arrow. If dest column is fragmented we handle this with multiple calls to `convert_column`. 

Was this the main reason these C++ tests were added? If so we don't get much benefit over the existing python tests.

python/tests/unit/arcticdb/version_store/test_arrow_read.py
IvoDD: It would be also good to parametrize the row_range to come from querybuilder. This way we'll handle both the memcpy and the no memcpy codepaths.

cpp/arcticdb/arrow/arrow_handlers.cpp
ognyanstoimenov: This is for the case where `convert_type` is not called from `handle_type`. Eg when it is called from `copy_frame_data_to_buffer` `column_data` has more than 1 contiguous buffer

cpp/arcticdb/arrow/test/test_arrow_read.cpp
ognyanstoimenov: Yes, cpp tests are added only to cover the logic going through multiple chunks of the source column. But lets discuss offline

python/tests/unit/arcticdb/version_store/test_arrow_read.py
ognyanstoimenov: Yes, I wanted to test with a higher segment count. Will parameterize

python/tests/unit/arcticdb/version_store/test_arrow_read.py
```diff
@@ -184,8 +184,12 @@ def test_strings_with_nones_and_nans(lmdb_version_store_tiny_segment, row_range,
 
 
 @pytest.mark.parametrize("row_range", [None, (2, 3), (2, 5), (2, 6), (3, 5)])
-def test_datetime_col_with_nats(lmdb_version_store_arrow, row_range):
-    lib = lmdb_version_store_arrow
+@pytest.mark.parametrize("segment_size", [None, 1, 2, 1000])
+@pytest.mark.parametrize("use_query_builder", [False, True])
+def test_datetime_col_with_nats(version_store_factory, row_range, segment_size, use_query_builder):
+    slicing_params = {"segment_row_size": segment_size, "column_group_size": segment_size} if segment_size is not None else {}
```
IvoDD: Nit: `column_group_size` doesn't really matter for a single column dataframe

python/tests/unit/arcticdb/version_store/test_arrow_read.py
```diff
@@ -203,14 +207,20 @@ def test_datetime_col_with_nats(lmdb_version_store_arrow, row_range):
         }
     )
     lib.write("arrow", df)
-    table = lib.read("arrow", row_range=row_range).data
-    expected = lib.read("arrow", row_range=row_range, output_format=OutputFormat.PANDAS).data
+    if use_query_builder and row_range is not None:
+        q = QueryBuilder().row_range(row_range)
+        table = lib.read("arrow", query_builder=q).data
+        expected = lib.read("arrow", query_builder=q, output_format=OutputFormat.PANDAS).data
```
IvoDD, line 212: Might be better to construct expected from the input df like:
```
expected  = df
if row_range is not None:
    expected = expected.iloc[row_range[0]:row_range[1]].reset_index(drop=True)
```

This will ensure we catch regressions happening both on pandas and arrow path together.

We'll also have a single expected calculation for the two cases.

IvoDD (approved):

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -522,4 +523,105 @@ void ArrowBoolHandler::default_initialize(ChunkedBuffer&, size_t, size_t, const
...
+    const auto end_idx = positions.end_idx_after_truncation.value_or(m.num_rows_);
+    using TimestampTag = ScalarTagType<DataTypeTag<DataType::NANOSECONDS_UTC64>>;
+    util::BitSet validity;
+    util::BitSet::bulk_insert_iterator inserter(validity);
+
+    if (is_sparse) {
+        for_each_enumerated_flattened<TimestampTag>(
+                source_column,
+                [&] ARCTICDB_LAMBDA_INLINE(const auto& en) {
+                    dest[en.idx()] = en.value();
```
alexowens90: Don't know if this will have a measurable impact on the performance, but I think this could be inside the if statement below?

cpp/arcticdb/arrow/arrow_handlers.cpp
```diff
@@ -522,4 +523,105 @@ void ArrowBoolHandler::default_initialize(ChunkedBuffer&, size_t, size_t, const
...
+        convert_type(const Column& source_column, Column& dest_column, const ColumnMapping& m, const DecodePathData&, std::any&, const std::shared_ptr<StringPool>&, const ReadOptions&)
+                const {
+    auto* dest = reinterpret_cast<timestamp*>(dest_column.bytes_at(m.offset_bytes_, m.dest_bytes_));
+    const auto positions = get_positions_after_truncation(m);
+
+    const bool is_sparse = source_column.opt_sparse_map().has_value();
+    const auto first_idx = positions.first_idx_after_truncation.value_or(0);
+    const auto end_idx = positions.end_idx_after_truncation.value_or(m.num_rows_);
+    using TimestampTag = ScalarTagType<DataTypeTag<DataType::NANOSECONDS_UTC64>>;
+    util::BitSet validity;
```
alexowens90, line 569: nit: can presize `validity` here rather than resizing at the end

alexowens90 (approved):

PR #3034, opened 2026-04-15
Bug: `read(as_of=<timestamp>)` ignores snapshot-protected deleted versions when any live version exists

jamesblackburn:
## Bug: `read(as_of=<timestamp>)` ignores snapshot-protected deleted versions when any live version exists   


This is an interesting edge-case, and I can see the argument both ways:                

 ### Minimal reproduction

```
  write(sym, 0)          # V0 — remains alive
  write(sym, 1)          # V1, timestamp captured                                                                                           
  snapshot("snap")       # snap protects V1
  delete_version(sym, 1) # V1 tombstoned; snap keeps its data keys alive                                                                    
                                                                             
  read(sym, as_of=1).data            == 1  # OK   — version-id path searches snapshots                                                      
  read(sym, as_of=ts_after_V1).data  == 1  # FAIL — returns V0 (data=0) instead
```

* Version is available via snapshot, but isn't via timestamp read as it's been 'deleted'
* However prune_previous_versions would automatically be expected to clean up non-snapshotted versions

... and versions which exist under snapshot shouldn't really be considered as deleted... Certainly from the point of view reads with timestamp.
                                                          
  ### Root cause                                                             

  The snapshot fallback only fires when `load_index_key_from_time` returns `nullopt` — i.e. when
  *no* undeleted version exists before the requested timestamp. If an older live version is present,
  it is returned silently, even if a newer version was current at that time and is protected by a                                           
  snapshot:
                                                                                                                                            
  ```cpp                                                                     
  auto index_key = load_index_key_from_time(store(), version_map(), stream_id, as_of);
  if (!index_key && ...iterate_snapshots_if_tombstoned) {                                                                                   
      // snapshot search — never reached if any live version exists
  }                                                                                                                                         
 ```                                                                            
                                                                                                                                            
###  Additional finding                                                         
                                                                                                                                            
  The V2 Library API hardcodes iterate_snapshots_if_tombstoned=False at every call site in
  library.py, so the snapshot fallback is disabled entirely for V2 users — both for timestamp and
  version-id reads. No comments explain this choice.                                                                                        
   
###  Coverage                                                                                                                                  
                                                                             
  This behaviour doesn't appear to be documented, or noted anywhere in the codebase.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -337,15 +337,27 @@ std::optional<VersionedItem> LocalVersionedEngine::get_specific_version(
 std::optional<VersionedItem> LocalVersionedEngine::get_version_at_time(
         const StreamId& stream_id, timestamp as_of, const VersionQuery& version_query
 ) {
+    LoadStrategy load_strategy{LoadType::FROM_TIME, LoadObjective::UNDELETED_ONLY, as_of};
+    auto entry = version_map()->check_reload(store(), stream_id, load_strategy, __FUNCTION__);
 
-    auto index_key = load_index_key_from_time(store(), version_map(), stream_id, as_of);
-    if (!index_key && std::get<TimestampVersionQuery>(version_query.content_).iterate_snapshots_if_tombstoned) {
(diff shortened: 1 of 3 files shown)


jamesblackburn:
Working as designed.

PR #3036, opened 2026-04-15
Fold bi_benchmarks.py in to query_builder.py

poodlewars:
bi_benchmarks.py tested with 1M rows and 10M rows, like query_builder.py. All it covered that query_builder.py did not were:

- String columns with a high null count
- String equality comparisons
- Count aggregations

so I've just added benchmarks for these in query_builder.py.

test_basic_operations_scenarios.py was using the very relaxed `from benchmarks.bi_benchmarks import assert_frame_equal`

 ```
 def assert_frame_equal(pandas_df: pd.DataFrame, arctic_df: pd.DataFrame):
     arctic_df.sort_index(inplace=True)
     test.assert_frame_equal(pandas_df, arctic_df, check_column_type=False, check_dtype=False)
```

 I've replaced it with a stricter assertion.

Example flaky BI run: https://chat-man.slack.com/archives/C02NEG3V38X/p1776241610547619

### CODE REVIEW

File: python/benchmarks/query_builder.py [hunk 1]
```diff
@@ -61,91 +80,163 @@ def setup(self, lib_for_storage, num_rows, storage):
...
+        result = self.lib.read(self.null_symbol, columns=["v3"], query_builder=q)
+        assert not result.data.empty
+
+    def peakmem_filtering_string_isin_with_nulls(self, lib_for_storage, num_rows, storage):
+        k = num_rows // 1000
+        string_set = [f"id{str(i).zfill(3)}" for i in range(1, k + 1)]
+        q = QueryBuilder()
+        q = q[q["id1"].isin(string_set)]
+        result = self.lib.read(self.null_symbol, columns=["v3"], query_builder=q)
+        assert not result.data.empty
```
vasil-pashov: There's some repetition here and in a few other function. What do you think about pulling it into a function?

vasil-pashov (approved):

alexowens90 (approved):

PR #3038, opened 2026-04-16
Column stats: col-col comparisons

poodlewars:
#### Reference Issues/PRs
11649170310 

Support column stats in cross column comparisons like `q["a"] < q["b"]`.

#### Any other comments?

The interesting part of this is the dynamic schema handling. If we compare `q["a"] < q["b"]` then our engine drops segments where either of those two columns are missing.

We plan to make changes so that users have to create stats over all numeric columns - https://man312219.monday.com/boards/7852509418/pulses/11803431071 . It will not be possible to have stats in a given row range for some numeric columns and not others.

We can then exploit the sparsity of the stats columns - min and max are both sparse. If we see missing entries for min and max for a given column at a given start_index and end_index, we infer that that column is missing in that range and we can safely exclude it when evaluating column stats.

There was a bug here where stats segments "lost" their sparsity when they were merged together, which this PR fixes.


cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -8,7 +8,7 @@
 namespace arcticdb {
 
 SegmentInMemory merge_column_stats_segments(const std::vector<SegmentInMemory>& segments) {
-    SegmentInMemory merged;
+    SegmentInMemory merged(Sparsity::PERMITTED);
```
poodlewars, line 11: This was a bug in stats creation. In the `MinMaxAggregator` we already created sparse stats columns, but we lost the sparsity when merging stats together.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2158,6 +2158,7 @@ FrameAndDescriptor read_column_stats_impl(const std::shared_ptr<Store>& store, c
     try {
         auto segment = store->read_compressed(column_stats_key).get().segment_ptr();
         auto segment_in_memory = decode_segment(*segment, AllocationType::DETACHABLE);
+        segment_in_memory.unsparsify();
```
alexowens90: We should probably always return Arrow structures for this, zeros in missing int columns will be confusing

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -505,30 +567,32 @@ struct EqualsOperator {
...
+
+    template<typename T, typename U>
+    StatsComparison operator()(ValueRange<T> lhs, ValueRange<U> rhs) const {
+        if (auto r = check_range_for_nan(lhs, StatsComparison::NONE_MATCH))
+            return *r;
+        if (auto r = check_range_for_nan(rhs, StatsComparison::NONE_MATCH))
+            return *r;
+
+        if ((*this)(lhs.min, rhs.min) && (*this)(lhs.max, rhs.max) && (*this)(lhs.min, lhs.max) &&
+            (*this)(rhs.min, rhs.max))
```
alexowens90: nit: This has one more comparison than necessary, unlike `NotEqualsOperator`

cpp/arcticdb/pipeline/test/test_column_stats_data.cpp
```diff
@@ -207,3 +207,126 @@ TEST(ColumnStatsData, DuplicateIndexPairDoesNotAffectOtherRows) {
...
+    min_vol_col->mark_absent_rows(1);
+    max_vol_col->mark_absent_rows(1);
+
+    ssize_t last_row = 1;
+    start_col->set_row_data(last_row);
+    end_col->set_row_data(last_row);
+    min_price_col->set_row_data(last_row);
+    max_price_col->set_row_data(last_row);
+    min_vol_col->set_row_data(last_row);
+    max_vol_col->set_row_data(last_row);
```
alexowens90: These can be omitted, calling `set_row_data` on the segment in memory will also set it on it's columns

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_filtering_hypothesis.py
```diff
@@ -310,12 +310,15 @@ def test_filter_not(lmdb_version_store_v1, any_output_format, column_stats_filte
...
-def test_filter_more_columns_than_fit_in_one_segment(lmdb_version_store_tiny_segment, any_output_format, df):
+def test_filter_more_columns_than_fit_in_one_segment(
+    lmdb_version_store_tiny_segment, any_output_format, column_stats_filtering_enabled_and_disabled, df
+):
     assume(not df.empty)
     lib = lmdb_version_store_tiny_segment
     lib._set_output_format_for_pipeline_tests(any_output_format)
     symbol = "test_filter_more_columns_than_fit_in_one_segment"
     lib.write(symbol, df)
+    lib.create_column_stats(symbol, {"a": {"MINMAX"}, "b": {"MINMAX"}, "c": {"MINMAX"}})
```
vasil-pashov, line 321: Is it intentional that we generate column stats even in the case when we don't need them?

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -1372,6 +1372,447 @@ def test_column_stats_duplicate_timestamp_index_still_prunes_unique_keys(
...
+        pytest.param(lambda q: q["a"] <= q["b"], 1, id="lte_nat_prunes_seg1"),
+        pytest.param(lambda q: q["a"] >= q["b"], 1, id="gte_nat_prunes_seg0"),
+    ],
+)
+def test_column_stats_cross_column_comparison_nat(
+    in_memory_version_store, clear_query_stats, column_stats_filtering_enabled, query_expr, expected_reads
+):
+    """Test that NaT is treated as a normal value (int64_min) for column-vs-column stats pruning.
+
+    ArcticDB treats NaT as int64_min in all comparisons (unlike Pandas which treats NaT as missing).
```
vasil-pashov, line 1536: This doesn't seem right to me. In resampling and groupby NaT behaves like a a NaN. We also decided that when we have sparse data NaT will be handled as missing value

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -1372,6 +1372,447 @@ def test_column_stats_duplicate_timestamp_index_still_prunes_unique_keys(
...
+    assert_frame_equal(result, expected, check_dtype=False)
+    assert table_data_reads == expected_reads, f"Expected {expected_reads} TABLE_DATA read(s), got {table_data_reads}"
+
+
+def test_column_stats_dynamic_schema_cross_column_backfilled_values_int(
+    in_memory_version_store_dynamic_schema,
+    clear_query_stats,
+    column_stats_filtering_enabled,
+):
+    """When we read with dynamic schema, we backfill missing columns with a default, like NaN or 0.
```
vasil-pashov, line 1743: Nice!

cpp/arcticdb/pipeline/column_stats_dispatch.hpp
```diff
@@ -136,6 +139,41 @@ StatsComparison stats_comparator(const ColumnStatsValues& stats_lhs, const Value
     });
 }
 
+template<typename Func>
+StatsComparison stats_comparator(const ColumnStatsValues& stats_lhs, const ColumnStatsValues& stats_rhs, Func&& func) {
+    if (stats_lhs.column_absent || stats_rhs.column_absent) {
+        return StatsComparison::NONE_MATCH;
+    }
+    if (!stats_lhs.min || !stats_rhs.min) {
```
vasil-pashov, line 147: Shouldn't there be the same but for `max`?

python/tests/unit/arcticdb/version_store/test_filtering_hypothesis.py
poodlewars: Yeah, I want to make sure the test goes through the column stats code path so we know if column stats start creating wrong results.

cpp/arcticdb/pipeline/column_stats_dispatch.hpp
poodlewars: We have checks when we load the stats that `min` is present iff `max` is present, so we only need to check one here.

python/tests/unit/arcticdb/test_column_stats_optimisation.py
poodlewars: Raised 11688915824 and picking it up this week

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -1772,13 +1772,14 @@ def test_column_stats_dynamic_schema_cross_column_backfilled_values_int(
     assert table_data_reads == 1
 
 
+@pytest.mark.xfail(reason="Need a NaN count or flag, Monday=11866077783")
```
poodlewars, line 1775: This started failing after rebasing on my changes to how we write NaN stats (min or max is NaN iff entire block is NaN). To implement this we need to add a NaN count statistic, which we'll pick up after re-working the `MinMaxAggregator`.

vasil-pashov (approved):

PR #3039, opened 2026-04-16
Reduce size of the column stats wide filter benchmark so it executes faster

poodlewars:
This is the slowest benchmark we have at the moment.

After this change, it still shows a ~40% difference with and without stats.

In https://github.com/man-group/ArcticDB/actions/runs/24465438801 this benchmark took ~8 minutes of setup time and ~9 minutes of execution time, contributing about 10% of the total benchmarking runtime, so we need to speed it up.

This speeds it up by a factor of 5 to about 5 minutes.


alexowens90 (approved):

maxim-morozov (approved):
LGTM

PR #3042, opened 2026-04-16
Remove unused libiconv (lgpl) dependency

jamesblackburn:
The EncodingConversion class wrapping iconv() was never instantiated — only PortableEncodingConversion is used (in string_reducers.hpp), which does manual UTF-8 to UTF-32 widening without calling iconv.

Remove the dead class, the #include <iconv.h>, and the libiconv vcpkg dependency and version override.

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -103,11 +103,7 @@ add_library(xxHash STATIC IMPORTED)
 set_target_properties(xxHash PROPERTIES IMPORTED_LOCATION ${xxHash_LIBRARY})
 set_target_properties(xxHash PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${xxHash_INCLUDE_DIR})
 
-# Libraries baked into the internal and external Linux images are found by searching various paths
-# On Windows, vcpkg will provide a "Config" which takes precedence
-if (WIN32)
-    find_package(Iconv)
-else ()
+if (NOT WIN32)
     find_library(Sasl2_LIBRARY NAMES sasl2 libsasl2 PATHS /usr/local/lib/libsasl2.a REQUIRED)
 endif ()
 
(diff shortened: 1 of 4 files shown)


poodlewars:
Also needs to be removed from the Conda dev recipe https://github.com/man-group/ArcticDB/blob/master/environment-dev.yml#L60 and from the feedstock `arcticdb-feedstock` during the next release.

Have you confirmed that `libiconv` is not installed transitively after this change?

jamesblackburn:
> Also needs to be removed from the Conda dev recipe https://github.com/man-group/ArcticDB/blob/master/environment-dev.yml#L60 and from the feedstock `arcticdb-feedstock` during the next release.

@poodlewars I did have that originally, but it blows up the whole conda build, so didn't include for now 😬

jamesblackburn:
> Have you confirmed that `libiconv` is not installed transitively after this change?

I did statically, but also followed up with a dynamic test too. The main thought is this is Linux only and may be different on other platforms - but then it would be a transitive dependency, so I don't see that this can be critical:

```  
  No libiconv dependency — confirmed at all levels:

  1. .so NEEDED libs — objdump -p arcticdb_ext.so shows only libc.so.6, libstdc++.so.6, libm.so.6,
  libresolv.so.2, libgcc_s.so.1. No libiconv.so.
  2. iconv symbols — the iconv, iconv_open, iconv_close symbols that remain are versioned @GLIBC_2.2.5 — they
  come from glibc's built-in iconv, not from the standalone libiconv package.
  3. vcpkg install — libiconv_1.18_x64-linux.list only installed cmake wrapper files (no .a or .so). The
  wrapper's Iconv_IS_BUILT_IN path was taken, meaning it used the system libc.
  4. Python deps — setup.cfg install_requires has no iconv entry; none of the runtime deps (numpy, pandas,
  protobuf, etc.) pull it in.
  5. Process map — /proc/maps shows no libiconv.so or libcharset.so loaded before or after import arcticdb.

```

maxim-morozov (approved):
LGTM, can you please check the conda eenvironment

poodlewars (changes requested):
Marking needs work due to the Conda failures

jamesblackburn:
Depends on: https://github.com/man-group/ArcticDB/pull/3052

poodlewars (approved):

PR #3043, opened 2026-04-16
Use UnsortedDataException not SortingException in update

poodlewars:
Use UnsortedDataException (the specific exception type) not SortingException (the category).

This is a very minor API break. UnsortedDataException is a sub-type of SortingException so this is backwards compatible, except that users catching UnsortedDataException will now catch unsorted data from update calls too. Marked "minor" to avoid scaring users with a major version bump for something so trivial.

Also update tests to check for UnsortedDataException rather than the more general SortingException.

Customer request: https://arcticdb.slack.com/archives/C05ATNAR3PA/p1776368548164199

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -37,7 +37,7 @@ from arcticdb.exceptions import (
     ArcticNativeException,
     ArcticDbNotYetImplemented,
     NormalizationException,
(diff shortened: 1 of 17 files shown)


G-D-Petrov (approved):
LGTM but Claude's comments about the duplicated imports are sensible.

alexowens90 (approved):

PR #3045, opened 2026-04-17
Implement POW function

markovskipetar:
# POW Function Implementation

The `**` (power) operator was added to ArcticDB's `QueryBuilder` as a binary arithmetic projection,
allowing users to raise a column's values to a power inside a query.

## Usage

```python
from arcticdb import QueryBuilder
import numpy as np

q = QueryBuilder()

# column ** column
q = q.apply("RESULT", q["BASE"] ** q["EXP"])

# column ** scalar
q = q.apply("CUBED", q["VALUE"] ** np.uint8(3))

# scalar ** column  (rpow)
q = q.apply("TWO_TO_N", np.uint8(2) ** q["N"])

data = lib.read("my_symbol", query_builder=q).data
```

## Result Type Rules

The output data type depends on the signedness of both operands and  data type promotion is performed if required 

| Base type         | Exponent type | Result type |
|-------------------|---------------|-------------|
| unsigned integer  | unsigned int  | `uint64`    |
| signed integer    | unsigned int  | `int64`     |
| unsigned integer  | signed int    | `float64`   |
| signed integer    | signed int    | `float64`   |
| float             | integer       | `float64`   |

The signed-int exponent cases promote to `float64` because a negative exponent produces a
fractional result (e.g. `2 ** -3 = 0.125`).

### Type-promotion rules (compile-time)

`binary_operation_promoted_type` was extended to handle `PowOperator` as a special case before
the general integer rules:

```
if Func == PowOperator:
    if exponent is unsigned:
        base unsigned  → uint64
        base signed    → int64
    else (exponent signed):
        → double
else:
    (existing +/-/*/÷ rules unchanged)
```

## Restrictions

- **Float exponents are not allowed.** Passing a `float32`/`float64` column or scalar as the
  exponent raises `UserInputException`.
- **String operands are not allowed.** Both operands must be numeric; strings raise
  `UserInputException`.
- Only the **exponent** drives the float-promotion decision — the base can be any numeric type.

The integer paths use a naive loop instead of `std::pow` to avoid floating-point rounding
errors that appear for large integers (e.g. `2^63`).

## Python Layer

`ExpressionNode` in `python/arcticdb/version_store/processing.py` gained two dunder methods:

```python
def __pow__(self, right):
    return self._apply(right, _OperationType.POW)

def __rpow__(self, left):
    return self._rapply(left, _OperationType.POW)
```

### Scalar value handling fix

`create_value()` had a bug where a positive `np.integer` scalar with a signed dtype could be
stored in a type too narrow to represent it (e.g. `np.int16(128)` was mapped to `int8` because
`np.min_scalar_type(-128) = int8`, but `int8.max = 127 < 128`). The fix:

1. For positive signed integers, compute `min_scalar_type` from `-value` (to get the signed
   range), then check if the positive value still fits; if not, widen by one step.

## Tests

Tests live in `python/tests/unit/arcticdb/version_store/test_projection.py`:

| Test | What it checks |
|---|---|
| `test_project_pow_col_col` | All four type-promotion combinations (uint^uint, int^uint, uint^int, int^int) |
| `test_project_pow_col_value` | Column raised to a scalar constant |
| `test_project_pow_float_base` | Float base raised to integer exponent |
| `test_project_pow_string_raises` | `UserInputException` when either operand is a string |
| `test_project_pow_float_exponent_raises` | `UserInputException` for float exponent column or scalar |
| `test_project_pow_dynamic` | Pow under dynamic schema |

Hypothesis-based property tests are in
`python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py` and
`python/arcticdb/util/hypothesis.py`.


jamesblackburn:
Will need some docs

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -189,62 +175,80 @@ struct binary_operation_promoted_type {
...
-                                             * e.g. 0 - 255 (both uint8_t's) = -255, requiring int16_t to hold the
-                                             * result */
-                                            arithmetic_promoted_type::details::signed_width_t<2 * max_width>,
-                                            // IsIn/IsNotIn operators, just use the type of the widest input
-                                            arithmetic_promoted_type::details::unsigned_width_t<max_width>>>,
+                                            std::is_unsigned_v<LHS>,
+                                            // uint ^ uint -> uint64
+                                            uint64_t,
+                                            // int ^ uint -> int64
+                                            int64_t>,
```
alexowens90, line 240: This isn't correct if LHS is a float

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -373,6 +377,73 @@ struct DivideOperator {
     }
 };
 
+struct PowOperator {
+    template<typename T, typename U, typename V = typename binary_operation_promoted_type<T, U, PowOperator>::type>
+    V apply(T t, U u) {
+        return static_cast<V>(std::pow(static_cast<V>(t), static_cast<V>(u)));
```
alexowens90, line 625: This will not be exact for uint**uint for large output values

cpp/arcticdb/processing/test/test_output_schema_ast_validity.cpp
```diff
@@ -179,6 +179,96 @@ TEST_F(AstParsingOutputTypesTest, FilterIsInStringColStringValueSet) {
...
+            "root", std::make_shared<ExpressionNode>(ColumnName("int32"), ColumnName("uint8"), OperationType::POW)
+    );
+    ProjectClause project_clause{{"int32", "uint8"}, "root", ec};
+    auto output_schema = project_clause.modify_schema(initial_schema());
+    auto stream_desc = initial_schema().stream_descriptor();
+    stream_desc.add_scalar_field(DataType::INT64, "root");
+    ASSERT_EQ(output_schema.stream_descriptor(), stream_desc);
+}
+
+TEST_F(AstParsingOutputTypesTest, ProjectionPowUintIntCols) {
```
alexowens90, line 206: Needs removing

cpp/arcticdb/processing/test/test_output_schema_ast_validity.cpp
```diff
@@ -179,6 +179,96 @@ TEST_F(AstParsingOutputTypesTest, FilterIsInStringColStringValueSet) {
...
+    ASSERT_THROW(project_clause.modify_schema(initial_schema()), UserInputException);
+}
+
+TEST_F(AstParsingOutputTypesTest, ProjectionPowNumericColBitset) {
+    ec.add_expression_node("bitset", bitset_node());
+    ec.add_expression_node(
+            "root", std::make_shared<ExpressionNode>(ColumnName("int32"), ExpressionName("bitset"), OperationType::POW)
+    );
+    ProjectClause project_clause{{"int32", "uint8"}, "root", ec};
+    ASSERT_THROW(project_clause.modify_schema(initial_schema()), UserInputException);
```
alexowens90, line 258: bitset**int should also throw

cpp/arcticdb/processing/test/test_output_schema_ast_validity.cpp
```diff
@@ -179,6 +179,96 @@ TEST_F(AstParsingOutputTypesTest, FilterIsInStringColStringValueSet) {
...
+
+TEST_F(AstParsingOutputTypesTest, ProjectionPowNumericColValueSet) {
+    std::unordered_set<uint8_t> raw_set{1, 2, 3};
+    auto value_set = std::make_shared<ValueSet>(std::make_shared<std::unordered_set<uint8_t>>(std::move(raw_set)));
+    ec.add_value_set("value_set", value_set);
+    ec.add_expression_node(
+            "root", std::make_shared<ExpressionNode>(ColumnName("uint8"), ValueSetName("value_set"), OperationType::POW)
+    );
+    ProjectClause project_clause{{"uint8"}, "root", ec};
+    ASSERT_THROW(project_clause.modify_schema(initial_schema()), UserInputException);
```
alexowens90, line 269: valueset**number should also throw

cpp/arcticdb/processing/operation_types.hpp
markovskipetar: 10 lines above there is this check: `std::is_floating_point_v<LHS> || std::is_floating_point_v<RHS>` if it's true -> double Is used

cpp/arcticdb/processing/test/test_output_schema_ast_validity.cpp
markovskipetar: after I review the code with the help of Claude, the ai said that:

The shared case OperationType::ADD/SUB/MUL/DIV/POW block (lines 176–190) already:

  - Throws on bitset ** int (left operand check, line 176–180)

cpp/arcticdb/processing/test/test_output_schema_ast_validity.cpp
markovskipetar: after I review the code with the help of Claude, the ai said that:

The shared case OperationType::ADD/SUB/MUL/DIV/POW block (lines 176–190) already:

  - Throws on valueset ** number (right value-set check, line 186–190)

cpp/arcticdb/processing/test/test_arithmetic_type_promotion.cpp
```diff
@@ -569,6 +569,111 @@ TEST(ArithmeticTypePromotion, Divide) {
     static_assert(std::is_same_v<binary_operation_promoted_type<double, int64_t, DivideOperator>::type, double>);
 }
 
+TEST(ArithmeticTypePromotion, Pow) {
+    using namespace arcticdb;
+    // Floating point types should promote to the larger type width
```
alexowens90: This comment is now out of date

cpp/arcticdb/processing/operation_types.hpp
alexowens90: Good point

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -373,6 +377,85 @@ struct DivideOperator {
...
+            uint64_t result = 1;
+            uint64_t base = static_cast<uint64_t>(t);
+            uint64_t exponent = static_cast<uint64_t>(u);
+
+            for (uint64_t i = 0; i < exponent; ++i) {
+                result *= base;
+            }
+
+            return result;
+        }
```
alexowens90, line 623: This needs an
```
else if constexptr (std::is_signed_v<T> && std::is_unsigned_v<U>) {...}
```
block so that we get exact results for operations like `(-10)**15` as well

cpp/arcticdb/processing/test/test_output_schema_ast_validity.cpp
alexowens90: Sorry, I believe that it does, what this comment really meant was that we should have the equivalent test to this with the operands the other way around, which I see has now been added below

cpp/arcticdb/processing/test/test_output_schema_ast_validity.cpp
alexowens90: Same reply as above

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
```diff
@@ -36,8 +36,14 @@
...
     val=numeric_type_strategies(),
 )
+
 def test_project_numeric_binary_operation(lmdb_version_store_v1, any_output_format, df, val):
     assume(not df.empty)
+    # Guards for POW: avoid overflow and NaN from negative-base-fractional-exponent
+    assume(df["b"].abs().max() < 50)
+    assume(abs(val) < 50)
+    assume((df["a"] >= 0).all())
+    assume(val >= 0)
```
alexowens90: These restrictions are quite heavy for the other operators being tested here. I would split hypothesis testing of the `pow` operator into their own tests so that we don't lose coverage of the other operators

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
```diff
@@ -123,8 +132,14 @@ def test_project_numeric_unary_operation(lmdb_version_store_v1, any_output_forma
...
     val=numeric_type_strategies(),
 )
+
 def test_project_numeric_binary_operation_dynamic(lmdb_version_store_dynamic_schema_v1, any_output_format, df, val):
     assume(len(df) >= 3)
+    # Guards for POW: avoid overflow and NaN from negative-base-fractional-exponent
+    assume(df["b"].abs().max() < 50)
+    assume(abs(val) < 50)
+    assume((df["a"] >= 0).all())
+    assume(val >= 0)
```
alexowens90: Same comment as above

python/tests/unit/arcticdb/version_store/test_projection.py
```diff
@@ -109,99 +148,176 @@ def test_docstring_example_query_builder_apply(lmdb_version_store_v1, any_output
...
+
     q = QueryBuilder()
-    q = q.apply("result", getattr(q["a"], op)(value))
-    with pytest.raises(UserInputException):
-        lib.read(symbol, query_builder=q)
+    q = q.apply("RESULT", q["BASE"] ** q["EXP"])
+    data = lib.read("pow_uint_uint", query_builder=q).data
+
+    df["RESULT"] = df["BASE"].astype(np.uint64) ** df["EXP"].astype(np.uint64)
+    assert_frame_equal(df.astype({"RESULT": "uint64"}), data)
```
alexowens90: I think this `astype` can be removed? And in the tests below as well

python/tests/unit/arcticdb/version_store/test_projection.py
```diff
@@ -109,99 +148,176 @@ def test_docstring_example_query_builder_apply(lmdb_version_store_v1, any_output
...
+        {
+            "BASE": np.arange(1, 11, dtype=np.uint16),
+            "EXP": np.arange(1, 11, dtype=np.uint8),
+        },
+        index=np.arange(10),
+    )
+
+    expected, slices = make_dynamic(df)
+    for df_slice in slices:
+        lib.append(symbol, df_slice, write_if_missing=True)
```
alexowens90:
```suggestion
        lib.append(symbol, df_slice)
```
Calling `append` if the symbol doesn't exist writes the data anyway by default

python/tests/unit/arcticdb/version_store/test_projection.py
alexowens90, line 1: A couple more tests I think would be good to add:

- Check that a sensible exception is thrown with a floating point exponent (column or value)
- Some Python tests raising a floating point base to unsigned and signed exponents

alexowens90 (changes requested):

python/arcticdb/version_store/processing.py
alexowens90, line 1: The docstring for the `QueryBuilder` class needs to be updated with what we do (and don't!) support with this new operator

cpp/arcticdb/processing/test/test_arithmetic_type_promotion.cpp
markovskipetar: done!

cpp/arcticdb/processing/operation_types.hpp
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection.py
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection.py
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection.py
markovskipetar: done!

python/arcticdb/version_store/processing.py
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
markovskipetar: done!

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_projection.py
```diff
@@ -109,6 +148,209 @@ def test_docstring_example_query_builder_apply(lmdb_version_store_v1, any_output
     df["ADJUSTED"] = df["ASK"] * df["VOL_ACC"] + 7
     assert_frame_equal(df.astype({"ADJUSTED": "int64"}), data)
 
+###############################################
+# POW OPERATOR TESTS                          #
+###############################################
```
ognyanstoimenov, line 115: Most of the tests below have a similar structure. Can we parametarize them instead of having them duplicated?

python/tests/unit/arcticdb/version_store/test_projection.py
```diff
@@ -88,6 +88,45 @@ def test_project_value_set():
         QueryBuilder().apply("new_col", [0, 1, 2])
 
 
+def test_pow_type_coercion(lmdb_version_store_v1, any_output_format):
```
ognyanstoimenov: Could be more clean if you fold this in the tests below. No strong opinion though

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
```diff
@@ -123,6 +170,7 @@ def test_project_numeric_unary_operation(lmdb_version_store_v1, any_output_forma
     ),
     val=numeric_type_strategies(),
 )
+
```
ognyanstoimenov: Random line

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
```diff
@@ -36,6 +36,7 @@
...
```
ognyanstoimenov: Random line

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
```diff
@@ -77,6 +78,52 @@ def test_project_numeric_binary_operation(lmdb_version_store_v1, any_output_form
...
+@given(
+    df=dataframe_strategy(
+        [
+            column_strategy("a", supported_numeric_dtypes(), restrict_range=True),
+            column_strategy("b", supported_numeric_dtypes(), restrict_range=True),
+        ],
+    ),
+    val=numeric_type_strategies(),
+)
+def test_project_pow_numeric_binary_operation(lmdb_version_store_v1, any_output_format, df, val):
```
ognyanstoimenov, line 96: Add hypothesis tests for dynamic schema?

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -356,6 +377,37 @@ struct DivideOperator {
...
+        else if constexptr (std::is_signed_v<T> && std::is_unsigned_v<U>) {
```
ognyanstoimenov: Typo `constexptr`. This probably wont compile

ognyanstoimenov (commented):
Mostly test suggestions and a typo that won't compile. Also, there are some merge conflicts

cpp/arcticdb/processing/operation_types.hpp
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection_hypothesis.py
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection.py
markovskipetar: done!

python/tests/unit/arcticdb/version_store/test_projection.py
markovskipetar: done!

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -373,6 +394,97 @@ struct DivideOperator {
...
+            }
+
+            return result;
+        }
+        else if constexpr (std::is_signed_v<T> && std::is_unsigned_v<U>) {
+            int64_t result = 1;
+            int64_t base = static_cast<int64_t>(t);
+            uint64_t exponent = static_cast<uint64_t>(u);
+
+            for (uint64_t i = 0; i < exponent; ++i) {
```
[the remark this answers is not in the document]
ognyanstoimenov: Claude may be right here?

ognyanstoimenov (approved):
Dynamic schema hypothesis tests fail but otherwise LGTM

PR #3048, opened 2026-04-20
Improve batch reads performance on Windows

G-D-Petrov:
#### Reference Issues/PRs
Monday ticket ref: 11777245573

On a wide-schema read benchmark (20 symbols × 35 000 columns × 320 rows,
V1 encoding), `Library.read_batch` was running ~**2× slower** than calling
`Library.read` in a serial loop. 

`FieldCollection::add_field` calls `Buffer::ensure(n_bytes)` once per
appended field. `Buffer::ensure` resizes the underlying allocation to
*exactly* `bytes + preamble_bytes_` on every growth, so each append
triggers a fresh `Allocator::realloc` + memcpy of the existing
`body_bytes_`. N appends therefore cost O(N²) in copy work.

The same O(N²) pattern exists in two places:

1. **`FieldCollection::add_field`** — one per column on the V1 decode path,
   when converting protobuf descriptors into an in-memory `FieldCollection`.
2. **`EncodedFieldCollection::regenerate_offsets`** — one per column on
   **both** V1 and V2 read paths, called from `SegmentHeader::set_body_fields`
   to rebuild the offset index from the stored field data.

## Fix

Rather than changing `Buffer::ensure` itself (a general-purpose primitive
used throughout the codebase), the fix is localised to the two callers
that exhibit the append-per-item pattern and can cheaply know the final
size up front.

### `FieldCollection` — new sizing constructor

[`cpp/arcticdb/entity/field_collection.hpp`](cpp/arcticdb/entity/field_collection.hpp)

```cpp
explicit FieldCollection(std::size_t num_fields, std::size_t total_data_bytes = 0) {
    if (num_fields == 0)
        return;
    const std::size_t index_bytes = num_fields * sizeof(shape_t);
    allocate_shapes(index_bytes);   // Buffer -> one reserve
    allocate_offsets(index_bytes);  // Buffer -> one reserve
    if (total_data_bytes > 0)
        buffer_.buffer().reserve(total_data_bytes);  // ChunkedBuffer -> one block
}
```

- `shapes_` and `offsets_` are `Buffer`-backed. Sized up front to exactly
  `num_fields * 8` bytes, so subsequent `add_field` calls hit the fast
  `total_size <= capacity_` path in `Buffer::ensure` with no reallocation.
- `buffer_` is `ChunkedBuffer`-backed. It does *not* have the O(N²)
  realloc problem (new blocks are appended, old blocks aren't copied), but
  a single `reserve(total_data_bytes)` block is still nicer than 100+
  small 4 KB blocks. Uses the existing `ChunkedBufferImpl::reserve` which
  adds one block without touching `bytes_`/cursor state, so `add_field`'s
  cursor-based write pattern keeps working.

`field_collection_from_proto` in
[`cpp/arcticdb/entity/stream_descriptor.hpp`](cpp/arcticdb/entity/stream_descriptor.hpp)
does a one-pass `Field::calc_size` sum to get `total_data_bytes`, then
constructs the collection with both sizes known.

### `EncodedFieldCollection::regenerate_offsets` — count first, reserve once

[`cpp/arcticdb/codec/encoded_field_collection.hpp`](cpp/arcticdb/codec/encoded_field_collection.hpp)

The loop was growing `offsets_` by 8 bytes per iteration via `ensure`.
Changed to a two-pass form: walk the field data once to count fields,
`offsets_.reserve(count * 8)` once, then populate. The extra walk is cheap
(no allocations, just pointer arithmetic) and saves O(N²) memcpy work.

## Impact

Release build, Windows, 10 symbols × 30 000 columns × 100 rows, LMDB,
V1 encoding. Regression test metric is
`sum(LMDB_SegmentFromBytes total_time_ms across threads)`, so a ratio of
batch/serial ≈ 1 means per-segment decode takes the same wall time alone
as concurrent — i.e. no cross-thread contention inside the timer.

| state                                                    | serial total | batch total | batch/serial |
|----------------------------------------------------------|-------------:|------------:|-------------:|
| pre-fix                                                  |          —   |         —   |   16× – 49×  |
| `Buffer::ensure` geometric growth (tried, reverted)      |          —   |         —   |  2.2× – 4.6× |
| localised `FieldCollection` sizing constructor           |       313 ms |      491 ms |       1.57×  |
| + `FieldCollection` data-bytes preallocation             |       281 ms |      472 ms |       1.68×  |
| + `EncodedFieldCollection::regenerate_offsets` reserve   |   **231 ms** |   **393 ms**|   **1.70×**  |

The last row is the landing state. Versus the middle row (all
field-collection fixes applied), the `regenerate_offsets` change shaves
~17-18% off both serial *and* batch totals but barely moves the ratio.
Interpretation: removing that O(N²) loop is real per-thread work saved,
but it's not the contention bottleneck any more. The residual 1.7× is
coming from allocations *elsewhere* in the decode path (codec scratch
buffers, Python binding interop, etc.) still hitting the shared Windows
heap.

Headroom below the 8× regression threshold: comfortable, ~4.7×.

## Wall-clock wide-schema read benchmark

End-to-end Python timings, same 10 × 30k × 100 config, release build,
Windows. Each measurement is the median of 8 runs after 3 warmup rounds,
`gc.collect()` between runs, interleaved across formats:

| format  | serial (s) | batch (s) | serial/batch |
|---------|-----------:|----------:|-------------:|
| pandas  |      4.51  |     6.97  |        0.65× |
| pyarrow |      5.51  |     7.34  |        0.75× |
| polars  |      6.79  |     7.95  |        0.85× |

**At default thread settings, batch was slower than serial across all
formats.** The Buffer::ensure / FieldCollection / EncodedFieldCollection
fixes closed the parallel-decode gap enormously (batch pandas: pre-fix
~27.6 s → now 6.97 s, a ~4× improvement), but batch was still losing
to serial by ~30%.

**This turned out to be a thread-pool tuning artifact, not a fundamental
characteristic of the batch path.** See the "Thread-pool sweep" section
below: at the default `(20 CPU, 30 IO)` thread counts, the heap mutex
contends on every concurrent allocation. Re-tuning to `(2 CPU, 4 IO)`
brings batch and serial to parity (~0.87× ratio) for LMDB. On S3, batch
is *always* faster than serial regardless of thread count (1.04× – 2.03×).

## Thread-pool sweep (separate but related finding)

While diagnosing why batch was still slower than serial after the allocation
fixes, a thread-pool sweep turned up a much bigger lever than any further
field-collection work: **the default `TaskScheduler` thread counts are
wrong-sized for both LMDB and S3 workloads on this hardware/OS** (20-logical-core
Windows). Defaults derive from `hardware_concurrency()` for CPU and 1.5×
that for IO, giving (CPU=20, IO=30).

### LMDB sweep (10 syms × 30 000 cols × 100 rows, pandas batch, all fixes)

Reordered runs (default first) to expose first-run / OS-cache anomalies:

| cpu | io | serial med | batch med | s/b ratio |
|----:|---:|-----------:|----------:|----------:|
|  20 | 30 |   12.30 *  |     6.79  | (1st-run noise) |
|   8 | 12 |    5.34    |     6.74  | 0.79× |
|   4 |  6 |    4.10    |     5.85  | 0.70× |
| **2** | **4** | **4.18** | **4.73** | **0.88×** |
|  16 | 24 |    4.17    |     6.16  | 0.68× |
|  32 | 48 |    4.50    |     6.79  | 0.66× |
|   8 | 12 |    4.26    |     6.79  | 0.63× |
| **2** | **4** | **4.08** | **4.66** | **0.87×** |
|  20 | 30 |    4.18    |     6.23  | 0.67× |

`(2 CPU, 4 IO)` reproducibly wins for LMDB by ~30%. Reason: this is a
memory-bound decode, and at 20+ concurrent threads the Windows heap mutex
serialises the small allocations anyway. With 6 threads total, mutex
contention drops to near zero and per-thread useful work climbs
proportionally.

### S3 sweep (real S3, eu-west-1, same workload, pandas batch, all fixes)

| cpu | io | serial med | batch med | s/b ratio |
|----:|---:|-----------:|----------:|----------:|
|   2 |  4 |   33.83    |    28.89  | 1.17× |
|   4 |  8 |   53.95    |    26.60  | **2.03×** |
|   8 | 16 |   45.69    |    29.19  | 1.57× |
|   8 | 32 |   35.89    |    34.67  | 1.04× |
| **8** | **64** | 45.48 | **24.96** | **1.82×** |
|  16 | 64 |   40.61    |    31.60  | 1.28× |
|  16 | 128|   37.03    |    31.34  | 1.18× |
|  32 | 128|   34.81    |    30.61  | 1.14× |

**Opposite of LMDB**: batch is *always* faster than serial on S3 (1.04× –
2.03×), and the best batch wall time comes from `(8 CPU, 64 IO)` —
roughly the inverse of the LMDB optimum. Reason: S3 GETs are network-bound
(~50-100 ms RTT), so IO threads hide latency by overlapping requests.
More CPU threads beyond ~8 just add heap-mutex contention without adding
parallelism.

### Practical tuning recommendation

```python

# Local LMDB (memory-bound)
set_config_int("VersionStore.NumCPUThreads", 2)
set_config_int("VersionStore.NumIOThreads", 4)

# Real S3 (network-bound)
set_config_int("VersionStore.NumCPUThreads", 8)
set_config_int("VersionStore.NumIOThreads", 64)
```

The current defaults are over-provisioned for both backends on Windows.
A per-backend default — small pool for in-process storage, large IO pool
for object stores — would be a worthwhile follow-up upstream.

### What this rules out about the "batch slower than serial" symptom

With the allocation fixes plus *correctly sized* pools:

- **LMDB**: serial ≈ 4.1 s, batch ≈ 4.7 s, ratio 0.87× — within noise
  of parity. Earlier "batch is 30% slower than serial" was a thread-pool
  artifact, not a real characteristic of the decoder.
- **S3**: serial ≈ 33 s, batch ≈ 25 s — batch wins by 25-30%. The
  parallelism behaves as you'd expect once IO threads can saturate the
  network.

So further field-collection work or per-thread allocators are not where
the next win lies on this workload — it's already at parity for LMDB and
ahead for S3 once the pools are right-sized.

## Platform applicability

Algorithmic, not platform-specific — the same O(N²) memcpy pattern
exists on every OS. Severity differs:

- **Linux (glibc)**: `realloc` can sometimes extend large allocations
  in place via `mremap`, slightly softening the worst case.
- **Windows (HeapReAlloc)**: usually falls back to alloc+copy+free and
  has more heap-lock contention under multi-threaded load — the worst
  case, and what we measured.
- **macOS** (nano/magazine malloc): between the two.


cpp/arcticdb/codec/encode_v1.cpp
```diff
@@ -121,9 +121,9 @@ using EncodingPolicyV1 = EncodingPolicyType<EncodingVersion::V1, ColumnEncoderV1
     auto descriptor_data = in_mem_seg.descriptor().data_ptr();
     descriptor_data->uncompressed_bytes_ = uncompressed_size;
 
-    EncodedFieldCollection encoded_fields;
-    if (in_mem_seg.row_count() > 0) {
-        encoded_fields.reserve(encoded_buffer_size, in_mem_seg.num_columns());
+    const bool has_rows = in_mem_seg.row_count() > 0;
+    EncodedFieldCollection encoded_fields(has_rows ? encoded_buffer_size : 0, has_rows ? in_mem_seg.num_columns() : 0);
```
alexowens90: I don't follow why these are the correct values here?

cpp/arcticdb/entity/field_collection.hpp
```diff
@@ -120,12 +138,27 @@ class FieldCollection {
...
     std::string_view add(const FieldRef& field) { return add_field(field.type(), field.name()); }
 
+    [[nodiscard]] std::optional<std::size_t> find_field(std::string_view view) const {
+        auto it = std::find_if(begin(), end(), [&](const auto& field) { return field.name() == view; });
+
+        if (it == end())
+            return std::nullopt;
+
+        return std::distance(begin(), it);
+    }
```
alexowens90: Random move

cpp/arcticdb/util/buffer.hpp
```diff
@@ -174,7 +174,16 @@ struct Buffer : public BaseBuffer<Buffer, true> {
...
+            // Grow geometrically (at least 1.5x) rather than to the exact request. Callers
+            // like FieldCollection::add_field append one field at a time, so `ensure` is
+            // invoked with monotonically increasing `bytes`. Sizing to exactly `total_size`
+            // would trigger a fresh realloc + memcpy of body_bytes_ on every call, making N
+            // appends cost O(N^2) in copy work (~35k fields on a wide schema = 35k reallocs).
+            // With 1.5x growth, realloc happens O(log N) times and amortized append is O(1),
+            // matching the std::vector strategy. The max() with `total_size` guarantees we
+            // still satisfy the request if capacity_ is 0 or the jump exceeds 1.5x.
+            const size_t new_alloc = std::max(total_size, capacity_ + (capacity_ / 2));
+            resize(new_alloc);
```
alexowens90: This change should be reverted now

python/tests/unit/arcticdb/test_wide_df_perf.py
alexowens90, line 1: I don't think this should be checked in as a test. Can augment ASV benchmarks to protect against regressions in this area instead

alexowens90 (changes requested):

cpp/arcticdb/codec/encode_v1.cpp
G-D-Petrov: Added some clarification

python/tests/unit/arcticdb/test_wide_df_perf.py
G-D-Petrov: We already have ASV benchmarks for the wide_dfs.
The problem with the ASV is that it only runs against Linux, and this test is a way to protect against regressions on Windows/Mac.

cpp/arcticdb/codec/encoded_field_collection.hpp
```diff
@@ -50,13 +50,17 @@ class EncodedFieldCollection {
         data_(std::move(data)),
         offsets_(std::move(offsets)) {}
 
-    void reserve(size_t bytes, size_t num_fields) {
+    void reserve(size_t num_fields, size_t bytes) {
```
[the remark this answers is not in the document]
G-D-Petrov: Fixed

alexowens90 (approved):

ognyanstoimenov (approved):

PR #3037, opened 2026-04-16
Polars write support and change index_column to bool

IvoDD:
#### Reference Issues/PRs
Monday ref: 18362798253

#### What does this implement or fix?
Two main changes:

#### 1. `index_column` change from `str` to `bool`
Changes the `index_column` argument from `str` to `bool` meaning we now only accept the first column to be the timeseries index.

This is NOT a breaking change because it only affects api strictly labeled as experimental.

This is done for two reaons:
- Simplifies drastically denormalization logic. We don't have to worry about where to put the index column after schema changes.
- It is future proof. We can re-enable the custom index column position in the future without being a braking change. The `index_column` arg could just start accepting strings alongside bools. We would not be able to change our minds in the opisite direction.

Change involves:
- Modification of experimental protobuf. This is ok as arrow writes are experimental
- Removal of all column reordering logic
- Updates tests and docs

#### 2. Accept `polars.DataFrame`s to write like methods if arrow input is enabled

Make `ArrowTableNormalizer` work with `polars.DataFrame`s. This allows all write methods to work with polars input.

Currently this works by converting `polars.DataFrame` to `pyarrow.Table` and works the same as pyarrow tables going forward. In the future this will be better done via the `PyCapsule` polars interface, when sparrow fully supports it.

Change invloves:
- Normalizer accepting polars dataframes
- update with date_range truncation of polars dataframes via pyarrow (was benchmarked faster than polars because pyarrow can work with views)
- Parametrizing a subset of tests to work with polars (done in a separate commit for easier review)

#### Note to reviewers
The polars test parametrization is done in a separate commit for easier review.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [ ] Have you updated the relevant docstrings, documentation and copyright notice?
 - [ ] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)?
 - [ ] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)?
 - [ ] Are API changes highlighted in the PR description?
 - [ ] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes?
</details>

PR #3049, opened 2026-04-20
Propagate SortedValue to Polars set_sorted() on index column

G-D-Petrov:
#### Reference Issues/PRs
Monday ticket ref: 10659398340

When reading data from ArcticDB with `output_format="polars"`, the resulting Polars DataFrame had no sorted flag set on the index column, even though ArcticDB tracks sort order internally via `SortedValue` in the C++ `StreamDescriptor`. This means Polars couldn't take advantage of optimized operations (e.g., fast joins, binary search filtering) that rely on knowing a column is sorted.

### Changes

**C++ layer — propagate `SortedValue` through the read path:**

- `cpp/arcticdb/entity/read_result.hpp` — Added a `SortedValue sorted` field to both `NodeReadResult` and `ReadResult`, defaulting to `UNKNOWN`.
- `cpp/arcticdb/pipeline/pipeline_utils.hpp` — In `create_python_read_result`, extracts the sorted value from the stream descriptor and passes it into `NodeReadResult` and the top-level `ReadResult`.
- `cpp/arcticdb/python/adapt_read_dataframe.hpp` — Includes `sorted` (cast to `int`) in the tuple returned to Python from `adapt_read_df`.
- `cpp/arcticdb/python/python_utils.hpp` — Includes `sorted` in the tuples produced by `node_results_to_python_list` and `adapt_read_dfs`.

**Python layer — receive sorted value and apply Polars flag:**

- `python/arcticdb/version_store/read_result.py` — `ReadResult` and `NodeReadResult` now accept and store a `sorted` parameter (converted from int to `SortedValue` enum).
- `python/arcticdb/version_store/_store.py`:
  - Added `_get_index_col_from_norm()` helper to extract the index column name from normalization metadata (handles pandas DataFrames, Series, and experimental Arrow formats).
  - Added `_apply_polars_sorted_flag_to_index()` static method that calls `polars.col(...).set_sorted()` on the index column when the stored sort order is `ASCENDING` or `DESCENDING`.
  - Integrated the flag application into `_adapt_read_res()` so it runs automatically on every read.

**Tests:**

- `python/tests/unit/arcticdb/version_store/test_polars_set_sorted.py` — New test file with 6 tests covering:
  - Unnamed `DatetimeIndex` → `__index__` column gets `SORTED_ASC`
  - Named `DatetimeIndex` → named column gets `SORTED_ASC`
  - `RangeIndex` (not physically stored) → no sorted flag set
  - Pandas output path still works without issues
  - Stage + finalize workflow preserves sorted flag
  - Value columns do not get the sorted flag (only the index does)
  
### Before vs After (stock ArcticDB 6.13.0 vs branch dev build)

The benchmark script reads 1M rows as Polars and measures various operations. On stock ArcticDB, the index has no sorted flag (`SORTED_ASC: False`). With the branch's dev build, the sorted flag is automatically applied (`SORTED_ASC: True`).

**Comparing stock (6.13.0) vs branch (dev) on key index operations:**

| Benchmark | Stock 6.13.0 (ms) | Branch dev (ms) | Speedup |
|---|--:|--:|--:|
| sort_by_index | 18.90 | 3.06 | **6.18x** |
| unique_index | 3.71 | 0.69 | **5.38x** |
| upsample | 25.20 | 8.32 | **3.03x** |
| filter_range | 1.93 | 1.89 | 1.02x |
| filter_after | 1.30 | 1.27 | 1.02x |
| group_by_dynamic | 9.40 | 8.83 | 1.06x |
| rolling_mean | 25.59 | 24.87 | 1.03x |

The branch delivers significant speedups on `sort_by_index` (6.2x), `unique_index` (5.4x), and `upsample` (3x). These are operations where Polars can skip work entirely when it knows the column is sorted.

### Branch dev build: detailed benchmark (Polars 1.35.2, 1M rows, 10 iters)

Note: In the branch build, "Current (no flag)" already has `SORTED_ASC: True` on the index (the branch's feature), so applying `set_sorted()` again is a no-op. This confirms the feature is working automatically.

#### Part 2: Value Columns — Future Work

These measure the *potential* benefit of also setting `set_sorted()` on value columns, which would require populating `FieldStats.sorted_` in C++. **Not implemented in this branch.**

| Benchmark | Current/branch (ms) | + index + values (ms) | Speedup | Notes |
|---|--:|--:|--:|---|
| sort_by_price | 22.71 | 3.42 | **6.63x** | Big win |
| filter_price_range | 0.91 | 0.79 | 1.14x | |
| unique_price | 4.07 | 0.70 | **5.82x** | Big win |
| search_sorted_price | 0.11 | 0.10 | 1.10x | |
| join_asof_on_price | 5.81 | 6.42 | 0.90x | |
| sort_by_volume | 20.77 | 4.19 | **4.96x** | Big win |
| unique_volume | 3.31 | 0.64 | **5.14x** | Big win |

### Overhead

The `set_sorted()` call adds negligible overhead (~0.3 ms) relative to the read cost (~399 ms debug build):

| Operation | Time (ms) |
|---|--:|
| `set_sorted` (index only) | 0.28 |
| `set_sorted` (index + 2 cols) | 0.65 |

### DIFF
--- a/cpp/arcticdb/entity/read_result.hpp
+++ b/cpp/arcticdb/entity/read_result.hpp
@@ -27,14 +27,16 @@ using OutputFrame = std::variant<pipelines::PandasOutputFrame, ArrowOutputFrame>
 struct ARCTICDB_VISIBILITY_HIDDEN NodeReadResult {
     NodeReadResult(
             const StreamId& symbol, OutputFrame&& frame_data,
-            arcticdb::proto::descriptors::NormalizationMetadata&& norm_meta
+            arcticdb::proto::descriptors::NormalizationMetadata&& norm_meta, SortedValue sorted = SortedValue::UNKNOWN
     ) :
         symbol_(symbol),
         frame_data_(std::move(frame_data)),
-        norm_meta_(std::move(norm_meta)) {};
+        norm_meta_(std::move(norm_meta)),
+        sorted_(sorted) {};
     StreamId symbol_;
     OutputFrame frame_data_;
     arcticdb::proto::descriptors::NormalizationMetadata norm_meta_;
+    SortedValue sorted_ = SortedValue::UNKNOWN;
 
     ARCTICDB_MOVE_ONLY_DEFAULT(NodeReadResult)
 };
@@ -47,7 +49,7 @@ struct ARCTICDB_VISIBILITY_HIDDEN ReadResult {
                     arcticdb::proto::descriptors::UserDefinedMetadata,
                     std::vector<arcticdb::proto::descriptors::UserDefinedMetadata>>& user_meta,
             const arcticdb::proto::descriptors::UserDefinedMetadata& multi_key_meta,
-            std::vector<NodeReadResult>&& node_results = {}
+            std::vector<NodeReadResult>&& node_results = {}, SortedValue sorted = SortedValue::UNKNOWN
     ) :
         item(versioned_item),
         frame_data(std::move(frame_data)),
@@ -55,7 +57,8 @@ struct ARCTICDB_VISIBILITY_HIDDEN ReadResult {
         norm_meta(norm_meta),
         user_meta(user_meta),
         multi_key_meta(multi_key_meta),
-        node_results(std::move(node_results)) {}
+        node_results(std::move(node_results)),
+        sorted(sorted) {}
     std::variant<VersionedItem, std::vector<VersionedItem>> item;
     OutputFrame frame_data;
     OutputFormat output_format;
@@ -66,6 +69,7 @@ struct ARCTICDB_VISIBILITY_HIDDEN ReadResult {
             user_meta;
     arcticdb::proto::descriptors::UserDefinedMetadata multi_key_meta;
     std::vector<NodeReadResult> node_results;
+    SortedValue sorted = SortedValue::UNKNOWN;
 
     ARCTICDB_MOVE_ONLY_DEFAULT(ReadResult)
 };
--- a/cpp/arcticdb/entity/read_result.hpp
+++ b/cpp/arcticdb/entity/read_result.hpp
@@ -27,7 +27,7 @@ using OutputFrame = std::variant<pipelines::PandasOutputFrame, ArrowOutputFrame>
 struct ARCTICDB_VISIBILITY_HIDDEN NodeReadResult {
     NodeReadResult(
             const StreamId& symbol, OutputFrame&& frame_data,
-            arcticdb::proto::descriptors::NormalizationMetadata&& norm_meta, SortedValue sorted = SortedValue::UNKNOWN
+            arcticdb::proto::descriptors::NormalizationMetadata&& norm_meta, SortedValue sorted
     ) :
         symbol_(symbol),
         frame_data_(std::move(frame_data)),
@@ -36,7 +36,7 @@ struct ARCTICDB_VISIBILITY_HIDDEN NodeReadResult {
     StreamId symbol_;
     OutputFrame frame_data_;
     arcticdb::proto::descriptors::NormalizationMetadata norm_meta_;
-    SortedValue sorted_ = SortedValue::UNKNOWN;
+    SortedValue sorted_;
 
     ARCTICDB_MOVE_ONLY_DEFAULT(NodeReadResult)
 };
@@ -49,7 +49,7 @@ struct ARCTICDB_VISIBILITY_HIDDEN ReadResult {
                     arcticdb::proto::descriptors::UserDefinedMetadata,
                     std::vector<arcticdb::proto::descriptors::UserDefinedMetadata>>& user_meta,
             const arcticdb::proto::descriptors::UserDefinedMetadata& multi_key_meta,
-            std::vector<NodeReadResult>&& node_results = {}, SortedValue sorted = SortedValue::UNKNOWN
+            std::vector<NodeReadResult>&& node_results, SortedValue sorted
     ) :
         item(versioned_item),
         frame_data(std::move(frame_data)),
@@ -69,7 +69,7 @@ struct ARCTICDB_VISIBILITY_HIDDEN ReadResult {
             user_meta;
     arcticdb::proto::descriptors::UserDefinedMetadata multi_key_meta;
     std::vector<NodeReadResult> node_results;
-    SortedValue sorted = SortedValue::UNKNOWN;
+    SortedValue sorted;
 
     ARCTICDB_MOVE_ONLY_DEFAULT(ReadResult)
 };
--- a/cpp/arcticdb/pipeline/input_frame.cpp
+++ b/cpp/arcticdb/pipeline/input_frame.cpp
@@ -16,10 +16,39 @@
 
 namespace arcticdb::pipelines {
 
+namespace {
+SortedValue compute_index_sortedness(const Column& index_column) {
+    using TimeseriesTDT = stream::TimeseriesIndex::TypeDescTag;
+    auto column_data = index_column.data();
+    auto it = column_data.cbegin<TimeseriesTDT>();
+    const auto end = column_data.cend<TimeseriesTDT>();
+    if (it == end) {
+        return SortedValue::ASCENDING;
+    }
+    bool ascending = true;
+    bool descending = true;
+    auto prev = *it;
+    for (++it; it != end; ++it) {
+        const auto curr = *it;
+        if (curr < prev) {
+            ascending = false;
+        } else if (curr > prev) {
+            descending = false;
+        }
+        if (!ascending && !descending) {
+            return SortedValue::UNSORTED;
+        }
+        prev = curr;
+    }
+    return ascending ? SortedValue::ASCENDING : SortedValue::DESCENDING;
+}
+} // namespace
+
 InputFrame::InputFrame() : index(stream::empty_index()) {}
 
 void InputFrame::set_from_columns(
-        std::vector<Column>&& cols, StreamDescriptor&& desc, std::vector<sparrow::record_batch>&& arrow_buffer_owners
+        std::vector<Column>&& cols, StreamDescriptor&& desc, std::vector<sparrow::record_batch>&& arrow_buffer_owners,
+        SortednessScan sortedness_scan
 ) {
     util::check(norm_meta.has_experimental_arrow(), "Unexpected non-Arrow norm metadata provided with Arrow data");
     desc_ = std::move(desc);
@@ -34,9 +63,11 @@ void InputFrame::set_from_columns(
         );
         desc_.set_index({IndexDescriptorImpl::Type::TIMESTAMP, 1});
         index = stream::TimeseriesIndex{std::string(desc_.field(0).name())};
-        // We do not verify monotonicity of Arrow index columns, so the sort order is UNKNOWN. This avoids the
-        // O(n) check on write; the polars read path will not claim the column is sorted as a result.
-        seg.descriptor().set_sorted(SortedValue::UNKNOWN);
+        // Arrow input does not record sortedness up front (unlike pandas), so we have to compute it when required.
+        seg.descriptor().set_sorted(
+                sortedness_scan == SortednessScan::SCAN_IF_UNKNOWN ? compute_index_sortedness(seg.column(0))
+                                                                   : SortedValue::UNKNOWN
+        );
     } else {
         desc_.set_index({IndexDescriptorImpl::Type::ROWCOUNT, 0});
         index = stream::RowCountIndex{};
--- a/cpp/arcticdb/pipeline/input_frame.cpp
+++ b/cpp/arcticdb/pipeline/input_frame.cpp
@@ -64,8 +64,8 @@ void InputFrame::set_from_columns(
         desc_.set_index({IndexDescriptorImpl::Type::TIMESTAMP, 1});
         index = stream::TimeseriesIndex{std::string(desc_.field(0).name())};
         // Arrow input does not record sortedness up front (unlike pandas), so we have to compute it when required.
-        seg.descriptor().set_sorted(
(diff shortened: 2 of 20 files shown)


python/tests/unit/arcticdb/version_store/test_polars_set_sorted.py
```diff
@@ -0,0 +1,141 @@
...
+    df = pd.DataFrame({"val": np.arange(10)})
+    lib.write("sym", df)
+    result = lib.read("sym", output_format="polars").data
+
+    assert isinstance(result, pl.DataFrame)
+    for col in result.columns:
+        assert result[col].flags["SORTED_ASC"] is False, f"Column {col} should not have SORTED_ASC"
+
+
+def test_sorted_flag_not_set_for_pandas_output(lmdb_library):
```
IvoDD: Do we need this test? We have a lot of other pandas coverage of pandas reads.

python/tests/unit/arcticdb/version_store/test_polars_set_sorted.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def test_unknown_sort_order_does_not_set_sorted_flag():
+    """When sort_order is UNKNOWN no sorted flag should be applied even with a physical index."""
+    data = pl.DataFrame({"__index__": pd.date_range("2024-01-01", periods=5, freq="h")})
+    read_result = _make_mock_read_result(SortedValue.UNKNOWN)
+
+    result = NativeVersionStore._apply_polars_sorted_flag_to_index(data, read_result)
+
+    assert result["__index__"].flags["SORTED_ASC"] is False
```
IvoDD, line 194: A meta point worth discussing is whether we should treat `UNKNOWN` as `SORTED_ASC`.

For all other purposes we treat `UNKOWN` (written by old arctcdb versions before sortedness was introduced) as sorted. (e.g. for update we use [this code](https://github.com/man-group/ArcticDB/blob/a6b7b4e50ea4fafc52c53bcabc2b1a0b0f7e8471/cpp/arcticdb/version/version_core.cpp#L417-L418)).

I'm personally in favor of treating unknown constistently and set sorted even if unkown.

We should instead add a test for `UNSORTED`

python/arcticdb/version_store/read_result.py
```diff
@@ -6,28 +6,30 @@
...
         self.frame_data = (
             FrameData(*frame_data.extract_numpy_arrays()) if isinstance(frame_data, PandasOutputFrame) else frame_data
         )
         self.norm = norm
+        self.sort_order = SortedValue(sort_order) if isinstance(sort_order, int) else sort_order
 
 
 class ReadResult:
-    def __init__(self, version, frame_data, norm, udm, mmeta, node_read_results):
+    def __init__(self, version, frame_data, norm, udm, mmeta, node_read_results, sort_order=SortedValue.UNKNOWN):
```
IvoDD: I'm slightly worried if we're using a default value we might have missed some place where this is used incorrectly.

python/arcticdb/version_store/library.py
```diff
@@ -2074,6 +2074,9 @@ def read(
             Output format for the returned dataframe.
             If `None`, uses the output format from the `Library` instance.
             See `OutputFormat` documentation for details on available formats.
+            When using ``POLARS`` output format, the index column (if physically stored) will automatically have
```
IvoDD: We should probably move these docs to `OutputFormat` docs. The same description applies to methods like `head` and `tail` all of which reference the `OutputFormat` docs.

python/arcticdb/version_store/_store.py
```diff
@@ -351,6 +341,23 @@ def _assume_false(name, kwargs):
         return True
 
 
+def _get_index_col_from_norm(norm):
```
IvoDD: This will get drastically simpler after #3037 as the index column will always be first.

python/arcticdb/version_store/_store.py
```diff
@@ -2908,8 +2780,23 @@ def _adapt_frame_data(self, frame_data, norm, output_format):
 
         return data
 
+    @staticmethod
+    def _apply_polars_sorted_flag_to_index(data, read_result):
+        if pl is None or not isinstance(data, pl.DataFrame):
```
IvoDD: Nit: I think it would be slightly cleaner if we call this only for polars DataFrames [here](https://github.com/man-group/ArcticDB/blob/a6b7b4e50ea4fafc52c53bcabc2b1a0b0f7e8471/python/arcticdb/version_store/_store.py#L2903), then this check would not be needed.

python/arcticdb/version_store/_store.py
```diff
@@ -351,6 +341,23 @@ def _assume_false(name, kwargs):
...
+def _get_index_col_from_norm(norm):
+    """Get the name of the index column from normalization metadata, or None if no physically stored index."""
+    input_type = norm.WhichOneof("input_type")
+    if input_type == "experimental_arrow":
+        if norm.experimental_arrow.has_index:
+            return norm.experimental_arrow.index_column_name
+    elif input_type in ("df", "series"):
+        common = norm.df.common if input_type == "df" else norm.series.common
+        index_type = common.WhichOneof("index_type")
+        # multi_index intentionally excluded: multi-level sorting semantics are more complex
```
IvoDD: We probably still want to include multi_index as long as first index is sorted.

cpp/arcticdb/python/adapt_read_dataframe.hpp
```diff
@@ -36,7 +36,13 @@ inline py::tuple adapt_read_df(ReadResult&& ret, std::any* const handler_data) {
...
     auto node_results = python_util::node_results_to_python_list(std::move(ret.node_results));
     return py::make_tuple(
-            ret.item, std::move(ret.frame_data), pynorm, pyuser_meta, multi_key_meta, std::move(node_results)
+            ret.item,
+            std::move(ret.frame_data),
+            pynorm,
+            pyuser_meta,
+            multi_key_meta,
+            std::move(node_results),
+            static_cast<int>(ret.sorted)
```
IvoDD: Why cast to int? Can't we expose `SortedValue` in python bindings?

cpp/arcticdb/entity/read_result.hpp
```diff
@@ -27,14 +27,16 @@ using OutputFrame = std::variant<pipelines::PandasOutputFrame, ArrowOutputFrame>
 struct ARCTICDB_VISIBILITY_HIDDEN NodeReadResult {
     NodeReadResult(
             const StreamId& symbol, OutputFrame&& frame_data,
-            arcticdb::proto::descriptors::NormalizationMetadata&& norm_meta
+            arcticdb::proto::descriptors::NormalizationMetadata&& norm_meta, SortedValue sorted = SortedValue::UNKNOWN
```
IvoDD: Nit: Again slightly worried about this having a default value, we might have missed some constructions which should specify a sorted value.

This only matters if we decide to tread `UNKNOWN` as sorted. Otherwise it's probably fine

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -776,7 +818,14 @@ VersionedItem PythonVersionStore::append(
...
     return append_internal(
             stream_id,
-            convert::py_ndf_to_frame(stream_id, item, norm, user_meta, cfg().write_options().empty_types()),
+            convert::py_ndf_to_frame(
+                    stream_id,
+                    item,
+                    norm,
+                    user_meta,
+                    cfg().write_options().empty_types(),
+                    sortedness_scan_for(validate_index)
```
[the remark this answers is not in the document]
IvoDD: That's incorrect. The existing data check is:
```
    bool existing_data_is_sorted = index_segment_reader.sorted() == SortedValue::ASCENDING ||
                                   index_segment_reader.sorted() == SortedValue::UNKNOWN;
```
The strict sortedness with `validate_index` is only requred on the new segment being added. Otherwise append/update would not work with old UNKOWN data.

IvoDD:
The last commit added later adds an explicit check on the sortedness of the input arrow index when `validate_index=True`

`validate_index=True` for:
- All V2 APIs
- `update` always sets it to `True`

The walk is a single threaded pass over the index column but as benchmarks show below it does not have significant impact on write speed even if we are writing many segments in parallel. For 10M row column it takes just `10ms`.

Cost of the walk relative to the write (release build, single timestamp index):

In-memory (zero-copy write, so this is the worst case for relative overhead):

| rows | batches | no-verify | verify   | overhead |
|------|---------|-----------|----------|----------|
| 1M   | 1       | 3.36 ms   | 4.49 ms  | +34%     |
| 1M   | 100     | 4.12 ms   | 5.30 ms  | +29%     |
| 10M  | 1       | 5.53 ms   | 15.87 ms | +187%    |
| 10M  | 100     | 6.07 ms   | 16.16 ms | +166%    |

Real pure-S3 (network-bound, parallel segment encode/write):

| library          | rows | no-verify  | verify     | overhead |
|------------------|------|------------|------------|----------|
| default segment  | 1M   | 81.7 ms    | 82.6 ms    | +1.1%    |
| default segment  | 10M  | 214.2 ms   | 226.1 ms   | +5.6%    |
| 5000-row segment | 1M   | 329.9 ms   | 346.2 ms   | +1.8%    |
| 5000-row segment | 10M  | 1699.9 ms  | 1726.8 ms  | +1.6%    |

The walk is only paid when validate_index requests it; on real storage it is a
low single-digit percentage of the write, and more segment parallelism dilutes
it further rather than amplifying it.

python/arcticdb/version_store/_store.py
```diff
@@ -352,6 +352,20 @@ def _assume_false(name, kwargs):
...
+    input_type = norm.WhichOneof("input_type")
+    if input_type == "experimental_arrow":
+        return norm.experimental_arrow.has_index
+    if input_type in ("df", "series"):
+        common = norm.df.common if input_type == "df" else norm.series.common
+        index_type = common.WhichOneof("index_type")
+        if index_type == "index":
+            return common.index.is_physically_stored
+        if index_type == "multi_index":
+            return True
```
alexowens90, line 365: I know it's not used here as this is only called with Arrow data, but this is will not be correct for multiindex data where the top level is not a timeseries

python/arcticdb/options.py
```diff
@@ -185,6 +185,10 @@ class OutputFormat(str, Enum):
         Dataframes are returned as `polars.DataFrame` objects using Apache Arrow's columnar memory format.
         Provides better performance than `PANDAS`, especially for dataframes containing many string columns.
         String format can be customized via `ArrowOutputStringFormat`.
+        If ArcticDB has verified the index column is monotonic (ascending or descending), its Polars sorted
+        flag is set automatically, enabling Polars sort-aware optimizations such as fast joins and group-bys.
+        Symbols whose sort order was never verified (Arrow writes, or symbols written by ArcticDB versions
```
alexowens90: The "Arrow writes" part of this comment is no longer accurate with the additions Ivo's made

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -14,9 +14,39 @@
...
+    return ascending ? SortedValue::ASCENDING : SortedValue::DESCENDING;
+}
+} // namespace
+
 InputFrame::InputFrame() : index(stream::empty_index()) {}
 
-void InputFrame::set_segment(SegmentInMemory&& seg, std::vector<sparrow::record_batch>&& arrow_buffer_owners) {
+void InputFrame::set_segment(
+        SegmentInMemory&& seg, std::vector<sparrow::record_batch>&& arrow_buffer_owners, SortednessScan sortedness_scan
+) {
```
alexowens90, line 52: This is going to have some merge conflicts with @ognyanstoimenov 's refactor of this class

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -14,9 +14,39 @@
 
 namespace arcticdb::pipelines {
 
+namespace {
+SortedValue compute_index_sortedness(const Column& index_column) {
```
alexowens90, line 20: We could parallelise this check by pushing it down to the `WriteToSegmentTask`, although this will add a lot of complexity. I'd suggest we don't do this now, and add a low-priority ticket to do it in future

IvoDD: Added a ticket, monday ref: 12335257870

python/arcticdb/version_store/_store.py
IvoDD: It is actually going to get used for pandas norm metadata. E.g. if we:
```
lib.write("sym", pandas_df_with_multiindex)
lib.read("sym", output_format="polars").data
```

I think the current code is correct because `_has_physically_stored_index` will return `True` for multiindex, because we indeed store the multiindex columns as columns on disk.

The `_apply_polars_sorted_flag_to_index` would also do the correct thing and it will look at the `SortedValue`. For non timeseries top level index it will be `UNKOWN` (see [here](https://github.com/man-group/ArcticDB/blob/61dc7187737bbeb700e05a48b36566a4e6a1b05b/python/arcticdb/version_store/_normalization.py#L1346-L1357))

alexowens90: Ah I see, I thought we were just trying to extract [this](https://github.com/man-group/ArcticDB/blob/61dc7187737bbeb700e05a48b36566a4e6a1b05b/cpp/proto/arcticc/pb2/descriptors.proto#L129) protobuf field.

Just to check my own understanding:
 `_apply_polars_sorted_flag_to_index` will do the right thing with a multiindex dataframe where the top level is not a timeseries because it will hit the return on line 2947?

alexowens90 (approved):

python/arcticdb/version_store/_store.py
IvoDD: yes exactly, added a test to prove and assure we don't regress

PR #3050, opened 2026-04-20
Fix intermittent hang in test_overwrite_append_data

G-D-Petrov:
#### Reference Issues/PRs
Monday ticket ref: 11803660445

`test_overwrite_append_data` intermittently hangs in CI, causing the pytest-xdist worker to crash. The faulthandler trace shows the main thread blocked inside various `LibraryTool` C++ calls. The hang point changes between runs — it has been observed in `read_timeseries_descriptor`, `read_descriptor`, and `segment_in_memory_to_read_result`.

This reproduces across multiple branches and Python versions (3.9, 3.10, 3.11) but only in CI environments.

### Root Cause

`LibraryTool` methods used the async pipeline machinery — submitting tasks to the IO and CPU thread pools via `submit_io_task`/`submit_cpu_task` and then blocking the calling thread on `.get()`. In CI containers, the thread pools intermittently fail to process submitted tasks, causing `.get()` to block forever.
The suspected root cause is that the `LibraryTool` is not releasing the GIL properly.
This will be addressed by Monday ticket 10841098589

The affected code paths were:
- `read_metadata` — `read_and_continue` → IO pool → `.get()`
- `read_descriptor` — `read_and_continue` → IO pool → `.get()`
- `read_timeseries_descriptor` — `read_and_continue` → IO pool → `.get()`
- `segment_in_memory_to_read_result` → `read_result_from_single_frame` → `reduce_and_fix_columns` → CPU pool → `.get()`

### Fix

Make all `LibraryTool` operations fully synchronous — no thread pool submissions:

1. **Read methods** (`read_metadata`, `read_descriptor`, `read_timeseries_descriptor`): Use `read_compressed_sync` + inline decode tasks instead of async `read_and_continue` + `.get()`.

2. **`segment_in_memory_to_read_result`**: Use new `read_result_from_single_frame_sync` which calls `reduce_and_fix_columns_sync` — runs `ReduceColumnTask` directly on the calling thread instead of submitting to the CPU pool.

Existing `PythonVersionStore` callers of `read_result_from_single_frame` are unchanged and continue using the async path. Common pipeline setup logic is extracted into `detail::prepare_single_frame_context` to avoid duplication between the async and sync variants.

The `LibraryTool` only processes single segments and doesn't benefit from parallelism, so there's no performance downside.

### DIFF
--- a/cpp/arcticdb/pipeline/pipeline_utils.hpp
+++ b/cpp/arcticdb/pipeline/pipeline_utils.hpp
@@ -141,4 +141,25 @@ inline ReadResult read_result_from_single_frame(
     return create_python_read_result(VersionedItem{key}, output_format, std::move(frame_and_desc));
 }
 
+inline ReadResult read_result_from_single_frame_sync(
+        FrameAndDescriptor& frame_and_desc, const AtomKey& key, std::any& handler_data, OutputFormat output_format
+) {
+    auto pipeline_context = std::make_shared<PipelineContext>(frame_and_desc.frame_.descriptor());
+    SliceAndKey sk{FrameSlice{frame_and_desc.frame_}, key};
+    pipeline_context->slice_and_keys_.emplace_back(std::move(sk));
+    util::BitSet bitset(1);
+    bitset.flip();
+    pipeline_context->fetch_index_ = std::move(bitset);
+    pipeline_context->ensure_vectors();
+
+    generate_filtered_field_descriptors(pipeline_context, {});
+    pipeline_context->begin()->set_string_pool(frame_and_desc.frame_.string_pool_ptr());
+    auto descriptor = std::make_shared<StreamDescriptor>(frame_and_desc.frame_.descriptor());
+    pipeline_context->begin()->set_descriptor(std::move(descriptor));
+    std::shared_ptr<std::any> handler_data_ptr(std::shared_ptr<std::any>{}, &handler_data);
+    reduce_and_fix_columns_sync(pipeline_context, frame_and_desc.frame_, ReadOptions{}, handler_data_ptr);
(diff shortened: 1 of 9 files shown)


cpp/arcticdb/pipeline/pipeline_utils.hpp
```diff
@@ -136,9 +137,26 @@ inline ReadResult read_result_from_single_frame(
     auto descriptor = std::make_shared<StreamDescriptor>(frame_and_desc.frame_.descriptor());
     pipeline_context->begin()->set_descriptor(std::move(descriptor));
     std::shared_ptr<std::any> handler_data_ptr(std::shared_ptr<std::any>{}, &handler_data);
+    return {std::move(pipeline_context), std::move(handler_data_ptr)};
+}
+} // namespace detail
+
+inline ReadResult read_result_from_single_frame(
```
IvoDD, line 144: Is there any benefit to the async version `read_result_from_single_frame` over the sync `read_result_from_single_frame_sync`?

I can imagine the async version being useful if it returned a `folly::Future` but now it just calls `get` imidately on the returned future.

IvoDD (approved):

vasil-pashov (approved):

cpp/arcticdb/pipeline/pipeline_utils.hpp
G-D-Petrov: The only reason the async version exists is that reduce_and_fix_columns submits ReduceColumnTask to the CPU thread pool and returns a Future. Even though read_result_from_single_frame immediately calls .get() on it, the column reduction work still runs in parallel across CPU  threads via folly::window batching.

PR #3053, opened 2026-04-21
Fixes for CI instability

poodlewars:
The first two commits are by me. 

2edf29 The get column stats info benchmark is flaky because it runs in less than 1ms. We don't really care about the performance of these management APIs, and I think we might as well just drop the benchmark.

c37d93 Georgi recently excluded Mac from this test for the same reason - we've recently seen this same failure on Windows too.

The other three are by Claude. It chose what was worth fixing by looking through the last 30 days of master CI failures. There were Monday tickets already open for the first two.

9fbd1c disables the version map cache to guarantee failed reads, which makes sense in the test.
913c8d disables another `SparseTestStore` test on Windows.
ddb545 adds a retry for a test that seems to fail occasionally on the Mac CI.

Finally while working on this, the CI failed due to download errors from one of the VCPKG mirrors. So 9f9be6 adds VCPKG asset caching - we save sources we download using vcpkg to an S3 bucket that we manage. That should save us from most instability with vcpkg. I haven't considered cache eviction here as we have good monitoring of S3 costs and our vcpkg deps change slowly.


.github/workflows/build_steps.yml
```diff
@@ -415,6 +427,15 @@ jobs:
           CXX_COMPILER_LAUNCHER: sccache
           SCCACHE_DIR: ${{env.SCCACHE_DIR || ''}}
 
+      - name: Upload vcpkg assets to S3 cache
+        if: inputs.job_type == 'build-python-wheels' && matrix.os == 'linux'
+        run: |
+          # setup.py copies vcpkg downloads to /host/tmp/vcpkg_downloads inside the
+          # CIBW container, which is /tmp/vcpkg_downloads on the host.
+          bash build_tooling/vcpkg_upload_assets.sh /tmp/vcpkg_downloads
+          sudo rm -rf /tmp/vcpkg_downloads
```
G-D-Petrov, line 436: nit: we probably don't need to explicitly remove them, as they will be dropped when the run ends.
And this will be wasting time on the run itself.

G-D-Petrov (approved):

.github/workflows/build_steps.yml
poodlewars: The cleanup was here before and was very deliberate, cos we kept running out of disk during the test execution.

build_tooling/vcpkg_caching.sh
```diff
@@ -24,3 +24,8 @@ if [[ -n "$VCPKG_MAN_NUGET_TOKEN" ]] ; then
 fi
 
 popd
+
+# S3 asset cache: caches source archives so builds survive upstream outages
+if [[ -n "$AWS_ACCESS_KEY_ID" ]] ; then
+    export X_VCPKG_ASSET_SOURCES="x-script,aws s3 cp s3://arcticdb-ci-vcpkg-assets/{sha512} {dst} --region eu-west-1 --only-show-errors"
```
IvoDD, line 30: Have you tested building from the cache? I.e.:
- Running one build to populate the cache
- Running a second build to verify build from cache works (and is indeed pulling from the cache)

Out of curiosity how much time is spent on downloads and how does `download from mirror` compare against `download from s3`?

How much data are we talking about? Downloading from aws s3 to outside aws can be expensive. Especially when done multiple times per day on multiple builds.

Do we know where the github runners are located? Are they close to `eu-west-1`, should we consider using a different aws region

.github/workflows/analysis_workflow.yml
```diff
@@ -221,6 +221,15 @@ jobs:
       - name: Build project for ASV
         run: |
           python -m pip install -ve .[Testing]
+        env:
+          ARCTICDB_KEEP_VCPKG_SOURCES: "1"
+
+      - name: Upload vcpkg assets to S3 cache
```
IvoDD, line 227: Analisys workflow now uploads the vcpkg assets as well, but does it use them during build?

python/tests/hypothesis/arcticdb/test_merge_update.py
```diff
@@ -140,16 +149,24 @@ def test_rowrange_merge_update(lmdb_version_store_v1, merge_args):
...
+        try:
+            lib.merge_experimental(symbol, source, strategy=strategy, on=on)
+        except UserInputException as e:
+            if "Multiple source rows match the same target row" not in str(e):
+                raise
+            with pytest.raises(ValueError, match="Multiple source rows match the same target row"):
+                merge(target[0], source, strategy=strategy, on=on)
+            return
+        result = lib.read(symbol).data
+    except DuplicateKeyException:
```
IvoDD: What do you think about just skipping these tests on mac and windows. We have probably good enough coverage with the non-hypothesis tests for windows and mac

cpp/arcticdb/util/test/test_storage_lock.cpp
```diff
@@ -352,6 +352,7 @@ class StorageLockWithAndWithoutRetry : public ::testing::TestWithParam<bool> {
 };
 
 TEST(StorageLock, ConcurrentWritesWithRetrying) {
+    SKIP_MAC("StorageLock is not supported");
```
IvoDD: Nit: I think skip reason is a bit misleading. It's not that "StorageLock is not supported". We do actually use it on mac for symbol list compaction. It's just that it's way less reliable because of the lower granularity clock.

I guess this particular issue we can fix more easily because we write the timestamp inside the lock ref key. We can easily add some random number and it will be way more reliable on macos/windows.

build_tooling/vcpkg_caching.sh
poodlewars: Yes, I've tested building from the cache locally. I didn't gather timings.

We are not actually talking about much data here, because we usually hit the nuget binary cache. We only use these asset sources if there is a binary cache miss.

The Github runners are in various US regions, but the rest of our CI infra (eg the sccache bucket) is in `eu-west-1` so I've kept that here. Might be sensible for us to migrate away to US infra.

I think it's sensible to try this. We have good alarms over AWS costs. Our S3 costs are actually very low (despite the ASV buckets, and sccache bucket), most of our AWS costs are from renting EC2 runners.

.github/workflows/analysis_workflow.yml
poodlewars: It's set up in this step,

```
      - name: Extra envs
        shell: bash -l {0}
        run: |
          . build_tooling/prep_cpp_build.sh
          . build_tooling/vcpkg_caching.sh
          echo -e "VCPKG_BINARY_SOURCES=$VCPKG_BINARY_SOURCES
          VCPKG_ROOT=$PLATFORM_VCPKG_ROOT
          ${X_VCPKG_ASSET_SOURCES:+X_VCPKG_ASSET_SOURCES=$X_VCPKG_ASSET_SOURCES}" | tee -a $GITHUB_ENV
          cmake -P cpp/CMake/CpuCount.cmake | sed 's/^-- //' | tee -a $GITHUB_ENV
        env:
          CMAKE_BUILD_PARALLEL_LEVEL: ${{vars.CMAKE_BUILD_PARALLEL_LEVEL}}
```

python/tests/hypothesis/arcticdb/test_merge_update.py
poodlewars: Yeah, that's much tidier.

IvoDD (approved):

PR #3056, opened 2026-04-23
Bug: add_to_snapshot does not work with negative version

markovskipetar:
# Bug Fix: `add_to_snapshot` with Negative Version Numbers

## Problem

`add_to_snapshot("snap", ["s1"], [-1])` silently did nothing instead of adding
the requested version to the snapshot.

## Root Cause

Negative version IDs (e.g. `-1`) are stored in `SpecificVersionQuery::version_id_`
as `SignedVersionId` (`int64_t`). In `get_multiple_sym_versions_from_query`, they
were pushed into `VersionVectorType` (`std::vector<VersionId>` = `std::vector<uint64_t>`).
This implicit conversion corrupts the value:

```
-1  (int64_t)  →  UINT64_MAX  (uint64_t)
-7  (int64_t)  →  UINT64_MAX - 6  (uint64_t)
```

The corrupted value was then passed to `batch_get_specific_versions`, which called
`find_index_key_for_version_id(UINT64_MAX, ...)`. No such version exists, so the symbol
was silently dropped and never added to the snapshot.

## Fix

**`get_multiple_sym_versions_from_query` (`local_versioned_engine.cpp`)** — Added an
explicit validation guard that rejects negative version IDs with a `UserInputException`
before any conversion occurs. Negative versions are not supported in `add_to_snapshot`.

**Duplicate symbol detection** — Added a check in `get_stream_index_map` that raises a
`UserInputException` if the same symbol appears more than once in a single
`add_to_snapshot` call (regardless of whether the versions differ).

**`resolve_version_id` refactor (`version_utils.hpp`)** — The repeated negative-index
resolution pattern (branching on `signed_version_id >= 0`, then calling
`get_version_id_negative_index`) that existed in three separate places
(`version_functions.hpp`, `version_map_batch_methods.cpp`, `version_map.hpp`) was
extracted into a single `resolve_version_id(SignedVersionId, const VersionMapEntry&)`
helper. All three call sites now use it.

**`batch_get_specific_versions` (`version_map_batch_methods.hpp/.cpp`)** — Updated the
doc comment to make explicit that only non-negative versions are supported; the caller
is responsible for resolving or rejecting negatives upstream.

## Tests

- `test_add_to_snapshot_with_negative_numbers` — updated: previously `xfail`, now
  asserts that a `UserInputException` matching `"Negative version"` is raised.
- `test_add_to_snapshot_negative_version_raises` — new test covering a negative version
  mixed with a valid version in the same call.
- `test_add_to_snapshot_duplicate_symbol_different_versions_raises` — new test: same
  symbol, different versions → raises.
- `test_add_to_snapshot_duplicate_symbol_same_version_raises` — new test: same symbol,
  same version twice → raises.
- `test_add_to_snapshot_mismatched_symbols_and_versions_raises` — new test: mismatched
  lengths of symbols and versions lists → raises.


cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -277,6 +277,22 @@ inline std::shared_ptr<std::unordered_map<StreamId, AtomKey>> batch_get_specific
 
 using VersionVectorType = std::vector<VersionId>;
 
+inline std::optional<VersionId> resolve_version_id(
+        VersionId version, const std::shared_ptr<VersionMapEntry>& entry
+) {
+    const auto signed_version = static_cast<SignedVersionId>(version);
```
IvoDD: Doesn't this result in the following data flow `SignedVersionId` -> `VersionId` -> `SignedVersionId` of casts?

I think this is more confusing than just preserving the `SignedVersionId` throughout. I.e. I think we should make `VersionVectorType = std::vector<SignedVersionId>`

and even more specifically rename it as well to e.g. `SignedVersionVectorType`.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -277,6 +277,22 @@ inline std::shared_ptr<std::unordered_map<StreamId, AtomKey>> batch_get_specific
...
+
+    if (signed_version >= 0)
+        return version;
+
+    auto opt_latest = entry->get_first_index(true).first;
+
+    if (!opt_latest.has_value())
+        return std::nullopt;
+
+    return get_version_id_negative_index(opt_latest->version_id(), signed_version);
```
IvoDD: This feels like it duplicates the code [here](https://github.com/man-group/ArcticDB/blob/d9218d7631bd6db617f92f25cc12b3952d37ae2a/cpp/arcticdb/version/version_functions.hpp#L69-L83). Can't we deduplicate the two by abstracting a common function in a common place?

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -306,11 +322,18 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
                 auto sym_it = sym_versions.find(sym_version.first);
                 util::check(sym_it != sym_versions.end(), "Failed to find versions for symbol {}", sym_version.first);
                 const auto& versions = sym_it->second;
+
                 for (auto version : versions) {
-                    auto index_key = find_index_key_for_version_id(version, entry, include_deleted);
+                    auto opt_resolved = resolve_version_id(version, entry);
```
IvoDD: This can have somewhat surprising semantics:

E.g. if a symbol's version chain looks like:
```
sym:
v3 -> v2 -> v1 -> v0
```
And we request both versions 2 and -2 by calling:
```
batch_get_specific_versions({`sym`,  {2, -2}})
```

This would end up searching for the same version twice and the resulting map will contain only one deduplicated entry:
```
{{"sym", 2}, index_key_for_version_2}
```

I think this is the correct behavior (especially since this is only used for the snapshot codepaths), but it's worth adding a comment explaingin this like:
```
// batch_get_specific_versions allows requesting both positive and negative versions for the same symbol. If they end up corresponding to the same version they will be deduplicated in the output.
// For example if symbol has versions `v0, v1, v2 and v3`. And we request both for `[2, -2]`, method will return only `{2: index_key_for_version_2}`
```

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -203,6 +203,18 @@ inline std::optional<VersionId> get_version_id_negative_index(VersionId latest,
                                      : std::nullopt;
 }
 
+inline std::optional<VersionId> resolve_version_id(SignedVersionId signed_version_id, const VersionMapEntry& entry) {
```
markovskipetar, line 206: This is the general-purpose function

cpp/arcticdb/version/version_map_batch_methods.cpp
```diff
@@ -52,24 +52,12 @@ std::optional<AtomKey> get_specific_version_from_entry(
...
-                version_id = *opt_version_id;
-            } else {
-                return std::nullopt;
-            }
-        } else {
-            return std::nullopt;
-        }
-    }
-    return find_index_key_for_version_id(version_id, version_map_entry, include_deleted);
+    auto opt_version_id = resolve_version_id(specific_version.version_id_, *version_map_entry);
```
markovskipetar, line 55: reusing version_utils.hpp's resolve_version_id()

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -277,22 +277,6 @@ inline std::shared_ptr<std::unordered_map<StreamId, AtomKey>> batch_get_specific
 
 using VersionVectorType = std::vector<VersionId>;
 
-inline std::optional<VersionId> resolve_version_id(
```
markovskipetar: this function is moved to version_utils.hpp

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -324,13 +308,13 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
                 const auto& versions = sym_it->second;
 
                 for (auto version : versions) {
-                    auto opt_resolved = resolve_version_id(version, entry);
+                    auto opt_resolved = resolve_version_id(static_cast<SignedVersionId>(version), *entry);
```
markovskipetar: reusing version_utils.hpp's resolve_version_id()

cpp/arcticdb/version/version_map.hpp
```diff
@@ -865,20 +865,16 @@ class VersionMapImpl {
...
-                    ARCTICDB_DEBUG(
-                            log::version(),
-                            "Loaded as far as required value {}, have {} and there are {} total versions",
-                            requested_version_id,
-                            entry.load_progress_.oldest_loaded_index_version_,
-                            opt_latest->version_id()
-                    );
-                    return true;
-                }
+            auto opt_version_id = resolve_version_id(requested_version_id, entry);
```
markovskipetar: reusing version_utils.hpp's resolve_version_id()

cpp/arcticdb/version/version_functions.hpp
```diff
@@ -65,23 +65,12 @@ inline std::optional<AtomKey> get_specific_version(
...
-                version_id = *opt_version_id;
-            } else {
-                return std::nullopt;
-            }
-        } else {
-            return std::nullopt;
-        }
-    }
-    return find_index_key_for_version_id(version_id, entry, include_deleted);
+    auto opt_version_id = resolve_version_id(signed_version_id, *entry);
```
markovskipetar, line 68: reusing version_utils.hpp's resolve_version_id()

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -294,9 +297,7 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
             std::move(tasks_input),
             [store, version_map](auto sym_version) {
                 auto first_version = *std::min_element(std::begin(sym_version.second), std::end(sym_version.second));
```
IvoDD, line 297: Ah sorry I have just seen that there is a single call per symbol which finds `first_version = std::min_element`. This will actually be incorrect a lot of times.

We should either:
- Request both the earliest positive and the earliest negative
- Revert the SignedVersionId change and just raise on negative number inputs

I'm leaning more towards just raising for negative versions as a user error, but let's discuss offline. 
In this case the refactor to reuse `resolve_version_id` would still be helpful.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -865,20 +865,16 @@ class VersionMapImpl {
...
+            
+            if (opt_version_id.has_value() && entry.load_progress_.oldest_loaded_index_version_ <= *opt_version_id) {
+                ARCTICDB_DEBUG(
+                        log::version(),
```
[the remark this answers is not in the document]
IvoDD: +1 we can leave the error message as it was.

cpp/arcticdb/version/version_map.hpp
IvoDD: Good reuse but can't we refactor the `if` keys on this branch as well with your new method?

cpp/arcticdb/version/version_map.hpp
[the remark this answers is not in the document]
markovskipetar: done!

cpp/arcticdb/version/version_map.hpp
```diff
@@ -854,34 +854,20 @@ class VersionMapImpl {
...
-                            opt_latest->version_id()
-                    );
-                    return true;
-                }
-            }
-        }
-        return false;
+        auto opt_version_id = resolve_version_id(requested_version_id, entry);
+
+        if (!opt_version_id.has_value() || entry.load_progress_.oldest_loaded_index_version_ > *opt_version_id)
```
IvoDD: Have you rebased? I suspect there will be merge conflicts with #2894. Let's resolve the merge conflicts together.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -306,11 +308,18 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
                 auto sym_it = sym_versions.find(sym_version.first);
                 util::check(sym_it != sym_versions.end(), "Failed to find versions for symbol {}", sym_version.first);
                 const auto& versions = sym_it->second;
+
                 for (auto version : versions) {
-                    auto index_key = find_index_key_for_version_id(version, entry, include_deleted);
+                    auto opt_resolved = resolve_version_id(version, *entry);
```
IvoDD: I think this change is no longer required. It is guaranteed that all inputs are positive so we can just:
```
auto index_key = find_index_key_for_version_id(version, entry, include_deleted);
```

python/tests/integration/arcticdb/version_store/test_snapshot.py
```diff
@@ -408,6 +409,56 @@ def test_add_to_snapshot_multiple(basic_store_tombstone_and_pruning):
...
+
+
+@pytest.mark.storage
+def test_add_to_snapshot_mismatched_symbols_and_versions_raises(basic_store_tombstone_and_pruning):
+    lib = basic_store_tombstone_and_pruning
+    lib.write("s1", 1)
+    lib.snapshot("snap")
+    lib.write("s2", 2)
+
+    with pytest.raises(Exception):
```
IvoDD: What kind of exception does this raise? It should probably be a `UserInputException`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -2054,17 +2023,54 @@ struct WarnVersionTypeNotHandled {
...
+    if (check_if_version_id_already_exists(version_id, it->second)) {
+        return; // exisintg version_id was provided for the symbol -> do nothing
+    }
+
+    user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            "Only one version per symbol is allowed in snapshots. Symbol '{}' appears more than once", stream_id
+    );
+}
+
 std::map<StreamId, VersionVectorType> get_multiple_sym_versions_from_query(
```
IvoDD, line 2058: `get_multiple_sym_versions_from_query` will now end up returning a mapping StreamId -> vectpr<version_id> but the vector will always have a single element.

I think the old API was useful and might be helpful in the future. I would instead just check where that method is called that every map entry has a single element like:
```
auto sym_versions = get_multiple_sym_versions_from_query(stream_ids, version_queries);
auto symbol_with_more_than_one_version = std::ranges::find_if(sym_versions, [](const auto& [sym, versions]) {
    return versions.size() > 1;
});
if (symbol_with_more_than_one_version != sym_versions.end()) {
    auto stream_id = symbol_with_more_than_one_version->first;
    user_input::raise("Only one version per symbol is allowed in snapshots. Symbol '{}' appears more than once", stream_id);
}
```

python/tests/integration/arcticdb/version_store/test_snapshot.py
markovskipetar: done

cpp/arcticdb/version/version_map.hpp
markovskipetar: done

cpp/arcticdb/version/version_map_batch_methods.hpp
markovskipetar: done

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -278,8 +278,10 @@ inline std::shared_ptr<std::unordered_map<StreamId, AtomKey>> batch_get_specific
 using VersionVectorType = std::vector<VersionId>;
 
 /**
- * Returns multiple versions for the same symbol
+ * Returns a specific version per symbol.
```
IvoDD: I think the old description is correct after we reverted the change.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -862,17 +862,10 @@ class VersionMapImpl {
...
-            }
-        } else {
-            resolved_version_id = static_cast<VersionId>(requested_version_id);
+        auto resolved_version_id = resolve_version_id(requested_version_id, entry);
+
+        if (!resolved_version_id.has_value()) {
+            return false;
         }
 
         if (resolved_version_id <= opt_latest->version_id() &&
```
IvoDD: `resolved_version_id` is now an optional. Let's use `*resolved_version_id` or `resolved_version_id.value()` to extract its value. I think `std::optinal<>` comparisons can be surprising

python/tests/integration/arcticdb/version_store/test_snapshot.py
```diff
@@ -408,6 +409,56 @@ def test_add_to_snapshot_multiple(basic_store_tombstone_and_pruning):
...
+    lib = basic_store_tombstone_and_pruning
+    lib.write("s1", 1)
+    lib.write("s1", 2)
+    lib.snapshot("snap")
+    lib.write("s1", 3)
+
+    lib.add_to_snapshot("snap", ["s1", "s1"], as_ofs=[2, 2])
+    lib.write("s1", 99)
+
+    assert lib.read("s1", as_of="snap").data == 3  # v2 (0-indexed) holds data=3
```
IvoDD: Looks like we forgot to update the test with the new behavior

IvoDD (commented):
The PR descriptions seems out of date. Also CI is failing

cpp/arcticdb/version/version_map_batch_methods.hpp
markovskipetar: done!

cpp/arcticdb/version/version_map.hpp
markovskipetar: done!

python/tests/integration/arcticdb/version_store/test_snapshot.py
markovskipetar: done

IvoDD (approved):

python/tests/integration/arcticdb/version_store/test_basic_operations_scenarios.py
```diff
@@ -837,20 +837,14 @@ def test_add_to_snapshot_and_remove_from_snapshots_scenarios(basic_store):
             lib.read(symbol, as_of="snap")
 
 
-@pytest.mark.xfail(True, reason="Negative version numbers does not work, issue 10060901137")
 def test_add_to_snapshot_with_negative_numbers(basic_store):
```
ognyanstoimenov, line 840: This test duplicates the one in python/tests/integration/arcticdb/version_store/test_snapshot.py. Can we remove this one?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -2059,12 +2060,25 @@ std::map<StreamId, VersionVectorType> get_multiple_sym_versions_from_query(
...
+        const auto version_id = std::get<SpecificVersionQuery>(query).version_id_;
+
+        user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+                version_id >= 0,
+                "Negative version numbers are not supported in add_to_snapshot. Got version {} for symbol '{}'",
+                version_id,
+                *stream_id
+        );
+
+        sym_versions[*stream_id].push_back(std::get<SpecificVersionQuery>(query).version_id_);
```
ognyanstoimenov:
```suggestion
        sym_versions[*stream_id].emplace_back(version_id);
```

ognyanstoimenov (commented):
Minor comments. Lots of tests are failing on the CI though

ognyanstoimenov (approved):

PR #3062, opened 2026-04-30
Resampling performance improvement and sparse aggregation columns support

IvoDD:
#### Reference Issues/PRs
Monday ref: 11679866800

Depends on PRs #3091 and #3110 

### Issues

- There is complicated bucket hopping logic in three places: `generate_output_index_column`, `generate_resampling_output_column`, `SortedAggregator::aggregate`
- The bucket hopping logic involves many branches with loads of checks

### Changes (split per commit for easier review)

0. Adds C++ benchmarks which measures the CPU intensive part of resampling
1. Pure move of the `generate_output_index_column` to `sorted_aggregation.cpp`. 
   - This way all bucket hopping logic is in one place.
2. Construct a `ResampleMapping` in `generate_output_index_column` and use it directly in other methods. 
   - `ResampleMapping` just has a mapping from `output_row` to `(start_column_index, start_column_offset), (end_column_index, end_column_offset)`.
   - Resolves the 3 places with similar logic.
   - Makes the implementation of sparse aggregation easier.
3. Use [galloping search](https://en.wikipedia.org/wiki/Exponential_search) in `generate_output_index_column` to skip past all rows in a single bucket at once.
     - Index column construction was the bottleneck: aggregation vectorises well but index iteration does not.
     - Changes complexity from `O(num_input_rows + num_buckets)` to `O(num_buckets × log(rows_per_bucket))`.
     - Always ≤ `O(num_input_rows + num_buckets)` even when `num_buckets ≥ num_input_rows`.
  4. Preallocate the output index column to `min(num_buckets, num_input_rows)` instead of `num_buckets`.
     - Galloping search has a higher constant than linear scan and regresses at low rows per bucket.
     - Slightly improves the case where most buckets are empty due to smaller allocation.
  5. Use a runtime heuristic to choose between linear scan and galloping search.
     - Linear scan is faster below ~32 rows/bucket (because of smaller constant and better branch prediction); galloping search is faster above.
     - Threshold determined empirically from benchmarks at intermediate bucket counts. Extra benchmarking was done with more parametrization of the existing benchmark. Not kept in PR to avoid a huge amount of benchmarking code.
     - Recovers the Dense-100k and Empty regressions from commit 3 while retaining all gains elsewhere.
  6. Implement sparse resampling.
     - Small change made straightforward by the `ResampleMapping` from commit 2.
     - Minimal overhead for the dense case.

### Resample benchmark timings
                                               
`BM_resample/<rows_per_seg>/<num_segs>/<num_buckets>/<num_cols>`. Total rows ~1M.
  Source: `cpp/arcticdb/processing/test/benchmark_resample.cpp`. Times in **ms**, `--benchmark_min_time=2s`.

  | Regime | Args | rows/bucket | Description |
  |---|---|---|---|
  | Dense-1k | `100k × 10, 1k buckets` | ~1000 | Many rows/bucket, single row-slice |
  | Dense-100 | `100k × 10, 10k buckets` | ~100 | Medium rows/bucket, single row-slice |
  | Dense-10 | `100k × 10, 100k buckets` | ~10 | Few rows/bucket, single row-slice |
  | Spanning | `2k × 500, 100 buckets` | ~10k | Buckets span multiple row-slices |
  | Empty | `100k × 10, 10M buckets` | <1 | Bucket smaller than row spacing; most empty |

  **1 aggregation column**

  | # | Change | D-1k | D-100 | D-10 | Spanning | Empty |
  |---|---|---|---|---|---|---|
  | 0 | Baseline | 1.27 | 1.34 | 1.47 | 1.65 | 11.1 |
  | 1 | Code move | 1.02 (−20%) | 1.12 (−16%) | 1.27 (−14%) | 1.40 (−15%) | 11.1 (0%) |
  | 2 | ResampleMapping | 1.02 (−20%) | 1.12 (−16%) | 1.32 (−10%) | 1.40 (−15%) | 11.8 (+6%) |
  | 3 | Galloping search | 0.059 (−95%) | 0.385 (−71%) | 2.94 (+100%) | 0.285 (−83%) | 21.9 (+97%) |
  | 4 | Bounded allocation | 0.058 (−95%) | 0.396 (−70%) | 2.91 (+98%) | 0.291 (−82%) | 21.5 (+94%) |
  | 5 | Heuristic (lin/EUB) | 0.059 (−95%) | 0.383 (−71%) | 1.27 (−14%) | 0.293 (−82%) | 11.5 (+4%) |
  | 6 | Sparse-input support | 0.068 (−95%) | 0.449 (−66%) | 1.28 (−13%) | 0.296 (−82%) | 11.5 (+4%) |

  **100 aggregation columns**

  | # | Change | D-1k | D-100 | D-10 | Spanning | Empty |
  |---|---|---|---|---|---|---|
  | 0 | Baseline | 1.37 | 1.43 | 1.56 | 6.22 | 48.0 |
  | 1 | Code move | 1.11 (−19%) | 1.18 (−17%) | 1.34 (−14%) | 5.92 (−5%) | 46.2 (−4%) |
  | 2 | ResampleMapping | 1.11 (−19%) | 1.19 (−17%) | 1.39 (−11%) | 5.87 (−6%) | 50.4 (+5%) |
  | 3 | Galloping search | 0.148 (−89%) | 0.471 (−67%) | 2.96 (+90%) | 4.65 (−25%) | 63.1 (+31%) |
  | 4 | Bounded allocation | 0.148 (−89%) | 0.480 (−66%) | 2.95 (+89%) | 4.67 (−25%) | 44.1 (−8%) |
  | 5 | Heuristic (lin/EUB) | 0.149 (−89%) | 0.477 (−67%) | 1.33 (−15%) | 4.70 (−24%) | 35.9 (−25%) |
  | 6 | Sparse-input support | 0.158 (−88%) | 0.537 (−62%) | 1.35 (−13%) | 4.94 (−21%) | 36.0 (−25%) |

  Deltas vs baseline (row 0).

  #### Notes on benchmark results

  - Load average varied across runs so there are some artifacts in results like "Code move" improvements.
  - Galloping search improves the speed when there are more rows in a single bucket significantly. Thorough benchmarking showed exponential upper bound (EUB) becomes faster than linear search at ~32 rows per bucket. Hence we see some performance regressions in the 10 rows per bucket and in the mostly empty bucket cases.
  - Bounded allocation mostly helps the empty case as expected
  - Using the heuristic to choose between EUB and linear search helps when rows_per_bucket < 32. It is even more efficient than the baseline due to slightly better branch prediction (improved use of `ARCTICDB_LIKELY` and `ARCTICDB_UNLIKELY`).
  - Final state: every regime at or faster than baseline; Dense 1000 rows per bucket is the biggest winner with 20x improvement; Mostly empty bucket is the only usecase with no improvement and remains around baseline (+4%)

### DIFF
--- a/cpp/arcticdb/processing/sorted_aggregation.cpp
+++ b/cpp/arcticdb/processing/sorted_aggregation.cpp
@@ -87,15 +87,20 @@ std::pair<std::shared_ptr<Column>, ResampleMapping> generate_output_index_column
     using IndexTDT = ScalarTagType<DataTypeTag<data_type>>;
 
     // The output row count is bounded by both the number of buckets and the number of input rows
-    size_t total_input_rows{0};
-    for (const auto& col : input_index_columns) {
-        total_input_rows += col->row_count();
-    }
+    const size_t total_input_rows = std::accumulate(
+            input_index_columns.begin(),
+            input_index_columns.end(),
+            size_t{0},
+            [](size_t acc, const auto& col) { return acc + col->row_count(); }
+    );
     const auto max_output_rows = std::min(bucket_boundaries.size() - 1, total_input_rows);
     // Below this rows/bucket threshold, element-by-element iteration benchmarked to beat galloping search.
+    // The threshold assumes rows are distributed uniformly across buckets; the less uniform the distribution, the more
+    // galloping search wins, so this is a conservative threshold.
     constexpr size_t linear_scan_threshold = 32;
-    const size_t num_buckets = bucket_boundaries.size() > 1 ? bucket_boundaries.size() - 1 : 1;
-    const bool use_linear_scan = (total_input_rows / num_buckets) < linear_scan_threshold;
+    const size_t num_buckets = bucket_boundaries.size() - 1;
+    // Equivalent to `total_input_rows / num_buckets < linear_scan_threshold` and avoids potential division by zero.
+    const bool use_linear_scan = total_input_rows < linear_scan_threshold * num_buckets;
     const auto max_index_column_bytes = max_output_rows * get_type_size(data_type);
     auto output_index_column = std::make_shared<Column>(
             TypeDescriptor(data_type, Dimension::Dim0),
--- a/cpp/arcticdb/processing/sorted_aggregation.cpp
+++ b/cpp/arcticdb/processing/sorted_aggregation.cpp
@@ -74,15 +74,7 @@ SortedAggregatorOutputColumnInfo SortedAggregator<aggregation_operator, closed_b
 }
 
 template<ResampleBoundary closed_boundary>
-bool value_past_bucket_start(const timestamp bucket_start, const timestamp value) {
-    if constexpr (closed_boundary == ResampleBoundary::LEFT) {
-        return value >= bucket_start;
-    }
-    return value > bucket_start;
-}
-
-template<ResampleBoundary closed_boundary>
-std::shared_ptr<Column> generate_output_index_column(
+std::pair<std::shared_ptr<Column>, ResampleMapping> generate_output_index_column(
         const std::vector<std::shared_ptr<Column>>& input_index_columns,
         const std::vector<timestamp>& bucket_boundaries, const TimestampRange& date_range,
         const ResampleBoundary label_boundary
@@ -100,30 +92,42 @@ std::shared_ptr<Column> generate_output_index_column(
     auto output_index_column_it = output_index_column_data.template begin<IndexTDT>();
     size_t output_index_column_row_count{0};
 
+    ResampleMapping mapping;
+    mapping.reserve(bucket_boundaries.size());
+
     auto bucket_end_it = std::next(bucket_boundaries.cbegin());
     Bucket<closed_boundary> current_bucket{*std::prev(bucket_end_it), *bucket_end_it};
     bool current_bucket_added_to_index{false};
-    // Only include buckets that have at least one index value in range
-    for (const auto& input_index_column : input_index_columns) {
+    bool reached_end_of_buckets{false};
+    ResampleInputCursor close_cursor{input_index_columns.size(), 0};
+    for (auto&& [col_idx, input_index_column] : folly::enumerate(input_index_columns)) {
+        if (reached_end_of_buckets) {
+            break;
+        }
         auto index_column_data = input_index_column->data();
-        const auto cend = index_column_data.cend<IndexTDT>();
-        auto it = index_column_data.cbegin<IndexTDT>();
+        const auto cend = index_column_data.template cend<IndexTDT>();
+        auto it = index_column_data.template cbegin<IndexTDT>();
+        size_t offset_in_col{0};
         // In case the passed date_range does not span the whole segment we need to skip the index values
         // which are before the date range start.
         while (it != cend && *it < date_range.first) {
             ++it;
+            ++offset_in_col;
         }
-        for (; it != cend && *it <= date_range.second; ++it) {
+        for (; it != cend && *it <= date_range.second; ++it, ++offset_in_col) {
             if (ARCTICDB_LIKELY(current_bucket.contains(*it))) {
                 if (ARCTICDB_UNLIKELY(!current_bucket_added_to_index)) {
                     *output_index_column_it++ =
                             label_boundary == ResampleBoundary::LEFT ? *std::prev(bucket_end_it) : *bucket_end_it;
                     ++output_index_column_row_count;
+                    mapping.push_back({col_idx, offset_in_col});
                     current_bucket_added_to_index = true;
                 }
             } else {
                 advance_boundary_past_value<closed_boundary>(bucket_boundaries, bucket_end_it, *it);
                 if (ARCTICDB_UNLIKELY(bucket_end_it == bucket_boundaries.end())) {
+                    close_cursor = {col_idx, offset_in_col};
+                    reached_end_of_buckets = true;
                     break;
                 } else {
                     current_bucket.set_boundaries(*std::prev(bucket_end_it), *bucket_end_it);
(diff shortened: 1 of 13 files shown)


cpp/arcticdb/processing/test/test_resample.cpp
```diff
@@ -462,7 +462,7 @@ std::optional<Column> compute_output_column(
         const std::vector<timestamp>& bucket_boundaries, ResampleBoundary label_boundary
 ) {
     const std::vector<std::shared_ptr<Column>> idx_cols(input_index_columns.begin(), input_index_columns.end());
-    const TimestampRange full_range{std::numeric_limits<timestamp>::min(), std::numeric_limits<timestamp>::max()};
+    const TimestampRange full_range{std::numeric_limits<timestamp>::min() + 1, std::numeric_limits<timestamp>::max()};
```
alexowens90, line 466: Is this to avoid `NaT`? If so a comment would be useful

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -83,7 +83,13 @@ std::pair<std::shared_ptr<Column>, ResampleMapping> generate_output_index_column
     constexpr auto data_type = DataType::NANOSECONDS_UTC64;
     using IndexTDT = ScalarTagType<DataTypeTag<data_type>>;
 
-    const auto max_index_column_bytes = (bucket_boundaries.size() - 1) * get_type_size(data_type);
+    // The output row count is bounded by both the number of buckets and the number of input rows
+    size_t total_input_rows{0};
+    for (const auto& col : input_index_columns) {
+        total_input_rows += col->row_count();
+    }
```
alexowens90: nit: accumulate

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -89,6 +89,10 @@ std::pair<std::shared_ptr<Column>, ResampleMapping> generate_output_index_column
         total_input_rows += col->row_count();
     }
     const auto max_output_rows = std::min(bucket_boundaries.size() - 1, total_input_rows);
+    // Below this rows/bucket threshold, element-by-element iteration benchmarked to beat galloping search.
+    constexpr size_t linear_scan_threshold = 32;
+    const size_t num_buckets = bucket_boundaries.size() > 1 ? bucket_boundaries.size() - 1 : 1;
```
alexowens90: When is this condition not true?

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -89,6 +89,10 @@ std::pair<std::shared_ptr<Column>, ResampleMapping> generate_output_index_column
         total_input_rows += col->row_count();
     }
     const auto max_output_rows = std::min(bucket_boundaries.size() - 1, total_input_rows);
+    // Below this rows/bucket threshold, element-by-element iteration benchmarked to beat galloping search.
+    constexpr size_t linear_scan_threshold = 32;
```
alexowens90, line 100: I guess this is assuming a uniform distribution of rows in each bucket?

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -214,7 +219,30 @@ std::optional<Column> SortedAggregator<aggregation_operator, closed_boundary>::g
...
+            }
+            const size_t range_start = (col_idx == start.input_column_idx) ? start.offset : 0;
+            const size_t range_end_exclusive =
+                    (col_idx == end.input_column_idx) ? end.offset : static_cast<size_t>(col.last_row()) + 1;
+            if (range_start >= range_end_exclusive) {
+                continue;
+            }
+            const auto cnt = col.sparse_map().count_range_no_check(
+                    bv_size(range_start), bv_size(range_end_exclusive - 1), rs_index
+            );
```
alexowens90, line 249: An alternative is to use `get_next` and check if it is less than `range_end_exclusive`, although this cannot leverage the rank-select index.
Or we could compare the rank of range start and of range end. If they are the same, we know there are no set bits in the interval, and it might be faster.

cpp/arcticdb/processing/test/test_resample.cpp
IvoDD: `timestamp::min` was causing underflow issues [here](https://github.com/man-group/ArcticDB/blob/5e4edb720c03a4aa29be47a9bbf96f080b6e06c2/cpp/arcticdb/processing/sorted_aggregation.cpp#L136) when we use `date_range.first - 1` to find the `upper_bound`.

Added a comment to the test explaining the issue. We could also add an explicit if for `min int64` but I don't think it's that valueable as I don't imagine people will be actually using so low timestamps.

I guess there is a general weirdness to be handled for any method which takes `pd.NaT` in a date range. `(pd.NaT, some_timestamp)` will mostly work but `(some_timestamp, pd.NaT)` would not. Added a ticket for this [here](https://man312219.monday.com/boards/7852509418/pulses/12072902843)

cpp/arcticdb/processing/sorted_aggregation.cpp
IvoDD: was just worried in case of empty buckets or something I'd end up with division by zero. Switched to a multiplication check so removed the case.

cpp/arcticdb/processing/sorted_aggregation.cpp
IvoDD: yes, added to the comment.

For non uniform distribution galloping search will perform better. I was pesimistic with assuming uniform distribution so that after this change we only use the galloping search only when it is guaranteed to be faster.

cpp/arcticdb/processing/sorted_aggregation.cpp
IvoDD: `count_range_no_check` is essentially doing `count_to(end) - count_to(start)==0` which I expect will be as fast as `count_to(end)==count_to(start)`.

The `get_next` idea is worthwhile. It would require some benchmarking to proove.

In any case some profiling showed that the bottleneck for sparse data is in `aggregate` where we need to iterate the entire sparse map for each input column vs here where we only check which bucket ranges are empty. So since not on hot path I'll keep the current state.

cpp/arcticdb/processing/test/test_resample.cpp
alexowens90: Makes sense. Alex is also NaT-wrangling for the column stats stuff, so we should coordinate if changing things in this area

cpp/arcticdb/processing/sorted_aggregation.cpp
alexowens90: Yeh no problem, was just thinking out loud about other possible implementations. The performance characteristics of bitmagic aren't always obvious, but if this is never the bottleneck definitely fine to leave as is.

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_arrow_sparse.py
vasil-pashov, line 1: It would be nice if we can get hypothesis tests covering some basic scenarios against polars, no need to test all supported parameters as some are quite painful to test.

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -73,76 +78,183 @@ SortedAggregatorOutputColumnInfo SortedAggregator<aggregation_operator, closed_b
...
+    using IndexTDT = ScalarTagType<DataTypeTag<data_type>>;
+
+    // The output row count is bounded by both the number of buckets and the number of input rows
+    const size_t total_input_rows = std::accumulate(
+            input_index_columns.begin(),
+            input_index_columns.end(),
+            size_t{0},
+            [](size_t acc, const auto& col) { return acc + col->row_count(); }
+    );
+    const auto max_output_rows = std::min(bucket_boundaries.size() - 1, total_input_rows);
```
vasil-pashov, line 96: How can you end up with bucket_boundaries.size() - 1 > total_input_rows

vasil-pashov (approved):

cpp/arcticdb/processing/sorted_aggregation.cpp
IvoDD: If we have loads of empty buckets. E.g. use `resample("1h")` on a table which has a 24h frequency like `2026-01-01, 2026-01-02, 2026-01-03`

IvoDD:
The `time_compact_data` benchmark failure is unrelated. ColumnDataIterator changes were from a previous PR.

PR #3063, opened 2026-05-01
Enhancement 11518164303: method for checking fragmentation

alexowens90:
#### Reference Issues/PRs
[11518164303](https://man312219.monday.com/boards/7852509418/pulses/11518164303)

Adds a method to check what the effect of calling `compact_data_experimental` would be. Returned information includes whether any compaction would take place, the number of row-slices before and after compaction, and the exact distribution of row-slices pre and post compaction.

### DIFF
--- a/cpp/arcticdb/entity/compact_data_info.cpp
+++ b/cpp/arcticdb/entity/compact_data_info.cpp
@@ -1,0 +1,58 @@
+/* Copyright 2026 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software
+ * will be governed by the Apache License, version 2.0.
+ */
+
+#include <arcticdb/entity/compact_data_info.hpp>
+
+namespace arcticdb {
+namespace entity {
+
+CompactDataInfo::CompactDataInfo(
+        std::vector<uint64_t>&& row_slices_before, std::vector<uint64_t>&& row_slices_after,
+        VersionId version_id_before, VersionId version_id_after
+) :
+    row_slices_before_(std::move(row_slices_before)),
+    row_slices_after_(std::move(row_slices_after)),
+    version_id_before_(version_id_before),
+    version_id_after_(version_id_after) {}
+
+std::vector<uint64_t> CompactDataInfo::row_slices_before() const { return row_slices_before_; }
+
+std::vector<uint64_t> CompactDataInfo::row_slices_after() const { return row_slices_after_; }
+
+size_t CompactDataInfo::num_row_slices_before() const {
+    // By construction, row_slices_before_.size() != 1. If there is a single row-slice, the vector will look like [0, n]
+    // where n is the number of rows
+    return row_slices_before_.empty() ? 0 : row_slices_before_.size() - 1;
+}
+
+size_t CompactDataInfo::num_row_slices_after() const {
+    // By construction, row_slices_after_.size() != 1. If there is a single row-slice, the vector will look like [0, n]
+    // where n is the number of rows
+    return row_slices_after_.empty() ? 0 : row_slices_after_.size() - 1;
+}
+
+VersionId CompactDataInfo::version_id_before() const { return version_id_before_; }
+
+VersionId CompactDataInfo::version_id_after() const { return version_id_after_; }
+
+bool CompactDataInfo::will_do_work() const { return version_id_after_ != version_id_before_; }
+
+std::string CompactDataInfo::to_string() const {
+    return fmt::format(
+            "CompactDataInfo(will_do_work={}, version_id_before={}, version_id_after={}, num_row_slices_before={}, "
+            "num_row_slices_after={})",
+            will_do_work(),
+            version_id_before(),
+            version_id_after(),
+            num_row_slices_before(),
+            num_row_slices_after()
+    );
+}
+
+} // namespace entity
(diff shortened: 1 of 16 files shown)


python/arcticdb/version_store/library.py
```diff
@@ -3185,6 +3186,62 @@ def compact_symbol_list(self) -> None:
         """
         return self._nvs.compact_symbol_list()
 
+    def compact_data_dry_run_experimental(
```
alexowens90: Open to suggestions on better names for this

poodlewars: Maybe something like `explain_plan_compact_data`? I was thinking that when column stats is done we might want to have `explain_plan_read` showing which blocks we'd visit.

python/tests/unit/arcticdb/version_store/test_compact_data.py
```diff
@@ -73,6 +115,17 @@ def generic_compact_data_test_noop(lib, sym, rows_per_segment=None):
...
+    with qs.query_stats():
+        compact_data_info = lib.compact_data_dry_run_experimental(sym, rows_per_segment=rows_per_segment)
+        stats = qs.get_query_stats()
+    qs.reset_stats()
+    assert compact_data_info.num_row_slices_before == compact_data_info.num_row_slices_after
+    assert compact_data_info.row_slices_before == compact_data_info.row_slices_after
+    assert compact_data_info.version_id_before == compact_data_info.version_id_after
+    assert not compact_data_info.will_do_work
+    # No data keys read
+    assert query_stats_operation_count(stats, "Memory_GetObject", "TABLE_DATA") == 0
```
poodlewars, line 128: Could also check there were no "Memory_PutObject" operations, to check it really is a dry run

python/tests/unit/arcticdb/version_store/test_compact_data.py
```diff
@@ -91,8 +144,55 @@ def generic_compact_data_test_noop(lib, sym, rows_per_segment=None):
...
+        compact_data_info, vit_before_compaction.version, compacted_version, pre_compaction_index, post_compaction_index
+    )
+
+
+def test_compact_data_dry_run(in_memory_store_factory):
+    lib = in_memory_store_factory(segment_row_size=10)
+    sym = "test_compact_data_dry_run"
+    lib.write(sym, pd.DataFrame({"col": [0, 1, 2, 3, 4]}))
+    lib.append(sym, pd.DataFrame({"col": [5, 6, 7, 8, 9]}))
+    compact_data_info = lib.compact_data_dry_run_experimental(sym)
```
poodlewars: Could run this twice in a loop, again just to check it doesn't actually do any compacting

cpp/arcticdb/entity/compact_data_info.cpp
```diff
@@ -0,0 +1,54 @@
...
+    row_slices_after_(std::move(row_slices_after)),
+    version_id_before_(version_id_before),
+    version_id_after_(version_id_after) {}
+
+std::vector<uint64_t> CompactDataInfo::row_slices_before() const { return row_slices_before_; }
+
+std::vector<uint64_t> CompactDataInfo::row_slices_after() const { return row_slices_after_; }
+
+size_t CompactDataInfo::num_row_slices_before() const {
+    return row_slices_before_.empty() ? 0 : row_slices_before_.size() - 1;
```
poodlewars, line 30: Confused by the `- 1`, in particular that empty `row_slices_before_` and single element `row_slices_before_` both return 0?

poodlewars (approved):

python/arcticdb/version_store/library.py
alexowens90: Yeh that would work, although I'd probably flip it to `compact_data_explain_plan` for consistency with modifiers being suffixes like `read->read_batch`

python/arcticdb/version_store/library.py
```diff
@@ -3185,6 +3186,62 @@ def compact_symbol_list(self) -> None:
...
+            If invalid rows_per_segment is provided
+        SchemaException
+            If the existing data is recursively normalized
+
+        Examples
+        --------
+
+        >>> df = pd.DataFrame({"col": np.arange(100_000)})
+        >>> for idx in range(100):
+        >>>     lib.append("sym", df[idx * 1_000: (idx + 1) * 1_000])
```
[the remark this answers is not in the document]
alexowens90: This is demonstrably wrong and documented in the docstring

cpp/arcticdb/entity/compact_data_info.cpp
alexowens90: Single element `row_slices_before_` is impossible, if you have 1 row slice then the vector will be `[0, n]` where `n` is the number of rows. I'll add a comment

cpp/arcticdb/entity/compact_data_info.hpp
```diff
@@ -0,0 +1,44 @@
...
+class CompactDataInfo {
+  public:
+    CompactDataInfo(
+            std::vector<uint64_t>&& row_slices_before, std::vector<uint64_t>&& row_slices_after,
+            VersionId version_id_before, VersionId version_id_after
+    );
+
+    ARCTICDB_MOVE_COPY_DEFAULT(CompactDataInfo);
+
+    std::vector<uint64_t> row_slices_before() const;
```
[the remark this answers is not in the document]
alexowens90: Not a problem with the more pythonic `for i in compact_data_info.row_slices_before` loop implementation

phoebusm (approved):

PR #3069, opened 2026-05-05
Performance improvements for reading column stats

poodlewars:
In this PR we:

- Only decompress and load statistics for columns that are in the user's query builder filters
- Only load statistics within the user specified `date_range`
- Iterate over the column stats segment column-wise rather than row-wise
- Load the column stats in to a column-oriented layout in memory

The profiling script,
[bench.py](https://github.com/user-attachments/files/27394513/bench.py) writes a column stats segment with ~200 columns and 100k rows, which corresponds to a 10B row symbol with our normal row slicing. It then times a read query for which column stats can prune everything, filtering on 10 of the 100 columns in the dataframe.

After all changes:

```
lib.read zero-match (rows=20000, num_cols=100, rows_per_segment=2, filter_cols=10, iters=20, column_stats=on):
  total block      = 197.2 ms
  mean             = 9.86 ms
  median           = 9.91 ms
  min              = 9.41 ms
  max              = 10.51 ms
```


Master:

```
(tree-two-venv) aseaton@dlonapatcs502:~/source/ArcticDB-worktree/tree-two/python DEV $ python ../scripts/profiling/profile_many_columns.py --rows 20000 --num-cols 100 --rows-per-segment 2
warmup x2
timed x20

lib.read zero-match (rows=20000, num_cols=100, rows_per_segment=2, filter_cols=10, iters=20, column_stats=on):
  total block      = 5281.3 ms
  mean             = 264.06 ms
  median           = 258.79 ms
  min              = 253.26 ms
  max              = 288.51 ms
```

Without the "flattened structure" in ColumnStatsData we get these results, so it seems worth keeping, although it will make things more complex when we add strings and Bloom filters:

```
lib.read zero-match (rows=20000, num_cols=100, rows_per_segment=2, filter_cols=10, iters=20, column_stats=on):
  total block      = 505.4 ms
  mean             = 25.27 ms
  median           = 23.28 ms
  min              = 21.86 ms
  max              = 42.47 ms
```

I also tried a more complex approach with type erased uint8_t buffers for the stats, with these results. The improvement does not seem worth the complexity here (it will make strings and bloom filters harder).

```
lib.read zero-match (rows=20000, num_cols=100, rows_per_segment=2, filter_cols=10, iters=20, column_stats=on):
  total block      = 171.2 ms
  mean             = 8.56 ms
  median           = 8.51 ms
  min              = 8.04 ms
  max              = 9.30 ms
```

I've also added an ASV benchmark to regression-test the improvement, which had these results:

  | Benchmark | `column_stats_enabled` | Master `668b267d` | HEAD `45dacae7` | Ratio |                                                                                                                                                                                                                              
  | --- | :---: | ---: | ---: | :---: |
  | `time_filter_10_columns_zero_match` | `True`  | 157 ± 1 ms   | **6.15 ± 0.1 ms**  | **0.04** |                                                                                                                                                                                                                  
  | `time_filter_10_columns_zero_match_with_date_range` | `True`  | 154 ± 2 ms   | **5.30 ± 0.05 ms** | **0.03** |                                                                                                                                                                                                  
  | `time_filter_10_columns_zero_match` | `False` | 161 ± 9 ms   | 166 ± 10 ms        | 1.03 |                                                                                                                                                                                                                      
  | `time_filter_10_columns_zero_match_with_date_range` | `False` | 92.8 ± 3 ms  | 82.8 ± 3 ms        | 0.89 |

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -190,6 +190,8 @@ When working on a feature branch, maintain a work log in `docs/claude/plans/<bra
 
 Code style is enforced by `make lint` **Always run `make lint` after making code changes.**
 
+Prefer `vec.at(i)` over `vec[i]` for `std::vector` and `std::array` access.
+
 ### Git Commits
 
 - Do not add "Generated with AI" or "Co-Authored-By" lines to commit messages
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -190,8 +190,6 @@ When working on a feature branch, maintain a work log in `docs/claude/plans/<bra
 
 Code style is enforced by `make lint` **Always run `make lint` after making code changes.**
 
-Prefer `vec.at(i)` over `vec[i]` for `std::vector` and `std::array` access.
-
 ### Git Commits
 
 - Do not add "Generated with AI" or "Co-Authored-By" lines to commit messages
--- a/cpp/arcticdb/entity/descriptor_item.hpp
+++ b/cpp/arcticdb/entity/descriptor_item.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <arcticdb/entity/atom_key.hpp>
+#include <arcticdb/storage/key_segment_pair.hpp>
 #include <fmt/format.h>
 #include <string>
 
@@ -31,7 +32,7 @@ struct DescriptorItem {
     std::optional<timestamp> end_index_;
     std::optional<TimeseriesDescriptor> timeseries_descriptor_;
     std::optional<SegmentInMemory> index_segment_;
-    std::optional<SegmentInMemory> column_stats_segment_;
+    std::optional<storage::KeySegmentPair> column_stats_segment_;
 
     std::string symbol() const { return fmt::format("{}", key_.id()); }
     uint64_t version() const { return key_.version_id(); }
@@ -41,6 +42,6 @@ struct DescriptorItem {
     std::optional<TimeseriesDescriptor> timeseries_descriptor() const { return timeseries_descriptor_; }
     entity::AtomKey key() const { return key_; }
     std::optional<SegmentInMemory> index_segment() const { return index_segment_; }
-    std::optional<SegmentInMemory> column_stats_segment() const { return column_stats_segment_; }
+    std::optional<storage::KeySegmentPair> column_stats_segment() const { return column_stats_segment_; }
 };
 } // namespace arcticdb
--- a/cpp/arcticdb/entity/descriptor_item.hpp
+++ b/cpp/arcticdb/entity/descriptor_item.hpp
@@ -8,8 +8,8 @@
 
 #pragma once
 
+#include <arcticdb/codec/segment.hpp>
 #include <arcticdb/entity/atom_key.hpp>
-#include <arcticdb/storage/key_segment_pair.hpp>
 #include <fmt/format.h>
 #include <string>
 
@@ -32,7 +32,7 @@ struct DescriptorItem {
     std::optional<timestamp> end_index_;
     std::optional<TimeseriesDescriptor> timeseries_descriptor_;
     std::optional<SegmentInMemory> index_segment_;
-    std::optional<storage::KeySegmentPair> column_stats_segment_;
+    std::shared_ptr<Segment> column_stats_segment_;
 
     std::string symbol() const { return fmt::format("{}", key_.id()); }
     uint64_t version() const { return key_.version_id(); }
@@ -42,6 +42,6 @@ struct DescriptorItem {
     std::optional<TimeseriesDescriptor> timeseries_descriptor() const { return timeseries_descriptor_; }
     entity::AtomKey key() const { return key_; }
     std::optional<SegmentInMemory> index_segment() const { return index_segment_; }
-    std::optional<storage::KeySegmentPair> column_stats_segment() const { return column_stats_segment_; }
+    std::shared_ptr<Segment> column_stats_segment() const { return column_stats_segment_; }
 };
 } // namespace arcticdb
--- a/cpp/arcticdb/pipeline/column_stats_filter.cpp
+++ b/cpp/arcticdb/pipeline/column_stats_filter.cpp
@@ -9,6 +9,8 @@
 #include <arcticdb/pipeline/column_stats_filter.hpp>
 #include <arcticdb/pipeline/column_stats_dispatch.hpp>
 
+#include <arcticdb/codec/codec.hpp>
+#include <arcticdb/entity/stream_descriptor.hpp>
 #include <arcticdb/pipeline/column_stats.hpp>
 #include <arcticdb/pipeline/index_fields.hpp>
 #include <arcticdb/pipeline/value.hpp>
@@ -16,86 +18,40 @@
 #include <arcticdb/stream/stream_utils.hpp>
 #include <arcticdb/processing/query_planner.hpp>
 
+#include <iterator>
+#include <unordered_set>
+
 namespace arcticdb {
-
-namespace {
-
-std::optional<Value> extract_value_from_column(
-        const SegmentInMemory::iterator& it, size_t col_idx, DataType data_type
-) {
-    return details::visit_type(data_type, [&it, col_idx](auto tag) -> std::optional<Value> {
-        using type_info = ScalarTypeInfo<decltype(tag)>;
-        // Only numeric types are supported at the moment.
-        if constexpr (is_numeric_type(type_info::data_type) || is_time_type(type_info::data_type) ||
-                      is_bool_type(type_info::data_type)) {
-            auto opt_val = it->scalar_at<typename type_info::RawType>(col_idx);
-            if (opt_val.has_value()) {
-                return Value{*opt_val, type_info::data_type};
-            }
-        }
-        return std::nullopt;
-    });
-}
-
-} // anonymous namespace
 
 bool is_column_stats_enabled() { return ConfigsMap::instance()->get_int("ColumnStats.UseForQueries", 0) == 1; }
 
-bool should_try_column_stats_read(const ReadQuery& read_query) {
-    if (!is_column_stats_enabled()) {
-        return false;
-    }
-    if (read_query.clauses_.empty()) {
-        return false;
-    }
-    for (const auto& clause : read_query.clauses_) {
-        auto& clause_type = folly::poly_type(*clause);
-        if (clause_type == typeid(DateRangeClause) || clause_type == typeid(RowRangeClause)) {
-            continue;
-        }
-        if (clause_type == typeid(FilterClause)) {
-            return true;
-        }
-        break;
-    }
-    return false;
+bool ColumnStatsQueryMetadata::should_try_column_stats_read() const {
+    return is_column_stats_enabled() && !filter_expressions.empty();
 }
 
 StatsVariantData evaluate_ast_node_against_stats(
-        const VariantNode& node, const ExpressionContext& expression_context, const StatsRowVector& stats_rows
+        const VariantNode& node, const ExpressionContext& expression_context, const StatsRowIndices& row_indices,
+        const ColumnStatsData& column_stats
 ) {
     return util::variant_match(
             node,
             [&](const ColumnName& column_name) -> StatsVariantData {
-                std::vector<ColumnStatsValues> result;
-                result.reserve(stats_rows.size());
-                for (const auto& row : stats_rows) {
-                    if (!row) {
-                        result.emplace_back(std::nullopt, std::nullopt);
-                        continue;
-                    }
-                    if (auto it = row->stats_for_column.find(column_name.value); it != row->stats_for_column.end()) {
-                        result.push_back(it->second);
-                    } else {
-                        result.emplace_back(std::nullopt, std::nullopt);
-                    }
+                if (auto slot = column_stats.slot_for_column(column_name.value); slot.has_value()) {
+                    return column_stats.materialize_slot(*slot, row_indices);
                 }
-                util::check(
-                        result.size() == stats_rows.size(), "Expected the result to have the same size as stats_rows"
-                );
-                return result;
+                return std::vector<ColumnStatsValues>(row_indices.size());
             },
             [&](const ValueName& value_name) -> StatsVariantData {
                 return expression_context.values_.get_value(value_name.value);
             },
             [&](const ExpressionName& expression_name) -> StatsVariantData {
                 auto expr = expression_context.expression_nodes_.get_value(expression_name.value);
-                return compute_stats(expression_context, *expr, stats_rows);
+                return compute_stats(expression_context, *expr, row_indices, column_stats);
             },
             [&](const ValueSetName& value_set_name) -> StatsVariantData {
                 return expression_context.value_sets_.get_value(value_set_name.value);
             },
-            [&](const auto&) -> StatsVariantData { return std::vector(stats_rows.size(), StatsComparison::UNKNOWN); }
+            [&](const auto&) -> StatsVariantData { return std::vector(row_indices.size(), StatsComparison::UNKNOWN); }
     );
 }
 
@@ -157,27 +113,30 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
 }
 
 StatsVariantData compute_stats(
-        const ExpressionContext& expression_context, const ExpressionNode& node, const StatsRowVector& stats_rows
+        const ExpressionContext& expression_context, const ExpressionNode& node, const StatsRowIndices& row_indices,
+        const ColumnStatsData& column_stats
 ) {
     if (is_binary_operation(node.operation_type_)) {
-        auto left = evaluate_ast_node_against_stats(node.left_, expression_context, stats_rows);
-        auto right = evaluate_ast_node_against_stats(node.right_, expression_context, stats_rows);
+        auto left = evaluate_ast_node_against_stats(node.left_, expression_context, row_indices, column_stats);
+        auto right = evaluate_ast_node_against_stats(node.right_, expression_context, row_indices, column_stats);
         return dispatch_binary_stats(left, right, node.operation_type_);
     }
     if (is_unary_operation(node.operation_type_)) {
-        auto left = evaluate_ast_node_against_stats(node.left_, expression_context, stats_rows);
+        auto left = evaluate_ast_node_against_stats(node.left_, expression_context, row_indices, column_stats);
         return dispatch_unary_stats(left, node.operation_type_);
     }
-    return std::vector(stats_rows.size(), StatsComparison::UNKNOWN);
+    return std::vector(row_indices.size(), StatsComparison::UNKNOWN);
 }
 
-ColumnStatsData::ColumnStatsData(SegmentInMemory&& segment, const TimeseriesDescriptor& tsd) {
+ColumnStatsData::ColumnStatsData(
+        SegmentInMemory&& segment, const TimeseriesDescriptor& tsd,
+        std::optional<std::pair<timestamp, timestamp>> date_range
+) {
     using namespace arcticc::pb2::column_stats_pb2;
     if (segment.row_count() == 0) {
         return;
     }
 
-    // Build reverse lookup: stats_seg_offset -> (column_name, stat_type)
     ColumnStatsHeader header;
     auto* metadata = segment.metadata();
     util::check(metadata != nullptr, "Column stats segment has no metadata");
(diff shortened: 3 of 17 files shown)


scripts/profiling/profile_zero_match.py
poodlewars, line 1: Do not merge this

cpp/arcticdb/pipeline/column_stats_filter.hpp
```diff
@@ -12,12 +12,17 @@
...
+#include <arcticdb/util/bitset.hpp>
 
 #include <memory>
 #include <optional>
+#include <unordered_map>
 #include <unordered_set>
 #include <utility>
 #include <vector>
+#include <gtest/gtest_prod.h>
+#include <out/linux-release-py314-build/proto/arcticc/pb2/proto/column_stats.pb.h>
```
poodlewars: **Critical**: This is a hardcoded build-output path containing the preset name (`linux-release-py314-build`). It will only resolve for that one preset and will break builds for `linux-debug`, conda, Windows, macOS, and any other Python ABI. The other files including this header use `<column_stats.pb.h>` directly (see `column_stats.hpp`, `version_core.cpp`, `unsorted_aggregation.cpp`). Replace with:

```cpp
#include <column_stats.pb.h>
```

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -751,33 +751,6 @@ def test_column_stats_with_row_range(
     assert_frame_equal(result, expected)
 
 
-def test_column_stats_with_date_range(in_memory_version_store, clear_query_stats, column_stats_filtering_enabled):
```
poodlewars, line 754: This just duplicated test_column_stats_query_optimisation_with_date_range

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,234 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
-                    "MIN and MAX should both be present or both be absent, col_name={}",
-                    col_name
+                    start_index_ts >= prev_start && end_index_ts >= prev_end,
+                    "Column stats segment start_index/end_index must be monotonically increasing "
+                    "(violated at row {})",
+                    r
             );
-            if (!stats.min) {
-                stats.column_absent = true;
+        }
```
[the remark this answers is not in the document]
poodlewars: I think Claude is right about this. I'll relax this.

python/benchmarks/column_stats.py
```diff
@@ -537,3 +537,93 @@ def teardown(self, *args):
...
+    def time_filter_10_columns_zero_match(self, *args):
+        self.lib.read(self.symbol, columns=self.read_columns, query_builder=self.query)
+
+    def time_filter_10_columns_zero_match_with_date_range(self, *args):
+        self.lib.read(
+            self.symbol,
+            columns=self.read_columns,
+            date_range=self.date_range_half,
+            query_builder=self.query,
+        )
```
alexowens90, line 629: Can include peakmem for completeness

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2794,17 +2799,20 @@ static folly::Future<VersionIdentifier> fetch_index_and_column_stats(
...
+    if (query_metadata.should_try_column_stats_read()) {
         auto column_stats_key = index_key_to_column_stats_key(versioned_item.key_);
         storage::ReadKeyOpts stats_read_opts{.dont_warn_about_missing_key = true};
-        column_stats_future =
-                store->read(column_stats_key, stats_read_opts)
-                        .thenValue([](std::pair<VariantKey, SegmentInMemory>&& key_seg) -> OptionalKeySeg {
-                            return std::move(key_seg.second);
-                        });
+        column_stats_future = store->read_compressed(column_stats_key, stats_read_opts)
+                                      .thenValue(
```
alexowens90: `read_compressed` will execute on an IO thread, and therefore so will this lambda (although possibly on a different thread in the pool).
If the lambda is cheap, use `thenValueInline` to ensure it stays on the same thread. If not, switch to the CPU executor.

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -338,38 +399,124 @@ FilterQuery<index::IndexSegmentReader> create_column_stats_filter(
     };
 }
 
-FilterQuery<index::IndexSegmentReader> create_column_stats_filter(
-        SegmentInMemory&& column_stats_segment, const TimeseriesDescriptor& tsd,
-        const std::vector<std::shared_ptr<Clause>>& clauses
-) {
-    std::vector<std::shared_ptr<ExpressionContext>> filter_expressions;
-
+ColumnStatsQueryMetadata column_stats_query_metadata(const std::vector<std::shared_ptr<Clause>>& clauses) {
```
alexowens90: What's the advantage of this being a free function rather than a ctor of `ColumnStatsQueryMetadata`?

cpp/arcticdb/pipeline/column_stats_filter.hpp
```diff
@@ -44,41 +49,85 @@ struct ColumnStatsValues {
...
+    ) const;
 
   private:
-    std::vector<ColumnStatsRow> rows_;
-    // (start_index, end_index) -> row index
-    // The index values in the column stats segment are just rowcounts if the symbol is string-indexed
+    FRIEND_TEST(ColumnStatsDataTest, FindStatsAllRowsPresent);
+    FRIEND_TEST(ColumnStatsDataTest, DateRangePrunesNonOverlappingRows);
+    FRIEND_TEST(ColumnStatsDataTest, DuplicateIndexPairDoesNotAffectOtherRows);
+    FRIEND_TEST(ColumnStatsDataTest, SparseColumnAbsentMarkedCorrectly);
```
alexowens90: I didn't know about this, cool

cpp/arcticdb/pipeline/column_stats_filter.hpp
```diff
@@ -94,21 +143,43 @@ FilterQuery<index::IndexSegmentReader> create_column_stats_filter(
...
+        ColumnStatsQueryMetadata&& query_metadata
+);
 
 /**
- * Create a column stats filter from an already-read column stats segment and query clauses.
+ * Create a column stats filter from an already-decoded column stats segment.
+ *
+ * Used when the column stats segment has been pre-loaded (e.g. via PreloadedIndexQuery) and the
+ * compressed bytes are no longer available. Date-range row pruning still applies, but column-set
+ * filtering does not — the caller has already paid the full decode cost.
```
alexowens90: I _think_ this is overkill. `collect_schema` always returns to storage:
https://github.com/man-group/ArcticDB/blob/master/python/tests/unit/arcticdb/version_store/test_collect_schema.py#L303-L309
so we always know the full query when calling `collect` with a preloaded index, and we could always use the decode-only-required columns approach.

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,234 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
-        for (size_t col_idx = end_index_column_offset + 1; col_idx < fields.size(); ++col_idx) {
-            const auto& field = fields[col_idx];
-            auto lookup_it = stats_at_column_index.find(col_idx);
-            if (lookup_it == stats_at_column_index.end()) {
+    // Gather metadata about the statistics we're interested in
+    std::vector<StatsMetadataForColumn> stats_metadata;
+    stats_metadata.reserve(header.stats_by_column().size());
+    for (const auto& [data_col_offset, entry_list] : header.stats_by_column()) {
+        StatsMetadataForColumn stats_metadata_for_column;
+        stats_metadata_for_column.col_name = std::string{tsd.fields().at(data_col_offset).name()};
```
alexowens90, line 144: Could add a `util::check` that `data_col_offset < tsd.fields().size()`, if it somehow isn't the error message will be very confusing

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,234 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
-    std::unordered_map<size_t, std::pair<std::string, ColumnStatsType>> stats_at_column_index;
-    for (const auto& [data_col_offset, entry_list] : header.stats_by_column()) {
-        std::string col_name{tsd.fields().at(data_col_offset).name()};
-        for (const auto& entry : entry_list.entries()) {
-            stats_at_column_index.emplace(entry.stats_seg_offset(), std::make_pair(col_name, entry.type()));
-        }
-    }
-
-    // Future aseaton consider iterating column-wise rather than row-wise?
+    segment.init_column_map();
```
alexowens90: This is `O(n)` in the number of columns in the segment, which could dominate for wide dfs in the case where we have decoded all columns (which I think we don't need, see other comments)

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,234 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
+            if constexpr (is_numeric_type(type_info::data_type) || is_time_type(type_info::data_type) ||
+                          is_bool_type(type_info::data_type)) {
+                using RawType = type_info::RawType;
+
+                for (const auto& entry : stats_metadata_for_column.entries) {
+                    const auto& column = segment.column(static_cast<position_t>(entry.segment_col_idx));
+                    const bool is_min = entry.stat_type == MIN_V1;
+                    auto& dest = is_min ? stats_for_column.mins : stats_for_column.maxes;
+
+                    if (!column.is_sparse() && static_cast<size_t>(column.row_count()) == segment_row_count) {
```
alexowens90: If constructed correctly (i.e. by calling `set_row_id` on the column at read time with the segment row count) then the second condition here is unnecessary

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,234 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
+                        // Dense
+                        auto it = column.begin<typename type_info::TDT>();
```
alexowens90: The column data iterators are more robust than these, and then you can use the stuff like `for_each_enumerated` in the column algorithms to avoid such duplicated logic and `scalar_at`

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,234 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
     }
 
-    for (const auto& key : duplicate_keys_) {
+    index_to_row_.reserve(num_rows_);
+    std::unordered_set<std::pair<timestamp, timestamp>, util::PairHasher> duplicate_keys;
+    for (size_t r = 0; r < num_rows_; ++r) {
+        auto key = std::make_pair(start_indices_.at(r), end_indices_.at(r));
+        if (auto [_, inserted] = index_to_row_.emplace(key, r); !inserted) {
+            // Duplicate (start_index, end_index) — can happen with timestamp indices when multiple
+            // segments span the same time range.
```
alexowens90, line 272: Hmmm this is an interesting case we hadn't considered before. I guess basically means that start/end index values aren't great unique identifiers of row-slices. Not for this PR, but maybe we should just switch to using a row-slice index?

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,234 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
+            result_entry.min = stats.mins.at(r);
+            result_entry.max = stats.maxes.at(r);
+        } else {
+            result_entry.column_absent = true;
+        }
+    }
+    return result;
+}
+
+ColumnStatsValues ColumnStatsData::stats_for(const std::string& col_name, size_t row) const {
```
alexowens90: Looks like this is just for testing? Could pull it out to a free function utility method compiled only for tests?

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -338,38 +399,124 @@ FilterQuery<index::IndexSegmentReader> create_column_stats_filter(
...
-        SegmentInMemory&& column_stats_segment, const TimeseriesDescriptor& tsd,
-        const std::vector<std::shared_ptr<Clause>>& clauses
-) {
-    std::vector<std::shared_ptr<ExpressionContext>> filter_expressions;
-
+ColumnStatsQueryMetadata column_stats_query_metadata(const std::vector<std::shared_ptr<Clause>>& clauses) {
+    // The clauses eligible for column stats use:
+    // - FilterClauses contribute filter expressions and columns of interest
+    // - DateRangeClauses contribute their range
+    // - RowRangeClauses are skipped
```
alexowens90, line 427: I guess we can't use these to prune at the moment because the stats segment doesn't have start row/end row information?
With the other comment on start/end index not working as unique identifiers, maybe we need to reconsider the stats key structure slightly

poodlewars: 11990503071

cpp/arcticdb/pipeline/column_stats_filter.cpp
poodlewars: Yes I think I agree. 11990503071

CLAUDE.md
```diff
@@ -190,6 +190,8 @@ When working on a feature branch, maintain a work log in `docs/claude/plans/<bra
 
 Code style is enforced by `make lint` **Always run `make lint` after making code changes.**
 
+Prefer `vec.at(i)` over `vec[i]` for `std::vector` and `std::array` access.
```
vasil-pashov: I dissagree, especially in tight loops.. `at` has explicit bounds check, operator[] has that only in debug so I think it's good enough. When we switch to new compiler we can also use hardend mode of the standard library in the tests so that in tests operator[] has bounds checks but in release it doesnt.

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,221 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
+    // Construct start_indices and end_indices with date range pruning.
+    // Also construct the interval [first_kept, last_kept_excl) to quickly skip stats for row ranges
+    // we are not interested in when we read the statistics themselves.
+    size_t first_kept = segment_row_count;
+    size_t last_kept_excl = 0;
+    start_indices_.reserve(segment_row_count);
+    end_indices_.reserve(segment_row_count);
+    using TsTDT = ScalarTagType<DataTypeTag<DataType::NANOSECONDS_UTC64>>;
+    auto start_it = start_index_col.begin<TsTDT>();
+    auto end_it = end_index_col.begin<TsTDT>();
```
vasil-pashov: I suggest using the ColumnDataIterator @alexowens90 can confirm but AFAIK there were some differences in performance. I think the TypedColumnIterator is not widely used and might have a bug or two. When I worked on resampling I tried to used them and there was some weird issue happening only on Windows that I couldn't figure out, when I switched to ColumnDataIterator the issue was gone.

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,221 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
-        auto left = evaluate_ast_node_against_stats(node.left_, expression_context, stats_rows);
+        auto left = evaluate_ast_node_against_stats(node.left_, expression_context, row_indices, column_stats);
         return dispatch_unary_stats(left, node.operation_type_);
     }
-    return std::vector(stats_rows.size(), StatsComparison::UNKNOWN);
+    return std::vector(row_indices.size(), StatsComparison::UNKNOWN);
 }
 
-ColumnStatsData::ColumnStatsData(SegmentInMemory&& segment, const TimeseriesDescriptor& tsd) {
+ColumnStatsData::ColumnStatsData(
```
vasil-pashov: This constructor became quite large. Can we split?

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,221 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
+std::vector<ColumnStatsValues> ColumnStatsData::values_for_column(
+        const std::string& col_name, const std::vector<std::optional<size_t>>& row_indices
+) const {
+    std::vector<ColumnStatsValues> result(row_indices.size());
+    if (num_rows_ == 0) {
+        return result;
+    }
+    auto it = stats_by_column_.find(col_name);
+    if (it == stats_by_column_.end()) {
+        return result;
```
vasil-pashov, line 340: I find it a bit odd we're returning non-empty vector in case the column name is not in the map. Maybe it would be better to `return {}` or an optional

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -157,123 +110,221 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
...
     }
-    return nullptr;
+    return std::nullopt;
 }
 
-bool ColumnStatsData::empty() const { return rows_.empty(); }
+std::vector<ColumnStatsValues> ColumnStatsData::values_for_column(
+        const std::string& col_name, const std::vector<std::optional<size_t>>& row_indices
+) const {
+    std::vector<ColumnStatsValues> result(row_indices.size());
```
vasil-pashov, line 334: nit: I'd rather reserve and emplace_back

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -439,9 +439,14 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
             version_query.content_,
             [&](const std::shared_ptr<PreloadedIndexQuery>& preloaded_index_query) -> VersionIdentifier {
                 // Clone because Python holds the PreloadedIndexQuery across multiple collect() calls.
-                auto column_stats = preloaded_index_query->column_stats_seg_.has_value()
-                                            ? std::optional{preloaded_index_query->column_stats_seg_->clone()}
-                                            : std::nullopt;
+                std::optional<ColumnStatsSource> column_stats;
+                if (preloaded_index_query->column_stats_seg_) {
+                    const auto& key_seg = *preloaded_index_query->column_stats_seg_;
+                    column_stats = ColumnStatsSource{
```
vasil-pashov: nit: `column_stats.emplace`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -519,14 +524,14 @@ folly::Future<DescriptorItem> LocalVersionedEngine::get_descriptor(AtomKey&& k,
...
         auto column_stats_key = index_key_to_column_stats_key(key);
         storage::ReadKeyOpts stats_read_opts{.dont_warn_about_missing_key = true};
-        column_stats_future = store()->read(column_stats_key, stats_read_opts)
-                                      .thenValue(
-                                              [](std::pair<VariantKey, SegmentInMemory>&& key_seg
-                                              ) -> std::optional<SegmentInMemory> { return std::move(key_seg.second); }
+        column_stats_future = store()->read_compressed(column_stats_key, stats_read_opts)
+                                      .thenValueInline(
+                                              [](storage::KeySegmentPair&& key_seg
+                                              ) -> std::optional<storage::KeySegmentPair> { return std::move(key_seg); }
```
vasil-pashov: Why is the return type optional?

cpp/arcticdb/pipeline/column_stats_filter.cpp
poodlewars: I've left this as is - although this section is a bit more complex, it keeps downstream code a bit simpler if it doesn't have to handle an empty vector / missing optional special case.

cpp/arcticdb/version/local_versioned_engine.cpp
poodlewars: To get the correct typing for `column_stats_future`, which is constructed even if stats are not enabled.

cpp/arcticdb/pipeline/column_stats_filter.cpp
poodlewars: https://github.com/man-group/ArcticDB/pull/3069#discussion_r3265292581

vasil-pashov (approved):

cpp/arcticdb/entity/descriptor_item.hpp
```diff
@@ -31,7 +32,7 @@ struct DescriptorItem {
     std::optional<timestamp> end_index_;
     std::optional<TimeseriesDescriptor> timeseries_descriptor_;
     std::optional<SegmentInMemory> index_segment_;
-    std::optional<SegmentInMemory> column_stats_segment_;
+    std::optional<storage::KeySegmentPair> column_stats_segment_;
```
alexowens90: I can't tell why the key is needed

alexowens90 (approved):

PR #3076, opened 2026-05-06
Fix flaky os exit and flaky test duplication across Windows/Linux

G-D-Petrov:
#### Reference Issues/PRs
Fixes #3075 

### Problem

The failure notification workflow creates duplicate GitHub issues for the same test when it fails on both Windows and Linux runners. Windows CI produces backslash paths (`tests\stress\arcticdb\...`) while Linux produces forward slashes (`tests/stress/arcticdb/...`). These are treated as different test IDs, creating separate issues (e.g. #3070 and #3074 for `test_os_exit_exits_within_timeout`).

Additionally, `_sanitise_search_query()` in `track_ci_issues.py` did not strip backslashes, so GitHub Search failed to find existing issues with backslash paths (backslash acts as an escape character in GitHub's search syntax).

### Fix

1. Added `normalise_path_separators()` in `parse_ci_failures.py` — converts `\` to `/` in parsed test IDs so Windows and Linux produce identical IDs.
2. Added `\` to the character class in `_sanitise_search_query()` in `track_ci_issues.py` as a safety net.

---

### Problem

`test_os_exit_exits_within_timeout[False-True]` hangs on Windows CI. The child process calls `lib.read("sym")` then `os._exit(0)`, but the process never exits within the 15-second timeout.

### Root cause

On Windows, `os._exit(0)` → `_exit()` → `ExitProcess()`. `ExitProcess` first terminates all threads via `TerminateThread`, then runs `DllMain(DLL_PROCESS_DETACH)` for loaded DLLs. If a terminated CPU thread pool thread was mid-allocation (holding a CRT heap lock), the DLL detach deadlocks trying to acquire the same lock. This is the classic Windows "loader lock" deadlock.

On Linux, `_exit()` is a clean syscall — no DLL cleanup, no deadlock.

There is no mechanism to hook into `ExitProcess` before it terminates threads — `Py_AtExit`, C `atexit`, none fire during `_exit()`.

### Potential Fix

Wrap `os._exit` on Windows when arcticdb is imported (`__init__.py`). The wrapper calls `arcticdb_ext.shutdown()` before the real `os._exit()`. Since arcticdb must be imported for the library object to be unpickled in the child process, the wrapper is always in place.

`shutdown()` does two things:

1. **`stop()`** (via `shutdown_globals()`) — signals thread pools to stop accepting work. This is best-effort and non-blocking (folly's `CPUThreadPoolExecutor::stop()` does not guarantee threads have exited).
2. **`join()`** — blocks until all threads have fully exited. This is the critical step: without it, threads may still be alive when `ExitProcess` runs.

After `join()` returns, there are no thread pool threads, so `ExitProcess` has nothing to terminate and no orphaned CRT locks to deadlock on.

**Why `join()` won't hang here:** After `lib.read()` returns, all tasks are complete and threads are idle (waiting on the task queue). `stop()` wakes them up, and `join()` returns immediately. `join()` would only hang if threads were genuinely stuck (deadlocked, blocked on I/O), which is a separate bug — not the normal exit scenario.

**Why two shutdown paths:** `Py_AtExit` (normal exit) calls only `stop()` — lightweight, non-blocking, safe even if threads are mid-work. The `os._exit` wrapper calls `stop()` + `join()` because it needs threads fully gone before `ExitProcess`.

`shutdown_globals()` is idempotent (`std::call_once`), so it's safe to call both from the wrapper and from `Py_AtExit` during normal exit.

### DIFF
--- a/.github/scripts/parse_ci_failures.py
+++ b/.github/scripts/parse_ci_failures.py
@@ -75,6 +75,16 @@ def parse_gtest_failures(log_text: str) -> set[str]:
     return results
 
 
+def normalise_path_separators(test_id: str) -> str:
+    """Normalise Windows backslash path separators to forward slashes.
+
+    Windows CI runners produce paths like ``tests\\stress\\arcticdb\\...``
+    while Linux/macOS produce ``tests/stress/arcticdb/...``. Normalising
+    ensures the same test produces the same ID regardless of platform.
+    """
+    return test_id.replace("\\", "/")
+
+
 def strip_parametrize(test_id: str) -> str:
     """Strip pytest parametrize suffix ``[...]`` from a test node ID.
 
@@ -98,7 +108,8 @@ def parse_pytest_failures(log_text: str) -> set[str]:
     parametrizations of the same test are grouped together.
     """
     pattern = r"(?:FAILED|ERROR)\s+(\S+::\S+)"
-    return {strip_parametrize(m.group(1)) for m in re.finditer(pattern, log_text)}
+    return {normalise_path_separators(strip_parametrize(m.group(1)))
+            for m in re.finditer(pattern, log_text)}
 
 
 def filter_infra_steps(all_steps: list[str]) -> list[str]:
--- a/.github/scripts/parse_ci_failures.py
+++ b/.github/scripts/parse_ci_failures.py
@@ -113,15 +113,17 @@ def parse_pytest_failures(log_text: str) -> set[str]:
 
 
 def filter_infra_steps(all_steps: list[str]) -> list[str]:
-    """Filter out test-runner steps, keeping only infrastructure steps.
+    """Filter out test-runner and benchmark steps, keeping only infrastructure steps.
 
-    Test-runner steps (e.g. "Run test", "Run pytest") are always removed:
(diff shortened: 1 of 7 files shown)


poodlewars:
The test that we're looking at is,

```
@pytest.mark.parametrize("io_threads_spawned_in_child", [True, False])
@pytest.mark.parametrize("cpu_threads_spawned_in_child", [True, False])
def test_os_exit_exits_within_timeout(
    lmdb_storage, lib_name, io_threads_spawned_in_child, cpu_threads_spawned_in_child
):
    lib = lmdb_storage.create_arctic().create_library(lib_name)
    df = pd.DataFrame()
    lib.write("sym", df)
    lib.snapshot("snap")
    proc = Process(target=killed_worker, args=(lib, io_threads_spawned_in_child, cpu_threads_spawned_in_child))
    proc.start()
    proc.join(timeout=15)

    if proc.is_alive():
        proc.terminate()
        pytest.fail("os._exit did not exit within 15 seconds")

    assert proc.exitcode == 0
```

Can you explain the relationship between this and `os._exit`?

ognyanstoimenov (approved):

poodlewars (approved):

PR #3080, opened 2026-05-06
Performance 11460919917: defrag profiling and optimisation

alexowens90:
#### Reference Issues/PRs
[11460919917](https://man312219.monday.com/boards/7852509418/pulses/11460919917)

Adds micro and ASV benchmarks for `compact_data`, and improves the performance in several ways.

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -430,6 +430,8 @@ class AsyncStore : public Store {
         ARCTICDB_RUNTIME_DEBUG(log::version(), "Reading {} keys", ranges_and_keys.size());
         // Window the reads by default (but with a kill switch) for reasons detailed in the PR description:
         // https://github.com/man-group/ArcticDB/pull/3086
+        // x2 IO threadpool size is a balance to keep the IO workers busy, without reading data in faster than we can
+        // process it
         // TODO 11961775873: remove this kill switch
         if (window_processing_pipeline_reads_) {
             return folly::window(
--- a/cpp/arcticdb/column_store/column_reslicer.cpp
+++ b/cpp/arcticdb/column_store/column_reslicer.cpp
@@ -15,7 +15,10 @@
 
 namespace arcticdb {
 
-ColumnReslicer::ColumnReslicer(const ReslicingInfo& reslicing_info) : reslicing_info_(reslicing_info) {}
+ColumnReslicer::ColumnReslicer(const size_t num_input_slices, const ReslicingInfo& reslicing_info) :
+    reslicing_info_(reslicing_info) {
+    cols_or_row_counts_.reserve(num_input_slices);
+}
 
 void ColumnReslicer::push_back(std::shared_ptr<Column> column, std::shared_ptr<StringPool> string_pool) {
     if (column->is_sparse()) {
@@ -174,6 +177,9 @@ std::vector<Column> ColumnReslicer::reslice_by_iteration(std::vector<StringPool>
         for (auto& col_or_row_count : cols_or_row_counts_) {
             if (auto* col_with_strings = std::get_if<ColumnWithStrings>(&col_or_row_count)) {
                 details::visit_type(col_with_strings->column_->type().data_type(), [&](auto input_tag) {
+                    // Maps offsets in the input string pool to offsets in the output string pool
+                    // Provides a small speed boost for low cardinality string columns
+                    ankerl::unordered_dense::map<StringPool::offset_t, StringPool::offset_t> offsets_map;
                     using input_type_info = ScalarTypeInfo<decltype(input_tag)>;
                     auto input_data = col_with_strings->column_->data();
                     auto input_end_it = input_data.cend<typename input_type_info::TDT>();
@@ -183,6 +189,7 @@ std::vector<Column> ColumnReslicer::reslice_by_iteration(std::vector<StringPool>
                             ++output_col;
                             ++string_pool;
                             advance_output_col();
+                            offsets_map.clear();
                             if (output_col != output_columns.end()) {
                                 output_data = output_col->data();
                                 output_it = output_data.begin<typename output_type_info::TDT>();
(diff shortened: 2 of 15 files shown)


cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -15,7 +15,10 @@
 
 namespace arcticdb {
 
-ColumnReslicer::ColumnReslicer(const ReslicingInfo& reslicing_info) : reslicing_info_(reslicing_info) {}
+ColumnReslicer::ColumnReslicer(const size_t num_input_slices, const ReslicingInfo& reslicing_info) :
+    reslicing_info_(reslicing_info) {
+    cols_or_row_counts_.reserve(num_input_slices);
```
alexowens90, line 20: Final size of `cols_or_row_counts_` is known a priori, so for wide data with many segments this eliminates a lot of reallocs

cpp/arcticdb/async/async_store.hpp
```diff
@@ -430,6 +430,8 @@ class AsyncStore : public Store {
         ARCTICDB_RUNTIME_DEBUG(log::version(), "Reading {} keys", ranges_and_keys.size());
         // Window the reads by default (but with a kill switch) for reasons detailed in the PR description:
         // https://github.com/man-group/ArcticDB/pull/3086
+        // x2 IO threadpool size is a balance to keep the IO workers busy, without reading data in faster than we can
+        // process it
```
alexowens90, line 434: Addressing https://github.com/man-group/ArcticDB/pull/3086/changes#r3216695683

cpp/arcticdb/processing/clause.cpp
alexowens90, line 1: Changes in this and other `clause*` files are not performance related. They fix up the previously incorrect `clause_info_` fields of both the compact data and write clauses, and makes the write clause more flexible, as it can accept any input structure now

python/benchmarks/compact_data.py
```diff
@@ -0,0 +1,166 @@
...
+from arcticdb.util.test import random_strings_of_length
+
+rng = np.random.default_rng()
+
+
+class CompactDataBase:
+    def __init__(self):
+        self.logger = get_logger()
+        self.SYM = "sym"
+        self.repeat = 5
```
IvoDD: I think a time based repeat might be less flaky (given that we below vary the column size of the dataframes a lot). See [docs](https://asv.readthedocs.io/en/stable/benchmarks.html#timing-benchmarks) which allow `repeat = (min_repeat, max_repeat, max_time)`. 

Do you have rough timings of the benchmarks?

python/benchmarks/compact_data.py
```diff
@@ -0,0 +1,166 @@
...
+        del self.lib
+        del self.ac
+
+    def _setup(self, row_params, num_columns, column_slicing, dfs):
+        _, initial_rows_per_segment, target_rows_per_segment = row_params
+        self.ac = Arctic(self.CONNECTION_STRING)
+        self.lib = self.ac[self.LIB_NAME]
+        self.ac.modify_library_option(self.lib, ModifiableLibraryOption.ROWS_PER_SEGMENT, initial_rows_per_segment)
+        if column_slicing:
+            self.ac.modify_library_option(self.lib, ModifiableLibraryOption.COLUMNS_PER_SEGMENT, num_columns // 2)
```
IvoDD: `column_slicing` can be misleading. If `true` it results in exactly `2` column slices. If `false` it produces `num_columns // 127` slices which might be more or less.

IvoDD: Reading code below this is fine for now as static schema benchmarks don't use >100 columns. Might be surprising if we decide to extend num_columns further but ok to keep as is.

python/benchmarks/compact_data.py
```diff
@@ -0,0 +1,166 @@
...
+class CompactDataStringsStaticSchema(CompactDataBase):
+    def __init__(self):
+        super().__init__()
+        self.LIB_NAME = "compact_data_strings_static_schema"
+        self.CONNECTION_STRING = "lmdb://compact_data_strings_static_schema"
+        self.DYNAMIC_SCHEMA = False
+        self.param_names = self.base_param_names + ["num_unique_strings"]
+        self.params = [
+            [
+                (1_000_000, 10_000, 100_000),
```
IvoDD, line 153: Would be interesting to add parametrization where we need to slicing and not only joining

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -267,9 +280,7 @@ std::vector<Column> ColumnReslicer::initialise_output_columns() const {
                                 inserter = *set_bit + idx;
                             }
                         } else {
-                            for (auto local_idx = 0; local_idx < col.last_row() + 1; ++local_idx) {
-                                inserter = local_idx + idx;
-                            }
+                            bitset.set_range(idx, idx + col.last_row());
```
IvoDD, line 283: It's not imediately obvious to me that a combination of `set_range` calls and a `inserter.flush` is faster than just a single `inserter.flush` with more `inserter=value`.

Have you ran benchmarks for this? Or `set_range` source code makes it obvious?

cpp/arcticdb/column_store/column_reslicer.cpp
```diff
@@ -174,6 +177,9 @@ std::vector<Column> ColumnReslicer::reslice_by_iteration(std::vector<StringPool>
         for (auto& col_or_row_count : cols_or_row_counts_) {
             if (auto* col_with_strings = std::get_if<ColumnWithStrings>(&col_or_row_count)) {
                 details::visit_type(col_with_strings->column_->type().data_type(), [&](auto input_tag) {
+                    // Maps offsets in the input string pool to offsets in the output string pool
+                    // Provides a small speed boost for low cardinality string columns
+                    ankerl::unordered_dense::map<StringPool::offset_t, StringPool::offset_t> offsets_map;
```
IvoDD, line 182: Have you measured the memory usage for columns with no unique values?

Presumably two `int64_t`s per unique string value is much less than the strings themselves.

cpp/arcticdb/processing/clause_utils.hpp
```diff
@@ -26,7 +26,14 @@ namespace arcticdb {
...
 using SliceAndKey = pipelines::SliceAndKey;
 
-enum class ProcessingStructure { ROW_SLICE, TIME_BUCKETED, HASH_BUCKETED, ALL, MULTI_SYMBOL };
+enum class ProcessingStructure {
+    ROW_SLICE,
+    TIME_BUCKETED,
+    HASH_BUCKETED,
+    ALL,
+    MULTI_SYMBOL,
+    SINGLE_COLUMN_MULTI_ROW_SLICES
```
IvoDD: `SINGLE_COLUMN_MULTI_ROW_SLICES` sound like a single column and not a single column slice

What do you think about e.g. `ONE_COL_SLICE_MANY_ROW_SLICES`?

IvoDD (approved):

cpp/arcticdb/column_store/test/benchmark_column_reslicer.cpp
```diff
@@ -0,0 +1,284 @@
...
+        for (auto& column : columns) {
+            reslicer.push_back(column, string_pool);
+        }
+        state.ResumeTiming();
+        reslicer.reslice_columns(string_pools);
+    }
+}
+
+BENCHMARK(BM_column_reslicer_combine_dense_numeric_type_promotion<uint8_t, int8_t>)->Args({10, 10'000});
+BENCHMARK(BM_column_reslicer_combine_dense_numeric_type_promotion<uint32_t, int32_t>)->Args({10, 10'000});
```
poodlewars, line 121: Would a benchmark for promotion between different widths be interesting? Update: Oh that's exactly what the next one does

python/benchmarks/compact_data.py
```diff
@@ -0,0 +1,166 @@
...
+            self.SYM, rows_per_segment=target_rows_per_segment, prune_previous_versions=False
+        )
+
+    def _peakmem_compact_data(self, target_rows_per_segment):
+        self.lib.compact_data_experimental(
+            self.SYM, rows_per_segment=target_rows_per_segment, prune_previous_versions=False
+        )
+
+
+class CompactDataNumericStaticSchema(CompactDataBase):
```
poodlewars: Would it be interesting to also have a float column, or we think there's no difference for floats vs integer types?

poodlewars (approved):

python/benchmarks/compact_data.py
alexowens90: Static schema are very fast, about 1s each for time and peakmem for numeric, and 10s for strings.
Dynamic schema are now the slowest total runtime at 6 minutes, but most of this is setup cost, the actual compactions are fast.
I thought we had a pattern for basically writing once and then copying the underlying filesystem around on LMDB for destructive benchmarks like these, and there's some [unused import evidence](https://github.com/man-group/ArcticDB/blob/master/python/benchmarks/modification_functions.py#L12) for this, but it seems to have been ripped out.
I'll move to the min/max/time model and give each benchmark different parameters. If the dynamic schema one turns out to be flaky, I'll add back in a filesystem based copy/delete setup/teardown

python/benchmarks/compact_data.py
alexowens90: Added some comments next to the parameter definitions to make this more obvious

cpp/arcticdb/column_store/column_reslicer.cpp
alexowens90: Yeh it's faster, the `BM_initialise_output_columns_dense_dynamic_schema` benchmarks go from ~165us to ~2us

cpp/arcticdb/column_store/column_reslicer.cpp
alexowens90: Not explicitly, but yeh, for a 100k row column it would be ~1.6MB + map overhead, so nothing concerning

cpp/arcticdb/column_store/test/benchmark_column_reslicer.cpp
alexowens90: Yeh maybe it's not obvious, but `uint<N>` combined with `int<N>` will produce an output column of type `int<2N>`

python/benchmarks/compact_data.py
alexowens90: Yeh no difference, numeric static schema data is just memcopied so underlying type doesn't matter

PR #3085, opened 2026-05-06
Be consistent with NaN when filtering on NaT values

poodlewars:
Monday: 11688915824 9871694127

⚠️ Technical API Break ⚠️ 

Make changes to our query builder filtering so that it handles `NaT` as a special case, like Pandas does. We move to NaN-like semantics - `NaT OP ANYTHING` is false for any OP other than `!=`,  for which `NaT != ANYTHING` is true.

The main example of this behaviour change,

```
# col a is datetime
q["a"] > pd.NaT
```

used to return every value in "a" greater than `min(int64_t)` and now returns no values.

We also change the `isin` and `isnotin` behaviour. With NaN we have:

```
q["a"].isin([NaN])
```

matching nothing. Note that this is different to Pandas, which would return any NaN rows in "a". With this change we have:

```
# Matches nothing - used to match any NaT rows in a
q["a"].isin([NaT])
```

## Other Changes (not user-facing)

For column stats, do the same as with NaN and write `min = max = NaT` iff the entire block is `NaT`, and use this when pruning with stats. This currently has some bugs for `!=` filtering (since a NaT might be hiding in a block with non-NaT min and max) that we will fix up with a NaN count stat in due course.

Add test coverage for missing datetime columns in resample and groupby aggregations, which we should treat as NaT (this already worked fine).

If we have a missing column in a segment (with dynamic schema) then our query builder filters drop it, rather than comparing against the backfilled value. This is inconsistent with a `lib.read` and then filtering in Pandas, but is sane, so I've just added test coverage for the existing behaviour.


cpp/arcticdb/pipeline/column_stats_dispatch.cpp
```diff
@@ -80,6 +80,17 @@ StatsComparison stats_membership_comparator(const ColumnStatsValues& stats, Valu
                 auto set_min_raw = static_cast<RightType>(set_min->get<SetRawType>());
                 auto set_max_raw = static_cast<RightType>(set_max->get<SetRawType>());
 
+                if constexpr (is_time_type(StatsTag::data_type)) {
```
poodlewars, line 83: I have the NaT handling up here (rather than in the operators like NaN) because otherwise every single comparison operator would neet NaT logic (whereas for NaN most of them do the right thing out of the box).

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -24,12 +25,31 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
         using type_info = ScalarTypeInfo<decltype(col_tag)>;
         using RawType = typename type_info::RawType;
         if constexpr (!is_sequence_type(type_info::data_type)) {
+            auto is_nan = [](RawType v) {
+                if constexpr (is_floating_point_type(type_info::data_type)) {
+                    return std::isnan(v);
+                } else if constexpr (is_time_type(type_info::data_type)) {
+                    return v == NaT;
+                } else {
+                    (void)v;
```
alexowens90: nit: `[[maybe_unused]]`

alexowens90:
Doesn't need to be in this PR, but while we're changing filtering behaviour for timestamp columns, we should disallow comparisons between timestamp and other numeric type columns, as well as projections involving timestamp columns in general

cpp/arcticdb/pipeline/column_stats_dispatch.cpp
```diff
@@ -80,6 +80,17 @@ StatsComparison stats_membership_comparator(const ColumnStatsValues& stats, Valu
...
+                    const bool nat_in_set = static_cast<timestamp>(set_min_raw) == NaT;
```
alexowens90: Maybe worth a comment that this works because `NaT == std::numeric_limits<timestamp>::min()`

cpp/arcticdb/pipeline/column_stats_dispatch.cpp
```diff
@@ -80,6 +80,17 @@ StatsComparison stats_membership_comparator(const ColumnStatsValues& stats, Valu
...
+                    const bool stats_all_nat = static_cast<timestamp>(stats_max) == NaT;
+                    if (stats_all_nat) {
+                        return is_isin == nat_in_set ? StatsComparison::ALL_MATCH : StatsComparison::NONE_MATCH;
```
alexowens90:
```suggestion
                        return is_isin ? StatsComparison::ALL_MATCH : StatsComparison::NONE_MATCH;
```
`stats_all_nat => nat_in_set`

alexowens90: Are the 2nd and 3rd arguments to the ternary the right way round?

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -185,30 +185,71 @@ def test_column_stats_nat_values(lmdb_version_store, any_output_format):
...
+        index=pd.date_range("2000-01-12", periods=4),
     )
     lib.write(sym, df0)
     lib.append(sym, df1)
     lib.append(sym, df2)
     lib.append(sym, df3)
+    lib.append(sym, df4)
+    lib.append(sym, df5)
+
+    # We store NaT stats iff the whole block is NaT
```
alexowens90: IIRC for floats we store NaN in the stats if **any** of the values are NaN. Why the difference?

poodlewars: This is consistent with NaN https://github.com/man-group/ArcticDB/pull/3019 . It does mean that we definitely need a NaN count stat.

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -238,6 +239,14 @@ VariantData binary_comparator(const ColumnWithStrings& left, const ColumnWithStr
...
                         sparse_missing_value_output,
                         [&func](auto left_value, auto right_value) -> bool {
+                            if constexpr (is_time_type(left_type_info::data_type)) {
+                                if (static_cast<timestamp>(left_value) == NaT)
+                                    return std::is_same_v<std::remove_reference_t<Func>, NotEqualsOperator>;
+                            }
+                            if constexpr (is_time_type(right_type_info::data_type)) {
+                                if (static_cast<timestamp>(right_value) == NaT)
+                                    return std::is_same_v<std::remove_reference_t<Func>, NotEqualsOperator>;
+                            }
```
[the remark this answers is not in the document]
poodlewars: This is gibberish

poodlewars:
> Doesn't need to be in this PR, but while we're changing filtering behaviour for timestamp columns, we should disallow comparisons between timestamp and other numeric type columns, as well as projections involving timestamp columns in general



> Doesn't need to be in this PR, but while we're changing filtering behaviour for timestamp columns, we should disallow comparisons between timestamp and other numeric type columns, as well as projections involving timestamp columns in general

11988010482

cpp/arcticdb/pipeline/column_stats_dispatch.cpp
poodlewars: Ah I was going for consistency with Pandas here (which includes NaT values in `.isin([NaT])`, which was dumb because with `.isin([NaN])` we do *not* match Pandas and exclude NaN rows. I'll make changes so that we do the same as for NaN here.

`stats_all_nat => nat_in_set` is wrong, `stats_all_nat` means the block is nat, `nat_in_set` means there's a `NaT` in the `ValueSet`.

alexowens90 (approved):

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -24,12 +25,30 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
         using type_info = ScalarTypeInfo<decltype(col_tag)>;
         using RawType = typename type_info::RawType;
         if constexpr (!is_sequence_type(type_info::data_type)) {
+            auto is_nan = []([[maybe_unused]] RawType v) {
```
ognyanstoimenov: Nit: could use better naming as it checks for NaT as well

cpp/arcticdb/processing/operation_types.hpp
```diff
@@ -62,6 +62,15 @@ std::optional<StatsComparison> check_scalar_for_nan(T value, StatsComparison nan
     return std::nullopt;
 }
 
+inline std::optional<StatsComparison> check_time_stats_for_nat(
+        timestamp min_val, timestamp max_val, timestamp query_value, StatsComparison nat_result
+) {
+    if (min_val == NaT || max_val == NaT || query_value == NaT) {
```
ognyanstoimenov: Isn't the max_val check redundant? Worth a comment why it is kept

cpp/arcticdb/pipeline/column_stats_dispatch.hpp
```diff
@@ -163,6 +176,21 @@ StatsComparison stats_comparator(const ColumnStatsValues& stats_lhs, const Colum
                 auto lhs_max = static_cast<typename comp::left_type>(stats_lhs.max->get<LhsRawType>());
                 auto rhs_min = static_cast<typename comp::right_type>(stats_rhs.min->get<RhsRawType>());
                 auto rhs_max = static_cast<typename comp::right_type>(stats_rhs.max->get<RhsRawType>());
+                if constexpr (is_time_type(LhsTag::data_type) || is_time_type(RhsTag::data_type)) {
```
ognyanstoimenov, line 176: Couldn't find a test for this overload?

ognyanstoimenov (approved):

PR #3086, opened 2026-05-06
Processing pipeline perf improvements

alexowens90:
While profiling and optimising the `compact_data` API, I noticed an unusual pattern. With 1 CPU/IO thread, peak memory usage was stable and low, but with 2 or more, it jumped massively. After [much digging](https://chat-man.slack.com/archives/C02NEG3V38X/p1778148513774729) it turned out to be a vicious cycle problem.
To avoid context switching and cache-thrashing, segment decodes happen in the IO thread that read the segment. Segments that are required to be together for the first clause's `process` call are then `folly::collect`ed. The important point is that they are collected in the IO executor as well, and the following lambda which does the processing is `thenValueInline`d. In practice, this means that whichever IO thread finished reading+decoding it's segment last also then executes the processing lambda.
Prior to this, `batch_read_uncompressed` scheduled all of the IO up front. Work scheduled to the IO threadpool is just round-robinned around the available threads, and so every IO thread will receive an equal number of segments to read. However, whichever thread had to execute the processing lambda, which was already the last to finish the read+decode will finish later than the others. This then compounds, so that for the next `collect`, it will almost certainly be the same IO thread that finishes last again, as it started reading later than the others.
The result is that in an IO threadpool of size `n`, `n-1` of the threads just do read+decode tasks, and 1 does ALL of the processing (and writing in the `compact_data` case), so most of the wall clock time is the tail latency of this 1 thread.
The solution is two changes:

1. Switch to the CPU executor for the processing lambda. This will then mean there isn't a single IO thread doing all of the work.
2. `folly::window` the initial reads. This is now required, because the IO of writing back to storage for clauses like compact and merge used to be (basically) inlined when it was scheduled from an IO thread. Now we want these writes to be interleaved with the reads, otherwise they will be at the back of the IO thread's work queues, and so the flush to storage won't happen until the end.

There is one case where this could have a negative performance impact, and that is when the clause processing function needs only a single segment from storage (e.g. a filter with non column sliced data). Previously this would have been executed inline in the thread that read+decoded the data for maximal cache efficiency. It is possible there will now be a delay between the decoding and the processing, leading to the segment being evicted from the CPU cache. We will need to see if this is a problem in practice. We could use the IO executor after single-segment collects, and the CPU executor for multi-segment collects, but this seems like a recipe for deadlocks and debugging nightmares.

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -427,14 +427,16 @@ class AsyncStore : public Store {
             std::shared_ptr<std::unordered_set<std::string>> columns_to_decode
     ) override {
         ARCTICDB_RUNTIME_DEBUG(log::version(), "Reading {} keys", ranges_and_keys.size());
-        std::vector<folly::Future<pipelines::SegmentAndSlice>> output;
-        for (auto&& ranges_and_key : ranges_and_keys) {
-            const auto key = ranges_and_key.key_;
-            output.emplace_back(read_and_continue(
-                    key, library_, storage::ReadKeyOpts{}, DecodeSliceTask{std::move(ranges_and_key), columns_to_decode}
-            ));
-        }
-        return output;
+        return folly::window(
+                std::move(ranges_and_keys),
+                [this, columns_to_decode](pipelines::RangesAndKey&& ranges_and_key) {
+                    const auto key = ranges_and_key.key_;
+                    return read_and_continue(
+                            key, library_, storage::ReadKeyOpts{}, DecodeSliceTask{std::move(ranges_and_key), columns_to_decode}
+                    );
+                },
+                2 * async::TaskScheduler::instance()->io_thread_count()
+                );
     }
 
     std::vector<folly::Future<bool>> batch_key_exists(const std::vector<entity::VariantKey>& keys) override {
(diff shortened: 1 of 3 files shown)


poodlewars (approved):

cpp/arcticdb/async/async_store.hpp
```diff
@@ -427,14 +428,36 @@ class AsyncStore : public Store {
...
+                    [this, columns_to_decode](pipelines::RangesAndKey&& ranges_and_key) {
+                        const auto key = ranges_and_key.key_;
+                        return read_and_continue(
+                                key,
+                                library_,
+                                storage::ReadKeyOpts{},
+                                DecodeSliceTask{std::move(ranges_and_key), columns_to_decode}
+                        );
+                    },
+                    2 * async::TaskScheduler::instance()->io_thread_count()
```
IvoDD, line 446: `2x` is probably a good middle ground between: - keep all io thread queues (relatively) full
- allow scheduling io threads to happen sooner

And if we ever decide to switch the IO pool to another `CPUThreadPool` with a different max queue we should probably consider making this `1x` since it will be enough to keep all workers busy.

Maybe an inline comment would be good.

IvoDD (approved):

cpp/arcticdb/async/async_store.hpp
alexowens90: Fair, will address in the follow up PR with `compact_data` specific perf improvements to avoid holding this one up on CI

PR #3090, opened 2026-05-08
[Bugfix] Fix wrong column naming after append and update when the DataFrame has non-string column name

vasil-pashov:
#### Reference Issues/PRs
Monday 10029194063

When we perform append or update we write back the new normalization metadata to disk. For most cases this is OK because we have strict requirements for matching indexes. However in case the DataFrame has non-string column names the normalization metadata contains additional metadata needed to denormalize these column names.

In order for this to work we need to merge the map of column_name -> column_name_metadata instead of writing back the new map. This is implemented by the function `accumulate_norm_metadata_column_names`. It happens in two places `accumulate_norm_metadata` which handles generation of normalization metadata for concat (concat is also affected by this bug) and in `get_merged_tsd` which handles appends and updates. 

The second issue is in `_normalization.py` when we set the columns. We must explicitly set the dtype of the pd.Index containing  the column names, otherwise Pandas will infer a dtype and might change the name of the columns. E.g. if the columns are [1, None] pandas will infer dtype=float64 renaming them to [1.0, np.nan]

This PR is split into two commits.
1. The first only moves code around and adds unit tests. No functional changes.
2. The second implements the fixes described above.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/pipeline/index_utils.cpp
```diff
@@ -7,6 +7,9 @@
  */
 
 #include <arcticdb/pipeline/index_utils.hpp>
+
+#include "python/normalization_utils.hpp"
```
alexowens90: nit: angle brackets

cpp/arcticdb/python/normalization_utils.cpp
```diff
@@ -228,6 +226,40 @@ void fix_normalization_or_throw(
...
+                    "settings for columns named the same way.",
+                    col_name
+            );
+        }
+    }
+    accumulated_col_names->insert(new_entry.common().col_names().begin(), new_entry.common().col_names().end());
+}
+
+void accumulate_norm_metadata_column_names(NormalizationMetadata& accumulated, const NormalizationMetadata& new_entry) {
+    if (accumulated.has_df()) {
```
alexowens90, line 256: Is it checked upstream that `new_entry.has_df()` in this case?

alexowens90 (approved):

cpp/arcticdb/pipeline/index_utils.cpp
```diff
@@ -74,6 +74,34 @@ std::pair<index::IndexSegmentReader, std::vector<SliceAndKey>> read_index_to_vec
...
 
+proto::descriptors::NormalizationMetadata merge_normalization_metadata(
+        const TimeseriesDescriptor& existing_tsd, const InputFrame& new_frame
+) {
+    if (existing_tsd.index().type() == IndexDescriptor::Type::EMPTY) {
+        return new_frame.norm_meta;
+    }
+
+    proto::descriptors::NormalizationMetadata result = existing_tsd.normalization();
+    accumulate_norm_metadata_column_names(result, new_frame.norm_meta);
```
IvoDD, line 139: This only merges the column names metadata. There are other metadata items to be considered:
- index timezones
- different multiindex names (presumably checked upstream and raises?)

Can you add some tests that cover the above to make sure we're doing the correct thing?

I'm ok with this PR handling just column names but would be good to know the

IvoDD, line 139: Also we now start using the `existing_tsd.normalization()` as baseline to add on top.

Doesn't this mean that we will change e.g. timezone behavior and preserve the existing timezone instead of using the new timezone (which is what we were doing before)?

cpp/arcticdb/python/normalization_utils.cpp
```diff
@@ -228,6 +226,40 @@ void fix_normalization_or_throw(
...
+) {
+    return lhs.is_empty() == rhs.is_empty() && lhs.is_int() == rhs.is_int() && lhs.is_none() == rhs.is_none() &&
+           lhs.original_name() == rhs.original_name();
+}
+
+template<typename ColumnNameMapParent>
+requires util::any_of<
+        ColumnNameMapParent, NormalizationMetadata_NormalisedTimeSeries, NormalizationMetadata_PandasDataFrame>
+void accumulate_norm_metadata_column_names(ColumnNameMapParent& accumulated, const ColumnNameMapParent& new_entry) {
+    accumulated.set_has_synthetic_columns(accumulated.has_synthetic_columns() && new_entry.has_synthetic_columns());
```
IvoDD, line 240: I think `has_synthetic_index` has always been broken and I'm not sure what the best behavior is here. I will open a separate ticket for this.

`<rant>`
I was looking into what `has_synthetic_columns` means and I think it is pretty broken.

It treats every ranged index the same [here](https://github.com/man-group/ArcticDB/blob/36122bcee7fdf72aa718fd20bf383fed3eb754da/python/arcticdb/version_store/_normalization.py#L1309-L1310) and result is always step=1, start=0 [here](https://github.com/man-group/ArcticDB/blob/36122bcee7fdf72aa718fd20bf383fed3eb754da/python/arcticdb/version_store/_normalization.py#L503-L504).

I'm wondering whether we should always `set_has_synthetic_columns(false)`? Otherwise we end up with some pretty messed up stuff like:
```
df_1 = pd.DataFrame(np.random.rand(5, 4), columns=pd.RangeIndex(4))
df_2 = pd.DataFrame(np.random.rand(3, 4), columns=pd.RangeIndex(0, 8, 2))
lib.write(sym, df_1)
lib.append(sym, df_2)
lib.read(sym).data
# Columns are 0, 1, 2, 3, 4, 5
```
`</rant>`

cpp/arcticdb/python/normalization_utils.cpp
vasil-pashov: Yes, one check in `fix_normalization_or_throw` and another (implicit) here https://github.com/man-group/ArcticDB/blob/08182e0118c9ecbd3587ae817ae33e9345d2fb7d/cpp/arcticdb/processing/clause_utils.cpp#L353

cpp/arcticdb/version/schema_checks.cpp
```diff
@@ -29,13 +19,56 @@ IndexDescriptor::Type get_common_index_type(const IndexDescriptor::Type& left, c
...
+        NormalizationOperation operation, const pipelines::index::IndexSegmentReader& existing_isr,
+        const pipelines::InputFrame& frame, bool empty_types
 ) {
-    const IndexDescriptor::Type old_idx_kind = old_descriptor.index().type();
+    const IndexDescriptor::Type old_idx_kind = existing_isr.tsd().as_stream_descriptor().index().type();
     const IndexDescriptor::Type new_idx_kind = frame.desc().index().type();
+
+    // In case of multiindex we require the multiindex fields in the input to match the mutliindex fields on disk even
+    // when dynamic schema is used. This is because it's too hard to keep the normalization metadata in sync.
+    if (existing_isr.tsd().normalization().has_df()) {
```
alexowens90: Series can also have a multiindex

alexowens90 (changes requested):

cpp/arcticdb/pipeline/index_utils.cpp
```diff
@@ -96,7 +96,30 @@ proto::descriptors::NormalizationMetadata merge_normalization_metadata(
         if (result.has_series()) {
             *result.mutable_series()->mutable_common()->mutable_index() = new_frame.norm_meta.series().common().index();
         } else if (result.has_df()) {
-            *result.mutable_df()->mutable_common()->mutable_index() = new_frame.norm_meta.df().common().index();
+            if (result.df().common().has_index()) {
+                *result.mutable_df()->mutable_common()->mutable_index() = new_frame.norm_meta.df().common().index();
+            } else if (result.df().common().has_multi_index()) {
+                *result.mutable_df()->mutable_common()->mutable_multi_index() =
```
IvoDD: Similar to [Alex's comment](https://github.com/man-group/ArcticDB/pull/3090#discussion_r3257515114) here we should also treat series and df the same with respect to multiindex.

cpp/arcticdb/version/schema_checks.cpp
```diff
@@ -18,6 +18,84 @@ IndexDescriptor::Type get_common_index_type(const IndexDescriptor::Type& left, c
     return IndexDescriptor::Type::UNKNOWN;
 }
 
+/// Checks if the two multiindex are compatible for append/update. For that to be true the filed name, count and order
```
IvoDD: Nit: Typo `filed name`

cpp/arcticdb/version/schema_checks.cpp
```diff
@@ -18,6 +18,84 @@ IndexDescriptor::Type get_common_index_type(const IndexDescriptor::Type& left, c
...
+template<typename CommonNormalization>
+requires util::any_of<
+        CommonNormalization, proto::descriptors::NormalizationMetadata_Pandas,
+        proto::descriptors::NormalizationMetadata_NormalisedTimeSeries>
+void check_multiindex_matches(
+        const CommonNormalization& existing_common, const StreamDescriptor& existing_stream_descriptor,
+        const CommonNormalization& new_common, const StreamDescriptor& new_stream_descriptor
+) {
+    normalization::check<ErrorCode::E_INCOMPATIBLE_INDEX>(
+            !(existing_common.has_multi_index() ^ new_common.has_multi_index()),
```
IvoDD: Nit: I find `==` to be more readable than `!^`

IvoDD (approved):

alexowens90 (approved):

PR #3091, opened 2026-05-08
Binary search utils

IvoDD:
#### Reference Issues/PRs
Monday ref: 11960725194

Add `lower_bound` and `upper_bound` methods that work on `ColumnDataIterators`.

The change includes:
- Efficient binary search implementation for all column layouts (presized, regular, irregular)
- Implementation of [exponential search](https://en.wikipedia.org/wiki/Exponential_search) methods. To be used in follow up resampling PR.
- Makes `ColumnDataIterator` improvements to allow efficient use with search methods:
  - `block_pos_` now refers to the "current" block instead of the "next" block
  - new constructor from `block_idx` and `in_block_offset` for dense iterators
  - new `ColumnData::iterator_at(position)`  for dense iterators
- Replaces the less efficient usages of `search_sorted` with the new `lower_bound` or `upper_bound` methods.

The change is split in 3 commits:
1. Unoptimized implementation of search methods. Constructs iterators at each step which turns out to be expensive.
2. Optimize implementation of search methods to bypass iterators and access memory directly. Can result in up to 2x improvements in some cases.
3. Replace usages of `search_sorted` with new more efficient methods.  

I recommend reviewing commits 1 and 2 together unless you want to first get a more code concise feel of the algorithm implementations. Commit 3 can be reviewed separately.

#### Any other comments?
Benchmark results after the first 2 commits:

  100k `NANOSECONDS_UTC64` rows. `--benchmark_min_time=2s` on a 128-core 3.67 GHz host.                                                                                
                                                                                                                                                  
  | Layout | search_sorted (old) | lower_bound | exponential_lb | exponential_lb (answer in first 100) |                                                                
  |---|---:|---:|---:|---:|                                       
  | single block | 487 ns | **404 ns** (1.21×) | 441 ns (1.10×) | **351 ns** (1.39×) |                                                            
  | regular blocks | 932 ns | **439 ns** (2.12×) | 498 ns (1.87×) | **362 ns** (2.57×) |                                                          
  | irregular blocks (1000) | 689 ns | **435 ns** (1.58×) | 487 ns (1.41×) | **355 ns** (1.94×) |                                                 
  | irregular blocks (1) | 1323 ns | **582 ns** (2.27×) | 668 ns (1.98×) | **447 ns** (2.96×) |

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

<!--
Thanks for contributing a Pull Request to ArcticDB! Please ensure you have taken a look at:
 - ArcticDB's Code of Conduct: https://github.com/man-group/ArcticDB/blob/master/CODE_OF_CONDUCT.md
 - ArcticDB's Contribution Licensing: https://github.com/man-group/ArcticDB/blob/master/docs/mkdocs/docs/technical/contributing.md#contribution-licensing
-->


cpp/arcticdb/column_store/column_data.hpp
```diff
@@ -205,13 +226,31 @@ struct ColumnData {
...
+        // Minimal accessors used by the search algorithms in column_algorithms.hpp.
+        [[nodiscard]] const ColumnData* parent() const { return parent_; }
+        [[nodiscard]] const std::optional<TypedBlockData<TDT>>& current_block() const { return opt_block_; }
+        [[nodiscard]] const RawType* current_ptr() const { return data_.ptr_; }
+        // Index of the block this iterator currently points into. For end iterators this is num_blocks.
+        [[nodiscard]] size_t current_block_index() const { return block_pos_; }
+        // TODO: Might be better to refactor storing the state as block_pos_ + in_block_offset_
+        // Maybe not because current impl has a very efficient == which just compares a single pair of pointers
+        // Although that has downsides as well (how do you iterate a column with external blocks pointing to the same
+        // memory?!?)
```
alexowens90: Would be interesting to know if changing the state to include the block position has a big performance negative, as you say, it would currently do something garbage for Arrow data backed by the same underlying buffer.

cpp/arcticdb/column_store/column_algorithms.hpp
```diff
@@ -327,4 +327,232 @@ static void transform(
...
+}
+
+} // namespace search_detail
+
+// Returns an iterator to the first element in [begin, end) that is not less than `value`.
+// Complexity is `O(log(std::distance(begin, end)))`.
+// Mirrors std::lower_bound semantics.
+template<typename TDT, IteratorType IT, IteratorDensity ID>
+requires SortedSearchInputs<TDT, ID>
+ColumnData::ColumnDataIterator<TDT, IT, ID, true> lower_bound(
```
alexowens90, line 503: Is there a reason to disallow returning non-const iterators?

cpp/arcticdb/column_store/column_algorithms.hpp
```diff
@@ -327,4 +327,232 @@ static void transform(
...
+        const ColumnData::ColumnDataIterator<TDT, IT, ID, true>& begin,
+        const ColumnData::ColumnDataIterator<TDT, IT, ID, true>& end, typename TDT::DataTypeTag::raw_type value
+) {
+    using RawType = typename TDT::DataTypeTag::raw_type;
+    return search_detail::bound_search<TDT, IT, ID>(
+            begin,
+            end,
+            value,
+            [](RawType probe, RawType v) { return probe < v; },
```
alexowens90, line 512: If we really wanted to mirror the STL API we could lift this lambda up to be an argument (with this as the default), although if we don't have a use case for anything more complicated yet then I'm fine with leaving it here

cpp/arcticdb/column_store/column_algorithms.hpp
```diff
@@ -327,4 +327,232 @@ static void transform(
...
+ColumnData::ColumnDataIterator<TDT, IT, ID, true> upper_bound(
+        const ColumnData::ColumnDataIterator<TDT, IT, ID, true>& begin,
+        const ColumnData::ColumnDataIterator<TDT, IT, ID, true>& end, typename TDT::DataTypeTag::raw_type value
+) {
+    using RawType = typename TDT::DataTypeTag::raw_type;
+    return search_detail::bound_search<TDT, IT, ID>(
+            begin,
+            end,
+            value,
+            [](RawType probe, RawType v) { return !(v < probe); },
```
alexowens90: I'm sure the compiler will optimise it out, but why not `v >= probe`?

cpp/arcticdb/column_store/column_algorithms.hpp
```diff
@@ -327,4 +327,232 @@ static void transform(
...
+    static_assert(ID == IteratorDensity::DENSE, "Sorted search currently supports DENSE only");
+    static_assert(TDT::dimension() == Dimension::Dim0, "Sorted search supports Dim0 only");
+    util::check(begin.parent() == end.parent(), "bound_search: begin and end have different parents");
+
+    if (begin == end) {
+        // This to covers the case of empty column or empty range
+        return begin;
+    }
+
+    const ColumnData* data = begin.parent();
```
alexowens90, line 373: It feels wrong that we're having to extract the undlerying `ColumnData` from the provided iterators, although I get that it's necessary to mimic the standard libraries APIs

cpp/arcticdb/column_store/test/test_column.cpp
```diff
@@ -402,6 +406,187 @@ TEST(ColumnStats, DoubleColumn) {
...
+            "all-1-blocks",
+            from,
+            to
+    );
+}
+
+} // namespace
+
+TEST(ColumnSearch, BasicRegular) {
+    std::vector<int64_t> values{0, 5, 10, 15, 20, 20, 25, 25, 30, 35};
```
alexowens90, line 583: This doesn't stress most features of regular sized blocks, need at least 8k elements really

alexowens90 (approved):

cpp/arcticdb/column_store/column_data.hpp
IvoDD: I added two new commits. One with benchmarks and one with addressing the TODO. It does not result in any regressions. In fact it slightly speeds up block jumping:

  | Layout                | Baseline (ptr) mean | cv    | Current (offset) mean | cv    | Δ      | Significant?               |
  |-----------------------|---------------------|-------|------------------------|-------|--------|----------------------------|
  | single_block          | 27,505 ns           | 0.47% | 27,316 ns              | 0.43% | -0.7%  | borderline (within noise)  |
  | regular_blocks        | 30,347 ns           | 0.70% | 29,145 ns              | 0.27% | -4.0%  | yes                        |
  | irregular_blocks_1000 | 28,819 ns           | 0.25% | 28,333 ns              | 0.27% | -1.7%  | yes                        |
  | irregular_blocks_1    | 481,096 ns          | 0.15% | 410,798 ns             | 3.52% | -14.6% | yes                        |

cpp/arcticdb/column_store/column_algorithms.hpp
IvoDD: we don't need non-const iterators and I didn't want to introduce too many template arguments

cpp/arcticdb/column_store/column_algorithms.hpp
IvoDD: agreed, not used now so I'll avoid the extra template arguments, can easily be added later on

cpp/arcticdb/column_store/test/test_column.cpp
```diff
@@ -275,6 +279,71 @@ TEST(ColumnData, Iterator) {
...
+    // All-empty column: a single zero-size external block. begin must compare equal to end.
+    Column empty_col(static_cast<TypeDescriptor>(TDT{}), 0, AllocationType::DYNAMIC, Sparsity::PERMITTED);
+    empty_col.set_external_block(0, static_cast<int64_t*>(nullptr), 0);
+    ASSERT_EQ(empty_col.buffer().num_blocks(), 1u);
+    ASSERT_EQ(empty_col.buffer().blocks()[0]->logical_size(), 0u);
+    auto empty_data = empty_col.data();
+    auto begin = empty_data.cbegin<TDT, IteratorType::REGULAR, IteratorDensity::SPARSE>();
+    auto end = empty_data.cend<TDT, IteratorType::REGULAR, IteratorDensity::SPARSE>();
+    EXPECT_EQ(begin, end);
+}
```
vasil-pashov, line 315: nit: these seem like two separate tests

cpp/arcticdb/column_store/test/test_column.cpp
```diff
@@ -402,6 +471,195 @@ TEST(ColumnStats, DoubleColumn) {
...
+    for (auto v : probes) {
+        auto std_lb = std::lower_bound(std_begin, std_end, v);
+        auto std_ub = std::upper_bound(std_begin, std_end, v);
+        auto lb_it = lower_bound<SearchTDT, IteratorType::ENUMERATED, IteratorDensity::DENSE>(begin, end, v);
+        auto ub_it = upper_bound<SearchTDT, IteratorType::ENUMERATED, IteratorDensity::DENSE>(begin, end, v);
+        auto ex_lb_it =
+                exponential_lower_bound<SearchTDT, IteratorType::ENUMERATED, IteratorDensity::DENSE>(begin, end, v);
+        auto ex_ub_it =
+                exponential_upper_bound<SearchTDT, IteratorType::ENUMERATED, IteratorDensity::DENSE>(begin, end, v);
+        ASSERT_EQ(res_idx(lb_it), std_lb - values.begin()) << label << " lower_bound(" << v << ")";
```
vasil-pashov, line 556: This is cool! I didn't know we can append to the error messages like so.

cpp/arcticdb/column_store/test/test_column.cpp
```diff
@@ -402,6 +471,195 @@ TEST(ColumnStats, DoubleColumn) {
...
+
+} // namespace
+
+TEST(ColumnSearch, BasicRegular) {
+    std::vector<int64_t> values{0, 5, 10, 15, 20, 20, 25, 25, 30, 35};
+    Column col = make_regular_blocks(values);
+    auto column_data = col.data();
+    auto begin = column_data.cbegin<SearchTDT, IteratorType::REGULAR, IteratorDensity::DENSE>();
+    auto end = column_data.cend<SearchTDT, IteratorType::REGULAR, IteratorDensity::DENSE>();
+    auto lb = lower_bound<SearchTDT, IteratorType::REGULAR, IteratorDensity::DENSE>(begin, end, int64_t{20});
```
vasil-pashov, line 588: Do we have guarantees that this lower_bound will find the 20 at index 4 and not the one on index 5? We can assert also lb == begin + 4. Same thing for upper_bound

cpp/arcticdb/column_store/test/rapidcheck_column.cpp
vasil-pashov, line 146: I think it's worth having similar rapid checks for regular sized blocks and for irregular sized blocks. Maybe for sparse data too.

cpp/arcticdb/column_store/test/benchmark_column.cpp
```diff
@@ -7,59 +7,202 @@
...
+    size_t remaining = data.size();
+    while (remaining > 0) {
+        const size_t alloc = std::min(remaining, block_size);
+        col.allocate_data(alloc * sizeof(timestamp));
+        col.advance_data(alloc * sizeof(timestamp));
+        remaining -= alloc;
+    }
+    populate(col, data);
+    return col;
+}
```
vasil-pashov, line 85: It loos like there's some code duplicatoin here and in test_column.cpp. Can we extract this in a common place

cpp/arcticdb/column_store/column_algorithms.hpp
```diff
@@ -327,4 +327,321 @@ static void transform(
...
+
+// Iterator-pair binary search.
+// `is_before_answer(probe, value)` returns true while a probe is strictly before the answer (so we move begin past it).
+// For lower_bound that is `probe < value`; for upper_bound it is `probe <= value`.
+// `within_block_bisect` is std::lower_bound or std::upper_bound run on the contiguous block memory.
+template<typename TDT, IteratorType IT, IteratorDensity ID, typename IsBeforeAnswer, typename WithinBlockBisect>
+ColumnData::ColumnDataIterator<TDT, IT, ID, true> bound_search(
+        const ColumnData::ColumnDataIterator<TDT, IT, ID, true>& begin,
+        const ColumnData::ColumnDataIterator<TDT, IT, ID, true>& end, typename TDT::DataTypeTag::raw_type value,
+        IsBeforeAnswer is_before, WithinBlockBisect bisect
```
vasil-pashov, line 361: Use forwarding reference for `is_before` and `bisect`

nit: I think it's useful to have a requires block `is_before` and `bisect`. Is before can be std::predicate<TDT::DataTypeTag::raw_type value, TDT::DataTypeTag::raw_type value>, bisect can use std::invocable.

nit: I'd also move the static asserts in the requires block

same applies for the gallop function below

cpp/arcticdb/column_store/column_algorithms.hpp
```diff
@@ -327,4 +327,321 @@ static void transform(
...
+
+    if (begin == end) {
+        // This to covers the case of empty column or empty range
+        return begin;
+    }
+
+    const ColumnData* data = begin.parent();
+    const auto& blocks = data->buffer().blocks();
+    auto block_data_at = [&blocks](size_t idx) { return reinterpret_cast<const RawType*>(blocks[idx]->data()); };
+    auto block_row_count_at = [&blocks](size_t idx) { return blocks[idx]->logical_size() / sizeof(RawType); };
```
vasil-pashov, line 376: These two seem to be repeated in the gallop_search below. Can we make them methods of the chunked buffer or free functions in this file?

cpp/arcticdb/column_store/column_algorithms.hpp
```diff
@@ -327,4 +327,321 @@ static void transform(
...
+            return end;
+        }
+        return ColumnData::ColumnDataIterator<TDT, IT, ID, true>(data, block, offset);
+    };
+
+    // Probe within the first block at first_offset + 2**n
+    // We iterate until `first_offset+step < up_to - 1` because we'll later explicitly probe at
+    // the last element of the first block
+    const size_t up_to = end_block_idx > first_block_idx ? first_block_row_count : end_in_block_offset;
+    size_t step = 1;
```
vasil-pashov, line 466: nit: step is not used outside the loop, so is in the loop below

Does it make sense to check if the last element of the first block is > value? Then we can directly skip to checking last elements of each block.

Reading the above I came up with another idea. Right now we perform exponential search withing the values of the first block and after that we perform exponential search on block level looking at the last elements. I think we can merge those roughly like so

```
for block_idx = first_block_idx; block_idx + step < end_block_idx; step = step << 1:
    if !probe last in block:
       for step = 1; step < current_block_size; step = step << 1;
         if !probe in block at step:
           return iterator
```
Does it make sense?

cpp/arcticdb/column_store/test/test_column.cpp
vasil-pashov, line 1: It would be nice to add some tests where there are multiple segments having the same value to make sure the block based binary search works (it looks correct but this way it'd be more sustainable to changes)

vasil-pashov (approved):

cpp/arcticdb/column_store/column_algorithms.hpp
IvoDD: Discussed offline. Not much point in galloping within a block after the first one. At that point regular binary search is better suited.

Will still address the nit.

IvoDD:
I agree with most of the comments. To save some time I will merge and address them in a follow up PR

PR #3097, opened 2026-05-11
Fix CI flakiness and improve resilience

G-D-Petrov:
### Problem

Several CI failures are caused by infrastructure flakiness rather than code issues:

1. **Windows CI timeouts** -- The `stress` Windows jobs were timing out at 120 minutes.
   A 1,023-append stress test consumed the entire time budget.

2. **Linux stress tests OOM-killing the runner** -- Stress tests run with `-n 2` (2 parallel workers),
   doubling memory usage on GitHub runners with ~7GB RAM.

3. **Windows Azurite connection timeouts during test teardown** -- The Azure storage test fixture uses
   a 1-second connection timeout and 3-second cleanup timeout. On loaded Windows CI runners, Azurite
   responses can exceed these limits, causing spurious test errors
   (e.g. `ServiceRequestTimeoutError` on `test_write_append_update_read_scenario`).

4. **Flaky `test_symbol_names_with_all_chars`** -- A transient network error (curlCode 7) during
   write can leave a symbol persisted server-side while the client sees failure. The test's exact
   set equality assertion then fails because `list_symbols()` returns symbols the client didn't
   track as written.

5. **Windows micromamba crash** -- `micromamba.exe` crashes with `STATUS_DLL_NOT_FOUND`
   (`0xC0000135`) on Windows runners due to a broken binary in recent micromamba releases
   (mamba-org/setup-micromamba#306).

#### Test fixes

- **Move recursive normalizer compat tests to `tests/compat/`** -- The compat tests are moved from
  `tests/unit/.../test_recursive_normalizers.py` to a new
  `tests/compat/arcticdb/test_recursive_normalizers_compat.py`. This means they run in the dedicated
  `compat` CI job rather than competing for time in the `unit` job.

- **Reduce append count in `test_write_large_df_in_chunks`** -- Changed from 2^22 rows x 2^10 chunks
  (1,023 appends) to 2^27 rows x 2^5 chunks (31 appends). Total row count stays at 2^32.

- **Fix flaky `test_symbol_names_with_all_chars`** (`test_lib_naming.py`) -- Replaced fragile
  exact set equality with directional assertions: all successfully written symbols must be present,
  and any extra symbols on the server must be from writes that hit transient errors.

#### Stress test OOM fix

- **Run stress tests serially** (`build_steps.yml`) -- Changed `PYTEST_XDIST_MODE` for stress tests
  from `-n 2 --dist worksteal` to `-n 0` (serial). Eliminates the 2x memory multiplier from
  parallel workers on memory-constrained GitHub runners.

- **Remove OOM retry logic** (`parallel_test.sh`) -- Removed the retry-on-exit-137 logic. With
  stress tests now running serially, retrying with the same configuration won't help.

#### Azurite timeout fixes

- **Increase connection timeouts** (`storage_fixtures/azure.py`) -- Connection timeout 1s -> 5s,
  read timeout 2s -> 10s, container cleanup timeout 3s -> 10s.

#### Subprocess timeout fix

- **Increase subprocess timeout in `test_set_log_level`** (`test_config.py`) -- 10s -> 30s.

#### Micromamba version pin

- **Pin `micromamba-version: '2.1.0-0'`** across all workflows using `setup-micromamba` to avoid
  the broken binary in recent releases (mamba-org/setup-micromamba#306).

#### CI improvements

- **Add session-level faulthandler watchdog** -- A session-level timer is armed in
  `pytest_configure` on xdist workers with a 10-minute timeout. It fires if a worker hangs before
  any test starts running, dumping a traceback and killing the worker. The per-test timer replaces
  it once the first test begins.

- **Unbuffered pytest output** -- Added `python -u` and `sed -u` to `parallel_test.sh` so test
  results stream to CI logs in real-time instead of being buffered until the session ends.

- **Shallow clone for non-release builds** -- Changed the compile checkout to use `fetch-depth: 1`
  for non-tag builds, with submodules initialized separately. Full history is only fetched for
  release builds from tags.


poodlewars (approved):

phoebusm:
>Test fixes
Move recursive normalizer compat tests to tests/compat/ -- The compat tests are moved from
tests/stress/.../test_recursive_normalizers.py to a new
tests/compat/arcticdb/test_recursive_normalizers_compat.py. This means they run in the dedicated
compat CI job rather than competing for time in the stress job.

That doesn't look matching with the content of the changes. 
`python/tests/unit/arcticdb/version_store/test_recursive_normalizers.py`  to `python/tests/compat/arcticdb/test_recursive_normalizers_compat.py`

python/tests/conftest.py
phoebusm, line 192: Will there be any concern of race condition with the timer at `pytest_configure`?

phoebusm (approved):

python/tests/conftest.py
G-D-Petrov: No race condition concern here. The timer is only armed on xdist **workers** (guarded by `PYTEST_XDIST_WORKER` env var check), and each worker is a separate process with its own PID. The flow is:

1. `pytest_configure` arms a 10-minute session-level timer — this covers the collection/fixture-setup phase where no per-test timer exists yet
2. `pytest_runtest_protocol` fires before the first test, re-opens the same crash file and calls `dump_traceback_later()` again — this **replaces** the session timer with the per-test timer (CPython's faulthandler only supports one active `dump_traceback_later` at a time; a new call cancels the previous one)

So there's no window where two timers are active simultaneously. The session timer is either running (during collection) or has been replaced by the per-test timer (during test execution).

G-D-Petrov:
Good catch, fixed — it's `tests/unit/`, not `tests/stress/`. Updated the PR description.

PR #3106, opened 2026-05-13
Add benchmark script to measure column stats creation

markovskipetar:
# Column Stats Benchmarks

Standalone benchmarks for measuring the performance and memory usage of ArcticDB column stats creation. These are **not** part of the ASV framework — they are designed for manual profiling and regression checks without ASV overhead.

## Scripts

| Script | Role |
|--------|------|
| `run_bench_col_stats.py` | Orchestrator — runs all scenarios and prints a summary table |
| `bench_col_stats.py` | Worker — runs a single operation in isolation and prints JSON results |

## How It Works

`run_bench_col_stats.py` iterates over a set of `(rows, cols)` scenarios. For each scenario it:

1. Writes a `float64` dataframe of the given shape to LMDB via a subprocess.
2. Runs `create_column_stats` (MINMAX on all columns) 10 times, each in its own subprocess.
3. Collects elapsed time and peak RSS delta from each run.
4. Prints a summary table with min, max, and variance for both time and memory.

Each operation runs in a **fresh subprocess** (`bench_col_stats.py`) so that:

- RSS measurements start from a clean baseline, unaffected by previous runs or accumulated Python/ArcticDB state.
- A crash or OOM kill in one run does not abort the entire benchmark — the orchestrator catches the failure, cleans up, and reports a descriptive error.

Results are returned as JSON over stdout and parsed by the orchestrator.

## Usage

```bash
# Run all scenarios
python python/benchmarks/non_asv/run_bench_col_stats.py

# Run a single operation manually (useful for profiling one case)
python python/benchmarks/non_asv/bench_col_stats.py --scenario 500x500 --operation write_symbol
python python/benchmarks/non_asv/bench_col_stats.py --scenario 500x500 --operation create_stats
```

`--scenario` is `ROWSxCOLS`. `--operation` is either `write_symbol` or `create_stats`.

## Output

```
        rows    cols      write_time_s    stats_time_min    stats_time_max    stats_time_var    stats_rss_min_mb    stats_rss_max_mb    stats_rss_var_mb
---------------------------------------------------------------------------------------------------------------------------------------------------------
         500     500              0.42              1.23              1.45            0.0031                45.2                52.1                1.23
```

- `write_time_s` — wall-clock time in seconds for the symbol write (one run).
- `stats_time_*` — min, max, and variance of elapsed time across 10 stats-creation runs.
- `stats_rss_*` — min, max, and variance of peak RSS increase in MB across the same 10 runs. Sampled at 10 ms intervals inside the worker process.

High variance in either column is a signal of instability worth investigating.

## Scenarios

Default scenarios in `run_bench_col_stats.py`:

| rows | cols |
|------|------|
| 500 | 500 |
| 500 | 100 |
| 600 | 600 |
| 700 | 700 |
| 800 | 500 |
| 900 | 800 |
| 1,000 | 100 |

A commented-out block of larger scenarios (`5,000×5,000` up to `10,000×10,000`) is also present for heavier profiling — uncomment and replace `SCENARIOS` to use them.

## Cleanup

The LMDB store at `/tmp/arcticdb_bench_col_stats` is cleaned up automatically:

- After each scenario (symbol deleted so the next write starts fresh).
- On normal exit and on `SIGINT`/`SIGTERM` via `atexit` and signal handlers.


python/benchmarks/non_asv/bench_col_stats.py
```diff
@@ -0,0 +1,79 @@
...
+
+LMDB_PATH = "/tmp/arcticdb_bench_col_stats"
+
+
+def _cleanup():
+    shutil.rmtree(LMDB_PATH, ignore_errors=True)
+
+
+atexit.register(_cleanup)
+signal.signal(signal.SIGINT, lambda *_: exit(130))
```
poodlewars: You shouldn't do cleanup like this. This file should end with:

```
if __name__ == "__main__":
  try:
    run()
  finally:
    cleanup()
```

or you could put the same idea in your `run()`. There's no need for exit handlers.

python/benchmarks/non_asv/bench_col_stats.py
```diff
@@ -0,0 +1,79 @@
...
+        columns=[f"col_{i}" for i in range(cols)],
+    )
+
+    t0 = time.time()
+    lib.write(sym, df)
+    write_time = time.time() - t0
+
+    col_stats = {f"col_{i}": {"MINMAX"} for i in range(cols)}
+
+    mem_before = psutil.Process().memory_info().rss / 1e6
```
poodlewars: These memory measurements are wrong, we care about peak RSS for the function calls, not RSS after they're done. This means that each experiment must be run in its own process. We also care about measuring peak RSS for the write call. This means that in each run of the script you should just do a write or create stats for a single example, then you can write a bash script to run your code across the different examples and summarize the results in a readable way.

python/benchmarks/non_asv/bench_col_stats.py
```diff
@@ -0,0 +1,79 @@
...
+    t0 = time.time()
+    nvs.create_column_stats(sym, col_stats)
+    stats_time = time.time() - t0
```
poodlewars: Unless test cases are very slow, it's always best to run timings with a couple of un-measured warmup iterations, then N measured iterations, to get more stable results. Then report the mean, median and max across the measured iterations.

poodlewars (commented):
PR titles should always be more descriptive than this, eg `Add script to benchmark column stats creation` rather than `Column stats improvements`. You should also always include a description.

poodlewars:
Claude is right about the very large examples. These should be created by chunking up the dataframe with a series of `append` calls rather than ever loading the large dataframe in to memory.

markovskipetar:
=== scenario 500x500 ===
  [write_symbol] 500x500
  [create_stats] run 1/10
  [create_stats] run 2/10
  [create_stats] run 3/10
  [create_stats] run 4/10
  [create_stats] run 5/10
  [create_stats] run 6/10
  [create_stats] run 7/10
  [create_stats] run 8/10
  [create_stats] run 9/10
  [create_stats] run 10/10

=== scenario 500x100 ===
  [write_symbol] 500x100
  [create_stats] run 1/10
  [create_stats] run 2/10
  [create_stats] run 3/10
  [create_stats] run 4/10
  [create_stats] run 5/10
  [create_stats] run 6/10
  [create_stats] run 7/10
  [create_stats] run 8/10
  [create_stats] run 9/10
  [create_stats] run 10/10

=== scenario 600x600 ===
  [write_symbol] 600x600
  [create_stats] run 1/10
  [create_stats] run 2/10
  [create_stats] run 3/10
  [create_stats] run 4/10
  [create_stats] run 5/10
  [create_stats] run 6/10
  [create_stats] run 7/10
  [create_stats] run 8/10
  [create_stats] run 9/10
  [create_stats] run 10/10

=== scenario 700x700 ===
  [write_symbol] 700x700
  [create_stats] run 1/10
  [create_stats] run 2/10
  [create_stats] run 3/10
  [create_stats] run 4/10
  [create_stats] run 5/10
  [create_stats] run 6/10
  [create_stats] run 7/10
  [create_stats] run 8/10
  [create_stats] run 9/10
  [create_stats] run 10/10

=== scenario 800x500 ===
  [write_symbol] 800x500
  [create_stats] run 1/10
  [create_stats] run 2/10
  [create_stats] run 3/10
  [create_stats] run 4/10
  [create_stats] run 5/10
  [create_stats] run 6/10
  [create_stats] run 7/10
  [create_stats] run 8/10
  [create_stats] run 9/10
  [create_stats] run 10/10

=== scenario 900x800 ===
  [write_symbol] 900x800
  [create_stats] run 1/10
  [create_stats] run 2/10
  [create_stats] run 3/10
  [create_stats] run 4/10
  [create_stats] run 5/10
  [create_stats] run 6/10
  [create_stats] run 7/10
  [create_stats] run 8/10
  [create_stats] run 9/10
  [create_stats] run 10/10

=== scenario 1000x100 ===
  [write_symbol] 1000x100
  [create_stats] run 1/10
  [create_stats] run 2/10
  [create_stats] run 3/10
  [create_stats] run 4/10
  [create_stats] run 5/10
  [create_stats] run 6/10
  [create_stats] run 7/10
  [create_stats] run 8/10
  [create_stats] run 9/10
  [create_stats] run 10/10

        rows    cols      write_time_s    stats_time_min    stats_time_max    stats_time_var  stats_rss_min_mb  stats_rss_max_mb  stats_rss_var_mb
         500     500              0.06              0.13              0.14            0.0000              26.1              26.2              0.00
         500     100              0.02              0.03              0.04            0.0000              11.9              12.0              0.00
         600     600              0.07              0.16              0.17            0.0000              30.6              30.7              0.00
         700     700              0.08              0.19              0.20            0.0000              36.2              36.3              0.00
         800     500              0.06              0.15              0.15            0.0000              28.7              28.8              0.00
         900     800              0.09              0.23              0.25            0.0000              42.5              42.6              0.00
       1,000     100              0.02              0.04              0.04            0.0000              12.7              12.8              0.00

python/benchmarks/non_asv/bench_col_stats.py
```diff
@@ -0,0 +1,67 @@
...
+
+import numpy as np
+import pandas as pd
+import psutil
+from arcticdb import Arctic
+
+LMDB_PATH = "/tmp/arcticdb_bench_col_stats"
+SYMBOL_NAME = "test_symbol"
+
+signal.signal(signal.SIGINT, lambda *_: exit(130))
```
poodlewars: You don't need this??

markovskipetar: done

python/benchmarks/non_asv/bench_col_stats.py
markovskipetar: done

python/benchmarks/non_asv/bench_col_stats.py
markovskipetar: done

PR #3111, opened 2026-05-14
Record statistics during deletion, for delayed deletes to log

poodlewars:
For enterprise's delayed deletes logging https://github.com/man-group/arcticdb-enterprise/pull/323

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -52,10 +52,15 @@ struct KeySizesInfo {
     size_t compressed_size; // bytes
 };
 
+/**
+ * These statistics record the number of deletions *attempted*. In particular note that we attempt
+ * to delete column stats keys without checking whether they exist first.
+ */
 struct DeleteTreesStats {
     size_t index_keys_considered = 0;
     size_t index_keys_protected_by_snapshots = 0;
     size_t index_keys_deleted = 0;
+    size_t column_stats_keys_deleted = 0;
     size_t data_keys_deleted = 0;
 };
 
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -61,6 +61,7 @@ struct DeleteTreesStats {
     size_t index_keys_protected_by_snapshots = 0;
     size_t index_keys_deleted = 0;
     size_t column_stats_keys_deleted = 0;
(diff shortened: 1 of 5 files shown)


cpp/arcticdb/version/snapshot.hpp
```diff
@@ -95,6 +95,20 @@ std::unordered_map<SnapshotId, std::optional<VariantKey>> get_keys_for_snapshots
...
 
+struct MasterSnapshotMapWithStats {
+    MasterSnapshotMap map;
+    size_t total_snapshots = 0;
+};
+
+/**
+ * Map the index keys in every snapshot, additionally reporting the total number of snapshots iterated.
+ */
+MasterSnapshotMapWithStats get_master_snapshots_map_with_stats(
```
poodlewars, line 108: I should add a comment about where the stats are used here

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -52,7 +52,15 @@ struct KeySizesInfo {
     size_t compressed_size; // bytes
 };
 
-folly::Future<folly::Unit> delete_trees_responsibly(
+struct DeleteTreesStats {
+    size_t index_keys_considered = 0;
+    size_t index_keys_protected_by_snapshots = 0;
+    size_t index_keys_deleted = 0;
+    size_t data_keys_deleted = 0;
```
G-D-Petrov, line 65: Shouldn't we also count the column stats keys?

G-D-Petrov (approved):

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -52,7 +52,20 @@ struct KeySizesInfo {
...
+/**
+ * These statistics record the number of deletions *attempted*. In particular note that we attempt
+ * to delete column stats keys without checking whether they exist first.
+ */
+struct DeleteTreesStats {
+    size_t index_keys_considered = 0;
+    size_t index_keys_protected_by_snapshots = 0;
+    size_t index_keys_deleted = 0;
+    size_t column_stats_keys_deleted = 0;
+    size_t data_keys_deleted = 0;
```
IvoDD, line 65: We could also include `data_keys_considered` (i.e. `data_keys_to_be_deleted.size()` before the exclusion of `data_keys_not_to_be_deleted`). 

Could be useful for investigating deletion slowness/high storage use related to very fragmented append only data

IvoDD (approved):

PR #3113, opened 2026-05-15
Reject adding deleted versions to snapshot

G-D-Petrov:
#### Reference Issues/PRs
Monday ticket ref: 11959817519

- **Bug fix**: `add_to_snapshot` now rejects tombstoned/deleted versions that are not preserved in any other snapshot, instead of silently adding them. Previously, deleted versions could be included in the snapshot; delayed deletes could then physically remove the data, corrupting the snapshot. The fix uses `BatchGetVersionOption::LIVE_AND_TOMBSTONED_VER_REF_IN_OTHER_SNAPSHOT` and raises `NoSuchVersionException` listing all rejected versions. Tombstoned versions that are still referenced by another snapshot are allowed.

- **Bug fix**: When `LIVE_AND_TOMBSTONED_VER_REF_IN_OTHER_SNAPSHOT` is used, the version chain is now loaded with `LoadObjective::INCLUDE_DELETED`. Previously, `UNDELETED_ONLY` was used, which caused the chain walk to stop at `tombstone_all` entries, making versions behind them appear as `NEVER_EXISTED` even when they were protected by another snapshot.

- **Bug fix**: Both `add_to_snapshot` and `remove_from_snapshot` now write the updated `SNAPSHOT_REF` before deleting old keys. Previously, old keys were deleted first — a crash between deletion and write would leave the snapshot referencing non-existent data. This matches the safe ordering already used by `snapshot()`.

- **Refactor**: Deduplicated `batch_get_specific_version` (singular) by making it delegate to `batch_get_specific_versions` (plural). Both now return a `BatchGetVersionResult<FoundKey>` struct with `found` and `rejected` fields instead of a raw shared pointer.

- **Refactor**: Simplified `get_stream_index_map` to use the singular `batch_get_specific_version` (returning `unordered_map<StreamId, AtomKey>`) since each symbol has at most one version. Simplified `SpecificAndLatestVersionKeys` type accordingly.

- **Refactor**: Simplified `add_to_snapshot` in `version_store_api.cpp` by removing the intermediate `affected_keys` set — uses `specific_versions_index_map->contains()` directly.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -2211,7 +2211,21 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
             );
         }
 
-        specific_versions = batch_get_specific_versions(store(), version_map(), sym_versions);
+        specific_versions = batch_get_specific_versions(store(), version_map(), sym_versions, false);
+        std::vector<std::string> missing;
+        for (const auto& [symbol, versions] : sym_versions) {
+            for (auto version : versions) {
+                if (!specific_versions->count(std::make_pair(symbol, version))) {
```
poodlewars: `contains` not `count`?

python/tests/integration/arcticdb/version_store/test_snapshot.py
```diff
@@ -540,6 +540,71 @@ def test_snapshot_tombstoned_but_referenced_in_other_snapshot_version(basic_stor
...
+
+    # Snapshot should be unchanged - still contains only "existing_sym"
+    snap_versions = lib.list_versions(snapshot="snap")
+    assert len(snap_versions) == 1
+    assert snap_versions[0]["symbol"] == "existing_sym"
+    assert lib.read("existing_sym", as_of="snap").data == 100
+
+
+@pytest.mark.storage
+def test_add_to_snapshot_rejects_tombstoned_version_kept_alive_by_other_snapshot(basic_store_delayed_deletes_v1):
```
poodlewars: @alexowens90 reminded me in the meeting this morning that `.snapshot` allows snapshotting versions that are tombstoned but are held alive by a different snapshot. People might be relying on this behaviour for `add_to_snapshot`. We could allow this case but log a deprecation warning when it happens, and rip it out after inspecting those logs like we've done for a few things recently?

G-D-Petrov: I've kept the same warning as the one from `.snapshot` - https://github.com/man-group/ArcticDB/pull/3113/changes#diff-6fedeaa3bbd887e3d71024c0c1033efb682727ef0b3bafae3979ce9a53144a00R325

poodlewars (approved):

python/tests/integration/arcticdb/version_store/test_snapshot.py
```diff
@@ -540,6 +540,71 @@ def test_snapshot_tombstoned_but_referenced_in_other_snapshot_version(basic_stor
...
+
+    # Snapshot should be unchanged - still contains only "existing_sym"
+    snap_versions = lib.list_versions(snapshot="snap")
+    assert len(snap_versions) == 1
+    assert snap_versions[0]["symbol"] == "existing_sym"
+    assert lib.read("existing_sym", as_of="snap").data == 100
+
+
+@pytest.mark.storage
+def test_add_to_snapshot_allows_tombstoned_version_kept_alive_by_other_snapshot(basic_store_delayed_deletes_v1):
```
IvoDD: Would be nice to make sure these tests also pass with `delayed_deletes=False` and `prune_previous=True`

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -301,24 +304,49 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
-                    if (index_key) {
+                    auto version_details = find_index_key_for_version_id_and_tombstone_status(version, entry);
+                    if ((option == BatchGetVersionOption::ALL_VER_FOUND_IN_STORAGE &&
+                         version_details.version_status_ == VersionStatus::TOMBSTONED) ||
+                        version_details.version_status_ == VersionStatus::LIVE) {
                         std::lock_guard lock{*mutex};
-                        (*output)[std::pair(sym_version.first, version)] = *index_key;
+                        (*output)[std::pair(sym_version.first, version)] = version_details.key_.value();
+                    } else if (option == BatchGetVersionOption::LIVE_AND_TOMBSTONED_VER_REF_IN_OTHER_SNAPSHOT &&
+                               version_details.version_status_ == VersionStatus::TOMBSTONED) {
```
IvoDD, line 267: I think a deleted version can appear as `NEVER_EXISTED` if it is deleted by a tombstone all.

We should use `LoadObjective::INCLUDE_DELETED` when `LIVE_AND_TOMBSTONED_VER_REF_IN_OTHER_SNAPSHOT` or accept `VersionStatus::NEVER_EXISTED`.

We should also make sure we cover this in tests by writing a version chain like `v0 <- v1 <- tomstone_all(1) <- v2` and load up to v0. The current code would load up to v1 and report v0 as `NEVER_EXISTED`.

cpp/arcticdb/version/version_map_batch_methods.hpp
```diff
@@ -301,24 +304,49 @@ inline std::shared_ptr<std::unordered_map<std::pair<StreamId, VersionId>, AtomKe
...
-
-                    if (index_key) {
+                    auto version_details = find_index_key_for_version_id_and_tombstone_status(version, entry);
+                    if ((option == BatchGetVersionOption::ALL_VER_FOUND_IN_STORAGE &&
+                         version_details.version_status_ == VersionStatus::TOMBSTONED) ||
+                        version_details.version_status_ == VersionStatus::LIVE) {
                         std::lock_guard lock{*mutex};
-                        (*output)[std::pair(sym_version.first, version)] = *index_key;
+                        (*output)[std::pair(sym_version.first, version)] = version_details.key_.value();
+                    } else if (option == BatchGetVersionOption::LIVE_AND_TOMBSTONED_VER_REF_IN_OTHER_SNAPSHOT &&
```
IvoDD, line 266: Looking at `BatchGetVersionOption` it looks the `batch_get_specific_versions` is pretty duplicated with `batch_get_specific_version` above. Can't we make `batch_get_specific_version` reuse the plural variant by constructing single element vectors.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -2211,7 +2211,27 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
...
+                store(),
+                version_map(),
+                sym_versions,
+                BatchGetVersionOption::LIVE_AND_TOMBSTONED_VER_REF_IN_OTHER_SNAPSHOT
+        );
+
+        std::vector<std::string> missing;
+        for (const auto& [symbol, versions] : sym_versions) {
+            for (auto version : versions) {
+                if (!specific_versions->contains(std::make_pair(symbol, version))) {
```
IvoDD: Theoretically this `contains` check would be `O(num_versions_for_symbol)`. I know in this case it is `1` because of the check above. Still I think this bit would be clearer if we use the `batch_get_specific_version` variant which returns a single version per symbol.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -2197,39 +2197,42 @@ SpecificAndLatestVersionKeys LocalVersionedEngine::get_stream_index_map(
...
+            std::vector<std::string> rejected_strs;
+            for (const auto& [symbol, version] : result.rejected) {
+                rejected_strs.push_back(fmt::format("{}:{}", symbol, version));
             }
-        }
-        if (!missing.empty()) {
             missing_data::raise<ErrorCode::E_NO_SUCH_VERSION>(
-                    "add_to_snapshot: the following versions do not exist or have been deleted: {}",
-                    fmt::join(missing, ", ")
+                    "add_to_snapshot: the following versions do not exist or have been deleted "
```
IvoDD: Nit: we no longer discard versions that do not exist

Honestly I was fine with the raising behavior as well. In any case doesn't matter much.

IvoDD (approved):

PR #3116, opened 2026-05-15
Performance 11900115518: use atom key packed in version map cache

alexowens90:
#### Reference Issues/PRs
[11900115518](https://man312219.monday.com/boards/7852509418/pulses/11900115518)

Uses `AtomKeyPacked` instead of `AtomKey` in the two collections inside `VersionMapEntry` in order to reduce the memory footprint of the version map cache. Also uses `ankerl::unordered_dense::map` over `std::unordered_map` for the same reason. Full `AtomKey`s are only materialised when necessary.

### DIFF
--- a/cpp/arcticdb/entity/atom_key.hpp
+++ b/cpp/arcticdb/entity/atom_key.hpp
@@ -89,12 +89,16 @@ class AtomKeyImpl {
     friend bool operator!=(const AtomKeyImpl& l, const AtomKeyImpl& r) { return !(l == r); }
 
     friend bool operator<(const AtomKeyImpl& l, const AtomKeyImpl& r) {
-        const auto lt = std::tie(l.id_, l.version_id_, l.index_start_, l.index_end_, l.creation_ts_);
-        const auto rt = std::tie(r.id_, r.version_id_, r.index_start_, r.index_end_, r.creation_ts_);
+        const auto lt = std::tie(
+                l.id_, l.version_id_, l.index_start_, l.index_end_, l.creation_ts_, l.key_type_, l.content_hash_
+        );
+        const auto rt = std::tie(
+                r.id_, r.version_id_, r.index_start_, r.index_end_, r.creation_ts_, r.key_type_, r.content_hash_
+        );
         return lt < rt;
     }
 
-    friend bool operator>(const AtomKeyImpl& l, const AtomKeyImpl& r) { return !(l < r) && (l != r); }
+    friend bool operator>(const AtomKeyImpl& l, const AtomKeyImpl& r) { return r < l; }
 
     size_t get_cached_hash() const {
         if (!hash_) {
@@ -236,6 +240,7 @@ struct AtomKeyPacked {
     AtomKeyPacked(const AtomKey& atom_key) :
         version_id_(atom_key.version_id()),
         creation_ts_(atom_key.creation_ts()),
+        content_hash_(atom_key.content_hash()),
         key_type_(atom_key.type()),
         index_start_(atom_key.start_time()),
         index_end_(atom_key.end_time()) {}
@@ -243,6 +248,13 @@ struct AtomKeyPacked {
     AtomKey to_atom_key(const StreamId& stream_id) const {
         return AtomKey(stream_id, version_id_, creation_ts_, content_hash_, index_start_, index_end_, key_type_);
     }
+
+    const auto& version_id() const { return version_id_; }
+    const auto& creation_ts() const { return creation_ts_; }
+    const auto& content_hash() const { return content_hash_; }
+    const auto& type() const { return key_type_; }
+    const auto& start_index() const { return index_start_; }
+    const auto& end_index() const { return index_end_; }
 
     VersionId version_id_ = 0;
     timestamp creation_ts_ = 0;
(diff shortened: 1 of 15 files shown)


cpp/arcticdb/version/version_functions.hpp
```diff
@@ -222,7 +222,7 @@ inline folly::Future<version_store::TombstoneVersionResult> finalize_tombstone_a
     if (version_map->validate())
         entry->validate();
 
-    version_store::TombstoneVersionResult res{true, entry->head_->id()};
+    version_store::TombstoneVersionResult res{true, entry->stream_id_};
```
alexowens90, line 226: This was a pre-existing bug that dereferenced `std::nullopt` when trying to delete a symbol that had never existed (i.e. had no vref key), it started segfaulting with the other changes.

cpp/arcticdb/entity/atom_key.hpp
```diff
@@ -94,7 +94,7 @@ class AtomKeyImpl {
         return lt < rt;
     }
 
-    friend bool operator>(const AtomKeyImpl& l, const AtomKeyImpl& r) { return !(l < r) && (l != r); }
+    friend bool operator>(const AtomKeyImpl& l, const AtomKeyImpl& r) { return r < l; }
```
IvoDD, line 101: Noticed that this now breaks [std::weak_ordering](https://en.cppreference.com/cpp/utility/compare/weak_ordering) which requires exactly one of `a < b`, `a == b` and `a > b` to be true. And we now have only [std::partial_ordering](https://en.cppreference.com/cpp/utility/compare/partial_ordering)

E.g. if two atom keys are the same but have a different `type` all of the above would be false.

To be honest the previous version was also broken because `a > b && b > a` could have been true.

In any case I think if not worried about performance we should include all fields in the `std::tie` comparison. I think it would make reasoning about this class easier because it would give us [std::strong_ordering](https://en.cppreference.com/cpp/utility/compare/strong_ordering)

poodlewars:
Excellent point Claude 

```
Title is awkward (Performance 11900115518: use atom key packed in version map cache) - leading with the Monday ticket number is uninformative. Something like Use AtomKeyPacked in version map cache reads better.
```

poodlewars (approved):

cpp/arcticdb/version/version_map.hpp
```diff
@@ -38,6 +38,8 @@
 #include <arcticdb/version/version_utils.hpp>
 #include <arcticdb/util/lock_table.hpp>
 
+#include "storage/storage_utils.hpp"
```
IvoDD: Nit: include with `<>`

cpp/arcticdb/version/version_map_entry.hpp
```diff
@@ -411,15 +415,7 @@ struct VersionMapEntry {
     void check_stream_id() const {
         if (empty())
             return;
-
-        std::unordered_map<StreamId, std::vector<VersionId>> id_to_version_id;
-        if (head_)
-            id_to_version_id[head_->id()].push_back(head_->version_id());
-        for (const auto& k : keys_)
-            id_to_version_id[k.id()].push_back(k.version_id());
```
IvoDD, line 419: Here we used to check that all keys have the same stream_id.

We should move these assertions to methods like `unsift_key` or `try_set_tombstone_all` which get `AtomKey` arguments.

IvoDD (approved):

cpp/arcticdb/entity/atom_key.hpp
alexowens90: Yeh, comparing keys of different type is pretty meaningless, but so is having `index_start_` and `index_end_` before `creation_ts_` in the `std::tie`, so I'm fine with making things simple.

IvoDD (approved):

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -103,6 +103,7 @@ def test_special_chars(object_version_store):
             object_version_store.write(sym, df)
             vitem = object_version_store.read(sym)
             assert_equal(vitem.data, df)
+            object_version_store.version_store.clear()
```
alexowens90, line 106: Azurite silently maps `\` characters to `/` in this test. It was passing previously as we did not validate that keys on version/version ref segments had the same stream id as the version map entry. In practice, because `/` is before `\` in the character list, this meant that the `write` call for the `\` symbol actually loaded the version map entry for the `/` symbol, but so did the read, so everything appeared to work correctly.
The change is safe for real users because one of the following is true:

1. This is a bug in Azurite, real Azure blob storage fully supports `\` characters in blob paths (the [docs](https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#blob-names) don't mention a limitation like this, although a [7 year old stack overflow answer](https://stackoverflow.com/questions/54805654/do-we-have-list-of-unsupported-characters-for-azure-blob-file-names) does), and so this is only a problem in the testing environment.
2. Real Azure behaves the same way as Azurite. In this case, if a user has two symbols `a/b` and `b\a`, then it is possible they were previously accidentally reading the wrong one, and so raising an exception is better behaviour. If this is the case, we should disallow `\` characters in symbol names on Azure.

cpp/arcticdb/version/test/test_symbol_list.cpp
```diff
@@ -973,13 +973,11 @@ struct CheckSymbolsTask {
...
     log::version().set_pattern("%Y%m%d %H:%M:%S.%f %t %L %n | %v");
     std::vector<Future<Unit>> futures;
-    std::optional<AtomKey> previous_key;
     for (auto x = 0; x < 1000; ++x) {
         auto symbol = fmt::format("symbol_{}", x);
         SymbolList::add_symbol(store_, symbol, 0);
         auto key = atom_key_builder().build(symbol, KeyType::TABLE_INDEX);
-        version_map_->write_version(store_, key, previous_key);
-        previous_key = key;
+        version_map_->write_version(store_, key, std::nullopt);
```
alexowens90, line 980: This was wrong previously, it was setting the previous key to one with a different stream id to `key`.
Before this change stream ids were not checked to be identical in the version map entry, so it worked anyway.
With this change, it only failed on the sanitizer runs because they are slower than regular test runs, so the version map cache expired before the call to `get_symbol_set` below, which then raised in `read_segment_with_keys`.

PR #3123, opened 2026-05-19
Submit delete batches in parallel

poodlewars:
Discovered while working on delayed deletes. I created a library on PURE with 1000 symbols, with ~200 versions each and 10 snapshots.

Then running delayed deletes had these results:

| Side                                    | Elapsed (s) | Peak RSS (MiB) |
| --------------------------------------- | ----------: | -------------: |
| arcticdb HEAD (47bba8011, parallel batches) | 349.9648 |          780.3 |
| arcticdb 2d1e60edf (pre-batching)           | 399.6425 |          792.9 |
| arcticc                                     | 333.3081 |          421.3 |

Notably running delayed deletes in "dry run" mode (which skips the physical deletion) showed no performance difference between arcticc and ArcticDB, for example on a small example:

| Side     | Mode    | Elapsed (s) | Peak RSS (MiB) |
| -------- | ------- | ----------: | -------------: |
| arcticdb | dry-run |      6.4449 |          388.2 |
| arcticdb | real    |     19.2177 |          427.9 |
| arcticc  | dry-run |      8.6100 |          329.0 |
| arcticc  | real    |     11.2915 |          330.4 |

arcticc submitted `RemoveBatchTask` in chunks of 1000 https://mangit.maninvestments.com/projects/DATA/repos/arcticc/browse/arcticcxx_impl/async/async_store.hpp#238 but https://github.com/man-group/ArcticDB/pull/70 changed it to this implementation.

If we delete 10k keys, the old implementation in arcticc would submit 10 tasks to delete 1000 keys in parallel, whereas ArcticDB currently submits 10 HTTP requests to delete 1000 keys in serial.

This explains most of the performance difference between arcticc delayed deletes and enterprise delayed deletes.

To stop storage specific deletion batch size limits leaking out of the storage layer, this PR has each storage declare its maximum deletion batch size, which is then used by the `AsyncStore` when it calculates the batches.

### DIFF
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -383,26 +383,53 @@ class AsyncStore : public Store {
     folly::Future<std::vector<RemoveKeyResultType>> remove_keys(
             const std::vector<entity::VariantKey>& keys, storage::RemoveOpts opts
     ) override {
-        return keys.empty() ? std::vector<RemoveKeyResultType>()
-                            : async::submit_io_task(RemoveBatchTask{keys, library_, opts});
+        return remove_keys(std::vector<entity::VariantKey>{keys}, opts);
     }
 
     folly::Future<std::vector<RemoveKeyResultType>> remove_keys(
             std::vector<entity::VariantKey>&& keys, storage::RemoveOpts opts
     ) override {
-        return keys.empty() ? std::vector<RemoveKeyResultType>()
-                            : async::submit_io_task(RemoveBatchTask{std::move(keys), library_, opts});
+        if (keys.empty()) {
+            return std::vector<RemoveKeyResultType>{};
+        }
+        const auto batch_size = library_->max_delete_batch_size();
+        if (!batch_size.has_value() || keys.size() <= *batch_size) {
+            return async::submit_io_task(RemoveBatchTask{std::move(keys), library_, opts});
+        }
+        auto chunks = chunk_keys(std::move(keys), *batch_size);
+        auto futs = folly::window(
+                std::move(chunks),
+                [this, opts](std::vector<entity::VariantKey>&& chunk) {
+                    return async::submit_io_task(RemoveBatchTask{std::move(chunk), library_, opts});
+                },
+                async::TaskScheduler::instance()->io_thread_count()
+        );
+        return folly::collect(std::move(futs)).via(&async::io_executor()).thenValue([](auto&&) {
+            return std::vector<RemoveKeyResultType>{};
+        });
     }
 
     std::vector<RemoveKeyResultType> remove_keys_sync(
             const std::vector<entity::VariantKey>& keys, storage::RemoveOpts opts
     ) override {
-        return keys.empty() ? std::vector<RemoveKeyResultType>() : RemoveBatchTask{keys, library_, opts}();
+        return remove_keys_sync(std::vector<entity::VariantKey>{keys}, opts);
     }
 
     std::vector<RemoveKeyResultType> remove_keys_sync(std::vector<entity::VariantKey>&& keys, storage::RemoveOpts opts)
             override {
-        return keys.empty() ? std::vector<RemoveKeyResultType>() : RemoveBatchTask{std::move(keys), library_, opts}();
+        if (keys.empty()) {
+            return {};
+        }
+        const auto batch_size = library_->max_delete_batch_size();
+        if (!batch_size.has_value() || keys.size() <= *batch_size) {
+            return RemoveBatchTask{std::move(keys), library_, opts}();
+        }
+        std::vector<RemoveKeyResultType> result;
+        for (auto& chunk : chunk_keys(std::move(keys), *batch_size)) {
+            auto sub = RemoveBatchTask{std::move(chunk), library_, opts}();
+            result.insert(result.end(), std::make_move_iterator(sub.begin()), std::make_move_iterator(sub.end()));
+        }
+        return result;
     }
 
     std::vector<folly::Future<VariantKey>> batch_read_compressed(
@@ -513,6 +540,19 @@ class AsyncStore : public Store {
 
   private:
     friend class arcticdb::toolbox::apy::LibraryTool;
+
+    static std::vector<std::vector<entity::VariantKey>> chunk_keys(
+            std::vector<entity::VariantKey>&& keys, size_t batch_size
+    ) {
+        std::vector<std::vector<entity::VariantKey>> chunks;
+        chunks.reserve((keys.size() + batch_size - 1) / batch_size);
+        for (size_t i = 0; i < keys.size(); i += batch_size) {
+            const size_t end = std::min(i + batch_size, keys.size());
+            chunks.emplace_back(std::make_move_iterator(keys.begin() + i), std::make_move_iterator(keys.begin() + end));
+        }
+        return chunks;
+    }
+
     std::shared_ptr<storage::Library> library_;
     std::shared_ptr<arcticdb::proto::encoding::VariantCodec> codec_;
     const EncodingVersion encoding_version_;
--- a/cpp/arcticdb/async/async_store.hpp
+++ b/cpp/arcticdb/async/async_store.hpp
@@ -372,25 +372,22 @@ class AsyncStore : public Store {
         return async::submit_io_task(WriteCompressedBatchTask(std::move(kvs), library_));
     }
 
-    folly::Future<RemoveKeyResultType> remove_key(const entity::VariantKey& key, storage::RemoveOpts opts) override {
+    folly::Future<folly::Unit> remove_key(const entity::VariantKey& key, storage::RemoveOpts opts) override {
         return async::submit_io_task(RemoveTask{key, library_, opts});
     }
 
-    RemoveKeyResultType remove_key_sync(const entity::VariantKey& key, storage::RemoveOpts opts) override {
-        return RemoveTask{key, library_, opts}();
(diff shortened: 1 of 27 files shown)


cpp/arcticdb/storage/s3/nfs_backed_storage.cpp
```diff
@@ -218,6 +218,13 @@ void NfsBackedStorage::do_remove(std::span<VariantKey> variant_keys, RemoveOpts)
     s3::detail::do_remove_impl(std::span(enc), root_folder_, bucket_name_, *s3_client_, NfsBucketizer{});
 }
 
+std::optional<size_t> NfsBackedStorage::max_delete_batch_size() const {
+    return std::min(
+            s3::detail::DELETE_OBJECTS_LIMIT,
+            static_cast<size_t>(ConfigsMap::instance()->get_int("S3Storage.DeleteBatchSize", 1000))
```
G-D-Petrov: nit: the default value should be `DELETE_OBJECTS_LIMIT`, same goes for `S3Storage::max_delete_batch_size`

G-D-Petrov (approved):

cpp/arcticdb/entity/variant_key.hpp
```diff
@@ -48,6 +50,22 @@ inline KeyType variant_key_type(const VariantKey& vk) {
     return std::visit([](const auto& key) { return key.type(); }, vk);
 }
 
+template<std::ranges::range R>
+requires std::same_as<std::remove_cvref_t<std::ranges::range_value_t<R>>, VariantKey>
+std::vector<KeyType> key_types(R&& keys) {
```
alexowens90: Feels like the return type of this should be an unordered set?

cpp/arcticdb/storage/mongo/mongo_storage.cpp
```diff
@@ -146,35 +145,31 @@ bool MongoStorage::do_fast_delete() {
...
-                    keys_not_found.push_back(k);
-                }
-            } catch (const mongocxx::operation_exception& ex) {
-                // mongo delete does not throw exception if key not found, it returns 0 as delete count
-                std::string object_name = std::string(variant_key_view(k));
-                raise_mongo_exception(ex, object_name);
+    for (auto& k : variant_keys) {
+        auto collection = collection_name(variant_key_type(k));
+        try {
+            auto result = client_->remove_keyvalue(db_, collection, k);
```
alexowens90, line 154: Does the Mongo client not have a method to delete multiple documents (at least in the same collection) at once?

cpp/arcticdb/storage/s3/s3_storage.hpp
```diff
@@ -103,6 +105,8 @@ class GCPXMLStorage : public S3Storage {
   public:
     GCPXMLStorage(const LibraryPath& lib, OpenMode mode, const GCPXMLSettings& conf);
 
+    std::optional<size_t> max_delete_batch_size() const override { return std::nullopt; }
```
alexowens90, line 108: Google has no limit!?

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -298,63 +299,44 @@ void do_remove_impl(
...
-                        to_delete.clear();
-                    }
-                }
-            });
+    std::vector<std::string> to_delete;
+    to_delete.reserve(ks.size());
+    for (const auto& k : ks) {
+        auto key_type_dir = key_type_folder(root_folder, variant_key_type(k));
+        to_delete.emplace_back(object_path(bucketizer.bucketize(key_type_dir, k), k));
+    }
```
alexowens90, line 317: Could use `std::ranges::transform`

cpp/arcticdb/storage/s3/detail-inl.hpp
```diff
@@ -298,63 +299,44 @@ void do_remove_impl(
...
+
+    std::vector<std::optional<query_stats::RAIIAddTime>> stat_timers;
+    if (query_stats::QueryStats::instance()->is_enabled()) {
+        auto distinct_key_types = key_types(ks);
+        stat_timers.reserve(distinct_key_types.size());
+        for (auto kt : distinct_key_types) {
+            stat_timers.emplace_back(query_stats::add_task_count_and_time(query_stats::TaskType::S3_DeleteObjects, kt));
```
alexowens90, line 324: I guess the query stats will now look odd for a batch-delete of mixed key types, but I'm not sure users can even trigger this in practice so not worth worrying about

cpp/arcticdb/async/async_store.hpp
```diff
@@ -530,6 +555,19 @@ class AsyncStore : public Store {
 
   private:
     friend class arcticdb::toolbox::apy::LibraryTool;
+
+    static std::vector<std::vector<entity::VariantKey>> chunk_keys(
+            std::vector<entity::VariantKey>&& keys, size_t batch_size
+    ) {
```
alexowens90, line 538: This is a pattern we have dotted all over the place, could template the element type and stick it in the utils folder somewhere

cpp/arcticdb/async/async_store.hpp
```diff
@@ -384,26 +384,53 @@ class AsyncStore : public Store {
...
+        auto futs = folly::window(
+                std::move(chunks),
+                [this, opts](std::vector<entity::VariantKey>&& chunk) {
+                    return async::submit_io_task(RemoveBatchTask{std::move(chunk), library_, opts});
+                },
+                async::TaskScheduler::instance()->io_thread_count()
+        );
+        return folly::collect(std::move(futs)).via(&async::io_executor()).thenValue([](auto&&) {
+            return std::vector<RemoveKeyResultType>{};
+        });
```
alexowens90, line 406: https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/stream/stream_sink.hpp#L49-L54
Don't have to, but if you want to rip out this pointless abstraction while you're here I won't object

cpp/arcticdb/util/key_utils.hpp
```diff
@@ -21,25 +21,51 @@ inline void delete_keys_of_type_if(
...
+                 flush_threshold](VariantKey&& key) {
+                    if (!predicate(key))
+                        return;
+                    keys.emplace_back(std::move(key));
+                    if (keys.size() >= flush_threshold) {
+                        auto batch = std::move(keys);
+                        keys = {};
+                        keys.reserve(flush_threshold);
+                        const auto batch_size = batch.size();
+                        store->remove_keys(std::move(batch)).get();
```
alexowens90, line 47: There is a slightly more elegant model that would remove the need for `flush_threshold `. Instead of calling `get` immediately, we maintain a set of `remove_keys` futures we've submitted. Folly let's you check if a future is done without calling `get()` using `isReady()`, so the size of this collection could be kept bounded. It's quite a bit more complicated though, and probably doesn't add much in this case.
At least worth a comment about why `remove_keys` is better than `remove_keys_sync` in this case though.

python/tests/integration/arcticdb/version_store/test_bulk_delete.py
```diff
@@ -0,0 +1,63 @@
...
+@pytest.mark.storage
+def test_bulk_delete_batching(object_and_mem_and_lmdb_version_store, sym):
+    lib = object_and_mem_and_lmdb_version_store
+    num_versions = 12
+    batch_size = 5
+
+    with (
+        config_context("S3Storage.DeleteBatchSize", batch_size),
+        config_context("AzureStorage.DeleteBatchSize", batch_size),
+    ):
```
alexowens90: There's a `config_context_multi` for setting multiple options at once

python/tests/integration/arcticdb/version_store/test_bulk_delete.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def _write_versions(lib, symbol, num_versions):
+    for i in range(num_versions):
+        df = pd.DataFrame({"x": np.arange(i, i + 5, dtype=np.int64)})
+        lib.write(symbol, df)
+
+
+@pytest.mark.storage
+def test_bulk_delete_batching(object_and_mem_and_lmdb_version_store, sym):
```
alexowens90: This is quite an indirect test of the batching. I get that we don't have query stats for Azure right now, but the test below is much more explicit

alexowens90 (approved):

cpp/arcticdb/storage/s3/s3_storage.hpp
poodlewars: It doesn't actually do batching, so we just submit one request at a time.

cpp/arcticdb/entity/variant_key.hpp
poodlewars: This does less allocation and keeps things ordered, so I think the vector is OK. I'll rename to `unique_key_types`.

cpp/arcticdb/storage/s3/detail-inl.hpp
poodlewars: I don't think that would make things any clearer

cpp/arcticdb/storage/mongo/mongo_storage.cpp
poodlewars: It does, `collection.delete_many`. I'd prefer not to fiddle around with the Mongo client (which is barely used) here though.

cpp/arcticdb/async/async_store.hpp
poodlewars: I can't find many instances of this, there some in the `StorageMover` and some in `batch_read_internal`, plus a bunch of `folly::window`.

PR #3131, opened 2026-05-22
Partial implementation insertion for numeric columns

vasil-pashov:
Phase one of insertion has some limitations that are going to be addressed in future PRs. The limitations are:
* Works for timeseries only
* String columns are not supported
* Does not support a single value spanning multiple segments
* Can insert only if the source index value lies in the middle of some existing segment, e.g. if there are 3 segments [5, 20) [20, 25) [30, 50) source rows with indexes 0, 4, 26, 29, 55 cannot be inserted
* Segments are not sliced. Insertion can produce large segments.

Matching on multiple columns is supported.

The algorithm is merges two sorted lists in O(m + n) where m is the size of the segment and n is the size of the source data that lies within this segment.

This PR also handles mixed strategy update on match and insert when source row is not matched
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/processing/clause_merge_update.cpp
+++ b/cpp/arcticdb/processing/clause_merge_update.cpp
@@ -52,8 +52,7 @@ void filter_selected_ranges_and_keys_and_reindex_entities(
     ranges_and_keys = std::move(new_ranges_and_keys);
 }
 
-template<typename TDT>
-requires util::instantiation_of<TDT, TypeDescriptorTag>
+template<util::type_descriptor_tag TDT>
 position_t write_py_string_to_pool_or_throw(
         PyObject* const py_string_object, const size_t row_in_segment, const RowRange& row_range,
         std::optional<ScopedGILLock>& gil_lock, StringPool& new_string_pool, const std::string_view column_name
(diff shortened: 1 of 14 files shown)


cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -95,81 +93,8 @@ void rebuild_sequence_column_in_new_pool(
...
-                if (is_a_string(row.value())) {
-                    const std::string_view string_in_target = target_string_pool.get_const_view(row.value());
-                    const OffsetString offset_in_new_pool = new_string_pool.get(string_in_target);
-                    row.value() = offset_in_new_pool.offset();
-                }
-            }
-        });
-    }
-    return is_column_changed;
-}
```
vasil-pashov, line 162: This relied on the invariant that sorted(flatten(rows_to_update)) = true) which is not true in general. Example
```
target:

idx  a
0    1
1    2
1    3
1    4
1    5
2    6

source:
idx   a
1     4
1     3
1     5
```

PR #3138, opened 2026-05-27
Add optional parameter for processing_on_cpu_executor for column stats

markovskipetar:
## Summary

Adds an optional `process_on_cpu_executor` parameter to the read/processing pipeline functions so the callers can choose which executor to use for clause processing.
  
Call chain where the parameter was threaded through (in the context of column stats):

  `create_column_stats_impl` → 
  `read_process_and_collect` → 
  `read_and_schedule_processing` → 
  `schedule_clause_processing` →
  `schedule_first_iteration_of_clauses`

The change does not affect any other pipelines that reuse any of the mentioned-above functions due to default value of `false` to the optional flag

  Previously the choice was controlled globally by the `Storage.ProcessOnCpuExecutor` config (read once into a `static` and applied to all reads). That kill-switch is removed in favor of a per-call argument that defaults to `false` (IO executor), preserving existing behavior for normal reads.

  `create_column_stats_impl` is the first caller to opt in: it now runs on the CPU executor by default, gated by a new
  `VersionStore.UseIOExecutorForColstats` config (set to `1` to fall back to the IO executor).


vasil-pashov:
I agree with claude on this one. This changes the defaults. Let's keep them the same.

poodlewars:
Closing in favour of https://github.com/man-group/ArcticDB/pull/3191

PR #3139, opened 2026-05-28
Add NaN and NULL count to MIN/MAX stats

markovskipetar:
## Add NaN and null counts to MINMAX column stats
  
 > [!IMPORTANT]
  > **The regression in the benchmark commit is EXPECTED and was reviewed by me and Ivo Dilov.**
  > It happened because the size of the `COLUMN_STATS_KEY` became **~2× bigger** due to the two
  > new columns added for **`null_count`** and **`nan_count`**.

  ### Summary
  Extends the `MINMAX` column statistic to also record per-segment counts of `NaN` (floating point) and null values (NaT timestamps, plus  sparse-map gaps such as those from Arrow validity bitmaps), alongside the existing min/max values.

  ### Changes
  - **Proto**: Added `NAN_COUNT_V1 = 3` and `NULL_COUNT_V1 = 4` entries to the `ColumnStatsType` enum in `column_stats.proto`.
  - **Aggregation** (`unsorted_aggregation.{hpp,cpp}`): `MinMaxAggregatorData` now tracks `nan_count_` (floating-point NaNs) and `null_count_` (NaT cells plus sparse-map gaps). For sparse columns, the gap count `last_row()+1 - row_count()` is added to `null_count_` so validity-bitmap nulls that the dense iterator never visits are still counted. `MinMaxAggregator` takes two additional output column names and `finalize` produces 4 output columns (`MIN`, `MAX`, `NAN_COUNT`, `NULL_COUNT`) instead of 2.
  - **Pipeline** (`column_stats.cpp`): `MINMAX` external stat now maps to the four internal stat types; `v1_NAN_COUNT` / `v1_NULL_COUNT` operator strings and segment column naming wired through.
  - **Tests**: Added `test_column_stats_nan_and_null_counts` covering multi-segment symbols with mixed NaN, NaT, and valid values across floating-point and timestamp columns. Widened the shared `assert_stats_equal` helper to subselect the received frame to the columns the test cares about (so existing min/max-only tests don't have to spell out all-zero count columns), and updated `test_column_stats_header_metadata` for the 4-entries-per-column header layout.

### DIFF
--- a/cpp/arcticdb/pipeline/column_stats.cpp
+++ b/cpp/arcticdb/pipeline/column_stats.cpp
@@ -97,6 +97,10 @@ std::string type_to_operator_string(ColumnStatTypeInternal type) {
         return "v1_MIN";
     case ColumnStatTypeInternal::MAX_V1:
         return "v1_MAX";
+    case ColumnStatTypeInternal::NAN_COUNT_V1:
+        return "v1_NAN_COUNT";
+    case ColumnStatTypeInternal::NULL_COUNT_V1:
+        return "v1_NULL_COUNT";
     default:
         internal::raise<ErrorCode::E_ASSERTION_FAILURE>("Unknown column stat type requested");
     }
@@ -146,7 +150,9 @@ ColumnStats::ColumnStats(
             switch (entry.type()) {
             case MIN_V1:
             case MAX_V1:
-                external_type = ColumnStatType::MINMAX;
+            case NAN_COUNT_V1:
+            case NULL_COUNT_V1:
+                external_type = ColumnStatType::MINMAX; // null and nan are calculated inline with minmax
                 break;
             case UNKNOWN:
             default:
@@ -255,7 +261,10 @@ namespace {
 std::unordered_set<ColumnStatTypeInternal> external_to_internal(ColumnStatType type) {
     switch (type) {
     case ColumnStatType::MINMAX:
-        return {ColumnStatTypeInternal::MIN_V1, ColumnStatTypeInternal::MAX_V1};
+        return {ColumnStatTypeInternal::MIN_V1,
+                ColumnStatTypeInternal::MAX_V1,
+                ColumnStatTypeInternal::NAN_COUNT_V1,
+                ColumnStatTypeInternal::NULL_COUNT_V1};
     default:
         internal::raise<ErrorCode::E_ASSERTION_FAILURE>("Unknown column stat type");
     }
@@ -339,7 +348,13 @@ std::optional<Clause> ColumnStats::clause() const {
                         ),
                         ColumnName(
                                 to_segment_column_name(name_and_stat_types.mangled_name, ColumnStatTypeInternal::MAX_V1)
-                        )
+                        ),
+                        ColumnName(to_segment_column_name(
+                                name_and_stat_types.mangled_name, ColumnStatTypeInternal::NAN_COUNT_V1
+                        )),
+                        ColumnName(to_segment_column_name(
+                                name_and_stat_types.mangled_name, ColumnStatTypeInternal::NULL_COUNT_V1
+                        ))
                 ));
                 break;
             default:
--- a/cpp/arcticdb/pipeline/column_stats_dispatch.hpp
+++ b/cpp/arcticdb/pipeline/column_stats_dispatch.hpp
@@ -105,15 +105,48 @@ struct FlippedComparator<NotEqualsOperator> {
 };
 
 template<typename Func>
+StatsComparison adjust_for_missing(StatsComparison result, uint64_t missing_count) {
+    if (missing_count == 0) { // no nulls or nans
+        return result;
+    }
+
+    using F = std::remove_cvref_t<Func>;
+
+    if constexpr (std::is_same_v<F, NotEqualsOperator>) {
+        if (result == StatsComparison::NONE_MATCH) {
+            // col = [3, NaN, NaN, 3], due to min=3 and max=3:
+            // Filter [col != 3] returns NONE_MATCH, but [NaN != 3] is a match, downgrade to UNKNOWN
+            return StatsComparison::UNKNOWN;
+        }
+    } else {
+        if (result == StatsComparison::ALL_MATCH) {
+            // col = [3, NaN, NaN, 3], due to min=3 and max=3:
+            // Filter [col == 3] returns ALL_MATCH, but [NaN == 3] is not a match, downgrade to UNKNOWN
+
+            // col = [5, 8, NaN], due to min=5 and max=8:
+            // Filter [col < 100] returns ALL_MATCH, but [NaN < 100] is not a match, downgrade to UNKNOWN
+            return StatsComparison::UNKNOWN;
+        }
+    }
+
+    return result;
+}
+
+template<typename Func>
 StatsComparison stats_comparator(const ColumnStatsValues& stats_lhs, const Value& val_rhs, Func&& func) {
     if (stats_lhs.column_absent) {
         return StatsComparison::NONE_MATCH;
+    }
+    if (stats_lhs.only_nulls()) {
+        // Every row is null: `!=` matches all of them, every other comparison matches none.
+        return std::is_same_v<std::remove_cvref_t<Func>, NotEqualsOperator> ? StatsComparison::ALL_MATCH
+                                                                            : StatsComparison::NONE_MATCH;
     }
     if (!stats_lhs.min) {
         return StatsComparison::UNKNOWN;
     }
 
-    return details::visit_type(stats_lhs.min->data_type(), [&](auto stats_tag) -> StatsComparison {
+    auto result = details::visit_type(stats_lhs.min->data_type(), [&](auto stats_tag) -> StatsComparison {
         using StatsTag = std::remove_reference_t<decltype(stats_tag)>;
 
         return details::visit_type(val_rhs.data_type(), [&](auto val_tag) -> StatsComparison {
@@ -152,6 +185,7 @@ StatsComparison stats_comparator(const ColumnStatsValues& stats_lhs, const Value
             return StatsComparison::UNKNOWN;
         });
     });
+    return adjust_for_missing<Func>(result, stats_lhs.nan_count + stats_lhs.null_count);
 }
 
 template<typename Func>
@@ -163,7 +197,7 @@ StatsComparison stats_comparator(const ColumnStatsValues& stats_lhs, const Colum
         return StatsComparison::UNKNOWN;
     }
 
-    return details::visit_type(stats_lhs.min->data_type(), [&](auto lhs_tag) -> StatsComparison {
+    auto result = details::visit_type(stats_lhs.min->data_type(), [&](auto lhs_tag) -> StatsComparison {
         using LhsTag = std::remove_reference_t<decltype(lhs_tag)>;
 
         return details::visit_type(stats_rhs.min->data_type(), [&](auto rhs_tag) -> StatsComparison {
@@ -202,6 +236,9 @@ StatsComparison stats_comparator(const ColumnStatsValues& stats_lhs, const Colum
             return StatsComparison::UNKNOWN;
         });
     });
+    return adjust_for_missing<Func>(
+            result, stats_lhs.nan_count + stats_lhs.null_count + stats_rhs.nan_count + stats_rhs.null_count
+    );
 }
 
 template<typename Func>
--- a/cpp/arcticdb/pipeline/column_stats_filter.cpp
+++ b/cpp/arcticdb/pipeline/column_stats_filter.cpp
@@ -143,32 +143,40 @@ std::vector<StatsMetadataForColumn> calculate_stats_metadata(
         );
         stats_metadata_for_column.col_name = std::string{tsd.fields().at(data_col_offset).name()};
         for (const auto& entry : entry_list.entries()) {
-            if (entry.type() != arcticc::pb2::column_stats_pb2::MIN_V1 &&
-                entry.type() != arcticc::pb2::column_stats_pb2::MAX_V1) {
+            const auto entry_type = entry.type();
+            const bool is_min_max = entry_type == arcticc::pb2::column_stats_pb2::MIN_V1 ||
+                                    entry_type == arcticc::pb2::column_stats_pb2::MAX_V1;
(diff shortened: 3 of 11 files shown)


cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -46,9 +55,15 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
...
-                    // Skip NaN/NaT as they don't generate a stable ordering
+                if constexpr (is_floating_point_type(type_info::data_type)) {
+                    if (is_nat_or_nan(curr)) {
+                        ++nan_count_;
+                        any_nan = true;
+                        return;
+                    }
+                } else if constexpr (is_time_type(type_info::data_type)) {
                     if (is_nat_or_nan(curr)) {
+                        ++null_count_;
```
poodlewars: I think `nat` should contribute to the `nan` count

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -160,6 +164,8 @@ ColumnStats::ColumnStats(
             switch (entry.type()) {
             case MIN_V1:
             case MAX_V1:
+            case NAN_COUNT_V1:
+            case NULL_COUNT_V1:
                 external_type = ColumnStatType::MINMAX;
```
poodlewars: Worth a comment here, that `nan` and `null` are calculated inline with `minmax`

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -274,6 +280,116 @@ def test_column_stats_only_nat_values(lmdb_version_store, any_output_format):
...
+    lib.create_column_stats(sym, {"float_col": {"MINMAX"}, "ts_col": {"MINMAX"}})
+
+    cs = pl.from_arrow(lib.read_column_stats(sym))
+    assert cs["v1_NAN_COUNT(float_col)"].to_list() == [3]
+    assert cs["v1_NULL_COUNT(float_col)"].to_list() == [0]
+    assert cs["v1_NAN_COUNT(ts_col)"].to_list() == [2]
+    assert cs["v1_NULL_COUNT(ts_col)"].to_list() == [0]
+
+
+def test_column_stats_null_count_sparse_floats(lmdb_version_store, any_output_format):
```
poodlewars: This test should cover multiple segments, else we'd have no coverage of the null calculation in the case that there is more than one segment.

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -2046,3 +2072,113 @@ def test_column_stats_bool_vs_int_cross_type(
...
+# level (e.g. min == max == v says "all == v" but in-band NaN rows fail `==`), the verdict
+# must be downgraded to UNKNOWN so the segment is still read.
+
+
+@pytest.mark.parametrize(
+    "query_expr,pandas_expr",
+    [
+        pytest.param(lambda q: q["a"] != 3.0, lambda df: df["a"] != 3.0, id="ne"),
+        pytest.param(lambda q: ~(q["a"] == 3.0), lambda df: ~(df["a"] == 3.0), id="not_eq"),
+        pytest.param(lambda q: q["a"].isnotin([3.0]), lambda df: ~df["a"].isin([3.0]), id="isnotin"),
```
poodlewars: Test for a negated `isin` as well?

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -2046,3 +2072,113 @@ def test_column_stats_bool_vs_int_cross_type(
...
+    qs.reset_stats()
+    result = lib.read(sym, query_builder=q).data
+    table_data_reads = get_table_data_read_count()
+
+    expected = df[pandas_expr(df)]
+    assert_frame_equal(expected, result)
+    assert table_data_reads == 1, "Segment must be read because in-band NaN rows match the filter"
+
+
+def test_column_stats_no_prune_ne_when_nat(in_memory_version_store, clear_query_stats, column_stats_filtering_enabled):
```
poodlewars: Can't we cover the different query builder expressions like in the `test_column_stats_no_prune_when_in_band_nan` parameterization?

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -2046,3 +2072,113 @@ def test_column_stats_bool_vs_int_cross_type(
...
+    qs.reset_stats()
+    result = lib.read(sym, query_builder=q).data
+    table_data_reads = get_table_data_read_count()
+
+    expected = df[df["a"] != ts]
+    assert_frame_equal(expected, result)
+    assert table_data_reads == 1
+
+
+def test_column_stats_no_prune_ne_when_sparse_null(
```
poodlewars, line 2232: Can't we cover the different query builder expressions like in the test_column_stats_no_prune_when_in_band_nan parameterization?

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -2046,3 +2072,113 @@ def test_column_stats_bool_vs_int_cross_type(
...
+    qs.reset_stats()
+    result = lib.read(sym, query_builder=q).data
+    table_data_reads = get_table_data_read_count()
+
+    expected = df[df["a"] != 3.0]
+    assert_frame_equal(expected, result)
+    assert table_data_reads == 1
+
+
+def test_column_stats_still_prunes_when_nan_does_not_save_row(
```
poodlewars, line 2266: Can't we cover the different query builder expressions like in the test_column_stats_no_prune_when_in_band_nan parameterization? Here we could cover all the comparision operators besides (in)equality?

cpp/arcticdb/pipeline/column_stats_dispatch.hpp
```diff
@@ -104,6 +99,27 @@ struct FlippedComparator<NotEqualsOperator> {
     using type = NotEqualsOperator;
 };
 
+template<typename Func>
+StatsComparison adjust_for_missing(StatsComparison result, uint64_t missing_count) {
```
poodlewars, line 108: Worth a comment to explain

poodlewars (approved):

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -2046,3 +2072,142 @@ def test_column_stats_bool_vs_int_cross_type(
...
+    qs.reset_stats()
+    result = lib.read(sym, query_builder=q).data
+    table_data_reads = get_table_data_read_count()
+
+    expected = df[pandas_expr(df)]
+    assert_frame_equal(expected, result)
+    assert table_data_reads == 1, "Segment must be read because in-band NaN rows match the filter"
+
+
+_NAT_TS = pd.Timestamp("2025-01-01")
```
vasil-pashov: nit: I think calling this NAT is confusing

python/tests/unit/arcticdb/test_column_stats_creation.py
vasil-pashov, line 1: 1. Can this be tested with arrow writes. It can have both nan and null in the same float column which makes the case quite interesting
2. This is a good candidate for hypothesis testing. We can append several segments to a symbol, count the nan/none elements and compare to the input

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -93,7 +140,7 @@ StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationTyp
         ARCTICDB_DEBUG(log::version(), "Unsupported unary operator for stats {}", operation);
         return util::variant_match(
                 left,
-                [](const std::vector<StatsComparison>& comparisons) -> StatsVariantData {
+show                [](const std::vector<StatsComparison>& comparisons) -> StatsVariantData {
```
vasil-pashov: What is this `show` at the beginning of the row? Looks like a typo

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -25,6 +25,15 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
         using type_info = ScalarTypeInfo<decltype(col_tag)>;
         using RawType = typename type_info::RawType;
         if constexpr (!is_sequence_type(type_info::data_type)) {
+            // null_count_ tracks rows that are genuinely absent (sparse-map gaps from Arrow
+            // validity bitmaps). nan_count_ tracks in-band sentinel values found while iterating
+            // the dense values (NaN for floats, NaT for time types) - see the for_each below.
+            if (input_column.column_->is_sparse()) {
+                const auto sparse_gap_count = input_column.column_->last_row() + 1 - input_column.column_->row_count();
+                if (sparse_gap_count > 0) {
```
vasil-pashov: Why are you testing if this is greater than 0. Shouldn't it always be non-negative? I think that we don't need the conditional here.

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -46,9 +55,11 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
             [[maybe_unused]] bool any_nan{false};
             arcticdb::for_each<typename type_info::TDT>(*input_column.column_, [&](auto value) {
                 const auto& curr = static_cast<RawType>(value);
```
vasil-pashov: Appreciate it's not your change, but I don't see a reason for having this static_cast here. Value should be of RawType

cpp/arcticdb/pipeline/column_stats_dispatch.hpp
```diff
@@ -104,6 +99,34 @@ struct FlippedComparator<NotEqualsOperator> {
...
+    if (missing_count == 0) { // no nulls or nans
+        return result;
+    }
+
+    using F = std::remove_reference_t<Func>;
```
vasil-pashov: We want either `std::remove_cvref_t` or `std::decay_t` because we need to remove both reference and const qualifiers

cpp/arcticdb/pipeline/test/test_column_stats_data.cpp
```diff
@@ -411,4 +411,95 @@ TEST(ColumnStatsDataTest, SparseColumnAbsentMarkedCorrectly) {
...
+    seg.set_metadata(std::move(any));
+
+    ColumnStatsData data(std::move(seg), build_test_tsd());
+    ASSERT_FALSE(data.empty());
+
+    auto row0 = data.find_row(100, 200);
+    ASSERT_TRUE(row0.has_value());
+    auto v0 = stats_for(data, "price", *row0);
+    ASSERT_TRUE(v0.min.has_value());
+    ASSERT_EQ(v0.min->get<int64_t>(), 10);
```
vasil-pashov, line 493: nit: add same asserts for the max column for completeness

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -181,32 +189,56 @@ std::unordered_map<std::string, StatsForColumn> load_stats_by_column(
...
-                    for_each_enumerated<typename type_info::TDT>(
-                            column,
-                            [&](const ColumnData::Enumeration<typename type_info::RawType>& enumerating_it) {
-                                auto idx = static_cast<size_t>(enumerating_it.idx());
-                                if (idx >= first_kept && idx < last_kept_excl) {
-                                    dest.at(idx - first_kept) = Value{enumerating_it.value(), type_info::data_type};
+        stats_for_column.nan_counts.resize(num_rows, 0);
+        stats_for_column.null_counts.resize(num_rows, 0);
+
+        if (stats_metadata_for_column.data_type != DataType::UNKNOWN) {
```
vasil-pashov: Why did you add this check? Did you get UNKNOWN in some codepath. It's most likely an error. It'd would be better to add internal::check that it's never unknown

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -264,6 +274,189 @@ def test_column_stats_only_nat_values(lmdb_version_store, any_output_format):
...
+    assert cs["v1_MAX(col_1)"].to_list() == [2.0, 4.0]
+
+
+def test_column_stats_arrow_nan_and_null_same_column(lmdb_version_store_arrow):
+    lib = lmdb_version_store_arrow
+    lib._cfg.write_options.segment_row_size = 100
+    sym = "test_column_stats_arrow_nan_and_null_same_column"
+
+    def table(values):
+        return pa.table({"f": pa.array(values, pa.float64())})
```
vasil-pashov, line 403: This definition is repeated in at least two places. You should extract it as a function. Ideally I'd prefer not having it at all `pa.table({"f": pa.array(values, pa.float64())})` is not that long and having it inline in the code makes it more obvious what's going on

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -264,6 +274,189 @@ def test_column_stats_only_nat_values(lmdb_version_store, any_output_format):
...
+
+    lib.write(sym, table(segments[0]))
+    for seg in segments[1:]:
+        lib.append(sym, table(seg))
+
+    lib.create_column_stats_experimental(sym)
+    cs = pl.from_arrow(lib.read_column_stats_experimental(sym))
+
+    # NaN is the only float not equal to itself; null is None.
+    expected_nan = sum(1 for seg in segments for v in seg if v is not None and v != v)
```
vasil-pashov: np.isnan

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -264,6 +274,189 @@ def test_column_stats_only_nat_values(lmdb_version_store, any_output_format):
...
+    expected_null = sum(1 for seg in segments for v in seg if v is None)
+
+    assert cs["v1_NAN_COUNT(f)"].sum() == expected_nan
+    assert cs["v1_NULL_COUNT(f)"].sum() == expected_null
```
vasil-pashov: It would be better to check the count for each segment instead of the total count

vasil-pashov (approved):

PR #3141, opened 2026-06-01
Rename APIs for column stats to _experimental and simplify their signatures

markovskipetar:
# Experimental: rename `create_column_stats` and default to all numeric columns

  ## Summary

  Rename `NativeVersionStore.create_column_stats` to `create_column_stats_experimental` (Python-only — no C++ rename) and make the `column_stats` argument **optional**. When omitted, the call computes `MINMAX` for every column of the symbol, instead of forcing the user to specify columns. 

  ## Before: column dict was required

  ```python
  lib.write(sym, df)
  lib.create_column_stats(sym, {"col_1": {"MINMAX"}, "col_2": {"MINMAX"}, ...})   # mandatory dict
  stats = lib.read_column_stats(sym)
  ```

  The user had to (a) remember to call it after `write`, **and** (b) hand-roll the per-column dict.

  ## Now: dict is optional, default = every supported column

  ```python
  lib.write(sym, df)
  lib.create_column_stats_experimental(sym)              # auto-MINMAX over every numeric/temporal col
  stats = lib.read_column_stats(sym)
  ```

  Explicit dicts still work unchanged:

  ```python
  lib.create_column_stats_experimental(sym, {"col_1": {"MINMAX"}})
  ```

  ## Changes

  ### `python/arcticdb/version_store/_store.py`

  - `create_column_stats` → `create_column_stats_experimental`.
  - `column_stats: Dict[str, Set[str]]` is now `Optional[..] = None`.
  - New private helper `_auto_column_stats(symbol, as_of)`: reads the symbol descriptor via `get_info` and returns `{col: {"MINMAX"}}` for every column
  - Cross-references in `drop_column_stats` and `get_column_stats_info` docstrings updated to point at the new name.


python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -477,64 +349,8 @@ def test_column_stats_pickled_symbol(version_store_factory, lib_name, encoding_v
     lib.write(sym, 1)
     assert lib.is_symbol_pickled(sym)
 
-    column_stats_dict = {"col_1": {"MINMAX"}}
-    with pytest.raises(SchemaException):
-        lib.create_column_stats(sym, column_stats_dict)
-
-
-def test_column_stats_multiple_creates(version_store_factory, lib_name, encoding_version, any_output_format):
```
poodlewars, line 485: Why removing this one?

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -1084,44 +850,18 @@ def test_column_stats_col_called_index(
...
 )
 def test_column_stats_multiindex(
-    version_store_factory, lib_name, encoding_version, any_output_format, level_name, index_level_name, stored_col_name
+    version_store_factory, lib_name, encoding_version, any_output_format, index_level_name, stored_col_name
 ):
-    """Column stats on a multiindex DataFrame: stats on inner index levels and data columns.
-    The inner index level can be referenced by the user-facing name ("level"), the mangled name ("__idx__level"),
-    or the __fkidx__ name for unnamed index levels."""
+    """Column stats on a multiindex DataFrame: auto-discovery picks up data columns; the inner
+    index level is not included by default."""
```
poodlewars: This is wrong, we definitely need to create stats over the inner index levels.

poodlewars (changes requested):

python/tests/unit/arcticdb/test_column_stats_creation.py
markovskipetar: `test_column_stats_multiple_creates` was removed because its purpose was to call create column stats with different column across different calls: `{"col_1": {"MINMAX"}}`, then `{"col_2": {"MINMAX"}}` and verify the merge was successful. But whit the new updated API, we cannot achieve this behavior because we create stats over all columns, without specifying particular columns.

`test_column_stats_string_column_minmax` was removed because its purpose was to test that creating stats over string column raises `SchemaException`. But with the new update API, the python layer iterates over all column and collects only the eligible columns for stats (numeric and bools)

python/tests/unit/arcticdb/test_column_stats_creation.py
poodlewars: I don't think this comment has been addressed properly, we still want to create stats over the multi-index columns,

```
    expected = index_columns_to_pl(lib, sym).with_columns(
        pl.Series(f"v1_MIN({stored_col_name})", [10, 30]),
        pl.Series(f"v1_MAX({stored_col_name})", [20, 40]),
        pl.Series("v1_MIN(val)", [100, 300]),
        pl.Series("v1_MAX(val)", [200, 400]),
    )
```

But this test just has:

```
    expected = index_columns_to_pl(lib, sym).with_columns(
        pl.Series("v1_MIN(val)", [100, 300]),
        pl.Series("v1_MAX(val)", [200, 400]),
    )
```

markovskipetar: I think we do have the correct inner multi-index col stats calculation: https://github.com/man-group/ArcticDB/pull/3141/changes#diff-fd55a7274e8a49f65e2f62074974d20809bde05f639f4d55fbe7f8430f772358R1304

poodlewars: Yes I thought that look right too, but we have https://github.com/man-group/ArcticDB/pull/3141#discussion_r3441217749

poodlewars (approved):

python/arcticdb/version_store/_store.py
```diff
@@ -1266,59 +1266,110 @@ def _batch_update_internal(
...
+            TypeDescriptor.ValueType.NANOSECONDS_UTC,
+        }
+        timeseries_descriptor = self._get_info(symbol, as_of).timeseries_descriptor
+        fields = list(timeseries_descriptor.fields)
+        num_index_fields = self._num_physically_stored_index_fields(timeseries_descriptor)
+
+        start = 1 if num_index_fields > 0 else 0
+        eligible_fields = [field for field in fields[start:] if field.type.value_type in numeric_value_types]
+
+        self._reject_duplicated_data_columns(
```
poodlewars, line 1319: I don't think that we actually need to reject duplicated data columns any more, now that the stats creation is across all columns? We could create stats for these?

python/arcticdb/version_store/_store.py
```diff
@@ -1266,59 +1266,110 @@ def _batch_update_internal(
...
+            timeseries_descriptor,
+        )
+
+        return {field.name: {"MINMAX"} for field in eligible_fields}
+
+    @staticmethod
+    def _num_physically_stored_index_fields(timeseries_descriptor) -> int:
+        normalization = timeseries_descriptor.normalization
+        if normalization.WhichOneof("input_type") != "df":
+            return 0
```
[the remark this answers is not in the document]
poodlewars: It looks like you've fixed this case. Please add a test for this case, or show me where one already exists.

python/arcticdb/version_store/_store.py
poodlewars: I think the problem here is that you have left the C++ layer alone. I think that's actually fine for now while we wait for the API to settle down. Let's leave this as is.

poodlewars (approved):

PR #3145, opened 2026-06-02
Retriable reads: check version ref before retrying on pruned-key errors

jamesblackburn:

  When a reader resolves a symbol from its cached version chain just before a concurrent writer
  supersedes and eagerly prunes the old version's keys, the read fails with `KeyNotFoundException`
  or `NoDataFoundException`. Rather than surfacing this error, the retry loop now recovers
  transparently:

  1. Reads the `VERSION_REF` key to compare the storage head against the cached head.
  2. **If unchanged**: the data is genuinely missing (not a race), so the exception is propagated
     immediately without consuming a retry slot.
  3. **If changed**: invalidates the stale cache entry. The retry repopulates it via the existing
     `LOAD_LATEST` shortcut, reading only the `VERSION_REF` — not the full version chain.

  Net cost per retry: **2 `VERSION_REF` reads** and **0 `VERSION` reads**, regardless of how many
  live versions exist for the symbol.

  #### Tests added

  - `test_read_retry.py` — five Python unit tests covering the happy path, the no-retry path
    (genuinely missing version), per-symbol cache scoping (unrelated symbols unaffected), and O(1)
    read count with N=15 live versions asserted via `query_stats`.
  - `InvalidateIfVersionRefChangedReturnsTrueWhenChangedFalseWhenUnchanged` — C++ unit test
    verifying the method returns `true` and invalidates the cache on a ref change, and `false` when
    unchanged.
  - `test_concurrent_read_write_eager_prune` — stress test using two LMDB handles with concurrent
    readers and eager-pruning writers.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -437,46 +437,78 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
         const ReadOptions& read_options, std::shared_ptr<std::any> handler_data
 ) {
     py::gil_scoped_release release_gil;
-    const auto identifier = util::variant_match(
-            version_query.content_,
-            [&](const std::shared_ptr<PreloadedIndexQuery>& preloaded_index_query) -> VersionIdentifier {
-                std::optional<ColumnStatsSource> column_stats;
-                if (preloaded_index_query->column_stats_seg_) {
-                    column_stats.emplace(ColumnStatsSource{
-                            preloaded_index_query->column_stats_seg_, ColumnStatsQueryMetadata(read_query->clauses_)
-                    });
-                }
-                return std::make_shared<IndexInformation>(
-                        std::pair{preloaded_index_query->index_key_, preloaded_index_query->index_seg_.clone()},
-                        std::move(column_stats)
-                );
-            },
-            [&](const auto&) -> VersionIdentifier {
-                auto version = get_version_to_read(stream_id, version_query);
-                return get_version_identifier(stream_id, version_query, read_options, version);
-            }
-    );
 
-    auto root_result = read_frame_for_version(store(), identifier, read_query, read_options, handler_data).get();
-    auto& keys = root_result.frame_and_descriptor_.keys_;
-    if (keys.empty()) {
-        return {std::move(root_result), {}};
-    } else {
-        std::vector<folly::Future<ReadVersionOutput>> node_futures;
-        node_futures.reserve(keys.size());
-        for (const auto& key : keys) {
-            node_futures.emplace_back(read_frame_for_version(store(), key, read_query, read_options, handler_data));
+    // A concurrent writer using prune_previous_version without background deletion physically deletes
+    // a superseded version's keys. A reader can resolve a version just before it is pruned and then
+    // fail to fetch its (now deleted) index/data keys. Rather than surface that race, drop the cached
+    // version chains so we re-resolve to the current version, and retry. A genuinely missing version
+    // raises E_NO_SUCH_VERSION (not caught here) and still fails fast. Preloaded-index reads carry
+    // their own index segment, so re-resolution cannot help them.
+    const bool is_preloaded = std::holds_alternative<std::shared_ptr<PreloadedIndexQuery>>(version_query.content_);
+    const int64_t max_attempts = is_preloaded ? 1 : ConfigsMap::instance()->get_int("VersionStore.ReadRetries", 3) + 1;
+
+    for (int64_t attempt = 1;; ++attempt) {
+        try {
+            const auto identifier = util::variant_match(
+                    version_query.content_,
+                    [&](const std::shared_ptr<PreloadedIndexQuery>& preloaded_index_query) -> VersionIdentifier {
+                        std::optional<ColumnStatsSource> column_stats;
+                        if (preloaded_index_query->column_stats_seg_) {
+                            column_stats.emplace(ColumnStatsSource{
+                                    preloaded_index_query->column_stats_seg_,
+                                    ColumnStatsQueryMetadata(read_query->clauses_)
+                            });
+                        }
+                        return std::make_shared<IndexInformation>(
+                                std::pair{preloaded_index_query->index_key_, preloaded_index_query->index_seg_.clone()},
+                                std::move(column_stats)
+                        );
+                    },
+                    [&](const auto&) -> VersionIdentifier {
+                        auto version = get_version_to_read(stream_id, version_query);
+                        return get_version_identifier(stream_id, version_query, read_options, version);
+                    }
+            );
+
+            auto root_result =
+                    read_frame_for_version(store(), identifier, read_query, read_options, handler_data).get();
+            auto& keys = root_result.frame_and_descriptor_.keys_;
+            if (keys.empty()) {
+                return {std::move(root_result), {}};
+            } else {
+                std::vector<folly::Future<ReadVersionOutput>> node_futures;
+                node_futures.reserve(keys.size());
+                for (const auto& key : keys) {
+                    node_futures.emplace_back(
+                            read_frame_for_version(store(), key, read_query, read_options, handler_data)
+                    );
+                }
+                auto node_trys = folly::collectAll(node_futures).get();
+                std::vector<ReadVersionOutput> node_results;
+                node_results.reserve(node_trys.size());
+                std::transform(
+                        std::make_move_iterator(node_trys.begin()),
+                        std::make_move_iterator(node_trys.end()),
+                        std::back_inserter(node_results),
+                        [](auto&& try_result) { return std::move(try_result).value(); }
+                );
+                return {std::move(root_result), std::move(node_results)};
+            }
+        } catch (const storage::NoDataFoundException&) {
+            if (attempt >= max_attempts || !version_map()->invalidate_if_version_ref_changed(store(), stream_id))
+                throw;
+        } catch (const storage::KeyNotFoundException&) {
+            if (attempt >= max_attempts || !version_map()->invalidate_if_version_ref_changed(store(), stream_id))
+                throw;
         }
-        auto node_trys = folly::collectAll(node_futures).get();
-        std::vector<ReadVersionOutput> node_results;
-        node_results.reserve(node_trys.size());
-        std::transform(
-                std::make_move_iterator(node_trys.begin()),
-                std::make_move_iterator(node_trys.end()),
-                std::back_inserter(node_results),
-                [](auto&& try_result) { return std::move(try_result).value(); }
+        ARCTICDB_DEBUG(
+                log::version(),
+                "Read of symbol '{}' raced with a concurrent prune; version ref changed, retrying "
+                "(attempt {} of {})",
+                stream_id,
+                attempt,
+                max_attempts
         );
-        return {std::move(root_result), std::move(node_results)};
     }
 }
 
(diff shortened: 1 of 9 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -437,46 +437,78 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
...
-    auto root_result = read_frame_for_version(store(), identifier, read_query, read_options, handler_data).get();
-    auto& keys = root_result.frame_and_descriptor_.keys_;
-    if (keys.empty()) {
-        return {std::move(root_result), {}};
-    } else {
-        std::vector<folly::Future<ReadVersionOutput>> node_futures;
-        node_futures.reserve(keys.size());
-        for (const auto& key : keys) {
-            node_futures.emplace_back(read_frame_for_version(store(), key, read_query, read_options, handler_data));
+        } catch (const storage::NoDataFoundException&) {
```
jamesblackburn: This is the change - the rest is whitespace.

maxim-morozov (commented):
This has a real potential to DDos the storage. In case of storage slowdown or some network blip, we should check what exception we are getting, to make sure we dont retry in those cases. Otherwise, we will make things pretty bad in situations like this. The AWS sdk has retries as well, so it will quickly scale exponentially in terms of storage requests.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -437,46 +437,78 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
...
+            if (attempt >= max_attempts || !version_map()->invalidate_if_version_ref_changed(store(), stream_id))
```
IvoDD: On a retry we'll end up reading the version ref twice:
- Once in `invalidate_if_version_ref_changed` to check whether it changed
- Once in `read_frame_for_version` to read the version chain

Probably not a big deal since retries will be fairly infrequent.

Ideally we could read it only once and short circuit if the same, but I think the current is good enough

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -437,46 +437,78 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
...
-                    });
+
+    // A concurrent writer using prune_previous_version without background deletion physically deletes
+    // a superseded version's keys. A reader can resolve a version just before it is pruned and then
+    // fail to fetch its (now deleted) index/data keys. Rather than surface that race, drop the cached
+    // version chains so we re-resolve to the current version, and retry. A genuinely missing version
+    // raises E_NO_SUCH_VERSION (not caught here) and still fails fast. Preloaded-index reads carry
+    // their own index segment, so re-resolution cannot help them.
+    const bool is_preloaded = std::holds_alternative<std::shared_ptr<PreloadedIndexQuery>>(version_query.content_);
+    const int64_t max_attempts = is_preloaded ? 1 : ConfigsMap::instance()->get_int("VersionStore.ReadRetries", 3) + 1;
```
IvoDD: What do you think about retrying only on latest version queries?

It is unlikely for a failing a version chain v2->v1->v0 a `read(as_of=0)` to get deleted. People usually either always prune previous or never do it.

IvoDD: Also I think the default retries should be fewer. Even just 1.
If someone races with their reads so frequently people are unlikely to get the result they want either way.

cpp/arcticdb/version/version_map.hpp
```diff
@@ -216,6 +216,25 @@ class VersionMapImpl {
...
+    // Returns true if the VERSION_REF head for stream_id differs from the cached head, and
+    // invalidates the cache entry so the next check_reload fetches the current version.
+    // Returns false if the ref is unchanged (missing data is genuine, not a write race).
+    bool invalidate_if_version_ref_changed(std::shared_ptr<Store> store, const StreamId& stream_id) {
+        VersionMapEntry ref_entry(stream_id);
+        read_symbol_ref(store, stream_id, ref_entry);
+
+        std::lock_guard lock(map_mutex_);
+        auto it = map_.find(stream_id);
+        const std::optional<AtomKey> cached_head = (it != map_.end()) ? it->second->head_ : std::nullopt;
```
IvoDD: I think this logic would be easier to follow if we short circuit the case where `it == map_.end()` with `return true`. There was no ref key to have changed.

Theoretically it is possible for the `invalidate_if_version_ref_changed` to return `false` if there was no cached entry for the symbol and the version ref contains no link to a version key (which should not be possible) but still makes reasoning about this harder imo

python/tests/stress/arcticdb/version_store/test_concurrent_read_and_write.py
```diff
@@ -47,3 +57,27 @@ def test_concurrent_read_write(writer_store, reader_store):
...
+    try:
+        reader.start()
+        writer.start()
+        reader.join(5)
+        writer.join(0.001)
+    finally:
+        writer.terminate()
+        reader.terminate()
+
+    assert exceptions_in_reader.empty()
```
IvoDD, line 88: It feels like this test might be flaky. It doesn't seem impossible for the writer process to invalidate the reader multiple times.

cpp/arcticdb/version/local_versioned_engine.cpp
IvoDD: I guess the `NoDataFoundException` is needed because of exception reraises like [this](https://github.com/man-group/ArcticDB/blob/bc47b3ef5483398736011c5f0aa4114f54ba4612/cpp/arcticdb/version/version_core.cpp#L2829)

I agree it's the correct thing for this PR but I think we should leave the more precise KeyNotFound exception in those reraises.

python/tests/unit/arcticdb/version_store/test_read_retry.py
```diff
@@ -0,0 +1,152 @@
...
+        qs.reset_stats()
+        other_result = reader.read("other")  # must still be a pure cache hit
+        other_stats = qs.get_query_stats()
+        qs.disable()
+
+    assert raced_result.data["a"][0] == 1
+    # One retry: one ref read for the changed-ref check, one more for storage_reload's LOAD_LATEST
+    # shortcut. Both are VERSION_REF reads; no full VERSION-chain traversal.
+    assert _version_ref_reads(raced_stats) == 2
+    assert _version_reads(raced_stats) == 0
```
IvoDD, line 112: It would be useful to show that the index key is read just once

python/tests/unit/arcticdb/version_store/test_read_retry.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # Warm the reader's cache so it resolves latest = v{N-1} without re-reading the ref.
+        assert reader.read(sym).data["a"][0] == N - 1
+
+        # Concurrent writer creates v_N and eagerly prunes v_{N-1}.
+        writer.write(sym, _df(N), prune_previous_version=True)
+
+        qs.enable()
+        qs.reset_stats()
+        result = reader.read(sym)  # stale cache -> one retry
```
IvoDD, line 141: It would be nice to also test a similar version chain with `read(as_of=0)` depending on what we decide for [this comment](https://github.com/man-group/ArcticDB/pull/3145/changes#r3346750313) this could mean no retries or a different exception being raised.

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Agreed, reading as_of specific versions, timestamps and snapshots will not benefit from retrying

alexowens90: Also agreed, if writes are happening so fast that a single retry doesn't help, then we create a thundering herd problem by retrying more times

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Unfortunately the current API can raise either depending on exactly when the failure occurs

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -437,46 +437,78 @@ ReadVersionWithNodesOutput LocalVersionedEngine::read_dataframe_version_internal
...
         }
-        auto node_trys = folly::collectAll(node_futures).get();
-        std::vector<ReadVersionOutput> node_results;
-        node_results.reserve(node_trys.size());
-        std::transform(
-                std::make_move_iterator(node_trys.begin()),
-                std::make_move_iterator(node_trys.end()),
-                std::back_inserter(node_results),
-                [](auto&& try_result) { return std::move(try_result).value(); }
+        ARCTICDB_DEBUG(
```
alexowens90: This macro gets compiled out of release builds. I would make the log info level - we want to know when this is happening

cpp/arcticdb/version/version_map.hpp
```diff
@@ -216,6 +216,25 @@ class VersionMapImpl {
...
+
+        if (cached_head == ref_entry.head_)
+            return false;
+
+        if (it != map_.end())
+            map_.erase(it);
```
alexowens90: The symbol should also be erased from the lock table if it is present

python/tests/stress/arcticdb/version_store/test_concurrent_read_and_write.py
alexowens90: Agreed, this test will not be reliable, better to use the storage failure simulator for this sort of thing

python/tests/unit/arcticdb/version_store/test_read_retry.py
```diff
@@ -0,0 +1,152 @@
...
+
+import pandas as pd
+import pytest
+
+import arcticdb.toolbox.query_stats as qs
+from arcticdb.exceptions import KeyNotFoundException, NoDataFoundException
+from arcticdb.util.test import config_context, query_stats_operation_count
+
+# Large enough that the reader's cached version chain never expires during a test.
+STICKY_RELOAD_INTERVAL = 2_000_000_000_000
```
alexowens90: This is in nanoseconds, so isn't very long at all

python/tests/unit/arcticdb/version_store/test_read_retry.py
```diff
@@ -0,0 +1,152 @@
...
+    """Guard that the retry is load-bearing: with VersionStore.ReadRetries=0 the same race surfaces
+    the missing-key error instead of recovering."""
+    writer = in_memory_store_factory()
+    reader = in_memory_store_factory(reuse_name=True)
+    sym = "sym"
+
+    with (
+        config_context("VersionMap.ReloadInterval", STICKY_RELOAD_INTERVAL),
+        config_context("VersionStore.ReadRetries", 0),
+    ):
```
alexowens90: `config_context_multi`

python/tests/unit/arcticdb/version_store/test_read_retry.py
```diff
@@ -0,0 +1,152 @@
+"""Deterministic tests for the read-retry behaviour in read_dataframe_version_internal.
```
alexowens90: Missing copyright

alexowens90:
This doesn't handle `read_batch`

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Yeh `invalidate_if_version_ref_changed` isn't the right way to do this. If the ref key has changed, we should continue `LOAD_LATEST_UNDELETED` and make this the new entry

PR #3154, opened 2026-06-05
Add creation of col stats from row_range or date_range

markovskipetar:
## Summary
                                
Adds optional `date_range` and `row_range`. When provided, column stats are only (re)computed for row-slices overlapping the given range — existing stats for out-of-range slices are kept, in-range slice stats are replaced. The two parameters are mutually exclusive.

### DIFF
--- a/cpp/arcticdb/pipeline/column_stats.cpp
+++ b/cpp/arcticdb/pipeline/column_stats.cpp
@@ -90,6 +90,51 @@ SegmentInMemory merge_column_stats_segments(const std::vector<SegmentInMemory>& 
     return merged;
 }
 
+SegmentInMemory merge_column_stats_with_range_replacement(
+        const SegmentInMemory& old_segment, SegmentInMemory new_segment, const entity::TimestampRange& range_replaced
+) {
+    const size_t old_row_count = old_segment.row_count();
+    std::vector<SegmentInMemory> to_merge;
+    to_merge.reserve(3);
+
+    size_t first_in_range = old_row_count;
+    size_t last_in_range = 0;
+    bool found_overlap = false;
+
+    for (size_t row = 0; row < old_row_count; ++row) {
+        auto start_opt = old_segment.scalar_at<entity::timestamp>(row, start_index_column_offset);
+        auto end_opt = old_segment.scalar_at<entity::timestamp>(row, end_index_column_offset);
+
+        util::check(
+                start_opt.has_value() && end_opt.has_value(),
+                "Missing start/end index in old column stats segment at row {}",
+                row
+        );
+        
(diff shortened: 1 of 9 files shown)


cpp/arcticdb/pipeline/column_stats.hpp
```diff
@@ -15,6 +15,14 @@ namespace arcticdb {
 
 SegmentInMemory merge_column_stats_segments(const std::vector<SegmentInMemory>& segments);
 
+// Builds a merged column-stats segment that keeps every row of `old_segment` whose
+// [start_index, end_index] lies fully outside `range_replaced`, and inserts the rows from
+// `new_segment` (covering the in-range row-slices). Both segments must share the same
+// stat-column schema. Used by the range-restricted RMW path in create_column_stats_impl.
+SegmentInMemory merge_column_stats_with_range_replacement(
```
poodlewars, line 22: I think let's just leave this merging with the old stats stuff out, at least for now, as it's a bit complicated and I'm not certain that we will need it.

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -90,6 +90,51 @@ SegmentInMemory merge_column_stats_segments(const std::vector<SegmentInMemory>&
...
+    } else {
+        if (first_in_range > 0) {
+            to_merge.emplace_back(old_segment.truncate(0, first_in_range, false));
+        }
+        if (last_in_range + 1 < old_row_count) {
+            to_merge.emplace_back(old_segment.truncate(last_in_range + 1, old_row_count, false));
+        }
+    }
+    to_merge.emplace_back(std::move(new_segment));
+    return merge_column_stats_segments(to_merge);
```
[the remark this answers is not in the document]
poodlewars: I think point 2 here is valid. Need careful testing of these boundary cases, and how this behaves with dynamic schema where new columns may have been appended to the symbol, or where columns may have been type promoted.

python/arcticdb/version_store/_store.py
```diff
@@ -1266,14 +1266,27 @@ def _batch_update_internal(
                 _log_warning_on_writing_empty_dataframe(data_vector[idx], symbols[idx])
         return result
 
-    def create_column_stats(
-        self, symbol: str, column_stats: Dict[str, Set[str]], as_of: Optional[VersionQueryInput] = None
+    def create_column_stats_experimental(
```
poodlewars: This has picked up changes from another of your branches somehow. The rename shouldn't be in this PR. You should just add date_range and row_range params (no rename in this PR).

poodlewars:
There are no new tests, and the rename has broken the CI.

PR #3155, opened 2026-06-05
WIP Enhancement 12036593840: append compact_data_inline

alexowens90:
#### Reference Issues/PRs
[12036593840](https://man312219.monday.com/boards/7852509418/pulses/12036593840)

Adds a `compact_data_inline` argument to `append` (on both the V1 and V2 APIs). This is roughly (although not exactly) equivalent to calling `append` followed by `compact_data`, but only creates a single new version. The invariant that every row slice will have `segment_row_size` ±33% rows after calling `append` with `compact_data_inline=True` still holds, but the exact on disk structure may differ when calling the two methods separately due to the slicing that can happen in `append` when `compact_data_inline=False`. The exception to this invariant is when the symbol does not exist (i.e. `upsert`). In this case the usual slice+write path is taken, which can result in very small segments due to the `FixedSlicer` greedily consuming rows. A better approach would be to have the `FixedSlicer` logic be more like `ReslicingInfo`, so that rows are more evenly distributed between segments, which would then benefit all modification methods.

### DIFF
--- a/cpp/arcticdb/entity/native_tensor.hpp
+++ b/cpp/arcticdb/entity/native_tensor.hpp
@@ -241,22 +241,34 @@ struct TypedTensor : public NativeTensor {
         // The new column shape * the column stride tells us how far to move the data pointer from the origin
 
         ptr = reinterpret_cast<const uint8_t*>(tensor.data()) + (slice_num * stride_offset);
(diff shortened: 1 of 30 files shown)


alexowens90:
Superseded by https://github.com/man-group/ArcticDB/pull/3208

PR #3160, opened 2026-06-08
Banning backslash in azure storage

phoebusm:
#### Reference Issues/PRs

https://man312219.monday.com/boards/7852509418/pulses/12053019058
Highlight:
1. Rename and update `do_is_path_valid` to `do_is_library_path_valid` as its sole purpose is to validate library path
2. Add new `is_path_valid` function for storage specific key's path, global validation. Currently only `azure` storage uses it
* This is the part which backslash validation is added
* Some pumping is done to make the check azure backend specific
3. Remove `is_symbol_key_valid` python bindings. It is added recently but not used in our codebase nor in any internal repo
4. Consolidate `verify_symbol_key` and `verify_snapshot_id` as they are essentially doing the same check
#### Any other comments?
There is an existing check for library path for **all** storage backends against a list of unsupported s3 characters (`UNSUPPORTED_S3_CHARS`). I leave it as it is since the check is in place for 2+ years
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/storage/mongo/mongo_storage.cpp
+++ b/cpp/arcticdb/storage/mongo/mongo_storage.cpp
@@ -19,8 +19,6 @@
 
 namespace arcticdb::storage::mongo {
 
-const auto UNSUPPORTED_MONGO_CHARS = std::unordered_set<char>{'/'};
-
 std::string MongoStorage::collection_name(KeyType k) { return (fmt::format("{}{}", prefix_, k)); }
 
 /*
@@ -201,8 +199,12 @@ bool MongoStorage::do_iterate_type_until_match(
     return false;
 }
 
-bool MongoStorage::do_is_path_valid(std::string_view path) const {
-    return std::none_of(path.cbegin(), path.cend(), [](auto c) { return UNSUPPORTED_MONGO_CHARS.contains(c); });
+std::optional<char> MongoStorage::do_is_library_path_valid(std::string_view path) const {
+    if (auto unsupported = is_path_valid(path)) {
+        return unsupported;
+    }
+    const auto pos = path.find('/');
+    return pos == std::string_view::npos ? std::nullopt : std::optional<char>{path[pos]};
 }
 
 bool MongoStorage::do_key_exists(const VariantKey& key) {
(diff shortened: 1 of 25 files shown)


cpp/arcticdb/util/name_validation.cpp
```diff
@@ -101,42 +101,51 @@ const auto UNSUPPORTED_S3_CHARS = std::set<char>{'*', '<', '>'};
     return std::monostate{};
 }
 
-CheckOutcome verify_symbol_key(const StreamId& symbol_key) {
+[[nodiscard]] CheckOutcome verify_name_for_backend(
+        const std::string& name_type_for_error, const VariantId& id, const std::shared_ptr<Store>& store
```
G-D-Petrov, line 105: nit: passing the whole store just to print the name seems weird.
Consider passing just the reference to the store name.

cpp/arcticdb/storage/mongo/mongo_storage.cpp
```diff
@@ -201,8 +199,12 @@ bool MongoStorage::do_iterate_type_until_match(
     return false;
 }
 
-bool MongoStorage::do_is_path_valid(std::string_view path) const {
-    return std::none_of(path.cbegin(), path.cend(), [](auto c) { return UNSUPPORTED_MONGO_CHARS.contains(c); });
+std::optional<char> MongoStorage::do_is_library_path_valid(std::string_view path) const {
+    if (auto unsupported = is_path_valid(path)) {
+        return unsupported;
+    }
+    const auto pos = path.find('/');
```
G-D-Petrov, line 208: nit: would be good to have a comment on why this character is not supported on Mongo, like we have for the `lmdb` and `azure` characters above.

G-D-Petrov (approved):

cpp/arcticdb/util/name_validation.cpp
phoebusm: It calls `store->is_path_valid()`, which ultimately calls storage specific `AzureStorage::do_is_path_valid`. I prefer my approach as the corresponding limitation is storage specific so it is better to make it stay in corresponding storage class, to make the approach more OO.

cpp/arcticdb/storage/mongo/mongo_storage.cpp
phoebusm: Added

PR #3165, opened 2026-06-10
Reduce symbol list peak heap usage

G-D-Petrov:
## 1. 63ebc8b3 - Change load approach to stream the entries rather than keeping all of them

**What it does:**

Replaces the two-pass approach (`get_all_symbol_list_keys` + `load_journal_keys`) with a single
streaming pass (`load_journal_streaming`). Instead of materialising all N `AtomKey`s into a
`vector<AtomKey>` and then building the update map from it, the streaming pass builds the
`MapType` (symbol → `vector<SymbolEntryData>`) directly during iteration.

**Result:**

Scan phase no longer requires a temporary `vector<AtomKey>`. The update map is built in one
iteration rather than two. The `all_keys` list (compaction path only) still holds N×AtomKey
(~160 B each), so peak memory during compaction is unchanged at this stage.

---

## 2. 0760d950 - Refactor the SL update map to store a more compacted version of the SL keys

**What it does:**

The main memory reduction commit. Replaces all `SymbolEntryData` / `MapType` / `vector<AtomKey>`
storage for journal entries with a single 32-byte `JournalEntryData` struct and unified
`JournalMapType`.

Key changes:

- **`JournalEntryData`** (32 B): stores `key_version_id`, `creation_ts`, `content_hash`,
  `action`, `is_new_style` — enough to reconstruct the full `AtomKey` for deletion without
  holding the symbol string (the map key provides that)
- **Unified map**: `MapType` (symbol → `vector<SymbolEntryData>`) replaced by `JournalMapType`
  (symbol → `vector<JournalEntryData>`) on all paths; removes the `collect_keys` boolean and
  its branching throughout `load_journal_streaming` and `attempt_load`
- **`LoadResult`**: `symbol_list_keys_: vector<VariantKey>` (N×~160 B) replaced by
  `update_map_: JournalMapType` (N×32 B) + `old_compaction_keys_: vector<VariantKey>` (0–1 entries)
- **Lazy batch deletion** in `compact_internal`: reconstructs `AtomKey`s from `JournalEntryData`
  in batches of 10 000, erasing each symbol's entry after dispatch to free memory incrementally
- **Removed dead code**: `key_sort_comparator`, `add_update_map_entry`, `sort_update_map_entries`,
  and the old `merge_existing_with_journal_map` (MapType version); `merge_existing_with_journal_map`
  is now the single merge function operating on the `JournalMapType`

**Result (measured with `BM_symbol_list_compaction` release benchmarks):**

| Scenario | Before | After | Reduction |
|----------|--------|-------|-----------|
| 1 000 symbols × 1 000 versions (1 M entries) | 68.4 MB | 33.5 MB | −51 % |
| 10 000 symbols × 100 versions (1 M entries)  | 76.5 MB | 43.7 MB | −43 % |

The 1K×1K peak of 33.5 MB is close to the theoretical N×32 B = 32 MB floor (small overhead
from `unordered_map` buckets and the `seen_in_existing` set in the merge path).

---

## 4. bbc5f3db - Add C++ benchmarks

**What it does:**

Adds `cpp/arcticdb/version/test/benchmark_symbol_list.cpp` with the
`BM_symbol_list_compaction` Google Benchmark suite. The benchmark:

- Writes N symbols × V versions of journal entries to an in-memory LMDB store
- Runs `list_symbols` (which triggers compaction) and records wall time and peak RSS memory
- Parameterised over `(N_symbols, N_versions)` pairs: (100×100), (1K×100), (1K×1K), (10K×10),
  (10K×100), plus an S3-mock variant for (1K×100) - most are commented to not interfere with the CI but still be usable for local benchmarking

### DIFF
--- a/cpp/arcticdb/version/symbol_list.cpp
+++ b/cpp/arcticdb/version/symbol_list.cpp
@@ -799,7 +799,7 @@ void delete_keys(const std::shared_ptr<Store>& store, std::vector<AtomKey>&& rem
         // Corner case: if the newly written Compaction key (exclude) has the same timestamp as an existing one
         // (e.g. when a previous compaction round failed in the deletion step), we don't want to delete the former
         if (atom_key != exclude)
-            to_remove.emplace_back(std::move(atom_key));
+            variant_keys.emplace_back(std::move(atom_key));
     }
 
     store->remove_keys_sync(variant_keys);
--- a/cpp/arcticdb/version/symbol_list.cpp
+++ b/cpp/arcticdb/version/symbol_list.cpp
@@ -91,59 +91,6 @@ std::vector<SymbolListEntry> load_previous_from_version_keys(
     return symbols;
 }
 
-std::vector<AtomKey> get_all_symbol_list_keys(
-        const std::shared_ptr<StreamSource>& store, SymbolListData& data, WillAttemptCompaction will_attempt_compaction
-) {
-    std::vector<AtomKey> output;
-    uint64_t uncompacted_keys_found = 0;
-    store->iterate_type(
-            KeyType::SYMBOL_LIST,
-            [&data, &output, &uncompacted_keys_found, will_attempt_compaction](auto&& key) {
-                auto atom_key = to_atom(std::forward<decltype(key)>(key));
-                if (atom_key.id() != compaction_id) {
-                    uncompacted_keys_found++;
-                }
-                if (uncompacted_keys_found == warning_threshold() && !data.warned_expected_slowdown_) {
-                    log::symbol().warn(
-                            "`list_symbols` may take longer than expected as there have been many modifications "
-                            "since `list_symbols` was last called. \n\n"
-                            "See here for more information: "
-                            "https://docs.arcticdb.io/latest/technical/on_disk_storage/#symbol-list-caching\n\n"
-                            "To resolve, run `list_symbols` through to completion frequently. "
-                            "Note: write access to storage is required for compaction. "
-                            "{}.\n"
-                            "Note: This warning will only appear once.\n",
-                            will_attempt_compaction
-                    );
-
-                    data.warned_expected_slowdown_ = true;
-                }
-
-                output.push_back(atom_key);
-            }
-    );
-
-    std::sort(output.begin(), output.end(), [](const AtomKey& left, const AtomKey& right) {
-        // Some very old symbol list keys have a non-zero version number, but with different semantics to the new style,
-        // so ignore it. See arcticdb-man#116. Most old style symbol list keys have version ID 0 anyway.
-        auto left_version = is_new_style_key(left) ? left.version_id() : 0;
-        auto right_version = is_new_style_key(right) ? right.version_id() : 0;
-        return std::tie(left.start_index(), left_version, left.creation_ts()) <
-               std::tie(right.start_index(), right_version, right.creation_ts());
-    });
-    return output;
-}
-
-MaybeCompaction last_compaction(const std::vector<AtomKey>& keys) {
-    auto pos = std::find_if(keys.rbegin(), keys.rend(), [](const auto& key) { return key.id() == compaction_id; });
-
-    if (pos == keys.rend()) {
-        return std::nullopt;
-    } else {
-        return {(pos + 1).base()}; // reverse_iterator -> forward itr has an offset of 1 per docs
-    }
-}
-
 // The below string_at and scalar_at functions should be used for symbol list cache segments instead of the ones
 // provided in SegmentInMemory, because the symbol list structure is the only place where columns can have more entries
 // than the segment has rows. Hence, we need to bypass the checks inside SegmentInMemory's function and directly call
@@ -269,20 +216,72 @@ std::vector<SymbolListEntry> read_from_storage(const std::shared_ptr<StreamSourc
         return read_new_style_list_from_storage(seg);
 }
 
-MapType load_journal_keys(const std::vector<AtomKey>& keys) {
-    MapType map;
-    for (const auto& key : keys) {
-        const auto& action_id = key.id();
-        if (action_id == compaction_id)
-            continue;
+void add_update_map_entry(MapType& update_map, const AtomKey& key) {
+    const auto& symbol = key.start_index();
+    const auto version_id = is_new_style_key(key) ? key.version_id() : unknown_version_id;
+    const auto ts = key.creation_ts();
+    ActionType action = std::get<StringId>(key.id()) == DeleteSymbol ? ActionType::DELETE : ActionType::ADD;
+    update_map[symbol].emplace_back(version_id, ts, action);
+}
 
-        const auto& symbol = key.start_index();
-        const auto version_id = is_new_style_key(key) ? key.version_id() : unknown_version_id;
-        const auto timestamp = key.creation_ts();
-        ActionType action = std::get<StringId>(action_id) == DeleteSymbol ? ActionType::DELETE : ActionType::ADD;
-        map[symbol].emplace_back(version_id, timestamp, action);
+void sort_update_map_entries(MapType& update_map) {
+    for (auto& [symbol, entries] : update_map) {
+        std::sort(entries.begin(), entries.end(), [](const SymbolEntryData& a, const SymbolEntryData& b) {
+            auto a_ver = a.reference_id_ == unknown_version_id ? VersionId{0} : a.reference_id_;
+            auto b_ver = b.reference_id_ == unknown_version_id ? VersionId{0} : b.reference_id_;
+            return std::tie(a_ver, a.timestamp_) < std::tie(b_ver, b.timestamp_);
+        });
     }
-    return map;
+}
+
+auto key_sort_comparator = [](const AtomKey& l, const AtomKey& r) {
+    auto l_ver = is_new_style_key(l) ? l.version_id() : VersionId{0};
+    auto r_ver = is_new_style_key(r) ? r.version_id() : VersionId{0};
+    return std::tie(l.start_index(), l_ver, l.creation_ts()) < std::tie(r.start_index(), r_ver, r.creation_ts());
+};
+
+/// Single-pass iteration over SYMBOL_LIST keys: builds the update map directly and locates the
+/// latest compaction key, optionally collecting all keys for later deletion.
+JournalResult load_journal_streaming(
+        const std::shared_ptr<Store>& store, SymbolListData& data, WillAttemptCompaction will_attempt_compaction,
+        bool collect_keys
+) {
+    JournalResult result;
+    size_t uncompacted_keys_found = 0;
+
+    store->iterate_type(KeyType::SYMBOL_LIST, [&](auto&& key) {
+        auto atom_key = to_atom(std::forward<decltype(key)>(key));
+        result.total_key_count++;
+
+        if (atom_key.id() == compaction_id) {
+            if (!result.compaction_key || atom_key.creation_ts() > result.compaction_key->creation_ts())
+                result.compaction_key = atom_key;
+        } else {
+            ++uncompacted_keys_found;
+            if (uncompacted_keys_found == warning_threshold() && !data.warned_expected_slowdown_) {
+                log::symbol().warn(
+                        "`list_symbols` may take longer than expected as there have been many modifications "
+                        "since `list_symbols` was last called. \n\n"
+                        "See here for more information: "
+                        "https://docs.arcticdb.io/latest/technical/on_disk_storage/#symbol-list-caching\n\n"
+                        "To resolve, run `list_symbols` through to completion frequently. "
+                        "Note: write access to storage is required for compaction. "
+                        "{}.\n"
+                        "Note: This warning will only appear once.\n",
+                        will_attempt_compaction
+                );
+                data.warned_expected_slowdown_ = true;
+            }
+            add_update_map_entry(result.update_map, atom_key);
+        }
+
+        if (collect_keys)
+            result.all_keys.emplace_back(std::move(atom_key));
+    });
+
+    sort_update_map_entries(result.update_map);
+
+    return result;
 }
 
 auto tail_range(const std::vector<SymbolEntryData>& updated) {
@@ -393,12 +392,11 @@ ProblematicResult is_problematic(
     return ProblematicResult{latest};
 }
 
-CollectionType merge_existing_with_journal_keys(
-        const std::shared_ptr<VersionMap>& version_map, const std::shared_ptr<Store>& store,
-        const std::vector<AtomKey>& keys, std::vector<SymbolListEntry>&& existing
+CollectionType merge_existing_with_journal_map(
+        const std::shared_ptr<VersionMap>& version_map, const std::shared_ptr<Store>& store, MapType& update_map,
+        std::vector<SymbolListEntry>&& existing
 ) {
     auto existing_keys = std::move(existing);
-    auto update_map = load_journal_keys(keys);
 
     CollectionType symbols;
     std::map<StreamId, std::pair<VersionId, timestamp>> problematic_symbols;
@@ -489,55 +487,46 @@ CollectionType merge_existing_with_journal_keys(
     return symbols;
 }
 
-CollectionType load_from_symbol_list_keys(
-        const std::shared_ptr<VersionMap>& version_map, const std::shared_ptr<Store>& store,
-        const std::vector<AtomKey>& keys, const Compaction& compaction
-) {
-    ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Loading symbols from symbol list keys");
-
-    auto previous_compaction = read_from_storage(store, *compaction);
-    return merge_existing_with_journal_keys(version_map, store, keys, std::move(previous_compaction));
-}
-
-CollectionType load_from_version_keys(
-        const std::shared_ptr<VersionMap>& version_map, const std::shared_ptr<Store>& store,
-        const std::vector<AtomKey>& keys, SymbolListData& data, WillAttemptCompaction will_attempt_compaction
-) {
-    ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Loading symbols from version keys");
-    auto previous_entries = load_previous_from_version_keys(store, data, will_attempt_compaction);
-    return merge_existing_with_journal_keys(version_map, store, keys, std::move(previous_entries));
-}
-
 LoadResult attempt_load(
         const std::shared_ptr<VersionMap>& version_map, const std::shared_ptr<Store>& store, SymbolListData& data,
         WillAttemptCompaction will_attempt_compaction
 ) {
     ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Symbol list load attempt");
+    const bool collect_keys = will_attempt_compaction == WillAttemptCompaction::YES;
+    auto journal = load_journal_streaming(store, data, will_attempt_compaction, collect_keys);
+
     LoadResult load_result;
-    load_result.symbol_list_keys_ = get_all_symbol_list_keys(store, data, will_attempt_compaction);
-    load_result.maybe_previous_compaction = last_compaction(load_result.symbol_list_keys_);
+    load_result.compaction_key_ = journal.compaction_key;
+    load_result.total_key_count_ = journal.total_key_count;
+    load_result.symbol_list_keys_ = std::move(journal.all_keys);
 
-    if (load_result.maybe_previous_compaction)
-        load_result.symbols_ = load_from_symbol_list_keys(
-                version_map, store, load_result.symbol_list_keys_, *load_result.maybe_previous_compaction
-        );
-    else {
-        load_result.symbols_ = load_from_version_keys(
-                version_map, store, load_result.symbol_list_keys_, data, will_attempt_compaction
-        );
-        std::unordered_set<StreamId> keys_in_versions;
-        for (const auto& entry : load_result.symbols_)
-            keys_in_versions.emplace(entry.stream_id_);
+    if (journal.compaction_key) {
+        ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Loading symbols from symbol list keys");
+        auto existing = read_from_storage(store, *journal.compaction_key);
+        load_result.symbols_ =
+                merge_existing_with_journal_map(version_map, store, journal.update_map, std::move(existing));
+    } else {
+        ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Loading symbols from version keys");
+        auto previous_entries = load_previous_from_version_keys(store, data, will_attempt_compaction);
+        load_result.symbols_ =
+                merge_existing_with_journal_map(version_map, store, journal.update_map, std::move(previous_entries));
 
-        for (const auto& key : load_result.symbol_list_keys_)
-            util::check(
-                    keys_in_versions.find(StreamId{std::get<StringIndex>(key.start_index())}) != keys_in_versions.end(),
-                    "Would delete unseen key {}",
-                    key
-            );
+        // Verify every journal key we'd delete during compaction corresponds to a symbol in the
+        // merged output. Guards against silent data loss from merge bugs.
+        if (collect_keys) {
+            std::unordered_set<StreamId> symbols_in_merge;
+            for (const auto& entry : load_result.symbols_)
+                symbols_in_merge.emplace(entry.stream_id_);
+
+            for (const auto& key : load_result.symbol_list_keys_)
+                util::check(
+                        symbols_in_merge.count(StreamId{std::get<StringIndex>(to_atom(key).start_index())}) > 0,
+                        "Would delete unseen key {}",
+                        key
+                );
+        }
     }
 
-    load_result.timestamp_ = store->current_timestamp();
     return load_result;
 }
 
@@ -624,12 +613,12 @@ StreamDescriptor delete_symbol_stream_descriptor(const StreamId& stream_id, cons
 }
 
 bool SymbolList::needs_compaction(const LoadResult& load_result) const {
-    if (!load_result.maybe_previous_compaction) {
+    if (!load_result.compaction_key_) {
         log::version().debug("Symbol list: needs_compaction=[true] as no previous compaction");
         return true;
     }
 
-    auto n_keys = static_cast<int64_t>(load_result.symbol_list_keys_.size());
+    auto n_keys = static_cast<int64_t>(load_result.total_key_count_);
     if (auto fixed = ConfigsMap::instance()->get_int("SymbolList.MaxDelta")) {
         auto result = n_keys > *fixed;
         log::version().debug(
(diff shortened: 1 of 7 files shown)


cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -31,13 +42,21 @@ enum class WillAttemptCompaction : uint8_t {
...
 };
 
+struct JournalResult {
+    std::optional<AtomKey> compaction_key;
+    JournalMapType update_map; // JournalEntryData (32B/entry) for all journal entries
+    size_t total_key_count = 0;
+    std::vector<VariantKey> compaction_keys; // VariantKeys of all compaction keys found during scan
+};
+
 struct LoadResult {
```
alexowens90, line 78: It feels like this should just have a `JournalResult` as a member, given the overlap and that it is always just constructed from the elements of `JournalResult` anyway?

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -17,12 +17,23 @@
 #include <util/storage_lock.hpp>
 
 namespace arcticdb {
-struct SymbolListEntry;
-struct SymbolEntryData;
+enum class ActionType : uint8_t { ADD, DELETE };
+
+// Compact 32-byte representation of a SYMBOL_LIST journal AtomKey.
```
alexowens90: The size won't be consistent across architectures (but would be 40 for most).
If we really want to push it, we could pack the struct (see `AtomKeyPacked`) down to 26 bytes.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -269,20 +216,90 @@ std::vector<SymbolListEntry> read_from_storage(const std::shared_ptr<StreamSourc
         return read_new_style_list_from_storage(seg);
 }
 
-MapType load_journal_keys(const std::vector<AtomKey>& keys) {
-    MapType map;
-    for (const auto& key : keys) {
-        const auto& action_id = key.id();
-        if (action_id == compaction_id)
-            continue;
+JournalEntryData journal_entry_from_atom(const AtomKey& key) {
```
alexowens90: Could be a constructor?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -393,15 +410,66 @@ ProblematicResult is_problematic(
     return ProblematicResult{latest};
 }
 
-CollectionType merge_existing_with_journal_keys(
+void resolve_problematic_symbols(
```
alexowens90, line 411: Is this just a lift and shift, or are there implementation changes?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -418,126 +486,85 @@ CollectionType merge_existing_with_journal_keys(
                 );
         } else {
             util::check(!updated->second.empty(), "Unexpected empty entry for symbol {}", updated->first);
-            if (auto problematic_entry = is_problematic(previous_entry, updated->second, min_allowed_interval);
+            seen_in_existing.insert(stream_id);
+            std::vector<SymbolEntryData> entries;
+            entries.reserve(updated->second.size());
+            for (const auto& ck : updated->second)
+                entries.push_back(to_symbol_entry_data(ck));
+            if (auto problematic_entry = is_problematic(previous_entry, entries, min_allowed_interval);
```
alexowens90: Couldn't we make `is_problematic` work with `std::vector<JournalEntryData>` and avoid this temporary vector?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -393,15 +410,66 @@ ProblematicResult is_problematic(
...
+CollectionType merge_existing_with_journal_map(
         const std::shared_ptr<VersionMap>& version_map, const std::shared_ptr<Store>& store,
-        const std::vector<AtomKey>& keys, std::vector<SymbolListEntry>&& existing
+        const JournalMapType& update_map, std::vector<SymbolListEntry>&& existing
 ) {
     auto existing_keys = std::move(existing);
-    auto update_map = load_journal_keys(keys);
 
     CollectionType symbols;
     std::map<StreamId, std::pair<VersionId, timestamp>> problematic_symbols;
```
alexowens90: Do you know why this is ordered? After we process it we just have to sort the `symbols` again anyway

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -418,126 +486,85 @@ CollectionType merge_existing_with_journal_keys(
...
+                symbols.emplace_back(updated->first, last.reference_id_, last.timestamp_, last.action_);
             }
-            update_map.erase(updated);
         }
     }
 
-    for (const auto& [symbol, entries] : update_map) {
-        ARCTICDB_DEBUG(log::symbol(), "{} {}", symbol, entries);
+    for (const auto& [symbol, ck_entries] : update_map) {
+        if (seen_in_existing.count(symbol) > 0)
```
alexowens90: nit: `contains`

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -418,126 +486,85 @@ CollectionType merge_existing_with_journal_keys(
...
+            continue;
+        std::vector<SymbolEntryData> entries;
+        entries.reserve(ck_entries.size());
+        for (const auto& ck : ck_entries)
+            entries.push_back(to_symbol_entry_data(ck));
         if (auto problematic_entry = is_problematic(entries, min_allowed_interval); problematic_entry) {
```
alexowens90: Same comment https://github.com/man-group/ArcticDB/pull/3165/changes#r3475315188

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -418,126 +486,85 @@ CollectionType merge_existing_with_journal_keys(
...
     LoadResult load_result;
-    load_result.symbol_list_keys_ = get_all_symbol_list_keys(store, data, will_attempt_compaction);
-    load_result.maybe_previous_compaction = last_compaction(load_result.symbol_list_keys_);
+    load_result.compaction_key_ = journal.compaction_key;
+    load_result.total_key_count_ = journal.total_key_count;
+
+    if (journal.compaction_key) {
+        ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Loading symbols from symbol list keys");
+        auto existing = read_from_storage(store, *journal.compaction_key);
+        if (journal.update_map.empty()) {
```
alexowens90: Why this special case when there was a previous compaction but not when loading from version chains?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -418,126 +486,85 @@ CollectionType merge_existing_with_journal_keys(
...
+        }
+    } else {
+        ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Loading symbols from version keys");
+        auto previous_entries = load_previous_from_version_keys(store, data, will_attempt_compaction);
+        load_result.symbols_ =
+                merge_existing_with_journal_map(version_map, store, journal.update_map, std::move(previous_entries));
+        if (will_compact) {
+            // Verify every journal symbol we'd delete corresponds to a symbol in the merged output.
+            // Guards against silent data loss from merge bugs.
+            std::unordered_set<StreamId> symbols_in_merge;
```
alexowens90: Will `symbols_in_merge.size() == load_result.symbols_.size()`?
If so, could use `ankerl` and reserve the size up front, will involve way fewer allocations

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -744,7 +771,7 @@ SegmentInMemory create_empty_segment(const StreamId& stream_id) {
 }
 
 VariantKey write_symbols(
-        const std::shared_ptr<Store>& store, const CollectionType& symbols, const StreamId& stream_id,
+        const std::shared_ptr<Store>& store, CollectionType symbols, const StreamId& stream_id,
```
alexowens90: `CollectionType&&` would make the intent clearer

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -830,18 +856,52 @@ size_t SymbolList::compact(const std::shared_ptr<Store>& store) {
...
+    auto written = write_symbols(store, std::move(load_result.symbols_), compaction_id, data_.type_holder_);
+    const auto& written_key = std::get<AtomKey>(written);
+
+    // Delete old compaction keys (typically 0–1 entries; exclude the newly written one).
+    auto& old_ck = load_result.old_compaction_keys_;
+    old_ck.erase(
+            std::remove_if(
+                    old_ck.begin(),
+                    old_ck.end(),
+                    [&written_key](const VariantKey& vk) { return to_atom(vk) == written_key; }
```
alexowens90: When could this be true? We loaded the old compaction keys prior to writing `written_key`

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -830,18 +856,52 @@ size_t SymbolList::compact(const std::shared_ptr<Store>& store) {
...
         // legacy arcticc symbol list entries don't get correctly listed when doing `iterate_type`, so can mess
         // up racing symbol list compaction detection.
         ARCTICDB_RUNTIME_DEBUG(
                 log::symbol(),
                 "Symbol list compaction will be skipped: either a concurrent compaction was detected "
                 "or there are legacy arcticc symbol list entries that cannot be verified."
         );
-    } else {
-        auto written = write_symbols(store, load_result.symbols_, compaction_id, data_.type_holder_);
-        delete_keys(store, load_result.detach_symbol_list_keys(), std::get<AtomKey>(written));
```
alexowens90, line 843: I get that the API will be different now than the version needed by the compat test, but I think having the code below `write_symbols` in it's own method would make things clearer.

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -830,18 +856,52 @@ size_t SymbolList::compact(const std::shared_ptr<Store>& store) {
...
+    old_ck.erase(
+            std::remove_if(
+                    old_ck.begin(),
+                    old_ck.end(),
+                    [&written_key](const VariantKey& vk) { return to_atom(vk) == written_key; }
+            ),
+            old_ck.end()
+    );
+    if (!old_ck.empty())
+        store->remove_keys_sync(std::move(old_ck));
```
alexowens90: There's probably only one element in here, should be batched up with the keys below

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -830,18 +856,52 @@ size_t SymbolList::compact(const std::shared_ptr<Store>& store) {
...
+    std::vector<VariantKey> batch;
+    batch.reserve(kBatchSize);
+    for (auto it = load_result.update_map_.begin(); it != load_result.update_map_.end();) {
+        const auto& [symbol, ck_entries] = *it;
+        for (const auto& ck : ck_entries) {
+            batch.emplace_back(atom_key_from_journal_entry(symbol, ck));
+            if (batch.size() == kBatchSize) {
+                store->remove_keys_sync(std::move(batch));
+                batch.clear();
+                batch.reserve(kBatchSize);
```
alexowens90: nit: `clear()` explicitly does not change the capacity, so this isn't needed

cpp/arcticdb/version/symbol_list.cpp
alexowens90: And even if not, `ankerl` will allocate less

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -418,126 +486,85 @@ CollectionType merge_existing_with_journal_keys(
...
-}
-
-CollectionType load_from_version_keys(
-        const std::shared_ptr<VersionMap>& version_map, const std::shared_ptr<Store>& store,
-        const std::vector<AtomKey>& keys, SymbolListData& data, WillAttemptCompaction will_attempt_compaction
-) {
-    ARCTICDB_RUNTIME_DEBUG(log::symbol(), "Loading symbols from version keys");
-    auto previous_entries = load_previous_from_version_keys(store, data, will_attempt_compaction);
-    return merge_existing_with_journal_keys(version_map, store, keys, std::move(previous_entries));
-}
```
alexowens90, line 509: I quite liked this structure, is there a reason to remove it?

cpp/arcticdb/version/test/test_symbol_list.cpp
```diff
@@ -1057,3 +1063,95 @@ TEST_P(SymbolListRace, Run) {
...
+        SymbolList::remove_symbol(store_, StreamId{symbol}, 1);
+    }
+
+    ConfigsMap::instance()->unset_int("SymbolList.MaxDelta");
+
+    // Load via compaction-eligible path (will not actually compact since threshold is default)
+    SymbolList sl1{version_map_};
+    auto compaction_result = sl1.load<std::set<StreamId>>(version_map_, store_, false);
+
+    // Load via direct path (no_compaction=true)
```
alexowens90: I would read "direct-path" here to mean from version chains, but that isn't what the `no_compaction` flag controls?

cpp/arcticdb/version/test/test_symbol_list.cpp
```diff
@@ -1057,3 +1063,95 @@ TEST_P(SymbolListRace, Run) {
...
+    // Setup: add symbols and force compaction
+    for (int i = 0; i < 50; ++i) {
+        auto symbol = fmt::format("sym_{}", i);
+        SymbolList::add_symbol(store_, StreamId{symbol}, 0);
+        auto key = atom_key_builder().build(symbol, KeyType::TABLE_INDEX);
+        version_map_->write_version(store_, key, std::nullopt);
+    }
+
+    ConfigsMap::instance()->set_int("SymbolList.MaxDelta", 0);
+    {
```
alexowens90: nit: Why is this scoped?

cpp/arcticdb/version/test/test_symbol_list.cpp
```diff
@@ -1057,3 +1063,95 @@ TEST_P(SymbolListRace, Run) {
...
+    // Add journal entries: new symbols + deletes of existing ones
+    for (int i = 40; i < 60; ++i) {
+        auto symbol = fmt::format("sym_{}", i);
+        SymbolList::add_symbol(store_, StreamId{symbol}, 1);
+        auto key = atom_key_builder().version_id(1).build(symbol, KeyType::TABLE_INDEX);
+        version_map_->write_version(store_, key, std::nullopt);
+    }
+    for (int i = 0; i < 10; ++i) {
+        auto symbol = fmt::format("sym_{}", i);
+        SymbolList::remove_symbol(store_, StreamId{symbol}, 1);
```
alexowens90: The reference id should be 0? IIUC that's the case in all of these tests

cpp/arcticdb/version/test/test_symbol_list.cpp
```diff
@@ -1057,3 +1063,95 @@ TEST_P(SymbolListRace, Run) {
...
+    SymbolList sl1{version_map_};
+    auto compaction_result = sl1.load<std::set<StreamId>>(version_map_, store_, false);
+
+    // Load via direct path (no_compaction=true)
+    SymbolList sl2{version_map_};
+    auto direct_result = sl2.load<std::set<StreamId>>(version_map_, store_, true);
+
+    EXPECT_EQ(compaction_result, direct_result);
+    // 50 original + 10 new = 60. The remove_symbol journal entries are resolved
+    // as ADD because the version map still shows those symbols as existing.
```
alexowens90: Isn't this only true because all the timestamps will be super close together here?

alexowens90 (changes requested):

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -830,18 +856,52 @@ size_t SymbolList::compact(const std::shared_ptr<Store>& store) {
...
-    } else {
-        auto written = write_symbols(store, load_result.symbols_, compaction_id, data_.type_holder_);
-        delete_keys(store, load_result.detach_symbol_list_keys(), std::get<AtomKey>(written));
+        return;
+    }
+
+    auto written = write_symbols(store, std::move(load_result.symbols_), compaction_id, data_.type_holder_);
+    const auto& written_key = std::get<AtomKey>(written);
+
+    // Delete old compaction keys (typically 0–1 entries; exclude the newly written one).
```
poodlewars: Why do we need logic for the exclusion, how could the new one possibly be in the old set?

cpp/arcticdb/version/symbol_list.cpp
```diff
@@ -830,18 +856,52 @@ size_t SymbolList::compact(const std::shared_ptr<Store>& store) {
...
+                    [&written_key](const VariantKey& vk) { return to_atom(vk) == written_key; }
+            ),
+            old_ck.end()
+    );
+    if (!old_ck.empty())
+        store->remove_keys_sync(std::move(old_ck));
+
+    // Reconstruct and delete journal keys in batches from the JournalMapType, freeing each symbol's
+    // entries as they are processed. Journal keys can never equal written_key (different id field).
+    static constexpr size_t kBatchSize = 10'000;
```
poodlewars: This will limit us to 10 concurrent deletes on S3 which could be very limiting, I think num IO threads * (storage delete batch size) would be a better default? Also better to use a config key rather than hard-coding.

cpp/arcticdb/version/test/benchmark_symbol_list.cpp
poodlewars, line 1: I'm not sure how useful this is given it's using a mock (ie, in memory) storage.

What ASV coverage do we have for the symbol list? Do we need to extend it before making changes here?

cpp/arcticdb/version/test/test_symbol_list.cpp
```diff
@@ -1057,3 +1063,95 @@ TEST_P(SymbolListRace, Run) {
 INSTANTIATE_TEST_SUITE_P(SymbolListSource, SymbolListRace, Combine(Values('S'), Bool(), Bool(), Bool()));
 // For version keys source (initial compaction), there's no old compaction key to remove:
 INSTANTIATE_TEST_SUITE_P(VersionKeysSource, SymbolListRace, Combine(Values('V'), Values(false), Bool(), Bool()));
+
+TEST_F(SymbolListSuite, DirectPathEquivalence) {
```
poodlewars: I don't know what `DirectPath` means

cpp/arcticdb/version/test/test_symbol_list.cpp
```diff
@@ -1057,3 +1063,95 @@ TEST_P(SymbolListRace, Run) {
...
+    }
+
+    ConfigsMap::instance()->set_int("SymbolList.MaxDelta", 0);
+    {
+        SymbolList sl{version_map_};
+        sl.load<std::set<StreamId>>(version_map_, store_, false);
+    }
+
+    // Add journal entries: new symbols + deletes of existing ones
+    for (int i = 40; i < 60; ++i) {
```
poodlewars: This overlaps with some of the previous for loop?

I don't really understand what this test is trying to check...

poodlewars (changes requested):

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: We try to load from version chains only when there are no SL entries (entries and/or competed keys), so there is no special case then.

This special case, when we have only a compacted key but no SL entries.
So there is nothing to merge, and we can just parse out the symbol list based on the compacted key.

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: It was easier for me to check the interaction between the 2 when they were laid out flat in the `attemp_load` function. I will bring them back.

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: removed

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: batched

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: Added the method

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: Added a comment about this

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: done

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: refactored to user `ankerl`

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: Refactored to use unsorted set

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: Refactored as suggested

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: Pure lift-and-shift - extracted into its own function with an early return; no logic changes.

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: Refactored

cpp/arcticdb/version/symbol_list.hpp
G-D-Petrov: Refactored as suggested

cpp/arcticdb/version/symbol_list.hpp
G-D-Petrov: Refactored as suggested

cpp/arcticdb/version/test/test_symbol_list.cpp
G-D-Petrov: removed

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: Added a comment about this

cpp/arcticdb/version/symbol_list.cpp
G-D-Petrov: refactored as suggested

cpp/arcticdb/version/test/benchmark_symbol_list.cpp
G-D-Petrov: We do have ASV coverage here but it is not very useful for this particular case because:

1. It measures peak memory during that execution; this is a problem when we want to measure just the memory consumed by `list_symbols` because:
a.  it includes the memory consumption from other sources which makes the baseline memory quite high (~200MB) - https://man-group.github.io/ArcticDB/#list_symbols.ListSymbolsWithoutCache.peakmem_list_symbols
b. it uses lmdb, where doing the listing operation of the SL entries loads ALL of them in memory, which further spikes the memory

2. to get an accurate/non-noisy benchmark of the memory, we need to write many entries (on the scale of millions due to the problems mentioned above), which is quite slow against a real storage (or even a simulated one, if we implement one as we don't have that option at the moment), which will slow down the ASV benchmarks a lot

All of this makes implementing such a benchmark infeasible/undesirable.
I have run a manual ASV run against a real storage with 1 million entries(took multiple hours) and it produced similar results as the ones reported by this benchmark (in terms of raw MB difference, not percentage obv)

cpp/arcticdb/version/symbol_list.hpp
```diff
@@ -17,12 +17,44 @@
...
+
+    // Logical constructor from a (reference id, time, action) triple. content_hash is left zero as it
+    // is not part of the logical identity is_problematic cares about; use the AtomKey constructor when
+    // the entry must round-trip back to a deletable key.
+    JournalEntryData(entity::VersionId reference_id, timestamp creation_time, ActionType action_type) :
+        key_version_id(reference_id),
+        creation_ts(creation_time),
+        content_hash(0),
+        action(action_type),
+        is_new_style(reference_id != unknown_version_id) {}
```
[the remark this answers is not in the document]
G-D-Petrov: Updated in commit b11e109f8a5e44b65c766ff72fee05d875fded6c

cpp/arcticdb/version/test/test_symbol_list.cpp
G-D-Petrov: Both tests are essentially trying to make sure that both the compaction and non-compaction paths produce the same symbol list set.
Both are trying to have a mix of entries for symbols that are already present, new symbols and ones that are deleted.
The difference is that the first test tries to simulate the case with problematic symbols (by just adding delete-symbol list entries, without removing the m from the version chain), and the second one simulates the case with regular deletes + remove entries.

cpp/arcticdb/version/test/test_symbol_list.cpp
G-D-Petrov: Added more meaningful name and comments to explain the intention.

cpp/arcticdb/version/test/test_symbol_list.cpp
poodlewars: I still don't really get this, why are we writing add keys on top of add keys and why do the overlaps use the same reference ID as the new symbols?

cpp/arcticdb/version/test/test_symbol_list.cpp
poodlewars: This hasn't been addressed

poodlewars (dismissed):

cpp/arcticdb/version/test/test_symbol_list.cpp
G-D-Petrov: yes, that is partly the point of the test - to make sure that the is_problematic logic is correct for both paths.

cpp/arcticdb/version/test/test_symbol_list.cpp
G-D-Petrov: We are writing add keys on top of write keys so we can test that the 2 load paths cover the 3 expected cases which are:
1. totally new symbols and add keys for them
2. existing symbols with new versions/add keys for them
3. existing symbols that are deleted (or only have the delete SL entries)

I've changed the tests to use realistic reference IDs.
But it is important to note that the reference IDs don't really matter, as the symbol list logic will either:
1. use them as is, in the general case
2. ignore them and use the version chain, in the `is_problematic` case

P.S.: there is technically a 4th case - delete SL entries for symbols that don't exist.
But this is impossible in practice, and even if it were, the symbol list logic would resolve them to nothing/no-op.

cpp/arcticdb/version/test/test_symbol_list.cpp
G-D-Petrov: I've changed it to 0 but you can check this comment for an explanation on why the test work even with 1 - https://github.com/man-group/ArcticDB/pull/3165#discussion_r3519636431

alexowens90 (approved):

PR #3167, opened 2026-06-11
Adapt to msgpack 1.2.0

phoebusm:
#### Reference Issues/PRs

Fix `test_deep_nesting_metastruct_size_over_limit` and `test_deep_nesting_metastruct_size_under_limit`
Culprit: `msgpack 1.2.0` has just released. It includes [Raise DEFAULT_RECURSE_LIMIT from 511 to 1024 by Copilot · Pull Request #676 · msgpack/msgpack-python](https://github.com/msgpack/msgpack-python/pull/676)
which changes `DEFAULT_RECURSE_LIMIT` from 511 to 1024
`python 3.9` is unaffected as the new msgpack version has dropped the support

Hard-pin the arcticdb limit to `511` as msgpack <1.2.0
#### Any other comments?
msgpack <1.2.0 has the recurse limit set at 511. arcticdb imports the limit from msgpack to calculate the metastruct layer limit
msgpack 1.2.0 has raised limit from 511 to 1024. Since reconstructing the metastruct from is done by recursion, increasing the no. of layers in the metastruct will make arcticdb hit the python recursion limit.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/python/arcticdb/flattener.py
+++ b/python/arcticdb/flattener.py
@@ -27,13 +27,16 @@ from arcticdb.version_store._normalization import MsgPackNormalizer, CompositeNo
 from arcticdb.preconditions import check
 from arcticdb_ext import get_config_int
 
-# The binding constraint is the read path: reconstructing the data recurses ~3 Python stack frames
-# per nesting level (see create_original_obj_from_metastruct), so with the default recursion limit of
-# 1000 reads fail above ~330 levels. We pin the cap at 255 - the value it was effectively fixed at for
-# years, when msgpack's DEFAULT_RECURSE_LIMIT was 511 (511 // 2 == 255) - which both packs on write and
-# round-trips on read. The min() keeps msgpack's packer as a ceiling too. Tying the cap directly to
-# DEFAULT_RECURSE_LIMIT broke when msgpack 1.2.0 raised it to 1024, letting writes through at depths the
-# reader could not reconstruct.
+# Two separate limits govern maximum nesting depth:
+#   Write (msgpack packer): uses ~2 Python stack frames per level; capped at DEFAULT_RECURSE_LIMIT // 2.
+#   Read (create_original_obj_from_metastruct): uses ~3 Python stack frames per level, so reads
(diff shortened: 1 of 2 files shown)


python/arcticdb/flattener.py
```diff
@@ -27,6 +27,16 @@
...
 
+# The binding constraint is the read path: reconstructing the data recurses ~3 Python stack frames
+# per nesting level (see create_original_obj_from_metastruct), so with the default recursion limit of
+# 1000 reads fail above ~330 levels. We pin the cap at 255 - the value it was effectively fixed at for
+# years, when msgpack's DEFAULT_RECURSE_LIMIT was 511 (511 // 2 == 255) - which both packs on write and
+# round-trips on read. The min() keeps msgpack's packer as a ceiling too. Tying the cap directly to
+# DEFAULT_RECURSE_LIMIT broke when msgpack 1.2.0 raised it to 1024, letting writes through at depths the
+# reader could not reconstruct.
+# https://man312219.monday.com/boards/7852509418/pulses/12254825163
+DEFAULT_RECURSE_LIMIT = min(511, _DEFAULT_RECURSE_LIMIT)
```
IvoDD, line 41: I find this explanation confusing. The recursion limit has always been 1000

As far as I understand the problem is that our metastruct reader uses 3 stackframe recursion so we reach the limit at around 330.

Should we just set the threshold for nesting to something like `(DEFAULT_RECURSION_LIMIT - 10) // 3`?  The `10` is for the stack frames which call the decoding code.

After we fix our metastruct reading to not be recursive we will be able to return it to the `(DEFAULT_RECURSION_LIMIT - 10) // 2`

IvoDD: Anyway I'm fine with the current fix to unblock the CI and it essentially preserves the old limit.

IvoDD (approved):
Can you only add `no-release-notes` label as this is not user facing.

python/arcticdb/flattener.py
phoebusm: I have updated the comment to better explain the write and read limits

IvoDD (approved):

PR #3174, opened 2026-06-15
Refactor InputFrame to keep a variant of InputTensors and Columns

ognyanstoimenov:
#### Reference Issues/PRs

Monday ref. 12217223987

* Refactor InputFrame to keep a variant of InputTensors and Columns
* Rework `WriteToSegmentTask`'s `slice_tensor` and `slice_segment` methods into a single `slice()` to incorporate the variant
* Other reworks around that code
  * Including removing the Aggregator from the `slice()` by exposing `set_*` functions that work on SegmentsInMemory instead of Aggregators

This is done in preparation to handle the arrow-backed string `dtype` introduced in pandas 3.
The actual string dtype handling will be done in a following PR.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


ognyanstoimenov:
One benchmark is failing, will take look:
```
| +        | 55.1±3ms                         | 79.9±0.7ms          |    1.45 | compact_data.CompactDataNumericStaticSchema.time_compact_data((1000000, 10000, 100000), 2, False) |
```

cpp/arcticdb/pipeline/input_frame.hpp
```diff
@@ -85,29 +99,14 @@ struct InputFrame {
...
-                if (field.type().data_type() == DataType::UTF_DYNAMIC32) {
-                    field.mutable_type() = TypeDescriptor(DataType::UTF_DYNAMIC64, field.type().dimension());
-                }
-            }
-        }
-    };
-    std::variant<InputTensors, InputSegment> input_data;
+    std::vector<FieldData> columns_;
+    std::vector<sparrow::record_batch> arrow_buffer_owners_;
+    std::optional<NativeTensor> index_tensor_;
```
alexowens90: I would just make this a bool saying whether or not there is an index column, and then everywhere else we can assume that the first element of `columns_` is the index if this is `true`

cpp/arcticdb/pipeline/input_frame.hpp
```diff
@@ -85,29 +99,14 @@ struct InputFrame {
...
+    StreamDescriptor desc_;
+    std::optional<StreamDescriptor> desc_for_tsd_;
+
+    bool has_only_tensors_{true};
+    bool has_only_arrow_columns_{true};
```
alexowens90: I haven't seen how these are used yet, but a 3 element enum seems more appropriate? Only tensors, only arrow, mixture?

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -10,52 +10,55 @@
...
     }
-    input_data.emplace<InputSegment>(std::move(seg), std::move(arrow_buffer_owners));
-}
 
-StreamDescriptor& InputFrame::desc() {
-    if (has_tensors()) {
-        return std::get<InputTensors>(input_data).desc;
-    } else {
-        return std::get<InputSegment>(input_data).seg.descriptor();
+    num_rows = cols.empty() ? 0 : cols[0].row_count();
```
alexowens90, line 45: I don't think this will be correct if the first column contains strings, as the buffer then has 1 more element than there are rows

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -111,23 +119,27 @@ void InputFrame::set_index_range() {
...
+const InputFrame::FieldData& InputFrame::field_data(size_t idx) const {
+    util::check(idx < columns_.size(), "InputFrame::field_data index {} out of range (size {})", idx, columns_.size());
+    return columns_[idx];
 }
 
-const std::vector<entity::NativeTensor>& InputFrame::field_tensors() const {
-    util::check(has_tensors(), "InputFrame field_tensors requested but holds SegmentInMemory");
-    return std::get<InputTensors>(input_data).field_tensors;
+const NativeTensor& InputFrame::get_tensor(size_t idx) const {
+    util::check(idx < columns_.size(), "InputFrame::get_tensor index {} out of range (size {})", idx, columns_.size());
```
alexowens90: Should also check that `columns_[idx]` actually contains a `NativeTensor` (and similarly in `get_column`)

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -420,31 +421,27 @@ DataType arcticdb_type_from_arrow_array(const sparrow::array& array) {
...
             !has_index || first_batch.nb_columns() > 0, "Cannot use index_column=True on a table with no columns"
     );
-    uint64_t total_rows = std::accumulate(
-            record_batches.cbegin(),
-            record_batches.cend(),
-            uint64_t(0),
-            [](const uint64_t& accum, const sparrow::record_batch& record_batch) {
-                return accum + record_batch.nb_rows();
-            }
-    );
```
alexowens90, line 439: Given that we still need this value (see comment in `input_frame.cpp`) it might make sense to still return a segment from here, pass it into `InputFrame::set_segment`, and then extract the row count, columns vector, and descriptor in that method

cpp/arcticdb/pipeline/frame_utils.hpp
```diff
@@ -182,24 +183,24 @@ std::optional<convert::StringEncodingError> set_sequence_type(
...
-        AggregatorType& agg, const entity::NativeTensor& tensor, size_t col, size_t rows_to_write, size_t row,
+        SegmentInMemory& seg, const entity::NativeTensor& tensor, size_t col, size_t rows_to_write, size_t row,
         size_t slice_num, size_t regular_slice_size, bool sparsify_floats
 ) {
     constexpr auto dt = TagType::DataTypeTag::data_type;
     auto ptr = tensor.template ptr_cast<RawType>(row);
     if (sparsify_floats) {
         if constexpr (is_floating_point_type(dt)) {
-            agg.set_sparse_block(col, ptr, rows_to_write);
+            seg.set_sparse_block(col, ptr, rows_to_write);
```
alexowens90, line 195: I take it the previous implementation just had the aggregator call `set_*` on it's contained segment for all of these calls?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -450,19 +450,21 @@ std::vector<EntityId> MergeUpdateClause::process(std::vector<EntityId>&& entity_
...
         return result;
     }();
+    if (!source_->has_only_tensors()) {
+        user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>("Arrow format is not supported as input for merge update"
+        );
+    }
+    std::vector<NativeTensor> source_tensors;
+    source_tensors.reserve(source_->num_columns());
+    for (size_t i = 0; i < source_->num_columns(); ++i) {
+        source_tensors.push_back(source_->get_tensor(i));
```
alexowens90, line 463: Presumably this copies pointers rather than the underlying data?

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -343,78 +318,76 @@ Column WriteToSegmentTask::slice_column(
...
+        const auto col_idx = slice_.absolute_field_col(col) + columns_index_offset;
+        util::variant_match(
+                frame_->field_data(col_idx),
+                [&](const NativeTensor& tensor) { add_tensor_column(abs_col, tensor, fd, sparsify_floats_); },
+                [&](const Column& source_column) { add_arrow_column(source_column, fd); }
+        );
+    }
 
-        agg.end_block_write(rows_to_write);
+    seg.end_block_write(rows_to_write);
```
alexowens90, line 384: Does this call `set_row_data`?

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -323,7 +296,9 @@ Column WriteToSegmentTask::slice_column(
         const auto& slice = contiguous_slices.front();
         ChunkedBuffer chunked_buffer;
         chunked_buffer.add_external_block(slice.block->ptr(slice.start_pos * type_size), slice.size * type_size);
-        return {source_column.type(), Sparsity::NOT_PERMITTED, std::move(chunked_buffer)};
+        Column dest{source_column.type(), Sparsity::NOT_PERMITTED, std::move(chunked_buffer)};
+        dest.set_row_data(slice_.rows().diff() - 1);
```
alexowens90, line 300: I think this is unnecessary, we should just be calling `set_row_data` on the whole segment at the end

cpp/arcticdb/pipeline/input_frame.hpp
alexowens90: OK it seems like we just need `has_only_tensors_` for operations that don't support Arrow yet, and we'll be able to get rid of them when everything does

cpp/arcticdb/pipeline/frame_utils.hpp
ognyanstoimenov: Yes

cpp/arcticdb/arrow/arrow_utils.cpp
ognyanstoimenov: I actually did this on purpose because I find it easier to follow what is going on. It's a bit confusing to me to have this intermediate SegmentInMemory just to hold the columns and descriptor before extracting them again during the construction of InputFrame. And then creating a "real" SegmentInMemory again during the WriteToSegmentTask::slice() which is what is later passed downstream. I feel it makes more sense for the WriteToSegmentTask (given its name) to be the first place in the pipeline where SegmentInMemory is created. As for the [other comment and the num_rows](https://github.com/man-group/ArcticDB/pull/3174#discussion_r3429594741): the old code (`seg.set_row_data`) just does `col.set_row_data` for each `col` it holds, and also sets its own row_id: [here](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/column_store/memory_segment_impl.cpp#L285-L289) And the column's `row_count()` correctly attributes for strings: it calculates it based on the ChunkedBuffer's bytes_ and the extra row when working with strings is in extra_bytes. Set [here](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/column_store/chunked_buffer.hpp#L520-L529) and read [here](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/column_store/column.cpp#L857-L867). Happy to discuss this lengthy comment offline.

cpp/arcticdb/pipeline/input_frame.cpp
ognyanstoimenov: Answered here https://github.com/man-group/ArcticDB/pull/3174#discussion_r3436251379

cpp/arcticdb/processing/clause_merge_update.cpp
ognyanstoimenov: [Yes](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/entity/native_tensor.hpp#L70-L82)

cpp/arcticdb/pipeline/write_frame.cpp
ognyanstoimenov: This sets the segment's `row_id_` which is used for when calling `segment.row_count`. The columns already have their row_count correctly set: in the NativeTensor case its from [here](https://github.com/man-group/ArcticDB/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/pipeline/write_frame.cpp#L340) -> [this](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/pipeline/frame_utils.hpp#L203) -> [this](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/column_store/column.hpp#L316-L327) and in the column case it's [here](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/pipeline/write_frame.cpp#L300) (or [here](https://github.com/man-group/arcticdb/blob/4849f6159174016d01ad8646912d9fd5b6b1f3f7/cpp/arcticdb/pipeline/write_frame.cpp#L312))

cpp/arcticdb/pipeline/write_frame.cpp
ognyanstoimenov: We could but this way we can prevent iterating through all the columns at the end. Also the NativeTensors will already have correctly set row count: https://github.com/man-group/ArcticDB/pull/3174#discussion_r3436369210

cpp/arcticdb/pipeline/input_frame.hpp
ognyanstoimenov: Right, yes, I'll keep only `has_only_tensors_` as a temporary measure before we support arrow in other cases, eg merge updates

cpp/arcticdb/pipeline/input_frame.hpp
ognyanstoimenov: Oh yes I think it will be nicer! Will do

alexowens90 (approved):

PR #3181, opened 2026-06-17
Support loading only specific symbols when loading snapshots

poodlewars:
Implementing Ivo's idea here: https://chat-man.slack.com/archives/C02NEG3V38X/p1781081726477429

Support loading snapshots only for particular stream IDs to control memory use. Change add_to_snapshot, remove_from_snapshot and the deletion code paths to supply stream IDs for the snapshot load. Add benchmarks for deletion in the presence of snapshots.

ASV results:

```
  | Benchmark | 140ede53e (baseline) | HEAD | Change |
  |---|---|---|---|
  | `peakmem_delete` (LMDB) | 318M | 310M | −8M (−2.5%) |
  | `time_delete` (LMDB) | 67.2±4ms | 34.6±0.2ms | −32.6ms (≈1.94× faster) |
```

(the memory change is quite low to keep the example at a practical size, I could tune this. The enterprise PR shows a more significant change).

Enterprise PR https://github.com/man-group/arcticdb-enterprise/pull/339 adopts this new API for delayed deletes and has some benchmarking results.

This PR also removes this `const std::optional<const std::tuple<const SnapshotVariantKey&, std::vector<IndexTypeKey>&>>&
                get_keys_in_snapshot = std::nullopt` argument from `get_master_snapshots_map`. It is modified in place to accumulate keys in a particular snapshot, and used only when deleting a snapshot to say that we can still delete these keys as long as no other snapshot holds them live. I just pulled out a dedicated function for that code path that returns the master snapshot map and these keys. I also added a test to cover the case where the snapshot being deleted shares some keys but not others with other snapshots.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -194,6 +194,7 @@ class LocalVersionedEngine : public VersionedEngine {
             const PreDeleteChecks& checks = default_pre_delete_checks
     ) override {
         std::unordered_set<StreamId> stream_ids;
+        stream_ids.reserve(idx_to_be_deleted.size());
         for (const auto& key : idx_to_be_deleted) {
             stream_ids.insert(key.id());
         }
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -193,7 +193,11 @@ class LocalVersionedEngine : public VersionedEngine {
             const std::vector<IndexTypeKey>& idx_to_be_deleted,
             const PreDeleteChecks& checks = default_pre_delete_checks
     ) override {
-        auto snapshot_map = get_master_snapshots_map(store());
+        std::unordered_set<StreamId> stream_ids;
+        for (const auto& key : idx_to_be_deleted) {
+            stream_ids.insert(key.id());
+        }
+        auto snapshot_map = get_master_snapshots_map(store(), stream_ids);
         delete_trees_responsibly(store(), version_map(), idx_to_be_deleted, snapshot_map, std::nullopt, checks).get();
     };
 
--- a/cpp/arcticdb/version/snapshot.cpp
+++ b/cpp/arcticdb/version/snapshot.cpp
@@ -396,36 +396,40 @@ SnapshotMap get_versions_from_snapshots(const std::shared_ptr<Store>& store, con
     return snapshot_map;
 }
 
-MasterSnapshotMapWithStats get_master_snapshots_map_with_stats(
-        std::shared_ptr<Store> store,
-        const std::optional<const std::tuple<const SnapshotVariantKey&, std::vector<IndexTypeKey>&>>&
-                get_keys_in_snapshot
-) {
+MasterSnapshotMapWithStats get_master_snapshots_map_with_stats(std::shared_ptr<Store> store) {
     MasterSnapshotMapWithStats out;
-    iterate_snapshots(store, [&get_keys_in_snapshot, &out, &store](const VariantKey& sk) {
+    iterate_snapshots(store, [&out, &store](const VariantKey& sk) {
         ++out.total_snapshots;
         auto snapshot_id = variant_key_id(sk);
         auto snapshot_segment = store->read_sync(sk).second;
         for (size_t idx = 0; idx < snapshot_segment.row_count(); idx++) {
             auto stream_index = read_key_row(snapshot_segment, static_cast<ssize_t>(idx));
             out.map[stream_index.id()][stream_index].insert(snapshot_id);
-            if (get_keys_in_snapshot) {
-                auto [wanted_snap_key, sink] = *get_keys_in_snapshot;
-                if (wanted_snap_key == sk) {
-                    sink.push_back(stream_index);
-                }
-            }
         }
     });
     return out;
 }
 
-MasterSnapshotMap get_master_snapshots_map(
-        std::shared_ptr<Store> store,
-        const std::optional<const std::tuple<const SnapshotVariantKey&, std::vector<IndexTypeKey>&>>&
-                get_keys_in_snapshot
+MasterSnapshotMap get_master_snapshots_map(std::shared_ptr<Store> store) {
+    return get_master_snapshots_map_with_stats(std::move(store)).map;
+}
+
+MasterSnapshotMapAndKeys get_master_snapshots_map_and_keys_in_given_snapshot(
+        std::shared_ptr<Store> store, const SnapshotVariantKey& given_snapshot
 ) {
-    return get_master_snapshots_map_with_stats(std::move(store), get_keys_in_snapshot).map;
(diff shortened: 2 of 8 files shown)


poodlewars:
> `fix_symbol_trees`

I left this alone deliberately as it is little used and poorly tested and I didn't see the value in touching it.

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -397,35 +397,45 @@ SnapshotMap get_versions_from_snapshots(const std::shared_ptr<Store>& store, con
...
-        const std::optional<const std::tuple<const SnapshotVariantKey&, std::vector<IndexTypeKey>&>>&
-                get_keys_in_snapshot
+        std::shared_ptr<Store> store, const std::optional<std::unordered_set<StreamId>>& stream_ids
 ) {
     MasterSnapshotMapWithStats out;
-    iterate_snapshots(store, [&get_keys_in_snapshot, &out, &store](const VariantKey& sk) {
+    iterate_snapshots(store, [&out, &store, &stream_ids](const VariantKey& sk) {
         ++out.total_snapshots;
         auto snapshot_id = variant_key_id(sk);
         auto snapshot_segment = store->read_sync(sk).second;
```
IvoDD, line 406: Future optimization: We can read snapshot keys in parallel in a folly::window at the cost of slightly higher peak mem usage

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -397,35 +397,45 @@ SnapshotMap get_versions_from_snapshots(const std::shared_ptr<Store>& store, con
...
         for (size_t idx = 0; idx < snapshot_segment.row_count(); idx++) {
             auto stream_index = read_key_row(snapshot_segment, static_cast<ssize_t>(idx));
```
IvoDD, line 408: Future optimization: `read_key_row` does a bunch of `scalar_at` calls for each column (which I've seen can be cpu intensive). We could:
- Use `ColumnDataIterator`s
- Build the AtomKey only when the stream_id is in the `stream_ids` set

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -193,7 +193,11 @@ class LocalVersionedEngine : public VersionedEngine {
             const std::vector<IndexTypeKey>& idx_to_be_deleted,
             const PreDeleteChecks& checks = default_pre_delete_checks
     ) override {
-        auto snapshot_map = get_master_snapshots_map(store());
+        std::unordered_set<StreamId> stream_ids;
+        for (const auto& key : idx_to_be_deleted) {
+            stream_ids.insert(key.id());
+        }
```
IvoDD, line 200: Nit: I think constructing the `unordered_set` with `bucket_count = idx_to_be_deleted.size()` upper bound will avoid rehashing.

Also with C++20 we could do (which I'm not feeling strongly about):
```
auto stream_ids_view = idx_to_be_deleted | std::views::transform([](const IndexTypeKey& key) { return key.id(); });

std::unordered_set<StreamId> stream_ids(stream_ids_view.begin(), stream_ids_view.end(), idx_to_be_deleted.size());
```

Similar suggestion applies to other places in this PR where we construct the stream_ids set in a for loop.

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1022,14 +1034,16 @@ void PythonVersionStore::delete_snapshot(const SnapshotId& snap_name) {
 void PythonVersionStore::delete_snapshot_sync(const SnapshotId& snap_name, const VariantKey& snap_key) {
     ARCTICDB_DEBUG(log::version(), "Deleting data of Snapshot {}", snap_name);
 
-    std::vector<AtomKey> index_keys_in_current_snapshot;
-    auto snap_map = get_master_snapshots_map(store(), std::tie(snap_key, index_keys_in_current_snapshot));
+    auto snap_map_and_keys = get_master_snapshots_map_and_keys_in_given_snapshot(store(), snap_key);
```
IvoDD, line 1039: `get_master_snapshots_map_and_keys_in_given_snapshot` can still OOM since it loads all snapshots without any filtering.

I wonder whether it would be better to upfront read the `snap_key` and then to just use `get_master_snapshots_map(, stream_ids_within_this_snapshot)`.

This would also help us remove the new `get_master_snapshots_map_and_keys_in_given_snapshot` method.

I don't feel strongly about this since most snapshots would contain most symbols so this wouldn't help much.

python/benchmarks/modification_functions.py
```diff
@@ -288,3 +288,40 @@ def teardown(self, *args):
...
         self.lib.delete_batch(self.delete_requests)
+
+
+class DeleteSymbolWithSnapshots:
+    """Benchmark deleting a single symbol in a library where every version is pinned by snapshots."""
+
+    number = 1  # the deleted symbol is gone after one run, so setup rebuilds the library for each run
+    warmup_time = 0
+    timeout = 600
+    repeat = 3  # each setup writes NUM_SYMBOLS symbols and NUM_SNAPSHOTS snapshots
```
IvoDD: I wonder whether we should keep `repeat` on the [default](https://asv.readthedocs.io/en/stable/benchmarks.html#benchmark-attributes).

Just 3 repeats feels like it might be flaky.

IvoDD (approved):

cpp/arcticdb/version/version_store_api.cpp
poodlewars: I think the current code behaves as it does to avoid reading the existing snap_key twice. Since we haven't had issues reported about delete snapshot memory use, I think I prefer leaving this (fragile) stuff alone.

cpp/arcticdb/version/snapshot.cpp
```diff
@@ -397,35 +397,45 @@ SnapshotMap get_versions_from_snapshots(const std::shared_ptr<Store>& store, con
...
 
 MasterSnapshotMapWithStats get_master_snapshots_map_with_stats(
-        std::shared_ptr<Store> store,
-        const std::optional<const std::tuple<const SnapshotVariantKey&, std::vector<IndexTypeKey>&>>&
-                get_keys_in_snapshot
+        std::shared_ptr<Store> store, const std::optional<std::unordered_set<StreamId>>& stream_ids
 ) {
     MasterSnapshotMapWithStats out;
-    iterate_snapshots(store, [&get_keys_in_snapshot, &out, &store](const VariantKey& sk) {
+    iterate_snapshots(store, [&out, &store, &stream_ids](const VariantKey& sk) {
```
G-D-Petrov, line 403: What do you think about adding some logs here to help future debugging?
Maybe a couple of info logs for the number of symbols at the beginning and the number of accumulated snapshots at the end?

G-D-Petrov (approved):

cpp/arcticdb/version/snapshot.cpp
poodlewars: The return type includes the snapshot count which can be used for logging at the call site (delayed deletes does this).

PR #3182, opened 2026-06-17
Replace SegmentInMemory With Simple Struct When Calculating Column Stats For Single Slice

markovskipetar:
### Before
The protobuf `ColumnStatsHeader` (which records, per data-column offset, the list of stat
entries and their offsets within the stats segment) was constructed and carried at
**three** stages:

1. `MinMaxAggregatorData::finalize()` returned a full `SegmentInMemory` for each stats that were calculated per single slice that already embedded a `ColumnStatsHeader` in its metadata, with the input column's `data_col_offset_` baked in.
2. `ColumnStatsGenerationClause::process()` unpacked each finalized segment's header, re-based every `stats_seg_offset` by the running field count, merged them into a combined header, and `concatenate`d the segments.
3. `merge_column_stats_segments()` re-read each segment's protobuf header, built a reverse lookup `stats_seg_offset → (data_col_offset, type)`, and reconstructed the header again while preserving offsets.

The offset of the original data column was effectively passed along inside protobuf at each hop.

### After
The protobuf header is built **once**, at the final merge step, by **parsing the segment column names** and resolving the data-column offset from the symbol layout:

1. `MinMaxAggregatorData::finalize()` now returns a lightweight POD struct
   (`ColumnStatsAggregatorOutput`) — just `min`, `max`, `nan_count`, `null_count`,
   `input_column_offset`. No segment, no protobuf.
2. `ColumnStatsGenerationClause::process()` builds the per-slice result segment directly,
   naming each stat column via `column_and_stat_to_segment_name` (`v1_MIN(col)`,
   `v1_MAX(col)`, `v1_NAN_COUNT(col)`, `v1_NULL_COUNT(col)`). **No protobuf header is built
   here anymore.**
3. `merge_column_stats_segments()` reconstructs the header from the merged column names:
   each name is parsed via `parse_segment_column_name` into `(column_name, stat_type)`,
   and the data-column offset is looked up in a `symbol_fields_to_offsets` map derived from
   the symbol's `TimeseriesDescriptor`.

### DIFF
--- a/cpp/arcticdb/pipeline/column_stats.cpp
+++ b/cpp/arcticdb/pipeline/column_stats.cpp
@@ -164,14 +164,6 @@ std::string stat_to_name(ColumnStatType stat) {
     }
 }
 
-std::optional<ColumnStatType> stat_name_to_stat(const std::string& statName) {
-    if (statName == "MINMAX") {
-        return ColumnStatType::MINMAX;
-    }
-
-    user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>("Unknown column stat type provided: {}", statName);
-}
-
 std::string column_and_stat_to_segment_name(const std::string& column, ColumnStatTypeInternal stat) {
     return fmt::format("{}({})", stat_to_operator_string(stat), column);
 }
@@ -310,16 +302,6 @@ ColumnStats::ColumnStats(const TimeseriesDescriptor& tsd) {
         );
     }
     offset_to_stat_info_set_ = true;
-}
-
-std::unordered_set<std::string> get_unmangled_column_names(
-        const std::map<std::string, std::set<ColumnStatType>>& column_name_to_stats
-) {
-    std::unordered_set<std::string> unmangled_names;
-    for (const auto& k : column_name_to_stats | ranges::views::keys) {
-        unmangled_names.insert(k);
-    }
-    return unmangled_names;
 }
 
 }
--- a/cpp/arcticdb/pipeline/column_stats.cpp
+++ b/cpp/arcticdb/pipeline/column_stats.cpp
@@ -131,12 +131,12 @@ std::string stat_to_name(ColumnStatType stat) {
     }
 }
 
-std::optional<ColumnStatType> stat_name_to_stat(const std::string& name) {
-    if (name != "MINMAX") {
-        user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>("Unknown column stat type provided: {}", name);
+std::optional<ColumnStatType> stat_name_to_stat(const std::string& statName) {
+    if (statName == "MINMAX") {
+        return ColumnStatType::MINMAX;
     }
 
-    return ColumnStatType::MINMAX;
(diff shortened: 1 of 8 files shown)


poodlewars:
For some reason the CI hasn't run, so I kicked it off. Running build and test: https://github.com/man-group/ArcticDB/actions/runs/27814800009 

and Analyis: https://github.com/man-group/ArcticDB/actions/runs/27814843871

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -105,24 +121,54 @@ std::string type_to_operator_string(ColumnStatTypeInternal type) {
...
         return "MINMAX";
     default:
         internal::raise<ErrorCode::E_ASSERTION_FAILURE>("Unknown column stat type requested");
     }
 }
 
-std::optional<ColumnStatType> name_to_type(const std::string& name) {
-    if (name == "MINMAX") {
-        return ColumnStatType::MINMAX;
+std::optional<ColumnStatType> stat_name_to_stat(const std::string& name) {
```
poodlewars: The old implementation was correct and we shouldn't have this change. We need to make sure that we don't crash or raise if this version of ArcticDB reads columns stats written by a newer version, that might write stats this version does not understand.

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -105,24 +121,54 @@ std::string type_to_operator_string(ColumnStatTypeInternal type) {
...
+    return ColumnStatType::MINMAX;
 }
 
-std::string to_segment_column_name(const std::string& column, ColumnStatTypeInternal type) {
-    return fmt::format("{}({})", type_to_operator_string(type), column);
+std::string column_and_stat_to_segment_name(const std::string& column, ColumnStatTypeInternal stat) {
+    return fmt::format("{}({})", stat_to_operator_string(stat), column);
+}
+
+ParsedSegmentColumnName parse_segment_column_name(std::string_view segment_column_name) {
```
poodlewars: We shouldn't need this string parsing - all the information to understand each column is in the segment header.

Stopping doing this string parsing was the purpose of https://github.com/man-group/ArcticDB/pull/2990

cpp/arcticdb/processing/aggregation_interface.hpp
```diff
@@ -10,9 +10,22 @@
...
 #include <folly/Poly.h>
 #include <arcticdb/entity/types.hpp>
+#include <arcticdb/pipeline/value.hpp>
+#include <arcticdb/column_store/memory_segment.hpp>
+#include <arcticdb/processing/expression_node.hpp>
+#include <optional>
 
 namespace arcticdb {
 
+// Lightweight per slice and per aggregator struct, replacing the previous SegmentInMemory
```
poodlewars, line 20: Get rid of this comment. In general these "historical" comments are just confusing (comments like "this replaces the previous X that is not in the codebase any more")

poodlewars (changes requested):

PR #3187, opened 2026-06-22
Bugfix 12337436632: do not orphan data keys in append/update when schemas mismatch

alexowens90:
#### Reference Issues/PRs
[12337436632](https://man312219.monday.com/boards/7852509418/pulses/12337436632)

Part of the compatible schema checks for dynamic schema happen in `get_merged_tsd`. Previously, this was called after `slice_and_write`, resulting in orphaned data keys if an exception was then raised.

### DIFF
--- a/cpp/arcticdb/pipeline/write_frame.cpp
+++ b/cpp/arcticdb/pipeline/write_frame.cpp
@@ -508,33 +508,14 @@ folly::Future<entity::AtomKey> write_frame(
 
 folly::Future<AtomKey> append_frame(
         IndexPartialKey&& key, const std::shared_ptr<InputFrame>& frame, const SlicingPolicy& slicing,
-        index::IndexSegmentReader& index_segment_reader, const std::shared_ptr<Store>& store, bool dynamic_schema,
-        bool ignore_sort_order
+        index::IndexSegmentReader& index_segment_reader, const std::shared_ptr<Store>& store, bool dynamic_schema
 ) {
     ARCTICDB_SAMPLE_DEFAULT(AppendFrame)
-    util::variant_match(
-            frame->index,
-            [&index_segment_reader, &frame, ignore_sort_order](const TimeseriesIndex&) {
-                util::check(frame->has_index(), "Cannot append timeseries without index");
-                if (index_segment_reader.tsd().total_rows() != 0 && frame->num_rows != 0) {
-                    auto first_index = NumericIndex{frame->index_value_at(0)};
-                    auto prev = std::get<NumericIndex>(index_segment_reader.last()->key().end_index());
-                    util::check(
-                            ignore_sort_order || prev - 1 <= first_index,
-                            "Can't append dataframe with start index {} to existing sequence ending at {}",
-                            util::format_timestamp(first_index),
-                            util::format_timestamp(prev)
-                    );
-                }
-            },
-            [](const auto&) {
-                // Do whatever, but you can't range search it
-            }
-    );
-
     auto existing_slices = unfiltered_index(index_segment_reader);
+    const auto tsd =
+            index::get_merged_tsd(frame->num_rows + frame->offset, dynamic_schema, index_segment_reader.tsd(), frame);
     auto keys_fut = slice_and_write(frame, slicing, IndexPartialKey{key}, store);
-    return std::move(keys_fut).thenValue([dynamic_schema,
+    return std::move(keys_fut).thenValue([tsd = std::move(tsd),
                                           slices_to_write = std::move(existing_slices),
                                           frame = frame,
                                           index_segment_reader = std::move(index_segment_reader),
@@ -546,9 +527,6 @@ folly::Future<AtomKey> append_frame(
                 std::make_move_iterator(std::end(slice_and_keys_to_append))
         );
         ranges::sort(slices_to_write);
-        const auto tsd = index::get_merged_tsd(
-                frame->num_rows + frame->offset, dynamic_schema, index_segment_reader.tsd(), frame
-        );
         return index::write_index(
                 index_type_from_descriptor(tsd.as_stream_descriptor()), tsd, std::move(slices_to_write), key, store
         );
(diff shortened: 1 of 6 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -148,6 +148,90 @@ void sorted_data_check_append(const InputFrame& frame, index::IndexSegmentReader
     );
 }
 
+void check_index_match(const Index& index, const IndexDescriptorImpl& desc) {
```
alexowens90, line 151: Just moved

cpp/arcticdb/version/version_core.cpp
```diff
@@ -148,6 +148,90 @@ void sorted_data_check_append(const InputFrame& frame, index::IndexSegmentReader
...
+                "Index mismatch, cannot update a non-timeseries-indexed frame with a timeseries"
+        );
+    else
+        util::check(
+                desc.type() == IndexDescriptorImpl::Type::ROWCOUNT,
+                "Index mismatch, cannot update a timeseries with a non-timeseries-indexed frame"
+        );
+}
+
+void check_update_data_is_sorted(const InputFrame& frame, const index::IndexSegmentReader& index_segment_reader) {
```
alexowens90, line 164: Just moved

cpp/arcticdb/version/version_core.cpp
```diff
@@ -148,6 +148,90 @@ void sorted_data_check_append(const InputFrame& frame, index::IndexSegmentReader
...
+                    );
+                }
+            },
+            [](const auto&) {
+                // Do whatever, but you can't range search it
+            }
+    );
+}
+
+static void check_can_update(
```
alexowens90, line 214: Just moved

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -508,33 +508,14 @@ folly::Future<entity::AtomKey> write_frame(
...
-                            "Can't append dataframe with start index {} to existing sequence ending at {}",
-                            util::format_timestamp(first_index),
-                            util::format_timestamp(prev)
-                    );
-                }
-            },
-            [](const auto&) {
-                // Do whatever, but you can't range search it
-            }
-    );
```
alexowens90, line 533: Moved into `check_can_append`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -148,6 +148,90 @@ void sorted_data_check_append(const InputFrame& frame, index::IndexSegmentReader
...
+    );
+}
+
+static void check_can_append(
+        const InputFrame& frame, const index::IndexSegmentReader& index_segment_reader, const UpdateInfo& update_info,
+        const WriteOptions& write_options, bool validate_index, bool empty_types
+) {
+    util::check(
+            update_info.previous_index_key_.has_value(),
+            "Cannot append as there is no previous index key to append into"
```
IvoDD: I agree with this check but isn't it technically an API break?

```
syms = lib.list_symbols()
new_sym = "test_nonexistent_12354"
new_sym in syms
> False
lib.append(new_sym, pd.DataFrame({"col": [1, 2, 3]}))
> VersionedItem(symbol='test_nonexistent_12354', library='idilov.test', data=n/a, version=0, metadata=None, host='mktdatad', timestamp=1782201047208429166)
```

cpp/arcticdb/version/version_core.cpp
```diff
@@ -148,6 +148,90 @@ void sorted_data_check_append(const InputFrame& frame, index::IndexSegmentReader
...
+    );
+    util::check_rte(!index_segment_reader.is_pickled(), "Cannot append to pickled data");
+    fix_descriptor_mismatch_or_throw(APPEND, write_options.dynamic_schema, index_segment_reader, frame, empty_types);
```
IvoDD, line 188: I'm not a fan of this modifying the mutable `norm_meta` on the `const InputFrame& frame`.

It has been like this before but I think `norm_meta` in `InputFrame` should not be mutable. Probably more a philosophical point and would require a lot of code changes so probably better to leave as is.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -161,13 +245,8 @@ folly::Future<AtomKey> async_append_impl(
     auto index_segment_reader = index::get_index_reader(*(update_info.previous_index_key_), store);
     bool bucketize_dynamic = index_segment_reader.bucketize_dynamic();
     auto row_offset = index_segment_reader.tsd().total_rows();
-    util::check_rte(!index_segment_reader.is_pickled(), "Cannot append to pickled data");
+    check_can_append(*frame, index_segment_reader, update_info, options, validate_index, empty_types);
     frame->set_offset(static_cast<ssize_t>(row_offset));
```
IvoDD, line 241: Nit: 
I find it a bit odd that we `check_can_append` in between of calculating and setting the `row_offset` on the frame.

cpp/arcticdb/pipeline/write_frame.cpp
```diff
@@ -508,33 +508,14 @@ folly::Future<entity::AtomKey> write_frame(
...
-
     auto existing_slices = unfiltered_index(index_segment_reader);
+    const auto tsd =
+            index::get_merged_tsd(frame->num_rows + frame->offset, dynamic_schema, index_segment_reader.tsd(), frame);
```
IvoDD, line 516: Note to other reviewers: `frame->offset` is calculated [here](https://github.com/man-group/ArcticDB/pull/3187/changes#r3458120434) before `append_frame` is called

cpp/arcticdb/version/version_core.cpp
alexowens90: It isn't, because of an earlier check:
https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/version/local_versioned_engine.cpp#L1870
I added it for consistency with `check_can_update`, where it is also redundant for similar reasons. In both cases the `IndexSegmentReader` argument was constructed from the key we are checking exists here, so I think I'll just remove it from both.

cpp/arcticdb/version/version_core.cpp
alexowens90: Yeh it's a bit all over the place atm, but I don't want to delve into fixing it right now

IvoDD (approved):

PR #3188, opened 2026-06-23
Make index key reading async in append

alexowens90:
Use `async_get_index_reader` in `async_append_impl` so that the index key reading, and therefore the full function, is actually asynchronous.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -1923,6 +1923,7 @@ std::vector<std::variant<VersionedItem, DataError>> LocalVersionedEngine::batch_
     for (const auto&& [idx, stream_update_info_fut] : folly::enumerate(stream_update_info_futures)) {
         append_versions_futs.push_back(
                 std::move(stream_update_info_fut)
+                        .via(&async::cpu_executor())
                         .thenValue(
                                 [this,
                                  frame = std::move(frames[idx]),
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -2007,6 +2007,7 @@ std::vector<std::variant<VersionedItem, DataError>> LocalVersionedEngine::batch_
     for (const auto&& [idx, stream_update_info_fut] : enumerate(stream_update_info_futures)) {
         update_versions_futs.push_back(
                 std::move(stream_update_info_fut)
+                        .via(&async::cpu_executor())
                         .thenValue(
                                 [this,
                                  frame = std::move(frames[idx]),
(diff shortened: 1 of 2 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -228,27 +228,32 @@ folly::Future<AtomKey> async_append_impl(
...
-            frame,
-            slicing_arg,
-            index_segment_reader,
-            store,
-            options.dynamic_schema
+    ARCTICDB_DEBUG(
+            log::version(), "append stream_id: {} , version_id: {}", frame->desc().id(), update_info.next_version_id_
     );
+    return index::async_get_index_reader(*(update_info.previous_index_key_), store)
+            .via(&async::io_executor())
```
IvoDD: Is there a benefit to the `via(io)` here? The `async_get_index_reader` returns a future already associated with the IOPool, so even without via it would get executed inline on the same IOPool worker.

This would only matter if in the future we change `async_get_index_reader` to bind via a cpu pool.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1923,6 +1923,7 @@ std::vector<std::variant<VersionedItem, DataError>> LocalVersionedEngine::batch_
     for (const auto&& [idx, stream_update_info_fut] : folly::enumerate(stream_update_info_futures)) {
         append_versions_futs.push_back(
                 std::move(stream_update_info_fut)
+                        .via(&async::cpu_executor())
```
IvoDD, line 1926: I assume the reasoning here is to run the `async_append_impl` from a cpu pool so that the `submit_io_pool` used there to read the index key gets fairly distributed across workers? Otherwise the index key read would end up sticking on the same IO worker.

I aggree with that reasoning and I think it is a good change. Should we do it for `batch_update` as well?

Or maybe make the `batch_get_latest_undeleted_version_and_next_version_id_async` do the `.via(cpu)` in the end so all users of those futures start from the cpu pool which seems more correct.

cpp/arcticdb/version/version_core.cpp
alexowens90: Yeh it isn't doing anything, it was just to make it explicit that we are currently in the IO pool

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: >I assume the reasoning here is to run the async_append_impl from a cpu pool so that the submit_io_pool used there to read the index key gets fairly distributed across workers?

Yeh exactly.

I'll check over batch update and make sure it behaves the same way.

IvoDD (approved):

PR #3191, opened 2026-06-23
Bound processing pipeline memory use

poodlewars:
In the processing pipeline we currently kick off reads eagerly, meaning that if the in-memory processing does not complete quickly, large numbers of `SegmentInMemory` objects accumulate in the CPU executor's queue. Creating column stats manually considers all the data in a symbol, so we can end up holding the entire symbol's data in memory.

# Implementation

This PR adds a new "admission handler" to control memory use in the processing pipeline. This affects all QueryBuilder operations, not just column stats. We now create a special "admission handler" object that provides segment read futures. Importantly, it creates these as promises and only kicks off work for them later, unlike the current eager read pipeline. This design means that everything downstream of the segment load futures in the processing pipeline can be used as-is.

The admission handler only allows K processing units to be live in memory at a time. As each unit finishes it calls `on_unit_complete` (when the `MemSegmentProcessingTask` completes). This kicks off the load of the next unit's segments. Importantly this is not a blocking operation, else it could deadlock.

To give an example, with processing units with an overlap:

```
P = [{0, 1}, {1, 2}]
```

we first call `admit_initial` to load processing unit `0`, i.e. segments `0, 1`. When this unit is complete, the component manager frees segment `0` but not segment `1`. At this point `on_unit_complete` fires. This calls `admit(processing_unit=1)` which calls `fire(segment=1)` (no-op) and `fire(segment=2)` which kicks off a segment load.

We also make sure that we effectively `folly::window` over the IO work that we kick off, rather than letting it all pile up on the IO queues when K is large, which gives worse queuing behaviour and regressed performance compared to `folly::window`.

For testing, I add a class `SegmentResidencyTracker` that measures how many segments were live in memory over an operation. This is compiled in to release builds but is a no-op - it only does anything in testing. Its cost is just reading a boolean that's always false.

# Impact

## Measurements

Each table cell is **peak `ru_maxrss` (MiB) / wall time (s)** for one operation. Every cell is a separate process running a single operation, so `ru_maxrss` (peak RSS over the process lifetime) is a clean high-water mark for that one operation.
`master` has no admission bound, so it has a single column. `fixes` varies the residency budget `K` (`VersionStore.NumProcessingUnitsLive`, the max processing units in flight):
`K=0` is the kill switch (bound disabled), `K=8` a tight bound, `K=default` the computed default; the read window `W` (`VersionStore.SegmentReadWindow`) is left at its default of `2*io`.
Each build is run under three process allocators: glibc default, glibc with `MALLOC_ARENA_MAX=4`, and tcmalloc (`LD_PRELOAD`). `IO:CPU` is the IO/CPU threadpool sizes (`VersionStore.NumIOThreads` / `NumCPUThreads`).

## Data

One symbol, `f64_100m_100c_zeros`, on PURE (S3): **100,000,000 rows x 100 `float64` data columns** (`f_0`..`f_99`), all zeros, plus an `int64` `category` column cycling through values 1-5,
indexed by a nanosecond `DatetimeIndex` at 1-second frequency. Written with default slicing (100k rows/segment, all columns in one column slice), giving 1000 row-slice segments.

Each segment is about 80MiB decoded, the symbol is about 80GiB decoded.

## Operations

- **filter**: `read` with `QueryBuilder()[q["f_0"] == -1.0]`. `-1.0` never occurs (data is zeros), so the result is empty and does not contribute to RSS
- **column_stats**: build MINMAX column statistics for each row-slice
- **resample**: `read` with `QueryBuilder().resample("30D").agg(sum)` summing all 100 float columns into ~30-day buckets (a few dozen output rows).
- **groupby**: `read` with `QueryBuilder().groupby("category").agg(sum)` summing all 100 float columns across the 5 `category` groups (5 output rows).

## Findings

- *Allocator makes a big difference* glibc default inflates peak RSS badly as there are so many arenas on the build servers, arena cap or tcmalloc cut RSS significantly. arena4 means glibc with MALLOC_ARENA_MAX=4
- *K=0 reproduces master timings so the kill switch works*
- Shows improved memory use for default K without a penalty on wall clock time

### 1. No Time Penalty with default or kill switch

Similarly performance with K=0 also matches current master.

Timing results with the default value:

| measure | io:cpu | glibc | arena4 | tcmalloc |
|---|---|---|---|---|
| filter | 8:12 | -7.0% / +2.5% | +3.2% / +2.9% | -3.6% / +1.0% |
| filter | 16:24 | -18.2% / +0.9% | -4.1% / -1.2% | -17.2% / +0.2% |
| filter | 64:96 | +26.6% / -0.4% | -0.6% / +3.4% | +1.2% / +10.5% |
| column_stats | 8:12 | -6.5% / +6.1% | -17.3% / +4.9% | -13.5% / +1.0% |
| column_stats | 16:24 | -6.1% / +22.8% | -10.1% / +3.7% | -6.0% / +9.9% |
| column_stats | 64:96 | -4.7% / -44.3% | +1.9% / -3.4% | -2.7% / -0.6% |
| resample | 8:12 | -3.7% / +8.2% | -5.4% / +6.9% | -3.1% / -0.2% |
| resample | 16:24 | +0.0% / +10.4% | +4.9% / -1.6% | +0.9% / +0.6% |
| resample | 64:96 | -1.1% / +2.6% | -1.6% / +0.5% | -0.4% / +2.4% |
| groupby | 8:12 | -6.1% / -17.6% | +0.9% / -0.9% | +18.6% / -0.4% |
| groupby | 16:24 | -9.0% / -21.6% | -5.0% / -0.1% | -12.0% / -0.9% |
| groupby | 64:96 | -11.7% / -19.0% | -9.9% / -0.1% | +7.3% / +0.4% |

### 2. The default only reduces memory for column stats on large hosts

I tried several other defaults, but they imposed a performance penalty on `filter`. I think that at least to start with, a default that does not affect wall-clock time will be easier to roll out.

Timing results:

| measure | io:cpu | glibc | arena4 | tcmalloc |
|---|---|---|---|---|
| filter | 8:12 | +1% | +1% | +1% |
| filter | 16:24 | -1% | -3% | +7% |
| filter | 64:96 | +1% | +1% | +2% |
| column_stats | 8:12 | -0% | +3% | -1% |
| column_stats | 16:24 | +8% | +7% | +2% |
| column_stats | 64:96 | -52% | -3% | +5% |
| resample | 8:12 | -14% | -4% | -2% |
| resample | 16:24 | +2% | -5% | +1% |
| resample | 64:96 | +1% | +2% | +2% |
| groupby | 8:12 | -21% | -1% | -0% |
| groupby | 16:24 | -23% | +0% | -1% |
| groupby | 64:96 | -23% | +0% | +0% |

### 3. `K=8` shows reduced memory use vs `K=0`:

Memory:

| measure | io:cpu | glibc | arena4 | tcmalloc |
|---|---|---|---|---|
| filter | 8:12 | -2% | -1% | +14% |
| filter | 16:24 | -6% | -26% | -38% |
| filter | 64:96 | -24% | -79% | -83% |
| column_stats | 8:12 | -31% | -23% | -16% |
| column_stats | 16:24 | -45% | -42% | -39% |
| column_stats | 64:96 | -87% | -83% | -79% |
| resample | 8:12 | -42% | -37% | -33% |
| resample | 16:24 | -56% | -48% | -46% |
| resample | 64:96 | -72% | -64% | -56% |
| groupby | 8:12 | -23% | -3% | -2% |
| groupby | 16:24 | -26% | -4% | -6% |
| groupby | 64:96 | -31% | -4% | -9% |

The next sections show the raw results:

### glibc default - peak MiB / wall s

| measure | io:cpu | pre-PR | K=0 | K=8 | K=default |
|---|---|---|---|---|---|
| filter | 8:12 | 771.1 / 3.45 | 782.2 / 3.28 | 770.1 / 3.93 | 790.0 / 3.21 |
| filter | 16:24 | 1315.2 / 2.75 | 1335.2 / 2.20 | 1254.8 / 4.00 | 1326.4 / 2.25 |
| filter | 64:96 | 4829.5 / 1.54 | 4772.2 / 2.01 | 3632.8 / 4.76 | 4811.0 / 1.95 |
| column_stats | 8:12 | 2362.4 / 4.94 | 2510.9 / 4.70 | 1724.2 / 8.14 | 2506.4 / 4.62 |
| column_stats | 16:24 | 3517.2 / 3.79 | 4007.0 / 3.22 | 2216.3 / 8.36 | 4319.5 / 3.56 |
| column_stats | 64:96 | 29444.4 / 2.77 | 33952.8 / 2.61 | 4572.4 / 8.84 | 16407.3 / 2.64 |
| resample | 8:12 | 24297.3 / 4.81 | 30551.6 / 4.31 | 17675.5 / 5.71 | 26278.1 / 4.63 |
| resample | 16:24 | 37070.6 / 3.46 | 40023.7 / 3.49 | 17784.2 / 5.53 | 40934.7 / 3.46 |
| resample | 64:96 | 67539.6 / 2.61 | 68373.3 / 2.51 | 19150.5 / 5.54 | 69289.1 / 2.58 |
| groupby | 8:12 | 97413.2 / 18.55 | 102016.0 / 17.48 | 78824.1 / 18.59 | 80275.6 / 17.41 |
| groupby | 16:24 | 104967.1 / 19.52 | 106864.1 / 19.74 | 79279.2 / 19.34 | 82289.8 / 17.77 |
| groupby | 64:96 | 112363.5 / 22.26 | 117741.9 / 21.78 | 81718.5 / 19.67 | 90961.0 / 19.66 |

### MALLOC_ARENA_MAX=4 - peak MiB / wall s

| measure | io:cpu | pre-PR | K=0 | K=8 | K=default |
|---|---|---|---|---|---|
| filter | 8:12 | 728.3 / 3.42 | 743.5 / 3.45 | 737.4 / 5.70 | 749.5 / 3.53 |
| filter | 16:24 | 1235.9 / 2.19 | 1258.1 / 2.07 | 936.1 / 3.40 | 1220.6 / 2.10 |
| filter | 64:96 | 3608.7 / 1.70 | 3704.0 / 1.59 | 780.2 / 3.11 | 3730.1 / 1.69 |
| column_stats | 8:12 | 2170.2 / 5.85 | 2219.8 / 4.89 | 1701.4 / 7.73 | 2276.5 / 4.84 |
| column_stats | 16:24 | 3222.2 / 4.07 | 3121.2 / 3.89 | 1795.0 / 7.78 | 3341.3 / 3.66 |
| column_stats | 64:96 | 10458.1 / 3.19 | 10412.7 / 3.12 | 1756.8 / 7.90 | 10098.1 / 3.25 |
| resample | 8:12 | 24621.5 / 5.14 | 27348.3 / 4.97 | 17238.9 / 5.96 | 26319.0 / 4.86 |
| resample | 16:24 | 32312.4 / 3.90 | 33639.9 / 3.77 | 17477.6 / 5.92 | 31804.6 / 4.09 |
| resample | 64:96 | 47186.8 / 3.87 | 46374.2 / 3.63 | 16925.7 / 6.15 | 47429.8 / 3.81 |
| groupby | 8:12 | 81737.9 / 17.72 | 81832.9 / 23.38 | 79250.6 / 19.79 | 81027.6 / 17.88 |
| groupby | 16:24 | 82548.4 / 18.06 | 82222.0 / 18.98 | 79323.1 / 20.41 | 82486.2 / 17.15 |
| groupby | 64:96 | 83289.1 / 20.80 | 82928.8 / 18.51 | 79212.4 / 26.04 | 83237.5 / 18.74 |

### tcmalloc - peak MiB / wall s

| measure | io:cpu | pre-PR | K=0 | K=8 | K=default |
|---|---|---|---|---|---|
| filter | 8:12 | 679.4 / 3.05 | 682.0 / 2.61 | 774.2 / 3.10 | 686.0 / 2.94 |
| filter | 16:24 | 1278.0 / 2.38 | 1201.7 / 2.10 | 741.5 / 3.29 | 1280.8 / 1.97 |
| filter | 64:96 | 4133.9 / 1.67 | 4474.0 / 1.68 | 745.7 / 2.83 | 4566.4 / 1.69 |
| column_stats | 8:12 | 1970.2 / 5.98 | 2005.1 / 5.16 | 1677.0 / 7.68 | 1989.9 / 5.17 |
| column_stats | 16:24 | 2635.0 / 4.30 | 2829.1 / 3.85 | 1716.4 / 7.82 | 2896.5 / 4.04 |
| column_stats | 64:96 | 8333.5 / 2.58 | 7890.6 / 2.48 | 1691.8 / 7.73 | 8282.7 / 2.51 |
| resample | 8:12 | 24522.3 / 5.76 | 24855.6 / 6.01 | 16537.7 / 6.77 | 24461.8 / 5.58 |
| resample | 16:24 | 30446.7 / 4.62 | 30418.5 / 4.64 | 16525.2 / 6.38 | 30635.5 / 4.66 |
| resample | 64:96 | 37634.2 / 4.48 | 37847.3 / 4.46 | 16494.6 / 6.11 | 38528.1 / 4.46 |
| groupby | 8:12 | 85290.0 / 18.46 | 85312.0 / 17.87 | 83639.0 / 23.53 | 84987.6 / 21.89 |
| groupby | 16:24 | 89116.7 / 18.23 | 88824.0 / 18.07 | 83568.6 / 21.20 | 88325.0 / 16.05 |
| groupby | 64:96 | 92362.7 / 16.08 | 92379.7 / 15.44 | 83730.6 / 23.94 | 92752.0 / 17.25 |

### Aside: Slow Processing

I accidentally did one benchmark run against a debug build `K=default` vs `K=0` peak, glibc, debug build, which had these memory results:

| measure | 8:12 | 16:24 | 64:96 |
|---|---|---|---|
| filter | -2% | +1% | -2% |
| column_stats | -97% | -94% | -82% |
| resample | -67% | -37% | +0% |
| groupby | -27% | -40% | -27% |

This shows a big performance difference with the default K when processing is slow.

# Alternatives

#3138 - discarded as it leaves the CPU pool idle, and is difficult to read (in that the mechanism that bounds memory use is not explicit).

Originally I wanted to pass a semaphore in to `batch_read_uncompressed`, and acquire it for segment reads, and release it after a segment is processed, like compacting staged data does, always allowing at least `n_col_slices` to be loaded. This did not work because we kick off all the segment reads in parallel so your `n_col_slices` budget can be split across segments in different processing units while leaving no processing unit able to load completely and continue. This PR is instead explicit about allowing loads per processing unit.

### DIFF
--- a/cpp/arcticdb/column_store/memory_segment_impl.hpp
+++ b/cpp/arcticdb/column_store/memory_segment_impl.hpp
@@ -523,6 +523,10 @@ class SegmentInMemoryImpl {
 
     SegmentInMemoryImpl clone() const;
 
+    // Marks this segment as having been decoded from storage, so SegmentResidencyTracker counts it from here until
+    // this object is destroyed. clone() and processing outputs are left unmarked.
+    void mark_from_disk();
+
     void set_string_pool(std::shared_ptr<StringPool> string_pool);
 
     std::shared_ptr<SegmentInMemoryImpl> get_output_segment(size_t num_values, bool pre_allocate = true) const;
@@ -577,6 +581,22 @@ class SegmentInMemoryImpl {
     bool compacted_ = false;
     util::MagicNum<'M', 'S', 'e', 'g'> magic_;
     std::optional<TimeseriesDescriptor> tsd_;
+
+    // Just used for memory tracking in tests, see SegmentResidencyTracker
+    // Clears on move so we do not double count the moved-from instance
+    struct OnDiskFlag {
+        bool value_ = false;
+        OnDiskFlag() = default;
+        OnDiskFlag(OnDiskFlag&& other) noexcept : value_(other.value_) { other.value_ = false; }
+        OnDiskFlag& operator=(OnDiskFlag&& other) noexcept {
+            value_ = other.value_;
+            other.value_ = false;
+            return *this;
+        }
+        OnDiskFlag(const OnDiskFlag&) = delete;
+        OnDiskFlag& operator=(const OnDiskFlag&) = delete;
+    };
+    OnDiskFlag from_disk_;
 };
 
 } // namespace arcticdb
--- a/cpp/arcticdb/column_store/memory_segment_impl.hpp
+++ b/cpp/arcticdb/column_store/memory_segment_impl.hpp
@@ -17,6 +17,8 @@
 #include <arcticdb/util/constructors.hpp>
 #include <boost/iterator/iterator_facade.hpp>
 
+#include <utility>
+
 namespace py = pybind11;
 
 namespace arcticdb {
@@ -524,7 +526,8 @@ class SegmentInMemoryImpl {
     SegmentInMemoryImpl clone() const;
 
     // Marks this segment as having been decoded from storage, so SegmentResidencyTracker counts it from here until
-    // this object is destroyed. clone() and processing outputs are left unmarked.
+    // this object is destroyed. A no-op unless the tracker is enabled.
+    // Processing outputs are left unmarked.
     void mark_from_disk();
 
     void set_string_pool(std::shared_ptr<StringPool> string_pool);
@@ -587,10 +590,10 @@ class SegmentInMemoryImpl {
     struct OnDiskFlag {
         bool value_ = false;
         OnDiskFlag() = default;
-        OnDiskFlag(OnDiskFlag&& other) noexcept : value_(other.value_) { other.value_ = false; }
+        OnDiskFlag(OnDiskFlag&& other) noexcept : value_(std::exchange(other.value_, false)) {}
+        // std::exchange rather than read-then-clear so that self-move-assignment leaves value_ set
         OnDiskFlag& operator=(OnDiskFlag&& other) noexcept {
-            value_ = other.value_;
-            other.value_ = false;
+            value_ = std::exchange(other.value_, false);
             return *this;
         }
         OnDiskFlag(const OnDiskFlag&) = delete;
--- a/cpp/arcticdb/stream/stream_source.hpp
+++ b/cpp/arcticdb/stream/stream_source.hpp
@@ -68,10 +68,11 @@ struct StreamSource {
     [[nodiscard]] virtual std::vector<folly::Future<bool>> batch_key_exists(const std::vector<entity::VariantKey>& keys
     ) = 0;
 
-    virtual std::vector<folly::Future<pipelines::SegmentAndSlice>> batch_read_uncompressed(
-            std::vector<pipelines::RangesAndKey>&& ranges_and_keys,
-            std::shared_ptr<std::unordered_set<std::string>> columns_to_decode
-    ) = 0;
+    virtual std::function<folly::Future<pipelines::SegmentAndSlice>(pipelines::RangesAndKey&&)>
+    make_uncompressed_reader(std::shared_ptr<std::unordered_set<std::string>> /*columns_to_decode*/
+    ) {
+        util::raise_rte("make_uncompressed_reader not implemented");
+    }
 
     virtual folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> read_metadata(
             const entity::VariantKey& key, storage::ReadKeyOpts opts = storage::ReadKeyOpts{}
--- a/cpp/arcticdb/stream/stream_source.hpp
+++ b/cpp/arcticdb/stream/stream_source.hpp
@@ -69,10 +69,7 @@ struct StreamSource {
     ) = 0;
 
     virtual std::function<folly::Future<pipelines::SegmentAndSlice>(pipelines::RangesAndKey&&)>
-    make_uncompressed_reader(std::shared_ptr<std::unordered_set<std::string>> /*columns_to_decode*/
-    ) {
-        util::raise_rte("make_uncompressed_reader not implemented");
-    }
+    make_uncompressed_reader(std::shared_ptr<std::unordered_set<std::string>> columns_to_decode) = 0;
 
     virtual folly::Future<std::pair<std::optional<VariantKey>, std::optional<google::protobuf::Any>>> read_metadata(
             const entity::VariantKey& key, storage::ReadKeyOpts opts = storage::ReadKeyOpts{}
--- a/cpp/arcticdb/util/segment_residency_tracker.hpp
+++ b/cpp/arcticdb/util/segment_residency_tracker.hpp
@@ -1,0 +1,56 @@
+/* Copyright 2026 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software
+ * will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+
+namespace arcticdb::util {
+
+// Test-only instrumentation for the number of segments decoded from storage that are resident in memory at once.
+class SegmentResidencyTracker {
+  public:
+    static SegmentResidencyTracker& instance() {
+        static SegmentResidencyTracker tracker;
+        return tracker;
+    }
+
+    void set_enabled(bool enabled) { enabled_.store(enabled, std::memory_order_relaxed); }
+
+    bool enabled() const { return enabled_.load(std::memory_order_relaxed); }
+
+    void reset() {
+        live_.store(0, std::memory_order_relaxed);
+        high_water_.store(0, std::memory_order_relaxed);
+    }
+
+    void on_segment_resident() {
+        if (!enabled_.load(std::memory_order_relaxed))
+            return;
+        const auto now = live_.fetch_add(1, std::memory_order_relaxed) + 1;
+        auto prev = high_water_.load(std::memory_order_relaxed);
+        while (now > prev && !high_water_.compare_exchange_weak(prev, now, std::memory_order_relaxed)) {
+        }
+    }
+
+    void on_segment_released() {
+        if (!enabled_.load(std::memory_order_relaxed))
+            return;
+        live_.fetch_sub(1, std::memory_order_relaxed);
+    }
+
+    int64_t high_water() const { return high_water_.load(std::memory_order_relaxed); }
+
+  private:
+    std::atomic<bool> enabled_{false};
+    std::atomic<int64_t> live_{0};
+    std::atomic<int64_t> high_water_{0};
+};
+
+} // namespace arcticdb::util
--- a/cpp/arcticdb/util/segment_residency_tracker.hpp
+++ b/cpp/arcticdb/util/segment_residency_tracker.hpp
@@ -33,19 +33,21 @@ class SegmentResidencyTracker {
     void on_segment_resident() {
         if (!enabled_.load(std::memory_order_relaxed))
             return;
+        // high_water_ = max(high_water_, now), handling concurrent on_segment_resident calls
+        // FUTURE C++26 - replace with std::atomic::fetch_max
         const auto now = live_.fetch_add(1, std::memory_order_relaxed) + 1;
         auto prev = high_water_.load(std::memory_order_relaxed);
         while (now > prev && !high_water_.compare_exchange_weak(prev, now, std::memory_order_relaxed)) {
         }
     }
 
-    void on_segment_released() {
-        if (!enabled_.load(std::memory_order_relaxed))
-            return;
-        live_.fetch_sub(1, std::memory_order_relaxed);
-    }
+    // Deliberately not gated on enabled_: callers only reach here for a segment on_segment_resident counted, so
+    // releasing unconditionally keeps live_ correct when tracking is disabled while counted segments are still alive.
+    void on_segment_released() { live_.fetch_sub(1, std::memory_order_relaxed); }
 
     int64_t high_water() const { return high_water_.load(std::memory_order_relaxed); }
+
+    int64_t live() const { return live_.load(std::memory_order_relaxed); }
 
   private:
     std::atomic<bool> enabled_{false};
--- a/cpp/arcticdb/version/admission_handler.hpp
+++ b/cpp/arcticdb/version/admission_handler.hpp
@@ -1,0 +1,155 @@
+/* Copyright 2026 Man Group Operations Limited
+ *
+ * Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source License, use of this software
+ * will be governed by the Apache License, version 2.0.
+ */
+
+#pragma once
+
+#include <arcticdb/pipeline/frame_slice.hpp>
+#include <folly/futures/Future.h>
+#include <folly/futures/Promise.h>
+#include <atomic>
+#include <deque>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <vector>
+
+namespace arcticdb::version_store {
+
+using SegmentReader = std::function<folly::Future<pipelines::SegmentAndSlice>(pipelines::RangesAndKey&&)>;
+
+// Feeds segment reads to the IO pool under two independent controls.
+//
+// Residency budget (memory): at most `max_processing_units_in_flight` processing units are admitted but not yet
+// processed at once. A unit is admitted initially or when an outstanding one finishes processing
+// (on_processing_unit_complete), which bounds the decoded segments resident in memory. A value >= the number of units
+// disables the bound.
+//
+// Read window (throughput): at most `read_window` reads are submitted-but-not-decoded at once. A slot frees on read
+// completion, so an admitted unit's segments enter a queue that fill_read_window drains it into the IO pool
+// the way folly::window does. When the residency budget is not the limiting factor, this
+// reduces to folly::window(read_window). This is important to avoid scheduling all the IO work upfront,
+// which can lead to very poor queue fairness as the Folly IO pool schedules work round robin across its threads.
+//
+// Never waits within a unit, so progress continues for any budget >= 1 and window >= 1 regardless of the processing
+// unit structure (for example, if there is overlap between units). Construct via make_shared (reads capture
+// shared_from_this to keep the handler alive across the async reads).
+class ProcessingUnitAdmissionHandler : public std::enable_shared_from_this<ProcessingUnitAdmissionHandler> {
+  public:
+    ProcessingUnitAdmissionHandler(
+            SegmentReader reader, std::vector<pipelines::RangesAndKey>&& ranges_and_keys,
+            std::vector<std::vector<size_t>>&& processing_units, size_t max_processing_units_in_flight,
+            size_t read_window
+    ) :
+        reader_(std::move(reader)),
+        ranges_and_keys_(std::make_shared<std::vector<pipelines::RangesAndKey>>(std::move(ranges_and_keys))),
+        promises_(std::make_shared<std::vector<folly::Promise<pipelines::SegmentAndSlice>>>(ranges_and_keys_->size())),
+        i_th_segment_enqueued_(ranges_and_keys_->size(), false),
+        read_window_(std::max<size_t>(1, read_window)),
+        processing_units_(std::move(processing_units)),
+        max_processing_units_in_flight_(max_processing_units_in_flight),
+        next_unit_(max_processing_units_in_flight) {}
+
+    // Used to chain downstream processing, possibly before any of the work is actually executing.
+    // The work then gets kicked off by admit_initial_processing_units or on_processing_unit_complete.
+    std::vector<folly::Future<pipelines::SegmentAndSlice>> futures() const {
+        std::vector<folly::Future<pipelines::SegmentAndSlice>> res;
+        res.reserve(promises_->size());
+        for (auto& promise : *promises_) {
+            res.emplace_back(promise.getFuture());
+        }
+        return res;
+    }
+
+    const std::vector<std::vector<size_t>>& processing_units() const { return processing_units_; }
+
+    void admit_initial_processing_units() {
+        for (size_t u = 0; u < std::min(max_processing_units_in_flight_, processing_units_.size()); ++u) {
+            enqueue_processing_unit(u);
+        }
+        fill_read_window();
+    }
+
+    void on_processing_unit_complete() {
+        const auto u = next_unit_.fetch_add(1);
+        if (u < processing_units_.size()) {
+            enqueue_processing_unit(u);
+        }
+        fill_read_window();
+    }
+
+  private:
+    // Make the u-th unit's segments eligible to read. A segment may belong to two units (e.g. resample boundaries), so
+    // it is enqueued at most once. Segments are enqueued unit-contiguously so folly::collect on a unit is not left
+    // waiting on segments submitted far apart.
+    void enqueue_processing_unit(const size_t u) {
+        std::lock_guard lock{window_mutex_};
+        for (const auto i : processing_units_.at(u)) {
+            if (!i_th_segment_enqueued_.at(i)) {
+                i_th_segment_enqueued_.at(i) = true;
+                segments_queued_for_read_.push_back(i);
+            }
+        }
+    }
+
+    // Launch eligible reads until the read window is full. Called whenever a read completes or a new processing unit
+    // is admitted.
+    void fill_read_window() {
+        std::vector<size_t> to_launch;
+        {
+            std::lock_guard lock{window_mutex_};
+            while (active_reads_ < read_window_ && !segments_queued_for_read_.empty()) {
+                to_launch.push_back(segments_queued_for_read_.front());
+                segments_queued_for_read_.pop_front();
+                ++active_reads_;
+            }
+        }
+        for (const auto i : to_launch) {
+            dispatch_read(i);
+        }
+    }
+
+    void dispatch_read(size_t i) {
+        auto self = shared_from_this();
+        reader_(pipelines::RangesAndKey{ranges_and_keys_->at(i)})
+                .thenTryInline([self, i](folly::Try<pipelines::SegmentAndSlice>&& t) {
+                    self->promises_->at(i).setTry(std::move(t));
+                    self->on_read_complete();
+                });
+    }
+
+    void on_read_complete() {
+        {
+            std::lock_guard lock{window_mutex_};
+            --active_reads_;
+        }
+        fill_read_window();
+    }
+
+    SegmentReader reader_;
+    std::shared_ptr<std::vector<pipelines::RangesAndKey>> ranges_and_keys_;
+
+    // We accumulate promises for all segments up front, but they do not start to execute until the read is dispatched.
+    // This lets us set up chains of futures while controlling how many are actually executing at a given time.
+    std::shared_ptr<std::vector<folly::Promise<pipelines::SegmentAndSlice>>> promises_;
+
+    // State to hand-roll a folly::window over the segment loads on the IO pool
+    std::mutex window_mutex_; // protects the 3 fields below
+    std::vector<bool> i_th_segment_enqueued_;
+    std::deque<size_t> segments_queued_for_read_;
+    size_t active_reads_{0};
+    size_t read_window_;
+
+    // State to manage the number of processing units being processed at a time
+    std::vector<std::vector<size_t>> processing_units_;
+    size_t max_processing_units_in_flight_;
+    std::atomic<size_t> next_unit_;
+};
+
+size_t max_resident_processing_units(const std::vector<std::vector<size_t>>& processing_unit_indexes);
+
+} // namespace arcticdb::version_store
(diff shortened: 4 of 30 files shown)


cpp/arcticdb/version/test/test_admission_handler.cpp
poodlewars, line 1: This file has a weird name, should be called `test_column_stats_memory.cpp`

cpp/arcticdb/processing/component_manager.cpp
```diff
@@ -27,6 +28,7 @@ void ComponentManager::decrement_entity_fetch_count(EntityId id) {
         // shared_mutex, so just decrement the ref count of the only sizeable component, so that when the shared pointer
         // goes out of scope in the calling function, the memory is freed
         registry_.get<std::shared_ptr<SegmentInMemory>>(id).reset();
+        util::SegmentResidencyTracker::instance().on_segment_released();
```
poodlewars: Could go in `SegmentInMemory` ctors? (but then more overhead)

cpp/arcticdb/util/segment_residency_tracker.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+// Test-only instrumentation for the number of decoded segments resident in the ComponentManager at once.
+class SegmentResidencyTracker {
+  public:
+    static SegmentResidencyTracker& instance() {
+        static SegmentResidencyTracker tracker;
+        return tracker;
+    }
+
+    void set_enabled(bool enabled) { enabled_.store(enabled, std::memory_order_relaxed); }
```
poodlewars, line 24: Could measure the overhead and consider enabling by default and adding some debug logging?

cpp/arcticdb/processing/component_manager.cpp
poodlewars: Or does entt have something useful for this? Should we track in the component manager itself?

poodlewars:
_Do some more convinving testing of the new memory use_
Done, see PR description.
*We may want to tune the default* so the memory bound kicks in sooner.

_Always use the admission handler (polymorphic or semaphore size = n processing units)_
  Done, outside of the batch path

_Is it valid to use IO thread count for the number of processing units to load? Should we also consider the CPU thread count?_
  I think so. This matches the old behaviour. Since users can configure the IO thread count and CPU thread count independently, I think we have to control the bound off the IO thread count for that configuration to make any sense.

We could take the max of the current setting and the CPU pool size, or twice the CPU pool size to give slack like the current approach (the idea being to keep the hopper full, try not to have consumers sat waiting on an empty queue). This would help if processing units have a large size (very col sliced) such that the IO pool creates only one processing unit at a time. The CPU pool works on units, so would be mostly idle with this workload.

_Later_
(Singleton?) admission handler across symbols so it works properly with batch reads. Could either change the non-batch path or not.

cpp/arcticdb/processing/component_manager.cpp
poodlewars: Tracking entirely in the component manager doesn't really work because we add entities to it in the CPU executor after they are decoded. If segments are stuck in the CPU queue they use memory but would not contribute to the count. This is the problem with column stats creation at the moment.

ENTT does have some hooks:

```
  registry.on_construct<std::shared_ptr<SegmentInMemory>>().connect<&inc>();
  registry.on_destroy  <std::shared_ptr<SegmentInMemory>>().connect<&dec>();
  registry.on_update   <std::shared_ptr<SegmentInMemory>>().connect<&...>();
```

but they are invoked on `.erase(entity_id)` - we are just calling `.reset()` as a performance optimization. So these hooks aren't available to us ATM.

For our testing here I care about the residence of segments that were loaded off disk, rather than created during processing. I'm now tracking this in `SegmentInMemoryImpl`. I think we should keep this tracking off outside of testing because the component manager decrement path is worried enough about performance to have this trick,

```
        // This entity will never be accessed again
        // Ideally would call registry_.destroy(id), or at least registry_.erase<std::shared_ptr<SegmentInMemory>>(id)
        // at this point. However, they are both slower than this, and would require taking a unique_lock on the
        // shared_mutex, so just decrement the ref count of the only sizeable component, so that when the shared pointer
        // goes out of scope in the calling function, the memory is freed
        registry_.get<std::shared_ptr<SegmentInMemory>>(id).reset();

```

but I will try out the impact of turning it on in a separate PR, see https://github.com/man-group/ArcticDB/pull/3191#discussion_r3482759464

cpp/arcticdb/util/segment_residency_tracker.hpp
poodlewars: I'll do this separately later - I can turn it on by default and observe the impact on ASV benchmarks in a new PR.

cpp/arcticdb/util/segment_residency_tracker.hpp
```diff
@@ -0,0 +1,56 @@
...
+        live_.store(0, std::memory_order_relaxed);
+        high_water_.store(0, std::memory_order_relaxed);
+    }
+
+    void on_segment_resident() {
+        if (!enabled_.load(std::memory_order_relaxed))
+            return;
+        const auto now = live_.fetch_add(1, std::memory_order_relaxed) + 1;
+        auto prev = high_water_.load(std::memory_order_relaxed);
+        while (now > prev && !high_water_.compare_exchange_weak(prev, now, std::memory_order_relaxed)) {
```
alexowens90, line 40: IIUC, `high_water_` is tracking the maximum number of segments live at any one time? And this is equivalent to `high_water_.fetch_add(1)` in the absence of any racing of `on_segment_resident/on_segment_released`? And this implementation is more accurate in the case that `on_segment_released` is called and finishes after `now` is loaded, but before `high_water_` is incremented?
If so, needs a comment, and if not, also needs a comment with what its actually doing 😆

cpp/arcticdb/util/segment_residency_tracker.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+
+namespace arcticdb::util {
+
+// Test-only instrumentation for the number of segments decoded from storage that are resident in memory at once.
+class SegmentResidencyTracker {
```
alexowens90, line 17: Did you consider making release builds even more of a no-op by ifdeffing out the method implementations?

cpp/arcticdb/column_store/memory_segment_impl.hpp
```diff
@@ -577,6 +581,22 @@ class SegmentInMemoryImpl {
...
     util::MagicNum<'M', 'S', 'e', 'g'> magic_;
     std::optional<TimeseriesDescriptor> tsd_;
+
+    // Just used for memory tracking in tests, see SegmentResidencyTracker
+    // Clears on move so we do not double count the moved-from instance
+    struct OnDiskFlag {
+        bool value_ = false;
+        OnDiskFlag() = default;
+        OnDiskFlag(OnDiskFlag&& other) noexcept : value_(other.value_) { other.value_ = false; }
+        OnDiskFlag& operator=(OnDiskFlag&& other) noexcept {
```
alexowens90, line 595: Classically not handling the `this == &other` case

cpp/arcticdb/column_store/memory_segment_impl.hpp
```diff
@@ -577,6 +581,22 @@ class SegmentInMemoryImpl {
     bool compacted_ = false;
     util::MagicNum<'M', 'S', 'e', 'g'> magic_;
     std::optional<TimeseriesDescriptor> tsd_;
+
+    // Just used for memory tracking in tests, see SegmentResidencyTracker
+    // Clears on move so we do not double count the moved-from instance
+    struct OnDiskFlag {
```
alexowens90, line 590: I guess an equivalent alternative implementation would be for each segment to get a unique ID in response to `on_segment_resident`, which is then provided in the call to `on_segment_released` to avoid double free-counting. Would involve more bookkeeping, but this will always be disabled in release builds so doesn't really matter

cpp/arcticdb/version/admission_handler.hpp
```diff
@@ -0,0 +1,155 @@
...
+        i_th_segment_enqueued_(ranges_and_keys_->size(), false),
+        read_window_(std::max<size_t>(1, read_window)),
+        processing_units_(std::move(processing_units)),
+        max_processing_units_in_flight_(max_processing_units_in_flight),
+        next_unit_(max_processing_units_in_flight) {}
+
+    // Used to chain downstream processing, possibly before any of the work is actually executing.
+    // The work then gets kicked off by admit_initial_processing_units or on_processing_unit_complete.
+    std::vector<folly::Future<pipelines::SegmentAndSlice>> futures() const {
+        std::vector<folly::Future<pipelines::SegmentAndSlice>> res;
```
alexowens90, line 72: nit: These could be calculated in the ctor, although as this method is only called once it doesn't matter in practice

cpp/arcticdb/version/admission_handler.hpp
```diff
@@ -0,0 +1,155 @@
...
+    const std::vector<std::vector<size_t>>& processing_units() const { return processing_units_; }
+
+    void admit_initial_processing_units() {
+        for (size_t u = 0; u < std::min(max_processing_units_in_flight_, processing_units_.size()); ++u) {
+            enqueue_processing_unit(u);
+        }
+        fill_read_window();
+    }
+
+    void on_processing_unit_complete() {
```
alexowens90, line 89: Can this deadlock if `read_window` is set to 1, but every processing unit has more than 1 segment in it?

alexowens90 (changes requested):

cpp/arcticdb/util/segment_residency_tracker.hpp
poodlewars: Yes, but I think the one relaxed atomic load per segment decode is nothing, and its nice to have cheap instrumentation in the release build that we can turn on for investigation without making a special wheel.

cpp/arcticdb/version/admission_handler.hpp
poodlewars: The window slots are managed based on when reads complete (rather than processing) so after 1 segment is read, there's a slot free in the window to read the next. Added a test for this. `on_read_complete` touches the window, but not the permits for processing units.

Update: If submitting the read future throws, things could still hang. I'll add a test and fix for this.

cpp/arcticdb/util/segment_residency_tracker.hpp
poodlewars: I've added a comment here - it isn't a `fetch_add` because we might be below the high water mark, it's a `fetch_max`.

cpp/arcticdb/version/admission_handler.hpp
```diff
@@ -0,0 +1,175 @@
...
+    ProcessingUnitAdmissionHandler(
+            SegmentReader reader, std::vector<pipelines::RangesAndKey>&& ranges_and_keys,
+            std::vector<std::vector<size_t>>&& processing_units, size_t max_processing_units_in_flight,
+            size_t read_window
+    ) :
+        reader_(std::move(reader)),
+        ranges_and_keys_(std::make_shared<std::vector<pipelines::RangesAndKey>>(std::move(ranges_and_keys))),
+        promises_(std::make_shared<std::vector<folly::Promise<pipelines::SegmentAndSlice>>>(ranges_and_keys_->size())),
+        i_th_segment_enqueued_(ranges_and_keys_->size(), false),
+        read_window_(std::max<size_t>(1, read_window)),
```
IvoDD: Nit: `max(1, read_window)` is a bit weird. I think I prefer asserting `read_window>=1` and `max_processing_units_in_flight>=1`. A user passing read_window=0 might expect some default to be picked for them.

cpp/arcticdb/version/admission_handler.hpp
```diff
@@ -0,0 +1,175 @@
...
+            while (active_reads_ < read_window_ && !segments_queued_for_read_.empty()) {
+                to_launch.push_back(segments_queued_for_read_.front());
+                segments_queued_for_read_.pop_front();
+                ++active_reads_;
+            }
+        }
+        for (const auto i : to_launch) {
+            // A reader that throws instead of returning a failed future would otherwise leave the window slot used
+            // and the promise unfulfilled, hanging every consumer of this segment. Catch everything, since losing a
+            // promise here cannot be recovered from.
```
IvoDD: If the reader throws instead of returning a failed future shouldn't we reraise that instead of catching it silently and trying to handle it gracefully?

If we still want to handle it gracefully I think we should do the catch inside `dispatch_read` and call `on_read_complete` instead of just `--active_threads`. 
Otherwise we could still deadlock if we have `read_window=1` a single processing unit with two segments `{0, 1}` and fail on the `0`. Then `1` will never be kicked off.

cpp/arcticdb/column_store/memory_segment_impl.hpp
```diff
@@ -523,6 +525,11 @@ class SegmentInMemoryImpl {
 
     SegmentInMemoryImpl clone() const;
 
+    // Marks this segment as having been decoded from storage, so SegmentResidencyTracker counts it from here until
+    // this object is destroyed. A no-op unless the tracker is enabled. clone() and processing outputs are left
```
IvoDD: Leaving a clone unmarked seems slightly weird.
We could:
```
1. decode into segment A
2. B = A.clone()
3. Memory use of decoded segments is now 2x. Recorded one is 1x.
4. Destruct A, we decrement the counter for it.
4. Memory use is 1x and counter is 0x. The decoded segment essentially still lives but we record it as freed.
```

I think calling mark_from_disk on clones if the source segment is marked would make sense.

Also if we strictly don't want to count copies but just the total number of decoded segments kept in memory we should probably use something like refcounting or AlexO's idea [here](https://github.com/man-group/ArcticDB/pull/3191#discussion_r3613419139)

In any case I hope we don't use `clone` much and this probably doesn't matter much

cpp/arcticdb/stream/stream_source.hpp
```diff
@@ -68,10 +68,11 @@ struct StreamSource {
     [[nodiscard]] virtual std::vector<folly::Future<bool>> batch_key_exists(const std::vector<entity::VariantKey>& keys
     ) = 0;
 
-    virtual std::vector<folly::Future<pipelines::SegmentAndSlice>> batch_read_uncompressed(
-            std::vector<pipelines::RangesAndKey>&& ranges_and_keys,
-            std::shared_ptr<std::unordered_set<std::string>> columns_to_decode
-    ) = 0;
+    virtual std::function<folly::Future<pipelines::SegmentAndSlice>(pipelines::RangesAndKey&&)>
+    make_uncompressed_reader(std::shared_ptr<std::unordered_set<std::string>> /*columns_to_decode*/
```
IvoDD: Nit: Slight personal preference to an abstract `=0` method.

This way if we ever introduce a new descendant of `StreamSource` we will know compile time if we fail to implement it instead of only encountering a runtime error in the processing pipeline

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1227,17 +1193,39 @@ folly::Future<std::vector<EntityId>> read_and_schedule_processing(
...
-                   std::move(processing_unit_indexes),
-                   std::make_shared<std::vector<std::shared_ptr<Clause>>>(read_query->clauses_)
-    )
-            .via(&async::cpu_executor());
+    auto admission = std::make_shared<ProcessingUnitAdmissionHandler>(
+            std::move(reader),
+            std::move(ranges_and_keys),
+            std::move(processing_unit_indexes),
+            max_processing_units_in_flight,
+            read_window
```
IvoDD, line 1384: Nit: What do you think about making the AdmissionHandler have a default nullopt argument value for max_processing_units_in_flight and read_window and make the constructor call `max_resident_processing_units` and `segment_read_window` and overre the default values only for tests?

IvoDD (commented):
What do you think about adding more end to end python testing around the admission handler?

Current coverage:
- Extensive local C++ unit tests for the admission handler itself
- All processing pipeline tests run with default admission handler settings
- A few column stats tests run with non default admission handler

What do you think about also:
- Parametrized CI run to run all tests with `max_processing_units_in_memory=1`. Can be only for one python version similar to other such parametrizations like version chain cache disabled
- `SegmentResidency` counter tests in python which assert the high_watermark is less than a given value for a full read with processing. This would require exposing the residency tracker in python but I think that could be pretty useful for some tests. It could also be used as a `track_` asv benchmark as some alternative to some `peakmem` benchmarks.

cpp/arcticdb/version/version_core.cpp
poodlewars: I like that the admission handler is its own standalone thing, that doesn't depend on the TaskScheduler, so I'm inclined to leave this as is. Especially since this is the only place that makes one atm.

cpp/arcticdb/column_store/memory_segment_impl.hpp
poodlewars: I've added counting for clones.

cpp/arcticdb/version/admission_handler.hpp
poodlewars: Added a test against your example, it doesn't deadlock because of the `.ensure` in version_core.cpp:864.

Suppose that we reraise here, and say that we have your example but segment 1 raises. Then the promise at index 1 will never resolve, but the `.ensure` keeps the chain alive. So when we eventually `folly::collect` we just hang.

poodlewars:
> What do you think about adding more end to end python testing around the admission handler?
> 
> Current coverage:
> 
> * Extensive local C++ unit tests for the admission handler itself
> * All processing pipeline tests run with default admission handler settings
> * A few column stats tests run with non default admission handler
> 
> What do you think about also:
> 
> * Parametrized CI run to run all tests with `max_processing_units_in_memory=1`. Can be only for one python version similar to other such parametrizations like version chain cache disabled
> * `SegmentResidency` counter tests in python which assert the high_watermark is less than a given value for a full read with processing. This would require exposing the residency tracker in python but I think that could be pretty useful for some tests. It could also be used as a `track_` asv benchmark as some alternative to some `peakmem` benchmarks.

I've added the Python tests you mentioned, they look pretty cool.

I'll try the parameterized CI run in its own commit and see how slow it is, it might be that it's more worthwhile just to do a single run with it before merging this PR rather than doing it all the time.

poodlewars:
@IvoDD Added the parametrized run. Linux 3.11 only

## Timings

From [run 31208626729](https://github.com/man-group/ArcticDB/actions/runs/31208626729), `Run test` step. All six `Serial` jobs pass.

| test dir | Default | Serial | ratio | Default job | Serial job |
|---|---|---|---|---|---|
| `integration` | 36m24s | 35m58s | 0.99x | 39m49s | 38m26s |
| `{hypothesis,nonreg,scripts,sanitizers}` | 27m42s | 38m17s | 1.38x | 31m24s | 42m02s |
| `stress` | 22m46s | 23m26s | 1.03x | 25m16s | 25m54s |
| `unit` | 8m53s | 10m27s | 1.18x | 11m09s | 13m06s |
| `compat` | 10m22s | 9m42s | 0.94x | 13m33s | 13m21s |
| `enduser` | 0m06s | 0m05s | 0.83x | 2m43s | 3m23s |
| **total** | **106m13s** | **117m55s** | **1.11x** | | |

docs/mkdocs/docs/runtime_config.md
alexowens90, line 1: I'd omit mention of column stats creation in the website docs until its a fully supported feature

alexowens90 (approved):

cpp/arcticdb/version/admission_handler.hpp
IvoDD: You're right there is no deadlock in version core but it is not only because of the `ensure` but because of the collect [here](https://github.com/man-group/ArcticDB/blob/a9b8d61e62d8b4313292227b571d998aa751eac9/cpp/arcticdb/version/version_core.cpp#L880) short circuiting early. If that `collect` were switched to `collectAll` it would deadlock I think.

~~Also the new test you added is not an accurate representation of what's happening in version_core. It calls `ensure` on each segment future, not on each processing unit and that's why it passes. I think if you use `collectAll(processing_unit_segment_futures).ensure(on_processing_unit_complete)` you'd be able to repro the deadlock.~~

I think this all would be way less confusing if instead of `--active_reads` we did `on_read_complete()`.

Also more generally the `collect().ensure(on_processing_unit_complete)` has some weird semantics. Example:
```
Segments per processing unit:
Proc 1: [1, 2, 3, 4]
Proc 2: [5, 6, 7, 8]
Proc 3: [9, 10, 11, 12]
Proc 4: [13, 14, 15, 16]

and we have a capacity of 1 and assume 1, 5, 9, and 13 all fail.

Events:
1. schedule Proc 1 and enque [1, 2, 3, 4]
2. 1 fails. The collect over [1, 2, 3, 4] short circuits and we call `on_processing_unit_complete()` while 2, 3, 4 are still working
3. We enqueu [5, 6, 7, 8] so the queue is now [2, 3, 4, 5, 6, 7, 8].
4. Now assume 5 fails before others complete
5. collect for proc 2 short circuits and we call `on_processing_unit_complete` again
6. We enqueu 9, 10, 11, 12 even though all of [2, 3, 4, 6, 7, 8] are running
7. So on and so on we can end up with unbounded number of reads running in parallel
```
I appreciate that is a very contrived example (almost impossible to happen in practice) but I still think it would make the semantics easier to follow if we use `collectAll` instead of `collect`. The difference wouldn't matter for all successful reads.

IvoDD (approved):

PR #3193, opened 2026-06-23
Refactor WriteToSegmentTask for compact_data_inline

alexowens90:
Refactors `WriteToSegmentTask` in ways that are needed for the `compact_data_inline` argument to `append`, plus some other cleanups:

- `SlicingPolicy` and `slice_num_for_column` were being used in a roundabout way to calculate the row, which was already available in the `set_*_type` functions
- `get_slice_and_rowcount` was only used to generate the `slice_num_for_column` above, and so is no longer needed
- The `partial_key_gen_` callback was the same everywhere, so this logic can just be part of the `WriteToSegmentTask` class. The `TypedStreamVersion` argument that replaces it is optional as it will not be needed with inline defrag
- The function returned by `get_partial_key_gen` commented on the inefficiency of the implementation, and that it could be improved by using the segment generated in `WriteToSegmentTask`, so this is now how it works
- `sparsify_floats` is always `false` when the task is used for anything except explicit `write` calls, so make this a default

### DIFF
--- a/cpp/arcticdb/entity/native_tensor.hpp
+++ b/cpp/arcticdb/entity/native_tensor.hpp
@@ -241,22 +241,36 @@ struct TypedTensor : public NativeTensor {
         // The new column shape * the column stride tells us how far to move the data pointer from the origin
 
         ptr = reinterpret_cast<const uint8_t*>(tensor.data()) + (slice_num * stride_offset);
-        check_ptr_within_bounds(tensor, slice_num, stride_offset);
+        check_ptr_within_bounds(tensor, slice_num * stride_offset);
+    }
+
+    TypedTensor(const NativeTensor& tensor, size_t row, ssize_t nvalues) :
+        NativeTensor(
+                nvalues * itemsize(), tensor.ndim(), tensor.strides(), tensor.shape(), tensor.data_type(),
+                tensor.elsize(), nullptr, tensor.expanded_dim()
+        ) {
+        // The above ctor that uses slice_num and regular_slice_size rather than just row is for ndarray column support,
+        // which is not yet available. It is only used in test_tensor.cpp.
+        util::check(ndim() <= 1, "TypedTensor rows ctor should not be called with dim 2 data");
+        shapes_[0] = nvalues;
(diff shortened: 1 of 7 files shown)


cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -241,22 +241,34 @@ struct TypedTensor : public NativeTensor {
         // The new column shape * the column stride tells us how far to move the data pointer from the origin
 
         ptr = reinterpret_cast<const uint8_t*>(tensor.data()) + (slice_num * stride_offset);
-        check_ptr_within_bounds(tensor, slice_num, stride_offset);
+        check_ptr_within_bounds(tensor, slice_num * stride_offset);
+    }
+
+    TypedTensor(const NativeTensor& tensor, size_t row, ssize_t nvalues) :
```
alexowens90, line 247: I'm not confident that this is sufficient for multidimensional tensors, so I've separated this out for now. Can be recombined later if necessary.

IvoDD: We have replaced all usages in frame_utils.hpp and namely the [assignment in WriteToSegmentTask](https://github.com/man-group/ArcticDB/blob/31fbaf2a170da34a9c38a27e40690a9c72ce71fa/cpp/arcticdb/pipeline/write_frame.cpp#L349) with this constructor.

Do you know if anyone is using multi-dimensional tensors or if we have tests for those?

Mostly worried to not break someone relying on multi dimensional inputs without being confident in the change.

Maybe an inline comment would be useful as well.

alexowens90: >Do you know if anyone is using multi-dimensional tensors or if we have tests for those?

They're not supported at all. I think there are some skipped tests for them because they were segfaulting.

The other constructor is only used in `set_array_type`, which itself is currently unused, and the `tensor_at` methods of `Column` and `SegmentInMemory`, which also seem to be unused, so its safe. Agree it could used a comment though.

IvoDD (approved):

ognyanstoimenov (approved):
Nice

PR #3194, opened 2026-06-24
Reimplement FixedSlicer in terms of SpecificSlicer

alexowens90:
Adds a new `SpecificSlicer` that takes collections of row-ranges and column-ranges, and slices input frames using the cartesian product of these ranges.

`FixedSlicer` is then just a special case of `SpecificSlicer`, and so now just constructs appropriate collections of row-ranges and column-ranges and then uses `SpecificSlicer`.

This is in preparation for the `compact_data_inline` argument to `append`, which requires slicing more flexibly than `FixedSlicer` in order to maintain the invariant that every resulting row slice has `rows_per_segment ± 33%` rows.

### DIFF
--- a/cpp/arcticdb/pipeline/slicing.cpp
+++ b/cpp/arcticdb/pipeline/slicing.cpp
@@ -58,23 +58,24 @@ std::pair<size_t, size_t> get_first_and_last_row(const arcticdb::pipelines::Inpu
 std::vector<FrameSlice> FixedSlicer::operator()(const arcticdb::pipelines::InputFrame& frame) const {
     const auto [first_row, last_row] = get_first_and_last_row(frame);
     std::vector<RowRange> row_ranges;
+    row_ranges.reserve((last_row - first_row + row_per_slice_ - 1) / row_per_slice_);
     for (std::size_t r = first_row, end = last_row; r < end; r += row_per_slice_) {
         auto rdist = std::min(last_row - r, row_per_slice_);
         row_ranges.emplace_back(r, r + rdist);
     }
+    if (row_ranges.empty()) {
+        return {};
+    }
     std::vector<ColRange> col_ranges;
     const auto [index_count, total_field_count] = get_index_and_field_count(frame);
     auto start_col = index_count;
+    // We do not support writing frames with zero columns, and so there will always be at least one column range here
     do {
         auto col_count = std::min(total_field_count - start_col, col_per_slice_);
         col_ranges.emplace_back(start_col, start_col + col_count);
         start_col += col_count;
     } while (start_col < total_field_count);
-    if (row_ranges.empty() || col_ranges.empty()) {
-        return {};
-    } else {
-        return SpecificSlicer{std::move(row_ranges), std::move(col_ranges)}(frame);
-    }
+    return SpecificSlicer{std::move(row_ranges), std::move(col_ranges)}(frame);
 }
 
 SpecificSlicer::SpecificSlicer(std::vector<RowRange>&& row_ranges, std::vector<ColRange>&& col_ranges) :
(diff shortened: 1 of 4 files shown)


cpp/arcticdb/pipeline/slicing.cpp
```diff
@@ -56,25 +56,59 @@ std::pair<size_t, size_t> get_first_and_last_row(const arcticdb::pipelines::Inpu
...
+    std::vector<ColRange> col_ranges;
     const auto [index_count, total_field_count] = get_index_and_field_count(frame);
-    auto field_count = total_field_count - index_count;
+    auto start_col = index_count;
+    do {
+        auto col_count = std::min(total_field_count - start_col, col_per_slice_);
+        col_ranges.emplace_back(start_col, start_col + col_count);
+        start_col += col_count;
+    } while (start_col < total_field_count);
+    if (row_ranges.empty() || col_ranges.empty()) {
```
ognyanstoimenov: `col_ranges` is never empty

cpp/arcticdb/pipeline/slicing.cpp
```diff
@@ -56,25 +56,59 @@ std::pair<size_t, size_t> get_first_and_last_row(const arcticdb::pipelines::Inpu
...
+    std::vector<RowRange> row_ranges;
+    for (std::size_t r = first_row, end = last_row; r < end; r += row_per_slice_) {
+        auto rdist = std::min(last_row - r, row_per_slice_);
+        row_ranges.emplace_back(r, r + rdist);
+    }
+    std::vector<ColRange> col_ranges;
     const auto [index_count, total_field_count] = get_index_and_field_count(frame);
-    auto field_count = total_field_count - index_count;
+    auto start_col = index_count;
+    do {
```
ognyanstoimenov, line 73: I know this is from before, but maybe worth a comment explaining why we need to produce at least one column slice here

ognyanstoimenov (approved):
Maybe now it's a good time to add some tests for the new slicer, as this PR is small enough and also the follow up PRs can rely on it working correctly

cpp/arcticdb/pipeline/slicing.cpp
```diff
@@ -56,25 +56,59 @@ std::pair<size_t, size_t> get_first_and_last_row(const arcticdb::pipelines::Inpu
 }
 
 std::vector<FrameSlice> FixedSlicer::operator()(const arcticdb::pipelines::InputFrame& frame) const {
+    const auto [first_row, last_row] = get_first_and_last_row(frame);
+    std::vector<RowRange> row_ranges;
```
IvoDD, line 60: Nit: Can `reserve((last_row - first_row + rows_per_slice_ - 1) / rows_per_slice_)` to avoid multiple allocations

cpp/arcticdb/pipeline/slicing.cpp
```diff
@@ -56,25 +56,59 @@ std::pair<size_t, size_t> get_first_and_last_row(const arcticdb::pipelines::Inpu
...
+    for (std::size_t r = first_row, end = last_row; r < end; r += row_per_slice_) {
+        auto rdist = std::min(last_row - r, row_per_slice_);
+        row_ranges.emplace_back(r, r + rdist);
+    }
+    std::vector<ColRange> col_ranges;
```
IvoDD, line 69: Nit: Similar `reserve(std::min((total_field_count - start_col + cols_per_slice - 1) / cols_per_slice_, 1))`

Although maybe a bit too ugly :/. Not on hot path so probably shouldn't bother.

IvoDD (approved):

IvoDD (approved):

PR #3196, opened 2026-06-24
Refactor PipelineContext to keep an IndexSegmentReader as a member

alexowens90:
Maintains an `IndexSegmentReader` in the `PipelineContext`, as this will be needed for schema checks with `compact_data_inline`. 

This eliminates the need for a separate `UserDefinedMetadata` field, as when present, this was always just moved out from an `IndexSegmentReader` anyway.

Also abstracts access to the pipeline context's normalization metadata behind some methods to avoid copying as far as possible. As a side effect, this prevents double-moving from the norm metadata field.


alexowens90:
Not the right way to do this, will incorporate a proper change into the `compact_data_inline` branch

PR #3201, opened 2026-06-25
Bug fixes for query builder filtering with ISIN filters over sets with duplicate str representations

poodlewars:
## Bug 1

Fix for OR filters with isin with value sets with clashing str representations:

```
q = q[(q["a"].isin(vals1)) | (q["a"].isin(vals2))]
```

str(np.ndarray) truncates like [0, 1, 2, ... 99997, 99998, 99999] so it is easy for ValueSet str representations to clash. We have special handling for this to suffix -vX in the ValueSet leaf node, but there was no such handling for the ExpressionNode,

```
ISIN([VALUE_SET_1])
```

This meant that filters,

```
q["a"].isin(V1) | q["a"].isin(V2)
```

would incorrectly create a single node key, `ISIN [1, 2 ... 999]` and one of the two filters would effectively be dropped at evaluation time, since we used these strings to describe the edges in the AST.

## Bug 2

Drop release notes for this part - not user facing as column stats is feature flagged. Fixes expressions like this with column stats in the case where the two value sets have the same `str` representation:

```
    q = QueryBuilder()
    q = q[q["col_1"].isin(value_set_one)]
    q = q[q["col_1"].isnotin(value_set_two)]
```

Column stats AND-s filters together in to one in `and_filter_expression_contexts`., which merged the trees in two ExpressionContext objects together in to a graph, sharing leaf nodes where they matched on names.

These names in the original filter AST come from this in `processing.py`:

```
def to_string(leaf):
    if isinstance(leaf, (np.ndarray, list)):
        # Truncate value set keys to first 100 characters
        key = str(leaf)[:100]
    elif isinstance(leaf, _RegexGeneric):
        key = "Regex({})".format(leaf.text())
    else:
        if isinstance(leaf, str):
            key = "Str({})".format(leaf)
        elif isinstance(leaf, bool):
            key = "Bool({})".format(leaf)
        else:
            key = "Num({})".format(leaf)
    return key
```

so different ValueSets can clash, leading to an incorrect merged tree . Note `str(np.ndarray)` also truncates like `[0, 1, 2, ... 99997, 99998, 99999]`, so the "first 100 chars" is not the only problem.

## Fix

We now form the query AST as a graph in the frontend, rather than relying on string references to encode edges. This gets rid of the need to make the two fixes below.

We keep the concept of memoizing results, ie making sure we only evaluate subtrees once in cases like:

```
  # only want to calculate the isin once
  big = list(range(100_000))
  q = q[q["id"].isin(big) & (q["t"] > 0) | q["id"].isin(big)]

  # only want to calculate the spread once
  spread = q["bid"] - q["ask"]
  q = q[(spread > 0) & (spread < limit)]
```

We do this by looking up expressions that we are going to evaluate in a map of string label name to expression node and result. To guard against conflicts, we do a deep comparison against the node that created the result.

### DIFF
--- a/cpp/arcticdb/processing/expression_context.hpp
+++ b/cpp/arcticdb/processing/expression_context.hpp
@@ -9,111 +9,21 @@
 #pragma once
 
 #include <memory>
-#include <string>
-#include <unordered_map>
 
 #include <arcticdb/processing/expression_node.hpp>
-#include <arcticdb/pipeline/value.hpp>
-#include <arcticdb/pipeline/value_set.hpp>
 #include <arcticdb/util/preconditions.hpp>
-#include <arcticdb/util/variant.hpp>
 
 namespace arcticdb {
 
-// ColumnName refers to a real data column and must never be prefixed. The other leaf kinds name
-// entries in an ExpressionContext's maps, so they are prefixed to match a merge of expression
-// contexts that has `prefix`.
-inline VariantNode prefix_internal_names(const VariantNode& node, const std::string& prefix) {
-    return util::variant_match(
-            node,
-            [](std::monostate empty) -> VariantNode { return empty; },
-            [](const ColumnName& column_name) -> VariantNode { return column_name; },
-            [&](const ValueName& value_name) -> VariantNode { return ValueName{prefix + value_name.value}; },
-            [&](const ValueSetName& value_set_name) -> VariantNode {
-                return ValueSetName{prefix + value_set_name.value};
-            },
-            [&](const ExpressionName& expression_name) -> VariantNode {
-                return ExpressionName{prefix + expression_name.value};
-            },
-            [&](const RegexName& regex_name) -> VariantNode { return RegexName{prefix + regex_name.value}; }
-    );
-}
-
 /*
- * Contains an expression tree as a map of string node names to nodes. Nodes contain leaves which may be references
- * to node names.
- *
- * ExpressionContext also contains the name of the root node as well as a map of node names to constant values.
- *
- * This class effectively contains as AST - there is likely a frontend implementation in a higher level language (for
- * example Python).
+ * An expression tree (an AST).
  */
 struct ExpressionContext {
     ExpressionContext() = default;
 
     ARCTICDB_MOVE_COPY_DEFAULT(ExpressionContext)
 
-    template<class T>
-    class ConstantMap {
-        std::unordered_map<std::string, std::shared_ptr<T>> map_;
-
-      public:
-        void set_value(const std::string& name, std::shared_ptr<T> val) { map_.try_emplace(name, val); }
-        std::shared_ptr<T> get_value(const std::string& name) const { return map_.at(name); }
-        bool contains(const std::string& name) const { return map_.contains(name); }
-
-        template<class Func>
-        void for_each(Func&& func) const {
-            for (const auto& [name, val] : map_) {
-                func(name, val);
-            }
-        }
-    };
-
-    void add_expression_node(const std::string& name, std::shared_ptr<ExpressionNode> expression_node) {
-        expression_nodes_.set_value(name, std::move(expression_node));
-    }
-
-    void add_value(const std::string& name, std::shared_ptr<Value> value) { values_.set_value(name, std::move(value)); }
-
-    void add_value_set(const std::string& name, std::shared_ptr<ValueSet> value_set) {
-        value_sets_.set_value(name, std::move(value_set));
-    }
-
-    void add_regex(const std::string& name, std::shared_ptr<util::RegexGeneric> regex) {
-        regex_matches_.set_value(name, std::move(regex));
-    }
-
-    // Merge another context's entries into this one, prefixing every name so that entries from
-    // separate contexts cannot collide.
-    void merge_from(const ExpressionContext& other, const std::string& prefix) {
-        other.values_.for_each([&](const std::string& name, const auto& val) { values_.set_value(prefix + name, val); }
-        );
-        other.value_sets_.for_each([&](const std::string& name, const auto& val) {
-            value_sets_.set_value(prefix + name, val);
-        });
-        other.regex_matches_.for_each([&](const std::string& name, const auto& val) {
-            regex_matches_.set_value(prefix + name, val);
-        });
-        other.expression_nodes_.for_each([&](const std::string& name, const auto& node) {
-            auto prefixed = std::make_shared<ExpressionNode>(*node);
-            prefixed->condition_ = prefix_internal_names(prefixed->condition_, prefix);
-            prefixed->left_ = prefix_internal_names(prefixed->left_, prefix);
-            prefixed->right_ = prefix_internal_names(prefixed->right_, prefix);
-            util::check(
-                    !expression_nodes_.contains(prefix + name),
-                    "Prefixed expression node {} already present",
-                    prefix + name
-            );
-            add_expression_node(prefix + name, std::move(prefixed));
-        });
-    }
-
-    ConstantMap<ExpressionNode> expression_nodes_;
-    ConstantMap<Value> values_;
-    ConstantMap<ValueSet> value_sets_;
-    ConstantMap<util::RegexGeneric> regex_matches_;
-    VariantNode root_node_name_;
+    std::shared_ptr<ExpressionNode> root_;
     bool dynamic_schema_{false};
 };
 
--- a/cpp/arcticdb/processing/expression_context.hpp
+++ b/cpp/arcticdb/processing/expression_context.hpp
@@ -11,7 +11,7 @@
 #include <memory>
 
 #include <arcticdb/processing/expression_node.hpp>
-#include <arcticdb/util/preconditions.hpp>
+#include <arcticdb/util/constructors.hpp>
 
 namespace arcticdb {
 
--- a/cpp/arcticdb/processing/expression_context.hpp
+++ b/cpp/arcticdb/processing/expression_context.hpp
@@ -15,8 +15,29 @@
 #include <arcticdb/processing/expression_node.hpp>
 #include <arcticdb/pipeline/value.hpp>
 #include <arcticdb/pipeline/value_set.hpp>
+#include <arcticdb/util/preconditions.hpp>
+#include <arcticdb/util/variant.hpp>
 
 namespace arcticdb {
+
+// ColumnName refers to a real data column and must never be prefixed. The other leaf kinds name
+// entries in an ExpressionContext's maps, so they are prefixed to match a merge of expression
+// contexts that has `prefix`.
+inline VariantNode prefix_internal_names(const VariantNode& node, const std::string& prefix) {
+    return util::variant_match(
+            node,
+            [](std::monostate empty) -> VariantNode { return empty; },
+            [](const ColumnName& column_name) -> VariantNode { return column_name; },
+            [&](const ValueName& value_name) -> VariantNode { return ValueName{prefix + value_name.value}; },
+            [&](const ValueSetName& value_set_name) -> VariantNode {
+                return ValueSetName{prefix + value_set_name.value};
+            },
+            [&](const ExpressionName& expression_name) -> VariantNode {
+                return ExpressionName{prefix + expression_name.value};
+            },
+            [&](const RegexName& regex_name) -> VariantNode { return RegexName{prefix + regex_name.value}; }
+    );
+}
 
 /*
  * Contains an expression tree as a map of string node names to nodes. Nodes contain leaves which may be references
@@ -41,9 +62,10 @@ struct ExpressionContext {
         std::shared_ptr<T> get_value(const std::string& name) const { return map_.at(name); }
         bool contains(const std::string& name) const { return map_.contains(name); }
 
-        void merge_from(const ConstantMap& other) {
-            for (const auto& [name, val] : other.map_) {
-                map_.try_emplace(name, val);
+        template<class Func>
+        void for_each(Func&& func) const {
+            for (const auto& [name, val] : map_) {
+                func(name, val);
             }
         }
     };
@@ -62,11 +84,29 @@ struct ExpressionContext {
         regex_matches_.set_value(name, std::move(regex));
     }
 
-    void merge_from(const ExpressionContext& other) {
-        expression_nodes_.merge_from(other.expression_nodes_);
-        values_.merge_from(other.values_);
-        value_sets_.merge_from(other.value_sets_);
-        regex_matches_.merge_from(other.regex_matches_);
+    // Merge another context's entries into this one, prefixing every name so that entries from
+    // separate contexts cannot collide.
+    void merge_from(const ExpressionContext& other, const std::string& prefix) {
+        other.values_.for_each([&](const std::string& name, const auto& val) { values_.set_value(prefix + name, val); }
+        );
+        other.value_sets_.for_each([&](const std::string& name, const auto& val) {
+            value_sets_.set_value(prefix + name, val);
+        });
+        other.regex_matches_.for_each([&](const std::string& name, const auto& val) {
+            regex_matches_.set_value(prefix + name, val);
+        });
+        other.expression_nodes_.for_each([&](const std::string& name, const auto& node) {
+            auto prefixed = std::make_shared<ExpressionNode>(*node);
+            prefixed->condition_ = prefix_internal_names(prefixed->condition_, prefix);
+            prefixed->left_ = prefix_internal_names(prefixed->left_, prefix);
+            prefixed->right_ = prefix_internal_names(prefixed->right_, prefix);
+            util::check(
+                    !expression_nodes_.contains(prefix + name),
+                    "Prefixed expression node {} already present",
+                    prefix + name
+            );
+            add_expression_node(prefix + name, std::move(prefixed));
+        });
     }
 
     ConstantMap<ExpressionNode> expression_nodes_;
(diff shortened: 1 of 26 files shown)


python/tests/unit/arcticdb/test_column_stats_isin.py
```diff
@@ -504,3 +504,136 @@ def test_column_stats_isin_int64_min_not_treated_as_nat(
...
+    qs.reset_stats()
+    result = lib.read(sym, query_builder=q).data
+    table_data_reads = get_table_data_read_count()
+
+    assert_frame_equal(df0, result)
+    expected_reads = 1
+    assert table_data_reads == expected_reads, f"Expected {expected_reads} TABLE_DATA read(s), got {table_data_reads}"
+
+
+def test_column_stats_isin_and_isnotin_colliding_value_sets_nonreg(
```
poodlewars, line 600: This is the only test that was failing before the change. I think the ones above are worth adding anyway.

cpp/arcticdb/processing/expression_context.hpp
```diff
@@ -9,71 +9,21 @@
 #pragma once
 
 #include <memory>
-#include <string>
-#include <unordered_map>
 
 #include <arcticdb/processing/expression_node.hpp>
-#include <arcticdb/pipeline/value.hpp>
-#include <arcticdb/pipeline/value_set.hpp>
+#include <arcticdb/util/preconditions.hpp>
```
alexowens90: ?

cpp/arcticdb/processing/expression_node.hpp
```diff
@@ -85,31 +69,56 @@ struct BitSetTag {};
...
-    OperationType operation_type_;
+    using Leaf = std::variant<
+            ColumnName, std::shared_ptr<Value>, std::shared_ptr<ValueSet>, std::shared_ptr<util::RegexGeneric>>;
+
+    struct Operation {
+        OperationType operation_type_;
+        std::shared_ptr<ExpressionNode> condition_;
+        std::shared_ptr<ExpressionNode> left_;
+        std::shared_ptr<ExpressionNode> right_;
+    };
```
alexowens90, line 89: nit: I think it would look slightly cleaner if the member ordering was `operation_type_, left_, right_, condition_`, with a ctor with defaulted `nullptr` values for `right_` and `condition_`

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+            [&seg, this](const Operation& op) -> VariantData {
+                if (auto it = seg.computed_data_.find(label_); it != seg.computed_data_.end()) {
+                    const auto& [other, cached] = it->second;
+                    if (other == this || nodes_equal(*this, *other)) {
+                        log::version().debug("Memoization hit for {}", label_);
+                        return cached;
+                    }
+                    log::version().debug("Memoization label collision for {}", label_);
+                }
+                log::version().debug("Memoization miss, computing {}", label_);
```
alexowens90: This will be very noisy, I'd either remove or compile out of debug builds

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+                        );
+                    } else if (is_binary_operation(op.operation_type_)) {
+                        return dispatch_binary(op.left_->compute(seg), op.right_->compute(seg), op.operation_type_);
+                    } else {
+                        return dispatch_unary(op.left_->compute(seg), op.operation_type_);
+                    }
+                }();
+                // On a label collision the slot is already held by a structurally-different node; try_emplace is a
+                // no-op there, so that node keeps the slot and this result simply isn't memoized.
+                seg.computed_data_.try_emplace(label_, this, result);
```
alexowens90: I think it _is_ worth logging if this emplace fails, as you could conceivably see unexpected bimodal performance for similar looking queries in this case

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+        return fmt::format("{}({})", op.operation_type_, op.left_->label_);
+    }
+}
+
+bool value_sets_equal(ValueSet& a, ValueSet& b) {
+    if (a.empty() != b.empty()) {
+        return false;
+    }
+    if (a.base_type() != b.base_type()) {
+        // This might cause some false negatives but the simplicity is worth it
```
alexowens90, line 115: Agreed

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+        const auto& leaf_b = std::get<ExpressionNode::Leaf>(b.kind_);
+        if (leaf_a.index() != leaf_b.index()) {
+            return false;
+        }
+        // Now we know that leaf_a and leaf_b hold the same variant type member
+        return util::variant_match(
+                leaf_a,
+                [&leaf_b](const ColumnName& a_col) { return a_col.value == std::get<ColumnName>(leaf_b).value; },
+                [&leaf_b](const std::shared_ptr<Value>& a_val) {
+                    return *a_val == *std::get<std::shared_ptr<Value>>(leaf_b);
```
alexowens90, line 156: SImilarly to the false negatives comment above, this wouldn't consider 2 semantically equal but representatively different values as equal (e.g. `np.uint8(0)` and `np.uint16(0)`, but agree it isn't worth worrying about

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+                [&leaf_b](const std::shared_ptr<ValueSet>& a_set) {
+                    return value_sets_equal(*a_set, *std::get<std::shared_ptr<ValueSet>>(leaf_b));
+                },
+                [&leaf_b](const std::shared_ptr<util::RegexGeneric>& a_regex) {
+                    return a_regex->text() == std::get<std::shared_ptr<util::RegexGeneric>>(leaf_b)->text();
+                }
+        );
+    }
+    const auto& oa = std::get<ExpressionNode::Operation>(a.kind_);
+    const auto& ob = std::get<ExpressionNode::Operation>(b.kind_);
```
alexowens90: nit: `operation_a` and `operation_b`

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+            [](const std::shared_ptr<Value>& value) {
+                return details::visit_type(value->data_type(), [&value](auto tag) {
+                    using info = ScalarTypeInfo<decltype(tag)>;
+                    return fmt::format(
+                            "Val({}:{})", value->data_type(), value->template to_string<typename info::RawType>()
+                    );
+                });
+            },
+            [](const std::shared_ptr<ValueSet>& value_set) {
+                return fmt::format("ValueSet({},n={})", value_set->base_type().data_type(), value_set->size());
```
alexowens90, line 94: Could include a few sample values, although the lack of stable ordering could make it confusing

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+            },
+            [](const std::shared_ptr<util::RegexGeneric>& regex) { return fmt::format("Regex({})", regex->text()); }
+    );
+}
+
+std::string label_for(const ExpressionNode::Operation& op) {
```
alexowens90, line 100: Guessing this works because by construction child nodes are always created before parent nodes?

cpp/arcticdb/processing/processing_unit.cpp
```diff
@@ -48,56 +48,29 @@ void ProcessingUnit::truncate(size_t start_row, size_t end_row) {
...
-                    auto data = expr->compute(*this);
-                    computed_data_.try_emplace(expression_name.value, data);
-                    return data;
-                }
-            },
-            [&](const std::monostate&) -> VariantData {
-                util::raise_rte("ProcessingUnit::get called with monostate VariantNode");
-            }
-    );
+    if (expression_context_ && !expression_context_->dynamic_schema_) {
```
alexowens90, line 66: Is this the only place the `dynamic_schema_` member of `ExpressionContext` is used? If so, clauses know whether dynamic schema is enabled or not via the `set_processing_config` method, so we could scrap `ExpressionContext` entirely

cpp/arcticdb/processing/processing_unit.hpp
```diff
@@ -47,7 +45,12 @@ struct ProcessingUnit {
     std::optional<std::vector<uint64_t>> entity_fetch_count_;
 
     std::shared_ptr<ExpressionContext> expression_context_;
-    std::unordered_map<std::string, VariantData> computed_data_;
+
+    /*
+     * Memoization of results after applying an ExpressionNode to this processing unit. Keyed on the node's label_ with
+     * a deep comparison on hits. One slot per label, colliding writes are dropped.
+     */
+    std::unordered_map<std::string, std::pair<const ExpressionNode*, VariantData>> computed_data_;
```
alexowens90, line 56: Maybe worth a comment that the shared pointer owning the `ExpressionNode` is guaranteed to outlive this map, and so the raw pointer is safe

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -33,34 +33,21 @@ ExpressionContext and_filter_expression_contexts(
...
     util::check(!expression_contexts.empty(), "Expression context cannot be empty");
-    std::optional<ExpressionName> overall_root_name;
-
-    ExpressionContext res;
-    for (auto&& [idx, expression_context] : folly::enumerate(expression_contexts)) {
+    std::shared_ptr<ExpressionNode> root;
+    for (const auto& expression_context : expression_contexts) {
         util::check(
-                std::holds_alternative<ExpressionName>(expression_context->root_node_name_),
+                expression_context->root_ && expression_context->root_->is_operation(),
```
alexowens90, line 40: I think the second condition is wrong, you can filter directly on a bool column for example

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -33,34 +33,21 @@ ExpressionContext and_filter_expression_contexts(
...
-                !res.expression_nodes_.contains(intermediate_name),
-                "Expression nodes already contains {}",
-                intermediate_name
-        );
-        res.add_expression_node(intermediate_name, std::make_shared<ExpressionNode>(and_node));
-        overall_root_name = ExpressionName{intermediate_name};
     }
 
-    res.root_node_name_ = *overall_root_name;
+    ExpressionContext res;
```
alexowens90, line 54: If we keep the `dynamic_schema_` field, then presumably should be inherited from the input expression contexts? Or are they not set yet?

python/arcticdb/version_store/processing.py
alexowens90: Think this is just used by the (Python) `ExpressionNode` class now, so could be a static method

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -555,6 +555,64 @@ def test_filter_isin_clashing_sets(
...
+    lib.write(sym, df)
+
+    limit = 3
+    q = QueryBuilder()
+    spread = q["bid"] - q["ask"]
+    q = q[(spread > 0) & (spread < limit)]
+
+    pandas_spread = df["bid"] - df["ask"]
+    expected = df[(pandas_spread > 0) & (pandas_spread < limit)].reset_index(drop=True)
+    received = lib.read(sym, query_builder=q).data
```
alexowens90, line 612: This will always be empty as the spread is always `-5`, might be more interesting to test a case with a non-empty result

cpp/arcticdb/processing/test/ast_test_helpers.hpp
alexowens90, line 1: nit: should be in `unit_test_srcs`

cpp/arcticdb/processing/expression_node.cpp
alexowens90: Ah I see this isn't part of the Python repr, just for the memoization, ignore me

cpp/arcticdb/processing/test/test_expression.cpp
```diff
@@ -28,19 +32,100 @@ TEST(ExpressionNode, AddBasic) {
...
+    for (auto j = 0; j < 20; ++j) {
+        auto v1 = proc.segments_->at(0)->scalar_at<uint64_t>(j, 1).value();
+        auto v2 = proc.segments_->at(0)->scalar_at<uint64_t>(j, 2).value();
+        ASSERT_EQ(result_col->scalar_at<uint64_t>(j), (v1 + v2) * (v1 + v2));
+    }
+
+    const std::string add_label = R"((Column["thing1"] ADD Column["thing2"]))";
+    ASSERT_EQ(add1->label_, add_label);
+    ASSERT_EQ(add2->label_, add_label);
+    ASSERT_EQ(proc.computed_data_.count(add_label), 1);
```
alexowens90: nit: `contains` over `count == 1` everywhere

cpp/arcticdb/processing/test/test_filter_and_project_sparse.cpp
```diff
@@ -26,86 +27,47 @@ class FilterProjectSparse : public testing::Test {
...
             std::string_view left_column_name,
             const std::variant<std::string_view, double, std::unordered_set<double>>& right_input, OperationType op
     ) {
-        const std::string root_node_name("binary filter");
-        const std::string value_name("value");
-        const std::string value_set_name("value set");
-        expression_context->root_node_name_ = ExpressionName(root_node_name);
-        std::shared_ptr<ExpressionNode> expression_node;
+        ExpressionChild right;
         util::variant_match(
```
alexowens90:
```suggestion
        ExpressionChild right = util::variant_match(
```

cpp/arcticdb/processing/test/test_expression.cpp
```diff
@@ -28,19 +32,100 @@ TEST(ExpressionNode, AddBasic) {
...
+    StreamId symbol("test_longhand");
+    auto wrapper =
+            SinkWrapper(symbol, {scalar_field(DataType::UINT64, "thing1"), scalar_field(DataType::UINT64, "thing2")});
+
+    for (auto j = 0; j < 20; ++j) {
+        wrapper.aggregator_.start_row(timestamp(j))([&](auto&& rb) {
+            rb.set_scalar(1, j);
+            rb.set_scalar(2, j + 1);
+        });
+    }
```
vasil-pashov: What do you think about using `create_dense_segment` from `segment_generation_utils.hpp` instead? I find it a bit more readable. It'll look something like

```cpp
SegmentInMemory seg = create_dense_segment(
  {scalar_field(DataType::UINT64, "thing1"), scalar_field(DataType::UINT64, "thing2")},
  std::ranges::views::iota<uint64_t, uitnt64_t>(0, 20),
  std::ranges::views::iota<uint64_t, uitnt64_t>(1, 21)
);
```

there's also `slice_data_into_segments` if you want to simulate how arcticdb slices segments (including getting the index as first column in every slice)

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -75,43 +76,186 @@ ColumnWithStrings::ColumnWithStrings(
...
+    if (std::holds_alternative<ExpressionNode::Leaf>(a.kind_)) {
+        const auto& leaf_a = std::get<ExpressionNode::Leaf>(a.kind_);
+        const auto& leaf_b = std::get<ExpressionNode::Leaf>(b.kind_);
+        if (leaf_a.index() != leaf_b.index()) {
+            return false;
+        }
+        // Now we know that leaf_a and leaf_b hold the same variant type member
+        return util::variant_match(
+                leaf_a,
+                [&leaf_b](const ColumnName& a_col) { return a_col.value == std::get<ColumnName>(leaf_b).value; },
```
vasil-pashov, line 154: No need to take any action, but this might break if we have columns with the same name. There are already quite a lot of things that don't work properly in this case, so I don't think we should worry about it.

vasil-pashov (approved):

cpp/arcticdb/processing/expression_node.cpp
poodlewars: Yeah, everything is built bottom up.

cpp/arcticdb/processing/query_planner.cpp
poodlewars: FilterClause already has this assertion,

```
    explicit FilterClause(
            std::unordered_set<std::string> input_columns, ExpressionContext expression_context,
            std::optional<PipelineOptimisation> optimisation
    ) :
        expression_context_(std::make_shared<ExpressionContext>(std::move(expression_context))),
        optimisation_(optimisation.value_or(PipelineOptimisation::SPEED)) {
                expression_context_->root_ && expression_context_->root_->is_operation(),
                "FilterClause AST would produce a column, not a bitset"
        );
        clause_info_.input_columns_ = std::move(input_columns);
    }
```

It works because the bare bool filter gets wrapped in an IDENTITY operation in the Python layer.

alexowens90: Ah cool

alexowens90 (approved):

PR #3202, opened 2026-06-25
Support staging arrow strings with sorting

IvoDD:
#### Reference Issues/PRs
Monday ref: 18190648152

Adds support for `lib.stage(arrow_with_strings, sort_by_index=True)` and other sorting on arrow strings.

Fixes a bug where `lib.stage(numpy_buffer, sort_by_index=True)` modifies the `numpy_buffer` in place.

This is done by replacing the `incomplete_segment_from_tensor_frame` with a new `incomplete_segment_from_frame` which works for both arrow and numpy buffers by using `WriteToSegmentTask`. 
Also introduces a new `enum CopyMode{ALWAYS, IF_NEEDED}`:
- `IF_NEEDED` is the old behavior where numeric contiguous buffers can end up represented as a view of the source data. - `ALWAYS` is a new option that allows explicitly requesting owned data so that the inplace sort does not modify user's input data.

Also refactors a few confusing places of code around the required change - inline comments in PR explain each of these changes. But in short the changes are:
- better `make_timeseries_descriptor` API
- use correct `stream_id` in `InputFrame::desc_for_tsd`

#### Any other comments?
The change is split in three commits which can be reviewed separately (but the mostly touch different lines so fine to review together as well)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/column_store/column.hpp
+++ b/cpp/arcticdb/column_store/column.hpp
@@ -53,6 +53,12 @@ struct JiveTable {
 
 enum class ExtraBufferType : uint8_t { OFFSET, STRING, ARRAY, BITMAP };
 
+// Controls whether external blocks get constructed via owning memcpy or with a non owning pointer assignment
+// ALWAYS - Always copies, result owns its memory
+// IF_NEEDED - Uses a non owning view when possible (e.g. numeric types). Copies when memory layout is different (e.g.
+// strings)
+enum class CopyMode : uint8_t { ALWAYS, IF_NEEDED };
+
 // Specifies a way to index extra buffers.
 // We can attach extra buffers to each offset and type. This is used for OutputFormat::ARROW to store the extra buffers
 // required to construct a string buffer.
@@ -313,7 +319,7 @@ class Column {
     }
 
     template<class T, std::enable_if_t<std::is_integral_v<T> || std::is_floating_point_v<T>, int> = 0>
-    void set_external_block(ssize_t row_offset, T* val, size_t size) {
+    void set_external_block(ssize_t row_offset, T* val, size_t size, CopyMode copy_mode = CopyMode::IF_NEEDED) {
         util::check_arg(
                 last_logical_row_ + 1 == row_offset,
                 "set_external_block expected row {}, actual {} ",
@@ -321,7 +327,15 @@ class Column {
                 row_offset
         );
         auto bytes = sizeof(T) * size;
-        const_cast<ChunkedBuffer&>(data_.buffer()).add_external_block(reinterpret_cast<const uint8_t*>(val), bytes);
+        const auto* val_bytes_ptr = reinterpret_cast<const uint8_t*>(val);
+        if (copy_mode == CopyMode::ALWAYS) {
+            auto owned_data = allocate_data(bytes);
+            memcpy(owned_data, val_bytes_ptr, bytes);
+        } else {
+            auto& buffer = const_cast<ChunkedBuffer&>(data_.buffer());
+            buffer.add_external_block(val_bytes_ptr, bytes);
+        }
+        advance_data(bytes);
         last_logical_row_ += static_cast<ssize_t>(size);
         last_physical_row_ = last_logical_row_;
     }
--- a/cpp/arcticdb/column_store/column.hpp
+++ b/cpp/arcticdb/column_store/column.hpp
@@ -319,10 +319,10 @@ class Column {
     }
 
     template<class T, std::enable_if_t<std::is_integral_v<T> || std::is_floating_point_v<T>, int> = 0>
-    void set_external_block(ssize_t row_offset, T* val, size_t size, CopyMode copy_mode = CopyMode::IF_NEEDED) {
+    void set_dense_block(ssize_t row_offset, T* val, size_t size, CopyMode copy_mode = CopyMode::IF_NEEDED) {
         util::check_arg(
                 last_logical_row_ + 1 == row_offset,
-                "set_external_block expected row {}, actual {} ",
+                "set_dense_block expected row {}, actual {} ",
                 last_logical_row_ + 1,
                 row_offset
         );
--- a/cpp/arcticdb/pipeline/frame_utils.cpp
+++ b/cpp/arcticdb/pipeline/frame_utils.cpp
@@ -75,7 +75,7 @@ TimeseriesDescriptor index_descriptor_from_frame(
 ) {
     return make_timeseries_descriptor(
             frame->num_rows + existing_rows,
-            frame->desc_for_tsd(),
+            frame->compute_desc_for_tsd(),
             frame->norm_meta,
             std::move(frame->user_meta),
             std::move(next_key),
--- a/cpp/arcticdb/pipeline/frame_utils.cpp
+++ b/cpp/arcticdb/pipeline/frame_utils.cpp
@@ -15,9 +15,9 @@ namespace arcticdb {
 TimeseriesDescriptor make_timeseries_descriptor(
         // TODO: It would be more explicit to use uint64_t instead of size_t. Not doing now as it involves a lot of type
         // changes and needs to be done carefully.
-        size_t total_rows, const StreamDescriptor& desc, proto::descriptors::NormalizationMetadata&& norm_meta,
-        std::optional<proto::descriptors::UserDefinedMetadata>&& um, std::optional<AtomKey>&& prev_key,
-        std::optional<AtomKey>&& next_key, bool bucketize_dynamic
+        size_t total_rows, const StreamDescriptor& desc, const proto::descriptors::NormalizationMetadata& norm_meta,
+        std::optional<proto::descriptors::UserDefinedMetadata>&& um, std::optional<AtomKey>&& next_key,
+        bool bucketize_dynamic
 ) {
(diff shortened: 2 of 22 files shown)


cpp/arcticdb/stream/incompletes.cpp
```diff
@@ -247,33 +201,15 @@ void do_sort(SegmentInMemory& mutable_seg, const std::vector<std::string> sort_c
...
-        }
-        return seg;
-    }();
+    // We use `CopyMode::Always` so that the resulting `segment` owns its buffers and we can sort inplace below.
+    // This is inefficient because `split` below would also copy the memory
+    // TODO: A better approach would be to:
+    // 1. Produce a JiveTable for the InputFrame directly
+    // 2. Modify WriteToSegmentTask to accept a JiveTable so it can copy from InputFrame and reorder according to the
+    // JiveTable This will avoid doing the extra copy and allocation we're doing here for sorting. It will also allow us
+    // to unify `write_incomplete_with_sorting` with `write_incomple_frame`
```
IvoDD: Monday ref: 18053150625

cpp/arcticdb/pipeline/input_frame.hpp
```diff
@@ -115,7 +116,6 @@ struct InputFrame {
     std::vector<FieldData> columns_;
     std::vector<sparrow::record_batch> arrow_buffer_owners_;
     StreamDescriptor desc_;
-    std::optional<StreamDescriptor> desc_for_tsd_;
```
IvoDD, line 118: The `desc_for_tsd_` state was removed because it was error prone to store a very similar descriptor in two places.

Namely [here](https://github.com/man-group/ArcticDB/blob/aa3c5c33badcf5b5e38dcfad7d542dd2304a917c/cpp/arcticdb/python/python_to_tensor_frame.cpp#L379) we were updating the `stream_id` on the `desc_` but not on the `desc_for_tsd_`.

So I switched from keeping it as a state, to computing it directly in `desc_for_tsd()` method.

cpp/arcticdb/pipeline/frame_utils.cpp
```diff
@@ -33,9 +33,6 @@ TimeseriesDescriptor make_timeseries_descriptor(
     if (user_meta)
         *proto->mutable_user_meta() = std::move(*user_meta);
 
-    if (prev_key)
-        proto->mutable_next_key()->CopyFrom(key_to_proto(prev_key.value()));
```
IvoDD, line 37: I removed all mentions of `prev_key`. It was just populating the same `mutable_next_key` on the timeseries descriptor. Now throughout the codebase we refer to the `next_key` when looking through the `APPEND_DATA` linked list.

cpp/arcticdb/pipeline/frame_utils.cpp
```diff
@@ -15,9 +15,9 @@ namespace arcticdb {
 TimeseriesDescriptor make_timeseries_descriptor(
         // TODO: It would be more explicit to use uint64_t instead of size_t. Not doing now as it involves a lot of type
         // changes and needs to be done carefully.
-        size_t total_rows, const StreamDescriptor& desc, proto::descriptors::NormalizationMetadata&& norm_meta,
-        std::optional<proto::descriptors::UserDefinedMetadata>&& um, std::optional<AtomKey>&& prev_key,
-        std::optional<AtomKey>&& next_key, bool bucketize_dynamic
+        size_t total_rows, const StreamDescriptor& desc, const proto::descriptors::NormalizationMetadata& norm_meta,
```
IvoDD, line 18: I made the `NormalicationMetadata` `const&` instead of `&&`. 

That better shows the ownership semantics we only `CopyFrom` the input `norm_meta`.

This allows to clear several places in incompletes code where we used `norm_meta` after passing it here via `std::move` which look like use after moves. E.g. this is fixed [use](https://github.com/man-group/ArcticDB/blob/1acdb08c7d1b140d4754ac4cd3cfe8ec10bfc4ce/cpp/arcticdb/stream/incompletes.cpp#L309) after [move](https://github.com/man-group/ArcticDB/blob/1acdb08c7d1b140d4754ac4cd3cfe8ec10bfc4ce/cpp/arcticdb/stream/incompletes.cpp#L297)

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -91,7 +84,17 @@ StreamDescriptor& InputFrame::desc() { return desc_; }
 
 const StreamDescriptor& InputFrame::desc() const { return const_cast<InputFrame*>(this)->desc(); }
 
-const StreamDescriptor& InputFrame::desc_for_tsd() { return desc_for_tsd_.has_value() ? *desc_for_tsd_ : desc_; }
+StreamDescriptor InputFrame::desc_for_tsd() const {
```
alexowens90: This is now a lot more expensive than it was. I agree the previous implementation was error prone though. Is the non-const `desc()` method widely used? If not, we could remove it and make all modifications go through methods that enforce consistency between `desc_` and `desc_for_tsd_`

python/tests/util/arrow.py
```diff
@@ -14,6 +14,14 @@ def to_format(table: Union[pa.Table, pl.DataFrame], arrow_output_format: OutputF
     return table
 
 
+def deep_copy(table: Union[pa.Table, pl.DataFrame]) -> Union[pa.Table, pl.DataFrame]:
+    """Deep copy of an arrow or polars table"""
+    if isinstance(table, pl.DataFrame):
+        return pl.from_arrow(deep_copy(table.to_arrow()))
+    # There is explicit no deep clone pyarrow api so we use `pyarrow.take(num_rows)`
+    return table.take(list(range(table.num_rows)))
```
alexowens90: I suspect this will silently change the record batch structure of the table? If so, could do this record batch by record batch.
In either case, I would expect `select` to be more efficient than `take`

python/tests/util/arrow.py
```diff
@@ -14,6 +14,14 @@ def to_format(table: Union[pa.Table, pl.DataFrame], arrow_output_format: OutputF
     return table
 
 
+def deep_copy(table: Union[pa.Table, pl.DataFrame]) -> Union[pa.Table, pl.DataFrame]:
+    """Deep copy of an arrow or polars table"""
+    if isinstance(table, pl.DataFrame):
+        return pl.from_arrow(deep_copy(table.to_arrow()))
+    # There is explicit no deep clone pyarrow api so we use `pyarrow.take(num_rows)`
```
alexowens90: Does `copy.deepcopy` not do the right thing?

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90, line 829: Could add one more test sorting on non-index columns

cpp/arcticdb/pipeline/input_frame.cpp
ognyanstoimenov: +1, also it now returns an unnecessary copy in the tensors_only case but the StreamDescriptor is just 2 pointers so that is fine

cpp/arcticdb/column_store/column.hpp
```diff
@@ -313,15 +319,23 @@ class Column {
     }
 
     template<class T, std::enable_if_t<std::is_integral_v<T> || std::is_floating_point_v<T>, int> = 0>
-    void set_external_block(ssize_t row_offset, T* val, size_t size) {
+    void set_external_block(ssize_t row_offset, T* val, size_t size, CopyMode copy_mode = CopyMode::IF_NEEDED) {
```
ognyanstoimenov: This looks a bit weird now, as set_external_block doesn't set an external block in some cases. Maybe a differrent API makes more sense for the CopyMode::ALWAYS path but that would probably require changes in lots of places, so no need to do it now.

ognyanstoimenov (approved):

cpp/arcticdb/pipeline/input_frame.cpp
IvoDD: I agree the method is more expensive now, but the overall work is the same. 

Previously we were doing the same work on InputFrame construction. Now we do it on `desc_for_tsd()` call, which we only do once per InputFrame (via `get_merged_tsd` for append and update and via `index_descriptor_for_frame` for write).

I'm just renaming it to `compute_desc_for_tsd` to make it more obvious that it is doing work and people don't end up calling it more than once for an `InputFrame`. I prefer avoiding the extra complexity in the state.

python/tests/util/arrow.py
IvoDD: Ah it does the correct thing for pyarrow (but surprisingly not for polars), so I used that for pyarrow and round trip through it if needed.

python/tests/util/arrow.py
IvoDD: Good catch, using copy.deepcopy as suggested below preserves the record batch structure as well.

cpp/arcticdb/column_store/column.hpp
IvoDD: Renamed to `set_dense_block` which makes it clearer with the `copy_mode` arg

alexowens90 (approved):

IvoDD:
Merging with failing benchmarks. Looks like flaky query builder benchmarks.

PR #3203, opened 2026-06-26
Add compiler argument handling to CMake commands

G-D-Petrov:
Updated CMake commands to use compiler arguments for better flexibility.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


phoebusm (approved):

vasil-pashov (changes requested):
I'd rather not merge this. I think we went too far with these Makefiles. It started as a way to make it easier for Claude to invoke commands, which although I don't like (I'd prefer CMake custom targets) I don't care much about. But I'm against configuring CMake via Makefiles because it makes it hard to reason about and also we get in the situation where we use make to generate cmake to generate make.

PR #3204, opened 2026-06-26
Merge filter clauses in to one expression during query builder evaluation

poodlewars:
11643672444

Only review https://github.com/man-group/ArcticDB/pull/3204/changes/412ee424d92e1a1aa334e0dc807a5500c35715ca - the rest of this diff is already in PR #3201 

- Merge filter clauses in to one to prevent materializing intermediate segments, and to unify with column stats evaluation which already does this. This improves performance and memory use for query builders with long chains of filters like:

```
        q = QueryBuilder()
        q = q[q["v3"] > 0.0]  # keeps ~all rows
        q = q[q["v3"] < 99.0]  # keeps ~99%
        q = q[q["v2"] > 1]  # keeps ~93%
        q = q[q["v2"] < 15]  # keeps ~93%
        q = q[q["v1"] > 1]  # keeps ~80%
        q = q[q["v1"] < 5]  # keeps ~80%
        q = q[q["v3"] < 50.0]  # narrows to ~50%
```

Benchmark results:

```
| Change   | Before [5077a00b] <master>   | After [b2bb249c] <aseaton/column-stats/11643672444/and-clauses-pipeline>   |   Ratio | Benchmark (Parameter)                                                                      |
|----------|------------------------------|----------------------------------------------------------------------------|---------|--------------------------------------------------------------------------------------------|
| -        | 258M                         | 208M                                                                       |    0.81 | query_builder.QueryBuilderFunctions.peakmem_filtering_chained(1000000, <Storage.LMDB: 2>)  |
| -        | 1.68G                        | 1.14G                                                                      |    0.68 | query_builder.QueryBuilderFunctions.peakmem_filtering_chained(10000000, <Storage.LMDB: 2>) |
| -        | 55.5±4ms                     | 35.4±2ms                                                                   |    0.64 | query_builder.QueryBuilderFunctions.time_filtering_chained(10000000, <Storage.LMDB: 2>)    |
| -        | 21.8±0.4ms                   | 12.6±0.08ms                                                                |    0.58 | query_builder.QueryBuilderFunctions.time_filtering_chained(1000000, <Storage.LMDB: 2>)     |
```

- Fix a bug with column stats evaluation where it was incorrectly evaluating filters that came after a RowRange filter (exclude from release notes)

### DIFF
--- a/cpp/arcticdb/pipeline/column_stats_filter.cpp
+++ b/cpp/arcticdb/pipeline/column_stats_filter.cpp
@@ -16,7 +16,6 @@
 #include <arcticdb/pipeline/value.hpp>
 #include <arcticdb/log/log.hpp>
 #include <arcticdb/stream/stream_utils.hpp>
-#include <arcticdb/processing/query_planner.hpp>
 
 #include <iterator>
 #include <unordered_set>
@@ -457,13 +456,16 @@ FilterQuery<index::IndexSegmentReader> create_column_stats_filter(
 }
 
 ColumnStatsQueryMetadata::ColumnStatsQueryMetadata(const std::vector<std::shared_ptr<Clause>>& clauses) {
-    // The clauses eligible for column stats use:
+    // We apply column stats filtering to a "prefix" of clauses that are eligible based on the rules below.
+    // Column stats are not used for any clauses after this "prefix".
     // - FilterClauses contribute filter expressions and columns of interest
     // - DateRangeClauses contribute their range
-    // - RowRangeClauses are skipped
+    // - A RowRangeClause ends the prefix unless it is the first clause, because a leading RowRangeClause
+    //   selects absolute row positions. A RowRangeClause anywhere else is over a changed dataset,
+    //   so a filter after it must not drive pruning (that would change which rows are "first/last N").
     // - Anything else (Resample / GroupBy / Project) ends the prefix because those clauses
     // transform the data so stats computed on the original segments are no longer valid.
-    for (const auto& clause : clauses) {
+    for (auto&& [idx, clause] : folly::enumerate(clauses)) {
         auto& clause_type = folly::poly_type(*clause);
         if (clause_type == typeid(DateRangeClause)) {
             const auto& date_range_clause = folly::poly_cast<DateRangeClause>(*clause);
@@ -476,7 +478,10 @@ ColumnStatsQueryMetadata::ColumnStatsQueryMetadata(const std::vector<std::shared
             continue;
         }
         if (clause_type == typeid(RowRangeClause)) {
-            continue;
+            if (idx == 0) {
+                continue;
+            }
+            break;
         }
         if (clause_type != typeid(FilterClause)) {
             break;
@@ -546,10 +551,12 @@ namespace {
 FilterQuery<index::IndexSegmentReader> build_filter_from_column_stats_data(
         ColumnStatsData&& column_stats, std::vector<std::shared_ptr<ExpressionContext>>&& filter_expressions
 ) {
-    util::check(!filter_expressions.empty(), "Expected at least one filter expression");
-    ARCTICDB_DEBUG(log::version(), "AND-ing expression contexts from filters");
-    ExpressionContext overall_context = and_filter_expression_contexts(filter_expressions);
-    ARCTICDB_DEBUG(log::version(), "Creating column stats filter");
+    util::check(
+            filter_expressions.size() == 1,
+            "Expected exactly one filter expression for column stats (filters are merged in plan_query), got {}",
+            filter_expressions.size()
+    );
+    ExpressionContext overall_context = *filter_expressions.front();
     return create_column_stats_filter(std::move(column_stats), std::move(overall_context));
 }
 
--- a/cpp/arcticdb/pipeline/column_stats_filter.cpp
+++ b/cpp/arcticdb/pipeline/column_stats_filter.cpp
@@ -25,7 +25,7 @@ namespace arcticdb {
 bool is_column_stats_enabled() { return ConfigsMap::instance()->get_int("ColumnStats.UseForQueries", 0) == 1; }
 
 bool ColumnStatsQueryMetadata::should_try_column_stats_read() const {
-    return is_column_stats_enabled() && !filter_expressions.empty();
+    return is_column_stats_enabled() && filter_expression.has_value();
 }
 
 StatsVariantData dispatch_unary_stats(const StatsVariantData& left, OperationType operation);
@@ -469,12 +469,12 @@ ColumnStatsQueryMetadata::ColumnStatsQueryMetadata(const std::vector<std::shared
         auto& clause_type = folly::poly_type(*clause);
         if (clause_type == typeid(DateRangeClause)) {
             const auto& date_range_clause = folly::poly_cast<DateRangeClause>(*clause);
-            if (date_range.has_value()) {
-                date_range->first = std::max(date_range->first, date_range_clause.start());
-                date_range->second = std::min(date_range->second, date_range_clause.end());
-            } else {
-                date_range = std::make_pair(date_range_clause.start(), date_range_clause.end());
-            }
+            util::check(
+                    !date_range.has_value(),
+                    "Expected at most one DateRangeClause in the column stats prefix (date ranges are merged in "
+                    "plan_query)"
+            );
+            date_range = std::make_pair(date_range_clause.start(), date_range_clause.end());
             continue;
         }
         if (clause_type == typeid(RowRangeClause)) {
@@ -487,12 +487,12 @@ ColumnStatsQueryMetadata::ColumnStatsQueryMetadata(const std::vector<std::shared
             break;
         }
         const auto& filter = folly::poly_cast<FilterClause>(*clause);
-        filter_expressions.emplace_back(filter.expression_context_);
         util::check(
-                filter.clause_info().input_columns_.has_value(),
-                "FilterClause is missing input_columns_ — Python bindings should always populate this"
+                !filter_expression.has_value(),
+                "Expected at most one FilterClause in the column stats prefix (filters are merged in plan_query)"
         );
-        for (const auto& col : *filter.clause_info().input_columns_) {
+        filter_expression = filter.expression_context_;
+        for (const auto& col : filter.clause_info().input_columns_) {
             columns_of_interest.insert(col);
         }
     }
@@ -546,22 +546,6 @@ SegmentInMemory partial_decode_column_stats_segment(
     return partial;
 }
 
-namespace {
-
-FilterQuery<index::IndexSegmentReader> build_filter_from_column_stats_data(
-        ColumnStatsData&& column_stats, std::vector<std::shared_ptr<ExpressionContext>>&& filter_expressions
-) {
-    util::check(
-            filter_expressions.size() == 1,
-            "Expected exactly one filter expression for column stats (filters are merged in plan_query), got {}",
-            filter_expressions.size()
-    );
-    ExpressionContext overall_context = *filter_expressions.front();
-    return create_column_stats_filter(std::move(column_stats), std::move(overall_context));
-}
-
-} // namespace
-
 FilterQuery<index::IndexSegmentReader> create_column_stats_filter(
         std::shared_ptr<Segment> column_stats_compressed, const TimeseriesDescriptor& tsd,
         ColumnStatsQueryMetadata&& query_metadata
@@ -573,7 +557,8 @@ FilterQuery<index::IndexSegmentReader> create_column_stats_filter(
     SegmentInMemory partial_segment =
             partial_decode_column_stats_segment(*column_stats_compressed, tsd, query_metadata.columns_of_interest);
     ColumnStatsData column_stats{std::move(partial_segment), tsd, query_metadata.date_range};
-    return build_filter_from_column_stats_data(std::move(column_stats), std::move(query_metadata.filter_expressions));
+    ExpressionContext expression_context = *query_metadata.filter_expression.value();
+    return create_column_stats_filter(std::move(column_stats), std::move(expression_context));
 }
 
 } // namespace arcticdb
--- a/cpp/arcticdb/pipeline/column_stats_filter.cpp
+++ b/cpp/arcticdb/pipeline/column_stats_filter.cpp
@@ -25,7 +25,7 @@ namespace arcticdb {
 bool is_column_stats_enabled() { return ConfigsMap::instance()->get_int("ColumnStats.UseForQueries", 0) == 1; }
 
 bool ColumnStatsQueryMetadata::should_try_column_stats_read() const {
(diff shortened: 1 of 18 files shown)


cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -505,10 +510,12 @@ namespace {
 FilterQuery<index::IndexSegmentReader> build_filter_from_column_stats_data(
         ColumnStatsData&& column_stats, std::vector<std::shared_ptr<ExpressionContext>>&& filter_expressions
 ) {
-    util::check(!filter_expressions.empty(), "Expected at least one filter expression");
-    ARCTICDB_DEBUG(log::version(), "AND-ing expression contexts from filters");
-    ExpressionContext overall_context = and_filter_expression_contexts(filter_expressions);
-    ARCTICDB_DEBUG(log::version(), "Creating column stats filter");
+    util::check(
```
alexowens90: Feels like the API to this should now take a single `ExpressionContext` rather than a vector?

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -8,8 +8,90 @@
...
+
+// Move each DateRangeClause as far left as it can go, so that filters that were only separated by a
+// date range end up next to each other and can be merged. Each date range ends up at the front of the
+// run of clauses it is allowed to move past, keeping the original order, eg
+// [F1, F2, DR1, F3, F4, DR2] -> [DR1, DR2, F1, F2, F3, F4]
+void move_date_ranges_left(std::vector<ClauseVariant>& clauses) {
+    size_t insert_pos = 0; // location of the clause we can move a date range left to
+    // Move each date range we see left to insert_pos. If we have [F1, F2, DR1, F3] then we hit rotate(start=0,
+    // middle=2, last=3). This rotates within [0, 3) and makes index 2 the new start, that is it rotates [F1, F2, DR1]
+    // until DR1 is the new start. This results in [DR1, F1, F2, F3].
```
alexowens90: Not that it really matters, but my reading of the `std::rotate` docs is that this will result in `[DR1, F3, F1, F2]`

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -8,8 +8,90 @@
...
+            result.push_back(std::move(clause));
+        }
+    }
+    flush_run();
+    return result;
+}
+
+} // namespace
+
 std::vector<ClauseVariant> plan_query(std::vector<ClauseVariant>&& clauses) {
```
alexowens90, line 116: Doesn't need to change now, but the reason this is called so early is because I wasn't aware of `folly::poly_cast`, and thought this stuff needed to work with `ClauseVariant`. It could be moved to a more sensible place with a small refactor to use `poly_cast` instead of `holds_alternative`

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -8,8 +8,90 @@
...
+std::shared_ptr<FilterClause> merge_filter_run(const std::vector<std::shared_ptr<FilterClause>>& filters) {
+    if (filters.size() == 1) {
+        return filters.front();
+    }
+    std::vector<std::shared_ptr<ExpressionContext>> expression_contexts;
+    std::unordered_set<std::string> input_columns;
+    bool any_memory = false;
+    for (const auto& filter : filters) {
+        expression_contexts.push_back(filter->expression_context_);
+        if (filter->clause_info_.input_columns_.has_value()) {
```
alexowens90: Is this condition ever false?

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -22,7 +104,8 @@ std::vector<ClauseVariant> plan_query(std::vector<ClauseVariant>&& clauses) {
             }
         });
     }
-    return clauses;
+    move_date_ranges_left(clauses);
+    return merge_consecutive_filter_clauses(std::move(clauses));
```
alexowens90, line 129: As implemented, a pipeline of clauses like `[F1, P1, F2]` where `F` is filter and `P` is projection will not merge the filters together. I think an equivalent of `move_date_ranges_left` for `move_projections_right` will take care of this

alexowens90: We could also merge adjacent `DateRange` clauses together, although I doubt this happens much in practice

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1912,3 +1912,264 @@ def test_filter_regex_comma_separated_strings(lmdb_version_store_v1, sym, dynami
...
+    expected = expected[expected["b"] < 10]
+    assert not expected.empty
+    assert_frame_equal(expected, received)
+
+
+def test_filter_merge_date_range_between(
+    lmdb_version_store_tiny_segment, any_output_format, column_stats_filtering_enabled_and_disabled
+):
+    """We move DateRangeClauses between two filters left in the query plan so we can merge the filters,
+    and in future restrict the data that the filter needs to evaluate against. This test just checks
```
alexowens90: What is "in future" referring to? Evaluating a date range clause before a filter will only perform the filter against data within the specified date range already

python/tests/unit/arcticdb/version_store/test_filtering.py
```diff
@@ -1912,3 +1912,264 @@ def test_filter_regex_comma_separated_strings(lmdb_version_store_v1, sym, dynami
...
+    lib._set_output_format_for_pipeline_tests(any_output_format)
+    symbol = "test_filter_merge_dynamic_schema_missing_column"
+    df0 = pd.DataFrame({"a": [1, 2], "b": [10, 20]}, index=pd.date_range("2000-01-01", periods=2))
+    df1 = pd.DataFrame({"a": [3, 4]}, index=pd.date_range("2000-01-03", periods=2))
+    lib.write(symbol, df0)
+    lib.append(symbol, df1)
+
+    q = QueryBuilder()
+    q = q[q["a"] > 1]
+    q = q[q["b"] < 100]
```
alexowens90, line 2010: Could use an expression that filters out some data

cpp/arcticdb/processing/query_planner.cpp
alexowens90: > As implemented, a pipeline of clauses like `[F1, P1, F2]` where `F` is filter and `P` is projection will not merge the filters together. I think an equivalent of `move_date_ranges_left` for `move_projections_right` will take care of this

Ah I see there's a comment in `test_filter_merge_projection_between_filters` about why we can't always do this. Worth putting that comment here as well to help the next reader (and in future we could move filters left that don't depend on a column created by a projection, but not for now)

alexowens90 (approved):

vasil-pashov (approved):
Just one question. Does this merge date ranges. E.g.

q.filter()
q.date_range()
q.filter()
q.date_range()

will push both date ranges left, but will it combine them into a single date range?

poodlewars:
> Just one question. Does this merge date ranges. E.g.
> 
> q.filter() q.date_range() q.filter() q.date_range()
> 
> will push both date ranges left, but will it combine them into a single date range?

It just pushes them left, it doesn't combine them. Happy to add a change to combine them, especially if there are already some utilities to do it. I'll have a look.

cpp/arcticdb/processing/query_planner.cpp
poodlewars: I see, we have

```
            .def("add_clauses", [](ReadQuery& self, std::vector<ClauseVariant> clauses) {
                clauses = plan_query(std::move(clauses));
                std::vector<std::shared_ptr<Clause>> _clauses;
                self.needs_post_processing = false;
                for (auto&& clause : clauses) {
                    util::variant_match(clause, [&](auto&& clause) {
                                !clause->clause_info().multi_symbol_,
                                "Multi-symbol clause cannot be used on a single symbol"
                        );
                        _clauses.emplace_back(std::make_shared<Clause>(*clause));
                    });
                }
                self.add_clauses(std::move(_clauses));
            });
```

in the pybind layer and you wanted to do the query planning before we type-erase in to `Clause`.

It would be good to clean this up - I'll do it in a standalone PR after this one.

python/tests/unit/arcticdb/version_store/test_filtering.py
poodlewars: Sorry this was baffling. Removed the "in future" part.

I think I was half thinking of those ideas about using column stats to decide the best order to evaluate the filter AST, running selective predicates first and presenting a smaller segment for other predicates to evaluate against, and had a loss of sanity.

cpp/arcticdb/processing/query_planner.cpp
poodlewars: Thanks, changed `input_columns_` to be non-optional.

cpp/arcticdb/processing/query_planner.cpp
poodlewars: I think the comment is correct, https://en.cppreference.com/cpp/algorithm/rotate . Substituting values:

We swap the elements in the range [0, 3) such that elements in [0, 2) are placed after [2, 3) while preserving order of elements in both ranges.

On [F1, F2, DR1, F3] this gives:

- Only consider [F1, F2, DR1]
- [F1, F2] are placed after [DR1]
- This gives [DR1, F1, F2]
- After putting the F3 back in we get [DR1, F1, F2, F3]

The last bound is non-inclusive so the F3 isn't touched by the `std::rotate`.

poodlewars: This has changed anyway now that I'm merging date range clauses together.

cpp/arcticdb/processing/query_planner.cpp
poodlewars: Drafted https://github.com/man-group/ArcticDB/pull/3254 will submit after this PR is in.

cpp/arcticdb/processing/query_planner.cpp
alexowens90: Oh I'd misread the last=3, thought it was end, yeh, you're right

cpp/arcticdb/processing/clause_utils.hpp
```diff
@@ -49,7 +49,8 @@ struct ClauseInfo {
     bool can_combine_with_column_selection_{true};
     // The names of the columns that are needed for this clause to make sense
     // Could either be on disk, or columns created by earlier clauses in the processing pipeline
-    std::optional<std::unordered_set<std::string>> input_columns_{std::nullopt};
+    // Empty if this clause does not have specific input columns (e.g. it operates on all columns)
+    std::unordered_set<std::string> input_columns_;
```
alexowens90, line 53: My original comment actually meant that `FilterClause` would always have some columns.
I thought there would have been a distinction between `std::nullopt` and an empty set, but I can't find any. I think the implementation in `read_pipeline.cpp` changed at some point after I added this so that the distinction didn't matter anymore.

cpp/arcticdb/pipeline/column_stats_filter.hpp
```diff
@@ -124,8 +124,8 @@ class ColumnStatsData {
 };
 
 struct ColumnStatsQueryMetadata {
-    // Filter expressions we can apply column stats to.
-    std::vector<std::shared_ptr<ExpressionContext>> filter_expressions;
+    // Filter expression we can apply column stats to.
+    std::optional<std::shared_ptr<ExpressionContext>> filter_expression;
```
alexowens90: nit: Prefer using nullity of `shared_ptr` to indicate `std::nullopt` equivalent unless that represents an explicit state

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -26,23 +27,43 @@ bool is_date_range(const ClauseVariant& clause) {
...
-            ++insert_pos;
-        } else if (!date_range_can_move_to_left_of(clauses.at(i))) {
-            insert_pos = i + 1;
+// Move each DateRangeClause as far left as it can go and merge any that end up adjacent into a single
+// DateRangeClause, so that filters that were only separated by a date range end up next to
+// each other (and can then be merged by merge_consecutive_filter_clauses), eg
+// [F1, F2, DR1, F3, F4, DR2] -> [DR_12, F1, F2, F3, F4]
+std::vector<ClauseVariant> move_and_merge_date_ranges_left(std::vector<ClauseVariant>&& clauses) {
+    std::vector<ClauseVariant> result;
+    result.reserve(clauses.size());
```
alexowens90: I add a [utility](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/util/collection_utils.hpp#L61-L66) for this (unbelievably common) pattern

cpp/arcticdb/processing/query_planner.cpp
```diff
@@ -26,23 +27,43 @@ bool is_date_range(const ClauseVariant& clause) {
...
+            result.push_back(std::move(clause));
+        }
+        run.clear();
+    };
+    for (auto& clause : clauses) {
+        if (is_date_range(clause)) {
+            auto& date_range = std::get<std::shared_ptr<DateRangeClause>>(clause);
+            merged_date_range = merged_date_range ? std::make_shared<DateRangeClause>(
+                                                            std::max(merged_date_range->start_, date_range->start_),
+                                                            std::min(merged_date_range->end_, date_range->end_)
```
alexowens90, line 56: I expect everything will work, but this could result in disjoint date ranges producing an odd `DateRangeClause`
e.g. `[0, 10]` and `[20, 30]` would combine to `[20, 10]`.
There don't seem to be any checks for this anywhere when constructing a date range clause "normally"

alexowens90: Ah I see you added exactly this as a C++ unit test, plus a Python test, ignore me

python/tests/unit/arcticdb/version_store/test_query_builder.py
```diff
@@ -477,6 +508,26 @@ def test_querybuilder_empty_date_range_then_groupby(lmdb_version_store_v1, any_o
...
+        index=pd.date_range("2000-01-01", periods=10),
+    )
+    lib.write(symbol, df)
+
+    q = QueryBuilder()
+    q = q.date_range((pd.Timestamp("2000-01-02"), pd.Timestamp("2000-01-09")))
+    q = q.groupby("col1").agg({"col2": "sum"})
+    q = q.date_range((pd.Timestamp("2000-01-04"), pd.Timestamp("2000-01-06")))
+    with pytest.raises(SchemaException):
+        lib.read(symbol, query_builder=q)
```
alexowens90, line 568: Out of curiosity, does it work if the grouping column contains timestamps?

alexowens90 (approved):

vasil-pashov (approved):

python/tests/unit/arcticdb/version_store/test_query_builder.py
poodlewars: No it doesn't at the moment (I've added a test).

PR #3208, opened 2026-06-29
Enhancement 12036593840: append inline defrag post refactors

alexowens90:
#### Reference Issues/PRs
[12036593840](https://man312219.monday.com/boards/7852509418/pulses/12036593840)

Adds a new `compact_data_inline` argument to `append` (defaulting to `False`). When set to `True`, the `compact_data` invariant that all row-slices must have the lib config `rows_per_segment` rows ±33% will hold. This is similar to calling `append` with `compact_data_inline=False`, followed by `compact_data`, but creating only one new version. Due to implementation details, the on-disk structure may be different than calling the two methods separately.

Manually benchmarked - performance when no data needs to be read from disk is the same whether `compact_data_inline` is set or not. When compaction is needed, the pathological case of appending a single row to a 100k row symbol results in a x3 slowdown with 10 columns of numeric or string data. This decreases to x2 slower when appending 50k rows.

### DIFF
--- a/cpp/arcticdb/pipeline/frame_utils.cpp
+++ b/cpp/arcticdb/pipeline/frame_utils.cpp
@@ -42,16 +42,6 @@ TimeseriesDescriptor make_timeseries_descriptor(
     };
 }
 
-TimeseriesDescriptor make_timeseries_descriptor(
-        size_t total_rows, StreamDescriptor&& desc, const proto::descriptors::NormalizationMetadata& norm_meta,
-        std::optional<proto::descriptors::UserDefinedMetadata>&& um, std::optional<AtomKey>&& next_key,
-        bool bucketize_dynamic
-) {
-    return make_timeseries_descriptor(
-            total_rows, desc, norm_meta, std::move(um), std::move(next_key), bucketize_dynamic
-    );
-}
-
 TimeseriesDescriptor timeseries_descriptor_from_pipeline_context(
         const std::shared_ptr<pipelines::PipelineContext>& pipeline_context, std::optional<AtomKey>&& next_key,
         bool bucketize_dynamic
@@ -59,11 +49,8 @@ TimeseriesDescriptor timeseries_descriptor_from_pipeline_context(
     return make_timeseries_descriptor(
             pipeline_context->total_rows_,
             pipeline_context->descriptor(),
-            *pipeline_context->norm_meta_,
-            pipeline_context->user_meta_ ? std::make_optional<arcticdb::proto::descriptors::UserDefinedMetadata>(
-                                                   std::move(*pipeline_context->user_meta_)
-                                           )
-                                         : std::nullopt,
+            pipeline_context->normalization(),
+            pipeline_context->release_opt_user_defined_metadata(),
             std::move(next_key),
             bucketize_dynamic
     );
--- a/cpp/arcticdb/pipeline/test/test_rollback.cpp
+++ b/cpp/arcticdb/pipeline/test/test_rollback.cpp
@@ -10,7 +10,8 @@
 #include <optional>
 #include <type_traits>
 #include <util/test/generators.hpp>
-#include <pipeline/write_frame.hpp>
+#include <arcticdb/pipeline/write_frame.hpp>
+#include <arcticdb/version/versioned_engine.hpp>
 
 using namespace arcticdb;
 
@@ -96,7 +97,7 @@ TestTensorFrame append_with_three_segments(version_store::PythonVersionStore& st
     constexpr size_t update_val{1};
     auto append_frame =
             get_test_frame<TimeseriesIndex>(stream_id, get_test_timeseries_fields(), 30, start_index, update_val);
-    store.append_internal(stream_id, append_frame.frame_, false, false, false);
+    store.append_internal(stream_id, append_frame.frame_, version_store::AppendOptions{});
     start_index += 30; // To avoid sameappends
     return append_frame;
 }
--- a/cpp/arcticdb/processing/clause_compact_data.cpp
+++ b/cpp/arcticdb/processing/clause_compact_data.cpp
@@ -11,14 +11,21 @@
 #include <cstdint>
 #include <set>
 
+#include <arcticdb/column_store/column_reslicer.hpp>
 #include <arcticdb/column_store/segment_reslicer.hpp>
 #include <arcticdb/pipeline/frame_slice.hpp>
+#include <arcticdb/pipeline/input_frame.hpp>
+#include <arcticdb/pipeline/write_frame.hpp>
 #include <arcticdb/processing/clause_utils.hpp>
 #include <arcticdb/util/collection_utils.hpp>
 
 namespace arcticdb {
 
-CompactDataClause::CompactDataClause(uint64_t rows_per_segment) : rows_per_segment_(rows_per_segment) {
+CompactDataClause::CompactDataClause(uint64_t rows_per_segment, std::shared_ptr<InputFrame> frame) :
+    rows_per_segment_(rows_per_segment) {
+    if (frame && frame->num_rows > 0) {
+        frame_ = std::move(frame);
+    }
     // Magic range of +-33% chosen so that:
     // - 2 row slices with < min_rows_per_segment_ cannot be combined into one row slice with > max_rows_per_segment_
     // - 1 row slice with a little more than max_rows_per_segment_ when split in half will still have
@@ -102,10 +109,22 @@ std::vector<std::vector<size_t>> CompactDataClause::structure_for_processing(std
     for (const auto& range_and_key : ranges_and_keys) {
         row_ranges.insert(range_and_key.row_range());
     }
+    if (frame_) {
+        // frame_ is non null when we have arrived here via the compact_data_inline argument to append[_batch]
+        // In this case, we treat the frame being appended as if it is ONE row-slice (even if it is larger than the
+        // library's default slicing policy). A consequence of this is that the resulting structure on-disk may not be
+        // identical to calling append with compact_data_inline=false, followed by an explicit compact_data call.
+        // However, the invariant that all row-slices will have rows_per_segments+-33% will be maintained in both cases
+        util::check(
+                frame_->offset == (row_ranges.empty() ? 0 : row_ranges.rbegin()->second),
+                "CompactDataClause expects the input frame offset to match the end of the existing data"
+        );
+        row_ranges.emplace(frame_->offset, frame_->offset + frame_->num_rows);
+    }
     // The greedy algorithm in structure_row_ranges can reslice data where all segments have an acceptable number of
     // rows, which is not desirable, so short-circuit out if this is the case
     if (row_ranges_all_acceptable_lengths(row_ranges)) {
-        log::version().info("No work to do in CompactDataClause, data is already compacted");
+        log::version().debug("No work to do in CompactDataClause, existing data is already compacted");
         ranges_and_keys.clear();
         return {};
     }
@@ -126,7 +145,7 @@ std::vector<std::vector<size_t>> CompactDataClause::structure_for_processing(std
             }
     );
     if (ranges_and_keys.empty()) {
-        log::version().info("No work to do in CompactDataClause, data is already compacted");
+        log::version().debug("No work to do in CompactDataClause, existing data is already compacted");
         return {};
     }
     // Order by column slice (i.e. all segments in first column slice from top to bottom, then second column slice, etc)
@@ -184,6 +203,12 @@ std::vector<EntityId> CompactDataClause::process(std::vector<EntityId>&& entity_
             proc.row_ranges_->back()->second
     );
     std::vector<SegmentInMemory> segments = util::extract_from_pointers(std::move(*proc.segments_));
+    // Presence of frame implies we are doing inline compaction in append. The second condition means we are processing
+    // the last row-slice of the existing data, and so it must be combined with the frame
+    if (frame_ && frame_->offset == proc.row_ranges_->back()->second) {
+        add_segment_from_frame(proc, col_range_start, segments);
+    }
+
     SegmentReslicer reslicer{max_rows_per_segment_};
     segments = reslicer.reslice_segments(std::move(segments));
 
@@ -214,9 +239,48 @@ std::vector<EntityId> CompactDataClause::process(std::vector<EntityId>&& entity_
     return push_entities(*component_manager_, std::move(proc));
 }
 
+void CompactDataClause::add_segment_from_frame(
+        const ProcessingUnit& proc, size_t col_range_start, std::vector<SegmentInMemory>& segments
+) const {
+    uint64_t total_rows_without_frame = std::accumulate(
+            segments.cbegin(),
+            segments.cend(),
+            uint64_t(0),
+            [](uint64_t n, const SegmentInMemory& segment) { return n + segment.row_count(); }
+    );
+    auto total_rows = total_rows_without_frame + frame_->num_rows;
+    ReslicingInfo reslicing_info{total_rows, max_rows_per_segment_};
+    // Work out how many rows of the frame we need to combine with segments from disk
+    uint64_t row_count{0};
+    for (size_t idx = 0; idx < reslicing_info.num_segments; ++idx) {
+        row_count += reslicing_info.rows_in_slice(idx);
+        if (row_count > total_rows_without_frame) {
+            break;
+        }
+    }
+    // Note that this RowRange does not automatically include all of the rows in the frame. This is for 3 reasons:
+    // - WriteToSegmentTask can be parallelised in the post-processing step that writes the remainder of the frame
+    // - At time of writing, WriteClause compresses all of the segments it receives in 1 thread, which can also be
+    //   parallelised if we do it later
+    // - If no compaction of existing data is needed, there is no obvious set of entity ids to call
+    //   CompactDataClause::process with to slice+write the frame
+    const SpecificSlicer slicer{
+            {RowRange{frame_->offset, frame_->offset + (row_count - total_rows_without_frame)}},
+            {ColRange{
+                    col_range_start, dynamic_schema_ ? frame_->desc().field_count() : proc.col_ranges_->front()->second
+            }}
+    };
+    // There is a check in SpecificSlicer that exactly 1 slice will be produced in this case
+    auto frame_slice = slicer(*frame_)[0];
+    WriteToSegmentTask write_to_segment_task{frame_, std::move(frame_slice)};
+    segments.emplace_back(std::get<SegmentInMemory>(write_to_segment_task()));
+}
+
 const ClauseInfo& CompactDataClause::clause_info() const { return clause_info_; }
 
-void CompactDataClause::set_processing_config(const ProcessingConfig&) {}
+void CompactDataClause::set_processing_config(const ProcessingConfig& processing_config) {
+    dynamic_schema_ = processing_config.dynamic_schema_;
+}
 
 void CompactDataClause::set_component_manager(std::shared_ptr<ComponentManager> component_manager) {
     component_manager_ = std::move(component_manager);
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -135,7 +135,7 @@ folly::Future<entity::AtomKey> async_write_dataframe_impl(
     return write_frame(std::move(partial_key), frame, slicing_policy, store, de_dup_map, sparsify_floats);
 }
 
-void sorted_data_check_append(const InputFrame& frame, const index::IndexSegmentReader& index_segment_reader) {
+void sorted_data_check_append(const InputFrame& frame, const TimeseriesDescriptor& existing_tsd) {
     if (!index_is_not_timeseries_or_is_sorted_ascending(frame)) {
         sorting::raise<ErrorCode::E_UNSORTED_DATA>(
                 "When calling append with validate_index enabled, input data must be sorted"
@@ -143,7 +143,7 @@ void sorted_data_check_append(const InputFrame& frame, const index::IndexSegment
     }
     sorting::check<ErrorCode::E_UNSORTED_DATA>(
             !std::holds_alternative<stream::TimeseriesIndex>(frame.index) ||
-                    index_segment_reader.tsd().sorted() == SortedValue::ASCENDING,
+                    existing_tsd.sorted() == SortedValue::ASCENDING,
             "When calling append with validate_index enabled, the existing data must be sorted"
     );
 }
@@ -181,21 +181,28 @@ void check_update_data_is_sorted(const InputFrame& frame, const index::IndexSegm
 }
 
 static void check_can_append(
-        const InputFrame& frame, const index::IndexSegmentReader& index_segment_reader,
-        const WriteOptions& write_options, bool validate_index, bool empty_types
+        const InputFrame& frame, const TimeseriesDescriptor& existing_tsd,
+        const std::optional<IndexValue>& last_existing_index_value, const WriteOptions& write_options,
+        bool validate_index, bool empty_types
 ) {
-    util::check_rte(!index_segment_reader.is_pickled(), "Cannot append to pickled data");
-    fix_descriptor_mismatch_or_throw(APPEND, write_options.dynamic_schema, index_segment_reader, frame, empty_types);
+    const bool is_pickled = existing_tsd.proto().normalization().input_type_case() ==
+                            arcticdb::proto::descriptors::NormalizationMetadata::InputTypeCase::kMsgPackFrame;
+    util::check_rte(!is_pickled, "Cannot append to pickled data");
+    fix_descriptor_mismatch_or_throw(APPEND, write_options.dynamic_schema, existing_tsd, frame, empty_types);
     if (validate_index) {
-        sorted_data_check_append(frame, index_segment_reader);
+        sorted_data_check_append(frame, existing_tsd);
     }
     util::variant_match(
             frame.index,
-            [&index_segment_reader, &frame, &write_options](const TimeseriesIndex&) {
+            [&existing_tsd, &last_existing_index_value, &frame, &write_options](const TimeseriesIndex&) {
                 util::check(frame.has_index(), "Cannot append timeseries without index");
-                if (index_segment_reader.tsd().total_rows() != 0 && frame.num_rows != 0) {
+                if (existing_tsd.total_rows() != 0 && frame.num_rows != 0) {
+                    util::check(
+                            last_existing_index_value.has_value(),
+                            "Cannot append timeseries: last index value of existing data is unknown"
+                    );
                     auto first_index = NumericIndex{frame.index_value_at(0)};
-                    auto prev = std::get<NumericIndex>(index_segment_reader.last()->key().end_index());
+                    auto prev = std::get<NumericIndex>(*last_existing_index_value);
                     util::check(
                             write_options.ignore_sort_order || prev - 1 <= first_index,
                             "Can't append dataframe with start index {} to existing sequence "
@@ -221,7 +228,7 @@ static void check_can_update(
     util::check(index::is_timeseries_index(index_desc), "Update not supported for non-timeseries indexes");
     check_update_data_is_sorted(frame, index_segment_reader);
     (void)check_and_mark_slices(index_segment_reader, false, std::nullopt);
-    fix_descriptor_mismatch_or_throw(UPDATE, dynamic_schema, index_segment_reader, frame, empty_types);
+    fix_descriptor_mismatch_or_throw(UPDATE, dynamic_schema, index_segment_reader.tsd(), frame, empty_types);
 }
 
 folly::Future<AtomKey> async_append_impl(
@@ -239,7 +246,17 @@ folly::Future<AtomKey> async_append_impl(
             .thenValueInline([store, update_info, frame, options, validate_index, empty_types](
                                      index::IndexSegmentReader&& index_segment_reader
                              ) {
-                check_can_append(*frame, index_segment_reader, options, validate_index, empty_types);
+                const std::optional<IndexValue> last_existing_index_value =
+                        index_segment_reader.tsd().total_rows() == 0 ? std::optional<IndexValue>()
+                                                                     : index_segment_reader.last()->key().end_index();
+                check_can_append(
+                        *frame,
+                        index_segment_reader.tsd(),
+                        last_existing_index_value,
+                        options,
+                        validate_index,
+                        empty_types
+                );
                 bool bucketize_dynamic = index_segment_reader.bucketize_dynamic();
                 auto row_offset = index_segment_reader.tsd().total_rows();
                 frame->set_offset(static_cast<ssize_t>(row_offset));
@@ -951,9 +968,8 @@ void set_output_descriptors(
                 (*clause)->clause_info().index_,
                 [](const KeepCurrentIndex&) { return false; },
                 [&](const KeepCurrentTopLevelIndex&) {
-                    if (pipeline_context->norm_meta_->mutable_df()->mutable_common()->has_multi_index()) {
-                        const auto& multi_index =
-                                pipeline_context->norm_meta_->mutable_df()->mutable_common()->multi_index();
+                    if (pipeline_context->normalization().df().common().has_multi_index()) {
+                        const auto& multi_index = pipeline_context->normalization().df().common().multi_index();
                         auto name = multi_index.name();
                         auto tz = multi_index.tz();
                         bool fake_name{false};
@@ -964,7 +980,8 @@ void set_output_descriptors(
                             }
                         }
                         auto mutable_index =
-                                pipeline_context->norm_meta_->mutable_df()->mutable_common()->mutable_index();
+                                pipeline_context->mutable_normalization().mutable_df()->mutable_common()->mutable_index(
+                                );
                         mutable_index->set_tz(tz);
                         mutable_index->set_is_physically_stored(true);
                         mutable_index->set_name(name);
@@ -974,7 +991,8 @@ void set_output_descriptors(
                 },
                 [&](const NewIndex& new_index) {
                     index_column = new_index;
-                    auto mutable_index = pipeline_context->norm_meta_->mutable_df()->mutable_common()->mutable_index();
+                    auto mutable_index =
+                            pipeline_context->mutable_normalization().mutable_df()->mutable_common()->mutable_index();
                     mutable_index->set_name(new_index);
                     mutable_index->clear_fake_name();
                     mutable_index->set_is_physically_stored(true);
@@ -1157,9 +1175,9 @@ static StreamDescriptor generate_initial_output_schema_descriptor(const Pipeline
 
 static OutputSchema create_initial_output_schema(PipelineContext& pipeline_context) {
     internal::check<ErrorCode::E_ASSERTION_FAILURE>(
-            pipeline_context.norm_meta_, "Normalization metadata should not be missing during read_and_process"
+            pipeline_context.has_normalization(), "Normalization metadata should not be missing during read_and_process"
     );
-    return OutputSchema{generate_initial_output_schema_descriptor(pipeline_context), *pipeline_context.norm_meta_};
+    return OutputSchema{generate_initial_output_schema_descriptor(pipeline_context), pipeline_context.normalization()};
 }
 
 static OutputSchema generate_output_schema(PipelineContext& pipeline_context, const ReadQuery& read_query) {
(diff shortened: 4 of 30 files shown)


python/tests/hypothesis/arcticdb/test_append.py
alexowens90, line 1: Changes in this file are just to parametrize relevant tests over `compact_data_inline=[True, False]`

python/tests/unit/arcticdb/version_store/test_append.py
alexowens90, line 1: Changes in this file are just to parametrize relevant tests over `compact_data_inline=[True, False]`

cpp/arcticdb/processing/clause_compact_data.cpp
```diff
@@ -102,10 +107,22 @@ std::vector<std::vector<size_t>> CompactDataClause::structure_for_processing(std
...
+        // In this case, we treat the frame being appended as if it is ONE row-slice (even if it is larger than the
+        // library's default slicing policy). A consequence of this is that the resulting structure on-disk may not be
+        // identical to calling append with compact_data_inline=false, followed by an explicit compact_data call.
+        // However, the invariant that all row-slices will have rows_per_segments+-33% will be maintained in both cases
+        util::check(
+                frame_->offset == row_ranges.rbegin()->second,
+                "CompactDataClause expects the input frame offset to match the end of the existing data"
+        );
+        row_ranges.emplace(frame_->offset, frame_->offset + frame_->num_rows);
+    }
```
alexowens90, line 123: This implementation defragments the whole symbol. An alternative would be to only defragment the tail of the symbol, in which case we would just need to modify the logic here. I think the current implementation is better though, as I would expect the usage pattern to be to always call with `compact_data_inline=True`. Therefore, the first call may be slower as it defragments the whole symbol, but all subsequent calls will only touch the tail of the data anyway.

alexowens90:
Will add ASV benchmarks in a separate PR once this is merged

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+        read_query->clauses_.push_back(
+                std::make_shared<Clause>(CompactDataClause(rows_per_segment, compact_data_frame->frame_))
+        );
+    } else {
+        read_query->clauses_.push_back(std::make_shared<Clause>(CompactDataClause(rows_per_segment)));
+    }
+
+    VersionIdentifier resolved = std::make_shared<IndexInformation>(
+            read_index_key_without_column_stats(store, *update_info.previous_index_key_)
+    );
```
alexowens90: This should be async

python/arcticdb/version_store/library.py
```diff
@@ -1399,6 +1400,12 @@ def append(
         index_column: bool, default=False
             Only applicable when data is a PyArrow Table or Polars DataFrame. If True, the first column
             is treated as the timeseries index.
+        compact_data_inline: bool, default=False
+            If False, the data being appended will be sliced and written to disk without consideration for how
+            fragmented this may make the data.
+            If True, the data will also be compacted at the same time (see the `compact_data` method for more details).
+            Note that this will usually involve reading some data segments from disk and doing some in-memory
+            processing, and so will generally be slower than when this argument is False.
```
alexowens90: Add a comment that upsert will not get "nicely sliced", it will follow the usual slicing policy

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -0,0 +1,548 @@
...
+    # There might be fewer rows in total than min_rows_per_segment
+    min_rows_per_segment = min(min_rows_per_segment, len(expected))
+    assert row_counts.min() >= min_rows_per_segment
+    assert row_counts.max() <= max_rows_per_segment
+
+    post_compaction_data_keys = len(post_compaction_index)
+    new_data_keys = len(post_compaction_index[post_compaction_index["version_id"] > vit_before_compaction.version])
+    compacted_data_keys = pre_compaction_data_keys - (post_compaction_data_keys - new_data_keys)
+    assert query_stats_operation_count(stats, "Memory_GetObject", "TABLE_DATA") == compacted_data_keys
+    assert query_stats_operation_count(stats, "Memory_PutObject", "TABLE_DATA") == new_data_keys
```
poodlewars, line 69: Should we have some checks on the total number of new data keys written, at least for some constructed examples? It seems valuable to have a test that protects us from doing excessive write IO.

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -0,0 +1,548 @@
...
+    sym = "test_read_previous_version"
+    df = pd.DataFrame({"col": np.arange(10)})
+    lib.write(sym, df[:5])
+    generic_compact_data_inline_test(lib, sym, df[5:])
+    assert_frame_equal(df[:5], lib.read(sym, as_of=0).data)
+    assert_frame_equal(df, lib.read(sym, as_of=1).data)
+    assert_frame_equal(df, lib.read(sym).data)
+
+
+def test_schema_mismatch_static(in_memory_store_factory):
```
poodlewars, line 471: Surprising this isn't already covered by the `test_append.py` that you parameterized

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -0,0 +1,548 @@
...
+    remaining_rows = num_rows
+    first_iteration = True
+    while remaining_rows > 0:
+        rows_to_take = rng.integers(1, 2 * rows_per_segment)
+        if first_iteration:
+            lib.write(sym, table.slice(length=rows_to_take))
+            first_iteration = False
+        else:
+            generic_compact_data_inline_test(lib, sym, table.slice(length=rows_to_take))
+        # lib.append(sym, table.slice(length=rows_to_take), compact_data_inline=True)
```
poodlewars: commented out

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
poodlewars, line 1: Worth having some coverage for appending `pd.Series` or numpy arrays?

python/arcticdb/version_store/library.py
```diff
@@ -1399,6 +1400,12 @@ def append(
         index_column: bool, default=False
             Only applicable when data is a PyArrow Table or Polars DataFrame. If True, the first column
             is treated as the timeseries index.
+        compact_data_inline: bool, default=False
+            If False, the data being appended will be sliced and written to disk without consideration for how
+            fragmented this may make the data.
+            If True, the data will also be compacted at the same time (see the `compact_data` method for more details).
```
poodlewars, line 1406: If the historical data is badly fragmented, could this end up doing a full table scan? Could the memory requirements and time to do this make it difficult for people to adopt the `compact_data_inline` option?

I would have expect this feature to work by just considering the tail of the index, folding the new data in to existing data to try to make sure the new block is ~100k rows.

cpp/arcticdb/pipeline/test/test_rollback.cpp
```diff
@@ -96,7 +96,7 @@ TestTensorFrame append_with_three_segments(version_store::PythonVersionStore& st
     constexpr size_t update_val{1};
     auto append_frame =
             get_test_frame<TimeseriesIndex>(stream_id, get_test_timeseries_fields(), 30, start_index, update_val);
-    store.append_internal(stream_id, append_frame.frame_, false, false, false);
+    store.append_internal(stream_id, append_frame.frame_, false, false, false, false);
```
poodlewars: `false, false, false, false` is a tad hard to read

cpp/arcticdb/processing/clause_compact_data.cpp
poodlewars: Ah, this duplicates a point I raised in another comment. Please ignore that one.

cpp/arcticdb/processing/clause_compact_data.cpp
```diff
@@ -214,9 +237,42 @@ std::vector<EntityId> CompactDataClause::process(std::vector<EntityId>&& entity_
...
+            break;
+        }
+    }
+    const SpecificSlicer slicer{
+            {RowRange{frame_->offset, frame_->offset + (row_count - total_rows_without_frame)}},
+            {ColRange{
+                    col_range_start, dynamic_schema_ ? frame_->desc().field_count() : proc.col_ranges_->front()->second
+            }}
+    };
+    // There is a check in SpecificSlicer that exactly 1 slice will be produced in this case
```
poodlewars, line 273: Still reading, but I don't get why we construct a special SpecificSlicer to get this property rather than skipping this slicing altogether.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+    std::set<RowRange> row_ranges;
+    for (const auto& slice_and_key : slices_and_keys) {
+        row_ranges.insert(slice_and_key.slice().row_range);
+    }
+    return std::vector<RowRange>(
+            std::make_move_iterator(row_ranges.begin()), std::make_move_iterator(row_ranges.end())
+    );
+}
+
+static void calc_on_disk_slices(
```
poodlewars: Quite confused by this. Why does the binary search only consider the first row in each row range, doesn't it need to consider both? Could this change to return a `std::vector<SliceAndKey>` of the new slices_and_keys we've emplaced, rather than void? I think a comment with a small worked example would help.

Ah didn't realise this was just moved. Still a bit of explanation would help.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+                                    const std::vector<RowRange> new_row_ranges =
+                                            find_unique_row_ranges(slices_and_keys);
+                                    size_t last_row_on_disk = new_row_ranges.empty() ? 0 : new_row_ranges.back().second;
+                                    calc_on_disk_slices(
+                                            new_row_ranges,
+                                            pipeline_context->slice_and_keys_,
+                                            last_row_on_disk,
+                                            slices_and_keys
+                                    );
+                                    // This implies there are more rows in frame that have not yet been written to disk
```
poodlewars, line 3399: When could the rows in frame have been written to disk already?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+                                                              last_row_on_disk,
+                                                              folly::poly_cast<CompactDataClause>(
+                                                                      *read_query->clauses_.front()
+                                                              )
+                                                                      .max_rows_per_segment_,
+                                                              write_options,
+                                                              target_partial_index_key
+                                                      )
+                                                    : folly::makeFuture(std::vector<SliceAndKey>{});
+                                    return (std::move(remaining_slice_and_keys_fut))
```
poodlewars, line 3414: Like we spoke about in the meeting it's a shame we fan in first with `read_modify_write_data_keys` before then doing the remaining keys strictly after, in principle we could do all this IO together in parallel. Doesn't have to change now.

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -21,147 +21,69 @@
...
 
-    lmdb_version_store.snapshot("my_snap")
+    for _ in range(100):
+        lib = lmdb_version_store_tiny_segment_dynamic
+        lib.write(sym, df0).version
+        lib.append(sym, df1).version
+        lib.append(sym, df2).version
+        lib.append(sym, df3).version
+        assert lib.get_info(sym)["sorted"] == "ASCENDING"
+        lib.version_store.defragment_symbol_data(sym, None)
```
IvoDD, line 58: We can remove this old xfailed test which uses `defragment_symbol_data` or maybe better make the last append `compact_data_inplace=True`

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -232,406 +154,6 @@ def test_sort_index(version_store_factory, dynamic_schema, prune_previous_versio
...
-
-        err = get_stderr()
-        assert err.count("W arcticdb.storage | Failed to find segment for key") == 0
-        assert err.count("D arcticdb.storage | Failed to find segment for key") == 1
-    finally:
-        set_log_level()
-
-
 @pytest.mark.parametrize("prune_previous_versions", [True, False])
 def test_defragment_read_prev_versions(sym, lmdb_version_store, prune_previous_versions):
```
IvoDD, line 159: There are other `defragment_symbol_data` tests in this file.
We should add a ticket to remove the `defragment_symbol_data` API alongside these tests in next major release.

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -21,147 +21,69 @@
...
+        err = get_stderr()
+        assert err.count("W arcticdb.storage | Failed to find segment for key") == 0
+        assert err.count("D arcticdb.storage | Failed to find segment for key") == 1
+    finally:
+        set_log_level()
 
 
+# sort_index assumes segments are internally sorted, which is not the case when appending with compact_data_inline=True
 @pytest.mark.parametrize("prune_previous_versions", [True, False])
 def test_append_out_of_order_and_sort(lmdb_version_store_ignore_order, prune_previous_versions):
```
IvoDD, line 89: I see `test_append_out_of_order_and_sort` and `test_sort_index` where not parametrized with `compact_data_inline`. Do you think it's worth parametrizing them as well just to make sure `sort_index` works after `append(compact_data_inline=True)`.

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -0,0 +1,548 @@
...
+pytestmark = pytest.mark.pipeline
+
+
+def generic_compact_data_inline_test(lib, sym, df, **append_kwargs):
+    qs.reset_stats()  # Clear any leftover stats from a previous failed run
+    vit_before_compaction = lib.read(sym, output_format="PANDAS" if isinstance(df, pd.DataFrame) else "POLARS")
+    oracle_sym = sym + "_oracle"
+    lib.write(oracle_sym, vit_before_compaction.data)
+    lib.append(oracle_sym, df, compact_data_inline=False, **append_kwargs)
+    # Use Polars so that sparse data checking is proper
```
IvoDD, line 40: This means we will lose pandas normalization metadata. E.g. for `test_multindex` it would be better to compare the output pandas dataframes vs the polars dataframes.

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
IvoDD: +1, it would also be nice to confirm that for the frequent use case of "repeated small appends" we do "read at most 1 data key" and "write at most 2 data keys"

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -0,0 +1,548 @@
...
+    sym = "test_append_empty_frame_no_work_to_do"
+    lib.write(sym, pd.DataFrame({"col": np.arange(10)}), metadata="0")
+    # Schema checks happen after empty input frame checks, so we don't need the same column set
+    # This metadata behaviour isn't ideal, but we should be consistent whether compact_data_inline is True or False
+    # See Monday issue 18103677039
+    lib.append(sym, pd.DataFrame(), metadata="1")
+    vit = lib.read(sym)
+    assert vit.version == 0
+    assert vit.metadata == "0"
+    lib.append(sym, pd.DataFrame(), metadata="1", compact_data_inline=True)
```
IvoDD: query stats would be useful to confirm that in this case we don't read the data key if it's fully compacted

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -0,0 +1,548 @@
...
+
+def test_append_empty_frame_compacts_existing_data(in_memory_store_factory):
+    lib = in_memory_store_factory(segment_row_size=10)
+    sym = "test_append_empty_frame_compacts_existing_data"
+    lib.write(sym, pd.DataFrame({"col": np.arange(5)}))
+    lib.append(sym, pd.DataFrame({"col": np.arange(5, 10)}))
+    # Schema checks happen after empty input frame checks, so we don't need the same column set
+    lib.append(sym, pd.DataFrame())
+    assert lib.read(sym).version == 1
+    lib.append(sym, pd.DataFrame(), compact_data_inline=True)
```
IvoDD: Query stats here would also be useful to confirm we read 2 data keys and write 1.

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
IvoDD: Maybe we can also fit some check in the generic case which does something like:
```
first_n_rows_already_compacted = 0
for data_key in pre_compacted_index:
    if not is_well_compacted(data_key):
        break;
    first_n_rows_already_compacted += 1

for data_key in post_compacted_index[:first_n_rows_already_compacted]:
    # I.e. if the symbol is starts with n compacted data keys we should not touch the first n-1.
    assert data_key.version_id == version_id_before 
```

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
IvoDD: I think the append would be done by the `generic_compact_data_inline_test`. We should instead comment that "Equivalent to: ...."

python/arcticdb/version_store/_store.py
```diff
@@ -985,6 +986,12 @@ def append(
         index_column: bool, default=False
             Only applicable when data is a PyArrow Table or Polars DataFrame. If True, the first column
             is treated as the timeseries index.
+        compact_data_inline: bool, default=False
+            If False, the data being appended will be sliced and written to disk without consideration for how
+            fragmented this may make the data.
+            If True, the data will also be compacted at the same time (see the `compact_data` method for more details).
+            Note that this will usually involve reading some data segments from disk and doing some in-memory
+            processing, and so will generally be slower than when this argument is False.
```
IvoDD: Maybe worth adding that it will make subsequent reads faster

cpp/arcticdb/pipeline/frame_utils.cpp
```diff
@@ -68,11 +68,8 @@ TimeseriesDescriptor timeseries_descriptor_from_pipeline_context(
     return make_timeseries_descriptor(
             pipeline_context->total_rows_,
             pipeline_context->descriptor(),
-            std::move(*pipeline_context->norm_meta_),
-            pipeline_context->user_meta_ ? std::make_optional<arcticdb::proto::descriptors::UserDefinedMetadata>(
-                                                   std::move(*pipeline_context->user_meta_)
-                                           )
-                                         : std::nullopt,
+            pipeline_context->release_normalization(),
```
IvoDD: Note: Will have some merge conflicts with #3202 depending on which gets merged first. We'll probably need to use `pipeline_context->normalization()` as it starts taking a `const&`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
     });
 }
 
+static folly::Future<std::vector<SliceAndKey>> slice_and_write_frame_remainder(
+        const std::shared_ptr<Store>& store, std::shared_ptr<InputFrame> frame, size_t start_row,
+        uint64_t max_rows_per_segment, const WriteOptions& write_options, IndexPartialKey target_partial_index_key
+) {
+    auto remaining_rows_to_write = (frame->offset + frame->num_rows) - start_row;
+    ReslicingInfo reslicing_info{remaining_rows_to_write, max_rows_per_segment};
+    std::vector<RowRange> row_ranges;
```
IvoDD: Nit: `reserve(reslicing_info.num_segments)`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+            &async::io_executor(),
+            [store, frame, slicer = std::move(slicer), key = std::move(target_partial_index_key)]() mutable {
+                // These rows are being appended, so dedup isn't possible
+                auto de_dup_map = std::make_shared<DeDupMap>();
+                return slice_and_write(frame, slicer, std::move(key), store, de_dup_map, false);
+            }
+    );
+}
+
+static std::vector<RowRange> find_unique_row_ranges(const std::vector<SliceAndKey>& slices_and_keys) {
```
IvoDD: Nit: Might rename to signal result is sorted e.g. `find_sorted_unique_row_ranges`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+                std::make_shared<Clause>(CompactDataClause(rows_per_segment, compact_data_frame->frame_))
+        );
+    } else {
+        read_query->clauses_.push_back(std::make_shared<Clause>(CompactDataClause(rows_per_segment)));
+    }
+
+    VersionIdentifier resolved = std::make_shared<IndexInformation>(
+            read_index_key_without_column_stats(store, *update_info.previous_index_key_)
+    );
+    return async::submit_io_task(SetupPipelineContextTask{store, std::move(resolved), read_query, {}})
```
IvoDD: Isn't `SetupPipelineContextTask` going to do only non-blocking IndexSegmentReader construction from the segment is read above in `read_index_key_without_column_stats`? If so running the `setup_pipeline_context` on the IOThread might just cause us to wait on some random io thread's queue and loose some cache locality.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+                    const auto& existing_tsd = pipeline_context->tsd();
+                    check_can_append(
+                            *frame,
+                            existing_tsd,
+                            pipeline_context->last_existing_index_value_,
+                            write_options,
+                            compact_data_frame->validate_index_,
+                            compact_data_frame->empty_types_
+                    );
+                    frame->set_offset(existing_tsd.total_rows());
```
IvoDD: On the regular append path we also `set_bucketize_dynamic` [here](https://github.com/man-group/ArcticDB/blob/0a3300a3ba3a03eb457745955b3f2114dfa9fe5d/cpp/arcticdb/version/version_core.cpp#L263).

I'm wondering how hard would be split the `compact_data_inplace=False` vs `compact_data_inplace=True` codepaths slightly lower in the call stack so we have to do the `frame->set_offset` and `check_can_append` in a single place before the split. E.g. we could split inside `async_append_impl`?

cpp/arcticdb/version/version_core.cpp
IvoDD: Yeah I can vaguely see how we can shift the responsibility of writing the data keys to the Compact + Write clauses. I.e. what is stopping Compact clause to give slices of the InputFrame alongiside the combined SegmentInMemorys to the Write clause. Then the Write clause could be capable of writing both SegmentInMemory and InputFrame slices.

cpp/arcticdb/processing/clause_compact_data.cpp
```diff
@@ -214,9 +237,42 @@ std::vector<EntityId> CompactDataClause::process(std::vector<EntityId>&& entity_
...
+    auto frame_slice = std::move(slicer(*frame_)[0]);
+    WriteToSegmentTask write_to_segment_task{frame_, std::move(frame_slice)};
+    segments.emplace_back(std::get<SegmentInMemory>(write_to_segment_task()));
```
IvoDD, line 276: Can't we here just inlcude the entire to be appended `frame_` as new segments here?

E.g. if we are splitting to 10 rows and on disk we have segments of sizes `[1, 1]`
And we try to append a frame of size `[20]`

Now we are producing segments like `[1, 1, 9]` and leave the remaining `[11]` from `frame_` to be sliced and written strictly later in version_core.cpp.

Instead if we just add the entire frame we'd have segments like `[1, 1, 20]` which will naturally produce segments to write `[11, 11]`.

I guess there are a couple of problems with this approach:
- Work to restructure in `WriteToSegmentTask` would not be parallelized (e.g. converting from an arrow string to internal string pool representation)
- What to do if the last data key from the previous index key is perfectly sliced. Then we don't have an obvious processing unit to which to attach the work to slice the append segment.

IvoDD (commented):
Also I think it would be good to add:

- arrow tests with `append(arrow_table, compact_data_inline=True)`
- ASV benchmarks to do repeated appends (e.g. 100 appends of a 10 rows each) with or without `compact_data_inline`. Also a benchmark on the follow up read

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -0,0 +1,548 @@
...
+    lib.write(sym, df[:5])
+    lib.append(sym, df[5:10])
+    lib.append(sym, df[10:20])
+    assert len(lib.read_index(sym)) == 3
+    generic_compact_data_inline_test(lib, sym, df[20:])
+
+
+@pytest.mark.parametrize("index", [None, "ts"])
+@pytest.mark.parametrize("segment_row_size", [100_000, 10, 5, 2])
+def test_basic_dynamic_schema(in_memory_store_factory, clear_query_stats, index, segment_row_size):
```
vasil-pashov: Is there a test with dynamic schema, were type promotion happens?

cpp/arcticdb/processing/clause_compact_data.cpp
```diff
@@ -102,10 +107,22 @@ std::vector<std::vector<size_t>> CompactDataClause::structure_for_processing(std
...
         row_ranges.insert(range_and_key.row_range());
     }
+    if (frame_) {
+        // frame_ is non null when we have arrived here via the compact_data_inline argument to append[_batch]
+        // In this case, we treat the frame being appended as if it is ONE row-slice (even if it is larger than the
+        // library's default slicing policy). A consequence of this is that the resulting structure on-disk may not be
+        // identical to calling append with compact_data_inline=false, followed by an explicit compact_data call.
+        // However, the invariant that all row-slices will have rows_per_segments+-33% will be maintained in both cases
+        util::check(
+                frame_->offset == (row_ranges.empty() ? 0 : row_ranges.rbegin()->second),
```
vasil-pashov, line 119: Apparently this works but I don't get why we're checking against `row_ranges.begin()` and not `row_ranges.end()`

cpp/arcticdb/processing/clause_compact_data.cpp
```diff
@@ -214,9 +237,42 @@ std::vector<EntityId> CompactDataClause::process(std::vector<EntityId>&& entity_
...
+        }
+    }
+    const SpecificSlicer slicer{
+            {RowRange{frame_->offset, frame_->offset + (row_count - total_rows_without_frame)}},
+            {ColRange{
+                    col_range_start, dynamic_schema_ ? frame_->desc().field_count() : proc.col_ranges_->front()->second
+            }}
+    };
+    // There is a check in SpecificSlicer that exactly 1 slice will be produced in this case
+    auto frame_slice = std::move(slicer(*frame_)[0]);
```
vasil-pashov: nit: no need to call move. The result is an rvalue already

cpp/arcticdb/processing/clause_compact_data.cpp
vasil-pashov: I agree. In case we stay with the current approach a comment would help a lot. I took me quite some time to figure out what's going on.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3255,97 +3274,228 @@ folly::Future<CompactDataInfo> compact_data_explain_plan_impl(
...
+        col_ranges.emplace_back(frame->desc().index().field_count(), frame->desc().field_count());
+    } else {
+        ColRange col_range{
+                frame->desc().index().field_count(),
+                frame->desc().index().field_count() + write_options.column_group_size
+        };
+        col_ranges.emplace_back(col_range);
+        while (col_range.second < frame->desc().field_count()) {
+            col_range.first += write_options.column_group_size;
+            col_range.second += write_options.column_group_size;
```
vasil-pashov, line 3224: shouldn't this be the `min(col_range.second + write_options.column_group_size, col_range.second + remaining_columns)`

cpp/arcticdb/version/version_core.cpp
vasil-pashov: If I followed the code correctly - in the case when the last segment had `n` rows until it reached the tipping point for splitting and the frame had less than `n` rows

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
alexowens90: Will add some more explicit tests of data keys read/written.
@IvoDD that invariant won't necessarily hold:
```
# Writes one well compacted data key
lib.write(sym, pd.DataFrame({"col": np.arange(100_000)}))
# Appending 1 row will combine the new frame with the existing well compacted data key to produce 1 data key total in version 1
lib.append(sym, pd.DataFrame({"col": np.arange(1)}))
```

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
alexowens90: Only the hypothesis test, will add something more explicit

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
alexowens90: Indeed, I thought I might have duplicated but just double checked and there's nothing like this or `test_schema_mismatch_dynamic` in there. I thought about moving it there, but it doesn't parametrize nicely because we're explicitly testing that the exceptions raised with `compact_data_inline=True/False` are the same, so will leave them here

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
alexowens90: 👍

python/tests/unit/arcticdb/version_store/test_append.py
alexowens90: I'd rather just leave all things `defragment_symbol_data` related as is for now, and remove it all in `7.0.0` as you suggested [here](https://github.com/man-group/ArcticDB/pull/3208/changes#r3511777923)

python/tests/unit/arcticdb/version_store/test_append.py
alexowens90: See comment added above the test

python/tests/unit/arcticdb/version_store/test_append.py
alexowens90: https://man312219.monday.com/boards/7852509418/pulses/12466578772

python/arcticdb/version_store/library.py
alexowens90: Yeh, if the historical data is all fragmented it could load everything. It is pipelined so that writes back to disk are scheduled interleaved with the reads, but as your column stats work showed, that can be insufficient if the processing takes too long. The semaphore work you've been doing should fix this though.

Also worth noting that a lot of append only workflows do a historical backfill using `write`, followed by the small appends, so often the head of the data isn't fragmented

cpp/arcticdb/pipeline/test/test_rollback.cpp
alexowens90: Fair, will combine into an options struct

cpp/arcticdb/processing/clause_compact_data.cpp
alexowens90: It's using `row_ranges.rbegin()`

cpp/arcticdb/processing/clause_compact_data.cpp
alexowens90: I don't get how we could skip the slicing?
Row-wise, we don't necessarily need all of the rows from the frame, we might only need the first `n`.
Column-wise, with static schema, we are only dealing with one column-slice here, so we need to slice by columns, which also "slices" the relevant part of the descriptor out from the frame.

cpp/arcticdb/processing/clause_compact_data.cpp
alexowens90: Yeh, the reasoning was both of the bullets Ivo outlined. I'll add a comment to this effect

cpp/arcticdb/version/version_core.cpp
alexowens90: It could be, the slicing implementation doesn't care if the col range is wider than the actual number of columns in the frame though. I'll add a comment that this is the case

cpp/arcticdb/version/version_core.cpp
alexowens90: Ah, yeh, I'd got confused about when the index key IO was happening (see my comment immediately above). Will sort it out

cpp/arcticdb/version/version_core.cpp
alexowens90: Yeh, say you had 1 row on disk, and you're appending 199,999 rows in the new frame. Then the first 99,999 rows of the new frame get combined with the 1 row on disk in `CompactDataClause::process`, and we have 100,000 rows of the frame we still haven't written at this point

cpp/arcticdb/version/version_core.cpp
alexowens90: I agree it would be nice to get to this point. The issue with that approach Ivo is that `WriteToSegment` task is then not parallelised

cpp/arcticdb/version/version_core.cpp
alexowens90: Its complicated by 2 things:

1. The empty-frame metadata-writing path
2. `async_append_impl` is keyed off an `IndexSegmentReader`, whereas `compact_data_inline` is keyed off a `PipelineContext`

There is a similar problem with duplicated code copy-pasted between `append_internal` and `batch_append_internal`. I think I'll look at tackling all of this as part of the batch implementation of `compact_data_inline` rather than dragging this PR out even further.

cpp/arcticdb/version/version_core.cpp
alexowens90: Looking at it, `read_index_key_without_column_stats` does a `read_sync`, which is almost certainly wrong in most use cases. I'll circle back round and fix that in a follow up PR as well.

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
IvoDD: > `# I.e. if the symbol is starts with n compacted data keys we should not touch the first n-1.`

But it should still hold for the first n-1 right?

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
```diff
@@ -81,6 +86,103 @@ def test_basic(in_memory_store_factory, clear_query_stats, index):
...
+            lib.append(sym, df[idx * 1_000 : (idx + 1) * 1_000], compact_data_inline=True)
+            stats = qs.get_query_stats()
+        qs.reset_stats()
+        received = lib.read(sym).data
+        assert_frame_equal(df[: (idx + 1) * 1_000], received)
+        assert query_stats_operation_count(stats, "Memory_GetObject", "TABLE_INDEX") == (0 if idx == 0 else 1)
+        assert query_stats_operation_count(stats, "Memory_GetObject", "TABLE_DATA") == (0 if idx == 0 else 1)
+        assert query_stats_operation_count(stats, "Memory_PutObject", "TABLE_DATA") == 1
+        assert query_stats_operation_count(stats, "Memory_PutObject", "TABLE_INDEX") == 1
+        assert len(lib.read_index(sym)) == 1
```
IvoDD: I think this test would be more interesting if we end up with > 1 row slice.

I think the number of data key reads should always be <=1 if we are appending very small tables.
And I think it will be a useful assertion for the case where I have a 10 million row dataframe to which I append to every day and want the result to remain compacted.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3203,7 +3203,8 @@ static folly::Future<std::vector<SliceAndKey>> slice_and_write_frame_remainder(
 ) {
     auto remaining_rows_to_write = (frame->offset + frame->num_rows) - start_row;
     ReslicingInfo reslicing_info{remaining_rows_to_write, max_rows_per_segment};
-    std::vector<RowRange> row_ranges;
+    auto row_ranges = util::reserve_vector<RowRange>(reslicing_info.num_segments);
+    row_ranges.reserve(reslicing_info.num_segments);
```
IvoDD: Nit: Double reserve

IvoDD (approved):

python/tests/unit/arcticdb/version_store/test_append_compact_data_inline.py
alexowens90: I thought so too, so I added the check. The hypothesis test showed that there are corner cases where this isn't true though. The one I dug into had 10 rows per slice (so min/max 6/13 rows per slice). Simplified flow:
```
Write 6 rows
Append 7 rows - doesn't trigger a compaction due to the row_ranges_all_acceptable_lengths check
Append 10 rows - doesn't trigger a compaction due to the row_ranges_all_acceptable_lengths check
Append 1 row - triggers a compaction. The greedy algo then combines the first 2 row slices, as well as the last 2
```
So while I think the invariant will hold in 99.99% of cases in practice, it isn't strictly a property of the system

PR #3214, opened 2026-06-30
Bugfixes 18103677039 & 9682041273: persist metadata when appending or updating empty frames, and skip writing unnecessary symbol list keys

alexowens90:
#### Reference Issues/PRs
[18103677039](https://man312219.monday.com/boards/7852509418/pulses/18103677039)
[9682041273](https://man312219.monday.com/boards/7852509418/pulses/9682041273)

First commit:
Previously, when an empty dataframe was provided to `append` or `update` (or batch variations thereof), a warning would be logged and no new version was created. In particular, this meant that the provided metadata was not persisted.

Second commit:
Reinstated a test removed as part of the first commit, and then fix the bug where symbol list keys were added on batch append/update even if a live version of the symbol already existed

### DIFF
--- a/python/tests/integration/arcticdb/test_arctic_batch.py
+++ b/python/tests/integration/arcticdb/test_arctic_batch.py
@@ -732,7 +732,7 @@ def test_append_batch_missing_keys(arctic_library):
     assert_frame_equal(read_dataframe.data, pd.concat([df2_write, df2_append]))
 
 
-def test_append_batch_empty_dataframe_does_not_increase_version(lmdb_version_store_v1):
+def test_append_batch_empty_dataframe_increases_version(lmdb_version_store_v1):
     lib = lmdb_version_store_v1
     lib.batch_write(["sym1", "sym2"], [pd.DataFrame({"a": [1, 2, 3]}), pd.DataFrame({"b": [1, 2, 3, 4]})])
     lib_tool = lib.library_tool()
@@ -746,7 +746,7 @@ def test_append_batch_empty_dataframe_does_not_increase_version(lmdb_version_sto
 
     append_result = lib.batch_append(["sym1", "sym2"], [pd.DataFrame({"a": [5, 6, 7]}), pd.DataFrame({"b": []})])
     assert append_result[0].version == 1
-    assert append_result[1].version == 0
+    assert append_result[1].version == 1
 
     sym_1_vit, sym_2_vit = lib.read("sym1"), lib.read("sym2")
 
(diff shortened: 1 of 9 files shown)


python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -756,16 +756,15 @@ def test_append_batch_empty_dataframe_does_not_increase_version(lmdb_version_sto
...
+    assert len(lib_tool.find_keys_for_symbol(KeyType.VERSION, "sym2")) == 2
+    assert len(lib_tool.find_keys_for_symbol(KeyType.TABLE_INDEX, "sym2")) == 2
     assert len(lib_tool.find_keys_for_symbol(KeyType.TABLE_DATA, "sym2")) == 1
 
     # This result is wrong. The correct value is 2. This is due to a bug Monday: 9682041273, append_batch and
-    # update_batch should not create symbol list keys for already existing symbols. Since append_batch is noop when
-    # the input is empty, there is no key for sym_2, but there is a new key for sym_1 and that is wrong.
-    assert len(lib_tool.find_keys(KeyType.SYMBOL_LIST)) == 3
+    # update_batch should not create symbol list keys for already existing symbols.
+    assert len(lib_tool.find_keys(KeyType.SYMBOL_LIST)) == 4
```
poodlewars: It's a shame we're creating even more needless symbol list keys, would it be worth looking at 9682041273 in your bug fixing rampage before this one?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -798,3 +798,25 @@ def test_append_after_delete_range(sym, lmdb_version_store):
...
+
+
+@pytest.mark.parametrize("batch", [True, False])
+def test_append_empty_frame_metadata(lmdb_version_store_v1, batch):
+    lib = lmdb_version_store_v1
+    sym = "test_append_empty_frame_metadata"
+    write_df = pd.DataFrame({"col": np.arange(1)})
+    metadata_v0 = "v0"
+    lib.write(sym, write_df, metadata=metadata_v0)
+    empty_df = pd.DataFrame({"col": np.arange(0)})  # Using arange guarantees the dtype matches the written df
```
poodlewars: Any value in doing this test for empty series as well?

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -1038,3 +991,28 @@ def test_update_new_data_contains_old(version_store_factory):
...
+
+@pytest.mark.parametrize("batch", [True, False])
+def test_update_empty_frame_metadata(lmdb_library, batch):
+    # Need to use a V2 API fixture as there is no batch_update on the V1 API
+    lib = lmdb_library
+    sym = "test_update_empty_frame_metadata"
+    write_df = pd.DataFrame({"col": np.arange(1)}, index=[pd.Timestamp("2026-01-01")])
+    metadata_v0 = "v0"
+    lib.write(sym, write_df, metadata=metadata_v0)
+    empty_df = pd.DataFrame(
```
poodlewars: Similar to append, any value in a variant of this test that uses an empty series?

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -872,53 +872,6 @@ def test_non_normalizable_data_throws(self, lmdb_library):
...
-        # Update happens in the middle of the dataframe. Data prior the update range (value 0) is in one segment, then
-        # there's one segment for the new data (values 10, 20) then there's one segment for the data that's pas the
-        # update range (values 3, 4). The fourth segment is the original data segment
-        assert len(lib_tool.find_keys_for_symbol(KeyType.TABLE_DATA, "symbol_2")) == 4
-        assert len(lib_tool.read_index("symbol_2")) == 3
-
-        # This result is wrong. The correct value is 2. This is due to a bug Monday: 9682041273, append_batch and
-        # update_batch should not create symbol list keys for already existing symbols. Since update_batch is noop when
-        # the input is empty, there is no key for symbol_1, but there is a new key for symbol_2 and that is wrong.
-        assert len(lib_tool.find_keys(KeyType.SYMBOL_LIST)) == 3
```
poodlewars, line 920: We seem to have lost one of the tests for 9682041273 here

poodlewars (approved):

python/tests/integration/arcticdb/test_arctic_batch.py
alexowens90: Ah yeh, that one didn't show up on my radar because it's marked as `performance` rather than `bug`. Will rampage through that next

vasil-pashov (commented):
1. I think this is technically a break in the API. We should at least make it visible in the notes that this now creates a new version
2. I think it's worth having a test for append/update with empty data frame and no metadata. If I followed the code it should create a new version and "clear" the current metadata.

vasil-pashov (approved):

PR #3215, opened 2026-07-01
Harden parsing in transform_asv_results.py and update library access in Resample benchmark

G-D-Petrov:

1. A benchmark gets skipped upstream, in the "Benchmark given commit" step:
``` bash
[66.67%] ··· Setting up resample:129   failed
  ...
  arcticdb.exceptions.LibraryNotFound: resample_wide
  asv: setup_cache failed (exit status 1)
[66.67%] ··· resample.ResampleWide.time_resample_wide skipped (setup_cache failed)ResampleWide.setup() 
```
python/benchmarks/resample.py:148 unconditionally does `self.ac[self.LIB_NAME]`, but ASV appears to invoke setup() once before setup_cache() has actually created the resample_wide library - so it always fails, and both time_resample_wide and peakmem_resample_wide end up marked "skipped" for every run.

2. That skip is what actually breaks the tracked step. In that case it stores result=None and trims trailing None fields, producing a short list instead of the usual 13-element [result, params, version, started_at, duration, ...] row.

So this PR implements 2 fixes:
1. Handles the library access better
2. Hardens the duration parsing to prevent such errors in the future

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/build_tooling/transform_asv_results.py
+++ b/build_tooling/transform_asv_results.py
@@ -186,6 +186,19 @@ def extract_most_recent_result(lib, json_path) -> int:
         return 1
 
 
+def extract_time(r):
+    """r looks like the "results" mentioned in the docstring of analyze_asv_results. Using eval as the results can
+    contain nan, inf etc which json.loads cannot parse.
+
+    ASV drops trailing fields (including "duration" at index 4) for benchmarks that did not produce a result, e.g.
+    because setup_cache failed, so the list can be shorter than expected.
+    """
+    as_list = eval(r)
+    if as_list is None or len(as_list) <= 4:
+        return nan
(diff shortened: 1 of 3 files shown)


IvoDD (approved):
happy to merge since it fixes the CI


>  but ASV appears to invoke setup() once before setup_cache() 

This seems quite surprising to me. I think we have many benchmarks which rely on this not being true (not only the `resample` one).
This does not look like expected behavior see [docs](https://asv.readthedocs.io/en/stable/writing_benchmarks.html#setup-and-teardown-functions). Maybe we should open a github issue if we can reproduce this.

PR #3222, opened 2026-07-03
Read Index Key Only Once When Creating Column Stats

markovskipetar:
## Summary
  
  `create_column_stats_experimental` was reading the symbol's `TABLE_INDEX` (index) key
  **twice**: once in the Python layer to work out which columns are eligible for MINMAX
  stats, and again in the C++ stats-generation path. The C++ layer already reads the index
  key and has the `TimeseriesDescriptor` in hand, so this moves eligible-column detection
  into C++ and reads the index key **once**.

  ## Before
  
  - **Python** (`_store.py`): `create_column_stats_experimental` called
    `_get_eligible_column_stats_spec` → `_get_info(...)`, which **read the index key** to
    fetch the `timeseries_descriptor`, computed eligible columns, rejected duplicated
    data-column names, and passed an explicit `ColumnStats` down to C++.
  - **C++** (`create_column_stats_impl`): **read the index key again** to get the same
    `tsd`, then mapped the supplied column names to offsets.

  Query stats for a 2-row-slice symbol:

  ```json
  "storage_operations": {
      "S3_GetObject": {
          "COLUMN_STATS": { "count": 1, "size_bytes": 0,    "total_time_ms": 18 },
          "TABLE_DATA":   { "count": 2, "size_bytes": 703,  "total_time_ms": 30 },
          "TABLE_INDEX":  { "count": 2, "size_bytes": 2550, "total_time_ms": 31 } 
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      },
      "S3_PutObject": {
          "COLUMN_STATS": { "count": 1, "size_bytes": 677,  "total_time_ms": 14 }
      }
  }
  ```
   ##  After

  - Python: create_column_stats_experimental no longer reads the index — it just calls
  create_column_stats_version(symbol, version_query).
  - C++: a new ColumnStats(const TimeseriesDescriptor&) constructor derives the eligible
  columns directly from the tsd that create_column_stats_impl already reads. The
  column_stats argument is dropped through the whole create chain.

  Same scenario, index key now read once:

 ```json
  "storage_operations": {
      "S3_GetObject": {
          "COLUMN_STATS": { "count": 1, "size_bytes": 0,    "total_time_ms": 18 },
          "TABLE_DATA":   { "count": 2, "size_bytes": 703,  "total_time_ms": 32 },
          "TABLE_INDEX":  { "count": 1, "size_bytes": 1275, "total_time_ms": 15 }
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      },
      "S3_PutObject": {
          "COLUMN_STATS": { "count": 1, "size_bytes": 518,  "total_time_ms": 14 }
      }
  }
  ```
  TABLE_INDEX GetObject: 2 → 1 (and size_bytes ~halved).

### DIFF
--- a/cpp/arcticdb/pipeline/column_stats.cpp
+++ b/cpp/arcticdb/pipeline/column_stats.cpp
@@ -251,31 +251,6 @@ ColumnStats::ColumnStats(const TimeseriesDescriptor& tsd) {
     offset_to_stat_info_set_ = true;
 }
 
-void ColumnStats::calculate_offsets(const TimeseriesDescriptor& tsd, utils::MissingColumnsBehavior missing_columns) {
-    util::check(!offset_to_stat_info_set_, "Should not calculate offsets twice");
-    std::unordered_set<std::string> unmangled_names;
-    for (const auto& k : user_specified_column_stats_ | ranges::views::keys) {
-        unmangled_names.insert(k);
-    }
-    auto offsets_and_mangled_names =
-            utils::find_offset_and_mangled_name(unmangled_names, tsd.as_stream_descriptor(), missing_columns);
-    util::check(
-            missing_columns == utils::MissingColumnsBehavior::IGNORE_MISSING ||
-                    offsets_and_mangled_names.size() == user_specified_column_stats_.size(),
-            "Expect calculated offsets_and_mangled_names to match column_stats_"
-    );
-    for (auto& offset_and_mangled_name : offsets_and_mangled_names) {
-        offset_to_stat_info_.emplace(
-                offset_and_mangled_name.offset,
-                NameAndStatTypes{
-                        offset_and_mangled_name.mangled_name,
-                        user_specified_column_stats_.at(offset_and_mangled_name.unmangled_name)
-                }
-        );
-    }
-    offset_to_stat_info_set_ = true;
-}
-
 namespace {
 std::unordered_set<ColumnStatTypeInternal> external_to_internal(ColumnStatType type) {
     switch (type) {
@@ -328,7 +303,6 @@ std::vector<std::string> ColumnStats::drop(const ColumnStats& to_drop, bool warn
             ++it;
         }
     }
-    user_specified_column_stats_ = {};
     return dropped_names;
 }
 
@@ -376,14 +350,10 @@ std::optional<Clause> ColumnStats::clause() const {
     return ColumnStatsGenerationClause(std::move(input_columns), index_generation_aggregators);
 }
 
-bool ColumnStats::empty() const { return user_specified_column_stats_.empty() && offset_to_stat_info_.empty(); }
+bool ColumnStats::empty() const { return offset_to_stat_info_.empty(); }
 
 bool ColumnStats::operator==(const ColumnStats& right) const {
-    if (offset_to_stat_info_set_) {
-        return offset_to_stat_info_ == right.offset_to_stat_info_;
-    } else {
-        return user_specified_column_stats_ == right.user_specified_column_stats_;
-    }
+    return offset_to_stat_info_ == right.offset_to_stat_info_;
 }
 
 } // namespace arcticdb
--- a/cpp/arcticdb/pipeline/column_stats.cpp
+++ b/cpp/arcticdb/pipeline/column_stats.cpp
@@ -2,6 +2,7 @@
 #include <arcticdb/pipeline/index_fields.hpp>
 #include <arcticdb/processing/aggregation_interface.hpp>
 #include <arcticdb/processing/unsorted_aggregation.hpp>
+#include <arcticdb/entity/timeseries_descriptor.hpp>
 #include <arcticdb/entity/type_utils.hpp>
 #include <arcticdb/util/preconditions.hpp>
 
@@ -133,21 +134,6 @@ void validate_column_stats_header_version(const arcticc::pb2::column_stats_pb2::
     }
 }
 
-ColumnStats::ColumnStats(const std::unordered_map<std::string, std::unordered_set<std::string>>& column_stats) {
-    for (const auto& [column, column_stat_names] : column_stats) {
-        if (!column_stat_names.empty()) {
-            user_specified_column_stats_[column] = {};
-            for (const auto& column_stat_name : column_stat_names) {
-                auto opt_index_type = name_to_type(column_stat_name);
-                user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
-                        opt_index_type.has_value(), "Unknown column stat type provided: {}", column_stat_name
-                );
-                user_specified_column_stats_[column].emplace(*opt_index_type);
-            }
-        }
-    }
-}
-
 ColumnStats::ColumnStats(
         const arcticc::pb2::column_stats_pb2::ColumnStatsHeader& header, const TimeseriesDescriptor& tsd
 ) {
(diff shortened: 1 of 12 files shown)


python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -766,6 +766,42 @@ def test_query_stats_in_mem_delete(in_memory_version_store, clear_query_stats):
...
+    #     "storage_operations": {
+    #         "S3_PutObject": {
+    #             "COLUMN_STATS": {
+    #                 "count": 1,
+    #                 "size_bytes": 320,
+    #                 "total_time_ms": 15
+    #             }
+    #         }
+    #     }
+    # }
```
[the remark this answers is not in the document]
IvoDD: +1 to claude comment

python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -766,6 +766,42 @@ def test_query_stats_in_mem_delete(in_memory_version_store, clear_query_stats):
         assert lists[key_type]["size_bytes"] == 0
 
 
+def test_query_stats_column_stats(s3_version_store_v1, clear_query_stats, sym):
```
IvoDD: Better to use the `in_memory_version_store` as it is faster to run.

python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -766,6 +766,42 @@ def test_query_stats_in_mem_delete(in_memory_version_store, clear_query_stats):
...
+    assert "storage_operations" in stats
+    storage_operations = stats["storage_operations"]
+
+    assert "S3_PutObject" in storage_operations, storage_operations
+    assert "COLUMN_STATS" in storage_operations["S3_PutObject"], storage_operations
+
+    column_stats = storage_operations["S3_PutObject"]["COLUMN_STATS"]
+    assert column_stats["count"] == 1, column_stats
+    assert column_stats["size_bytes"] > 0, column_stats
+    assert column_stats["total_time_ms"] < 8000, column_stats
```
IvoDD: The `total_time_ms` assertion is very hard to control (especially with s3). This means tests like this can be flaky (if e.g. the GitHub CI runners slows down for any reason)

In general for time measurements it is better to use ASV benchmarks. They run on specially provisioned software from AWS and allow for 15% variance in timings by default.

So my suggestion is to remove any `time_ms` query stats assertions from unit tests.

python/tests/integration/toolbox/test_query_stats.py
```diff
@@ -766,6 +766,42 @@ def test_query_stats_in_mem_delete(in_memory_version_store, clear_query_stats):
...
+    #         }
+    #     }
+    # }
+    assert "storage_operations" in stats
+    storage_operations = stats["storage_operations"]
+
+    assert "S3_PutObject" in storage_operations, storage_operations
+    assert "COLUMN_STATS" in storage_operations["S3_PutObject"], storage_operations
+
+    column_stats = storage_operations["S3_PutObject"]["COLUMN_STATS"]
```
IvoDD: Here we only measure that we are writing a single COLUMN_STATS key. What do you think about folding this assertion in in your other test in `test_column_stats_optimization.py` and removing this test here?

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -24,9 +24,49 @@ def get_column_stats_read_count():
...
+    index = pd.MultiIndex.from_arrays([pd.date_range("2000-01-01", periods=2), [10, 20]], names=["dt", "level"])
+    df = pd.DataFrame({"col_1": [1, 2], "col_2": ["a", "b"]}, index=index)
+    lib.write(sym, df)
+
+
+@pytest.mark.parametrize(
+    "writer",
+    [_write_single_segment, _write_multi_segment, _write_multi_index],
+    ids=["single_segment", "multi_segment", "multi_index"],
+)
```
alexowens90: Parametrizing this test is overkill, the `_write_multi_segment` is a sufficient test by itself

python/tests/integration/toolbox/test_query_stats.py
alexowens90: This test can be removed entirely

python/arcticdb/version_store/_store.py
```diff
@@ -1306,62 +1305,10 @@ def create_column_stats_experimental(
         -------
         None
         """
-        column_stats = self._get_eligible_column_stats_spec(symbol, as_of)
-        if not column_stats:
-            return
-
-        column_stats = self._convert_to_native_column_stats(column_stats)
+        # Eligible-column detection is done in the C++ layer so that the symbol's index key is
+        # read only once (during stats generation) rather than an extra time here.
```
alexowens90: This comment can go as well, this change is just bringing this method in line with everything else where we only read index keys once

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -569,9 +569,7 @@ void register_bindings(py::module& version, py::exception<arcticdb::ArcticExcept
             .value("ASCENDING", SortedValue::ASCENDING)
             .value("DESCENDING", SortedValue::DESCENDING);
 
-    py::class_<ColumnStats>(version, "ColumnStats")
-            .def(py::init<std::unordered_map<std::string, std::unordered_set<std::string>>>())
-            .def("to_map", &ColumnStats::to_map);
+    py::class_<ColumnStats>(version, "ColumnStats").def("to_map", &ColumnStats::to_map);
```
IvoDD, line 572: Is this class used in python anymore? If not we can remove it from the python bindings

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -180,6 +166,65 @@ ColumnStats::ColumnStats(
...
+} // namespace
+
+// Build MINMAX stats for every eligible column, computed directly from the TSD.
+// The outer/primary index is skipped.
+// Rejects symbols with duplicated data-column names.
+ColumnStats::ColumnStats(const TimeseriesDescriptor& tsd) {
+    const auto& fields = tsd.fields();
+    const auto& norm = tsd.normalization();
+
+    const bool has_outer_index = tsd.index().field_count() > 0;
```
alexowens90: I don't think we refer to this as an "outer" index anywhere else
```suggestion
    const bool has_timeseries_index = tsd.index().field_count() > 0;
```

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -180,6 +166,65 @@ ColumnStats::ColumnStats(
...
+    const size_t start_filed_index = has_outer_index ? 1 : 0;
```
alexowens90:
```suggestion
    const size_t start_field_index = has_outer_index ? 1 : 0;
```

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -180,6 +166,65 @@ ColumnStats::ColumnStats(
...
+
+    std::unordered_set<std::string> seen_user_names;
+
+    for (size_t field_index = start_filed_index; field_index < fields.size(); ++field_index) {
```
alexowens90: Because of how the fields collection is stored, `.at` is not very efficient. Better to use `fields.begin()` and `fields.end()`

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -180,6 +166,65 @@ ColumnStats::ColumnStats(
...
+        if (!is_col_eligible_for_stats(fields.at(field_index).type().data_type())) {
+            continue;
+        }
+
+        std::string field_name{fields.at(field_index).name()};
+
+        if (norm.has_df()) {
+            user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
```
alexowens90: I know that you haven't changed the behaviour here, but long-term we should still create stats over the non-duplicated column names. Can you open a ticket in the column stats epic so we don't forget?

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -180,6 +166,65 @@ ColumnStats::ColumnStats(
     offset_to_stat_info_set_ = true;
 }
 
+namespace {
+bool is_col_eligible_for_stats(DataType col_data_type) { return is_numeric_type(col_data_type) || is_bool_type(col_data_type); }
+
+std::string to_user_facing_name_key(
+        const std::string& field_name, const arcticdb::proto::descriptors::NormalizationMetadata::Pandas& common
```
alexowens90:
```suggestion
        std::string_view field_name, const arcticdb::proto::descriptors::NormalizationMetadata::Pandas& common
```

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -180,6 +166,65 @@ ColumnStats::ColumnStats(
...
+// The outer/primary index is skipped.
+// Rejects symbols with duplicated data-column names.
+ColumnStats::ColumnStats(const TimeseriesDescriptor& tsd) {
+    const auto& fields = tsd.fields();
+    const auto& norm = tsd.normalization();
+
+    const bool has_outer_index = tsd.index().field_count() > 0;
+    const size_t start_filed_index = has_outer_index ? 1 : 0;
+
+    std::unordered_set<std::string> seen_user_names;
```
alexowens90, line 227: I think this set can use `std::string_view`?

cpp/arcticdb/pipeline/column_stats.cpp
alexowens90: I see that you need the `field_index` below, so you can use `begin/end` together with `folly::enumerate`

alexowens90 (changes requested):

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -133,21 +134,6 @@ void validate_column_stats_header_version(const arcticc::pb2::column_stats_pb2::
     }
 }
 
-ColumnStats::ColumnStats(const std::unordered_map<std::string, std::unordered_set<std::string>>& column_stats) {
```
IvoDD, line 136: As far as I can see we have removed the only constructor that allows setting the `user_specified_column_stats_`.

If it is always going to be empty should we just remove it entirely from the state? Or do we plan to keep using it in the future?

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -180,6 +166,65 @@ ColumnStats::ColumnStats(
...
+            continue;
+        }
+
+        std::string field_name{fields.at(field_index).name()};
+
+        if (norm.has_df()) {
+            user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+                    seen_user_names.insert(to_user_facing_name_key(field_name, norm.df().common())).second,
+                    "Cannot create column stats: symbol has duplicated data column name [{}]",
+                    field_name
```
IvoDD: Shouldn't the error message use the user facing name instead of `field_name`?

E.g. if we have two columns named `col` this error message would throw something like `col_2` which will be confusing for the user.

cpp/arcticdb/pipeline/column_stats.cpp
IvoDD: I'm not sure because I think this `unordered_set` is the only place we store the user facing names after the calls to `to_user_facing_name`.

cpp/arcticdb/version/python_bindings.cpp
markovskipetar: It is used in Python `get_column_stats_info_experimental` which calls `get_column_stats_info_version` that returns `ColumnStats` object, on which the `.to_map()` function is called

cpp/arcticdb/pipeline/column_stats.cpp
alexowens90: Ah I thought they were just field names from the TSD, ignore me

IvoDD (approved):

cpp/arcticdb/pipeline/column_stats.cpp
markovskipetar: done!

alexowens90 (approved):

PR #3224, opened 2026-07-06
[[DO NOT MERGE]] Arrow norm meta RFC

IvoDD:

An RFC for discussing design of Arrow normalization metadata.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
+++ b/cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
@@ -1,0 +1,188 @@
+# Arrow normalization metadata
+
+## Arrow specific metadata
+
+### Per column
+
+Each arrow column will have several properties:
+- Timezone (e.g. None, London, New_York)
+- String format (e.g. LARGE_STRING, CATEGORICAL)
+
+This dictates a straightforward:
+```
+message ArrowArray {
+    string tz = 1;            // It makes sense to reuse the Pandas format
+    string string_format = 2; // A `oneof` or `uint8` might be more efficient but I feel like string is more future proof without the need to further modify protobufs. 
+}
+```
+
+### Arrow Table vs Array primitives
+
+A single arrow array can be represented as a single `ArrowArray` message. 
+It is also fundamentally unindexed (a timeseries index would require a second column). 
+
+A table is a collection of columns, so we'll use a dictionary from column name to `ArrowArray` metadata.
+An alternative would be to index by column index which would be more efficient but would make the problem of merging dynamic schema normalization metadatas harder.
+
+```
+message ArrowTable {
+    bool has_index = 1;
+    map<string, ArrowArray> columns = 2;
+}
+```
+
+## Arrow -> Pandas
+
+Here we discuss the basic flow:
+```
+lib.write(arrow)
+lib.read(output_format="pandas")
+```
+
+- Arrow arrays -> Pandas Series with RangeIndex
+- Arrow table without index -> Pandas DataFrame with RangeIndex
+- Arrow table with index -> Pandas DataFrame with TimeseriesIndex
+- string format metadata is lost
+- timezone metadata is non-trivial
+  - Option A: Apply timezone metadata only to index column. Gives a straight translation from arrow to pandas metadata
+  - Option B: Apply all timezone columns. Will require pandas denoramalization code to work directly with Arrow metadata
+  - I am prefering option B as it is more correct and less surprising for users.
+
+## Pandas -> Arrow
+
+Here we discuss the basic flow:
+```
+lib.write(pandas)
+lib.read(output_format="polars")
+```
+
+- Pandas Series with RangeIndex -> Arrow array
+- Pandas Series with other index -> Arrow Table
+- Pandas DataFrame -> Arrow Table
+
+## Combining pandas with arrow via append/update or concat
+
+Here we discuss flows like:
+```
+lib.write(arrow)
+lib.append(pandas)
+lib.read()
+```
+or
+```
+lib.write(arrow)
+lib.write(pandas)
+lib.read_batch_concat([arrow, pandas])
+```
+
+### How should we handle index mismatch?
+- Pandas RangeIndex is compatible only with `has_index=False`
+- Pandas TimeseriesIndex is compabitible only with `has_index=True`
+- Pandas MultiIndex is compatible only with `has_index=True`
+
+### How should we handle unnamed indices and duplicate columns?
+We have specified a conversion from unnamed pandas indices to arrow on read. However making append work with that convention will be challenging:
+```
+lib.write(pandas_df_with_multiindex) # Multindex level names [None, None], column name "col"
+lib.append(arrow_table)              # Column names ["__index_level_0__", "__index_level_1__", "col"]
+```
+The above example will be challenging to make work because the two `StreamDescriptor` would be different.
+
+This becomes even more complicated if we consider duplicated columns and dynamic schema.
+```
+lib.write(pandas_df)     # Index [None], columns ["index"] 
+lib.append(arrow_table)  # Columns: ["_index_", "index", "__index__"], "__index__" is a new column.
+lib.append(pandas_df)    # Index [None], columns ["index", "__index___", "_index_"], which column corresponds to which?
+```
+In the above it is very unclear which column should be considered the new one in the last append.
+
+I'm in favor with the first version of this to:
+- Gracefully handle unnamed indices without duplicates. I.e. allow combining unnamed pandas index or multiindex with an arrow table which has the correct column names
+- Raise when trying to combine a pandas dataframe with duplicate column names with arrow. If people care about this we can fix in the future.
+
+
+### How should we handle timezone mismatch?
+- [Existing pandas behavior](https://github.com/man-group/ArcticDB/blob/b9a8b861c950377c15104bd135c7635a827dc673/python/tests/unit/arcticdb/version_store/test_symbol_concatenation.py#L659-L660) is to allow any mismatch and clear the timezone on the result
+- I'm personally in favor of raising an exception always whenever timezones are different (specified or not), but that will need to be future 7.0.0 change. So, for now we'll allow combining any timezone with any other timezone.
+
+### How should we store a mix of pandas specific metadata and arrow specific metadata?
+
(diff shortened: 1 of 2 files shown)


cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -224,6 +224,22 @@ message NormalizationMetadata {
         repeated uint64 shape = 1;
     }
 
+    message ArrowArray {
+        string tz = 1;
+        string string_format = 2;
```
alexowens90, line 229: I made the tz optional in my timezones PR, but we agree on the principle.
We could more elegantly express that timezones and string formats (and probably other things we add in future) are mutually exclsuive by using a `oneof`? And make it extensible by having a temporal message (that would just contain an optional tz right now), a string message, etc

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+- Arrow arrays -> Pandas Series with RangeIndex
+- Arrow table without index -> Pandas DataFrame with RangeIndex
+- Arrow table with index -> Pandas DataFrame with TimeseriesIndex
+- string format metadata is lost
+- timezone metadata is non-trivial
+  - Option A: Apply timezone metadata only to index column. Gives a straight translation from arrow to pandas metadata
+  - Option B: Apply all timezone columns. Will require pandas denoramalization code to work directly with Arrow metadata
+  - I am prefering option B as it is more correct and less surprising for users.
```
alexowens90, line 49: Agreed

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+```
+lib.write(arrow)
+lib.write(pandas)
+lib.read_batch_concat([arrow, pandas])
+```
+
+### How should we handle index mismatch?
+- Pandas RangeIndex is compatible only with `has_index=False`
+- Pandas TimeseriesIndex is compabitible only with `has_index=True`
+- Pandas MultiIndex is compatible only with `has_index=True`
```
alexowens90, line 81: I think this is only the case if the top-level index is a timeseries?

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+This becomes even more complicated if we consider duplicated columns and dynamic schema.
+```
+lib.write(pandas_df)     # Index [None], columns ["index"] 
+lib.append(arrow_table)  # Columns: ["_index_", "index", "__index__"], "__index__" is a new column.
+lib.append(pandas_df)    # Index [None], columns ["index", "__index___", "_index_"], which column corresponds to which?
+```
+In the above it is very unclear which column should be considered the new one in the last append.
+
+I'm in favor with the first version of this to:
+- Gracefully handle unnamed indices without duplicates. I.e. allow combining unnamed pandas index or multiindex with an arrow table which has the correct column names
```
alexowens90, line 100: So in this case, the index would acquire the name of the first column from the Arrow table on future reads? I think this makes sense

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+- Raise when trying to combine a pandas dataframe with duplicate column names with arrow. If people care about this we can fix in the future.
```
alexowens90, line 101: Yeh, let's start here and see if it is a problem in practice

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+### How should we handle timezone mismatch?
+- [Existing pandas behavior](https://github.com/man-group/ArcticDB/blob/b9a8b861c950377c15104bd135c7635a827dc673/python/tests/unit/arcticdb/version_store/test_symbol_concatenation.py#L659-L660) is to allow any mismatch and clear the timezone on the result
+- I'm personally in favor of raising an exception always whenever timezones are different (specified or not), but that will need to be future 7.0.0 change. So, for now we'll allow combining any timezone with any other timezone.
```
alexowens90, line 106: In my PR I made it a static/dynamic schema split. Static schema requires an exact match, dynamic makes the column timezone-naive on a mismatch. I think it's worth doing 7.0.0 now if it means that we never have any behaviour other than this with Arrow

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+- Pandas Series with RangeIndex -> Arrow array
+- Pandas Series with other index -> Arrow Table
+- Pandas DataFrame -> Arrow Table
+
+## Combining pandas with arrow via append/update or concat
+
+Here we discuss flows like:
+```
+lib.write(arrow)
+lib.append(pandas)
```
alexowens90, line 68: This way around (i.e. writing Arrow first and then combining with Pandas) should actually not be too bad, as Arrow guarantees we have well formed schema, string/non-duplicated column names, etc. Appends and updates to this with Pandas data then also need well formed schema to combine without us throwing.
The other way round (and more common use case given all the legacy Pandas data) is much harder. I think we should start off making the minimal subset work, and then consider adding support for more esoteric Pandas features on a case-by-case basis. For me, the minimal set of Pandas features that do not map cleanly to Arrow that we should support are:

1. Unnamed indexes (inherits from the Arrow table)
2. Multiindex, including unnamed levels (again, column names come from the Arrow table)
3. `has_synthetic_columns` - similarly to how if you don't specify and index, Pandas will give you a `RangeIndex`, if you don't specify column names, Pandas will also make them a `RangeIndex`, amounting to naming them `0, 1, ..., n`. If you create an Arrow table with columns called `"0", "1", ..., "n"`, we should allow this to be combined with Pandas data where `has_synthetic_columns == True`

The main features this would exclude is

1. Duplicated/non-string column names:  it is hard to think of what an API would even look like to handle this
2. Categoricals: we don't support append/update with these in Pandas anyway, so this isn't really an additional restriction

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+
+message ArrowTable {
+    bool has_index = 1;
+    map<string, ArrowArray> columns = 2;
+    optional PandasDataFrame pandas_meta = 3;
+}
+```
+
+Downsides:
+- Allows for inconsistent state (e.g. arrow timezone vs pandas index timezone)
```
alexowens90, line 127: For single-symbol (i.e. non-`concat`) use cases we can guard against this at append/update time.
With `concat`, we are generally as permissive as possible about what can be combined

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
```diff
@@ -0,0 +1,188 @@
...
+- Also allows for inconsistent state
+- Preserves the messy PandasDataFrame metadata even for arrow only flows
+
+Upsides:
+- Can use the same normalization metadata format for both Pandas and Arrow
+- Arrow tables written with new arcticdb will be readable by older clients
+
+#### Ivo's take
+
+I'd go with option A.
```
alexowens90, line 182: I agree that this is the best of the proposed options. If we refactor all of our various normalization metadata checking/combining into `accumulate_norm_metadata`, then merging two (or more) norm metadatas should be easy enough to reason about.

There is a fourth option - we add a new protobuf message that goes into the `input_type oneof` that is something like `ArrowPandasCombo`, that is only produced when you append/update Pandas with Arrow or vice-versa. This would only contain valid combinations of fields from the Pandas/Arrow messages used to produce it. We could then have functions:

- ArrowPandasCombo to_combo(const PandasDataFrame&)
- ArrowPandasCombo to_combo(const ExperimentalArrow&)
- PandasDataFrame to_pandas(const ArrowPandasCombo&)
- ExperimentalArrow to_arrow(const ArrowPandasCombo&)

This has the advantage of keeping Pandas crap out of the Arrow messages, and pure Arrow/pure Pandas use cases remain easy to reason about (e.g. if you read as Arrow, you might still need to check the `optional PandasDataFrame` field with option A). Then reading pure Arrow as Pandas becomes `to_pandas(to_combo(experimental_arrow))`, and vice-versa.

cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -224,6 +224,22 @@ message NormalizationMetadata {
...
+    message ArrowSeries {
+        ArrowArray arrow_meta = 1;
+        optional PandasDataFrame pandas_meta = 2;
+    }
+
+    message ArrowTable {
+        bool has_index = 1;
+        map<string, ArrowArray> columns = 2;
+        optional PandasDataFrame pandas_meta = 3;
+    }
```
alexowens90, line 241: When I was writing up the tickets I suggested we support pretty much all the primitives:

- Polars series and dataframe
- pyarrow table/record batch/array/chunked array
I think we should always just return a list of record batches to the Python layer though, and then zero-copy transform in `_normalization.py`. Might need some additional metadata (e.g. Polars Series can have a name like Pandas)

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
alexowens90: Static schema case we agree is correct.
Dynamic schema is more subtle, and will probably make someone unhappy regardless of what we pick

cpp/proto/arcticc/pb2/descriptors.proto
alexowens90: Record batch and array will never be part of the norm metadata, as at read time they would still come back as tables and chunked arrays due to row slicing.

cpp/proto/arcticc/pb2/arrow_norm_meta_rfc.md
maxim-morozov: It is definitely a complex decision, the challenge that I would like us to think about is that users generally not [fully] aware of those differences, which makes them hard to transition to arrow and understand the schema expectations. If we can come up with some magic covers 90% of cases or implicit conversion, it would help users to move quicker, but also remove the knowledge barrier.

PR #3225, opened 2026-07-06
Enhancement 9929831600: arrow timezone support

alexowens90:
#### Reference Issues/PRs
[9929831600](https://man312219.monday.com/boards/7852509418/pulses/9929831600)

Adds support for storing timestamp columns' timezones with data written as Arrow. Behaviour with append/update is as follows:

- Static schema - timezone of column in the new table must match that of the existing data exactly (or both be timezone-naive)
- Dynamic schema - timezones can differ arbitrarily. If they are different, the result will be timezone-naive

### DIFF
--- a/cpp/arcticdb/arrow/arrow_utils.cpp
+++ b/cpp/arcticdb/arrow/arrow_utils.cpp
@@ -12,11 +12,15 @@
 #include <arcticdb/column_store/column.hpp>
 #include <arcticdb/column_store/memory_segment.hpp>
 #include <arcticdb/util/allocator.hpp>
+#include <sparrow/layout/array_access.hpp>
 #include <sparrow/layout/primitive_data_access.hpp>
+#include <sparrow/utils/temporal.hpp>
 #include <sparrow/record_batch.hpp>
 #include <utility>
 
 namespace arcticdb {
+
+using ArrowMeta = proto::descriptors::NormalizationMetadata::ExperimentalArrow;
 
 std::optional<sparrow::validity_bitmap> create_validity_bitmap(
         size_t offset, const Column& column, size_t bitmap_size
@@ -37,14 +41,14 @@ std::optional<sparrow::validity_bitmap> create_validity_bitmap(
 }
 
 template<typename T>
-sparrow::primitive_array<T> create_primitive_array(
+sparrow::array create_primitive_array(
         T* data_ptr, size_t data_size, std::optional<sparrow::validity_bitmap>&& validity_bitmap
 ) {
     sparrow::u8_buffer<T> buffer(data_ptr, data_size, get_detachable_allocator());
     if (validity_bitmap) {
-        return sparrow::primitive_array<T>{std::move(buffer), data_size, std::move(*validity_bitmap)};
+        return sparrow::array{sparrow::primitive_array<T>{std::move(buffer), data_size, std::move(*validity_bitmap)}};
     } else {
-        return sparrow::primitive_array<T>{std::move(buffer), data_size};
+        return sparrow::array{sparrow::primitive_array<T>{std::move(buffer), data_size}};
     }
 }
 
@@ -80,43 +84,57 @@ sparrow::array create_packed_bool_array(
     return arr;
 }
 
-template<typename T>
-sparrow::timestamp_without_timezone_nanoseconds_array create_timestamp_array(
-        T* data_ptr, size_t data_size, std::optional<sparrow::validity_bitmap>&& validity_bitmap
+sparrow::array create_timestamp_array(
+        timestamp* data_ptr, size_t data_size, std::optional<sparrow::validity_bitmap>&& validity_bitmap,
+        const date::time_zone* tz = nullptr
 ) {
-    static_assert(sizeof(T) == sizeof(sparrow::zoned_time_without_timezone_nanoseconds));
-    // We default to using timestamps without timezones. If the normalization metadata contains a timezone it will be
-    // applied during normalization in python layer.
-    sparrow::u8_buffer<sparrow::zoned_time_without_timezone_nanoseconds> buffer(
-            reinterpret_cast<sparrow::zoned_time_without_timezone_nanoseconds*>(data_ptr),
-            data_size,
-            get_detachable_allocator()
-    );
-    if (validity_bitmap) {
-        return sparrow::timestamp_without_timezone_nanoseconds_array{
-                std::move(buffer), data_size, std::move(*validity_bitmap)
-        };
+    static_assert(sizeof(timestamp) == sizeof(sparrow::zoned_time_without_timezone_nanoseconds));
+    if (tz == nullptr) {
+        // Timezone naive
+        sparrow::u8_buffer<sparrow::zoned_time_without_timezone_nanoseconds> buffer(
+                reinterpret_cast<sparrow::zoned_time_without_timezone_nanoseconds*>(data_ptr),
+                data_size,
+                get_detachable_allocator()
+        );
+        if (validity_bitmap) {
+            return sparrow::array{sparrow::timestamp_without_timezone_nanoseconds_array{
+                    std::move(buffer), data_size, std::move(*validity_bitmap)
+            }};
+        } else {
+            return sparrow::array{sparrow::timestamp_without_timezone_nanoseconds_array{std::move(buffer), data_size}};
+        }
     } else {
-        return sparrow::timestamp_without_timezone_nanoseconds_array{std::move(buffer), data_size};
+        // Timezone aware. Arrow stores the timestamp in UTC regardless of timezone, so no transformations are required
+        // from our internal format
+        sparrow::u8_buffer<timestamp> buffer(
+                reinterpret_cast<timestamp*>(data_ptr), data_size, get_detachable_allocator()
+        );
+        if (validity_bitmap) {
+            return sparrow::array{
+                    sparrow::timestamp_nanoseconds_array{tz, std::move(buffer), std::move(*validity_bitmap)}
+            };
+        } else {
+            return sparrow::array{sparrow::timestamp_nanoseconds_array{tz, std::move(buffer)}};
+        }
     }
 }
 
 template<typename T>
-sparrow::dictionary_encoded_array<T> create_dict_array(
+sparrow::array create_dict_array(
         sparrow::array&& dict_values_array, sparrow::u8_buffer<T>&& dict_keys_buffer,
         std::optional<sparrow::validity_bitmap>&& validity_bitmap
 ) {
     if (validity_bitmap) {
-        return sparrow::dictionary_encoded_array<T>{
+        return sparrow::array{sparrow::dictionary_encoded_array<T>{
                 typename sparrow::dictionary_encoded_array<T>::keys_buffer_type(std::move(dict_keys_buffer)),
                 std::move(dict_values_array),
                 std::move(*validity_bitmap)
-        };
+        }};
     } else {
-        return sparrow::dictionary_encoded_array<T>{
+        return sparrow::array{sparrow::dictionary_encoded_array<T>{
                 typename sparrow::dictionary_encoded_array<T>::keys_buffer_type(std::move(dict_keys_buffer)),
                 std::move(dict_values_array),
-        };
+        }};
     }
 }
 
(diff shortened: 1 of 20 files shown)


python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -401,27 +401,183 @@ def test_write_owned_and_non_owned_buffers(in_memory_version_store_tiny_segment_
...
+@pytest.mark.parametrize("method", ["append", "update"])
+@pytest.mark.parametrize(
+    "col_with_tz_timezone,col_without_tz_timezone",
+    [
+        pytest.param("America/New_York", None, id="col_with_tz_has different timezone"),
+        pytest.param(None, None, id="col_with_tz has no timezone"),
+        pytest.param("Europe/Amsterdam", "America/New_York", id="col_without_tz has a timezone"),
+    ],
+)
+def test_append_and_update_changing_timezones_dynamic(
```
IvoDD, line 535: Another aspect of dynamic schema we should test is that reordering columns still preserves the timezones. E.g.
```
write {"col_a": "Europe/Amsterdam", "col_b": "America/New_York"}
append {"col_b": "America/New_York", "col_a": "Europe/Amsterdam", "col_c": "UTC"}

read should result in:
{"col_a": "Europe/Amsterdam", "col_b": "America/New_York", "col_c": "UTC"}
```

cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -225,7 +225,12 @@ message NormalizationMetadata {
     }
 
     message ExperimentalArrow {
+        message ColumnMeta {
+            optional string timezone = 1;
```
IvoDD, line 229: `optional` keyword is avaialbe with `protobuf >= 3.15` but we pin against `3.5` [here](https://github.com/man-group/ArcticDB/blob/388294a36ad7d97ddcec3cec60654f0f3d3f97d3/setup.cfg#L41)

3.15 has been released in 2021, so we can probably just update our pin and use optional.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -81,42 +85,56 @@ sparrow::array create_packed_bool_array(
...
 template<typename T>
-sparrow::timestamp_without_timezone_nanoseconds_array create_timestamp_array(
-        T* data_ptr, size_t data_size, std::optional<sparrow::validity_bitmap>&& validity_bitmap
+sparrow::array create_timestamp_array(
+        T* data_ptr, size_t data_size, std::optional<sparrow::validity_bitmap>&& validity_bitmap,
+        const date::time_zone* tz = nullptr
 ) {
     static_assert(sizeof(T) == sizeof(sparrow::zoned_time_without_timezone_nanoseconds));
-    // We default to using timestamps without timezones. If the normalization metadata contains a timezone it will be
-    // applied during normalization in python layer.
```
IvoDD, line 89: Have you thought about the decision to apply the timezones in python _normalization.py vs in C++?

On one side, now we have 2 places where we apply timezones. We apply it it here in C++ if we read arrow data as arrow. If we read pandas data as arrow we apply the timezone in Python.

On the other side, doing everything in C++ moves us closer to having a C++ only arrow API.

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -401,27 +401,183 @@ def test_write_owned_and_non_owned_buffers(in_memory_version_store_tiny_segment_
...
+    received = lib.read(sym).data
+    # pyarrow will not concat timezone-aware and timezone-naive arrays together, so cast to result timezone first
+    first_col_timezone = "Europe/Amsterdam" if col_with_tz_timezone == "Europe/Amsterdam" else None
+    write_table = write_table.set_column(
+        0, "col_with_tz", write_table.column(0).cast(pa.timestamp("ns", tz=first_col_timezone))
+    )
+    append_table = append_table.set_column(
+        0, "col_with_tz", append_table.column(0).cast(pa.timestamp("ns", tz=first_col_timezone))
+    )
+    append_table = append_table.set_column(1, "col_without_tz", append_table.column(1).cast(pa.timestamp("ns")))
```
IvoDD, line 577: This is correct, but it took me time to figure out that casting timezone aware to a timezone naive returns the time as it was in UTC.

Maybe a comment explaining this would be useful.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -81,42 +85,56 @@ sparrow::array create_packed_bool_array(
...
+        // Timezone aware
+        sparrow::u8_buffer<timestamp> buffer(
+                reinterpret_cast<timestamp*>(data_ptr), data_size, get_detachable_allocator()
+        );
+        if (validity_bitmap) {
+            return sparrow::array{
+                    sparrow::timestamp_nanoseconds_array{tz, std::move(buffer), std::move(*validity_bitmap)}
+            };
+        } else {
+            return sparrow::array{sparrow::timestamp_nanoseconds_array{tz, std::move(buffer)}};
```
IvoDD, line 117: Maybe an inline comment explaining that arrow always stores the timestamps in UTC irrespective of timezone would help.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -253,20 +277,14 @@ sparrow::array arrow_array_from_block(
...
-    auto arr = [&]() {
-        if constexpr (is_time_type(TagType::DataTypeTag::data_type)) {
-            auto timestamp_array = create_timestamp_array<RawType>(data_ptr, data_size, std::move(maybe_bitmap));
-            return sparrow::array{std::move(timestamp_array)};
-        } else {
-            auto primitive_array = create_primitive_array<RawType>(data_ptr, data_size, std::move(maybe_bitmap));
-            return sparrow::array{std::move(primitive_array)};
-        }
-    }();
+    auto arr = create_primitive_array<RawType>(data_ptr, data_size, std::move(maybe_bitmap));
```
IvoDD, line 281: Now that we no longer accept time types, what do you think adding a `requires` statement on the template that enforces it.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -279,37 +297,44 @@ sparrow::array empty_arrow_array_for_column(const Column& column, std::string_vi
...
+        using DataTagType = typename TagType::DataTypeTag;
+        using RawType = typename DataTagType::raw_type;
         if (column_data.num_blocks() == 0) {
             // For empty columns we want to return one empty array instead of no arrays.
-            vec.emplace_back(empty_arrow_array_for_column(column, name));
+            vec.emplace_back(empty_arrow_array_for_column(column, name, opt_column_meta));
+            return;
         }
+        // Only used with timestamp columns
+        std::optional<const date::time_zone*> opt_tz;
```
IvoDD: Nit: What do you think about simplifying this with:
```
const date::time_zone* tz = timezone(opt_column_meta)
```
and save ourselfes the if below where we check if the timezone was already computed.

For non timeseries columns this would be a nullptr because we would not be able to set timezone metadata to a non timezone column.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -433,25 +479,49 @@ DataType arcticdb_type_from_arrow_array(const sparrow::array& array) {
     }
 }
 
+std::optional<ArrowMeta::ColumnMeta> generate_column_metadata(const sparrow::array& array, DataType data_type) {
+    std::optional<ArrowMeta::ColumnMeta> opt_column_meta;
+    switch (data_type) {
+    case DataType::NANOSECONDS_UTC64: {
+        const auto& proxy = sparrow::detail::array_access::get_arrow_proxy(array);
+        if (const date::time_zone* dtz = sparrow::get_timezone(proxy); dtz != nullptr) {
+            opt_column_meta.emplace();
```
IvoDD: Nit: Instead of `emplace()` we can just construct the opt_column_meta here directly with an empty ColumnMeta.
Then in the bottom we can just `return std::nullopt`

cpp/arcticdb/pipeline/index_utils.cpp
```diff
@@ -140,6 +140,27 @@ proto::descriptors::NormalizationMetadata merge_normalization_metadata(
...
+        for (auto& [col_name, new_col_meta] : new_col_meta_map) {
+            if (auto it = old_col_meta_map.find(col_name); it != old_col_meta_map.end()) {
+                // Column has metadata in both the old and new norm metadata
+                if (!google::protobuf::util::MessageDifferencer::Equals(new_col_meta, it->second)) {
+                    // Column metadata differed between the old and new, so set to timezone naive
+                    new_col_meta.clear_timezone();
+                }
+            } else {
+                // Column has metadata in the new, but not the old norm metadata, so set to timezone naive
+                new_col_meta.clear_timezone();
```
IvoDD: In this case it could also be that the old didn't have the column to begin with.
E.g. we should be able to differentiate between:
```
write {"col_a": tz_naive}
append {"col_a": tz_aware}
```
and
```
write {"col_b": tz_aware}
append {"col_a": tz_aware}
```

we should also add tests for this.

For the proto storage this means that if a timeseries column is missing a timezone we should still store it in the mapping.
E.g. if we have `{"col_a": tz_aware, "col_b": tz_naive}`
The protobuf currently would only contain `{"col_a": ColumnMeta{timezone}}`
Where it should contain `{"col_a": ColumnMeta{timezone}, "col_b": ColumnMeta{none}}`

cpp/arcticdb/pipeline/index_utils.cpp
```diff
@@ -140,6 +140,27 @@ proto::descriptors::NormalizationMetadata merge_normalization_metadata(
...
+            }
+        }
+        // We do not need a similar loop through the old_col_meta_map. If any columns had metadata in the old norm meta,
+        // but do not in the new norm meta, then all we would do is set the column to timezone naive for that column,
+        // which is the current state of the new norm meta anyway
```
IvoDD: We would need to preserve the old norm metadata for columns that existed in the old norm but not in the new.
E.g.
```
write {"col_a": tz_aware}
append {"col_b": tz_aware}
```
result should be
```
{"col_a": ColumnMeta{tz}, "col_b": ColumnMeta{tz}}
```

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: Added a test with reordering, adding, and removing columns, which exposed some bugs you identified in other comments. All good now.

cpp/proto/arcticc/pb2/descriptors.proto
alexowens90: Updated the pin

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: Discussed in person, non-Python language support makes applying in C++ cleaner, and we could push the Pandas norm metadata handling down into the C++ layer for Arrow outputs if we wanted too

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: Added

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: Added

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: `timestamp` is just an alias for `int64_t` so we can't `requires` away `timestamp` support

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: The point was to have one return statement, which I'd messed up. Fixed now

cpp/arcticdb/pipeline/index_utils.cpp
alexowens90: Yeh this was broken for adding and removing columns with dynamic schema, fixed now

cpp/arcticdb/arrow/arrow_utils.cpp
IvoDD: The method templates on `TagType` [here](https://github.com/man-group/ArcticDB/pull/3225/changes/fad2a71df369265800b69804ac4f302a51f54e27#diff-8b01d82fb580cb2019d961fd12a93cc3249e3b9c99b218ea64d523d331dc468bR272), so we can do something like:
```
requires !is_time_type(TagType::DataTypeTag::data_type)
```

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -433,25 +473,48 @@ DataType arcticdb_type_from_arrow_array(const sparrow::array& array) {
...
+std::optional<ArrowMeta::ColumnMeta> generate_column_metadata(const sparrow::array& array, DataType data_type) {
+    std::optional<ArrowMeta::ColumnMeta> opt_column_meta;
+    switch (data_type) {
+    case DataType::NANOSECONDS_UTC64: {
+        opt_column_meta.emplace();
+        const auto& proxy = sparrow::detail::array_access::get_arrow_proxy(array);
+        if (const date::time_zone* dtz = sparrow::get_timezone(proxy); dtz != nullptr) {
+            opt_column_meta->set_timezone(dtz->name());
+        }
+    }
```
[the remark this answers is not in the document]
IvoDD: +1

IvoDD (approved):

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: Ah, you meant on `arrow_array_from_block`, sorry, I thought you were asking for the `requires` on `create_primitive_array`

PR #3228, opened 2026-07-09
Add arrow and pandas interop normalization tests

IvoDD:

It introduces a comprehensive test suite for expected behavior for arrow interoperability with pandas.

This will serve as the base to design arrow normalization metadata.

This PR also includes comprehensive pandas normalization metadata tests, which serve to find all usecases of each field of pandas normalization metadata.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+    lib.write(sym, table, index_column=True)
+
+    received = lib.read(sym, output_format="pandas").data
+    expected = pd.DataFrame({"col": np.arange(2, dtype=np.int64)}, index=_dt(2, tz="America/New_York"))
+    expected.index.name = "ts"
+    assert_frame_equal(expected, received)
+
+
+@pytest.mark.xfail(reason="Arrow column timezone not propagated to pandas norm metadata", strict=True)
+def test_write_arrow_column_timezone_read_pandas(in_memory_version_store_arrow):
```
IvoDD, line 540: Alongside arrow column timezones, we should consider implementing pandas timezone writes for non index columns

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+@pytest.mark.parametrize("first_fmt, second_fmt", FORMATS_ORDER)
+@pytest.mark.parametrize("op", UNINDEXED_OPS)
+def test_combine_matching_schema_rowcount(arrow_library, op, first_fmt, second_fmt):
+    first = _maybe_arrow(pd.DataFrame({"col": [0, 1]}), first_fmt)
+    second = _maybe_arrow(pd.DataFrame({"col": [2, 3]}), second_fmt)
+    received = _combine(arrow_library, op, first, second)
+    assert_frame_equal_with_arrow(received, pd.DataFrame({"col": [0, 1, 2, 3]}))
+
+
+# --- unnamed pandas index / multiindex inherit the arrow name(s) ----------
```
IvoDD: The unnamed pandas index promotion to named column after arrow append is going to be the trickiest part of this proposal to support natively because:

- Existing pandas data will have existing data keys with a predefined column names for the unnamed indices.
- We don't want to rewrite all previous data keys on append to rename the columns
- Static schema is challenging because we could end up breaking invariants (i.e. all slices having the same StreamDescriptor)
- Dynamic schema can also be challenging. We might end up with diffenetly named columns in different rowslices which correspond to the same underlying column.

Only reasonable way to implement this as is will be to ensure the to be appended data preserves the on disk column names and store any renames required in normalization metadata.

A reasonable alternative:
- Raise exceptions on such unnamed indices appends
- Provide users with an API to rename columns on disk so they can make their unnamed indices appendable

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+    df = pd.DataFrame({"col": np.arange(3, dtype=np.int64)}, index=index)
+    lib.write(sym, df)
+
+    received = lib.read(sym).data
+    assert received.column_names == ["__index_level_0__", "__index_level_1__", "col"]
+    assert_frame_equal_with_arrow(received, df)
+
+
+# Tests dealing with column name duplication and invalid arrow names
+# Such pandas dataframes initially won't support combining with arrow (but still must be readable)
```
IvoDD, line 253: The question here is what to advise users that want to append arrow data to enormous pandas dataframes with e.g. duplicate columns.

Option 1: Just override the data with `write(read(output_format="pyarrow").data)`
- Very expensive - full rewrite

Option 2: Provide users with a more efficient API to rename their duplicate columns
- Requires dev work
- Still expensive, we will probably still need to rewrite every data key which contains the column to update its `StreamDescriptor`

cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -121,31 +121,33 @@ message NormalizationMetadata {
...
-        int64 start = 5; // Used for RangeIndex
-        int64 step = 6; // Used for RangeIndex
-        bool is_int = 7;
+        int64 start = 5; // RangeIndex start (used when not physically stored).
+        int64 step = 6; // RangeIndex step (used when not physically stored).
+        bool is_int = 7; // The index name was an integer; cast back from its string form on read.
     }
 
+    // Metadata for a pandas MultiIndex. The first level is the primary (physically stored) index;
+    // higher levels are stored as additional columns.
```
alexowens90: Worth noting that the top-level index column is not necessarily a timeseries

python/tests/unit/arcticdb/version_store/test_pandas_normalization.py
```diff
@@ -0,0 +1,559 @@
...
+    descriptor = lib.version_store.read_descriptor(sym, version_query)
+    return descriptor.timeseries_descriptor.normalization
+
+
+# ---------------------------------------------------------------------------
+# input_type oneof + Pandas.mark
+# ---------------------------------------------------------------------------
+
+
+def test_dataframe_uses_df_input_type(lmdb_version_store_v1):
```
alexowens90: Any reason not to use in-memory for these?

python/tests/unit/arcticdb/version_store/test_pandas_normalization.py
```diff
@@ -0,0 +1,559 @@
...
+
+
+def test_default_range_index_not_physically_stored(lmdb_version_store_v1):
+    """A default RangeIndex is not physically stored; it is rebuilt from ``start``/``step``."""
+    lib = lmdb_version_store_v1
+    sym = "test_default_range_index_not_physically_stored"
+    df = pd.DataFrame({"col": np.arange(4, dtype=np.int64)})
+    lib.write(sym, df)
+
+    index_meta = _read_norm_meta(lib, sym).df.common.index
```
alexowens90: Should assert it has `index` and not `multi_index` first

python/tests/unit/arcticdb/version_store/test_pandas_normalization.py
```diff
@@ -0,0 +1,559 @@
...
+    lib = lmdb_version_store_v1
+    sym = "test_named_datetime_index_physically_stored"
+    df = pd.DataFrame({"col": np.arange(3, dtype=np.int64)}, index=pd.date_range("2025-01-01", periods=3))
+    df.index.name = "ts"
+    lib.write(sym, df)
+
+    index_meta = _read_norm_meta(lib, sym).df.common.index
+    assert index_meta.is_physically_stored
+    assert index_meta.name == "ts"
+    assert not index_meta.fake_name
```
alexowens90, line 157: Can also assert `tz` is empty and `is_int` is false

python/tests/unit/arcticdb/version_store/test_pandas_normalization.py
```diff
@@ -0,0 +1,559 @@
...
+    lib.write(sym, df)
+
+    index_meta = _read_norm_meta(lib, sym).df.common.index
+    assert index_meta.name == "my_range"
+    assert_frame_equal(df, lib.read(sym).data)
+
+
+# ---------------------------------------------------------------------------
+# PandasIndex — physically stored index (is_physically_stored / name / fake_name / is_int / tz)
+# ---------------------------------------------------------------------------
```
alexowens90, line 140: All of these tests could be parametrized over df/series

python/tests/unit/arcticdb/version_store/test_pandas_normalization.py
```diff
@@ -0,0 +1,559 @@
...
+    sym = "test_series_int_name"
+    series = pd.Series(np.arange(3, dtype=np.int64), name=3)
+    lib.write(sym, series)
+
+    common = _read_norm_meta(lib, sym).series.common
+    assert common.has_name
+    assert common.name == "3"
+    received = lib.read(sym).data
+    assert received.name == 3
+    assert_series_equal(series, received)
```
alexowens90, line 322: Could also test that df `has_name` is always false and `name` is always an empty string

python/tests/unit/arcticdb/version_store/test_pandas_normalization.py
```diff
@@ -0,0 +1,559 @@
...
+
+def test_multiindex_higher_level_tz(lmdb_version_store_v1):
+    """Timezones on levels >= 1 go into the per-level ``multi_index.timezone`` map."""
+    lib = lmdb_version_store_v1
+    sym = "test_multiindex_higher_level_tz"
+    df = _multiindex_df(names=["l0", "l1"], tzs=[None, "Europe/London"])
+    lib.write(sym, df)
+
+    timezone = dict(_read_norm_meta(lib, sym).df.common.multi_index.timezone)
+    assert timezone[1] == "Europe/London"
```
alexowens90:
```suggestion
    timezone == {1: "Europe/London"}
```
Asserts no extra entries as well

python/tests/unit/arcticdb/version_store/test_pandas_normalization.py
```diff
@@ -0,0 +1,559 @@
...
+    assert received.index.names[0] == 7
+
+
+# ---------------------------------------------------------------------------
+# Legacy / unimplemented fields
+# ---------------------------------------------------------------------------
+
+
+def test_multiindex_version_field_unused(lmdb_version_store_v1):
+    """``PandasMultiIndex.version`` is legacy and stays at the default 0."""
```
alexowens90: I assume this was intended to version the protobuf so the format could be changed later, by somebody who didn't understand how protobuf compatibility works

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
 import pytest
 
-from arcticdb.exceptions import InternalException, NormalizationException
-from arcticdb.util.test import assert_frame_equal
+from arcticdb import concat
+from arcticdb.exceptions import ArcticException, InternalException, NormalizationException, SchemaException
+from arcticdb.util.test import assert_frame_equal, assert_series_equal, assert_frame_equal_with_arrow
+
+
+def _dt(periods, tz=None, start="2025-01-01"):
```
alexowens90: Personal taste, but I find this kind of abstraction unhelpful. I know what the line below does at a glance, but `_dt(1)` could mean anything

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+
+
+def _indexed_arrow_table(index_name, values, tz=None, start="2025-01-01"):
+    """Arrow table whose first column is a (optionally tz-aware) timestamp index."""
+    index = _dt(len(values), tz=tz, start=start)
+    return pa.table({index_name: _ts_array(index), "col": pa.array(values, pa.int64())})
+
+
+def _pandas_ts(values, start="2025-01-01", tz=None, name="ts"):
+    """pandas DataFrame with a named (optionally tz-aware) DatetimeIndex and a single int column."""
```
alexowens90: Surprised there's no use case in the testing for unnamed indexes?

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
alexowens90: I think we cross this bridge when we're asked to, and only if the rewrite with more sensible column names really isn't an option.
Column-renaming functionality has been a feature request before. I think we would implement it as a mapping in the index layer rather than data key re-write, so wouldn't be too expensive.

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+    assert received.column_names == [
+        "____index_level_0____",
+        "col",
+        "__index_level_0__",
+        "___index_level_0___",
+        "_____index_level_0_____",
+        "_col_",
+        "__col__",
+    ]
+    assert_frame_equal_with_arrow(received, df)
```
alexowens90, line 486: I added a bunch of tests like these in https://github.com/man-group/ArcticDB/pull/2759
Worth removing duplicates, and pulling anything covered by those not covered here into this file

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
alexowens90: I've been thinking about this. Right now, we have Pandas and Arrow norm metadata, but actually what we have added for Arrow so far (i.e. `has_index` and timezones) are not actually output-format specific, they're really core to the schema of the data. I wonder if there should be a base-layer to the norm metadata that does not depend on the writing format, and then the Arrow-specific metadata is just things that are really Arrow specific (like string format)

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
-    received = lib.read(sym).data
-    assert_frame_equal(df, received)
+
+@pytest.mark.xfail(reason="pyarrow Array input should read back as a pandas Series, not be pickled", strict=True)
+def test_write_arrow_array_read_pandas(in_memory_version_store_arrow):
+    lib = in_memory_version_store_arrow
+    sym = "test_write_arrow_array_read_pandas"
+    lib.write(sym, pa.array([1, 2, 3], pa.int64()))
+    received = lib.read(sym, output_format="pandas").data
+    assert isinstance(received, pd.Series)
```
alexowens90, line 589: One could (but I am not) argue that array and chunked array are more analogous to a 1D numpy array than a Series, but that way madness lies

alexowens90: Implementing the inverse (write 1D numpy array, read as Arrow chunked array) is more sensible, but I would wait for a user request as I'm not sure how widely used numpy arrays are in practice

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
-    lib = lmdb_version_store_arrow
-    sym = "test_update_arrow_with_pandas"
-    df = pd.DataFrame({"col1": np.arange(2, dtype=np.int64)}, index=pd.date_range("2025-01-01", periods=2))
+@pytest.mark.parametrize("first_fmt, second_fmt", FORMATS_ORDER)
+@pytest.mark.parametrize("op", INDEXED_OPS)
+def test_combine_matching_schema_indexed(arrow_library, op, first_fmt, second_fmt):
+    first = _maybe_arrow(_pandas_ts([0, 1]), first_fmt)
+    second = _maybe_arrow(_pandas_ts([2, 3], start="2025-01-03"), second_fmt)
+    received = _combine(arrow_library, op, first, second, index_column=True)
+    assert_frame_equal_with_arrow(received, _pandas_ts([0, 1, 2, 3]))
```
alexowens90, line 693: For these kind of tests, is it worth being more explicit about what the output looks like, rather than relying in generic equality functions like these?

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
alexowens90: The more I think about this, the more I think that offering a column rename API makes more sense than massively complicating our implementation just to support this use case. By definition, if users are swapping from Pandas to Arrow in their writes, they have to come up with column names for those indexes to use in the Arrow tables. Any readers still using Pandas are unlikely to be relying on the [multi[_index columns being named `None`, and so will silently continue working when those columns suddenly get names.

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+    assert_frame_equal_with_arrow(received, _pandas_ts([0, 1, 2, 3], tz="America/New_York"))
+
+
+@pytest.mark.parametrize("first_fmt, second_fmt", FORMATS_ORDER)
+@pytest.mark.parametrize("op", ["append", "update"])
+@pytest.mark.xfail(
+    reason="tz-mismatch under static schema should raise SchemaException once cross-format combine lands", strict=True
+)
+def test_combine_index_tz_mismatch_static_raises(arrow_library, op, first_fmt, second_fmt):
+    """Static schema: a timezone mismatch on the index must raise SchemaException."""
```
alexowens90, line 838: I agree for append/update, but I think concat should allow this

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+
+
+@pytest.mark.parametrize("first_fmt, second_fmt", FORMATS_ORDER)
+@pytest.mark.parametrize("op", ["append", "update"])
+@pytest.mark.xfail(
+    reason="non-index column tz mismatch handling for pandas<->arrow combine is not implemented yet", strict=True
+)
+def test_combine_column_tz_mismatch_static_raises(arrow_library, op, first_fmt, second_fmt):
+    """Static schema: a non-index column that is tz-aware in arrow but tz-naive in pandas must
+    raise."""
```
alexowens90, line 892: Same comment as above on concat

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+@pytest.mark.parametrize("op", UNINDEXED_OPS)
+@pytest.mark.xfail(reason="merging a Series with a polars Series is not supported yet", strict=True)
+def test_combine_series_with_polars_series(arrow_library, op):
+    received = _combine(arrow_library, op, pd.Series(np.array([0, 1], dtype=np.int64)), pl.Series("col", [2, 3]))
+    assert received.to_pylist() == [0, 1, 2, 3]
+
+
+@pytest.mark.parametrize("op", INDEXED_OPS)
+def test_combine_series_with_index_and_table(arrow_library, op):
+    """An indexed Series (stored as a one-column frame) combines with a matching arrow table."""
```
alexowens90, line 930: Side-note, what happens at the moment if you try and combine a Series with a 1-column dataframe? (with and without an index) I would expect some mental stuff with dynamic schema, and maybe something sane if the series name matches the column name in the df? As I think we store Series as a df with the only column sharing the name of the Series?

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
-        {
-            "ts": pa.Array.from_pandas(pd.date_range("2025-01-02", periods=2), type=pa.timestamp("ns")),
-            "col1": pa.array([2, 3], pa.int64()),
-        }
+    with pytest.raises(ArcticException):
+        lib.append("sym", df)
+
+
+def test_combine_pandas_empty_column_raises(arrow_library):
+    """Merging arrow with a pandas frame that has an empty column name must raise"""
```
alexowens90: I think an empty string is a valid column name in Arrow?
Right now this is testing a generic schema mismatch

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
```diff
@@ -6,97 +6,737 @@
...
+def test_append_single_index_pandas_with_unindexed_arrow_raises(arrow_library_any_schema):
+    """A single-index (timeseries) pandas symbol cannot be appended with an unindexed arrow table."""
+    lib = arrow_library_any_schema
+    lib.write("sym", _pandas_ts([0, 1]))
+    with pytest.raises(ArcticException):
+        lib.append("sym", pa.table({"col": pa.array([2, 3], pa.int64())}))
+
+
+def test_append_multiindex_pandas_with_unindexed_arrow_raises(arrow_library_any_schema):
+    """A MultiIndex pandas symbol cannot be appended with an unindexed arrow table."""
```
alexowens90: I think if the top-level of the multiindex is not a timeseries then this should be allowed?

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
IvoDD: I agree, however reusing the `CommonMetadata` in pandas metadata would be a challenge. We'll still need to maintain the old written data with pandas metadata.

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
IvoDD: We raise `NormalizationException`. Added tests for this `test_pandas_normalization`. For now I made the arrow pandas interop preserve behavior and added tests for this.

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
IvoDD: Sparrow seems to have a bug with accepting "" as a column name. Added a separate test for that issue and ammended this test.

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
IvoDD: Agreed, added a new test for this

python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py
IvoDD: It is used but index name is optionally cleared later [here](https://github.com/man-group/ArcticDB/blob/63758ef635fad02a8ab3dca9b01f22f261aa2b8d/python/tests/unit/arcticdb/version_store/test_arrow_pandas_interop.py#L715). Also passing name=None will keep the index unnamed

alexowens90 (approved):

PR #3237, opened 2026-07-13
Performance 12498089633: Make more index key reading async

alexowens90:
#### Reference Issues/PRs
[12498089633](https://man312219.monday.com/boards/7852509418/pulses/12498089633)

`read_index_key_without_column_stats` called `read_sync` to get the index key in many async contexts that effectively added a blocking IO to what were meant to be completely asynchronous functions.
This makes `read_index_key_without_column_stats` async, and chains futures from it where necessary (synchronous callers just call `.get()` immediately. The sync version `read_index_key_without_column_stats_sync` has been added for use in `merge_update_impl` to avoid rebase nightmares with ongoing work, @vasil-pashov will switch it to using the async version and remove the sync version as part of one of his PRs in this area.

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -1980,7 +1980,27 @@ folly::Future<SegmentInMemory> prepare_output_frame(
             .thenValue([frame](auto&&) { return frame; });
 }
 
-IndexInformation read_index_key_without_column_stats(const std::shared_ptr<Store>& store, const AtomKey& key) {
+// This future will complete on an IO thread
+folly::Future<IndexInformation> read_index_key_without_column_stats(
+        const std::shared_ptr<Store>& store, const AtomKey& key
+) {
+    return store->read(key).thenTryInline([key = std::move(key)](auto&& key_seg_pair_try) {
+        if (key_seg_pair_try.hasValue()) {
+            auto column_stats = std::nullopt;
+            return IndexInformation{std::move(key_seg_pair_try.value()), column_stats};
+        } else {
+            throw storage::NoDataFoundException(fmt::format(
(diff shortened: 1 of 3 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -1980,7 +1980,27 @@ folly::Future<SegmentInMemory> prepare_output_frame(
...
+// This future will complete on an IO thread
+folly::Future<IndexInformation> read_index_key_without_column_stats(
+        const std::shared_ptr<Store>& store, const AtomKey& key
+) {
+    return store->read(key).thenTryInline([key = std::move(key)](auto&& key_seg_pair_try) {
+        if (key_seg_pair_try.hasValue()) {
+            auto column_stats = std::nullopt;
+            return IndexInformation{std::move(key_seg_pair_try.value()), column_stats};
+        } else {
+            throw storage::NoDataFoundException(fmt::format(
```
IvoDD, line 1992: I know this was the case before, but it seems odd to reraise everything as a `NoDataFoundException`.

I.e. surely we don't want to mask a network error as `NoDataFound`. 
I'd be in favor of reraising the `key_seg_pair_try.exception()` type. 

There are probably other places we do this so this should probably be a separate PR.

IvoDD (approved):

cpp/arcticdb/version/version_core.cpp
alexowens90: Agreed. Raising consistent exception types in general across all code paths for versions not existing/keys not existing would be a nice thing to do next time we're doing a major version bump for something more important. I found [this crap](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/version/local_versioned_engine.cpp#L1915) as part of my refactor (which is the same for `update`), which is correctly a `NoSuchVersionException` for the batch methods.

PR #3238, opened 2026-07-13
Homogenise batch and non-batch implementations in LocalVersionedEngine, and fix 12548384234, 12548383860, 12584371808, and 12548367093

alexowens90:
#### Reference Issues/PRs
[12548384234](https://man312219.monday.com/boards/7852509418/pulses/12548384234)
[12548383860](https://man312219.monday.com/boards/7852509418/pulses/12548383860)
[12548371808](https://man312219.monday.com/boards/7852509418/pulses/12548371808)
[12548367093](https://man312219.monday.com/boards/7852509418/pulses/12548367093)

The main part of the change refactors write, append, update, write_metadata, and their batch variants in `LocalVersionedEngine`. There was previously a lot of duplicated code between both the different modification methods, and between modification methods and their batch versions. There were also many inconsistencies in how they initially loaded version chains, handled empty frames, wrote back to the version chain, and added symbol list keys, which led to a lot of the bugs listed above.
While performing the refactor, I noticed a few other opportunities to cleanup:

- The `UpdateMetadataTask` was a different pattern to the other modification methods. This has been moved to `async_write_metadata_impl`, and it's threading model improved.
- The sync `*_impl` methods in `version_core.cpp` just called the async methods and `.get()` immediately. This responsibility has been pushed up to methods in `LocalVersionedEngine`
- `get_write_options()` was called a lot to convert protobuf to a struct. However, the protobuf is never modified after library initialization, so this struct is now just generated once and stored as a member of `LocalVersionedEngine`
- The `sparsify_floats` member of `WriteOptions` was redundant, and has been removed

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -150,7 +150,7 @@ folly::Future<folly::Unit> LocalVersionedEngine::delete_unreferenced_pruned_inde
         std::vector<AtomKey>&& pruned_indexes, const AtomKey& key_to_keep
 ) {
     try {
-        if (!pruned_indexes.empty() && !cfg().write_options().delayed_deletes()) {
+        if (!pruned_indexes.empty() && !write_options_.delayed_deletes) {
             // TODO: the following function will load all snapshots, which will be horrifyingly inefficient when called
             // multiple times from batch_*
             auto [not_in_snaps, in_snaps] = get_index_keys_partitioned_by_inclusion_in_snapshots(
@@ -488,7 +488,6 @@ VersionedItem LocalVersionedEngine::read_modify_write_internal(
             source_version.has_value(), "Could not find requested version for symbol \"{}\"", source_stream
     );
 
-    const WriteOptions write_options = get_write_options();
     auto [maybe_prev, deleted] = ::arcticdb::get_latest_version(store(), version_map(), target_stream);
     const auto target_version = get_next_version_from_key(maybe_prev);
 
@@ -502,7 +501,7 @@ VersionedItem LocalVersionedEngine::read_modify_write_internal(
                                            store(),
                                            read_query,
                                            read_options,
-                                           write_options,
+                                           write_options_,
                                            target_partial_index_key,
                                            pipeline_context,
                                            std::move(user_meta_proto)
@@ -633,17 +632,17 @@ std::vector<std::variant<DescriptorItem, DataError>> LocalVersionedEngine::batch
 void LocalVersionedEngine::flush_version_map() { version_map()->flush(); }
 
 std::shared_ptr<DeDupMap> LocalVersionedEngine::get_de_dup_map(
-        const StreamId& stream_id, const std::optional<AtomKey>& maybe_prev, const WriteOptions& write_options
+        const StreamId& stream_id, const UpdateInfo& update_info, const WriteOptions& write_options
 ) {
     auto de_dup_map = std::make_shared<DeDupMap>();
-    if (write_options.de_duplication) {
-        if (auto latest_undeleted_index_key = get_latest_undeleted_version(store(), version_map(), stream_id)) {
-            const auto data_keys =
-                    get_data_keys(store(), {std::move(*latest_undeleted_index_key)}, storage::ReadKeyOpts{});
+    // Next version id == 0 => this is the first version, so there is nothing to dedup against
+    if (write_options.de_duplication && update_info.next_version_id_ > 0) {
+        if (update_info.previous_index_key_.has_value()) {
+            const auto data_keys = get_data_keys(store(), *update_info.previous_index_key_, storage::ReadKeyOpts{});
             for (const auto& data_key : data_keys) {
                 de_dup_map->insert_key(data_key);
             }
-        } else if (maybe_prev && write_options.snapshot_dedup) {
+        } else if (write_options.snapshot_dedup) {
             // This means we don't have any live versions(all tombstoned), so will try to dedup from snapshot versions
             auto snap_versions = get_index_keys_in_snapshots(store(), stream_id);
             auto latest_snapshot_it = ranges::max_element(snap_versions, [](const auto& k1, const auto& k2) {
@@ -663,7 +662,7 @@ std::shared_ptr<DeDupMap> LocalVersionedEngine::get_de_dup_map(
 VersionedItem LocalVersionedEngine::sort_index(
         const StreamId& stream_id, bool dynamic_schema, bool prune_previous_versions
 ) {
-    auto update_info = get_latest_undeleted_version_and_next_version_id(store(), version_map(), stream_id);
+    auto update_info = get_next_version_id_and_optionally_latest_undeleted_version(store(), version_map(), stream_id);
     util::check(update_info.previous_index_key_.has_value(), "Cannot sort_index a non-existent symbol {}", stream_id);
     auto [index_segment_reader, slice_and_keys] =
             index::read_index_to_vector(store(), *update_info.previous_index_key_);
@@ -708,9 +707,8 @@ VersionedItem LocalVersionedEngine::sort_index(
 VersionedItem LocalVersionedEngine::delete_range_internal(
         const StreamId& stream_id, const UpdateQuery& query, const DeleteRangeOptions& option
 ) {
-    auto update_info = get_latest_undeleted_version_and_next_version_id(store(), version_map(), stream_id);
-    auto versioned_item =
-            delete_range_impl(store(), stream_id, update_info, query, get_write_options(), option.dynamic_schema_);
+    auto update_info = get_next_version_id_and_optionally_latest_undeleted_version(store(), version_map(), stream_id);
+    auto versioned_item = delete_range_impl(store(), stream_id, update_info, query, option.dynamic_schema_);
     write_version_and_prune_previous(
             option.prune_previous_versions_, versioned_item.key_, update_info.previous_index_key_
     );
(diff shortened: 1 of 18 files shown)


python/tests/integration/toolbox/test_query_stats.py
alexowens90, line 1: These tests are a bit odd in that they verify empirically measured current behaviour, rather than desired behaviour. I considered replacing equality checks with `<=` for object get/put counts (since reducing IO counts without changing functionality is always a net positive), but that runs the risk that one change reduces an IO count (with the tests continuing to pass), and then a subsequent change increases it again (a regression), but the tests still pass.

cpp/arcticdb/version/version_functions.hpp
```diff
@@ -38,16 +38,21 @@ inline std::pair<std::optional<AtomKey>, bool> get_latest_version(
...
 // The next version ID returned will be 0 for brand new symbols, or one greater than the largest ever version created so
 // far
-inline version_store::UpdateInfo get_latest_undeleted_version_and_next_version_id(
-        const std::shared_ptr<Store>& store, const std::shared_ptr<VersionMap>& version_map, const StreamId& stream_id
+inline version_store::UpdateInfo get_next_version_id_and_optionally_latest_undeleted_version(
+        const std::shared_ptr<Store>& store, const std::shared_ptr<VersionMap>& version_map, const StreamId& stream_id,
+        bool get_latest_undeleted_version = true
 ) {
-    ARCTICDB_SAMPLE(GetLatestUndeletedVersionAndHighestVersionId, 0)
-    LoadStrategy load_strategy{LoadType::LATEST, LoadObjective::UNDELETED_ONLY};
```
alexowens90, line 45: `UNDELETED_ONLY` was overkill when calling `[batch_]write` with dedup disabled. See `test_write_version_chain_minimal_io` for an example

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -633,17 +632,16 @@ std::vector<std::variant<DescriptorItem, DataError>> LocalVersionedEngine::batch
 void LocalVersionedEngine::flush_version_map() { version_map()->flush(); }
 
 std::shared_ptr<DeDupMap> LocalVersionedEngine::get_de_dup_map(
-        const StreamId& stream_id, const std::optional<AtomKey>& maybe_prev, const WriteOptions& write_options
+        const StreamId& stream_id, const UpdateInfo& update_info, const WriteOptions& write_options
 ) {
     auto de_dup_map = std::make_shared<DeDupMap>();
-    if (write_options.de_duplication) {
-        if (auto latest_undeleted_index_key = get_latest_undeleted_version(store(), version_map(), stream_id)) {
```
alexowens90, line 640: Previously, this would have reloaded the version chain. Now, by construction, if dedup is enabled `update_info.previous_index_key_` will contain the previous live version if any exists

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -633,17 +632,16 @@ std::vector<std::variant<DescriptorItem, DataError>> LocalVersionedEngine::batch
...
-            const auto data_keys =
-                    get_data_keys(store(), {std::move(*latest_undeleted_index_key)}, storage::ReadKeyOpts{});
+    if (write_options.de_duplication && update_info.next_version_id_ > 0) {
+        if (update_info.previous_index_key_.has_value()) {
+            const auto data_keys = get_data_keys(store(), *update_info.previous_index_key_, storage::ReadKeyOpts{});
             for (const auto& data_key : data_keys) {
                 de_dup_map->insert_key(data_key);
             }
-        } else if (maybe_prev && write_options.snapshot_dedup) {
```
alexowens90, line 646: Previously, `maybe_prev` contained the latest live or deleted index key. This check was therefore just used to ensure we didn't try and dedup version 0, which is inherently dedupable. Replace by earlier `update_info.next_version_id_ > 0`.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -862,22 +788,15 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
     ARCTICDB_SAMPLE(WriteVersionedDataFrame, 0)
     py::gil_scoped_release release_gil;
     ARCTICDB_RUNTIME_DEBUG(log::version(), "Command: write_versioned_dataframe");
-    auto [maybe_prev, deleted] = ::arcticdb::get_latest_version(store(), version_map(), stream_id);
```
alexowens90, line 865: This was a previous inconsistency between `write` and `batch_write`. `write` would just load the next version ID (i.e. not necessarily the latest live version). This is good for IO, except for when dedup is enabled, when the version chain must then be traversed again to get the latest live version. `batch_write` always loaded the previous live version, which could cause excessive IO. Now everything has been homogenised around `[batch_]get_next_version_id_and_optionally_latest_undeleted_version[_async]`

python/tests/integration/toolbox/test_query_stats.py
alexowens90: This comment pretty much exactly also applies to `test_num_storage_operations.py`

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -792,12 +779,14 @@ VersionedItem PythonVersionStore::write_versioned_composite_data(
...
-    auto multi_key = write_multi_index_entry(store(), index_keys, stream_id, metastruct, user_meta, version_id);
+    auto multi_key = write_multi_index_entry(
+            store(), index_keys, stream_id, metastruct, user_meta, update_info.next_version_id_
+    );
     auto versioned_item = VersionedItem(to_atom(std::move(multi_key)));
-    write_version_and_prune_previous(prune_previous_versions, versioned_item.key_, maybe_prev);
+    write_version_and_prune_previous(prune_previous_versions, versioned_item.key_, update_info.previous_index_key_);
 
     if (cfg().symbol_list())
-        symbol_list().add_symbol(store(), stream_id, version_id);
```
poodlewars, line 800: I'm happy you've kept the logic the same, but I thought the reference ID on the symbol list key was supposed to be the version number we saw when we added the symbol, not the new version number? Is that not actually how it works?

cpp/arcticdb/version/version_utils.hpp
```diff
@@ -356,14 +357,17 @@ inline bool key_exists_in_ref_entry(
         const LoadStrategy& load_strategy, const VersionMapEntry& ref_entry,
         std::optional<AtomKey>& cached_penultimate_key
 ) {
+    // If we are looking for the latest version, deleted or undeleted, then an index key in the ref key is always
+    // the answer. By construction, any index or tombstone keys further down the chain will have lower version IDs than
+    // this one
```
poodlewars, line 362: I don't quite follow this. If we're looking for the latest undeleted version, how do we know that it must be in the ref key?

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
poodlewars, line 684: Is there a similar test for `write_metadata` singular?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -862,22 +789,15 @@ VersionedItem LocalVersionedEngine::write_versioned_dataframe_internal(
...
-    auto version_id = get_next_version_from_key(maybe_prev);
-    ARCTICDB_DEBUG(
-            log::version(), "write_versioned_dataframe for stream_id: {} , version_id = {}", stream_id, version_id
+    // We don't need to load the latest live version unless dedup is enabled. If dedup is disabled, we only care about
+    // the next version ID, which can save some IO traversing the version chain
```
poodlewars, line 793: Was there a test for this? (IO performed for writes with dedup enabled)

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1734,20 +1654,185 @@ folly::Future<std::vector<AtomKey>> LocalVersionedEngine::batch_write_internal(
...
+                    "throwing this exception.",
+                    stream_id
+            );
+            if (batch) {
+                missing_data::raise<ErrorCode::E_NO_SUCH_VERSION>(error_msg);
+            } else {
+                util::raise_rte(error_msg);
+            }
+        }
+        // Replace above with this as part of Monday ticket 12551552848 for 7.0.0
```
poodlewars, line 1782: Why do we need to wait for 7.0.0 to do this? Is the non-batch method raising an `InternalException` up to Python? I think we can argue that's a bug, and returning a more specific exception type is unlikely to break existing code.

poodlewars (approved):

cpp/arcticdb/version/version_store_api.cpp
alexowens90: I've just checked and it seems like with `__add__` symbol list entries the reference ID is the version number of the index key being written (both before and after this change)
I think it's with `__delete__` entries that we use the highest version number used so far.

cpp/arcticdb/version/version_utils.hpp
alexowens90: We don't, but the condition below is explicitly that the ref key contains an index key, i.e. that the last operation on the version chain was to write a new version. By definition this must be the latest undeleted version.

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
alexowens90: Not obviously, I'll add one

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: No, I can add one

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Exception types are a part of our API, even when buggy. e.g. users could be running this code in production:
```python
try:
    lib.update(sym, df, upsert=False)
except InternalExcepttion:
   # Handle symbol not existing in some other way
```
Which would break if we change the type, even though we consider it to be wrong. This is particularly sensitive because this has been the exception type for this inside Man for a very long time.

poodlewars: Ah my mistake, I thought this was true in which case the API break would have been marginal

```
In [1]: from arcticdb.exceptions import InternalException

In [2]: from arcticdb.exceptions import NoDataFoundException

In [3]: isinstance(NoDataFoundException, InternalException)
Out[3]: False
```

PR #3239, opened 2026-07-13
Implement insertion strategy for merge update in timeseries

vasil-pashov:
# Merge update - Insert
## High level overview
This PR implements both strategies `MergeStrategy("do_nothing", "insert")` and `MergeStrategy("update", "insert")`. Both the source and the target must be DateTime-indexed and ordered. Insertion is stable, meaning that new values will come after values already in the target.
### Structure by time slice
Only source rows that are not matched by anything in the target are inserted. This means that if an index value appears in more than one segment all segments containing that index value must be loaded by the processing unit. This is implemented in `structure_by_time_slice`. This function structures together all segments that contain the same index values. With this change one segment can appear in multiple groups. Example
```
[[1, 2, 3],[4, 4, 4], [4, 5, 5], [5, 5, 5], [5, 6, 7], [7, 8, 9]]
```
Will produce the following groups:
```
[[[1, 2, 3]], [[4, 4, 4], [4, 5, 5]], [[4, 5, 5], [5, 5, 5], [5, 6, 7]], [[5, 6, 7], [7, 8, 9]]]
```
With corresponding time ranges
```
[[1, 4), [4, 5), [4, 8), [7, 10)]
```
**Note 1**. that the procedure intentionally splits the groups so that when segments are chained like so `[a, b] [b, c] [c, d], [d, e]` not all segments are loaded in the same processing unit.
**Note 2.** With this approach different processing unit not only can have overlapping index ranges. Two different processing units can have the same index range. Example:
```
[[10], [10, 10, 30], [30, 30]]
```
Will create two groups
```
[[[10], [10, 10, 30]], [[10, 10, 30], [30, 30]]]
```
Both groups span the same index range `[10;31)`
**Note 3.** Only the first and last row slice of a group can be requested by more than one processing unit
**Note 4.** Two processing units will never operate on the same set of data (see `get_target_start_end`). If the first row slice of a group has entity fetch count > 1 this means that it's the last row slice of a previous group. This means that the first index value of this row slice was repeated in the previous group. That group will handle this index value. This processing unit must skip the repeated index value. Analogous if the last segment has entity fetch count > 1, it must be cut at the end. Using the example above. 
```
[[4, 5, 5], [5, 5, 5], [5, 6, 7]] // [5, 6, 7] has entity fetch count > 1. The processing unit will process all vaues in range [4;6)
[[5, 6, 7], [7, 8, 9]] // [5, 6, 7] has entity fetch count > 1. The processing unit will process all values in range [6; 10)
```
#### Performance consideration
Structure for processing still marks all segments in a time sliced group for loading if there's an index value appearing in the time range of the index group. When this is the value that spans multiple segments this will create more work, as some segments that are not matched will be loaded. Using the example above if the source contains the value `5` we'll end up with the following processing units
```
[4, 4, 4], [4, 5, 5]
[4, 5, 5], [5, 5, 5], [5, 6, 7]
[5, 6, 7], [7, 8, 9]
```
Clearly only the one in the middle needs actual work.
## Merge
The `merge` function performs both inserting new data and updating matched rows. It produces **a single one** column. Essentially it is merging of two sorted lists, with the caveat that the target is actually list of lists, not a single list. However all values in `concatenate(target_index)` are sorted. The nuisance is that there's additional bookkeeping of which is the current target list.
Since the target rows to be updated can be in arbitrary order updating happens in two steps
1. Write all data values corresponding to index value `x` (including new values)
2. Iterate over the rows to update and overwrite with the new value
	1. This must keep track of how many values were inserted up to now in order to get the proper offset
	2. This makes strings sub-optimal. It first adds all strings in the global merged pool but overwriting might orphan some string values. The pool needs rebuilding at the end.
Merge will also truncate columns so that the output contains only values for which the processing unit is responsible. Using the the example above if the source contains value `5` the processing units will be
```
[4, 4, 4], [4, 5, 5]
[4, 5, 5], [5, 5, 5], [5, 6, 7]
[5, 6, 7], [7, 8, 9]
```
assuming no insertion the output will be
```
[4, 4, 4, 4]
[5, 5, 5, 5, 5, 5]
[6, 7, 7, 8, 9]
```
### Further work
Currently merging uses the same algorithm for the index and the data columns. A better approach would be to run the current algorithm for the index once and make it produce a recipe of the sort:
```c++
enum DataSource {SOURCE, TARGET};
struct Entry {
	DataSource src;
	size_t offset_from_source_start;
	size_t count;
};
std::vector<Entry> merge_recipe;
```
and apply this to all columns.
## Index reconstruction
The reconstruction of the index key has changed a bit. Up to now there was a `1:1` mapping in the slices and keys of the old keys and the new keys because update did not create new data. Now the processing unit will emit different row ranges. The new row ranges will represent the combined row range of the whole group of row slices processed by the segment, it will **not** account the added rows. The added rows are reported separately in the `MergeUpdateInsertedRowsEntity` structure fetched from the component manager.
## MatchRecord
Helper structure to handle all the matching logic together and encapsulated. The biggest change is that the structure of matched rows is no longer a list of lists (for every source row all target rows that match it). It's list of list of lists. For each row slice for each source source row all target rows that match it.

### DIFF
--- a/cpp/arcticdb/entity/native_tensor.hpp
+++ b/cpp/arcticdb/entity/native_tensor.hpp
@@ -141,6 +141,55 @@ struct NativeTensor {
         return (&(reinterpret_cast<const T*>(ptr)[signed_pos]));
     }
 
+    template<typename T>
+    std::span<const T> span(size_t offset = 0, size_t count = std::dynamic_extent) const {
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE, ndim() == 1, "Cannot create a span from a multi-dimensional tensor"
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                elsize() == sizeof(T),
+                "Mismatched tensor byte size {} and span element type size {}",
+                elsize(),
+                sizeof(T)
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                count == std::dynamic_extent || count <= nbytes() / sizeof(T) - offset,
+                "Span count {} is out of bounds for tensor of size {}",
+                count,
+                nbytes() / sizeof(T)
+        );
+        return std::span<const T>(
+                static_cast<const T*>(ptr) + offset,
+                count == std::dynamic_extent ? nbytes() / sizeof(T) - offset : count
+        );
+    }
+
+    template<typename T>
+    std::span<T> span(size_t offset = 0, size_t count = std::dynamic_extent) {
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE, ndim() == 1, "Cannot create a span from a multi-dimensional tensor"
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                elsize() == sizeof(T),
+                "Mismatched tensor byte size {} and span element type size {}",
+                elsize(),
+                sizeof(T)
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                count == std::dynamic_extent || count <= nbytes() / sizeof(T) - offset,
+                "Span count {} is out of bounds for tensor of size {}",
+                count,
+                nbytes() / sizeof(T)
+        );
+        return std::span<T>(
+                static_cast<T*>(ptr) + offset, count == std::dynamic_extent ? nbytes() / sizeof(T) - offset : count
+        );
+    }
+
     // returns number of elements, not bytesize
     [[nodiscard]] ssize_t size() const { return calc_elements(shape(), ndim()); }
 
@@ -259,6 +308,7 @@ struct TypedTensor : public NativeTensor {
     }
 
     const T* data() const { return static_cast<const T*>(NativeTensor::data()); }
+    std::span<const T> span() const { return std::span<const T>(data(), size()); }
 
   private:
     void check_ptr_within_bounds(const NativeTensor& tensor, size_t rows) {
--- a/cpp/arcticdb/entity/native_tensor.hpp
+++ b/cpp/arcticdb/entity/native_tensor.hpp
@@ -141,53 +141,27 @@ struct NativeTensor {
         return (&(reinterpret_cast<const T*>(ptr)[signed_pos]));
     }
 
+    /// @brief Returns a std::span over the tensor data.
+    /// @param offset Number of elements to skip from the start of the tensor.
+    /// @param count Number of elements after offset to include in the span. If std::dynamic_extent, the span will
+    /// include all elements from the offset to the end of the tensor.
     template<typename T>
-    std::span<const T> span(size_t offset = 0, size_t count = std::dynamic_extent) const {
-        ARCTICDB_DEBUG_CHECK(
-                ErrorCode::E_ASSERTION_FAILURE, ndim() == 1, "Cannot create a span from a multi-dimensional tensor"
-        );
-        ARCTICDB_DEBUG_CHECK(
-                ErrorCode::E_ASSERTION_FAILURE,
-                elsize() == sizeof(T),
-                "Mismatched tensor byte size {} and span element type size {}",
-                elsize(),
-                sizeof(T)
-        );
-        ARCTICDB_DEBUG_CHECK(
-                ErrorCode::E_ASSERTION_FAILURE,
-                count == std::dynamic_extent || count <= nbytes() / sizeof(T) - offset,
-                "Span count {} is out of bounds for tensor of size {}",
-                count,
-                nbytes() / sizeof(T)
-        );
+    std::span<const T> span(const size_t offset = 0, const size_t count = std::dynamic_extent) const {
+        validate_can_take_span<T>(offset, count);
         return std::span<const T>(
                 static_cast<const T*>(ptr) + offset,
                 count == std::dynamic_extent ? nbytes() / sizeof(T) - offset : count
         );
     }
 
+    /// @brief Returns a std::span over the tensor data.
+    /// @param offset Number of elements to skip from the start of the tensor.
+    /// @param count Number of elements after offset to include in the span. If std::dynamic_extent, the span will
+    /// include all elements from the offset to the end of the tensor.
     template<typename T>
-    std::span<T> span(size_t offset = 0, size_t count = std::dynamic_extent) {
-        ARCTICDB_DEBUG_CHECK(
-                ErrorCode::E_ASSERTION_FAILURE, ndim() == 1, "Cannot create a span from a multi-dimensional tensor"
-        );
-        ARCTICDB_DEBUG_CHECK(
-                ErrorCode::E_ASSERTION_FAILURE,
-                elsize() == sizeof(T),
-                "Mismatched tensor byte size {} and span element type size {}",
-                elsize(),
-                sizeof(T)
-        );
-        ARCTICDB_DEBUG_CHECK(
-                ErrorCode::E_ASSERTION_FAILURE,
-                count == std::dynamic_extent || count <= nbytes() / sizeof(T) - offset,
-                "Span count {} is out of bounds for tensor of size {}",
-                count,
-                nbytes() / sizeof(T)
-        );
-        return std::span<T>(
-                static_cast<T*>(ptr) + offset, count == std::dynamic_extent ? nbytes() / sizeof(T) - offset : count
-        );
+    std::span<T> span(const size_t offset = 0, const size_t count = std::dynamic_extent) {
+        const std::span<const T> as_const = std::as_const(*this).span<T>(offset, count);
+        return {const_cast<T*>(as_const.data()), as_const.size()};
     }
 
     // returns number of elements, not bytesize
@@ -207,6 +181,43 @@ struct NativeTensor {
     /// API providing the strides and shapes arrays, expanded_dim is what ArcticDB thinks of the tensor and using it
     /// can lead to out of bounds reads from strides and shapes.
     int expanded_dim_;
+
+  private:
+    template<typename T>
+    void validate_can_take_span([[maybe_unused]] const size_t offset, [[maybe_unused]] const size_t count) const {
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE, ndim() == 1, "Cannot create a span from a multi-dimensional tensor"
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                elsize() == sizeof(T),
+                "Mismatched tensor byte size {} and span element type size {}",
+                elsize(),
+                sizeof(T)
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                size() == 0 || strides(ndim() - 1) == elsize(),
+                "Converting a fortran style NativeTensor to span is not possible"
+        );
+        // offset == size is allowed: it yields an empty span at one-past-the-end, and keeps nbytes() / sizeof(T) -
+        // offset from underflowing below.
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                offset <= nbytes() / sizeof(T),
+                "Span offset {} is out of bounds for tensor of size {}",
+                offset,
+                nbytes() / sizeof(T)
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                count == std::dynamic_extent || count <= nbytes() / sizeof(T) - offset,
+                "Span count {} is out of bounds for tensor of size {} with offset {}",
+                count,
+                nbytes() / sizeof(T),
+                offset
+        );
+    }
 };
 
 template<ssize_t>
@@ -308,7 +319,6 @@ struct TypedTensor : public NativeTensor {
     }
 
     const T* data() const { return static_cast<const T*>(NativeTensor::data()); }
-    std::span<const T> span() const { return std::span<const T>(data(), size()); }
 
   private:
     void check_ptr_within_bounds(const NativeTensor& tensor, size_t rows) {
--- a/cpp/arcticdb/entity/native_tensor.hpp
+++ b/cpp/arcticdb/entity/native_tensor.hpp
@@ -15,6 +15,7 @@
 
 // for std::accumulate
 #include <numeric>
+#include <span>
 
 #include <pybind11/numpy.h>
 
--- a/cpp/arcticdb/processing/clause.hpp
+++ b/cpp/arcticdb/processing/clause.hpp
@@ -8,8 +8,6 @@
 
 #pragma once
 
-#include <arcticdb/pipeline/write_options.hpp>
-
 #include <arcticdb/pipeline/frame_slice.hpp>
 #include <arcticdb/processing/expression_context.hpp>
 #include <arcticdb/processing/expression_node.hpp>
@@ -23,6 +21,7 @@
 #include <arcticdb/pipeline/pipeline_common.hpp>
 #include <arcticdb/version/merge_options.hpp>
 #include <arcticdb/util/string_utils.hpp>
+#include <arcticdb/pipeline/input_frame.hpp>
 
 #include <vector>
 #include <string>
@@ -31,6 +30,10 @@
 #include <ranges>
 
 namespace arcticdb {
+
+struct MergeUpdateInsertedRowsEntity {
+    size_t inserted_rows;
+};
 
 using ResampleOrigin = std::variant<std::string, timestamp>;
 
@@ -885,44 +888,70 @@ struct MergeUpdateClause {
     OutputSchema join_schemas(std::vector<OutputSchema>&&) const;
 
     [[nodiscard]] std::string to_string() const;
+    class MatchRecord {
+      public:
+        MatchRecord(std::span<ProcessingUnit> row_slices, size_t num_source_rows);
+        void add_match(size_t source_row, size_t target_row_slice, size_t target_row);
+        void add_match(size_t source_row, size_t target_row_slice, std::span<size_t> target_rows);
+        void filter_matching_rows(
+                std::string_view column_name, DataType source_type, DataType target_type, size_t source_offset,
+                std::span<const std::byte> source_data_bytes
+        );
+        void clone_source_match(size_t source_row_src, size_t source_row_dst, size_t row_slice);
+        void validate_rows_to_update(const MergeStrategy& strategy) const;
+        [[nodiscard]] size_t total_unmatched_source_rows() const;
+        [[nodiscard]] std::span<const std::vector<size_t>> matched_rows(size_t target_row_slice) const;
+        [[nodiscard]] bool is_source_row_matched(size_t source_row) const;
+
+      private:
+        /// For each row slice, for each source row, store all target rows that match it
+        std::vector<std::vector<std::vector<size_t>>> matched_target_rows_;
+        std::span<ProcessingUnit> row_slices_;
+        std::vector<size_t> source_row_matched_count_;
+        size_t total_matched_target_rows_count_ = 0;
+    };
 
   private:
-    bool update_and_insert(
-            const std::span<const NativeTensor> source_tensors, const StreamDescriptor& source_descriptor,
-            const ProcessingUnit& proc, const std::span<const std::vector<size_t>> rows_to_update
+    std::pair<std::vector<ProcessingUnit>, bool> update_and_insert(
+            const MatchRecord& match_record, const StreamDescriptor& target_descriptor,
+            std::vector<ProcessingUnit>&& row_slices
+    ) const;
+
+    std::pair<std::vector<ProcessingUnit>, bool> update(
+            const MatchRecord& match_record, std::vector<ProcessingUnit>&& row_slices
     ) const;
 
     /// Filter segments which will be affected by the merge. The complexity is O(m * log(n)) where n is the number
     /// of rows in the source data and m is the number of row slices in the library
     std::vector<std::vector<size_t>> structure_for_processing_log(std::vector<RangesAndKey>& ranges_and_keys);
 
-    /// @return Vector of size equal to the number of source data rows that are within the row slice being
-    /// processed. Each element is a vector of the rows from the target data that has the same index as the
-    /// corresponding source row
-    std::vector<std::vector<size_t>> filter_index_match(
-            const Column& target_index, const std::span<const timestamp> source_index, const ProcessingUnit& proc
-    ) const;
+    MatchRecord match(std::span<ProcessingUnit> row_slices) const;
 
-    std::vector<std::vector<size_t>> filter_on_additional_columns_match(
+    MatchRecord filter_on_additional_columns_match(
             const StreamDescriptor& source_descriptor, const StreamDescriptor& target_descriptor,
-            const std::span<const NativeTensor> source_tensors, ProcessingUnit& proc,
-            std::optional<std::vector<std::vector<size_t>>>&& index_match
+            std::span<ProcessingUnit> proc, std::optional<MatchRecord>&& match_record
     ) const;
 
-    std::vector<std::vector<size_t>> initialize_rows_to_update_for_rowrange_indexed_data(
-            ProcessingUnit& proc, const StreamDescriptor& source_descriptor
+    MatchRecord initialize_rows_to_update_for_row_range_indexed_data(
+            std::span<ProcessingUnit> row_slices, const StreamDescriptor& source_descriptor
     ) const;
 
     size_t field_index_for_matching_on_column(std::string_view name, const StreamDescriptor& descriptor) const;
 
     bool is_update_only() const;
 
-    /// For each timestamp range stores the first and last row in the source that overlaps with the row range. The
-    /// interval is closed in the start and open in the end: [start, end)
-    ankerl::unordered_dense::map<TimestampRange, std::pair<size_t, size_t>, folly::hasher<TimestampRange>>
-            source_start_end_for_row_range_;
+    /// For each processing group, identified by its row range, stores the first and last row in the source that
+    /// overlaps with it. The interval is closed in the start and open in the end: [start, end). The key is the row
+    /// range rather than the timestamp range because overlapping-window groups can share a timestamp range when
+    /// boundary index values repeat.
+    ankerl::unordered_dense::map<RowRange, std::pair<size_t, size_t>> source_start_end_for_row_range_;
+    std::pair<size_t, size_t> get_source_start_end(std::span<const ProcessingUnit> row_slice) const;
 
-    std::pair<size_t, size_t> get_source_start_end(const ProcessingUnit& proc) const;
+    [[nodiscard]] bool must_structure_by_time_slice() const;
+
+    std::span<const timestamp> get_source_index(std::span<const ProcessingUnit> row_slices) const;
+
+    std::span<const std::byte> get_source_data_bytes(size_t field_index, std::pair<size_t, size_t> range) const;
 };
 
 struct CompactDataClause {
--- a/cpp/arcticdb/processing/clause.hpp
+++ b/cpp/arcticdb/processing/clause.hpp
@@ -31,10 +31,6 @@
 
 namespace arcticdb {
 
-struct MergeUpdateInsertedRowsEntity {
-    size_t inserted_rows;
-};
-
 using ResampleOrigin = std::variant<std::string, timestamp>;
 
 using RangesAndKey = pipelines::RangesAndKey;
@@ -900,8 +896,9 @@ struct MergeUpdateClause {
         void clone_source_match(size_t source_row_src, size_t source_row_dst, size_t row_slice);
         void validate_rows_to_update(const MergeStrategy& strategy) const;
         [[nodiscard]] size_t total_unmatched_source_rows() const;
-        [[nodiscard]] std::span<const std::vector<size_t>> matched_rows(size_t target_row_slice) const;
+        [[nodiscard]] const std::vector<std::vector<size_t>>& matched_rows(size_t target_row_slice) const;
         [[nodiscard]] bool is_source_row_matched(size_t source_row) const;
+        [[nodiscard]] bool has_matched_target_rows() const;
 
       private:
         /// For each row slice, for each source row, store all target rows that match it
@@ -912,28 +909,31 @@ struct MergeUpdateClause {
     };
 
   private:
-    std::pair<std::vector<ProcessingUnit>, bool> update_and_insert(
+    std::vector<ProcessingUnit> update_and_insert(
             const MatchRecord& match_record, const StreamDescriptor& target_descriptor,
-            std::vector<ProcessingUnit>&& row_slices
+            std::vector<ProcessingUnit>&& row_slices, std::pair<size_t, size_t> source_start_end
     ) const;
 
-    std::pair<std::vector<ProcessingUnit>, bool> update(
-            const MatchRecord& match_record, std::vector<ProcessingUnit>&& row_slices
+    std::vector<ProcessingUnit> update(
+            const MatchRecord& match_record, std::vector<ProcessingUnit>&& row_slices,
+            std::pair<size_t, size_t> source_start_end
     ) const;
 
     /// Filter segments which will be affected by the merge. The complexity is O(m * log(n)) where n is the number
     /// of rows in the source data and m is the number of row slices in the library
     std::vector<std::vector<size_t>> structure_for_processing_log(std::vector<RangesAndKey>& ranges_and_keys);
 
-    MatchRecord match(std::span<ProcessingUnit> row_slices) const;
+    MatchRecord match(std::span<ProcessingUnit> row_slices, std::pair<size_t, size_t> source_start_end) const;
 
     MatchRecord filter_on_additional_columns_match(
             const StreamDescriptor& source_descriptor, const StreamDescriptor& target_descriptor,
-            std::span<ProcessingUnit> proc, std::optional<MatchRecord>&& match_record
+            std::span<ProcessingUnit> proc, std::optional<MatchRecord>&& match_record,
+            std::pair<size_t, size_t> source_start_end
     ) const;
 
     MatchRecord initialize_rows_to_update_for_row_range_indexed_data(
-            std::span<ProcessingUnit> row_slices, const StreamDescriptor& source_descriptor
+            std::span<ProcessingUnit> row_slices, const StreamDescriptor& source_descriptor,
+            std::pair<size_t, size_t> source_start_end
     ) const;
 
     size_t field_index_for_matching_on_column(std::string_view name, const StreamDescriptor& descriptor) const;
@@ -949,7 +949,7 @@ struct MergeUpdateClause {
 
     [[nodiscard]] bool must_structure_by_time_slice() const;
 
-    std::span<const timestamp> get_source_index(std::span<const ProcessingUnit> row_slices) const;
+    std::span<const timestamp> get_source_index(std::pair<size_t, size_t> source_start_end) const;
 
     std::span<const std::byte> get_source_data_bytes(size_t field_index, std::pair<size_t, size_t> range) const;
 };
--- a/cpp/arcticdb/processing/clause_merge_update.cpp
+++ b/cpp/arcticdb/processing/clause_merge_update.cpp
@@ -22,6 +22,16 @@
 
 namespace {
 using namespace arcticdb;
+
+template<util::type_descriptor_tag TDT>
+using SourceRawType =
+        std::conditional_t<is_sequence_type(TDT::data_type()), PyObject* const, typename TDT::DataTypeTag::raw_type>;
+
+struct TargetRange {
+    size_t start_row_in_first_row_slice{};
+    size_t end_row_in_last_row_slice{};
+};
+
 /// Remove all row slice entities and ranges and keys whose indexes do not appear in row_slices_to_keep
 /// @param offsets Must be structured by row slice with ranges and keys
 /// @param ranges_and_keys Must be structured by row slice with offsets
@@ -43,17 +53,22 @@ void filter_selected_ranges_and_keys_and_reindex_entities(
     size_t new_entity_id = 0;
     std::vector<RangesAndKey> new_ranges_and_keys;
     new_ranges_and_keys.reserve(ranges_and_keys.size());
+    ankerl::unordered_dense::map<size_t, size_t> offset_to_entity;
     for (std::span<size_t> row_slice : offsets) {
         for (size_t& entity_id : row_slice) {
-            new_ranges_and_keys.emplace_back(std::move(ranges_and_keys[entity_id]));
-            entity_id = new_entity_id++;
+            auto [it, inserted] = offset_to_entity.emplace(entity_id, new_entity_id);
+            if (inserted) {
+                new_ranges_and_keys.emplace_back(std::move(ranges_and_keys[entity_id]));
+                ++new_entity_id;
+            }
+
+            entity_id = it->second;
         }
     }
     ranges_and_keys = std::move(new_ranges_and_keys);
 }
 
-template<typename TDT>
-requires util::instantiation_of<TDT, TypeDescriptorTag>
+template<util::type_descriptor_tag TDT>
 position_t write_py_string_to_pool_or_throw(
         PyObject* const py_string_object, const size_t row_in_segment, const RowRange& row_range,
         std::optional<ScopedGILLock>& gil_lock, StringPool& new_string_pool, const std::string_view column_name
@@ -68,16 +83,15 @@ position_t write_py_string_to_pool_or_throw(
     );
 }
 
-template<typename TDT>
-requires util::instantiation_of<TDT, TypeDescriptorTag>
+template<util::type_descriptor_tag TDT>
 void rebuild_sequence_column_in_new_pool(
         Column& target_column, const StringPool& old_string_pool, StringPool& new_string_pool,
-        const util::BitSet* rows_to_add_in_new_pool = nullptr
+        const util::BitSet* target_rows_to_add_in_new_pool = nullptr
 ) {
-    if (rows_to_add_in_new_pool) {
+    if (target_rows_to_add_in_new_pool) {
         ColumnData target_column_data = target_column.data();
         auto accessor = random_accessor<TDT>(&target_column_data);
-        for (auto row = rows_to_add_in_new_pool->first(); row != rows_to_add_in_new_pool->end(); ++row) {
+        for (auto row = target_rows_to_add_in_new_pool->first(); row != target_rows_to_add_in_new_pool->end(); ++row) {
             if (is_a_string(accessor[*row])) {
                 const std::string_view string_value = old_string_pool.get_const_view(accessor[*row]);
                 const OffsetString& new_offset = new_string_pool.get(string_value);
@@ -95,86 +109,13 @@ void rebuild_sequence_column_in_new_pool(
     }
 }
 
-/// When merge update is performed on a string column the string pool is rebuild from scratch, thus we must iterate over
-/// all rows in the column. When the column is a part of a timeseries all target rows stored in rows_to_update will be
-/// ordered (i.e. sorted(flatten(rows_to_update)) = true). This means we can iterate over the target column row and
-/// at the same time forward iterate on the source rows. It's a nested loop but the complexity is O(n + m + k)
-/// where n is the total number of rows in the target, m are the rows in the source, and k are the total matches between
-/// source and target (i.e. sum(rows_to_update[i] for i in range(len(rows_to_update))))
-template<typename TDT>
-requires util::instantiation_of<TDT, TypeDescriptorTag>
-bool merge_update_string_column_timeseries(
-        Column& target_column, const std::span<const std::vector<size_t>> rows_to_update, const Field& target_field,
+template<util::type_descriptor_tag TDT>
+bool merge_update_string_column(
+        Column& target_column, const std::span<const std::vector<size_t>> rows_to_update, std::string_view column_name,
         const RowRange& row_range, const StringPool& target_string_pool,
         const std::span<PyObject* const> source_column_view, StringPool& new_string_pool
 ) {
-    bool is_column_changed = false;
-    if constexpr (is_fixed_string_type(TDT::data_type())) {
-        user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>(
-                "Fixed string sequences are not supported for merge update"
-        );
-    } else if constexpr (is_dynamic_string_type(TDT::data_type())) {
-        // GIL will be acquired if there is a string that is not pure ASCII/UTF-8
-        // In this case a PyObject will be allocated by convert::py_unicode_to_buffer
-        // If such a string is encountered in a column, then the GIL will be held until that whole column has
-        // been processed, on the assumption that if a column has one such string it will probably have many.
-        std::optional<ScopedGILLock> gil_lock;
-        // TODO: Create new column in case of insertion
-        auto source_row_it = rows_to_update.begin();
-        auto next_target_row_to_update = source_row_it->begin();
-        // TODO: Handle sparse target columns Monday 10756321294
-        arcticdb::for_each_enumerated<TDT>(target_column, [&](auto row) {
-            source_row_it = std::find_if(
-                    source_row_it,
-                    rows_to_update.end(),
-                    [&, new_row = false](const std::vector<size_t>& vec) mutable {
-                        // TODO: Handle inserts. Empty vec means insert.
-                        next_target_row_to_update = std::find_if(
-                                new_row ? vec.begin() : next_target_row_to_update,
-                                vec.end(),
-                                [&](const int64_t target_row_to_update) { return target_row_to_update >= row.idx(); }
-                        );
-                        if (next_target_row_to_update == vec.end()) {
-                            new_row = true;
-                        }
-                        return next_target_row_to_update != vec.end();
-                    }
-            );
-            const bool current_target_row_is_matched = source_row_it != rows_to_update.end() &&
-                                                       next_target_row_to_update != source_row_it->end() &&
-                                                       static_cast<int64_t>(*next_target_row_to_update) == row.idx();
-            if (current_target_row_is_matched) {
-                is_column_changed = true;
-                const auto py_string_object = source_column_view[source_row_it - rows_to_update.begin()];
-                row.value() = write_py_string_to_pool_or_throw<TDT>(
-                        py_string_object, row.idx(), row_range, gil_lock, new_string_pool, target_field.name()
-                );
-            } else {
-                if (is_a_string(row.value())) {
-                    const std::string_view string_in_target = target_string_pool.get_const_view(row.value());
-                    const OffsetString offset_in_new_pool = new_string_pool.get(string_in_target);
-                    row.value() = offset_in_new_pool.offset();
-                }
-            }
-        });
-    }
-    return is_column_changed;
-}
-
-/// When merge update is performed on a string column the string pool is rebuild from scratch, thus we must iterate over
-/// all rows in the column. When the column is *not* part of a timeseries the rows in rows_to_update are in random
-/// order. To avoid having quadratic complexity (i.e. O(n * k) where n are the rows in the target and k are all matches
-/// i.e. sum(rows_to_update[i] for i in range(len(rows_to_update)))) we invert the loops and iterate over all
-/// rows_to_update perform the update and then add additional loop over all rows in target to rebuild the string pool.
-/// Asymptotically this will yield the same complexity as the timeseries case but practically is a bit slower.
-template<typename TDT>
-requires util::instantiation_of<TDT, TypeDescriptorTag>
-bool merge_update_string_column_rowrange(
-        Column& target_column, const std::span<const std::vector<size_t>> rows_to_update, const Field& target_field,
-        const RowRange& row_range, const StringPool& target_string_pool,
-        const std::span<PyObject* const> source_column_view, StringPool& new_string_pool
-) {
-    bool is_column_changed = false;
+    bool column_is_changed = false;
     if constexpr (is_fixed_string_type(TDT::data_type())) {
         user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>(
                 "Fixed string sequences are not supported for merge update"
@@ -192,9 +133,9 @@ bool merge_update_string_column_rowrange(
             for (size_t target_row : rows_to_update[source_row]) {
                 const auto py_string_object = source_column_view[source_row];
                 target_column_accessor[target_row] = write_py_string_to_pool_or_throw<TDT>(
-                        py_string_object, target_row, row_range, gil_lock, new_string_pool, target_field.name()
+                        py_string_object, target_row, row_range, gil_lock, new_string_pool, column_name
                 );
-                is_column_changed = true;
+                column_is_changed = true;
                 target_rows_not_matched_by_source.set(target_row);
             }
         }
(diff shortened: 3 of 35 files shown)


python/tests/hypothesis/arcticdb/test_merge_update.py
```diff
@@ -184,3 +188,163 @@ def test_rowrange_merge_update(s3_version_store_v1, merge_args):
...
+            target.iloc[updated_positions_in_target, target.columns.get_loc(col)] = replacement.iloc[picks][
+                col
+            ].to_numpy()
+            # Restore the dtype in case the positional assignment upcast it (object columns keep None/NaN as-is).
+            if merged_dataframe[col].dtype != object:
+                target[col] = target[col].astype(merged_dataframe[col].dtype)
+
+    # expected = merged reordered so inserted rows follow the target rows at the same timestamp; updates stay in place.
+    order = sorted(range(len(merged_dataframe)), key=lambda i: (merged_dataframe.index[i], bool(is_inserted[i])))
+    expected = merged_dataframe.iloc[order]
```
IvoDD: There is some duplicated code between `merge_update_insert_dataframes` and `merge_insert_dataframes`.

What do you think about using `merge_insert_dataframes` in `merge_update_insert_dataframes`?
I.e. something like:
```
target, source, value = draw(merge_insert_dataframes)
# Choose updatable positions in target. Requires recalculating the target_keys
target_keys = pd.Series(_row_keys(target, on), index=merged_dataframe.index)
target_key_counts = Counter(target_keys)
# Populate is_updated as in this code (will need some minor tweaks)

# Add the updatable positions in target to source
source = pd.concat([source, target[is_updated]])
source.sort_index(inplace=True, kind="stable") # stable is needed so that order of multiple insertions with same index is preserved

# Randomize the target non on columns for is_updated. Same code as above

return target, source, value
```

python/tests/hypothesis/arcticdb/test_merge_update.py
```diff
@@ -184,3 +188,163 @@ def test_rowrange_merge_update(s3_version_store_v1, merge_args):
...
+    ac = s3_storage.create_arctic()
+    name = "test_datetime_insert"
+    if name in ac.list_libraries():
+        ac.delete_library(name)
+    lib = ac.create_library(
+        name, library_options=LibraryOptions(rows_per_segment=rows_per_segment, columns_per_segment=cols_per_segment)
+    )
+    lib.write("symbol", target)
+    lib.merge_experimental(
+        "symbol", source, strategy=MergeStrategy(matched="do_nothing", not_matched_by_target="insert"), on=on
```
IvoDD: Can also parametrize that on `matched = do_nothing | update` both should give the same result by construction.

python/arcticdb/version_store/library.py
```diff
@@ -3438,8 +3438,6 @@ def merge_experimental(
         source : pandas.DataFrame or pandas.Series
             The new data to merge. In the case of timeseries, the index must be sorted.
         strategy : Optional[MergeStrategy], default=MergeStrategy(matched="update", not_matched_by_target="insert")
-            !!! warning
-                Only `MergeStrategy(matched="update", not_matched_by_target="do_nothing")` is implemented
```
IvoDD, line 3454: Maybe worth still noting that rowrange insert is not supported?

python/arcticdb/util/test.py
```diff
@@ -156,10 +156,8 @@ def dataframe_dump_to_log(label_for_df, df: pd.DataFrame):
     we could use to reproduce something or further analyze failures caused by
     a problems in test code or arctic
     """
-    print("-" * 80)
-    if SHORTER_LOGS:
-        print("No full dataframes dumps when shorter logs is activated")
-    else:
+    if not SHORTER_LOGS:
```
IvoDD: Nit: I prefer avoiding extra identation with:
```
if SHORTER_LOGS:
    return
# Rest of code
```

cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -141,6 +141,55 @@ struct NativeTensor {
...
+                nbytes() / sizeof(T)
+        );
+        return std::span<const T>(
+                static_cast<const T*>(ptr) + offset,
+                count == std::dynamic_extent ? nbytes() / sizeof(T) - offset : count
+        );
+    }
+
+    template<typename T>
+    std::span<T> span(size_t offset = 0, size_t count = std::dynamic_extent) {
```
IvoDD: Nit: To avoid debug check duplication we can make one span use the other with const_cast

cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -141,6 +141,55 @@ struct NativeTensor {
...
+                elsize(),
+                sizeof(T)
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                count == std::dynamic_extent || count <= nbytes() / sizeof(T) - offset,
+                "Span count {} is out of bounds for tensor of size {}",
+                count,
+                nbytes() / sizeof(T)
+        );
```
IvoDD: Maybe also a debug check that offset itself is not out of bounds. Otherwise `nbytes() / sizeof(T) - offset` could underflow

cpp/arcticdb/version/version_core.hpp
```diff
@@ -194,9 +194,10 @@ folly::Future<VersionedItem> read_modify_write_impl(
 );
 
 folly::Future<VersionedItem> merge_update_impl(
-        const std::shared_ptr<Store>& store, const VersionIdentifier& version_info, const ReadOptions& read_options,
+        const std::shared_ptr<Store>& store, const UpdateInfo& version_info, const ReadOptions& read_options,
```
IvoDD: Nit: Rename arg to `update_info`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2979,18 +3022,18 @@ folly::Future<ReadVersionOutput> read_frame_for_version(
 
 folly::Future<std::vector<SliceAndKey>> read_modify_write_data_keys(
         const std::shared_ptr<Store>& store, std::shared_ptr<ReadQuery> read_query, const ReadOptions& read_options,
-        const IndexPartialKey& target_partial_index_key, const std::shared_ptr<PipelineContext>& pipeline_context
+        const IndexPartialKey& target_partial_index_key, const std::shared_ptr<PipelineContext>& pipeline_context,
+        std::shared_ptr<ComponentManager> component_manager, std::shared_ptr<DeDupMap> de_dup_map
```
IvoDD: If we're now correctly passing the dedup map should we add a test where a no-op update doesn't create new data keys?

E.g. if we have:
```
target:
index value
1     1
2     2
3     3

source:
index value
1     1
3     3
```

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3079,6 +3153,11 @@ folly::Future<VersionedItem> merge_update_impl(
     user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
             !write_options.dynamic_schema, "Cannot merge update with dynamic schema"
     );
+    internal::check<ErrorCode::E_ASSERTION_FAILURE>(
```
IvoDD: Nit: `E_NOT_IMPLEMENTED` instead?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -920,4 +1338,153 @@ size_t MergeUpdateClause::field_index_for_matching_on_column(std::string_view na
     return std::distance(descriptor.fields().begin(), field_it);
 }
 
+bool MergeUpdateClause::must_structure_by_time_slice() const {
+    return source_->has_index() && !on_.empty() && strategy_.insert();
```
IvoDD, line 1323: I feel like branching off here on insert vs update creates quite some extra complexity.

I appreciate that structuring by row slice is more efficient (because row slices with the same index value will be processed in parallel) but this feels a bit like premature optimization.

Don't have a strong opinon since we'll anyway need to structure by row slice for rowrange (however that's mostly a separate codepath so unifying that wouldn't give us much)

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -203,50 +144,35 @@ bool merge_update_string_column_rowrange(
...
-                    target_field,
-                    row_range,
-                    target_string_pool,
-                    source_column_view,
-                    new_string_pool
-            );
-        }
     }
+
+    return merge_update_string_column<TDT>(
```
IvoDD: Nice, this is a lot simpler!

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
 /// This means that the ordering of the columns in MergeUpdateClause::on_ matters and it would be more efficient to
 /// start with the columns that have a lesser chance of matching.
```
IvoDD: I think the above comment needs some updates to mention the `MatchRecord` struct instead of the vector of vectors.

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -322,17 +248,423 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+/// For each row of source that falls in the row slice in proc find all rows whose index matches the source index
+/// value. The matching rows will be sorted in increasing order. Since both source and target are timestamp indexed
+/// and ordered, only forward iteration on both source and target is needed and binary search can be used to check
+/// if a source index value exists in the target index. At the end some vectors in the output can be empty which
+/// means that that particular row in source did not match anything in the target. It is allowed for one row in
+/// target to be matched by multiple rows in source only if MergeUpdateClause::on_ is not empty. If
+/// MergeUpdateClause::on_ is not empty, there will be further filtering that might remove some matches. Otherwise,
+/// one row will be updated multiple times, which is not allowed.
+///
+/// Complexity: $$O(m * log_2(n) + n)$$ where: m is the count of source rows in the bounds of the segment, n is the
```
IvoDD: exponential binary search actually gives us a linear time if `m==n`. Actual complexity is `O(m*log(n/m) + n)` because the average step exponential search would need to take is `n/m` and exponential search is `log(distance_to_target)` not `log(length_of_range)`.

To be fully precise the complexity is actually `sum(log(distance_i) for distance_i in distances_between_matches)`. But due to Jensen inequality that is actually `<= m*log(average_distance)` because log is a concave function

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -322,17 +248,423 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+        return result;
+    }
+    for (size_t i = 0; i < row_slices.size(); ++i) {
+        const ProcessingUnit& proc = row_slices[i];
+        const Column& target_index = proc.segments_->front()->column(0);
+        ColumnData target_index_column_data = target_index.data();
+        auto last_target_row_it = exponential_upper_bound<IndexType, IteratorType::ENUMERATED>(
+                target_index_column_data, source_index.back()
+        );
+        size_t source_row{};
```
IvoDD: We can be smarter than starting from `source_row=0` every time.

E.g. if we have row slices with index values:
```
target:
Slice 1: [1, 2, 3, 3]
Slice 2: [3, 3, 3]
Slice 3: [3, 4, 5]

source:
[1, 2, 3, 3, 4, 4, 5]
```

Slice 3 doesn't need to start from the `source_row=0`.
It can safely start from `source_row=2`

This can make the current `O(num_row_slices * source_len)` to a `O(source_len)`

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -920,4 +1338,153 @@ size_t MergeUpdateClause::field_index_for_matching_on_column(std::string_view na
...
+
+void MergeUpdateClause::MatchRecord::add_match(size_t source_row, size_t target_row_slice, size_t target_row) {
+    ++source_row_matched_count_[source_row];
+    matched_target_rows_[target_row_slice][source_row].push_back(target_row);
+    ++total_matched_target_rows_count_;
+}
+void MergeUpdateClause::MatchRecord::add_match(
+        size_t source_row, size_t target_row_slice, std::span<size_t> target_rows
+) {
+    ++source_row_matched_count_[source_row];
```
IvoDD: Shouldn't we `source_row_matched_count_[source_row] += target_rows.size()`?

I think this leaves the opportunity for some nasty underflow bugs.

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -920,4 +1338,153 @@ size_t MergeUpdateClause::field_index_for_matching_on_column(std::string_view na
...
+            // TODO: Relax for dynamic schema
+            if constexpr (std::same_as<std::decay_t<SourceDataTypeTag>, std::decay_t<TargetDataTypeTag>>) {
+                for (size_t row_slice_idx = 0; row_slice_idx < matched_target_rows_.size(); ++row_slice_idx) {
+                    ProcessingUnit& row_slice = row_slices_[row_slice_idx];
+                    const ColumnWithStrings target_column =
+                            std::get<ColumnWithStrings>(row_slice.get(ColumnName{column_name}));
+                    ColumnData col_data = target_column.column_->data();
+                    auto target_column_accessor = random_accessor<TargetTDT>(&col_data);
+                    std::vector<std::vector<size_t>>& matched_rows = matched_target_rows_[row_slice_idx];
+                    for (size_t source_row_idx = 0; source_row_idx < matched_rows.size(); ++source_row_idx) {
```
IvoDD, line 1382: Iterating over all source rows for a given row slice seems inefficient. We could easily preserve a `target_slice -> source_range` mapping which we can use here.

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -920,4 +1338,153 @@ size_t MergeUpdateClause::field_index_for_matching_on_column(std::string_view na
...
+        });
+    });
+}
+
+void MergeUpdateClause::MatchRecord::clone_source_match(
+        size_t source_row_src, size_t source_row_dst, size_t row_slice
+) {
+    source_row_matched_count_[source_row_dst] = source_row_matched_count_[source_row_src];
+    matched_target_rows_[row_slice][source_row_dst] = matched_target_rows_[row_slice][source_row_src];
+    total_matched_target_rows_count_ += source_row_matched_count_[source_row_dst];
```
IvoDD, line 1433: Either assert that source_row_dst is empty upfront, or do `total_matched_target_rows_count_ -= source_row_matched_count_[source_row_dst]` first.

cpp/arcticdb/processing/clause_merge_update.cpp
IvoDD: The main difference this results in is the branching between `update` and `update_and_insert` methods. We could probably live with only the `update_and_insert` one.

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -322,17 +248,423 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+            advance_output();
+            advance_target();
+        }
+
+        if (target_index_is_exhausted()) {
+            break;
+        }
+
+        // Target index values are equal to the source index values. Since the index is matching, it's possible to
+        // perform both an update and an insert. First, copy all target values to the output buffer and then append
```
IvoDD: I guess it would be somewhat easier to follow if we do one pass to apply the updates inplace on the target and then do another pass to do the actual `merge` where it will be just a simple loop which adds either from target or from unmatched source.

This way we could also reuse `update_column` potentially.

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -322,17 +248,423 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+                        const size_t target_offset = i == 0 ? target.range.start_row_in_first_row_slice : 0;
+                        // matched_rows holds indexes into the target segment. But the output segment can start at
+                        // an offest in the target in case the work is shared between two processing units. Only
+                        // the first and last row slices of a time slice can be requested by more than one
+                        // processing unit
+                        const size_t output_row_to_update = target_row_to_update_in_slice +
+                                                            row_slice_start_in_output[i] + inserted_rows[i] -
+                                                            target_offset;
+                        if constexpr (is_sequence_type(TargetColumnTypeDescriptorTag::data_type())) {
+                            set_string_from_source_at(source_row_idx, output_row_to_update);
```
IvoDD: This backfilling is somewhat complicated:
- Makes the flow in this loop somewhat unintuitive, we have to take a stap back.
- We end up adding two strings to the string pool. Once when we `set_string_from_source` with the target's value. Once here with the source's value. So resulting stringpool would have more fields than necassary.

What do you think of instead reordering `source` upfront so that inserted values are at the end of equal range. That should be quite cheap `O(n)`. For example from source:
```
index value operation
1     1     update
2     2     update
2     3     insert
2     4     update
2     5     insert
2     6     update
3     7     insert
3     8     update
```
we go to:
```
index value operation
1     1     update
2     2     update
2     4     update
2     6     update
2     3     insert
2     5     insert
3     8     update
3     7     insert
```

python/tests/unit/arcticdb/version_store/test_merge_update.py
IvoDD, line 1: Would be nice to add some tests which assert the data_key structure after update_and_insert.

I.e. how the row ranges change. Also show that column slicing is preserved.

cpp/arcticdb/processing/clause_utils.cpp
```diff
@@ -46,6 +46,59 @@ std::vector<std::vector<EntityId>> structure_by_row_slice(
...
+                }
+                previous_time_range = first_overlap;
+            } else {
+                previous_time_range = current_time_range;
+            }
+            overlapping_ranges = 0;
+        }
+        overlapping_ranges += current_time_range.second != previous_time_range.second;
+        res.back().emplace_back(idx);
+    }
```
alexowens90, line 91: I find this algorithm a little hard to follow. Is it equivalent to:
```
for (size_t idx = 0; idx < ranges.size();) {
    res.emplace_back();
    res.back().emplace_back(idx);
    auto last_ts = ranges[idx].key_.time_range().second - 1;
    while (++idx < ranges.size()) {
        if (ranges[idx].key_.time_range().first == last_ts) {
            res.back().emplace_back(idx);
        } else {
            break;
        }
    }
}
```

cpp/arcticdb/processing/expression_node.hpp
```diff
@@ -49,7 +49,6 @@ struct ColumnWithStrings {
     std::shared_ptr<Column> column_;
     const std::shared_ptr<StringPool> string_pool_;
     std::string column_name_;
-
```
alexowens90, line 36: nit: revert

cpp/arcticdb/processing/component_manager.hpp
```diff
@@ -128,6 +128,18 @@ class ComponentManager {
         return get_entities_impl<Args...>(ids, false);
     }
 
+    template<typename ProcessComponents>
+    void process_entities(ProcessComponents&& process_fn) {
```
alexowens90, line 143: This isn't how the component manager should be used. IIUC, this will apply `process_fn` to all components of a certain type, rather than just those with specific entity IDs? Right now this will work because you have separate component manager instances for each call to `merge`, but this isn't the only usage pattern (e.g. `concat` has one component manager across all of the symbols).

cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -141,6 +141,55 @@ struct NativeTensor {
...
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE, ndim() == 1, "Cannot create a span from a multi-dimensional tensor"
+        );
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                elsize() == sizeof(T),
+                "Mismatched tensor byte size {} and span element type size {}",
+                elsize(),
+                sizeof(T)
+        );
```
alexowens90: Does this always imply Fortran-ordered data? If so, we should include that in the error message

cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -141,6 +141,55 @@ struct NativeTensor {
...
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                count == std::dynamic_extent || count <= nbytes() / sizeof(T) - offset,
+                "Span count {} is out of bounds for tensor of size {}",
+                count,
+                nbytes() / sizeof(T)
+        );
```
alexowens90: I'd include the offset in this, otherwise you could get a message that is not clear why it failed like
```
"Span count 10 is out of bounds for tensor of size 20"
```

cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -141,6 +141,55 @@ struct NativeTensor {
         return (&(reinterpret_cast<const T*>(ptr)[signed_pos]));
     }
 
+    template<typename T>
+    std::span<const T> span(size_t offset = 0, size_t count = std::dynamic_extent) const {
```
alexowens90: Took me a second to work out what was going on, worth adding a comment that `count == std::dynamic_extent` implies span the whole remainder of the tensor from `offset`

cpp/arcticdb/entity/native_tensor.hpp
```diff
@@ -259,6 +308,7 @@ struct TypedTensor : public NativeTensor {
     }
 
     const T* data() const { return static_cast<const T*>(NativeTensor::data()); }
+    std::span<const T> span() const { return std::span<const T>(data(), size()); }
```
alexowens90: This doesn't have the same dimension checking etc as the other new methods
```suggestion
    std::span<const T> span() const { return span(); }
```

cpp/arcticdb/processing/test/test_split_by_time_slice.cpp
```diff
@@ -0,0 +1,1188 @@
...
+                    ColRange{1, 2},
+                    AtomKeyBuilder().start_index(40).end_index(50).build<KeyType::TABLE_DATA>("t")
+            },
+            RangesAndKey{
+                    RowRange{15, 20},
+                    ColRange{1, 2},
+                    AtomKeyBuilder().start_index(60).end_index(70).build<KeyType::TABLE_DATA>("t")
+            },
+    }};
+    shuffle(input);
```
alexowens90: I'm not a fan of this kind of randomness in tests, can we just pick a deterministic (out of order) ordering of the input?

cpp/arcticdb/processing/test/test_split_by_time_slice.cpp
```diff
@@ -0,0 +1,1188 @@
...
+
+class TestSplitByTimeSlice : public testing::Test {
+  protected:
+    TestSplitByTimeSlice() : g(rd()) {}
+    void shuffle(std::span<RangesAndKey> input) { std::ranges::shuffle(input, g); }
+    std::random_device rd;
+    std::mt19937 g;
+};
+
+TEST_F(TestSplitByTimeSlice, DisjointTimeRanges_OneColPerRowSlice) {
```
alexowens90: nit
```suggestion
TEST_F(TestStructureByTimeSlice, DisjointTimeRanges_OneColPerRowSlice) {
```

cpp/arcticdb/processing/test/test_split_by_time_slice.cpp
```diff
@@ -0,0 +1,1188 @@
...
+                    AtomKeyBuilder().start_index(19).end_index(25).build<KeyType::TABLE_DATA>("t")
+            },
+            RangesAndKey{
+                    RowRange{15, 20},
+                    ColRange{1, 2},
+                    AtomKeyBuilder().start_index(30).end_index(50).build<KeyType::TABLE_DATA>("t")
+            },
+    }};
+    shuffle(input);
+    const std::array<std::vector<size_t>, 3> expected{{{0}, {1, 2}, {3}}};
```
alexowens90: Ah I've misunderstood the structuring algo slightly. I thought this would be:
```
    const std::array<std::vector<size_t>, 3> expected{{{0}, {1, 2}, {2}, {3}}};
```
In resampling, a call to `process` being responsible for the first bucket in a row-slice doesn't automatically mean it also processes all of the others

cpp/arcticdb/util/buffer.hpp
```diff
@@ -220,6 +220,8 @@ struct Buffer : public BaseBuffer<Buffer, true> {
 
     [[nodiscard]] size_t available() const { return capacity_ >= preamble_bytes_ ? capacity_ - preamble_bytes_ : 0; }
 
+    void clear() { deallocate(); }
```
alexowens90, line 223: Why not just call `deallocate()` directly?

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -2542,13 +2542,14 @@ VersionedItem LocalVersionedEngine::merge_internal(
...
-                                               VersionedItem{*update_info.previous_index_key_},
+                                               update_info,
                                                read_options,
                                                write_options,
                                                IndexPartialKey{stream_id, update_info.next_version_id_},
                                                std::move(on),
                                                strategy,
-                                               std::move(source)
+                                               std::move(source),
+                                               get_de_dup_map(stream_id, update_info.previous_index_key_, write_options)
```
alexowens90: This will conflict with my `LocalVersionedEngine` refactor (the second argument is now an `UpdateInfo`)
At a higher level though, do we need the dedup map with `merge`? Of the other modification methods, it is only used with `[batch_]write`, as with append and update we have inherently already loaded the previous version's index key

cpp/arcticdb/version/version_core.cpp
alexowens90: These new arguments could be defaulted rather than forcing all callers to add 2 `make_shared` calls

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3054,20 +3105,43 @@ folly::Future<VersionedItem> read_modify_write_impl(
...
+    VersionIdentifier resolved = VersionedItem{*update_info.previous_index_key_};
     if (auto* vi = std::get_if<VersionedItem>(&resolved)) {
         resolved = std::make_shared<IndexInformation>(read_index_key_without_column_stats(store, vi->key_));
     }
     std::shared_ptr<PipelineContext> pipeline_context =
             setup_pipeline_context(store, std::move(resolved), *read_query, read_options);
+    // The target is empty.
+    if (pipeline_context->rows_ == 0) {
+        if (strategy.insert()) {
+            return write_dataframe_impl(store, update_info.next_version_id_, source, write_options, de_dup_map);
```
alexowens90: This will also conflict with my refactor

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3054,20 +3105,43 @@ folly::Future<VersionedItem> read_modify_write_impl(
...
+        } else if (strategy.update_only()) {
+            const TimeseriesDescriptor tsd = make_timeseries_descriptor(
+                    0,
+                    pipeline_context->descriptor(),
+                    pipeline_context->release_normalization(),
+                    pipeline_context->release_opt_user_defined_metadata(),
```
alexowens90: This should come from `source`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3089,25 +3168,32 @@ folly::Future<VersionedItem> merge_update_impl(
...
-                        ++new_slice;
-                    } else {
-                        merged_ranges_and_keys.push_back(std::move(slice));
-                    }
-                }
+                ankerl::unordered_dense::map<RowRange, size_t> inserted_rows_per_row_range;
+                component_manager->process_entities([&](const MergeUpdateInsertedRowsEntity& inserted_rows,
+                                                        const std::shared_ptr<RowRange>& row_range) {
+                    inserted_rows_per_row_range.emplace(*row_range, inserted_rows.inserted_rows);
+                });
```
alexowens90, line 3187: Ah OK I see why you've implemented it this way now. I think it would be cleaner to change `read_modify_write_data_keys` to return futures on entity ids. Most callers will then just use
```
auto [slice_and_keys] = component_manager_->get_entities<SliceAndKey>(entity_ids)
```
and continue as they do now. But here we can do:
```
`auto [slice_and_keys, inserted_rows, row_ranges] = component_manager_->get_entities<SliceAndKey, MergeUpdateInsertedRowsEntity, std::shared_ptr<RowRange>>(entity_ids)`
```
(and maybe not need `std::shared_ptr<RowRange>` as I guess it's the same as the one in the `SliceAndKey`?)

cpp/arcticdb/version/version_core.cpp
```diff
@@ -415,6 +415,49 @@ bool is_fake_index_name(const arcticc::pb2::descriptors_pb2::NormalizationMetada
...
+        ) {
+            if (new_slice_and_key_it == new_slices.end() ||
+                old_slice_and_key_it->slice() < new_slice_and_key_it->slice()) {
+                old_slice_and_key_it->slice().row_range.first += total_inserted_rows;
+                old_slice_and_key_it->slice().row_range.second += total_inserted_rows;
+                merged_ranges_and_keys.emplace_back(std::move(*old_slice_and_key_it));
+                ++old_slice_and_key_it;
+            } else {
+                const RowRange new_row_range = new_slice_and_key_it->slice().row_range;
+                const size_t inserted_rows = inserted_rows_per_row_range[new_row_range];
```
alexowens90: We should either use `.at` or check the the key is in the map, otherwise will get some very weird behaviour f this access causes an insertion

cpp/arcticdb/processing/clause.hpp
```diff
@@ -32,6 +31,10 @@
 
 namespace arcticdb {
 
+struct MergeUpdateInsertedRowsEntity {
```
alexowens90: If `size_t` isn't taken as a component yet, I would just add another `using` statement [here](https://github.com/man-group/ArcticDB/blob/b3189eefe683ed80b4c458ebed381ef1820993cf/cpp/arcticdb/processing/component_manager.hpp#L25)
I'd consider putting it `component_manager.hpp` anyway, as it is clearly a construct only required because of the component manager class.
I _think_ ENTT supports multiple entities of the same type representing semantically different things, but I don't recall how it works

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -389,26 +719,83 @@ std::vector<std::vector<size_t>> MergeUpdateClause::structure_for_processing_log
         ranges_and_keys.clear();
         return {};
     }
-    std::span index(static_cast<const IndexType::DataTypeTag::raw_type*>(index_tensor.data()), source_->num_rows);
+
+    std::span source_index(index_tensor.data(), source_->num_rows);
```
alexowens90: Weren't some utility methods added to `NativeTensor` to create spans from them?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -389,26 +719,83 @@ std::vector<std::vector<size_t>> MergeUpdateClause::structure_for_processing_log
...
     for (size_t row_slice_idx = 0; row_slice_idx < offsets.size(); ++row_slice_idx) {
```
alexowens90: `row_slice_idx` is a bit misleading when structured by time slice

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -389,26 +719,83 @@ std::vector<std::vector<size_t>> MergeUpdateClause::structure_for_processing_log
...
-        }
-        const auto source_range_start = std::ranges::lower_bound(index, time_range.first);
-        if (source_range_start == index.end()) {
+        const std::vector<std::vector<size_t>>& row_slice_offsets =
+                must_structure_by_time_slice() ? ::split_by_row_slice(ranges_and_keys, offsets[row_slice_idx])
+                                               : std::vector<std::vector<size_t>>{offsets[row_slice_idx]};
+        const TimestampRange time_range{
+                ranges_and_keys[row_slice_offsets.front().front()].key_.time_range().first,
+                ranges_and_keys[row_slice_offsets.back().back()].key_.time_range().second
+        };
```
alexowens90: How is this different from:

```suggestion
const TimestampRange time_range{ranges_and_keys.at(offsets.at(idx).front()).key_.time_range().first, ranges_and_keys.at(offsets.at(idx).back()).key_.time_range().second};
```
?

cpp/arcticdb/processing/clause_merge_update.cpp
alexowens90: I think this mega loop could benefit from being broken up into some smaller functions

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -389,26 +719,83 @@ std::vector<std::vector<size_t>> MergeUpdateClause::structure_for_processing_log
...
+                    const timestamp next_row_slice_start =
+                            ranges_and_keys[offsets[row_slice_idx + 1].front()].key_.time_range().first;
+                    const timestamp extended_segment_end = std::max(time_range.second, next_row_slice_start) - 1;
+                    source_range_end =
+                            ranges::upper_bound(source_range_start, source_index.end(), extended_segment_end);
+                }
+            } else {
+                source_range_end = ranges::upper_bound(source_range_start, source_index.end(), time_range.second - 1);
+            }
         }
```
alexowens90: For breaking this loop up into smaller functions, this whole block just calculates `source_range_end`

cpp/arcticdb/processing/clause_merge_update.cpp
alexowens90, line 42: Again, `row_slices_to_keep` is misleading

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -22,6 +22,16 @@
...
+        std::conditional_t<is_sequence_type(TDT::data_type()), PyObject* const, typename TDT::DataTypeTag::raw_type>;
+
+struct TargetRange {
+    size_t start_row_in_first_row_slice{};
+    size_t end_row_in_last_row_slice{};
+};
+
 /// Remove all row slice entities and ranges and keys whose indexes do not appear in row_slices_to_keep
 /// @param offsets Must be structured by row slice with ranges and keys
 /// @param ranges_and_keys Must be structured by row slice with offsets
```
alexowens90: These comments are also out of date

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
 
-std::pair<size_t, size_t> MergeUpdateClause::get_source_start_end(const ProcessingUnit& proc) const {
+std::pair<size_t, size_t> MergeUpdateClause::get_source_start_end(std::span<const ProcessingUnit> row_slices) const {
     if (source_->has_index()) {
-        const std::span<const std::shared_ptr<AtomKey>> atom_keys = *proc.atom_keys_;
-        const auto source_row_range_it = source_start_end_for_row_range_.find(atom_keys.front()->time_range());
-        internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+        const pipelines::RowRange row_range = get_row_range(row_slices);
+        const auto source_row_range_it = source_start_end_for_row_range_.find(row_range);
+        ARCTICDB_DEBUG_CHECK(
```
alexowens90: No need to compile this out of release builds

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
-                "Missing mapping between AtomKey timerange {} {} and source row start/end",
-                atom_keys.front()->time_range().first,
-                atom_keys.front()->time_range().second
+                "Source row range for row range [{}, {}] not found",
+                row_range.first,
+                row_range.second
         );
-        return source_row_range_it->second;
+        const std::pair<size_t, size_t> source_range = source_row_range_it->second;
+        if (!row_slices.empty() && row_slices.front().entity_fetch_count_) {
```
alexowens90: When would `row_slices` be empty? And I believe `row_slices.front().entity_fetch_count_.has_value()` is true by construciton?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
+            // An index value is spanning more than one row slices.
```
alexowens90: Isn't this only true if `row_slices.front().entity_fetch_count_ > 1` or `row_slices.back().entity_fetch_count_ > 1`?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
+                    source_->get_tensor(0).span<timestamp>(result.first, result.second - result.first);
+            if (row_slices.front().entity_fetch_count_->front() > 1) {
+                // This is the last row slice containing the index value that spans multiple row slices. It appears at
+                // the beginning of the row slices, which means that the other processing unit will handle all rows of
+                // the segment containing the index value; this processing unit must handle only the rows that do not
+                // contain index value. Example:
+                //     0         1         2         3
+                // [a, b, c] [c, c, c] [c, d, e] [e, f, g]
+                // When the source contains value "c," the structure for processing will be [0, 1, 2] [2, 3]. This case
+                // handles processing [2, 3].
```
alexowens90: nit: This structure is wrong since the latest commit

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
+            });
         });
-    });
+    }
+
     return result;
 }
 
-std::pair<size_t, size_t> MergeUpdateClause::get_source_start_end(const ProcessingUnit& proc) const {
+std::pair<size_t, size_t> MergeUpdateClause::get_source_start_end(std::span<const ProcessingUnit> row_slices) const {
```
alexowens90, line 925: This function is called quite a few times, and presumably always with the same `row_slices` for each call to `process`? But it is also a pure function, so we shouldn't call it repeatedly with the same inputs given that it is doing some binary searching

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -322,17 +248,423 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+            const timestamp source_ts = source_index[source_row];
+            auto target_match_it = exponential_lower_bound(target_row_it, last_target_row_it, source_ts);
+            if (target_match_it == last_target_row_it) {
+                break;
+            }
+            target_row_it = target_match_it;
+            while (target_row_it != last_target_row_it && target_row_it->value() == source_ts) {
+                result.add_match(source_row, i, target_row_it->idx());
+                ++target_row_it;
+            }
```
alexowens90, line 643: This will have poor complexity when the target contains a lot of repeated index values that match an index value in the source

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
+        const Field& source_field = source_descriptor.field(source_field_position);
+        details::visit_type(source_field.type().data_type(), [&](auto source_field_dt) {
+            using SourceTDT = ScalarTagType<std::decay_t<decltype(source_field_dt)>>;
+            using SourceType = SourceRawType<SourceTDT>;
+            const ColumnWithStrings target_column = std::get<ColumnWithStrings>(proc.get(ColumnName{column_name}));
+            details::visit_type(target_column.column_->type().data_type(), [&](auto target_field_dt) {
+                using TargetTDT = ScalarTagType<decltype(target_field_dt)>;
+                if constexpr (std::same_as<std::decay_t<SourceTDT>, std::decay_t<TargetTDT>>) {
+                    auto target_values_to_rows = map_column_values_to_rows<TargetTDT>(target_column);
+                    std::span source_data = source_->get_tensor(source_field_position).span<SourceType>();
```
alexowens90, line 882: Couldn't this be restricted to `source_range`?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -322,17 +248,423 @@ auto map_column_values_to_rows(const ColumnWithStrings& column) {
...
+    } else {
+        if (is_matching_column) {
+            return false;
+        }
+        ColumnData target_column_data = target_column.data();
+        auto target = random_accessor<ScalarType>(&target_column_data);
+        for (size_t source_row_idx = 0; source_row_idx < source_data.size(); ++source_row_idx) {
+            row_slice_changed |= !rows_to_update[source_row_idx].empty();
+            for (const size_t target_row_idx : rows_to_update[source_row_idx]) {
+                target[target_row_idx] = source_data[source_row_idx];
```
alexowens90, line 561: It's possible `target[target_row_idx] == source_data[source_row_idx]` for every row, which would produce a false positive in `row_slice_changed`, but this is an optimisation, not a bug

cpp/arcticdb/processing/clause_merge_update.cpp
alexowens90, line 102: Might be worth homogenising in the else case on setting to `None`, instead of having a mixture of `None/NaN` based on what was in the target, ignoring what was in the source

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -192,9 +133,9 @@ bool merge_update_string_column_rowrange(
             for (size_t target_row : rows_to_update[source_row]) {
                 const auto py_string_object = source_column_view[source_row];
                 target_column_accessor[target_row] = write_py_string_to_pool_or_throw<TDT>(
-                        py_string_object, target_row, row_range, gil_lock, new_string_pool, target_field.name()
+                        py_string_object, target_row, row_range, gil_lock, new_string_pool, column_name
                 );
-                is_column_changed = true;
+                column_is_changed = true;
```
alexowens90: Similar to previous comment, this could produce false positives, but I wouldn't worry about it

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
+    if (strategy_.update_only()) {
+        if (auto [new_row_slices, time_slice_changed] = update(matched, std::move(row_slices)); time_slice_changed) {
+            std::vector<EntityId> res;
+            for (ProcessingUnit& row_slice : new_row_slices) {
+                const size_t entity_count = row_slice.segments_->size();
+                std::vector<EntityId> entts = component_manager_->add_entities(
+                        std::move(*row_slice.segments_),
+                        std::move(*row_slice.row_ranges_),
+                        std::move(*row_slice.col_ranges_),
+                        std::vector<EntityFetchCount>(entity_count, 1)
```
alexowens90: Could some downstream logic be simplified if we also added a `MergeUpdateInsertedRowsEntity` in this case where they are all zero?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
+            "MergeUpdateClause::update_and_insert should only be called for strategies that allow insertion"
+    );
+    // A group with nothing to insert can be left untouched only if it is self-contained. If it
+    // co-owns a boundary slice (entity_fetch_count > 1) with a neighbouring group that does insert,
+    // that neighbour rewrites part of the shared slice; this group must therefore also re-emit its
+    // owned portion, otherwise the stale old slice overlaps the neighbour's new slice during index
+    // reconstruction (merge_slices_and_keys) and produces a non-contiguous, unreadable index.
+    // Only the first and last row slices of a group can be shared with a neighbour (a group is a
+    // contiguous window of row slices), matching how get_source_start_end/get_target_start_end trim.
+    const bool co_owns_shared_slice = !row_slices.empty() && (row_slices.front().entity_fetch_count_->front() > 1 ||
```
alexowens90: There are several `!row_slices.empty()` checks littered around, which I think should be impossible by construction, but if not, is already short-circuited by this:
```
    if (entity_ids.empty()) {
        return {};
    }
```

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -428,425 +815,456 @@ std::vector<std::vector<EntityId>> MergeUpdateClause::structure_for_processing(s
...
+                    row_slices, [&](const auto& proc) { return proc.col_ranges_->size() == num_col_slices; }
+            ),
+            "All row slices should have the same number of column ranges"
+    );
+
+    ProcessingUnit result{};
+    result.segments_.emplace();
+    result.segments_->reserve(num_col_slices);
+    result.col_ranges_ = row_slices.front().col_ranges_;
+    std::vector<ColumnWithStrings> target_datas, target_index_datas;
```
alexowens90: nit: formatting

cpp/arcticdb/processing/test/test_merge_update.cpp
```diff
@@ -1598,12 +1651,1678 @@ TEST(MergeClauseDateRange, SourceMatchesAllIndexRangesButDoesNotMatchAnyRow) {
...
+            std::array<timestamp, 3>{-10, -5, -4},
+            std::array<int8_t, 3>{100, 101, 102},
+            std::array<unsigned, 3>{1000, 2000, 3000},
+            std::array<bool, 3>{true, false, true},
+            std::array<float, 3>{500.f, 600.f, 700.f},
+            std::array<timestamp, 3>{111, 222, 333}
+    );
+    MergeUpdateClause clause = create_clause(strategy, component_manager, std::move(input_frame));
+    const std::vector<std::vector<size_t>> structure_indices = clause.structure_for_processing(ranges_and_keys);
+    // Inserting in the beginning of the DataFrame picks the first row slice
```
alexowens90, line 1707: Is this to simplify the implementation? In theory, we don't need to read any segments from disk in this case?

cpp/arcticdb/processing/test/test_merge_update.cpp
```diff
@@ -1598,12 +1651,1678 @@ TEST(MergeClauseDateRange, SourceMatchesAllIndexRangesButDoesNotMatchAnyRow) {
...
+            std::array<int8_t, 3>{100, 101, 102},
+            std::array<unsigned, 3>{1000, 2000, 3000},
+            std::array<bool, 3>{true, false, true},
+            std::array<float, 3>{500.f, 600.f, 700.f},
+            std::array<timestamp, 3>{111, 222, 333}
+    );
+    MergeUpdateClause clause = create_clause(GetParam(), component_manager, std::move(input_frame));
+
+    const std::vector<std::vector<size_t>> structure_indices = clause.structure_for_processing(ranges_and_keys);
+    // Inserting in the beginning of the DataFrame picks the first row slice
```
alexowens90: copy/paste fail

docs/mkdocs/docs/notebooks/ArcticDB_merge.ipynb
alexowens90, line 1: The new code has print and display statements that haven't been executed/rendered in the notebook, which is out of sync with the rest of the notebook

cpp/arcticdb/entity/native_tensor.hpp
vasil-pashov: Sure, I used it because STL also uses the same pattern.

cpp/arcticdb/entity/native_tensor.hpp
vasil-pashov: The check in particular doesn't imply it, but it's true that this will not work for fortran styled data. I'll add additional check.

cpp/arcticdb/processing/test/test_merge_update.cpp
vasil-pashov: Yes, so that everything happens via the clause. Otherwise there must an additional step after the pipeline exists that writes the new segment.

cpp/arcticdb/entity/native_tensor.hpp
vasil-pashov: That's actually not used. Removing

cpp/arcticdb/processing/test/test_split_by_time_slice.cpp
vasil-pashov: To be fare it's not really needed. I removed all shuffling

cpp/arcticdb/processing/clause.hpp
vasil-pashov: Double buffering (having two different components with the same type) is not supported.

size_t isn't taken, but I prefer the idea of wrapping POD in structs as it's a bit more descriptive and it's more obvious when iterating over all components.

I put it here as it's related to insertion and it's not supposed to be used by anything else. Don't mind putting it in the more general component manager class.

cpp/arcticdb/processing/component_manager.hpp
vasil-pashov: Discussed on slack. The Write clause does not know about the `MergeUpdateInsertedRowsEntity` class and cannot forward custom components of clauses that precede it. Issue opened on Monday for a general solution to the problem Monday 12618296803

cpp/arcticdb/version/version_core.cpp
vasil-pashov: see https://github.com/man-group/ArcticDB/pull/3239/changes#r3645051996

cpp/arcticdb/processing/clause_utils.cpp
vasil-pashov: The proposed algorithm is not equivalent. 
Example 1: Time ranges [0;5) [4; 10). The suggested algorithm will produce one group [0,1]. The correct groups are [0,1] working on time range of [0;4] and [1] working on time range [5;9]
Example 2: Time ranges [0;5) [4; 10) [9; 15) [14;20). The suggested algorithm will produce two groups [0,1] [2,3] but segment 1 and 2 share a value and must be in the same group. The correct groups are [0,1], [1,2], [2,3], [3]

I agree it's not easy to read. Added a new equivalent version I believe is more readable.

cpp/arcticdb/util/buffer.hpp
vasil-pashov: CursoredBuffer has a clear method. Calling clear on the buffer which is a template argument. So when the underlying buffer happens to be `Buffer` it results in a compilation error

cpp/arcticdb/version/local_versioned_engine.cpp
vasil-pashov: Segments can be potentially deduplicated. Simple example would be an update on column sliced data. Imagine two column slices and the update happens only in the first.

We don't strictly need it but I don't think it hurts to have it. See example in `test_dedup_map_deduplicates_unchanged_column_slice`

cpp/arcticdb/version/version_core.cpp
vasil-pashov: Added `test_dedup_map_deduplicates_unchanged_column_slice`

cpp/arcticdb/version/version_core.cpp
vasil-pashov: Fixed a PR build on top of this: https://github.com/man-group/ArcticDB/pull/3247/changes#diff-495e358e10c46902ba449e362a292f3ebe9a10c49e15741071079c1301c2b340

But I'll apply it here as well and handle the conflict

python/tests/unit/arcticdb/version_store/test_merge_update.py
vasil-pashov: I consider than an implementation detail which I wouldn't test in end-to-end tests like this. We're likely to change it so that performs some slicing. As long as the result matches the expected I consider it to be functionally correct.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: For insertion processing units can share segments. Which means that we must guard string pools with a lock. I wanted to avoid all synchronization between units.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: Wouldn't call it premature as the code was already there. The insert codepath also joints segments together at the moment which is not ideal. Also the row ranges will need this anyway even when we start inserting in row ranges. The plan to insert in row range is to run the update in the processing pipeline and record which rows were updated and then after the pipeline ends write the last segment containing all new values. Maybe we can give it a try after we implement the slicing

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: Note that there's intermediate string pool and the final string pool contains only the used values. Anyway I did something similar to you sugestion. I'm not quite convinced the code became simpler, but at least we're not rebuilding the string pool now.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: Yeh, I though it's not worth the added complexity. But it turned out easier than expected.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: This does not iterate over the full source. The structure is created in process (one per unit) and it uses `get_source_start_end` to restrict the source.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: Leftover - fixed.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: We don't need this anymore. Nice catch

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: Indeed

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: Yeh, I decided to remove this tracking. When the dedup map is enabled this doesn't do much besides complicating the code.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: I could be missing something, but the overall complexity is O(m * log(n/m) + n) where: m is the count of source rows in the bounds of the segment, n is the number of target rows in the segment.

The inner loop will iterate over all source_rows that fall into the slice. It doesn't matter if there are lots of repeated values, the work should be the same.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: This function is used only for row ranged data. The source range is the whole source.

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: I don't think so. Do you have something in mind?

cpp/arcticdb/processing/clause_merge_update.cpp
vasil-pashov: I like the idea of storing exactly the same thing the user passed. Not a hill I'm willing to die on, though.

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: I guess, although that is also true of `update`, and we don't do this there

cpp/arcticdb/processing/clause_merge_update.cpp
alexowens90: Yep my bad, I was reading this wrong

cpp/arcticdb/processing/clause_merge_update.cpp
alexowens90: I see this got simplified anyway

alexowens90 (approved):

PR #3242, opened 2026-07-13
Read and write support for arrow-backed string dtype

ognyanstoimenov:
#### Reference Issues/PRs

This introduces read and write support for the new arrow-backed string dtype introduced in pandas2.3 and made the default in pandas3.

**Note for reviewers**: I think it would be easiest for this PR to be reviewed one commit at a time. I've split them into logical parts so they shouldn't overlap. More info on each commit below.

### 1. Refactor of ReadOptions (already reviewed)

* `ReadOptionsData` now holds a variant of PythonOutputConfig and ArrowOutputConfig. Old output_format enum removed. The output format for the frame is determined by checking which of the two config variants are used.
* Per column outputformat getters in ReadOptions
* Per column Arrow format getters in ReadOptions
* Renamed `output_format` to `output_format_for_frame` so it is easily distinguishable from `output_format_for_column`
* Changed some call sites that were building columns to use the per column OutputFormat rather than the frame one.

### 3. Implementation of reading str dtype (Reviewed by @alexowens90)
* Based on the availability of the new string dtype sets the option in the new ReadOptions 
* Check the option in read_frame and uses the existing ArrowStringHandler to handle the string dtype
* Creates a detachable chunked column for arrow backed columns
* Truncation handling for arrow-backed columns
* Passes the column to python as an arrow array (not as record batch) but still uses RecordBatchData wrapper. And is zero-copy as seen by the benchmarks below 
* With the help of claude I found existing tests that have string columns and parameterized them with read_type[object, str] and write_type (which is currently just object, as str is still not implemented; that will be done in the following PR). That should handle most use cases.
  * Added some cases I thought were missing, but might have missed some others though 
  * Instead of parametarizing existing tests, I initially thought about having a full workflow that runs our current suite with the new dtype, but decided against that as it is a bit of an overkill, because not all tests are string related and also it is nice to test combinations of writing as str/reading as obj etc. **Edit: after discussion we decided to go the route of running the full suite on the CI with the new string dtype. (python311 and 314, all platforms) **

<details> 
<summary> Benchmarks show similar performance to native arrow </summary>

## Read benchmarks: pandas object vs pandas str vs native arrow

Source: `python/benchmarks/arrow.py`, class `ArrowStrings`. Data: 10 string columns,
`arrow_string_format=LARGE_STRING` (format doesn't affect the pandas output, so only
LARGE_STRING rows are shown for the pandas comparisons).

## Peak memory (read), MiB

Benchmarks: `peakmem_read` (native arrow) / `peakmem_read_pandas_object` / `peakmem_read_pandas_str`

| rows      | date_range | unique_strings | pandas object | pandas str | native arrow |
|-----------|------------|-----------------|---------------|------------|--------------|
| 100,000   | None       | 100             | 184           | 199        | 198          |
| 100,000   | None       | 10,000          | 193           | 202        | 199          |
| 100,000   | middle     | 100             | 185           | 206        | 206          |
| 100,000   | middle     | 10,000          | 194           | 209        | 209          |
| 1,000,000 | None       | 100             | 507           | 632        | 632          |
| 1,000,000 | None       | 10,000          | 586           | 644        | 642          |
| 1,000,000 | middle     | 100             | 507           | 649        | 649          |
| 1,000,000 | middle     | 10,000          | 585           | 659        | 660          |

## Time (read), ms

Benchmarks: `time_read` (native arrow) / `time_read_pandas_object` / `time_read_pandas_str`

| rows      | date_range | unique_strings | pandas object | pandas str | native arrow |
|-----------|------------|-----------------|---------------|------------|--------------|
| 100,000   | None       | 100             | 39.3          | 50.2       | 46.8         |
| 100,000   | None       | 10,000          | 56.1          | 52.4       | 49.3         |
| 100,000   | middle     | 100             | 39.5          | 66.1       | 64.0         |
| 100,000   | middle     | 10,000          | 57.6          | 70.9       | 66.4         |
| 1,000,000 | None       | 100             | 59.8          | 56.3       | 57.0         |
| 1,000,000 | None       | 10,000          | 171           | 73.5       | 55.6         |
| 1,000,000 | middle     | 100             | 58.7          | 70.1       | 63.6         |
| 1,000,000 | middle     | 10,000          | 168           | 78.3       | 69.8         |

</details>

### 7. Implementation of writing str dtype
* In python _to_primitive now returns `Union[np.ndarray, List[RecordBatchData]]` for a column
  * it is a list because a column may have > 1 chunk.  
* `PandasData` which is what crosses python -> cpp now holds a variant of py::array and list of shared_ptr<RecordBatchData>  
* Adds new `set_from_frame_data` method to input_frame that accepts a vector of FrameData(which is a variant of NativeTensors and columns. And that is used when converting in python_to_tensor_frame.cpp. Also handles string dtype index writes
<details> 
<summary> Comparisons of reading and writing different types </summary>

| scenario | format | dtype        | col mem (MiB) | write peak (MiB) | read peak (MiB) |
|----------|--------|--------------|---------------|-------------------|------------------|
| small    | object | object       | 139.2         | 589.7             | 387.8            |
| small    | str    | str          | 45.8          | 454.9             | 343.5            |
| small    | arrow  | large_string | 45.8          | 432.5             | 337.5            |
| small    | polars | String       | 30.5          | 494.9             | 375.0            |
| large    | object | object       | 101.6         | 627.1             | 509.9            |
| large    | str    | str          | 96.9          | 617.1             | 507.6            |
| large    | arrow  | large_string | 96.9          | 612.5             | 508.4            |
| large    | polars | String       | 96.1          | 715.0             | 509.0            |

- **small**: 2,000,000 rows x 1 string col (8 chars) + 1 numeric col
- **large**: 100,000 rows x 1 string col (1000 chars) + 1 numeric col

</details>

### DIFF
--- a/cpp/arcticdb/arrow/arrow_utils.cpp
+++ b/cpp/arcticdb/arrow/arrow_utils.cpp
@@ -379,6 +379,17 @@ std::optional<ArrowMeta::ColumnMeta> column_metadata(
     return std::nullopt;
 }
 
+std::vector<std::shared_ptr<RecordBatchData>> column_to_arrow_arrays(const Column& column, std::string_view name) {
+    auto column_arrays = arrow_arrays_from_column(column, name);
+    std::vector<std::shared_ptr<RecordBatchData>> output;
+    output.reserve(column_arrays.size());
+    for (auto& column_array : column_arrays) {
+        auto [arr, schema] = sparrow::extract_arrow_structures(std::move(column_array));
+        output.emplace_back(std::make_shared<RecordBatchData>(arr, schema));
+    }
+    return output;
+}
+
 std::vector<sparrow::record_batch> segment_to_arrow_data(
         SegmentInMemory& segment, const proto::descriptors::NormalizationMetadata& norm_meta
 ) {
--- a/cpp/arcticdb/arrow/arrow_utils.cpp
+++ b/cpp/arcticdb/arrow/arrow_utils.cpp
@@ -12,6 +12,7 @@
 #include <arcticdb/column_store/column.hpp>
 #include <arcticdb/column_store/memory_segment.hpp>
 #include <arcticdb/util/allocator.hpp>
+#include <arcticdb/util/collection_utils.hpp>
 #include <sparrow/layout/array_access.hpp>
 #include <sparrow/layout/primitive_data_access.hpp>
 #include <sparrow/utils/temporal.hpp>
@@ -381,10 +382,11 @@ std::optional<ArrowMeta::ColumnMeta> column_metadata(
 
 std::vector<std::shared_ptr<RecordBatchData>> column_to_arrow_arrays(const Column& column, std::string_view name) {
     auto column_arrays = arrow_arrays_from_column(column, name);
-    std::vector<std::shared_ptr<RecordBatchData>> output;
-    output.reserve(column_arrays.size());
+    auto output = util::reserve_vector<std::shared_ptr<RecordBatchData>>(column_arrays.size());
     for (auto& column_array : column_arrays) {
-        auto [arr, schema] = sparrow::extract_arrow_structures(std::move(column_array));
+        sparrow::record_batch batch{};
+        batch.add_column(std::string{name}, std::move(column_array));
+        auto [arr, schema] = sparrow::extract_arrow_structures(std::move(batch));
         output.emplace_back(std::make_shared<RecordBatchData>(arr, schema));
     }
     return output;
--- a/cpp/arcticdb/arrow/arrow_utils.cpp
+++ b/cpp/arcticdb/arrow/arrow_utils.cpp
@@ -534,6 +534,9 @@ std::pair<std::vector<Column>, entity::StreamDescriptor> record_batches_to_colum
 
     uint64_t start_row{0};
     for (const auto& batch : record_batches) {
+        if (batch.nb_rows() == 0) {
+            continue;
+        }
         for (size_t idx = 0; idx < batch.nb_columns(); ++idx) {
             auto& column = columns[idx];
             const auto data_type = column.type().data_type();
--- a/cpp/arcticdb/pipeline/input_frame.cpp
+++ b/cpp/arcticdb/pipeline/input_frame.cpp
@@ -46,12 +46,28 @@ SortedValue compute_index_sortedness(const Column& index_column) {
 
 InputFrame::InputFrame() : index(stream::empty_index()) {}
 
+void InputFrame::set_from_tensors(StreamDescriptor&& desc, std::vector<NativeTensor>&& field_tensors) {
+    std::vector<FieldData> columns(
+            std::make_move_iterator(field_tensors.begin()), std::make_move_iterator(field_tensors.end())
+    );
+    set_from_frame_data(std::move(desc), std::move(columns), {}, /*has_only_tensors=*/true);
+}
+
+void InputFrame::set_from_frame_data(
+        StreamDescriptor&& desc, std::vector<FieldData>&& columns,
+        std::vector<sparrow::record_batch>&& arrow_buffer_owners, bool has_only_tensors
+) {
+    desc_ = std::move(desc);
+    columns_ = std::move(columns);
+    arrow_buffer_owners_ = std::move(arrow_buffer_owners);
+    has_only_tensors_ = has_only_tensors;
+}
+
 void InputFrame::set_from_columns(
         std::vector<Column>&& cols, StreamDescriptor&& desc, std::vector<sparrow::record_batch>&& arrow_buffer_owners,
         SortednessScan sortedness_scan
 ) {
     util::check(norm_meta.has_experimental_arrow(), "Unexpected non-Arrow norm metadata provided with Arrow data");
-    desc_ = std::move(desc);
     if (norm_meta.experimental_arrow().has_index()) {
         user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
                 !cols.empty(), "Arrow index column specified but there are zero columns"
@@ -61,23 +77,22 @@ void InputFrame::set_from_columns(
                 "Specified Arrow index column has non-time type {}",
                 cols[0].type().data_type()
         );
-        desc_.set_index({IndexDescriptorImpl::Type::TIMESTAMP, 1});
-        index = stream::TimeseriesIndex{std::string(desc_.field(0).name())};
+        desc.set_index({IndexDescriptorImpl::Type::TIMESTAMP, 1});
+        index = stream::TimeseriesIndex{std::string(desc.field(0).name())};
         // Arrow input does not record sortedness up front (unlike pandas), so we have to compute it when required.
-        desc_.set_sorted(
+        desc.set_sorted(
                 sortedness_scan == SortednessScan::SCAN_IF_UNKNOWN ? compute_index_sortedness(cols[0])
                                                                    : SortedValue::UNKNOWN
         );
     } else {
-        desc_.set_index({IndexDescriptorImpl::Type::ROWCOUNT, 0});
+        desc.set_index({IndexDescriptorImpl::Type::ROWCOUNT, 0});
         index = stream::RowCountIndex{};
-        desc_.set_sorted(SortedValue::UNKNOWN);
+        desc.set_sorted(SortedValue::UNKNOWN);
     }
 
     num_rows = cols.empty() ? 0 : cols[0].row_count();
-    columns_ = std::vector<FieldData>(std::make_move_iterator(cols.begin()), std::make_move_iterator(cols.end()));
-    arrow_buffer_owners_ = std::move(arrow_buffer_owners);
-    has_only_tensors_ = false;
+    std::vector<FieldData> columns(std::make_move_iterator(cols.begin()), std::make_move_iterator(cols.end()));
+    set_from_frame_data(std::move(desc), std::move(columns), std::move(arrow_buffer_owners), false);
 }
 
 StreamDescriptor& InputFrame::desc() { return desc_; }
--- a/cpp/arcticdb/pipeline/output_config.hpp
+++ b/cpp/arcticdb/pipeline/output_config.hpp
@@ -17,7 +17,7 @@ namespace arcticdb {
 
 enum class ArrowOutputStringFormat : uint8_t { CATEGORICAL, LARGE_STRING, SMALL_STRING };
 
-enum class PandasStringFormat : uint8_t { OBJECT, ARROW };
+enum class PandasStringFormat : uint8_t { OBJECT, ARROW_LARGE_STRING };
 
 struct PandasOutputConfig {
     PandasStringFormat default_string_format_ = PandasStringFormat::OBJECT;
--- a/cpp/arcticdb/pipeline/output_config.hpp
+++ b/cpp/arcticdb/pipeline/output_config.hpp
@@ -8,13 +8,26 @@
 
 #pragma once
 
+#include <cstdint>
+#include <string>
+#include <unordered_map>
+#include <variant>
+
 namespace arcticdb {
 
 enum class ArrowOutputStringFormat : uint8_t { CATEGORICAL, LARGE_STRING, SMALL_STRING };
+
+enum class PandasStringFormat : uint8_t { OBJECT, ARROW };
+
+struct PandasOutputConfig {
+    PandasStringFormat default_string_format_ = PandasStringFormat::OBJECT;
+};
 
 struct ArrowOutputConfig {
     ArrowOutputStringFormat default_string_format_ = ArrowOutputStringFormat::LARGE_STRING;
     std::unordered_map<std::string, ArrowOutputStringFormat> per_column_string_format_;
 };
+
+using OutputConfig = std::variant<PandasOutputConfig, ArrowOutputConfig>;
 
 } // namespace arcticdb
--- a/cpp/arcticdb/pipeline/read_frame.cpp
+++ b/cpp/arcticdb/pipeline/read_frame.cpp
@@ -62,7 +62,9 @@ std::pair<StreamDescriptor, BlockConfigPerColumn> get_filtered_descriptor_and_bl
                 auto handlers = TypeHandlerRegistry::instance();
 
                 for (auto& field : *fields) {
-                    if (auto handler = handlers->get_handler(read_options.output_format(), field.type())) {
+                    if (auto handler = handlers->get_handler(
+                                read_options.output_format_for_column_type(field.type()), field.type()
+                        )) {
                         auto [output_type, block_config] =
                                 handler->output_type_and_block_config(field.type(), field.name(), read_options);
                         block_config_per_column.emplace_back(block_config);
@@ -155,7 +157,7 @@ SegmentInMemory allocate_contiguous_frame(
 }
 
 SegmentInMemory allocate_frame(const std::shared_ptr<PipelineContext>& context, const ReadOptions& read_options) {
-    if (read_options.output_format() == OutputFormat::ARROW)
+    if (read_options.output_format_for_frame() == OutputFormat::ARROW)
         return allocate_chunked_frame(context, read_options);
     else
         return allocate_contiguous_frame(context, read_options);
@@ -345,7 +347,9 @@ void decode_or_expand(
 ) {
     const auto source_type_desc = mapping.source_type_desc_;
     const auto dest_type_desc = mapping.dest_type_desc_;
-    if (auto handler = get_type_handler(read_options.output_format(), source_type_desc, dest_type_desc)) {
+    if (auto handler = get_type_handler(
+                read_options.output_format_for_column_type(dest_type_desc), source_type_desc, dest_type_desc
+        )) {
         handler->handle_type(
                 data,
                 dest_column,
@@ -369,7 +373,7 @@ void decode_or_expand(
         // hence `ndarray().sparse_map_bytes() > 0` is not enough. We check whether the uncompressed size matches
         // the expected number of rows.
         const bool is_sparse = dense_num_rows < mapping.num_rows_;
-        const bool is_arrow = read_options.output_format() == OutputFormat::ARROW;
+        const bool is_arrow = read_options.output_format_for_column_type(dest_type_desc) == OutputFormat::ARROW;
 
         if (is_sparse) {
             // We explicitly allocate an intermediate dense_buffer to decode the dense data into.
@@ -470,7 +474,7 @@ ColumnTruncation get_truncate_range(
     const auto& row_range = context.slice_and_key().slice().row_range;
     const auto& first_row_offset = frame.offset();
     auto adjusted_row_range = RowRange(row_range.first - first_row_offset, row_range.second - first_row_offset);
-    if (read_options.output_format() == OutputFormat::ARROW) {
+    if (read_options.output_format_for_frame() == OutputFormat::ARROW) {
         util::variant_match(
                 read_query.row_filter,
                 [&truncate_rows,
@@ -998,6 +1002,7 @@ struct ReduceColumnTask : async::BaseTask {
             return {};
         }();
 
+        auto output_format_for_column = read_options_.output_format_for_column_type(column.type());
         if (dynamic_schema && column_data == slice_map_->columns_.end()) {
             if (is_fixed_string_type(field_type)) {
                 // Special case where we have a fixed-width string column that is all null (e.g. dynamic schema
@@ -1014,26 +1019,14 @@ struct ReduceColumnTask : async::BaseTask {
                 swap(prev_buffer, new_buffer);
             } else {
                 NullValueReducer null_reducer{
-                        column,
-                        context_,
-                        frame_,
-                        shared_data_,
-                        handler_data_,
-                        read_options_.output_format(),
-                        default_value
+                        column, context_, frame_, shared_data_, handler_data_, output_format_for_column, default_value
                 };
                 null_reducer.finalize();
             }
         } else if (column_data != slice_map_->columns_.end()) {
             if (dynamic_schema) {
                 NullValueReducer null_reducer{
-                        column,
-                        context_,
-                        frame_,
-                        shared_data_,
-                        handler_data_,
-                        read_options_.output_format(),
-                        default_value
+                        column, context_, frame_, shared_data_, handler_data_, output_format_for_column, default_value
                 };
                 for (const auto& row : column_data->second) {
                     PipelineContextRow context_row{context_, row.second.context_index_};
--- a/cpp/arcticdb/pipeline/read_frame.cpp
+++ b/cpp/arcticdb/pipeline/read_frame.cpp
@@ -112,55 +112,35 @@ void finalize_segment_setup(
     handle_modified_descriptor(context, output);
 }
 
-SegmentInMemory allocate_chunked_frame(
-        const std::shared_ptr<PipelineContext>& context, const ReadOptions& read_options
-) {
-    ARCTICDB_SAMPLE_DEFAULT(AllocContiguousFrame)
+SegmentInMemory allocate_frame(const std::shared_ptr<PipelineContext>& context, const ReadOptions& read_options) {
+    ARCTICDB_SAMPLE_DEFAULT(AllocateFrame)
     auto [offset, row_count] = offset_and_row_count(context);
     auto block_row_counts = output_block_row_counts(context);
-    ARCTICDB_DEBUG(log::version(), "Allocated chunked frame with offset {} and row count {}", offset, row_count);
     auto [desc, block_config_per_column] = get_filtered_descriptor_and_block_config(context, read_options);
     SegmentInMemory output{
             std::move(desc), 0, AllocationType::DETACHABLE, Sparsity::NOT_PERMITTED, block_config_per_column
     };
-
     for (auto& column : output.columns()) {
         const auto data_size = data_type_size(column->type());
-        for (auto block_row_count : block_row_counts) {
-            if (block_row_count > 0) {
-                // We can end up with empty segments from the processing pipeline, e.g. when:
-                // - Filtering a data key to the empty set (e.g. date_range = (3, 3) in a data key with no index=3)
-                // - Resampling with a date range with a bucket slice containing no indices
-                // 0 sized memory blocks would break the offset assumptions in chunked buffers, and it is fine to have
-                // number of memory blocks not equal number of segments because follow-up methods like
-                // `copy_frame_data_to_buffer` rely on offsets rather than block indices.
-                const auto bytes = block_row_count * data_size;
-                column->allocate_and_advance_by(bytes);
+        if (read_options.output_format_for_column_type(column->type()) == OutputFormat::ARROW) {
+            for (auto block_row_count : block_row_counts) {
+                if (block_row_count > 0) {
+                    // We can end up with empty segments from the processing pipeline, e.g. when:
+                    // - Filtering a data key to the empty set (e.g. date_range = (3, 3) in a data key with no index=3)
+                    // - Resampling with a date range with a bucket slice containing no indices
+                    // 0 sized memory blocks would break the offset assumptions in chunked buffers, and it is fine to
+                    // have number of memory blocks not equal number of segments because follow-up methods like
+                    // `copy_frame_data_to_buffer` rely on offsets rather than block indices.
+                    const auto bytes = block_row_count * data_size;
+                    column->allocate_and_advance_by(bytes);
+                }
             }
+        } else if (row_count > 0) {
+            column->allocate_and_advance_by(row_count * data_size);
         }
     }
-
     finalize_segment_setup(output, offset, row_count, context);
     return output;
-}
-
-SegmentInMemory allocate_contiguous_frame(
-        const std::shared_ptr<PipelineContext>& context, const ReadOptions& read_options
-) {
-    ARCTICDB_SAMPLE_DEFAULT(AllocChunkedFrame)
-    auto [offset, row_count] = offset_and_row_count(context);
-    auto desc = get_filtered_descriptor_and_block_config(context, read_options).first;
-    // block_config_per_column is not used for contiguous frame allocation
-    SegmentInMemory output{std::move(desc), row_count, AllocationType::DETACHABLE, Sparsity::NOT_PERMITTED};
-    finalize_segment_setup(output, offset, row_count, context);
-    return output;
-}
-
-SegmentInMemory allocate_frame(const std::shared_ptr<PipelineContext>& context, const ReadOptions& read_options) {
-    if (read_options.output_format_for_frame() == OutputFormat::ARROW)
-        return allocate_chunked_frame(context, read_options);
-    else
-        return allocate_contiguous_frame(context, read_options);
 }
 
 size_t get_index_field_count(const SegmentInMemory& frame) { return frame.descriptor().index().field_count(); }
@@ -347,9 +327,8 @@ void decode_or_expand(
 ) {
     const auto source_type_desc = mapping.source_type_desc_;
     const auto dest_type_desc = mapping.dest_type_desc_;
-    if (auto handler = get_type_handler(
-                read_options.output_format_for_column_type(dest_type_desc), source_type_desc, dest_type_desc
-        )) {
+    const auto column_format = read_options.output_format_for_column_type(dest_type_desc);
+    if (auto handler = get_type_handler(column_format, source_type_desc, dest_type_desc)) {
         handler->handle_type(
                 data,
                 dest_column,
@@ -373,7 +352,7 @@ void decode_or_expand(
         // hence `ndarray().sparse_map_bytes() > 0` is not enough. We check whether the uncompressed size matches
         // the expected number of rows.
         const bool is_sparse = dense_num_rows < mapping.num_rows_;
-        const bool is_arrow = read_options.output_format_for_column_type(dest_type_desc) == OutputFormat::ARROW;
+        const bool is_arrow = column_format == OutputFormat::ARROW;
 
         if (is_sparse) {
             // We explicitly allocate an intermediate dense_buffer to decode the dense data into.
@@ -419,7 +398,10 @@ void decode_or_expand(
             data += decode_field(source_type_desc, encoded_field_info, data, sink, bv, encoding_version);
             handle_type_promotion(mapping, dest_column);
         }
-        handle_truncation(dest_column, mapping.truncate_);
+
+        if (is_arrow) {
+            handle_truncation(dest_column, mapping.truncate_);
+        }
     }
 }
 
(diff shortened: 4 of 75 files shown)


cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -60,40 +67,78 @@ struct ReadOptions {
 
     [[nodiscard]] const std::optional<bool>& incompletes() const { return data_->incompletes_; }
 
-    void set_output_format(OutputFormat output_format) { data_->output_format_ = output_format; }
+    void set_output_config(OutputConfig output_config) { data_->output_config_ = std::move(output_config); }
```
IvoDD: Nit: What do you think about constructing by rvalue? `OutputConfig` is not very big now but it will probably become bigger over time

cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -60,40 +67,78 @@ struct ReadOptions {
...
+                            return *fmt;
+                    }
+                    return arrow.default_string_format_;
+                },
+                [](const PandasOutputConfig& pandas) -> ArrowOutputStringFormat {
+                    util::check(
+                            pandas.default_string_format_ == PandasStringFormat::ARROW,
+                            "Arrow string format requested for a non-arrow column"
+                    );
+                    return ArrowOutputStringFormat::LARGE_STRING;
```
IvoDD, line 123: Nit: An inline comment explaining that pandas arrow string format only supports large_string would be useful

cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -134,6 +178,9 @@ struct BatchReadOptions {
     void set_output_format(OutputFormat output_format) { data_->output_format_ = output_format; }
```
IvoDD, line 178: What do you think about removing the `BatchReadOptionsData.output_format` as well? 
Instead we could use the `read_options_per_symbol`'s first element. I don't remember where this was used but theoretically nothing should be stopping us from having a mix of arrow and pandas output formats in which case the batch level property doesn't make much sense.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -875,9 +895,10 @@ void register_bindings(py::module& version, py::exception<arcticdb::ArcticExcept
                         const VersionQuery& version_query,
                         const std::shared_ptr<ReadQuery>& read_query,
                         const ReadOptions& read_options) {
-                        auto handler_data = std::make_shared<std::any>(
-                                TypeHandlerRegistry::instance()->get_handler_data(read_options.output_format())
-                        );
+                        auto handler_data =
+                                std::make_shared<std::any>(TypeHandlerRegistry::instance()->get_handler_data(
+                                        read_options.output_format_for_frame()
```
IvoDD, line 913: I suppose this is for a follow up PR but we'll also need to rework the `handler_data`. If we have a mix of arrow and numpy columns we'll need to store the handler_data for both arrow and pandas. In reality that's not too complicated because arrow's handler data is empty.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1706,7 +1706,8 @@ void copy_frame_data_to_buffer(
             dst_column.type(),
             destination.field(target_index).name()
     );
-    if (auto handler = get_type_handler(read_options.output_format(), src_column.type(), dst_column.type()); handler) {
+    auto output_format_for_column = read_options.output_format_for_column_type(dst_column.type());
```
IvoDD, line 1706: Here we open the door to use `handler_data` which doesn't match the column output format. As per previous comment this needs to be addressed in a future PR.

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -155,7 +157,7 @@ SegmentInMemory allocate_contiguous_frame(
 }
 
 SegmentInMemory allocate_frame(const std::shared_ptr<PipelineContext>& context, const ReadOptions& read_options) {
-    if (read_options.output_format() == OutputFormat::ARROW)
+    if (read_options.output_format_for_frame() == OutputFormat::ARROW)
```
IvoDD: Allocation should become column format specific. Ok to do in a follow up PR. 
Currently if we ever specify PandasOutputConfig{string_format=ARROW} this would break.

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -470,7 +474,7 @@ ColumnTruncation get_truncate_range(
     const auto& row_range = context.slice_and_key().slice().row_range;
     const auto& first_row_offset = frame.offset();
     auto adjusted_row_range = RowRange(row_range.first - first_row_offset, row_range.second - first_row_offset);
-    if (read_options.output_format() == OutputFormat::ARROW) {
+    if (read_options.output_format_for_frame() == OutputFormat::ARROW) {
```
IvoDD: Out of curiousity how do we plan to handle truncation for pandas with arrow strings?

cpp/arcticdb/stream/test/test_incompletes.cpp
```diff
@@ -39,7 +39,7 @@ TEST(Append, Simple) {
     generate_filtered_field_descriptors(pipeline_context, {});
 
     auto read_options = ReadOptions{};
-    read_options.set_output_format(OutputFormat::NATIVE);
```
alexowens90, line 42: What is `OutputFormat::NATIVE` used for again?

cpp/arcticdb/pipeline/output_config.hpp
```diff
@@ -8,13 +8,26 @@
...
+#include <cstdint>
+#include <string>
+#include <unordered_map>
+#include <variant>
+
 namespace arcticdb {
 
 enum class ArrowOutputStringFormat : uint8_t { CATEGORICAL, LARGE_STRING, SMALL_STRING };
 
+enum class PandasStringFormat : uint8_t { OBJECT, ARROW };
```
alexowens90: Maybe `ARROW_LARGE_STRING` would make it clearer which of the Arrow string formats Pandas supports

cpp/arcticdb/pipeline/read_options.hpp
alexowens90: Yeh I don't remember why this is here, but smells like it shouldn't be

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -230,6 +231,27 @@ void register_bindings(py::module& version, py::exception<arcticdb::ArcticExcept
             .value("LARGE_STRING", ArrowOutputStringFormat::LARGE_STRING)
             .value("SMALL_STRING", ArrowOutputStringFormat::SMALL_STRING);
 
+    py::enum_<PandasStringFormat>(version, "InternalPandasStringFormat")
+            .value("OBJECT", PandasStringFormat::OBJECT)
+            .value("ARROW", PandasStringFormat::ARROW);
```
alexowens90: I think we can drop the `Internal` prefix from this and `InternalArrowOutputStringFormat`. `InternalOutputFormat` has it because in the Python layer there is an output format enum with Pandas, PyArrow, and Polars, but these enums don't have that conflict.

cpp/arcticdb/pipeline/read_frame.cpp
alexowens90: I think `handle_truncation` means it should just work?

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -379,6 +379,17 @@ std::optional<ArrowMeta::ColumnMeta> column_metadata(
     return std::nullopt;
 }
 
+std::vector<std::shared_ptr<RecordBatchData>> column_to_arrow_arrays(const Column& column, std::string_view name) {
```
ognyanstoimenov, line 383: This reuses the RecordBatchData struct although it cointains arrays instead of record batches. Might make more sense to rename it or/and enclude a type enum in the RecordBatchData

python/arcticdb/version_store/_normalization.py
```diff
@@ -1067,6 +1068,41 @@ def _consolidate(self):
...
+    # The pandas arrow-backed "str" dtype (StringDtype with na_value) was added in pandas 2.3. future.infer_string
+    # exists from pandas 2.1, so the option can be set on 2.1/2.2 but there this dtype cannot be constructed.
+    try:
+        pd.StringDtype(storage="pyarrow", na_value=np.nan)
+        return True
+    except (TypeError, ImportError):
+        return False
+
+
+_ARROW_BACKED_STR_DTYPE_SUPPORTED = _arrow_backed_str_dtype_supported()
```
ognyanstoimenov: Could make more sense to have this as disabled until we merge the write support

cpp/arcticdb/pipeline/test/test_frame_allocation.cpp
alexowens90, line 1: We could add some more tests to this file, covering pure-numpy and numpy/arrow mixed output frames

python/arcticdb/version_store/_store.py
```diff
@@ -2696,7 +2712,8 @@ def _compute_filter_start_end_row(self, read_result, read_query):
                 end_idx = index.searchsorted(datetime64(read_query.row_filter.end_ts, "ns"), side="right")
         else:
             raise ArcticNativeException("Unrecognised row_filter type: {}".format(type(read_query.row_filter)))
-        return (start_idx, end_idx)
+        start, stop, _ = slice(start_idx, end_idx).indices(read_result.frame_data.row_count)
```
alexowens90, line 2740: I don't understand what's happening here

python/arcticdb/version_store/_normalization.py
```diff
@@ -1067,6 +1068,41 @@ def _consolidate(self):
...
+
+
+def _use_pyarrow_strings_in_pandas():
+    if not _PYARROW_AVAILABLE or not _ARROW_BACKED_STR_DTYPE_SUPPORTED:
+        return False
+    return bool(pd.get_option("future.infer_string"))
```
alexowens90: Is this the idiomatic way to identify the string format in Pandas `3.X.X` as well?

python/arcticdb/version_store/_normalization.py
alexowens90: I don't think we need to wait, data written as object columns will be readable as Arrow immediately

alexowens90: In fact, when we release this into Man we can encourage people to try it by setting the `future.infer_string` flag, with the caveat that data read in this way will not be writeable back. Will only work with 3.14, 3.11 envs use Pandas `2.1.4`

python/arcticdb/version_store/_normalization.py
```diff
@@ -1067,6 +1068,41 @@ def _consolidate(self):
...
 
 
+def _arrow_backed_str_dtype_supported():
+    # The pandas arrow-backed "str" dtype (StringDtype with na_value) was added in pandas 2.3. future.infer_string
+    # exists from pandas 2.1, so the option can be set on 2.1/2.2 but there this dtype cannot be constructed.
+    try:
+        pd.StringDtype(storage="pyarrow", na_value=np.nan)
+        return True
+    except (TypeError, ImportError):
+        return False
```
alexowens90: Is there a more direct way to check if this will work? Like `_PYARROW_AVAILABLE and pandas_version >= (x,y,z)`?

python/arcticdb/version_store/_normalization.py
```diff
@@ -1096,6 +1132,10 @@ def gen_blocks():
                 for idx, a in enumerate(arrays):
                     if idx < n_ind:
                         continue
+                    if _is_arrow_string_column(a):
```
alexowens90, line 1122: I haven't got to the testing yet, but is there an explicit test that reading Arrow-format strings with dataframe consolidation both enabled and disabled?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -780,13 +762,25 @@ void decode_into_frame_dynamic(
...
+            if (read_options.output_format_for_column_type(type) != OutputFormat::ARROW) {
+                return no_truncation;
+            }
+            if (!truncate_range) {
+                truncate_range = get_truncate_range(
+                        frame, context, read_query, encoding_version, index_field, index_field_offset
+                );
+            }
+            return *truncate_range;
+        };
```
alexowens90, line 778: I think this is identical in `decode_into_frame_static`, can we factor it out into a separate function?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -803,8 +797,10 @@ void decode_into_frame_dynamic(
             auto& column = frame.column(static_cast<position_t>(dst_col));
             ColumnMapping mapping{frame, dst_col, field_col, context};
             check_mapping_type_compatibility(mapping);
-            mapping.set_truncate(truncate_range);
-            if (requires_truncation) {
+            const auto& column_truncation = truncate_range_for(mapping.dest_type_desc_);
+            mapping.set_truncate(column_truncation);
+
+            if (column_truncation.start_.has_value() || column_truncation.end_.has_value()) {
```
alexowens90: We can still use `column_truncation.requires_truncation()`?

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -379,6 +379,17 @@ std::optional<ArrowMeta::ColumnMeta> column_metadata(
...
+    auto column_arrays = arrow_arrays_from_column(column, name);
+    std::vector<std::shared_ptr<RecordBatchData>> output;
+    output.reserve(column_arrays.size());
```
alexowens90: I added a [utility](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/util/collection_utils.hpp#L62) for this very comment pattern

cpp/arcticdb/python/python_utils.cpp
alexowens90: This function name could now use an update

cpp/arcticdb/python/python_utils.cpp
```diff
@@ -26,14 +34,18 @@ py::tuple extract_numpy_arrays(PandasOutputFrame& pandas_output_frame) {
...
-    std::vector<py::array> arrays;
+    std::vector<PandasColumn> arrays;
     std::vector<std::string> index_column_names;
     std::vector<std::string> column_names;
     arrays.reserve(field_count);
     index_column_names.reserve(index_field_count);
     column_names.reserve(field_count - index_field_count);
     for (std::size_t c = 0; c < field_count; ++c) {
-        arrays.emplace_back(arcticdb::detail::array_at(frame, c));
+        if (frame.column(c).buffer().has_extra_bytes_per_block()) { // Checks if this column decoded as arrow string
```
alexowens90, line 44: This is a bit indirect, is there anything more explicit available?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -462,89 +448,81 @@ ColumnTruncation get_truncate_range_from_index(
 }
 
 ColumnTruncation get_truncate_range(
```
alexowens90, line 450: Is all of the delta in this function just an indentation change from the removal of the `if (read_options.output_format() == OutputFormat::ARROW) {` line?

python/arcticdb/version_store/_normalization.py
```diff
@@ -1067,6 +1068,41 @@ def _consolidate(self):
...
+def _arrow_string_arrays_to_pd_array(arrays):
+    dtype = pd.StringDtype(storage="pyarrow", na_value=np.nan)
+    if not arrays:
+        return pd.array([], dtype=dtype)
+    imported = [pa.Array._import_from_c(a.array(), a.schema()) for a in arrays]
+    return pd.array(pa.chunked_array(imported), dtype=dtype)
+
+
+def _adopt_arrow_strings(column):
+    return _arrow_string_arrays_to_pd_array(column) if _is_arrow_string_column(column) else column
```
alexowens90: This file is already super bloated, can we move all this stuff to it's own file? (similar to `dependencies.py`)

python/arcticdb/version_store/_normalization.py
```diff
@@ -1067,6 +1068,41 @@ def _consolidate(self):
         return self.blocks
 
 
+def _arrow_backed_str_dtype_supported():
+    # The pandas arrow-backed "str" dtype (StringDtype with na_value) was added in pandas 2.3. future.infer_string
+    # exists from pandas 2.1, so the option can be set on 2.1/2.2 but there this dtype cannot be constructed.
+    try:
+        pd.StringDtype(storage="pyarrow", na_value=np.nan)
```
alexowens90: Does this mean that pyarrow is now an optional dependency of Pandas, and required for Arrow-backed string support from their side as well as ours?

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -110,57 +112,37 @@ void finalize_segment_setup(
...
+                    // - Resampling with a date range with a bucket slice containing no indices
+                    // 0 sized memory blocks would break the offset assumptions in chunked buffers, and it is fine to
+                    // have number of memory blocks not equal number of segments because follow-up methods like
+                    // `copy_frame_data_to_buffer` rely on offsets rather than block indices.
+                    const auto bytes = block_row_count * data_size;
+                    column->allocate_and_advance_by(bytes);
+                }
             }
+        } else if (row_count > 0) {
+            column->allocate_and_advance_by(row_count * data_size);
```
alexowens90, line 139: Is this the same method that was being used previously when the column was constructed preallocated?

cpp/arcticdb/pipeline/read_frame.cpp
ognyanstoimenov: Yes

cpp/arcticdb/pipeline/read_frame.cpp
ognyanstoimenov: It uses the same call underneath : `add_detachable_block`. Before, that happenned at the constructor. now it happens explictily in this function during `buffer.ensure` [here](https://github.com/man-group/ArcticDB/blob/235cff2cc381bb0e963452ad9d626e0a0650c594/cpp/arcticdb/column_store/chunked_buffer.hpp#L236)

python/arcticdb/version_store/_normalization.py
ognyanstoimenov: yes

cpp/arcticdb/arrow/arrow_utils.cpp
ognyanstoimenov: I've changed this. It now contains a record batch with 1 column.

python/arcticdb/version_store/_store.py
ognyanstoimenov: Added an inline comment to explain

cpp/arcticdb/python/python_utils.cpp
ognyanstoimenov: I didn't find anything. Could thread the ReadOptions to here but I don't think it is worth it.

python/arcticdb/version_store/_normalization.py
ognyanstoimenov: Yes exactly

cpp/arcticdb/pipeline/input_frame.cpp
```diff
@@ -46,12 +46,28 @@ SortedValue compute_index_sortedness(const Column& index_column) {
 
 InputFrame::InputFrame() : index(stream::empty_index()) {}
 
+void InputFrame::set_from_tensors(StreamDescriptor&& desc, std::vector<NativeTensor>&& field_tensors) {
```
ognyanstoimenov, line 49: this method was just moved from the .hpp as it is no longer templated

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -254,69 +257,136 @@ void pandas_data_to_frame(const PandasData& pandas_data, const bool empty_types,
...
-        if (tensor.expanded_dim() == 1) {
-            desc.add_field(scalar_field(tensor.data_type(), col_names[i]));
-        } else if (tensor.expanded_dim() == 2) {
-            desc.add_field(FieldRef{TypeDescriptor{tensor.data_type(), Dimension::Dim1}, col_names[i]});
-        }
-        field_tensors.push_back(std::move(tensor));
+        util::variant_match(
+                col_vals[i],
+                [&](const py::array& arr) {
+                    auto tensor = [&] {
```
ognyanstoimenov, line 333: Pandas path identical to before, with the exception of the error handling, which handles the new variant type

python/arcticdb/version_store/_normalization.py
```diff
@@ -240,6 +243,8 @@ def _to_primitive(arr, arr_name, dynamic_strings, string_max_len=None, coerce_co
             norm_meta.common.categories[arr_name].category.extend(arr.categories)
         return arr.codes
 
+    if "str" in arr_dtype_as_str:
```
alexowens90: What if the "storage" of the `StringDtype` is `python`?

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -100,8 +101,6 @@ static std::tuple<char, int> parse_array_descriptor(PyObject* obj) {
 
 struct PyArrayDescriptor {
     PyArrayDescriptor(PyObject* ptr, bool empty_types) {
-        auto& api = pybind11::detail::npy_api::get();
-        util::check(api.PyArray_Check_(ptr), "Expected Python array");
```
alexowens90, line 104: Isn't this still needed?

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -254,69 +257,136 @@ void pandas_data_to_frame(const PandasData& pandas_data, const bool empty_types,
     );
 
     StreamDescriptor desc;
-    std::vector<entity::NativeTensor> field_tensors;
-    std::optional<entity::NativeTensor> opt_index_tensor;
+    std::vector<InputFrame::FieldData> columns;
+    columns.reserve(idx_names.size() + col_vals.size());
```
alexowens90: nit: `reserve_vector`

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -254,69 +257,136 @@ void pandas_data_to_frame(const PandasData& pandas_data, const bool empty_types,
...
+                    auto tensor = obj_to_tensor(arr.ptr(), empty_types);
+                    util::check(tensor.ndim() == 1, "Multi-dimensional indexes not handled");
+                    util::check(tensor.shape() != nullptr, "Index tensor expected to contain shapes");
+                    TypeDescriptor type{tensor.data_type(), Dimension::Dim0};
+                    auto rows = static_cast<size_t>(tensor.shape(0));
+                    return {std::move(tensor), type, rows};
+                },
+                [&](const std::vector<std::shared_ptr<RecordBatchData>>& chunks
+                ) -> std::tuple<InputFrame::FieldData, TypeDescriptor, size_t> {
+                    // Arrow-backed sring column
```
alexowens90: typo

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -254,69 +257,136 @@ void pandas_data_to_frame(const PandasData& pandas_data, const bool empty_types,
...
+                    std::vector<sparrow::record_batch> owners;
+                    owners.reserve(chunks.size());
```
alexowens90: reserve_vector

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -254,69 +257,136 @@ void pandas_data_to_frame(const PandasData& pandas_data, const bool empty_types,
...
+                    for (const auto& rbd : chunks)
+                        owners.emplace_back(std::move(rbd->array_), std::move(rbd->schema_));
+                    proto::descriptors::NormalizationMetadata::ExperimentalArrow arrow_meta;
+                    auto [cols, _] = record_batches_to_columns(owners, arrow_meta);
```
alexowens90: `auto cols = record_batches_to_columns(owners, arrow_meta).first;`

cpp/arcticdb/pipeline/read_frame.cpp
ognyanstoimenov: Hmm, it would need to carry state as well (for the lazy calculation) so it would have to be some sort of struct. Not sure if it is worth it?

cpp/arcticdb/python/python_to_tensor_frame.cpp
```diff
@@ -254,69 +257,136 @@ void pandas_data_to_frame(const PandasData& pandas_data, const bool empty_types,
...
+                    );
+                    frame.num_rows = std::max(frame.num_rows, static_cast<size_t>(cols[0].row_count()));
+                    desc.add_field(FieldRef{cols[0].type(), col_names[i]});
+                    columns.emplace_back(std::move(cols[0]));
+                    arrow_owners.insert(
+                            arrow_owners.end(),
+                            std::make_move_iterator(owners.begin()),
+                            std::make_move_iterator(owners.end())
+                    );
+                    has_only_tensors = false;
```
alexowens90, line 381: Could factor out this code with the duplicate in the index handling?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3078,9 +3078,17 @@ folly::Future<ReadVersionOutput> read_frame_for_version(
                 .via(&async::cpu_executor())
                 .thenValue([store,
                             read_query,
-                            read_options,
+                            read_options = read_options,
```
alexowens90, line 2945: I don't think this does anything?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3078,9 +3078,17 @@ folly::Future<ReadVersionOutput> read_frame_for_version(
...
                             res_versioned_item = std::move(res_versioned_item),
                             handler_data](auto&& pipeline_context) mutable {
+                    // TODO: Handle ndarray data when requesting ARROW output
+                    if (pipeline_context->has_normalization() && pipeline_context->normalization().has_np() &&
+                        read_options.output_format_for_frame() == OutputFormat::PANDAS) {
+                        read_options = read_options.clone();
```
alexowens90, line 2951: Why?

python/arcticdb/version_store/_normalization.py
ognyanstoimenov: I had that in the tests with [filtering](https://github.com/man-group/ArcticDB/blob/9fb2e957c0e2f622ad481bb31fb5a8cc21cdbfb3/python/tests/unit/arcticdb/test_string.py#L221) but [now](https://github.com/man-group/ArcticDB/blob/9fb2e957c0e2f622ad481bb31fb5a8cc21cdbfb3/python/tests/unit/arcticdb/test_string.py#L261) added it to pure read as well

python/tests/integration/arcticdb/version_store/test_basic_version_store.py
```diff
@@ -1614,8 +1621,12 @@ def test_dynamic_strings_with_all_nones(basic_store):
...
     basic_store.write("strings", df, dynamic_strings=True)
     data = basic_store.read("strings")
-    assert data.data["x"][0] is None
-    assert data.data["x"][1] is None
+    if _use_pyarrow_strings_in_pandas():
+        assert pd.isna(data.data["x"][0])
+        assert pd.isna(data.data["x"][1])
+    else:
+        assert data.data["x"][0] is None
+        assert data.data["x"][1] is None
```
alexowens90: I noticed the `StringDtype` has a `na_value` option. Out of curiosity, what impact does it have?

python/arcticdb/version_store/_normalization.py
ognyanstoimenov: Yeah, changed

python/tests/unit/arcticdb/version_store/test_ternary.py
```diff
@@ -132,6 +133,9 @@ def test_project_ternary_column_column_dynamic_strings(lmdb_version_store_v1, an
 
 
 @pytest.mark.skipif(WINDOWS, reason="We do not support fixed-width strings on Windows")
+@pytest.mark.skipif(
+    _use_pyarrow_strings_in_pandas(), reason="fixed-width strings not applicable under future.infer_string"
```
alexowens90: Why not?

python/tests/unit/arcticdb/test_string.py
```diff
@@ -214,6 +216,209 @@ def test_write_dynamic_simple(lmdb_version_store_v2):
...
+    with arrow_string_read(read_string_dtype):
+        col = lib.read("s").data["x"]
+    assert list(col.isna()) == [False, True, True, False]
+    assert col.iloc[0] == "a" and col.iloc[3] == "b"
+    assert np.isnan(col.iloc[2])
+    if read_string_dtype:
+        assert str(col.dtype) == "str"
+        assert np.isnan(col.iloc[1])
+    else:
+        assert col.iloc[1] is None
```
alexowens90: This checking for `None` with object dtype and `np.isnan/pd.isna` with string dtype seems to be repeated quite a lot, perhaps could be factored out to a utility?

python/tests/unit/arcticdb/test_string.py
```diff
@@ -214,6 +216,209 @@ def test_write_dynamic_simple(lmdb_version_store_v2):
...
+
+
+def test_none_vs_nan_null_distinction(lmdb_version_store_v2, read_string_dtype):
```
alexowens90, line 277: This seems to duplicate the test above

cpp/arcticdb/python/python_to_tensor_frame.cpp
ognyanstoimenov: I changed the `py::object` to `py::array` [here](https://github.com/man-group/ArcticDB/pull/3242/changes#diff-8e603385e60e53592eeb4c30c860df64938aae8f247f5c1a6c6bb24053e8e1e6) so pybind should reject anything that isn't an array

python/tests/unit/arcticdb/test_string.py
```diff
@@ -241,6 +446,10 @@ def test_mixed_types_errors(lmdb_version_store_v1, first_value, second_value):
     df = pd.DataFrame({"col": [first_value, second_value]})
     if first_value == second_value:
         pytest.skip()
+    if _use_pyarrow_strings_in_pandas() and {type(first_value), type(second_value)} == {StringInheritingClass, str}:
+        pytest.skip(
+            "future.infer_string coerces a str-subclass/plain-str mix to a clean str column, so no error is raised"
+        )
```
alexowens90, line 461: What a non-intuitive behaviour. So if `StringInheritingClass` had another field (say an `int`), that would just be silently dropped when creating a `StringDtype` column containing a `StringInheritingClass` object?

python/arcticdb/util/_versions.py
```diff
@@ -16,6 +16,8 @@
 IS_PANDAS_ZERO = PANDAS_VERSION < version.Version("1.0")
 IS_PANDAS_ONE = PANDAS_VERSION >= version.Version("1.0") and PANDAS_VERSION < version.Version("2.0")
 IS_PANDAS_TWO = PANDAS_VERSION >= version.Version("2.0")
+# The StringDtype(storage=..., na_value=...) arrow-backed "str" dtype was added in pandas 2.3.
+IS_PANDAS_TWO_THREE = PANDAS_VERSION >= version.Version("2.3")
```
alexowens90:
```suggestion
IS_AT_LEAST_PANDAS_TWO_THREE = PANDAS_VERSION >= version.Version("2.3")
```

python/arcticdb/util/test.py
```diff
@@ -296,6 +296,31 @@ def random_unicode_string(length: int) -> str:
...
 
+@contextmanager
+def arrow_string_read(enabled: bool):
+    """Read strings back as the pandas arrow-backed ``str`` dtype (``future.infer_string``) or as ``object``.
+
+    Wrap only the read call, not the write: writing that dtype is not supported yet. Skips when ``enabled`` is
+    requested on pandas that lacks either ``future.infer_string`` (< 2.1) or the ``str`` dtype (StringDtype na_value,
+    added in 2.3).
+    """
+    from arcticdb.version_store._string_dtype import _ARROW_BACKED_STR_DTYPE_SUPPORTED
```
alexowens90: nit: imports at top of file (a Claude classic I believe)

python/arcticdb/util/test.py
```diff
@@ -296,6 +296,31 @@ def random_unicode_string(length: int) -> str:
...
+
+        pytest.skip("pandas too old for the arrow-backed str dtype (StringDtype na_value, added in 2.3)")
+    try:
+        with pd.option_context("future.infer_string", enabled):
+            yield
+    except pd.errors.OptionError:
+        if enabled:
+            import pytest
+
+            pytest.skip("pandas too old for future.infer_string")
```
alexowens90: Is this reachable? I think the `_ARROW_BACKED_STR_DTYPE_SUPPORTED` guard already covers this?

python/tests/unit/arcticdb/test_string.py
```diff
@@ -214,6 +216,211 @@ def test_write_dynamic_simple(lmdb_version_store_v2):
...
+    assert list(received.index) == list(range(5))
+    assert list(received["x"]) == values[3:8]
+    assert (str(received["x"].dtype) == "str") == read_string_dtype
+
+
+def test_read_string_column_dtype(lmdb_version_store_v2, read_string_dtype, skip_consolidation):
+    if skip_consolidation:
+        lmdb_version_store_v2._normalizer.df.set_skip_df_consolidation()
+    values = ["Aaba", "A", "B", "C", "Baca", "CABA", "dog", "cat", "here is a very long one"]
+    lmdb_version_store_v2.write("strings", pd.DataFrame({"x": values}), dynamic_strings=True)
```
alexowens90: Because of what consolidation does, the test would be better if it had multiple adjacent string columns

python/arcticdb/version_store/_string_dtype.py
```diff
@@ -0,0 +1,60 @@
...
+    batches = []
+    for chunk in chunked.chunks:  # each chunk is a pa.Array (large_string)
+        record_batch = pa.RecordBatch.from_arrays([chunk], names=[str(arr_name)])
+        rbd = RecordBatchData()
+        record_batch._export_to_c(rbd.array(), rbd.schema())
+        batches.append(rbd)
+    return batches
+
+
+__all__ = [
```
alexowens90: We don't normally bother with this, particularly for private modules

cpp/arcticdb/pipeline/read_frame.cpp
alexowens90: Yeh fair enough

cpp/arcticdb/python/python_to_tensor_frame.cpp
alexowens90: I would just leave the check in, otherwise we should try and test the behaviour and that seems like more pain than its worth

python/arcticdb/version_store/_string_dtype.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import List
+
+import numpy as np
+import pandas as pd
+
+from arcticdb.dependencies import _PYARROW_AVAILABLE, pyarrow as pa
+from arcticdb.util._versions import IS_PANDAS_TWO_THREE
+from arcticdb_ext.version_store import RecordBatchData
+
+_ARROW_BACKED_STR_DTYPE_SUPPORTED = _PYARROW_AVAILABLE and IS_PANDAS_TWO_THREE
```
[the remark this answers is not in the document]
alexowens90: I think this is valid

alexowens90:
How experimental is this feature still from Pandas side? Default behaviour in `3.x` suggests fully supported in `3.x`, but [docs](https://pandas.pydata.org/docs/reference/api/pandas.StringDtype.html) still say experimental.
Also need to know if it is fully supported in `2.3.x` with the infer string option set.
Either way, need some docs to explain the different behaviours.

ognyanstoimenov:
As discussed offline several things are not implemented in this PR
* The `pandas` storage variant of the StringDType which is using the old `object` dtype underneath and is used as fallback when pyarrow dependancy is not installed - this will be done in a follow up PR
* Support for pd.NA null value - this is the old "string" dtype - we won't be adding support for that.
* Support for fixed-width strings with the new dtype - won't implement at this time

alexowens90 (approved):

PR #3244, opened 2026-07-14
Add xfail tests for inconsistencies between batch and non-batch methods

alexowens90:
While refactoring the modification methods `write/append/update/write_metadata` and their batch equivalents to remove code duplication and homogenise implementations, I came across various bugs and inconsistencies, that this PR captures through xfail tests.

### DIFF
--- a/python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
+++ b/python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
@@ -15,7 +15,7 @@ import sys
 from arcticdb import QueryBuilder
 from arcticdb.exceptions import UserInputException
 import arcticdb.toolbox.query_stats as qs
-from arcticdb.util.test import assert_frame_equal, assert_series_equal, config_context_multi
+from arcticdb.util.test import assert_frame_equal, assert_series_equal, config_context
 from arcticdb.version_store.library import Library
 from arcticdb_ext import set_config_int
 import arcticdb_ext.cpp_async as adb_async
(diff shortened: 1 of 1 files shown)


python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
```diff
@@ -579,3 +579,112 @@ def test_dynamic_schema_incompatible_types_do_not_orphan_data_keys(in_memory_sto
...
+        else lib.update(sym, pd.DataFrame({"col": [1]}, index=[pd.Timestamp(0)]), upsert=True)
+    )
+    lt = lib.library_tool()
+    # Batch writes a symbol list key
+    assert lt.count_keys(KeyType.SYMBOL_LIST) == 0
+
+
+@pytest.mark.parametrize(
+    "batch",
+    [pytest.param(True, marks=pytest.mark.xfail(reason="Latest live version is always loaded", strict=True)), False],
```
IvoDD, line 667: I'm not sure I understand the xfail reason. The comment below explainst that the non batch write also reads the lastest version.

How many version reads does the batch_write do?

IvoDD (approved):

python/tests/nonreg/arcticdb/version_store/test_nonreg_specific.py
alexowens90: The non-batch write loads the latest version including deleted versions, which is all that is needed.
The batch version loads the latest live version, so in this case loads the entire version chain.

PR #3265, opened 2026-07-28
Correctly filter against NaT as a timestamp rather than an int64 with the query builder

poodlewars:
Correctly filter against `NaT` as a timestamp rather than an int64 with the query builder, so `q["col"] > pd.NaT` will now return nothing (like Pandas). This is because we are now tracking that `pd.NaT` is a time type, whereas before we treated it as `int64`.

This also fixes the str() formatting of query builder expression nodes to have special formatting for strings, which regressed in PR #3201 and broke some internal tests. Before #3201 we used Python str representations of values to encode edges in the AST graph when we created it. These were reflected in the `str()` form of `ExpressionNode` objects. After that PR, no such strings are passed in to the C++ layer, what is presented to the user is all calculated in the C++ layer.

Note there are some limitations where we have still "regressed" from #3201 but I think minorly:

- We format the datetimes at nanosecond precision, whereas we used to strip trailing zeroes (in particular on dates)
- We don't have special formatting for `timedelta` objects

We also now format string values properly in error messages.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_query_builder_str.py
+++ b/python/tests/unit/arcticdb/version_store/test_query_builder_str.py
@@ -1,0 +1,171 @@
+"""
+Copyright 2026 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+import datetime
+import numpy as np
+import pandas as pd
+import pytest
+
+from arcticdb.version_store.processing import QueryBuilder, where
+
+pytestmark = pytest.mark.pipeline
+
+
+def test_to_strings():
+    q = QueryBuilder().row_range((1, 10))
+    assert str(q) == "ROWRANGE: RANGE, start=1, end=10"
+
+    q = QueryBuilder().head(10)
+    assert str(q) == "ROWRANGE: HEAD, n=10"
+
+    q = QueryBuilder().tail(9)
+    assert str(q) == "ROWRANGE: TAIL, n=9"
+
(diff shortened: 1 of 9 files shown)


python/tests/unit/arcticdb/version_store/test_query_builder_str.py
```diff
@@ -0,0 +1,171 @@
...
+import numpy as np
+import pandas as pd
+import pytest
+
+from arcticdb.version_store.processing import QueryBuilder, where
+
+pytestmark = pytest.mark.pipeline
+
+
+def test_to_strings():
```
poodlewars, line 19: This one is just moved over.

alexowens90 (approved):

poodlewars:
> ## ArcticDB Code Review Summary
> **Correctness**
> 
> * [ ]  ❌ Non-time numeric columns vs `pd.NaT`: the row-level fast path in `binary_comparator` keys off the _value_ type, but `stats_comparator(stats, value)` keys the same NaT rule off the _stats_ type, so the two paths can disagree. With `!=` against `NaT` on an int64/float column whose segment min == max == `INT64_MIN` (and no nulls), stats returns `NONE_MATCH` and prunes a segment whose rows the row-level path would all have matched. See the inline comment on `processing.py:1266`.
> 
> **Testing**
> 
> * [ ]  ❌ No coverage for comparing a **non-time** numeric column against `pd.NaT`, which is the case whose semantics changed most (previously an `INT64_MIN` numeric comparison, now NaT semantics).
> 
> **PR Description**
> 
> * [ ]  ❌ The description covers time columns but not the behaviour change for non-time numeric columns: `q["int_col"] > pd.NaT` (and `<`, `>=`, `<=`, `==`) now match nothing instead of comparing against `INT64_MIN`, and `!=` matches everything. Worth calling out explicitly since it is user-visible.

I'm ignoring these points - we have a ticket open to ban this mixed time -> int comparisons.

PR #3270, opened 2026-07-29
Enhancement 12036601732: support compact_data arg in append_batch

alexowens90:
#### Reference Issues/PRs
[12036601732](https://man312219.monday.com/boards/7852509418/pulses/12036601732)

Adds the `compact_data` argument to `append_batch/batch_append` methods.

### DIFF
--- a/cpp/arcticdb/version/version_store_api.cpp
+++ b/cpp/arcticdb/version/version_store_api.cpp
@@ -107,10 +107,13 @@ std::vector<std::variant<VersionedItem, DataError>> PythonVersionStore::batch_wr
 std::vector<std::variant<VersionedItem, DataError>> PythonVersionStore::batch_append(
         const std::vector<StreamId>& stream_ids, const std::vector<convert::InputItem>& items,
         const std::vector<py::object>& norms, const std::vector<py::object>& user_metas, bool prune_previous_versions,
-        bool validate_index, bool upsert, bool throw_on_error
+        bool validate_index, bool upsert, bool throw_on_error, bool compact_data
 ) {
     AppendOptions append_options{
-            .upsert = upsert, .prune_previous_versions = prune_previous_versions, .validate_index = validate_index
+            .upsert = upsert,
+            .prune_previous_versions = prune_previous_versions,
+            .validate_index = validate_index,
+            .compact_data = compact_data
     };
     auto frames = create_input_tensor_frames(
             stream_ids,
--- a/python/tests/integration/arcticdb/test_arctic_batch.py
+++ b/python/tests/integration/arcticdb/test_arctic_batch.py
@@ -690,6 +690,7 @@ def test_append_batch(library_factory, compact_data):
         assert_frame_equal(read_dataframe.data, original_dataframe)
 
 
+# See test_basic_version_store.py::test_batch_append_missing_keys for equivalent V1 API test
 @pytest.mark.storage
 @pytest.mark.parametrize("compact_data", [False, True])
 def test_append_batch_missing_keys(arctic_library, compact_data):
--- a/python/tests/integration/arcticdb/test_arctic_batch.py
+++ b/python/tests/integration/arcticdb/test_arctic_batch.py
@@ -642,7 +642,8 @@ def test_snapshots_with_delete_batch(arctic_library):
 
 
 @pytest.mark.storage
-def test_append_batch(library_factory):
+@pytest.mark.parametrize("compact_data", [False, True])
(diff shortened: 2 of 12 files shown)


python/tests/unit/arcticdb/version_store/test_append_compact_data.py
```diff
@@ -72,440 +77,535 @@ def generic_append_compact_data_test(lib, sym, df, **append_kwargs):
...
+    generic_append_compact_data_test(lib, sym, df[5:], coerce_columns={"col": object})
 
 
-def test_frequent_append_io_counts_compact_every_time(in_memory_store_factory, clear_query_stats):
-    lib = in_memory_store_factory()
-    sym = "test_frequent_append_io_counts_compact_every_time"
-    df = pd.DataFrame({"col": np.arange(200_000)}, index=pd.date_range("2026-01-01", freq="s", periods=200_000))
-    for idx in range(100):
+@pytest.mark.parametrize("batch", [False, True])
+class TestAppendCompactData:
```
alexowens90, line 90: All the tests in this class already existed, they have just been parametrized to test both batch and non-batch

cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -107,10 +107,13 @@ std::vector<std::variant<VersionedItem, DataError>> PythonVersionStore::batch_wr
 std::vector<std::variant<VersionedItem, DataError>> PythonVersionStore::batch_append(
         const std::vector<StreamId>& stream_ids, const std::vector<convert::InputItem>& items,
         const std::vector<py::object>& norms, const std::vector<py::object>& user_metas, bool prune_previous_versions,
-        bool validate_index, bool upsert, bool throw_on_error
+        bool validate_index, bool upsert, bool throw_on_error, bool compact_data
```
poodlewars, line 110: 5 bools in a row :(

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -732,10 +735,13 @@ def test_append_batch_missing_keys(arctic_library):
     assert_frame_equal(read_dataframe.data, pd.concat([df2_write, df2_append]))
 
 
-def test_append_batch_empty_dataframe_increases_version(lmdb_version_store_v1):
-    lib = lmdb_version_store_v1
-    lib.batch_write(["sym1", "sym2"], [pd.DataFrame({"a": [1, 2, 3]}), pd.DataFrame({"b": [1, 2, 3, 4]})])
-    lib_tool = lib.library_tool()
+@pytest.mark.parametrize("compact_data", [False, True])
+def test_append_batch_empty_dataframe_increases_version(lmdb_library, compact_data):
```
poodlewars, line 740: Is there a similar test for append non-batch?

poodlewars (commented):
Claude's correctness comment is worth adding a test for.

The Claude performance comment I suppose would be addressed by a batch version of my "admission handler" work. I can look at that after its singular form is done.

I think it would be worthwhile to have some ASV coverage of these APIs?

poodlewars (approved):

alexowens90:
> Claude's correctness comment is worth adding a test for.
> 
> The Claude performance comment I suppose would be addressed by a batch version of my "admission handler" work. I can look at that after its singular form is done.
> 
> I think it would be worthwhile to have some ASV coverage of these APIs?

Yeh I'll sort out the exception type (although it looks like the current implementation is also probably inconsistent between V1 and V2 APIs).

Yeh general memory backpressure should be handled in one place by an admission handler I think.

ASV benchmarks to follow in a separate PR once the APIs all exist.

cpp/arcticdb/version/version_store_api.cpp
alexowens90: I agree its a bit sad (although consistent with our other Python-exposed methods). I could expose `AppendOptions` to Python, but that will just result in 4 bools in a row in the Python layer, which I don't think is a vast improvement.

python/tests/integration/arcticdb/test_arctic_batch.py
alexowens90: Yep https://github.com/man-group/ArcticDB/blob/master/python/tests/unit/arcticdb/version_store/test_append.py#L803

cpp/arcticdb/version/version_store_api.cpp
poodlewars: There's still no need for the bools to leak out of our python bindings even if we leave this signature in the Python layer?

PR #3280, opened 2026-07-30
Benchmark 12499851988: Add ASV benchmarks for compact_data argument to append and append_batch

alexowens90:
#### Reference Issues/PRs
[12499851988](https://man312219.monday.com/boards/7852509418/pulses/12499851988)

### DIFF
--- a/python/benchmarks/compact_data.py
+++ b/python/benchmarks/compact_data.py
@@ -14,7 +14,7 @@ import numpy as np
 import pandas as pd
 import random
 
-from arcticdb import Arctic, LibraryOptions
+from arcticdb import Arctic, LibraryOptions, WritePayload
 from arcticdb.options import ModifiableLibraryOption
 from arcticdb.util.logger import get_logger
 from arcticdb.util.test import random_strings_of_length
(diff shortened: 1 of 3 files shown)


python/benchmarks/modification_functions.py
```diff
@@ -40,9 +40,10 @@ class ModificationFunctions:
 
     rows_and_cols = [(1_000_000, 2), (10_000_000, 2), (5_000, 30_000)]
     storages = [Storage.LMDB, Storage.AMAZON]
-    param_names = ["rows_and_cols", "storage"]
+    compact_data = [False, True]
+    param_names = ["rows_and_cols", "storage", "compact_data"]
```
vasil-pashov, line 44: When update gets implemented this will double the time during testing. I don't know if it's a problem at the moment but in future if it becomes too slow we might trim it down further or create a specific class testing for compaction.

python/benchmarks/compact_data.py
```diff
@@ -254,3 +254,208 @@ def time_compact_data(self, row_params, num_columns):
...
+        self.logger.info(f"SETUP_CACHE TIME: {time.time() - start}")
+
+    def _setup_cache(self):
+        # Populate the base libraries only once
+        ac = Arctic(self.CONNECTION_STRING_BASE)
+        num_rows = 1_000_000
+        df = pd.DataFrame({f"col_{i}": np.arange(i * num_rows, (i + 1) * num_rows) for i in range(self.NUM_COLUMNS)})
+        for existing_data_fragmented in self.params[1]:
+            if existing_data_fragmented:
+                dfs = [df[i * 10_000 : (i + 1) * 10_000] for i in range(num_rows // 10_000)]
```
vasil-pashov, line 345: 1. Does the fact that the segment size is divisible by the size of the "fragmented" segments? Will things change if we pick something more irregular?
2. Does it make sense to benchmark smaller fragments like 1 row. I imagine people might start appending one row and occasionally perform a compaction.

vasil-pashov (approved):

python/benchmarks/modification_functions.py
alexowens90: We're not planning to implement `compact_data` for `update` (at least in the near future).
I think all of this stuff needs separating out and tidying up, see slack thread [here](https://chat-man.slack.com/archives/C02NEG3V38X/p1785405223301909)

python/benchmarks/compact_data.py
alexowens90: 1. It does, but funkier merge/split combinations are benchmarked for the explicit `compact_data` method rather than duplicated here
2. I'm expecting most usage to always have the flag set

PR #3281, opened 2026-07-30
Reject mixed timestamp/numeric types in QueryBuilder expressions

poodlewars:
#### Reference Issues/PRs
Fixes 8065794446.

`QueryBuilder` expressions that mixed a timestamp (`pd.Timestamp`/`datetime64` column) with a plain numeric type (`int`, `uint`, `float`) previously either silently compared/computed against the timestamp's raw nanosecond-since-epoch integer, or raised a confusing, unrelated error. This PR makes all mismatches raise a `UserInputException`:

```
Timestamp and numeric types cannot be mixed
```

This affects comparisons, `isin`/`isnotin`, `where(...)` (ternary), and arithmetic, in any
combination of column/column, column/value, or value/value operands. The one exception is
adding or subtracting a plain integer to/from a timestamp, which is unchanged and still treated
as a nanosecond offset (see below), because ArcticDB has no separate timedelta type.

##### Comparisons and filtering

Before, this would silently compare the raw nanosecond count against `5`, filtering on an
almost-certainly unintended condition, rather than raising:

```python
q = QueryBuilder()
q = q[q["timestamp_col"] > 5]
lib.read("sym", query_builder=q)
```

Now this raises:

```
UserInputException: ... timestamp_col is of type NANOSECONDS_UTC64 and cannot be compared to 5,
which is of type INT64. Timestamp and numeric types cannot be mixed.
```

The same applies to `isin`/`isnotin` when the value set mixes timestamps and plain numbers:

```python
q = q[q["timestamp_col"].isin([pd.Timestamp("2024-01-01"), 5])]  # now raises
```

##### Projections (`where`, arithmetic)

```python
q = q.apply("result", where(q["cond"], q["timestamp_col"], 0))       # now raises
q = q.apply("result", q["timestamp_col"] * 2)                        # now raises
q = q.apply("result", q["timestamp_col"] / 2)                        # now raises
```

Two timestamp columns/values on both sides of a `where(...)` continue to work, and now correctly
produce a `datetime64[ns]` result column (previously this path could produce an `int64` column
instead, since a timestamp's underlying raw type is `int64`):

```python
q = q.apply("result", where(q["cond"], q["dt1"], q["dt2"]))
# result column dtype: datetime64[ns]
```

##### Timestamp +/- integer offsets keep working

Adding or subtracting an integer to/from a timestamp column or value is unaffected by this
change and continues to be treated as a nanosecond offset (there is no distinct timedelta type
in ArcticDB's processing pipeline), and the result is correctly typed as a timestamp rather than
an integer:

```python
q = q.apply("result", q["timestamp_col"] + 1)          # result: timestamp_col + 1 nanosecond
q = q.apply("result", q["timestamp_col"] - q["int_col"])  # result: timestamp column
```

Subtracting two timestamp columns/values is also unaffected, and returns an `int64` nanosecond
difference (as before), since there is no timedelta type to represent the result:

```python
q = q.apply("result", q["dt_col_a"] - q["dt_col_b"])   # result: int64 nanoseconds
```

An integer minus a timestamp still raises, matching pandas' behaviour for the equivalent
`Series` operation (a `TypeError` in pandas):

```python
q = q.apply("result", 5 - q["timestamp_col"])          # now raises
```

##### Unary operations on timestamp columns

`abs()` and unary negation on a timestamp column now raise a clearer message
(`Cannot perform unary operation ...`) rather than the previous, less specific wording. This was
already rejected before; only the message text has changed.

### DIFF
--- a/cpp/arcticdb/entity/types.hpp
+++ b/cpp/arcticdb/entity/types.hpp
@@ -123,8 +123,10 @@ constexpr bool is_sequence_type(ValueType v) {
 constexpr bool is_time_type(ValueType v) { return uint8_t(v) == uint8_t(ValueType::NANOSECONDS_UTC); }
 
 constexpr bool is_numeric_type(ValueType v) {
-    return is_time_type(v) || (uint8_t(v) >= uint8_t(ValueType::UINT) && uint8_t(v) <= uint8_t(ValueType::FLOAT));
+    return uint8_t(v) >= uint8_t(ValueType::UINT) && uint8_t(v) <= uint8_t(ValueType::FLOAT);
 }
+
+constexpr bool is_numeric_or_time_type(ValueType v) { return is_numeric_type(v) || is_time_type(v); }
 
 constexpr bool is_floating_point_type(ValueType v) { return uint8_t(v) == uint8_t(ValueType::FLOAT); }
 
@@ -249,6 +251,10 @@ constexpr bool is_sequence_type(DataType v) { return is_sequence_type(slice_valu
 
 constexpr bool is_numeric_type(DataType v) { return is_numeric_type(slice_value_type(v)); }
 
+constexpr bool is_time_type(DataType v) { return is_time_type(slice_value_type(v)); }
+
+constexpr bool is_numeric_or_time_type(DataType v) { return is_numeric_type(v) || is_time_type(v); }
+
 constexpr bool is_bool_type(DataType dt) { return slice_value_type(dt) == ValueType::BOOL; }
 
 constexpr bool is_bool_object_type(DataType dt) { return slice_value_type(dt) == ValueType::BOOL_OBJECT; }
@@ -258,8 +264,6 @@ constexpr bool is_unsigned_type(DataType dt) { return slice_value_type(dt) == Va
 constexpr bool is_signed_type(DataType dt) { return slice_value_type(dt) == ValueType::INT; }
 
 constexpr bool is_floating_point_type(DataType v) { return is_floating_point_type(slice_value_type(v)); }
-
-constexpr bool is_time_type(DataType v) { return is_time_type(slice_value_type(v)); }
 
 constexpr bool is_integer_type(DataType v) { return is_integer_type(slice_value_type(v)); }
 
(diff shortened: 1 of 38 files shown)


cpp/arcticdb/entity/types.hpp
```diff
@@ -122,10 +122,12 @@ constexpr bool is_sequence_type(ValueType v) {
 
 constexpr bool is_time_type(ValueType v) { return uint8_t(v) == uint8_t(ValueType::NANOSECONDS_UTC); }
 
-constexpr bool is_numeric_type(ValueType v) {
+constexpr bool is_numeric_or_time_type(ValueType v) {
     return is_time_type(v) || (uint8_t(v) >= uint8_t(ValueType::UINT) && uint8_t(v) <= uint8_t(ValueType::FLOAT));
 }
 
+constexpr bool is_numeric_type(ValueType v) { return is_numeric_or_time_type(v) && !is_time_type(v); }
```
poodlewars: I started by renaming this to `is_numeric_or_time_type` and compiling to make sure all call sites were updated.

I then introduced this new implementation of the function and updated three places to `is_numeric_type` deliberately:

- expression_node.cpp#274
- operation_dispatch_unary.hpp#43#72

to stop unary operators (ie ABS) on timestamps deliberately.

It also survives in sorted_aggregation.cpp#400, but that had already excluded the time case.

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -335,12 +339,13 @@ std::variant<BitSetTag, DataType> ExpressionNode::compute(
...
             );
-            details::visit_type(std::get<DataType>(left_type), [operation_type_, &res, right_type](auto left_tag) {
+            details::visit_type(std::get<DataType>(left_type), [&](auto left_tag) {
                 using left_type_info = ScalarTypeInfo<decltype(left_tag)>;
-                details::visit_type(std::get<DataType>(right_type), [operation_type_, &res](auto right_tag) {
+                details::visit_type(std::get<DataType>(right_type), [&](auto right_tag) {
                     using right_type_info = ScalarTypeInfo<decltype(right_tag)>;
-                    if constexpr (is_numeric_type(left_type_info::data_type) &&
-                                  is_numeric_type(right_type_info::data_type)) {
+                    if constexpr (numeric_or_time_types_compatible(
```
IvoDD, line 346: This does still end up allowing `time * time`, `time ** time` and `time / time` all of which don't really make sense.

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -504,12 +587,13 @@ VariantData binary_operator(const Value& left, const Value& right, Func&& func)
...
             auto right_value = right.get<typename right_type_info::RawType>();
             auto left_value = left.get<typename left_type_info::RawType>();
-            using TargetType = typename binary_operation_promoted_type<
-                    typename left_type_info::RawType,
-                    typename right_type_info::RawType,
-                    std::remove_reference_t<Func>>::type;
             *output_value =
-                    Value{TargetType{func.apply(left_value, right_value)}, data_type_from_raw_type<TargetType>()};
+                    Value{TargetType{func.apply(left_value, right_value)},
+                          binary_arithmetic_output_type<
```
IvoDD, line 592: This allows `time * time` and the result is `int64`. Instead we should definetely raise for `*`, `/`, `**`.

I agree we should allow `time - time = int64`.

I am not so sure however about `time + time`. I can definetely see someone doing `(time + time) / 2` and expecting it to work. In this case however we would return an `int` and not a `time` which is also surprising.

I am in favor of raising for `time + time` and forcing users who want the average of two times to do `time1 + (time2 - time1) / 2` instead.

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -527,19 +611,38 @@ VariantData binary_operator(const ColumnWithStrings& left, const ColumnWithStrin
...
         details::visit_type(right.column_->type().data_type(), [&](auto right_tag) {
             using right_type_info = ScalarTypeInfo<decltype(right_tag)>;
-            if constexpr (!is_numeric_type(left_type_info::data_type) || !is_numeric_type(right_type_info::data_type)) {
+            using TargetType = typename binary_operation_promoted_type<
+                    typename left_type_info::RawType,
+                    typename right_type_info::RawType,
+                    std::remove_reference_t<decltype(func)>>::type;
+            constexpr bool offset_arithmetic =
+                    is_offset_arithmetic<Func, left_type_info::data_type, right_type_info::data_type, TargetType>;
+            if constexpr (is_time_numeric_mismatch(left_type_info::data_type, right_type_info::data_type) &&
```
IvoDD, line 620: There is some duplication between the different `binary_operator` variants.

cpp/arcticdb/processing/operation_dispatch_ternary.cpp
```diff
@@ -261,13 +275,16 @@ VariantData ternary_operator(const util::BitSet& condition, const ColumnWithStri
...
                                  (is_bool_type(col_type_info::data_type) && is_bool_type(val_type_info::data_type))) {
                 using TargetType = typename ternary_operation_promoted_type<
                         typename col_type_info::RawType,
                         typename val_type_info::RawType>::type;
-                constexpr auto output_data_type = data_type_from_raw_type<TargetType>();
+                // timestamp is int64_t, so data_type_from_raw_type would give INT64 for a time column and time value
+                constexpr auto output_data_type =
+                        is_time_type(col_type_info::data_type) && is_time_type(val_type_info::data_type)
+                                ? DataType::NANOSECONDS_UTC64
+                                : data_type_from_raw_type<TargetType>();
```
IvoDD, line 287: Again there is some duplication between the different overloads. Here it is not so bad though. I think just a single method for the output type would be enough to abstract the:
```
                constexpr auto output_data_type =
                        is_time_type(col_type_info::data_type) && is_time_type(val_type_info::data_type)
                                ? DataType::NANOSECONDS_UTC64
                                : data_type_from_raw_type<TargetType>();
```

cpp/arcticdb/processing/sorted_aggregation.cpp
```diff
@@ -397,7 +397,7 @@ void SortedAggregator<aggregation_operator, closed_boundary>::check_aggregator_s
 ) const {
     schema::check<ErrorCode::E_UNSUPPORTED_COLUMN_TYPE>(
             (is_time_type(data_type) && aggregation_operator != AggregationOperator::SUM) ||
-                    (is_numeric_type(data_type) && !is_time_type(data_type)) || is_bool_type(data_type) ||
+                    is_numeric_type(data_type) || is_bool_type(data_type) ||
```
IvoDD, line 400: I think we disallowed `MIN` and `MAX` for time types here which I think is a mistake.

python/arcticdb/version_store/processing.py
```diff
@@ -1282,6 +1298,9 @@ def _visit_child(node):
                 return _ExpressionNode(_ColumnName(node.left))
             return _visit(node)
         elif isinstance(node, (np.ndarray, list)):
+            # "M" is numpy's dtype kind code for datetime64
+            if isinstance(node, np.ndarray) and node.dtype.kind == "M":
+                return _ExpressionNode(_ValueSet.from_nanoseconds(node.view("int64")))
```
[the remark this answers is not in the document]
IvoDD: +1 to this claude comment

cpp/arcticdb/processing/sorted_aggregation.cpp
poodlewars: There's no logical change here? I'll just check what test coverage we have for the non-SUM aggregators of time columns.

PR #3282, opened 2026-07-30
Implement upsert and write new version even when source is empty

vasil-pashov:
#### Reference Issues/PRs
Upsert: Monday 18041316370
Metadata: Monday 12518592426

**Upsert** When the strategy allows insertion and the symbol does not exist it will be created and filled with the source data. If the symbol does not exist, upsert is true and the strategy is update only an exception is thrown.

**Metadata** Increase the version and store metadata even when empty source dataframes are passed. Similar to append/update/write.
#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -2492,18 +2492,17 @@ VersionedItem LocalVersionedEngine::merge_internal(
     const bool add_new_symbol_list_entry = !update_info.previous_index_key_.has_value() && cfg().symbol_list();
     if (update_info.previous_index_key_.has_value()) {
         const ReadOptions read_options;
-        const WriteOptions write_options = get_write_options();
         index_key_fut = source->empty() ? async_write_metadata_impl(store(), update_info, std::move(source->user_meta))
                                         : merge_update_impl(
                                                   store(),
                                                   update_info,
                                                   read_options,
-                                                  write_options,
+                                                  write_options_,
                                                   IndexPartialKey{stream_id, update_info.next_version_id_},
                                                   std::move(on),
                                                   strategy,
                                                   std::move(source),
-                                                  get_de_dup_map(stream_id, update_info, write_options)
+                                                  get_de_dup_map(stream_id, update_info, write_options_)
                                           );
     } else if (upsert) {
         user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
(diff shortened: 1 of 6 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -2461,39 +2461,57 @@ VersionedItem LocalVersionedEngine::merge_internal(
...
-                        IndexPartialKey{.id = stream_id, .version_id = update_info.next_version_id_},
-                        std::move(on),
-                        strategy,
-                        std::move(source),
-                        get_de_dup_map(stream_id, update_info, write_options_)
-                )
-                        .get();
-        write_version_and_prune_previous(prune_previous_versions, versioned_item.key_, update_info.previous_index_key_);
-        return versioned_item;
+        const WriteOptions write_options = get_write_options();
```
alexowens90: This can be removed, and just use `write_options_` below

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -1782,52 +1688,64 @@ def test_slicing(self, lmdb_library_factory, monkeypatch, slicing_policy):
...
+            merge_vit = lib.merge_experimental("sym", source, strategy=strategy, upsert=upsert, metadata=metadata)
+            assert merge_vit.version == 0
+            assert merge_vit.metadata == metadata
+            assert lib.list_symbols() == ["sym"]
             read_vit = lib.read("sym")
             assert_vit_equals_except_data(merge_vit, read_vit)
-            assert_frame_equal(read_vit.data, expected)
+            assert_frame_equal(read_vit.data, source)
+
+    def test_upsert_with_existing_symbol(self, lmdb_library):
```
alexowens90, line 1712: This test is (basically) duplicated

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -1782,52 +1688,64 @@ def test_slicing(self, lmdb_library_factory, monkeypatch, slicing_policy):
     )
     @pytest.mark.parametrize("upsert", (True, False))
     @pytest.mark.parametrize("strategy", (MergeStrategy("update", "insert"), MergeStrategy("do_nothing", "insert")))
```
alexowens90, line 1690: Could also test that `MergeStrategy("update", "do_nothing")` raises

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_merge_update.py
vasil-pashov: There's already such a test in the update suite: `TestMergeTimeseriesUpdate::test_target_symbol_does_not_exist`

python/tests/unit/arcticdb/version_store/test_merge_update.py
vasil-pashov: They are slightly different. I've structured tests in suites based on strategy and index. So the other `test_upsert_with_existing_symbol` is for the `update` strategy. I agree the difference is small though it's in the output as this tests that both update and insert happened while the other tests that only update happened.

When I started writing the tests the choice was between some duplication with clear separation of concerns vs. less code duplication bug heavy on the parametrization which leads to what I found to be hared to read code. I went with the former.

PR #3284, opened 2026-07-31
Support writing metadata with a storage lock, for better tracing

poodlewars:
Support writing user defined metadata (similar to snapshot metadata) with an old-style storage lock. Add APIs to read that metadata, and to list out all the locks and their metadata.

This is to help tracing down what wrote storage locks, for example if an uexpected lock causes liveness issues. We anticipate users may want to write a job name, launch command, IP address, environment metadata, etc in this metadata.

Metadata is normalized the same way as symbol metadata, but capped at 1MB (rather than the general 16MB user-metadata limit), since the locking mechanism relies on reads and writes happening much faster than the pre-emption window.

Marked no release notes as this is not part of our published tooling.

See internal thread: https://chat-man.slack.com/archives/CKD4V6N0H/p1785431206969279

### DIFF
--- a/cpp/arcticdb/toolbox/library_tool.cpp
+++ b/cpp/arcticdb/toolbox/library_tool.cpp
@@ -175,14 +175,9 @@ std::vector<StorageLockInfo> LibraryTool::list_storage_locks() {
         auto name = fmt::format("{}", variant_key_id(key));
         try {
             auto kv = store()->read_sync(key, storage::ReadKeyOpts{.dont_warn_about_missing_key = true});
-            auto acquire_time = kv.second.template scalar_at<timestamp>(0, 0).value();
-            std::optional<google::protobuf::Any> metadata;
-            if (const auto* meta = kv.second.metadata()) {
-                metadata = *meta;
-            }
-            res.push_back(
-                    StorageLockInfo{std::move(name), acquire_time, (now - acquire_time) < ttl, std::move(metadata)}
-            );
+            auto data = extract_lock_segment_data(kv.second);
+            auto active = lock_is_active(data.acquire_time, now, ttl);
+            res.push_back(StorageLockInfo{std::move(name), data.acquire_time, active, std::move(data.metadata)});
         } catch (const storage::KeyNotFoundException&) {
         }
     });
--- a/cpp/arcticdb/toolbox/library_tool.cpp
+++ b/cpp/arcticdb/toolbox/library_tool.cpp
@@ -22,7 +22,6 @@
 #include <arcticdb/entity/protobuf_mappings.hpp>
 #include <arcticdb/python/adapt_read_dataframe.hpp>
 #include <arcticdb/util/storage_lock.hpp>
-#include <arcticdb/util/reliable_storage_lock.hpp>
 #include <cstdlib>
 
 namespace arcticdb::toolbox::apy {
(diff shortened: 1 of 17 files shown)


cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -29,10 +30,15 @@ inline StreamDescriptor lock_stream_descriptor(const StreamId& stream_id) {
     };
 }
 
-SegmentInMemory lock_segment(const StreamId& name, uint64_t timestamp) {
+SegmentInMemory lock_segment(
+        const StreamId& name, uint64_t timestamp, std::optional<google::protobuf::Any> metadata = std::nullopt
```
alexowens90, line 34: Presumably the metadata isn't mutated by this class, could it be passed by `const&` everywhere to avoid any accidental copies if a `std::move` is missed?

cpp/arcticdb/util/storage_lock.hpp
```diff
@@ -110,25 +123,44 @@ class StorageLock {
...
     }
 
+    // Returns the current holder's user-defined metadata, or nullopt if there is no active lock or it carries no
+    // metadata. Readable regardless of ownership, so it can be used to trace who holds the lock.
+    std::optional<google::protobuf::Any> read_metadata(const std::shared_ptr<Store>& store) const {
+        if (!exists_active_lock(store)) {
+            return std::nullopt;
+        }
+        try {
+            auto key_seg = store->read_sync(ref_key(), {.dont_warn_about_missing_key = true});
```
alexowens90, line 153: `exists_active_lock` already reads this key, we should refactor so it's only read once

cpp/arcticdb/toolbox/library_tool.cpp
```diff
@@ -165,6 +166,30 @@ std::vector<VariantKey> LibraryTool::find_keys_for_id(entity::KeyType kt, const
...
+
+    store()->iterate_type(KeyType::LOCK, [&](VariantKey&& key) {
+        auto name = fmt::format("{}", variant_key_id(key));
+        try {
+            auto kv = store()->read_sync(key, storage::ReadKeyOpts{.dont_warn_about_missing_key = true});
+            auto acquire_time = kv.second.template scalar_at<timestamp>(0, 0).value();
+            std::optional<google::protobuf::Any> metadata;
+            if (const auto* meta = kv.second.metadata()) {
+                metadata = *meta;
+            }
```
alexowens90: There's some duplicated code here with `storage_lock.hpp`, could be factored out into some free functions and reused

python/arcticdb/toolbox/library_tool.py
```diff
@@ -67,6 +71,24 @@ def dataframe_to_keys(
     def find_keys_for_symbol(self, key_type: KeyType, id: Union[str, int]) -> Union[List[AtomKey], List[RefKey]]:
         return self.find_keys_for_id(key_type, id)
 
+    def get_storage_lock(self, name: str):
+        """Return a dict-in/dict-out :class:`~arcticdb.toolbox.storage_lock.StorageLock` for the given lock name."""
+        from arcticdb.toolbox.storage_lock import StorageLock
```
alexowens90: Why here and not at top of file?

cpp/arcticdb/toolbox/python_bindings.cpp
```diff
@@ -16,9 +16,32 @@
...
+}
+
+py::object any_metadata_to_py(const std::optional<google::protobuf::Any>& metadata) {
+    if (!metadata.has_value()) {
+        return py::none();
+    }
+    arcticdb::proto::descriptors::UserDefinedMetadata udm;
+    metadata->UnpackTo(&udm);
+    return python_util::pb_to_python(udm);
+}
```
alexowens90: Could go in `python_utils.hpp`

python/arcticdb/toolbox/storage_lock.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Metadata passed at acquire time is stored on the lock and can be read back by any process via
+    ``read_metadata`` to trace which job holds the lock.
+    """
+
+    def __init__(self, ext_lock: _StorageLockImpl):
+        self._ext_lock = ext_lock
+
+    def lock(self, metadata: Optional[Mapping[str, Any]] = None) -> None:
+        self._ext_lock.lock(normalize_metadata(metadata))
```
alexowens90: As discussed we should cap the size of this metadata relatively small

python/tests/integration/arcticdb/test_storage_lock.py
```diff
@@ -82,3 +84,106 @@ def test_many_increments(real_storage_factory, lib_name, num_processes, max_slee
...
+        # No lock yet
+        assert reader.read_metadata() is None
+
+        # Locking without metadata reports no metadata
+        writer.lock()
+        assert reader.read_metadata() is None
+        writer.unlock()
+
+
+def test_list_storage_locks_unreliable(mem_library):
```
alexowens90, line 224: One test listing out multiple locks with and without metdata would be good

python/arcticdb/toolbox/storage_lock.py
poodlewars: I've had a look at PUT performance on VAST at different metadata sizes,

```
| metadata | median | min  | p95  | max   |
|---------:|-------:|-----:|-----:|------:|
| 2 MB     | 90.5   | 80.7 | 109.9| 112.9 |
| 3 MB     | 96.9   | 76.3 | 109.4| 116.3 |
| 4 MB     | 102.3  | 81.3 | 113.9| 114.8 |
| 5 MB     | 94.2   | 63.5 | 114.9| 314.9 |
| 6 MB     | 101.4  | 89.8 | 108.4| 108.4 |
| 7 MB     | 115.0  | 107.8| 128.5| 130.2 |
| 8 MB     | 126.1  | 103.2| 139.1| 159.7 |
| 9 MB     | 125.9  | 104.2| 158.5| 195.1 |
| 10 MB    | 139.1  | 97.7 | 160.0| 164.9 |
| 11 MB    | 129.5  | 104.9| 165.4| 167.8 |
| 12 MB    | 117.2  | 96.3 | 133.5| 136.6 |
| 13 MB    | 127.7  | 104.9| 157.4| 224.4 |
| 14 MB    | 133.7  | 106.6| 176.6| 178.7 |
| 15 MB    | 133.4  | 116.9| 176.4| 397.8 |
```

and GET performance,

```
| metadata | median | min | p95 | max |
|---------:|-------:|----:|----:|----:|
| none     | 6.5    | 4.3 | 8.4 | 8.4  |
| 1 MB     | 6.8    | 5.3 | 8.7 | 9.0  |
| 2 MB     | 11.3   | 9.3 | 13.1| 14.1 |
| 3 MB     | 10.1   | 8.1 | 11.9| 13.2 |
| 4 MB     | 8.5    | 6.8 | 13.2| 17.0 |
| 5 MB     | 7.7    | 6.3 | 10.8| 10.8 |
| 6 MB     | 10.9   | 9.6 | 12.5| 13.9 |
| 7 MB     | 10.7   | 9.1 | 11.9| 11.9 |
| 8 MB     | 10.0   | 8.9 | 11.7| 12.4 |
| 9 MB     | 13.0   | 9.1 | 14.7| 20.2 |
| 10 MB    | 13.4   | 11.0| 15.9| 19.4 |
| 11 MB    | 15.6   | 13.7| 18.0| 19.0 |
| 12 MB    | 15.1   | 12.6| 16.5| 16.6 |
| 13 MB    | 15.2   | 12.8| 20.3| 25.0 |
| 14 MB    | 24.0   | 21.5| 29.4| 32.0 |
| 15 MB    | 26.0   | 17.1| 33.3| 35.0 |
| ~16 MB   | 22.7   | 15.1| 25.0| 31.8 |
```

The logic is to PUT the lock, then wait `StorageLock.WaitMs` (default 1000), then check we still have the lock.

The case where the lock is most likely to fail is,

- T1: Writer A submits PUT
- T2: Writer A PUT completes
- T3: Writer A 1000ms sleep starts
- T4: Writer B submits PUT
- T5: Writer A sleep ends, submits and completes GET
- T6: Writer B PUT completes

I don't see such differences in PUT performance within the currently allowed metadata sizes that it's worth restricting further in this case.

poodlewars: CApped at 1MB rather than 16MB

alexowens90 (approved):

cpp/arcticdb/util/storage_lock.hpp
poodlewars: We can't do that, because we have ```        output.set_metadata(std::move(*metadata));``` which takes an rvalue reference

PR #3287, opened 2026-08-01
Fix date_range/row_range reads retaining full segment buffers (OOM)

### LINKED ISSUE #2348: Unexpected memory retention when reading slices of dataframes
### Describe the bug

When reading and slicing a subset of a large DataFrame:

1. The entire DataFrame appears to be loaded into memory.
2. A slice is taken and returned, likely as a view retaining a reference to the original.
3. If this operation is repeated in a loop and each slice is stored (e.g., in a list), the original large DataFrames are never deallocated.

This causes cumulative memory usage to increase continuously, eventually leading to an out-of-memory crash.

#### 1. Create demo data

```python
import pandas as pd
import numpy as np
from datetime import datetime


def generate_random_dataframe(n_rows=25, n_cols=10, start_date="2000-01-01", freq="D"):
    """
    Generate a random DataFrame with a datetime index.

    Args:
        n_rows (int): Number of rows in the DataFrame.
        n_cols (int): Number of columns in the DataFrame.
        start_date (str): Start date for the datetime index.
        freq (str): Frequency for the datetime index (e.g., 'D' for daily, 'H' for hourly).

    Returns:
        pd.DataFrame: A random DataFrame with datetime as the index.
    """
    # Generate column names
    cols = [f"COL_{i}" for i in range(n_cols)]

    # Generate random data
    data = np.random.randint(0, 100, size=(n_rows, n_cols))

    # Create a datetime index
    index = pd.date_range(start=start_date, periods=n_rows, freq=freq)

    # Create the DataFrame
    df = pd.DataFrame(data, columns=cols, index=index)

    return df
```
Write a 20 years times 10 000 columns DataFrame with random data to Arctic:
```python
df = generate_random_dataframe(
    n_rows=255 * 20, n_cols=10_000, start_date="1990-01-01", freq="D"
)
df.head()
```
```python
import arcticdb as adb

uri = "lmdb://tmp/arcticdb_leak"
ac = adb.Arctic(uri)

library = ac.get_library("demo_lib", create_if_missing=True)

library.write("test_frame", df)
```

#### 2. Read Data

Helper function to get memory usage of a list of DataFrames:

```python
def get_total_dataframe_size_gb(df_list):
    """
    Calculate total memory usage of a list of DataFrames in gigabytes.

    Parameters:
        df_list (list of pd.DataFrame): List of DataFrames.

    Returns:
        float: Total size in GB.
    """
    total_bytes = sum(df.memory_usage(deep=True).sum() for df in df_list)
    return total_bytes / (1024**3)
```

Read the full dataframe for reference:
```python
from_storage_df = library.read("test_frame").data

print(f"Shape of data: {from_storage_df.shape}")
print(f"Size of data: {get_total_dataframe_size_gb([from_storage_df]):.2f} GB")
```

Read only a slice:
```python
n_rows_to_read = 3

from_date = from_storage_df.index[1]
to_date = from_storage_df.index[n_rows_to_read]

small_df = library.read(
    "test_frame",
    date_range=(from_date, to_date),
).data

print(f"Shape of fetched subset of data: {small_df.shape}")
print(
    f"Size of fetched subset of data: {get_total_dataframe_size_gb([small_df]):.4f} GB"
)
```

Now read the small slice in a loop and save results in a list:
```python
retrieved_data = []
n_times_to_fetch = 100

for i in range(n_times_to_fetch):
    small_df = library.read(
        "test_frame",
        date_range=(
            from_date,
            to_date,
        ),
    ).data
    retrieved_data.append(small_df)

    if i % 10 == 0:
        print(f"Fetched small subset {i} times")
        print(
            f"    Total size of retrieved data so far: {get_total_dataframe_size_gb(retrieved_data):.2f} GB"
        )

print()
print(
    f"Total size of retrieved data: {get_total_dataframe_size_gb(retrieved_data):.2f} GB"
)
```

### Expected Results

Memory usage should in total increase by ca 0.0002 GB * 300 = 0.06 GB -- instead, it increases by several hundred MB per iteration (until out of memory).

Image

- Linux
- Python 3.10.12
- ArcticDB 5.2.3

### Backend storage used

LMDB

### Additional Context

We're able to bypass the problem using `adb.QueryBuilder().date_range((from_date, to_date))` as an argument instead to `library.read`, but it's not clear if this is the intended way to do it and why the most obvious way to read a slice of a DataFrame would cause this memory leak.

vedjaw:
## Summary

Fixes **#2348**: `lib.read(symbol, date_range=...)` (and `row_range` / `read_batch`) could OOM when callers looped over small windows of a large symbol.

### Why

Intersecting segments (default ~100k rows) were loaded, then Python post-filtering returned **numpy views** onto those full buffers. Storing each “small” slice kept the parent segment alive → cumulative memory growth.

`QueryBuilder.date_range` already avoided this via C++ `DateRangeClause` truncate. `batch_read_and_join` already forced ranges onto that path. Single `read` / `read_batch` did not.

### What changed

1. **Default:** `_get_read_query` / `_get_read_queries` now route `date_range` / `row_range` through the processing pipeline (`force_ranges_to_queries=True`), matching the QueryBuilder / join behaviour so segments are truncated in C++.
2. **Safety net:** any remaining Python post-filter path **copies** sliced column arrays instead of returning views.
3. **Docs:** `read` / `QueryBuilder.date_range` no longer advertise the old “views are leaner” tradeoff.
4. **Tests:** correctness vs QueryBuilder / iloc, batch parity, legacy post-filter ownership, and a tiny-segment loop regression.

`force_ranges_to_queries=False` remains available for callers that intentionally want the legacy post-filter path (still copy-safe).

## Test plan

- [ ] `python/tests/unit/arcticdb/version_store/test_date_range_memory.py`
- [ ] Existing date_range / row_range / QueryBuilder / batch_read suites
- [ ] CI green

Fixes #2348

Made with [Cursor](https://cursor.com)

### DIFF
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -1640,11 +1640,9 @@ class NativeVersionStore:
         version_queries = self._get_version_queries(len(symbols), as_ofs, **kwargs)
         # Take a copy as _get_read_queries can modify the input argument, which makes reusing the input counter-intuitive
         per_symbol_query_builders = copy.deepcopy(per_symbol_query_builders)
-        # Needed to force date_range and row_range arguments to go through the read_and_process path rather than the
-        # direct read path if no explicit query is provided for a symbol
-        force_ranges_to_queries = True
+        # Ranges default to the processing-pipeline path (see _get_read_query / #2348).
         read_queries = self._get_read_queries(
-            len(symbols), date_ranges, row_ranges, columns, per_symbol_query_builders, force_ranges_to_queries
+            len(symbols), date_ranges, row_ranges, columns, per_symbol_query_builders, force_ranges_to_queries=True
         )
         read_options, output_format = self._get_read_options_and_output_format(**kwargs)
         return self._adapt_read_res(
@@ -2228,9 +2226,27 @@ class NativeVersionStore:
         row_range: Tuple[int, int],
         columns: Optional[List[str]],
         query_builder: Optional[QueryBuilder],
+        force_ranges_to_queries: bool = True,
     ):
+        """
+        Build a ReadQuery for a single symbol.
+
+        When ``force_ranges_to_queries`` is True (the default) and a ``date_range`` /
+        ``row_range`` is supplied without an existing QueryBuilder, the range is
+        converted into a processing clause. That path truncates segment buffers in
+        C++ so returned frames do not retain unused rows from intersecting segments
+        (see issue #2348). Set ``force_ranges_to_queries=False`` only for callers that
+        intentionally want the legacy zero-copy Python post-filter view.
+        """
         check(date_range is None or row_range is None, "Date range and row range both specified")
         read_query = _PythonVersionStoreReadQuery()
+
+        if (
+            force_ranges_to_queries
+            and query_builder is None
+            and (date_range is not None or row_range is not None)
+        ):
+            query_builder = QueryBuilder()
 
         if date_range is not None and query_builder is not None:
             query_builder.prepend(QueryBuilder().date_range(date_range))
@@ -2258,7 +2274,7 @@ class NativeVersionStore:
         row_ranges: Optional[List[Optional[Tuple[int, int]]]],
         columns: Optional[List[List[str]]],
         query_builder: Optional[Union[QueryBuilder, List[QueryBuilder]]],
-        force_ranges_to_queries: bool = False,
+        force_ranges_to_queries: bool = True,
     ):
         read_queries = []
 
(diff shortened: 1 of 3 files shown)


ognyanstoimenov:
Thanks for contributing. I see that this idea has been [explored](https://github.com/man-group/ArcticDB/issues/2348#issuecomment-2866150850) before and I think this change needs additional benchmarking numbers for this so that we know we haven't regressed in performance in certain situations. E.g. when the filter matches the majority of the dataframe the .copy will use ~ double the memory and this may cause OOMs for large dataframes. Also now all filtering goes throguh the query builder path which may be slower. And based on results we can decide if copying should be the default.

PR #3291, opened 2026-08-04
Enhancement 11555212497: compact_data_batch method

alexowens90:
#### Reference Issues/PRs
[11555212497](https://man312219.monday.com/boards/7852509418/pulses/11555212497)

Adds `compact_data_batch` method, a multi-symbol equivalent of `compact_data` for defragmenting data chunks on disk.

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.cpp
+++ b/cpp/arcticdb/version/local_versioned_engine.cpp
@@ -37,6 +37,37 @@ bool is_valid_merge_strategy(const arcticdb::MergeStrategy& strategy) {
     };
     return std::ranges::find(valid_strategies, strategy) != std::ranges::end(valid_strategies);
 }
+
+void check_for_duplicated_symbols(const std::vector<StreamId>& stream_ids, std::string_view method) {
+    std::unordered_set<StreamId> unique_stream_ids;
+    std::vector<StreamId> non_unique_stream_ids;
+    unique_stream_ids.reserve(stream_ids.size());
+    for (const auto& stream_id : stream_ids) {
+        if (!unique_stream_ids.emplace(stream_id).second) {
+            non_unique_stream_ids.emplace_back(stream_id);
+        }
+    }
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
+            non_unique_stream_ids.empty(),
+            "Duplicate symbols in {} request. Symbols submitted more than once [{}]",
+            method,
+            fmt::join(non_unique_stream_ids, ",")
+    );
+}
+
+void compact_data_checks(const StreamId& stream_id, const version_store::UpdateInfo& update_info) {
+    storage::check<ErrorCode::E_SYMBOL_NOT_FOUND>(
+            update_info.previous_index_key_.has_value(), "Cannot compact data of non-existent symbol \"{}\".", stream_id
+    );
+    // We could make this work by compacting the leaf nodes, but it is not obviously worth the effort at this stage
+    // Leaf nodes of recursively normalized data cannot become fragmented through appends/updates, so it would only be
+    // useful for changing the slicing by an explicitly provided rows_per_segment
+    schema::check<ErrorCode::E_OPERATION_NOT_SUPPORTED_WITH_RECURSIVE_NORMALIZED_DATA>(
+            update_info.previous_index_key_->type() != KeyType::MULTI_KEY,
+            "Cannot compact data of recursively normalized symbol {}",
+            stream_id
+    );
+}
 } // namespace
 
 namespace arcticdb::version_store {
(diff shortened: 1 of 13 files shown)


cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -37,6 +37,32 @@ bool is_valid_merge_strategy(const arcticdb::MergeStrategy& strategy) {
...
+        if (!unique_stream_ids.emplace(stream_id).second) {
+            user_input::raise<ErrorCode::E_INVALID_USER_ARGUMENT>(
+                    "Duplicated symbol name '{}' passed to {}", stream_id, method
+            );
+        }
+    }
+}
+
+void compact_data_checks(const StreamId& stream_id, const version_store::UpdateInfo& update_info) {
+    storage::check<ErrorCode::E_SYMBOL_NOT_FOUND>(
```
alexowens90, line 59: `E_NO_SUCH_VERSION` would be more in-line with other modification methods. This is technically an API break for non-batch `compact_data`, but given how new that method is, we could think about sneaking it into a minor release.

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -9,12 +9,11 @@
 import pytz
 from arcticdb_ext.exceptions import ErrorCode, ErrorCategory
 
-from arcticdb.version_store import VersionedItem as PythonVersionedItem
```
alexowens90, line 12: This was a pointless alias

python/tests/unit/arcticdb/version_store/test_compact_data.py
```diff
@@ -780,6 +642,275 @@ def test_compact_data_output_column_missing_from_slice_changing_types(in_memory_
     generic_compact_data_test(lib, sym)
 
 
+@pytest.mark.parametrize("batch", [False, True])
+class TestCompactData:
```
alexowens90, line 646: All of these tests already existed, and have just been parametrized over batch and non-batch methods

python/tests/unit/arcticdb/version_store/test_compact_data.py
alexowens90, line 1: Most of the diff in this file is tests being moved into `TestCompactData`. The new tests are:

- `test_batch_compact_data_*`
- `test_compact_data_batch_docstring_example_v2`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1301,6 +1323,41 @@ VersionedItem LocalVersionedEngine::compact_data_internal(
...
+    internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+            stream_ids.size() == update_info_futs.size(), "stream_ids and update_info_futs must be of the same size"
+    );
+    std::vector<folly::Future<VersionedItem>> version_futures;
+    auto resolved_rows_per_segment = rows_per_segment.value_or(write_options_.segment_row_size);
+    for (auto&& [idx, update_info_fut] : folly::enumerate(update_info_futs)) {
+        version_futures.push_back(std::move(update_info_fut)
+                                          .via(&async::cpu_executor())
+                                          .thenValue([this,
+                                                      &stream_id = stream_ids[idx],
```
poodlewars, line 1344: I think the capture by reference here is OK given the `collectAll` below

python/arcticdb/version_store/_store.py
```diff
@@ -4146,6 +4146,92 @@ def compact_data(
...
+            The i-th element of the returned list corresponds to the i-th entry of the input symbols argument. List of
+            structures containing information including the version number of the written symbol in the store. The data
+            and metadata attributes will not be populated.  If no compaction occurs because the data is already
+            compacted, the version field will be that of the latest live version for the symbol.
+
+        Raises
+        ------
+        StorageException
+            If any of the symbols don't exist
+        ArcticNativeException
```
poodlewars, line 4200: Is this used as a base type for any other exceptions? Could make handling a bit tricky if so

python/arcticdb/version_store/library.py
```diff
@@ -1217,7 +1217,7 @@ def write_pickle(
 
     @staticmethod
     def _raise_if_duplicate_symbols_in_batch(batch):
```
poodlewars: We have the same checks in the C++ layer no? Do those map on to an `ArcticDuplicateSymbolsInBatchException` as well?

python/tests/integration/arcticdb/test_arctic_batch.py
```diff
@@ -1680,3 +1679,68 @@ def dataframe_and_symbol(version_num, symbol_num):
...
+def test_compact_data_batch_prune_previous(lmdb_library, prune_previous_versions):
+    lib = lmdb_library
+    syms = [f"test_compact_data_batch_prune_previous_{i}" for i in range(3)]
+    lib.write_batch([WritePayload(sym, pd.DataFrame({"col": [0]})) for sym in syms])
+    lib.append_batch([WritePayload(sym, pd.DataFrame({"col": [1]})) for sym in syms])
+    lib_tool = lib._nvs.library_tool()
+    # 2 per symbol
+    assert len(lib_tool.find_keys(KeyType.TABLE_INDEX)) == 6
+    lib.compact_data_batch(syms, prune_previous_versions=prune_previous_versions)
+    assert len(lib_tool.find_keys(KeyType.TABLE_INDEX)) == (3 if prune_previous_versions else 9)
```
poodlewars, line 1695: Would also be good to check that the compaction does something in both cases when there is something large enough to compact.

poodlewars (approved):

poodlewars:
PR description needs some user-facing explanation as it will be used for release notes

cpp/arcticdb/version/local_versioned_engine.cpp
alexowens90: Yeh the vector element can't go out of scope while the reference is needed

python/arcticdb/version_store/_store.py
alexowens90: No, this is the most generic exception type we use from the Python layer

python/arcticdb/version_store/library.py
alexowens90: Annoyingly no. This is one of the places the V1 and V2 APIs diverge significantly. All V2 API batch modification methods perform this check immediately. V1 has no such checks in most cases (I opened a [ticket](https://man312219.monday.com/boards/7852509418/pulses/12701526171)), although the ones I tested did at least throw, just not very useful exceptions.
Now that I think about it more, I think the right thing to do is move `_raise_if_duplicate_symbols_in_batch` down into `NativeVersionStore`, and then only call it from `batch_compact_data` for now. Python import magic means importing `from arcticdb.version_store.library import ArcticDuplicateSymbolsInBatchException` will continue to work.
I'll leave the check in `local_versioned_engine.cpp` anyway, although it should become unreachable from Python.

python/tests/integration/arcticdb/test_arctic_batch.py
alexowens90: That's covered in `test_compact_data.py`?

poodlewars: Not with `prune_previous_versions`

PR #3295, opened 2026-08-04
Log a warning in a process using arcticdb that forks, from Python 3.12+ on Linux

poodlewars:
Forking a process that uses ArcticDB is unsafe: `fork()` duplicates only the calling thread, so
the folly pools and the storage SDKs' background threads do not exist in the child, and any mutex
they held stays locked. `Arctic`, `Library` and `NativeVersionStore` objects created before the
fork must not be used afterwards in the child.

CPython 3.12 added a `DeprecationWarning` for forking a multi-threaded process, but it is
invisible in the case that matters. It is attributed to `multiprocessing/popen_fork.py` rather
than `__main__`, so the default warning filters drop it. Users doing exactly the wrong thing get
no diagnostic. This PR logs an ArcticDB warning instead.

## New runtime configuration: `Fork.WarnOnFork`

**User-facing.** Controls whether the warning is logged. Documented in
`docs/mkdocs/docs/runtime_config.md`.

| | |
|---|---|
| Default | `1` (enabled) |
| Disable in code | `set_config_int("Fork.WarnOnFork", 0)` |
| Disable by environment | `ARCTICDB_Fork_WarnOnFork_int=0` |

The warning is logged at most once per process however many times it forks, and only once
ArcticDB has started its background threads.

No public API changes, and no on-disk format changes.

### DIFF
--- a/cpp/arcticdb/python/python_module.cpp
+++ b/cpp/arcticdb/python/python_module.cpp
@@ -206,7 +206,6 @@ void reinit_lmdb_warning() {
 
 // Windows has no fork(), so there is nothing to warn about there.
 #if !defined(WIN32) && PY_VERSION_HEX >= ARCTICDB_PY_FORK_DEPRECATED_HEX
-#include <unistd.h>
 
 static std::atomic_flag warned_about_fork;
 
@@ -214,6 +213,12 @@ static std::atomic_flag warned_about_fork;
 // hang: the sink's mutex may be inherited locked, and an async logger has no flusher thread there.
 void warn_about_fork() {
     using namespace arcticdb;
+    // Importing arcticdb, and even creating libraries, starts no pool threads, so there would be
+    // nothing to warn about. This also keeps a multiprocessing forkserver quiet: it forks every
+    // worker, but does no ArcticDB work itself.
+    if (!async::io_pool_thread_started.load(std::memory_order_relaxed)) {
+        return;
+    }
     if (ConfigsMap::instance()->get_int("Fork.WarnOnFork", 1) == 0) {
         return;
     }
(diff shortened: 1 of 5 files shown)


IvoDD (approved):

cpp/arcticdb/python/python_module.cpp
```diff
@@ -198,6 +204,37 @@ void reinit_lmdb_warning() {
     arcticdb::storage::lmdb::LmdbStorage::reset_warning_counter();
 }
 
+// Windows has no fork(), so there is nothing to warn about there.
+#if !defined(WIN32) && PY_VERSION_HEX >= ARCTICDB_PY_FORK_DEPRECATED_HEX
+#include <unistd.h>
+
+static std::atomic_flag warned_about_fork;
```
alexowens90, line 215: `std::call_once`?

python/tests/unit/arcticdb/version_store/test_fork.py
```diff
@@ -137,6 +146,80 @@ def test_configs_propagated_to_child_process(request, store_fixture, start_metho
...
+    return proc.stderr.count(FORK_WARNING)
+
+
+_FORK_ONCE = """
+    import os
+    import arcticdb
+
+    pid = os.fork()
+    if pid == 0:
+        os._exit(0)
```
alexowens90, line 189: What's the advantage of this over just letting the child process exit "normally"?

python/tests/unit/arcticdb/version_store/test_fork.py
```diff
@@ -137,6 +146,80 @@ def test_configs_propagated_to_child_process(request, store_fixture, start_metho
         unset_config_int("TestPropagation")
 
 
+def _count_fork_warnings(body, env=None):
+    """Run body in a fresh interpreter and count the fork warnings it logs to stderr."""
+    child_env = dict(os.environ)
+    child_env.update(env or {})
+    proc = subprocess.run([sys.executable, "-c", textwrap.dedent(body)], capture_output=True, text=True, env=child_env)
```
alexowens90: I hadn't thought about this before, but we started skipping all our `capfs` tests before as they were unreliable. Running in a subprocess like this would be a viable alternative solution.

alexowens90 (approved):

python/tests/unit/arcticdb/version_store/test_fork.py
poodlewars: The tear down of the child process runs cleanup (eg the C++ static destructors) which can deadlock - the category of problems that we're warning against.

cpp/arcticdb/python/python_module.cpp
poodlewars: That has a mutex in it which is a bit worrying when forking.

PR #3297, opened 2026-08-04
[throwaway, do not merge] Test docs-only CI skip

poodlewars:
Temporary PR to verify detect_docs_only.yml correctly skips the build/test matrix on a docs-only diff. Will be closed without merging once verified.


poodlewars:
Verified: detect_docs_only correctly reported docs_only=true, the full build/conda matrix skipped, and can_merge / All conda jobs completed both went green in under a minute. Closing without merging — this branch only existed for testing.

PR #3300, opened 2026-08-04
Throwaway: verify docs-only detection after review fixes

poodlewars:
Throwaway PR to verify `detect_docs_only.yml` still correctly classifies a docs-only change as `docs_only=true` using the new PR-files-API approach (no full checkout). Will be closed immediately after checking.


poodlewars:
Confirmed: docs_only=true correctly detected via PR-files API (no checkout needed), matrix skipped, can_merge + All conda jobs completed both passed.

PR #3301, opened 2026-08-04
Throwaway: verify non-docs detection after review fixes

poodlewars:
Throwaway PR to verify `detect_docs_only.yml` still correctly classifies a .cpp change as `docs_only=false` using the new PR-files-API approach. Will be closed immediately after checking.


poodlewars:
Confirmed: docs_only=false correctly detected for a .cpp-only change via the PR-files API (no checkout needed for pull_request events), matrix started running normally. Cancelled the in-progress matrix runs after confirming classification to save CI cost.

PR #3302, opened 2026-08-04
Implement insertion strategy for merge update with row range index

vasil-pashov:
#### Reference Issues/PRs

New data is always appended as a singe segment.

If the strategy allows update the update code path is reused without any changes.

The MatchRecord already has information about all matched source rows. Now it can emit an ordered vector containing all source rows that are not matched. Each processing unit emits such an entity similar to how each processing unit emits `MergeUpdateInsertedRowsComponent` tracking the number of inserted rows when datetime index is used. At the end of the processing all such sets are intersected. The result is a set of the rows which must be inserted.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/processing/component_manager.hpp
+++ b/cpp/arcticdb/processing/component_manager.hpp
@@ -30,11 +30,19 @@ using namespace entt::literals;
 /// clause adding this entity and then after read_modify_write it reads it in order to compute the correct row ranges
 /// accounting for insertion. If any other piece of code adds this entity to the component manager the merge updates
 /// will iterate over them as well, producing wrong results. See merge_update_impl and Monday 12618296803
-struct MergeUpdateInsertedRowsEntity {
-    MergeUpdateInsertedRowsEntity() = default;
-    MergeUpdateInsertedRowsEntity(const size_t inserted_rows) : inserted_rows(inserted_rows) {}
+struct MergeUpdateInsertedRowsComponent {
+    MergeUpdateInsertedRowsComponent() = default;
+    MergeUpdateInsertedRowsComponent(const size_t inserted_rows) : inserted_rows(inserted_rows) {}
     operator size_t() const { return inserted_rows; }
     size_t inserted_rows = 0;
+};
+
+struct MergeUpdateNotMatchedSourceRowsComponent {
+    MergeUpdateNotMatchedSourceRowsComponent() = default;
+    MergeUpdateNotMatchedSourceRowsComponent(std::shared_ptr<util::BitSet> unmatched_source_rows) :
+        unmatched_source_rows(std::move(unmatched_source_rows)) {}
+    operator util::BitSet&() const { return *unmatched_source_rows; }
+    std::shared_ptr<util::BitSet> unmatched_source_rows;
 };
 
 class ComponentManager {
@@ -140,7 +148,7 @@ class ComponentManager {
     }
 
     template<typename ProcessComponents>
-    void process_entities(ProcessComponents&& process_fn) {
+    auto process_entities(ProcessComponents&& process_fn) const {
         using ArgTypes = util::function_arg_types<std::decay_t<ProcessComponents>>::args_t;
         // Derive the component tuple type (references stripped) without default-constructing it, as
         // the component types need not be default-constructible.
--- a/cpp/arcticdb/processing/component_manager.hpp
+++ b/cpp/arcticdb/processing/component_manager.hpp
@@ -37,11 +37,16 @@ struct MergeUpdateInsertedRowsComponent {
     size_t inserted_rows = 0;
 };
 
+/// Used only by the MergeUpdateClause, and only for row-count indexed targets. After the pipeline finishes,
+/// write_inserted_row_range_data (via source_rows_to_insert_for_row_range_merge_update) intersects this component
+/// across *all* entities that carry it to decide which source rows to append. If any other piece of code adds this
+/// component the intersection will silently drop bits and rows will not be inserted. All instances added within a
+/// single merge must have the same size.
 struct MergeUpdateNotMatchedSourceRowsComponent {
     MergeUpdateNotMatchedSourceRowsComponent() = default;
     MergeUpdateNotMatchedSourceRowsComponent(std::shared_ptr<util::BitSet> unmatched_source_rows) :
         unmatched_source_rows(std::move(unmatched_source_rows)) {}
-    operator util::BitSet&() const { return *unmatched_source_rows; }
+    operator util::BitSet&() { return *unmatched_source_rows; }
     std::shared_ptr<util::BitSet> unmatched_source_rows;
 };
 
--- a/cpp/arcticdb/util/bitset.hpp
+++ b/cpp/arcticdb/util/bitset.hpp
@@ -25,6 +25,7 @@ using BitSetSizeType = bm::bvector<>::size_type;
 using BitIndex = bm::bvector<>::rs_index_type;
 
 template<typename functor>
+requires std::is_invocable_r_v<void, functor, size_t>
 struct BitSetVisitor {
     functor& f;
     int add_bits(BitSet::size_type off, const unsigned char* bits, unsigned n) {
--- a/cpp/arcticdb/util/bitset.hpp
+++ b/cpp/arcticdb/util/bitset.hpp
@@ -11,6 +11,7 @@
 #include <arcticdb/util/magic_num.hpp>
 
 #include <bitmagic/bm.h>
+#include <bitmagic/bmalgo.h>
 
 namespace arcticdb {
 
@@ -22,6 +23,23 @@ using BitMagicStart = SmallMagicNum<'M', 's'>;
 using BitMagicEnd = SmallMagicNum<'M', 'e'>;
 using BitSetSizeType = bm::bvector<>::size_type;
 using BitIndex = bm::bvector<>::rs_index_type;
+
+template<typename functor>
+struct BitSetVisitor {
+    functor& f;
+    int add_bits(BitSet::size_type off, const unsigned char* bits, unsigned n) {
+        for (unsigned i = 0; i < n; ++i) {
+            f(off + bits[i]);
+        }
+        return 0;
+    }
+    int add_range(BitSet::size_type off, BitSet::size_type n) {
+        for (BitSet::size_type i = 0; i < n; ++i) {
+            f(off + i);
+        }
+        return 0;
+    }
+};
 
 /// @brief Get the combined size of the magic words used as delimiters for the sparse bitmaps
 /// When sparse bitmaps are encoded we use two different magic words to mark the start and the end of the bitmap
@@ -50,10 +68,20 @@ void copy_packed_bits(const uint8_t* src, size_t src_bit_offset, size_t num_bits
 template<typename functor>
 requires std::is_invocable_r_v<void, functor, size_t>
 void iterate_over_set_positions(const bm::bvector<>& bv, size_t from, size_t to, functor&& f) {
-    // TODO: Investigate performance of `get_enumerator`. Maybe having an rs_index can speed it up.
-    for (auto en = bv.get_enumerator(from); *en < to; ++en) {
-        f(*en);
+    util::check(to >= from, "Invalid bit set iteration range: from {} to {}", from, to);
+    if (from == to) {
+        return;
     }
+    util::BitSetVisitor visitor{f};
+    // the range that bitmagic requires is inclusive
+    bm::for_each_bit_range(bv, from, to - 1, visitor);
+}
+
+template<typename functor>
+requires std::is_invocable_r_v<void, functor, size_t>
+void iterate_over_set_positions(const bm::bvector<>& bv, functor&& f) {
+    util::BitSetVisitor visitor{f};
+    bm::for_each_bit(bv, visitor);
 }
 
 } // namespace arcticdb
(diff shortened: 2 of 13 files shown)


python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -3026,6 +3018,144 @@ def test_row_range_index_name_does_not_count_as_duplicate(self, lmdb_library):
...
+        self.strategy = MergeStrategy(MergeAction.DO_NOTHING, MergeAction.INSERT)
+
+    def test_basic_append_unmatched(self, lmdb_library):
+        lib = lmdb_library
+        target = pd.DataFrame({"a": [1, 2, 3], "b": [1.0, 2.0, 3.0]})
+        # a=1 matches (do_nothing leaves it untouched), a=99 and a=100 are unmatched and appended in source order.
+        source = pd.DataFrame({"a": [1, 99, 100], "b": [10.0, 99.0, 100.0]})
+        lib.write("sym", target)
+        lib.merge_experimental("sym", source, strategy=self.strategy, on=["a"])
+        expected = pd.DataFrame({"a": [1, 2, 3, 99, 100], "b": [1.0, 2.0, 3.0, 99.0, 100.0]})
```
IvoDD, line 3041: Would be nice to have `query_stats` tests for the rowrange case. E.g. here to verify that a single data key was written

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -2652,8 +2642,10 @@ def test_merge_matched_update_with_metadata(self, lmdb_library, strategy):
         assert_vit_equals_except_data(merge_vit, read_vit)
 
         lt = lib._dev_tools.library_tool()
-
-        assert len(lt.find_keys_for_symbol(KeyType.TABLE_DATA, "sym")) == 2
+        expected_data_keys = (
+            3 if strategy.not_matched_by_target == MergeAction.INSERT and strategy.matched == MergeAction.UPDATE else 2
```
IvoDD, line 2647: I find this slightly confusing because we actually write only `2` or `1` data keys. The reason we see `3` or `2` is because we also count the `v0` data key. 

A comment explaining 3 and 2 or using `prune_previous_versions` would be good I think

cpp/arcticdb/version/version_core.cpp
```diff
@@ -417,6 +418,13 @@ std::vector<SliceAndKey> merge_slices_and_keys(
                 }
             }
         }
+        while (new_slice_and_key_it != new_slices.end() && new_slice_and_key_it->slice().col_range == current_col_range
+        ) {
+            new_slice_and_key_it->slice().row_range.first += total_inserted_rows;
+            new_slice_and_key_it->slice().row_range.second += total_inserted_rows;
+            merged_ranges_and_keys.emplace_back(std::move(*new_slice_and_key_it));
+            ++new_slice_and_key_it;
+        }
```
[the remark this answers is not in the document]
IvoDD: I think this claude comment might not be correct (rows after the last are considered alongside the last chunk so this while loop would never be reached from the timeseries path).

Still I think we should add some tests on the index key structure (i.e. row start and row end is correct after insertions)

cpp/arcticdb/version/version_core.cpp
```diff
@@ -3182,37 +3284,70 @@ folly::Future<AtomKey> merge_update_impl(
...
-                        tsd,
-                        std::move(merged_ranges_and_keys),
                         target_partial_index_key,
-                        store
-                );
+                        strategy](std::vector<SliceAndKey>&& data_keys_and_slices) {
+                const StreamDescriptor& target_descriptor = pipeline_context->descriptor();
+                folly::SemiFuture<std::vector<SliceAndKey>> inserted_row_slices_fut =
+                        (target_descriptor.index().type() == IndexDescriptor::Type::ROWCOUNT && strategy.insert())
+                                ? write_inserted_row_range_data(
```
IvoDD, line 3294: Doesn't this introduce a dependency for the inserted data key on the other data keys?

I.e. the last inserted data key at the end starts writing only after all previous data keys finished writing.

This means that the flow is for example if we have 3 updated data keys:
```
reading affected data keys [1, 2, 3] in parallel
        |
        v
processing data keys [1, 2, 3] in parallel
        |
        v
processing finishes
        |
        v
write updated data keys [1, 2, 3] in parallel
        |
        v
write final inserted data key     
```
Instead we could schedule the inserted data key write alongside the data keys.

Sorry if this was already discussed.

cpp/arcticdb/util/bitset.hpp
```diff
@@ -23,6 +24,23 @@ using BitMagicEnd = SmallMagicNum<'M', 'e'>;
 using BitSetSizeType = bm::bvector<>::size_type;
 using BitIndex = bm::bvector<>::rs_index_type;
 
+template<typename functor>
+struct BitSetVisitor {
+    functor& f;
+    int add_bits(BitSet::size_type off, const unsigned char* bits, unsigned n) {
```
IvoDD, line 31: This is a surprising method signature. Why the bits are an `unsigned char`s and `n` is unsigned? Is this signature required from bitmagic because it seems quite limiting with so small bits sizes? If so what do you think about using `uint8_t` instead of unsigned char?

cpp/arcticdb/processing/component_manager.hpp
```diff
@@ -30,13 +30,21 @@ using namespace entt::literals;
...
     size_t inserted_rows = 0;
 };
 
+struct MergeUpdateNotMatchedSourceRowsComponent {
+    MergeUpdateNotMatchedSourceRowsComponent() = default;
+    MergeUpdateNotMatchedSourceRowsComponent(std::shared_ptr<util::BitSet> unmatched_source_rows) :
+        unmatched_source_rows(std::move(unmatched_source_rows)) {}
+    operator util::BitSet&() const { return *unmatched_source_rows; }
+    std::shared_ptr<util::BitSet> unmatched_source_rows;
+};
```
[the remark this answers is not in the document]
IvoDD: I think claude's suggestion for comment and const makes sense

cpp/arcticdb/version/version_core.cpp
```diff
@@ -426,6 +434,105 @@ std::vector<SliceAndKey> merge_slices_and_keys(
...
+            descriptors.emplace_back();
+            descriptors.back().set_index(descriptor.index());
+            descriptors.back().set_id(descriptor.id());
+        }
+        descriptors.back().add_field(descriptor.field(field_index));
+    }
+    return descriptors;
+}
+
+folly::SemiFuture<std::vector<SliceAndKey>> write_inserted_row_range_data(
```
IvoDD, line 468: This will need to do row slicing (not only column slicing) in the future.

This will need to be reconcieled with `WriteClause`'s own slicing implemented in the future.

Obviously not ideal but what do you think of handling insert with row range the same way as we would if we had a timeseries index which is just the same across all values? 

I.e. have one processing unit responsible for the entire target and produce one segment in the end which contains the updated and inserted values. Then the code for timeseries insert would handle the output structure correctly.
Obviously this will be a lot slower if we have many rowslices due to single threaded processing

cpp/arcticdb/version/version_core.cpp
vasil-pashov: You need all calls to process to end so that you can compute the global bit set of matched rows.

cpp/arcticdb/util/bitset.hpp
vasil-pashov: This is required by bitmagic.

cpp/arcticdb/version/version_core.cpp
vasil-pashov: I don't think the `WriteClause` should do any slicing.

If a time series is the same across all values the processing unit will load all segments in a single unit and a single process will execute them all. This will OOM with large dataframes.

Also we already have a precedent of writing segments after the pipeline ends (defrag)

cpp/arcticdb/version/version_core.cpp
```diff
@@ -426,6 +434,105 @@ std::vector<SliceAndKey> merge_slices_and_keys(
...
+            rows_to_insert = *unmatched_source_rows.unmatched_source_rows;
+        }
+    });
+    return rows_to_insert;
+}
+
+std::vector<StreamDescriptor> split_rowrange_descriptor(
+        const StreamDescriptor& descriptor, const size_t column_slice_size
+) {
+    auto descriptors = util::reserve_vector<StreamDescriptor>(descriptor.field_count() / column_slice_size + 1);
```
alexowens90: nit: This is one element larger than necessary when `descriptor.field_count() % column_slice_size == 0`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -426,6 +434,105 @@ std::vector<SliceAndKey> merge_slices_and_keys(
...
+    for (size_t col_slice = 0; col_slice < descriptors.size(); ++col_slice) {
+        SegmentInMemory new_segment(
+                std::move(descriptors[col_slice]), num_rows_to_insert, AllocationType::PRESIZED, Sparsity::NOT_PERMITTED
+        );
+        const StreamDescriptor& slice_descriptor = new_segment.descriptor();
+        const ColRange col_range{
+                col_slice * columns_per_slice,
+                std::min((col_slice + 1) * columns_per_slice, target_descriptor.field_count())
+        };
+        for (size_t column_in_segment = 0; column_in_segment < slice_descriptor.field_count(); ++column_in_segment) {
```
alexowens90, line 491: This could be parallelised across column slices, but I think we should wait until we're row-slicing as well.
Me and @IvoDD have previously discussed a more flexible `WriteToSegmentTask` that doesn't just deal with contiguous rows. We discussed it in the context of sorting segments on staging, which currently forms a fully sorted segment and then slices it. The proposed change would instead just generate the jive table, and have the new variant of `WriteToSegmentTask` just extract the rows it needs. A similar approach could be taken here. 
In fact, I think we could do both by having a `std::vector<size_t>` parameter to the task which is an ordered list of rows from the `InputFrame` to write into this slice.

cpp/arcticdb/version/version_core.cpp
alexowens90, line 1: There are now quite a lot of free functions in here that relate only to merging. What do you think about moving them to `clause_merge_update.cpp` to make their specificity more explicit? Can introduce `clause_merge_update.hpp` at the same time, I think we only left everything in `clause.hpp` to avoid too many merge conflicts originally. Can be in a follow on PR as I know others are stacked on this one.

cpp/arcticdb/util/bitset.hpp
```diff
@@ -50,10 +68,20 @@ void copy_packed_bits(const uint8_t* src, size_t src_bit_offset, size_t num_bits
 template<typename functor>
 requires std::is_invocable_r_v<void, functor, size_t>
 void iterate_over_set_positions(const bm::bvector<>& bv, size_t from, size_t to, functor&& f) {
-    // TODO: Investigate performance of `get_enumerator`. Maybe having an rs_index can speed it up.
-    for (auto en = bv.get_enumerator(from); *en < to; ++en) {
-        f(*en);
+    util::check(to >= from, "Invalid bit set iteration range: from {} to {}", from, to);
+    if (from == to) {
+        return;
     }
```
alexowens90, line 75: Was this a bottleneck before?

cpp/arcticdb/util/bitset.hpp
```diff
@@ -23,6 +24,23 @@ using BitMagicEnd = SmallMagicNum<'M', 'e'>;
 using BitSetSizeType = bm::bvector<>::size_type;
 using BitIndex = bm::bvector<>::rs_index_type;
 
+template<typename functor>
+struct BitSetVisitor {
```
alexowens90, line 29: I guess the same `requires` as for `iterate_over_set_positions` applies here?

docs/mkdocs/docs/notebooks/ArcticDB_merge.ipynb
alexowens90, line 1: For completeness, can include a matched="update", not_matched_by_target="do_nothing" example for non-timeseries data

alexowens90 (approved):

docs/mkdocs/docs/notebooks/ArcticDB_merge.ipynb
alexowens90, line 1: Bit of a nit, but instead of saying `RowRange` anywhere, can we just use timeseries and non-timeseries? This then extends to Arrow, and is more descriptive of how we actually store data

cpp/arcticdb/version/version_core.cpp
vasil-pashov: Yeh, I wanted to separate this for quite some time, but PRs kept stacking up. I'll open a follow up once the related stuff is merged.

cpp/arcticdb/util/bitset.hpp
vasil-pashov: No, I just found the comment interesting.

cpp/arcticdb/version/version_core.cpp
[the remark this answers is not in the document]
vasil-pashov: Claude's wrong here, yes. I added a check on in the hypothesis.

IvoDD (approved):

alexowens90 (approved):

PR #3317, opened 2026-08-06
Generate schema output before processing in read_modify_write and refactor PipelineContext

vasil-pashov:
This commit implements two changes.

Monday: 10997979275

## Generate the output schema before the processing pipeline is run in `read_modify_write`

Currently `read_modify_write` calls `generate_output_schema_and_save_to_pipeline` **after** the pipeline ends. This is the functions that applies modify schema to compute the output schema and performs the checks for schema validity for all clauses. This is wasteful for two reasons:
1. It'll run the pipeline anyway even if modify_schema forbids it and only at the it will check the preconditions
2. If write clause is part of the pipeline it'll write the segments and throw, orphaning the written segments

merge_update needs to know the output schema in advance. This will be important for the implementation of dynamic schema.

This PR moved this computation before the pipeline is run.

## Refactor `PipelineContext`

The `PipelineContext` now holds the output schema and `merge_update` can use it (will be needed for dynamic schema in future).

The `PipelineContext` holds multiple descriptors, some of which poorly named which makes the working with the context confusing and error prone. The descriptors are now renamed and made private. For most of the part this is renaming

`orig_desc_` -> `on_disk_descriptor_`
`desc_` -> `string_coerced_descriptor_`

The string coerced descriptor is used at run time to produce different string types. On the type of the string is changed, names, positions and count of the fields stays the same.

A new utility method for getting the output schema is added which uses 3 way fallback. If the processing pipeline set the output schema it's the result descriptor, if during a read string coercion was required it's the correct user facing output, otherwise return the on disk descriptor.

### DIFF
--- a/cpp/arcticdb/pipeline/pipeline_context.hpp
+++ b/cpp/arcticdb/pipeline/pipeline_context.hpp
@@ -159,6 +159,10 @@ struct PipelineContext : public std::enable_shared_from_this<PipelineContext> {
 
     const proto::descriptors::NormalizationMetadata& output_normalization() const;
 
+    /// Number of leading required (index) fields in the output, accounting for multi-index/series normalization.
+    /// Returns the count with no normalization applied when the context carries neither an output schema nor a TSD.
+    uint32_t output_required_fields_count() const;
+
     const ankerl::unordered_dense::map<std::string, Value>& output_default_values() const;
 
     void set_selected_columns(const std::optional<std::vector<std::string>>& columns);
--- a/cpp/arcticdb/pipeline/pipeline_context.hpp
+++ b/cpp/arcticdb/pipeline/pipeline_context.hpp
@@ -12,6 +12,7 @@
 #include <arcticdb/entity/index_range.hpp>
 #include <arcticdb/entity/timeseries_descriptor.hpp>
 #include <arcticdb/pipeline/frame_slice.hpp>
+#include <arcticdb/pipeline/read_options.hpp>
 #include <arcticdb/util/bitset.hpp>
 #include <memory>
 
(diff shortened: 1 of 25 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -985,106 +956,6 @@ folly::Future<std::vector<EntityId>> schedule_clause_processing(
     });
 }
 
-void set_output_descriptors(
```
vasil-pashov, line 1126: this was dead code, deleted

cpp/arcticdb/version/version_core.cpp
```diff
@@ -53,35 +53,6 @@ namespace arcticdb::version_store {
 
 namespace ranges = std::ranges;
 
-static void modify_descriptor(
```
vasil-pashov, line 60: Moved as a member function of PipelineContext and renamed to `generate_string_coerced_descriptor`

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -1454,7 +1454,7 @@ std::shared_ptr<PipelineContext> setup_join_pipeline_context(
 ) {
     auto output_schema = modify_schema(clauses.front()->join_schemas(std::move(input_schemas)), clauses);
     auto pipeline_context = std::make_shared<PipelineContext>();
-    pipeline_context->set_descriptor(output_schema.stream_descriptor());
+    pipeline_context->set_on_disk_descriptor(output_schema.stream_descriptor());
```
alexowens90: I think `set_output_schema` would be more appropriate? Not that it will change the behaviour in this case, just more descriptive

alexowens90 (approved):

cpp/arcticdb/processing/operation_dispatch_binary.hpp
```diff
@@ -74,6 +74,11 @@ inline std::string binary_operation_with_types_to_string(
     );
 }
 
+std::string binary_operation_with_types_to_string(
```
alexowens90: Is this specialisation needed? Won't the above do the right thing if `Func == OperationType`?

cpp/arcticdb/processing/operation_dispatch_unary.hpp
```diff
@@ -31,10 +31,12 @@ VariantData unary_boolean(FullResult, OperationType operation);
 VariantData visit_unary_boolean(const VariantData& left, OperationType operation);
 
 template<typename Func>
-inline std::string unary_operation_to_string(Func&& func, std::string_view operand_str) {
+std::string unary_operation_to_string(Func&& func, std::string_view operand_str) {
     return fmt::format("{}({})", func, operand_str);
 }
 
+std::string unary_operation_to_string(OperationType func, std::string_view operand_str);
```
alexowens90: Same comment as on the binary version

cpp/arcticdb/processing/expression_node.cpp
```diff
@@ -486,10 +504,11 @@ std::variant<BitSetTag, DataType> ExpressionNode::compute(
                          is_sequence_type(std::get<DataType>(right_type))) ||
                                 (is_numeric_type(std::get<DataType>(left_type)) &&
                                  is_numeric_type(std::get<DataType>(right_type))),
-                        "Unexpected data types {} {} input to {}",
-                        std::get<DataType>(left_type),
-                        std::get<DataType>(right_type),
-                        operation_type_
+                        "Cannot check membership '{}' of {} {} in set of {}",
```
alexowens90: nit: the types are in brackets elsewhere when they come after a label, so
`"Cannot check membership '{}' of {} ({}) in set of {}",`

cpp/arcticdb/pipeline/pipeline_context.hpp
```diff
@@ -159,6 +159,10 @@ struct PipelineContext : public std::enable_shared_from_this<PipelineContext> {
 
     const proto::descriptors::NormalizationMetadata& output_normalization() const;
 
+    /// Number of leading required (index) fields in the output, accounting for multi-index/series normalization.
```
alexowens90: The `index` here is a bit misleading, as it is `num_index_columns + 1` for `Series`

alexowens90 (approved):

PR #3319, opened 2026-08-12
Add per key type timing and partial-failure handling to the object size scan

G-D-Petrov:

`LocalVersionedEngine::scan_object_sizes` now reports how long each key type took to scan, and no longer
fails the whole scan when a single key type cannot be listed.

- `ObjectSizes` carries `scan_duration_ns_`, exposed to Python as `ObjectSizes.scan_duration_ns`.
  `scan_object_sizes_for_stream` reports it too.
- `OnScanFailure::Raise` keeps the existing behaviour. `OnScanFailure::Skip` logs and omits a key type that
  cannot be listed, so one unreadable key type costs only its own numbers rather than the whole scan. This is
  reached through `lib._nvs.version_store.scan_object_sizes`; `AdminTools.get_sizes` does not take the option.
- Adds `COLUMN_STATS` to `arcticdb.KeyType` and to the library-wide size scan. The per-symbol scans
  (`get_sizes_by_symbol`, `get_sizes_for_symbol`) are unchanged.

Driven by a downstream need: Man's `populate_library_size` background job computes per-library size statistics
with the legacy `arcticc` stack's `S3StorageTool::get_prefix_info`, which records a per-key-type duration.
Moving that job onto ArcticDB needs the same timings.

Not in this PR: letting callers choose which key types to scan. `populate_library_size` covers 25 key types and
its Grafana dashboards are keyed on a `key_type` label, so the job cannot move over on this PR alone - that
needs a follow-up.

#### API changes

Both additive, but the `KeyType` addition changes `get_sizes` output, so code asserting on the exact key set or
on `len(sizes)` needs updating.

- `arcticdb.KeyType` gains `COLUMN_STATS`, which now appears in `AdminTools.get_sizes`.
- `arcticdb.OnScanFailure` is newly exported.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>

</details>

### DIFF
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -41,6 +41,9 @@ struct KeySizesInfo {
     size_t compressed_size; // bytes
 };
 
+/// What a size scan does with a key type it cannot list.
+enum class OnScanFailure { Raise, Skip };
+
 /**
  * These statistics record the number of deletions *attempted*. In particular note that we attempt
  * to delete column stats keys without checking whether they exist first.
@@ -387,16 +390,13 @@ class LocalVersionedEngine : public VersionedEngine {
             const std::vector<std::pair<StreamId, VersionId>>& symbols_to_delete
     );
 
-    /** Scan sizes for the given key types, or for a default set of the most important ones when not specified. */
     /**
-     * Sizes and counts, one entry per key type. Defaults to the ten most important key types.
+     * Sizes and counts, one entry per key type scanned.
      *
-     * With raise_on_failure=false a key type that cannot be listed is logged and left out of the result instead of
+     * With OnScanFailure::Skip a key type that cannot be listed is logged and left out of the result instead of
      * failing the whole scan, so one bad key type costs only its own numbers.
      */
-    std::vector<storage::ObjectSizes> scan_object_sizes(
-            const std::optional<std::vector<KeyType>>& key_types = std::nullopt, bool raise_on_failure = true
-    );
+    std::vector<storage::ObjectSizes> scan_object_sizes(OnScanFailure on_failure = OnScanFailure::Raise);
 
     std::vector<storage::ObjectSizes> scan_object_sizes_for_stream(const StreamId& stream_id);
 
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -387,7 +387,10 @@ class LocalVersionedEngine : public VersionedEngine {
             const std::vector<std::pair<StreamId, VersionId>>& symbols_to_delete
     );
 
-    std::vector<storage::ObjectSizes> scan_object_sizes();
+    /** Scan sizes for the given key types, or for a default set of the most important ones when not specified. */
+    std::vector<storage::ObjectSizes> scan_object_sizes(
+            const std::optional<std::vector<KeyType>>& key_types = std::nullopt
+    );
 
     std::vector<storage::ObjectSizes> scan_object_sizes_for_stream(const StreamId& stream_id);
 
--- a/cpp/arcticdb/version/local_versioned_engine.hpp
+++ b/cpp/arcticdb/version/local_versioned_engine.hpp
@@ -388,8 +388,14 @@ class LocalVersionedEngine : public VersionedEngine {
     );
 
     /** Scan sizes for the given key types, or for a default set of the most important ones when not specified. */
+    /**
+     * Sizes and counts, one entry per key type. Defaults to the ten most important key types.
+     *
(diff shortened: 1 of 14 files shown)


cpp/arcticdb/version/test/test_object_sizes.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include <arcticdb/storage/test/in_memory_store.hpp>
+#include <arcticdb/util/test/generators.hpp>
+#include <arcticdb/version/local_versioned_engine.hpp>
+
+namespace arcticdb {
+
+namespace {
+
+/// Sizes one key type cannot be listed. Real storages fail this way per key type - a bucket policy denying a
```
poodlewars: `Sizes one key type cannot be listed` doesn't make sense to me

cpp/arcticdb/version/test/test_object_sizes.cpp
```diff
@@ -0,0 +1,75 @@
...
+    const std::vector<KeyType> key_types{KeyType::TABLE_DATA, KeyType::TABLE_INDEX};
+
+    EXPECT_ANY_THROW(engine.scan_object_sizes(key_types));
+}
+
+TEST(ScanObjectSizes, OmitsFailedKeyTypesWhenNotRaising) {
+    auto engine = engine_with_failing_key_type(KeyType::TABLE_INDEX);
+    const std::vector<KeyType> key_types{KeyType::TABLE_DATA, KeyType::TABLE_INDEX, KeyType::VERSION};
+
+    const auto sizes = engine.scan_object_sizes(key_types, false);
```
poodlewars: I don't like the boolean on this API, we could use an enum class, or just forget about this bit of error handling and raise if one key type fails. Silently dropping some results is a problem - if the scan is running without the permissions it needs, I think most users would rather know about it and set up their permissions properly.

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -387,7 +387,16 @@ class LocalVersionedEngine : public VersionedEngine {
             const std::vector<std::pair<StreamId, VersionId>>& symbols_to_delete
     );
 
-    std::vector<storage::ObjectSizes> scan_object_sizes();
+    /** Scan sizes for the given key types, or for a default set of the most important ones when not specified. */
```
poodlewars: Weird extra comment

cpp/arcticdb/version/local_versioned_engine.hpp
```diff
@@ -387,7 +387,16 @@ class LocalVersionedEngine : public VersionedEngine {
...
+    /**
+     * Sizes and counts, one entry per key type. Defaults to the ten most important key types.
```
poodlewars: "ten" could rot quickly

python/arcticdb/version_store/admin_tools.py
```diff
@@ -36,7 +36,15 @@ def sum_sizes(sizes: Iterable[Size]) -> Size:
...
-    This is a subset of all the key types that ArcticDB uses, covering the most important types.
+    The key types that ArcticDB can report sizes for, i.e. those that exist as objects in storage.
+
+    The first ten are the most important ones, and are the only types scanned by `AdminTools.get_sizes` unless you
+    ask for others explicitly. The rest are historical, specialized or transient; each one costs an extra listing
+    operation to scan, and most are empty in a typical library.
+
+    This is the complete set - `get_sizes(key_types=list(KeyType))` accounts for every object in the library.
+    ArcticDB has a few internal key types that are absent because they are never objects in storage: TOMBSTONE and
+    TOMBSTONE_ALL live inside VERSION key segments, and STREAM_GROUP is not a real key type at all.
```
poodlewars: We could have a test checking that the set difference between all key types and this list is what we expect, so it doesn't rot if we add more key types.

python/arcticdb/version_store/admin_tools.py
```diff
@@ -76,30 +84,74 @@ class KeyType(Enum):
...
+
+
+# Every member is named after its counterpart in the extension, so the mapping is derived rather than
+# written out. ATOMIC_LOCK is the one exception - the extension binds it as SLOW_LOCK. A member with no
+# counterpart fails here at import rather than on first use.
+_NATIVE_NAMES = {"ATOMIC_LOCK": "SLOW_LOCK"}
+
+_TO_NATIVE = {key_type: getattr(NativeKeyType, _NATIVE_NAMES.get(key_type.name, key_type.name)) for key_type in KeyType}
+
+_FROM_NATIVE = {native: key_type for key_type, native in _TO_NATIVE.items()}
```
poodlewars: Doing this work in the module isn't right, it should be in the `get_sizes` function. Really I think the switch statement version that existed before this diff is a cleaner way to do this.

poodlewars (changes requested):

poodlewars (approved):

PR #3321, opened 2026-08-14
1 - Column stats write side re-write

poodlewars:
This is split in to separate working commits, but they trample over each other so reviewing all together is likely best. This is why we have these all together in a single PR.

## Step 1: Do not create `SegmentInMemory` for each column stats row

This is a pure refactor. Rather than creating the `SegmentInMemory` repeatedly during column stats creation (we were creating one for each row of column stats, then merging them together in to one bigger `SegmentInMemory`). This had a lot of overhead because each of these needs a protobuf header `ColumnStatsHeader`, and was just needlessly confusing.

Now in the `ColumnStatsGenerationClause` the process returns entities that are `ColumnStatsRow` rather than `SegmentInMemory`. This is just:

```
struct ColumnStatValue {
    // to_segment_column_name(<name of the column at data_col_offset>, type), e.g. "v1_MIN(col)"
    std::string segment_column_name;
    ColumnStatTypeInternal type;
    size_t data_col_offset;
    Value value;
};

/// One per row slice. Added to the ComponentManager as a standalone entity by
/// ColumnStatsGenerationClause::process and read back by create_column_stats_impl via
/// process_entities, which scans the whole registry — no other code may add this component.
struct ColumnStatsRow {
    uint64_t start_row;
    uint64_t end_row;
    std::vector<ColumnStatValue> stats;
};
```

and models a single row in the column stats segment.

`create_column_stats_impl` now consumes them by,

```
    read_and_schedule_processing(store, pipeline_context, read_query, read_options, component_manager).get();

    std::vector<ColumnStatsRow> components;
    component_manager->process_entities([&components](const ColumnStatsRow& component) {
        components.emplace_back(component);
    });
```

and constructs the final `SegmentInMemory`: `    SegmentInMemory new_segment = build_column_stats_segment(std::move(components), pipeline_context->descriptor());`.

I've also made changes to calculate the types of the column stats rows based on the index TSD, rather than applying our type promotion logic to them at the point we merge them in to the `SegmentInMemory`.

## Step 2 - Store column stats by (start_row, end_row) rather than (start_index, end_index)

This new layout can handle duplicated index values.

When we come to do automatic stats maintenance at write time, I was worried about how this might work with `update`. The new layout gives us a reasonable way to do this - `flatten_and_fix_rows` takes `SliceAndKey`s with their old row range and stamps the new post-update row range on them. This gives us a natural place to map between old row range and new row range, which is what we need to take the pre-update column stats segment and map unchanged slices on to the new row ranges, without recalculating stats.

## Step 3 - Support creating stats with row_range or date_range args

Support creating stats with `row_range=` or `date_range=` while preserving any existing stats, so you can built up stats incrementally with successive calls with different ranges. The idea of this is to make it tractable to create stats over very large historical symbols, at least for more recent more relevant data.

## Review Fixes

Drop a check that column stats creation should not be attempted over string indexed columns. The commit message explains why we never needed this check with either the old or the new format. There was in fact already a test in test_column_stats_optimisation.py `test_column_stats_query_optimisation_index_types` showing that this worked.

### DIFF
--- a/cpp/arcticdb/pipeline/column_stats.cpp
+++ b/cpp/arcticdb/pipeline/column_stats.cpp
@@ -1,45 +1,35 @@
 #include <arcticdb/pipeline/column_stats.hpp>
-#include <arcticdb/pipeline/index_fields.hpp>
 #include <arcticdb/processing/aggregation_interface.hpp>
 #include <arcticdb/processing/unsorted_aggregation.hpp>
 #include <arcticdb/entity/timeseries_descriptor.hpp>
 #include <arcticdb/entity/type_utils.hpp>
 #include <arcticdb/util/preconditions.hpp>
 
+#include <algorithm>
+#include <tuple>
+
 namespace arcticdb {
 
-SegmentInMemory merge_column_stats_segments(const std::vector<SegmentInMemory>& segments) {
-    SegmentInMemory merged(Sparsity::PERMITTED);
-    merged.init_column_map();
-    merged.descriptor().set_index(IndexDescriptorImpl{IndexDescriptor::Type::ROWCOUNT, 0});
+namespace {
+struct StatColumn {
+    std::string name;
+    ColumnStatTypeInternal type;
+    size_t data_col_offset;
+    TypeDescriptor type_descriptor;
+};
 
-    // Maintain the order of the columns in the input segments
-    ankerl::unordered_dense::map<std::string, size_t> field_name_to_index;
-    std::vector<TypeDescriptor> type_descriptors;
-    std::vector<std::string> field_names;
-    std::vector<arcticc::pb2::column_stats_pb2::ColumnStatsType> stat_types;
-    std::vector<size_t> data_col_offsets;
-    for (auto& segment : segments) {
-        arcticc::pb2::column_stats_pb2::ColumnStatsHeader header;
-        auto metadata = segment.metadata();
-        util::check(metadata != nullptr, "Column stats segment has no metadata");
-        bool unpacked = metadata->UnpackTo(&header);
-        util::check(unpacked, "Could not unpack column stats metadata?");
-
-        // Build reverse lookup: stats_seg_offset -> (data_col_offset, type)
-        ankerl::unordered_dense::map<uint32_t, std::pair<uint32_t, arcticc::pb2::column_stats_pb2::ColumnStatsType>>
-                offset_lookup;
-        for (const auto& [data_col_offset, entry_list] : header.stats_by_column()) {
-            for (const auto& entry : entry_list.entries()) {
-                offset_lookup[entry.stats_seg_offset()] = {data_col_offset, entry.type()};
-            }
-        }
-
-        for (const auto& [idx, field] : folly::enumerate(segment.descriptor().fields())) {
-            auto new_type = field.type();
-
-            if (auto it = field_name_to_index.find(std::string{field.name()}); it != field_name_to_index.end()) {
-                auto& merged_type = type_descriptors.at(field_name_to_index.at(std::string{field.name()}));
+// Distinct stat columns in first-appearance order, each widened to a type that can hold every
+// component's value for it. Dynamic schema can give a different type per row slice.
+std::vector<StatColumn> collect_stat_columns(
+        const std::vector<ColumnStatsComponent>& components,
+        ankerl::unordered_dense::map<std::string, size_t>& name_to_index
+) {
+    std::vector<StatColumn> stat_columns;
+    for (const auto& component : components) {
+        for (const auto& stat : component.stats) {
+            auto new_type = make_scalar_type(stat.value.data_type());
+            if (auto it = name_to_index.find(stat.segment_column_name); it != name_to_index.end()) {
+                auto& merged_type = stat_columns.at(it->second).type_descriptor;
                 auto opt_common_type = has_valid_common_type(merged_type, new_type);
                 internal::check<ErrorCode::E_ASSERTION_FAILURE>(
                         opt_common_type.has_value(),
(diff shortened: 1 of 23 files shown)


python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -152,24 +152,6 @@ def test_column_stats_nan_values(lmdb_version_store, any_output_format):
     assert_stats_equal(column_stats, expected_column_stats)
 
 
-def test_column_stats_only_nan_values(lmdb_version_store, any_output_format):
```
poodlewars, line 169: Covered by `test_column_stats_nan_values`

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -248,32 +230,6 @@ def test_column_stats_nat_values(lmdb_version_store, any_output_format):
     assert raw_stats["v1_MAX(col_1)"].values.view("int64")[3] == nat_sentinel
 
 
-def test_column_stats_only_nat_values(lmdb_version_store, any_output_format):
```
poodlewars, line 265: Covered by `test_column_stats_nat_values`

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -1,94 +1,149 @@
...
-                    // Skip start_index and end_index which are not statistics
-                    auto& [dcol, stype] = offset_lookup.at(idx);
-                    stat_types.emplace_back(stype);
-                    data_col_offsets.emplace_back(dcol);
-                }
-            }
+// The target is the descriptor's field type, which merge_descriptors resolved as a common type over
+// every slice's type for that column, so the static_cast is OK. The source type is needed to
+// interpret Value's raw bytes, hence the nested visit.
+void set_stat_value(Column& column, size_t row, const Value& value) {
```
poodlewars, line 74: I'm a bit sad about this approach to populating the column stats segment. Ideas welcome...

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -1,94 +1,149 @@
...
+    SegmentInMemory seg(Sparsity::PERMITTED);
+    seg.init_column_map();
+    seg.descriptor().set_index(IndexDescriptorImpl{IndexDescriptor::Type::ROWCOUNT, 0});
+
+    auto start_row_col = std::make_shared<Column>(make_scalar_type(DataType::UINT64), Sparsity::NOT_PERMITTED);
+    auto end_row_col = std::make_shared<Column>(make_scalar_type(DataType::UINT64), Sparsity::NOT_PERMITTED);
+    for (const auto& component : components) {
+        start_row_col->push_back<uint64_t>(component.start_row);
+        end_row_col->push_back<uint64_t>(component.end_row);
+    }
```
vasil-pashov: Nit
1. I think we can pre-allocate the columns
2. I'd place this inside the for loop above (the one checking for correctness) to avoid multiple iterations over the components.

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -1299,3 +1319,372 @@ def test_column_stats_drop_tiny_thread_pool(
...
+            "start_row": pl.Series([s for s, _ in expected_slices], dtype=pl.UInt64),
+            "end_row": pl.Series([e for _, e in expected_slices], dtype=pl.UInt64),
+            "v1_MIN(col_1)": pl.Series([int(v) for v in col_1_min], dtype=pl.Int64),
+            "v1_MAX(col_1)": pl.Series([int(v) for v in col_1_max], dtype=pl.Int64),
+            "v1_NAN_COUNT(col_1)": pl.Series(col_1_nan, dtype=pl.UInt64),
+            "v1_NULL_COUNT(col_1)": pl.Series(null_counts, dtype=pl.UInt64),
+            "v1_MIN(col_2)": pl.Series(col_2_min, dtype=pl.Float64),
+            "v1_MAX(col_2)": pl.Series(col_2_max, dtype=pl.Float64),
+            "v1_NAN_COUNT(col_2)": pl.Series(col_2_nan, dtype=pl.UInt64),
+            "v1_NULL_COUNT(col_2)": pl.Series(null_counts, dtype=pl.UInt64),
```
alexowens90, line 1499: I've got out of the loop, what's the reasoning behind storing separate NaN and null counts?

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -1299,3 +1319,372 @@ def test_column_stats_drop_tiny_thread_pool(
...
+        ([(9, 10), (0, 4)], [(0, 3), (3, 6), (9, 11)]),
+        # Fill the hole
+        ([(0, 4), (9, 10), (6, 7)], [(0, 3), (3, 6), (6, 9), (9, 11)]),
+        # Repeating a range replaces its rows rather than duplicating them
+        ([(0, 4), (0, 4)], [(0, 3), (3, 6)]),
+        # You need all segments to cover (2,10)
+        ([(0, 4), (2, 10)], [(0, 3), (3, 6), (6, 9), (9, 11)]),
+        # [3, 6) is recalculated, [0, 3) is kept, [6, 9) is new
+        ([(0, 4), (5, 9)], [(0, 3), (3, 6), (6, 9)]),
+    ],
```
alexowens90, line 1577: Probably the most common case would be to use the output of `read_index` and then do 1 (or more) row slice at a time, so worth testing `row_ranges=[(0, 3), (3, 6), (6, 9), (9, 11)]` and `row_ranges=[(0, 6), (6, 11)]`

python/tests/unit/arcticdb/test_column_stats_creation.py
```diff
@@ -1299,3 +1319,372 @@ def test_column_stats_drop_tiny_thread_pool(
     lib.drop_column_stats_experimental(sym)
     with pytest.raises(StorageException):
         lib.get_column_stats_info_experimental(sym)
+
+
+def test_column_stats_duplicate_index_values_across_slice_boundary(
+    lmdb_version_store_tiny_segment, column_stats_filtering_enabled
```
alexowens90: Any reason not to use in-memory store across these tests?

python/tests/unit/arcticdb/test_column_stats_optimisation.py
```diff
@@ -2323,3 +2369,33 @@ def test_column_stats_still_prunes_when_nan_does_not_save_row(
...
+
+
+def write_six_segments(lib):
+    """Six single-slice segments, col_1 ascending in pairs, two rows each."""
+    for i in range(6):
+        df = pd.DataFrame(
+            {"col_1": [2 * i + 1, 2 * i + 2]},
+            index=pd.date_range(pd.Timestamp("2000-01-01") + pd.Timedelta(days=2 * i), periods=2),
+        )
+        lib.append(sym, df)
```
alexowens90: Why not just use a lib with 2 rows per segment?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2162,21 +2162,29 @@ AtomKey index_key_to_column_stats_key(const IndexTypeKey& index_key) {
...
-            folly::collectAll(std::move(index_future), std::move(column_stats_future)).get();
+    folly::Try<std::pair<VariantKey, SegmentInMemory>> index_try;
+    std::optional<folly::Try<std::pair<VariantKey, SegmentInMemory>>> column_stats_try;
+    if (column_stats_future) {
+        auto [idx_try, stats_try] = folly::collectAll(std::move(index_future), std::move(*column_stats_future)).get();
+        index_try = std::move(idx_try);
+        column_stats_try = std::move(stats_try);
+    } else {
+        index_try = std::move(index_future).getTry();
+    }
```
alexowens90: Can this be simplified to:
```
    std::optional<folly::Try<std::pair<VariantKey, SegmentInMemory>>> column_stats_try;
    if (range_given) {
        column_stats_try = store->read(column_stats_key, stats_read_opts).getTry();
    }
    auto index_try = std::move(index_future).getTry();
```

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2200,103 +2208,76 @@ void create_column_stats_impl(
                     arcticdb::proto::descriptors::NormalizationMetadata::InputTypeCase::kMsgPackFrame,
             "Cannot create column stats on pickled data"
     );
+    user_input::check<ErrorCode::E_INVALID_USER_ARGUMENT>(
```
alexowens90, line 2067: These checks could be pre-IO

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2200,103 +2208,76 @@ void create_column_stats_impl(
...
         log::version().warn("Cannot create empty column stats");
         return;
     }
 
+    std::vector<ColumnStatsRow> old_components;
+    if (column_stats_try) {
+        if (column_stats_try->hasException<storage::KeyNotFoundException>()) {
+            // no stats yet, nothing to preserve
+        } else {
+            column_stats_try->throwUnlessValue();
```
alexowens90, line 2089: Doesn't the `hasException` guarantee this won't throw?

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2200,103 +2208,76 @@ void create_column_stats_impl(
...
-        }
-    }
-
     auto clause = column_stats.clause();
     if (!clause.has_value()) {
         log::version().warn("Cannot create empty column stats");
         return;
     }
 
+    std::vector<ColumnStatsRow> old_components;
```
alexowens90: Why are these called`*_components`? If it's just related to the involvement of the `ComponentManager` then they're bad names

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2200,103 +2208,76 @@ void create_column_stats_impl(
...
+        }
+    }
+
     auto pipeline_context = std::make_shared<PipelineContext>();
     pipeline_context->stream_id_ = versioned_item.key_.id();
     auto read_query = std::make_shared<ReadQuery>(std::vector{std::make_shared<Clause>(std::move(*clause))});
+    read_query->row_range = row_range;
+    if (date_range) {
+        read_query->row_filter = *date_range;
+    }
```
alexowens90: We could use `NativeVersionStore._get_read_query` to align more closely with other methods

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -106,6 +161,82 @@ std::string type_to_operator_string(ColumnStatTypeInternal type) {
...
+                    static_cast<size_t>(start_row_col.row_count()) == segment.row_count() &&
+                    static_cast<size_t>(end_row_col.row_count()) == segment.row_count(),
+            "Column stats start_row/end_row columns must be dense and cover every row of the segment"
+    );
+
+    std::vector<ColumnStatsRow> components(segment.row_count());
+    for (size_t row = 0; row < segment.row_count(); ++row) {
+        components.at(row).start_row =
+                *segment.scalar_at<uint64_t>(static_cast<position_t>(row), start_row_column_offset);
+        components.at(row).end_row = *segment.scalar_at<uint64_t>(static_cast<position_t>(row), end_row_column_offset);
```
alexowens90: `Column` also has `scalar_at`

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -106,6 +161,82 @@ std::string type_to_operator_string(ColumnStatTypeInternal type) {
...
+            );
+            const auto stats_seg_offset = entry.stats_seg_offset();
+            internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+                    stats_seg_offset > end_row_column_offset && stats_seg_offset < fields.size(),
+                    "Column stats header entry stats_seg_offset {} is out of range for a segment with {} fields",
+                    stats_seg_offset,
+                    fields.size()
+            );
+
+            const auto field_name = fields.at(stats_seg_offset).name();
```
alexowens90: nit: Only used to construct `segment_column_name`

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -106,6 +161,82 @@ std::string type_to_operator_string(ColumnStatTypeInternal type) {
...
+
+            const auto& column = segment.column(stats_seg_offset);
+            const std::string segment_column_name{field_name};
+            details::visit_type(column.type().data_type(), [&]([[maybe_unused]] auto tag) {
+                using type_info = ScalarTypeInfo<decltype(tag)>;
+                for_each_enumerated<typename type_info::TDT>(column, [&](const auto& it) {
+                    components.at(static_cast<size_t>(it.idx()))
+                            .stats.emplace_back(ColumnStatValue{
+                                    segment_column_name,
```
alexowens90: Is this used? Presumably it is a deterministic function of `type` and `data_col_offset`?

cpp/arcticdb/pipeline/column_stats_types.hpp
```diff
@@ -0,0 +1,39 @@
...
+    size_t data_col_offset;
+    Value value;
+};
+
+/// One per row slice. Added to the ComponentManager as a standalone entity by
+/// ColumnStatsGenerationClause::process and read back by create_column_stats_impl via
+/// process_entities
+struct ColumnStatsRow {
+    uint64_t start_row;
+    uint64_t end_row;
```
alexowens90: `RowRange` would make the intent clearer

cpp/arcticdb/pipeline/column_stats.cpp
```diff
@@ -1,94 +1,149 @@
...
-                    data_col_offsets.emplace_back(dcol);
-                }
-            }
+// The target is the descriptor's field type, which merge_descriptors resolved as a common type over
+// every slice's type for that column, so the static_cast is OK. The source type is needed to
+// interpret Value's raw bytes, hence the nested visit.
+void set_stat_value(Column& column, size_t row, const Value& value) {
+    details::visit_type(column.type().data_type(), [&column, row, &value](auto target_tag) {
+        using TargetRaw = typename ScalarTypeInfo<decltype(target_tag)>::RawType;
+        details::visit_type(value.data_type(), [&column, row, &value](auto source_tag) {
```
alexowens90, line 77: I think `value.data_type() == column.type().data_type()` by construction? So can avoid the double visitation at least

cpp/arcticdb/pipeline/column_stats.cpp
alexowens90: I don't think you can do much better tbh. The `std::vector<ColumnStatsRow>` fundamentally represents a (sparse, i.e. possibly missing some columns) row of the output, so inverting it into our columnar format is always inefficient.
The closest thing we have is the `RowBuilder` pattern, although that is very much implemented with strong assumptions about what callers are doing, so probably won't extend naturally to this.

cpp/arcticdb/pipeline/column_stats_types.hpp
```diff
@@ -0,0 +1,39 @@
...
+    std::string segment_column_name;
+    ColumnStatTypeInternal type;
+    size_t data_col_offset;
+    Value value;
+};
+
+/// One per row slice. Added to the ComponentManager as a standalone entity by
+/// ColumnStatsGenerationClause::process and read back by create_column_stats_impl via
+/// process_entities
+struct ColumnStatsRow {
```
alexowens90, line 31: It feels like most usages of this amount to a `std::map<RowRange, std::vector<ColumnStatValue>>`, although that smells bad in the `ComponentManager` where each call to `process` would add a 1-element map

cpp/arcticdb/pipeline/test/test_column_stats_build_segment.cpp
```diff
@@ -0,0 +1,304 @@
...
+    return ColumnStatValue{name, ColumnStatTypeInternal::MIN_V1, price_data_col_offset, Value{value, data_type}};
+}
+
+ColumnStatsRow component(uint64_t start_row, uint64_t end_row, std::vector<ColumnStatValue> stats) {
+    return ColumnStatsRow{start_row, end_row, std::move(stats)};
+}
+
+position_t column_index(const SegmentInMemory& seg, std::string_view name) {
+    auto idx = seg.column_index(name);
+    EXPECT_TRUE(idx.has_value()) << "column " << name << " missing";
```
alexowens90: `ASSERT_TRUE`, otherwise the dereference below is just gonna crash anyway

alexowens90 (approved):

vasil-pashov (approved):

python/tests/unit/arcticdb/test_column_stats_creation.py
poodlewars: I think they can be folded together - added a note about this [here](https://manwiki.maninvestments.com/display/AlphaTech/%5BArcticDB+Planning%5D+Column+Stats). I didn't realise at the time that we always treated NaN and null similarly, and in the current pruning code we always look at the sum of the NaN and null counts.

cpp/arcticdb/version/version_core.cpp
poodlewars: No, that only guards against a `KeyNotFoundException`. We don't want to swallow any others.

cpp/arcticdb/pipeline/test/test_column_stats_build_segment.cpp
poodlewars: Changed to an `arcticdb::util` assertion. ASSERT_TRUE can only be used in void returning functions.

cpp/arcticdb/pipeline/column_stats.cpp
poodlewars: With dynamic schema the `value` can have a narrower type than the column. For example if a column has some slices of type int32 and some of int64, then the column type will be int64 but the value types will be a mix of int32 and int64.

cpp/arcticdb/pipeline/column_stats.cpp
poodlewars: Yeah I've changed to key off the type and data_col_offset rather than these strings. It's a huge cleanup, thank you!

python/tests/unit/arcticdb/test_column_stats_creation.py
poodlewars: Having said that, Arrow has `is_null(nan) == False` which is an argument for keeping the counts separate.

PR #3324, opened 2026-08-15
version: pipeline asynchronous reads during compaction in do_compact

### LINKED ISSUE #3248: finalize_staged_data() reads staged segments serially: 195k segments took 3.4h on S3
### Describe the bug

`finalize_staged_data` fetches staged segments one at a time through a blocking read loop on the calling thread. On remote storage, finalize time is roughly one storage round-trip plus decode per 100k-row segment. The writes on the same path are already parallel, so staging can finish quickly while finalize becomes the bottleneck.

In one production run, 19.5B rows were staged as 20,747 `StageResult` tokens (~195k segments). Staging across ~200 distributed writers took 334s. Finalize on one host took 12,226s (3.4h). During finalize, py-spy showed one thread inside `compact_incomplete` at moderate CPU, which matches a serial network-read + decompress loop.

## Root cause

`compact_incomplete_impl` delegates to `do_compact`
([version_core.cpp#L2620](https://github.com/man-group/ArcticDB/blob/229c5ab09a86dd096c877f16d668020d2479e14b/cpp/arcticdb/version/version_core.cpp#L2620)), whose loop calls `sk.segment(store)` once per staged segment
([version_core.hpp#L281-L292](https://github.com/man-group/ArcticDB/blob/229c5ab09a86dd096c877f16d668020d2479e14b/cpp/arcticdb/version/version_core.hpp#L281-L292)). That resolves to `SliceAndKey::ensure_segment`, which is a synchronous
`store->read_sync(*key_)`
([frame_slice.cpp#L19-L22](https://github.com/man-group/ArcticDB/blob/229c5ab09a86dd096c877f16d668020d2479e14b/cpp/arcticdb/pipeline/frame_slice.cpp#L19-L22)): one blocking GET + decode per segment, on the calling thread.

The write side of the same loop is already async and bounded: `store->write_maybe_blocking` futures, gated by a semaphore sized from `VersionStore.NumSegmentsLiveDuringCompaction`
([version_core.hpp#L275](https://github.com/man-group/ArcticDB/blob/229c5ab09a86dd096c877f16d668020d2479e14b/cpp/arcticdb/version/version_core.hpp#L275)). Only the reads are serial. The code is unchanged on current master.

## Observed results

On S3 (staged and finalized from one host):

| Token shape | ~Segments | Finalize | Per segment |
|---|---|---|---|
| 10 rows x 1 int col | 100 | 0.9s | 9ms |
| 10k rows x 1 int col | 100 | 1.0s | 10ms |
| 1M rows x 1 int col | 1,000 | 17.3s | 17ms |
| 100k rows x 13 cols (12 string) | 100 | 13.1s | 131ms |

Per-segment cost starts at a ~9-10ms round-trip floor and grows with segment byte size (~3MB int segments 17ms, ~12MB string segments 131ms; production 20-60MB segments ~63ms). On `lmdb://`, the same script gives 0.1-2ms/segment across the board, so the cost is storage round-trips and bytes rather than CPU.

Production scale, same pattern:

| Run | Tokens | ~Segments | Finalize | Per segment |
|---|---|---|---|---|
| 19.5B rows | 20,747 | ~195,000 | 12,226s | ~63ms |
| 426M rows | 456 | ~4,300 | 259-266s | ~61ms |

## Steps/Code to Reproduce

Point `ARCTIC_URI` at any networked backend. LMDB hides the issue because there is no remote round-trip:

```python
"""ARCTIC_URI="s3://..." python arcticdb_finalize_bench.py"""
import os
import time

import numpy as np
import pandas as pd
from arcticdb import Arctic

URI = os.environ.get("ARCTIC_URI", "lmdb:///tmp/finalize_bench_lmdb")
rng = np.random.default_rng(0)

ac = Arctic(URI)
lib = ac.get_library("finalize_bench", create_if_missing=True)


def make_df(rows: int, start: pd.Timestamp, n_str_cols: int = 0) -> pd.DataFrame:
    index = pd.date_range(start, periods=rows, freq="ms")
    data = {"value": np.arange(rows, dtype=np.int64)}
    for col in range(n_str_cols):
        data[f"s{col}"] = pd.array(rng.integers(10**8, 10**9, rows).astype(str), dtype=object)
    return pd.DataFrame(data, index=index)


def bench(n_tokens: int, rows_per_token: int, n_str_cols: int = 0) -> None:
    symbol = f"bench_{rows_per_token}_{n_str_cols}"
    tokens = []
    started = time.monotonic()
    for i in range(n_tokens):
        df = make_df(rows_per_token, pd.Timestamp("2000-01-01") + pd.Timedelta(days=i), n_str_cols)
        tokens.append(lib.stage(symbol, df, validate_index=True))
    stage_s = time.monotonic() - started

    started = time.monotonic()
    lib.finalize_staged_data(symbol, mode="WRITE", stage_results=tokens, delete_staged_data_on_failure=True)
    finalize_s = time.monotonic() - started

    segments = n_tokens * max(1, rows_per_token // 100_000)
    print(
        f"{n_tokens} tokens x {rows_per_token:>9,} rows x {1 + n_str_cols:>2} cols "
        f"(~{segments} segments): stage {stage_s:7.1f}s, "
        f"finalize {finalize_s:7.1f}s ({finalize_s / segments * 1000:6.0f} ms/segment)",
        flush=True,
    )
    lib.delete(symbol)


if __name__ == "__main__":
    for rows in (10, 10_000, 1_000_000):
        bench(100, rows)
    bench(100, 100_000, n_str_cols=12)
```

## Expected results

`do_compact` should overlap segment reads instead of issuing one blocking `read_sync` at a time. The aggregator consumes segments in order, so a bounded sliding-window read-ahead using the existing async `store->read()` should work: keep up to `NumSegmentsLiveDuringCompaction` reads in flight, then consume them in sequence. That preserves ordering and memory bounds while removing the per-segment latency stall. `defragment_symbol_data_impl` should benefit too, since it shares `do_compact`.

## OS / Environment

- ArcticDB 6.18.1, Python 3.11, Linux x86_64
- Storage: S3; library with default options (100k rows per segment)
- Staging via `Library.stage()` / `StageResult` tokens, finalize via `finalize_staged_data(symbol, mode="WRITE", stage_results=tokens, delete_staged_data_on_failure=True)`

### Backend storage used

VAST S3

### Additional Context

_No response_

patrickswedish:
﻿### Summary
This PR addresses issue #3248 by migrating `compact_incomplete_impl` (`finalize_staged_data`) to use ArcticDB's canonical `read_modify_write` processing pipeline (`read_process_and_collect` with `RemoveColumnPartitioningClause`), as requested by @phoebusm.

### Changes
1. **Pipeline Integration (`version_core.cpp`)**:
   - In `compact_incomplete_impl`, construct a `ReadQuery` with `RemoveColumnPartitioningClause{pipeline_context->incompletes_after()}` and call `read_process_and_collect(store, pipeline_context, pipeline_read_query, pipeline_read_options).get()`.
   - Asynchronously reads and decompresses incomplete segments in parallel on `io_executor()` and `cpu_executor()` with bounded admission (`Processing.SegmentReadWindow`).
   - Unifies the architecture of `compact_incomplete_impl` with `defragment_symbol_data_impl` and `sort_merge_impl`.

2. **Streamlined `do_compact` (`version_core.hpp`)**:
   - Removed the bespoke sliding-window prefetch queue (`prefetched_queue`, `PrefetchedSegment`, `fetch_next`), allowing `do_compact` to directly consume in-memory segments without remote I/O blocking.
   - Retained asynchronous streaming writes via `SegmentAggregator` and `NativeSemaphore(n_segments_live_during_compaction())`.
   - Maintained exception safety and cleanup with `cleanup_written_keys()`.

3. **Test Suite**:
   - Python tests in `test_stage.py` for varying pipeline concurrency, single-slot sequential control, and error rollback.
   - C++ tests in `test_version_store.cpp` validating parallel read admission and out-of-order storage completion FIFO ordering.

Closes #3248

### DIFF
--- a/cpp/arcticdb/version/test/test_version_store.cpp
+++ b/cpp/arcticdb/version/test/test_version_store.cpp
@@ -11,6 +11,7 @@
 #include <arcticdb/version/version_store_api.hpp>
 #include <arcticdb/entity/types.hpp>
 #include <arcticdb/storage/memory/memory_storage.hpp>
+#include <arcticdb/storage/test/in_memory_store.hpp>
 #include <arcticdb/stream/test/stream_test_common.hpp>
 #include <arcticdb/util/test/generators.hpp>
 #include <arcticdb/util/allocator.hpp>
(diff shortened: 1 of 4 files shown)


phoebusm:
Hi @patrickswedish thanks for you PR. Indeed you are right. We have identified the same issue as well.
We have a new `read_modify_write` pipeline and we have discussed moving `compact` over to, which then gets parallel reads+processing for free.
It would be great if you could update your implementation accordingly, if that works for you.

patrickswedish:
﻿Hi @phoebusm,

Thank you for the guidance! I have updated the PR to migrate `compact_incomplete_impl` over to ArcticDB's canonical `read_modify_write` processing pipeline using `read_process_and_collect` with `RemoveColumnPartitioningClause{incompletes_after}`.

Summary of updates:
1. **Pipeline Migration**: `compact_incomplete_impl` now uses `read_process_and_collect` to schedule parallel asynchronous reads and decompression on `io_executor` / `cpu_executor` with bounded admission (`Processing.SegmentReadWindow`), unifying its implementation with `defragment_symbol_data_impl` and `sort_merge_impl`.
2. **Simplified `do_compact`**: Removed the custom prefetch sliding-window queue from `do_compact()`, so it now directly consumes in-memory segments while continuing to stream asynchronous output writes via `SegmentAggregator` and `NativeSemaphore`.
3. **Tests & Clean Commit**: Added comprehensive unit tests in Python (`test_stage.py`) and C++ (`test_version_store.cpp`) for pipeline concurrency, sequential control, out-of-order completion, and failure rollback, reverted unrelated test noise, and kept the PR as a single clean atomic commit.

Ready for your review!

PR #3327, opened 2026-08-18
2- Column stats - apply stats to all null segments, create stats even if there are duplicate column names, create stats over index column

poodlewars:
- Apply column stats filtering to all-null slices. If a slice has min=max=null then all values in it are null, and those stats will have `DataType::UNKNOWN`. We can apply column stats filtering to these.
- Create column stats even if there are duplicate column names. We were wrongly blowing up and refusing to create stats if there were any duplicates. This is a hangover from the old string-based column stats layout, whereas now we encode column stats in the dedicated `ColumnStatsHeader` in the segment proto and duplicate column names are safe (although cause various downstream issues with the QueryBuilder).
- Create column stats over the index column. This helps users who have an out of order index to use stats to still get some amount of filtering, rather than having to do full table scans.

Depends on #3321


alexowens90 (approved):

poodlewars:
Closing in favour of #3387

PR #3326, opened 2026-08-18
Unify `combine_schema` method and have consistent error handling among `append`, `update`, `concat`

IvoDD:
### Fixed

| Ref | What |
|---|---|
| [9797097831](https://man312219.monday.com/boards/7852509418/pulses/9797097831) | A RangeIndex name mismatch on append silently let the last write win. Index names are now required to match whether or not the index is physically stored. |
| [12029540807](https://man312219.monday.com/boards/7852509418/pulses/12029540807) | A mismatched index timezone on append/update silently overwrote the stored one. |
| [12781487305](https://man312219.monday.com/boards/7852509418/pulses/12781487305) | Concat now works with empty dataframes similar to append (ignores empty symbols) |

### Opened

| Ref | What |
|---|---|
| [12821228270](https://man312219.monday.com/boards/7852509418/pulses/12821228270) | Use `schema_combine` for incomplete segments as well |
| [12781411945](https://man312219.monday.com/boards/7852509418/pulses/12781411945) | `concat` of Series with mixed signedness does not promote to float64, where the equivalent DataFrame column does. |

## What this does

Append, update and concat each had diverged in how they combine `StreamDescriptor` and `NormalizationMetadata`. I.e. `OutputSchema`. This PR replaces all three with a single `combine_schema`, driven by explicit policies:

| Policy | Values |
|---|---|
| `MissingColumnPolicy` | `STRICT` (static append/update), `DROP` (inner join), `KEEP` (dynamic append/update, outer join) |
| `TypePromotionPolicy` | `STATIC`, `DYNAMIC`, `MOST_PERMISSIVE` (concat only) |
| `RequiredNameMismatchPolicy` | `RAISE` (append/update), `RECONCILE_TO_UNNAMED` (concat) |

The new `schema_combine.hpp` replaces methods like: `fix_normalization_or_throw`,
`fix_descriptor_mismatch_or_throw`, `get_merged_tsd`, `merge_normalization_metadata`, `join_indexes`,
`generate_index_descriptor`, `add_index_fields`, `inner_join`, `outer_join`, `required_fields_count`.

Incomplete segments schema handling is left for a follow up PR.

## Structure

Two commits, deliberately:

1. **`Extra testing for schema combinations`** — tests only, and passes on `master`. Every test this PR adds
   is here, so the second commit can show all behavior changes.
2. **`Unify schema and normalization combining into combine_schema`** — the implementation. Every behaviour
   change is therefore visible as a *modification* to a test that passed in the first commit.

[...]

PR #3333, opened 2026-08-18
Enhancement 11864701485: arrow string formatting norm metadata

alexowens90:
#### Reference Issues/PRs
[11864701485](https://man312219.monday.com/boards/7852509418/pulses/11864701485)

Unless explicitly overridden at the library/read call/column level, string columns written using either pyarrow or polars will return as the same type (small string/large string/categorical). The exception is small string columns written with pyarrow being read back into polars, as polars does not support small strings. These are mapped to large string columns.

### DIFF
--- a/cpp/arcticdb/arrow/arrow_utils.cpp
+++ b/cpp/arcticdb/arrow/arrow_utils.cpp
@@ -498,6 +498,23 @@ std::optional<ArrowMeta::ColumnMeta> generate_column_metadata(const sparrow::arr
         }
         break;
     }
+    case DataType::UTF_DYNAMIC32:
+    case DataType::UTF_DYNAMIC64: {
+        opt_column_meta.emplace();
+        if (array.dictionary().has_value()) {
+            opt_column_meta->set_string_format(proto::descriptors::ArrowStringFormat::CATEGORICAL);
+        } else if (array.data_type() == sparrow::data_type::LARGE_STRING) {
+            opt_column_meta->set_string_format(proto::descriptors::ArrowStringFormat::LARGE_STRING);
+        } else if (array.data_type() == sparrow::data_type::STRING) {
+            opt_column_meta->set_string_format(proto::descriptors::ArrowStringFormat::SMALL_STRING);
+        } else {
+            internal::raise<ErrorCode::E_ASSERTION_FAILURE>(
+                    "Unexpected Arrow data type in generate_column_metadata {}",
+                    sparrow::data_type_to_format(array.data_type())
+            );
+        }
+        break;
+    }
     default:
         break;
     }
@@ -611,12 +628,13 @@ std::pair<std::vector<Column>, entity::StreamDescriptor> record_batches_to_colum
 }
 
 RecordBatchData empty_record_batch_from_descriptor(
-        const entity::StreamDescriptor& stream_desc, const ArrowOutputConfig& arrow_output_config,
+        const entity::StreamDescriptor& stream_desc, ArrowOutputConfig&& arrow_output_config,
         const std::optional<ankerl::unordered_dense::set<std::string_view>>& columns,
         const proto::descriptors::NormalizationMetadata& norm_meta
 ) {
     // The logic here is similar to empty_arrow_array_for_column, but there the string format is dictated by the
     // column's buffers, whereas here we rely on the ArrowOutputConfig
+    modify_arrow_output_config_from_norm_meta(norm_meta, arrow_output_config);
     const auto& default_string_format = arrow_output_config.default_string_format_;
     const auto& per_column_string_format = arrow_output_config.per_column_string_format_;
     sparrow::record_batch record_batch;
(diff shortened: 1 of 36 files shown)


alexowens90:
The changes in `ingestion_stress_test.cpp/test_memory_storage.cpp/test_sparse.cpp/test_version_store.cpp` just reflect some signature changes from `const&` to `&` or `&&`

cpp/arcticdb/pipeline/read_options.hpp
```diff
@@ -125,26 +125,28 @@ class ReadOptions {
...
+    }
+
     [[nodiscard]] ReadOptions clone() const { return ReadOptions(std::make_shared<ReadOptionsData>(*data_)); }
 
   private:
     explicit ReadOptions(std::shared_ptr<ReadOptionsData> data) : data_(std::move(data)) {}
     std::shared_ptr<ReadOptionsData> data_ = std::make_shared<ReadOptionsData>();
 };
 
-using ReadOptionsPerSymbol = std::variant<ReadOptions, std::vector<ReadOptions>>;
```
alexowens90, line 144: The non-vector variant was unused except accidentally in a C++ test

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -1773,3 +1781,564 @@ def test_validate_index_arrow_batch(lmdb_version_store_arrow, method, validate_i
...
+            lib.batch_read([sym], date_ranges=[date_range])[sym].data
+            if batch
+            else lib.read(sym, date_range=date_range).data
+        )
+        assert received.schema.field(0).type == pa.timestamp("ns")
+        assert received.schema.field(1).type == pa.string()
+        assert received.schema.field(2).type == pa.large_string()
+        assert received.schema.field(3).type == pa.dictionary(pa.int32(), pa.large_string())
+        assert_arrow_equal(table.slice(10), received)
+
```
IvoDD, line 1840: After the #3326 PR it might be worth to make `combine(large_string, string) == large_string`.
What do you think?

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -1773,3 +1781,564 @@ def test_validate_index_arrow_batch(lmdb_version_store_arrow, method, validate_i
     else:
         result = ops[method]()
         assert all(not isinstance(r, DataError) for r in result)
+
+
+@pytest.mark.parametrize("batch", [False, True])
+class TestStringFormatRoundtrip:
```
IvoDD, line 1751: Apart from adding the new test coverage what do you think about simplifying our existing tests?

E.g. see TODO [here](https://github.com/man-group/ArcticDB/blob/e2e130e5b9b2cf898fbdaeb55f9b26959783f6c3/python/tests/unit/arcticdb/version_store/test_arrow_writes.py#L725)

A quick ask for claude which tests can be simplified with this change yielded:
```
  Fully redundant — drop the kwarg entirely

  test_arrow_writes.py (single write, requested format = written format):
  - 118 test_basic_write_strings
  - 174 test_write_multiple_record_batches
  - 192 test_write_with_index
  - 249 test_write_sliced
  - 341-342 test_many_record_batches_many_slices
  - 372 test_many_record_batches_edge_cases
  - 388 test_write_view
  - 408 test_write_view_strings
  - 729 test_write_categorical_strings (already has a # TODO: Remove string format... comment)
  - 749 test_write_mixed_categorical_and_variable_length_columns (same TODO)
  - 784 test_append
  - 876 test_update
  - 1063 test_staging_without_sorting
  - 1103 test_staging_with_sorting
  - 1129 test_staging_with_sorting_secondary_column
  - 1417 test_sparse_write_different_types
  - 1442 test_sparse_write_multiple_columns_different_types
  - 1478 test_sparse_many_different_size_batches
  - 1540 test_sparse_write_many_batches_many_slices
  - 1570 test_sparse_write_view
  - 1609 test_sparse_write_with_index

  test_arrow_sparse.py — all 9 sites using module-level STRING_FORMAT_PER_COLUMN (lines 71, 74, 117, 120, 164, 168, 447, 557, 626): str_col/large_str_col entries always match what's written. Simplest fix is shrinking the constant itself:
  STRING_FORMAT_PER_COLUMN = {"cat_str_col": pa.dictionary(pa.int32(), pa.large_string())}

  The string_format_kwargs helper in tests/util/arrow.py (lines 54-64) — used only at the 8 sites above (test_arrow_writes.py: 192, 249, 784, 876, 1103, 1129; test_arrow_sparse.py: 447, 557) plus _to_large_string (48-51). Once those call sites are simplified, this helper and
  _to_large_string become dead code and can be deleted along with their import.

```

cpp/arcticdb/pipeline/read_frame.cpp
```diff
@@ -112,8 +113,23 @@ void finalize_segment_setup(
...
+
+void modify_read_options_from_norm_meta(const PipelineContext& context, ReadOptions& read_options) {
+    if (context.has_normalization()) {
+        modify_read_options_from_norm_meta(context.normalization(), read_options);
+    }
+}
+
+SegmentInMemory allocate_frame(const std::shared_ptr<PipelineContext>& context, ReadOptions& read_options) {
     ARCTICDB_SAMPLE_DEFAULT(AllocateFrame)
+    modify_read_options_from_norm_meta(*context, read_options);
```
IvoDD: This feels like a fairly arbitrary place to modify the read_options in place.

I think it is correct since all uses of the read_options's arrow string config are after this but it has a couple of downsides:
- We need to remember to clone the ReadOptions coming from python
- I find it harder to reason about `ReadOptions&` over `const ReadOption&` (opens the door for further mutations)
- If in the future for some reason we want to use the string format before `allocate_frame` we'll need to move the `modify_read_options_from_norm_meta` as well.

I think it might be better to either: 
1. make handler methods like `output_type_and_block_config`, `handle_type` etc. take the `norm_meta`.
2. Do the `modify_read_options_from_meta` directly after normalization metadata is read.

I think I prefer 1 but it's going to change a lot of API calls so I'm also fine with leaving as is.

cpp/arcticdb/arrow/arrow_utils.cpp
```diff
@@ -675,4 +693,90 @@ RecordBatchData empty_record_batch_from_descriptor(
...
+                                                   : arrow_config.default_string_format_;
+        for (auto& [col_name, final_string_format] : arrow_config.per_column_string_format_) {
+            if (final_string_format == ArrowOutputStringFormat::UNSPECIFIED) {
+                if (auto it = columns_norm_meta.find(col_name); it != columns_norm_meta.end()) {
+                    const auto& col_meta = it->second;
+                    if (col_meta.has_string_format()) {
+                        final_string_format =
+                                convert_arrow_string_format_enum(col_meta.string_format(), arrow_config.output_format);
+                    } else {
+                        final_string_format = default_string_format;
```
IvoDD: A little duplication with the ternary operator here.

Might be a bit clearer to first populate the `arrow.default_string_format_` with the if below and then iterate over the per_column_string_format:
```
        if (arrow_config.default_string_format_ == ArrowOutputStringFormat::UNSPECIFIED) {
            // Add any columns from the norm meta NOT already present in per_column_string_format_
            for (const auto& [col_name, col_meta] : columns_norm_meta) {
                if (!arrow_config.per_column_string_format_.contains(col_name) && col_meta.has_string_format()) {
                    arrow_config.per_column_string_format_[col_name] =
                            convert_arrow_string_format_enum(col_meta.string_format(), arrow_config.output_format);
                }
            }
            arrow_config.default_string_format_ = ArrowOutputStringFormat::LARGE_STRING;
        }

        for (auto& [col_name, final_string_format] : arrow_config.per_column_string_format_) {
            if (final_string_format == ArrowOutputStringFormat::UNSPECIFIED) {
                if (auto it = columns_norm_meta.find(col_name); it != columns_norm_meta.end()) {
                    const auto& col_meta = it->second;
                    if (col_meta.has_string_format()) {
                        final_string_format =
                                convert_arrow_string_format_enum(col_meta.string_format(), arrow_config.output_format);
                    } else {
                        final_string_format = arrow_config.default_string_format_;
                    }
                } else {
                    final_string_format = arrow_config.default_string_format_;
                }
            }
        }
```

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -1245,10 +1267,6 @@ void register_bindings(py::module& version, py::exception<arcticdb::ArcticExcept
         }
     });
 
-    version.def("write_dataframe_to_file", &write_dataframe_to_file);
-
-    version.def("read_dataframe_from_file", &read_dataframe_from_file);
```
IvoDD, line 1250: Nit: I doubt anyone will complain but from a v1 user perspective this is removing a "public" API

It's available through `nvs.version_store.read_dataframe_from_file`. We should at some point rename `version_store` to `_version_store`

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -912,8 +922,14 @@ void register_bindings(py::module& version, py::exception<arcticdb::ArcticExcept
                                 std::make_shared<std::any>(TypeHandlerRegistry::instance()->get_handler_data(
                                         read_options.output_format_for_frame()
                                 ));
+                        // ReadOptions copy-ctor is shallow, so passing by copy does not achieve the same thing as the
```
IvoDD: I guess it would be easier to not forget `clone` if we override the copy constructor to do a deep copy and pass by value.

cpp/arcticdb/version/python_bindings.cpp
```diff
@@ -247,11 +252,15 @@ void register_bindings(py::module& version, py::exception<arcticdb::ArcticExcept
...
+                             ArrowOutputFormat output_format) {
+                     return ArrowOutputConfig{
+                             default_string_format, std::move(per_column_string_format), output_format
+                     };
                  }),
-                 py::arg("default_string_format") = ArrowOutputStringFormat::LARGE_STRING,
-                 py::arg("per_column_string_format") = std::unordered_map<std::string, ArrowOutputStringFormat>{});
+                 py::arg("default_string_format") = ArrowOutputStringFormat::UNSPECIFIED,
+                 py::arg("per_column_string_format") = std::unordered_map<std::string, ArrowOutputStringFormat>{},
+                 py::arg("output_format") = ArrowOutputFormat::PYARROW);
```
IvoDD: The output_format feels like it should be a required kwarg.

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: So the `large_to_small` column in `test_modify_string_type` would stay as large instead of swapping to small? Potentially, although the downside is then not being able to "modify" the column type to small string

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: Good spot. Turns out `STRING_FORMAT_PER_COLUMN` also isn't needed so removed that as well

cpp/arcticdb/pipeline/read_frame.cpp
alexowens90: I agree the placement is fairly arbitrary. There are also issues with recursively normalized data which I hadn't considered:
https://github.com/man-group/ArcticDB/pull/3333#discussion_r3823141052
I'm not sure the `[Batch]ReadOptions` pimpl pattern is correct anymore. It was only there to make the object cheap to copy so folly threads could keep underlying objects alive after the API call returned in error scenarios.
I'll address the rest of the comments in one commit, and then think about the best approach to this in a follow up.
I don't like 1. as you end up doing the same work repeatedly.

cpp/arcticdb/arrow/arrow_utils.cpp
alexowens90: I agree that impl is a bit clearer, will swap it out. I don't understand the ternary operator reference?

cpp/arcticdb/version/python_bindings.cpp
alexowens90: They were [duplicated definitions](https://github.com/man-group/ArcticDB/blob/16644c0ad4a28d5cf1ec2a2354cfdc3b14c04fb2/cpp/arcticdb/version/python_bindings.cpp#L284-L298), I just noticed because the `read_dataframe_from_file` signature changed so it stopped compiling.

cpp/arcticdb/version/python_bindings.cpp
alexowens90: Will take care of as part of [this](https://github.com/man-group/ArcticDB/pull/3333#discussion_r3855025884)

alexowens90:
> The changes in `ingestion_stress_test.cpp/test_memory_storage.cpp/test_sparse.cpp/test_version_store.cpp` just reflect some signature changes from `const&` to `&` or `&&`

This comment no longer applies after addressing review comments

cpp/arcticdb/pipeline/read_frame.cpp
IvoDD: Nice, I like how you combined the pandas object string type handling with the new arrow modification.

IvoDD (approved):

PR #3337, opened 2026-08-20
Enhancement 18053144964: fix get description output with arrow written data

alexowens90:
#### Reference Issues/PRs
[18053144964](https://man312219.monday.com/boards/7852509418/pulses/18053144964)

Correctly populates all fields returned by `get_info` (V1 API) and `get_description` (V2 API) when data is written with Arrow.

### DIFF
--- a/python/tests/unit/arcticdb/test_arrow_api.py
+++ b/python/tests/unit/arcticdb/test_arrow_api.py
@@ -400,3 +400,88 @@ def test_read_batch_and_join_strings(mem_storage, lib_name, default, per_column)
     assert result.schema.field(2).type == per_column or default or pa.large_string()
     expected_df = pd.concat([df_1, df_2]).reset_index(drop=True)
     assert_frame_equal_with_arrow(expected_df, result)
+
+
+# See test with the same name in test_arrow_read.py for V1 API equivalent
+def test_arrow_written_data_get_info_non_timeseries(mem_library):
+    lib = mem_library
+    sym = "test_arrow_written_data_get_info_non_timeseries"
+    lib._nvs._set_allow_arrow_input()
+    table = pa.table({"col": pa.array(np.arange(10), type=pa.int64())})
+    lib.write(sym, table)
+    desc = lib.get_description(sym)
+    assert len(desc.columns) == 1
+    assert desc.columns[0].name == "col"
+    assert len(desc.index) == 0
+    assert "INT64" in str(desc.columns[0].dtype)
+    assert np.isnat(desc.date_range[0]) and np.isnat(desc.date_range[1])
+    assert desc.index_type == "NA"
+    assert desc.row_count == 10
+    assert desc.sorted == "UNKNOWN"
+
+
+# See test with the same name in test_arrow_read.py for V1 API equivalent
+@pytest.mark.parametrize(
+    "table",
+    [
+        pytest.param(
+            pa.table(
+                {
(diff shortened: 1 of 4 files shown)


python/tests/unit/arcticdb/test_arrow_api.py
```diff
@@ -400,3 +400,88 @@ def test_read_batch_and_join_strings(mem_storage, lib_name, default, per_column)
...
+        assert "NANOSECONDS_UTC64" in str(desc.columns[0].dtype)
+    assert len(desc.index) == (1 if has_index else 0)
+    if has_index:
+        assert desc.index[0].name == "ts"
+        assert "NANOSECONDS_UTC64" in str(desc.index[0].dtype)
+    if has_index:
+        assert desc.date_range == (table.to_pandas()["ts"][0], table.to_pandas()["ts"][19])
+    else:
+        assert np.isnat(desc.date_range[0]) and np.isnat(desc.date_range[1])
+    assert desc.index_type == "NA"
```
IvoDD, line 482: Shouldn't `index_type=='index'` in case `has_index=True`?

python/tests/unit/arcticdb/test_arrow_api.py
```diff
@@ -400,3 +400,88 @@ def test_read_batch_and_join_strings(mem_storage, lib_name, default, per_column)
...
+    assert "UINT8" in str(desc.columns[-1].dtype)
+    if not has_index:
+        assert desc.columns[0].name == "ts"
+        assert "NANOSECONDS_UTC64" in str(desc.columns[0].dtype)
+    assert len(desc.index) == (1 if has_index else 0)
+    if has_index:
+        assert desc.index[0].name == "ts"
+        assert "NANOSECONDS_UTC64" in str(desc.index[0].dtype)
+    if has_index:
+        assert desc.date_range == (table.to_pandas()["ts"][0], table.to_pandas()["ts"][19])
```
IvoDD, line 479: This is somewhat surprising for `unsorted, has_index=True, validate_index=False` (it would return a crossed date range)

I guess we can use `NaT` if sort order is `UNKOWN` for arrow. I don't have a strong feeling about this.

python/tests/unit/arcticdb/test_arrow_api.py
alexowens90: Claude made the same comment, it's a bit of a mess. The `index` field with Pandas data is actually populated as `timeseries_descriptor.normalization.df.common.WhichOneof("index_type")`, which is always one of `index` or `multi_index`, even if the data isn't a timeseries. I think I prefer leaving the implementation as-is, and improving the `SymbolDescription` docstring to state that the field is only ever not `NA` for Pandas data, as it represents the Pandas concept of an index.

python/tests/unit/arcticdb/test_arrow_api.py
alexowens90: The root of this problem is that `validate_index` is overloaded. `compute_index_sortedness` is called only when `validate_index == True`, but the argument defaults to `False` on the V1 API. When this function isn't called, we set to `SortedValue::UNKNOWN`, but most V1 API users would probably rather see the date range here assuming their data was well constructed. At least the sortedness flag is also in the `SymbolDescription`, which gives a hint that the date range might not be reliable.

IvoDD (approved):

PR #3338, opened 2026-08-21
Implement row slicing for merge update

vasil-pashov:
This branch implements row slicing for merge update.
Monday Ticket: 12846079278

Row slicing can happen only if the strategy allows insertion. It's impossible to trigger for update-only strategy.

There are two cases based on the index of the dataframe. Both cases use the `ReslicingInfo` class and the strategy of max size = `4/3 * rows_per_segment` from the compaction algorithm. Neither of the two options allocates one big chunk and then slices it. Both options use the `ReslicingInfo` to see how much row slices are there and and create that many slices with the row count prescribed by `ReslicingInfo`

# Merge update
## `Datetime`

For `Datetime` the logic is in 2 places.

* `update_and_insert` finds the number of unmatched rows and computes the total row count for the insertion. It creates the `ReslicingInfo` and then calls `merge` column by column.
* `merge` is the function that performs the merging of source and target and the updates (if required). It's mostly unchanged. It still calls `advance_output` to move the tip of the output where new data is placed but now it tracks which row slice is used and if rows per slice is exceeded it moves to the next slice.

Keeping track of new row sliced becomes a bit harder. The non-sliced version had 1:1 mapping between input row range and output row range. Now one input row range can be split into several output row ranges. A new component is introduced which keeps track of this. Each entity produced by the MergeUpdate clause has `MergeUpdateRowSlicingInfoComponent` it tracks the number of segments the input row range was split into, the index of the row range this entity represents and the number of rows in the current row range. Version core then groups all entities by their input row range.

## `Rowrange`

Row range must be performed after the pipeline ends in order to gather all unmatched rows and combine them. All unmatched rows are still appended at the end, but the `ReslicingInfo` is used to compute the number of row slices. This has an optimization of the previous iteration that column slices are processed (data is copied from source to the result) in parallel

# Other changes
* WriteClause is changed to forward all entities it got from the previous clause instead of discarding them and creating new entities.

### DIFF
--- a/cpp/arcticdb/column_store/column_reslicer.hpp
+++ b/cpp/arcticdb/column_store/column_reslicer.hpp
@@ -18,6 +18,8 @@
 
 namespace arcticdb {
 
+[[nodiscard]] size_t min_rows_per_segment(size_t rows_per_segment);
+
 [[nodiscard]] size_t max_rows_per_segment(size_t rows_per_segment);
 
 // Helper class used by both the column and segment reslicer classes
--- a/cpp/arcticdb/column_store/column_reslicer.hpp
+++ b/cpp/arcticdb/column_store/column_reslicer.hpp
@@ -14,39 +14,62 @@
 
 #include <arcticdb/column_store/column.hpp>
 #include <arcticdb/column_store/string_pool.hpp>
-#include <arcticdb/entity/type_utils.hpp>
 #include <arcticdb/processing/expression_node.hpp>
 
 namespace arcticdb {
+
+[[nodiscard]] size_t max_rows_per_segment(size_t rows_per_segment);
 
 // Helper class used by both the column and segment reslicer classes
 class ReslicingInfo {
   public:
     ReslicingInfo(uint64_t _total_rows, uint64_t _rows_per_segment) :
-        total_rows(_total_rows),
-        num_segments((total_rows + _rows_per_segment - 1) / _rows_per_segment),
-        rows_per_segment(total_rows / num_segments),
-        num_remainder_segments(total_rows % num_segments),
-        num_exact_segments(num_segments - num_remainder_segments) {
+        total_rows_(_total_rows),
+        num_segments_((total_rows_ + _rows_per_segment - 1) / _rows_per_segment),
+        rows_per_segment(total_rows_ / num_segments_),
+        num_remainder_segments(total_rows_ % num_segments_),
+        num_exact_segments(num_segments_ - num_remainder_segments) {
         auto output_rows = num_exact_segments * rows_per_segment + num_remainder_segments * (rows_per_segment + 1);
         util::check(
-                output_rows == total_rows,
+                output_rows == total_rows_,
                 "SlicingInfo input rows does not match constructed output rows {} != {}",
-                total_rows,
+                total_rows_,
                 output_rows
         );
     }
 
     ARCTICDB_MOVE_COPY_DEFAULT(ReslicingInfo)
 
-    uint64_t rows_in_slice(uint64_t idx) const {
+    [[nodiscard]] uint64_t rows_in_slice(uint64_t idx) const {
         return idx < num_exact_segments ? rows_per_segment : rows_per_segment + 1;
     }
 
-    uint64_t total_rows;
-    uint64_t num_segments;
+    [[nodiscard]] uint64_t num_segments() const { return num_segments_; }
+
+    [[nodiscard]] uint64_t total_rows() const { return total_rows_; }
+
+    /// Inverse of ReslicingInfo::rows_in_slice: given a row index into the combined [0, total_rows) output, returns
+    /// the (slice index, offset within that slice) pair that rows_in_slice's slicing would place it in.
+    [[nodiscard]] std::pair<uint64_t, uint64_t> slice_and_offset_for_row(uint64_t global_row) const {
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                global_row < total_rows_,
+                "ReslicingInfo::slice_and_offset_for_row: row {} is out of bounds for {} total rows",
+                global_row,
+                total_rows_
+        );
+        const uint64_t exact_rows = num_exact_segments * rows_per_segment;
+        if (global_row < exact_rows) {
+            return {global_row / rows_per_segment, global_row % rows_per_segment};
+        }
+        const uint64_t remainder_row = global_row - exact_rows;
+        return {num_exact_segments + remainder_row / (rows_per_segment + 1), remainder_row % (rows_per_segment + 1)};
+    }
 
   private:
+    uint64_t total_rows_;
+    uint64_t num_segments_;
+
     // This is how many rows most segments will have
     uint64_t rows_per_segment;
     // This is how many segments will have rows_per_segment+1 rows
--- a/cpp/arcticdb/processing/clause_compact_data.cpp
+++ b/cpp/arcticdb/processing/clause_compact_data.cpp
@@ -26,13 +26,7 @@ CompactDataClause::CompactDataClause(uint64_t rows_per_segment, std::shared_ptr<
     if (frame && frame->num_rows > 0) {
         frame_ = std::move(frame);
     }
-    // Magic range of +-33% chosen so that:
-    // - 2 row slices with < min_rows_per_segment_ cannot be combined into one row slice with > max_rows_per_segment_
-    // - 1 row slice with a little more than max_rows_per_segment_ when split in half will still have
-    //   >= min_rows_per_segment_ in each resulting row slice
-    // If rows_per_segment_ == 1 min_rows_per_segment_ would be 0 without the std::max
-    min_rows_per_segment_ = std::max((2 * rows_per_segment_) / 3, uint64_t(1));
-    // If rows_per_segment_ == 2 max_rows_per_segment_ would be 2 without the std::max
+    min_rows_per_segment_ = min_rows_per_segment(rows_per_segment_);
     max_rows_per_segment_ = max_rows_per_segment(rows_per_segment_);
     clause_info_.input_structure_ = ProcessingStructure::ONE_COL_SLICE_MULTIPLE_ROW_SLICES;
     clause_info_.output_structure_ = ProcessingStructure::ONE_COL_SLICE_MULTIPLE_ROW_SLICES;
--- a/cpp/arcticdb/processing/clause_compact_data.cpp
+++ b/cpp/arcticdb/processing/clause_compact_data.cpp
@@ -33,7 +33,7 @@ CompactDataClause::CompactDataClause(uint64_t rows_per_segment, std::shared_ptr<
     // If rows_per_segment_ == 1 min_rows_per_segment_ would be 0 without the std::max
     min_rows_per_segment_ = std::max((2 * rows_per_segment_) / 3, uint64_t(1));
     // If rows_per_segment_ == 2 max_rows_per_segment_ would be 2 without the std::max
-    max_rows_per_segment_ = std::max((4 * rows_per_segment_) / 3, rows_per_segment_ + 1);
+    max_rows_per_segment_ = max_rows_per_segment(rows_per_segment_);
     clause_info_.input_structure_ = ProcessingStructure::ONE_COL_SLICE_MULTIPLE_ROW_SLICES;
     clause_info_.output_structure_ = ProcessingStructure::ONE_COL_SLICE_MULTIPLE_ROW_SLICES;
     clause_info_.can_combine_with_column_selection_ = false;
@@ -252,7 +252,7 @@ void CompactDataClause::add_segment_from_frame(
     ReslicingInfo reslicing_info{total_rows, max_rows_per_segment_};
     // Work out how many rows of the frame we need to combine with segments from disk
     uint64_t row_count{0};
-    for (size_t idx = 0; idx < reslicing_info.num_segments; ++idx) {
+    for (size_t idx = 0; idx < reslicing_info.num_segments(); ++idx) {
         row_count += reslicing_info.rows_in_slice(idx);
         if (row_count > total_rows_without_frame) {
             break;
--- a/cpp/arcticdb/processing/clause_merge_update.cpp
+++ b/cpp/arcticdb/processing/clause_merge_update.cpp
@@ -6,6 +6,8 @@
  * will be governed by the Apache License, version 2.0.
  */
 
+#include "util/collection_utils.hpp"
+
 #include <arcticdb/processing/clause.hpp>
 #include <arcticdb/processing/processing_unit.hpp>
 #include <arcticdb/column_store/string_pool.hpp>
@@ -15,6 +17,7 @@
 #include <arcticdb/version/schema_checks.hpp>
 #include <arcticdb/pipeline/slicing.hpp>
 #include <arcticdb/stream/index.hpp>
+#include <arcticdb/column_store/column_reslicer.hpp>
 #include <ankerl/unordered_dense.h>
 #include <boost/regex.hpp>
 
@@ -256,7 +259,7 @@ struct InsertTargetData {
     std::span<ColumnWithStrings> columns;
     TypeDescriptor type;
     TargetRange range;
-    StringPool& new_string_pool;
+    std::span<StringPool> new_string_pools;
 };
 
 std::vector<size_t> compute_target_slice_offset(const InsertTargetData& target) {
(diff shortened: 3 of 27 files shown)


cpp/arcticdb/version/version_core.cpp
```diff
@@ -1094,164 +1435,6 @@ folly::Future<std::vector<EntityId>> schedule_clause_processing(
...
-            read_window
-    );
-
-    auto processed = schedule_clause_processing(
-            component_manager, admission, std::make_shared<std::vector<std::shared_ptr<Clause>>>(read_query->clauses_)
-    );
-
-    return std::move(processed).via(&async::cpu_executor());
-}
-
```
vasil-pashov, line 1254: These are just moved to the anonymous namespace above

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2997,49 +3180,19 @@ folly::Future<ReadVersionOutput> read_frame_for_version(
...
-                            return std::move(*fut).thenValueInline([](SliceAndKey&& slice_and_key) {
-                                slice_and_key.unset_segment();
-                                return slice_and_key;
-                            });
-                        }
-                );
-                return folly::collect(std::move(write_segments_futures));
-            });
-}
-
```
vasil-pashov, line 3035: This is changed see the `read_modify_write_data_keys` above.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -439,76 +461,395 @@ std::vector<StreamDescriptor> split_rowrange_descriptor(
...
+            max_processing_units_in_flight,
+            read_window
+    );
+
+    auto processed = schedule_clause_processing(
+            component_manager, admission, std::make_shared<std::vector<std::shared_ptr<Clause>>>(read_query->clauses_)
+    );
+
+    return std::move(processed).via(&async::cpu_executor());
+}
```
vasil-pashov, line 786: Moved here without changes

cpp/arcticdb/version/version_core.cpp
```diff
@@ -439,76 +461,395 @@ std::vector<StreamDescriptor> split_rowrange_descriptor(
...
+
+folly::Future<std::vector<EntityId>> read_modify_write_data_keys(
```
vasil-pashov, line 788: Because the write clause now forwards entities this now returns the entityid the write clause forwarded. Other clauses can use these entities to to gather specific components. `SlicesAndKeys` are one of those components, merge update also needs `MergeUpdateRowSlicingInfoComponent`

python/arcticdb/util/test.py
```diff
@@ -1656,3 +1656,38 @@ def merge(
 
 def query_stats_operation_count(stats, operation, key_type):
     return stats.get("storage_operations", {}).get(operation, {}).get(key_type, {}).get("count", 0)
+
+
+def max_rows_per_segment(rows_per_segment: int) -> int:
```
IvoDD, line 1667: It looks like this is implemented in a few places. Can we move the min and max calculations here and reuse them across tests?

Other calculation [here](https://github.com/man-group/ArcticDB/blob/88375efdd3132683ede7ae5c98cb756b9225d6f1/python/tests/unit/arcticdb/version_store/test_compact_data.py#L100-L106)

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -3174,6 +3174,780 @@ def test_ambiguous_multi_match_raises(self, lmdb_library):
...
+                        "I",
+                    ],
+                }
+            )
+        assert_frame_equal(lib.read("sym").data, expected)
+        index_df = lib._dev_tools.library_tool().read_index("sym")
+        assert_index_key_structure_static_schema(index_df, rows_per_segment=3)
+
+    @pytest.mark.parametrize(
+        "num_unmatched, source_a, source_b, expected_a, expected_b, expected_num_segments, expected_data_key_writes",
```
IvoDD, line 3253: It looks like we don't use `num_unmatched` and `expected_num_segments`. I agree they are useful for reading of the test params. What do you think of also adding an assertion on `len(lib.read_index) == expected_num_segments`

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -3174,6 +3174,780 @@ def test_ambiguous_multi_match_raises(self, lmdb_library):
...
+                ]
+            ),
+        )
+        source = pd.DataFrame({"a": [999], "b": ["A"]}, index=pd.DatetimeIndex(["2024-01-03"]))
+        lib.write("sym", target)
+        qs.reset_stats()
+        with qs.query_stats():
+            lib.merge_experimental("sym", source, strategy=strategy, on=["a"])
+        stats = qs.get_query_stats()
+        assert query_stats_operation_count(stats, "Memory_PutObject", "TABLE_DATA") == 4
```
IvoDD: Strange that this is 4, aren't the first 2 row slices untouched by this? I get we process them all together but dedup should probably prevent us from writing the first two?

cpp/arcticdb/column_store/column_reslicer.hpp
```diff
@@ -14,39 +14,62 @@
...
+        ARCTICDB_DEBUG_CHECK(
+                ErrorCode::E_ASSERTION_FAILURE,
+                global_row < total_rows_,
+                "ReslicingInfo::slice_and_offset_for_row: row {} is out of bounds for {} total rows",
+                global_row,
+                total_rows_
+        );
+        const uint64_t exact_rows = num_exact_segments * rows_per_segment;
+        if (global_row < exact_rows) {
+            return {global_row / rows_per_segment, global_row % rows_per_segment};
```
IvoDD, line 65: Nit: Maybe `std::div`

python/tests/unit/arcticdb/version_store/test_merge_update.py
```diff
@@ -3174,6 +3174,780 @@ def test_ambiguous_multi_match_raises(self, lmdb_library):
...
+                    "2024-01-08",
+                    "2024-01-09",
+                    "2024-01-10",
+                    "2024-01-11",
+                ]
+            ),
+        )
+        assert_frame_equal(lib.read("sym").data, expected)
+        index_df = lib._dev_tools.library_tool().read_index("sym")
+        assert_index_key_structure_static_schema(index_df, rows_per_segment=3)
```
IvoDD, line 3559: Would be cool to assert the precise row structure here is `[3, 2, 3, 2]`. This shows that each row slice get sliced individually instead of calling defrag on the entire table which would probably give `[3, 3, 4]`

cpp/arcticdb/processing/component_manager.hpp
```diff
@@ -59,7 +63,7 @@ class ComponentManager {
 
     // Add a single entity with the components defined by args
     template<class... Args>
-    void add_entity(EntityId id, Args... args) {
+    void add_components(EntityId id, Args... args) {
```
IvoDD, line 66: Nit: Update inline comment to something like "Add components defined by args to entity id"

cpp/arcticdb/processing/clause_compact_data.cpp
```diff
@@ -33,7 +33,7 @@ CompactDataClause::CompactDataClause(uint64_t rows_per_segment, std::shared_ptr<
     // If rows_per_segment_ == 1 min_rows_per_segment_ would be 0 without the std::max
     min_rows_per_segment_ = std::max((2 * rows_per_segment_) / 3, uint64_t(1));
     // If rows_per_segment_ == 2 max_rows_per_segment_ would be 2 without the std::max
-    max_rows_per_segment_ = std::max((4 * rows_per_segment_) / 3, rows_per_segment_ + 1);
+    max_rows_per_segment_ = max_rows_per_segment(rows_per_segment_);
```
IvoDD, line 30: Nit: I think max_rows_per_segment calculate and min_rows_per_segment calculations should live in one place.

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -279,30 +281,32 @@ std::vector<size_t> compute_target_slice_offset(const InsertTargetData& target)
...
+        new_column_datas.emplace_back(new_columns.back()->data());
+        new_column_accessors.emplace_back(random_accessor<TargetColumnTypeDescriptorTag>(&new_column_datas.back()));
+    }
+    size_t new_column_row_slice_index{};
+    // Offset within the current output column
+    size_t new_column_row_idx{};
+    // Position in the combined [0, total_rows()) output
+    size_t output_row_idx{};
+    size_t rows_in_current_slice = reslicing_info.rows_in_slice(0);
+    auto new_column_it = new_column_datas.front().begin<TargetColumnTypeDescriptorTag>();
```
IvoDD, line 309: Just to confirm we can't end up with `num_segments()==0`? Theoretically if we have an empty rows slice `total_rows==0` would produce `num_segments==0`. Maybe an assertion to be sure would be good, unless this is checked somewhere higher up the stack.

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -508,10 +523,21 @@ Column merge(
             advance_output();
         }
     } else {
-        std::copy(source.data.begin() + source_row_idx, source.data.end(), new_column_it);
+        while (new_column_row_slice_index < new_columns.size()) {
+            const size_t free_elements_in_column =
+                    static_cast<size_t>(new_columns[new_column_row_slice_index]->row_count()) - new_column_row_idx;
+            std::copy_n(source.data.begin() + source_row_idx, free_elements_in_column, new_column_it);
+            source_row_idx += free_elements_in_column;
```
IvoDD, line 540: What do you think about reusing `advance_output()` here by adding `size_t step_size=1` argument to it and here calling with `step_size=free_elements_in_column`?

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -568,11 +596,13 @@ TargetRange get_target_start_end(std::span<const ProcessingUnit> row_slices) {
     if (row_slices.front().entity_fetch_count_) {
         if (row_slices.front().entity_fetch_count_->front() > 1) {
             const ColumnData index = row_slices.front().segments_->front()->column(0).data();
-            result.start_row_in_first_row_slice = first_different_index_value_position(index);
+            result.start_row_in_first_row_slice = first_different_index_value_position(index).value_or(0);
```
IvoDD: This change makes it so that if the first segment is of same value [5, 5, 5] we return the full date_range [0, 3) instead of what previously we returned [3, 3).

In practice this should not happen because all equally valued indices should get assigned only to one time slice. Not shared between two. I think an assert that `has_value()` would be better.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -439,76 +461,395 @@ std::vector<StreamDescriptor> split_rowrange_descriptor(
     return descriptors;
 }
 
+/// One instance is the single output data segment covering one column slice's columns for one output row slice's
+/// rows.
+struct MergeUpdateRowRangeInsertInfo {
+    /// Check which source rows within this range must be inserted. Note, not all source rows within this range will be
+    /// inserted.
+    RowRange source_range_to_check_for_appends{};
```
IvoDD: The name `_for_appends` got me a bit confused. 
I think `insert_candidates_source_range` or something might be better. As far asI understand it's the rowrange in the source we need to later filter on the bitset to produce the inserted segment.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -439,76 +461,395 @@ std::vector<StreamDescriptor> split_rowrange_descriptor(
     return descriptors;
 }
 
+/// One instance is the single output data segment covering one column slice's columns for one output row slice's
+/// rows.
+struct MergeUpdateRowRangeInsertInfo {
```
IvoDD: It's a bit sad that "RowRange" means two different things:
- Index type (timeseries vs row range)
- Row slice (represented by the RowRange type)

Hence this name is slightly confusing. Maybe something like `MergeUpdateSliceInfo` and comment it's only used for rowrange index or something. I don't feel strongly.

IvoDD (commented):
This PR introduces some extra complexity to "directly construct sliced segments" instead of "construct one segment and slice later". 

It is probably worth the extra complexity, but would be good to have some ASV benchmarks which measure merge_update with smaller segment_row_size so we stress this codepath, to make sure we don't regress

cpp/arcticdb/column_store/column_reslicer.hpp
alexowens90: It doesn't have unsigned overloads

cpp/arcticdb/version/version_core.cpp
```diff
@@ -439,76 +461,395 @@ std::vector<StreamDescriptor> split_rowrange_descriptor(
...
+                                         : read_query->clauses_.back()->clause_info().output_structure_;
+    read_query->clauses_.push_back(std::make_shared<Clause>(
+            WriteClause(target_partial_index_key, std::move(de_dup_map), store, write_clause_processing_structure)
+    ));
+    pipeline_context->set_output_schema(generate_output_schema(*pipeline_context, *read_query));
+
+    return read_and_schedule_processing(store, pipeline_context, read_query, read_options, component_manager)
+            .thenValue([component_manager = std::move(component_manager),
+                        read_query = std::move(read_query)](std::vector<EntityId>&& processed_entity_ids) {
+                std::vector<std::shared_ptr<folly::Future<SliceAndKey>>> slice_components =
```
alexowens90:
```suggestion
                std::vector<std::shared_ptr<folly::Future<SliceAndKey>>> slice_futures =
```

cpp/arcticdb/version/version_core.cpp
```diff
@@ -439,76 +461,395 @@ std::vector<StreamDescriptor> split_rowrange_descriptor(
...
+                                processed_entity_ids
+                        ));
+                auto write_segments_futures = util::reserve_vector<folly::Future<EntityId>>(slice_components.size());
+                for (auto&& [index, slice_future] : folly::enumerate(slice_components)) {
+                    write_segments_futures.push_back(
+                            std::move(*slice_future)
+                                    .thenValueInline([component_manager,
+                                                      entt = processed_entity_ids[index]](SliceAndKey&& slice_and_key) {
+                                        component_manager
+                                                ->remove_component<std::shared_ptr<folly::Future<SliceAndKey>>>(entt);
```
alexowens90: Could augment `get_entities` to optionally remove the component from entt as well to avoid this separate call

cpp/arcticdb/version/version_core.cpp
```diff
@@ -439,76 +461,395 @@ std::vector<StreamDescriptor> split_rowrange_descriptor(
...
+                std::vector<std::shared_ptr<folly::Future<SliceAndKey>>> slice_components =
+                        std::get<0>(component_manager->get_entities<std::shared_ptr<folly::Future<SliceAndKey>>>(
+                                processed_entity_ids
+                        ));
+                auto write_segments_futures = util::reserve_vector<folly::Future<EntityId>>(slice_components.size());
+                for (auto&& [index, slice_future] : folly::enumerate(slice_components)) {
+                    write_segments_futures.push_back(
+                            std::move(*slice_future)
+                                    .thenValueInline([component_manager,
+                                                      entt = processed_entity_ids[index]](SliceAndKey&& slice_and_key) {
```
alexowens90:
```suggestion
                                                      entity_id = processed_entity_ids[index]](SliceAndKey&& slice_and_key) {
```

cpp/arcticdb/version/version_core.cpp
alexowens90: Please please please put this sort of change in it's own commit

cpp/arcticdb/processing/component_manager.hpp
```diff
@@ -144,16 +144,26 @@ class ComponentManager {
...
-        return get_entities_impl<Args...>(ids, false);
+    std::tuple<std::vector<Args>...> get_components(const std::vector<EntityId>& ids) {
+        return get_components_impl<Args...>(ids, false, false);
+    }
+
+    // Get the requested components for a collection of entities and remove them from the entities. Returns a tuple of
+    // vectors, one for each component requested via Args
+    template<class... Args>
+    requires(!std::same_as<Args, EntityFetchCount> && ...)
+    std::tuple<std::vector<Args>...> get_components_and_remove_components(const std::vector<EntityId>& ids) {
```
alexowens90: `get_and_remove_components`

cpp/arcticdb/processing/clause_merge_update.cpp
IvoDD: Ah looks like you added an assertion higher up the stack [here](https://github.com/man-group/ArcticDB/pull/3338/changes/79404739a04c8659afea6966c3fb44210bdca158#diff-142c49046a430dcb6e5fe2c06139df92b0d1f700f622a42e1097dbb79984faf4R1151)

python/benchmarks/merge_functions.py
```diff
@@ -22,6 +22,10 @@
 WORK_BUCKET = "arcticdb-merge-work"
 ROWS_PER_SEGMENT = 100_000
 START_DATE = pd.Timestamp("1960-01-01")
+# Target of exactly one row slice, merged with a source large enough that reslicing splits it into MIN_SPLIT_SEGMENTS
+SPLIT_TARGET_ROWS = ROWS_PER_SEGMENT
+SPLIT_SOURCE_ROWS = 600_000
```
IvoDD: How long does the new benchmark take? A

lt this could be faster to run and excercise the slicing logic harder if we had configured a smaller ROWS_PER_SEGMENT for the slicing benchmark.

python/benchmarks/merge_functions.py
```diff
@@ -514,3 +526,204 @@ def time_merge(self, scenario, strategy, on_count, source_size):
...
+        super().__init__()
+        self.value_dtype = "string"
+        self.param_names = ["scenario", "strategy", "on_count", "source_size", "matched_slices", "num_unique_strings"]
+        self.params = [
+            [(SPLIT_TARGET_ROWS, 3)],  # scenario: (num_rows, num_value_cols) — exactly one row slice
+            ["update_and_insert"],  # strategy
+            [1],  # on_count
+            [SPLIT_SOURCE_ROWS],  # source_size (source row count)
+            [1],  # matched_slices: the whole source lands in the target's single row slice
+            [5_000, 100_00],  # num_unique_strings (size of the pool each column is drawn from)
```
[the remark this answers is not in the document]
IvoDD: +1 to claude

python/benchmarks/merge_functions.py
```diff
@@ -514,3 +526,204 @@ def time_merge(self, scenario, strategy, on_count, source_size):
 
     def peakmem_merge(self, scenario, strategy, on_count, source_size):
         self.merge(strategy)
+
+
+class MergeSplitSegmentsThinDatetime(MergeBase):
```
IvoDD, line 530: I remember there were some issues in using inheritance in ASV benchmarks, can you check the produced json is sensible?

I guess if the json doesn't get updated if we update `MergeBase` that could be a problem.

IvoDD (approved):

python/tests/unit/arcticdb/version_store/test_merge_update.py
vasil-pashov: Dedup is off by default. There was `test_index_value_spanning_three_slices_dedup` just below this one where it's enabled and the put object is 2 but I decided to parametrize this test.

python/benchmarks/merge_functions.py
vasil-pashov: Yes, that's true but it's the same if it uses free functions because ASV hashes only the body of the time_*. peakmem_.* methods but doesn't pull in code recursively.

This is modeled after what @alexowens90 did for defrag.

I remember there was an issue that if time_.* was in the base then inherited classes wouldn't run that and it won't be benchmarked but it's not an issue here.

python/benchmarks/merge_functions.py
vasil-pashov: It takes about 1min on local machine with 16 CPU and 24 I/O threads. I reduced the rows per segment and increased the source rows

PR #3342, opened 2026-08-24
Change The Exception When append() or update() Is Called On Symbol With Missing Version

markovskipetar:
## Problem
  
  Calling `append` or `update` on a symbol that has **no existing version**, with the
  "create if missing" toggle disabled, raised the wrong exception:

  - `append(..., write_if_missing=False)` → `InternalException`
  - `update(..., upsert=False)` → `InternalException`

  `InternalException` signals an internal assertion failure, not an ordinary user
  error. The **batch** equivalents (`append_batch` / `update_batch`) already raised
  the correct `NoSuchVersionException` (`E_NO_SUCH_VERSION`), so the two code paths
  were inconsistent.

  ## Fix
  
  The non-batch and batch paths share the internal functions
  `async_append_internal` / `async_update_internal`, which branched on a `batch`
  bool to pick the exception type (batch → `NoSuchVersionException`, non-batch →
  `InternalException`).

  - Both paths now raise `NoSuchVersionException` via
    `missing_data::check<E_NO_SUCH_VERSION>`.
  - The now-redundant `batch` parameter is removed from both functions and their
    call sites.

  Files: `cpp/arcticdb/version/local_versioned_engine.cpp`,
  `cpp/arcticdb/version/local_versioned_engine.hpp`.

  ## ⚠️ API break (targeted at 7.0.0)

  Non-batch `append` / `update` now raise **`NoSuchVersionException`** instead of
  **`InternalException`** when the symbol does not exist and
  `write_if_missing`/`upsert` is `False`. Code that caught `InternalException` for
  this case must be updated. Affects both V1 (`NativeVersionStore`) and V2
  (`Library`) APIs. No on-disk format change.

  ## Tests

  Added to `test_update.py` and `test_append.py`:

  - `update` on a missing symbol (`upsert=False`) raises `NoSuchVersionException`;
    `upsert=True` creates it.
  - `append` on a missing symbol (`write_if_missing=False`) raises
    `NoSuchVersionException`; `write_if_missing=True` creates it (V1 & V2 stores).

  Existing batch missing-symbol tests are unchanged and still pass.

### DIFF
--- a/python/tests/unit/arcticdb/version_store/test_append.py
+++ b/python/tests/unit/arcticdb/version_store/test_append.py
@@ -40,21 +40,6 @@ def create_random_data(at_date: datetime, num_cols: int = 5) -> pd.DataFrame:
     return pd.DataFrame(data=data, index=date_range, columns=[f"c{i + 1}" for i in range(num_cols)])
 
 
-def test_append_missing_symbol_no_write_if_missing_raises(lmdb_version_store):
-    lib = lmdb_version_store
-    df = pd.DataFrame({"a": [1, 2]}, index=pd.date_range("2024-01-01", periods=2))
-    with pytest.raises(NoSuchVersionException) as ex_info:
-        lib.append("does_not_exist", df, write_if_missing=False)
-    assert all(s in str(ex_info.value) for s in ["write_if_missing", "Cannot append", "does_not_exist"])
-
-
-def test_append_missing_symbol_write_if_missing_creates(lmdb_version_store):
-    lib = lmdb_version_store
-    df = pd.DataFrame({"a": [1, 2]}, index=pd.date_range("2024-01-01", periods=2))
-    lib.append("new_symbol", df, write_if_missing=True)
-    assert_frame_equal(lib.read("new_symbol").data, df)
-
-
 @pytest.mark.xfail(reason="Needs to be fixed with issue #496")
 def test_append_with_cont_mem_problem(sym, lmdb_version_store_tiny_segment_dynamic):
     set_config_int("SymbolDataCompact.SegmentCount", 1)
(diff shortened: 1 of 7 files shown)


python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -39,6 +40,29 @@ def create_random_data(at_date: datetime, num_cols: int = 5) -> pd.DataFrame:
...
 
+def test_append_missing_symbol_no_write_if_missing_raises(lmdb_version_store):
+    lib = lmdb_version_store
+    df = pd.DataFrame({"a": [1, 2]}, index=pd.date_range("2024-01-01", periods=2))
+    with pytest.raises(NoSuchVersionException) as ex_info:
+        lib.append("does_not_exist", df, write_if_missing=False)
+    assert all(s in str(ex_info.value) for s in ["write_if_missing", "Cannot append", "does_not_exist"])
+
+
+def test_append_missing_symbol_write_if_missing_creates(lmdb_version_store):
```
alexowens90: This and the test above belong in the `TestAppend` class so we cover `batch_append` as well

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -39,6 +40,29 @@ def create_random_data(at_date: datetime, num_cols: int = 5) -> pd.DataFrame:
...
+    lib = lmdb_version_store
+    df = pd.DataFrame({"a": [1, 2]}, index=pd.date_range("2024-01-01", periods=2))
+    lib.append("new_symbol", df, write_if_missing=True)
+    assert_frame_equal(lib.read("new_symbol").data, df)
+
+
+def test_append_missing_symbol_creates_v2(lmdb_library):
```
alexowens90: Move to `test_arctic.py`, and check if there is a test of the `append_batch` behaviour in `test_arctic_batch.py`

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -628,6 +634,22 @@ def test_update_not_sorted_range_index_exception(lmdb_version_store):
...
+        lib.update("does_not_exist", df)
+    assert all(s in str(ex_info.value) for s in ["upsert", "Cannot update", "does_not_exist"])
+
+
+def test_update_missing_symbol_upsert_creates(lmdb_library):
+    lib = lmdb_library
+    df = pd.DataFrame({"a": [1, 2]}, index=pd.date_range("2024-01-01", periods=2))
+    lib.update("new_symbol", df, upsert=True)
+    assert "new_symbol" in lib.list_symbols()
+    assert_frame_equal(lib.read("new_symbol").data, df)
```
alexowens90: V2 API tests?

python/tests/unit/arcticdb/version_store/test_append.py
```diff
@@ -212,6 +213,31 @@ def test_defragment_no_work_to_do(sym, lmdb_version_store):
...
+    def test_append_missing_symbol_write_if_missing_creates(self, lmdb_version_store_v1, batch, compact_data):
+        lib = lmdb_version_store_v1
+        sym = "new_symbol"
+        df = pd.DataFrame({"a": [1, 2]}, index=pd.date_range("2024-01-01", periods=2))
+        (
+            lib.batch_append([sym], [df], write_if_missing=True, compact_data=compact_data)
+            if batch
+            else lib.append(sym, df, write_if_missing=True, compact_data=compact_data)
+        )
+        assert_frame_equal(lib.read(sym).data, df)
```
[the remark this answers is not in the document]
alexowens90: The test in `test_append_compact_data.py` is a bit more comprehensive, it should be moved to here

alexowens90 (approved):

PR #3343, opened 2026-08-25
Consolidate Unfilterable-data Checks With Typed Error Codes

markovskipetar:
# Consolidate unprocessable-data checks with typed error codes

## Summary

Some symbols cannot have processing (row/column filtering, projection, aggregation,
resampling, `head`/`tail`/`row_range`, etc.) applied to them — pickled data, recursively
normalized data, and raw numpy arrays. Previously these cases were guarded by a scattered
mix of generic `util::check` / `util::check_rte` calls at different points on the read path
(and, for pickled data, by a check that was explicitly deferred to the next major release),
so the error you got depended on which entry point you came in through, and several entry
points had no guard at all.

This PR routes every processing entry point through a single guard,
`check_can_perform_processing`, and gives each unprocessable data kind its own dedicated
error code so callers get a meaningful `SchemaException` instead of an incidental or generic
error.

### One choke point for all processing entry points

- `check_can_perform_processing` is now called from `read_indexed_keys_to_pipeline` at two
  sites: the multi-key early return, and immediately after `pipeline_context->set_tsd(...)`
  (the checks need the tsd in order to inspect `has_normalization()` / `is_pickled()` /
  `is_numpy_array()`).
- The multi-key call previously made from `read_frame_for_version` is removed — it is now
  covered by the choke point.
- Every processing entry point therefore reaches the check: `setup_pipeline_context` (reads,
  lazy reads, batch reads, concat, `SetupPipelineContextTask`), `create_column_stats_impl`,
  `compact_data_explain_plan_impl`, `async_compact_data_impl`, `merge_update_impl`,
  `read_modify_write_internal` and `read_and_process`.

### Add `PipelineContext::is_numpy_array()` helper (no behavioural change)

Introduces a named `is_numpy_array()` accessor on `PipelineContext`, mirroring `is_pickled()`,
and replaces the inline `normalization().has_np()` check with it.

### Typed error codes

Each data kind now raises its dedicated error code (all `SchemaException`):

| Data kind              | Error code                                                        |
|------------------------|-------------------------------------------------------------------|
| Pickled                | `E_OPERATION_NOT_SUPPORTED_WITH_PICKLED_DATA` (4004)              |
| Recursively normalized | `E_OPERATION_NOT_SUPPORTED_WITH_RECURSIVE_NORMALIZED_DATA` (4005) |
| Numpy array            | `E_OPERATION_NOT_SUPPORTED_WITH_NUMPY_ARRAY` (4006)               |

The guard already raised 4005/4006 for recursive and numpy data; the pickled case and the
paths below are what change:

- Fold the pickled check into the guard (4004), replacing the generic `util::check` whose
  typed replacement was deferred to the next major release (ticket 18038782559).
- Remove the now-redundant `Cannot filter pickled data` and
  `Cannot use head/tail/row_range with pickled data` `check_rte`s in
  `do_direct_read_or_process`.
- Column stats on recursively normalized symbols: `E_UNSUPPORTED_INDEX_TYPE` → 4005.
- Multi-symbol join: recursively normalized data `user_input::check<E_INVALID_USER_ARGUMENT>`
  → `schema::check<4005>`; pickled data `E_DESCRIPTOR_MISMATCH` → 4004.
- Batch reads: `transform_batch_items_or_throw` now maps `PickledDataException` and
  `NumpyArrayDataException` onto the matching `DataError` codes (only the recursive one was
  mapped before), so `read_batch` reports 4004/4006 with `ErrorCategory.SCHEMA` rather than a
  generic code. Adds the `PickledDataException` / `NumpyArrayDataException` aliases and their
  `throw_error` specializations in `error_code.hpp`.

### Reject unprocessable data before query validation

`check_can_perform_processing` is reordered so unprocessable data is rejected with its
dedicated error code **before** the `date_range` / sorted-index validation runs. This means a
`date_range` read of pickled data reports `E_OPERATION_NOT_SUPPORTED_WITH_PICKLED_DATA`
rather than an incidental "non-timestamp index" complaint.

The `date_range`-on-non-timestamp-index check itself is also promoted from a generic
`util::check` to `schema::check<E_UNSUPPORTED_INDEX_TYPE>`, so it raises `SchemaException`
instead of `InternalException`.

### Allow compaction of pickled data

The consolidated check would otherwise reject all processing on pickled data, but
`compact_data` (directly, or via `append`'s `compact_data` argument) must work on pickled
symbols, just as it already does for numpy arrays. `CompactDataClause` is the one exempt
clause: the pickled branch is skipped during compaction, mirroring the existing numpy-array
handling.

## Tests

- New `python/tests/unit/arcticdb/version_store/test_unprocessable_data.py`, parametrized over
  the three data kinds × four processing kinds (filter, projection, groupby+agg, resample),
  covering: `read` with `query_builder` / `columns` / `date_range` / `row_range`, `head`,
  `tail`, lazy reads, `read_batch` `DataError` codes and `ErrorCategory.SCHEMA`,
  `read_batch_and_join` with per-symbol processing, read-modify-write, `merge` (including
  that a failed merge leaves the symbol untouched), and `create_column_stats`. It also pins
  the two exemptions: `concat` of plain numpy arrays still works.
- Removed the per-feature duplicates now superseded by the consolidated file:
  `test_partial_read_pickled_df` (`test_basic_version_store.py`), `test_group_pickled_symbol`
  and `test_group_pickled_symbol_dynamic` (`test_aggregation.py`),
  `test_filter_pickled_symbol` and `test_filter_date_range_pickled_symbol`
  (`test_filtering.py`), `test_head_pickled_symbol`, `test_tail_pickled_symbol` and
  `test_row_range_pickled_symbol`.


cpp/arcticdb/pipeline/pipeline_context.hpp
```diff
@@ -238,6 +238,12 @@ struct PipelineContext : public std::enable_shared_from_this<PipelineContext> {
                arcticdb::proto::descriptors::NormalizationMetadata::InputTypeCase::kMsgPackFrame;
     }
 
+    bool is_numpy_array() const {
+        util::check(tsd_.has_value(), "No normalization metadata defined");
+        return tsd_->proto().normalization().input_type_case() ==
+               arcticdb::proto::descriptors::NormalizationMetadata::InputTypeCase::kNp;
+    }
```
vasil-pashov: Flagging that this is branched from an older master and will conflict with https://github.com/man-group/ArcticDB/pull/3317 no need to take action now but be aware. The referred PR introduces a new function output_normalization. That is the function that this must use when conflicts are resolved.

nit: I prefer having most function implementations in .cpp files. Exceptions being performance critical functions in headers that must be inlined

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1552,8 +1536,28 @@ void check_can_perform_processing(
...
+        } else if (is_pickled && !is_compaction) {
+            schema::raise<ErrorCode::E_OPERATION_NOT_SUPPORTED_WITH_PICKLED_DATA>(
+                    "Cannot perform processing such as row/column filtering, projection, aggregation, resampling, "
+                    "etc.. on pickled data"
+            );
         }
     }
+
+    if (pipeline_context->has_on_disk_descriptor()) {
+        util::check(
```
alexowens90, line 1548: This should not be an internal exception, I would recommend `E_UNSUPPORTED_INDEX_TYPE`

cpp/arcticdb/version/version_core.cpp
```diff
@@ -57,6 +57,10 @@ namespace arcticdb::version_store {
 
 namespace ranges = std::ranges;
 
+void check_can_perform_processing(
+        const std::shared_ptr<PipelineContext>& pipeline_context, const ReadQuery& read_query
+);
```
alexowens90, line 62: Can just move the definition up here, no need for a forward declaration

cpp/arcticdb/version/version_core.cpp
alexowens90, line 1: I think the implementation of `check_can_perform_processing` is sound, but it's a bad code smell that it is called from 5 different places. There are also multiple code paths where it is called more than once. The old location of `read_indexed_keys_to_pipeline` seemed like the right one, why was it removed from there?

There are basically 3 code paths that need this kind of checking:

1. Normal reads (with or without `QueryBuilder`, as some of the checks are things like no date-range filter over recursively normalized data)
2. `concat` - this is really just a special case of read, that never goes through the direct-read path available in `do_direct_read_or_process`.
3. read-modify-write - Used by `compact_data` and `merge_update`.

python/tests/unit/arcticdb/version_store/test_unprocessable_data.py
```diff
@@ -0,0 +1,236 @@
...
+    assert result.error_code == ERROR_CODE_FOR[kind]
+    assert result.error_category == ErrorCategory.SCHEMA
+
+
+# --- multi-symbol join -------------------------------------------------------------------------
+
+
+@all_data_kinds
+@all_processing_kinds
+def test_read_batch_and_join_per_symbol_processing(in_memory_library, kind, processing):
```
alexowens90: It would be better not to parametrize this over `all_processing_kinds`, and check that a plain `concat` involving a non-supported data format raises even if there is no per-symbol processing

python/tests/unit/arcticdb/version_store/test_unprocessable_data.py
```diff
@@ -0,0 +1,236 @@
...
+
+@all_data_kinds
+def test_merge(in_memory_library, kind):
+    lib = in_memory_library
+    sym = write_unprocessable(lib._nvs, kind)
+    with expect_refusal(kind):
+        lib.merge(sym, MERGE_SOURCE)
+
+
+def test_merge_on_recursive_data_leaves_symbol_untouched(in_memory_library):
```
alexowens90: Can move to `test_nonreg_specific.py`

alexowens90 (approved):

PR #3344, opened 2026-08-25
Control matching NaN/None/NaT via parameter

vasil-pashov:
#### Reference Issues/PRs
Monday issue: 12896784915

Currently if a source column used for matching contains NaN/None/NaT that value is going to be equal to other NaN/None/NaT in a target column. This behavior makes sense, however there are two considerations

SQL does not consider such values equal. NaN is different than anything including NaN

It is more future proof to have these values different. With dynamic schema when a column is missing it makes sense to not match anything on that column. It will be easier to explain to users why nothing was matched if the default is to not match NaNs. The problem with int columns being backfilled to 0 stays, but there are already places in the processing pipeline where such columns behave like NaN

Actually making NaN/None/NaT match each other required additional code. Making them behave like they are different is the easier to implement (because a == b will always return false if a or b is NaN). Make this an option to the API and use the additional code to match such values if it's required.

This unifies the places in the code that check for n/a values. It's all now done in a single function is_na. The `NaNAwareComparator` became `NaAwaraComparator` and can now handle all N/A values.

Hypothesis tests are also unified. Now the tests for update only use the same strategy for generation of the data as the insertion tests. The final dataframe is generated first, then the we go backward from there by selecting some random rows which will be the rows that are inserted and some random rows which will be rows that are updated. See [this](https://github.com/man-group/ArcticDB/blob/16f406fa376451aad953d2cf1420c24ada25aa0e/python/tests/hypothesis/arcticdb/test_merge_update.py#L205)

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>

### DIFF
--- a/cpp/arcticdb/processing/clause_merge_update.cpp
+++ b/cpp/arcticdb/processing/clause_merge_update.cpp
@@ -26,14 +26,14 @@ namespace {
 using namespace arcticdb;
 
 template<typename TDT, typename T>
-concept sequience_type_raw_value =
+concept sequence_type_raw_value =
         util::type_descriptor_tag<TDT> && is_sequence_type(TDT::data_type()) &&
         util::any_of<std::remove_cvref_t<T>, PyObject*, typename TDT::RawType, std::optional<std::string_view>>;
 
 template<typename TDT, typename T>
 concept raw_value_for_type_descriptor =
         util::type_descriptor_tag<TDT> &&
-        (std::same_as<T, typename TDT::DataTypeTag::raw_type> || sequience_type_raw_value<TDT, T>);
+        (std::same_as<T, typename TDT::DataTypeTag::raw_type> || sequence_type_raw_value<TDT, T>);
 
 template<util::type_descriptor_tag TDT>
 using SourceRawType =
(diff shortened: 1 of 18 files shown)


cpp/arcticdb/version/version_store_api.cpp
```diff
@@ -1553,7 +1553,7 @@ void PythonVersionStore::force_delete_symbol(const StreamId& stream_id) {
 VersionedItem PythonVersionStore::merge(
         const StreamId& stream_id, const std::shared_ptr<convert::PandasData>& source, const py::object& norm,
         const py::object& user_meta, const bool prune_previous_versions, const bool upsert,
-        const py::tuple& py_strategy, std::vector<std::string> on
+        const py::tuple& py_strategy, std::vector<std::string> on, const bool match_na
 ) {
     const MergeStrategy strategy{
```
alexowens90, line 1560: What do you think about folding the `match_na` bool into the `MergeStrategy` class?

python/arcticdb/version_store/library.py
```diff
@@ -3540,9 +3541,19 @@ def merge_experimental(
             IMPORTANT: For date-time indexed data, the index is always included in matching and cannot be excluded.
 
             Note on equality semantics:
+                By default (`match_na=False`), a missing value in an `on` column never matches anything, on either
+                side:
```
alexowens90: I don't understand the "on either side" bit of this docstring

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -25,10 +25,49 @@
 namespace {
 using namespace arcticdb;
 
+template<typename TDT, typename T>
+concept sequience_type_raw_value =
```
alexowens90: nit: typo

cpp/arcticdb/processing/clause_merge_update.cpp
```diff
@@ -162,85 +201,74 @@ void merge_update_string_column(
...
-        const StringPool& target_string_pool, std::optional<ScopedGILLock>& scoped_gil_lock
+        SourceRawType<SourceTDT> source_value, typename TargetTDT::DataTypeTag::raw_type target_value,
+        const StringPool& target_string_pool, std::optional<ScopedGILLock>& scoped_gil_lock, bool match_na
 ) {
+
+    const bool source_is_na = is_na<SourceTDT>(source_value);
+    const bool target_is_na = is_na<TargetTDT>(target_value);
+    if (source_is_na || target_is_na) {
+        return match_na && source_is_na && target_is_na;
+    }
```
alexowens90: Can reuse `NaAwareComparator` here

docs/mkdocs/docs/notebooks/ArcticDB_merge.ipynb
alexowens90, line 1: In the string matching example, I would add another row to the source to get rid of the `Column Exchange does not have non null elements.` log (or I guess set the log level so it doesn't show up)

alexowens90 (approved):

PR #3345, opened 2026-08-26
Performance Improvement on VAST storage: Set CURLOPT_DNS_SHUFFLE_ADDRESSES in our S3 client

poodlewars:
Full scripts are available in [this branch](https://github.com/man-group/ArcticDB/tree/aseaton/connection-pool-sizing).

## S3: improve CNode distribution on VAST

VAST's DNS answers describe CNode IPs. When we make DNS requests in short bursts (like when ArcticDB establishes its connection pool) we get a non-uniform distribution of connections across CNodes, meaning that connections in our pool actually talk to the same CNodes, and their requests might queue on the CNode. This also leads to a poor distribution of connections across CNodes, which are a limited resource per CNode.

### Root cause

Running `dig +short <vast_hostname>` in a loop gives a uniform distribution, however the DNS server results seem to change on a relatively slow timescale and a C++ client can get "duplicate" results out of the DNS.

A minimal repro ([`dns_clustering_repro.cpp`](https://github.com/man-group/ArcticDB/blob/aseaton/connection-pool-sizing/pool_sizes_testing/dns_clustering_repro.cpp)) completes in ~29ms and produces only 4 distinct answer sets across 100 lookups (repeat counts 32, 32, 29, 7).

### Fix

Apply [CURLOPT_DNS_SHUFFLE_ADDRESSES ](https://curl.se/libcurl/c/CURLOPT_DNS_SHUFFLE_ADDRESSES.html) to the DNS results across all S3 backends. This is discussed below - it simplifies the implementation and did not harm PURE or AWS S3.

Each connection we establish does its own DNS lookup already (the ttl on this DNS is zero, deliberately so we get at least the slower shuffling described above), and this change shuffles those results so we end up with a uniform distribution.

### Measured impact: CNode distribution (VAST, real dev bucket)

Measured with [`profile_s3_dns_shuffle_cnode_trace.py`](https://github.com/man-group/ArcticDB/blob/aseaton/connection-pool-sizing/python/benchmarks/non_asv/profile_s3_dns_shuffle_cnode_trace.py), which parses libcurl TRACE logs for `Trying <ip>...` lines to count distinct CNode IPs touched per trial, with a large connection pool size.

| Metric | Before | After |
|---|---|---|
| Distinct CNode IPs touched | 22 | 41 |
| Top-1 IP share | 18.8% | 7.9% |
| Top-3 IP share | 37.6% | 18.8% |
| Top-5 IP share | 50.5% | 28.7% |

| Pool size | Mean distinct CNodes, before | Mean distinct CNodes, after |
|---|---|---|
| 12 | 3.75 | 9.40 |
| 72 | 22 | 41 |
| 96 | 18.55 | 41.95 |

### Measured impact on VAST

Measured with [`profile_s3_dns_shuffle_ab.py`](https://github.com/man-group/ArcticDB/blob/aseaton/connection-pool-sizing/python/benchmarks/non_asv/profile_s3_dns_shuffle_ab.py). One subprocess per measurement, with the connection pool warmed with a large read first. Interleaved runs with and without the fix.

Pool size = 12 used 8 CPU and 12 IO threads. Pool size = 96 used 64 CPU and 96 IO threads.

#### Testing with large reads

`batch_read_large_multiseg` (8 columns, 100 segments, 10M rows/unit, ~640MB/unit, batch size 32, ~20.5GB/measurement).

These show a significant improvement from the fix on reads.

**`batch_read_large_multiseg`**

| Pool size | n | mean(on), s | mean(off), s | mean(diff=on-off), s | median(diff), s | relative effect | p-value |
|---|---|---|---|---|---|---|---|
| 12 | 15 pairs | 6.7872 | 8.7917 | -2.0044 | -2.1243 | -22.8% | **0.0019** |
| 96 | 20 pairs | 1.8041 | 2.0098 | -0.2057 | -0.0749 | -10.2% | **0.0231** |

**`batch_write_large_multiseg`**

This shows mized results - 10% faster on reasonable pool sizes, 10% slower on extremely large pool sizes.

| Pool size | n | mean(on), s | mean(off), s | mean(diff=on-off), s | median(diff), s | relative effect | p-value |
|---|---|---|---|---|---|---|---|
| 12 | 50 pairs |  | 27.02 | -2.83 | -1.39 | -10.5% | **0.0021** |
| 96 | 50 pairs | | 7.48 | 0.67 | 0.83 | 8.9% | **0.0008** |

#### Testing with relatively small objects

Cells:
- `read_fanout` — single-symbol `read()` fanning out over 2x pool size 2000 row segments (2,000 rows/segment)
- `read_fanout_large` Same but with 200k rows and 32 cols
- `batch_read_small` / `batch_write_small` — `batch_read`/`batch_write` over 2x pool size 2000 row single-segment objects
- `batch_read_large` / `batch_write_large` — `batch_read`/`batch_write` over 24 objects (200,000 rows, 32 columns, ~51MB each)

##### pool_size=12, full run (n=50 pairs/cell)

| Cell | mean(on), s | mean(off), s | mean(diff=on-off), s | p-value | Significant? |
|---|---|---|---|---|---|
| `read_fanout` | 0.1587 | 0.1341 | +0.0246 | 0.0771 | no |
| `batch_read_small` | 0.1958 | 0.1727 | +0.0231 | 0.1599 | no |
| `batch_write_small` | 0.7181 | 0.8058 | -0.0877 | 0.0822 | no |
| `batch_read_large` | 0.4996 | 0.5343 | -0.0347 | 0.0603 | no |
| `batch_write_large` | 1.2371 | 1.4219 | -0.1848 | 0.0017 | **yes** |
| `read_fanout_large` | 0.3329 | 0.3788 | -0.0459 | 0.0001 | **yes** |

###### pool_size=96, full run (n=50 pairs/cell)

| Cell | mean(on), s | mean(off), s | mean(diff=on-off), s | p-value | Significant? |
|---|---|---|---|---|---|
| `read_fanout` | 0.1291 | 0.1256 | +0.0035 | 0.6456 | no |
| `batch_read_small` | 0.3598 | 0.3661 | -0.0063 | 0.6219 | no |
| `batch_write_small` | 1.8413 | 1.7136 | +0.1276 | 0.0218 | **yes** |
| `batch_read_large` | 0.2922 | 0.3032 | -0.0109 | 0.3241 | no |
| `batch_write_large` | 1.2550 | 1.1835 | +0.0715 | 0.1190 | no |
| `read_fanout_large` | 0.3518 | 0.3357 | +0.0162 | 0.1967 | no |

### PURE and AWS S3 are unaffected

- **PURE**: resolves to a fixed set of 5 IPs. Distribution is already near-uniform before the fix (top-1/top-3 = 20.5%/60.9%). Performance measurements on PURE showed no difference with or without this fix.
- **AWS S3**: AWS's own DNS gives back a uniform distribution of DNS records

### DIFF
--- a/cpp/arcticdb/storage/test/test_s3_storage.cpp
+++ b/cpp/arcticdb/storage/test/test_s3_storage.cpp
@@ -515,7 +515,7 @@ TEST(TestS3Storage, curl_http_client_factory_is_registered) {
 
 TEST(TestS3Storage, dns_shuffle_addresses_enabled_by_default) {
     using namespace arcticdb::storage::s3;
-    ConfigsMap::instance()->unset_int("S3Storage.DnsShuffleAddresses");
+    ScopedConfig::unset_int("S3Storage.DnsShuffleAddresses");
     ASSERT_TRUE(dns_shuffle_addresses_enabled());
 }
 
--- a/cpp/arcticdb/storage/test/test_s3_storage.cpp
+++ b/cpp/arcticdb/storage/test/test_s3_storage.cpp
@@ -16,14 +16,19 @@
 #include <arcticdb/entity/variant_key.hpp>
 #include <arcticdb/storage/test/common.hpp>
 #include <arcticdb/util/test/gtest_utils.hpp>
+#include <arcticdb/util/configs_map.hpp>
 
 #include <arcticdb/log/log.hpp>
 
 #include <aws/core/Aws.h>
 #include <aws/core/utils/memory/stl/AWSStringStream.h>
+#ifndef WIN32
+#include <aws/core/http/HttpClientFactory.h>
+#endif
 
 #include <spdlog/sinks/ostream_sink.h>
 
+#include <set>
 #include <sstream>
 
 struct EnvFunctionShim : ::testing::Test {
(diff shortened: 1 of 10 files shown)


cpp/arcticdb/storage/test/test_s3_storage.cpp
```diff
@@ -499,3 +504,54 @@ TEST(S3LogSystem, RoutesToSpdlogWithLevelAndTag) {
...
+TEST(TestS3Storage, curl_http_client_factory_is_registered) {
+    arcticdb::storage::s3::S3ApiInstance::instance();
+    Aws::Client::ClientConfiguration client_configuration;
+    auto http_client = Aws::Http::CreateHttpClient(client_configuration);
+    ASSERT_NE(dynamic_cast<arcticdb::storage::s3::ArcticCurlHttpClient*>(http_client.get()), nullptr);
+}
+
+TEST(TestS3Storage, dns_shuffle_addresses_enabled_by_default) {
+    using namespace arcticdb::storage::s3;
+    ConfigsMap::instance()->unset_int("S3Storage.DnsShuffleAddresses");
```
alexowens90: `ScopedConfig`

alexowens90 (approved):

PR #3356, opened 2026-08-30
CI: cut the build from ~100 to ~46 minutes wall (draft: verifying the PR path)

G-D-Petrov:
**Draft — opened to verify the pull-request path of the matrix changes. Not for review yet.**

Cuts the build from ~100 min wall / ~4,900 job-min to ~46 min / ~2,080 job-min. Two of the changes are real bug
fixes; the rest are CI-only.

### Bug fixes

- **Windows console log sinks wrote into whatever file had recycled fd 1/2.** spdlog's stdout/stderr sinks cache the
  Win32 HANDLE at construction. `arcticdb_ext` links the CRT statically, so when another CRT in the process (Python's)
  `dup2`s over fd 2 — as pytest's capture does — our fd table still names the closed handle, whose value Windows had
  reused for LMDB's `data.mdb`. Log lines were written into the database's meta pages, which is the source of the
  `MDB_MAP_RESIZED` / `MDB_CORRUPTED` flakes seen on master. `write_to_console()` now resolves the handle through the
  shared CRT on every write. Regression test: `python/tests/unit/arcticdb/test_log_capture.py`.
- **`AzureStorage.HttpKeepAlive`** (default unchanged): the Azure C++ SDK blocks for seconds when it reuses a pooled
  connection the server has already closed. Regression test in `test_azure_transport.py`.

### CI

| change | effect |
|---|---|
| `MDB_NOSYNC\|MDB_NOMETASYNC` on Windows test jobs | Windows unit 40+ → 10-11 min |
| Test fixtures addressed by IPv4 literal, not `localhost` | Windows integration 45 → 16-18 min |
| Windows Defender exclusions in test jobs | −10% job-min |
| Stateful hypothesis test sharded 4 ways | hypothesis job 34-42 → 12-22 min |
| Stress suite parallelised on Linux (`-n 4 --dist loadgroup`) | 22.8 → 16-17 min |
| `enduser` folded into the hypothesis job, redundant matrix variants dropped | −60 jobs per run |
| `arcticdb_ext` links the core objects, skipping a ~1GB archive | −112-385s per Windows compile |
| Mongo retry timeouts scoped in the test | 906 → 180s |

Why `localhost` mattered: on Windows it resolves to `::1` first and the moto/GCP/azurite fixtures bind IPv4 only, so
every connection wasted ~2s failing over — measured at 205s per 100 connections versus 0.5s via `127.0.0.1`.

Stress runs on master, the nightly and manual dispatches but not on pull requests: over 176 master runs since
2026-06-01 it failed once in 2,465 jobs, and that one was "the runner has received a shutdown signal", while it is
the second most expensive type in the run.

### Not yet cleaned up

Contains an added-then-reverted Defender experiment and superseded intermediate commits; will be tidied before review.

https://claude.ai/code/session_014NcbXiRyX2crd3w8bpV13s


G-D-Petrov:
Superseded by #3376, which carries this work as one commit per change on a single branch off master, with the measured CI effect of each commit in the description. Nothing here is dropped: the trees were diffed file by file and match.

PR #3358, opened 2026-08-30
Add an LMDBStorage.ExtraFlags knob and LMDB corruption diagnostics

G-D-Petrov:
First of a stack that takes the `build.yml` run from ~100 min wall / ~4,900 job-min to **40 min / 1,686 job-min**, verified green end to end. This PR adds the LMDB knob and the diagnostics that found the root cause; the CI value stays **0** here, and is turned on in the next PR, which fixes what made it unsafe.

### Why

Windows `unit` jobs burned 144–172 CPU-min against 36 on Linux, and every slow test was a small-write-heavy LMDB workload. The env was opened with only `MDB_NOTLS`, so every commit fsyncs, and `FlushFileBuffers` on NTFS costs about a millisecond.

### What is here

- **`LMDBStorage.ExtraFlags`** (`ConfigsMap`, env `ARCTICDB_LMDBStorage_ExtraFlags_int`), ORed into the flags passed to `mdb_env_open`. Opt-in, **default 0**, so nothing changes for users or for CI in this PR.
- **`LmdbEnvDiagnostics`** — on `MDB_MAP_RESIZED`/`PAGE_NOTFOUND`/`CORRUPTED`/`BAD_TXN`/`PANIC`/`INVALID`, the raised message and the storage log now carry `mdb_env_info`/`mdb_stat` plus both meta pages read straight from `data.mdb` via pread/ReadFile, bypassing the map. This is what distinguishes "map and file disagree" from "both garbage", and it is what found the root cause behind the long-standing Windows flakes.

### Tests

`cpp/arcticdb/storage/test/test_lmdb_flags.cpp` — `LmdbFlags.*` and `LmdbDiagnostics.*`, passing locally on a debug build.

### Measured (with the flag set, which happens in the next PR)

`3.9 Windows / unit` 51 → 17 min; per-test CPU sum 171 → 34.5 min against Linux's 36; `test_realistic` 684 s → 20 s against Linux's 25 s; all 20,174 tests passed. Linux is indifferent — fsync on ext4 is cheap.

### Reviewer notes

- No behaviour change without the env var, and no API change.
- `MDB_WRITEMAP` is documented as **not** usable on Windows runners: a writable mapping makes Windows back the whole `map_size` (2 GiB per library, and one fixture asks for 100 GB), which filled `C:` to 100% and lost pages silently.
- `LmdbStorage::diagnostics()` is public so tooling can call it.

---

| | PR | |
|---|---|---|

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -794,7 +794,7 @@ jobs:
           # Skipping the fsync per commit makes the Windows unit jobs ~3x faster, but neither variant is safe
           # there: MDB_NOSYNC|MDB_NOMETASYNC gave sporadic MDB_MAP_RESIZED, and MDB_WRITEMAP|MDB_NOSYNC makes
           # Windows allocate the whole map_size per library, filling the disk and silently losing pages.
-          # Kept at 0 until that is understood; details in docs/claude/plans/gpetrov/ci_lmdb_nosync/.
+          # Kept at 0 until that is understood; details in docs/claude/plans/gpetrov/ci_lmdb_extra_flags/.
           ARCTICDB_LMDBStorage_ExtraFlags_int: '0'
 
       - name: Collect crash dumps (Windows)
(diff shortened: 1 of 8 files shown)


G-D-Petrov:
Superseded by #3376. This change is commit 964c0a247 there, which also folds in the console-sink branch's CI enablement of the knob.

PR #3359, opened 2026-08-30
Windows: stop console log writes landing inside LMDB files (fixes long-standing CI flakes)

G-D-Petrov:
Stacked on #3358. **Fixes a real Windows bug**, and on the strength of that turns the LMDB knob from the previous PR on for Windows test jobs.

### The bug

spdlog's `stderr_sink_mt` — ArcticDB's default sink — caches `_get_osfhandle(_fileno(stderr))` at construction on Windows and writes with `WriteFile(handle_, ...)`. pytest capture `dup2`s a temp file over fd 2, the CRT closes the original `HANDLE`, Windows recycles that handle value for the next `CreateFileW` — which in these tests is LMDB's `data.mdb` — and **every later log line is written into the database** at its current file offset, i.e. over the meta pages.

Windows-only and, in practice, test-only. It explains the ~1-in-50 `MDB_CORRUPTED` / `MDB_PAGE_NOTFOUND` flakes seen on master for months (`MDB_CORRUPTED` on run 30895908105, `MDB_BAD_TXN` on 30527747377 and 30466219731), and why `MDB_NOSYNC` made them far more frequent: commits get ~20× faster, so more log lines land between syncs.

The evidence is direct — the diagnostics added in the previous PR dumped ASCII log text (`arcticdb | ... V1 ... deprecated`) sitting inside LMDB's meta pages, both in memory and on disk.

### The fix, in two steps

1. `ConsoleSink` writes through `fwrite(stderr)` + `fflush` rather than a cached handle. **Not sufficient alone** — 20 of 56 Windows jobs still failed, and the meta page still held `arcticdb | ... DB v7.0.0 release`.
2. `arcticdb_ext` is built `/MT` (`x64-windows-static-msvc`), so it has its **own CRT fd table**. pytest's `os.dup2` goes through Python's `ucrtbase.dll`: it closes the OS handle behind *that* fd 2, while our CRT's fd 2 still names the recycled handle. So `write_to_console()` resolves the fd through the shared CRT's `_get_osfhandle` when `ucrtbase.dll` is loaded, falling back to `fwrite` when it is not (a static-CRT executable, where nothing can `dup2` our fds anyway).

The `GetModuleHandle`/`GetProcAddress` lookup is cached in a function-local static, since it takes the loader lock and would serialise logging threads. The **handle** is still resolved per write — that is the entire point of the fix.

### Tests

- `TestLog.ConsoleSinkFollowsStderrRedirection` — `dup2`s a file over fd 2 after the sink exists, asserts the line lands there.
- `python/tests/unit/arcticdb/test_log_capture.py` — asserts a C++ log line reaches pytest's `capfd`. Fails on Windows before the fix, passes on Linux either way.

### Result

Run 33206552356: **0 of 54** Windows jobs with LMDB errors, against 20 of 56 before; the capfd test passed 108/108 on Windows. With that established, `ARCTICDB_LMDBStorage_ExtraFlags_int=327680` (`MDB_NOSYNC|MDB_NOMETASYNC`) is enabled for Windows test jobs, taking Windows `unit` from 40–107 min to **10–11**.

### Reviewer notes

- Non-Windows behaviour is unchanged; the shared-CRT path is `#ifdef`ed to Windows.
- ANSI colour sinks stay off on Windows only, as before.

---

| | PR | |
|---|---|---|

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -791,11 +791,10 @@ jobs:
           ARCTICDB_LMDBSTORAGE_WARNIFOPENED_STR: all
           ARCTICDB_PYTEST_INFER_STRING: ${{ matrix.infer_string_id == '-inferstr' && '1' || '0' }}
           # Extra flags for mdb_env_open (see LMDBStorage.ExtraFlags in docs/mkdocs/docs/runtime_config.md).
-          # Skipping the fsync per commit makes the Windows unit jobs ~3x faster, but neither variant is safe
-          # there: MDB_NOSYNC|MDB_NOMETASYNC gave sporadic MDB_MAP_RESIZED, and MDB_WRITEMAP|MDB_NOSYNC makes
-          # Windows allocate the whole map_size per library, filling the disk and silently losing pages.
-          # Kept at 0 until that is understood; details in docs/claude/plans/gpetrov/ci_lmdb_extra_flags/.
-          ARCTICDB_LMDBStorage_ExtraFlags_int: '0'
+          # MDB_NOSYNC | MDB_NOMETASYNC (0x50000): skip the fsync per LMDB commit; the Windows unit jobs are ~3x
+          # faster and durability is irrelevant on CI. Not MDB_WRITEMAP, which makes Windows allocate the whole
+          # map_size per library on disk. See docs/claude/plans/gpetrov/ci_win_console_sink/.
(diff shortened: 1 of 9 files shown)


G-D-Petrov:
Superseded by #3376. This change is commit 12441ca79 there (the two console-sink commits squashed into one), with the `MDB_NOSYNC` enablement moved into 964c0a247.

PR #3360, opened 2026-08-30
Cut three stalls out of the integration suite (IPv4 fixtures, mongo timeouts, Azure keep-alive)

G-D-Petrov:
Stacked on #3359. Three unrelated stalls found while chasing the integration jobs, which were 45–47 min and the critical path once Windows was fixed. **16–18 min after this.**

### 1. `localhost` resolves to `::1` first

On Windows the moto, GCP and azurite fixtures bind IPv4 only, so every connection through the name wastes ~2 s failing over: **100 connects take 205 s via `localhost` and 0.5 s via `127.0.0.1`.** The fixtures now use the literal, and the test CA covers it.

The moto change has to be in the same commit: moto keys one Flask app, and one set of buckets, per value returned by `get_backend_for_host`. Once the S3 and IAM endpoints stopped agreeing on a host string they got separate backends, and every bucket created through one was invisible to the other (`NoSuchBucket`). Splitting these two would put a commit on master where the S3 tests 404.

### 2. `test_mongo_retryable_network_error`, ~1,000 s in every run

The test kills mongod and runs six operations, each waiting the default 120 s `SelectionTimeoutMs` per backoff attempt; teardown then spent 72 s sending `shutdown` to a server that was already dead. The timeouts are now scoped to the test, and the shutdown is skipped when the process has already exited.

Not runnable locally (no mongod); verified on CI.

### 3. Six azurite tests, ~123 s or ~186 s each — but under 1 s alone

Debug timing of the `Submitting/Submitted DeleteBlob batch` pairs showed 967 batch deletes in a run: 966 instant, **exactly one taking 122.4 s**.

It is not azurite — 60 concurrent batch deletes driven from Python peak at 0.06 s. It is the Azure C++ SDK's curl pool reusing a connection azurite (node, 5 s `keepAliveTimeout`) has already closed. Deterministic repro: write, idle 6 s, write with prune → 5–6 s instead of 0.08 s.

So this adds **`AzureStorage.HttpKeepAlive`** (default `1` — unchanged for users) wired to `CurlTransportOptions::HttpKeepAlive`. With it off, the local xdist repro goes from "1 of 967 batches takes 122.4 s" to "1002 batches, slowest 0.0 s". CI test jobs set it to 0.

Worth reporting upstream to azure-sdk-for-cpp.

### Reviewer notes

- The only production-code change is the new Azure config knob, and its default preserves today's behaviour.
- Everything else is test fixtures.

---

| | PR | |
|---|---|---|

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -795,6 +795,9 @@ jobs:
           # faster and durability is irrelevant on CI. Not MDB_WRITEMAP, which makes Windows allocate the whole
           # map_size per library on disk. See docs/claude/plans/gpetrov/ci_win_console_sink/.
           ARCTICDB_LMDBStorage_ExtraFlags_int: ${{ matrix.os == 'windows' && '327680' || '0' }}
+          # The Azure C++ SDK pools connections; reusing one azurite has already closed (node closes idle
+          # connections after 5s) blocks the next request for seconds, or ~2 minutes when several are in flight.
+          ARCTICDB_AzureStorage_HttpKeepAlive_int: '0'
 
       - name: Collect crash dumps (Windows)
         if: matrix.os == 'windows' && failure()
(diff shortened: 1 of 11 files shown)


G-D-Petrov:
Superseded by #3376, commits f38d39115 (mongo timeouts), 2ca3cff40 (`AzureStorage.HttpKeepAlive`) and d080a2fee (IPv4 fixture hosts).

PR #3361, opened 2026-08-30
CI: Defender exclusions, hypothesis sharding and matrix trims

G-D-Petrov:
Stacked on #3360. Matrix and job-shape changes. **No production code.**

- **Defender exclusions on Windows test jobs.** Real-time scanning of the thousands of small LMDB and temp files the tests create is worth 5–10× per test. Note this helps *test* jobs only — the same exclusions on the compile job were measured and made no difference, so that experiment was dropped rather than shipped.
- **enduser tests move into the hypothesis job.** Both were short jobs each paying a full setup.
- **The stateful hypothesis test is sharded.** It was a single ~20 minute test in a job that otherwise finished in three, so it set the length of the hypothesis job on every platform. Parametrising over `HYPOTHESIS_STATEFUL_SHARDS` (default 4) gives xdist something to distribute. Fewer examples per shard raises the share of overruns, so `data_too_large` is suppressed next to the `filter_too_much` that was already there. **Group: 34–42 min → 12–22.**
- **Benchmark smoke run** shortened on pull requests, where the job only needs to prove the benchmarks still run.

### The one thing to actually decide

**Matrix trims.** `inferstr` and the version-cache variants differ only in Python-level behaviour, so this runs them on Linux only rather than on all three platforms. That is a **coverage decision**, not purely a speed one, and it should be agreed explicitly rather than waved through with the rest. If you would rather keep them everywhere, that commit can be dropped and the remainder of the stack is unaffected.

---

| | PR | |
|---|---|---|

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -388,12 +388,12 @@ jobs:
           - python_deps_ids: [""]
             infer_string_ids: [""]
             matrix_override: ${{fromJson(needs.common_config.outputs.windows_matrix)}}
+          # infer_string is a Python/format-layer flag with no OS-specific behaviour, so the -inferstr variants
+          # run on Linux only (they cost 20 jobs / ~286 job-min across Windows and macOS).
           - python3: 11
             python_deps_ids: [""]
-            infer_string_ids: ${{ needs.storage_type.outputs.storage == 'no' && fromJSON('["", "-inferstr"]') || fromJSON('[""]') }}
+            infer_string_ids: [""]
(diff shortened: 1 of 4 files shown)


G-D-Petrov:
Superseded by #3376, commits 2fc395fcf (Defender exclusions), d30aa6832 (enduser in the hypothesis job), 2484e8e2b (hypothesis sharding) and e3f8a31bf (matrix trims). The matrix-trim coverage decision is called out in that PR's description and can still be dropped on its own.

PR #3362, opened 2026-08-30
CI: run the stress suite in parallel on Linux

G-D-Petrov:
Stacked on #3361. **Linux stress: 22.8 → 16–17 min.**

The stress suite ran with `-n 0` everywhere because a few of its tests peak at several gigabytes and two at once would evict the runner. This adds `xdist_group(name="stress_heavy")` to exactly those tests and runs the suite `-n 4 --dist loadgroup` on Linux, so the heavy tests stay serialised against each other while everything else runs four-up.

**`--dist loadgroup` is required here.** The other jobs use `worksteal`, which ignores `xdist_group` entirely — picking it would schedule the multi-gigabyte tests concurrently, which is the exact thing the grouping exists to prevent. Worth keeping in mind if the xdist mode is ever unified across jobs.

Also:
- drops stress from the `compat311` matrix, where it duplicates the main run;
- excludes stress from the `Serial` admission limit, which it no longer needs now that it is not the serial job.

`test_stress_merge_update.py`'s test is renamed from `stress_merge_update` to `merge_update` so it is collected under the same convention as its neighbours.

---

| | PR | |
|---|---|---|

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -579,6 +579,10 @@ jobs:
           # difference (~18% slower, same 152 tests), and it is the longest job in the run.
           - {python_deps_id: -compat311, type: stress}
           - {infer_string_id: -inferstr, admission_limit: Serial}
+          # NumProcessingUnitsLive=1 collapses read concurrency, which costs the stress suite 15 min for no signal:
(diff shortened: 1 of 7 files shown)


G-D-Petrov:
Superseded by #3376, commits 095465722 and c95b0177f.

PR #3363, opened 2026-08-30
CI: run the stress suite on master and the nightly, not on pull requests

G-D-Petrov:
Stacked on #3362, and deliberately **last and alone** — it is the only change in the stack that reduces coverage, so it should live or die on its own evidence rather than ride in behind six performance PRs. Nothing else in the stack depends on it.

### The change

Run the stress suite on master pushes and the nightly, not on pull requests.

### The evidence

Over 90 days and 176 master runs of `build.yml`, job-level conclusions from the Actions API:

| job type | failed jobs |
|---|---:|
| unit | 63 |
| compile | 47 |
| integration | 46 |
| **stress** (2,465 jobs) | **1** |

And that single stress failure was a **runner shutdown**, not a test.

So the suite has not caught a regression on master in three months, while costing a job slot on every pull request. Master and the nightly still run it, so a regression is caught before a release — just not before a merge.

### The call for the reviewer

Whether "one failure in 2,465, and that one infrastructural" is enough to move a suite off the pull-request path. Reasonable people can say no; if so, drop this PR and the other seven still stand.

The numbers came from a reusable script (job-level failure rates by suite over an arbitrary window) which I kept local rather than adding to the repo — happy to contribute it if it would be useful to others.

---

| | PR | |
|---|---|---|

### DIFF
--- a/.github/workflows/build_steps.yml
+++ b/.github/workflows/build_steps.yml
@@ -509,13 +509,18 @@ jobs:
         # There are so few nonreg and enduser tests, run them in the hypothesis runner: a job of its own costs
         # ~2.5 min of runner setup for ~6s of tests, and one less job per matrix entry means less runner contention.
         run: |
+          # The stress suite has caught nothing in 90 days: 1 failure in 2465 jobs across 176 master runs, and that
+          # one was "the runner has received a shutdown signal". It is the second most expensive type in the run, so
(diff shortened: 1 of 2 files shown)


G-D-Petrov:
Superseded by #3376, commit d45047696. It is still the one change in that PR that reduces coverage, and the 90-day evidence is repeated in the description so it can be dropped on its own.

PR #3364, opened 2026-08-30
Link arcticdb_ext against the core objects instead of the static archive

G-D-Petrov:
Independent of the CI stack — branched straight off master, touches nothing else.

`arcticdb_ext` linked `arcticdb_core_static`, so the archive was built and then largely re-read at link time. It now consumes `$<TARGET_OBJECTS:arcticdb_core_object>` through an INTERFACE target that forwards the usage requirements off `arcticdb_core_static` via `$<TARGET_PROPERTY:...>`, keeping include directories, compile definitions and link libraries identical while skipping the archive step.

20 lines of CMake. `arcticdb_core_static` itself is unchanged and still used by the C++ test binaries.

---

| | PR | |
|---|---|---|

### DIFF
--- a/cpp/arcticdb/CMakeLists.txt
+++ b/cpp/arcticdb/CMakeLists.txt
@@ -921,6 +921,19 @@ target_include_directories(arcticdb_core_static PUBLIC
 
 target_link_libraries(arcticdb_core_static PUBLIC ${arcticdb_core_libraries})
 
+# The objects of arcticdb_core_static without the archive step, carrying the same usage requirements. The generator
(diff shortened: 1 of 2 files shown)


G-D-Petrov:
Superseded by #3376, commit ad17d3d1c. Note that commit now carries a comment explaining why the link libraries are forwarded as `${arcticdb_core_libraries}` and not as `$<TARGET_PROPERTY:arcticdb_core_static,INTERFACE_LINK_LIBRARIES>` — the property holds unevaluated generator expressions (zstd) and the property form fails to configure on every platform.

PR #3365, opened 2026-08-30
Fix the pytest-data publish job, and report the critical path in the run summary

G-D-Petrov:
Independent of the CI stack — branched straight off master. **Tooling only**, no test or production code.

### The publish job was slow *and* silently lossy

It took **21 minutes** and processed **29 of 113 artifacts**, so most test timings never reached the database — which defeats the point of keeping the history at all.

Two bugs: the artifact list was fetched without pagination, so only the first page existed as far as the job was concerned; and the downloads ran serially. Now paginated and pulled through a thread pool.

A third surfaced once the first two were fixed: the job publishes the run it is *part of*, and that run is not `completed` while it is still running, so the status check rejected it.

**21 min → 1 min, all 113 artifacts.** Verified locally by simulating the Actions environment (92 XML files in 21.7 s) and then on CI.

### Where the time actually goes

Wall time is not the sum of the work — with runners to spare it is the longest *chain* of jobs — and that was not visible from the run page. `build_tooling/ci_critical_path.py` appends to `$GITHUB_STEP_SUMMARY`: when the run became merge-ready, the last jobs to finish, a mermaid gantt of the tail, and "without X the run would end at Y".

It also runs standalone against any run id (`--run-id`), which is how the numbers throughout this stack were measured.

The job is `continue-on-error`, so a reporting failure can never fail a build.

---

| | PR | |
|---|---|---|

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -553,6 +553,22 @@ jobs:
       - run: echo Dummy job to simplify PR merge checks configuration
       # FUTURE: add some test stats/reporting
 
+  # Wall time is the longest chain of jobs, not the sum of the work, so this reports which chain it was. It runs
+  # even when the build failed, since a run that fails slowly is worth seeing too.
+  critical_path:
+    needs: [cpp-test-linux, cpp-test-windows, cpp-test-macos, build-python-wheels-linux, build-python-wheels-windows, build-python-wheels-macos]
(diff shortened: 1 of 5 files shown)


G-D-Petrov:
Superseded by #3376, commits 6c6749d98 (publish job) and 6eff3c2be (critical-path summary).

PR #3373, opened 2026-08-31
fix(python): pickling warning says "metadata" when data is being pickled

### LINKED ISSUE #2336: Pickling data produces a misleading warning
### Describe the bug

Pickling data produces a warning that we are pickling _metadata_ which is misleading for users.

### Steps/Code to Reproduce

``` python
from arcticdb import Arctic

ac = Arctic("lmdb://test_metadata")
lib = ac.get_library("test_metadata", create_if_missing=True)

lib._nvs.write("b", float)
```

### Expected Results

Should be something related to the fact that we are pickling data, instead of:
`W arcticdb | Pickling metadata - may not be readable by other clients`

Python: 3.11.11 (main, Dec  4 2024, 08:55:08) [GCC 13.2.0]
OS: Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.39
ArcticDB: 5.3.3

### Backend storage used

Any backend

### Additional Context

_No response_

WatchTree-19:
### Reference Issues/PRs
Closes #2336.

### What does this implement or fix?

`Pickler.write` hardcodes "metadata" in its message:

```python
@staticmethod
def write(obj):
    log.log(get_pickled_metadata_loglevel(), f"Pickling metadata - may not be readable by other clients")
```

but its only caller is `MsgPackNormalizer._custom_pack`, and that class is the fallback normalizer for **user data** as well as the serializer for metadata. So pickling data reports that metadata was pickled.

Reproduced on 6.23.0, passing no metadata argument at all:

```
A) write_pickle(DATA), no metadata passed : Pickling metadata - may not be readable by other clients
B) write(df) with unserializable metadata : Pickling metadata - may not be readable by other clients
```

### The change

`MsgPackNormalizer` gains a `pickled_subject`, defaulting to `"data"`. `_init_msgpack_metadata`, which builds the single module-level instance `normalize_metadata` uses, sets it to `"metadata"`. `Pickler.write` names it.

After:

```
A) write_pickle(DATA), no metadata passed : Pickling data - may not be readable by other clients
B) write(df) with unserializable metadata : Pickling metadata - may not be readable by other clients
```

The metadata wording is unchanged, so anyone grepping logs for it still finds the metadata case.

### Tests

Two added beside `test_warns_on_large_meta`, which already patches the same logger, so the pattern matches what is there:

```
                                             master   this branch
test_pickling_metadata_says_metadata          PASS       PASS
test_pickling_data_does_not_say_metadata      FAIL       PASS
```

The metadata one passes either way on purpose, it is there to pin that the existing wording did not regress.

### Note on verification

I do not have a working C++ toolchain, so I could not run the repo's suite directly. I verified by applying the identical four edits to the installed 6.23.0 wheel and running the two tests' logic against it, which is the before/after shown above. The change is pure Python and touches no bindings.

### Checklist
<details>
  <summary>Checklist for code changes...</summary>

- [ ] Is the PR labelled as enhancement or bug if it can be added to our release notes?
</details>

### DIFF
--- a/python/arcticdb/version_store/_normalization.py
+++ b/python/arcticdb/version_store/_normalization.py
@@ -1397,8 +1397,11 @@ class MsgPackNormalizer(Normalizer):
     Fall back plan for the time being to store arbitrary data
     """
 
-    def __init__(self, cfg=None):
+    def __init__(self, cfg=None, pickled_subject="data"):
         self.strict_mode = cfg.strict_mode if cfg is not None else False
+        # What this instance serialises. Only used so the pickling log message names the
+        # right thing: this class is the fallback for user data as well as for metadata.
+        self._pickled_subject = pickled_subject
 
     def normalize(self, obj, **kwargs):
         disallow_pickle = kwargs.get("disallow_pickle", None)
@@ -1451,7 +1454,7 @@ class MsgPackNormalizer(Normalizer):
         if disallow_pickle:
             raise TypeError("Normalizing data by pickling has been disabled.")
         else:
-            return ExtType(MsgPackSerialization.PY_PICKLE_3, packb(Pickler.write(obj)))
+            return ExtType(MsgPackSerialization.PY_PICKLE_3, packb(Pickler.write(obj, self._pickled_subject)))
 
     def _ext_hook(self, code, data):
         if code == MsgPackSerialization.PD_TIMESTAMP:
@@ -1506,8 +1509,8 @@ class Pickler(object):
         return pd.read_pickle(io.BytesIO(data))
 
     @staticmethod
-    def write(obj):
-        log.log(get_pickled_metadata_loglevel(), f"Pickling metadata - may not be readable by other clients")
+    def write(obj, subject="data"):
+        log.log(get_pickled_metadata_loglevel(), f"Pickling {subject} - may not be readable by other clients")
         return pickle.dumps(obj, protocol=PICKLE_PROTOCOL)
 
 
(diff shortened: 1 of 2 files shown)


WatchTree-19:
Could a maintainer approve the workflow runs on this and the three sibling PRs when convenient?

All four are sitting at `action_required` — GitHub's first-time fork-contributor gate — so no job has executed and there's no CI signal either way:

- #3369 — `will_item_be_pickled` raises `UnboundLocalError` instead of returning `True`
- #3370 — honour `fail_on_missing_custom_normalizer` when unpickling the normalizer
- #3371 — `is_server_type` raises `UnboundLocalError` when the server is unreachable
- #3373 — pickling warning says "metadata" when data is being pickled

Each is a small independent fix in the Python layer with a test, and they don't touch each other, so they can be taken in any order or individually.

No rush, and happy to squash, split or rebase any of them if that makes review easier. Also let me know if `Check Pull Request Labels` expects a label I should be adding myself — I couldn't tell from the workflow whether that's a contributor or maintainer step.

WatchTree-19:
Gentle follow-up on this — no rush if now is not the time.

These four are still at zero check-runs, so CI has never executed on any of them and there is no signal either way. It is the first-time fork-contributor gate rather than anything about the changes.

@poodlewars @vasil-pashov — you have both been merging this week, so tagging you rather than the room. One "Approve and run workflows" on each would unblock them:

- #3369 — `will_item_be_pickled` raises `UnboundLocalError` instead of returning `True`
- #3370 — `fail_on_missing_custom_normalizer` not honoured when unpickling the normalizer
- #3371 — `is_server_type` raises `UnboundLocalError` when the server is unreachable
- #3373 — this one, pickling warning says "metadata" when it is the data being pickled

Each is small and independent, so they can be taken or dropped individually. If any of them is not worth the review time, say so and I will close it rather than leave it sitting in your queue.

PR #3375, opened 2026-09-01
Import all exceptions from arcticdb.exceptions in tests

markovskipetar:
## What was the exceptions import logic:

  ArcticDB exceptions can be imported from many different places — `arcticdb_ext.exceptions`,
  `arcticdb_ext.storage`, `arcticdb_ext.version_store`, `arcticdb.version_store.library`, etc.
  This makes exceptions hard to discover and easy to import inconsistently.

  This PR adds a test that enforces a single allowed source for them: every ArcticDB exception
  used anywhere under `python/` must be imported from `arcticdb.exceptions`.

  ## What is the new logic:

The test reads every `.py` file under `python/` (skipping `__pycache__` and the test file
  itself) and checks how each file imports ArcticDB exceptions. It catches both ways an exception
  can be pulled from the wrong place:

  - an import, e.g. `from arcticdb_ext.exceptions import UserInputException`
  - a direct inline use, e.g. `raise arcticdb_ext.exceptions.UserInputException(...)`

  In both cases, if the exception came from anywhere other than `arcticdb.exceptions`, it's a
  violation.

  To avoid false alarms, the test confirms the name really is an ArcticDB exception (a subclass
  of `ArcticException`) before flagging it.

### DIFF
--- a/python/tests/unit/arcticdb/test_exception_imports.py
+++ b/python/tests/unit/arcticdb/test_exception_imports.py
@@ -1,0 +1,108 @@
+"""
+Copyright 2026 Man Group Operations Limited
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+import ast
+import os
+
+import arcticdb.exceptions as arcticdb_exceptions
+
+# The one canonical place tests are allowed to import exceptions from.
+CANONICAL_MODULE = "arcticdb.exceptions"
+
+# Native submodules the exceptions are also (still) reachable from. Used to catch
+# fully-qualified attribute access such as ``arcticdb_ext.exceptions.UserInputException``.
+NATIVE_EXCEPTION_MODULES = {
+    "arcticdb_ext.exceptions",
+    "arcticdb_ext.storage",
+    "arcticdb_ext.version_store",
+}
+
+
+def canonical_exception_names():
+    """Every exception re-exported from ``arcticdb.exceptions``.
+
+    A name qualifies if it resolves to a subclass of ``BaseException``. This is name-agnostic
+    (so it catches exceptions like ``StreamDescriptorMismatch`` or ``UnknownLibraryOption`` that
+    do not end in ``Exception``/``Error``) and excludes error-metadata enums such as
+    ``ErrorCode``/``ErrorCategory`` and non-exception data containers such as ``DataError``.
+    Names that are *not* re-exported from ``arcticdb.exceptions`` are out of scope and never
+    flagged.
+    """
+    names = set()
+    for name in dir(arcticdb_exceptions):
+        obj = getattr(arcticdb_exceptions, name)
+        if isinstance(obj, type) and issubclass(obj, BaseException):
+            names.add(name)
+    return names
+
+
+def dotted_name(node):
+    """Return the dotted source of an attribute chain, e.g. ``a.b.C`` -> "a.b.C"."""
+    parts = []
+    while isinstance(node, ast.Attribute):
+        parts.append(node.attr)
+        node = node.value
+    if isinstance(node, ast.Name):
+        parts.append(node.id)
+    return ".".join(reversed(parts))
+
+
+def find_tests_root():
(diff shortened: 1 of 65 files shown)


python/tests/unit/arcticdb/test_exception_imports.py
```diff
@@ -0,0 +1,108 @@
...
+    parts = []
+    while isinstance(node, ast.Attribute):
+        parts.append(node.attr)
+        node = node.value
+    if isinstance(node, ast.Name):
+        parts.append(node.id)
+    return ".".join(reversed(parts))
+
+
+def find_tests_root():
```
IvoDD, line 68: Nice to have such automated assertion!

I'd go one step further and assert that all python imports across all files (apart from exceptions.py), not only tests, should import exceptions only from `arcticdb.exceptions`

python/tests/unit/arcticdb/test_exception_imports.py
```diff
@@ -0,0 +1,108 @@
...
+    this_file = os.path.abspath(__file__)
+
+    violations = []
+    for path in iter_test_python_files(tests_root):
+        if os.path.abspath(path) == this_file:
+            continue
+        with open(path, encoding="utf-8") as file:
+            tree = ast.parse(file.read(), filename=path)
+        relative_path = os.path.relpath(path, tests_root)
+        for node in ast.walk(tree):
```
IvoDD: How long does it take to walk over all ast trees?

python/tests/unit/arcticdb/test_exception_imports.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_exceptions_only_imported_from_arcticdb_exceptions():
+    """Every exception used in the test suite must be imported from ``arcticdb.exceptions``.
+
+    We don't stop the exceptions being importable from their original locations (that would
+    break backwards compatibility), but the tests themselves should use the single canonical
+    module so there is one obvious place to find them.
+    """
+    exception_names = canonical_exception_names()
```
IvoDD: We're only searching for exceptions in `CANONICAL_MODULE`.

This means that if we add a new exception e.g. `arcticdb_ext.exceptions.NewException` and do an import somewhere to `from arcticdb_ext.exceptions import NewException` this code would not catch it.

What do you think about also checking all exception names exported from `NATIVE_EXCEPTION_MODULES`?

python/tests/unit/arcticdb/test_exception_imports.py
vasil-pashov: +1

vasil-pashov (commented):
I think there's a cleaner way. All arcticdb exceptions inherit from `ArcticException`. We can check the imports of each python file and check if there's something inheriting from `ArcticException` if so check where it's imported from.

Since there are multiple files I suggest using multi-threading.

python/tests/unit/arcticdb/test_exception_imports.py
```diff
@@ -0,0 +1,112 @@
...
+# The one canonical place tests are allowed to import ArcticDB exceptions from.
+CANONICAL_MODULE = "arcticdb.exceptions"
+
+
+@functools.lru_cache(maxsize=None)
+def _load_module(module):
+    try:
+        return importlib.import_module(module)
+    except Exception:
+        return None
```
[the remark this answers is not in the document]
IvoDD: +1 to claude comment

python/tests/unit/arcticdb/test_exception_imports.py
```diff
@@ -0,0 +1,112 @@
...
+def find_exception_import_violations(source, filename="<source>"):
+    tree = ast.parse(source, filename=filename)
+    violations = []
+    for node in ast.walk(tree):
+        if isinstance(node, ast.ImportFrom):
+            if node.module is None or node.module == CANONICAL_MODULE:
+                continue
+            for alias in node.names:
+                if _is_arctic_exception(node.module, alias.name):
+                    violations.append((node.lineno, alias.name, node.module))
```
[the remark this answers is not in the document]
IvoDD: I think this claude comment has a point, but it will overcomplicate this. Let's not do this for now but worth adding at least a comment explaining the current code is not perfect with respect to `*` imports.

IvoDD (approved):

PR #3376, opened 2026-09-01
CI: cut build.yml from ~100 to 40 minutes, one commit per change

G-D-Petrov:
Takes `build.yml` from **~100-150 min wall to 39-59 min**, verified green end to end (run 33731135306, 154 jobs,
all passing).

After all the changes, Wall time is set by how long the macOS test jobs wait for hosted runners, not by compute, so a single headline
figure would be misleading. Three measurements of this pipeline:

| run | wall | critical path | queueing |
|---|---|---|---|
| 33648621710, 15:27 UTC | 74 min | macOS `integration`, queued **50 min**, ran 22 min | ~45 min |
| 33648621710 re-run, 06:02 UTC | **39 min** | Linux `integration`, queued 9 min, ran 27 min | ~12 min |
| 33731135306, 07:2x UTC | 59 min | macOS `integration`, queued **36 min**, ran 21 min | ~37 min |

The compute is stable across all three; the queueing is not. When macOS runners are available the pipeline is
compute-bound at 39 min with Linux `integration` as the ceiling. Master runs over the same period took 109-210 min.

Reducing the macOS job count is the next lever and hopefully, by merging this and reducing the total MacOS jobs in flight, the wall time might come down even more.

Two of the commits are **bug fixes, not CI configuration**: the Windows console sink writing log lines into
`data.mdb`, and the Azure C++ SDK stalling on connections the server has already closed. Everything else is CI
configuration or test fixtures. Those were kept here because they also improve the test times.

## What each commit costs or saves

Measured on `man-group/ArcticDB` runs during August 2026. "job-min" is the sum of every job's duration; wall time is the preamble plus the slowest Windows compile plus the longest test job, so only the tail job moves it.

| # | Commit | Effect on CI |
|---|--------|--------------|
| 1 | aab3df181 LMDB `ExtraFlags` knob, corruption diagnostics, `MDB_NOSYNC` on Windows CI | Windows `unit` **40–107 → 13–15 min**; per-test CPU **387 → ~35 min**, i.e. Linux parity. The diagnostics are what found the bug in commit 2 |
| 2 | 0a6f63f95 stop console log writes landing inside LMDB files | **20 of 56 → 0 of 54** Windows jobs with LMDB errors (run 33206552356); this is what makes commit 1 safe |
| 3 | 3f80585a3 mongo selection timeout scoped to the test | `test_mongo_retryable_network_error` **1,007 s (+72 s teardown) → 201 s** Linux, 107 s Windows |
| 4 | 898510c0e `AzureStorage.HttpKeepAlive`, off in CI | six azurite tests **123–186 s each → <1 s**; with 3, Linux `integration` **38–39 → 26–27 min** |
| 5 | e0ec3b6b0 fixtures addressed by IPv4 literal | 100 connects on Windows **205 s → 0.5 s**; Windows `integration` **45–46 → 16–18 min**, Windows `unit` 13–15 → **10–11** |
| 6 | 61bfbee16 Defender exclusions on Windows test jobs | **−10% job-min** over the run; cp313 Windows `unit` CPU 387 → 172 min. (Measured as *no* benefit on compile jobs, so it is not applied there) |
| 7 | 9f134fe99 enduser tests folded into the hypothesis job | **−27 jobs per run** |
| 8 | 8bf5d4ee2 stateful hypothesis test sharded 4 ways | hypothesis group **34–42 → 12–22 min** — it was one ~20 min test setting the job length |
| 9 | cc8268a74 matrix trims, shorter benchmark smoke run | with 7, **207 → 173 jobs per run**. **A coverage decision** — see below |
| 10 | 81502b70c stress `-n 4 --dist loadgroup` on Linux | Linux `stress` **22.8 → 16–17 min** |
| 11 | ba823a977 stress out of the `Serial` admission limit | scheduling only; it is no longer the serial job |
| 12 | 353ad3a35 stress on master and nightly, not on PRs | superseded in part by 16: PRs now keep Linux stress. **A coverage decision** — see below |
| 13 | a2133aa1c paginate and parallelise the artifact download | publish job **21 min → 5.7 s** of downloads, and **29 → 92** XMLs published (run 33648621710). De-duplicates the listing: paging over a run that is still uploading returns a boundary artifact twice, and two threads then raced on one directory |
| 14 | 26d833913 critical-path summary in the run summary | reporting only; this is how the numbers above were measured |
| 15 | 6d640c22f `printf` instead of `echo -e` for the Windows env | the workspace path `D:\a\...` had its `\a` turned into a BEL and the trailing newline dropped, so `VCPKG_ROOT` was corrupt and `CMAKE_BUILD_PARALLEL_LEVEL` **never got set** - every `cmake --build -j $CMAKE_BUILD_PARALLEL_LEVEL` ran as a bare `-j`, unbounded on a 4-core runner. Fixing it re-enabled `run-cmake`'s `vcpkg env` MSVC setup, which leaks a truncated `X_VCPKG_RECURSIVE_DATA` and breaks the configure, so the same commit defines `CC`/`CXX` - the action's documented way to skip that block, and the exact values it sets itself |
| 16 | 8b74bf1fb PRs keep the stress suite on Linux only | softens 12: pull requests keep the 4 Linux `stress` jobs (~46 runner-min, cheapest lane, races are not OS-specific) and skip the Windows/macOS ones (~42 runner-min + per-job setup); master, the nightly and dispatches still run stress on every OS |

Runner queueing is small on Linux and Windows, so cutting job count there saves cost rather than wall time - only
the tail job shortens the wall. macOS is the exception and is now the binding constraint under contention: its test
jobs waited 50 min for runners in the 15:27 run while their wheels had been ready since minute 16. 

## Reviewer notes

- The two new config knobs, `LMDBStorage.ExtraFlags` and `AzureStorage.HttpKeepAlive`, both default to today's
  behaviour. No API change, no on-disk format change.
- `MDB_WRITEMAP` is documented as **not** usable on the CI runners: a writable mapping makes Windows back the whole
  `map_size` on disk, which filled `C:` and lost pages silently.


cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
```diff
@@ -405,6 +420,145 @@ T or_else(T val, T or_else_val, T def = T()) {
...
+        throw std::runtime_error(fmt::format("pread failed with errno {}", errno));
+    }
+    buf.resize(static_cast<size_t>(got));
+#endif
+    return buf;
+}
+
+} // namespace
+
+LmdbEnvDiagnostics lmdb_env_diagnostics(::lmdb::env& env) {
```
alexowens90: If you've scoured the LMDB docs/source code and verified this is all correct, then kudos.
If not and this is pure-vibes, then I'm fine with checking it in, but it absolutely has to be disabled by default and only enabled by config variable (which we can enable in the CI).

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
alexowens90, line 1: Can we move all of the new stuff to a self contained translation unit to keep the actual LMDB impl easier to read? `lmdb_error_handling.?pp` or something

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
```diff
@@ -405,6 +420,145 @@ T or_else(T val, T or_else_val, T def = T()) {
...
+                out.file_read_error = fmt::format("short read: {} bytes", bytes.size());
+            }
+        }
+    } catch (const std::exception& ex) {
+        out.file_read_error = ex.what();
+    }
+    return out;
+}
+
+std::string LmdbEnvDiagnostics::to_string() const {
```
alexowens90: Any reason not to implement a fmt formatter for this class?

python/tests/unit/arcticdb/test_log_capture.py
```diff
@@ -0,0 +1,20 @@
...
+
+Use of this software is governed by the Business Source License 1.1 included in the file licenses/BSL.txt.
+
+As of the Change Date specified in that file, in accordance with the Business Source License, use of this software will be governed by the Apache License, version 2.0.
+"""
+
+from arcticdb_ext.log import log, LogLevel, LoggerId
+
+
+def test_cpp_log_lines_follow_fd_capture(capfd):
```
alexowens90: We've had issues with `capfd` tests being flaky before, although it's possible this change will fix that. I propose we leave this in, and if it remains non-flaky then we can unskip the other `capfd` tests.

python/tests/unit/arcticdb/test_log_capture.py
```diff
@@ -0,0 +1,20 @@
...
+    # pytest's capfd dup2()s a temp file over fd 2 through Python's CRT. On Windows arcticdb_ext has its own static
+    # CRT, so its fd 2 still pointed at the original stderr handle, which Python's CRT had closed and Windows had reused
+    # for the next opened file (LMDB's data.mdb) - log lines were written into the database. The log line must land
+    # in the capture.
+    marker = "arcticdb-log-capture-marker-7e1c"
+    log(LoggerId.ROOT, LogLevel.ERROR, marker)
```
alexowens90, line 18: Flush?

cpp/arcticdb/log/log.cpp
```diff
@@ -32,11 +33,24 @@ std::shared_ptr<Loggers> loggers_instance_;
 std::once_flag loggers_init_flag_;
 } // namespace
 
+std::shared_ptr<spdlog::sinks::sink> make_console_sink(bool std_err, bool color) {
```
alexowens90, line 39: This change has a potentially infinite blast radius. How have you verified the correctness of the implementation?
IIUC the problem is Windows specific? In which case why has the non-Windows impl also changed?

cpp/arcticdb/storage/azure/azure_client_impl.cpp
```diff
@@ -111,6 +112,9 @@ Azure::Storage::Blobs::BlobClientOptions RealAzureClient::get_client_options(con
     // On Linux, use libcurl with CA cert configuration
     ARCTICDB_RUNTIME_DEBUG(log::storage(), "Using libcurl transport");
     Azure::Core::Http::CurlTransportOptions curl_transport_options;
+    // Reusing a pooled connection that the server has already closed makes the next request hang for ~2 minutes
+    // before it is retried. Azurite (node) closes idle connections after 5s, so tests hit this regularly.
```
alexowens90, line 116: Would changing our Azurite configuration have the same impact without modifying production code?

.github/workflows/build_steps.yml
```diff
@@ -795,6 +795,9 @@ jobs:
           # faster and durability is irrelevant on CI. Not MDB_WRITEMAP, which makes Windows allocate the whole
           # map_size per library on disk.
           ARCTICDB_LMDBStorage_ExtraFlags_int: ${{ matrix.os == 'windows' && '327680' || '0' }}
+          # The Azure C++ SDK pools connections; reusing one azurite has already closed (node closes idle
+          # connections after 5s) blocks the next request for seconds, or ~2 minutes when several are in flight.
+          ARCTICDB_AzureStorage_HttpKeepAlive_int: '0'
```
alexowens90: It would be better if this was patched into our test files rather than here, so it also speeds up local development without needing to set this env var

.github/workflows/build_steps.yml
```diff
@@ -651,6 +651,17 @@ jobs:
...
+      # Real-time scanning of the thousands of small LMDB/temp files the tests create makes
+      # Windows test jobs 5-10x slower per test than Linux.
+      - name: Exclude test paths from Windows Defender
+        if: matrix.os == 'windows'
+        shell: pwsh
+        run: |
+          Add-MpPreference -ExclusionPath "$env:GITHUB_WORKSPACE", "$env:RUNNER_TEMP", "$env:TEMP", "$env:LOCALAPPDATA\Temp", "C:\hostedtoolcache", "C:\tmp", "D:\tmp"
+          Add-MpPreference -ExclusionProcess "python.exe", "pytest.exe", "node.exe", "azurite.exe"
+          Set-MpPreference -DisableRealtimeMonitoring $true
+          Get-MpPreference | Select-Object DisableRealtimeMonitoring, ExclusionPath, ExclusionProcess
```
alexowens90, line 699: If possible, we could just disable Defender entirely, it isn't useful to us

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
```diff
@@ -281,17 +281,31 @@ def test_list_versions_latest_only_snapshot(self):
...
+# this was ~20 minutes and set the wall time of the whole hypothesis job, while other workers sat idle. Examples are
+# generated independently, so N shards of M examples explore as much as one run of N*M.
+HYPOTHESIS_STATEFUL_SHARDS = int(os.getenv("HYPOTHESIS_STATEFUL_SHARDS", 4))
+
+
 @pytest.mark.parametrize("lib_type", ["lmdb_version_store_delayed_deletes_v1", "lmdb_version_store_delayed_deletes_v2"])
+@pytest.mark.parametrize("shard", range(HYPOTHESIS_STATEFUL_SHARDS))
 @SLOW_TESTS_MARK
-def test_stateful(lib_type, request):
+def test_stateful(lib_type, shard, request):
```
alexowens90, line 293: The argument against this change is that test is fundamentally changed by the sharding process.
The argument for it is that this test is hot garbage that has never exposed a bug not also found by another test.

.github/workflows/build_steps.yml
```diff
@@ -340,9 +341,11 @@ jobs:
...
       - name: Run C++ Benchmarks
         if: inputs.job_type == 'cpp-tests' && env.ARCTICDB_USE_SANITIZER == ''
-        # We set a benchmark_min_time=20x because running until a time limit is reached can take a lot of time on
-        # benchmarks which pause the timer.
-        run: cpp/out/${ARCTIC_CMAKE_PRESET}-build/arcticdb/benchmarks --benchmark_time_unit=ms --benchmark_min_time=20x
+        # Nothing consumes the timings, this only checks the benchmarks still run, so one iteration is enough on
+        # pull requests. 20 iterations took ~5.5 min of the C++ test job; the full count still runs on master and
+        # on the nightly schedule. (Running until a time limit is reached is much slower for benchmarks that pause
+        # the timer, hence a fixed iteration count rather than a duration.)
+        run: cpp/out/${ARCTIC_CMAKE_PRESET}-build/arcticdb/benchmarks --benchmark_time_unit=ms --benchmark_min_time=${{ github.event_name == 'pull_request' && '1x' || '20x' }}
```
alexowens90: I'd go `x2` for the PR runs. It's easy to write a benchmark that succeeds exactly once and fails/crashes on the second pass because of accidental data reuse.

.github/workflows/build_steps.yml
alexowens90: Hard disagree with this one. I would rather tune down the parameters the stress tests use to speed them up than have to add a nightly-check.

.github/workflows/publish_pytest_data.yml
alexowens90, line 1: 21 minutes is insane, why does it take so long?

.github/workflows/build.yml
```diff
@@ -556,6 +556,22 @@ jobs:
       - run: echo Dummy job to simplify PR merge checks configuration
       # FUTURE: add some test stats/reporting
 
+  # Wall time is the longest chain of jobs, not the sum of the work, so this reports which chain it was. It runs
+  # even when the build failed, since a run that fails slowly is worth seeing too.
+  critical_path:
```
alexowens90, line 561: Cool! Where does this get published? Can't see it in the CI for this PR

alexowens90:
I'm a bit on the fence about the stress tests. They were written, and therefore time-tuned with only linux in mind. Instead of not running them at all on Windows/Mac on PRs, we could make the parameters they use OS-dependent.

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
G-D-Petrov: Verified against LMDB 0.9.x mdb.c: the offsets are the field offsets of MDB_meta inside a meta page.

Gated it anyway in 4f6ea06c4 as it is decoding a struct layout LMDB does not promise to keep. LMDBStorage.Diagnostics, default 0, and the test jobs set ARCTICDB_LMDBStorage_Diagnostics_int=1.

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
G-D-Petrov: Done in 13c864789 
I left lmdb_extra_env_flags() where it was, as it belongs next to the `mdb_env_open` call it feeds

cpp/arcticdb/storage/lmdb/lmdb_storage.cpp
G-D-Petrov: None, just missed it. af57474b9 - formatters for LmdbEnvDiagnostics and for the nested Meta, which was a lambda.

python/tests/unit/arcticdb/test_log_capture.py
G-D-Petrov: Agreed. One thing though - I went looking for the other skipped capfd tests and could not find any: the only capfd
users in the tree now are this test and the get_stderr fixture in conftest.py. If you remember which ones, I'll dig
them out of the history.

The fd-capture failure mode this test covers is the same one behind the MDB_CORRUPTED/MDB_PAGE_NOTFOUND flakes
(run 33206552356: 0 of 54 Windows jobs with LMDB errors vs 20 of 56 before), so if capfd was flaky for that reason
it should be fixed now; if it was flaky for another, this test will tell us.

python/tests/unit/arcticdb/test_log_capture.py
G-D-Petrov: 428ae7450. Switched it to the get_stderr fixture in conftest.py, which does flush_all() and then reads capfd - same thing test_lmdb.py and test_append.py already use.

cpp/arcticdb/log/log.cpp
G-D-Petrov: You're right on the second half - fixed in a0aa20495, the replacement sink is now Windows-only.

On verification: the Windows sink is spdlog's own stdout_sink_base with one line changed (handle looked up per write instead of cached), and 50912d059 restores spdlog's process-wide ConsoleMutex, which my base_sink version had narrowed to a per-sink lock.

cpp/arcticdb/storage/azure/azure_client_impl.cpp
G-D-Petrov: Partly - Azurite does have `--blobKeepAliveTimeout (default 5s)`, so raising it would make the suite faster without the C++ change. I'd still keep the knob: the stall is not only an Azurite artifact (anything with an idle timeout in front of the SDK's pooled connections does this - Azure's own load balancer drops them at 4 minutes - and a user has no way to turn reuse off today), and the default is unchanged, so nothing differs for anyone who doesn't set it.

Raising Azurite's timeout would also stop the tests from exercising the reconnect path.

.github/workflows/build_steps.yml
G-D-Petrov: Agreed - fixed in bf75789f1.

.github/workflows/build_steps.yml
G-D-Petrov: Just disabling Defender doesn't work - i.e., there are some actions and commands that say they do it, but they don't produce the same effect.
I assume that the runners have some protection against something like that.
The steps here were the only ones that had any substantial effect.

python/tests/hypothesis/arcticdb/test_hypothesis_version_store.py
G-D-Petrov: The difference is the we won't be able to catch  a bug that needs a database hammered for 20 minutes rather than 5.
I went over the pytest date from the last couple of months and this test hasn't caught any actual failures (only timeouts or lmdb corruptions due to the problem that this PR solved).
So I think that it is find to shard it.

.github/workflows/build_steps.yml
G-D-Petrov: Good point - fixed in 2a5f4ff89, 2x on pull requests.

.github/workflows/publish_pytest_data.yml
G-D-Petrov: There were 2 problems that overlapped:
- we were downloading the artifacts sequentially
- sometimes GH throttles these downloads

So the 21 min is a bit of an anomaly, but at least now we are downloading them in parallel.

.github/workflows/build.yml
G-D-Petrov: The report writes to $GITHUB_STEP_SUMMARY, so it renders like this on the summary page:
image

G-D-Petrov:
Agreed that is the better end state. In the meantime the PR is already softer than the commit you're replying to -
8b74bf1fb keeps Linux stress on PRs and drops only Windows/macOS, which still run on master, the nightly and
workflow_dispatch, so an OS-specific regression is caught on merge rather than never.

OS-dependent parameters mean retuning row and iteration counts for five modules against three OSes, so I'd rather
raise it as a follow-up issue, if we want to go down this road.

cpp/arcticdb/log/test/test_log.cpp
```diff
@@ -45,3 +45,161 @@ TEST(TestLog, TestFormatBytes) {
...
+    const HANDLE capture_handle = create_file(capture_path);
+    ASSERT_NE(capture_handle, INVALID_HANDLE_VALUE);
+    const int capture_fd = io.open_osfhandle(reinterpret_cast<intptr_t>(capture_handle), 0);
+    ASSERT_GE(capture_fd, 0);
+    ASSERT_EQ(io.dup2(capture_fd, 2), 0);
+    const HANDLE decoy_handle = create_file(decoy_path);
+    ASSERT_NE(decoy_handle, INVALID_HANDLE_VALUE);
+    const bool decoy_took_stale_handle = io.shared && decoy_handle == original_handle;
+    RecordProperty("decoy_took_stale_stderr_handle", decoy_took_stale_handle ? "yes" : "no");
+    ASSERT_EQ(io.close(capture_fd), 0);
```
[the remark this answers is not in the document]
G-D-Petrov: Right, and an assertion failure leaving the process in exactly the state the PR is about is not an acceptable failure mode for this test. The redirect now lives in its own scope with a `RestoreStderr` guard: its destructor puts fd 2 back (through ucrtbase and through the static CRT on Windows, `dup2` on POSIX), closes the saved fds and the decoy, and reports a failed restore with `EXPECT_EQ`. The checks after the `dup2` are `EXPECT_*`s; the ones before it stay `ASSERT_*`s, since nothing has been redirected yet and each call needs the previous one's result (a `dup2` of a -1 fd through ucrtbase is itself an invalid-parameter fast-fail). The scope closes before the capture and decoy files are read and removed, so the restore still precedes the read as before.

Two details of the guard that are not in your comment but follow from it:

- It restores only when the `dup2` actually ran. Putting the static CRT's fd 2 back after a failure before the redirect would `CloseHandle` the handle the shared table still holds for fd 2, which is the stale-handle state this test exists to catch, created by the test itself.
- The static CRT's restore now runs before ucrtbase's, not after. Its `_dup2` `CloseHandle`s the stale value first, and with ucrtbase restored first that value could by then have been the handle ucrtbase's `_dup2` had just been given. In the old order that was a matter of which free slot the kernel handed out; in this order the value is either the decoy's or nobody's.

Folded into the console sink commit, which is where this block came from; its message's last paragraph now describes the guard.

cpp/arcticdb/log/console_sink.cpp
```diff
@@ -0,0 +1,130 @@
...
+    for (const wchar_t* name : {L"ucrtbase.dll", L"ucrtbased.dll"}) {
+        HMODULE module = nullptr;
+        if (!::GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_PIN, name, &module) || module == nullptr)
+            continue;
+        CrtFdTable crt;
+        crt.get_osfhandle = reinterpret_cast<GetOsfHandleFn>(::GetProcAddress(module, "_get_osfhandle"));
+        crt.set_thread_local_invalid_parameter_handler = reinterpret_cast<SetInvalidParameterHandlerFn>(
+                ::GetProcAddress(module, "_set_thread_local_invalid_parameter_handler")
+        );
+        if (crt.get_osfhandle != nullptr)
```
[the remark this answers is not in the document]
G-D-Petrov: Agreed, that was a gap. `shared_crt()` now requires both entry points before it returns ucrtbase's table, so a ucrtbase that exports `_get_osfhandle` without `_set_thread_local_invalid_parameter_handler` is not used and `console_crt()` falls through to `static_crt()`, whose handler setter is the binary's own and always there. The comment above `shared_crt()` and the commit message say why. Folded into the console sink commit rather than added on top, per the one-commit-per-change rule of this PR.

alexowens90 (approved):

PR #3378, opened 2026-09-02
Bugfix 12852897792: fix [sort and] finalize staged data with sparse inputs and outputs

alexowens90:
#### Reference Issues/PRs
[12852897792](https://man312219.monday.com/boards/7852509418/pulses/12852897792)

### DIFF
--- a/cpp/arcticdb/column_store/column.cpp
+++ b/cpp/arcticdb/column_store/column.cpp
@@ -540,7 +540,7 @@ void Column::append(const Column& other, position_t at_row) {
 
     util::check(last_physical_row_ + 1 == row_count(), "Row count calculation incorrect after dense append");
 
-    if (at_row == initial_row_count && !other.is_sparse() && !is_sparse()) {
+    if (at_row == initial_row_count && !other.is_sparse() && !was_sparse) {
         util::check(
                 last_logical_row_ == last_physical_row_,
                 "Expected logical and physical rows to line up in append of non-sparse columns"
--- a/cpp/arcticdb/stream/merge.hpp
+++ b/cpp/arcticdb/stream/merge.hpp
@@ -117,22 +117,6 @@ void do_merge(QueueType& input_streams, AggregatorType& agg, bool add_symbol_col
                                 }
                             });
                         }
-                    } else {
-                        if constexpr (is_sequence_type(row_field_descriptor_tag.data_type())) {
-                            // When the value is std::nullopt this means we're dealing with sparse data. For string
-                            // values this means that the values in the string pool are placeholders either for NaN or
-                            // None. We write the placeholder into the column, otherwise if the whole column contains
-                            // None/NaN the encoding throws an exception see Monday tickets: 9712849046 and 9712629170
-                            using TDT = decltype(row_field_descriptor_tag);
-                            using RawType = typename TDT::DataTypeTag::raw_type;
-                            const RawType& raw_value = val->template value<RawType>();
-                            ARCTICDB_DEBUG_CHECK(
-                                    ErrorCode::E_ASSERTION_FAILURE,
-                                    nan_placeholder() == raw_value || not_a_string() == raw_value,
-                                    "Expected NaN or None placeholders to represent missing value."
-                            );
-                            rb.set_scalar_by_name(name, raw_value, row_field_descriptor_tag.data_type());
-                        }
                     }
                 });
             }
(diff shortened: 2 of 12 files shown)


cpp/arcticdb/column_store/column.cpp
```diff
@@ -540,7 +540,7 @@ void Column::append(const Column& other, position_t at_row) {
 
     util::check(last_physical_row_ + 1 == row_count(), "Row count calculation incorrect after dense append");
 
-    if (at_row == initial_row_count && !other.is_sparse() && !is_sparse()) {
+    if (at_row == initial_row_count && !other.is_sparse() && !was_sparse) {
```
alexowens90, line 543: This was just wrong before. `!is_sparse() => last_logical_row_ == last_physical_row_`, so the check inside this if would never fail

cpp/arcticdb/stream/merge.hpp
```diff
@@ -117,22 +117,6 @@ void do_merge(QueueType& input_streams, AggregatorType& agg, bool add_symbol_col
...
-                            using TDT = decltype(row_field_descriptor_tag);
-                            using RawType = typename TDT::DataTypeTag::raw_type;
-                            const RawType& raw_value = val->template value<RawType>();
-                            ARCTICDB_DEBUG_CHECK(
-                                    ErrorCode::E_ASSERTION_FAILURE,
-                                    nan_placeholder() == raw_value || not_a_string() == raw_value,
-                                    "Expected NaN or None placeholders to represent missing value."
-                            );
-                            rb.set_scalar_by_name(name, raw_value, row_field_descriptor_tag.data_type());
-                        }
```
alexowens90, line 135: This whole block was a fudge around writing zero-byte blocks, which is no longer needed.
In addition, the debug check failed (and the process could crash in general) with genuinely sparse data (not just string columns containing Nones and NaNs) as `val->template value<RawType>()` calls `Column::reference_at`. This will return the wrong value if the row index is <= the number of physical rows in the column, and ['splode](https://github.com/man-group/ArcticDB/blob/master/cpp/arcticdb/column_store/column.hpp#L543) if the row index is > number of physical rows.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -2751,8 +2565,7 @@ std::variant<VersionedItem, CompactionError> compact_incomplete_impl(
             index,
             dynamic_schema ? VariantSchema{DynamicSchema::default_schema(index, stream_id)}
                            : VariantSchema{FixedSchema::default_schema(index, stream_id)},
-            compaction_parameters.sparsify_ ? VariantColumnPolicy{SparseColumnPolicy{}}
-                                            : VariantColumnPolicy{DenseColumnPolicy{}}
+            VariantColumnPolicy{SparseColumnPolicy{}}
```
alexowens90, line 2568: `sparsify_` only ever appears to be true in the tick compactor, as it was not possible to write sparse data from batch APIs prior to Arrow

cpp/arcticdb/version/version_core.hpp
```diff
@@ -240,6 +240,30 @@ struct CompactionOptions {
     bool perform_schema_checks{true};
 };
 
+template<typename IndexType>
+folly::Future<entity::VariantKey> write_compacted_segment(
+        Store& store, const StreamId stream_id, VersionId version_id, SegmentInMemory&& segment,
+        std::shared_ptr<folly::NativeSemaphore> semaphore
+) {
+    auto local_index_start = IndexType::start_value_for_segment(segment);
+    auto local_index_end = pipelines::end_index_generator(IndexType::end_value_for_segment(segment));
```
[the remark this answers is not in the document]
IvoDD: So sad that we had such bugs up until now :/. 
I guess this is not a big deal since most of the codepaths already account for this potentially broken type of error due to e.g. update with date_range.

What do you think about adding a nonreg test for the end index value?

What do you think about also adding a compatibility test that e.g.:
```
old_arcticdb_version.sort_and_finalize(sym) # Writes several data keys all with incorrect end timestamp
current_arcticdb.read(sym, date_range=<>) # Checks that reading with date range works even with broken end index values.
```

Also have you scanned the code for any other places where `end_value_for_segment` is used incorrectly to populate the end value.

ognyanstoimenov (approved):

cpp/arcticdb/version/version_core.hpp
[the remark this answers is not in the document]
alexowens90: Agree with the nonreg tests.
That compatability test is functionally equivalent to `test_compat_update_old_updated_data`, as the bug was the same.
I've fixed up all the other locations that used raw `end_value_for_segment` to wrap them in `end_index_generator`, with the exception of `append_incomplete_segment`, as this is what the tick collector uses. I've opened a [ticket](https://man312219.monday.com/boards/7852509418/pulses/12989519294) for that to be fixed separately, as it might have knock-on effects on how the compactor behaves, and I don't want to hold up Arrow when it must already work in practice.

PR #3380, opened 2026-09-02
Allow deleting staged data by StageResult

markovskipetar:
Edit the two staged-data deletion APIs so their argument can be a **symbol string**, a
  **single `StageResult`**, or a **`List[StageResult]`**:

  - **V1:** `NativeVersionStore.remove_incomplete(symbol)`
  - **V2:** `Library.delete_staged_data(symbol)`

  | Argument | Behaviour |
  |----------|-----------|
  | `str` | Delete **all** `APPEND_DATA` (incomplete) keys for the symbol — *unchanged* |
  | `StageResult` | Delete **only** the `APPEND_DATA` keys named by that receipt |
  | `List[StageResult]` | Delete **only** the `APPEND_DATA` keys named by those receipts |

  Deleting by stage result(s) is **idempotent** — keys already gone (e.g. already finalized or
  already deleted) are silently ignored.

  ## What was changed

  - Extracted the existing flatten-and-remove logic out of the finalize-failure cleanup
    (`DeleteIncompleteKeysOnExit::~DeleteIncompleteKeysOnExit`) into a shared helper
    `delete_incomplete_keys_for_stage_results` in `incompletes.{hpp,cpp}`. It flattens all
    receipts' keys and issues a **single batched** `remove_keys(..., ignores_missing_key_ = true)`
    call. The destructor now calls this helper (behaviour-preserving).
  - Added `LocalVersionedEngine::remove_incompletes_for_stage_results` and a pybind binding
    (`PythonVersionStore` inherits the engine method, so no `version_store_api` change).
  - Type dispatch lives in **one** place — V1 `remove_incomplete`. `str` keeps the old C++ path;
    a lone `StageResult` is wrapped in a one-element list; both non-`str` forms hit the new
    binding. V2 `delete_staged_data` stays a thin forwarder to V1.
  - A single `StageResult` and a list both funnel to the same batched delete, so ~500 receipts is
    **one** storage round-trip, not 500.

  ## API change (backwards compatible)
  
  Both **V1 and V2 public APIs** widen their parameter from `str` to
  `Union[str, StageResult, List[StageResult]]`. The `str` form is unchanged, so existing callers
  are unaffected.

### DIFF
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -2325,7 +2325,6 @@ DeleteIncompleteKeysOnExit::~DeleteIncompleteKeysOnExit() {
         return;
 
     try {
-        storage::RemoveOpts opts{.ignores_missing_key_ = true};
         if (context_->incompletes_after_) {
             delete_incomplete_keys(*context_, *store_);
         } else {
@@ -2334,6 +2333,7 @@ DeleteIncompleteKeysOnExit::~DeleteIncompleteKeysOnExit() {
             if (stage_results_) {
                 delete_incomplete_keys_for_stage_results(store_, *stage_results_);
             } else {
+                storage::RemoveOpts opts{.ignores_missing_key_ = true};
                 auto keys_to_delete = read_incomplete_keys_for_symbol(store_, context_->stream_id_, via_iteration_);
                 store_->remove_keys(keys_to_delete, opts).get();
             }
--- a/cpp/arcticdb/version/version_core.cpp
+++ b/cpp/arcticdb/version/version_core.cpp
@@ -2331,15 +2331,12 @@ DeleteIncompleteKeysOnExit::~DeleteIncompleteKeysOnExit() {
         } else {
             // If an exception is thrown before read_incompletes_to_pipeline the keys won't be placed inside the
             // context thus they must be read manually.
-            std::vector<VariantKey> keys_to_delete;
             if (stage_results_) {
-                auto keys_to_delete_view =
-                        *stage_results_ | std::views::transform(&StageResult::staged_segments) | std::views::join;
-                keys_to_delete = std::vector<VariantKey>(keys_to_delete_view.begin(), keys_to_delete_view.end());
+                delete_incomplete_keys_for_stage_results(store_, *stage_results_);
             } else {
-                keys_to_delete = read_incomplete_keys_for_symbol(store_, context_->stream_id_, via_iteration_);
+                auto keys_to_delete = read_incomplete_keys_for_symbol(store_, context_->stream_id_, via_iteration_);
+                store_->remove_keys(keys_to_delete, opts).get();
             }
-            store_->remove_keys(keys_to_delete, opts).get();
         }
     } catch (const std::exception& e) {
         // Don't emit exceptions from destructor
--- a/python/arcticdb/version_store/_store.py
+++ b/python/arcticdb/version_store/_store.py
@@ -2842,11 +2842,13 @@ class NativeVersionStore:
             If a `StageResult` or list of them, removes only the APPEND_DATA keys named by
             those stage result(s). Missing keys are ignored.
         """
-        if isinstance(symbol, str):
-            self.version_store.remove_incomplete(symbol)
+        if isinstance(symbol, StageResult):
+            self.version_store.remove_incompletes_for_stage_results([symbol])
+        elif isinstance(symbol, (list, tuple)):
+            self.version_store.remove_incompletes_for_stage_results(list(symbol))
         else:
-            stage_results = [symbol] if isinstance(symbol, StageResult) else list(symbol)
-            self.version_store.remove_incompletes_for_stage_results(stage_results)
+            # str or numeric StreamId
+            self.version_store.remove_incomplete(symbol)
 
     def compact_incomplete(
         self,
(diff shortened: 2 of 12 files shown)


python/arcticdb/version_store/library.py
```diff
@@ -1737,19 +1741,30 @@ def update_batch(
...
         write
             Documentation on the ``staged`` parameter explains the concept of staged data in more detail.
+        stage
+            Returns the ``StageResult`` receipts accepted by this method.
+
+        Examples
+        --------
+        >>> receipt_1 = lib.stage("symbol", df1)
+        >>> receipt_2 = lib.stage("symbol", df2)
+        >>> lib.delete_staged_data(receipt_1)  # delete only the first batch's staged keys
```
alexowens90: I'm not convinced by this "receipts" terminology, I would just go with `stage_result_1/stage_result_2`

python/arcticdb/version_store/_store.py
```diff
@@ -2818,17 +2830,25 @@ def list_symbols_with_incomplete_data(self) -> List[str]:
...
+            those stage result(s). Missing keys are ignored.
         """
-        self.version_store.remove_incomplete(symbol)
+        if isinstance(symbol, StageResult):
+            self.version_store.remove_incompletes_for_stage_results([symbol])
+        elif isinstance(symbol, (list, tuple)):
+            self.version_store.remove_incompletes_for_stage_results(list(symbol))
+        else:
+            # str or numeric StreamId
+            self.version_store.remove_incomplete(symbol)
```
alexowens90: This doesn't follow our usual pattern. We should change the C++ implementation of `remove_incomplete` to accept `std::variant<StreamId, std::vector<StageResult>>`, and branch on the type of argument received in `local_versioned_engine.cpp`

python/arcticdb/version_store/_store.py
```diff
@@ -2818,17 +2830,25 @@ def list_symbols_with_incomplete_data(self) -> List[str]:
...
-            Symbol name.
+        symbol : `str`, `StageResult`, or `List[StageResult]`
+            If a symbol name (`str`), removes all incomplete segments for that symbol.
+            If a `StageResult` or list of them, removes only the APPEND_DATA keys named by
+            those stage result(s). Missing keys are ignored.
         """
-        self.version_store.remove_incomplete(symbol)
+        if isinstance(symbol, StageResult):
+            self.version_store.remove_incompletes_for_stage_results([symbol])
+        elif isinstance(symbol, (list, tuple)):
```
alexowens90: Why the special handling for `tuple`?

python/arcticdb/version_store/_store.py
```diff
@@ -2818,17 +2830,25 @@ def list_symbols_with_incomplete_data(self) -> List[str]:
...
+            self.version_store.remove_incompletes_for_stage_results(list(symbol))
+        else:
```
alexowens90: With this kind of polymorphism in Python, it's better to explicitly check `isinstance(symbol, str)`, and have an `else` that throws for all other received types. Otherwise users get a very confusing pybind error about incompatible types

cpp/arcticdb/version/version_core.cpp
alexowens90: nit: This is now only used on line 2338, so can be moved there

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -72,6 +72,129 @@ def test_staging_doesnt_write_append_ref(lmdb_version_store_v1):
...
+    stage_result_2 = lib.stage(sym, df2)
+    stage_result_3 = lib.stage(sym, df3)
+
+    lib.remove_incomplete([stage_result_1, stage_result_2])
+
+    remaining = {str(k) for k in get_append_keys(lib, sym)}
+    assert remaining == {str(k) for k in stage_result_3.staged_segments}
+
+
+def test_delete_staged_data_by_string_removes_all(lmdb_version_store_v1):
```
alexowens90: This is probably already tested somewhere else?

python/tests/unit/arcticdb/version_store/test_parallel.py
```diff
@@ -72,6 +72,129 @@ def test_staging_doesnt_write_append_ref(lmdb_version_store_v1):
...
+    keep_a = lib.stage(sym_a, df)
+
+    lib.remove_incomplete([stage_result_a, stage_result_b])
+
+    remaining_a = {str(k) for k in get_append_keys(lib, sym_a)}
+    assert remaining_a == {str(k) for k in keep_a.staged_segments}
+    assert len(get_append_keys(lib, sym_b)) == 0
+
+
+def test_delete_staged_data_v2_by_stage_result(lmdb_library):
```
alexowens90: Move to `test_arctic.py`

alexowens90 (changes requested):

python/arcticdb/version_store/_store.py
```diff
@@ -2842,13 +2842,15 @@ def remove_incomplete(self, symbol: Union[str, StageResult, List[StageResult]])
...
-            # str or numeric StreamId
+        if isinstance(symbol, (str, int, list)):
             self.version_store.remove_incomplete(symbol)
+        elif isinstance(symbol, StageResult):
+            self.version_store.remove_incomplete([symbol])
+        else:
+            raise UserInputException(
+                "remove_incomplete expected a symbol (str), a StageResult, or a list of StageResults, "
+                f"but got {type(symbol).__name__}"
+            )
```
alexowens90: I've changed my mind about [this](https://github.com/man-group/ArcticDB/pull/3380#discussion_r3923246650). We don't generally perform argument type checking in the Python layer. So the cleanest approach (that also handles Claude's comment above would be:
```
if isinstance(symbol, StageResult):
    symbol = [symbol]
self.version_store.remove_incomplete(symbol)
```

alexowens90 (approved):

PR #3382, opened 2026-09-02
Fix the data race on the slice_added guard that aborts read workers

### LINKED ISSUE #3381: Data race on the slice_added guard aborts read workers (part of #3233)
## Summary

`test_resample_dynamic_schema` crashes its xdist worker because two threads add the same entity to the
`ComponentManager` registry, tripping a live EnTT assertion. The guard meant to prevent that is a
`std::vector<bool>` written under per-element mutexes, which is a data race: `std::vector<bool>` is bit-packed, so
elements sharing a word are read-modify-written concurrently under *different* locks.

This is the crash mode behind part of #3233. Note that #3233 is an umbrella for at least three distinct crashes of
the same test id (see "Not all of #3233" below).

## Evidence

`PYTHONFAULTHANDLER=1` captured a full C stack in run 33626292855, job 100267024755 (`3.14 macOS`), reading
inside-out:

```
abort   <- __assert_rtn
entt::basic_sparse_set<entity>::try_emplace(...)
entt::basic_storage<shared_ptr<arcticdb::SegmentInMemory>,entity>::emplace_element
entt::basic_sigh_mixin<...>::emplace
arcticdb::ComponentManager::add_entity<shared_ptr<SegmentInMemory>, shared_ptr<RowRange>, ...>
arcticdb::version_store::add_slice_to_component_manager
  ... folly continuation from arcticdb::version_store::schedule_first_iteration
folly::CPUThreadPoolExecutor::threadRun
```

`cpp/third_party/entt/src/entt/entity/sparse_set.hpp:353` is `ENTT_ASSERT(elem == null, "Slot not available")`,
i.e. the entity already carries that component. `ENTT_ASSERT` is a plain `assert()` and ArcticDB never defines
`ENTT_DISABLE_ASSERT`.

## Root cause

`cpp/arcticdb/version/version_core.cpp:935`:

```cpp
auto slice_added_mtx = std::make_shared<std::vector<std::mutex>>(num_segments);
auto slice_added     = std::make_shared<std::vector<bool>>(num_segments, false);   // <-- bit-packed
...
std::lock_guard lock{slice_added_mtx->at(pos)};      // per-position lock
if (!(*slice_added)[pos]) {
    add_slice_to_component_manager(entity_id, segment_and_slice, component_manager, ...);
    (*slice_added)[pos] = true;                      // non-atomic RMW of a shared word
}
```

Positions `p` and `q` in the same 64-bit word are written under different mutexes. `(*slice_added)[p] = true` is a
read-modify-write of the whole word, so a concurrent write to `q` can write back a stale word and lose the `true`
for `p`. A later work unit sharing entity `p` then sees `false`, calls `add_slice_to_component_manager` a second
time on an id that already has its components, and EnTT aborts. Bits only ever go false -> true, so tearing can
only *lose* a `true`, which matches the observed double-add exactly.

Why this test above all others: `ResampleClause::structure_for_processing` ends in `structure_by_time_bucket`, one
of the few structurers that assigns the same segment to multiple processing units - precisely what the guard
exists for. The hypothesis strategy appends several generated frames and draws `rule` from 1s to 10000min over a
1960-1980 range, so segments routinely span many buckets and many units contend at once.

Introduced in #1950 ("Fully parallelise processing in read_batch"), 2024.

## Suggested fix

```cpp
auto slice_added = std::make_shared<std::vector<std::atomic_flag>>(num_segments);
// test_and_set(std::memory_order_acq_rel)
```

That removes the race and the whole `num_segments`-sized mutex vector with it.

## Related: assertions are live in shipped CI binaries

`cpp/CMakeLists.txt:249` sets `CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -ggdb"` with **no `-DNDEBUG`**, while
`_RELEASE` has it. CI wheels are RelWithDebInfo, so third-party `assert()`s are live in an otherwise optimised
build. That is why this surfaces as a worker abort rather than silent corruption - arguably a good thing, since it
caught a real race - but it is currently an accident of the flag string rather than a decision. Worth making
explicit either way.

## Second, related race

`ComponentManager::decrement_entity_fetch_count` (`cpp/arcticdb/processing/component_manager.cpp:22`) calls
`.reset()` on the `shared_ptr<SegmentInMemory>` component - a mutation - while `get_entities_impl` holds only a
`std::shared_lock`. Two units sharing an entity can have one thread reset the pointer while another copies it.
This is a plausible source of the segfault-flavoured variants of the same test failure.

## Not all of #3233

The three runs cited in #3233 are three different crashes:

| Date | Job | Param | Crash |
|---|---|---|---|
| 2026-08-04 | 3.14 macOS (92032483301) | PANDAS | SIGSEGV in `mdb_txn_begin` <- `LmdbStorage::do_fast_delete` <- `PythonVersionStore::clear` |
| 2026-08-18 | linux_aarch64 (95695985904) | PANDAS | `Fatal Python error: Aborted`, no C stack |
| 2026-09-02 | 3.14 macOS (100267024755) | PYARROW | this one |

So it is not PYARROW-specific (two of three are PANDAS), not macOS-specific (one is linux_aarch64) and not 3.14-
specific. Dynamic schema is incidental - it raises the segment count and therefore the contention.

Previous PRs #2710 and #2924 were tolerance/oracle fixes in `generic_resample_test` for *assertion* failures, a
different failure mode of the same test id; neither could have affected this crash.

G-D-Petrov:
Fixes #3381.

## The race

`schedule_first_iteration` must add each segment's components to the `ComponentManager` exactly once,
even though a segment shared by several processing units (resample bucket boundaries, for example) is
handed to continuations that run concurrently. The guard was a `std::vector<bool>` under a
`std::vector<std::mutex>`, one entry per position. `std::vector<bool>` is bit-packed, so positions in
the same word are read-modify-written under *different* mutexes: a write to `q` can write back a word
it read before `p` was set, losing the `true` for `p`. The next unit sharing `p` calls
`add_slice_to_component_manager` again for an id that already has its components, and EnTT's
`ENTT_ASSERT(elem == null, "Slot not available")` aborts the worker.

## The fix

`slice_added` becomes a `std::vector<uint8_t>`; the per-position `std::vector<std::mutex>` stays,
and the lock is still held across `add_slice_to_component_manager`. Net against master the guard is
`std::vector<bool>` → `std::vector<uint8_t>`, `= true` → `= 1`, and a comment spelling out why both
halves matter.

Both properties are load-bearing. `uint8_t` instead of `bool` removes the tearing. Holding the lock
*across* the add orders the loser after the winner: a unit that finds the flag set falls straight
through into `MemSegmentProcessingTask`, whose clauses call `gather_entities` on those same entity
ids with no other synchronisation in between. A claim-first `atomic_flag` would let it read components
the winner had not added yet — `ComponentManager`'s `shared_mutex` serialises those accesses but does
not order them.

`ComponentManager::add_components` now checks (debug builds only, `ARCTICDB_DEBUG_CHECK`) that the
entity does not already have the component, so a double add is a catchable ArcticDB error rather than
a third-party `assert()` that aborts the process. Release builds lose nothing: EnTT's own check is a
plain `assert()`, already compiled out under `NDEBUG`. Every existing caller either adds a component
the entity lacks or removes it first, so no live path changes.


cpp/arcticdb/version/version_core.cpp
```diff
@@ -1313,24 +1313,21 @@ std::shared_ptr<std::vector<folly::Future<std::vector<EntityId>>>> schedule_firs
...
                                           slice_added,
                                           clauses,
                                           entity_ids = std::move(entity_ids
                                           )](std::vector<pipelines::SegmentAndSlice>&& segment_and_slices) mutable {
                             for (auto&& [idx, segment_and_slice] : folly::enumerate(segment_and_slices)) {
                                 auto entity_id = entity_ids[idx];
                                 auto pos = id_to_pos->at(entity_id);
-                                std::lock_guard lock{slice_added_mtx->at(pos)};
-                                if (!(*slice_added)[pos]) {
+                                if (slice_added->claim(pos)) {
```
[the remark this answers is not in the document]
G-D-Petrov: Agreed at the time it was written, and this is what commit 5a15821 fixed: `ClaimFlags` is gone and `util::OnceFlags` holds the per-position `std::mutex` across the callable, exactly the "keep the mutex, only fix the storage" option. `std::vector<std::once_flag>` was the alternative; a plain mutex plus a `uint8_t` was kept because it is the same shape as the original guard and is directly testable. `OnceFlags.LaterCallersWaitForTheFirst` pins the publication ordering you describe and fails immediately against a claim-before-work implementation.

G-D-Petrov:
Thanks — going through the three open points from the review:

1. **Vacuous `ASSERT_EQ(calls.load(), num_positions)`** — correct, and now fixed in 4eaf2a3. The strided pass hands each position to exactly one thread, so the count held whether or not a flag was lost, and only the single-threaded pass afterwards caught it. Each thread now sweeps every position after its strided pass, so a lost flag is run a second time and fails the count directly. Verified standalone against both implementations (200 iterations each): with the extra sweep the pre-fix `std::vector<bool>` guard fails the count assertion 200/200, while it failed it 0/200 before; the fixed implementation passes both assertions 200/200.

2. **`OnceFlags.*` unverified in the build** — still true locally, for the reason in the description (a CMake reconfigure fails on `azure-storage-common-cpp:x64-linux` here). `test_once_flags.cpp` was compile-checked with `g++ -fsyntax-only` against the real headers and the real `gtest`, and run standalone, but the `CMakeLists.txt` addition is only exercised by CI. CI on this PR is currently still running; I will not merge until the C++ test jobs are green and `OnceFlags.*` shows up in them.

3. **Labels** — added `bug` and `patch`, which also clears the failing `check_labels` job. No `no-release-notes`.

The only other failing check is `Benchmark commit 45ce153b / start_ec2_runner`, which is EC2 runner provisioning and unrelated to the change.

cpp/arcticdb/processing/component_manager.hpp
```diff
@@ -67,6 +68,13 @@ class ComponentManager {
         std::unique_lock lock(mtx_);
         (
                 [&] {
+                    // Without this the double add trips a plain assert() deep inside EnTT, which aborts the
+                    // process rather than raising something a caller can catch. See #3381.
+                    internal::check<ErrorCode::E_ASSERTION_FAILURE>(
+                            !registry_.all_of<Args>(id),
```
alexowens90, line 77: This seems like it could be a performance bottleneck, I would include it in debug builds only

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1271,9 +1272,9 @@ std::shared_ptr<std::vector<folly::Future<std::vector<EntityId>>>> schedule_firs
     internal::check<ErrorCode::E_ASSERTION_FAILURE>(
             static_cast<bool>(admission), "schedule_first_iteration requires an admission handler"
     );
-    // Used to make sure each entity is only added into the component manager once
-    auto slice_added_mtx = std::make_shared<std::vector<std::mutex>>(num_segments);
-    auto slice_added = std::make_shared<std::vector<bool>>(num_segments, false);
+    // Used to make sure each entity is only added into the component manager once, and that a unit
+    // sharing an entity does not proceed to process it until it has been added
+    auto slice_added = std::make_shared<util::OnceFlags>(num_segments);
```
alexowens90: I agree with the solution, but not the abstraction. I can't imagine it being reused, and so the additional context needed to go check that `OnceFlags` behaves as you want increases rather than decreases the cognitive load. Prefer:
```suggestion
    auto slice_added = std::make_shared<std::vector<uint8_t>>(num_segments, 0);
```

alexowens90 (changes requested):

koriyoshi2041 (approved):
Validated the revised guard at `d0398773` against both concurrency invariants independently: a packed `vector<bool>` control lost neighbouring flags while the `uint8_t` form did not, and an early-publish control let a later caller observe missing state while holding the per-position mutex through the work prevented it. The inlined production guard keeps that ordering and the new unconditional tests cover the same two properties. I also checked the release/debug boundary on the duplicate-component assertion and diff hygiene. I did not run the full ArcticDB CMake matrix locally; hosted C++ jobs should remain the merge gate.

alexowens90 (approved):

koriyoshi2041 (changes requested):
The latest head removes the only unconditional regression coverage for the two properties this fix depends on. The remaining `AddSameComponentTwiceRaises` test is behind `DEBUG_BUILD`, while the hosted C++ matrix builds the release preset, so reverting `uint8_t` to the bit-packed `vector<bool>` form would still leave CI green.

Please keep a release-built regression that reaches the production guard, or an equivalent test whose failure is causally tied to the storage/ordering fix. My approval at `d0398773` no longer applies after `2cc31d2` removed that coverage.

PR #3392, opened 2026-09-04
Vectorize timestamp conversion in NativeVersionStore.list_versions

G-D-Petrov:
#### Reference Issues/PRs

None. Found while profiling the existing `python/benchmarks/list_versions.py` ASV benchmark.

#### What does this implement or fix?

`NativeVersionStore.list_versions` called `pandas.to_datetime` on a **scalar** epoch-nanosecond `int` once per returned version:

```python
"date": to_datetime(version_result[2], unit="ns", utc=True),
```

pandas' scalar branch (`pandas/core/tools/datetimes.py:1106`) is not a fast path. For every single call it allocates a throwaway one-element numpy array, constructs a full `DatetimeIndex`, tz-localizes it (a second `DatetimeArray._simple_new`), and then unboxes element `[0]`. That is create a lot of overhead and it scales lenearly with the number of versions.

This replaces the per-row scalar call with a single columnar conversion: gather the `creation_ts` values into an `int64` buffer, view it as `M8[ns]`, and build one `DatetimeIndex` with `tz="UTC"`, then zip it against the rows.

```python
result = self.version_store.list_versions(symbol, snapshot, latest_only, skip_snapshots)
dates = pd.DatetimeIndex(
    np.fromiter((version_result[2] for version_result in result), dtype=np.int64, count=len(result)).view("M8[ns]"),
    tz="UTC",
)
return [
    {
        "symbol": version_result[0],
        "version": version_result[1],
        "date": date,
        ...
    }
    for version_result, date in zip(result, dates)
]
```

**Micro-benchmark** of the returning comprehension in isolation (pandas 2.3.3), best-of-5:

| rows | before | after | speedup |
|-----:|-------:|------:|--------:|
| 1 | 11.9 µs | 6.2 µs | 1.9× |
| 10 | 118.0 µs | 8.4 µs | 14.0× |
| 100 | 1182 µs | 29 µs | 40.8× |
| 1,000 | 11.8 ms | 0.26 ms | 45.1× |
| 10,000 | 118.6 ms | 3.2 ms | 40.8× |

It is faster at **every** size, including a single row, so there is no length threshold or guard to tune.

and here are the results from the ASV benchmarks:

| Change   | Before [46c5ecdd]    | After [232d7b50]    |   Ratio | Benchmark (Parameter)                                                                                         |
|----------|----------------------|---------------------|---------|---------------------------------------------------------------------------------------------------------------|
[...]

PR #3394, opened 2026-09-05
Raise the Resample benchmarks' timeout so setup_cache fits hosted runners

G-D-Petrov:
#### Reference Issues/PRs

No linked issue. Evidence is a benchmarking run on this repo:
[run 33889446573](https://github.com/man-group/ArcticDB/actions/runs/33889446573),
job `bench windows_pilot 6.24.0` (2026-09-04).

`resample.Resample.setup_cache` builds the benchmark datasets (five column
types x 100 segments of 100k rows each) and runs under asv's default 60s
budget, which it now sits right on top of. On a hosted `windows-latest`
runner it goes over and asv kills it:

```
[11.11%] ··· Setting up resample:53                    failed
[11.11%] ···· asv: setup_cache failed (exit status -256)
[11.11%] ··· resample.Resample.time_resample skipped (setup_cache failed)
...
                    <setup_cache resample:53>          1.00m
                    <setup_cache resample:127>         9.99s
```

`resample:53` is `Resample.setup_cache`, killed at exactly the 60s deadline
(`-256` is asv's timeout exit code); `resample:127` is
`ResampleWide.setup_cache`, which finishes in ~10s. Losing the cache skips
every `Resample` time and peakmem benchmark for that job. This is variance,
not a regression: nine sibling jobs on the same hardware in the same run
came in just under the limit.

This sets `timeout = 600` on the `Resample` class, matching the value
`basic_functions`, `query_builder`, `column_stats`, `arrow` and
`modification_functions` already declare for their heavy suites.

#### Any other comments?

A note on why `timeout` is the right knob, since asv's handling of
`setup_cache` is not obvious:

- `asv_runner` only reads a `timeout` attribute off the `setup_cache`
  *function object* (`benchmarks/_base.py`:
  `setup_cache_timeout = _get_first_attr([self._setup_cache], "timeout", None)`).
- When that is absent, asv falls back to the benchmark's own timeout
  (`asv/runner.py`: `benchmark.get('setup_cache_timeout', benchmark['timeout'])`),
  which *is* resolved from the class. So a class-level `timeout` does govern
  `setup_cache` here.
- A class-level `setup_cache_timeout` attribute would be silently ignored —
  it never reaches the benchmark JSON.

Verified against asv 0.6.6 / asv_runner 0.3.1 by running the discovery and
budget-resolution code paths directly: no timeout -> 60s, class
`timeout = 600` -> 600s, class `setup_cache_timeout = 600` -> 60s.

`Resample` sets no `repeat`, so `asv_runner`'s timing loop takes the
`repeat == 0` branch with its own `max_time = 20.0` and never consults
`timeout`. Raising it therefore changes only the kill deadline, not how the
benchmarks are sampled. `ResampleWide` is left alone.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
 - [x] Have you updated the relevant docstrings, documentation and copyright notice? — n/a, benchmark configuration only; the change carries an explanatory comment.
 - [x] Is this contribution tested against [all ArcticDB's features](../docs/mkdocs/docs/technical/contributing.md)? — n/a, no library code is touched.
 - [x] Do all exceptions introduced raise appropriate [error messages](https://docs.arcticdb.io/error_messages/)? — n/a, no exceptions introduced.
 - [x] Are API changes highlighted in the PR description? — n/a, no API changes.
 - [x] Is the PR labelled as enhancement or bug so it appears in autogenerated release notes? — labelled `patch` / `no-release-notes`, matching prior benchmark-only PRs (#3217, #3215, #3016).
</details>

### DIFF
--- a/python/.asv/results/benchmarks.json
+++ b/python/.asv/results/benchmarks.json
@@ -5046,7 +5046,8 @@
                 "'count'"
             ]
         ],
-        "setup_cache_key": "resample:53",
+        "setup_cache_key": "resample:57",
+        "timeout": 600,
         "type": "peakmemory",
         "unit": "bytes",
         "version": "e0d4d4d70e841616329b33431eb4753443e3143e7d7c458daa965b15278c1015"
@@ -5092,7 +5093,8 @@
         "repeat": 0,
         "rounds": 2,
         "sample_time": 0.01,
-        "setup_cache_key": "resample:53",
+        "setup_cache_key": "resample:57",
+        "timeout": 600,
         "type": "time",
         "unit": "seconds",
         "version": "19cf418c76226bff4b3c6f56f5a8e84707551066a9b995c8f83f5876c869714b",
@@ -5103,7 +5105,7 @@
         "name": "resample.ResampleWide.peakmem_resample_wide",
         "param_names": [],
         "params": [],
-        "setup_cache_key": "resample:127",
+        "setup_cache_key": "resample:131",
         "type": "peakmemory",
         "unit": "bytes",
         "version": "93455e3ea01b28c847b53f4ed1ecbf6ea0faa06d1b08631f863399fe9a9ee8b3"
@@ -5118,7 +5120,7 @@
         "repeat": 0,
         "rounds": 2,
         "sample_time": 0.01,
-        "setup_cache_key": "resample:127",
+        "setup_cache_key": "resample:131",
         "type": "time",
         "unit": "seconds",
         "version": "b91bf134b20358ef57d42d66f4578d48c3908d31422ab5ecba4fc0dccf3c5427",
--- a/python/benchmarks/resample.py
+++ b/python/benchmarks/resample.py
@@ -23,6 +23,10 @@ from asv_runner.benchmarks.mark import skip_for_params
 
 class Resample:
     number = 5
+    # setup_cache takes right around asv's 60s default timeout on a shared
+    # windows-latest box, so it flakes on the slower draws; 600 matches the
+    # timeout the other heavy suites in this directory already declare.
+    timeout = 600
 
     LIB_NAME = "resample"
     CONNECTION_STRING = "lmdb://resample"


G-D-Petrov:
### The `benchmark_commit` failure here is gate noise, not a regression

This PR adds a `timeout = 600` class attribute to the `Resample` benchmarks (plus the regenerated `benchmarks.json` metadata that follows from it). It changes no library code and no benchmark body — the three rows the gate flagged are `compact_data` and `merge_functions`, which this PR does not touch at all.

I checked each flagged row against the published master benchmark history for the same machine (`ArcticDB-Medium-Runner`, py3.11). All three sit inside the scatter master already shows against itself.

| Benchmark (parameter) | Before | After | Ratio | Master range (same machine) | Master's own consecutive-commit spread | Verdict |
|---|---|---|---|---|---|---|
| `compact_data.CompactDataStringsStaticSchema.time_compact_data((100000, 100000, 10000), 10, False, 100000)` | 243 ms | 317 ms | 1.31 | 221–336 ms | median 0.97, p95 1.27, max 1.39; 25% of master-to-master pairs exceed 1.15 | noise |
| `compact_data.CompactDataStringsStaticSchema.time_compact_data((1000000, 10000, 100000), 10, False, 100000)` | 367 ms | 483 ms | 1.32 | 347–522 ms | median 0.97, p95 1.35, max 1.43; 22% of master-to-master pairs exceed 1.15 | noise |
| `merge_functions.MergeSplitSegmentsThinRowRange.time_merge((10000, 3), 'update_and_insert', 1, 1000000)` | 177 ms | 212 ms | 1.20 | 174–213 ms | median 1.00, max 1.19 over a short history | noise |

Every "after" value lands inside the range master has already recorded for that benchmark with no relevant code change. There were no ASV execution failures in this run — the job is red purely on these three rows. The same run also reports two *improvements* of 0.72x and 0.86x on benchmarks this PR likewise does not touch, which is the same scatter pointing the other way.

### The three rows move together, and master moves them too

These three measurements are strongly correlated across master history (pairwise r = 0.87–0.92): the runner alternates between a "fast" and a "slow" state from one master commit to the next, and all three rows swing together. That is a machine-state effect, not a code effect.

The clearest control is master commit `bdb3d7f4` ("Vectorize timestamp conversion in `NativeVersionStore.list_versions`", #3392). It touches only `_store.py` and one test — it cannot affect string compaction or thin-row-range merge. Measured against the *same* baseline this PR was measured against, it recorded:

| Benchmark row | Baseline `46c5ecdd` | master `bdb3d7f4` | ratio | this PR | ratio |
|---|---|---|---|---|---|
| compact_data `(100000, 100000, 10000), 10, False, 100000` | 243 ms | 311 ms | **1.28** | 317 ms | 1.31 |
| compact_data `(1000000, 10000, 100000), 10, False, 100000` | 367 ms | 489 ms | **1.33** | 483 ms | 1.32 |
| merge_functions `(10000, 3), 'update_and_insert', 1, 1000000` | 177 ms | 211 ms | **1.19** | 212 ms | 1.20 |

An unrelated master commit reproduced this PR's exact failure — same three rows, same ratios — against the same baseline. Master commit `84c4c759` (#3382) shows the same pattern (317 / 468 / 213 ms), and the commits either side of both sit back down at the low values. This is the runner oscillating, not code regressing.

### On the baseline

For this run the baseline is correct: `46c5ecdd` is the actual merge base of this PR (`gh api compare` reports `behind_by: 0`), so nothing is misattributed here. Worth noting for future readers, though, that the baseline numbers are not measured in this job — they are the stored results from the separate master benchmark run (recorded 2026-09-07 05:56, on a different runner instance from the one that measured the PR). The comparison is therefore always across two machine draws, which is where this scatter comes from. In this run the baseline happened to land at roughly the 20th percentile of that benchmark's own master distribution, so anything measured against it looks slow.

### Why this is expected rather than unlucky

The gate compares two single runs at `-f 1.15`. Applying that threshold to the observed master-to-master variation across the benchmark suite gives an expectation of about **8.2 false "worse" rows on a PR that changes nothing** (dropping to ~3.7 at 1.20 and ~0.2 at 2.0), i.e. a clean PR is essentially certain to go red. The three rows flagged here are, by historical trip rate, ranked #3, #7 and #13 out of ~1,470 benchmark/parameter combinations — the gate reddened on precisely the noisiest rows in the suite. For context, genuine regressions in this project's history have a median magnitude of 1.18x, which is below the noise floor at this factor and is why the threshold cannot separate the two.

I'm treating this check as non-blocking for this PR. Happy to re-run it if a second opinion is wanted, but a re-run is a fresh coin flip rather than a confirmation.

PR #3399, opened 2026-09-08
Enhancement 18362804801: Support for round-tripping pyarrow record batches, chunked arrays, and arrays, and polars series

alexowens90:
#### Reference Issues/PRs
[9360502457](https://man312219.monday.com/boards/7852509418/pulses/9360502457)

Adds support for writing `pyarrow.RecordBatch/ChunkedArray/Array` and `polars.Series`. Due to our on-disk slicing, `RecordBatch` always reads back as a `Table`, and `Array` always reads back as a `ChunkedArray`, even if they only have 1 chunk per column.

### DIFF
--- a/cpp/arcticdb/processing/schema_combine.cpp
+++ b/cpp/arcticdb/processing/schema_combine.cpp
@@ -707,10 +707,30 @@ NormalizationMetadata accumulate_arrow_and_arrow_norm(
             "Cannot {}: cannot combine indexed arrow data with unindexed arrow data",
             options.name()
     );
+    normalization::check<ErrorCode::E_INCOMPATIBLE_OBJECTS>(
+            accumulated.experimental_arrow().one_dimensional() == other.experimental_arrow().one_dimensional(),
+            "Cannot {}: cannot combine single-array arrow data with multi-array arrow data",
+            options.name()
+    );
+    auto res = accumulated;
+    // This correctly allows an unnamed Polars Series to be combined with [Chunked]Array as well as other unnamed Polars
+    // Series for append/update
+    if (res.experimental_arrow().polars_series_name() != other.experimental_arrow().polars_series_name() &&
+        options.name_mismatch == RequiredNameMismatchPolicy::RECONCILE_TO_UNNAMED) {
+        res.mutable_experimental_arrow()->clear_polars_series_name();
+    } else {
+        schema::check<ErrorCode::E_DESCRIPTOR_MISMATCH>(
+                res.experimental_arrow().polars_series_name() == other.experimental_arrow().polars_series_name(),
+                "Cannot {}: cannot combine single-array arrow data with mismatching names {}",
+                options.name(),
+                names_differ(
+                        res.experimental_arrow().polars_series_name(), other.experimental_arrow().polars_series_name()
+                )
+        );
+    }
     // Per-column metadata is merged rather than taking only the base schema's, so that a column only a later
     // schema has keeps what that schema says about it. Presence is checked per column rather than by comparing
     // map sizes, so that a metadata field a future client adds does not make existing data un-appendable.
-    auto res = accumulated;
     auto& res_columns = *res.mutable_experimental_arrow()->mutable_columns();
     util::for_each_key_union(
             accumulated.experimental_arrow().columns(),
--- a/cpp/proto/arcticc/pb2/descriptors.proto
+++ b/cpp/proto/arcticc/pb2/descriptors.proto
@@ -240,6 +240,11 @@ message NormalizationMetadata {
 
         bool has_index = 1;
         map<string, ColumnMeta> columns = 2;
+        // True for Polars series, and PyArrow chunked arrays and arrays. False for Polars dataframes, and PyArrow
+        // tables and record batches
+        bool one_dimensional = 3;
+        // Set only when data written is a Polars Series
+        optional string polars_series_name = 4;
     }
 
     oneof input_type {
--- a/python/arcticdb/util/arrow.py
+++ b/python/arcticdb/util/arrow.py
@@ -49,8 +49,8 @@ def convert_arrow_to_pandas_for_tests(table):
 
 
 def to_pyarrow_table(
-    arrow_structure: Union[pa.Table, pa.RecordBatch, pa.ChunkedArray, pa.Array, pl.DataFrame, pl.Series],
-) -> pa.Table:
+    arrow_structure: "Union[pa.Table, pa.RecordBatch, pa.ChunkedArray, pa.Array, pl.DataFrame, pl.Series]",
+) -> "pa.Table":
     if isinstance(arrow_structure, NORMALIZABLE_POLARS_TYPES):
         if not _PYARROW_AVAILABLE:
             raise ModuleNotFoundError(
--- a/python/arcticdb/util/arrow.py
+++ b/python/arcticdb/util/arrow.py
@@ -1,4 +1,10 @@
-from arcticdb.dependencies import pyarrow as pa
+from typing import Union
+
+from arcticdb.dependencies import _PYARROW_AVAILABLE, _POLARS_AVAILABLE, pyarrow as pa, polars as pl
+from arcticdb.exceptions import ArcticUnsupportedDataTypeException
+
+NORMALIZABLE_PYARROW_TYPES = (pa.Table, pa.RecordBatch, pa.ChunkedArray, pa.Array) if _PYARROW_AVAILABLE else tuple()
+NORMALIZABLE_POLARS_TYPES = (pl.DataFrame, pl.Series) if _POLARS_AVAILABLE else tuple()
 
 
 def cast_string_columns(table, string_type=None):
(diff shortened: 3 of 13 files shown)


python/arcticdb/version_store/_normalization.py
```diff
@@ -760,29 +761,46 @@ def apply_pyarrow_operations(self, table, op: ArrowNormalizationOperations):
...
+            if isinstance(arrow_structure, pl.Series):
+                norm_metadata.experimental_arrow.polars_series_name = arrow_structure.name
+                # For some reason to_frame() on a pl.DataFrame maintains the chunking and so is zero-copy, but on a
+                # pl.Series it copies into a pa.Array if the Series was chunked
+                arrow_structure = arrow_structure.to_frame().to_arrow().column(0)
+            else:  # pl.DataFrame
+                arrow_structure = arrow_structure.to_arrow()
+        if isinstance(arrow_structure, (pa.ChunkedArray, pa.Array)):
+            norm_metadata.experimental_arrow.one_dimensional = True
+            arrow_structure = pa.Table.from_arrays([arrow_structure], names=["__array__"])
```
alexowens90: The dummy name for this column will live forever, so should be considered carefully

cpp/proto/arcticc/pb2/descriptors.proto
```diff
@@ -240,6 +240,11 @@ message NormalizationMetadata {
 
         bool has_index = 1;
         map<string, ColumnMeta> columns = 2;
+        // True for Polars series, and PyArrow chunked arrays and arrays. False for Polars dataframes, and PyArrow
+        // tables and record batches
+        bool one_dimensional = 3;
+        // Set only when data written is a Polars Series
+        optional string polars_series_name = 4;
```
alexowens90, line 247: This may not be the final form we want, see [previous discussion](https://github.com/man-group/ArcticDB/pull/3224/changes#diff-0c560a9ebc8d0bb59b0262ef28aa5318cbbad4c953cfc7eab80cb55c09258b78), but will reduce merge conflicts with other in-flight work.

Worth noting that compared to the format proposed in the linked PR, this format requires minimal C++ knowledge of the written format, which is architecturally cleaner. The proposed format would need to introduce the need for an `arrow_common` style function comparable to `pandas_common`, but more complex as the Arrow protobuf messages have different members.

python/arcticdb/util/arrow.py
```diff
@@ -40,3 +46,27 @@ def convert_arrow_to_pandas_for_tests(table):
...
+                "ArcticDB's pyarrow optional dependency is missing and is required for working with polars DataFrames."
+            )
+        if isinstance(arrow_structure, pl.Series):
+            # For some reason to_arrow() on a pl.DataFrame maintains the chunking and so is zero-copy, but on a
+            # pl.Series it copies into a pa.Array if the Series was chunked
+            arrow_structure = arrow_structure.to_frame().to_arrow().column(0)
+        else:  # pl.DataFrame
+            arrow_structure = arrow_structure.to_arrow()
+    if isinstance(arrow_structure, (pa.ChunkedArray, pa.Array)):
+        arrow_structure = pa.Table.from_arrays([arrow_structure], names=["__array__"])
```
alexowens90, line 66: An alternative to this would be to store the Polars series name if provided, or an empty string (or some other placeholder) if an unnamed Polars series or pyarrow array type is provided here.
This implementation keeps `combine_schema` much simpler, as it avoids all of the required columns nonsense of Pandas by having all one dimensional structures have the same stream descriptors, and names to be a purely metadata concern.

IvoDD: I think storing the series name in the descriptor and treating it a required column would make arrow <> pandas interop easier.

E.g. if we're doing:
```
lib.write(pd.Series([1, 2, 3], name="pandas_name"))
lib.append(pl.Series([4, 5, 6], name="polars_name"))
```
This presumably must raise due to a mismatched series name (corresponding to the column name).

Probably it is fine to leave as is in this PR and I can change it later in the interop PR.

cpp/proto/arcticc/pb2/descriptors.proto
IvoDD: I guess the tradoff here:
- this format is easier to work with in C++, as you say
- this format allows for producing inconsitent state. E.g. `one_dimensional=True` alongside `len(columns)>1`.

Going with the proposed format seems sensible to me even if it allows for inconsistent state.

python/arcticdb/util/arrow.py
alexowens90: Worth noting that Pandas and Polars Series are only interoperable if the Pandas Series doesn't have a timeseries index

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -52,6 +56,299 @@ def assert_inputs_not_modified(*inputs):
...
+        or (first_type == "NamedSeries" and second_type != "NamedSeries")
+        or (first_type != "NamedSeries" and second_type == "NamedSeries")
+    ):
+        with pytest.raises(StreamDescriptorMismatch):
+            lib.append(sym, second_input)
+    else:
+        lib.append(sym, second_input)
+        received = lib.read(sym).data
+        expected = create_1d_arrow_structure("NamedSeries" if first_type == "NamedSeries" else "UnnamedSeries", data)
+        polars_assert_series_equal(received, expected)
```
IvoDD, line 265: Nit: Maybe also explicit assertion on the result type?

cpp/arcticdb/processing/schema_combine.cpp
```diff
@@ -707,10 +707,30 @@ NormalizationMetadata accumulate_arrow_and_arrow_norm(
...
+            accumulated.experimental_arrow().one_dimensional() == other.experimental_arrow().one_dimensional(),
+            "Cannot {}: cannot combine single-array arrow data with multi-array arrow data",
+            options.name()
+    );
+    auto res = accumulated;
+    // This correctly allows an unnamed Polars Series to be combined with [Chunked]Array as well as other unnamed Polars
+    // Series for append/update
+    if (res.experimental_arrow().polars_series_name() != other.experimental_arrow().polars_series_name() &&
+        options.name_mismatch == RequiredNameMismatchPolicy::RECONCILE_TO_UNNAMED) {
+        res.mutable_experimental_arrow()->clear_polars_series_name();
```
IvoDD, line 720: This is good for this PR.

Note that in follow up PR I'll likely reroute this through the `RequiredNameMismatch` struct for easier interop with pandas.

python/arcticdb/version_store/_store.py
```diff
@@ -3885,6 +3892,10 @@ def _process_info(
                 columns = pd.RangeIndex(0, len(columns))
         elif input_type == "experimental_arrow":
             arrow_meta = timeseries_descriptor.normalization.experimental_arrow
+            if arrow_meta.one_dimensional:
+                # This correctly gives an empty string for Array/ChunkedArray written data as well as Series with empty
+                # names
+                columns[0] = arrow_meta.polars_series_name
```
IvoDD, line 3898: It feels like this belongs in `_normalization.py`

python/arcticdb/util/arrow.py
```diff
@@ -40,3 +46,27 @@ def convert_arrow_to_pandas_for_tests(table):
             new_col = new_table.column(i).fill_null(False)
             new_table = new_table.set_column(i, name, new_col)
     return new_table.to_pandas()
+
+
+def to_pyarrow_table(
+    arrow_structure: Union[pa.Table, pa.RecordBatch, pa.ChunkedArray, pa.Array, pl.DataFrame, pl.Series],
+) -> pa.Table:
```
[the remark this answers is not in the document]
IvoDD: I think claude is right. For the type signatures we should use strings throughout.

python/arcticdb/version_store/_normalization.py
```diff
@@ -760,29 +761,30 @@ def apply_pyarrow_operations(self, table, op: ArrowNormalizationOperations):
...
-                    "ArcticDB's pyarrow optional dependency is missing and is required for working with polars DataFrames."
-                )
-            table = table.to_arrow()
-        if table.num_rows == 0:
+    def normalize(self, arrow_structure, **kwargs):
+        norm_metadata = NormalizationMetadata()
+        norm_metadata.experimental_arrow.one_dimensional = isinstance(
+            arrow_structure, (pa.ChunkedArray, pa.Array, pl.Series)
+        )
+        if isinstance(arrow_structure, pl.Series):
```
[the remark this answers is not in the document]
IvoDD: +1 to claude. This here could blow up in case we are writing a pa.Table with only `pyarrow` installed without `polars`

python/arcticdb/version_store/_normalization.py
```diff
@@ -1904,18 +1913,14 @@ def _strip_tz(s, e):
...
-                )
-            # PyArrow binary search + slice is benchmarked faster than polars native filtering
-            # (filter/is_between with set_sorted), which materializes a boolean mask over the
-            # entire column. The result is a zero-copy pa.Table view which the caller's
-            # normalizer accepts directly, avoiding a needless round-trip back to polars.
-            data = data.to_arrow()
+    elif isinstance(data, NORMALIZABLE_PYARROW_TYPES + NORMALIZABLE_POLARS_TYPES):
         check(index_column, "Cannot update with pyarrow Table without specifying index_column=True")
+        original_type = type(data)
+        original_name = data.name if original_type == pl.Series else None
```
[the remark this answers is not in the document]
IvoDD: We should probably add ourselves a ticket to add specific CI with various combinations of `pyarrow` and `polars` installed before the full release.

IvoDD (approved):

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
alexowens90: This is a Polars function, it will raise if the inputs aren't both Series already

python/arcticdb/version_store/_store.py
alexowens90: I kind of agree (although the `input_type == "df"` branch is much more egregious).
However, this stuff is so specific to the weird output format of `get_info` that keeping it localized to this method seems less polluting.

PR #3403, opened 2026-09-09
Create only one AwsEventLoop thread rather than num cores

poodlewars:
Create only one AwsEventLoop thread. These are only used for data transfer by the CRT client, whereas we use the cUrl client, so they are just wasting threads on the host machine. This is especially true when running on cgroup'd platforms like k8s where the host machine may have hundreds of cores - the AWS SDK's calculation is not cgroup aware - it uses the number of cores on the machine divided by two.

We do use these threads for auth interactions - looking up creds from a profile, or STS credentials refresh. A single thread should be plenty for these.

Testing on PURE and VAST:

## Method

| | |
|---|---|
| Dataset | 100 symbols x 100,000 rows x 5 float64 columns, 400 MB per batch read |
| Read with | `read_batch` of 100 `ReadRequest`s |
| Concurrency | ArcticDB defaults: ~192 IO threads and `S3Storage.MaxConnections` 192 on a 128-thread host |
| Repetitions | 5 per backend per configuration, A/B interleaved within each repetition |
| Per process | two `read_batch` passes, the first on a cold connection pool and the second warm |

## Results

These show no difference on the read timings, and something like a 15% improvement in the time taken to create the `Arctic()` instance.

Median with [min-max] across the 5 repetitions. Times in seconds.

| backend / config | `AwsEventLoop` threads | `Arctic()` construct | read pass 0 (cold) | read pass 1 (warm) | RSS MB |
|---|---|---|---|---|---|
| PURE / pinned (1) | 1 | **0.213** [0.213-0.214] | 0.280 [0.248-0.352] | 0.149 [0.141-0.175] | 1677 |
| PURE / SDK default (0) | 64 | 0.244 [0.235-0.263] | 0.335 [0.238-0.383] | 0.185 [0.146-0.195] | 1679 |
| VAST / pinned (1) | 1 | **0.214** [0.214-0.214] | 0.319 [0.301-0.578] | 0.218 [0.185-0.493] | 1696 |
| VAST / SDK default (0) | 64 | 0.247 [0.236-0.269] | 0.326 [0.278-0.497] | 0.226 [0.189-0.365] | 1719 |

Throughput at the median: PURE 1429 MB/s cold and 2685 MB/s warm when pinned, against 1194 and
2162 at the SDK default. VAST 1254 and 1835 when pinned, against 1227 and 1770.

### DIFF
--- a/cpp/arcticdb/storage/s3/s3_api.cpp
+++ b/cpp/arcticdb/storage/s3/s3_api.cpp
@@ -12,7 +12,9 @@
 #include <arcticdb/util/configs_map.hpp>
 #include <arcticdb/log/log.hpp>
 #include <arcticdb/storage/s3/ec2_utils.hpp>
+#include <algorithm>
 #include <cstdarg>
+#include <limits>
 #include <vector>
 #ifndef WIN32
 #include <aws/core/http/standard/StandardHttpRequest.h>
@@ -23,6 +25,8 @@
 namespace arcticdb::storage::s3 {
 
 namespace {
+constexpr const char* EventLoopAllocationTag = "ArcticDBS3EventLoop";
+
 spdlog::level::level_enum to_spdlog_level(Aws::Utils::Logging::LogLevel log_level) {
     switch (log_level) {
     case Aws::Utils::Logging::LogLevel::Fatal:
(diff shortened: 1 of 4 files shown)


poodlewars:
Running real storage benchmarks as a sanity check https://github.com/man-group/ArcticDB/actions/runs/34453868997

cpp/arcticdb/storage/s3/s3_api.cpp
```diff
@@ -134,13 +138,36 @@ void ArcticCurlHttpClientFactory::CleanupStaticState() {
...
 #endif // WIN32
 
-S3ApiInstance::S3ApiInstance(Aws::Utils::Logging::LogLevel log_level, bool log_to_file) :
+uint16_t event_loop_thread_count_from_config() {
+    const auto configured = ConfigsMap::instance()->get_int("AWS.EventLoopThreads", 1);
+    return static_cast<uint16_t>(std::clamp<int64_t>(configured, 0, std::numeric_limits<uint16_t>::max()));
+}
+
+ClientBootstrapFactory make_client_bootstrap_factory(uint16_t event_loop_thread_count) {
+    // Mirrors Aws::InitAPI's own default construction, apart from the explicit thread count.
```
alexowens90, line 147: This seems fragile to SDK version updates

alexowens90 (approved):

PR #3406, opened 2026-09-10
Unify combining empty schemas for `append`, `update` and `concat`

IvoDD:
## Reference Issues/PRs
Monday ref: 12911136883

## What does this implement or fix?

Consistantly handles combining an empty dataframe with any other dataframe.

An empty pandas dataframe can have an incorrect pandas index. E.g.:
`pd.DataFrame({"col": []})` gives either `RangeIndex` or `DatetimeIndex` depending on pandas version.

In a similar way we can't trust column datatypes for empty dataframes. E.g.:
`pd.DataFrame({"col": []})` has column `col` with dtype `object` or `float64` depending on pandas version.

So the only reasonable approach (without breaking a lot of existing code by disallowing some combinations) is to completely ignore schemas coming from empty pandas dataframes.

This PR handles that case consistently by adding an `inferred_from_empty_frame` property on the `OutputSchema` and uses it in `combine_schema` to unify the handling of the empty frames, instead of doing the filtering higher up the call stack.

## Notes for reviewers

The change is split in 3 commits:
- The first adds additional testing on top of the base branch
- The second makes `combine_schema` work with empty dataframes and applies the change for `append` and `concat`
- The third makes the change for `update` which might be more controversial, hence separated into its own commit.
- The fourth reverts the data column behavior changes as they are not desirable and simplifies the implementation

Reviewing the commits separately should make the most sense. Due to the later added 4th commit it is recommended to review 2, 3, 4 together. (unless you already looked at 2 and 3).

### 1. A symbol with no rows no longer constrains the index of what is appended to it

  ```python
  lib.write("sym", pd.DataFrame({"col": []}))     # stored with a DatetimeIndex on pandas 2
  lib.append("sym", pd.DataFrame({"col": [1.0]})) # RangeIndex

  # before: NormalizationException, E_INCOMPATIBLE_INDEX
  # after:  succeeds; the symbol is RangeIndexed, as its user intended
  ```

This also all index metadata related to the empty frame including:
- is multiindex
- index names
- rowrange `step` and `start`

All of these will be ignored on the empty dataframe and always accept the non empty one.

### 2. An `update` with a date range and no rows deletes that range

  ```python
  lib.update("sym", empty_df, date_range=(start, end))
  # before: no-op, new version with the new metadata only
  # after:  the rows in [start, end] are deleted
  ```

  An update with **no** date range is unchanged: there is no range to replace, so it still writes a
  new version carrying the new metadata and nothing else.


cpp/arcticdb/version/version_core.cpp
```diff
@@ -93,6 +93,11 @@ folly::Future<entity::AtomKey> async_write_dataframe_impl(
 }
 
 void sorted_data_check_append(const InputFrame& frame, const TimeseriesDescriptor& existing_tsd) {
+    if (frame.empty()) {
+        // Nothing to be out of order, and the index it was normalized with is not the one it was given, so neither
+        // side's sortedness has anything to say about it.
+        return;
+    }
```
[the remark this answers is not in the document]
IvoDD: We should skip because the existing_tsd might have e.g. RangeIndex and we're appending an empty one which appears as `DateRange`.

In any case `validate_index=True` not doing checks on an empty input frame makes sense. In above example user should have validate_index on the write if they cared about the index sortedness instead of relyin on the append to do it.

cpp/arcticdb/processing/test/test_schema_combine.cpp
```diff
@@ -596,3 +613,170 @@ TEST(CombineSchema, ThreeSchemasKeepFirstSeenColumnOrder) {
...
+    for (const bool arrow_has_index : {true, false}) {
+        auto arrow = arrow_has_index ? timeseries_df("ts", {{"a", DataType::FLOAT64}})
+                                     : rowcount_df({{"a", DataType::FLOAT64}});
+        arrow.norm_metadata_.mutable_experimental_arrow()->set_has_index(arrow_has_index);
+
+        for (auto schemas : {std::vector<OutputSchema>{empty, arrow}, std::vector<OutputSchema>{arrow, empty}}) {
+            auto combined = combine(schemas, append_options(true));
+            const auto& index = combined.norm_metadata_.df().common().index();
+            ASSERT_EQ(index.is_physically_stored(), arrow_has_index);
+            ASSERT_EQ(index.name(), "");
```
alexowens90: Shouldn't this be "ts" if `arrow_has_index == true`?

cpp/arcticdb/processing/schema_combine.cpp
```diff
@@ -298,26 +308,32 @@ void add_required_fields(
...
                 options.name(),
-                required_for_schema,
+                required_to,
                 desc.field_count()
         );
-        for (size_t idx = 0; idx < required_for_schema; ++idx) {
+        // offset_to_combined is non zero only for schemas which are inferred_from_empty.
+        // In that case it will align the positon of the series column (if one exists). It is possible for the offset to
+        // be negative.
+        const auto offset_to_combined = static_cast<ssize_t>(required_fields) - static_cast<ssize_t>(required_to);
```
alexowens90: Strictly speaking, `ssize_t` is only required to be able to store `-1`, although in practice it's almost always a typedef for a signed int type

cpp/arcticdb/processing/schema_combine.cpp
```diff
@@ -1055,20 +1035,29 @@ SortedValue deduce_sorted(SortedValue existing_frame, SortedValue input_frame) {
...
+    if (all_inferred_from_empty_pandas) {
+        // If all schemas are empty pandas we just use the first one's schema.
+        schemas = schemas.subspan(0, 1);
+    } else if (std::ranges::any_of(schemas, &OutputSchema::is_inferred_from_empty_pandas)) {
+        retained = util::reserve_vector<OutputSchema>(schemas.size());
+        std::ranges::copy_if(schemas, std::back_inserter(retained), [](const OutputSchema& schema) {
+            return !schema.is_inferred_from_empty_pandas();
+        });
+        schemas = retained;
+    }
```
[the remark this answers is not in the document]
IvoDD: Concat before 7.0.0 didn't work at all with empty symbols so this is not a behavior change.

cpp/arcticdb/python/normalization_utils.cpp
```diff
@@ -143,4 +143,12 @@ void update_rowrange_norm_for_append(
     }
 }
 
+namespace entity {
+
+bool OutputSchema::is_inferred_from_empty_pandas() const {
+    return inferred_from_empty_frame_ && pandas_common(norm_metadata_) != nullptr;
+}
```
alexowens90: Weird TU?

python/tests/unit/arcticdb/version_store/test_empty_writes.py
```diff
@@ -199,77 +199,55 @@ def test_append_empty_series(lmdb_version_store_dynamic_schema, sym, dtype, seri
...
-@pytest.mark.skipif(
-    not IS_PANDAS_TWO, reason="Pandas 1 normalizes an empty float column to a string one, leaving no float to promote"
-)
-def test_append_empty_dataframe_promotes_column_type_dynamic_schema(in_memory_version_store_dynamic_schema, sym):
-    # The type of an empty column is as much a statement as the column itself, so it promotes the stored one.
-    lib = in_memory_version_store_dynamic_schema
+# float64 under pandas 2, object under pandas 1, and a user reading from a vendor cannot always choose - so neither
+# displaces the stored type, promotable or not.
+@pytest.mark.parametrize("dtype", ["float64", "datetime64[ns]"])
+# TODO: parametrize over dynamic_schema [True, False] and use in_memory version store factory
```
alexowens90: nit: Completed TODO

alexowens90 (approved):

PR #3408, opened 2026-09-10
Single-hash insert for StringPool

G-D-Petrov:
`StringPool::get` hashed every string twice on the miss path: once for the `find` that established it was absent, and again for the `insert` of the same string.

The motivation is measured, from a CI flame graph of compaction on high-cardinality string columns: 2.02s of self time in `do_find`'s bucket probing under `StringPool::get`. Nothing about this is specific to compaction — every write path interns its strings through this function.

`try_emplace` gives the lookup and the insertion a single hash and probe. The key it stores is the caller's `string_view`, which need not outlive the pool, so it is rebound to the block-backed copy once the block holds one; the two compare equal, so the bucket the key was placed in stays correct.

**Exception safety.** `try_emplace` puts the key in the map *before* `block_.insert` runs, a window the old find-then-insert did not have: if the block insert threw, the map would be left holding a key that points into the caller's buffer. The block insert is wrapped so the entry is erased before the exception propagates, restoring the invariant — stated in a comment above the function — that a key is only in the map once the block owns a copy of it. This is not covered by a test: there is no seam to inject an allocation failure at, and a throwing allocator for this one call site was not worth the machinery.

Part 1 of a stack. Part 2 of the stack adds a `StringPool::reserve` capacity API and calls it from the reslicer, where capacity is known a priori.

https://claude.ai/code/session_01CGB2G1euu3wVQczjRbrsJv

### DIFF
--- a/cpp/arcticdb/column_store/string_pool.cpp
+++ b/cpp/arcticdb/column_store/string_pool.cpp
@@ -104,32 +104,32 @@ void StringPool::set_allow_sparse(Sparsity) {
 
 size_t StringPool::num_blocks() const { return block_.num_blocks(); }
 
+// try_emplace hashes and probes once for both the lookup and the insertion, where find followed by
+// insert hashes the same string twice on the miss path. The key it stores is the caller's view, which
+// need not outlive this pool, so it is rebound to the block-backed copy. The two compare equal, so the
+// bucket the key was placed in remains the right one.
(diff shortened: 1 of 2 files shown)


G-D-Petrov:
Parking alongside #3409 — measured ~5% on the high-cardinality reslicer path in isolation, below CI's noise floor end-to-end. The single-hash + exception-safety change is worth reviving with the stack. Branch stays.

PR #3409, opened 2026-09-10
Remap tables instead of per-row re-interning in the segment reslicer

G-D-Petrov:
At 100k unique strings, per-row `StringPool::get` was 62% of the compaction body's CPU in a CI flame graph. `reslice_by_iteration` reached it once per distinct offset per column per output row-slice: its `offsets_map` memo was per column, and it was cleared whenever the output segment changed.

Give each *input segment* a dense offset-to-offset table instead, shared by every column of that segment, so a string used by several columns is interned into an output pool at most once. Offsets are byte offsets and no string occupies fewer than `sizeof(StringHead)` bytes, so dividing by that indexes a vector without collisions. The table is filled lazily and clears itself when the output segment changes. Two cases keep the existing per-row path untouched: fixed-width string columns, whose strings are transformed before interning so their input offsets do not identify what lands in the output pool, and the None/NaN sentinels, which are out of range of any pool and so never get an entry.

This also adds a `StringPool::reserve` capacity API alongside the remap, and sizes each output pool with it at the `SegmentReslicer` call site: an output pool holds no more strings than all of the input pools do, and no more than the output row-slice has string values. Without it a pool that ends up with 100k strings rehashes its way there through every intermediate power of two.

Measured on the isolated reslicer benchmarks, this branch against a `master` benchmark binary, interleaved A/B, median of 5 rounds each:

| `BM_column_reslicer_...` | master | this branch | |
|---|---|---|---|
| `combine_dense_strings/10/10000/1` | 1.69 ms | 0.89 ms | **−47%** |
| `combine_dense_strings/10/10000/100000` | 5.94 ms | 4.13 ms | **−30%** |
| `split_dense_strings/100000/10/1` | 256 µs | 184 µs | **−28%** |

The single-unique case gains most because the old memo still cost a hash and a probe per row even when it hit. End-to-end `compact_data` was −29% at 10 columns and −8% at 2 when this was prototyped on a laptop; peak RSS unchanged.

**Why lazy, rather than pre-populating the output pool from the input pools** — which is what the comment this change deletes suggested. A decoded string pool is a single buffer that still contains the writer's block-boundary zero padding, so a linear walk over `StringHead` records desyncs from the real string starts. Enumerating an input pool's strings is not safe as the format stands; only the offsets a column actually references are known good, and those are exactly the ones the lazy table fills.

`reslice_columns` takes the remaps as a required parameter rather than gaining an overload: the signature already makes the caller own the output string pools and share them across a segment's columns, and the remaps are the exact analogue, where an overload that manufactured a private remap per column would compile and silently forgo the sharing that is the point.

Correctness: C++ unit suite green (1805 passed), Python unit suite green (18644 passed), with new tests covering the shared remap across two columns of one input segment and None/NaN passing through it across an output-slice boundary.

Stacked on #3408.

https://claude.ai/code/session_01CGB2G1euu3wVQczjRbrsJv

### DIFF
--- a/cpp/arcticdb/column_store/column_reslicer.cpp
+++ b/cpp/arcticdb/column_store/column_reslicer.cpp
@@ -76,13 +76,15 @@ void ColumnReslicer::push_back(size_t row_count) {
     sparse_ = true;
 }
 
-std::vector<Column> ColumnReslicer::reslice_columns(std::vector<StringPool>& string_pools) {
+std::vector<Column> ColumnReslicer::reslice_columns(
+        std::vector<StringPool>& string_pools, std::vector<StringOffsetRemap>& remaps
+) {
     util::check(type_.has_value(), "ColumnReslicer::reslice_columns called without any calls to push_back");
-    auto res = [this, &string_pools]() {
+    auto res = [this, &string_pools, &remaps]() {
         if (!is_sequence_type(type_->data_type()) && numeric_types_all_same_) {
             return reslice_by_memcpy();
         } else {
-            return reslice_by_iteration(string_pools);
+            return reslice_by_iteration(string_pools, remaps);
         }
     }();
     for (auto&& [idx, col] : folly::enumerate(res)) {
@@ -162,7 +164,9 @@ std::vector<Column> ColumnReslicer::reslice_by_memcpy() {
     return output_columns;
 }
 
-std::vector<Column> ColumnReslicer::reslice_by_iteration(std::vector<StringPool>& string_pools) {
+std::vector<Column> ColumnReslicer::reslice_by_iteration(
+        std::vector<StringPool>& string_pools, std::vector<StringOffsetRemap>& remaps
+) {
     auto output_columns = initialise_output_columns();
     util::check(
             output_columns.size() == string_pools.size(),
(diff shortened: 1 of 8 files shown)


G-D-Petrov:
The benchmark gate's three `MergeThinString*` rows (1.16–1.19) are cross-instance variance, not this PR: merge never executes the reslicer (`SegmentReslicer` is constructed only in `clause_compact_data.cpp`), and the only `string_pool` difference against #3408's head is the `reserve` method — called only from the reslicer — plus a `constexpr`. The merge code in #3408's and this PR's benchmark runs is therefore byte-identical, yet the same rows read 0.98–1.09 there and 1.16–1.19 here against the same stored baseline; that spread between two runner instances is the machine, not the change.

On the target benchmarks, this run's `CompactDataStringsStaticSchema` wall times are neutral-to-mixed (0.92–1.14 across the 100k-unique combos): on a 16-vCPU runner the reslicer's CPU sits off the critical path, so the win here is CPU time and the measured −28–47% on the isolated `BM_column_reslicer` benchmarks, not these wall clocks.

G-D-Petrov:
Parking this for now — the CPU win is real (−28–47% on the isolated reslicer benchmarks, analysis in the comments) but wall time on the 16-vCPU benchmark runners is neutral, so we'll revisit when compaction CPU matters for a concrete deployment shape. Branch stays for revival.

PR #3414, opened 2026-09-15
Use `combine_schema` for `incompletes` and `merge`

IvoDD:
### Reference Issues/PRs
Monday ref: 12821228270

### What does this implement or fix?

Uses the new `combine_schema` method for:
- `merge`
- `compact_incomplete`, `finalize_staged_data`
- Reading tick collector incompletes alognside existing compacted data

This allows deleting old APIs like `merge_descriptors`, `index_match`, `columns_match`.

### Review notes
The change is split into 2 commits:
- First adds extra testing
- Second makes the implementation and showcases the behavior changes.

### Behavior changes
All behavior changes can be seen in test changes in the second commit

#### 1. Finalizing staged data now accepts the same schemas as `append` and throws the same errors

- Compaction now raises on incompatible types:
  - rowrange with a daterange index
  - series with a dataframe
  - incompatible indices
- Same exception types and messages as `append`

#### 2. Finalizing staged data drops schemas of empty staged segments

```python
res_1 = lib.stage("sym", pd.DataFrame({"col1": [1, 2]}))
res_2 = lib.stage("sym", pd.DataFrame({"col2": []}))
lib.finalize_staged_data([res_1, res_2])
# Previously: result had an all empty `col2`
# Now: The empty schema is ignored and result only has the non-empty `col1`. Matching other APIs
```

If all empty dataframes are staged only the first schema will be kept:
```python
res_1 = lib.stage("sym", pd.DataFrame({"col1": []}))
res_2 = lib.stage("sym", pd.DataFrame({"col2": []}))
lib.finalize_staged_data([res_1, res_2])
# Previously: result had both `col1` and `col2`
# Now: Only the first staged result schema is kept.
```

#### 3. `read(incomplete=True)` now does a schema check

Previously reading tick streaming incomplete data with incompatible symbol raised `InternalException` lower down the stack. Now any schema incompatibility is raised with an appropriate error message.

#### 3. `merge` now rejects schemas in similar way to `append`

- Doesn't accept:
  - mixing `series` with `dataframe`
  - mixing index timezone

### Bugfixes

Again each bugfix change is visible in tests.

- `finalize_staged_data` could segfault when appending to empty frame with `validate_index=True`
- All identical index staged segments now preserve the staged_results order the user gave
- `convert_int_to_float` is now correctly applied only to staged data schema
- `read_incompletes_to_pipeline` took its `staged_descriptor` from `incomplete_segments[0]` while taking the
  normalization metadata and index columns from the first non-empty segment.
- `sort_merge` was doing one extra APPEND_DATA IO read.
- `required_fields_info` now infers the index count for empty pandas from the index_descriptor.field_count()

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


cpp/arcticdb/version/version_core.hpp
```diff
@@ -343,38 +337,12 @@ template<
                 segment.descriptor().uncompressed_bytes()
         );
 
-        if (!index_names_match(segment.descriptor(), pipeline_context->on_disk_descriptor())) {
-            auto written_keys = folly::collect(write_futures).get();
-            remove_written_keys(store.get(), std::move(written_keys));
```
IvoDD, line 348: Schema checks for incompletes are now done during load time in `read_incompletes_to_pipeline`. 
So no need to `remove_written_keys` on schema failure.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -920,10 +920,14 @@ VersionedItem LocalVersionedEngine::write_segment(
...
             tsd.set_stream_descriptor(segment.descriptor());
             tsd.set_total_rows(segment.row_count());
+            // The metadata has to describe the index the segment actually carries, otherwise a later append
+            // disagrees with it about whether the index is stored as a column.
             arcticdb::proto::descriptors::NormalizationMetadata norm_meta;
-            norm_meta.mutable_df()->mutable_common()->mutable_index()->set_is_physically_stored(false);
-            norm_meta.mutable_df()->mutable_common()->mutable_index()->set_start(0);
-            norm_meta.mutable_df()->mutable_common()->mutable_index()->set_step(1);
+            if (segment.descriptor().index().type() == IndexDescriptorImpl::Type::TIMESTAMP) {
+                ensure_timeseries_norm_meta(norm_meta, stream_id);
```
IvoDD, line 927: This `write_segment` API was only used in C++ tests, `lib_tool` and in the `health_checker` so this doesn't change structure on disk in any meaningful way.

cpp/arcticdb/stream/protobuf_mappings.hpp
```diff
@@ -27,9 +27,19 @@ arcticdb::proto::descriptors::NormalizationMetadata make_timeseries_norm_meta(co
...
 
+// Attaches a UTC timezone to an index that has none. The values are untouched, only labelled: a naive 12:00 becomes
+// 12:00+00:00.
+//
+// TODO (monday 11198274752): the caller that decides this should be the library's `set_tz` option, but:
+// - Nothing reads `set_tz`. `ReadOptions::set_tz_` has no getter.
+// - The compaction paths pass their `sparsify` flag here instead, reusing the parameter slot by coincidence.
+// - The tick compactor always sets `sparsify`, so tick-streamed symbols are always relabelled UTC.
+// Untangling this changes what users read back, so it needs its own rollout.
+void label_index_utc_if_unlabelled(arcticdb::proto::descriptors::NormalizationMetadata& norm_meta);
```
IvoDD, line 41: This was previously doen as part of `ensure_timeseries_norm_meta`. 

I separated it out as a different method to showcase the bug of passing `sparsify` instead of `set_tz`. As the ticket implies fixing this will probably require a way to revert to old behavior if old tick collector readers are relying on the odd behavior.

python/tests/unit/arcticdb/version_store/test_sort_merge.py
```diff
@@ -371,30 +371,7 @@ def test_mixing_empty_and_non_empty_columns(self, lmdb_library_dynamic_schema, m
...
-                    "c": np.array([0], dtype="int64"),
-                    "d": np.array([None], dtype="object"),
-                },
-                index=[pd.Timestamp("1970-01-01")],
-            )
-        if _use_pyarrow_strings_in_pandas():
-            # The all-null object column "d" reads back as the arrow-backed str dtype.
-            expected["d"] = expected["d"].astype(pd.StringDtype(storage="pyarrow", na_value=np.nan))
-        assert_frame_equal(lib.read("sym").data, expected)
+        assert_frame_equal(lib.read("sym").data, df)
```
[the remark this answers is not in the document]
IvoDD: Forgot to list it as a behavior change, now listed.

python/tests/hypothesis/arcticdb/test_sort_merge.py
```diff
@@ -107,6 +107,19 @@ def assert_staged_columns_are_incompatible(lib, symbol, mode):
...
+
+
+def assert_only_empty_staged_schema_survives(lib, symbol, df_list):
+    """With nothing but empty frames staged, one of their schemas stands. Which one is unspecified: schemas are ordered
+    by index range, and empty frames have no range to order by."""
+    data = lib.read(symbol).data
+    assert len(data) == 0
+    assert any(list(data.columns) == list(segment.columns) for segment in df_list)
+
+
```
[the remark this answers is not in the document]
IvoDD: Adding an explicit test for that case and a skip for the hypothesis test. 

The schema of an all empty staged result combination doesn't matter materially. The thing that matters is that the empty result can be appended with any schema, which the new test covers.

cpp/arcticdb/version/version_core.cpp
```diff
@@ -1653,116 +1641,54 @@ static std::variant<bool, CompactionError> read_incompletes_to_pipeline(
...
+        arcticdb::proto::descriptors::NormalizationMetadata norm_meta;
+        if (pipeline_context->has_normalization()) {
+            norm_meta.CopyFrom(pipeline_context->normalization());
+        }
+        ensure_timeseries_norm_meta(norm_meta, pipeline_context->stream_id_);
+        if (flags.sparsify) {
+            // Reaching a timezone decision through the sparsify flag is a bug. Monday ref 11198274752.
+            label_index_utc_if_unlabelled(norm_meta);
+        }
+        existing.emplace(pipeline_context->on_disk_descriptor(), std::move(norm_meta), pipeline_context->rows_ == 0);
```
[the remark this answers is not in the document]
IvoDD: The utc relabling is done only on `sparsify=True` which happens only on the tick compactor path. Also the `label_index_utc_if_unlabeled` is a no-op in case a timezone is previously set.

Since all tick compactor libraries are only populated by the tick compactor this doesn't result in a behavior change.

In any case I'll fix up the behavior to use `set_tz` as a follow up PR on the major-7-0-0 branch.

cpp/arcticdb/version/local_versioned_engine.cpp
```diff
@@ -920,10 +920,14 @@ VersionedItem LocalVersionedEngine::write_segment(
...
+            } else {
+                ensure_rowcount_norm_meta(norm_meta, stream_id);
+            }
             tsd.set_normalization_metadata(std::move(norm_meta));
```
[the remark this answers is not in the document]
IvoDD: My other comment explains that `lib_tool` is only a private API so no user-facing behavior change is seen.

python/tests/unit/arcticdb/version_store/test_stage.py
```diff
@@ -713,3 +713,34 @@ def test_delete_staged_data_on_failure_with_tokens_out_of_order_append(
...
+
+
+def empty_int_frame(column):
+    return pd.DataFrame({column: np.array([], dtype=np.int64)}, index=pd.DatetimeIndex([]))
+
+
+@pytest.mark.parametrize("dynamic_schema", [True, False], ids=["dynamic_schema", "static_schema"])
+@pytest.mark.parametrize("follow_up", ["append", "finalize"])
+def test_finalize_only_empty_segments(lmdb_library_factory, arctic_api, dynamic_schema, follow_up):
+    """The schema of the first stage_result stands, and binds nothing afterwards.
```
alexowens90, line 725: I guess "first" is a little ambiguous here if no stage results are passed. In practice (on S3) the lexicographical ordering of the listed `APPEND_DATA` keys means it would be the first key written by creation timestamp, so the statement holds, albeit by accident.

python/tests/hypothesis/arcticdb/test_sort_merge.py
```diff
@@ -162,15 +167,19 @@ def test_sort_merge_static_schema_write(lmdb_library, df_list):
     if len(df_list) == 0:
         assert_cannot_finalize_without_staged_data(lib, sym, StagedDataFinalizeMethod.WRITE)
         return
-    if not all(df_list[0].dtypes.equals(segment.dtypes) for segment in df_list):
+    contributing = staged_segments_that_contribute(df_list)
+    if not contributing:
```
vasil-pashov: nit: Does it make sense to change generate_dataframes so that it returns staged_segments_that_contribute?

Also I'd rather use hypothesis' `assume(len(staged_segments_that_contribute))` rather than early exiting

python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
```diff
@@ -505,3 +505,51 @@ def test_update_num_reads(s3_store_factory, clear_query_stats, dynamic_schema, u
...
+    TODO (monday 12955314293): the first read already holds the body. Keeping it would mean holding every staged
+    segment in memory at once instead of streaming them past the aggregator.
+    """
+    lib = in_memory_version_store
+    sym = "sym"
+    num_staged = 4
+    if append:
+        lib.write(sym, pd.DataFrame({"a": [0]}, index=pd.DatetimeIndex([pd.Timestamp(0)])))
+    for i in range(1, num_staged + 1):
+        lib.write(sym, pd.DataFrame({"a": [i]}, index=pd.DatetimeIndex([pd.Timestamp(i)])), parallel=True)
```
vasil-pashov: Why does this need the `apped` parameter but sort and finalize below doesn't?

python/tests/integration/arcticdb/version_store/test_num_storage_operations.py
```diff
@@ -505,3 +505,51 @@ def test_update_num_reads(s3_store_factory, clear_query_stats, dynamic_schema, u
...
+
+    qs.enable()
+    lib.compact_incomplete(sym, append=append, convert_int_to_float=False)
+    stats = qs.get_query_stats()
+    qs.reset_stats()
+
+    assert query_stats_operation_count(stats, "Memory_GetObject", "APPEND_DATA") == 2 * num_staged, pformat(stats)
+    # The keys are only listed once, and deleted once each.
+    assert query_stats_operation_count(stats, "Memory_ListObjects", "APPEND_DATA") == 1, pformat(stats)
+    assert query_stats_operation_count(stats, "Memory_DeleteObject", "APPEND_DATA") == num_staged, pformat(stats)
```
vasil-pashov: Why are these checked only here and not below?

PR #3416, opened 2026-09-15
Throw if date_range and update data have mixed timezone awareness

vasil-pashov:
Both the index of the date and the date_range must either be timezone aware or both must be timezone native. Mixing them was a bug in pre v7.0.0 version that allowed unexpected deletion of data. Exception will be thrown if timezone awareness is mixed.

This handles polars and arrow normalizable types. Whose first column is a datetime column.

This check if the objects have a timezone. Which means that timezone naive and something having utc timezone will not work. We don't need the complexity of handling this edge case. This can be relaxed easier in future.

#### Checklist

<details>
  <summary>
   Checklist for code changes...
  </summary>
 
</details>


python/arcticdb/version_store/library.py
```diff
@@ -399,7 +399,7 @@ def __init__(
         symbol: str,
         data: NormalizableType,
         metadata: Any = None,
-        date_range: Optional[Tuple[Optional[Timestamp], Optional[Timestamp]]] = None,
+        date_range: Optional[Union[DateRangeInput, Tuple[Optional[Timestamp], Optional[Timestamp]]]] = None,
```
alexowens90: For the V2 API I prefer the more limited type hint we had before.
If we do change it though, then the docstring for this class also needs updating, and it should also be updated in the `update` method type hint and docstring as well, as well in the `man.core` wrapper.

python/arcticdb/version_store/library.py
```diff
@@ -415,8 +415,11 @@ def __init__(
         metadata : Any, default=None
             Optional metadata to persist along with the new symbol version.
         date_range : Optional[Tuple[Optional[Timestamp], Optional[Timestamp]]], default=None
-            Restricts the update to the specified range in the stored data. Leaving either bound as ``None`` leaves that
-            side of the range open-ended.
+            If a range is specified, the existing data within that range is cleared and overwritten by data. This allows
+            the user to update a subset of the original data. Note that date_range is end-inclusive, and if either the
+            start or end is None, the range becomes open-ended on that side. Both date_range and data must be either
+            timezone-aware or timezone-naive. Only data with date_range will be modified, even if data covers a wider
+            date range.
```
alexowens90: I find this phrasing confusing (although it was terrible before). We should make clear what happens if the `date_range` argument is narrower or wider than the date range of the input data.

python/arcticdb/version_store/_store.py
```diff
@@ -4589,3 +4618,30 @@ def _log_warning_on_writing_empty_dataframe(dataframe, symbol):
             empty_column_type,
             current_dtypes,
         )
+
+
+def is_dataframe_index_tz_aware(data):
+    if hasattr(data, "loc"):
+        index = data.index.get_level_values(0)
+        return isinstance(index, pd.DatetimeIndex) and index.tz is not None
+    elif isinstance(data, NORMALIZABLE_POLARS_TYPES):
```
alexowens90, line 4632: For the Arrow types, this is only meaningful if `index_column == True`

alexowens90: It would have thrown later in `restrict_data_to_date_range_only`, but this could change the exception to something less meaningful

python/tests/unit/arcticdb/version_store/test_arrow_writes.py
```diff
@@ -1332,6 +1325,46 @@ def test_update_with_date_range_narrower_than_data(
     assert_frame_equal(expected, received)
 
 
+@pytest.mark.parametrize("start, end", DATE_RANGE_BOUNDS)
+@pytest.mark.parametrize("index_tz, date_range_tz", [("America/New_York", "Etc/UTC"), ("Etc/UTC", "America/New_York")])
+def test_update_with_date_range_in_different_timezone(
+    in_memory_version_store_arrow, start, end, index_tz, date_range_tz, arrow_output_format
+):
+    # Both timezone aware but in different zones, which update allows because the bounds are compared
+    # against the index as absolute instants, so the zone the caller writes them in does not matter.
```
alexowens90, line 1334: I agree with this behaviour, worth including in the docstrings though

python/tests/unit/arcticdb/version_store/test_update.py
```diff
@@ -336,23 +338,158 @@ def generate_dataframe(columns, dt, num_days, num_rows_per_day):
...
+    is the sole column, since these types carry no separate index/data distinction."""
+    if kind == "PandasSeries":
+        return pd.Series(data=range(len(index)), index=index, name="a")
+    if kind == "PandasDataFrame":
+        return pd.DataFrame({"a": range(len(index))}, index=index)
+    ts = pa.Array.from_pandas(index)
+    if kind == "Table":
+        return pa.table({"index": ts})
+    if kind == "RecordBatch":
+        return pa.RecordBatch.from_arrays([ts], names=["index"])
```
alexowens90: nit: `pa.record_batch({"index": ts})` works as well

alexowens90 (approved):

alexowens90 (approved):

PR #3418, opened 2026-09-16
WIP Enhancement 12844033169: arrow compatible column renaming

alexowens90:
#### Reference Issues/PRs
[12844033169](https://man312219.monday.com/boards/7852509418/pulses/12844033169)

#### Any other comments?


python/tests/unit/arcticdb/version_store/test_arrow_col_rename.py
```diff
@@ -0,0 +1,556 @@
...
+        assert not index.is_int
+        if index.is_physically_stored:
+            assert index.name == col_names[0]
+    else:  # multi_index
+        index = common.multi_index
+        assert not index.is_int
+        assert not len(index.fake_field_pos)
+        assert index.name == col_names[0]
+        for idx in range(1, index.field_count + 1):
+            assert col_names[idx].startswith("__idx__")
```
IvoDD, line 66: How is this arrow compatible? Wouldn't then the arrow table to be appended need to have this `__idx__` prefixes?

I think the `assert_norm_meta_arrow_compatible` should be done more along the way:
```
lib.write("sym", messy_pandas)
lib.rename_columns_for_arrow("sym")
arrow_table = lib.read("sym", output_format="pyarrow")
# Maybe update index so it can be appended
lib.append("sym", arrow_table)
```

I appreciate this will need my work on the arrow<>pandas interop for a lot of the metadata changes like e.g. timezones. But if no timezones or other arrow specific metadata is involved this should work at this state of the branch.

PR #3424, opened 2026-09-21
Stats for string columns

markovskipetar:
## How string stats are packed
  
  A stats cell is a fixed 8 bytes and the stats segment has no string pool, so we can't store the actual
  strings. Instead we pack each string into a `uint64`: **the first 7 UTF-8 bytes, then the byte length**.
  If the string is longer than 7 bytes, the length byte is `255` to mark it as truncated.

  ```
  "cat"        63 61 74 00 00 00 00 | 03
  "cab"        63 61 62 00 00 00 00 | 03
  "catalog"    63 61 74 61 6c 6f 67 | 07
  "elephant"   65 6c 65 70 68 61 6e | ff   (8 bytes, so the 't' is dropped)
  ```

  Text first, length last means comparing these as plain integers gives the same order as comparing the
  strings: the first differing byte decides it (`cat` > `cab` because `74` > `62`), and when one string is
  a prefix of the other the length byte breaks the tie (`cat` < `catalog`). So `min`/`max` over packed
  values is a real string range.


cpp/arcticdb/entity/types.hpp
```diff
@@ -63,6 +63,13 @@ constexpr size_t ASCII_WIDTH = 1;
 static_assert(UNICODE_WIDTH == 4, "Only support python platforms where unicode width is 4");
 #endif
 
+// The character width of a fixed-width UTF string as the string pool holds it. Fixed-width columns
+// are written from numpy `<U` arrays, whose characters are UCS-4 on every platform, and the pool keeps
+// those bytes verbatim. Code reading pool data must use this and not UNICODE_WIDTH, which follows
+// wchar_t and so is only two bytes on Windows.
+constexpr size_t UTF32_WIDTH = sizeof(char32_t);
+static_assert(UTF32_WIDTH == 4, "UCS-4 is four bytes per character on every platform");
```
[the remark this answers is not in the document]
alexowens90: Claude has a point here, but can be in a follow-up PR

cpp/arcticdb/pipeline/column_stats_filter.cpp
```diff
@@ -137,6 +138,13 @@ std::vector<StatsMetadataForColumn> calculate_stats_metadata(
                 tsd.fields().size()
         );
         stats_metadata_for_column.col_name = std::string{tsd.fields().at(data_col_offset).name()};
+        // Pruning on packed string stats is not implemented yet. Skipping only the MIN_STR/MAX_STR
+        // entries would leave the column min-less, which the comparator reads as "prune this slice".
```
alexowens90: I don't understand this comment. Isn't the condition for skipping that `tsd.field(data_col_offset).type()` is one of the string types?

cpp/arcticdb/util/string_stat_encoding.hpp
```diff
@@ -0,0 +1,61 @@
...
+// what bytewise string comparison does. Packing is therefore monotonic, so for a query q:
+//     pack(q) < stored_min  =>  q sorts below every value in the row slice
+//     pack(q) > stored_max  =>  q sorts above every value in the row slice
+// Either lets a reader prune the row slice without ever giving a wrong answer.
+constexpr size_t truncated_prefix_bytes = 7;
+
+// Length byte meaning "the source string was longer than truncated_prefix_bytes, so the prefix
+// stored here is incomplete". Deliberately the largest byte value so that a truncated stat sorts
+// above any exact-length value sharing its prefix.
+constexpr uint8_t truncated_string_length_marker = 255;
```
alexowens90: nit 
```suggestion
constexpr uint8_t truncated_string_length_marker = std::numeric_limits<uint8_t>::max();
```

cpp/arcticdb/util/string_stat_encoding.hpp
```diff
@@ -0,0 +1,61 @@
...
+// share a full-width prefix the length byte breaks the tie as shorter < longer < truncated, which is
+// what bytewise string comparison does. Packing is therefore monotonic, so for a query q:
+//     pack(q) < stored_min  =>  q sorts below every value in the row slice
+//     pack(q) > stored_max  =>  q sorts above every value in the row slice
+// Either lets a reader prune the row slice without ever giving a wrong answer.
+constexpr size_t truncated_prefix_bytes = 7;
+
+// Length byte meaning "the source string was longer than truncated_prefix_bytes, so the prefix
+// stored here is incomplete". Deliberately the largest byte value so that a truncated stat sorts
+// above any exact-length value sharing its prefix.
```
alexowens90: I don't think it matters for the implementation, but why not store the exact length if it is in the range `[8, 254]`?

cpp/arcticdb/util/string_stat_encoding.hpp
```diff
@@ -0,0 +1,61 @@
...
+constexpr uint8_t truncated_string_length_marker = 255;
+
+// Stencil for pulling the length byte out of a packed stat: ones in the low byte, zeros in the other
+// seven, so ANDing keeps that byte and erases everything above it.
+constexpr uint64_t rightmost_byte_only_mask = 0xFFULL;
```
alexowens90: `least_significant_byte_only_mask` is a bit more descriptive

cpp/arcticdb/util/string_stat_encoding.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace {
+constexpr size_t calculate_shift_bits(size_t i) { return (truncated_prefix_bytes - i) * 8; }
+
+// Fixed-width pools pad with null codepoints, so trailing nulls are dropped. Interior ones are data
+// and must survive, which is why this cannot use util::utf32_to_u8: that stops at the *first* null,
+// so "a\0b" would pack as "a", and since the engine's equality path does match an embedded null in a
+// fixed-width column, a query for the whole value would sort above the stored max and prune a row
+// slice that contains it.
+std::string utf32_to_utf8_stripping_padding(std::string_view raw_utf32_string) {
```
alexowens90: This duplicates `utf32_to_u8` in `string_utils.hpp`

cpp/arcticdb/util/string_stat_encoding.cpp
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    const auto last_non_zero_symbol_pos = utf32.find_last_not_of(char32_t{0});
+    utf32 = utf32.substr(0, last_non_zero_symbol_pos == std::u32string_view::npos ? 0 : last_non_zero_symbol_pos + 1);
+
+    return boost::locale::conv::utf_to_utf<char>(utf32.data(), utf32.data() + utf32.size());
+}
+
+// As above, only the padding goes: ASCII_FIXED64 pads a byte at a time, and an interior null is data.
+std::string_view strip_ascii_padding(std::string_view raw_ascii_string) {
```
alexowens90: This could also go in `string_utils.hpp`

cpp/arcticdb/util/string_stat_encoding.cpp
```diff
@@ -0,0 +1,95 @@
...
+} // namespace
+
+uint64_t pack_string_stat(std::string_view utf8_str) {
+    const auto bytes_kept = std::min(utf8_str.size(), truncated_prefix_bytes);
+    uint64_t packed{0};
+
+    for (size_t i = 0; i < bytes_kept; ++i) {
+        const uint64_t byte_in_packed = shift_to_byte_in_packed(utf8_str[i], i);
+        packed |= byte_in_packed;
+    }
```
alexowens90: I think this is correct, but possibly less efficient than:
```
uint64_t packed = *reinterpret_cast<const uint64_t*>(utf8_str.data());
```
and then masking it based on `bytes_kept`. Would be good to add a microbenchmark for `pack_string_stat` to confirm.

cpp/arcticdb/util/string_stat_encoding.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    if (raw_pool_string_data_type == entity::DataType::ASCII_FIXED64) {
+        return pack_string_stat(strip_ascii_padding(raw_pool_string));
+    }
+
+    return pack_string_stat(raw_pool_string);
+}
+
+UnpackedStringStat unpack_string(uint64_t packed) {
+    const auto length_byte = static_cast<size_t>(static_cast<uint8_t>(packed & rightmost_byte_only_mask));
```
alexowens90: `packed & rightmost_byte_only_mask` has type `uint64_t`, so these 2 static casts are unnecessary

cpp/arcticdb/util/string_stat_encoding.cpp
```diff
@@ -0,0 +1,95 @@
...
+    const bool was_truncated = (length_byte == truncated_string_length_marker);
+    const auto length = was_truncated ? truncated_prefix_bytes : std::min(length_byte, truncated_prefix_bytes);
```
alexowens90: The third operand to this ternary will always equal `length_byte` by construction

cpp/arcticdb/util/string_stat_encoding.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    std::string text(length, '\0');
+
+    for (size_t i = 0; i < length; ++i) {
+        const auto shift_right_bits = calculate_shift_bits(i);
+        const auto packed_shifted_right = (packed >> shift_right_bits);
+        text[i] = static_cast<char>(packed_shifted_right & rightmost_byte_only_mask);
+    }
```
alexowens90: Similarly to [this comment](https://github.com/man-group/ArcticDB/pull/3424/changes#r4093808112), I believe this implementation is correct, but would be more efficient to do something like:
```
uint64_t stripped = packed & !rightmost_byte_only_mask;
*text.data() = *reinterpret_cast<const char*>(&stripped);
```
Again, a microbenchmark for `unpack_string` to compare performance of different implementations would be good.

cpp/arcticdb/util/string_stat_encoding.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    if (raw_pool_string_data_type == entity::DataType::ASCII_FIXED64) {
+        return pack_string_stat(strip_ascii_padding(raw_pool_string));
+    }
+
+    return pack_string_stat(raw_pool_string);
+}
+
+UnpackedStringStat unpack_string(uint64_t packed) {
```
alexowens90: Am I right in saying this is only used for testing? If so, can move to `test_string_stat_encoding.cpp` with a comment that it is the inverse of `pack_string`.

cpp/arcticdb/util/string_stat_encoding.cpp
alexowens90: The microbenchmark is not needed if this function is only used for testing

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -77,21 +79,69 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
...
+                        offset_in_pool
+                );
+
+                return pack_string(*raw_pool_string, type_info::data_type);
+            };
+
+            ankerl::unordered_dense::set<RawType> seen_offsets_in_pool;
+
+            arcticdb::for_each<typename type_info::TDT>(*input_column.column_, [&](auto offset) {
+                const auto offset_in_pool = static_cast<entity::position_t>(offset);
```
alexowens90: Why do we need this `static_cast`?

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -77,21 +79,69 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
...
+
+                if (!is_a_string(offset_in_pool)) {
+                    if (offset_in_pool == nan_placeholder()) {
+                        ++nan_count_;
+                    } else {
+                        ++null_count_;
```
alexowens90: `None` values in string columns should contribute to the nan count, not the null count. Null count is strictly for missing values from Arrow sparse data or tick collectors.

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -77,21 +79,69 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
                 }
             }
         } else {
```
alexowens90: There seems to be a lot of duplication with the block above, can they be combined?
The if `(input_column.column_->is_sparse()) {...}` block is the same. The `is_nat_or_nan` lambda can be renamed to `adds_to_nan_count` and handle string types through `is_a_string`, and similar for the `missing_value` lambda

cpp/arcticdb/processing/unsorted_aggregation.cpp
```diff
@@ -77,21 +79,69 @@ void MinMaxAggregatorData::aggregate(const ColumnWithStrings& input_column) {
...
+        const auto min_type = packed_strings_ ? ColumnStatTypeInternal::MIN_STR_V1 : ColumnStatTypeInternal::MIN_V1;
+        const auto max_type = packed_strings_ ? ColumnStatTypeInternal::MAX_STR_V1 : ColumnStatTypeInternal::MAX_V1;
         res.reserve(4);
-        res.emplace_back(ColumnStatValue{ColumnStatTypeInternal::MIN_V1, data_col_offset_, *min_});
-        res.emplace_back(ColumnStatValue{ColumnStatTypeInternal::MAX_V1, data_col_offset_, *max_});
-    } else if (null_count_ == 0) {
-        // The column is absent from this slice entirely, so there is nothing to record
+        res.emplace_back(ColumnStatValue{min_type, data_col_offset_, *min_});
+        res.emplace_back(ColumnStatValue{max_type, data_col_offset_, *max_});
+    } else if (null_count_ == 0 && nan_count_ == 0) {
```
alexowens90: I think this check will go back to just `null_count_ == 0` if we treat `None` values in string columns as contributing to the `nan_count` rather than the `null_count`?

cpp/arcticdb/pipeline/test/test_column_stats_data.cpp
```diff
@@ -451,4 +452,98 @@ TEST(ColumnStatsDataTest, AllNullSliceKeepsCountsAndNotAbsent) {
...
+
+    ColumnStatsHeader header;
+    header.set_version(1);
+    auto& price_entries = (*header.mutable_stats_by_column())[price_data_col_offset];
+    price_entries.add_entries()->set_type(MIN_V1);
+    price_entries.mutable_entries(0)->set_stats_seg_offset(2);
+    price_entries.add_entries()->set_type(MAX_V1);
+    price_entries.mutable_entries(1)->set_stats_seg_offset(3);
+    auto& fruit_entries = (*header.mutable_stats_by_column())[fruit_data_col_offset];
+    for (const auto& [offset, type] : std::vector<std::pair<uint32_t, ColumnStatsType>>{
```
alexowens90: Would be clearer what is happening if the price stats were also added with this pattern

cpp/arcticdb/pipeline/test/test_column_stats_string_columns.cpp
```diff
@@ -0,0 +1,197 @@
...
+    util::check(idx.has_value(), "column {} missing", name);
+    return static_cast<position_t>(*idx);
+}
+} // namespace
+
+TEST(ColumnStatsStringColumns, EveryStringTypeIsEligibleForStats) {
+    for (const auto data_type : string_data_types) {
+        const auto desc = make_descriptor(data_type);
+        const auto map = ColumnStats{make_tsd(desc)}.to_map();
+        ASSERT_TRUE(map.contains("fruit")) << datatype_to_str(data_type);
```
alexowens90:
```suggestion
        EXPECT_TRUE(map.contains("fruit")) << datatype_to_str(data_type);
```

cpp/arcticdb/pipeline/test/test_column_stats_string_columns.cpp
```diff
@@ -0,0 +1,197 @@
...
+    const auto desc = make_descriptor(DataType::UTF_DYNAMIC32);
+    const auto map = ColumnStats{make_tsd(desc)}.to_map();
+    EXPECT_FALSE(map.contains("fruit"));
+    // The numeric index column is still eligible, so an empty map would prove nothing here.
+    EXPECT_TRUE(map.contains("index"));
+}
+
+// The reason the proto needed distinct MIN_STR_V1/MAX_STR_V1 values: the stat column's type is
+// resolved from the stat type alone. Inheriting the data column's string type would store the packed
+// value as a string pool offset into a segment that has no string pool.
```
alexowens90: It is not accurate to say that the segment has no string pool, it is just not used by the string column stats. This should also be corrected in the PR description.

cpp/arcticdb/pipeline/test/test_column_stats_string_columns.cpp
```diff
@@ -0,0 +1,197 @@
...
+        EXPECT_EQ(map.at("fruit"), std::unordered_set<std::string>{"MINMAX"}) << datatype_to_str(data_type);
+    }
+}
+
+// UTF_DYNAMIC32 only ever exists on the Arrow output path, so it never reaches stats generation and
+// admitting it would mean claiming stats for a column that cannot produce them.
+TEST(ColumnStatsStringColumns, ArrowOutputOnlyStringTypeIsNotEligible) {
+    const auto desc = make_descriptor(DataType::UTF_DYNAMIC32);
+    const auto map = ColumnStats{make_tsd(desc)}.to_map();
+    EXPECT_FALSE(map.contains("fruit"));
```
alexowens90: Generally prefer `ASSERT_` to `EXPECT_*`, except like in the test above where there is a loop over test cases

cpp/arcticdb/pipeline/test/test_column_stats_string_columns.cpp
```diff
@@ -0,0 +1,197 @@
...
+    EXPECT_EQ(to_segment_column_name("fruit", ColumnStatTypeInternal::MIN_STR_V1), "v1_MIN_STR(fruit)");
+    EXPECT_EQ(to_segment_column_name("fruit", ColumnStatTypeInternal::MAX_STR_V1), "v1_MAX_STR(fruit)");
+    // A reader must be able to tell a packed prefix from an ordinary numeric min by name alone.
+    EXPECT_EQ(to_segment_column_name("fruit", ColumnStatTypeInternal::MIN_V1), "v1_MIN(fruit)");
+    EXPECT_EQ(to_segment_column_name("fruit", ColumnStatTypeInternal::MAX_V1), "v1_MAX(fruit)");
+}
+
+// A string column can hold both packed min/max and the count stats, and the counts must keep their
+// own UINT64 type resolution rather than being confused with the packed ones.
+TEST(ColumnStatsStringColumns, CountStatsCoexistWithPackedStatsOnAStringColumn) {
```
alexowens90: This is not a very useful test, I would remove it

cpp/arcticdb/util/test/test_string_stat_encoding.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+#include <arcticdb/util/string_stat_encoding.hpp>
+#include <arcticdb/util/string_utils.hpp>
+
+#include <vector>
+
+namespace arcticdb {
+
+namespace {
+std::u32string as_utf32(std::string_view utf8) { return util::utf8_to_u32(utf8); }
```
alexowens90: Why not just call `utf8_to_u32` directly?

cpp/arcticdb/util/test/test_string_stat_encoding.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+TEST(StringStatEncoding, PrefixInHighBytesLengthInLowByte) {
+    ASSERT_EQ(pack_string_stat("a"), 0x6100000000000001ULL);
+    ASSERT_EQ(pack_string_stat("ab"), 0x6162000000000002ULL);
+    ASSERT_EQ(pack_string_stat("abc"), 0x6162630000000003ULL);
+    ASSERT_EQ(pack_string_stat("abcdef"), 0x6162636465660006ULL);
+    ASSERT_EQ(pack_string_stat("abcdefg"), 0x6162636465666707ULL);
+}
+
+TEST(StringStatEncoding, LengthByteCoversEveryExactLength) {
```
alexowens90: This is already covered by the test above

